Issue #1201, opened 2022-06-16
Improve synchronisation between tedge daemons


albinsuresh:
**Is your feature request related to a problem? Please describe.**

There are many interactions between different tedge daemons where one expects the other daemon to be up and running to respond to any requests that it sends. Here are some examples:

1. Once the c8y mapper starts, request software list from tedge-agent
2. Once the c8y-bridge is created, make c8y-config-plugin send the list of all supported config files
3. Once the c8y-bridge is created, make c8y-log-plugin send the list of all supported log files
4. Once the c8y-mapper starts, make tedge-agent send the status of the last "executing" operation, if any

Because of these dependencies, some of these messages could be lost if the requestee service is not up and running when the requester service sends a request.

Currently, we rely on MQTT broker's persistence-session feature to have these requests persisted, even if the requestee service is not up and running to receive those requests. But the broker keeps them persisted to deliver it to that service when it starts later. But, this persistence session feature is not very stable on the mosquitto broker that we use and hence we need a solution that doesn't fully rely on this feature.

**Describe the solution you'd like**

Make daemons request data from other daemons only once their liveness is validated. For example, c8y-mapper should request the the software list from tedge-agent only once  it can confirm that tedge-agent is up an running. The `tedge/health` endpoints of these daemons could be used to check this liveness.

**Describe alternatives you've considered**

Defining systemd service dependencies could be an alternative, but there are many cases where some service pairs have dependencies on each other, leading to cyclic dependencies. Even otherwise, it would have been a systemd specific solution.

didier-wenzek:
I want to highlight that the issue is beyond the stability issues we observed on mosquitto 2.0.

1. The agent can successfully receive a message published by the mapper even if it was down when the message was published. But the agent must have been launched *at least once* creating a subscription persisted in a named session. Messages sent just after the very first installation might be lost. This is why we introduced the `tedge_agent --init` option to create this persisted session on install ... with a major drawback : messages are consumed and discarded if the option is used after install.
2.  The issue with the bridge to Cumulocity is deeper. The bridge is created on `tedge connect c8y` and remove on `tedge disconnect c8y`.  After the first install and after a disconnect, there is *no* more bridge, i.e. the topics `c8y/#` are topics without any subscribers. Any message published on a `c8y/#` topic before a `tedge connect c8y` will be lost. Here the `--init` trick to create a session cannot work - because the subscriber is mosquitto itself.

I see the fix proposed here as the right approach.

[...]

PR #2065, opened 2023-07-10
Retry http requests to c8y fails with the response status code 401,403 and 404

PradeepKiruvale:
## Proposed changes

When an HTTP interaction with `Cumulocity` fails with 

- status code 401 or 403, then get the renewed `JWT` token and retry the HTTP request.
- status code 404 then get the new `internal-id` and retry the HTTP request.
- use map inside the `C8yEndPoint` to cache `child_internal_id` and `main_device_internal id`. Also, cache the `token` in `C8yEndPoint`

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/1201

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/http_proxy.rs
+++ b/crates/core/c8y_api/src/http_proxy.rs
@@ -6,35 +6,47 @@ use mqtt_channel::StreamExt;
 use mqtt_channel::Topic;
 use mqtt_channel::TopicFilter;
 use reqwest::Url;
+use std::collections::HashMap;
 use std::time::Duration;
 use tedge_config::mqtt_config::MqttConfigBuildError;
 use tedge_config::TEdgeConfig;
 use tracing::error;
 use tracing::info;
 
+#[derive(thiserror::Error, Debug)]
+pub enum C8yEndPointError {
+    #[error("Cumulocity internal id not found for the device: {0}")]
+    InternalIdNotFound(String),
+}
+
 /// Define a C8y endpoint
 #[derive(Debug)]
 pub struct C8yEndPoint {
-    pub c8y_host: String,
+    c8y_host: String,
     pub device_id: String,
-    pub c8y_internal_id: String,
+    pub token: Option<String>,
+    devices_internal_id: HashMap<String, String>,
 }
 
 impl C8yEndPoint {
-    pub fn new(c8y_host: &str, device_id: &str, c8y_internal_id: &str) -> C8yEndPoint {
+    pub fn new(c8y_host: &str, device_id: &str) -> C8yEndPoint {
         C8yEndPoint {
             c8y_host: c8y_host.into(),
             device_id: device_id.into(),
-            c8y_internal_id: c8y_internal_id.into(),
+            token: None,
+            devices_internal_id: HashMap::new(),
         }
     }
 
-    pub fn get_c8y_internal_id(&self) -> &str {
-        &self.c8y_internal_id
+    pub fn get_internal_id(&self, device_id: String) -> Result<String, C8yEndPointError> {
+        match self.devices_internal_id.get(&device_id) {
+            Some(internal_id) => Ok(internal_id.to_string()),
+            None => Err(C8yEndPointError::InternalIdNotFound(device_id.clone())),
+        }
     }
 
-    pub fn set_c8y_internal_id(&mut self, id: String) {
-        self.c8y_internal_id = id;
+    pub fn set_internal_id(&mut self, device_id: String, internal_id: String) {
+        self.devices_internal_id.insert(device_id, internal_id);
     }
 
     fn get_base_url(&self) -> String {
@@ -47,18 +59,17 @@ impl C8yEndPoint {
         url_get_id
     }
 
-    pub fn get_url_for_sw_list(&self) -> String {
+    pub fn get_url_for_sw_list(&self, internal_id: String) -> String {
         let mut url_update_swlist = self.get_base_url();
         url_update_swlist.push_str("/inventory/managedObjects/");
-        url_update_swlist.push_str(&self.c8y_internal_id);
-
+        url_update_swlist.push_str(&internal_id);
         url_update_swlist
     }
 
-    pub fn get_url_for_get_id(&self, device_id: Option<&str>) -> String {
+    pub fn get_url_for_internal_id(&self, device_id: String) -> String {
         let mut url_get_id = self.get_base_url();
         url_get_id.push_str("/identity/externalIds/c8y_Serial/");
-        url_get_id.push_str(device_id.unwrap_or(&self.device_id));
+        url_get_id.push_str(device_id.as_str());
 
         url_get_id
     }
@@ -178,13 +189,14 @@ pub enum JwtError {
 
 #[cfg(test)]
 mod tests {
+
     use super::*;
     use test_case::test_case;
 
     #[test]
     fn get_url_for_get_id_returns_correct_address() {
-        let c8y = C8yEndPoint::new("test_host", "test_device", "internal-id");
-        let res = c8y.get_url_for_get_id(None);
+        let c8y = C8yEndPoint::new("test_host", "test_device");
+        let res = c8y.get_url_for_internal_id("test_device".into());
 
         assert_eq!(
             res,
@@ -194,8 +206,11 @@ mod tests {
 
     #[test]
     fn get_url_for_sw_list_returns_correct_address() {
-        let c8y = C8yEndPoint::new("test_host", "test_device", "12345");
-        let res = c8y.get_url_for_sw_list();
+        let mut c8y = C8yEndPoint::new("test_host", "test_device");
+        c8y.devices_internal_id
+            .insert("test_device".to_string(), "12345".to_string());
+        let internal_id = c8y.get_internal_id("test_device".to_string()).unwrap();
+        let res = c8y.get_url_for_sw_list(internal_id);
 
         assert_eq!(res, "https://test_host/inventory/managedObjects/12345");
     }
@@ -210,7 +225,7 @@ mod tests {
     #[test_case("https://t1124124.test.com/path/to/file.test")]
     #[test_case("https://t1124124.test.com/path/to/file")]
     fn url_is_my_tenant_correct_urls(url: &str) {
-        let c8y = C8yEndPoint::new("test.test.com", "test_device", "internal-id");
+        let c8y = C8yEndPoint::new("test.test.com", "test_device");
         assert!(c8y.url_is_in_my_tenant_domain(url));
     }
 
@@ -220,7 +235,20 @@ mod tests {
     #[test_case("http://test.com:123456")]
     #[test_case("http://test.com::12345")]
     fn url_is_my_tenant_incorrect_urls(url: &str) {
-        let c8y = C8yEndPoint::new("test.test.com", "test_device", "internal-id");
+        let c8y = C8yEndPoint::new("test.test.com", "test_device");
         assert!(!c8y.url_is_in_my_tenant_domain(url));
+    }
+
+    #[test]
+    fn check_non_cached_internal_id_for_a_device() {
+        let mut c8y = C8yEndPoint::new("test_host", "test_device");
+        c8y.devices_internal_id
+            .insert("test_device".to_string(), "12345".to_string());
+        let end_pt_err = c8y.get_internal_id("test_child".into()).unwrap_err();
+
+        assert_eq!(
+            end_pt_err.to_string(),
+            "Cumulocity internal id not found for the device: test_child".to_string()
+        );
     }
 }
--- a/crates/core/c8y_api/src/json_c8y.rs
+++ b/crates/core/c8y_api/src/json_c8y.rs
@@ -17,7 +17,7 @@ use time::OffsetDateTime;
 
 const EMPTY_STRING: &str = "";
 
-#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
 #[serde(rename_all = "camelCase")]
 pub struct C8yCreateEvent {
     #[serde(skip_serializing_if = "Option::is_none")]
@@ -42,7 +42,7 @@ pub struct C8yEventResponse {
     pub id: String,
 }
 
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct C8yManagedObject {
     pub id: String,
@@ -98,7 +98,7 @@ impl From<SoftwareModule> for C8ySoftwareModuleItem {
     }
 }
 
-#[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Default)]
 #[serde(rename_all = "camelCase")]
 pub struct C8yUpdateSoftwareListResponse {
     #[serde(rename = "c8y_SoftwareList")]
--- a/crates/extensions/c8y_config_manager/src/upload.rs
+++ b/crates/extensions/c8y_config_manager/src/upload.rs
@@ -90,7 +90,7 @@ impl ConfigUploadManager {
                     self.upload_config_file(
                         Path::new(config_file_path.as_str()),
                         &config_upload_request.config_type,
-                        None,
+                        config_upload_request.device,
                         message_box,
                     )
                     .await
@@ -334,7 +334,7 @@ impl ConfigUploadManager {
             .upload_config_file(
                 Path::new(&uploaded_config_file_path),
                 &config_response.get_config_type(),
-                Some(config_response.get_child_id()),
+                config_response.get_child_id(),
                 message_box,
             )
             .await?;
@@ -357,12 +357,12 @@ impl ConfigUploadManager {
         &mut self,
         config_file_path: &Path,
         config_type: &str,
-        child_device_id: Option<String>,
+        device_id: String,
         message_box: &mut ConfigManagerMessageBox,
     ) -> Result<String, ConfigManagementError> {
         let url = message_box
             .c8y_http_proxy
-            .upload_config_file(config_file_path, config_type, child_device_id)
+            .upload_config_file(config_file_path, config_type, device_id)
             .await?;
         Ok(url)
     }
--- a/crates/extensions/c8y_http_proxy/src/actor.rs
+++ b/crates/extensions/c8y_http_proxy/src/actor.rs
@@ -7,6 +7,7 @@ use crate::messages::C8YRestRequest;
 use crate::messages::C8YRestResult;
 use crate::messages::DownloadFile;
 use crate::messages::EventId;
+use crate::messages::SoftwareListResponse;
 use crate::messages::Unit;
 use crate::messages::UploadConfigFile;
 use crate::messages::UploadLogBinary;
@@ -16,13 +17,13 @@ use c8y_api::http_proxy::C8yEndPoint;
 use c8y_api::json_c8y::C8yCreateEvent;
 use c8y_api::json_c8y::C8yEventResponse;
 use c8y_api::json_c8y::C8yManagedObject;
-use c8y_api::json_c8y::C8yUpdateSoftwareListResponse;
 use c8y_api::json_c8y::InternalIdResponse;
 use c8y_api::smartrest::error::SMCumulocityMapperError;
 use c8y_api::OffsetDateTime;
 use download::Auth;
 use download::DownloadInfo;
 use download::Downloader;
+use http::status::StatusCode;
 use log::debug;
 use log::error;
 use log::info;
@@ -45,7 +46,6 @@ const RETRY_TIMEOUT_SECS: u64 = 20;
 
 pub struct C8YHttpProxyActor {
     end_point: C8yEndPoint,
-    child_devices: HashMap<String, String>,
     peers: C8YHttpProxyMessageBox,
 }
 
@@ -80,16 +80,17 @@ impl Actor for C8YHttpProxyActor {
 
         while let Some((client_id, request)) = self.peers.clients.recv().await {
             let result = match request {
-                C8YRestRequest::GetJwtToken(_) => {
-                    self.get_jwt_token().await.map(|response| response.into())
-                }
+                C8YRestRequest::GetJwtToken(_) => self
+                    .get_and_set_jwt_token()
+                    .await
+                    .map(|response| response.into()),
 
                 C8YRestRequest::C8yCreateEvent(request) => self
                     .create_event(request)
                     .await
                     .map(|response| response.into()),
 
-                C8YRestRequest::C8yUpdateSoftwareListResponse(request) => self
+                C8YRestRequest::SoftwareListResponse(request) => self
                     .send_software_list_http(request)
                     .await
                     .map(|response| response.into()),
@@ -98,7 +99,6 @@ impl Actor for C8YHttpProxyActor {
                     .upload_log_binary(request)
                     .await
                     .map(|response| response.into()),
-
                 C8YRestRequest::UploadConfigFile(request) => self
                     .upload_config_file(request)
                     .await
@@ -117,20 +117,25 @@ impl Actor for C8YHttpProxyActor {
 
 impl C8YHttpProxyActor {
     pub fn new(config: C8YHttpConfig, message_box: C8YHttpProxyMessageBox) -> Self {
-        let unknown_internal_id = "";
-        let end_point = C8yEndPoint::new(&config.c8y_host, &config.device_id, unknown_internal_id);
-        let child_devices = HashMap::default();
+        let end_point = C8yEndPoint::new(&config.c8y_host, &config.device_id);
         C8YHttpProxyActor {
             end_point,
-            child_devices,
             peers: message_box,
         }
     }
 
     async fn init(&mut self) -> Result<(), C8YConnectionError> {
         info!(target: self.name(), "start initialisation");
-        while self.end_point.get_c8y_internal_id().is_empty() {
-            if let Err(error) = self.try_get_and_set_internal_id().await {
+
+        while self
+            .end_point
+            .get_internal_id(self.end_point.device_id.clone())
+            .is_err()
+        {
+            if let Err(error) = self
+                .get_and_set_internal_id(self.end_point.device_id.clone())
+                .await
+            {
                 error!(
                     "An error occurred while retrieving internal Id, operation will retry in {} seconds\n Error: {:?}",
                     RETRY_TIMEOUT_SECS, error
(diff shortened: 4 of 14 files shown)


crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -358,14 +455,44 @@ impl C8YHttpProxyActor {
...
-            .json(&c8y_event);
-        let http_result = self.execute(req_builder).await?;
+        let http_result = self.execute(None, req_builder_closure).await?;
         let http_response = http_result.error_for_status()?;
         let event_response: C8yEventResponse = http_response.json().await?;
         Ok(event_response.id)
     }
 }
+
+fn update_url_with_new_internal_id(internal_id: String, url: &str) -> Result<Url, C8YRestError> {
```
didier-wenzek, line 475: Why do we need this function that updates an url already built?

Why not directly call the primary function that build an url given an internal id?

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -186,11 +188,38 @@ impl C8YHttpProxyActor {
         &mut self,
         device_id: Option<&str>,
     ) -> Result<String, C8YRestError> {
-        let url_get_id = self.end_point.get_url_for_get_id(device_id);
+        let url_get_id: String = self.end_point.get_url_for_get_id(device_id);
+        let req_build_closure = |token: String| {
+            let request_internal_id = HttpRequestBuilder::get(&url_get_id);
+            request_internal_id.bearer_auth(token)
```
didier-wenzek, line 194: As adding a token is done for all and every request and as this can be done blindly, it's better to move this inside the `execute` method.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -199,19 +228,54 @@ impl C8YHttpProxyActor {
...
+                        self.end_point.token.clone().unwrap_or_default(),
+                        self.end_point.c8y_internal_id.clone(),
+                    );
+                    let request = request_builder?.build()?;
+                    Ok(self.peers.http.await_response(request).await?)
+                }
+                StatusCode::NOT_FOUND => {
+                    // get new internal id not the cached one
+                    self.end_point.c8y_internal_id = "".into();
+                    let internal_id = match child_device_id {
```
didier-wenzek, line 257: I would introduce a `get_fresh_internal_id()` or `renew_internal_id()` method.

crates/extensions/c8y_http_proxy/src/actor.rs
albinsuresh: As suggested in the previous PR, I still feel that those `C8yEndpoint` functions generating the URLs can be converted to accept the internal id as an `override_id: Option<String>` parameter so that the URLs can be created out of it. If the `override_id` is `None`, just use the once cached in that `C8yEndpoint` itself.

PradeepKiruvale: refactored, this is not required anymore.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -199,19 +228,54 @@ impl C8YHttpProxyActor {
...
-        // Get a JWT token to authenticate the device
-        let request_builder = if let Ok(token) = self.peers.jwt.await_response(()).await? {
-            request_builder.bearer_auth(token)
-        } else {
-            return Err(C8YRestError::CustomError("JWT token not available".into()));
-        };
+        let request_builder = req_builder_closure(
+            self.end_point.token.clone().unwrap_or_default(),
+            self.end_point.c8y_internal_id.clone(),
+        );
```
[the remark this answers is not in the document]
PradeepKiruvale: Refactored code to take `end_point` as input in closure. Caching of the token is moved to some other place.

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: refactored

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -186,11 +188,38 @@ impl C8YHttpProxyActor {
...
-        let request_internal_id = HttpRequestBuilder::get(url_get_id);
-        let res = self.execute(request_internal_id).await?;
-        let res = res.error_for_status()?;
+        let res = match self.peers.http.await_response(request).await? {
+            Ok(response) => match response.status() {
+                StatusCode::OK => Ok(Ok(response)),
+                StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN => {
+                    // get new token not the cached one
+                    self.end_point.token = None;
+                    self.get_jwt_token().await?;
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -358,14 +455,44 @@ impl C8YHttpProxyActor {
...
+    })?;
+
+    let new_url = url
+        .join(&format!("./{internal_id}"))
+        .map_err(|e| C8YRestError::CustomError(format!("failed to update the url due to: {e}")))?;
+
+    Ok(new_url)
+}
+
+fn update_event_with_new_internal_id(
```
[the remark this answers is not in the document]
PradeepKiruvale: Refactored and removed this code.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -186,56 +197,156 @@ impl C8YHttpProxyActor {
         &mut self,
         device_id: Option<&str>,
     ) -> Result<String, C8YRestError> {
-        let url_get_id = self.end_point.get_url_for_get_id(device_id);
+        let internal_id = if self.cached_identifiers.main_device_internal_id.is_none() {
```
didier-wenzek, line 200: There is a weird overlapping of responsibilities between the `try_get_and_set_internal_id` and `try_get_internal_id`. Both are manipulating the cache leading to useless and hard to follow sequences (as storing in the cache what has just been retrieved form the cache.

- The `try_get_internal_id` should not even be aware that a cache exist.
- The `try_get_and_set_internal_id` should call the former only when the id is not cached.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -186,56 +197,156 @@ impl C8YHttpProxyActor {
...
+                req_builder(self.cached_identifiers.token.clone().unwrap_or_default());
+
+            let request = request_builder.build()?;
+
+            let res = match self.peers.http.await_response(request).await? {
+                Ok(response) => match response.status() {
+                    StatusCode::OK => Ok(Ok(response)),
+                    StatusCode::UNAUTHORIZED | StatusCode::FORBIDDEN => {
+                        let request_builder =
+                            req_builder(self.cached_identifiers.token.clone().unwrap_or_default());
```
didier-wenzek, line 216: The very same token is used while the point is to retry with a fresh one.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -186,56 +197,156 @@ impl C8YHttpProxyActor {
...
     ) -> Result<String, C8YRestError> {
-        let url_get_id = self.end_point.get_url_for_get_id(device_id);
+        let internal_id = if self.cached_identifiers.main_device_internal_id.is_none() {
+            let url_get_id: String = self.end_point.get_url_for_get_id(device_id);
+            let req_builder =
+                |token: String| HttpRequestBuilder::get(&url_get_id).bearer_auth(token);
+
+            self.get_and_set_jwt_token().await?;
+            let request_builder =
+                req_builder(self.cached_identifiers.token.clone().unwrap_or_default());
```
didier-wenzek, line 207: As the closure is used along its definition, this seems to be useless.

didier-wenzek (commented):
The current design is too fragile.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -17,9 +18,9 @@ use tracing::info;
 pub struct C8yEndPoint {
     pub c8y_host: String,
     pub device_id: String,
+    // internal id that can be used to build url, this could be main/child device internal-id
     pub c8y_internal_id: String,
```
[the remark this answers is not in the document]
PradeepKiruvale: The `internal_id` here is for building the `http` request. This may be the `child or main` device internal ID based on the type of `http` request that is being built.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -164,20 +170,25 @@ impl C8YHttpProxyActor {
 
     async fn try_get_and_set_internal_id(&mut self) -> Result<(), C8YRestError> {
         let internal_id = self.try_get_internal_id(None).await?;
-        self.end_point.set_c8y_internal_id(internal_id);
+        self.end_point.set_c8y_internal_id(internal_id.clone());
+        self.cached_identifiers
+            .set_main_device_internal_id(internal_id);
```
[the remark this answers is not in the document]
PradeepKiruvale: Yes, one is for the `caching purpose` and other is for building the `http` request.

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: Thanks for pointing out., I did add `self.get_and_set_jwt_token().await?;` some how missed while refactoring,

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: Removed `self.get_and_set_jwt_token()`

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -104,6 +105,45 @@ impl C8yEndPoint {
     }
 }
 
+#[derive(Debug)]
+pub struct CachedIdentifiers {
```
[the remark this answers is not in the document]
PradeepKiruvale: Does `CachedDeviceIdentifiers` make sense?

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: Refactored

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -104,6 +105,45 @@ impl C8yEndPoint {
...
+    pub token: Option<String>,
+    pub child_devices_intenal_id: HashMap<String, String>,
+    pub main_device_internal_id: Option<String>,
```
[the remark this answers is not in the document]
PradeepKiruvale: There will be only one main device. Also, if we do that we need to pass the `device-id` always to get and set functions. So, I feel it's needed.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -17,24 +18,37 @@ use tracing::info;
...
     pub device_id: String,
-    pub c8y_internal_id: String,
+    pub token: Option<String>,
+    pub devices_internal_id: HashMap<String, String>,
 }
 
 impl C8yEndPoint {
     pub fn new(c8y_host: &str, device_id: &str, c8y_internal_id: &str) -> C8yEndPoint {
+        let mut devices_internal_id = HashMap::new();
+        if !c8y_internal_id.is_empty() {
```
didier-wenzek, line 28: I don't think there is a single place in the code where `C8yEndPoint::new()` is called with a meaningful internal id.

=> I would simply get rid of this 3rd argument.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -119,18 +120,20 @@ impl C8YHttpProxyActor {
...
     async fn init(&mut self) -> Result<(), C8YConnectionError> {
         info!(target: self.name(), "start initialisation");
-        while self.end_point.get_c8y_internal_id().is_empty() {
-            if let Err(error) = self.try_get_and_set_internal_id().await {
+
+        while self.end_point.get_internal_id(None).is_none() {
+            if let Err(error) = self
+                .try_get_and_set_internal_id(self.end_point.device_id.clone())
+                .await
+            {
```
didier-wenzek, line 136: Is this initialization loop still required, now that the code is ready to retry on obsolete internal ids?

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +165,146 @@ impl C8YHttpProxyActor {
...
+        };
 
-    async fn try_get_internal_id(
-        &mut self,
-        device_id: Option<&str>,
-    ) -> Result<String, C8YRestError> {
-        let url_get_id = self.end_point.get_url_for_get_id(device_id);
+        let request_builder = build_request(&self.end_point)?;
+
+        let request = request_builder.build()?;
```
didier-wenzek, line 193: I don't see the point to create and use the `build_request` closure in the same scope. Why not a simple method to build the request?

albinsuresh (commented):
Some cosmetic comments.

crates/core/c8y_api/src/http_proxy.rs
PradeepKiruvale: done

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -47,18 +61,18 @@ impl C8yEndPoint {
         url_get_id
     }
 
-    pub fn get_url_for_sw_list(&self) -> String {
+    pub fn get_url_for_sw_list(&self, device_id: Option<&str>) -> String {
         let mut url_update_swlist = self.get_base_url();
         url_update_swlist.push_str("/inventory/managedObjects/");
-        url_update_swlist.push_str(&self.c8y_internal_id);
+        url_update_swlist.push_str(&self.get_internal_id(device_id).unwrap_or_default());
```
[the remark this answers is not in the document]
PradeepKiruvale: Is it better to validate the `internal-id` before caching? Cache only if it's not an `empty` string? Or else this `get_url_for_software_list` has to return an error.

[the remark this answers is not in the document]
PradeepKiruvale: Fixed, it returns error if the `internal id` is empty.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -47,18 +61,18 @@ impl C8yEndPoint {
         url_get_id
     }
 
-    pub fn get_url_for_sw_list(&self) -> String {
+    pub fn get_url_for_sw_list(&self, device_id: Option<&str>) -> String {
```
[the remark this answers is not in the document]
PradeepKiruvale: I would keep this generic name.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -47,18 +61,18 @@ impl C8yEndPoint {
...
         let mut url_update_swlist = self.get_base_url();
         url_update_swlist.push_str("/inventory/managedObjects/");
-        url_update_swlist.push_str(&self.c8y_internal_id);
+        url_update_swlist.push_str(&self.get_internal_id(device_id).unwrap_or_default());
 
         url_update_swlist
     }
 
-    pub fn get_url_for_get_id(&self, device_id: Option<&str>) -> String {
+    pub fn get_url_for_get_id(&self, device_id: String) -> String {
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -17,24 +18,37 @@ use tracing::info;
...
             device_id: device_id.into(),
-            c8y_internal_id: c8y_internal_id.into(),
+            token: None,
+            devices_internal_id,
         }
     }
 
-    pub fn get_c8y_internal_id(&self) -> &str {
-        &self.c8y_internal_id
+    pub fn get_internal_id(&self, id: Option<&str>) -> Option<String> {
```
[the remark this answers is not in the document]
PradeepKiruvale: I would keep it as `device_id`, which is more generic.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +165,146 @@ impl C8YHttpProxyActor {
...
+        self.get_fresh_token().await?;
+        // build the request
+        let request_builder = build_request(&self.end_point);
+        let request = request_builder?.build()?;
+        // retry the request
         Ok(self.peers.http.await_response(request).await?)
     }
 
-    async fn create_event(
+    async fn retry_request_with_fresh_internal_id(
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +165,146 @@ impl C8YHttpProxyActor {
...
+            Err(e) => Err(C8YRestError::FromHttpError(e)),
+        }
+    }
+
+    async fn get_fresh_token(&mut self) -> Result<String, C8YRestError> {
+        self.end_point.token = None;
+        self.get_and_set_jwt_token().await
+    }
+
+    async fn retry_request_with_fresh_token(
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +165,146 @@ impl C8YHttpProxyActor {
...
-        }
-        self.send_event_internal(c8y_event).await
+        device_id: Option<String>,
+        build_request: impl Fn(&C8yEndPoint) -> Result<HttpRequestBuilder, C8YRestError>,
+    ) -> Result<HttpResult, C8YRestError> {
+        self.end_point
+            .clear_the_cached_internal_id(device_id.clone());
+        // get new internal id not the cached one
+        self.try_get_and_set_internal_id(device_id.unwrap_or(self.end_point.device_id.clone()))
+            .await?;
```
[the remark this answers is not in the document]
PradeepKiruvale: done, but named it as `get_and_set_internal_id`

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +165,146 @@ impl C8YHttpProxyActor {
...
+    async fn try_get_internal_id(&mut self, device_id: String) -> Result<String, C8YRestError> {
+        let url_get_id: String = self.end_point.get_url_for_get_id(device_id);
+        if self.end_point.token.is_none() {
+            self.get_fresh_token().await?;
         }
-    }
+        let build_request = |end_point: &C8yEndPoint| -> Result<HttpRequestBuilder, C8YRestError> {
+            Ok(HttpRequestBuilder::get(&url_get_id)
+                .bearer_auth(end_point.token.clone().unwrap_or_default()))
+        };
```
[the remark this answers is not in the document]
PradeepKiruvale: I made it like this just to reuse the `retry` code when it fails. So, the `try_with_fresh_jwt_token` function takes a `closure` that returns the `Result`. So, this has to be in consistent with that.

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: This `closure` is not just used inside this function. When the first attempt fails because of `401/403` then the retries with the new `token`. Then it calls the `try_with_fresh_token`, this function takes the `closure. So, I created a closure here.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -6,35 +6,46 @@ use mqtt_channel::StreamExt;
...
 use std::time::Duration;
 use tedge_config::mqtt_config::MqttConfigBuildError;
 use tedge_config::TEdgeConfig;
 use tracing::error;
 use tracing::info;
 
+#[derive(thiserror::Error, Debug)]
+pub enum C8yEndPointError {
+    #[error("Internal id is empty")]
+    EmptyInternalID,
```
didier-wenzek, line 19: Such an error message makes no sense for the end-user.

I would be better to get  rid of this error case and even error type. These are more a sign that the struct design is wrong (No internal should be encoded by `None`) and in any case the true error is more about "failing to get the c8y identifier for the device".

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -6,35 +6,46 @@ use mqtt_channel::StreamExt;
...
+}
+
 /// Define a C8y endpoint
 #[derive(Debug)]
 pub struct C8yEndPoint {
     pub c8y_host: String,
     pub device_id: String,
-    pub c8y_internal_id: String,
+    pub token: Option<String>,
+    pub devices_internal_id: HashMap<String, String>,
```
didier-wenzek, line 28: The asymmetry of this struct makes things fragile and confusing.

The asymmetry is that the main device id is stored in in its own field while the internal-id is stored in a map along all the child devices' ids and external ids.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -47,18 +58,22 @@ impl C8yEndPoint {
         url_get_id
     }
 
-    pub fn get_url_for_sw_list(&self) -> String {
+    pub fn get_url_for_sw_list(&self, device_id: Option<&str>) -> Result<String, C8yEndPointError> {
```
didier-wenzek, line 61: Using `None ` for the main device is a trick that makes the code not crystal clear.

A solution could be to introduce a specific enum with `MainDevice` and `ChildDevice("xyz")` cases.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +165,146 @@ impl C8YHttpProxyActor {
...
         let internal_id = internal_id_response.id();
+
         Ok(internal_id)
     }
 
     async fn execute(
         &mut self,
-        request_builder: HttpRequestBuilder,
+        device_id: Option<String>,
+        build_request: impl Fn(&C8yEndPoint) -> Result<HttpRequestBuilder, C8YRestError>,
```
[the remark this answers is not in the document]
didier-wenzek: > Yeah,its needed because get_software_list is called inside the closure and if it fails it will return an error.

There is no  `get_software_list` method nor function.

crates/core/c8y_api/src/http_proxy.rs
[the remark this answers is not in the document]
albinsuresh: I suggested `child_id` specifically since this API is expecting only child devices IDs as that argument and `None` for the main device. Keeping it generic would have been fine if this API accepted the main device's id as well. But this [suggestion](https://github.com/thin-edge/thin-edge.io/pull/2065#discussion_r1264984722) Didier seems like the best way to avoid all such confusion.

crates/core/c8y_api/src/http_proxy.rs
PradeepKiruvale: Refactored the error message and type. Renamed the type to `InternalIdNotFound`

crates/core/c8y_api/src/http_proxy.rs
PradeepKiruvale: Agreed, now refactored the map to hold `DeviceType` enum and `internal-id`.

crates/core/c8y_api/src/http_proxy.rs
PradeepKiruvale: refactored

crates/extensions/c8y_http_proxy/src/actor.rs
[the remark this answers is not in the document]
didier-wenzek: Now that you introduced methods that build urls from an internal id and never fail, I would try what is suggested by @albinsuresh . I.e. to remove the error case by passing to the closure the internal id.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +166,148 @@ impl C8YHttpProxyActor {
...
+            Ok(HttpRequestBuilder::get(&url_get_id)
+                .bearer_auth(end_point.token.clone().unwrap_or_default()))
+        };
 
-    async fn try_get_internal_id(
-        &mut self,
-        device_id: Option<&str>,
-    ) -> Result<String, C8YRestError> {
-        let url_get_id = self.end_point.get_url_for_get_id(device_id);
+        let request_builder = build_request(&self.end_point)?;
```
didier-wenzek, line 193: As already said, I would not use a closure to build the request to get the internal id of a device. I would simply use the url returned by `get_url_for_internal_id()` and not call `try_request_with_fresh_token()`. Sharing the later for regular requests and for the very specific request to get an internal id makes the code more complex than necessary.

didier-wenzek (commented):
The API for `C8yEndPoint` has now a nice shape. This should lead to further simplifications on `C8YHttpProxyActor`.

crates/extensions/c8y_http_proxy/src/actor.rs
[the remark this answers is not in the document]
PradeepKiruvale: Yes, that can be done by moving out the logic of getting the internal id into the `execute` and just sending the `internal-id` to the `closure`. However, this will make the code bit redundant. Passing the `end-point` and getting the `internal id` reduces the redundant code.

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: done

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +166,145 @@ impl C8YHttpProxyActor {
...
+        let create_event = |internal_id: String| -> C8yCreateEvent {
+            C8yCreateEvent {
+                source: Some(C8yManagedObject { id: internal_id }),
+                event_type: c8y_event.event_type.clone(),
+                time: c8y_event.time,
+                text: c8y_event.text.clone(),
+                extras: c8y_event.extras.clone(),
+            }
+        };
+        let device_type = self.end_point.get_device_type(None);
```
albinsuresh, line 288: The `device_type` will always be the main device? As you're passing `None` here. Why? What if an event has to be created for a child device?

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,80 +166,145 @@ impl C8YHttpProxyActor {
...
+                .get_internal_id(DeviceType::MainDevice)
+                .map_err(|e| C8YRestError::CustomError(e.to_string()))?;
+            let url = end_point.get_url_for_sw_list(internal_id);
+            Ok(HttpRequestBuilder::put(url)
+                .header("Accept", "application/json")
+                .header("Content-Type", "application/json")
+                .json(&software_list))
+        };
+
+        let http_result = self.execute(DeviceType::MainDevice, build_request).await?;
```
albinsuresh, line 307: Here also, hardcoded to `MainDevice`. Something looks wrong here. I understand that we don't support software list upload for child devices ATM. But this generic API assuming the same seems wrong. This API should be able to send software list for any device.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -327,43 +427,30 @@ impl C8YHttpProxyActor {
...
         &mut self,
-        c8y_event: C8yCreateEvent,
+        device_type: DeviceType,
+        create_event: impl Fn(String) -> C8yCreateEvent,
     ) -> Result<EventId, C8YRestError> {
-        let create_event_url = self.end_point.get_url_for_create_event();
+        // Get and set child device internal id
+        if !device_type.eq(&DeviceType::MainDevice) {
+            self.get_and_set_internal_id(device_type.clone()).await?;
+        }
```
albinsuresh, line 438: On a first look, it wasn't obvious to me as to why we're doing this here. But later understood.

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: Thats right. Since the `Software list` is supported only by the main device, I have hardcoded here. When the software list feature is supported by the child device, then one has to pass the `device ID` to `send_software_list_http()`, and based on that URL can be created.

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: Yes, this function `create_event` is only used by the `main` device. If a child device wants to use this function then it has to pass the `device` id, then the `DeviceType` can be derived out of this.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -102,6 +124,24 @@ impl C8yEndPoint {
         }
         false
     }
+
+    pub fn get_device_type(&self, device_id: Option<String>) -> DeviceType {
+        match device_id {
+            Some(device_id) => DeviceType::ChildDevice(device_id),
+            None => DeviceType::MainDevice,
+        }
+    }
```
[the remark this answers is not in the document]
PradeepKiruvale: I would not do that because the APIs have to be changed. I wanted to keep the `http_proxy` APIs as it is and do the mapping here. From the API pov, it's calling either for the `main device` or for the `child device`. 

Other way would be we must change the APIs to get the `device-id` always, irrespective of its for a child or the main device. For example for the `send_software_list_http()`` API, this must always take a `device-id`. That way things will be clearer.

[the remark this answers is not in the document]
albinsuresh: At some point we need to change those APIs to complete the cleanup what we've started here. The problem is that some of those APIs still accept request structs like `UploadConfigFile`, `UploadLogBinary` etc that has an `Option` field for the `device-id` which can be `None` for the main device. We should use the new `DeviceType` enum to those structs as well, so that we use it across the entire call stack and avoid this `None` magic completely.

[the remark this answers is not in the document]
didier-wenzek: @PradeepKiruvale I encourage you to follow the direction given by @albinsuresh

crates/extensions/c8y_http_proxy/src/actor.rs
didier-wenzek: I don't see what has been updated.

PradeepKiruvale: Removed the closure and used the `URL` returned by the `get_url_for_internal_id()`.

crates/extensions/c8y_http_proxy/src/actor.rs
albinsuresh: That doesn't sound right. Even if a child device sends an event with extra fragments, it has to go through this same API call. So, we're missing something here.

crates/extensions/c8y_config_manager/src/upload.rs
```diff
@@ -90,7 +90,7 @@ impl ConfigUploadManager {
                     self.upload_config_file(
                         Path::new(config_file_path.as_str()),
                         &config_upload_request.config_type,
-                        None,
+                        "main".into(),
```
albinsuresh, line 93: We should be passing the actual `device-id` here rather than this special key.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -117,20 +117,25 @@ impl Actor for C8YHttpProxyActor {
...
     }
 
     async fn init(&mut self) -> Result<(), C8YConnectionError> {
         info!(target: self.name(), "start initialisation");
-        while self.end_point.get_c8y_internal_id().is_empty() {
-            if let Err(error) = self.try_get_and_set_internal_id().await {
+
+        while self
+            .end_point
+            .get_internal_id(self.end_point.device_id.clone())
```
albinsuresh, line 132:
```suggestion
            .end_point
            .get_main_device_internal_id()
```
Such a function would be better.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -117,20 +117,25 @@ impl Actor for C8YHttpProxyActor {
...
+            .is_err()
+        {
+            if let Err(error) = self
+                .get_and_set_internal_id(self.end_point.device_id.clone())
```
albinsuresh, line 136:
```suggestion
                .get_and_set_main_device_internal_id()
```

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -162,149 +167,241 @@ impl C8YHttpProxyActor {
...
+        device_id: String,
+        build_request: impl Fn(&C8yEndPoint) -> Result<HttpRequestBuilder, C8YRestError>,
     ) -> Result<HttpResult, C8YRestError> {
-        // Get a JWT token to authenticate the device
-        let request_builder = if let Ok(token) = self.peers.jwt.await_response(()).await? {
-            request_builder.bearer_auth(token)
-        } else {
-            return Err(C8YRestError::CustomError("JWT token not available".into()));
-        };
+        let device_id = self.end_point.get_device_id(device_id);
```
albinsuresh, line 215: That looks really weird. Deriving the `device_id` from the `device_id`. I'm assuming that this is happening because of that special handling of "main" literal. If possible, I'd avoid passing that special char around.

crates/extensions/c8y_http_proxy/src/handle.rs
```diff
@@ -61,12 +61,12 @@ impl C8YHttpProxy {
         &mut self,
         log_type: &str,
         log_content: &str,
-        child_device_id: Option<String>,
+        source: String,
```
albinsuresh, line 64:
```suggestion
        device_id: String,
```

crates/extensions/c8y_http_proxy/src/handle.rs
```diff
@@ -79,12 +79,12 @@ impl C8YHttpProxy {
         &mut self,
         config_path: &Path,
         config_type: &str,
-        child_device_id: Option<String>,
+        source: String,
```
albinsuresh, line 82:
```suggestion
        device_id: String,
```

crates/extensions/c8y_http_proxy/src/messages.rs
```diff
@@ -47,14 +47,14 @@ pub struct GetJwtToken;
 pub struct UploadLogBinary {
     pub log_type: String,
     pub log_content: String,
-    pub child_device_id: Option<String>,
+    pub source: String,
```
albinsuresh, line 50:
```suggestion
    pub device_id: String,
```

crates/extensions/c8y_http_proxy/src/messages.rs
```diff
@@ -47,14 +47,14 @@ pub struct GetJwtToken;
...
 }
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct UploadConfigFile {
     pub config_path: PathBuf,
     pub config_type: String,
-    pub child_device_id: Option<String>,
+    pub source: String,
```
albinsuresh, line 57:
```suggestion
    pub device_id: String,
```

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -117,6 +118,8 @@ impl From<&SoftwareListResponse> for C8yUpdateSoftwareListResponse {
 
         Self {
             c8y_software_list: Some(new_list),
+            // Must derive the source from the topic
+            source: "main".into(),
```
albinsuresh, line 122: This source field **should not** have been implicit like this. It should have been injected/derived from outside. When we want to extend software list support for child devices, this will have to change.

PradeepKiruvale: Refactored, now it gets the device id from the caller.

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: refactored

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -102,6 +114,14 @@ impl C8yEndPoint {
         }
         false
     }
+
+    pub fn get_device_id(&self, device_id: String) -> String {
+        if device_id.eq("main") {
```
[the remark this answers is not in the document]
PradeepKiruvale: refactored

crates/extensions/c8y_config_manager/src/upload.rs
PradeepKiruvale: done

crates/extensions/c8y_http_proxy/src/messages.rs
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/messages.rs
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -90,17 +90,17 @@ impl Actor for C8YHttpProxyActor {
                     .await
                     .map(|response| response.into()),
 
-                C8YRestRequest::C8yUpdateSoftwareListResponse(request) => self
-                    .send_software_list_http(request.source.clone(), request)
+                C8YRestRequest::SoftwareListResponse(request) => self
+                    .send_software_list_http(request.device_id.clone(), request)
```
albinsuresh, line 94:
```suggestion
                    .send_software_list_http(request)
```
When the `request` instance already has the `device_id` in it, why extract it and pass it separately?

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -90,17 +90,17 @@ impl Actor for C8YHttpProxyActor {
...
                     .await
                     .map(|response| response.into()),
 
                 C8YRestRequest::UploadLogBinary(request) => self
-                    .upload_log_binary(request.source.clone(), request)
+                    .upload_log_binary(request.device_id.clone(), request)
```
albinsuresh, line 99:
```suggestion
                    .upload_log_binary(request)
```
Same as above.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -90,17 +90,17 @@ impl Actor for C8YHttpProxyActor {
...
                     .await
                     .map(|response| response.into()),
                 C8YRestRequest::UploadConfigFile(request) => self
-                    .upload_config_file(request.source.clone(), request)
+                    .upload_config_file(request.device_id.clone(), request)
```
albinsuresh, line 103:
```suggestion
                    .upload_config_file(request)
```
Same as above.

albinsuresh (approved):

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: fixed

crates/extensions/c8y_http_proxy/src/actor.rs
PradeepKiruvale: fixed

PR #2068, opened 2023-07-12
Fix bug in ConcurrentServerMessageBox

Bravo555:
## Proposed changes

The bug in ConcurrentServerMessageBox::await_idle_processor caused runtime shutdown requests to not have intended effect (closing the actor) when the message box was processing in-progress requests.

This fix works for this specific case, but the message box can still block in other places, e.g. when sending a response if the channel is full. When this happens, if the receiver is not freed before the sender, we block.
## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_actors/src/servers/message_boxes.rs
+++ b/crates/core/tedge_actors/src/servers/message_boxes.rs
@@ -3,6 +3,7 @@ use crate::ChannelError;
 use crate::Message;
 use crate::MessageReceiver;
 use crate::NoConfig;
+use crate::RuntimeRequest;
 use crate::Sender;
 use crate::ServiceConsumer;
 use crate::ServiceProvider;
@@ -10,6 +11,8 @@ use crate::SimpleMessageBox;
 use crate::SimpleMessageBoxBuilder;
 use futures::StreamExt;
 use std::fmt::Debug;
+use std::ops::ControlFlow;
+use tokio::select;
 
 /// A message box for a request-response server
 pub type ServerMessageBox<Request, Response> =
@@ -53,7 +56,10 @@ impl<Request: Message, Response: Message> ConcurrentServerMessageBox<Request, Re
     }
 
     async fn next_request(&mut self) -> Option<(usize, Request)> {
-        self.await_idle_processor().await;
+        if self.await_idle_processor().await.is_break() {
+            return None;
+        }
+
         loop {
             tokio::select! {
                 Some(request) = self.clients.recv() => {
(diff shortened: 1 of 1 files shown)


crates/core/tedge_actors/src/servers/message_boxes.rs
```diff
@@ -123,3 +145,165 @@ impl<Request: Message, Response: Message> ClientMessageBox<Request, Response> {
...
+
+        // Expect first 4 tasks to be in-progress
+        assert_eq!(rx.recv().await, Some((0usize, 0)));
+        assert_eq!(rx.recv().await, Some((1usize, 1)));
+        assert_eq!(rx.recv().await, Some((2usize, 2)));
+        assert_eq!(rx.recv().await, Some((3usize, 3)));
+
+        // Expect at this point in time that 5th task hasn't started executing
+        // yet
+        // TODO: remove waits
```
didier-wenzek, line 210: Why do you think these waits are useful in the current setting?

crates/core/tedge_actors/src/servers/message_boxes.rs
```diff
@@ -123,3 +145,165 @@ impl<Request: Message, Response: Message> ClientMessageBox<Request, Response> {
...
+        }
+
+        let mut runtime = Runtime::try_new(None).await.unwrap();
+
+        let mut server_actor_builder = TestServerBuilder::new();
+        let mut client_box = server_actor_builder.client_box();
+
+        runtime.spawn(server_actor_builder).await.unwrap();
+
+        // TODO: remove waits
```
didier-wenzek, line 287: Why do you think these waits are useful in the current setting? Is this to avoid sending the shutdown request even before the server started to handle requests? If so you can use a channel for the tasks to tell they have been launched (instead of `eprintln!("got request: {_request:?}");`)

Bravo555: My bad, forgot to delete the `eprintln!`, and indeed here it is very trivial to replace the waits with a channel.

crates/core/tedge_actors/src/servers/message_boxes.rs
```diff
@@ -69,11 +75,27 @@ impl<Request: Message, Response: Message> ConcurrentServerMessageBox<Request, Re
...
+            // a regular return, because then next_request wouldn't see it
+            //
+            // a better approach would be to do select on top-level entry point,
+            // then we'd be sure we're able to cancel when anything happens, not
+            // just when waiting for pending_responses, e.g. if send_result
+            // stalls
+            Some(RuntimeRequest::Shutdown) = self.clients.recv_signal() => {
+                ControlFlow::Break(())
             }
+            else => ControlFlow::Continue(())
```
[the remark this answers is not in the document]
Bravo555: fixed

crates/core/tedge_actors/src/servers/message_boxes.rs
Bravo555: The original attempt was to trade some of the perceived flakiness for delays. I was being overly cautious with assuming that spawned tasks might not begin executing immediately. I shuffled some channels, added some more asserts, and removed the waits, and now I'm reasonably confident that they shouldn't be flaky.

Bravo555:
The integration test that failed seems unrelated to the change, and it also works on my machine. Can we rerun the integration tests?

didier-wenzek (approved):
Approved

Issue #2048, opened 2023-06-26
Make log plugin usable for different clouds


Ruadhri17:
**Is your refactoring request related to a problem? Please describe.**
This ticket is part of #2017. It considers moving cloud-specific logic to the mapper or refactoring it to be cloud agnostic. 

**Describe the solution you'd like**

1. Move cloud specific logic from the c8y-log-plugin to the mapper
2. Rename the c8y-log-plugin to tedge-log-plugin once all cloud specific logic has been moved.

**Describe alternatives you've considered**

**Additional context**
See #2017

**Tasks**
- [x] tedge-log-plugin specification: #2069 
- [x] tedge-log-plugin code change: #2086 
- [x] c8y-mapper code change:  #2108 
- [x] Robot framework test: https://github.com/thin-edge/thin-edge.io/pull/2213
- [x] Update specs to align with implemenentation after tedge-log-plugin and c8y-mapper PR are merged. #2216 
- [x] Introduce `root.topic` and `device.topic` to tedge config so that tedge-log-plugin and c8y-mapper can use them. (later?)

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite. 
- [x] QA has tested the feature and it meets the required specifications.

rina23q:
@reubenmiller Why is this ticket still opened?

reubenmiller:
> @reubenmiller Why is this ticket still opened?

I just forgot to close it after marking it with the 0.13.0 milestone (apparently)

PR #2069, opened 2023-07-12
tedge-log-plugin specification

rina23q:
## Proposed changes
This is a PR for `tedge-log-plugin` specification.

### Open topics
* States. Need to sync with the state machine spec later.

## Types of changes

## Paste Link to the issue
#2048 

## Checklist

### DIFF
--- a/docs/src/references/tedge-log-management.md
+++ b/docs/src/references/tedge-log-management.md
@@ -1,5 +1,5 @@
 ---
-title: New Log File Management Plugin
+title: New Log File Management
 tags: [Reference, Log Files]
 sidebar_position: 7
 draft: true
@@ -22,13 +22,14 @@ Thin-edge provides an operation plugin to fetch log files from the device.
 * However, the plugin provides no direct connection to clouds, which is the responsibility of another component, i.e. the cloud mapper.
 * The plugin has a dependency on the `tedge.toml` configuration file to get the MQTT hostname, port, and device identifier.
 * The plugin establishes an MQTT connection to the broker using the `mqtt.bind.address` and `mqtt.bind.port` values from the `tedge.toml` configuration.
+* The `<root>` and `<identifier>` for the topic to publish and subscribe are defined in `tedge.toml` file as `root.topic` and `device.topic`.
 
 ## Installation
 
 As part of this plugin installation:
 * On systemd-enabled devices, the service definition file for this `tedge-log-plugin` daemon is also installed.
 
-Once installed, the `tedge-log-plugin` runs as a daemon on the device, listening to log update commands on the [`<root>/<identifier>/cmd/log_upload/+` MQTT topic](mqtt-topic-design.md).
+Once installed, the `tedge-log-plugin` runs as a daemon on the device, listening to log update commands on the [`<root>/<identifier>/cmd/log_upload/+` MQTT topic](mqtt-api.md).
 
 ## Configuration
 
@@ -48,9 +49,9 @@ files = [
 
 The `tedge-log-plugin` parses this configuration file on startup for all the `type` values specified,
 and sends the supported log types message to the MQTT local broker on the `<root>/<identifier>/cmd/log_upload` topic with a retained flag.
-The `<root>` and `<identifier>` are defined in `tedge.toml` file as `device.topic.root` and `device.topic.id`.
 
-Given that `device.topic.root` and `device.topic.id` are set to `te` and `device/main//` for the main device, the message to declare the supported log types is as follows.
+Given that `root.topic` and `device.topic` are set to `te` and `device/main//` for the main device,
+the message to declare the supported log types is as follows.
 
 ```sh te2mqtt
 tedge mqtt pub -r 'te/device/main///cmd/log_upload' '{
@@ -68,7 +69,8 @@ then a JSON message with an empty array for the `types` field is sent, indicatin
 
 ## Handling log upload commands
 
-The plugin is listening to log update commands on the [`<root>/<identifier>/cmd/log_upload/+` MQTT topic](mqtt-topic-design.md). For example, it subscribes to the following topic for the main device.
+The plugin is listening to log update commands on the [`<root>/<identifier>/cmd/log_upload/+` MQTT topic](mqtt-api.md).
+For example, it subscribes to the following topic for the main device.
 
 ```sh te2mqtt
 tedge mqtt sub 'te/device/main///cmd/log_upload/+'
@@ -79,7 +81,7 @@ A new log file upload command with the ID "1234" is published by another compone
 ```sh te2mqtt
 tedge mqtt pub -r 'te/device/main///cmd/log_upload/1234' '{
   "status": "init",
-  "url": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
+  "tedgeUrl": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
   "type": "mosquitto",
   "dateFrom": "2013-06-22T17:03:14.000+02:00",
   "dateTo": "2013-06-23T18:03:14.000+02:00",
@@ -93,7 +95,7 @@ retrieves the log files using the `path` glob pattern provided in the plugin con
 including only the ones modified within the date range(`2013-06-22T17:03:14.000+02:00` to `2013-06-23T18:03:14.000+02:00`),
 with the content filtered by the search text(`ERROR`) and the maximum line count(`1000`).
 
-This filtered content is then uploaded to the given URL as `url` via an HTTP PUT request.
+This filtered content is then uploaded to the given URL as `tedgeUrl` via an HTTP PUT request.
 
 During the process, the plugin updates the command status via MQTT
 by publishing a retained message to the same `<root>/<identifier>/cmd/log_upload/<id>` topic,
@@ -109,7 +111,7 @@ Thus, the operation status update message for the above example looks like below
 tedge mqtt pub -r 'te/device/main///cmd/log_upload/1234' '{
   "status": "failed",
   "reason": "The target log file for 'mosquitto' does not exist.",
-  "url": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
+  "tedgeUrl": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
   "type": "mosquitto",
   "dateFrom": "2013-06-22T17:03:14.000+02:00",
   "dateTo": "2013-06-22T18:03:14.000+02:00",
@@ -167,7 +169,7 @@ OPTIONS:
 
 The thin-edge `CONFIG_DIR` is used:
   * to find where to connect the MQTT bus: `tedge config get mqtt.bind.address` and `tedge config get mqtt.bind.port`
-  * to find where to publish/subscribe: `tedge config get device.topic`
+  * to find where to publish/subscribe: `tedge config get root.topic` and `tedge config get device.topic`
   * to store tedge-log-plugin.toml: the configuration file that specifies which logs to be retrieved
 ```
 
--- a/docs/src/references/tedge-log-management.md
+++ b/docs/src/references/tedge-log-management.md
@@ -69,7 +69,7 @@ then a JSON message with an empty array for the `types` field is sent, indicatin
 
 ## Handling log upload commands
 
-The plugin is listening to log update commands on the [`<root>/<identifier>/cmd/log_upload/+` MQTT topic](mqtt-api.md).
+The plugin subscribes to log upload commands on the [`<root>/<identifier>/cmd/log_upload/+` MQTT topic](mqtt-api.md).
 For example, it subscribes to the following topic for the main device.
 
 ```sh te2mqtt
@@ -90,12 +90,12 @@ tedge mqtt pub -r 'te/device/main///cmd/log_upload/1234' '{
 }'
 ```
 
-The plugin then checks the `tedge-log-plugin.toml` file for the log type in the incoming message (`mosquitto`),
+The plugin then checks the `tedge-log-plugin.toml` file for the log `type` in the incoming message (`mosquitto`),
 retrieves the log files using the `path` glob pattern provided in the plugin config file,
 including only the ones modified within the date range(`2013-06-22T17:03:14.000+02:00` to `2013-06-23T18:03:14.000+02:00`),
 with the content filtered by the search text(`ERROR`) and the maximum line count(`1000`).
 
-This filtered content is then uploaded to the given URL as `tedgeUrl` via an HTTP PUT request.
+This filtered content is then uploaded to the URL received in the command as `tedgeUrl` via an HTTP PUT request.
 
 During the process, the plugin updates the command status via MQTT
 by publishing a retained message to the same `<root>/<identifier>/cmd/log_upload/<id>` topic,
@@ -168,9 +168,10 @@ OPTIONS:
             Print version information
 
 The thin-edge `CONFIG_DIR` is used:
-  * to find where to connect the MQTT bus: `tedge config get mqtt.bind.address` and `tedge config get mqtt.bind.port`
-  * to find where to publish/subscribe: `tedge config get root.topic` and `tedge config get device.topic`
-  * to store tedge-log-plugin.toml: the configuration file that specifies which logs to be retrieved
+  * to find the `tedge.toml` where the following configs are defined:
+     ** `mqtt.bind.address` and `mqtt.bind.port`: to connect to the tedge MQTT broker
+     ** `root.topic` and `device.topic`: for the MQTT topics to publish to and subscribe from
+  * to find/store the `tedge-log-plugin.toml`: the configuration file that specifies which logs to be retrieved
 ```
 
 ## Logging
--- a/docs/src/references/tedge-log-management.md
+++ b/docs/src/references/tedge-log-management.md
@@ -1,36 +1,39 @@
 ---
-title: New Log File Management
+title: New Log File Management Plugin
 tags: [Reference, Log Files]
 sidebar_position: 7
+draft: true
 ---
 
-# Log file management from cloud
+# Log file management plugin
 
-Thin-edge provides an operation plugin to fetch log files from the device on to cloud.
+Thin-edge provides a command plugin to fetch log files from the device.
 
-* Log file management from cloud is provided with a `tedge-log-plugin` which runs as a daemon on thin-edge together with cloud mapper.
-* The device owner can define the list of log files that can be retrieved from cloud,
-  in the plugin's configuration file named `tedge-log-plugin.toml`.
+* Log file management feature is provided with a `tedge-log-plugin` which runs as a daemon on thin-edge.
+* The device owner can define the list of log files in the plugin's configuration file named `tedge-log-plugin.toml`.
 * Each entry in the the `tedge-log-plugin.toml` file contains a log `type` and a `path` pattern,
   where the `type` is used to represent the logical group of log files matching the `path` pattern.
-* On receipt of a log file request for a given `type`, 
+* On receipt of a log file upload command for a given `type`, 
   the log files for that type are retrieved using the `path` pattern defined in this `tedge-log-plugin.toml`,
   matched against the requested time range, search text and maximum line count.
-* The list of managed log files in `tedge-log-plugin.toml` can be updated both locally as well as from cloud, for instance,
-  using the configuration management feature of Cumulocity, combined with the `tedge-configuration-plugin` of thin-edge.
+* The plugin uploads the requested log file to the tedge file transfer repository,
+  which url is given by the received log upload command.
+* The list of managed log files in `tedge-log-plugin.toml` can be updated both locally as well as from cloud, for instance, by using the configuration management feature.
+* However, the plugin provides no direct connection to clouds, which is responsibility of another component, i.e. cloud mapper.
+* The plugin has dependency on the `tedge.toml` device configuration file to get the MQTT hostname, port and the device identifier.
 
 ## Installation
 
 As part of this plugin installation:
 * On systemd enabled devices, the service definition file for this `tedge-log-plugin` daemon is also installed as part of this plugin installation.
 
-Once installed, the `tedge-log-plugin` is run as a daemon on the device listening to log requests from cloud mapper on `te/device/main///cmd/log_upload/+` MQTT topic.
+Once installed, the `tedge-log-plugin` is run as a daemon on the device listening to log update commands on `<root>/<identifier>/cmd/log_upload/+` MQTT topic.
 
 ## Configuration
 
-The `tedge-log-plugin` configuration is stored by default under `/etc/tedge/plugins/tedge-log-plugin.toml`.
+The `tedge-log-plugin` configuration is stored by default under `/etc/tedge/plugins/tedge-log-plugin.toml`. If it does not exist on startup, the plugin creates the file with example contents.
 
-This [TOML](https://toml.io/en/) file defines the list of log files that can be retrieved from the cloud tenant.
+This [TOML](https://toml.io/en/) file defines the list of log files that should be managed by the plugin.
 The paths to these files can be represented using [glob](https://en.wikipedia.org/wiki/Glob_(programming)) patterns.
 The `type` given to these paths are used as the log type associated to a log path.
 
@@ -43,12 +46,12 @@ files = [
 ```
 
 The `tedge-log-plugin` parses this configuration file on startup for all the `type` values specified,
-and sends the supported log types message to MQTT local broker on `te/device/main///cmd/log_upload` topic as follows:
+and sends the supported log types message to MQTT local broker on `<root>/<identifier>/cmd/log_upload` topic with a retained flag as follows:
 
-```json
-{
+```sh te2mqtt
+tedge mqtt pub -r '<root>/<identifier>/cmd/log_upload' '{
   "types" : [ "mosquitto", "software-management", "c8y_CustomOperation" ]
-}
+}'
 ```
 
 The plugin continuously watches this configuration file for any changes and resends the JSON message with the `type`s in this file,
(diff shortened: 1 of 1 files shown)


docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+
+* Renaming the toml file from `c8y-log-plugin.toml` to `tedge-log-plugin.toml`.
+* Supported operation. Create automatically `c8y_LogfileRequest` as it is now? Or ask user to create it?
+* The location of the toml file. Change from `/etc/tedge/c8y/c8y-log-plugin.toml` to `/etc/tedge/plugins/tedge-log-plugin.toml`?
+
+## Check points (MQTT topic names and payloads)
+
+* Supported log type declaration format. CSV vs JSON? I prefer JSON as it's easy to extend fields in the future.
+* Type declaration topic: `te/device/main///cmd/log_upload`?
+* For receiving operation from mapper: `te/device/main///cmd/log_upload`?
```
reubenmiller, line 230: The command topic should have the command id as the last field (including on the initial creation of the command).

```suggestion
* For receiving operation from mapper: `te/device/main///cmd/log_upload/<cmd_id>`?
```

The command topic without the `/<cmd_id>` suffix, e.g. `te/device/main///cmd/log_upload`, is used by a component to signal that the `log_upload` capability is supported on the main device.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+
+## Logging
+
+The `tedge-log-plugin` reports progress and errors to the OS journal which can be retrieved using `journalctl`.
+
+
+## Not clear / undecided
+
+* Renaming the toml file from `c8y-log-plugin.toml` to `tedge-log-plugin.toml`.
+* Supported operation. Create automatically `c8y_LogfileRequest` as it is now? Or ask user to create it?
```
reubenmiller, line 223: I suspect this will be influence by the new proposal to support a declaration of thin-edge agnostic command support (e.g. by a component publishing a retained mqtt message to `te/<indentifier>/cmd/<cmd_type>`). The cloud mappers could register to each of these command meta topics to see which command type are supported on the agnostic level, and decide if it knows how to map them to the cloud specific name, e.g. `log_upload` --> `c8y_LogFileRequest`...so I'm not sure what value having the physical file has, though we'll need to figure out how to handle it in the transitional phase whilst we're still finalizing the command meta registration details.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+whenever it is updated.
+
+:::note
+If the file `/etc/tedge/plugins/tedge-log-plugin.toml` is ill-formed or cannot be read,
+then a JSON message with an empty array for the `types` field is sent, indicating no log files are tracked.
+:::
+
+## Startup
+
+The plugin declares **supported operation** and **supported log types** on startup.
```
albinsuresh, line 64:
```suggestion
The plugin declares its **supported operation type** and **supported log types** on startup by sending the supported log types message described above.
```

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+It's up to cloud mappers, if it converts the declaration messages to the corresponding cloud format or discard them.
+
+### Supported operation
+
+On startup, the plugin declares to the Cumulocity mapper that it supports `c8y_LogfileRequest` operation
+by creating an empty file at `/etc/tedge/operations/c8y/c8y_LogfileRequest`.
```
albinsuresh, line 70: The mapper should be creating that file on receipt of the tedge supported log types message described above and below.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+
+This filtered content is then uploaded to the given url as `uploadUrl` via HTTP PUT request.
+
+During this process, the plugin need to update the operation status via MQTT for `executing `, `uploading` and `failed` 
+by publishing a message to the topic `te/device/main///cmd/log_upload/<op_id>`.
+
+So, for example, the operation status update message should look like below. The topic is `te/device/main///cmd/log_upload/1234`.
+
+```json
+// Why do we need to repeat other fileds than status and reason?
```
albinsuresh, line 113: I don't think we need all of that, especially since we're sending the responses to a specific `op_id` topic. Just the `status` and `uploadUrl` are more than enough on success. Just for more context, we can add just the `logType` as well, if needed, but nothing else.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+
+  Cloud->>Mapper: Cloud-specific Log file upload request (Status: init)
+  Mapper->>Plugin: Cloud-agnostic Log file upload request (Status: init)
+  Plugin->>Mapper: Status: executing
+  alt if necessary
+    Mapper->>Cloud: Status: executing
+  end
+  Plugin->>Plugin: Extract log
+  Plugin->>Mapper: File transfer [HTTP]
+  Plugin->>Mapper: Status: uploading
```
albinsuresh, line 145:
```suggestion
  Plugin->>Mapper: Status: uploading
  Plugin->>Mapper: File transfer [HTTP]
  Plugin->>Mapper: Status: successful
```
The `uploading` status must be sent before starting the upload, right? No point in sending it after the upload completes. Similarly, the completion of upload must be notified to the mapper either with a `successful` or `failed` status.

docs/src/references/tedge-log-management.md
albinsuresh: It should be the mapper that creates this file for its own tracking when it receives the supported log types message from the plugin.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+* Renaming the toml file from `c8y-log-plugin.toml` to `tedge-log-plugin.toml`.
+* Supported operation. Create automatically `c8y_LogfileRequest` as it is now? Or ask user to create it?
+* The location of the toml file. Change from `/etc/tedge/c8y/c8y-log-plugin.toml` to `/etc/tedge/plugins/tedge-log-plugin.toml`?
+
+## Check points (MQTT topic names and payloads)
+
+* Supported log type declaration format. CSV vs JSON? I prefer JSON as it's easy to extend fields in the future.
+* Type declaration topic: `te/device/main///cmd/log_upload`?
+* For receiving operation from mapper: `te/device/main///cmd/log_upload`?
+* Do we need to include other fields in payload for status update message?
```
albinsuresh, line 231: I'd say no to everything, other than the fields absolutely needed for context like `logType`. IMO only the fields required for the next "state" to make progress is sufficient.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+  Plugin->>Mapper: File transfer [HTTP]
+  Plugin->>Mapper: Status: uploading
+  alt if necessary
+    Mapper->>Cloud: Status: uploading
+  end
+  Mapper->>Cloud: Upload log file [HTTP] (FAILED!)
+  Cloud-->> Mapper: Return the http endpoint of the uploaded file [HTTP] (FAILED CODE!)
+  alt if necessary
+    Mapper->>Cloud: Cloud-specific Status failed
+  end
```
albinsuresh, line 208:
```suggestion
  Plugin->>Mapper: Status: uploading
  Plugin->>Mapper: File transfer [HTTP]
  Plugin->>Mapper: Status: successful
  alt if necessary
    Mapper->>Cloud: Status: uploading
  end
  Mapper->>Cloud: Upload log file [HTTP] (FAILED!)
  alt if necessary
    Mapper->>Cloud: Cloud-specific Status failed
  end
```

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+    Mapper->>Cloud: Status: executing
+  end
+  Plugin->>Plugin: Extract log
+  Plugin->>Mapper: File transfer [HTTP]
+  Plugin->>Mapper: Status: uploading
+  alt if necessary
+    Mapper->>Cloud: Status: uploading
+  end
+  Mapper->>Cloud: Upload log file [HTTP] (FAILED!)
+  Cloud-->> Mapper: Return the http endpoint of the uploaded file [HTTP] (FAILED CODE!)
```
albinsuresh, line 205: I didn't really understand why the cloud would return a URL when the upload failed.

docs/src/references/tedge-log-management.md
didier-wenzek: I would even consider to more use such files.

Indeed, the `log_upload` capability will be published as a retained message on `te/device/main///cmd/log_upload`
for the main device and on `te/device/child-xyz///cmd/log_upload` for a child device. A mapper has to subscribe to all these retained messages and to translate them appropriately for the cloud.

So a `c8y_LogfileRequest` file can be used for caching purpose but not to detect that the capability is available or not.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+```toml title="file: /etc/tedge/plugins/tedge-log-plugin.toml"
+files = [
+  { type = "mosquitto", path = '/var/log/mosquitto/mosquitto.log' },
+  { type = "software-management", path = '/var/log/tedge/agent/software-*' },
+  { type = "c8y_CustomOperation", path = '/var/log/tedge/agent/c8y_CustomOperation/*' }
+]
+```
+
+The `tedge-log-plugin` parses this configuration file on startup for all the `type` values specified,
+and sends the supported log types message to MQTT local broker on `te/device/main///cmd/log_upload` topic as follows:
```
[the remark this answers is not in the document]
rina23q: All messages should be retained? I mean others than the this command registration message?

docs/src/references/tedge-log-management.md
rina23q: Yeah, your comment fits what I thought. I will update.

docs/src/references/tedge-log-management.md
rina23q: Since the states will be defined in Didier's state machine definition, I decided to keep status simplest as of now. So,
* Plugin sends only 3 status: executing, successful, and failed.
* The timing of successful is after uploading to http file transfer repository is done. (= that's the end of plugin's one cycle of handling an operation)
* Mapper still needs to upload the log file to cloud, then send a successful message to cloud. However, this "successful" is different from the "successful" from plugin.
I will update accordingly soon.

docs/src/references/tedge-log-management.md
rina23q: It's a copy & paste mistake!

docs/src/references/tedge-log-management.md
rina23q: I'll update this part as I mentioned above. So quite similar to what you suggested, but no longer "uploading", instead using "successful" as it's successful for plugin.

docs/src/references/tedge-log-management.md
didier-wenzek: I would only *add* new fields as the operation request advances. It's also important to not reject and keep unknown fields. These two requirements are important if we want to *extend* the workflow of an operation, as a software component added by the used might need *extra* fields we are not aware of.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+Topic (provided "1234" is operation ID):
+```text
+te/device/main///cmd/log_upload/1234
+```
+
+Payload:
+```json
+{
+  "status": "init",
+  "uploadUrl": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto",
```
didier-wenzek, line 91: I would use an url that is specific to this operation request. To avoid clash with concurrent requests.

```suggestion
  "uploadUrl": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
```

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,232 @@
...
+* The list of managed log files in `tedge-log-plugin.toml` can be updated both locally as well as from cloud, for instance,
+  using the configuration management feature of Cumulocity, combined with the `tedge-configuration-plugin` of thin-edge.
+
+## Installation
+
+As part of this plugin installation:
+* On systemd enabled devices, the service definition file for this `tedge-log-plugin` daemon is also installed as part of this plugin installation.
+
+Once installed, the `tedge-log-plugin` is run as a daemon on the device listening to log requests from cloud mapper on `te/device/main///cmd/log_upload/+` MQTT topic.
+
```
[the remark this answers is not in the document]
rina23q: I want to confirm the scope of the drop 1. Drop 1 should support more than `main` device? Because I thought it just needed to address for main as of now on `c8y-log-plugin`...

Anyway, then there are question point of the design.

Given that the plugin is installed in `main` device,
* it should subscribe to `te/device/main///cmd/log_upload/+` for sure.
* For child? Shouldn't it be installed directly on child?

Given that the plugin is installed in `child01` device,
* it should not subscribe to `te/device/main///cmd/log_upload/+` as the device is not main.
* it should subscribe precisely to `te/device/child01///cmd/log_upload/+`. However, how can the plugin know the child name?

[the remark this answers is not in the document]
didier-wenzek: > I want to confirm the scope of the drop 1. Drop 1 should support more than `main` device? Because I thought it just needed to address for main as of now on `c8y-log-plugin`...

The first implementation can focus on the main device. But the design must have a broader view.

The beauty of the MQTT topic re-design is that they are very little requirements to have the *same* implementation being able to run on the main device or a child device:
- The plugin needs to know the topic identifier of the device it's running on.
- All file transfers must be done over HTTP (as you propose here).

Note also, that the mapper doesn't need to change its interactions with the plugin depending on it's a child device or not (something really error prone in the current code base).


> Given that the plugin is installed in `main` device,
* it should subscribe to `te/device/main///cmd/log_upload/+` for sure.
* For child? Shouldn't it be installed directly on child?

The plugin must indeed be installed on the child device to get a direct access of the log files of this device.

> Given that the plugin is installed in `child01` device,
* it should not subscribe to `te/device/main///cmd/log_upload/+` as the device is not main.
* it should subscribe precisely to `te/device/child01///cmd/log_upload/+`. However, how can the plugin know the child name?

Sure, the plugin needs to know the topic identifier of the device it's running on. A new config setting is required for that.

docs/src/references/tedge-log-management.md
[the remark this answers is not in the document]
albinsuresh: Not all messages. Only the ones that updates the `entity store` like the supported capablities messages like this one. The regular command execution messages (`init`, `executing`, `successful` etc) continue to be just regular messages.

docs/src/references/tedge-log-management.md
rina23q: I decided to propose to include all information for a command inside the payload for the robustness in the future. If the component who is supposed to receive the message is not interested in other fields, they can just ignore. On the contrary, it's more difficult to get missing data.

didier-wenzek (commented):
I really like this proposal, notably the fact the very same plugin can run either on the main device or a child device.

However, the text is not always easy to read.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,195 @@
...
+```run command="tedge-log-plugin --help" lang="text" title="Output"
+Thin-edge device log file retriever
+
+USAGE:
+    tedge-log-plugin [OPTIONS]
+
+OPTIONS:
+        --config-dir <CONFIG_DIR>
+            [default: /etc/tedge/plugins]
```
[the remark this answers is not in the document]
rina23q: Ah true, missed the point that it requires `tedge.toml`, so should be `/etc/tedge`.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,195 @@
+---
+title: New Log File Management Plugin
+tags: [Reference, Log Files]
+sidebar_position: 7
+draft: true
+---
+
+# Log file management plugin
+
+Thin-edge provides a command plugin to fetch log files from the device.
```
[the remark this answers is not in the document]
rina23q: Yes, _operation_ fits this context.

However, I'm still not 100% confident where I should use _operation_ and _command_. For example,
* The plugin receives a commnad (I think this)? an operation?
* `successful` is a status of command? operation?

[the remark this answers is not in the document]
didier-wenzek: > However, I'm still not 100% confident where I should use _operation_ and _command_. For example,

(From my point of view):
- an operation is akin to a class / a category / a template of what can be done
- a command is an instance / a specific request on a target with specific parameters
 
>     * The plugin receives a commnad (I think this)? an operation?

The plugin receives a *command*, yes. This a request to do something now will all the parameters to do it.

>     * `successful` is a status of command? operation?

`successful` is a status of a command. The next one might fail.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,195 @@
...
+```sh te2mqtt
+tedge mqtt sub 'te/device/main///cmd/log_upload/+'
+```
+
+A new log file upload command with the ID "1234" is published by another component as below.
+
+```sh te2mqtt
+tedge mqtt pub -r 'te/device/main///cmd/log_upload/1234' '{
+  "status": "init",
+  "uploadUrl": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
```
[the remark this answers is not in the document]
rina23q: Right. This needs to be considered on the requester side for any upload operations.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+```sh te2mqtt
+tedge mqtt pub -r 'te/device/main///cmd/log_upload/1234' '{
+  "status": "failed",
+  "reason": "The target log file for 'mosquitto' does not exist.",
+  "uploadUrl": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
+  "logType": "mosquitto",
+  "dateFrom": "2013-06-22T17:03:14.000+02:00",
+  "dateTo": "2013-06-22T18:03:14.000+02:00",
+  "searchText": "ERROR",
+  "maximumLines": 1000
```
albinsuresh, line 120: Although I understand the need to keep these fields while an operation is transitioning between intermediate states like `executing` to `uploading` to `uploaded`, so that each state handler has the complete context, I'd at least expect these fields to be optional for the terminal states like `successful` and `failed`. Else, it may not be very user-friendly.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+```
+
+### Flow
+
+```mermaid
+sequenceDiagram
+  participant Mapper/others
+  participant Plugin
+
+  Mapper/others->>Plugin: Cloud-agnostic Log file upload command (Status: init)
```
albinsuresh, line 131:
```suggestion
  Mapper/others->>Plugin: tedge log_upload command (Status: init)
```

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+  Plugin->>Mapper/others: Status: executing
+  alt No error
+    Plugin->>Plugin: Extract log
+    Plugin->>Mapper/others: File transfer [HTTP]
+    Mapper/others-->>Plugin: Success status code [HTTP]
```
albinsuresh, line 136:
```suggestion
    Plugin->>Tedge Agent: File upload [HTTP]
    Tedge Agent-->>Plugin: Status OK [HTTP]
```

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+  "maximumLines": 1000
+}'
+```
+
+### Flow
+
+```mermaid
+sequenceDiagram
+  participant Mapper/others
+  participant Plugin
```
albinsuresh, line 129:
```suggestion
  participant Plugin
  participant Tedge Agent
```

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+files = [
+  { type = "mosquitto", path = '/var/log/mosquitto/mosquitto.log' },
+  { type = "software-management", path = '/var/log/tedge/agent/software-*' },
+  { type = "c8y_CustomOperation", path = '/var/log/tedge/agent/c8y_CustomOperation/*' }
+]
+```
+
+The `tedge-log-plugin` parses this configuration file on startup for all the `type` values specified,
+and sends the supported log types message to the MQTT local broker on the `<root>/<identifier>/cmd/log_upload` topic with a retained flag.
+The `<root>/<identifier>` is defined in `tedge.toml` file as `device.topic`.
```
albinsuresh, line 50: I'd have kept separate tedge config keys for the `root` prefix (te/) and the `device identifier` part(device/<device-id>).

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+
+Given that `device.topic` is set to `te/device/main//`, the message to declare the supported log types is as follows.
```
albinsuresh, line 52:
```suggestion
Given that `device.topic` is set to `te/device/main//` for the main device, the message to declare the supported log types is as follows.
```

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
+---
+title: New Log File Management Plugin
+tags: [Reference, Log Files]
```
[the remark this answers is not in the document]
didier-wenzek: These are working fine: https://thin-edge.github.io/thin-edge.io/tags/log-files

docs/src/references/tedge-log-management.md
rina23q: It's up to a cloud mapper how it converts the `successful`/`failed` message and the converted outcome can be more than one messages (e.g. c8y-mapper needs to update the software list after software update operation). In my opinion, having all information in the payload and let the mapper ignore uninterested fields is more user-friendly.

rina23q: Ah I think I get your point. In short, I also think no other fields should be included for `successful`/`failed`(the final states). 
However, the current proposal here is not fully synced with the state machine proposal. In this proposal's `successful` is actually `uploaded`, but I want to keep states the simplest until we finished to define state machine. So, your point will make sense.

docs/src/references/tedge-log-management.md
rina23q: Divided to `device.topic.root` and `device.topic.id`. If you have a better proposal of naming, please comment here.

docs/src/references/tedge-log-management.md
didier-wenzek: > Although I understand the need to keep these fields while an operation is transitioning between intermediate states like `executing` to `uploading` to `uploaded`, so that each state handler has the complete context, I'd at least expect these fields to be optional for the terminal states like `successful` and `failed`. Else, it may not be very user-friendly.

Why not user friendly? Because too verbose?

I see these fields as useful for many small reasons: generating events and alarms for instance. It can also useful for audit purposes: an audit plugin could be implemented to record all the init and final states.

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+```sh te2mqtt
+tedge mqtt sub 'te/device/main///cmd/log_upload/+'
+```
+
+A new log file upload command with the ID "1234" is published by another component as below.
+
+```sh te2mqtt
+tedge mqtt pub -r 'te/device/main///cmd/log_upload/1234' '{
+  "status": "init",
+  "uploadUrl": "http://127.0.0.1:8000/tedge/file-transfer/main/log_upload/mosquitto-1234",
```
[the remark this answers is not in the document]
rina23q: I may want to revert this change and even want to change the field names of config/firmware management. Just `url` is confusing for other operations. For downloading, we need to keep two URLs in a message payload. 
1. The original URL, where you download the file
2. File transfer repository URL
We need to find good names for these. 
For 1), `originalUrl`, `serverUrl`, ...
For 2), `internalUrl`, `downloadUrl`, ...

docs/src/references/tedge-log-management.md
reubenmiller: Having two independent settings makes it easier to change the `root` prefix without knowing anything about the `device identifier`.

For example, if there is only one setting and the default value is `te/device/main///`, if you want to just change the prefix, then the user needs to check what the default value is first, then modify the value to replace `te` with `mycompany`, e.g. `mycompany/device/main///`.

Where as there are two independent settings, then it allows a user to change just the prefix using the following config (without having to know anything about the identifier):

```toml
device.topic.root = "mycompany"
```

didier-wenzek: > Having two independent settings makes it easier to change the `root` prefix without knowing anything about the `device identifier`.

So let's go with 2 settings

docs/src/references/tedge-log-management.md
albinsuresh: > Why not user friendly? Because too verbose?

During the dev phase, I've always liked the simplicity of the SmartREST messages that lets you finish a dangling operation by simply sending a `503` or `504` message. Even while writing simple clients, it comes in handy. So, even for tedge, I'd appreciate the ability to finish an operation with minimal inputs like `{ "status": "successful" }`.

Of course there isn't much value in production as we'd probably recommend devs to always republish the input fields for all state transitions. That's why I was suggesting to make them optional rather than cutting them out completely.

docs/src/references/tedge-log-management.md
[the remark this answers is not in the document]
rina23q: For the firmware feature with c8y, c8y-mapper needs to update the installed firmware info including the original url in the end of operation.

 `cloudUrl` and `tedgeUrl` sounds good, although the location doesn't have to be in cloud platforms... maybe some random repository in the internet.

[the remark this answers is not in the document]
albinsuresh: Most, if not all, of the internet is in the cloud these days, right? ;)

But, to account even for instances like ThickEdge as well, may be something "cloud agnostic" is more appropriate. How about just `remoteUrl`?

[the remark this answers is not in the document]
rina23q: > Most, if not all, of the internet is in the cloud these days, right? ;)
Haha love to see this :D

I like `remoteUrl`! I'll use it!

docs/src/references/tedge-log-management.md
```diff
@@ -0,0 +1,180 @@
...
+    -h, --help
+            Print help information
+
+    -V, --version
+            Print version information
+
+The thin-edge `CONFIG_DIR` is used:
+  * to find where to connect the MQTT bus: `tedge config get mqtt.bind.address` and `tedge config get mqtt.bind.port`
+  * to find where to publish/subscribe: `tedge config get root.topic` and `tedge config get device.topic`
+  * to store tedge-log-plugin.toml: the configuration file that specifies which logs to be retrieved
```
albinsuresh, line 173:
```suggestion
The thin-edge `CONFIG_DIR` is used:
  * to find the `tedge.toml` where the following configs are defined:
     ** `tedge config get mqtt.bind.address` and `tedge config get mqtt.bind.port` to connect to the tedge MQTT broker
     ** `tedge config get root.topic` and `tedge config get device.topic`: for the MQTT topics to publish to and subscribe from
  * to find/store the `tedge-log-plugin.toml`: the configuration file that specifies which logs to be retrieved
```

albinsuresh (approved):

PR #2076, opened 2023-07-18
Remove hardcoded port in test_mqtt_server.rs

Bravo555:
## Proposed changes

This removes hard-coded port for the test MQTT broker, allowing the tests to be run concurrently and removes some flakiness. The difference can be observed by running `cargo nextest run -p mqtt_channel` - without the changes some or most tests failed, with the changes, all test finish successfully (at least on my machine).

Unfortunately neither the older `rumqttd` version we use, nor the latest version (that additionally drops QoS 2 support) provides a way to query the broker port after it started, so the workaround with binding a TCP listener just to drop it, has to be used for now.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/tests/mqtt_tests/src/test_mqtt_server.rs
+++ b/crates/tests/mqtt_tests/src/test_mqtt_server.rs
@@ -1,7 +1,5 @@
 use std::collections::HashMap;
-use std::net::Ipv4Addr;
-use std::net::SocketAddr;
-use std::net::SocketAddrV4;
+use std::net::TcpListener;
 use std::time::Duration;
 
 use futures::channel::mpsc::UnboundedReceiver;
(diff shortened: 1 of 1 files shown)


didier-wenzek (approved):
Approved. This works fine. Thank you.

> Unfortunately neither the older rumqttd version we use, nor the latest version (that additionally drops QoS 2 support) provides a way to query the broker port after it started, so the workaround with binding a TCP listener just to drop it, has to be used for now.

This is really unfortunate. Do you think it would be complicated to fix it and to push a PR on rumqttd? This can be used as a test of their willingness to improve rumqttd. If not it might the time for us to consider another alternative.

Bravo555:
> Unfortunately neither the older `rumqttd` version we use, nor the latest version (that additionally drops QoS 2 support)

So looks like this is no longer true as QoS 2 support was added in `rumqttd 0.16` on 24th of July.

Issue #2019, opened 2023-06-09
Refactor the internal config code to use new TedgeConfig


reubenmiller:
**Is your refactoring request related to a problem? Please describe.**

A new `TedgeConfig` internal feature was introduced in https://github.com/thin-edge/thin-edge.io/pull/1936. This should be utilized by all components and the older configuration model should be removed.

**Describe the solution you'd like**

* [x] Refactor existing code base to make use of `TedgeConfig`

  * [x] `tedge-mapper` https://github.com/thin-edge/thin-edge.io/pull/2013
  * [x] `tedge-agent` https://github.com/thin-edge/thin-edge.io/pull/2073
  * [x] `tedge-firmware-plugin` https://github.com/thin-edge/thin-edge.io/pull/2064
  * [x] `tedge-configuration-plugin` https://github.com/thin-edge/thin-edge.io/pull/2064
  * [x] `tedge-log-plugin` https://github.com/thin-edge/thin-edge.io/pull/2064
  * [x] `tedge` https://github.com/thin-edge/thin-edge.io/pull/2077

* [x] Remove the older configuration logic

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
@didier-wenzek Do we need to complete the last task in the list, or is this outdated information?

didier-wenzek:
> @didier-wenzek Do we need to complete the last task in the list before closing the ticket, or is this outdated information?

This outdated. I forgot to update it after https://github.com/thin-edge/thin-edge.io/pull/2143

PR #2077, opened 2023-07-19
Refactor tedge to use new config

PradeepKiruvale:
## Proposed changes

`tedge` must use the new tedge config APIs to parse the configurations. 

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2019

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/new.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/new.rs
@@ -83,6 +83,25 @@ impl TEdgeConfig {
 
         Ok(mqtt_config)
     }
+
+    pub fn mqtt_client_auth_config(&self) -> MqttAuthConfig {
+        let mut client_auth = MqttAuthConfig {
+            ca_dir: self.mqtt.client.auth.ca_dir.or_none().cloned(),
+            ca_file: self.mqtt.client.auth.ca_file.or_none().cloned(),
+            client: None,
+        };
+        // Both these options have to either be set or not set
+        if let Ok(Some((client_cert, client_key))) = all_or_nothing((
+            self.mqtt.client.auth.cert_file.as_ref(),
+            self.mqtt.client.auth.key_file.as_ref(),
+        )) {
+            client_auth.client = Some(MqttAuthClientConfig {
+                cert_file: client_cert.clone(),
+                key_file: client_key.clone(),
+            })
+        }
+        client_auth
+    }
 }
 
 #[derive(serde::Deserialize, serde::Serialize, Clone, Copy, PartialEq, Eq, Debug)]
@@ -691,6 +710,19 @@ where
     fn call(self, data: &T, location: &TEdgeConfigLocation) -> Self::Output {
         (self)(data, location)
     }
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct MqttAuthConfig {
+    pub ca_dir: Option<Utf8PathBuf>,
+    pub ca_file: Option<Utf8PathBuf>,
+    pub client: Option<MqttAuthClientConfig>,
+}
+
+#[derive(Debug, Clone, Default)]
+pub struct MqttAuthClientConfig {
+    pub cert_file: Utf8PathBuf,
+    pub key_file: Utf8PathBuf,
 }
 
 #[cfg(test)]
--- a/crates/core/tedge/src/cli/connect/command.rs
+++ b/crates/core/tedge/src/cli/connect/command.rs
@@ -12,6 +12,7 @@ use rumqttc::QoS::AtLeastOnce;
 use std::path::Path;
 use std::sync::Arc;
 use std::time::Duration;
+use tedge_config::new::TEdgeConfig;
 use tedge_config::system_services::*;
 use tedge_config::*;
 use tedge_utils::paths::create_directories;
@@ -53,7 +54,7 @@ impl Command for ConnectCommand {
     }
 
     fn execute(&self) -> anyhow::Result<()> {
-        let config = self.config_repository.load()?;
+        let config = self.config_repository.load_new()?;
         if self.is_test_connection {
             let br_config = self.bridge_config(&config)?;
             if self.check_if_bridge_exists(&br_config) {
@@ -84,29 +85,33 @@ impl Command for ConnectCommand {
             .common_mosquitto_config
             .clone()
             .with_internal_opts(
-                config.query(MqttPortSetting)?.into(),
-                config.query(MqttBindAddressSetting)?.to_string(),
+                config.mqtt.bind.port.into(),
+                config.mqtt.bind.address.to_string(),
             )
             .with_external_opts(
-                config.query(MqttExternalPortSetting).ok().map(|x| x.into()),
+                config.mqtt.external.bind.port.or_none().cloned(),
                 config
-                    .query(MqttExternalBindAddressSetting)
-                    .ok()
-                    .map(|x| x.to_string()),
-                config.query(MqttExternalBindInterfaceSetting).ok(),
-                config.query(MqttExternalCAPathSetting).ok(),
-                config.query(MqttExternalCertfileSetting).ok(),
-                config.query(MqttExternalKeyfileSetting).ok(),
+                    .mqtt
+                    .external
+                    .bind
+                    .address
+                    .or_none()
+                    .cloned()
+                    .map(|a| a.to_string()),
+                config.mqtt.external.bind.interface.or_none().cloned(),
+                config.mqtt.external.ca_path.or_none().cloned(),
+                config.mqtt.external.cert_file.or_none().cloned(),
+                config.mqtt.external.key_file.or_none().cloned(),
             );
 
-        let device_type = config.query(DeviceTypeSetting)?;
+        let device_type = &config.device.ty;
 
         new_bridge(
             &bridge_config,
             &updated_mosquitto_config,
             self.service_manager.as_ref(),
             &self.config_location,
-            &device_type,
+            device_type,
         )?;
 
         match self.check_connection(&config) {
@@ -136,7 +141,13 @@ impl Command for ConnectCommand {
         if let Cloud::C8y = self.cloud {
             check_connected_c8y_tenant_as_configured(
                 &config,
-                &config.query_string(C8yMqttSetting)?,
+                &config
+                    .c8y
+                    .mqtt
+                    .or_none()
+                    .cloned()
+                    .map(|u| u.to_string())
+                    .unwrap_or_default(),
             );
             enable_software_management(&bridge_config, self.service_manager.as_ref());
         }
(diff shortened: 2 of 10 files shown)


crates/common/tedge_config/src/tedge_config_cli/new.rs
```diff
@@ -433,7 +433,6 @@ define_tedge_config! {
             #[doku(as = "u16")]
             port: NonZeroU16,
 
-            #[tedge_config(reader(private))]
```
PradeepKiruvale, line 436: To access the `auth.cert and key files` had to remove this private.

Bravo555 (changes requested):

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -84,22 +85,34 @@ impl Command for ConnectCommand {
...
-                    .map(|x| x.to_string()),
-                config.query(MqttExternalBindInterfaceSetting).ok(),
-                config.query(MqttExternalCAPathSetting).ok(),
-                config.query(MqttExternalCertfileSetting).ok(),
-                config.query(MqttExternalKeyfileSetting).ok(),
+                    .mqtt
+                    .external
+                    .bind
+                    .address
+                    .clone()
```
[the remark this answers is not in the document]
PradeepKiruvale: Yup, done. I was expecting `clippy`  to catch them ;)

crates/common/tedge_config/src/tedge_config_cli/new.rs
didier-wenzek: Before removing this, I would ask @jarhodes314 about why he added this (taking the time to implement this macro only for this `auth` field and the c8y url). 

If this reader has to be kept private, a solution is to add methods in `new::TEdgeConfig` as this has been done to [get the `c8y_url`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/common/tedge_config/src/tedge_config_cli/new.rs#L54)

jarhodes314: The goal here was to stop people accessing these fields directly because `auth.key_file` and `auth.cert_file` must either both be set, or neither of them should be set. The idea here is that whenever we access the two fields, we need to return an error if only one of them is set. If you look at https://github.com/thin-edge/thin-edge.io/blob/b044aaebf0f643e353ce32e62518236e5de09b25/crates/common/tedge_config/src/tedge_config_cli/new.rs#L77-L80, for example, you'll see we use `all_or_nothing` to access the fields. I think the appropriate solution in this case is to create a method on TedgeConfigMqttReader (or whatever is called, I'd look at the `cargo doc` output to figure that out) called something like `cert_key_files()` that returns the result of calling all_or_nothing.

Once you've done that, you can probably modify https://github.com/thin-edge/thin-edge.io/pull/2077/files#diff-e932087e82e2186b8e86bcbdc49124d1e279770b5f055c33d69f832ff4f1101dR50-R55 to use `if let Some((cert_file, key_file)) = config.mqtt.cert_key_files() { ... }` rather than needing to unwrap in the where the cert and key and some.

PradeepKiruvale: Thanks for the clarification @jarhodes314

crates/core/tedge/src/cli/mqtt/cli.rs
```diff
@@ -44,41 +43,16 @@ pub enum TEdgeMqttCli {
...
+        let client_auth_config = if client_cert.is_none() && client_private_key.is_none() {
             None
         } else {
-            let client_cert = client_cert?;
-            let client_private_key = client_private_key?;
             Some(ClientAuthConfig {
-                cert_file: client_cert,
-                key_file: client_private_key,
+                cert_file: client_cert.unwrap_or_default(),
+                key_file: client_private_key.unwrap_or_default(),
```
rina23q, line 55: It's kinda related to what @jarhodes314 explained why `auth` is private.

The code has a problem. If `client_cert` is `Some` and `client_private_key` is `None`, it enters this `else` conditional branch. Actually, it should have been already an error for that case. But this code allows to have such config.
```rust
client_auth_config = Some(ClientAuthConfig {
    cert_file: <some value>,
    client_private_key: <default of "Utf8PathBuf">
})
```

In short, you need to address the pair of `client_cert` and `client_private_key` as James suggested.

rina23q (commented):
`auth` must be correctly addressed as @jarhodes314 suggested.

crates/core/tedge/src/error.rs
```diff
@@ -1,3 +1,5 @@
+use tedge_config::new::ConfigNotSet;
+
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/core/tedge/src/error.rs
```diff
@@ -27,4 +29,7 @@ pub enum TEdgeError {
 
     #[error(transparent)]
     FromTEdgeConfigRead(#[from] tedge_config::new::ReadError),
+
+    #[error(transparent)]
+    FromConfigNotSet(#[from] ConfigNotSet),
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/core/tedge/src/error.rs
```diff
@@ -1,3 +1,5 @@
+use tedge_config::new::ConfigNotSet;
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/core/tedge/src/cli/mqtt/cli.rs
PradeepKiruvale: fixed

crates/common/tedge_config/src/tedge_config_cli/new.rs
jarhodes314: The updated code looks much better.

crates/common/tedge_config/src/tedge_config_cli/new.rs
```diff
@@ -83,6 +84,28 @@ impl TEdgeConfig {
 
         Ok(mqtt_config)
     }
+
+    pub fn mqtt_client_auth_configs(
```
Bravo555, line 88: Under which conditions this function returns an error is not very clear.

1. It looks to me like the function doesn't throw errors in any case (there's no `?` or `return Err(..)`), so the function should just return a thing instead of result.
2. There is a comment (that I remember having written) saying that if either option is not set, we return an error. Right now, it looks like we don't return an error, so this is either a regression or the comment should be removed.
3. As more of a general note, if a function is public, and even more so if it returns a result, it should be documented, at least to tell when it's supposed to fail.

crates/core/tedge/src/cli/mqtt/cli.rs
```diff
@@ -44,43 +43,18 @@ pub enum TEdgeMqttCli {
...
-        };
+        let client_auth_config =
+            if auth_config.cert_file.is_none() && auth_config.key_file.is_none() {
+                None
+            } else {
+                Some(ClientAuthConfig {
+                    cert_file: auth_config.cert_file.clone().unwrap_or_default(),
+                    key_file: auth_config.key_file.unwrap_or_default(),
+                })
+            };
```
Bravo555, line 57: If a function `mqtt_client_auth_configs` got created in order to put this "if either client key or cert are not set, then both are not set" logic there, then couldn't we also move this bit there? Because it seems to me like we need to do a fairly simple transformation but go in circles instead.

What we essentially have is: `(ca_dir, ca_file, cert_file, key_file)`, all of them are optional. Then the logic is:

1. If either `ca_dir` or `ca_file` is set, we should use TLS to connect. The code that we call later is aware of this convention, so we should just pass these fields as they are
2. If either `cert_file` or `key_file` is not set while the other is set, set both to `None` and don't use client authentication (or if we want to be very correct, either print a warning or return an error)

I think all this behaviour could be encompassed by a single function, and the above could even be its documentation. The output could be descibed by a struct like this:

```
MqttAuthConfig {
  ca_dir: Utf8PathBuf,
  ca_file: Utf8PathBuf,
  client: MqttAuthClientConfig {
    cert_file: Utf8PathBuf,
    key_file: Utf8PathBuf\
  }
}
```

and the function signature could be `(ca_dir, ca_file, cert_file, key_file) -> MqttAuthConfig` (in this signature, I don't mean this 4-tuple to be a literal tuple, but just these 4 properties we get from tedge_config)

crates/core/tedge/src/error.rs
```diff
@@ -27,4 +27,10 @@ pub enum TEdgeError {
 
     #[error(transparent)]
     FromTEdgeConfigRead(#[from] tedge_config::new::ReadError),
+
+    #[error(transparent)]
+    FromConfigNotSet(#[from] tedge_config::new::ConfigNotSet),
+
+    #[error(transparent)]
+    CertificatesNotSet(#[from] tedge_config::CertificateError),
```
Bravo555, line 35: I don't see this error variant used anywhere, so it can probably be deleted.

Bravo555 (changes requested):

crates/core/tedge/src/error.rs
PradeepKiruvale: This error is used here     `config.mqtt_client_auth_configs()?`

Bravo555: Ah right, you're correct in that this variant would be constructed using the `From` conversion, had `mqtt_client_auth_configs()` returned an `Err(CertificateError)`, but looking at the function body, this function does not return an error in any branch, so as it is now, the `Result` there is useless, making this variant here useless as well.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -109,6 +110,14 @@ struct ClientAuthConfig {
     key: PrivateKey,
 }
 
+#[derive(Debug, Clone, Default)]
+pub struct MqttClientAuthConfigs {
```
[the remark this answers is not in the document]
PradeepKiruvale: Refactored it, now there is only one struct.

crates/common/tedge_config/src/tedge_config_cli/new.rs
PradeepKiruvale: Yes, the error is useless., removed it.

crates/core/tedge/src/cli/mqtt/cli.rs
PradeepKiruvale: Refactored it below

```
#[derive(Debug, Clone, Default)]
pub struct MqttAuthConfig {
    pub ca_dir: Option<Utf8PathBuf>,
    pub ca_file: Option<Utf8PathBuf>,
    pub client: Option<MqttAuthClientConfig>,
}

#[derive(Debug, Clone, Default)]
pub struct MqttAuthClientConfig {
    pub cert_file: Utf8PathBuf,
    pub key_file: Utf8PathBuf,
}
```

But I did not get what function you are talking about here. Because the `tedge config` provides these certificates, I populated these structs inside the `mqtt_client_auth_config` function, after some `validations` for the `MqttAuthClientConfig`

crates/core/tedge/src/error.rs
PradeepKiruvale: Yes, removed.

crates/core/tedge/src/cli/mqtt/cli.rs
Bravo555: `mqtt_client_auth_config()` function as implemented, does exactly what i wanted, just reading properties from the config and putting them into this struct so the later code doesn't have to run this logic again.

rina23q (approved):
Looks good!

Issue #2113, opened 2023-07-29
tedge-apt-plugin does not remove sm-plugin symlink when being uninstalled


reubenmiller:
**Describe the bug**

There is an error in the **tedge-apt-plugin** Debian `postrm` maintainer script which results in the symlink `/etc/tedge/sm-plugins/apt` not being removed when the package is uninstalled.

Introduced in https://github.com/thin-edge/thin-edge.io/pull/2027

**To Reproduce**

1. Install tedge-apt-plugin (e.g. example shows using the tedge debian repo)

    ```sh
    sudo apt-get install -y tedge-apt-plugin
    ```

2. Check that the `apt` symlink exists to the sm-plugins folder

    ```sh
    ls -l /etc/tedge/sm-plugins/
    ```
    
    You should see the `apt` in the list of files

3. Uninstall tedge-apt-plugin

    ```sh
    sudo apt-get remove tedge-apt-plugin
    ```

4. ERROR step: The symlink still exists (when it should have been removed)

    ```sh
    ls -l /etc/tedge/sm-plugins/
    ```

    The expected behaviour is that the `apt` file should have been removed from `/etc/tedge/sm-plugins`.

**Expected behavior**

The `/etc/tedge/sm-plugins/apt` symlink should be removed when uninstalling the tedge-apt-plugin.

**Screenshots**

Full log showing the above steps when using tedge 0.12.0.

*Step 1: Install tedge-apt-plugin*
```text
$ sudo apt-get install tedge-apt-plugin

Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following NEW packages will be installed:
  tedge-apt-plugin
0 upgraded, 1 newly installed, 0 to remove and 17 not upgraded.
Need to get 0 B/456 kB of archives.
After this operation, 1381 kB of additional disk space will be used.
apt-listchanges: Can't set locale; make sure $LC_* and $LANG are correct!
perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = (unset),
	LC_ALL = "en_US.UTF-8",
	LC_TERMINAL = "iTerm2",
	LANG = "en_GB.UTF-8"
    are supported and installed on your system.
perl: warning: Falling back to a fallback locale ("en_GB.UTF-8").
locale: Cannot set LC_CTYPE to default locale: No such file or directory
locale: Cannot set LC_MESSAGES to default locale: No such file or directory
[...]

PR #2116, opened 2023-07-31
fix(tedge-apt-plugin): remove symlink using rm -f instead of using conditional logic

reubenmiller:
## Proposed changes

Simplify the post remove maintainer script for `tedge-apt-plugin` to always remove the symlink.

Use `rm -f` which does not fail if the file does not exist (as this also avoid race conditions between the test and action).

The previous logic also was erroneous (as indicated by shellcheck rule [SC2107](https://www.shellcheck.net/wiki/SC2107)).

image

## Types of changes

## Paste Link to the issue
<br/>

[#2113](https://github.com/thin-edge/thin-edge.io/issues/2113)

## Checklist

### DIFF
--- a/configuration/debian/tedge-apt-plugin/postrm
+++ b/configuration/debian/tedge-apt-plugin/postrm
@@ -2,7 +2,4 @@
 
 set -e
 
-if [ ! -e /etc/tedge/sm-plugins/apt && ! -L /etc/tedge/sm-plugins/apt ];
-then
-   rm /etc/tedge/sm-plugins/apt
-fi
+rm -f /etc/tedge/sm-plugins/apt


configuration/debian/tedge-apt-plugin/postrm
```diff
@@ -2,7 +2,4 @@
 
 set -e
 
-if [ ! -e /etc/tedge/sm-plugins/apt && ! -L /etc/tedge/sm-plugins/apt ];
-then
-   rm /etc/tedge/sm-plugins/apt
-fi
+rm -f /etc/tedge/sm-plugins/apt
```
PradeepKiruvale, line 5: I added those checks to make sure that the link exists. If we try to remove it without checking, doesn't the command fail?

reubenmiller: I put a comment about this in the PR description ;)

> Use rm -f which does not fail if the file does not exist (as this also avoid race conditions between the test and action).

reubenmiller: Short answer, no `rm -f` does not fail if the file does not exist due to the `-f` option.

https://linux.die.net/man/1/rm

> -f, --force
>     ignore nonexistent files, never prompt

rina23q (approved):
Nice improvement. Actually in the previous logic, if non symbolic link (e.g. binary) exists as `rm /etc/tedge/sm-plugins/apt` by chance, the `postrm` didn't remove it. I think it was useless to consider if the target is a link or something else. So just remove :D

configuration/debian/tedge-apt-plugin/postrm
PradeepKiruvale: I got it, thanks!

PradeepKiruvale (approved):

Issue #2082, opened 2023-07-19
Update the mappers to consume MEA from the new topics


didier-wenzek:
With the [new MQTT topic scheme](https://github.com/thin-edge/thin-edge.io/pull/2030), the mappers will have to consume messages from the `te/# ` and no more the `tedge/#`.

Once implemented the [automatic conversion from the old to new topics](https://github.com/thin-edge/thin-edge.io/issues/2079), the c8y, az and aws mapper must be updated accordingly.
- This has to done for measurements, events and alarms.
- The payload of these messages is unchanged with an exception for the severity of an alarm (previously attached on the topic, now inside the message payload) 
- See [this issue for the details](https://github.com/thin-edge/thin-edge.io/issues/2079).
- For c8y, the auto-registration process must be adapted too. Hence, this task might depend on the [implementation of an inventory / digital twin representation](https://github.com/thin-edge/thin-edge.io/issues/2081).
  - As first step, this task can address only auto registration using the `te/device/+/service/+` scheme.
  - The inventory / digital twin representation can be used as a second step when available.
 
Tasks:

- [x] Measurements for c8y are consumed from `te/+/+/+/+/m/+`
- [x] Measurements are published on `te/+/+/+/+/m/+` by collectd mapper
- [x] #2215
- [x] #2214
- [x] #2243
- [x] https://github.com/thin-edge/thin-edge.io/pull/2304
- [x] #2244
- [x] #2274
- [x] #2351

To be done only later:
- [x] #2288
- [x] [Software List&Update operations  on child devices](https://github.com/thin-edge/thin-edge.io/pull/2307)

PradeepKiruvale:
`c8y-mapper` specific topics info is here https://github.com/thin-edge/thin-edge.io/pull/2030/files?short_path=9695a2a#diff-9695a2a09b64f1cfa0ed7f9e9ff632abdc04a7fd3dc4c3849cf5b53a461e4add

didier-wenzek:
Tasks:

- [x] Measurements for c8y are consumed from `te/+/+/+/+/m/+`
- [x] Measurements are published on `te/+/+/+/+/m/+` by collectd mapper
- [ ] Forward measurement type to c8y.
- [ ] Events for c8y  are consumed from `te/+/+/+/+/e/+`
- [ ] Alarms for c8y  are consumed from `te/+/+/+/+/a/+`
- [ ] Azure mapper consumes MEA from `te` topics.
[...]

Issue #2079, opened 2023-07-19
Implement a converter that bridges the old MEA topics to the new ones.


didier-wenzek:
As we plan a [major refactoring of all the MQTT topics](https://github.com/thin-edge/thin-edge.io/pull/2030) used by thin-edge, we have to provide some backward compatibilities. During a transition period, thin-edge will support the old as well as the new schemes, using __a converter that re-publishes any measurement, event or alarm received on a `tedge/#` topic to the appropriate new `te/#` topic__.

- All the messages must be re-published unchanged - same content, QoS and retain flag, from the old to the new topic.
- There is only one exception for the severity of the alarms (that has to be extracted from the `tedge` topic and pushed into the json payload of the message).
- The complete specifications of this conversion are [here](https://github.com/thin-edge/thin-edge.io/pull/2030/files#diff-af802a095f72bbf84e172455307666c297802f0376506349420c5c2576aeaa88R658).
- The simpler is to use a [`Converter`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_actors/src/converter.rs#L78) and to launch this actor inside the `tedge-agent`.

PradeepKiruvale:
The table below shows mapping the `tedge` topic to the `te` topic and if there are any changes needed.

| tedge topic | te topic | other updates
|---|---|---|
|tedge/measurements| te/device/main///m/ ||
| tedge/measurements/child | te/device/child///m/ ||
| tedge/alarms/severity/alarm_type|te/device/main///a/alarm_type|add severity into payload|
|tedge/alarms/severity/child/alarm_type|te/device/child///a/alarm_type|add severity into payload|
|tedge/events/event_type|te/device/main///e/event_type||
|tedge/events/child/event_type|te/device/child///e/event_type||
|tedge/commands/res/software/list|te/device/main/cmd/software_list/<cmd_id>||
|tedge/commands/res/software/update|te/device/main/cmd/software_update/<cmd_id>||
|tedge/+/commands/res/config_snapshot|te/device/main/cmd/config_snapshot/<cmd_id>||
|tedge/+/commands/res/config_snapshot|te/device/child/cmd/config_snapshot/<cmd_id>||
|tedge/+/commands/res/config_update|te/device/main/cmd/config_update/<cmd_id>||
|tedge/+/commands/res/config_update|te/device/child/cmd/config_update/<cmd_id>||
[...]

PR #2118, opened 2023-08-01
cumulocity mapper using new mqtt topics for telemetry

PradeepKiruvale:
## Proposed changes

This PR captures the code changes include

-  `Cumulocity mapper` to use the `new mqtt topics` for the `main` & `child` device `telemetry`
-  Updating the tests
-   Integration tests to test the `telemetry` with new topics

What is not addressed in this PR

- `Main & child` device Service telemetry
-  Measurement metadata 

I have introduced a new `code` path for processing the measurement, events, and alarms using the new MQTT topics. This segregation will help us easily maintain the code and also, it will be much easier to remove the earlier version of processing telemetry whenever we want to remove it. I agree that there is a bit of a code redundancy but this will help us to maintain the code easily.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2082

## Checklist


crates/core/tedge_api/src/device_id.rs
```diff
@@ -0,0 +1,36 @@
+pub fn new_get_child_id_from_topic(parent_device_name: String, topic: String) -> Option<String> {
+    if topic.starts_with("te/device") {
```
reubenmiller, line 2: We should make the root prefix (e.g. `te`) configurable. Though since we have multiple PRs going on at the moment, we can just put a `TODO` command above the prefix so that we know we need to change it (or alternatively, create a ticket for the follow up work).

reubenmiller: Though at the minimum we should at least create a constant for the root prefix until we are at a stage that we can read it from the tedge configuration.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -145,6 +150,47 @@ impl ThinEdgeAlarm {
...
+        } else {
+            new_get_child_id_from_topic(parent_device_name, mqtt_topic.into()).unwrap_or_default()
+        };
+
+        let severity = match alarm_data.clone() {
+            Some(data) => match data.severity {
+                Some(alarm_severity) => alarm_severity,
+                None => return Err(ThinEdgeJsonDeserializerError::AlarmSeverityNotFound),
+            },
+            None => return Err(ThinEdgeJsonDeserializerError::AlarmSeverityNotFound),
```
reubenmiller, line 184: Using a default severity will be more convenient for users. I would go with using a severity of `major` if it is not provided by the user.

We could even make this configurable later on.

crates/core/tedge_api/src/device_id.rs
```diff
@@ -0,0 +1,36 @@
+pub fn new_get_child_id_from_topic(parent_device_name: String, topic: String) -> Option<String> {
```
reubenmiller, line 1: The function name and signature need to change as it is a bit misleading.

Maybe something like:

```
get_external_identity_from_topic(device_common_name: String, topic: String)
```

crates/core/tedge_api/src/device_id.rs
```diff
@@ -0,0 +1,36 @@
...
+    if topic.starts_with("te/device") {
+        let topic_split = topic.split('/').collect::<Vec<_>>();
+        if topic_split.len() >= 5 {
+            if topic_split[2].eq("main") {
+                return None;
+            }
```
reubenmiller, line 7: There seems to be a logic error here.

With the above logic, the following two topics will return the same thing, even though they are very different (one is for the main device, and the other for a service on the main device).

```
te/device/main// => None
te/device/main/service/nodered => None
```

Also I would avoid using `None` as a reference to the device certificate's common name...it would just be better to return the `parent_device_name` variable instead (imo).

crates/core/tedge_api/src/alarm.rs
```diff
@@ -145,6 +150,47 @@ impl ThinEdgeAlarm {
             ))
         }
     }
+
+    pub fn new_try_from(
```
Bravo555, line 154: I assume this function is meant to replace `ThinEdgeAlarm::try_from` so the current name is temporary? If so, we'll need to remember to rename it when removing `ThinEdgeAlarm::try_from`, although we could perhaps use a different name than `try_from`, as usually it refers to `TryFrom::try_from`, so it can be very confusing.

crates/core/tedge_api/src/device_id.rs
reubenmiller: I would have expected a function more like this (though I'm not a Rust dev, so forgot any bad-pratices):

```rust
fn get_external_identity(topic: String, root_prefix: String, common_name: String) -> Option<String> {
    let sep = ":";
    
    let local_name = topic.split("/").take(5).filter(|&x| !x.is_empty()).collect::<Vec<&str>>();
    if local_name.len() == 0 {
        // ignore if all empty, as something is obviously wrong
        return None;
    }

    let main_topic = format!("{}/device/main", root_prefix);
    if local_name.join("/") == main_topic {
        return Some(common_name.to_string());
    }

    if local_name[0] == root_prefix {
        match local_name.len() {
            1 => {
                return Some(common_name.to_string());
            }
            _ => {
                let mut output = format!("{}{}", common_name, sep).to_string();
                if local_name.len() > 1 {
                    output.push_str(local_name.into_iter().skip(1).collect::<Vec<&str>>().join(sep).as_str());
                }
                return Some(output);
            }
            
        }
    } 
    return None;    
}
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -354,6 +467,27 @@ impl CumulocityConverter {
             }
             mqtt_messages.append(&mut messages);
             Ok(mqtt_messages)
+        } else if topic.name.starts_with("te/device/") && topic.name.contains("/a/") {
```
Bravo555, line 470: Is there anything that prevents devices or services from being named e.g. `a`? If not, then that condition is not sufficient.

Bravo555 (changes requested):
Some things to fix.

Small note: overall I feel like maybe having MQTT topic fragments be sprinkled over all these functions may not be desirable when we want to make these fragments configurable. Reading the fragments from the config may be better, but it still concerns these different functions with parsing MQTT topics.

So in effect, all the MQTT topic parsing logic is spread throughout all these different functions which makes modifications harder. That might not even be a problem, as after we migrate to the new MQTT scheme, it's supposed to be stable, but even still, maybe it'd be nice to divide MQTT messages from thin-edge messages that they encode.

Theoretically, we could parse/serialize MQTT messages as some abstract `ThinEdgeMessage` type in the type system, and functions related to alarms/events/measurements etc. would interact with them instead of MQTT messages directly.

But this could just move coupling instead of eliminating it, as then in order to add any functionality we would have to be adding new message schemas to this "MQTT message parser", and end up with something like old config system, or smartrest, or worse.

Something to think about once @didier-wenzek is back.

crates/core/tedge_api/src/topic.rs
```diff
@@ -78,6 +78,7 @@ pub fn get_child_id_from_child_topic(topic: &str) -> Option<String> {
 
 #[cfg(test)]
 mod tests {
+
```
rina23q, line 81: Remove this so that we can exclude this file from the change list.

crates/core/tedge_api/src/device_id.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use test_case::test_case;
+    #[test_case("test_device".to_string(), "te/device/main///m/".to_string(), None; "main device id")]
+    #[test_case("test_device".to_string(), "te/device/child///m/".to_string(), Some("test_device:device:child".to_string()); "child device id")]
+    #[test_case("test_device".to_string(), "te/device/child/a1/a2/m/".to_string(), Some("test_device:device:child:a1:a2".to_string()); "child with more components device id")]
+    #[test_case("test_device".to_string(), "foo/bar".to_string(), None; "wrong topic")]
```
rina23q, line 31: Let's define what the expected cases and  error cases for the target function as of now.
It needs to generate the c8y acceptable unique device name from the given topic name. However, this logic will be changed once we have the entity store. The mapper will just need to query.

My suggestion is to accept only `<root>/device/<something>/service/<something>` for the first 5 digits.

So coverage will be:
* te/device/main///m/ => test_device
* te/device/child///m/ => test_device:device:child
* te/device/main/service/nodered/m/ => test_device:device:main:service:nodered
* te/device//service/nodered/m => error
* te/device/main/service//m => error

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
```diff
@@ -99,6 +92,70 @@ impl AlarmConverter {
         Ok(output_messages)
     }
 
+    pub(crate) fn new_try_convert_alarm(
+        &mut self,
+        parent_device_name: String,
+        input_message: &Message,
+    ) -> Result<Vec<Message>, ConversionError> {
+        let mut output_messages: Vec<Message> = Vec::new();
+        let alarm_id = input_message.topic.name.split('/').last();
```
rina23q, line 101: Since `alarm_id` should have a value to proceed the following step, the error should be handled in this line. Then stop using `expect` later as it will panic.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
```diff
@@ -110,6 +167,15 @@ impl AlarmConverter {
         }
     }
 
+    pub(crate) fn new_get_c8y_alarm_topic(&self, topic: &str) -> Result<String, ConversionError> {
+        let topic_split: Vec<&str> = topic.split('/').collect();
+        if topic_split[2].eq("main") {
```
rina23q, line 172: Why 2? I thought the `topic` had something like `te/device/main///a/alarm_type`.

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -111,6 +111,12 @@ impl C8yMapperConfig {
             "tedge/events/+/+",
             "tedge/health/+",
             "tedge/health/+/+",
+            "te/device/+///m/",
+            "te/device/+///m/+",
+            "te/device/+///a/",
+            "te/device/+///a/+",
+            "te/device/+///e/",
+            "te/device/+///e/+",
```
rina23q, line 119: As I wrote in the comment in `crates/common/tedge_config/src/tedge_config_cli/new.rs`, I think it's wrong.
```suggestion
            "te/device/+/+/+/m/",
            "te/device/+/+/+/m/+",
            "te/device/+/+/+/a/+",
            "te/device/+/+/+/e/+",
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -273,6 +275,48 @@ impl CumulocityConverter {
         Ok(mqtt_messages)
     }
 
+    fn new_try_convert_measurement(
```
rina23q, line 278: What's the difference from the original `try_convert_measurement()`? Only how to get child ID?
Why not changing like this:

1. Create a new function `try_convert_measurement_with_maybe_id()` to take a `maybe_child_id` as an argument.
2. Call `try_convert_measurement_with_maybe_id()` inside `try_convert_measurement()`, so that we don't need to change anything on the function that call `try_convert_measurement()`.
3. Call `try_convert_measurement_with_maybe_id()` inside `new_try_convert_measurement()`.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -330,6 +375,74 @@ impl CumulocityConverter {
         Ok(messages)
     }
 
+    async fn new_try_convert_event(
```
rina23q, line 378: The same comment as above for measurements. What's the difference from the original function?

rina23q (commented):
Many things need to be changed...

crates/core/tedge_api/src/device_id.rs
PradeepKiruvale: I think the logic is right for the `telemetry` topics. When it comes to the `service` it might be wrong. That is a separate issue, not sure to address it here or later. IMO it's better to look into it later.

crates/core/tedge_api/src/device_id.rs
PradeepKiruvale: done

crates/core/tedge_api/src/alarm.rs
PradeepKiruvale: done

crates/core/tedge_api/src/device_id.rs
PradeepKiruvale: renamed it

crates/extensions/c8y_mapper_ext/src/config.rs
PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
PradeepKiruvale: Yes, here the index of `main` is  `2`.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
PradeepKiruvale: fixed

crates/common/tedge_config/src/tedge_config_cli/new.rs
```diff
@@ -373,7 +373,7 @@ define_tedge_config! {
 
         /// Set of MQTT topics the Cumulocity mapper should subscribe to
         #[tedge_config(example = "tedge/alarms/#,tedge/measurements/+")]
-        #[tedge_config(default(value = "tedge/measurements,tedge/measurements/+,tedge/alarms/+/+,tedge/alarms/+/+/+,tedge/events/+,tedge/events/+/+,tedge/health/+,tedge/health/+/+"))]
+        #[tedge_config(default(value = "tedge/measurements,tedge/measurements/+,tedge/alarms/+/+,tedge/alarms/+/+/+,tedge/events/+,tedge/events/+/+,tedge/health/+,tedge/health/+/+,te/device/+///m/,te/device/+///m/+,te/device/+///e/,te/device/+///e/+,te/device/+///a/,te/device/+///a/+"))]
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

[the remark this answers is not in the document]
PradeepKiruvale: This PR only addresses the `telemetry` part for the devices not for the `services`.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -145,6 +150,47 @@ impl ThinEdgeAlarm {
...
+            ));
+        }
+        let alarm_data: Option<ThinEdgeAlarmData> = if mqtt_payload.is_empty() {
+            None
+        } else {
+            Some(serde_json::from_str(mqtt_payload)?)
+        };
+
+        let external_source = if topic_split[2].eq("main") {
+            parent_device_name
```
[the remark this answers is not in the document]
PradeepKiruvale: This is needed for the complex `Alarm` messages. The complex alarm messages will be sent as the `json`. Have a look here,https://thin-edge.github.io/thin-edge.io/next/references/mappers/c8y-mapper/#alarms. The same applies to complex `event` messages as well.

crates/core/tedge_api/src/event.rs
```diff
@@ -91,6 +93,46 @@ impl ThinEdgeEvent {
...
+            };
+
+            let event_data = if mqtt_payload.is_empty() {
+                None
+            } else {
+                Some(serde_json::from_str(mqtt_payload)?)
+            };
+
+            let external_source = if topic_split[2].eq("main") {
+                parent_device_name
```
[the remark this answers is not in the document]
PradeepKiruvale: This is needed for complex `events`, Have a look here https://thin-edge.github.io/thin-edge.io/next/references/mappers/c8y-mapper/#events

crates/core/tedge_api/src/topic.rs
PradeepKiruvale: fixed

crates/core/tedge_api/src/device_id.rs
PradeepKiruvale: IMO the `telemetry` for the service can be addressed as part of another PR. As of now, thin-edge does not support the telemetry for services, isn't it?

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: My plan was to keep converting the `telemetry` with new mqtt topics completely separate. Sooner or later we need to remove the older topics, then it will be much easier.

crates/core/tedge_api/src/alarm.rs
PradeepKiruvale: Thats right, while replacing the existing ones, this needs to be renamed as `try_from`.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -145,6 +150,47 @@ impl ThinEdgeAlarm {
...
+                mqtt_topic.into(),
+            ));
+        }
+        let alarm_data: Option<ThinEdgeAlarmData> = if mqtt_payload.is_empty() {
+            None
+        } else {
+            Some(serde_json::from_str(mqtt_payload)?)
+        };
+
+        let external_source = if topic_split[2].eq("main") {
```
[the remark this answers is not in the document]
PradeepKiruvale: I just wanted to use the `function` instead of `operator` because it's more readable.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -145,6 +150,47 @@ impl ThinEdgeAlarm {
             ))
         }
     }
+
+    pub fn new_try_from(
+        parent_device_name: String,
+        mqtt_topic: &str,
+        mqtt_payload: &str,
+    ) -> Result<Self, ThinEdgeJsonDeserializerError> {
```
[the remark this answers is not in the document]
PradeepKiruvale: `parent_device_name` is used to create the `child` device external ID. I have added a comment here.

crates/core/tedge_api/src/device_id.rs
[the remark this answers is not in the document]
PradeepKiruvale: Added a doc comment, with an example.

crates/core/tedge_api/src/event.rs
```diff
@@ -91,6 +93,46 @@ impl ThinEdgeEvent {
             ))
         }
     }
+
+    pub fn new_try_from(
```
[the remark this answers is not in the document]
PradeepKiruvale: added a comment

crates/common/tedge_config/src/tedge_config_cli/new.rs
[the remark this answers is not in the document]
reubenmiller: Please implement for services as well as it is 99% the same logic. I'm ok if we ignore the auto-registration for services, but at least publishing telemetry data for everything makes more sense as the purpose of the PR is to "New topics for telemetry", so extending it to also support services would be useful as it will be touching the same function (so we should prevent double work here).

crates/core/tedge_api/src/alarm.rs
```diff
@@ -27,10 +28,11 @@ pub enum AlarmSeverity {
 }
 
 /// In-memory representation of ThinEdge JSON alarm payload
-#[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq, Clone)]
 pub struct ThinEdgeAlarmData {
     pub text: Option<String>,
 
+    pub severity: Option<String>,
```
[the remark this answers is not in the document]
PradeepKiruvale: Yes, the earlier code had the `severity` in the `ThinEdgeAlarm`, which was populated using the topic component. In the new topic scheme the `severity` will be part of the `payload`. So, I have added it to `ThinEdgeAlarmData`.  Now I refactored it and now it's just in `ThinEdgeAlarmData`.

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -156,12 +157,14 @@ impl TryFrom<ThinEdgeEvent> for C8yCreateEvent {
             Some(event_data) => {
                 text = event_data.text.unwrap_or_else(|| event_type.clone());
                 time = event_data.time.unwrap_or_else(OffsetDateTime::now_utc);
-                extras = event_data.extras;
+                extras = event_data.extras.clone();
+                if let Some(source) = event.source {
+                    if !event_data.extras.is_empty() {
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
```diff
@@ -110,6 +167,15 @@ impl AlarmConverter {
...
+            Ok(SMARTREST_PUBLISH_TOPIC.to_string())
+        } else {
+            Ok(format!("{SMARTREST_PUBLISH_TOPIC}/{}", topic_split[2]))
+        }
+    }
```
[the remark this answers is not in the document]
PradeepKiruvale: Fixed

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: That's right, nothing stops someone from naming a device or a service as `a`/`m`/`e`. Need extra check there, better we check exactly that field. Now changed it to check the exact index of the `a/m/e` substrings after the split.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -145,6 +150,47 @@ impl ThinEdgeAlarm {
...
+                None => return Err(ThinEdgeJsonDeserializerError::AlarmSeverityNotFound),
+            },
+            None => return Err(ThinEdgeJsonDeserializerError::AlarmSeverityNotFound),
+        };
+
+        Ok(Self {
+            name: alarm_name.into(),
+            severity: severity.as_str().try_into()?,
+            data: alarm_data,
+            source: Some(external_source),
```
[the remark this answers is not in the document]
PradeepKiruvale: It uses the same structure as the old API `try_from`. The `try_from` uses `None` for the `main` device. So, this has to be an `Option`.

crates/common/tedge_config/src/tedge_config_cli/new.rs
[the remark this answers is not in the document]
PradeepKiruvale: sure

didier-wenzek:
I will take the time for a proper review on Monday. Just reacting to some comments about the `new_xxx` functions.

I would really avoid to introduce such temporary code that will have to be refactored in a few weeks. This is why I proposed to start with [a mapper from the old to the new topics](https://github.com/thin-edge/thin-edge.io/issues/2079) and only then address [this work](https://github.com/thin-edge/thin-edge.io/issues/2082). Doing so the mappers can then focus on the `te/#` topics and no more subscribe to the `tedge/#` topics. i.e. no need to maintain two versions of the alarm struct.

didier-wenzek:
> Small note: overall I feel like maybe having MQTT topic fragments be sprinkled over all these functions may not be desirable when we want to make these fragments configurable. Reading the fragments from the config may be better, but it still concerns these different functions with parsing MQTT topics.

> So in effect, all the MQTT topic parsing logic is spread throughout all these different functions which makes modifications harder. That might not even be a problem, as after we migrate to the new MQTT scheme, it's supposed to be stable, but even still, maybe it'd be nice to divide MQTT messages from thin-edge messages that they encode.

> Theoretically, we could parse/serialize MQTT messages as some abstract ThinEdgeMessage type in the type system, and functions related to alarms/events/measurements etc. would interact with them instead of MQTT messages directly.

Definitely. The logic to parse and build topic along [thin-edge topic scheme](https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#topic-scheme) must be grouped in a single place. I would expect methods to:
- parse a topic extracting the root, the identifier and the channel, the latter being an enum (e.g. `Alarm("fire!")`)
- build a topic given a root, an identifier and a channel

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -111,6 +111,12 @@ impl C8yMapperConfig {
             "tedge/events/+/+",
             "tedge/health/+",
             "tedge/health/+/+",
+            "te/device/+/+/+/m/",
+            "te/device/+/+/+/m/+",
+            "te/device/+/+/+/a/",
+            "te/device/+/+/+/a/+",
+            "te/device/+/+/+/e/",
+            "te/device/+/+/+/e/+",
```
didier-wenzek, line 119: There is no points subscribing to `te/device/+/+/+/m/` and `te/device/+/+/+/m/+` as the latter includes the former. Furthermore, one should not assume the first identifier part to be the `device` constant.

```suggestion
            "te/+/+/+/+/m/+",
            "te/+/+/+/+/a/+",
            "te/+/+/+/+/e/+",
```

There is an ambiguity though. When a measurement has no type, do we have to publish it on `te/device/main///m` or `te/device/main///m/` ? I would prefer the latter to be more systematic and to avoid confusion with no vs empty measurement types.

reubenmiller: Yes if no type is provided, it should be published on `te/device/main///m/`.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -1,13 +1,14 @@
-use std::convert::TryFrom;
-use std::convert::TryInto;
-use std::fmt;
-
 use clock::Timestamp;
 use serde::Deserialize;
 use serde::Serialize;
+use std::convert::TryFrom;
+use std::convert::TryInto;
+use std::fmt;
```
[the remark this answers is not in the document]
PradeepKiruvale: done

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: Now the mapper supports only `te` topics. The `tedge to te` converter will handle messages on legacy topics and convert and republish them to corresponding `te` topics.

Issue #2081, opened 2023-07-19
Implement an inventory / digital twin representation


didier-wenzek:
With the [new MQTT topic scheme](https://github.com/thin-edge/thin-edge.io/pull/2030), the type of the target of an MQTT message can no more directly be derived from the message topic. Some auto-registration rules apply for the simple cases (when using the `te/device/+/service/+` scheme) but for the agent developer to be really free to define a domain-specific scheme, thin-edge must be enriched with a description of the misc entities, devices and services as well as their hierarchy. This is the aim of the [registration process](https://github.com/thin-edge/thin-edge.io/pull/2030/files#diff-9695a2a09b64f1cfa0ed7f9e9ff632abdc04a7fd3dc4c3849cf5b53a461e4addR13).

The point here is to implement a `struct DigitalTwin`  to keep an in-memory representation
of a device, its child-devices, and the services running on these devices.

The main feature of this struct is to map `+/+/+/+` entity identifiers to their:
- type: the main device? a child device? a service?
- parent, if any
- external device/service id: the id used on the cloud for this device/service. It can be the device certificate CN, an id given by the user or [derived from the entity identifier](https://thin-edge.github.io/thin-edge.io/next/references/mappers/c8y-mapper/)
- domain-specific data, such as a role or a location

The primary usage of this struct is for the c8y mapper, but not only as all the mappers might need it
as well as internal services as health checks.
- The c8y mapper needs to
  - get all the entities with their type, name and parent (to register these entities in the cloud)
  - be aware of any entity update (to register any new entity in the cloud)
  - get the device external id to which some telemetry data is related (`te/device/xyz///a/disk_full` => the child device named `child-xyz`)
  - get the entity identifier for from a device external id (a restart command for the device named `child-xyz` => `te/device/xyz///cmd/restart`)
- The health-check services needs to
  - get the list of services running on a device (in order to broadcast health-check messages)
- This map must be dynamic, with 
  - updates on registration messages received on [`te/+`, `te/+/+`, `te/+/+/+` or `te/+/+/+/+`](https://thin-edge.github.io/thin-edge.io/next/references/mappers/c8y-mapper/#registration)
  - updates on lookup request for an identifier in order to implement [auto-registration](https://thin-edge.github.io/thin-edge.io/next/contribute/design/mqtt-topic-design/#support-auto-registration-via-inference)
  - notification of any update on an identifier.

Hence, the needs are to:
[...]

PR #2122, opened 2023-08-03
Digital twin

Bravo555:
## TODO

- [x] adding entities
- [x] listing all entities
- [x] get the main device
- [x] get/set an entity struct for an entity identifier `+/+/+/+`
- [x] list the service running on a device
- [x] list child devices of a device or child device
- [x] provide a channel where updates are notified (`update()` returns a list of affected entities)
- [x] create empty twin with a function to derive a name from an (identifier, entity) pair
- [x] derive a name for each entity (indexing this names)
- [x] get an entity identifier from and entity name
- [x] ~~make `entity_store` use `entity` API~~ not a blocker, can be done later

## Proposed changes

Add `entity` and `entity_store` modules that represent entities and allow keeping track of registered entities, respectively.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2081

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity.rs
+++ b/crates/core/tedge_api/src/entity.rs
@@ -1,0 +1,284 @@
+//! A module defining entities, their types, and utilities for parsing MQTT
+//! topics following the default thin-edge MQTT scheme.
+
+use std::str::FromStr;
+
+// TODO: read from config
+const MQTT_ROOT: &str = "te";
+
+/// A thin-edge entity MQTT topic.
+///
+/// An entity topic consists of 3 groups: root, entity identifier, and
+/// optionally a channel. To be a valid entity topic, a topic must start with a
+/// root, and then have its entity identifier and channel (if present) groups
+/// successfully parsed.
+///
+/// ```
+/// # use tedge_api::entity::{EntityTopic, Channel, ChannelCategory};
+/// let entity_topic: EntityTopic =
+///     format!("te/device/child001/service/service001/m/measurement_type")
+///         .parse()
+///         .unwrap();
+/// assert_eq!(entity_topic.entity_id(), "device/child001/service/service001");
+/// assert_eq!(entity_topic.channel(), Some(&Channel {
+///     category: ChannelCategory::Measurement,
+///     r#type: "measurement_type".to_string(),
+///     suffix: "".to_string()
+/// }));
+/// ```
+///
+/// https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#topic-scheme
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EntityTopic {
+    entity_id: EntityId,
+    channel: Option<Channel>,
+}
+
+impl EntityTopic {
+    pub fn entity_id(&self) -> &str {
+        self.entity_id.0.as_str()
+    }
+
+    pub fn channel(&self) -> Option<&Channel> {
+        self.channel.as_ref()
+    }
+
+    /// Returns a device name if entity topic identifier is not using a custom
+    /// schema.
+    pub fn device_name(&self) -> Option<&str> {
+        match self.entity_id.0.split('/').collect::<Vec<&str>>()[..] {
+            ["device", device_id, "service", _] => Some(device_id),
+            ["device", device_id, "", ""] => Some(device_id),
+            _ => None,
+        }
+    }
+
+    /// Returns a service name if entity topic identifier is not using a custom
+    /// schema and the entity identifier refers to the service.
+    pub fn service_name(&self) -> Option<&str> {
+        match self.entity_id.0.split('/').collect::<Vec<&str>>()[..] {
+            ["device", _, "service", service_id] => Some(service_id),
+            _ => None,
+        }
+    }
+}
+
+impl FromStr for EntityTopic {
+    type Err = EntityTopicError;
+
+    fn from_str(topic: &str) -> Result<Self, Self::Err> {
+        const ENTITY_ID_SEGMENTS: usize = 4;
+
+        let (root, topic) = topic.split_once('/').ok_or(EntityTopicError::Root {
+            expected: MQTT_ROOT.to_string(),
+            got: topic.to_string(),
+        })?;
+
+        if root != MQTT_ROOT {
+            return Err(EntityTopicError::Root {
+                expected: MQTT_ROOT.to_string(),
+                got: root.to_string(),
+            });
+        }
+
+        let mut topic_separator_indices = topic.match_indices('/').map(|(i, _)| i);
+        let id_channel_separator_index = topic_separator_indices.nth(3).unwrap_or(topic.len());
+
+        let (entity_id, channel) = topic.split_at(id_channel_separator_index);
+
+        let entity_id_segments = entity_id.matches('/').count();
+        let missing_slashes = ENTITY_ID_SEGMENTS - entity_id_segments - 1;
+        let entity_id = format!("{entity_id}{:/<1$}", "", missing_slashes);
+
+        let channel = channel.trim_start_matches('/');
+        let channel = if !channel.is_empty() {
+            Some(Channel::new(channel)?)
+        } else {
+            None
+        };
+
+        Ok(EntityTopic {
+            entity_id: EntityId(entity_id.to_string()),
+            channel,
+        })
+    }
+}
+
+impl TryFrom<mqtt_channel::Topic> for EntityTopic {
+    type Error = EntityTopicError;
+
+    fn try_from(topic: mqtt_channel::Topic) -> Result<Self, Self::Error> {
+        topic.name.parse()
+    }
+}
+
+#[derive(Debug, thiserror::Error, PartialEq, Eq, Clone)]
+pub enum EntityTopicError {
+    #[error("Fist topic segment expected to be {expected:?}, got {got:?}")]
+    Root { expected: String, got: String },
+
+    #[error("Channel group invalid")]
+    Channel(#[from] ChannelError),
+}
+
+/// Represents an entity identifier group in thin-edge MQTT scheme.
+///
+/// An entity identifier is a fixed 4-segment group, as such any 4 topic
+/// segments that come after the root are considered a part of an identifier,
+/// even if they contain values usually present in the channel group, e.g.
+/// `/m/`.
+///
+/// If the topic ends before the expected 4 segments, the remaining segments are
+/// filled by empty segments (`//`).
+///
+/// # Example
+///
+///
+/// https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#group-identifier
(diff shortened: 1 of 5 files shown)


crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -110,6 +133,63 @@ impl C8yMapperActor {
...
+    ///
+    /// It returns MQTT register messages for the given entities to be published
+    /// by the mapper, so other components can also be aware of a new device
+    /// being registered.
+    fn auto_register_entity(&mut self, entity_id: &str) -> Vec<Message> {
+        let mut register_messages = vec![];
+        if let Some(("te", topic)) = entity_id.split_once('/') {
+            let (device_id, service_id) = match topic.split('/').collect::<Vec<&str>>()[..] {
+                ["device", device_id, "service", service_id, ..] => (device_id, Some(service_id)),
+                ["device", device_id, ..] => (device_id, None),
```
didier-wenzek, line 150: I would discard any message that strictly follow the default scheme.

```suggestion
                ["device", device_id, "", ""] => (device_id, None),
```

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -110,6 +133,63 @@ impl C8yMapperActor {
...
 
+    /// Performs auto-registration process for an entity under a given
+    /// identifier.
+    ///
+    /// If an entity is a service, its device is also auto-registered if it's
+    /// not already registered.
+    ///
+    /// It returns MQTT register messages for the given entities to be published
+    /// by the mapper, so other components can also be aware of a new device
+    /// being registered.
```
didier-wenzek, line 144: This is a good idea. But I would be cautious. Indeed, the measurement/alarm/event must not be sent *before* the device is actually registered on C8Y. But, the current code jumps into the translation.

didier-wenzek (commented):
In a good shape! Thanks

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -0,0 +1,478 @@
...
+/// assert_eq!(entity_store.main_device(), Some("te/device/main//"));
+/// ```
+#[derive(Debug, Clone)]
+pub enum EntityStore {
+    Uninitialised,
+    Initialised {
+        main_device: EntityId,
+        entities: HashMap<EntityId, EntityMetadata>,
+    },
+}
```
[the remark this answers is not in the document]
Bravo555: I talked with @reubenmiller about whether the main device also has to be registered, and whether or not it can be auto-registered, and I ended up rewriting this once again under the assumption that when the entity store is created, the main device is already known.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -0,0 +1,478 @@
...
+
+                Ok(Vec::new())
+            }
+
+            EntityStore::Initialised {
+                entities,
+                main_device,
+            } => {
+                if message.r#type == "device" {
+                    return Err(Error::MainDeviceAlreadyRegistered);
```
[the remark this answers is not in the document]
Bravo555: fixed

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -0,0 +1,539 @@
...
+#[derive(Debug, Clone)]
+pub struct EntityStore {
+    main_device: EntityId,
+    entities: HashMap<EntityId, EntityMetadata>,
+    external_id_index: HashMap<String, EntityId>,
+}
+
+impl EntityStore {
+    /// Creates a new entity store with a given main device.
+    pub fn with_main_device(main_device: EntityRegistrationMessage) -> Option<Self> {
```
didier-wenzek, line 69: This is okay for a first version  but will not work in a setting where the main-device topic is defined by the user using a registration message.

When the mapper starts, the id of the main device is known (This is the CN of its certificate) but not its topic (unless we are using the default topic scheme where the main device is attached to the well-known topic `te/device/main//`).

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// # Examples
+    /// - `device/main//` => `DEVICE_COMMON_NAME`
+    /// - `device/child001//` => `DEVICE_COMMON_NAME:device:child001`
+    /// - `device/child001/service/service001` => `DEVICE_COMMON_NAME:device:child001:service:service001`
+    /// - `factory01/hallA/packaging/belt001` => `DEVICE_COMMON_NAME:factory01:hallA:packaging:belt001`
+    fn external_id(&self, entity_topic: EntityIdRef) -> String {
+        let main_device_external_id = &self.get(&self.main_device).unwrap().external_id;
+        let external_id_suffix = entity_topic.replace('/', ":");
+        let external_id_suffix = external_id_suffix.trim_matches(':');
+        format!("{main_device_external_id}:{external_id_suffix}")
```
didier-wenzek, line 202: No. For the main device the answer is `DEVICE_COMMON_NAME` and not `DEVICE_COMMON_NAME:device:main`

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// The external id is generated by prefixing the id with main device name
+    /// (device_common_name) and then appending the MQTT entity topic with `/`
+    ///  characters replaced by `:`.
+    ///
+    /// # Examples
+    /// - `device/main//` => `DEVICE_COMMON_NAME`
+    /// - `device/child001//` => `DEVICE_COMMON_NAME:device:child001`
+    /// - `device/child001/service/service001` => `DEVICE_COMMON_NAME:device:child001:service:service001`
+    /// - `factory01/hallA/packaging/belt001` => `DEVICE_COMMON_NAME:factory01:hallA:packaging:belt001`
+    fn external_id(&self, entity_topic: EntityIdRef) -> String {
```
didier-wenzek, line 198: A better name for this function is `derive_external_id` as this is computation used when the registration message doesn't provide an id.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -0,0 +1,539 @@
...
+        // parent device is affected if new device is its child
+        if let Some(parent) = &parent {
+            if !self.entities.contains_key(parent) {
+                return Err(Error::NoParent(parent.clone().into_boxed_str()));
+            }
+
+            affected_entities.push(parent.clone());
+        }
+
+        let external_id = self.external_id(&message.mqtt_id);
```
didier-wenzek, line 160: If the message provides an `external_id` this one must be used.

```suggestion
        let external_id = message.external_id.unwrap_or_else(||self.external_id(&message.mqtt_id));
```

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct EntityMetadata {
+    parent: Option<EntityId>,
+    r#type: String,
+    external_id: String,
+    other: serde_json::Value,
+}
```
didier-wenzek, line 212: I would consider to use an enum here, ruling out the cases that makes no sense.

```suggestion
#[derive(Debug, Clone)]
pub struct EntityMetadata {
    r#type: EntityType,
    other: serde_json::Value,
}

pub enum EntityType {
    MainDevice{ device_id: String },
    ChildDevice( device_id: String, parent: EntityId },
    Service( service_id: String, device: EntityId },
}
```

didier-wenzek (commented):
We need to come with better names for the misc kind of identifier with have.
- We use in turn `identifier`, `entity_id`, `mqtt_id`, `external_id`,`device_id` for related but different things.
- `external_id` is confusing (my bad  as I proposed it) as this is used by the c8y mapper for the internal id given by c8y. A better name would be in fact `device_id`: for the main device this is the CN of its certificate , for a child device this is name assigned by the user or derived from the parent device id and the child identifier.
- You the term `mqtt_id` somewhere in this PR. I like it. This might be less confusing than `device_identifier` which rhymes with `device_id`. Other alternatives could be `device_topic` or `entity_topic`.

crates/core/tedge_api/src/entity_store.rs
Bravo555: Because `device_id` and `service_id` both refer to the external id, and in the case of child devices and services `parent` and `device` mean basically the same thing, I've opted to make the entity type just a plain enum.

crates/core/tedge_api/src/entity.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ChannelCategory {
+    Measurement,
+    Event,
+    Alarm,
+    Command,
+}
```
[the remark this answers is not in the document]
Bravo555: In this case, with all the `r#type`s being inside enum variants, it makes pattern matching on the channel harder. We'd be forced to match on every variant to extract the type, which is basically the same thing.
Perhaps the same type should not be used by multiple categories, so the set of possible values might be different for each variant, but the suggestion does not prevent that while making it a bit harder to match in the general case.

I'm also not very familiar with the idea of what's the impact of the `/meta` suffix and I'd like to keep it simple for now. I've aimed to make it a minimal wrapping over MQTT scheme, adding typing to very basic stuff. I'd rather see how this basic shape fits and then revisit this.

Bravo555:
All the tests should pass now, will squash after positive review

crates/common/tedge_config_macros/Cargo.toml
```diff
@@ -15,7 +15,7 @@ once_cell = "1.17"
 serde = "1.0"
 tedge_config_macros-macro = { path = "macro" }
 thiserror = "1.0"
+tracing = "0.1"
 url = "2.3"
 
 [dev-dependencies]
-tracing = "0.1"
```
didier-wenzek, line 21: If not moved into another PR this change must at least be extracted in a specific commit fixing `cargo doc`.

didier-wenzek (commented):
I'm okay to merge this PR despite some discussions still open. This work is expected to unblock the work on the c8y mappers (to interact with  tedge-log-plugin and to use the new MQTT topic scheme).

We still need to discuss:
- [Channel type](https://github.com/thin-edge/thin-edge.io/pull/2122#discussion_r1293640062)
- [EntityType](https://github.com/thin-edge/thin-edge.io/pull/2122#discussion_r1293640062)
- [Naming](https://github.com/thin-edge/thin-edge.io/pull/2122#pullrequestreview-1577115611)

@Bravo555 can you clean the gitlog history squashing the fixup commits.

crates/common/tedge_config_macros/Cargo.toml
Bravo555: extracted into #2160

didier-wenzek (approved):
Approved

PR #2123, opened 2023-08-03
fix(install): improve maintainer script compatibility in preparation for future package support

reubenmiller:
## Proposed changes

Various improvements and fixes to the debian maintainer scripts including. Some of the changes are in preparation for support other linux package formats, so they are intended to be more generic and remove debian specific assumptions.

* remove reliance on sudo (as maintainer scripts run as root anyway)
* avoid using recurse option when deleting files e.g. `rm -r` (as this can have unintended consequences if the path is a folder and not a file)
* don't delete operations folder when the mapper is removed (as the mapper is not the owner of the folder contents, and tedge will delete /etc/tedge on purge anyway)
* only add sudoer config if sudoers config folder exists
* fallback to systemctl commands if deb helper is not installed
* support user and group management using adduser or useradd for improved compatibility (to support Alpine linux, RHEL and Debian)
* improve editing of mosquitto conf file by not assuming the file already has a include_dir defined and only insert it once

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/configuration/debian/tedge/preinst
+++ b/configuration/debian/tedge/preinst
@@ -1,15 +1,31 @@
 #!/bin/sh
 set -e
 
+command_exists() {
+    command -V "$1" >/dev/null 2>&1
+}
+
 ### Create groups
 if ! getent group tedge >/dev/null; then
-    groupadd --system tedge
+    if command_exists groupadd; then
+        groupadd --system tedge
+    elif command_exists addgroup; then
+        addgroup -S tedge
+    else
+        echo "WARNING: Could not create group: tedge" >&2
+    fi
 fi
 
 ### Create users
 # Create user tedge with no home(--no-create-home), no login(--shell) and in group tedge(--gid)
 if ! getent passwd tedge >/dev/null; then
-    useradd --system --no-create-home --shell /sbin/nologin --gid tedge tedge
+    if command_exists useradd; then
+        useradd --system --no-create-home --shell /sbin/nologin --gid tedge tedge
+    elif command_exists adduser; then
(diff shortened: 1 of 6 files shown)


Bravo555 (approved):
lgtm :+1:

configuration/debian/tedge/preinst
```diff
@@ -1,69 +1,89 @@
...
+
 ### Create groups
 if ! getent group tedge >/dev/null; then
-    groupadd --system tedge
+    if command_exists groupadd; then
+        groupadd --system tedge
+    elif command_exists addgroup; then
+        addgroup -S tedge
+    else
+        echo "WARNING: Could not create group: tedge" >&2
```
rina23q, line 15: It's just a question. I'm wondering if failing user/group creation of `tedge` is "warning". Indeed people can create it later and if they want to run everything with `root` then no pronlem (e.g. in container). But other components like tedge-mapper will run by the `tedge` user as long as systemd is used. So, in some scenario, no existence of `tedge` is more like "error", although not all scenarios.

rina23q (approved):
Approved. But I have one question regarding failure of `tedge` user/group creation.

configuration/debian/tedge/preinst
reubenmiller: It is warning here because `tedge init` will most likely fail later on because it tries to create files/folders with the configured owner. But I didn't want the user creation to be coupled with the `tedge init` code, so that in case if we change the handling of `tedge init` to be more relaxed if the user is not present, then we don't have to change the maintainer script.

reubenmiller:
Rebasing to include zig fix.

PR #2129, opened 2023-08-04
fix(install): improve reliability of latest version detection by using non rate limited api

reubenmiller:
## Proposed changes

Improve the get thin-edge.io install script to avoid rate limiting enforced by GitHub. Sometimes the install script was failing due to the latest version detection logic which relied on the `api.github.com` endpoint which has fairly low limits for unauthenticated requests.

For example, if a user does too many requests, then the following response is returned by GitHub.

```
{"message":"API rate limit exceeded for 80.133.228.217. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

The error was mostly occurred in the integration test runs in GitHub as the rate limiting would be hit due to parallel workflow runs which are attributed to the same IP address.

Below shows the before and after the change.

**Before**

The requests are limited after about 60 attempts in a short period of time

```sh
i=0
while true; do (curl -Ls 'https://api.github.com/repos/thin-edge/thin-edge.io/releases/latest' | grep -vi 'rate limit exceeded' >/dev/null ) && echo " success - attempt $i" || echo " failed - attempt $i"; sleep 0.25; i=$((i+1)); done
```

*Output*

```sh
 success - attempt 56
 success - attempt 57
 failed - attempt 58
 failed - attempt 59
```

**After**

No rate limiting (or an at least the limit has not been found yet)

```sh
i=0
while true; do (curl -Ls -o /dev/null -w "%{url_effective}" 'https://github.com/thin-edge/thin-edge.io/releases/latest' | grep '0.12.0' >/dev/null ) && echo " success - attempt $i" || echo " failed - attempt $i"; sleep 0.25; i=$((i+1)); done
```

*Output*

```sh
#...
 success - attempt 116
 success - attempt 117
 success - attempt 118
 success - attempt 119
 success - attempt 120
 success - attempt 121
 success - attempt 122
 success - attempt 123
 success - attempt 124
 success - attempt 125
```

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/get-thin-edge_io.sh
+++ b/get-thin-edge_io.sh
@@ -244,23 +244,21 @@ install_tedge_plugins() {
 }
 
 get_latest_version() {
-    # Detect latest version from github api to avoid having a default version in the script
+    # Detect latest version from github using api to avoid having a default version in the script
+    # The request checks the redirected url behind the latest url.
+    # Example:
+    #   https://github.com/thin-edge/thin-edge.io/releases/latest => https://github.com/thin-edge/thin-edge.io/releases/tag/0.12.0
+    #
+    # Note: avoid using the api endpoint (api.github.com) as it is rate limited for unauthenticated requests.
     if command_exists curl; then
-        response=$(curl -s https://api.github.com/repos/thin-edge/thin-edge.io/releases/latest)
+        response=$(curl -Ls -o /dev/null -w "%{url_effective}" 'https://github.com/thin-edge/thin-edge.io/releases/latest')
     elif command_exists wget; then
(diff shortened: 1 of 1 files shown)


PradeepKiruvale (commented):
LGTM,
Why don't we use authenticated users for the GH installations?

reubenmiller:
> LGTM, Why don't we use authenticated users for the GH installations?

Because we would have to inject the github token into the containers where the tests are running from...but using an authenticated api request does not make sense to for the install script anyway so there is no point adding it for a special cause for the integration tests.

PR #2130, opened 2023-08-04
Allow log filtering with $RUST_LOG

Bravo555:
## Proposed changes

`--debug` flag used to increase log verbosity is very noisy, as it enables `trace` (lowest level) logs from all modules, as well as dependencies.
On the contrary, `RUST_LOG` environment variable can be used to set different logging levels on a per-module basis, allowing to e.g. set `trace` for a module one is currently working on and `info` on the rest, or limiting logging only to the current crate.

As such, `set_log_level` function was modified to not overide a `$RUST_LOG` filter if it is present. If this variable is not present, it sets log level globally as usual.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/system_services/log_config.rs
+++ b/crates/common/tedge_config/src/system_services/log_config.rs
@@ -19,14 +19,23 @@ pub fn get_log_level(
     }
 }
 
-/// Initialize a `tracing_subscriber`
-/// Reports all the log events sent either with the `log` crate or the `tracing` crate.
+/// Initializes a tracing subscriber with a given log level if environment
+/// variable `RUST_LOG` is not present.
+///
+/// Reports all the log events sent either with the `log` crate or the `tracing`
+/// crate.
 pub fn set_log_level(log_level: tracing::Level) {
-    tracing_subscriber::fmt()
+    let subscriber = tracing_subscriber::fmt()
         .with_writer(std::io::stderr)
-        .with_timer(tracing_subscriber::fmt::time::UtcTime::rfc_3339())
-        .with_max_level(log_level)
-        .init();
+        .with_timer(tracing_subscriber::fmt::time::UtcTime::rfc_3339());
+
+    if std::env::var("RUST_LOG").is_ok() {
+        subscriber
+            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+            .init();
+    } else {
+        subscriber.with_max_level(log_level).init();
+    }
 }
 
 #[cfg(test)]
--- a/crates/core/tedge_actors/src/message_boxes.rs
+++ b/crates/core/tedge_actors/src/message_boxes.rs
@@ -97,7 +97,7 @@ use crate::RuntimeRequest;
 use async_trait::async_trait;
 use futures::channel::mpsc;
 use futures::StreamExt;
-use log::info;
+use log::debug;
 use std::fmt::Debug;
 
 /// Either a message or a [RuntimeRequest]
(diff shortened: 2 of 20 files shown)


crates/common/tedge_config/src/system_services/log_config.rs
```diff
@@ -19,14 +19,23 @@ pub fn get_log_level(
     }
 }
 
-/// Initialize a `tracing_subscriber`
-/// Reports all the log events sent either with the `log` crate or the `tracing` crate.
+/// Initializes a tracing subscriber with a given log level if environment
```
PradeepKiruvale, line 22:
```suggestion
/// Initializes a tracing subscriber with a given log level through --debug, if the environment
```

PradeepKiruvale (commented):
IMO, it's better to mention in the documentation, which is a higher priority `RUST_LOG` Vs --debug flag.

didier-wenzek:
> Do we really need a crate-level granularity for filtering log levels?

I see that as useful, knowing that this can be done at the actor level:

```
RUST_LOG=C8yMapper=debug,MQTT=debug tedge-mapper c8y
```

> One could argue if TRACE would better suited for showing inter actor communication, and only the MQTT message are logged (or some variant of that).

Agree, however I vote for `DEBUG`, as `TRACE` is super (too) verbose.

I also see as an issue the fact that [`--debug` enable the `TRACE` level](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_mapper/src/main.rs#L94). `--debug` should be `DEBUG`.

> I think this is a byproduct to the real issue that we are logging in some instances to INFO instead of TRACE. For example, the software list messages result in multiple (very long) messages being logged essentially spamming the log output when there are a large number (>200) debian packages being detected by tedge-apt-plugin. 

Yes, `TRACE` would be better for this message (as published over MQTT).

crates/common/tedge_config/src/system_services/log_config.rs
Bravo555: It's the caller who chooses whether or not to call this function when `--debug` is present or not, and the function doesn't check if `--debug` is present or not, so the function doc should not mention it.

crates/core/tedge_actors/src/message_boxes.rs
```diff
@@ -184,25 +184,25 @@ impl<Input: Debug> LoggingReceiver<Input> {
 impl<Input: Send + Debug> MessageReceiver<Input> for LoggingReceiver<Input> {
     async fn try_recv(&mut self) -> Result<Option<Input>, RuntimeRequest> {
         let message = self.receiver.try_recv().await;
-        info!(target: &self.name, "recv {:?}", message);
+        trace!(target: &self.name, "recv {:?}", message);
```
didier-wenzek, line 187: I would prefer to use the `DEBUG` level to trace the message received by an actor.

From my point of view:
- `TRACE` makes sense only if you are involved in the development. i.e. you are a developer.
- A user should be able to understand what's is going on with `DEBUG`.

didier-wenzek (commented):
@Bravo555  and @reubenmiller We need to discuss the log level used for the messages received by the actors.

crates/core/tedge_actors/src/message_boxes.rs
Bravo555: Changed inter-actor level communication log levels to `DEBUG`, leaving `TRACE` for clear cases for printing payloads where previously-printed `DEBUG` level logs leave enough context for the user to know what's happening.

Bravo555:
> - configuration should use the following order of preference (following the 12 factor app guidelines), where the first non-empty value is used in order of: 
>     1. flag (if one exists)
>     2. environment variable
>     3. config file

Doing this would require rewriting more code in each crate that we do config, taking more time and effort that i don't think is worth it. It's just every time I start working on a feature, I need to comment out the `set_log_level` call to be able to filter out log messages by crate or level. If having this undocumented environment variable which overrides the `--debug` flag is a problem, then I can discard this commit, leaving the ones that tweak log levels of some messages, on which we seem to have a consensus.

> we should try to re-use the existing tedge config env variable/config file mechanism that we have (if possible) rather than creating another custom env variable mapping.

To be honest, I have no idea what options in our tedge config do we have, that would be relevant to this. It's just that crate based-filtering is very useful for developers and it's what you get with `log` and `tracing` crates for free, but from my point of view it seems that it was disabled unintentionally.

didier-wenzek (approved):
Approved

PR #2139, opened 2023-08-09
fix the order of child id in the tedge topics

PradeepKiruvale:
## Proposed changes

This PR fixes two issues
1. The misplaced `child-id` in the `alarm` and `event` topics. This has to be as per [this link ](https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#telemetry-child-device)
2. Output message pattern matching in robot tests. Especially for the `alarm` tests.
    Since the  `te converter` adds the `severity` into the payload, the order in which it adds will become very random. So, better to check if the output message contains the `severity` fragment or not, instead of the whole payload.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -179,8 +179,8 @@ impl Agent {
         runtime.spawn(restart_actor_builder).await?;
         runtime.spawn(software_update_builder).await?;
         runtime.spawn(converter_actor_builder).await?;
-        runtime.spawn(health_actor).await?;
         runtime.spawn(tedge_to_te_converter).await?;
+        runtime.spawn(health_actor).await?;
 
         runtime.run_to_completion().await?;
 
--- a/tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
+++ b/tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
@@ -18,32 +18,29 @@ Convert main device measurement topic
 
 Convert child device measurement topic
     Execute Command    tedge mqtt pub tedge/measurements/child '{"temperature":25}'
-    Should Have MQTT Messages    te/device/child///m/    message_pattern={"temperature":25}   
+    Should Have MQTT Messages    te/device/child///m/    message_pattern={"temperature":25}
 
 Convert main device event topic
     Execute Command    tedge mqtt pub tedge/events/login_event '{"text":"someone logedin"}'
     Should Have MQTT Messages    te/device/main///e/login_event    message_pattern={"text":"someone logedin"}
 
 Convert child device event topic
-    Execute Command    tedge mqtt pub tedge/events/child/login_event '{"text":"someone logedin"}'
+    Execute Command    tedge mqtt pub tedge/events/login_event/child '{"text":"someone logedin"}'
     Should Have MQTT Messages    te/device/child///e/login_event    message_pattern={"text":"someone logedin"}
 
 Convert main device alarm topic
-    Execute Command    tedge mqtt pub tedge/alarms/minor/test_alarm '{"text":"test alarm"}'
-    Should Have MQTT Messages    te/device/main///a/test_alarm    message_pattern={"text":"test alarm","severity":"minor"}    
+    Execute Command    tedge mqtt pub tedge/alarms/minor/test_alarm '{"text":"test alarm"}' -q 2 -r
+    Should Have MQTT Messages    te/device/main///a/test_alarm    message_contains=minor
 
 Convert child device alarm topic
-    Execute Command    tedge mqtt pub tedge/alarms/minor/child/test_alarm '{"text":"test alarm"}'
-    Should Have MQTT Messages    te/device/child///a/test_alarm    message_pattern={"text":"test alarm","severity":"minor"}    
+    Execute Command    tedge mqtt pub tedge/alarms/major/test_alarm/child '{"text":"test alarm"}' -q 2 -r
+    Should Have MQTT Messages    te/device/child///a/test_alarm    message_contains=major
 
 
 *** Keywords ***
 Custom Setup
     Setup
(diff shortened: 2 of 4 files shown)


crates/core/tedge_agent/src/agent.rs
```diff
@@ -179,8 +179,8 @@ impl Agent {
         runtime.spawn(restart_actor_builder).await?;
         runtime.spawn(software_update_builder).await?;
         runtime.spawn(converter_actor_builder).await?;
-        runtime.spawn(health_actor).await?;
         runtime.spawn(tedge_to_te_converter).await?;
+        runtime.spawn(health_actor).await?;
```
PradeepKiruvale, line 183: Start all the other actors before the health actor starts.

didier-wenzek: Why? What was the issue? How this fix improves things?

tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
```diff
@@ -18,32 +18,29 @@ Convert main device measurement topic
...
 Convert child device event topic
-    Execute Command    tedge mqtt pub tedge/events/child/login_event '{"text":"someone logedin"}'
+    Execute Command    tedge mqtt pub tedge/events/login_event/child '{"text":"someone logedin"}'
     Should Have MQTT Messages    te/device/child///e/login_event    message_pattern={"text":"someone logedin"}
 
 Convert main device alarm topic
-    Execute Command    tedge mqtt pub tedge/alarms/minor/test_alarm '{"text":"test alarm"}'
-    Should Have MQTT Messages    te/device/main///a/test_alarm    message_pattern={"text":"test alarm","severity":"minor"}    
+    Execute Command    tedge mqtt pub tedge/alarms/minor/test_alarm '{"text":"test alarm"}' -q 2 -r
+    Should Have MQTT Messages    te/device/main///a/test_alarm    message_contains=severity  message_contains=minor
```
didier-wenzek, line 33: @reubenmiller Do we have a way to compare json values? and not strings?

crates/core/tedge_agent/src/agent.rs
reubenmiller: There was a suspicion that some of the integration test failures where potentially caused by the converter actor not being "ready" before the messages were received by the broker (as the test would check the health status of the `tedge-agent` before publishing to the measurements topic.

tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
PradeepKiruvale: > No, but I'm happy to add something in that can accept a dictionary, and it will do a deep object comparison.

That would be super helpful for a scenario like this.

> 
> Btw, the suggested change won't work because `message_contains` only accepts a string (not an array of strings), below is the function signature.
> 
> ```python
> @keyword("Should Have MQTT Messages")
>     def mqtt_should_have_topic(
>         self,
>         topic: str,
>         date_from: relativetime_ = None,
>         date_to: relativetime_ = None,
>         message_pattern: str = None,
>         message_contains: str = None,
>         minimum: int = 1,
>         maximum: int = None,
>         **kwargs,
>     ) -> List[Dict[str, Any]]:
> ```

But, I can just check one string right?, say for example just `severity`, IMO that would be sufficient for now, if implementing a new API takes longer.

crates/core/tedge_agent/src/agent.rs
PradeepKiruvale: Yes, as @reubenmiller mentioned it's to make sure the `tedge-agent` is up. This is not completely fail-safe as the `health actor` does not check whether all the other actors are started successfully or not. But, at least we are sure that the `te converter` actor spawned (though it might hang and not be completely ready to serve the requests) before starting the `health` actor.

tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
didier-wenzek: > I'm happy to add something in that can accept a dictionary, and it will do a deep object comparison.

This can be done later. This fix is rather urgent.

> Btw, the suggested change won't work because message_contains only accepts a string (not an array of strings)

You mean that one can check either that `message_contains=severity` or `message_contains=minor` but not both?

reubenmiller: This is where the doc strings are helpful, yes, `message_contains=` does not support multiple values, however you can `message_pattern=` which accepts a regular expression (which you can then make to accept multiple values).

For example:

```
message_pattern="value":\s*\d+
```

crates/core/tedge_agent/src/agent.rs
didier-wenzek: Okay. So you suspect that:

1. The `health_actor` sends an `up` status before `tedge_to_te_converter` is spawned.
2. The test sends a measurement on `tedge` and expects the same on `te`.
3. The test gives up waiting for the message on `te` before `tedge_to_te_converter` got a chance to be spawned and notified of the measurement on `tedge`.

I see that as extremely unlikely unless `tedge_to_te_converter` hangs .i.e. is buggy. Indeed, the difference between the time scale of tokio context-switch and system test timeouts is of an order of magnitude 10^6 .
 
I have nothing against this change. However, if you observed such failures then the issue might be deeper.

tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
didier-wenzek: We need to move fast - as the system tests are broken on man.

=> I simplified the tests for alarms, only checking that the severity is inserted in the message payload.

didier-wenzek (approved):
Approved

PR #2148, opened 2023-08-10
docs(install): update installation guide to detail installing thin-edge.io on any distribution

reubenmiller:
## Proposed changes

Add new install guide which uses the brand new install.sh script which installs thin-edge.io using package managers.

The instructions detail how to install thin-edge.io on various linux distributions (not just Debian based).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


docs/src/install/index.md
```diff
@@ -4,200 +4,213 @@ tags: [Installation]
...
-sudo systemctl stop c8y-log-plugin
+```sh tab={"label":"Alpine"}
+sudo apk update
+sudo apk add --no-cache tedge-full
 ```
-:::
 
-Run `get-thin-edge_io.sh` script as below to upgrade to the latest version.
+:::info
+The update only works if the tarball package manager was not used. The tarball installation does not use a package manager, as the install.sh script handles the downloading and extracting of the thin-edge.io tarball's contents. In this case, the easiest option is to continue using the install.sh script.
```
didier-wenzek, line 54: One might add that the `install.sh` script works fine to update thin-edge when already installed.

By the way, what if thin-edge has been installed with a different script/method?

didier-wenzek (approved):
Really nice!  This demonstrares that improving the software makes the documentation easier to write and read!

I would also consider to remove `docs/src/operate/installation/install_and_enable_software_management.md` 

Fix the two typos in `install.sh`
  - "guranteed" should be "guaranteed"
  - "compatable" should be "compatible"
 
Broken links:
```
[WARNING] Docs markdown link couldn't be resolved: (../installation/installation_without_deb_support.md) in "/docusaurus/docs/operate/c8y/config_management_plugin.md" for version current
[WARNING] Docs markdown link couldn't be resolved: (../installation/installation_without_deb_support.md) in "/docusaurus/docs/operate/c8y/config_management_plugin.md" for version current
```

didier-wenzek:
> I would also consider to remove `docs/src/operate/installation/how_to_install_thin_edge_manually.md` 

I would even consider to:
- move `operate/installation/install_and_enable_software_management.md`  under `operate/operations`
- fully deprecate the section `operate/installation`

docs/src/install/index.md
```diff
@@ -4,200 +4,213 @@ tags: [Installation]
...
+curl -1sLf 'https://dl.cloudsmith.io/public/thinedge/tedge-main/setup.rpm.sh' | bash
 ```
 
-```sh title="Example"
-sudo dpkg -i tedge-mapper_0.9.0_armhf.deb
+```sh tab={"label":"Alpine"}
+curl -1sLf 'https://dl.cloudsmith.io/public/thinedge/tedge-main/setup.alpine.sh' | bash
 ```
 
-## Uninstall
```
PradeepKiruvale, line 143: Is this moved to some other doc?

docs/src/install/index.md
reubenmiller: > One might add that the install.sh script works fine to update thin-edge when already installed.

Though I would have assumed that wasn't necessary because the section header is called "Install/Update"...but if it is not clear, then I'm happy to add a sentence to that effect.

> By the way, what if thin-edge has been installed with a different script/method?

It might have some issues if you want from the tarball to auto detection, as the auto detection could find a package manager and install it via that. Though the tarball section does show an example how to explicitly install it via the tarball (rather than using auto detection).

Though technically, overwriting it should not be a problem, if it it was a problem, then the package manager would throw an error explaining why (at least for tarball -> package manager, not the other way around).

reubenmiller:
> > I would also consider to remove `docs/src/operate/installation/how_to_install_thin_edge_manually.md`
> 
> I would even consider to:
> 
> * move `operate/installation/install_and_enable_software_management.md`  under `operate/operations`
> * fully deprecate the section `operate/installation`

Yes these parts will be done in follow up PRs. I wanted to break it down into bite-size chunks (since our new "next" docs supports this way or working better than before).

docs/src/install/index.md
reubenmiller: I'll add a new uninstall guide in a follow up PR, as this has also changed. But more changes are needed, e.g. changing the script, or instructions.

We can do this work in a followup PR because user's generally don't run the uninstall very often, so just focussing on the install part makes more sense as a first step.

reubenmiller:
> move operate/installation/install_and_enable_software_management.md under operate/operations

Yes, I'll move this page in a follow up PR.

reubenmiller:
> Broken links:

I just pushed updates to fix the broken links (I forgot to commit that change before). All the links should be fine now.

didier-wenzek (approved):
Approve the latest updates

gligorisaev:

- [x] QA has tested the feature and it meets the required specifications.

Issue #2150, opened 2023-08-11
Restart operation stuck in executing


toewsar:
**Describe the bug**
We've installed thin-edge on a custom board with a custom package-manager. 
We also had to relocate the `tmp.path` to a non-volatile path:
```shell
$ tedge config list
...
tmp.path=/mnt/data/tedge/download
...
```

When I trigger a restart from c8y the device restarts, but the agent does not set the operation to successful.
I checked the log, but I could find anything pointing me to the error.

Log:
```shell
$ cat /var/log/tedge-agent.err
2023-08-11T08:13:24.09699871Z  INFO flockfile::unix: Lockfile created "/run/lock/tedge-agent.lock"
2023-08-11T08:13:24.097237419Z  INFO tedge_agent::agent: tedge-agent starting
2023-08-11T08:13:24.097401085Z  INFO sm-agent: tedge_agent::agent: Starting tedge agent
2023-08-11T08:13:24.104584794Z  INFO Runtime: Started
2023-08-11T08:13:24.10618296Z  INFO Runtime: Running Signal-Handler-0
2023-08-11T08:13:24.109716335Z  INFO Runtime: Running HttpFileTransferServer-1
2023-08-11T08:13:24.122435544Z  INFO Runtime: Running MQTT-2
2023-08-11T08:13:24.14085996Z  INFO Runtime: Running RestartManagerActor-3
2023-08-11T08:13:24.140973794Z  INFO Runtime: Running SoftwareManagerActor-4
2023-08-11T08:13:24.141049127Z  INFO Runtime: Running TedgeOperationConverter-5
2023-08-11T08:13:24.141117002Z  INFO Runtime: Running HealthMonitorActor-6
2023-08-11T08:13:24.141268585Z  INFO tedge-agent: send Message { topic: Topic { name: "tedge/health/tedge-agent" }, payload: "{"pid":2838,"status":"up","time":1691741604}", qos: AtLeastOnce, retain: true }
2023-08-11T08:13:24.621208585Z  INFO plugin_sm::plugin_manager: Plugin activated: /mnt/data/tedge/sm-plugins/cds
2023-08-11T08:13:33.55458817Z  INFO plugin_sm::plugin_manager: Plugin activated: /mnt/data/tedge/sm-plugins/ap
2023-08-11T08:13:33.554966295Z  INFO plugin_sm::plugin_manager: Default plugin type: ap
[...]

PR #2152, opened 2023-08-11
Sync file on atomically_write_file

didier-wenzek:
The `atomically_write_file_async` and `atomically_write_file_sync` functions correctly flush the temporary content before atomically moving it to its destination. However, for the content to reach the disk, these methods must also call `fsync` on the file and its directory.

This notably critical for the restart operation which use the file system to resume the command when the system reboot.
See https://github.com/thin-edge/thin-edge.io/issues/2150

## Proposed changes

- [x] Update  `atomically_write_file_async` and  `atomically_write_file_sync` to call `fsync` on the file and its directory.
- [ ] Fix `create_tmp_restart_file` using  `atomically_write_file_async`.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2150

## Checklist


crates/core/tedge_agent/src/restart_manager/restart_operation_handler.rs
```diff
@@ -3,38 +3,24 @@ pub mod restart_operation {
...
+    /// This file will be compared on restart with /proc/uptime,
+    /// to determine if a reboot actually happened.
+    ///
+    /// See [has_rebooted]
+    pub async fn create_tmp_restart_file(tmp_dir: &Utf8Path) -> Result<(), RestartManagerError> {
         let path = &tmp_dir.join(TEDGE_AGENT_RESTART);
-        let path = Utf8Path::new(path);
+        let date_utc = OffsetDateTime::now_utc().unix_timestamp().to_string();
+
+        tedge_utils::fs::atomically_write_file_sync(path, date_utc.as_bytes())?;
```
didier-wenzek, line 22: For a reason I fail to understand for now, using the async version of `atomically_write_file_sync` makes failing the test `restart_manager::tests::test_new_restart_operation`.

The function `atomically_write_file_sync` called to create and populate a `TEDGE_AGENT_RESTART` fails to open the file for write. This is not an access issue as this works with `std:fs:File`.

didier-wenzek: Okay I got it. The `atomically_write_file_async` function is correct, but the test is based on a wrong assumption: i.e. that the restart witness file is create *before* is sent over MQTT the executing message.

=> I updated the restart actor so the executing message is sent only when all the files, used to check whether or not the device has reboot, are actually created.

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -155,12 +155,18 @@ impl RestartManagerActor {
...
         }
 
+        if let Err(err) = create_tmp_restart_file(&self.config.tmp_dir).await {
+            error!(
+                "Fail to create a witness file in {} due to: {}",
+                self.config.tmp_dir, err
+            );
+            return response.with_status(OperationStatus::Failed);
+        }
+
```
didier-wenzek, line 165: The call to `create_tmp_restart_file` has been move here to ensure that all the reboot witness files are created before is sent the message  that notifies that the command is executing.

crates/common/tedge_utils/src/fs.rs
```diff
@@ -50,16 +60,31 @@ pub async fn atomically_write_file_async(
         return Err(err);
     }
 
+    // Ensure the content reach the disk
```
PradeepKiruvale, line 63: It's not just the `content` but also `meta data` right?

reubenmiller (approved):
Approved

PR #2156, opened 2023-08-14
feat(cli): use tedge instead of mosquitto as owner of mosquitto files/folders

reubenmiller:
## Proposed changes

Use tedge as the user and group owner of all folders created by `tedge init`. This removed the dependency on the mosquitto user when initializing the file system, as the initialization could be done on a device which does not have mosquitto installed and is using an external mqtt broker hosted on a different device.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/init.rs
+++ b/crates/core/tedge/src/cli/init.rs
@@ -34,8 +34,8 @@ impl TEdgeInitCmd {
         create_directory(
             config_dir.join("mosquitto-conf"),
             PermissionEntry::new(
-                Some("mosquitto".into()),
-                Some("mosquitto".into()),
+                Some(self.user.clone()),
+                Some(self.group.clone()),
                 Some(0o775),
             ),
         )?;
@@ -57,11 +57,7 @@ impl TEdgeInitCmd {
         )?;
         create_directory(
             config_dir.join("device-certs"),
-            PermissionEntry::new(
-                Some("mosquitto".into()),
-                Some("mosquitto".into()),
-                Some(0o775),
-            ),
+            PermissionEntry::new(Some("root".into()), Some("root".into()), Some(0o775)),
         )?;
 
         let config = self.context.config_repository.load()?;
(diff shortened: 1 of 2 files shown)


crates/core/tedge/src/cli/init.rs
```diff
@@ -58,8 +58,8 @@ impl TEdgeInitCmd {
         create_directory(
             config_dir.join("device-certs"),
             PermissionEntry::new(
-                Some("mosquitto".into()),
-                Some("mosquitto".into()),
+                Some(self.user.clone()),
+                Some(self.group.clone()),
                 Some(0o775),
             ),
         )?;
```
didier-wenzek, line 65: Here, I would be bit cautious, as this directory contains the private keys of the device.

If the point is only to remove an hard-coded dependency to the `mosquitto` user, I would consider to use `root` as the owner of this directory. If so, one has to double check that this doesn't break `tedge cert create`.

By the way,  [`tedge cert create`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/lib.rs#L7) is another place where we assume the user running the broker is `mosquitto`. Here it's critical that the generated primary key can only be read by the MQTT broker.
 
What about adding a setting in `/etc/tedge/system.toml` to configure the MQTT broker user and group?

reubenmiller: Yes, I'll change the owner of the `device-certs` directory to be owned by root (to prevent accidental permission changes to the folder (though the files will still be owned by `mosquitto`)).

> What about adding a setting in /etc/tedge/system.toml to configure the MQTT broker user and group?

Yes, having a configurable user in the `tedge cert create` command makes sense, though we'll cover this in a follow issue.

reubenmiller: I've updated to use "root" as the owner of the "device-certs" folder.

didier-wenzek (approved):
Approved

PR #2164, opened 2023-08-17
Write error log if updating entity store fails

Bravo555:
## Proposed changes

Remove `unwrap()`s in the c8y mapper actor that would crash the actor if updating the store failed.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -37,6 +37,7 @@ use tedge_timer_ext::SetTimeout;
 use tedge_timer_ext::Timeout;
 use tedge_utils::file::create_directory_with_defaults;
 use tedge_utils::file::FileError;
+use tracing::error;
 
 const MQTT_ROOT: &str = "te";
 const SYNC_WINDOW: Duration = Duration::from_secs(3);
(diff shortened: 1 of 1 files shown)


crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -108,19 +109,25 @@ impl C8yMapperActor {
...
+                    error!("Could not update device registration: {e}");
+                }
             } else {
                 // if device is unregistered register using auto-registration
                 if self.entity_store.get(entity_id).is_none() {
-                    let register_messages = self.auto_register_entity(entity_id);
+                    let register_messages = match self.auto_register_entity(entity_id) {
+                        Ok(register_messages) => register_messages,
+                        Err(e) => {
+                            error!("Could not update device registration: {e}");
```
PradeepKiruvale, line 126: I feel it's better to write this error message to the `error` topic. @didier-wenzek can clarify this.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -108,19 +109,25 @@ impl C8yMapperActor {
...
-        // to which entity is the message related
         if let Some(entity_id) = entity_store::entity_mqtt_id(&message.topic) {
-            // if message is registration message:
             if is_entity_register_message(&message) {
-                // update entity store with registration message
-                // TODO: emit error message on Err
-                self.entity_store
+                if let Err(e) = self
+                    .entity_store
                     .update(message.clone().try_into().unwrap())
```
PradeepKiruvale, line 116: Need to handle this as well

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -165,8 +175,7 @@ impl C8yMapperActor {
             .with_retain();
```
PradeepKiruvale, line 175: You can use `Topic::new_unchecked("")` to create the topic, no need to worry about `unwrap`.

crates/extensions/c8y_mapper_ext/src/actor.rs
didier-wenzek: To use or not the error topic is a broader topic that needs to be discussed with @reubenmiller .

Thin-edge emits error messages on too many channels. This is puzzling for the users (what streams  do I have to watch?) and the developers (where do I have to log this error?).

Ideally sending errors over MQTT should be done behind the scene by the `error!()` macro.

crates/extensions/c8y_mapper_ext/src/actor.rs
Bravo555: In this case, the message was previously checked to be a register message on line 113, so this unwrap never panics. That's not very obvious to see, so the API will be changed a bit to remove this unwrap in a later PR.

crates/extensions/c8y_mapper_ext/src/actor.rs
Bravo555: okay, looks like `new_unchecked` saves an extra check, so I will be using it from now on

PradeepKiruvale (approved):

PR #2169, opened 2023-08-18
Update c8y-mapper to consume `te` measurements

didier-wenzek:
__The main point of this PR is to implement a concrete use-case of the EntityStore and  be in a better position to evaluate and improve this new API.__

## Proposed changes

Update the c8y-mapper to translate the measurements received on the `te/+/+/+/+/m/+` topics.

- [x] Move the EntityStore from the mapper to its converter
- [x] Subscribe to `te/+/+/+/+/m/+`  and no more on `tedge/measurements/#`
- [x] Combine local device registration with cloud registration
- [ ] ~~Fix the behavior of device registration when the message triggering the registration is ill-formed~~
- [x] List suggestions of improvement for EntityStore: https://github.com/thin-edge/thin-edge.io/issues/2179
 
## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2082

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -10,6 +10,7 @@
 use std::collections::HashMap;
 
 use crate::entity::EntityTopic;
+use crate::entity_store;
 use mqtt_channel::Message;
 use mqtt_channel::Topic;
 
@@ -230,6 +231,64 @@ impl EntityStore {
 
             format!("{main_device_entity_id}:{entity_id_suffix}")
         }
+    }
+
+    /// Performs auto-registration process for an entity under a given
+    /// identifier.
+    ///
+    /// If an entity is a service, its device is also auto-registered if it's
+    /// not already registered.
+    ///
+    /// It returns MQTT register messages for the given entities to be published
+    /// by the mapper, so other components can also be aware of a new device
+    /// being registered.
+    pub fn auto_register_entity(
+        &mut self,
+        entity_id: &str,
+    ) -> Result<Vec<Message>, entity_store::Error> {
+        let mut register_messages = vec![];
+        let (device_id, service_id) = match entity_id.split('/').collect::<Vec<&str>>()[..] {
+            ["device", device_id, "service", service_id, ..] => (device_id, Some(service_id)),
+            ["device", device_id, "", ""] => (device_id, None),
+            _ => return Ok(register_messages),
+        };
+
+        // register device if not registered
+        let device_topic = format!("{MQTT_ROOT}/device/{device_id}//");
+        if self.get(&device_topic).is_none() {
+            let device_type = if device_id == "main" {
+                "device"
+            } else {
+                "child-device"
+            };
+            let device_name = if device_id == "main" {
+                self.main_device_name()
+            } else {
+                device_id
+            };
+            let device_register_payload =
+                format!("{{ \"@type\":\"{device_type}\", \"@id\":\"{device_name}\"}}");
+            let device_register_message =
+                Message::new(&Topic::new(&device_topic).unwrap(), device_register_payload)
+                    .with_retain();
+            register_messages.push(device_register_message.clone());
+            self.update(EntityRegistrationMessage::try_from(&device_register_message).unwrap())?;
+        }
+
+        // register service itself
+        if let Some(service_id) = service_id {
+            let service_topic = format!("{MQTT_ROOT}/device/{device_id}/service/{service_id}");
+            let service_register_payload = r#"{"@type": "service", "type": "systemd"}"#.to_string();
+            let service_register_message = Message::new(
+                &Topic::new(&service_topic).unwrap(),
+                service_register_payload,
+            )
+            .with_retain();
+            register_messages.push(service_register_message.clone());
+            self.update(EntityRegistrationMessage::try_from(&service_register_message).unwrap())?;
+        }
+
+        Ok(register_messages)
     }
 }
 
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -9,6 +9,7 @@
 
 use std::collections::HashMap;
 
+use crate::entity::EntityTopic;
 use mqtt_channel::Message;
 use mqtt_channel::Topic;
 
@@ -98,11 +99,25 @@ impl EntityStore {
         self.get(topic_id)
     }
 
+    /// Returns the entity attached to a topic, if any
+    pub fn get_entity_from_topic(&self, topic: &Topic) -> Option<&EntityMetadata> {
+        let entity_topic = EntityTopic::try_from(topic).ok()?;
+        self.get(entity_topic.entity_id())
+    }
+
     /// Returns the MQTT identifier of the main device.
     ///
     /// The main device is an entity with `@type: "device"`.
     pub fn main_device(&self) -> EntityTopicIdRef {
         self.main_device.as_str()
+    }
+
+    /// Returns the name of main device.
+    pub fn main_device_name(&self) -> &str {
+        self.get(self.main_device.as_str())
+            .unwrap()
+            .entity_id
+            .as_str()
     }
 
     /// Returns MQTT identifiers of child devices of a given device.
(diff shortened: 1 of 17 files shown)


crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -246,15 +160,6 @@ impl C8yMapperActor {
     }
 }
 
-/// Check if a message is an entity registration message.
-fn is_entity_register_message(message: &Message) -> bool {
```
didier-wenzek, line 250: This method has been moved into the Converter.

This will have to be updated after https://github.com/thin-edge/thin-edge.io/pull/2164

crates/extensions/c8y_mapper_ext/src/json.rs
```diff
@@ -30,28 +31,21 @@ pub enum CumulocityJsonError {
...
-    Ok(c8y_vec)
-}
-
-/// Converts from thin-edge Json to c8y_json with child id information
-pub fn from_thin_edge_json_with_child(
+pub fn from_thin_edge_json(
     input: &str,
-    child_id: &str,
+    entity: &EntityMetadata,
 ) -> Result<String, CumulocityJsonError> {
```
didier-wenzek, line 37: The changes in this file highlights how the new EntityStore API can simplify and make more robust the mappers.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1176,7 +1176,7 @@ async fn inventory_registers_unknown_entity_once() {
 
     let measurement_message = MqttMessage::new(
         &Topic::new("te/device/main/service/my_service/m/measurement").unwrap(),
-        "",
+        r#"{"foo":25}"#,
```
didier-wenzek, line 1179: A proper measurements is required otherwise the registration message is rejected by the converter.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -232,44 +241,94 @@ impl CumulocityConverter {
...
+            } else {
+                "child-device"
+            };
+            let device_name = if device_id == "main" {
+                self.entity_store.main_device_name()
+            } else {
+                device_id
+            };
+            let device_register_payload =
+                format!("{{ \"@type\":\"{device_type}\", \"@id\":\"{device_name}\"}}");
```
didier-wenzek, line 279: Compared to the original version https://github.com/thin-edge/thin-edge.io/pull/2169/files#diff-1d1eb076015e0658044f7b03b0b805fce4f0253f4048587797fc5dd42019f908L149

- the device type is now correct for the main device.
- the device name is compatible with the current version of thin-edge (to avoid breaking the tests)

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -232,44 +241,94 @@ impl CumulocityConverter {
...
-        let c8y_json_payload = match maybe_child_id {
-            Some(child_id) => {
-                // Need to check if the input Thin Edge JSON is valid before adding a child ID to list
-                let c8y_json_child_payload =
-                    json::from_thin_edge_json_with_child(input.payload_str()?, child_id.as_str())?;
-                add_external_device_registration_message(
-                    child_id,
-                    &mut self.children,
-                    &mut mqtt_messages,
-                );
```
didier-wenzek, line 254: This has been moved to the `try_convert()` method https://github.com/thin-edge/thin-edge.io/pull/2169/files#diff-1d1eb076015e0658044f7b03b0b805fce4f0253f4048587797fc5dd42019f908L149.

- Pros: The registration can then be done for all kind of messages (measurements/alarms/events)
- Cons: Now an ill-formed message trigger the auto-registration while it should not.
- Because of that the test `convert_first_measurement_invalid_then_valid_with_child_id` must be ignored https://github.com/thin-edge/thin-edge.io/pull/2169/files#diff-08a20a8877f5e1e1bce1aac2c6fc0e1a63a36cc9311e7e60cb3aacacf3c8c57cR1361

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -232,44 +241,94 @@ impl CumulocityConverter {
...
+                .unwrap();
+        }
+
+        register_messages
+    }
+
     fn try_convert_measurement(
         &mut self,
         input: &Message,
     ) -> Result<Vec<Message>, ConversionError> {
```
didier-wenzek, line 310: My goal is to reach a point where this method is no more processing a raw MQTT message, but a Measurement message that has been fully decoded by the `tedge-api` and gives:

- the topic identifier of the source
- the measurement type
- the measurement value

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -796,6 +888,15 @@ impl Converter for CumulocityConverter {
     }
 }
 
+/// Check if a message is an entity registration message.
+fn is_entity_register_message(message: &Message) -> bool {
```
Bravo555, line 892: Could be moved to `entity_store.rs`

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -885,6 +986,13 @@ fn add_external_device_registration_message(
     false
 }
 
+fn external_device_registration_message(entity: &EntityMetadata) -> Option<Message> {
```
Bravo555, line 989: Could be moved to `entity_store.rs`

Bravo555 (changes requested):

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -97,13 +98,27 @@ impl EntityStore {
...
+
     /// Returns the MQTT identifier of the main device.
     ///
     /// The main device is an entity with `@type: "device"`.
     pub fn main_device(&self) -> EntityIdRef {
         self.main_device.as_str()
     }
 
+    /// Returns the name of main device.
+    pub fn main_device_name(&self) -> &str {
```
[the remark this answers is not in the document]
didier-wenzek: > But probably 99% this is what we want, so `main_device` could be changed to just return `EntityMetadata`.

Yes, that would be better. Especially, because `entity_store.get(entity_store.main_device())` returns an `Option` despite there is always some main device.

> One problem is that `EntityMetadata` doesn't contain entity MQTT identifier, but that can perhaps be added later in a way that doesn't just duplicate the string

Yes, this will also be really convenient.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -216,19 +231,41 @@ impl EntityStore {
...
+    pub fn child_device(external_id: String) -> Self {
+        Self {
+            external_id,
+            r#type: EntityType::ChildDevice,
+            parent: Some("device/main//".to_string()),
+            other: serde_json::json!({}),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
didier-wenzek: > I see this functions are only used in tests

Indeed, currently only used for testing. Mostly because creating an `EntityRegistrationMessage` where a bit more painful.

>  Would you say that we'll need to construct `EntityMetadata` in normal code as well? My intent was for only `EntityStore` to be able to construct it, when parsing a registration message, but this can change. 

Not sure to see the benefits of  `EntityRegistrationMessage`.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -682,12 +742,41 @@ impl Converter for CumulocityConverter {
...
+                        self.children
+                            .insert(entity.external_id.to_string(), Operations::default());
+                        registration_messages.push(message);
+                    }
+                }
             }
+        }
+
+        let mut messages = match &message.topic {
+            topic if EntityTopic::is_measurement(topic) => self.try_convert_measurement(message),
```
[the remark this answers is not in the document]
didier-wenzek: Here, I would prefer to have a fully different `match` expression.

Instead of having methods working on raw topics and messages,  my goal is to do all the parsing inside the `tedge-api` crate with some function taking an MqttMessage and returning an error or a plain struct with the message source (e.g. `device/main//`) the channel (e.g. Measurement { r#type })  and the decoded message (e.g the measurement). The converter main method would be then a breeze focused on the translation into the format expected by the cloud and no more a mix of parsing / validation / translation.

crates/core/tedge_mapper/src/collectd/mapper.rs
```diff
@@ -12,7 +12,7 @@ use tedge_config::TEdgeConfig;
 
 const COLLECTD_MAPPER_NAME: &str = "tedge-mapper-collectd";
 const COLLECTD_INPUT_TOPICS: &str = "collectd/#";
-const COLLECTD_OUTPUT_TOPIC: &str = "tedge/measurements";
+const COLLECTD_OUTPUT_TOPIC: &str = "te/device/main///m/ ";
```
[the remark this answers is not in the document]
didier-wenzek: Again, it would be good to get this topic from the entity store.

`entity_store.main_device().measurement_topic()`

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: No, looking closer, this method is c8y related. The point is to register the device on the cloud.

didier-wenzek: The function that should be moved to `entity_store.rs` is `auto_register_entity()`.

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: Moved to `entity_store.rs`

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: > The function that should be moved to `entity_store.rs` is `auto_register_entity()`.

Done

Bravo555 (approved):

PR #2176, opened 2023-08-21
Entity store

Bravo555:
## Proposed changes

- Change `EntityTopic` and `EntityRegistrationMessage` `TryInto` impls to take references instead of values, removing 
needless clones
- Renamed `external_id` to `entity_id` (it was already used in thin-edge documentation as device id or service id) and `mqtt_id` to `mqtt_address` to differentiate it more from `device_id`

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/core/tedge_api/src/entity_store.rs
```diff
@@ -12,19 +12,19 @@ use std::collections::HashMap;
...
+/// Represents an "Entity MQTT address" portion of the MQTT topic
 ///
 /// Example:
-/// - topic: `te/device/dev1/service/myservice/m//my_measurement
+/// - topic: `te/device/dev1/service/myservice/m//my_measurement`
 /// - entity id: `device/dev1/service/myservice`
 ///
 /// <https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#group-identifier>
-type EntityId = String;
+type EntityMqttAddress = String;
```
didier-wenzek, line 22: @Bravo555, @reubenmiller, @albinsuresh 

I'm not fully happy with this proposal for a new name:

Pros:
- `EntityId` is definitely confusing, knowing that *device id* is already used with a [different meaning](https://thin-edge.github.io/thin-edge.io/next/references/mappers/c8y-mapper/#child-device-with-explicit-id): (the main device id is the CN of its certificate while its *entity identifier* is `device/main//`.

Cons:
- The specifications uses the term [*entity identifier*](https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#group-identifier).
- The *address* suffix of `EntityMqttAddress` might be confused with some kind of network address.
- This identifier is used only for MQTT topics. Hence, I would expect the term *topic* rather than *address*

What about `EntityTopicId` or `EntityMqttTopicId` ?

## Alternative

Another way to solve the issue is t:
- keep the terms `EntityId` and entity identifier, when speaking about topics.
- use the terms `EntityName` and entity name, when speaking about the CommonName (CN) of these devices.

## Important

In any case, the terms used in the code base must be the same as those used in the docs.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -161,12 +165,12 @@ impl EntityStore {
             affected_entities.push(parent.clone());
         }
 
-        let external_id = message
-            .external_id
-            .unwrap_or_else(|| self.derive_external_id(&message.mqtt_id));
+        let entity_id = message
+            .entity_id
+            .unwrap_or_else(|| self.derive_entity_id(&message.mqtt_id));
```
didier-wenzek, line 170: This change is really good one, as external id has a specific meaning in the context of c8y.

crates/core/tedge_api/src/entity_store.rs
didier-wenzek: We agreed to use the name `EntityTopicId` for entity identifiers in the new topic scheme.

Issue #2181, opened 2023-08-22
Inherit the dependencies from the workspace


didier-wenzek:
**Is your refactoring request related to a problem? Please describe.**

It's frustrating to update 36 `Cargo.toml` files to update the version used by thin-edge for `tokio`. 

**Describe the solution you'd like**

Let `cargo` manage this by [Inheriting a dependency from a workspace](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace).

- The versions are given in the main `Cargo.toml` 
- All the dependencies in `crates/*/*/Cargo.toml` and `plugins/*/cargo.toml` reference the workspace as in `serde_json.workspace = true` or  `tokio = { workspace = true, features = ["time"]}` where there are crate specific settings.

As there a lot of dependencies, we can start we the main ones: tokio, serde, rumqttc, thiserror, anyhow ...

**Describe alternatives you've considered**

**Additional context**

didier-wenzek:
3 crates still require more work to update their dependencies:

- `tedge`: clap = "3"
- `tedge_http_ext`: mockito = "1.0.2"
- `tedge_config`: toml = "0.7"

didier-wenzek:
Updating the `tedge` dependency on `clap` is required for https://github.com/thin-edge/thin-edge.io/issues/2051

PR #2186, opened 2023-08-24
Inherit the dependencies from the workspace

PradeepKiruvale:
## Proposed changes

It's possible to define the dependency version & features in `workspace Cargo.toml` instead of in every `crates Cargo.toml`.
Now the `crate's toml`  file can inherit the dependency version and feature from the `workspace Cargo.toml` .
For example
```
In `workspace Cargo.toml`

[workspace.dependencies]
serde = "1.0"

In the crate's Cargo.toml

`serde = { workspace =true }`.
```
This will make it easy to update the version.
The `<crate>.workspace = true` syntax is not supported by cargo sort, So, used <crate> = { workspace =true }

This PR only addresses the dependencies that were defined multiple times.

**Todo**

The following instances use different versions of crates for some features, to use the same version, the code needs to be changed.

- Clap version that is used in tedge
- Mockito in tedge-http-ext
- toml in tedge-config

- Define the dependencies that are defined only once

- We can also define the local crate dependencies in the Workspace Cargo.toml and inherit from the crate's cargo.toml.
For example
```
in Workspace Cargo.toml
tedge_actors = { path = "crates/core/tedge_actors" }

and inherit in the project's Cargo.toml as below
```

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2181

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,134 @@ license = "Apache-2.0"
 homepage = "https://thin-edge.io"
 repository = "https://github.com/thin-edge/thin-edge.io"
 
+[workspace.dependencies]
+anyhow =  "1.0"
+assert-json-diff = "2.0"
+assert_cmd = "2.0"
+assert_matches = "1.5"
+async-compat = "0.2.1"
+async-log = "2.0"
+async-trait = "0.1"
+async-tungstenite = { version = "0.18.0", features = [
+    "tokio-runtime",
+    "tokio-rustls-native-certs",
+] }
+atty = "0.2"
+aws_mapper_ext = { path = "crates/extensions/aws_mapper_ext" }
+az_mapper_ext = { path = "crates/extensions/az_mapper_ext" }
+backoff = { version = "0.4", features = ["tokio"] }
+base64 = "0.13"
+batcher = { path = "crates/common/batcher" }
+c8y_api = { path = "crates/core/c8y_api" }
+c8y_config_manager = { path = "crates/extensions/c8y_config_manager" }
+c8y_firmware_manager = { path = "crates/extensions/c8y_firmware_manager" }
+c8y_http_proxy = { path = "crates/extensions/c8y_http_proxy" }
+c8y_log_manager = { path = "crates/extensions/c8y_log_manager" }
+c8y_mapper_ext = { path = "crates/extensions/c8y_mapper_ext" }
+camino = "1.1"
+certificate = { path = "crates/common/certificate" }
+clap = { version = "4.1", features = ["cargo", "derive"] }
+clock = { path = "crates/common/clock" }
+collectd_ext = { path = "crates/extensions/collectd_ext" }
+csv = "1.1"
+darling = "0.20"
+doku = "0.21"
+download = { path = "crates/common/download" }
+easy_reader = "0.5"
+env_logger = "0.10"
+fastrand = "1.8"
+figment = { version = "0.10" }
+filetime = "0.2"
+flockfile = { path = "crates/common/flockfile" }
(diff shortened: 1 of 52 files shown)


crates/common/download/Cargo.toml
```diff
@@ -12,21 +12,21 @@ repository = { workspace = true }
 [dependencies]
 anyhow = { version = "1.0", features = ["backtrace"] }
```
didier-wenzek, line 13:
```suggestion
anyhow = { workspace = true, features = ["backtrace"] }
```

crates/common/log_manager/Cargo.toml
```diff
@@ -11,13 +11,13 @@ repository = { workspace = true }
 [dependencies]
 easy_reader = "0.5"
 glob = "0.3"
```
didier-wenzek, line 13: I would also consider to move the version to the workspace even for those dependencies used only once.

didier-wenzek (commented):
Thank you. This will ease our dependency management.

It would be good to also include in this PR:
- the dependencies that are defined only once
- local crate dependencies

It makes sense to address in specific PRs the following exceptions as the code must be upgraded:

- clap version that is used in tedge
- mockito in tedge-http-ext
- toml in tedge-config

PradeepKiruvale:
> Thank you. This will ease our dependency management.
> 
> It would be good to also include in this PR:
> 
> * the dependencies that are defined only once
> * local crate dependencies
> 
> It makes sense to address in specific PRs the following exceptions as the code must be upgraded:
> 
> * clap version that is used in tedge
> * mockito in tedge-http-ext
> * toml in tedge-config

I agree with the plan. So, I have updated the PR addressing the first two points. i.e the dependencies that are defined once and other local crate dependencies.

crates/common/log_manager/Cargo.toml
PradeepKiruvale: fixed

crates/common/download/Cargo.toml
PradeepKiruvale: fixed

Cargo.toml
```diff
@@ -11,7 +11,7 @@ members = [
     "plugins/c8y_remote_access_plugin",
     "plugins/tedge_apt_plugin",
     "plugins/tedge_dummy_plugin",
-    "plugins/tedge_log_plugin",
+    "plugins/tedge_log_plugin",
```
didier-wenzek, line 14:
```suggestion
    "plugins/tedge_log_plugin",
```

crates/extensions/tedge_log_manager/Cargo.toml
```diff
@@ -10,29 +10,29 @@ homepage = { workspace = true }
...
-anyhow = "1.0"
-assert-json-diff = "2.0"
-filetime = "0.2"
+anyhow = { workspace = true }
+assert-json-diff = { workspace = true }
+filetime = { workspace = true }
 tedge_actors = { path = "../../core/tedge_actors", features = ["test-helpers"] }
 tedge_http_ext = { path = "../../extensions/tedge_http_ext", features = [
     "test_helpers",
 ] }
```
didier-wenzek, line 36: These two have to be updated too.

didier-wenzek (approved):
Approved. Thank you.

crates/extensions/tedge_log_manager/Cargo.toml
PradeepKiruvale: done

Cargo.toml
PradeepKiruvale: fixed

Issue #2131, opened 2023-08-06
disable file locking mechansim when lock directory does not exist or is not writable


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The lock file mechanism used by some thin-edge.io components assumes there is always a `lock` folder under the `/run` path. This implicit requirement prevent the components from being able to start until the folder is created by some other process.

For example, the default `run.path` setting is `/run`, which means lock files are always placed under the `/run/lock/` directory. If the `/run` directory exists, but the `/run/lock` directory does not, then the tedge components which rely on the locking mechanism will fail to start with the following reason:

```text
$ tedge-agent

2023-08-06T07:40:47.166927176Z ERROR flockfile::unix: Lock file path: /run/lock/tedge-agent.lock
Error: Couldn't create file lock.

Caused by:
    No such file or directory (os error 2)
```

In addition, since the `/run` directory is typically a volatile one (e.g. it is not persisted across device reboots), then even if a process were to created the `/run/lock` directory, it would need to create it each time the device is rebooted.

This problem is most notably visible when using thin-edge.io in containers as some images (e.g. alpine linux) do not create a `/run/lock` directory (though `/run` does exist).

**Describe the solution you'd like**

The locking mechanism should be DISABLED in the following situations.

* If the configuration `run.lock_files` is set to `false` (this functionality is already implemented)
* If the `run.path` directory does not exist
* If the `{run.path}/lock` directory does not exist or is not writable

When the locking mechanism is disabled, a log entry (at INFO level) should be added to indicate that the locking mechanism is not active. This would help user's understand the behaviour of the system better.

**Describe alternatives you've considered**

An alternative solution would be to create the `lock` folder if it does not exist, however this would be lead to unpredictable behaviour as the logic to create the folder would have to exist in each tedge component which would lead to race conditions between the components. Whilst such race conditions might not have any adverse functional risks at first, it adds potential future pitfalls should the directory creation procedure change (e.g. change user/group ownership) during any kind of upgrade scenarios.

**Additional context**

rina23q:
The first proposal described under "Describe the solution you'd like" is implemented by #2190.

gligorisaev:
From QA perspective it is running as requested but, @reubenmiller we need to talk about it, I have not understand the point for this change

didier-wenzek:
[...]

PR #2190, opened 2023-08-24
Disable lock file if directory does not exist or no permissions

Ruadhri17:
## Proposed changes
This PR disable lock file mechanism when `{run.path}/lock` does not exist or directory is not writable.

## Types of changes

## Paste Link to the issue
#2131
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -102,12 +102,8 @@ impl Agent {
     pub fn try_new(name: &str, config: AgentConfig) -> Result<Self, FlockfileError> {
         let mut flock = None;
         if config.use_lock {
-            flock = Some(check_another_instance_is_not_running(
-                name,
-                config.run_dir.as_std_path(),
-            )?);
+            flock = check_another_instance_is_not_running(name, config.run_dir.as_std_path())?;
         }
-
         info!("{} starting", &name);
 
         Ok(Self {
--- a/crates/core/tedge_mapper/src/main.rs
+++ b/crates/core/tedge_mapper/src/main.rs
@@ -104,10 +104,7 @@ async fn main() -> anyhow::Result<()> {
     let mut _flock = None;
     if config.run.lock_files {
         let run_dir = config.run.path.as_std_path();
-        _flock = Some(check_another_instance_is_not_running(
-            &mapper_opt.name.to_string(),
-            run_dir,
-        )?);
+        _flock = check_another_instance_is_not_running(&mapper_opt.name.to_string(), run_dir)?;
     }
 
     if mapper_opt.init {
--- a/crates/common/flockfile/src/unix.rs
+++ b/crates/common/flockfile/src/unix.rs
@@ -1,3 +1,5 @@
+use nix::errno::Errno::EACCES;
+use nix::errno::Errno::EAGAIN;
 use nix::fcntl::flock;
 use nix::fcntl::FlockArg;
 use nix::unistd::write;
@@ -11,6 +13,7 @@ use std::path::Path;
 use std::path::PathBuf;
 use tracing::debug;
 use tracing::error;
+use tracing::info;
 use tracing::warn;
 
 const LOCK_CHILD_DIRECTORY: &str = "lock/";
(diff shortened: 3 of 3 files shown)


crates/core/tedge_agent/src/agent.rs
```diff
@@ -101,11 +102,13 @@ pub struct Agent {
...
     pub fn try_new(name: &str, config: AgentConfig) -> Result<Self, FlockfileError> {
         let mut flock = None;
-        if config.use_lock {
+        if config.use_lock && can_create_lock_file(config.run_dir.as_std_path()) {
             flock = Some(check_another_instance_is_not_running(
                 name,
                 config.run_dir.as_std_path(),
             )?);
+        } else {
+            info!("The locking mechanism is not active");
```
PradeepKiruvale, line 111:
```suggestion
            info!("The locking mechanism is not active, so it does not prevent running multiple instances of tedge-agent");
```

crates/core/tedge_mapper/src/main.rs
```diff
@@ -102,12 +104,14 @@ async fn main() -> anyhow::Result<()> {
...
     let mut _flock = None;
-    if config.run.lock_files {
+    if config.run.lock_files && can_create_lock_file(config.run.path.as_std_path()) {
         let run_dir = config.run.path.as_std_path();
         _flock = Some(check_another_instance_is_not_running(
             &mapper_opt.name.to_string(),
             run_dir,
         )?);
+    } else {
+        info!("The locking mechanism is not active");
```
PradeepKiruvale, line 114:
```suggestion
       info!("The locking mechanism is not active, so it does not prevent running multiple instances of tedge-mapper");
```

didier-wenzek (changes requested):
The proposed check is not working: if a user cannot write the lock while another user could, then this user will not be able to run the agent or the mapper.

A better approach is to leverage the error returned by [`Flockfile::new_lock()`](https://github.com/thin-edge/thin-edge.io/blob/3e37b354ed8937e6870ae9094667ffedeb5d414c/crates/common/flockfile/src/unix.rs#L161C18-L161C18), ignoring the error [`if !error.non_exclusive_access()`](https://github.com/thin-edge/thin-edge.io/blob/3e37b354ed8937e6870ae9094667ffedeb5d414c/crates/common/flockfile/src/unix.rs#L45)

didier-wenzek:
> A better approach is to leverage the error returned by [`Flockfile::new_lock()`](https://github.com/thin-edge/thin-edge.io/blob/3e37b354ed8937e6870ae9094667ffedeb5d414c/crates/common/flockfile/src/unix.rs#L161C18-L161C18), ignoring the error [`if !error.non_exclusive_access()`](https://github.com/thin-edge/thin-edge.io/blob/3e37b354ed8937e6870ae9094667ffedeb5d414c/crates/common/flockfile/src/unix.rs#L45)

The fix is a bit more involved as  [`non_exclusive_access()`](https://github.com/thin-edge/thin-edge.io/blob/3e37b354ed8937e6870ae9094667ffedeb5d414c/crates/common/flockfile/src/unix.rs#L45) should check [errno](https://linux.die.net/man/3/lockf) has been set to [`EAGAIN `](https://docs.rs/nix/latest/nix/errno/enum.Errno.html#variant.EAGAIN) or [`EACCES`](https://docs.rs/nix/latest/nix/errno/enum.Errno.html#variant.EACCES).

didier-wenzek (commented):
The code is correct and behaves as described by https://github.com/thin-edge/thin-edge.io/issues/2131

However, the warning message published when locks are disabled could be more useful for the user who intends to enable locks.

```
$ tedge config set run.path /tmp
$ target/debug/tedge-mapper --debug collectd 
2023-08-28T07:34:59.955514889Z  INFO flockfile::unix: The locking mechanism is not active, so it does not prevent running multiple instances of tedge-mapper-collectd.
```

This should tell the user *why* locking mechanism is not active.

```
The locking mechanism is not active, as `/tmp/lock` is not a directory with write access.
```

Might be a bit verbose though. @reubenmiller what's your point of view?

reubenmiller:
> ```
> The locking mechanism is not active, as `/tmp/lock` is not a directory with write access.
> ```

I really like the Didier's suggestion, so please use the statement.

```
The locking mechanism is not active, as `/tmp/lock` is not a directory with write access.
```

Issue #1926, opened 2023-04-24
Supporting filtering criteria in apt-sm-plugin list output


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Depending on your setup the `tedge-apt-plugin` can return a large list of packages, sometimes > 256 packages. The large list can effect other related Cumulocity IoT processes on the server side (e.g. Cumuloity Datahub offboarding of the data to a data lake).

A large number of packages (e.g > 50) generally has very limited value for users as the list is too big to get a good overview over.  Instead people are more interested in selected packages related to their use case (e.g. users care more about their user applications rather than which libc package is installed).

**Describe the solution you'd like**

The `tedge-apt-plugin` should support the following filtering criteria:

* name (regex pattern)
* maintainer (regex pattern)

In addition, the max number of packages should be configured to prevent large number of packages from being sent to the cloud
* Add configuration to set the max number of software packages reported by each software management (sm) plugin (configurable via `tedge.toml`). The limit should be applied to each sm-plugin so that one rogue plugin producing 1000's of entries does not bump the output of other sm-plugins.

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
Maybe switching from using the `apt list --installed` command, we could use the `dpkg-query` command. For example you can list the package, version and maintainer using:

```
dpkg-query -f '${Package}\t${Version}\t${Maintainer}\n' -W
```

Though more investigation is needed to see if it returns the appropriate data.

Ruadhri17:
We can replace `apt list --installed` with slight modification (`dpkg-query` does not support the `--installed` flag, but we can check package status and then `grep` all that are `installed` but print them without status information). We will also need to use grep when filtering by maintainers as `dpkg-query` filters only by names. However, I have a few questions: 

* Should we apply a maximum number of displayed packages when using a filter?  
* Should we set a maximum number of displayed packages by default? If so, what should be a default number?
* How do we want to provide filtering options? Initially, I wanted to add two flags: `--name` and `--maintainer`, but maybe you have a better suggestion.

reubenmiller:
> Should we apply a maximum number of displayed packages when using a filter?

The maximum number of items should be applied by the tedge component that runs the plugin filters (not the plugin itself), as this will be common logic which can be applied across different software management plugins.

> Should we set a maximum number of displayed packages by default? If so, what should be a default number?

Use a default of 1000 items. This limit should always be applied to the software management list.

[...]

PR #2198, opened 2023-08-29
Add test for sm plugin max packages (and fixes)

gligorisaev:
## Proposed changes

Adding test case that validates the max packages of sw plugin

* Resolves max packages problem. #2200
* Resolves unstable sorting of sm-plugin modules. #2199 

## Types of changes

## Paste Link to the issue

#1926
#2199
#2200

## Checklist

### DIFF
--- a/crates/core/plugin_sm/src/plugin.rs
+++ b/crates/core/plugin_sm/src/plugin.rs
@@ -445,27 +445,28 @@ impl Plugin for ExternalPluginCommand {
         let command = self.command(LIST, None)?;
         let output = self.execute(command, logger).await?;
         if output.status.success() {
-            if self.max_packages == 0 {
-                Ok(deserialize_module_info(
-                    self.name.clone(),
-                    output.stdout.as_slice(),
-                )?)
-            } else {
-                let (last_line, _) = String::from_utf8(output.stdout.as_slice().to_vec())
+            // If max_packages is set to an invalid value, use 0 which represents all
+            // of the content, don't bother filtering the content when all of it will
+            // be included anyway
+            let max_packages = usize::try_from(self.max_packages).unwrap_or(0);
+            let last_char = match max_packages {
+                0 => 0,
+                _ => String::from_utf8(output.stdout.as_slice().to_vec())
                     .unwrap_or_default()
                     .char_indices()
                     .filter(|(_, c)| *c == '\n')
-                    .nth(usize::try_from(self.max_packages).unwrap_or(usize::MAX))
-                    .unwrap_or_default();
+                    .nth(max_packages - 1)
+                    .map(|(i, _)| i)
+                    .unwrap_or_default(),
+            };
 
-                Ok(deserialize_module_info(
-                    self.name.clone(),
-                    match last_line {
-                        0 => output.stdout.as_slice(),
-                        _ => &output.stdout[..=last_line],
-                    },
-                )?)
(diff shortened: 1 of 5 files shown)


reubenmiller:
> dummy1-0006

This test failure is due to this bug, https://github.com/thin-edge/thin-edge.io/issues/2200

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -462,7 +462,7 @@ impl Plugin for ExternalPluginCommand {
                     self.name.clone(),
                     match last_line {
                         0 => output.stdout.as_slice(),
-                        _ => &output.stdout[..=last_line],
+                        _ => &output.stdout[..=last_line - 1],
```
Ruadhri17, line 465: The limitation of `max_packages -1` must be done here.

If we applied it there `.nth(usize::try_from(self.max_packages - 1).unwrap_or(usize::MAX))` there would be one case that doesn't work: If user set `max_packages = 1` then `last_line` would be 0 and we would return whole list instead of one package.

Ruadhri17 (approved):
One issue regarding `max_packages`, others approved.

crates/core/plugin_sm/src/plugin.rs
reubenmiller: Actually this didn't work as expected because the variable name is misleading as it should be `last_char` and not `last_line` as the it is taking a slice of the output which is just a vector of characters (not a vector of lines).

I reworked the example to make it a bit more clear, and there were some other minor refactoring to make the code a bit more readable.

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -451,18 +451,26 @@ impl Plugin for ExternalPluginCommand {
...
+                        .unwrap_or_default(),
+                };
 
                 Ok(deserialize_module_info(
                     self.name.clone(),
-                    match last_line {
-                        0 => output.stdout.as_slice(),
-                        _ => &output.stdout[..=last_line],
+                    match last_char {
+                        0 => &output.stdout[..],
```
reubenmiller, line 472: This is purely a cosmetic change to make it easier for users to compare the two match branches.

Ruadhri17 (approved):
approved

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -451,18 +451,26 @@ impl Plugin for ExternalPluginCommand {
...
-                    .unwrap_or_default()
-                    .char_indices()
-                    .filter(|(_, c)| *c == '\n')
-                    .nth(usize::try_from(self.max_packages).unwrap_or(usize::MAX))
-                    .unwrap_or_default();
+                // If max_packages is set to an invalid value, use 0 which represents all
+                // of the content, don't bother filtering the content when all of it will
+                // be included anyway
+                let max_packages = usize::try_from(self.max_packages).unwrap_or(0);
+                let last_char = match max_packages {
```
Ruadhri17, line 458: This part of code will now cover all cases so we can delete that if statement:
```rust
            if self.max_packages == 0 {
                Ok(deserialize_module_info(
                    self.name.clone(),
                    output.stdout.as_slice(),
                )?)
            } else {
```

reubenmiller: Removed the if statement.

Issue #2208, opened 2023-09-01
Fix `download` crate unit tests that fail or do not terminate due to lack of disk space on Github CI runner


Bravo555:
**TODO**

- [x] try to reproduce the issue locally
- [x] fix tests not terminating because of no disk space
- [x] find out how to reduce amount of disk space used on Github CI runner

**Describe the bug**

Some of the unit tests of the download crate fail or do not terminate, as can be seen in this CI run:

https://github.com/thin-edge/thin-edge.io/actions/runs/6026288086/job/16348875869

For now, failing tests were ignored in https://github.com/thin-edge/thin-edge.io/commit/3cf64981878ee3c65063b07631023ce3eda5f47e, and will be re-enabled once the issue has been fixed.

There are 2 problems to be solved:

- [x] ~Resolve lack of disk space on the CI runner, which is a problem in itself and could potentially cause other tests to fail as well~ fixed in #2217
- [x] ~Figure out why some tests do not terminate and make them terminate with a failure~ fixed in #2210 

**To Reproduce**

To reproduce the issue, one needs to re-enable ignored tests and run `pull-request-checks` CI workflow in this very repository. I tried running this workflow on my fork, but for some reason the amount of space used was lower, so writes did not fail (65GB used here vs 72GB used by the main repository workflow):

https://github.com/Bravo555/thin-edge.io/actions/runs/6035002974/job/16374496858

This can also be a clue how to reclaim disk space used in the main repository.

The issue can also be easily reproduced locally by filling up your disk:

1. remove `#[ignore ...]` directives from affected tests
1. run `cargo llvm-cov -p download` to generate artifacts
2. run `dh -h` to see how much free space there is in the filesystem
3. run `fallocate -l SIZE big_file` where `SIZE` is available space on your filesystem
4. run `cargo llvm-cov -p download` again
5. some of the tests should fail and some should not terminate

**Expected behavior**

The target state after resolution of this issue is the following:

1. fix the tests that do not terminate
6. reclaim enough disk space so that we're not running into any write errors
7. have some sort of way to monitor resource usage by the runners so that we can more quickly find issues caused by lack of resources on the CI runner
8. determine if Github hosted runners are enough for us, and whether or not should we set up self-hosted runners.

didier-wenzek:
I would be also sure that downloads don't block for ever when there is not enough space left. In other words, I want to be sure that only the tests are buggy.

[...]

PR #2210, opened 2023-09-01
Fix `download` crate unit tests not terminating

Bravo555:
## Proposed changes

This PR fixes `download` crate unit tests which didn't terminate when lack of disk space caused a failure.

The tests did not terminate because previously used version of mockito (0.31) had an issue where a destructor for objects created in spawned task would sometimes hang. It caused the tests to hang as well because when returning `Result` from a function (instead of `panic!`ing) the runtime would wait for all tasks to finish.

The issue was solved by upgrading to the latest version of mockito and rewriting tests to use panics. This version was already used in another crate, so this change also removes a duplicate version so that we use only 1 version of mockito.

After the upgrade to the latest mockito version, a test that checked `Content-Length` header stopped working because after mockito upgrade the `Content-Length` no longer overrides `Content-Length` derived from the response body. This appears to be a regression in the library and will be addressed later by submitting a PR to mockito.

The tests still fail because of lack disk space. This will be addressed in another PR.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2208

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1196,7 +1196,7 @@ dependencies = [
  "anyhow",
  "backoff",
  "log",
- "mockito 0.31.1",
+ "mockito",
  "nix",
  "regex",
  "reqwest",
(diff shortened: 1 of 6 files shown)


didier-wenzek (commented):
Approved.

As said by @Bravo555, one still needs to address:
- the issue with mockito when overriding Content-Length
- the CI disk space issue

didier-wenzek (approved):

Issue #2112, opened 2023-07-28
c8y-remote-access-plugin should support the --config-dir flag to allow for custom config directories


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When using a custom tedge configuration directory, `c8y-remote-access-plugin` does not support running the `--init` option using the custom location.

For example if the user wants to create a custom folder to store the tedge configuration in (e.g. `/custom`), then they would create the initial folder structure using the new `tedge init` command included in 0.12.0:

```sh
tedge --config-dir /custom init
```

However, when trying to init `c8y-remote-access-plugin`, it does not support the `--config-dir` flag, meaning that it will create a file under `/etc/tedge/operations/c8y/c8y_RemoteAccessConnect` instead of the custom directory.

```sh
c8y-remote-access-plugin --init
```

**Describe the solution you'd like**

* `c8y-remote-access-plugin` should support the `--config-dir` flag to be able to specify the base tedge configuration directory (similar to the behaviour of the other tedge binaries, e.g. `tedge-agent`, `c8y-log-plugin`, `c8y-configuration-plugin` etc.).

**Describe alternatives you've considered**

**Additional context**

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the bug and could not reproduce it anymore.

PR #2211, opened 2023-09-01
Add support for `config-dir` flag

Ruadhri17:
## Proposed changes
This PR add support for `config-dir` flag in c8y-remote-access-plugin

## Types of changes

## Paste Link to the issue
#2112
<br/>

## Checklist

### DIFF
--- a/plugins/c8y_remote_access_plugin/src/input.rs
+++ b/plugins/c8y_remote_access_plugin/src/input.rs
@@ -4,6 +4,9 @@ use miette::ensure;
 use miette::miette;
 use miette::Context;
 use serde::Deserialize;
+use std::path::PathBuf;
+use tedge_config::TEdgeConfigLocation;
+use tedge_config::DEFAULT_TEDGE_CONFIG_PATH;
 
 use crate::csv::deserialize_csv_record;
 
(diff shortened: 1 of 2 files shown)


Ruadhri17:
The main issue of ticket #2112 is fixed, however I am thinking about two other modifications:

1. This is the only place in our code when we use `miette` instead of `anyhow`. Do we want to replace this package to make plugin more consistent with thin-edge code?
2. This plugin does not parse subcommands in cli, instead only flags are used. Some time ago we replaced `--init` flag with `init` subcommand. Do we want to replace it here? Again, it is also about consistency, though we would have to replace all flags with subcommands.

@didier-wenzek @reubenmiller what do you think?

didier-wenzek:
> 1. This is the only place in our code when we use `miette` instead of `anyhow`. Do we want to replace this package to make plugin more consistent with thin-edge code?

I'm okay to have this crate using `miette` while the others are using `anyhow`.  We will have to reduce the number of dependencies, sure. However, I would prefer to do that in a second step, the first one being to reduce the number of executables.

didier-wenzek (approved):
Approved.

The small inconsistency with other tegde cli tools is okay as this plugin is not launched by the user but the c8y mapper.

PR #2220, opened 2023-09-05
Fix panic in tedge_config macro when `-` present in field name

Bravo555:
## Proposed changes

`#[tedge_config(rename = "name-with-hyphens")]` panicked because of hyphens in the rename field. This occured because a test that was being generated to test that field did not remove hyphens from the identifier name, e.g:

> error: proc macro panicked
> ...
>     = help: message:
`"example_value_can_be_deserialized_for_run_name-with-hyphens_example_0"` is not a valid identifier

Was fixed by replacing hyphens with underscores in identifier name.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config_macros/impl/src/lib.rs
+++ b/crates/common/tedge_config_macros/impl/src/lib.rs
@@ -43,7 +43,7 @@ pub fn generate_configuration(tokens: TokenStream) -> Result<TokenStream, syn::E
             field.examples.iter().enumerate().map(move |(n, example)| {
                 let name = quote::format_ident!(
                     "example_value_can_be_deserialized_for_{}_example_{n}",
-                    key.join("_")
+                    key.join("_").replace('-', "_")
                 );
                 let span = example.span();
                 let example = example.as_ref();
@@ -187,5 +187,19 @@ mod tests {
             },
         })
         .is_err());
+    }
+
+    #[test]
+    fn can_contain_hyphen_separated_fields() {
+        generate_configuration(quote! {
+            device: {
+                #[tedge_config(rename = "type")]
+                ty: String,
+
+                #[tedge_config(rename = "hyphen-separated-field", example = "hsf")]
+                hyphen_separated_field: String
+            },
+        })
+        .unwrap();
     }
 }


crates/common/tedge_config_macros/impl/src/lib.rs
```diff
@@ -188,4 +188,18 @@ mod tests {
...
     }
+
+    #[test]
+    fn can_contain_hyphen_separated_fields() {
+        generate_configuration(quote! {
+            device: {
+                #[tedge_config(rename = "type")]
+                ty: String,
+
+                #[tedge_config(rename = "hyphen-separated-field", example = "hsf")]
```
Bravo555, line 199: One thing I noticed is that the panic was not triggered with `#[tedge_config(rename = "hyphen-separated-field")]` and I also had to add `example = hsf` in the macro. Is this intended behaviour?

jarhodes314: Yes, the generated tests exist to verify the examples are valid, so there is no test if no example is provided, and if there was a second example two tests would be generated.

jarhodes314 (approved):
This looks good to me. Out of interest, what are you intending to use hyphens for? I'm just wondering as we use underscores elsewhere in tedge.toml, and using a different casing style may be confusing.

reubenmiller:
> This looks good to me. Out of interest, what are you intending to use hyphens for? I'm just wondering as we use underscores elsewhere in tedge.toml, and using a different casing style may be confusing.

Yes I would not advise to use hypens as this makes translating the configuration name to the environment variable equivalent more complicated as environment variables should only match the following regex (for compatibility): `^[A-Z][A-Z0-9]+$`

Bravo555:
> Out of interest, what are you intending to use hyphens for? I'm just wondering as we use underscores elsewhere in tedge.toml, and using a different casing style may be confusing.

I was just testing how the config options would look like with hyphens, and stumbled on this bug. I don't intend to use them for any configuration options.

jarhodes314:
> > This looks good to me. Out of interest, what are you intending to use hyphens for? I'm just wondering as we use underscores elsewhere in tedge.toml, and using a different casing style may be confusing.
> 
> Yes I would not advise to use hypens as this makes translating the configuration name to the environment variable equivalent more complicated as environment variables should only match the following regex (for compatibility): `^[A-Z][A-Z0-9_]*$`

Obviously this bug was never the right behaviour, but I'm wondering whether characters that can't be (widely) used in environment variable names should be explicitly rejected with an appropriate compiler error explaining why we don't support them, in case someone accidentally uses incorrect casing in the future.

Issue #2096, opened 2023-07-25
Custom restart should only use sudo if sudo is installed


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The custom restart command (defined in the `system.toml` file), always assumes that `sudo` needs to be used when calling the custom command. When using a container setup, this is generally not the case as certain privileged functions are available to a non-root user (e.g. restarting the container when using s6-overlay is allowed by a non root user.)

Ideally `sudo` should be an optional dependency, and only called if it is actually installed.

Below shows a snippet of the log entry when sudo is not installed and the restart manager tries to execute the script (notice it doesn't actually say that sudo is missing, but it was definitely caused by the missing sudo as shown in the rust snippet following the example).

```log
tedge-agent | 2023-07-25T19:56:47.185282174Z ERROR tedge_agent::restart_manager::actor: No such file or directory (os error 2)
```

The rust source code shows the problem with the reliance on `sudo`, and it blindly uses it without any prior checks if it exists or not.

*File: crates/core/tedge_agent/src/restart_manager/actor.rs*
```rust
let mut command = Command::new(SUDO);
command.args(system_config.system.reboot);
vec.push(command);
Ok(vec)
```

**Describe the solution you'd like**

* Only use sudo to run specific commands if it is installed. If `sudo` isn't installed just try running the command without it, and pass the error reason back to the user.

**Describe alternatives you've considered**

**Additional context**

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the bug and could not reproduce it anymore.

PR #2224, opened 2023-09-06
Check if sudo is available when using tedge agent operations

Ruadhri17:
## Proposed changes
This PR add improvements to `tedge agent` modules that run `command`. They run the `sudo` command only if it is available. If it is not, module try to run command without `sudo` and pass error if occurred. More in #2096 

## Types of changes

## Paste Link to the issue
* #2096 
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/restart_manager/actor.rs
+++ b/crates/core/tedge_agent/src/restart_manager/actor.rs
@@ -24,6 +24,7 @@ use tokio::process::Command;
 use tokio::time::timeout;
 use tracing::error;
 use tracing::info;
+use which::which;
 
 const SYNC: &str = "sync";
 
@@ -206,17 +207,34 @@ impl RestartManagerActor {
 
     async fn get_restart_operation_commands(&self) -> Result<Vec<Command>, RestartManagerError> {
         let mut vec = vec![];
-        // sync first
-        let mut sync_command = Command::new(SUDO);
-        sync_command.arg(SYNC);
-        vec.push(sync_command);
 
         // reading `config_dir` to get the restart command or defaulting to `["init", "6"]'
         let system_config = SystemConfig::try_new(&self.config.config_dir)?;
 
-        let mut command = Command::new(SUDO);
-        command.args(system_config.system.reboot);
-        vec.push(command);
+        // Check if sudo command is available
+        match which(SUDO) {
+            Ok(sudo) => {
+                let mut sync_command = Command::new(&sudo);
+                sync_command.arg(SYNC);
+                vec.push(sync_command);
+
+                let mut command = Command::new(sudo);
+                command.args(system_config.system.reboot);
+                vec.push(command);
+            }
+            Err(_) => {
+                let sync_command = Command::new(SYNC);
+                vec.push(sync_command);
+
+                let mut args = system_config.system.reboot.iter();
+
+                let mut command =
+                    Command::new(args.next().map(|s| s.to_string()).unwrap_or_default());
+                command.args(args);
+                vec.push(command);
+            }
+        }
+
         Ok(vec)
     }
 }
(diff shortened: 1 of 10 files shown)


crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -206,17 +208,38 @@ impl RestartManagerActor {
...
+                vec.push(sync_command);
+
+                let mut args = system_config.system.reboot.iter();
+
+                let mut command =
+                    Command::new(args.next().map(|s| s.to_string()).unwrap_or_default());
+                command.args(args);
+                vec.push(command);
+            }
+            _ => {
```
rina23q, line 232: This conditional arm includes the case when command `sudo` returns an error except NotFound. But the following code uses `sudo`, which sounds not correct to me.

I would remove `if err.kind() == ErrorKind::NotFound` and just not using `sudo` if any error occurs after execute `sudo`.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -77,10 +81,19 @@ impl Actor for SoftwareManagerActor {
         let operation_logs = OperationLogs::try_new(self.config.log_dir.clone().into())
             .map_err(SoftwareManagerError::FromOperationsLogs)?;
 
+        let sudo: Option<PathBuf> = match Command::new(SUDO)
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .spawn()
+        {
+            Err(err) if err.kind() == NotFound => None,
```
rina23q, line 89: The same as in `restart_manager/actor.rs`. The next conditional arm includes the case when command sudo returns an error except NotFound. But the following code uses sudo, which sounds not correct to me.

I would remove if err.kind() == ErrorKind::NotFound and just not using sudo if any error occurs after execute sudo.

rina23q (commented):
Small comments!

rina23q (approved):
Thanks! LGTM!

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -206,17 +207,38 @@ impl RestartManagerActor {
...
         let system_config = SystemConfig::try_new(&self.config.config_dir)?;
 
-        let mut command = Command::new(SUDO);
-        command.args(system_config.system.reboot);
-        vec.push(command);
+        // Check if sudo command is available
+        match Command::new(SUDO)
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .spawn()
```
reubenmiller, line 218: Maybe using the [which crate](https://docs.rs/which/latest/which/) would be a more simple approach which does not involve executing `sudo`. We are already using `which` in the `tedge connect` command.

reubenmiller:
I added some integration tests to ensure coverage

crates/core/tedge_agent/src/restart_manager/actor.rs
reubenmiller: @didier-wenzek What do you think about using `which` instead of calling sudo to test its existence?

didier-wenzek: I agree with @reubenmiller that using the [which crate](https://docs.rs/which/latest/which/) would less brittle.

didier-wenzek: I would even go a step further by adding a `get_sudo_path() -> Option<PathBuf>` function in `tedge_utils`.

Ruadhri17: Replaced with `which` crate

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -77,10 +79,15 @@ impl Actor for SoftwareManagerActor {
         let operation_logs = OperationLogs::try_new(self.config.log_dir.clone().into())
             .map_err(SoftwareManagerError::FromOperationsLogs)?;
 
+        let sudo: Option<PathBuf> = match which(SUDO) {
+            Ok(_) => Some(SUDO.into()),
+            Err(_) => None,
+        };
+
```
didier-wenzek, line 86: Use the path returned by `which`.

```suggestion
        let sudo = which(SUDO).ok();
```

didier-wenzek (approved):
Approved

Issue #2222, opened 2023-09-06
c8y-log-plugin fails to upload binary due to 401 error


reubenmiller:
**Describe the bug**

There is an intermittent error occurring when processing the Cumulocity IoT Log File Request operation, which results in the operation sometimes failing due to a HTTP 401 error (e.g. permission denied) when the log file is uploaded via HTTP to Cumulocity IoT.

In this error case, the following is shown in the "failureReason" property of the operation in Cumulocity IoT.
```
Handling of operation failed with Failed with HTTP error status 401 Unauthorized
```

However, subsequent log file request operations then work. It is unclear whether a 401 error triggers a token refresh, or the `c8y-log-plugin` service is not waiting for the requested token to be returned (or it receives an expired one).

**To Reproduce**

The exact steps have not been confirmed but I suspect it would be something like:

1. Start thin-edge.io and all of its components (including the `c8y-log-plugin`)
2. Leave the system running for over 1 hour
3. Request a log file via Cumulocity IoT. The operation should be successful. This is where I suspect the operation is failing (at least in some situations)
4. Request a second log file via Cumulocity IoT. Check if the second request passes or not

**Expected behavior**

The log file upload should be retried automatically by the component which is uploading the file to Cumulocity IoT. If a new token is required then it should fetch one. Relying on the user to resend the operation is results in a very negative user experience.

**Screenshots**

The journalctl logs were collected using the following commands:

```sh
journalctl -f -u "c8y-*" -u "tedge-*"
```

**Log Output**

```log
Sep 06 10:41:31 lumber tedge-mapper-template[577]: 2023-09-06T10:41:31+02:00 INF Publishing new message. topic=c8y/s/us message=503,c8y_SoftwareUpdate delay=0
Sep 06 10:57:17 lumber c8y-log-plugin[5980]: 2023-09-06T08:57:17.247352549Z  INFO c8y_log_manager::actor: Log request received on topic: c8y/s/ds
Sep 06 10:57:17 lumber c8y-log-plugin[5980]: 2023-09-06T08:57:17.445391887Z  INFO mqtt_channel::connection: MQTT connection established
Sep 06 10:57:17 lumber c8y-log-plugin[5980]: 2023-09-06T08:57:17.467384337Z  INFO c8y_api::http_proxy: JWT token requested
Sep 06 10:57:17 lumber tedge-mapper-template[577]: 2023-09-06T10:57:17+02:00 INF Received new c8y token len=790
[...]

PR #2226, opened 2023-09-06
c8y-log-plugin fails to upload binary due to 401 error

PradeepKiruvale:
## Proposed changes

While retrying the HTTP call on the expired jwt token, use the freshly retrieved token for authentication in the subsequent HTTP calls.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2222

## Checklist

### DIFF
--- a/crates/extensions/c8y_http_proxy/src/actor.rs
+++ b/crates/extensions/c8y_http_proxy/src/actor.rs
@@ -46,7 +46,7 @@ use tedge_http_ext::HttpResult;
 const RETRY_TIMEOUT_SECS: u64 = 20;
 
 pub struct C8YHttpProxyActor {
-    end_point: C8yEndPoint,
+    pub(crate) end_point: C8yEndPoint,
     peers: C8YHttpProxyMessageBox,
 }
 
@@ -182,7 +182,10 @@ impl C8YHttpProxyActor {
         Ok(())
     }
 
-    async fn try_get_internal_id(&mut self, device_id: String) -> Result<String, C8YRestError> {
+    pub(crate) async fn try_get_internal_id(
+        &mut self,
+        device_id: String,
+    ) -> Result<String, C8YRestError> {
         let url_get_id: String = self.end_point.get_url_for_internal_id(device_id);
         if self.end_point.token.is_none() {
             self.get_fresh_token().await?;
@@ -190,6 +193,7 @@ impl C8YHttpProxyActor {
         let request = HttpRequestBuilder::get(&url_get_id)
             .bearer_auth(self.end_point.token.clone().unwrap_or_default())
             .build()?;
+
         let res = match self.peers.http.await_response(request).await? {
             Ok(response) => match response.status() {
                 StatusCode::OK => Ok(Ok(response)),
@@ -224,6 +228,7 @@ impl C8YHttpProxyActor {
         let request = request_builder?
             .bearer_auth(self.end_point.token.clone().unwrap_or_default())
             .build()?;
+
         let resp = self.peers.http.await_response(request).await?;
         match resp {
             Ok(response) => match response.status() {
@@ -257,7 +262,9 @@ impl C8YHttpProxyActor {
         self.get_fresh_token().await?;
         // build the request
         let request_builder = build_request(&self.end_point);
-        let request = request_builder?.build()?;
+        let request = request_builder?
+            .bearer_auth(self.end_point.token.clone().unwrap_or_default())
+            .build()?;
         // retry the request
         Ok(self.peers.http.await_response(request).await?)
     }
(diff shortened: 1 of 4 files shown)


crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -257,7 +257,9 @@ impl C8YHttpProxyActor {
         self.get_fresh_token().await?;
         // build the request
         let request_builder = build_request(&self.end_point);
-        let request = request_builder?.build()?;
+        let request = request_builder?
+            .bearer_auth(self.end_point.token.clone().unwrap_or_default())
+            .build()?;
```
didier-wenzek, line 262: If  `self.get_fresh_token().await?` where returning a `&Token` explicitly passed to the build request method, then things would be simpler to get correct:

```
let token = self.get_fresh_token().await?;
let request_builder = build_request(&self.end_point, token);
let request = request_builder?.build()?;
```

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -46,7 +46,7 @@ use tedge_http_ext::HttpResult;
 const RETRY_TIMEOUT_SECS: u64 = 20;
 
 pub struct C8YHttpProxyActor {
-    end_point: C8yEndPoint,
+    pub end_point: C8yEndPoint,
```
didier-wenzek, line 49: I would avoid to make one field public just for some unit test.

crates/extensions/c8y_http_proxy/src/tests.rs
```diff
@@ -127,6 +133,58 @@ async fn retry_internal_id_on_expired_jwt() {
...
+    let c8y_proxy_actor = C8YHttpProxyBuilder::new(config, &mut http_actor, &mut jwt);
+    let mut jwt_actor = ServerActor::new(DynamicJwtRetriever { count: 0 }, jwt.build());
+
+    tokio::spawn(async move { http_actor.run().await });
+    tokio::spawn(async move { jwt_actor.run().await });
+    let mut proxy = c8y_proxy_actor.build();
+
+    target_url.push_str("/identity/externalIds/c8y_Serial/device-001");
+    let result = proxy.try_get_internal_id(device_id.into()).await;
+    assert_eq!(device_id, result.unwrap());
```
didier-wenzek, line 184: Isn't `try_get_internal_id` supposed to return an external_device_id and not a device_id ?

I was expecting the following:

```suggestion
    let result = proxy.try_get_internal_id(device_id.into()).await;
    assert_eq!(external_id, result.unwrap());
```

This test being green, it raises more concerns than it gives confidence.

didier-wenzek (changes requested):
I don't understand how the test can be green.

crates/extensions/c8y_http_proxy/src/tests.rs
reubenmiller: If this is is related to the events REST api then it is the Cumulocity global id (e.g. internal id, not the external identity; example 12345). The REST api users the global id exclusively (except for  the external identity loop up api).

reubenmiller: ah now I see the problem. The assertion is comparing the `device_id` with the result of the `try_get_internal_id` which should not be equal.

PradeepKiruvale: I wrongly populated the struct `InternalIdResponse::new(internal_id,external_id)`.

crates/extensions/c8y_http_proxy/src/tests.rs
```diff
@@ -135,11 +135,11 @@ async fn retry_internal_id_on_expired_jwt() {
...
 #[tokio::test]
 async fn retry_internal_id_on_expired_jwt_with_mock() {
-    let device_id = "device-001";
+    let external_id = "device-001";
     let tmp_dir = "/tmp";
-    let external_id = "external-device-001";
+    let internal_id = "external-device-001";
 
-    let response = InternalIdResponse::new(device_id, external_id);
+    let response = InternalIdResponse::new(internal_id,external_id);
```
didier-wenzek: This is really easy to mess-up, as this has been done for this test.

- The names are misleading: the device_id expected here in the internal device id (internal from the c8y perspective), while device_id usually means the device certificate CN in the context of thin-edge (external if from the c8y perspective).
- The types should not be Strings but specific types wrapping strings.

This will have to be done in a different PR.

crates/extensions/c8y_http_proxy/src/tests.rs
```diff
@@ -127,6 +138,126 @@ async fn retry_internal_id_on_expired_jwt() {
     .await;
 }
 
+#[tokio::test]
+async fn retry_internal_id_on_expired_jwt_with_mock() {
+    let external_id = "device-001";
+    let tmp_dir = "/tmp";
+    let internal_id = "external-device-001";
```
didier-wenzek, line 145: The value is confusing. Is this the internal or the external id?

didier-wenzek (approved):
Approved.

This code fix the bug but is still really fragile. This will have to be addressed as a follow task of https://github.com/thin-edge/thin-edge.io/pull/2239

Issue #2018, opened 2023-06-09
Support running tedge-agent on a child device


reubenmiller:
**Is your feature request related to a problem? Please describe.**

The `tedge-agent` provides a nice software management plugin interface. It would be useful if the `tedge-agent` could be modified to allow it to run both on the main device and on/as a child device. This would enable users to get out-of-the-box functionality for child devices with zero coding! In the future it could also be extended to the other plugins to support reusing all of the existing components as either a main device or a child device.

**Describe the solution you'd like**

There are two areas that will need addressing to enable this feature.

**Area 1: Changes to the tedge-agent to support different runtime roles**

* Not all features of the `tedge-agent` make sense running when running as a child device, for example the http server should not be started when running as a child device.
    * Expected functionality when running as a child
        * Software management
        * Restart operation
    * Disable functionality when running as a child
        * http service
* The MQTT client id should be configurable so that the client id does not conflict with the `tedge-agent` running on the main device
* The MQTT topics used by the `tedge-agent` should be configurable to use the appropriate topics depending on its current role (e.g. running as a main device or as a child device)

**Area 2: Changes to the tedge-mapper to support sending commands to child devices**
* Change the `tedge-mapper` to support routing the operations for child devices to `tedge/` topics. For example `tedge/child/{childid}/commands/...`. The specifics of the topic naming is dependent on #2005

**Notes**

* The [tedge-mapper-template](https://github.com/reubenmiller/tedge-mapper-template) PoC could be used to try out possible new MQTT topic structures before the changes are done to the `tedge-mapper`

**Describe alternatives you've considered**

**Additional context**

An [PoC](https://github.com/thin-edge/thin-edge.io/compare/main...reubenmiller:thin-edge.io:exp-topic-prefix) has already been created to show how the MQTT client id and the topics could be made configurable. A configurable topic prefix is used to control whether the `tedge-agent` should behave as a main device or a child device.

reubenmiller:
@didier-wenzek Can you please update the status of this ticket?

didier-wenzek:
Child-support has been implemented by https://github.com/thin-edge/thin-edge.io/pull/2307.
The agent documentation has been updated: https://github.com/thin-edge/thin-edge.io/pull/2235:
- [x] How to launch tedge-agent on the main device as well as on a child device
- [x] Tedge MQTT v1 protocol used by the agent for `restart` of device (main or child)
[...]

PR #2231, opened 2023-09-07
Add `mqtt.topic_root` and `mqtt.device_topic_id` options

Bravo555:
## Proposed changes

Settings `mqtt.topic_root` have been `mqtt.device_topic_id` added to tedge-config and used inside tedge-agent, enabling the agent to be run on a child device and connect to a parent's MQTT broker.
Also `--mqtt-topic-root` and `--device-topic-id` flags have been added to `tedge-agent`, which can be used to override these values if taken from a configuration file.
`tedge-log-manager` was updated to use new `mqtt.topic_root` and `mqtt.device_topic_id` config options, still allowing them to be overridden by cli flags. The naming of the flags was made consistent with `tedge-agent` and some code was cleaned up.

## Types of changes

## Paste Link to the issue
#2018

## Checklist

## Further comments

This PR completes the _Make tedge-agent running on any device_ section of [this TODO list](https://github.com/thin-edge/thin-edge.io/issues/2018#issuecomment-1705325553)

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -435,6 +435,14 @@ define_tedge_config! {
     },
 
     mqtt: {
+        /// MQTT topic root
+        #[tedge_config(default(value = "te"))]
+        topic_root: String,
+
+        /// The device MQTT topic identifier
+        #[tedge_config(default(value = "device/main//"))]
+        device_topic_id: String,
+
         bind: {
             /// The address mosquitto binds to for internal use
             #[tedge_config(example = "127.0.0.1", default(variable = "Ipv4Addr::LOCALHOST"))]
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -6,11 +6,13 @@ use crate::software_manager::builder::SoftwareManagerBuilder;
 use crate::software_manager::config::SoftwareManagerConfig;
 use crate::tedge_operation_converter::builder::TedgeOperationConverterBuilder;
 use crate::tedge_to_te_converter::converter::TedgetoTeConverter;
+use crate::AgentOpt;
 use camino::Utf8PathBuf;
 use flockfile::check_another_instance_is_not_running;
 use flockfile::Flockfile;
 use flockfile::FlockfileError;
 use std::fmt::Debug;
+use std::sync::Arc;
 use tedge_actors::ConvertingActor;
 use tedge_actors::ConvertingActorBuilder;
 use tedge_actors::MessageSink;
@@ -38,11 +40,14 @@ pub struct AgentConfig {
     pub use_lock: bool,
     pub log_dir: Utf8PathBuf,
     pub data_dir: Utf8PathBuf,
+    pub mqtt_device_topic_id: Arc<str>,
+    pub mqtt_topic_root: Arc<str>,
 }
 
 impl AgentConfig {
-    pub fn from_tedge_config(
+    pub fn from_config_and_cliopts(
         tedge_config_location: &tedge_config::TEdgeConfigLocation,
+        cliopts: AgentOpt,
     ) -> Result<Self, anyhow::Error> {
         let config_repository =
             tedge_config::TEdgeConfigRepository::new(tedge_config_location.clone());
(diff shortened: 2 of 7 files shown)


crates/core/tedge_agent/src/agent.rs
```diff
@@ -38,22 +40,35 @@ pub struct AgentConfig {
     pub use_lock: bool,
     pub log_dir: Utf8PathBuf,
     pub data_dir: Utf8PathBuf,
+    pub mqtt_device_topic_id: Arc<str>,
```
Bravo555, line 43: Using `Arc<str>` here because it's better for immutable data, as it doesn't allocate more memory when cloning, and also tells the reader that it's immutable. More information in [this video](https://www.youtube.com/watch?v=A4cKi7PTJSs).
I think most of the data we're working with in thin-edge is immutable, so I'm experimenting with using these immutable types for new code.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -38,22 +40,35 @@ pub struct AgentConfig {
...
+    pub mqtt_topic_root: Arc<str>,
 }
 
 impl AgentConfig {
-    pub fn from_tedge_config(
+    pub fn from_config_and_cliopts(
```
Bravo555, line 48: We should try to find a more systematic solution to merging CLI arguments into a common config, that can be applied across all of our binaries. I think that's what figment does and we already use it to handle environment variables, but for now I wanted to push this ASAP so I haven't looked into it yet.

plugins/tedge_log_plugin/src/main.rs
```diff
@@ -43,11 +44,11 @@ pub struct LogfilePluginOpt {
     #[clap(long = "config-dir", default_value = DEFAULT_TEDGE_CONFIG_PATH)]
     pub config_dir: PathBuf,
 
-    #[clap(long, default_value = "te")]
-    root: String,
+    #[clap(long)]
+    mqtt_topic_root: Option<Arc<str>>,
```
Bravo555, line 48: Renamed flags here to be consistent with `tedge-agent`, also because I think `--root` or `--device` do not give much context to the user of what these options are for, although probably this should be accompanied by help text.

crates/core/tedge_agent/src/agent.rs
didier-wenzek: Thank you. Indeed a nice hint, that can be used for all the `String`s we use as names or identifiers.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -163,7 +175,7 @@ impl Agent {
         let signal_actor_builder = SignalActor::builder(&runtime.get_handle());
 
         // Health actor
-        let health_actor = HealthMonitorBuilder::new(TEDGE_AGENT, &mut mqtt_actor_builder);
+        let health_actor = HealthMonitorBuilder::new("health-actor", &mut mqtt_actor_builder);
```
didier-wenzek, line 178: No, the name which is expected here is the name of the service.

This is one of the reasons why the system tests are failing. They expect the `"tedge-agent"` to be up, not some unknown `"health-actor"`. 

However, using `"tedge-agent"` here can only be a short term solution. To have the agent running concurrently on  several devices, one needs instance specific health status. The better would be to provide here the service topic id. Ignoring the issue with custom device topic id, the service topic id can be forged after the parent topic id.  For instance `device/child001/service/tedge-agent`.

didier-wenzek (changes requested):

Ruadhri17 (commented):
Looks good (Apart from the health actor service name)

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -435,6 +435,14 @@ define_tedge_config! {
     },
 
     mqtt: {
+        /// MQTT topic root
```
rina23q, line 438: This comment goes directly to the description showed by `tedge config --doc`. So I want to have a bit better description. (Another suggestion is welcomed!)
```suggestion
        /// The MQTT root/base topic prefix
```

crates/core/tedge_agent/src/agent.rs
rina23q: This is really nice! It's a great help to distinguish if a member is immutable or mutable by just looking at the struct data type.

plugins/tedge_log_plugin/src/main.rs
rina23q: Understood. Consistency is most important!

rina23q (commented):
Looks good in general!

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -435,6 +435,14 @@ define_tedge_config! {
...
+        #[tedge_config(default(value = "te"))]
+        topic_root: String,
+
+        /// The device MQTT topic identifier
+        #[tedge_config(default(value = "device/main//"))]
+        device_topic_id: String,
```
didier-wenzek, line 444: @reubenmiller you told me off-line that you where not happy with this name.
 
 What would be your proposal?

didier-wenzek (commented):
I'm okay to approve this PR. I just need a confirmation from @reubenmiller about the names for the two new config settings.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
reubenmiller: That's fine, we'll leave it as is

didier-wenzek (approved):
Approved

PR #2235, opened 2023-09-07
Document the agent and the device management API

didier-wenzek:
## Proposed changes

- [x] Create a reference section for the agent and operation plugins
- [x] Move under the agent section the documentation of the tedge-xxx-plugins
- [x] Document `tedge-agent`
- [x] Document the principles ruling the command APIs
- [ ] Document the software update API
- [ ] Document the restart API

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,295 @@
...
+  - When execution is successful, the `"status"` is set to `"successful"`.
+  - If execution fails for some reason, the `"status"` is set to `"failed"` and the reason given as `"reason"` field.
+  - An operation might define specific `"status"` values for extra steps, actions or checks.
+
+:::info
+Health checks is a notable exception to the rule "command status are published as retained messages".
+A health check is sent to the chanel `cmd/health/check` of a service (or a device) as a non-retained request `{}`,
+triggering a health message response published on the `status/health` channel of the same service
+(or respectively the channels of all the service running on that device).
+:::
```
didier-wenzek, line 136: I have a concern about this use of the `cmd` channel for health check.

I would really prefer to send  health check requests on a `req` or `request` channel.
There are indeed too many differences:
- Health check requests must not be retained
- There is no need for a specific id for each request
- The response is sent to another topic (`status/health`).

Having such exceptions will make the implementation of extensible command workflow a bit more complex for no added value.

Having a `req/<request-type>` channel can be used for other kind of requests: [setpoints](https://en.wikipedia.org/wiki/Setpoint_(control_system)), clearing alarms ...

docs/src/references/agent/restart-operation.md
```diff
@@ -0,0 +1,80 @@
...
+the agent or the `restart` plugin publishes the new state of the command,
+adding a `reason` text field with the error. 
+
+```sh te2mqtt formats="v1"
+tedge mqtt pub --retain 'te/device/child001///cmd/restart/c8y-2023-09-08T18:13:00' '{
+    "status": "failed",
+    "reason": "The device has not restarted within 5 minutes"
+}'
+```
+
```
didier-wenzek, line 79: It would be good to add details on how restart is implemented by the agent.
However, I would like first to clarify this design before documenting too much details.

docs/src/references/agent/software-management.md
```diff
@@ -0,0 +1,332 @@
...
+}'
+```
+
+### `init` state
+
+A `software-update` command is defined by an `"updateList"` array giving the packages to install, update or remove.
+
+- If `"updateList"` field is not provided or empty, then
+  - the command must change nothing on the device
+  - and simply return the list of software currently installed.
```
didier-wenzek, line 44: Currently, `tedge-agent` implements 2 operations: `software-list` and `software-update`.
Here, I propose to merge the two in a single  `software-update` used to query and update software.

Indeed, having two operations add some (little) complexities with no real benefits beyond mimicking c8y.
- If there are two operations, does it make sense for a device to support one but not the other?
- Have the list of supported package types to be same for the two operations?

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,325 @@
...
+  end
+```
+
+Where the groups are described as follows:
+
+| Group        | Description                                                                                                                                                                              |
+|--------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| root         | The [topic root prefix](../mqtt-api.md/#group-root) for this installation of thin-edge.                                                                                                  |
+| target       | The [topic identifier](../mqtt-api.md/#group-identifier) of the target device, possibly a [custom identifier](/contribute/design/mqtt-topic-design.md/#using-custom-identifier-schemas). |
+| `cmd`        | The [command channel](../mqtt-api.md/#channel-identifier) grouping of all the commands for this target.                                                                                  |
```
reubenmiller, line 60:
```suggestion
| `cmd`        | The [command channel](../mqtt-api.md/#channel-identifier) grouping all of the commands for this target.                                                                                  |
```

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,325 @@
...
+    cmd --/--- cmd_type["&lt;command_type&gt;"] --/--- cmd_id["&lt;cmd_id&gt;"]
+  end
+```
+
+| Group        | Description                               |
+|--------------|-------------------------------------------|
+| command_id   | The identifier of the operation instance. |
+
+The messages published on these topics represent each the current status of a running command.
+So, one can list __all the pending commands of any type across all the devices__:
```
reubenmiller, line 101: Though it is not just the pending operations but the in progress operations as well...so something like:

```suggestion
So, one can list __all the pending/in progress commands of any type across all the devices__:
```

docs/src/references/agent/index.md
```diff
@@ -0,0 +1,54 @@
...
+tags: [Reference, Agent]
+sidebar_position: 1
+---
+
+import DocCardList from '@theme/DocCardList';
+
+# The Agent and Operations Plugins
+
+The __agent__ and the __operation plugins__ are the main components used by thin-edge
+to enable edge device management from various cloud on diverse hardware,
```
reubenmiller, line 12:
```suggestion
to enable edge device management from various clouds on diverse hardware,
```

docs/src/references/agent/index.md
```diff
@@ -0,0 +1,54 @@
...
+giving the cloud operators a uniform interface to install software packages,
+update configuration files, inspect log files, restart devices ...
+
+The API between the mappers and the agent is designed to abstract device actual location, operating system, file system and hardware.
```
reubenmiller, line 16:
```suggestion
The API between the mappers and the agent is designed to abstract device location, operating system, file system and hardware.
```

docs/src/references/agent/restart-operation.md
```diff
@@ -0,0 +1,80 @@
...
+# Restart Operation
+
+Thin-edge defines a `restart` operation to restart a device, being the main device or a child device.
+
+- A restart is typically triggered by a [mapper](../mappers) on behalf of a cloud operator.
+- A restart can also be triggered from another operation (as a software update)
+  or service (detecting for instance some anomalies requesting a reboot).
+- `tedge-agent` is the reference implementation of the `restart` operation.
+- However, a custom `restart` plugin implementation can be installed on a device with specific requirements.
```
reubenmiller, line 15: This sentence can be a little misleading as it is not 100% clear that you need to display the restart function handled by the tedge-agent before doing your own implementation (assuming you still want to use other aspects of the tedge-agent, just not the restart functionality).

docs/src/references/agent/restart-operation.md
```diff
@@ -0,0 +1,80 @@
...
+
+## MQTT API
+
+The `restart` operation API follows the [generic thin-edge rules for operations](device-management-api.md):
+
+- The `te/<device-topic-id>/cmd/restart` topic is used to tell the device `<device-topic-id>` can be restarted.
+- Each `restart` request is given a `<command-id>` and a dedicated topic  `te/<device-topic-id>/cmd/restart/<command-id>`,
+  where all the subsequent states of the restart command are published during its execution.
+- The workflow is [generic with `"init"`, `"executing"`, `"successful"` and `"failed"` statutes](references/agent/device-management-api.md/#operation-workflow).
```
reubenmiller, line 24:
```suggestion
- The workflow is [generic with `"init"`, `"executing"`, `"successful"` and `"failed"` statuses](references/agent/device-management-api.md/#operation-workflow).
```

docs/src/references/agent/software-management.md
```diff
@@ -0,0 +1,332 @@
...
+- install, update and remove software packages on a device
+
+## MQTT API
+
+The `software-update` operation API follows the [generic thin-edge rules for operations](device-management-api.md):
+- The `te/<device-topic-id>/cmd/software-update` topic is used to publish the type of software packages
+  that can be managed on the device with the given topic identifier.
+- Each `te/<device-topic-id>/cmd/software-update/+` topic is dedicated to a software update command instance,
+  where all the subsequent states of this instance are published during its execution.
+- The workflow is [generic with `"init"`, `"executing"`, `"successful"` and `"failed"` statutes](references/agent/device-management-api.md/#operation-workflow).
```
reubenmiller, line 20:
```suggestion
- The workflow is [generic with `"init"`, `"executing"`, `"successful"` and `"failed"` statuses](references/agent/device-management-api.md/#operation-workflow).
```

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,325 @@
...
+
+  subgraph command["command request"]
+    direction LR
+    cmd --/--- cmd_type["&lt;command_type&gt;"] --/--- cmd_id["&lt;cmd_id&gt;"]
+  end
+```
+
+| Group        | Description                               |
+|--------------|-------------------------------------------|
+| command_id   | The identifier of the operation instance. |
```
reubenmiller, line 98: It might be worth mentioning that the `command_id` is an arbitrary string however it should be unique. It is recommended to either use a unique id generator, or add a unix timestamp as a suffix, e.g. `date +%s`

docs/src/references/agent/software-management.md
```diff
@@ -0,0 +1,332 @@
...
+The registration message of the `software-update` operation on a device:
+- must provide a `supportedTypes` list of the types of software package that can be installed on this device (e.g. `["apt", "docker"]`)
+- can provide a description of the operation and of each supported package type.
+
+```sh te2mqtt formats="v1"
+tedge mqtt pub --retain 'te/device/child001///cmd/software-update' '{
+    "description": "Install, update and remove software packages",
+    "supportedTypes": [
+      "apt",
+      "docker",
```
reubenmiller, line 33: Removing trailing comma as they are not supported in the vanilla JSON spec

```suggestion
      "docker"
```

Bravo555 (changes requested):

Bravo555:
Reading the PR, I started thinking about the command execution model. I have some questions:

- Are the all commands executed serially or could we allow some commands to be executed concurrently?
- I assume that we will need a queue for command requests. Is this queue bounded?
- If the queue is bounded and at full capacity, would it make sense to discard command requests, immediately moving them to the `"failure"` state without going through the execution state? This would be a clear signal of "i did not even attempt to execute your request, you're free to retry it later" for client applications, but it relies on the client application having noticed that there was no change to the `"execution"` state. Of course this information can also be included in the payload.
- What executes the commands, which is to say how many channels for commands do we need? One per service? One per device? One per thin-edge installation? Because one can imagine having many trivial commands which only invoke some process to read some data and return it in response, while also having "expensive" commands which can take a long time or perform side effects. One can imagine that a user would expect that e.g. restarting `childA` should not impact `childB`'s ability to handle commands. Likewise one would expect that `childB` can no longer process commands if we restart `main` which is `childB`'s parent.
- In such case, we would need some more semantics to express synchronization, e.g. when a command can be executed concurrently, and when it needs to be executed serially.

I think we can consider making a _Command/Command Execution Model_ specification which answers some of these questions.

docs/src/references/agent/software-management.md
didier-wenzek: Indeed, a counter argument to my proposal. However, I void avoid to run a `sofware-list` concurrently with a `software-update` as the result would not be deterministic.

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,325 @@
...
+  - When execution is successful, the `"status"` is set to `"successful"`.
+  - If execution fails for some reason, the `"status"` is set to `"failed"` and the reason given as `"reason"` field.
+  - An operation might define specific `"status"` values for extra steps, actions or checks.
+
+:::info
+Health checks is a notable exception to the rule "command status are published as retained messages".
+A health check is sent to the channel `cmd/health/check` of a service (or a device) as a non-retained request `{}`,
+triggering a health message response published on the `status/health` channel of the same service
+(or respectively the channels of all the service running on that device).
+:::
```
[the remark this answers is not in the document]
didier-wenzek: Health checks are implemented using a combination of 3 modes:
- passive mode: on start, the services publish an up status, on shutdown a down status
- crash mode: on crash, the MQTT broker publishes a down status on behalf of the dead process
- active mode: health check requests are sent on a regular pace (by a watchdog), and the services have to publish an up status.

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,325 @@
...
+
+## Operation workflow
+
+A specific workflow rules each operation type, with specific:
+- states
+- message payloads
+- states transitions.
+
+However, there are some common rules:
+- There at least for states: `"init"`, `"executing"`, `"successful"` and `"failed"`.
```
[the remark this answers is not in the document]
didier-wenzek: fixed

docs/src/references/agent/device-management-api.md
didier-wenzek: Added

docs/src/references/agent/index.md
didier-wenzek: fixed

docs/src/references/agent/index.md
didier-wenzek: fixed

docs/src/references/agent/index.md
```diff
@@ -0,0 +1,54 @@
...
+__Any service connected to the local MQTT bus can trigger commands as well as respond to commands.__
+
+- The agent and the operation plugins interact with the operating system,
+  running the appropriate sequences of checks and actions
+  in response to command requests initiated by [mappers](../mappers) on behalf of cloud operators.
+- `tedge-agent` is the agent reference implementation. It can run on the main device as well as on child devices,
+  but can also be substituted by a custom operation plugin to address specific requirements or hardware.
+
+```mermaid
```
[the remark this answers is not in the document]
didier-wenzek: Added this title: "System components enabling device management"

docs/src/references/agent/restart-operation.md
didier-wenzek: I don't understand what you mean by *need to display*.

docs/src/references/agent/software-management.md
didier-wenzek: fixed

docs/src/references/agent/software-management.md
didier-wenzek: fixed

docs/src/references/agent/restart-operation.md
didier-wenzek: fixed

docs/src/references/agent/restart-operation.md
```diff
@@ -0,0 +1,80 @@
...
+  or service (detecting for instance some anomalies requesting a reboot).
+- `tedge-agent` is the reference implementation of the `restart` operation.
+- However, a custom `restart` plugin implementation can be installed on a device with specific requirements.
+
+## MQTT API
+
+The `restart` operation API follows the [generic thin-edge rules for operations](device-management-api.md):
+
+- The `te/<device-topic-id>/cmd/restart` topic is used to tell the device `<device-topic-id>` can be restarted.
+- Each `restart` request is given a `<command-id>` and a dedicated topic  `te/<device-topic-id>/cmd/restart/<command-id>`,
```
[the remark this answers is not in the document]
didier-wenzek: I agree that adding an id to each restart command seems a bit heavy.

But it's better to be consistent for all operations. Keep also in mind that the next steps will be to introduce [extensible operation workflow](https://github.com/thin-edge/thin-edge.io/pull/2071) and for that consistency will be a key.

That said, the restart operation needs a redesign. Triggering reboots from `tedge-agent` kills the agent and leads to issues like this one: https://github.com/thin-edge/thin-edge.io/issues/2150

> That is assuming that commands are executed serially.

Yes this is the assumption. Unfortunately not explicitly stated by this doc. => I have to make this explicit somehow.

didier-wenzek:
> Reading the PR, I started thinking about the command execution model. Looking at the all the commands on the whole, I think these are important questions to understand how the command interface works in general:

Indeed, good points that must be addressed.

> - Are the all commands executed serially or could we allow some commands to be executed concurrently?

The assumption is that `tedge-agent` executes the operations sequentially.

> - I assume that we will need a queue for command requests. Is this queue bounded?

Yes, we need to enforce some maximum number of pending requests. But there is no queues, as the pending and in-progress requests are stored as retained messages.

This raises a point we missed: What is the expected order of execution? While `tedge-agent` is running it receives the request in a time-related order, but in case of a restart the pending request will be pushed in an arbitrary order to `tedge-agent`.
 
> If the queue is bounded and at full capacity, would it make sense to discard command requests, immediately moving them to the `"failure"` state without going through the execution state? This would be a clear signal of "i did not even attempt to execute your request, you're free to retry it later" for client applications, but it relies on the client application having noticed that there was no change to the `"execution"` state. Of course this information can also be included in the payload.

Agree, failing requests when overwhelmed  is the best approach.

> What executes the commands, which is to say how many channels for commands do we need? One per service? One per device? One per thin-edge installation? Because one can imagine having many trivial commands which only invoke some process to read some data and return it in response, while also having "expensive" commands which can take a long time or perform side effects. 

- What do you name a channel here ? An in-memory queue is some process or the topics `te/+/+/+/+/cmd/+` ?
- The assumption is to have a sequential execution independently on each device.

> One can imagine that a user would expect that e.g. restarting `childA` should not impact `childB`'s ability to handle commands. Likewise one would expect that `childB` can no longer process commands if we restart `main` which is `childB`'s parent.

Having some kind of synchronization at the scale of a main device and all the child devices is something not planned for now. However, this is a good point.

__This where the current design shines:__ it's easy to add a scheduler service per thin-edge installation:
- as currently proposed, the mappers create commands in the `init` state.
- but the agent and the operation plugins wait for  the commands to in a `scheduled` state.
- the scheduler service will be then the only process moving commands from the `init` state to the `scheduled` state
- the scheduler can also decide to move some commands to "failed" if the device is overwhelmed.

> In such case, we would need some more semantics to express synchronization, e.g. when a command can be executed concurrently, and when it needs to be executed serially.

On a single,  

> I think we can consider making a _Command/Command Execution Model_ specification which answers some of these questions.

This is definitely required, but to be honest, I would prefer to address this in another PR. The point of this PR was only to describe what has been implemented so far and to clarify some required changes at the topic level so `tedge-agent` can run on a child-device.

This must be addressed with  https://github.com/thin-edge/thin-edge.io/pull/2071

The approach I would take is along the lines of the "scheduler service" described above.

docs/src/references/agent/software-management.md
reubenmiller: I would also favour having independent `software-list` and `software-update` commands, as this would allow for a device to only support a software-list operation when user's want to manage their software via pure OS updates (for improved robustness, by having the rollback safety of A/B style updates).

didier-wenzek: Okay. I will rewrite this proposal with two independent `software-list` and `software-update` commands.

Bravo555:
> This is definitely required, but to be honest, I would prefer to address this in another PR. The point of this PR was only to describe what has been implemented so far and to clarify some required changes at the topic level so `tedge-agent` can run on a child-device.

Agreed, this spec should be the subject of another PR.

docs/src/references/agent/restart-operation.md
reubenmiller: Sorry it was a type/word-mixup....it should read `need to disable`

docs/src/references/agent/restart-operation.md
```diff
@@ -0,0 +1,80 @@
...
+adding a `reason` text field with the error. 
+
+```sh te2mqtt formats="v1"
+tedge mqtt pub --retain 'te/device/child001///cmd/restart/c8y-2023-09-08T18:13:00' '{
+    "status": "failed",
+    "reason": "The device has not restarted within 5 minutes"
+}'
+```
+
+
```
albinsuresh, line 80: Although it's part of the generic commands handling protocol, you might wanna add a few lines about clearing the retained cmd as well.

docs/src/references/agent/restart-operation.md
```diff
@@ -0,0 +1,80 @@
...
+When ready, but before actually restarting the device,
+the agent or the `restart` plugin publishes the new state of the command.
+
+```sh te2mqtt formats="v1"
+tedge mqtt pub --retain 'te/device/child001///cmd/restart/c8y-2023-09-08T18:13:00' '{
+    "status": "executing"
+}'
+```
+
```
albinsuresh, line 55: Although not part of the core API itself, it might be important to highlight that the `restart` plugin must be able to survive a device restart and push the state machine through the subsequent states as well. It shouldn't expect the agent to do the rest of the work after the work restart is done and can be verified. It also needs to differentiate between a simple process restart vs the actual device restart itself.

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+  - Initialized with all the required information to execute the command,
+    they are updated during the command execution with outcome data.
+  - A command request is cleared by publishing a retained empty string `""`.
+- The command status messages contain a mandatory `"status"` field.
+  - This status tells at which step the command execution is along its workflow. 
+  - A new command is created with a `"status"` set to `"init"`.
+  - When execution starts, the `"status"` is set to `"executing"`.
+  - When execution is successful, the `"status"` is set to `"successful"`.
+  - If execution fails for some reason, the `"status"` is set to `"failed"` and the reason given as `"reason"` field.
+  - An operation might define specific `"status"` values for extra steps, actions or checks.
```
albinsuresh, line 134:
```suggestion
  - Different operations might define other specific `"status"` values for extra steps, actions or checks.
```

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Command execution
+
+To request the execution of a command, a client must:
+- know the MQTT root prefix (per default: `te`),
+- know the entity topic identifier of the target device (e.g. `device/child001//`),
+- know the well-known name of the operation (e.g. `software-update`),
+- assign a unique id to its command (say `c8y-1234`)
+- use these target and command identifiers to forge the topic name for its request
+  (e.g. `te/device/child001///cmd/software-update/c8y-1234`)
```
albinsuresh, line 204: I'd just compress all that into a single statement: "subscribe to the command topic that includes the command id as well (e.g. `te/device/child001///cmd/software-update/c8y-1234`).

Using a term like command topic should be sufficient as we've already established its format in the previous section.

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+- await on the same topic for a response message with a `"status"` set to `"executing"`, `"successful"` or `"failed"`
+- ignore any message with an unknown status as the agent is free to use intermediate states
+- finally, clear the command topic by sending an empty retained message
+
+To implement an operation on behalf of a device, a service must:
+- know the MQTT root prefix (per default: `te`),
+- know the entity topic identifier of the target device (e.g. `device/child001//`),
+- know the well-known name of the operation (e.g. `software-update`),
+- subscribe to any command requests of this type and the target device (i.e. `te/device/child001///cmd/software-update/+`)
+- react to a well-formed requests by publishing the updated status when the step is done.
```
albinsuresh, line 216: Again, not worth repeating I feel.

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Example
+
+As an example, let's take software updates on a child device.
+
+When launched on the child device `device/child001//`,
+`tedge-agent` notifies that software packages can be updated on this device:
```
albinsuresh, line 223:
```suggestion
`tedge-agent` notifies that software packages of types: `apt` and `docker` can be updated on this device:
```
Just to highlight the main points in the example.

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+}'
+```
+
+The Cumulocity mapper, having subscribed to all software-update commands,
+monitors this command instance and notifies Cumulocity of its progress upto completion.  
+Finally, the Cumulocity mapper clear the command topic:
+
+```sh te2mqtt formats="v1"
+tedge mqtt pub --retain 'te/device/child001///cmd/software-update/c8y-123' ''
+```
```
albinsuresh, line 330: I'm assuming that the workflow API bits like the workflow definition file from [this PR](https://github.com/thin-edge/thin-edge.io/pull/2071) will also be added to the device-mangement-api doc at some point. Because, that'll give clarity to the users on how different components can work together with `tedge-agent` to drive an operation to completion.

docs/src/references/agent/device-management-api.md
albinsuresh: I wasn't a big fan of retrofitting health statuses by partially using the cmd APIs. But we did that at that point to avoid introducing yet another req/res mechanism just for health statuses. But, with additional things like setpoints or clear alarm requests, may be it's worth adding something.

But, I still wouldn't define an API that ties the response to the request in any way, as we want components to freely declare health status, setpoint values etc without any explicit request. Even for the clear alarm case, the prompt will be something and the resultant action will be just the regular "clear alarm" message. Will have to think about such a disjoint req/prompt mechanism.

docs/src/references/agent/restart-operation.md
didier-wenzek: I will add a note to clarify this point.

docs/src/references/agent/index.md
```diff
@@ -0,0 +1,57 @@
...
+giving the cloud operators a uniform interface to install software packages,
+update configuration files, inspect log files, restart devices ...
+
+The API between the mappers and the agent is designed to abstract device location, operating system, file system and hardware.
+__Any service connected to the local MQTT bus can trigger commands as well as respond to commands.__
+
+- The agent and the operation plugins interact with the operating system,
+  running the appropriate sequences of checks and actions
+  in response to command requests initiated by [mappers](../mappers) on behalf of cloud operators.
+- `tedge-agent` is the agent reference implementation. It can run on the main device as well as on child devices,
```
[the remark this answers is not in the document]
didier-wenzek: What would be your proposal?

docs/src/references/agent/device-management-api.md
didier-wenzek: Agree that this is bit verbose. But I think it helps to be explicit in a reference documentation on what needs to given to and then done by the operation plugins.

docs/src/references/agent/device-management-api.md
didier-wenzek: Indeed, I would not repeat that all and every operation. However, this must be said once.

docs/src/references/agent/device-management-api.md
didier-wenzek: Yes, definitely. I even think the whole [operation workflow specification](https://github.com/thin-edge/thin-edge.io/pull/2071) has to reworked on top of this PR.

docs/src/references/agent/device-management-api.md
didier-wenzek: Added

docs/src/references/agent/device-management-api.md
didier-wenzek: Updated along an intermediate suggestion:

```suggestion
  - Each operation might define other specific `"status"` values for extra steps, actions or checks.
```

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+
+The agent and the operation plugins enable device management on thin-edge,
+giving the cloud operators a uniform interface to install software packages,
+update configuration files, inspect log files, restart devices ...
+
+Each of these device management operation requires a different set of interactions between the participants,
+with specific steps, message payloads and expectations.
+
+Hence, each operation type comes with its own API, and, possibly, with different implementations,
+the `tedge-agent` service, operation plugins as the `tedge-log-plugin` or user-specific plugins.
```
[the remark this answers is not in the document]
didier-wenzek: updated

docs/src/references/agent/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+# The Agent and Operations Plugins
+
+The __agent__ and the __operation plugins__ are the main components used by thin-edge
+to enable edge device management from various clouds on diverse hardware,
+giving the cloud operators a uniform interface to install software packages,
+update configuration files, inspect log files, restart devices ...
+
+The API between the mappers and the agent is designed to abstract device location, operating system, file system and hardware.
+__Any service connected to the local MQTT bus can trigger commands as well as respond to commands.__
```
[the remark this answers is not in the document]
didier-wenzek: Added

docs/src/references/agent/restart-operation.md
didier-wenzek: Added

docs/src/references/agent/restart-operation.md
didier-wenzek: My plan was to describe the internals in a different PR as we need the current implementation needs some rework.

That said, I added a few lines along your proposal.

docs/src/references/agent/index.md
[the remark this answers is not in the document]
albinsuresh: Just skip that controversial word completely and just say: "`tedge-agent` can run on the main device as well as on child devices" ? And somewhere where we talk about workflow overriding stuff, we can mention that `tedge-agent` can be replaced with user developed components, if desired.

docs/src/references/agent/software-management.md
albinsuresh: > Have the list of supported package types to be same for the two operations?

I'm of the opinion that the `device software list` and `updatable software list` would mostly likely be different for many customers. They may want to view all the installed packages from the cloud but limit the updatable list to a much smaller subset only for the ones that are safe to be updated remotely. Unfortunately C8y has them merged, for simplicity probably. I'd argue the same for configs as well. Not all readable configs are updatable.

docs/src/references/agent/software-management.md
```diff
@@ -0,0 +1,376 @@
...
+    ]}
+     ```
+   - Each action is either to `"install"` or `"remove"` a software package.
+      - `"action": "install"` is to be understood as a goal:
+        "this package must be installed with this version" whatever the actual action to reach this goal,
+        being a fresh install, an upgrade or a downgrade.
+      - `"action": "remove"` is also to be understood as a goal: "that package must not be installed".
+   - An action provides:
+      - the package `"name"` (as known by the package packager),
+      - optionally a `"version"` (using the same conventions as the package packager),
```
albinsuresh, line 174:
```suggestion
      - optionally a `"version"` (using the same conventions as the package manager),
```

docs/src/references/agent/software-management.md
```diff
@@ -0,0 +1,376 @@
...
+                },
+                {
+                    "name": "mongodb",
+                    "version": "4.4.6",
+                    "action": "remove"
+                }
+            ]
+        }
+    ]
+}'
```
albinsuresh, line 270: I'm assuming that the omission of the `currentSoftwareList` from the update command response intentional. I'm juts curious why. Of course, the mapper can always trigger the `software_list` command immediately after it gets the `successful` response from the agent, to report the up-to-date software list to the cloud. The `currentSoftwareList` was included earlier just to avoid this extra hop.

docs/src/references/agent/software-management.md
didier-wenzek: done

docs/src/references/agent/software-management.md
```diff
@@ -0,0 +1,376 @@
...
+title: Software management
+tags: [Reference, Agent, Software Management]
+sidebar_position: 3
+---
+
+Thin-edge software management is implemented by two operations
+which give the ability to manage software packages of different types on the same device.
+
+- `software-list` is used to fetch a pertinent subset of the software packages installed on a device.
+- `software-update` let the user install, update and remove software packages on a device.
```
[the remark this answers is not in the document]
didier-wenzek: Done

docs/src/references/agent/software-management.md
```diff
@@ -4,29 +4,142 @@ tags: [Reference, Agent, Software Management]
...
+Thin-edge software management is implemented by two operations
+which give the ability to manage software packages of different types on the same device.
 
-## MQTT API
+- `software_list` is used to fetch a pertinent subset of the software packages installed on a device.
+- `software_update` let the user install, update and remove software packages on a device.
 
-The `software-update` operation API follows the [generic thin-edge rules for operations](device-management-api.md):
-- The `te/<device-topic-id>/cmd/software-update` topic is used to publish the type of software packages
+## `software_list` MQTT API
```
Bravo555, line 13: chore: code block in header
https://thin-edge.github.io/thin-edge.io/contribute/writing-documentation/markdown-guidelines/#avoid-in-line-code-blocks-in-headers

docs/src/references/agent/software-management.md
```diff
@@ -4,29 +4,142 @@ tags: [Reference, Agent, Software Management]
...
+              "type": "apt",
+              "modules": [ "..." ]
+            },
+            {
+              "type": "docker",
+              "modules": [ "..." ]
+            }
+         ]
+      }
+       ```
```
Bravo555, line 75: chore: unaligned codeblock tag

Bravo555 (changes requested):
No issues from me other than fixing the formatting.

albinsuresh (commented):
The changes LGTM.

docs/src/references/agent/software-management.md
```diff
@@ -0,0 +1,378 @@
...
+}'
+```
+
+### init state
+
+A `software_list` command has nothing to provide beyond a `status` field.
+This empty message stands for a request of the list of software currently installed.
+
+```sh te2mqtt formats="v1"
+tedge mqtt pub --retain 'te/device/child001///cmd/software_list/c8y-2023-09-25T14:34:00' '{
```
rina23q, line 43: [Out of topic] I found using cloud type and RFC3339 timestamp as a command ID brilliant and better than random ID generated by nanoid. It gives understanable information to users and also unique!

docs/src/references/agent/tedge-configuration-management.md
rina23q, line 1: [Minor] Moving the two files, tedge-configuration-management and tedge-log-management to under tedge-agent doc is correct in terms of our roadmap. However, be aware that there are still "usage" sections in the two page, which shows the output the plugins. These must be removed in the future.

rina23q (commented):
LGTM!

docs/src/references/agent/software-management.md
didier-wenzek: This has been addressed by https://github.com/thin-edge/thin-edge.io/pull/2235/commits/a48766243fe489a0bfdbe65b1022f145d007d1a2

docs/src/references/agent/device-management-api.md
didier-wenzek: For details, see https://github.com/thin-edge/thin-edge.io/issues/2372

docs/src/references/agent/software-management.md
didier-wenzek: See https://github.com/thin-edge/thin-edge.io/issues/2251

rina23q (approved):

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+
+### Capability declaration
+
+On start, a service that implements an operation for a device must:
+
+- know the MQTT root prefix (per default: `te`),
+- know the entity topic identifier of the target device (e.g. `device/child001//`),
+- know the well-known name of the operation (e.g. `software-update`),
```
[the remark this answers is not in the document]
didier-wenzek: Fixed all over the new pages.

albinsuresh (approved):
LGTM

docs/src/references/agent/device-management-api.md
```diff
@@ -0,0 +1,330 @@
...
+            "modules": [
+                {
+                    "name": "nodered",
+                    "version": "1.0.0",
+                    "action": "install"
+                }
+            ]
+        }
+    ],
+    "currentSoftwareList": [
```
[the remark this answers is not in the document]
didier-wenzek: Oops!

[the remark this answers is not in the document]
didier-wenzek: Fixed

PR #2108, opened 2023-07-28
C8y mapper handling c8y_LogfileRequest

albinsuresh:
### Todo (in this PR's scope)
- [x] C8y mapper todo the mapping of `c8y_LogfileRequest` to tedge `log_upload` command
- [x] C8y mapper converts supported log types message to c8y format
- [x] Supported operation file creation when it receives `log_upload` metadata topic
- [x] Combine file upload logic for both log and config -> Move to #2203 as it's refactoring of the existing code
- [x] Clear the retain messages after operations finishes
- [ ] Dedicated actor for http uploads to avoid blocking the mapper actor

### Todo out of this PR
- [x] Rename the filed `maximumLines` to `lines` in specs, both tedge-log-plugin and c8y-mapper. #2127
- [ ] Write a robot test for tedge-log-plugin and new c8y-mapper

### Much later (not in the drop 1 scope)
- [ ] Streaming file upload from C8yHttpProxy actor

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
#2048 

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [ ] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2245, opened 2023-09-12
Restart operation on child-devices

didier-wenzek:
Updating `tedge-agent` and `c8y-mapper` to support restart operation on child devices

## Proposed changes

- [x] use `mqtt.topic_root` and `mqtt.device_topic_id`
- [x] use ` te/device/main///cmd/restart/+` instead of `tedge/commands/+/control/restart`
- [x] move the command id from the payload to the topic
- [x] publish restart commands as retained messages
- [x] The c8y-mapper has to register child device restart capability
- [x] Add system tests: restarting a child device from c8y

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2018

## Checklist


reubenmiller:
@didier-wenzek is the PR in a state where we can add an integration test for the child device as well? Or is there still other work required for the tedge-agent to function on a child device (in the scope of the restart command only)

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -348,7 +348,7 @@ mod test {
         let tempdir_path = ttd.utf8_path_buf();
         let http_config = HttpConfig::default()
             .with_data_dir(tempdir_path)
-            .with_port(3000);
+            .with_port(3333);
```
albinsuresh, line 351: Any specific issue that triggered this change?

crates/core/tedge_api/src/messages.rs
```diff
@@ -27,12 +31,12 @@ where
         serde_json::from_slice(bytes)
     }
 
-    fn to_json(&self) -> Result<String, serde_json::Error> {
-        serde_json::to_string(self)
+    fn to_json(&self) -> String {
+        serde_json::to_string(self).unwrap() // all thin-edge data can be serialize to json
```
albinsuresh, line 35: Looked bold, but makes sense.

crates/core/tedge_api/src/messages.rs
```diff
@@ -473,65 +477,115 @@ impl From<SoftwareError> for Option<SoftwareModuleItem> {
...
 }
 
-impl<'a> Jsonify<'a> for RestartOperationRequest {}
+impl RestartCommand {
+    /// Create a [RestartCommand] to restart a target device.
+    ///
+    /// - use a fresh cmd id
+    /// - set the status to [CommandStatus::Init]
+    pub fn new(target: EntityTopicId) -> Self {
+        let cmd_id = nanoid!();
```
albinsuresh, line 494: ~Wouldn't it be better to pass that `cmd_id` as an argument, so that the caller has better control on whether to create unique IDs or pass something know? Especially for unit tests.~

Saw the `with_id` function below that solves this problem. This is better.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -981,35 +995,76 @@ fn create_inventory_fragments_message(
...
-
-    match response.status() {
-        OperationStatus::Executing => {
-            let smartrest_set_operation = SmartRestSetOperationToExecuting::new(
-                CumulocitySupportedOperations::C8yRestartRequest,
-            )
-            .to_smartrest()?;
-
-            Ok(vec![Message::new(&topic, smartrest_set_operation)])
+impl CumulocityConverter {
```
albinsuresh, line 998: I just realised that you can have multiple plain impl blocks for the same struct, which are not for any trait implementations and this module already had 3 different impl blocks. Looks fine, but I just don't understand the benefits though. I'm assuming it is for better code organization to isolate different responsibilities into different impl blocks instead of one big fat impl block.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -981,35 +995,76 @@ fn create_inventory_fragments_message(
...
+            None => {
+                error!("Fail to register `restart` operation for unknown device: {target}");
+                Ok(vec![])
+            }
+            Some(device) => {
+                let ops_dir = match device.r#type {
+                    EntityType::MainDevice => self.ops_dir.clone(),
+                    EntityType::ChildDevice => self.ops_dir.clone().join(&device.entity_id),
+                    EntityType::Service => {
+                        error!("Fail to register `restart` operation for a service: {target}");
```
albinsuresh, line 1013:
```suggestion
                        error!("Unsupported `restart` operation for a service: {target}");
```

albinsuresh (approved):
LGTM. I was surprised to see that no integration tests were updated. But then found out that none exists atm. I'm hoping that it is feasible now, since we have containers in place of real devices.

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
reubenmiller: Looking at the individual commits is helpful here ;)

didier-wenzek: See https://github.com/thin-edge/thin-edge.io/pull/2245/commits/4874c95f7e159d99510a6735c2fdb6a736399261

crates/core/tedge_api/src/messages.rs
didier-wenzek: I decided to move that way when I saw the impact for nothing on so many places: https://github.com/thin-edge/thin-edge.io/pull/2245/commits/162a83ba520ccd4862874b1d7eed7b0d89cc4a23.

reubenmiller:
This might be a bit premature, but I did a quick manual test to see if I could trigger a restart operation from Cumulocity on the child device, but
* c8y_Restart operation is not registered on the c8y device managed object
* the tedge-agent running on the child device does not respond to the operation, the operation stays in PENDING status in Cumulocity

So if this is not the expected behaviour (due to that part not being implemented yet), then I would suggest we wait to add some new integration tests for this new functionality before we merge it.

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: Indeed, something that needs to be fixed. But in a PR doing only that.

Why are we here? Previously, there was an implementations of a trait used *only* here. There were also functions which have been promoted to methods. When I introduced [these changes](https://github.com/thin-edge/thin-edge.io/pull/2108), I decided to go that way in order to help the code reviewers by minimizing code shuffling on an already huge refactoring.

Now, this can be cleaned up.

didier-wenzek:
> So if this is not the expected behaviour (due to that part not being implemented yet), then I would suggest we wait to add some new integration tests for this new functionality before we merge it.

Sure, we must add system tests with the 3 systems (c8y, c8y-mapper and tedge-agent). The integration tests are only checking partial exchanges between c8y-mapper and tedge-agent.

reubenmiller:
> > So if this is not the expected behaviour (due to that part not being implemented yet), then I would suggest we wait to add some new integration tests for this new functionality before we merge it.
> 
> Sure, we must add system tests with the 3 systems (c8y, c8y-mapper and tedge-agent). The integration tests are only checking partial exchanges between c8y-mapper and tedge-agent.

Then I'll add the system tests on Monday, though we definitely have issues here (as previously posted).

didier-wenzek:
> @didier-wenzek is the PR in a state where we can add an integration test for the child device as well? Or is there still other work required for the tedge-agent to function on a child device (in the scope of the restart command only)

The PR should be in a state where a restart of the child-device should work.

> I was also surprised to see that no integration tests were updated. But then found out that none exists atm. I'm hoping that it is feasible now, since we have containers in place of real devices.

There is at least this one for the main device:

https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/restart/restart_device.robot

albinsuresh:
> > I was also surprised to see that no integration tests were updated. But then found out that none exists atm. I'm hoping that it is feasible now, since we have containers in place of real devices.
> 
> There is at least this one for the main device:
> 
> https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/restart/restart_device.robot

Aah okay, using the Cumulocity APIs instead of triggering commands via `te/.../cmd/restart` topics. I was searching for a direct test using the `tedge/` or `te/` topics and that's why I didn't get any results.

didier-wenzek:
This PR has been rebased on top of https://github.com/thin-edge/thin-edge.io/pull/2265

Bravo555 (approved):

didier-wenzek:
I confirm something is going wrong :-(

Restart operations are pushed and executed but then the agent start looping while the mapper say to nothing to c8y.

```
```

didier-wenzek:
To be fixed:
- [ ] the agent must ignore commands which are not in the init state
- [ ] the mapper should use better cmd_id e.g. `c8y-{timestamp}`
- [ ] the mapper correctly send `[c8y/s/us] 502,c8y_Restart,"Restart Failed"` but this is ignored by the cloud. Why?
- [ ] the mapper must clear successful and failed requests
- [ ] the reason for a restart failure is not given
- [ ] on restart, the mapper should not create a new restart request (when pending in the cloud and locally) 
- [ ] it would be helpful to add to each message the name of the service which changed the state

didier-wenzek:
> The reason behind the status updates and failure reason not working is because they're being published on the incorrect c8y topic. For child devices it should be c8y/s/us/{child_id}

I started by testing the main device.

reubenmiller:
@didier-wenzek I've added a system test to the PR

tests/RobotFramework/tests/cumulocity/restart/restart_device_child.robot
```diff
@@ -0,0 +1,64 @@
...
+
+Suite Setup         Custom Setup
+Test Teardown       Get Logs
+
+Test Tags           theme:c8y    theme:troubleshooting    theme:childdevices
+
+
+*** Test Cases ***
+Support restarting the device
+    Cumulocity.Should Contain Supported Operations    c8y_Restart
```
didier-wenzek, line 14: @reubenmiller Where is this check defined?

tests/RobotFramework/tests/cumulocity/restart/restart_device_child.robot
```diff
@@ -0,0 +1,64 @@
...
+    Execute Command    sudo tedge config set mqtt.external.bind.address ${PARENT_IP}
+    Execute Command    sudo tedge config set mqtt.external.bind.port 1883
+
+    ThinEdgeIO.Connect Mapper    c8y
+    ThinEdgeIO.Service Health Status Should Be Up    tedge-mapper-c8y
+
+    # Child
+    ${child_sn}=    Setup    skip_bootstrap=${True}
+    Set Suite Variable    $CHILD_SN    ${child_sn}
+    Register Child Device    parent_name=${PARENT_SN}    child_name=${CHILD_SN}    supported_operations=c8y_LogfileRequest
```
didier-wenzek, line 61: Support of `c8y_LogfileRequest` is explicitly added here. Why? I would expect to have this added when `tedge_log_plugin` starts.

reubenmiller: Sorry that is my mistake (copy paste error). Maybe this is the source of the error. When I wrote the system tests for the tedge-log-plugin the supported operations mechanism was not implemented, so I compensated for it...I'll double check if this is the case now or not and update the tests accordingly.

tests/RobotFramework/tests/cumulocity/restart/restart_device_child.robot
reubenmiller: The assertion checks if the `c8y_Restart` operation type has been declared in Cumulocity IoT REST API.

The capabilities are registered fine on the tedge broker, however they are definitely not translated to cloud messages. I believe this functionality is missing as the c8y_LogFileRequest operation is also not sent to the cloud.

tests/RobotFramework/tests/cumulocity/restart/restart_device_child.robot
reubenmiller: I've repeated the checks, and tedge does not forward the locally registered capabilities to Cumulocity...so this functionality probably hasn't been implemented yet.

didier-wenzek: I'm missing something. I tested manual registration is working as I was expecting.

1. On start, `tedge-agent` publishes its capabilities `[te/device/child777-002///cmd/restart] {}`
2. On reception of this message, `c8y-mapper` publishes an auto-registration messages : `[te/device/child777-002//] { "@type":"child-device", "@id":"child777-002"}`.
3. On reception of this message, `c8y-mapper` also registers the child device and its capabilities:
     1. `[c8y/s/us] 101,child777-002,child777-002,thin-edge.io-child`
     2. `[c8y/s/us/child777-002] 114,c8y_Restart`

reubenmiller: Strange. I'll have another look, and try and strip back the test setup to a bare minimum, though I believe the test was already fairly minimal...

didier-wenzek: > though I believe the test was already fairly minimal...

The test looks good. However, it's not obvious for me to see which checks are running on the main device and which on the child device.

reubenmiller: Ok, I've fixed it...the child device and capability are being registered correctly without doing any manual file creation (which what the test was previously doing).

So the only thing which is broken is the updating of the c8y_Restart operation status when operating on a child device.

didier-wenzek: > Ok, I've fixed it...the child device and capability are being registered correctly without doing any manual file creation (which what the test was previously doing).

A file is created by the c8y mapper on the main device.

> So the only thing which is broken is the updating of the c8y_Restart operation status when operating on a child device.

Thanks. I will have a look.

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -91,89 +112,100 @@ impl RestartManagerActor {
...
     async fn handle_restart_operation(&mut self) -> Result<(), RestartManagerError> {
-        let commands = self.get_restart_operation_commands().await?;
+        let commands = self.get_restart_operation_commands()?;
         for mut command in commands {
             match command.status().await {
                 Ok(status) => {
-                    if !status.success() {
-                        return Err(RestartManagerError::CommandFailed);
+                    if status.code().is_none() {
+                        return Err(RestartManagerError::CommandInterrupted {
```
reubenmiller, line 202: I don't think we should treat this as an error here as the script/process can be stopped via a signal as a result of the reboot handling (e.g. systemd killing the services quickly after `shutdown -r now` or `reboot` is executed in the command).

Instead we can just treat it like it was successful, and let the failed restart logic handle the case where no restart happens.

crates/core/tedge_api/src/messages.rs
```diff
@@ -473,65 +477,139 @@ impl From<SoftwareError> for Option<SoftwareModuleItem> {
...
+
+/// Command to restart a device
+#[derive(Debug, Clone, Deserialize, Serialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct RestartCommandPayload {
+    pub status: CommandStatus,
+
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub reason: String,
+}
```
Bravo555, line 603: Is the `reason` field supposed to be not empty only when `status` is `CommandStatus::Failed`? If there are no other motivations for this, I suggest using [Serde's internally tagged enum representation](https://serde.rs/enum-representations.html#internally-tagged). Modified `CommandStatus` would look like this:

```rust
#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
#[serde(rename_all = "camelCase", tag = "status")]
pub enum CommandStatus {
    Init,
    Executing,
    Successful,
    Failed { reason: String },
}
```

And the `reason` field would automatically be serialised/deserialised and only be present when in `Failed` variant, which looks the same as the current structure.

One issue is that `CommandStatus` would no longer be copy, and also perhaps there would have to be more code changes.

`RestartCommandPayload would then simply be:

```suggestion
#[derive(Debug, Clone, Deserialize, Serialize, Eq, PartialEq)]
#[serde(rename_all = "camelCase")]
pub struct RestartCommandPayload {
    #[serde(flatten)]
    pub status: CommandStatus,
}
```

reubenmiller (changes requested):
The PRs looking very close now, I pushed some minor fixes and added some more system tests.

Though only remaining minor change is the following:

* Log entry is misleading when the restart command is killed by a signal. The entry with `Restart command has been interrupted by a signal` is good, but it is proceeded by an entry with the opposite meaning `The restart command has been successfully executed`.

    ```
    Sep 21 21:30:08 d2dcf7b3095f tedge-agent[189]: 2023-09-21T21:30:08.072623262Z  INFO tedge_agent::restart_manager::actor: Restart command has been interrupted by a signal: Command { std: "/usr/bin/sudo" "/usr/bin/on_shutdown_signal_interrupt.sh", kill_on_drop: false }
    Sep 21 21:30:08 d2dcf7b3095f tedge-agent[189]: 2023-09-21T21:30:08.072724928Z  INFO tedge_agent::restart_manager::actor: The restart command has been successfully executed
    ```

albinsuresh (commented):
The code looks fine except for a small issue that I've highlighted around failed state repository updates, that's not handled cleanly(even in the old code). Happy to approve once that's fixed.

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -51,24 +52,44 @@ impl Actor for RestartManagerActor {
...
+                            info!("As requested, a shutdown has been triggered");
+                            return Ok(());
+                        }
+                        Ok(None) | Err(_ /* timeout */) => {
+                            // Something went wrong. The process should have been shutdown by the restart.
+                            let error = "No shutdown has been triggered".to_string();
+                            error!(error);
+                            self.handle_error(request, error).await?;
+                        }
                     }
```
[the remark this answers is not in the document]
albinsuresh: > From his perspective, the successful execution of the script should have translated to a Success, right?

I guess this is the root of the problem. When it's a user-overridden script, we can't always assume that the device would have been restarted, and wait for the agent to reboot and really validate if the restart succeeded and then only send the `Successful` status. But, purely relying on the exit status is not sufficient for our default restart script either. So, it's like a chicken and egg problem.

If we can always assume that the user-provided scripts must restart the device, then this is fine. Else we should expect their custom script to send the `successful` status update MQTT message as well, at the end of their "non-restarting" logic.

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
```diff
@@ -96,26 +109,45 @@ impl TedgeOperationConverterActor {
...
-                        self.restart_sender.send(request).await?;
-                    }
-                    Err(err) => error!("Incorrect restart request payload: {err}"),
+            _ => {
+                // Not a tedge/commands !
+            }
+        }
+
+        if message.topic.name.as_str().starts_with("tedge") {
+            return Ok(());
```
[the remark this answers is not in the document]
didier-wenzek: No this is temporary code. The first part of the method is related to `tedge` topics, the second part to `te` topics.

I could have write this differently, but came with this not so nice return to avoid confusing diffs in the PR.

didier-wenzek:
>    * Log entry is misleading when the restart command is killed by a signal.
The entry with `Restart command has been interrupted by a signal` is good,
but it is proceeded by an entry with the opposite meaning `The restart command has been successfully executed`.

Addressed by https://github.com/thin-edge/thin-edge.io/pull/2245/commits/6cb2de650fcc4446512a93bece631817b829c21e

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -51,24 +52,44 @@ impl Actor for RestartManagerActor {
         }
 
         while let Some(request) = self.message_box.recv().await {
-            let executing_response = self.update_state_repository(&request).await;
+            if request.status() != CommandStatus::Init {
+                continue;
+            }
+            info!("Triggering a restart");
+            let executing_response = self.update_state_repository(request.clone()).await;
```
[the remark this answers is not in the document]
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/2245/commits/97a07a0e33a30e6ef745f71053f3cfecc67d8e96

crates/core/tedge_agent/src/restart_manager/actor.rs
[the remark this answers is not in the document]
didier-wenzek: > What if the user intentionally replaced the restart script with something that doesn't really restart the device but does something else?

The user should not be surprised that the *restart*  command fails ;-)

> If we can always assume that the user-provided scripts must restart the device, then this is fine.

We can and we do. Even more, we check that this is the case: if the agent doesn't shutdown than restart after a reboot of the device, then command execution is reported as failed.

tests/RobotFramework/tests/cumulocity/restart/restart_device_child.robot
```diff
@@ -0,0 +1,110 @@
...
+    ${operation}=    Cumulocity.Restart Device
+    Operation Should Be SUCCESSFUL    ${operation}    timeout=180
+
+Restart operation should be set to failed when an non-existent command is configured
+    Set Restart Command    ["/usr/bin/on_shutdown_does_not_exist.sh"]
+    Set Restart Timeout    60
+    ${operation}=    Cumulocity.Restart Device
+    Operation Should Be FAILED    ${operation}    failure_reason=Restart Failed: Fail to trigger a restart.*    timeout=30
+
+Restart operation should be set to failed when command is not allows by sudo
```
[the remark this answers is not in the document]
didier-wenzek: fixed

crates/core/tedge_api/src/messages.rs
```diff
@@ -473,65 +477,143 @@ impl From<SoftwareError> for Option<SoftwareModuleItem> {
...
-    pub fn with_status(self, status: OperationStatus) -> Self {
-        Self { status, ..self }
+    /// Return the MQTT topic for this command
+    fn topic(&self, schema: &MqttSchema) -> Topic {
+        schema.topic_for(self.topic_id(), &self.channel())
     }
 
-    pub fn topic() -> Topic {
-        Topic::new_unchecked(DEVICE_RESTART_RESPONSE_TOPIC)
+    /// Return the MQTT Message to register `restart` commands on a given target device
```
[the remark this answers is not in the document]
didier-wenzek: fixed

crates/core/tedge_api/src/messages.rs
```diff
@@ -473,65 +477,143 @@ impl From<SoftwareError> for Option<SoftwareModuleItem> {
...
+        );
+        let payload = "{}";
+        Message::new(&meta_topic, payload)
+            .with_retain()
+            .with_qos(QoS::AtLeastOnce)
     }
 
-    pub fn status(&self) -> OperationStatus {
-        self.status
+    /// Return the MQTT payload for this command
```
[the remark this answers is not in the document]
didier-wenzek: fixed

reubenmiller (approved):
The requested log message changes were resolved.

Approved

Issue #2236, opened 2023-09-07
tedge-mapper-aws prints errors for known non-json mosquitto bridge health status messages


reubenmiller:
**Describe the bug**

The tedge-mapper-aws service prints errors on mqtt payloads which aren't valid json. This is a problem because the mosquitto bridge health status is published as a single integer `1` or `0` which is the most likely root cause of these messages.

```sh
$ journalctl -fu tedge-mapper-aws -n 100
-- Journal begins at Tue 2023-08-15 14:24:36 CEST. --
Sep 08 00:54:04 lumber tedge-mapper[32138]: 2023-09-07T22:54:04.089756196Z ERROR aws_mapper_ext::converter: Mapping error: invalid type: integer `0`, expected a map at line 1 column 1
Sep 08 00:54:04 lumber tedge-mapper[32138]: 2023-09-07T22:54:04.08980177Z ERROR aws_mapper_ext::converter: Mapping error: invalid type: integer `0`, expected a map at line 1 column 1
```

**To Reproduce**

1. Install thin-edge.io
2. Configure the tedge-mapper-c8y to connect to AWS (this [gist](https://gist.github.com/reubenmiller/0aa9dfb6fabbfdbe76156e9c3af4f33f) can help guide you through the setup)
3. Publish a message to simulate the mosquitto bridge health message

    ```sh
    tedge mqtt pub tedge/health/mosquitto-aws-bridge 0
    ```
4. Inspect the tedge-mapper-aws logs for any error messages related to mapping/parsing errors
    ```sh
    journalctl -u tedge-mapper-aws -n 50
    ```

**Expected behavior**

The aws mapper should ignore parsing errors for known topics like `tedge/health/mosquitto-aws-bridge` and `tedge/health/mosquitto-c8y-bridge` etc.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `Debian 11 (bullseye)`
 - Hardware [incl. revision]: `Raspberry Pi 4`
 - System-Architecture [e.g. result of "uname -a"]: `Linux lumber 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux`
 - thin-edge.io version [e.g. 0.1.0]: `tedge 0.12.1~237+g7f58cd2`

**Additional context**

gligorisaev:
QA has thoroughly checked the feature and here are the results:

[...]

PR #2247, opened 2023-09-12
tedge-mapper-aws prints errors for known non-json mosquitto bridge health status messages

PradeepKiruvale:
## Proposed changes

Since Aws and Azure mappers do not support `health` messages. Respective converters must not translate them.
The health messages include the `bridge and daemon health` messages.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2236

## Checklist


PradeepKiruvale:
> @PradeepKiruvale Wait, we still want to support tedge health messages, we just have to handle the mosquitto bridge related messages properly (I believe we are just ignoring the bridge messages in the tedge-mapper-c8y) without throwing an error.

Agree, I have reverted those changes. Now only avoid traslating the bridge status.

crates/extensions/aws_mapper_ext/src/converter.rs
```diff
@@ -37,44 +37,51 @@ impl AwsConverter {
     }
 
     fn try_convert(&mut self, input: &MqttMessage) -> Result<Vec<MqttMessage>, ConversionError> {
-        let default_timestamp = self.add_timestamp.then(|| self.clock.now());
```
didier-wenzek, line 40: [Minor] It could be nice to minimize the diff.

Here, one way to do that is to return early when this is a bridge topic.

```
// Ignore health status message coming from the bridges
if input.topic.name.starts_with("tedge/health") && input.topic.name.ends_with("bridge") {
    return vec![];
}

// ... former code unchanged
```

crates/extensions/aws_mapper_ext/src/converter.rs
```diff
@@ -37,44 +37,51 @@ impl AwsConverter {
...
-            let mut serializer = ThinEdgeJsonSerializer::new_with_timestamp(default_timestamp);
-            tedge_api::parser::parse_str(input.payload_str()?, &mut serializer)?;
-
-            serializer.into_string()?
-        } else if input.topic.name.starts_with("tedge/events")
-            || input.topic.name.starts_with("tedge/alarms")
-            || input.topic.name.starts_with("tedge/health")
+        if input.topic.name.eq("tedge/health/mosquitto-az-bridge")
+            || input.topic.name.eq("tedge/health/mosquitto-aws-bridge")
+            || input.topic.name.eq("tedge/health/mosquitto-c8y-bridge")
```
didier-wenzek, line 42: Better to avoid exact matches on a the *current* list of bridge, as adding a new one would break again this mapper.

PradeepKiruvale: fixed, added a function to check the bridge dynamically.

PR #2249, opened 2023-09-12
Don't log authentication configuration when connecting to MQTT broker

Bravo555:
PR #2233 introduced some `info!` logging statements to MQTT connection code, so that the user could see what's happening.

However, this code printed out the contents of the AuthenticationConfig struct which contained root certificate store and a client private key if client authentication option was used.

The intent was to print out the certificate paths, but what was printed was the contents, potentially including secrets.

Fortunately it was caught quickly by @reubenmiller.

I think the biggest factor that contributed to this mistake was keeping secrets together with other safe-to-print fields, without using any marker type to denote that `ClientAuthConfig.key` is a secret, disabling it's `Debug` implementation so that it can't be accidentally printed out in the first place. Crates like [secrecy](https://docs.rs/secrecy/latest/secrecy/) can be used to replace `Display` and `Debug` impls so that sensitive information cannot be printed.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/config.rs
+++ b/crates/common/mqtt_channel/src/config.rs
@@ -5,12 +5,13 @@ use certificate::CertificateError;
 use log::debug;
 use rumqttc::tokio_rustls::rustls;
 use rumqttc::tokio_rustls::rustls::Certificate;
-use rumqttc::tokio_rustls::rustls::PrivateKey;
 use rumqttc::LastWill;
 use std::fmt::Debug;
 use std::fmt::Formatter;
+use std::ops::Deref;
 use std::path::Path;
 use std::sync::Arc;
+use zeroize::Zeroizing;
 
 /// Configuration of an MQTT connection
 #[derive(Debug, Clone)]
@@ -107,7 +108,7 @@ impl Default for AuthenticationConfig {
 #[derive(Clone)]
 struct ClientAuthConfig {
     cert_chain: Vec<Certificate>,
-    key: PrivateKey,
+    key: Zeroizing<PrivateKey>,
 }
 
 impl Debug for ClientAuthConfig {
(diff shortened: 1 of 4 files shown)


didier-wenzek (commented):
To avoid repeating this mistake, I would remove the culprit `derive(Debug)`.

```
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 struct ClientAuthConfig {
     cert_chain: Vec<Certificate>,
     key: PrivateKey,
 }
 
+impl Debug for ClientAuthConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        // Do not print the private key!
+        self.cert_chain.fmt(f)
+    }
+}

```

Bravo555:
> To avoid repeating this mistake, I would remove the culprit derive(Debug).

fixed in b6a13abcb556367c71efdee5a9ad6ee04767e72d.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -103,10 +105,27 @@ impl Default for AuthenticationConfig {
...
+impl Debug for ClientAuthConfig {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ClientAuthConfig")
+            .field("cert_chain", &self.cert_chain)
+            .finish()
+    }
+}
+
+#[derive(Debug, Clone)]
+struct PrivateKey(rustls::PrivateKey);
```
didier-wenzek, line 123:
```suggestion
#[derive(Clone)]
struct PrivateKey(rustls::PrivateKey);
```

didier-wenzek (approved):
Approved. Cargo.lock has to be updated before merge.

reubenmiller (approved):
Approved

The log entries look much better now:

```
Sep 13 06:47:08 ad0bf762eaeb c8y-firmware-plugin[969]: 2023-09-13T06:47:08.91651Z  INFO mqtt_channel::connection: MQTT connecting to broker: host=localhost:8883, session_name=Some("c8y-firmware-plugin")
```

PR #2250, opened 2023-09-12
Replace device.type config option with an enum

Bravo555:
## Proposed changes

This PR replaces `device.type` config option with an enum with 2 values: `ThinEdgeDevice` and `ChildDevice`.

I didn't know what this option is for, and if it could be used to determine whether a device we're running on is a child device or not, or if there are any other values valid or invalid for this option at all, as `type` name suggests that not every valid string may be a valid type. Tried to see if the set of possible values for this setting is in the thin-edge Documentation, I found [a table on a health monitoring page](https://thin-edge.github.io/thin-edge.io/operate/c8y/c8y_service_monitoring/#configuring-the-default-service-type) that seemed to suggest that `Thin-edge-device` and `Child-device` are valid values for this setting. But I also found [a Cumulocity Smartrest 2.0 reference section](https://cumulocity.com/guides/reference/smartrest-two/#100), according to which device type can be any string, but I still have no idea what it means.

The tests are failing though, so it seems this setting is also used in other ways.

As such, I think that we should discuss if this is intended state, and if so, maybe adding some documentation to clarify the purpose of this setting. Also if this setting should remain in its current state, where any string is a valid value, I'd like to ask if there is any other setting to determine if a current device is a child device, or should such settings be added?

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/models/device_type.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/models/device_type.rs
@@ -1,0 +1,49 @@
+use std::fmt::Display;
+use std::str::FromStr;
+
+use doku::Document;
+use serde::Deserialize;
+use serde::Serialize;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Document)]
+pub enum DeviceType {
+    ThinEdgeDevice,
+    ChildDevice,
+}
+
+impl Default for DeviceType {
+    fn default() -> Self {
+        Self::ThinEdgeDevice
+    }
+}
+
+impl FromStr for DeviceType {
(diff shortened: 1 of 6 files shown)


reubenmiller (changes requested):
No the device.type setting is used by users to add there own custom type names to the device. The other type to determine what thin-edge.io entity type it is (e.g. main device, child device) is something separate, which has nothing to do with the `device.type` setting.

The `device.type` should be a free-form setting so that users can encode whatever useful information they deem necessary.

didier-wenzek:
So `device.type` is a category name chosen by the user to classify devices of a large fleet. The user can choose to set this type to convey the device role, os, hardware ... Hence, this must be a `String`.

PR #2252, opened 2023-09-13
remove trailing full stop in log entries as it is unnecessary

reubenmiller:
## Proposed changes

Remove trailing full stop in log entries as it is unnecessary. Also when a full stop is after a value in the log entry, it makes it hard to copy from as the shell highlighting will include the full stop as well.

The only downside to not having the full stop is that if the value has a whitespace after the last visible character, then it will not be obvious in the log entry, however improving the copy/paste of the value trumps the less common trailing whitespace scenario.

**Before**

```
INFO tedge_log_manager::actor: Log request processed for log type: example.
```

**After**

```
INFO tedge_log_manager::actor: Log request processed for log type: example
```

And example showing that it is annoying to copy just the value by double-clicking in the shell to select the value shown in the log entry.

image

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_firmware_manager/src/actor.rs
+++ b/crates/extensions/c8y_firmware_manager/src/actor.rs
@@ -433,7 +433,7 @@ impl FirmwareManagerActor {
                     .insert(operation_key.clone(), ActiveOperationState::Executing);
             }
             None => {
-                info!("Received a response from {child_id} for unknown request {operation_id}.");
+                info!("Received a response from {child_id} for unknown request {operation_id}");
                 return Ok(());
             }
         }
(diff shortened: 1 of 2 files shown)


albinsuresh (approved):
LGTM. Any specific reason why you handpicked just these log statements? Or these are the only ones we've got with trailing `.`?

reubenmiller:
> LGTM. Any specific reason why you handpicked just these log statements? Or these are the only ones we've got with trailing `.`?

These were the only ones I found which were directly in a log message. There were some other possible hits but they were in error definitions and I was unsure how these get rendered together with other information. As some instances having a full stop is ok if it is a sentence (and some errors are more like sentences)

PR #2257, opened 2023-09-13
Add test to check if incorrect MQTT payload is going to close the MQTT connection

gligorisaev:
Add test to check if incorect MQTT payload is going to close the MQTT connection

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/MQTT_health_check/mqtt_connection_status_with_incorrect_payload.robot
+++ b/tests/RobotFramework/tests/MQTT_health_check/mqtt_connection_status_with_incorrect_payload.robot
@@ -1,0 +1,25 @@
+*** Settings ***
+Resource    ../../resources/common.resource
+Library    ThinEdgeIO
+
+Test Tags    theme:troubleshooting    theme:monitoring    theme:c8y
+Suite Setup       Setup
+Suite Teardown    Get Logs
+
+
+*** Test Cases ***
+
+Send incorrect mqtt payload
+    Execute Command    tedge mqtt pub 'tedge/commands/req/software/list' '{"ids": "100"}'
+    ${output}=    Execute Command    systemctl status tedge-agent.service
+    Should Not Contain    ${output}    INFO mqtt_channel::connection: MQTT connection closed
+
+Send incorrect mqtt payload again
+    Execute Command    tedge mqtt pub 'tedge/commands/req/software/list' '{"ids": "100"}'
+    ${output}=    Execute Command    systemctl status tedge-agent.service
+    Should Not Contain    ${output}    INFO mqtt_channel::connection: MQTT connection closed
+
+Send incorrect mqtt payload third time
+    Execute Command    tedge mqtt pub 'tedge/commands/req/software/list' '{"ids": "100"}'
+    ${output}=    Execute Command    systemctl status tedge-agent.service
+    Should Not Contain    ${output}    INFO mqtt_channel::connection: MQTT connection closed


tests/RobotFramework/tests/MQTT_health_check/mqtt_connection_status_with_incorrect_payload.robot
```diff
@@ -0,0 +1,25 @@
...
+Suite Setup       Setup
+Suite Teardown    Get Logs
+
+
+*** Test Cases ***
+
+Send incorrect mqtt payload
+    Execute Command    tedge mqtt pub 'tedge/commands/req/software/list' '{"ids": "100"}'
+    ${output}=    Execute Command    systemctl status tedge-agent.service
+    Should Not Contain    ${output}    INFO mqtt_channel::connection: MQTT connection closed
```
reubenmiller, line 15: This might be a little too fragile, as the `systemctl status` command only returns the last ~10 lines. A better approach would be to use `journalctl` and its date filtering options, though there are a lot of potential pitfalls here (which they should be addressed by the following suggestion).

```suggestion
    ${start_time}=    ThinEdgeIO.Get Unix Timestamp
    Execute Command    tedge mqtt pub 'tedge/commands/req/software/list' '{"ids": "100"}'
    ${output}=    Execute Command    systemctl status tedge-agent.service
    Sleep    5s    reason=Allow sufficient time for the service to fail before checking the logs
    ${output}=    Execute Command    journalctl -u tedge-agent --since @${start_time} -n 1000
    Should Not Contain    ${output}    INFO mqtt_channel::connection: MQTT connection closed
```

It is still not ideal because using log entries in assertions will eventually break if we change the formatting of them, however due to a lack of any other option, we'll have to live with this method for now.

And this same change should be extend to the other test cases here.

Ruadhri17 (commented):
Apart from the comment above it looks good to me, however I have mixed feelings about this PR:

The issue was caused by incorrect error handling (in this case it was the problem with the payload parsing). The similar issue may occur if ANY of the error handlings will be implemented incorrectly. Moreover, we only covered `tedge-agent` case while we should cover all modules that use MQTT actor system. And that is a lot of test cases.

reubenmiller:
> Apart from the comment above it looks good to me, however I have mixed feelings about this PR:
> 
> The issue was caused by incorrect error handling (in this case it was the problem with the payload parsing). The similar issue may occur if ANY of the error handlings will be implemented incorrectly. Moreover, we only covered `tedge-agent` case while we should cover all modules that use MQTT actor system. And that is a lot of test cases.

Yes I agree that it definitely isn't optimal, though I'm ok with only using an integration test for the tedge-agent as this component will be the main component (or only component) once we consolidate the components into one or two binaries, therefore increasing the coverage of this test.

If we can sufficiently prove that a unit test sufficiently covers this scenario then I'm open to excluding this integration test.

gligorisaev:
I have no problems to exclude this integration test (or also to extend it for the whole scope) if we have "some other" mechanisam to check this issue, but I will not exclude it only bacause " ANY of the error handlings will be implemented incorrectly" or especially not because "Moreover, we only covered tedge-agent case while we should cover all modules that use MQTT actor system. And that is a lot of test cases". So lets decide if this issue is coverend with, for example, unit test

Ruadhri17:
> I have no problems to exclude this integration test (or also to extend it for the whole scope) if we have "some other" mechanisam to check this issue, but I will not exclude it only bacause " _**ANY of the error handlings will be implemented incorrectly**_" or especially not because "**_Moreover, we only covered tedge-agent case while we should cover all modules that use MQTT actor system. And that is a lot of test cases_**". So lets decide if this issue is coverend with, for example, unit test

Definitely not exclude it. But as we cover those error handlings with unit tests I think it is worth checking if we can also use them to verify if MQTT connection is not closed on error, though it needs to be examined thoroughly. As I now work on `tedge-configuration-plugin` I can check if those approach is suitable for our problem.

didier-wenzek:
System tests can push more stress on thin-edge than unit tests can do.

Pushing a bush of random messages on te/# topics before a test case should not make this case fails.

gligorisaev:
Obsolite

Issue #2174, opened 2023-08-20
tedge-agent fails to download binary from Cumulocity IoT due to 401 error


reubenmiller:
**Describe the bug**

The thin-edge.io software management plugin fails to download a binary stored in Cumulocity IoT if the JWT has expired (and this error does not trigger a renewal of the token).

**To Reproduce**

1. Start thin-edge.io and its components (only mosquitto, tedge-agent and the tedge-mapper-c8y are required)
2. Wait until the Cumulocity IoT token requested on startup as expired. On Cumulocity eu-latest the expiration is set to 1 hour (though decode the JWT to check the expiration timestamp)
3. Create a software package and version in Cumulocity IoT and upload a binary which is hosted within Cumulocity IoT (don't use an external URL)
4. On the thin-edge.io device trigger a software update using the software package uploaded in the previous step
5. If no bug is observed, the operation should be set to SUCCESSFUL. With the bug the operation is set to FAILED, and the error "Could not make a successful request to the remote server" is shown in the software update logs, as well as a corresponding error ("HTTP status client error (401 Unauthorized) for url") in the tedge-agent logs.

**Expected behavior**

* the tedge-agent should renew the Cumulocity IoT token if required so that the binary hosted in Cumulocity IoT can be downloaded successfully

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `Alpine Linux 3.18`
 - Hardware [incl. revision]: Docker container running on `aarch64`
 - System-Architecture [e.g. result of "uname -a"]: `Linux cb069e118bbc 5.15.68-0-virt #1-Alpine SMP Fri, 16 Sep 2022 06:29:31 +0000 aarch64 Linux`
 - thin-edge.io version [e.g. 0.1.0]: `0.12.1~99+g3e0fe253`

**Additional context**

The following shows the logs from the failed attempt (with the tokens redacted):

**file: tedge-mapper-c8y**

```log
2023-08-20T09:20:35.983272265Z  INFO MQTT: send Message { topic: Topic { name: "c8y/s/ds" }, payload: "528,rmi_apk_demo001,example,1.2.3::apk,https://t493319102.eu-latest.cumulocity.com/inventory/binaries/29251966,install", qos: AtLeastOnce, retain: false }    
[...]

PR #2261, opened 2023-09-13
Add c8y auth proxy

jarhodes314:
## Proposed changes
This adds a proxy server for Cumulocity to the mapper which automatically authenticates requests using a JWT, intended to address the issues we've observed with 401s from JWT caching.

## Types of changes

## Paste Link to the issue
#2174 

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ async-tungstenite = { version = "0.23", features = [
     "tokio-rustls-native-certs",
 ] }
 aws_mapper_ext = { path = "crates/extensions/aws_mapper_ext" }
+axum = "0.6"
 az_mapper_ext = { path = "crates/extensions/az_mapper_ext" }
 backoff = { version = "0.4", features = ["tokio"] }
 base64 = "0.13"
@@ -45,6 +46,7 @@ c8y_api = { path = "crates/core/c8y_api" }
 c8y_config_manager = { path = "crates/extensions/c8y_config_manager" }
 c8y_firmware_manager = { path = "crates/extensions/c8y_firmware_manager" }
 c8y_http_proxy = { path = "crates/extensions/c8y_http_proxy" }
+c8y_auth_proxy = { path = "crates/extensions/c8y_auth_proxy" }
 c8y_log_manager = { path = "crates/extensions/c8y_log_manager" }
 c8y_mapper_ext = { path = "crates/extensions/c8y_mapper_ext" }
 camino = "1.1"
--- a/crates/core/c8y_api/src/http_proxy.rs
+++ b/crates/core/c8y_api/src/http_proxy.rs
@@ -101,7 +101,7 @@ impl C8yEndPoint {
 
         let (_, host) = url_host.split_once('.').unwrap_or((url_host, ""));
         let (_, c8y_host) = tenant_uri.split_once('.').unwrap();
-        (host == c8y_host).then(|| url)
+        (host == c8y_host).then_some(url)
     }
 }
 
--- a/crates/core/c8y_api/src/http_proxy.rs
+++ b/crates/core/c8y_api/src/http_proxy.rs
@@ -89,29 +89,19 @@ impl C8yEndPoint {
         url_event_binary
     }
 
-    pub fn url_is_in_my_tenant_domain(&self, url: &str) -> bool {
+    pub fn possible_tenant_url(&self, url: &str) -> Option<Url> {
         // c8y URL may contain either `Tenant Name` or Tenant Id` so they can be one of following options:
         // * <tenant_name>.<domain> eg: sample.c8y.io
         // * <tenant_id>.<domain> eg: t12345.c8y.io
         // These URLs may be both equivalent and point to the same tenant.
         // We are going to remove that and only check if the domain is the same.
         let tenant_uri = &self.c8y_host;
-        let url_host = match Url::parse(url) {
-            Ok(url) => match url.host() {
-                Some(host) => host.to_string(),
-                None => return false,
-            },
-            Err(_err) => {
-                return false;
-            }
-        };
+        let url = Url::parse(url).ok()?;
+        let url_host = url.domain()?;
 
-        let url_domain = url_host.splitn(2, '.').collect::<Vec<&str>>();
-        let tenant_domain = tenant_uri.splitn(2, '.').collect::<Vec<&str>>();
-        if url_domain.get(1) == tenant_domain.get(1) {
-            return true;
-        }
-        false
+        let (_, host) = url_host.split_once('.').unwrap_or((url_host, ""));
+        let (_, c8y_host) = tenant_uri.split_once('.').unwrap();
+        (host == c8y_host).then(|| url)
     }
 }
 
@@ -226,7 +216,7 @@ mod tests {
     #[test_case("https://t1124124.test.com/path/to/file")]
     fn url_is_my_tenant_correct_urls(url: &str) {
         let c8y = C8yEndPoint::new("test.test.com", "test_device");
-        assert!(c8y.url_is_in_my_tenant_domain(url));
+        assert_eq!(c8y.possible_tenant_url(url), Some(url.parse().unwrap()));
     }
 
     #[test_case("test.com")]
@@ -236,7 +226,7 @@ mod tests {
     #[test_case("http://test.com::12345")]
     fn url_is_my_tenant_incorrect_urls(url: &str) {
         let c8y = C8yEndPoint::new("test.test.com", "test_device");
-        assert!(!c8y.url_is_in_my_tenant_domain(url));
+        assert!(c8y.possible_tenant_url(url).is_none());
     }
 
     #[test]
--- a/crates/core/c8y_api/src/http_proxy.rs
+++ b/crates/core/c8y_api/src/http_proxy.rs
@@ -89,7 +89,7 @@ impl C8yEndPoint {
         url_event_binary
     }
 
-    pub fn possible_tenant_url(&self, url: &str) -> Option<Url> {
+    pub fn maybe_tenant_url(&self, url: &str) -> Option<Url> {
         // c8y URL may contain either `Tenant Name` or Tenant Id` so they can be one of following options:
         // * <tenant_name>.<domain> eg: sample.c8y.io
         // * <tenant_id>.<domain> eg: t12345.c8y.io
@@ -216,7 +216,7 @@ mod tests {
     #[test_case("https://t1124124.test.com/path/to/file")]
     fn url_is_my_tenant_correct_urls(url: &str) {
         let c8y = C8yEndPoint::new("test.test.com", "test_device");
-        assert_eq!(c8y.possible_tenant_url(url), Some(url.parse().unwrap()));
+        assert_eq!(c8y.maybe_tenant_url(url), Some(url.parse().unwrap()));
     }
 
     #[test_case("test.com")]
@@ -226,7 +226,7 @@ mod tests {
     #[test_case("http://test.com::12345")]
     fn url_is_my_tenant_incorrect_urls(url: &str) {
         let c8y = C8yEndPoint::new("test.test.com", "test_device");
-        assert!(c8y.possible_tenant_url(url).is_none());
+        assert!(c8y.maybe_tenant_url(url).is_none());
     }
 
     #[test]
--- a/crates/extensions/c8y_auth_proxy/src/actor.rs
+++ b/crates/extensions/c8y_auth_proxy/src/actor.rs
@@ -1,0 +1,104 @@
+use std::convert::Infallible;
+use std::net::IpAddr;
+
+use axum::async_trait;
+use c8y_http_proxy::credentials::C8YJwtRetriever;
+use c8y_http_proxy::credentials::JwtRetriever;
+use futures::channel::mpsc;
+use futures::StreamExt;
+use tedge_actors::Actor;
+use tedge_actors::Builder;
+use tedge_actors::DynSender;
+use tedge_actors::RuntimeError;
+use tedge_actors::RuntimeRequest;
+use tedge_actors::RuntimeRequestSink;
+use tedge_actors::Sequential;
+use tedge_actors::ServerActorBuilder;
+use tedge_config::ConfigNotSet;
+use tedge_config::TEdgeConfig;
+use tracing::info;
+
+use crate::server::AppState;
+use crate::server::Server;
+use crate::tokens::TokenManager;
+
+type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+pub struct C8yAuthProxyBuilder {
+    app_state: AppState,
+    bind_address: IpAddr,
+    bind_port: u16,
+    signal_sender: mpsc::Sender<RuntimeRequest>,
+    signal_receiver: mpsc::Receiver<RuntimeRequest>,
+}
+
+impl C8yAuthProxyBuilder {
+    pub fn try_from_config(
+        config: &TEdgeConfig,
+        jwt: &mut ServerActorBuilder<C8YJwtRetriever, Sequential>,
+    ) -> Result<Self, ConfigNotSet> {
+        let app_state = AppState {
+            target_host: format!("https://{}", config.c8y.http.or_config_not_set()?).into(),
+            token_manager: TokenManager::new(JwtRetriever::new("C8Y-PROXY => JWT", jwt)).shared(),
+        };
+        let bind = &config.c8y.proxy.bind;
+        let (signal_sender, signal_receiver) = mpsc::channel(10);
+
+        Ok(Self {
+            app_state,
+            bind_address: bind.address,
+            bind_port: bind.port,
+            signal_sender,
+            signal_receiver,
+        })
+    }
+}
+
+impl Builder<C8yAuthProxy> for C8yAuthProxyBuilder {
+    type Error = Infallible;
+
+    fn try_build(self) -> Result<C8yAuthProxy, Self::Error> {
+        Ok(C8yAuthProxy {
+            app_state: self.app_state,
+            bind_address: self.bind_address,
+            bind_port: self.bind_port,
+            signal_receiver: self.signal_receiver,
+        })
+    }
+}
+
+impl RuntimeRequestSink for C8yAuthProxyBuilder {
+    fn get_signal_sender(&self) -> DynSender<RuntimeRequest> {
+        Box::new(self.signal_sender.clone())
+    }
+}
+
+pub struct C8yAuthProxy {
+    app_state: AppState,
+    bind_address: IpAddr,
+    bind_port: u16,
+    signal_receiver: mpsc::Receiver<RuntimeRequest>,
+}
+
+#[async_trait]
+impl Actor for C8yAuthProxy {
+    fn name(&self) -> &str {
+        "C8yAuthProxy"
+    }
+
+    async fn run(&mut self) -> Result<(), RuntimeError> {
+        let server = Server::try_init(self.app_state.clone(), self.bind_address, self.bind_port)
+            .map_err(BoxError::from)?
+            .wait();
+        tokio::select! {
+            result = server => {
+                info!("Done");
+                Ok(result.map_err(BoxError::from)?)
+            },
+            Some(RuntimeRequest::Shutdown) = self.signal_receiver.next() => {
+                info!("Shutdown");
+                Ok(())
+            }
+        }
+    }
+}
--- a/crates/extensions/c8y_auth_proxy/src/server.rs
+++ b/crates/extensions/c8y_auth_proxy/src/server.rs
@@ -1,0 +1,307 @@
+use crate::tokens::*;
+use axum::body::Body;
+use axum::body::Full;
+use axum::body::StreamBody;
+use axum::extract::FromRef;
+use axum::extract::State;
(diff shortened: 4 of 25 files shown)


crates/extensions/c8y_auth_proxy/src/url.rs
```diff
@@ -0,0 +1,51 @@
...
+        }
+    }
+
+    pub fn proxy_url(&self, cumulocity_url: &str) -> miette::Result<url::Url> {
+        let mut parsed_url = cumulocity_url
+            .parse::<url::Url>()
+            .into_diagnostic()
+            .wrap_err_with(|| format!("failed to parse url: {cumulocity_url}"))?;
+        parsed_url
+            .set_host(Some("localhost"))
```
didier-wenzek, line 26: No, this must be an IP address accessible from child-devices. See comment on `pub struct ProxyUrlGenerator`.

crates/extensions/c8y_auth_proxy/src/tokens.rs
```diff
@@ -0,0 +1,43 @@
...
+        Self { recv, cached: None }
+    }
+
+    pub fn shared(self) -> SharedTokenManager {
+        SharedTokenManager(Arc::new(Mutex::new(self)))
+    }
+}
+
+impl TokenManager {
+    async fn not_matching(&mut self, input: Option<&str>) -> Arc<str> {
```
didier-wenzek, line 31: What is done by this method is mysterious.

Provided with the cached token, it returns a fresh one.
And provided with a silly token, it returns the cached one.

What is the point?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -465,22 +468,20 @@ impl CumulocityConverter {
...
-                        if self.c8y_endpoint.url_is_in_my_tenant_domain(url.url()) {
-                            module.url = module.url.as_ref().map(|s| {
-                                DownloadInfo::new(&s.url).with_auth(Auth::new_bearer(&token))
-                            });
+                    if let Some(url) = &mut module.url {
+                        *url = if self.c8y_endpoint.url_is_in_my_tenant_domain(url.url()) {
+                            // TODO what do we do if this fails?
+                            DownloadInfo::new(
+                                self.auth_proxy.proxy_url(url.url()).unwrap().as_ref(),
+                            )
```
didier-wenzek, line 481: What are the possible cause of failures preventing the generation of the appropriate url? This seems to be only a matter of substituting a host by another.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -0,0 +1,307 @@
...
+    let handle = get(respond_to)
+        .post(respond_to)
+        .put(respond_to)
+        .patch(respond_to)
+        .delete(respond_to)
+        .options(respond_to);
+    let app = Router::new()
+        .route("/", handle.clone())
+        .route("/*path", handle)
+        .with_state(state);
```
didier-wenzek, line 56: So the current proposal is to proxy `https://didier.latest.stage.c8y.io` directly on `http://local-thin-edge-device/`.

I would consider to add some `c8y` prefix. The idea being to use the same HTTP server to serve local file-transfer requests.

@reubenmiller @albinsuresh What do you think here.

didier-wenzek (commented):
Thank you. Definitely in the correct direction.

Questions we will have to address:

- Do we add some `c8y` prefix to thin-edge local urls.
- Can `c8y_http_proxy` delegate to `c8y_auth_proxy` all the authentications issues?
- Can `c8y_auth_proxy`  also implement the `file_transfer` service?

crates/extensions/c8y_auth_proxy/src/server.rs
reubenmiller: Yes having a `c8y` route makes sense, and it would be consistent to the way the mapper specific topics work, and allow for extending it to other cloud platforms in the future (if similar API exists).

albinsuresh: Agree that it's good to have that prefix to support different services in future as well.

But, question on the prospect of using the same server for file-transfer service as well: The file-transfer service is a generic cloud-agnostic service hosted rightly by tedge-agent, and **not** the mapper, right? And this service and any future cloud proxy services belongs to their own respective mappers. So, how can we combine both?

crates/extensions/c8y_mapper_ext/src/converter.rs
jarhodes314: I think in practice the only way this will fail is if the URL passed in isn't valid, but if that's the case, we won't consider it a Cumulocity URL and therefore this code will never be called. So I think unwrapping is good here, but it might be worth considering whether this method `proxy_url` wants to take a URL struct rather than parsing the URL internally. That way, we could do the conversion here and it would be clearer why to unwrap (or `url_is_in_my_tenant_domain` could be modified to return `Option<url::Url>` in the case where it is a Cumulocity URL so we entirely avoid parsing it twice).

crates/extensions/c8y_auth_proxy/src/tokens.rs
jarhodes314: I think I was trying to be a bit unnecessarily clever here. The goal was that if we receive a request and that request fails, we can call this method with the token that failed, and if another request has caused the token to be regenerated in the meantime, we will just return the new cached token rather than making a new JWT request.

didier-wenzek: I got it!

Indeed unnecessary clever. But it's okay. Just put you full response as a comment: "The goal is ...".

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: > it might be worth considering whether this method `proxy_url` wants to take a URL struct rather than parsing the URL internally.

This would indeed help error handling and to avoid confusion between "failing to parse the input url" and "failing to substitute the host".

Cargo.toml
```diff
@@ -37,6 +37,7 @@ async-tungstenite = { version = "0.23", features = [
     "tokio-rustls-native-certs",
 ] }
 aws_mapper_ext = { path = "crates/extensions/aws_mapper_ext" }
+axum = "0.6"
```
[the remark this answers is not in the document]
jarhodes314: `routerify` seems to have not been maintained since 2021, this has particular implications with https://github.com/routerify/stream-body, which depends on tokio 0.2. It's clearly possible to address this situation, ourselves, but I'm not convinced it's worth the effort just to avoid a new dependency. I also prefer axum due to its preference for the witness pattern around managed state, and the fact that is checked at compile time rather than at runtime. If we want the two services to use the same library, I think it makes more sense to use `axum` for the file_transfer service than routerify for this.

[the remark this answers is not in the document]
didier-wenzek: I'm okay for using `axum` for the `c8y_auth_proxy` and to consider (later) to rewrite the `file_transfer service` using `axum`.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -89,29 +89,19 @@ impl C8yEndPoint {
         url_event_binary
     }
 
-    pub fn url_is_in_my_tenant_domain(&self, url: &str) -> bool {
+    pub fn possible_tenant_url(&self, url: &str) -> Option<Url> {
```
didier-wenzek, line 92: The naming convention we have on thin-edge is to use the prefix `maybe` for optional values.

```suggestion
    pub fn maybe_tenant_url(&self, url: &str) -> Option<Url> {
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -474,10 +474,12 @@ impl CumulocityConverter {
...
                 modules.modules.iter_mut().for_each(|module| {
                     if let Some(url) = &mut module.url {
-                        *url = if self.c8y_endpoint.url_is_in_my_tenant_domain(url.url()) {
+                        *url = if let Some(cumulocity_url) =
+                            self.c8y_endpoint.possible_tenant_url(url.url())
+                        {
                             // TODO what do we do if this fails?
                             DownloadInfo::new(
-                                self.auth_proxy.proxy_url(url.url()).unwrap().as_ref(),
+                                self.auth_proxy.proxy_url(cumulocity_url).unwrap().as_ref(),
```
didier-wenzek, line 482: The `unwrap()` has to be removed

crates/extensions/c8y_firmware_manager/src/actor.rs
```diff
@@ -248,7 +248,8 @@ impl FirmwareManagerActor {
             let auth = if self
                 .config
                 .c8y_end_point
-                .url_is_in_my_tenant_domain(firmware_url)
+                .maybe_tenant_url(firmware_url)
+                .is_some()
             {
                 if let Ok(token) = self.message_box.jwt_retriever.await_response(()).await? {
                     Some(Auth::new_bearer(&token))
```
didier-wenzek, line 255: This is out of scope for this PR, but we need to consider using `c8y_auth_proxy` here instead of managing jwt tokens. This will be required when moving firmware support to a tedge plugin and no more a c8y plugin.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -428,7 +428,8 @@ impl C8YHttpProxyActor {
         // If the provided url is c8y, add auth
         if self
             .end_point
-            .url_is_in_my_tenant_domain(download_info.url())
+            .maybe_tenant_url(download_info.url())
+            .is_some()
         {
             let token = self.get_and_set_jwt_token().await?;
             download_info.auth = Some(Auth::new_bearer(token.as_str()));
```
didier-wenzek, line 435: This is out of scope for this PR, but we need to consider using `c8y_auth_proxy` here instead of managing jwt tokens.

@rina23q  and @Ruadhri17 You will have to consider using  `c8y_auth_proxy`  in your work on `tedge_configuration_plugin` (not in the plugin itself but for the urls forged by the c8y mapper for the plugin to download configuration files).

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -0,0 +1,307 @@
...
+    let (body, body_clone) = small_body.try_clone();
+    let send_request = |body, token| {
+        client
+            .request(method.to_owned(), destination.clone())
+            .headers(headers.clone())
+            .bearer_auth(&token)
+            .body(body)
+            .send()
+    };
+    let mut res = send_request(body, token.clone()).await.unwrap();
```
didier-wenzek, line 122: When can this unwrap go wrong?

There are many unwraps in this method and it's not clear which are safe and which are not.

didier-wenzek (commented):
I manually tested this PR. However, a system test is missing. The [current tests](https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/software_management/software.robot) are not testing software update where the package is downloaded from c8y.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -0,0 +1,307 @@
...
+struct TargetHost(Arc<str>);
+
+impl fmt::Display for TargetHost {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.0.fmt(f)
+    }
+}
+
+async fn respond_to(
+    State(TargetHost(host)): State<TargetHost>,
```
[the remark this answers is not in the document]
jarhodes314: In order for Axum to make state type-safe, there is a limitation that only one piece of state can be managed by the server, hence the use of an `AppState` struct. This is then [decomposed into "substates" using the `FromRef` trait](https://docs.rs/axum/latest/axum/extract/struct.State.html#substates)

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -0,0 +1,307 @@
...
+    method: Method,
+    headers: HeaderMap<HeaderValue>,
+    small_body: crate::body::PossiblySmallBody,
+) -> Response {
+    let path = match &path {
+        Some(Path(p)) => p.as_str(),
+        None => "",
+    };
+
+    if path.ends_with(".js") || path.starts_with("apps/") {
```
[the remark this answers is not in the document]
jarhodes314: Once the server is listening, it's easy to access `localhost:8001` in a browser, which will attempt to access these scripts, and something in these requests causes the JWT to be revoked (for reasons unknown to me), which means the following request will fail.

crates/extensions/c8y_auth_proxy/src/server.rs
jarhodes314: I should adjust this to be more explicit, but this should only fail in the case where the request fails entirely (e.g. connection refused, DNS errors, etc.), which will be turned into a 500 Internal Server Error response when the handler panics.

didier-wenzek: >  which will be turned into a 500 Internal Server Error response when the handler panics.

That's the point I was missing.

crates/extensions/c8y_auth_proxy/src/server.rs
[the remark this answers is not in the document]
didier-wenzek: Okay, a bit clearer. But, how `axum::routing::get(respond_to)` can know that the state has to be passed twice to this `respond_to` function? one reference for the first arg and a second reference for the second one?

Cargo.toml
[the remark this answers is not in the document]
albinsuresh: I know this that ship has already sailed, but still asking just for clarification. From my brief exploration of axum, the major USP of axum over hyper is its advanced routing capabilities, which seems to be the reason why we're using it as well. But, since our routing requirement is pretty simple with a very static rule such as `localhost:8080/c8y/*path` -> `c8y.http.url/*path`, did we really need axum for it? Or is there some other axum capability/feature that we're using that wouldn't have been possible or difficult to build with hyper directly?

albinsuresh (commented):
I already saw suggestions from @didier-wenzek to have follow-up PRs to migrate plugins like config plugin, firmware plugin etc to use this local proxy for any C8y interactions instead of doing them directly. As an extension of that, are we planning to change `c8y_http_proxy` also to use these local endpoints, as some of the token handling and all are replicated there as well.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -0,0 +1,307 @@
...
+
+    let body = if te_header.map_or(false, |h| {
+        h.to_str().unwrap_or_default().contains("chunked")
+    }) {
+        axum::body::boxed(StreamBody::new(res.bytes_stream()))
+    } else {
+        axum::body::boxed(Full::new(res.bytes().await.unwrap()))
+    };
+
+    build.body(body).unwrap()
```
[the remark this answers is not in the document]
jarhodes314: If the body is sufficiently large that we don't cache it in memory via `PossiblySmallBody`, then the body is simply preserved as a stream and just passed straight to `reqwest`. I haven't done any specific tests of this, but there's no reason why this shouldn't work. The issue currently would be that it would still succumb to expired JWTs. We deal with that by retrying a request transparently, but if we haven't stored the (potententially very large) body in memory, we cannot retry the request (but if the client were to retry the request to the proxy that would then work).

crates/extensions/c8y_auth_proxy/src/server.rs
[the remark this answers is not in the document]
jarhodes314: > Okay, a bit clearer. But, how `axum::routing::get(respond_to)` can know that the state has to be passed twice to this `respond_to` function? one reference for the first arg and a second reference for the second one?

(comment appeared to be placed in the wrong conversation, so quoting here)

It's all achieved through the magic of the [`Handler` trait](https://docs.rs/axum/latest/axum/handler/trait.Handler.html). Axum calls [`from_request_parts`](https://docs.rs/axum/latest/axum/extract/trait.FromRequestParts.html#tymethod.from_request_parts) on each argument, apart from the last one which uses [`FromRequest`](https://docs.rs/axum/latest/axum/extract/trait.FromRequest.html) instead (so you can consume the body). So as far as handlers are concerned, extracting managed state is no different from extracting the request method or headers.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -89,29 +89,19 @@ impl C8yEndPoint {
         url_event_binary
     }
 
-    pub fn url_is_in_my_tenant_domain(&self, url: &str) -> bool {
+    pub fn maybe_tenant_url(&self, url: &str) -> Option<Url> {
```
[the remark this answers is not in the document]
didier-wenzek: > So, that name with a bool return felt more natural.

The issue with functions, that check a shape and return a bool, is that if you want to go further you have to redo the parsing work a second time dealing with error cases which should no more happens.

If you only need a check this can be done by chaining `self.maybe_tenant_url(url).is_some()`.

crates/extensions/c8y_auth_proxy/src/actor.rs
```diff
@@ -0,0 +1,104 @@
...
+    app_state: AppState,
+    bind_address: IpAddr,
+    bind_port: u16,
+    signal_sender: mpsc::Sender<RuntimeRequest>,
+    signal_receiver: mpsc::Receiver<RuntimeRequest>,
+}
+
+impl C8yAuthProxyBuilder {
+    pub fn try_from_config(
+        config: &TEdgeConfig,
```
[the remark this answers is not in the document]
jarhodes314: I'm not sure when it would be valuable to create this actor in unit tests as this is essentially just a wrapper to listen for shutdown signals from the mapper. The unit tests that currently exist spin up the server directly, which is simpler (particularly for trying ports in sequence), and could even be made to not listen on a port at all (but they already rely on mockito to simulate Cumulocity's responses), so I didn't bother working out how to do that.

tests/RobotFramework/tests/cumulocity/software_management/dummy-plugin-2.sh
```diff
@@ -0,0 +1,62 @@
...
+      ;;
+    *)
+      echo "Unknown argument $1"
+      exit 1
+      ;;
+  esac
+done
+
+if [[ $(cat "$FILEPATH") != "Testing a thing" ]]; then
+    echo -e "Downloaded file (for package $NAME) does not have expected SHA256. Contents are:\n\n$(cat "$FILEPATH")"
```
didier-wenzek, line 56: I don't understand this error message

tests/RobotFramework/tests/cumulocity/software_management/sm-plugin.robot
```diff
@@ -47,6 +47,14 @@ sm-plugins should work without sudo installed and running as root
...
     Length Should Be    ${software}    2
 
+sm-plugins download files from Cumulocity
+    Execute Command    sudo tedge config set software.plugin.max_packages 1
+    Connect Mapper    c8y
+    Device Should Exist    ${DEVICE_SN}
+    ${file_url}=    Cumulocity.Create Inventory Binary    sm-plugin-test-file    software1    contents=Testing a thing
+    ${OPERATION}=    Install Software    dummy-software,1.0.0::dummy3,${file_url}
+    ${OPERATION}=    Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
+    File Should Exist    /tmp/dummy3/intalled_dummy-software
```
didier-wenzek, line 57: The test is correct but not really clear: the downloaded content is checked by the dummy plugin and not the test itself.

I suggest to add:

```
${downloaded}=    Execute Command    /tmp/dummy3/intalled_dummy-software
Should Be Equal    ${downloaded}    Testing a thing
```

tests/RobotFramework/tests/cumulocity/software_management/dummy-plugin-2.sh
```diff
@@ -0,0 +1,62 @@
...
+    exit 1
+else
+    mkdir -p "/tmp/$PLUGIN_NAME"
+    touch "/tmp/$PLUGIN_NAME/intalled_$NAME"
```
didier-wenzek, line 60:
```suggestion
    touch "/tmp/$PLUGIN_NAME/installed_$NAME"
```

didier-wenzek (approved):
Happy to have this merged with as small test improvement.

tests/RobotFramework/tests/cumulocity/software_management/dummy-plugin-2.sh
jarhodes314: Sorry, that's left over from when I'd made this even more complicated, although I think it wants removing entirely anyway given your other comment

didier-wenzek (approved):
Approving the updated test

Cargo.toml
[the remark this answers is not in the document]
jarhodes314: The changes I've since made in https://github.com/thin-edge/thin-edge.io/pull/2284/files do a better job of demonstrating the value of axum. Mainly this boils down to safety and the ability to write code that focuses on the task at hand, rather than having a bunch of boilerplate to convert incoming requests to responses, and handle errors at the other end. In the case of a proxy, it doesn't make a huge amount of difference, but as I demonstrated in this PR (i.e. stuff fixed in #2284), the APIs can have spurious errors that want `expect`ing/ at the best of times

gligorisaev:

PR #2266, opened 2023-09-15
C8y mapping of entity registration messages

albinsuresh:
## Proposed changes

Cumulocity mapping for:
- [x] Child device registration
- [x] Service registration
- [x] Nested child device/service registration
- [x] System tests

Deferred for later:
- [ ] Inventory updates for custom fragments
- [ ] Auto-registration for services
- [ ] Avoid resending all entity creation messages to c8y on restart

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


reubenmiller:
@albinsuresh I've been trying out the new settings,

* `name` is used instead of the `displayName` property defined in the MQTT spec. I'm ok with using `name` over `displayName` however then we need to update the spec to the implementation
* The Cumulocity IoT external identities for both the child devices and services are missing the common name prefix (which reduces naming conflicts between child devices on different tedge instances).
    * **child-device:** `device:TST_jawbone_direct_wood` -> `<common_name>:device:TST_jawbone_direct_wood`
    * **service:** `device:TST_jawbone_direct_wood:service:custom-app` -> `<common_name>:device:TST_jawbone_direct_wood:service:custom-app`

albinsuresh:
> * `name` is used instead of the `displayName` property defined in the MQTT spec. I'm ok with using `name` over `displayName` however then we need to update the spec to the implementation

Stuck with `name` over `displayName` as thin-edge on its own doesn't really have a "display" aspect to it by default, and that word felt more appropriate for the UI aspects of the cloud. So, `name` sounds more "agnostics" to my ears ;) And it was more consistent with `type` as well. Updated the specs accordingly.

> * The Cumulocity IoT external identities for both the child devices and services are missing the common name prefix (which reduces naming conflicts between child devices on different tedge instances).
>   
>   * **child-device:** `device:TST_jawbone_direct_wood` -> `<common_name>:device:TST_jawbone_direct_wood`
>   * **service:** `device:TST_jawbone_direct_wood:service:custom-app` -> `<common_name>:device:TST_jawbone_direct_wood:service:custom-app`

Fixed by appending the the main device's `device_id` to the generated external IDs of entities.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -171,6 +173,8 @@ pub struct CumulocityConverter {
     pub c8y_endpoint: C8yEndPoint,
     pub mqtt_schema: MqttSchema,
     pub entity_store: EntityStore,
+    pub external_id_map: HashMap<EntityTopicId, String>,
+    pub reverse_external_id_map: HashMap<String, EntityTopicId>,
```
albinsuresh, line 177: @didier-wenzek and I have agreed to avoid the duplication of these maps here and reuse the mapping in the entity store itself for the same.

The key thing is that, the mapping of `EntityTopicId` to its C8y representation would be done at the converter level, as it is a cloud specific logic, but the mapped id will be used inside the entity store as the `id` of the entity. From the entity store's perspective, it's just one unique opaque ID.  That way, we have clear separation of concerns. So, when different kinds of mappers/components use the `EntityStore` data structure, they are free to do different kinds of `id` mappings and store those mapped ids in the same struct, without needing yet another external mapping.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -224,9 +232,87 @@ impl CumulocityConverter {
...
+        self.external_id_map
+            .insert(entity_topic_id.clone(), external_id.clone());
+        self.reverse_external_id_map
+            .insert(external_id.clone(), entity_topic_id.clone());
+
+        match input.r#type {
+            EntityType::MainDevice => Err(ConversionError::MainDeviceRegistrationNotSupported),
+            EntityType::ChildDevice => {
+                let ancestors = self.entity_store.ancestors(entity_topic_id)?;
+                let ancestors_external_ids = self.ancestors_external_ids(&ancestors);
```
albinsuresh, line 264: Will create a new helper function over this piece of code that'll return the target entity's C8y topic (e.g: `c8y/s/us/entity-1/entity-2/entity-3`) directly from the `EntityTopicId`.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -224,9 +232,87 @@ impl CumulocityConverter {
...
+            mapped_ancestors.push(
+                self.external_id_map
+                    .get(entity)
+                    .expect("External ID should be present")
+                    .to_string(),
+            )
+        }
+
+        mapped_ancestors
+    }
```
didier-wenzek, line 314: This can be replaced by an `ancestors.iter().map(...).collect()`.

didier-wenzek (changes requested):
The duplications with the `EntityStore` must be resolved.

Bravo555 (commented):
No issues with the code, just requesting some additional comments

crates/core/c8y_api/src/smartrest/topic.rs
```diff
@@ -136,10 +136,22 @@ impl TryFrom<Topic> for MapperSubscribeTopic {
     }
 }
 
+pub fn publish_topic_from_ancestors(ancestors: &[String]) -> Topic {
```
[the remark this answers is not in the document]
albinsuresh: Will do the cleanup of the `c8y_api::topic` module in a follow-up PR, as replacing something like `C8yTopic` would touch a lot of files.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -702,15 +779,39 @@ impl CumulocityConverter {
         match &channel {
             Channel::EntityMetadata => {
                 if let Ok(register_message) = EntityRegistrationMessage::try_from(message) {
-                    if let Err(e) = self.entity_store.update(register_message) {
+                    // Generate the c8y external id, if an external id is not provided in the message
+                    let external_id = register_message
+                        .entity_id
+                        .as_ref()
+                        .map(|v| v.to_string().into())
+                        .unwrap_or_else(|| self.map_to_c8y_external_id(&source));
```
didier-wenzek, line 787: It's fragile to let this logic to the mapper. The only piece of code that is related to c8y is the `map_to_c8y_external_id`  method. See my comment on the `entity_store.update()` method.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -721,7 +822,7 @@ impl CumulocityConverter {
                     let entity = self.entity_store.get(&source).unwrap();
                     if let Some(message) = external_device_registration_message(entity) {
                         self.children
-                            .insert(entity.entity_id.to_string(), Operations::default());
+                            .insert(entity.entity_id.clone().into(), Operations::default());
                         registration_messages.push(message);
```
didier-wenzek, line 826: This are local registration messages. Don't we also to call `self.try_convert_entity_registration()` to register this entity on c8y?

- Is the expectation to have this mapper processing the local registration messages sent by itself?
- If so telemetry data might be sent to c8y before the child device registration.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -103,9 +133,54 @@ impl EntityStore {
...
-    pub fn main_device_name(&self) -> &str {
-        self.get(&self.main_device).unwrap().entity_id.as_str()
+    /// Returns the external id of the main device.
+    pub fn main_device_external_id(&self) -> EntityExternalId {
+        self.get(&self.main_device).unwrap().entity_id.clone()
+    }
+
+    /// Returns an ordered list of ancestors of the given entity
+    /// starting from the immediate parent all the way till the root main device.
+    /// The last parent in the list for any entity would always be the main device.
```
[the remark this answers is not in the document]
albinsuresh: It felt more natural, as even in the physical world when you think about your ancestors, you start with your immediate parents, right? That's why I kept it this way.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -144,6 +219,7 @@ impl EntityStore {
     pub fn update(
         &mut self,
         message: EntityRegistrationMessage,
+        external_id: EntityExternalId,
     ) -> Result<Vec<EntityTopicId>, Error> {
```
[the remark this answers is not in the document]
albinsuresh: Yeah, I was also initially thinking of accepting a closure to compute that external id when needed. But, since the external id computation relies on the state of the converter (the main device id), I thought it wouldn't work. That's why I decided to pass the computed value this way. And yes, it has the downside of unnecessarily computing the external id when an `@id` value is already present in the message. But, because of that closure limitation, I was forced to do it this way.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -168,13 +247,10 @@ impl EntityStore {
             affected_entities.push(parent.clone());
         }
 
-        let entity_id = message
-            .entity_id
-            .unwrap_or_else(|| self.derive_entity_id(&message.topic_id));
```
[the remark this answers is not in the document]
albinsuresh: It was working because of the way the `external_id` was computed and passed into the entity store on an `update`. That compute logic in the converter was using that `@id` as-is when one is provided. But, this is going to to be restored once the computation logic is passed in as a closure.

crates/core/tedge_api/src/entity_store.rs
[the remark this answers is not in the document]
rina23q: I feel it natural, starting with the device in focus, then the last is the main device.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -233,29 +285,35 @@ impl EntityStore {
     /// being registered.
     pub fn auto_register_entity(
```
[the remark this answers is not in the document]
albinsuresh: I've made an attempt to simplify it.

tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
```diff
@@ -29,13 +29,117 @@ Child device registration
...
+    # Check service registration
+    Check Service    child_sn=${DEVICE_SN}:device:${CHILD_SN}    service_sn=${DEVICE_SN}:device:${CHILD_SN}:service:custom-app    service_name=custom-app    service_type=custom-type    service_status=up
+
+
+Register devices using custom MQTT schema
+    [Documentation]    Complex example showing how to use custom MQTT topics to register devices/services using
+        ...            custom identity schemas
+
+    # Main device registration via MQTT is not supported at the moment
+    # Execute Command    tedge mqtt pub --retain 'te/base///' '{"@type":"main-device","name":"base","type":"te_gateway"}'
```
albinsuresh, line 83: Even though the registration of a new main device can't be supported atm, I'm assuming that the expected behaviour here was to just add this custom topic id also into the entity store, associated to the main device's external id itself so that data to the main device can be sent via this custom topic as well. That much can be done. The `@id`, `name` and `type` fields will be ignored until we support device fragment updates as well.

crates/core/tedge_api/src/entity_store.rs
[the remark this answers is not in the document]
didier-wenzek: Okay then - maybe just me :-)

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -168,13 +258,13 @@ impl EntityStore {
             affected_entities.push(parent.clone());
         }
 
-        let entity_id = message
-            .entity_id
-            .unwrap_or_else(|| self.derive_entity_id(&message.topic_id));
+        let external_id = message.entity_id.unwrap_or_else(|| {
```
albinsuresh, line 261: There was a suggestion from @reubenmiller to apply the same transformation rules(replace `/` with `:`) to the user provided `@id` value as well, rather than just using it as-is, which could mess up the external ID in SmartREST topics. I was initially thinking of just treating it as bad input if it contains "special forbidden chars" like `/, +, #`. But, this proposed scheme might be a bit more flexible, with the caveat of us overrriding even user overridden value. So,  @didier-wenzek, thoughts?

crates/core/tedge_api/src/entity_store.rs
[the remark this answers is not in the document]
didier-wenzek: Indeed simpler and clearer.

I think that we can go a step further by returning a vector of `RegistrationMessage` instead of `Mqtt::Message` (the latter can be easily built from the former by a mapper). => I will push a commit for that.

didier-wenzek (commented):
I'm happy to merge this PR once the conflicts fixed.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -920,7 +1106,7 @@ async fn mapper_dynamically_updates_supported_operations_for_child_device() {
     let cfg_dir = TempTedgeDir::new();
     create_thin_edge_child_operations(
         &cfg_dir,
-        "child1",
+        "test-device:device:child1",
```
didier-wenzek, line 1109: Having to fix this test (`mapper_dynamically_updates_supported_operations_for_child_device`) raises questions it would be good to address before merging.

The issue is that the new c8y external ids must be used only over the c8y topics but also for the c8y operation directories. The current behavior might be confusing: the same device being auto-registered differently after an operation file vs on reception of measurement.

didier-wenzek (changes requested):
One needs to check what are the impacts of using the new external ids on the c8y topics.

crates/extensions/c8y_mapper_ext/src/tests.rs
albinsuresh: Even though this directory mechanism won't be around for too long, as long as we support it, it should preserve its former behaviour IMO: where the external user just creates directories with the local device ID, and the mapper should generate the longer external ID with all `<MAIN_DEVICE_ID>:device` prefix. The user shouldn't be forced to create the directories with those prefixes as that is an internal mapper logic.  The child directory name should be treated exactly like the `device_id` in `te/device/<device_id>//` topic.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: The `external_device_registration_message` should have been replaced with `try_convert_entity_registration` message as the `external_device_registration_message` is redundant and incomplete(doesn't cover services). The C8y device/service registration should happen from this scope itself, and we can't wait till the mapper processes the auto-generated registration messages on the next attempt, to avoid the telemetry message going to the cloud before its device's registration message.

Although harmless in its current form, we still need to do something to avoid reprocessing those auto-generated messages. But, as mentioned in the description, I'd like to defer that for a follow-up PR, where can handle this case, and the duplicate processing of all the retained messages(including alarms) on a mapper restart with a single mechanism, if possible. I was thinking of persisting some of the reg information on the file system to tackle this. But, that is to be discussed and I'd like to keep that out of the scope of this PR.

crates/core/tedge_api/src/entity_store.rs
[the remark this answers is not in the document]
albinsuresh: I'll do that, as it'd allow the reuse of `try_convert_entity_registration` fn for both explicit registration as well as auto-registration.

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: > The C8y device/service registration should happen from this scope itself, and we can't wait till the mapper processes the auto-generated registration messages on the next attempt, to avoid the telemetry message going to the cloud before its device's registration message.

Correct. So do you plan to update this PR accordingly?

tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
didier-wenzek: > Even though the registration of a new main device can't be supported atm

I think the issue is to create the EntityStore from a main-device registration message: this message can only be a guess of the topic id used by the main device.

The EntityStore should be created only after a main-device CN. It's registration can then be done either explicitly with a registration message or automatically if some MEA is published on `device/main///#`.

crates/extensions/c8y_mapper_ext/src/tests.rs
didier-wenzek: > the external user just creates directories with the local device ID, and the mapper should generate thes  longer external ID with all <MAIN_DEVICE_ID>:device prefix. The user shouldn't be forced to create the directories with those prefixes as that is an internal mapper logic. The child directory name should be treated exactly like the device_id in te/device/<device_id>// topic.

Fully agree. And this needs to be done along this PR.

I think the `children` cache used by the mapper must also be updated to use external-ids for the keys.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -327,10 +329,10 @@ impl EntityStore {
                 let topic = Topic::new(&format!("{MQTT_ROOT}/{parent_device_id}")).unwrap();
                 let device_register_message =
                     Message::new(&topic, device_register_payload).with_retain();
+                let device_register_message =
+                    EntityRegistrationMessage::try_from(&device_register_message).unwrap();
```
didier-wenzek, line 333: I find weird the way this `EntityRegistrationMessage` is built, i.e. by parsing a message payload.

Why not directly building an `EntityRegistrationMessage`? This would avoid to:
- `unwrap()`
- Use an hard-coded `MQTT_ROOT`

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -501,6 +503,39 @@ impl TryFrom<&Message> for EntityRegistrationMessage {
...
+        if let Some(parent_id) = &value.parent {
+            register_payload.insert("@parent".into(), Value::String(parent_id.to_string()));
+        }
+
+        if let Value::Object(other_keys) = value.payload.clone() {
+            register_payload.extend(other_keys)
+        }
+
+        Message::new(
+            &Topic::new(&format!("{MQTT_ROOT}/{entity_topic_id}")).unwrap(),
```
didier-wenzek, line 532: The `MQTT_ROOT` should not be hard-coded.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -127,11 +128,17 @@ impl C8yMapperActor {
             FsWatchEvent::DirectoryCreated(path) => {
                 if let Some(directory_name) = path.file_name() {
                     let child_id = directory_name.to_string_lossy().to_string();
-                    let message = Message::new(
-                        &Topic::new_unchecked(SMARTREST_PUBLISH_TOPIC),
-                        format!("101,{child_id},{child_id},thin-edge.io-child"),
-                    );
-                    self.mqtt_publisher.send(message).await?;
+                    let child_topic_id = EntityTopicId::default_child_device(&child_id).unwrap();
```
didier-wenzek, line 131: This is the drawback of using *short* child device names in path: one has to assume a default topic scheme.

Ideally, I would use the same hierarchy for paths as for MQTT. i.e. with a `device` directory and there a sub directory per child plus a `main` directory. Or an `a/b/c/d` tree if the user prefer its own schema.

didier-wenzek: > we could do that in a follow-up PR

Sure.

crates/core/c8y_api/src/smartrest/topic.rs
```diff
@@ -96,7 +96,7 @@ impl From<&EntityMetadata> for C8yTopic {
     fn from(value: &EntityMetadata) -> Self {
         match value.r#type {
             EntityType::MainDevice => Self::SmartRestResponse,
-            EntityType::ChildDevice => Self::ChildSmartRestResponse(value.entity_id.clone()),
+            EntityType::ChildDevice => Self::ChildSmartRestResponse(value.entity_id.clone().into()),
             EntityType::Service => Self::SmartRestResponse, // TODO how services are handled by c8y?
```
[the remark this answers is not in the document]
albinsuresh: Got rid of that function.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Moved the this function to the converter where the schema is available.

tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
albinsuresh: Enabled overriding of the default schema using another registration message with custom topic scheme.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Done.

albinsuresh:
Follow-up tasks:
- [ ] Sanitise `external_id` provided via `@id` value in the registration message
- [ ] Review file system based child device registration API. Should the directory name be used as-is as the child device external id? How to support addition of more operations when someone is already using the default topic scheme and hence the default external_id generation logic.
- [ ] Avoid re-processing of auto-registered entity registration messages
- [ ] Avoid re-processing of all the retained entity registration messages on mapper restart. Use file system persistence to identify the delta? Or re-use the message sync mechanism of alarms (too complex)?
- [ ] Simplify `smartrest::topic` module by keeping `publish_topic_from_ancestors` and removing redundant helpers

didier-wenzek (approved):
Approved. Thank you!

Issue #2243, opened 2023-09-11
Alarms for c8y  are consumed from `te/+/+/+/+/a/+`


rina23q:
Cumulocity mapper should consume `te/+/+/+/+/a/+` topic instead of `tedge/alarms/+` and `tedge/alarms/+/+`.

The details is described in https://github.com/thin-edge/thin-edge.io/issues/2082.

rina23q:
Done with #2267

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite. 
- tests/RobotFramework/tests/cumulocity/telemetry/c8y_child_alarms_rpi.robot
- tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
- tests/RobotFramework/tests/cumulocity/telemetry/raise_alarms.robot
- tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot
- tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot
- [x] QA has tested the feature and it meets the required specifications.

Issue #2242, opened 2023-09-11
C8Y Mapper converts alarms/events to only JSON over MQTT


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
Cumulocity Mapper converts alarms and events to either SmartREST or JSON over MQTT up to the payload patterns.
My proposal is to stop converting to SmartREST in any case, and converts to only JSON over MQTT.

Why?
The cost of supporting SmartREST with correct validation is quite high compared to the efforts.
We can simplify our current c8y-mapper code and stronger and easier validation for payload by using external JSON crate.
SmartREST is **CSV-like** format, which causes a problem how to escape special characters. Also, since it's not completely CSV, we cannot use external CSV crate.

For example, if user publishes this payload to the event topic:
```json
{
   "type":"foo,bar",
   "text":"Someone clicked",
}
```
The event type `foo,bar` is valid, however, we need an escape for `,` in SmartREST, whilst we don't need any escape for JSON over MQTT.

Another example, if given type contains `"`:
```json
{
   "type":"foo\"bar",
   "text":"Someone clicked",
}
```
The event type `foo"bar` is valid, however, what do we need to escape `"` for SmartREST? Two double-quotes or backslash? If JSON over MQTT, this escape is surely backslash as it is given by input JSON.

**Describe the solution you'd like**

**Describe alternatives you've considered**
Nothing

**Additional context**
https://github.com/thin-edge/thin-edge.io/pull/2221#discussion_r1317499361

PR #2267, opened 2023-09-17
c8y-mapper consume the "te" new alarm mqtt scheme

rina23q:
## Proposed changes
Implement #2243.

Behaviour changes:
- Defined the default values for mandatory fields for Cumulocity if it's missing or wrongly given.
a. type: `ThinEdgeAlarm`
b. text: The alarm type (the same as previous behaviour)
c. severity: `minor`
- The definition of uniqueness of alarm. Now it's per alarm type, although it was the combination of alarm type and severity previously. So, alarm A `"type": "foo", "severity": "major"` and alarm B `"type": "foo", "severity": "critical"` are no longer considered as different alarms.

## Types of changes

## Paste Link to the issue
#2243 

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/json_c8y.rs
+++ b/crates/core/c8y_api/src/json_c8y.rs
@@ -1,21 +1,25 @@
+use download::DownloadInfo;
+use serde::Deserialize;
+use serde::Serialize;
+use serde_json::Value;
 use std::collections::HashMap;
-
+use std::fmt;
 use tedge_api::alarm::ThinEdgeAlarm;
+use tedge_api::alarm::ThinEdgeAlarmData;
+use tedge_api::entity_store::EntityMetadata;
+use tedge_api::entity_store::EntityType;
+use tedge_api::event::ThinEdgeEvent;
+use tedge_api::EntityStore;
 use tedge_api::Jsonify;
 use tedge_api::SoftwareListResponse;
 use tedge_api::SoftwareModule;
 use tedge_api::SoftwareType;
 use tedge_api::SoftwareVersion;
-
-use crate::smartrest::error::SMCumulocityMapperError;
-use download::DownloadInfo;
-use serde::Deserialize;
-use serde::Serialize;
-use serde_json::Value;
-use tedge_api::event::ThinEdgeEvent;
 use time::OffsetDateTime;
 
 const EMPTY_STRING: &str = "";
+const DEFAULT_ALARM_SEVERITY: AlarmSeverity = AlarmSeverity::Minor;
+const DEFAULT_ALARM_TYPE: &str = "ThinEdgeAlarm";
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
 #[serde(rename_all = "camelCase")]
@@ -218,8 +222,8 @@ fn update_the_external_source_event(extras: &mut HashMap<String, Value>, source:
     extras.insert("externalSource".into(), value.into());
 }
 
-fn make_c8y_source_fragment(source_name: &str) -> Option<SourceInfo> {
-    Some(SourceInfo::new(source_name.into(), "c8y_Serial".into()))
+fn make_c8y_source_fragment(source_name: &str) -> SourceInfo {
+    SourceInfo::new(source_name.into(), "c8y_Serial".into())
 }
 
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
@@ -236,96 +240,184 @@ impl SourceInfo {
         Self { id, source_type }
     }
 }
+
+/// Internal representation of c8y's alarm model.
+#[derive(Debug, PartialEq, Eq)]
+pub enum C8yAlarm {
+    Create(C8yCreateAlarm),
+    Clear(C8yClearAlarm),
+}
+
+/// Internal representation of creating an alarm in c8y.
+/// Note: text and time are optional for SmartREST, however,
+/// mandatory for JSON over MQTT. Hence, here they are mandatory.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
 pub struct C8yCreateAlarm {
+    /// Alarm type, default is "ThinEdgeAlarm".
+    #[serde(rename = "type")]
+    pub alarm_type: String,
+
+    /// None for main device, Some for child device.
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "externalSource")]
     pub source: Option<SourceInfo>,
 
-    pub severity: String,
+    pub severity: AlarmSeverity,
 
-    #[serde(rename = "type")]
-    pub alarm_type: String,
+    pub text: String,
 
     #[serde(with = "time::serde::rfc3339")]
     pub time: OffsetDateTime,
-
-    pub text: String,
 
     #[serde(flatten)]
     pub fragments: HashMap<String, Value>,
 }
 
-impl C8yCreateAlarm {
-    pub fn new(
-        source: Option<SourceInfo>,
-        severity: String,
-        alarm_type: String,
-        time: OffsetDateTime,
-        text: String,
-        fragments: HashMap<String, Value>,
-    ) -> Self {
-        Self {
-            source,
-            severity,
-            alarm_type,
-            time,
-            text,
-            fragments,
+/// Internal representation of clearing an alarm in c8y.
+#[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct C8yClearAlarm {
+    /// Alarm type, default is "ThinEdgeAlarm".
+    #[serde(rename = "type")]
+    pub alarm_type: String,
+
+    /// None for main device, Some for child device.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "externalSource")]
+    pub source: Option<SourceInfo>,
+}
+
+impl C8yAlarm {
+    pub fn try_from(
+        alarm: &ThinEdgeAlarm,
+        entity_store: &EntityStore,
+    ) -> Result<Self, C8yAlarmError> {
+        if let Some(entity) = entity_store.get(&alarm.source) {
+            let source = Self::convert_source(entity);
+            let alarm_type = Self::convert_alarm_type(&alarm.alarm_type);
+
+            let c8y_alarm = match alarm.data.as_ref() {
+                None => C8yAlarm::Clear(C8yClearAlarm { alarm_type, source }),
+                Some(tedge_alarm_data) => C8yAlarm::Create(C8yCreateAlarm {
+                    alarm_type: alarm_type.clone(),
+                    source,
+                    severity: C8yCreateAlarm::convert_severity(tedge_alarm_data),
+                    text: C8yCreateAlarm::convert_text(tedge_alarm_data, &alarm_type),
+                    time: C8yCreateAlarm::convert_time(tedge_alarm_data),
+                    fragments: C8yCreateAlarm::convert_extras(tedge_alarm_data),
+                }),
+            };
+            Ok(c8y_alarm)
+        } else {
+            Err(C8yAlarmError::UnsupportedDeviceTopicId(
+                alarm.source.to_string(),
+            ))
+        }
+    }
+
+    fn convert_source(entity: &EntityMetadata) -> Option<SourceInfo> {
+        match entity.r#type {
+            EntityType::MainDevice => None,
+            EntityType::ChildDevice => Some(make_c8y_source_fragment(&entity.entity_id.clone())),
+            EntityType::Service => Some(make_c8y_source_fragment(&entity.entity_id.clone())),
+        }
+    }
+
+    fn convert_alarm_type(alarm_type: &str) -> String {
+        if alarm_type.is_empty() {
+            DEFAULT_ALARM_TYPE.to_string()
+        } else {
+            alarm_type.to_string()
         }
     }
 }
 
-impl TryFrom<&ThinEdgeAlarm> for C8yCreateAlarm {
-    type Error = SMCumulocityMapperError;
-
-    fn try_from(alarm: &ThinEdgeAlarm) -> Result<Self, SMCumulocityMapperError> {
-        let severity = alarm.severity.to_string();
-        let alarm_type = alarm.name.to_owned();
-        let text;
-        let time;
-        let fragments;
-
-        match &alarm.to_owned().data {
-            None => {
-                text = alarm_type.clone();
-                time = OffsetDateTime::now_utc();
-                fragments = HashMap::new();
-            }
-            Some(data) => {
-                text = data.text.clone().unwrap_or_else(|| alarm_type.clone());
-                time = data.time.unwrap_or_else(OffsetDateTime::now_utc);
-                fragments = data.alarm_data.clone();
-            }
+impl C8yCreateAlarm {
+    fn convert_severity(alarm_data: &ThinEdgeAlarmData) -> AlarmSeverity {
+        match alarm_data.severity.clone() {
+            Some(severity) => match AlarmSeverity::try_from(severity.as_str()) {
+                Ok(c8y_severity) => c8y_severity,
+                Err(_) => DEFAULT_ALARM_SEVERITY,
+            },
+            None => DEFAULT_ALARM_SEVERITY,
         }
-
-        let source = if let Some(external_source) = &alarm.source {
-            make_c8y_source_fragment(external_source)
-        } else {
-            None
-        };
-
-        Ok(Self {
-            source,
-            severity,
-            alarm_type,
-            time,
-            text,
-            fragments,
-        })
     }
+
+    fn convert_text(alarm_data: &ThinEdgeAlarmData, alarm_type: &str) -> String {
+        alarm_data.text.clone().unwrap_or(alarm_type.to_string())
+    }
+
+    fn convert_time(alarm_data: &ThinEdgeAlarmData) -> OffsetDateTime {
+        alarm_data.time.unwrap_or_else(OffsetDateTime::now_utc)
+    }
+
+    /// Remove reserved keywords from extras.
+    /// "type", "time", "text", "severity" are ensured that they are not
+    /// in the hashmap of ThinEdgeAlarm because they are already members of the struct itself.
+    fn convert_extras(alarm_data: &ThinEdgeAlarmData) -> HashMap<String, Value> {
+        let mut map = alarm_data.extras.clone();
+        map.remove("externalSource");
+        map
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[serde(rename_all(serialize = "UPPERCASE"))]
+pub enum AlarmSeverity {
+    Critical,
+    Major,
+    Minor,
+    Warning,
+}
+
+impl TryFrom<&str> for AlarmSeverity {
+    type Error = C8yAlarmError;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        match value {
+            "critical" => Ok(AlarmSeverity::Critical),
+            "major" => Ok(AlarmSeverity::Major),
+            "minor" => Ok(AlarmSeverity::Minor),
+            "warning" => Ok(AlarmSeverity::Warning),
+            invalid => Err(C8yAlarmError::UnsupportedAlarmSeverity(invalid.into())),
+        }
+    }
+}
+
+impl fmt::Display for AlarmSeverity {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            AlarmSeverity::Critical => write!(f, "CRITICAL"),
+            AlarmSeverity::Major => write!(f, "MAJOR"),
+            AlarmSeverity::Minor => write!(f, "MINOR"),
+            AlarmSeverity::Warning => write!(f, "WARNING"),
+        }
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum C8yAlarmError {
+    #[error("Unsupported alarm severity in topic: {0}")]
+    UnsupportedAlarmSeverity(String),
+
+    #[error("Unsupported device topic ID in topic: {0}")]
+    UnsupportedDeviceTopicId(String),
 }
 
 #[cfg(test)]
 mod tests {
+    use crate::json_c8y::AlarmSeverity;
     use anyhow::Result;
     use assert_matches::assert_matches;
+    use mqtt_channel::Message;
+    use mqtt_channel::Topic;
     use serde_json::json;
-    use tedge_api::alarm::AlarmSeverity;
     use tedge_api::alarm::ThinEdgeAlarm;
     use tedge_api::alarm::ThinEdgeAlarmData;
+    use tedge_api::entity_store::EntityRegistrationMessage;
     use tedge_api::event::ThinEdgeEventData;
+    use tedge_api::mqtt_topics::EntityTopicId;
     use test_case::test_case;
     use time::macros::datetime;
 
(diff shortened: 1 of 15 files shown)


crates/core/tedge_api/src/alarm.rs
```diff
@@ -1,58 +1,84 @@
...
+    pub alarm_type: Option<String>,
 
-#[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
-pub enum AlarmSeverity {
-    Critical,
-    Major,
-    Minor,
-    Warning,
+    /// Alarm source retrieved from the MQTT topic.
+    pub source: EntityMetadata,
```
didier-wenzek, line 17: I think it would be more appropriate to use an [`EntityTopicId`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/mqtt_topics.rs#L213) instead of an [`EntityMetadata`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/entity_store.rs#L288).

- An `EntityTopicId` is just an identifier such as `device/xyz//`, while an `EntityMetadata` contains all the data attached to that entity.
- Using an `EntityMetadata` to identify the source of an alarm makes things a bit more involved (one needs an [`EntityStore`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/entity_store.rs#L58) to decode the `EntityTopicId` extracted from the alarm topic.
- One can not represent an alarm sent by a device that is not registered yet (as it is then unknown from the `EntityStore`).

crates/core/tedge_api/src/alarm.rs
```diff
@@ -1,58 +1,84 @@
...
+            data: None,
+        }
+    }
 
-    #[error("Unsupported alarm severity in topic: {0}")]
-    UnsupportedAlarmSeverity(String),
+    /// Override "type" when it is provided in payload.
+    pub fn with_alarm_data(self, mut data: ThinEdgeAlarmData) -> Self {
+        match data.extras.remove("type") {
+            Some(maybe_type_from_payload) => match maybe_type_from_payload.as_str() {
```
didier-wenzek, line 52: Why do we need this second match ` maybe_type_from_payload.as_str()` ?

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -236,95 +238,167 @@ impl SourceInfo {
...
-        Ok(Self {
-            source,
-            severity,
-            alarm_type,
-            time,
-            text,
-            fragments,
-        })
+    // TODO: Ask which one is better, using default or returning an error.
+    fn convert_severity(alarm_data: &ThinEdgeAlarmData) -> AlarmSeverity {
```
didier-wenzek, line 325: In that case, using a default value is appropriate and convenient for the users.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
```diff
@@ -33,62 +34,67 @@ impl AlarmConverter {
...
-                    output_messages.push(Message::new(&c8y_alarm_topic, cumulocity_alarm_json));
+                match c8y_alarm.data {
+                    Some(ref alarm_data) if !alarm_data.fragments.is_empty() => {
+                        // JSON over MQTT
+                        let cumulocity_alarm_json = serde_json::to_string(&c8y_alarm)?;
+                        let c8y_alarm_topic = Topic::new_unchecked(C8Y_JSON_MQTT_ALARMS_TOPIC);
+                        output_messages.push(Message::new(&c8y_alarm_topic, cumulocity_alarm_json));
+                    }
+                    _ => {
+                        // SmartREST
```
didier-wenzek, line 87: Why not sending all the alarms using C8Y JSON?

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
```diff
@@ -33,62 +34,67 @@ impl AlarmConverter {
...
+                    let tedge_alarm_data =
+                        ThinEdgeAlarmData::try_from(mqtt_payload).map_err(|e| {
+                            ThinEdgeAlarmDeserializerError::FailedToParseJsonPayload {
+                                topic: mqtt_topic,
+                                error: e.to_string(),
+                                payload: mqtt_payload.chars().take(50).collect(),
+                            }
+                        })?;
+                    ThinEdgeAlarm::new(alarm_type, source).with_alarm_data(tedge_alarm_data)
+                };
```
didier-wenzek, line 74: This logic should be in `tedge_api::alarms` as this is unrelated to Cumulocity.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -1,58 +1,84 @@
...
 pub struct ThinEdgeAlarm {
-    pub name: String,
-    pub severity: AlarmSeverity,
-    #[serde(flatten)]
-    pub data: Option<ThinEdgeAlarmData>,
-    pub source: Option<String>,
-}
+    /// Alarm type retrieved from the MQTT topic or payload.
+    /// The "type" given from the payload has priority to the one from the topic.
+    pub alarm_type: Option<String>,
```
didier-wenzek, line 14: Why not a `String` as the empty string is a valid alarm type (for thin-edge).

crates/core/tedge_api/src/alarm.rs
```diff
@@ -1,58 +1,84 @@
...
-
 use serde_json::Value;
 use std::collections::HashMap;
-/// In-memory representation of ThinEdge JSON alarm.
-#[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
+use std::convert::TryFrom;
+
+/// Internal representation of ThinEdge alarm model.
+#[derive(Debug, Eq, PartialEq)]
 pub struct ThinEdgeAlarm {
```
didier-wenzek, line 11: I really like this decomposition of an alarm in two parts: one extracted from the topic, the other from the payload.

- My initial idea was to have a single structure, but I agree that this leads to issues when parsing an MQTT message.
- Using an optional payload makes clear and less error prone the case of the empty messages used to clear alarms.

That said it would be good to find a generic approach here. The same pattern `(source, type, payload)` is used for alarms, events, measurements, commands ...

didier-wenzek (commented):
I have some questions

crates/core/tedge_api/src/alarm.rs
rina23q: The hashmap `data.extras` may have `type: true` (boolean), not the desirable type. In this case, `as_str()` returns `Err`. If it's `Err`, I want to use the default type `ThinEdgeAlarm`.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
rina23q: Agree! Still need to check how to clear an alarm by JSON over MQTT though. But I would address this later with https://github.com/thin-edge/thin-edge.io/issues/2242.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
rina23q: You are right. The logic should move to the `ThinEdgeAlarm` side.

crates/core/tedge_api/src/alarm.rs
rina23q: Actually it should be `String`. I was overthinking about the case like `te/device/main///a/` (empty type), but this must be `"".to_string()`, not `None`. In short, I'll fix it.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
```diff
@@ -33,62 +34,67 @@ impl AlarmConverter {
...
-                            payload: mqtt_payload.chars().take(50).collect(),
-                        }
-                    })?;
-                let c8y_alarm = C8yCreateAlarm::try_from(&tedge_alarm)?;
+                let tedge_alarm = if mqtt_payload.is_empty() {
+                    // Clearing an alarm
+                    ThinEdgeAlarm::new(alarm_type, source)
+                } else {
+                    // Creating an alarm
+                    let tedge_alarm_data =
```
albinsuresh, line 65: I'd just restore that old `ThinEdgeAlarm::try_from` function or something like `impl TryFrom<&Message> for ThinEdgeAlarm`

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
albinsuresh: Just legacy, I guess. But doing that would be much better, in terms of code complexity, as then we just need to create one `C8yCreateAlarm` structure and use that both for MQTT as well as HTTP.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -318,33 +318,19 @@ impl CumulocityConverter {
...
+            let mqtt_messages = self
+                .alarm_converter
+                .try_convert_alarm(entity, input, alarm_type)?;
             Ok(mqtt_messages)
-        } else if topic.name.starts_with(INTERNAL_ALARMS_TOPIC) {
-            self.alarm_converter.process_internal_alarm(message);
-            Ok(vec![])
         } else {
-            Err(ConversionError::UnsupportedTopic(topic.name.clone()))
+            Ok(vec![])
```
albinsuresh, line 333: I'd return an error about unknown entity so that this error is properly reported on `tedge/errors`.

crates/core/tedge_api/src/alarm.rs
```diff
@@ -1,58 +1,84 @@
...
+    /// All other alarm data retrieved from the MQTT payload.
+    /// None means the message is meant to clear an alarm.
+    pub data: Option<ThinEdgeAlarmData>,
 }
 
-/// In-memory representation of ThinEdge JSON alarm payload
+/// Internal representation of the JSON MQTT payload.
 #[derive(Debug, Serialize, Deserialize, Eq, PartialEq)]
 pub struct ThinEdgeAlarmData {
+    pub severity: Option<String>,
```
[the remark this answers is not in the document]
rina23q: It's thin-edge level. I thought the new scheme would allow full freedom for any fields. Rather, why do we need to have limited `severity` as Cumulocity?

tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot
```diff
@@ -69,12 +69,50 @@ Thin-edge devices support sending custom events without type in topic
...
 Thin-edge devices support sending custom alarms #1699
     [Tags]    \#1699
-    Execute Command    tedge mqtt pub tedge/alarms/critical/myCustomAlarmType '{ "text": "Some test alarm", "someOtherCustomFragment": {"nested":{"value": "extra info"}} }'
+    Execute Command    tedge mqtt pub te/device/main///a/myCustomAlarmType '{ "severity": "critical", "text": "Some test alarm", "someOtherCustomFragment": {"nested":{"value": "extra info"}} }'
     ${alarms}=    Device Should Have Alarm/s    expected_text=Some test alarm    severity=CRITICAL    minimum=1    maximum=1    type=myCustomAlarmType
     Should Be Equal    ${alarms[0]["someOtherCustomFragment"]["nested"]["value"]}    extra info
     Log    ${alarms}
 
 
+Thin-edge devices support sending custom alarms overriding the type
```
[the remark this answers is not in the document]
rina23q: Honestly I don't know about the use-cases. But this is the same way as we support for measurements and events.

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -236,95 +238,167 @@ impl SourceInfo {
...
+    }
+
+    /// Remove reserved keywords from extras.
+    /// "type", "time", "text", "severity" are ensured that they are not
+    /// in the hashmap of ThinEdgeAlarm because they are already members of the struct itself.
+    fn convert_extras(alarm_data: &ThinEdgeAlarmData) -> HashMap<String, Value> {
+        let mut map = alarm_data.extras.clone();
+        map.remove("externalSource");
+        map
     }
```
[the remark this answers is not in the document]
rina23q: These functions are private and called by `From`. Technically, we can break these functions but I didn't want to have very long implementation inside `From`.

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -236,95 +238,167 @@ impl SourceInfo {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(rename = "externalSource")]
     pub source: Option<SourceInfo>,
 
-    pub severity: String,
+    /// If None, the message is meant to clear the alarm,
+    /// Some means to create a new alarm.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde(flatten)]
+    pub data: Option<C8yCreateAlarmData>,
```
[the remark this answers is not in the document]
rina23q: Good point. It can be a flag `clear: true` for example. I'll consider.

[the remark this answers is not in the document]
didier-wenzek: @albinsuresh at first, I was aligned with you, asking @rina23q to group all related data into a single `struct`.

But I have to acknowledge that there is a nice property of this two-parts message. This gives a nice way to represent either an alarm or a message to clear the alarm. There is a similar issue with commands, where one needs to be cautious before [de-serializing a command message](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/converter.rs#L740) because the command structs are enable to represent the clearing case.

crates/core/tedge_api/src/alarm.rs
rina23q: I'm not sure if I understood @albinsuresh here correctly. So what I wanted to say here are
1. `Option<String>` is wrong. It should be `String` as @didier-wenzek suggested.
2. The value of `alarm_type` is always delivered by [Channel](https://github.com/thin-edge/thin-edge.io/blob/c0ded23e1927a156f3f83941eda8fc854c6c997f/crates/core/tedge_api/src/mqtt_topics.rs#L323). Since our MQTT scheme allows an empty string as an alarm type (at least this `Channel` implementation returns `String("")`), we need to remember this `alarm_type` may have `String("")`. 
3. If we don't want to allow `String("")`, the implementation should be done in the `Channel` level.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
rina23q: Just not that `impl TryFrom<&Message> for ThinEdgeAlarm` is not ideal as `alarm_type` and `source` are given by the upper level. Re-invention of Message to `alarm_type` and `source` are not smart.

Here we need 3 things to create `ThinEdgeAlarm`.
1. `alarm_type` (originally from topic, but here derived from upper level)
2. `source` (originally from topic, but here derived from upper level)
3. `mqtt_payload` (get from `Message`).

Technically we can create a generic struct and implement `TryFrom` for the struct as these 3 parameters are the same for measurements and events. However, I'd address this level of sophistication later.

In short, I will create a function which takes the 3 fields and return `Result<ThinEdgeAlarm, Err>` in this PR.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
rina23q: Done with ec55b7328b99294efd1e2e16b7430bb7bed5a371

crates/core/tedge_api/src/alarm.rs
rina23q: Done with ec55b7328b99294efd1e2e16b7430bb7bed5a371

crates/core/tedge_api/src/alarm.rs
```diff
@@ -64,316 +90,117 @@ pub enum ThinEdgeJsonDeserializerError {
...
-            AlarmSeverity::Critical => write!(f, "CRITICAL"),
-            AlarmSeverity::Major => write!(f, "MAJOR"),
-            AlarmSeverity::Minor => write!(f, "MINOR"),
-            AlarmSeverity::Warning => write!(f, "WARNING"),
-        }
-    }
-}
-
-impl ThinEdgeAlarm {
-    pub fn try_from(
```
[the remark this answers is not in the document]
rina23q: Done with ec55b7328b99294efd1e2e16b7430bb7bed5a371

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
rina23q: Done with ec55b7328b99294efd1e2e16b7430bb7bed5a371

crates/core/c8y_api/src/json_c8y.rs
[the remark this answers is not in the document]
rina23q: I think @albinsuresh and @didier-wenzek are on the same page here. But I believe that @albinsuresh says `C8yCreateAlarm` doesn't need to have a sub-struct as it's used only for **serializing** a struct to c8y's way.

In c8y's API, [`"status": "CLEARED"`](https://cumulocity.com/guides/10.4.6/reference/alarms/#alarm-applicationvndcomnsncumulocityalarmjson) is the key. We don't need to use it now as we clear an alarm by SmartREST, however, once we move to JSON over MQTT, we will need this field. 

So I'll change the current struct to align more with c8y.

tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot
```diff
@@ -69,12 +69,50 @@ Thin-edge devices support sending custom events without type in topic
...
+    Log    ${alarms}
+
+
+Thin-edge devices support sending custom alarms without severity in payload
+    Execute Command    tedge mqtt pub te/device/main///a/myCustomAlarmType2 '{ "text": "Some test alarm" }'
+    ${alarms}=    Device Should Have Alarm/s    expected_text=Some test alarm    severity=MINOR    minimum=1    maximum=1    type=myCustomAlarmType2
+    Log    ${alarms}
+
+
+Thin-edge devices support sending custom alarms with unknown severity in payload
```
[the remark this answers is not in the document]
reubenmiller: To be honest, I'm ok with a more lenient way handling the severity (e.g. using a default value when an invalid value is given).

crates/core/c8y_api/src/json_c8y.rs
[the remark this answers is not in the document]
albinsuresh: Aah okay. As long as there's an outer `From` that the consumer can directly use, that's sufficient. I missed that there was one, wrapping these.

crates/core/c8y_api/src/json_c8y.rs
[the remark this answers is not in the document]
albinsuresh: @didier-wenzek I was also fully aligned with you on that decomposition of the `ThinEdgeAlarm` struct, precisely for that reason that you highlighted. That's why even the former struct had this decomposition using a sub-struct representing the payload. But, here, as @rina23q also highlighted, my concern was about replicating the same style for the `C8yCreateAlarm` struct, which looked unnecessary, as the representation of a "clear alarm message" was not an empty sub-struct in C8y.

IMO, instead of overloading the `C8yCreateAlarm` struct, by making `severity` and `text` fields optional, to handle clearing as well, it'd be better have an entirely different struct, a `C8yClearAlarm`, to represent that case.

tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot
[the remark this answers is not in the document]
albinsuresh: Aah okay. I thought that rule applied only for measurements as the old regime did not support types in the topic, where-as it has always been mandatory for alarms and events. But, I guess this is fine for the sake of consistency.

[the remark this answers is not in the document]
rina23q: @albinsuresh Yes, I would really recommend that we should keep the consistency in all measurements, events, and alarms at this PR (the main goal is to support the new "te" topic), then later change/unchange all of them together if we want to change the detail level decisions..

crates/core/c8y_api/src/json_c8y.rs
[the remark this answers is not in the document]
rina23q: > as the representation of a "clear alarm message" was not an empty sub-struct in C8y.

Correct. I start thinking rather making two different struct for creating an alarm and clearing an alarm is the right way. Why?
1. As @albinsuresh said, "clearing an alarm" is not an empty sub-struct. It has ["status": "CLEARED"](https://cumulocity.com/guides/10.4.6/reference/alarms/#alarm-applicationvndcomnsncumulocityalarmjson) or not.
2. But as I said, having **optional** `severity`, `text` and `time` are also wrong as they are mandatory to create an alarm.  

So I start feeling we are talking about two different (but similar in terms of alarm) data models and trying to keep it in one struct. I would go to have two structs if others agree with me. Otherwise, I will not change.

crates/core/c8y_api/src/json_c8y.rs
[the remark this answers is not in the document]
rina23q: Exactly. As the test shows ;)

```rust
let c8y_alarm = C8yCreateAlarm::from(&tedge_alarm);
```

crates/core/c8y_api/src/smartrest/alarm.rs
```diff
@@ -1,168 +1,136 @@
...
+use crate::json_c8y::C8yCreateAlarm;
 use time::format_description::well_known::Rfc3339;
-use time::OffsetDateTime;
 
-/// Converts from thin-edge alarm to C8Y alarm SmartREST message
-pub fn serialize_alarm(alarm: ThinEdgeAlarm) -> Result<String, time::error::Format> {
-    match alarm.data {
-        None => Ok(format!("306,{}", alarm.name)),
+/// Serialize C8yCreateAlarm to SmartREST message
+pub fn serialize_alarm(alarm: C8yCreateAlarm) -> Result<String, time::error::Format> {
```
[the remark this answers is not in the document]
albinsuresh: Creating the SmartREST alarm directly from the `ThinEdgeAlarm` was intentional in the old regime, to avoid the creation of intermediate structs. That's why necessary defaults were also applied at this level itself, when the input didn't have some. But, now that you've made this change and the plan is to migrate JSON over MQTT anyway, which would need this `C8yCreateAlarm` struct anyway, let it be.

crates/core/c8y_api/src/json_c8y.rs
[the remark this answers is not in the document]
rina23q: Discussed with @albinsuresh, and this model satisfies both 1) and 2) as I mentioned in the above comment. I'll change.

```rust
enum C8yAlarm {
    Create(C8yCreateAlarm),
    Clear(C8yClearAlarm)
    
}

struct C8yCreateAlarm {
    // all necessary fields, and all are mandatory.
    
}

struct C8yClearAlarm {
    alarm_type: &str,
    source: Option<SourceInfo>,
}
```

```rust
let c8y_alarm = C8yAlarm::from(&tedge_alarm);
```

crates/extensions/c8y_mapper_ext/src/converter.rs
rina23q: As @didier-wenzek, changing from `EntityMetadata` to `EntityTopicId` solved this. https://github.com/thin-edge/thin-edge.io/pull/2267/commits/6536b54e6401841d88d0b20c3e24650ff8d9e894

crates/extensions/c8y_mapper_ext/src/error.rs
```diff
@@ -115,6 +115,9 @@ pub enum ConversionError {
 
     #[error(transparent)]
     FileError(#[from] FileError),
+
+    #[error(transparent)]
+    FileC8yAlarmError(#[from] c8y_api::json_c8y::C8yAlarmError),
```
didier-wenzek, line 120: Why a `File` prefix?

crates/core/c8y_api/src/smartrest/topic.rs
```diff
@@ -101,6 +102,21 @@ impl From<&EntityMetadata> for C8yTopic {
...
+                None => Self::SmartRestResponse,
+                Some(info) => Self::ChildSmartRestResponse(info.id.clone()),
+            },
+            C8yAlarm::Clear(alarm) => match alarm.source.as_ref() {
+                None => Self::SmartRestResponse,
+                Some(info) => Self::ChildSmartRestResponse(info.id.clone()),
+            },
+        }
+    }
+}
```
didier-wenzek, line 118: This will have to be updated, once merged this PR: https://github.com/thin-edge/thin-edge.io/pull/2266#discussion_r1330224805

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -286,41 +288,47 @@ pub struct C8yClearAlarm {
...
+                    source,
+                    severity: C8yCreateAlarm::convert_severity(tedge_alarm_data),
+                    text: C8yCreateAlarm::convert_text(tedge_alarm_data, &alarm_type),
+                    time: C8yCreateAlarm::convert_time(tedge_alarm_data),
+                    fragments: C8yCreateAlarm::convert_extras(tedge_alarm_data),
+                }),
+            };
+            Ok(c8y_alarm)
+        } else {
+            Err(C8yAlarmError::UnsupportedExternalDeviceId(
```
didier-wenzek, line 313: The error type is not appropriate: the source is device topic identifier not an external device id.

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -236,96 +240,184 @@ impl SourceInfo {
...
-            Some(data) => {
-                text = data.text.clone().unwrap_or_else(|| alarm_type.clone());
-                time = data.time.unwrap_or_else(OffsetDateTime::now_utc);
-                fragments = data.alarm_data.clone();
-            }
+    fn convert_source(entity: &EntityMetadata) -> Option<SourceInfo> {
+        match entity.r#type {
+            EntityType::MainDevice => None,
+            EntityType::ChildDevice => Some(make_c8y_source_fragment(&entity.entity_id.clone())),
+            EntityType::Service => None, // TODO!: What is for service?
```
didier-wenzek, line 323: A service is addressed the same way as for a child-device.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -318,34 +318,20 @@ impl CumulocityConverter {
...
-            }
-            mqtt_messages.append(&mut messages);
-            Ok(mqtt_messages)
-        } else if topic.name.starts_with(INTERNAL_ALARMS_TOPIC) {
-            self.alarm_converter.process_internal_alarm(message);
-            Ok(vec![])
-        } else {
-            Err(ConversionError::UnsupportedTopic(topic.name.clone()))
-        }
+        self.size_threshold.validate(input)?;
```
didier-wenzek, line 325: This check can be removed. Only the size of output messages must be checked.

didier-wenzek (commented):
Really close to be merged. However, I still have some questions and comments.

crates/extensions/c8y_mapper_ext/src/error.rs
rina23q: Lol. I thought it's `From` :D:D

crates/core/c8y_api/src/smartrest/topic.rs
rina23q: Also, alarm won't use this after #2242 is done.

crates/core/c8y_api/src/json_c8y.rs
rina23q: Done with 816a8cb06525fbaef157ef25690ba983ed647949

crates/core/c8y_api/src/json_c8y.rs
rina23q: Done with 816a8cb06525fbaef157ef25690ba983ed647949

crates/extensions/c8y_mapper_ext/src/error.rs
rina23q: Fix with 816a8cb06525fbaef157ef25690ba983ed647949

crates/extensions/c8y_mapper_ext/src/converter.rs
rina23q: I tried to move this logic after creating either SmartREST or JSON over MQTT message, but one unit test failed. Therefore, I'll leave it as it is now. 

P.S. If the input JSON is way too big, most likely the output mqtt message is also too big. So, checking input is not completely useless.

didier-wenzek: Okay, we will see later.

albinsuresh: > If the input JSON is way too big, most likely the output mqtt message is also too big. So, checking input is not completely useless.

Yes, that was the reason why it was done, to avoid an unnecessary conversion of a huge input that'd be thrown away a few steps later.

albinsuresh (approved):

PR #2271, opened 2023-09-19
tedge-agent: move to new health topic

Bravo555:
TODO
- [ ] add test checking tedge-agent works with new topics
- [ ] figure out how to handle health command not having a command_id

## Proposed changes

tedge-agent was updated to work with new health topics:

- it listens to health check command on a new topic: `te/[DEVICE_TOPIC_ID]/service/tedge-agent/cmd/health/check`
- it publishes its health status on a new topic: `te/[DEVICE_TOPIC_ID]/service/tedge-agent/status/health

### Converter

However, to still be observable and work with other components and the tests, the converter was modified to properly bridge new inputs and output topics of tedge-agent:

- `tedge/health-check` and `tedge/health-check/SERVICE_NAME` are converted into `te/device/main///cmd/health/check` and `te/device/main/service/SERVICE_NAME/cmd/health/check` respectively, so that tedge-agent receives a health check command on a topic it expects
- `te/device/main/service/tedge-agent/status/health` and `te/device/CHILD_ID/service/tedge-agent/status/health` are converted into `tedge/health/tedge-agent` and `tedge/health/CHILD_ID/tedge-agent` respectively.
- in order not to cause a loop, a previous `tedge/health/SERVICE_NAME` -> `te/...` conversion was modified to not happen if SERVICE_NAME is "tedge-agent"

### tedge-api

A new type, `ServiceHealthTopic` was introduced to denote health topics of services. Currently it is a simple string wrapper, handling both new and old health topics, but after all components are moved, it will handle only new topics.

`ServiceTopicId` and `DeviceTopicId` were added as simple wrappers of `EntityTopicId` that are confirmed to be either a service or a device.

Because `te/device/main/service/tedge-agent/cmd/health/check` doesn't have a command_id, an OperationType struct was modified to make command_id optional.
## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2018#issuecomment-1705325553

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
+++ b/crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
@@ -2,10 +2,14 @@ use serde_json::Value;
 use std::collections::HashMap;
 use std::convert::Infallible;
 use tedge_actors::Converter;
+use tedge_api::mqtt_topics::Channel;
+use tedge_api::mqtt_topics::MqttSchema;
 use tedge_mqtt_ext::MqttMessage;
 use tedge_mqtt_ext::Topic;
 
-pub struct TedgetoTeConverter {}
+pub struct TedgetoTeConverter {
+    mqtt_schema: MqttSchema,
+}
 
 impl Converter for TedgetoTeConverter {
     type Input = MqttMessage;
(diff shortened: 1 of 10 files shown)


crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -305,6 +305,75 @@ impl EntityTopicId {
...
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+/// Represents an entity topic identifier known to be a service.
+///
+/// It's most often in a format `device/DEVICE_NAME/service/SERVICE_NAME`.
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct ServiceTopicId(EntityTopicId);
```
didier-wenzek, line 342: Do we really need a type for `ServiceTopicId`? I don't think so: an entity topic id can be used for services as well as for devices.

crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
```diff
@@ -19,20 +23,56 @@ impl Converter for TedgetoTeConverter {
...
+            // to be able to move different services to new health topic at different times, we
+            // selectively map services either from old to new topics, and vice versa.
+            // tedge-agent publishes on new health topic, exclude it from old mapping not to produce
+            // a publish loop
+            topic
+                if topic.name.starts_with("tedge/health")
+                    && !topic.name.contains("tedge-agent") =>
+            {
                 self.convert_health_status_message(message)
             }
```
didier-wenzek, line 57: Would not be simpler to update all the services in a single PR?

Or even simpler:

- to start with the service *consuming* health check messages (the mapper and [tedge-watchdog](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_watchdog/src/systemd_watchdog.rs#L89), so all the health status are consumed from the `te` topics.
- to update one after the other the services (so they publish health status on the `te` topics).

I realized for that to work, one needs to update the `tedge_to_te_converter` to also converts health check requests. So, we will be able to update the services (as the agent) one after the other so they react to `te/+/+/+/+/cmd/health/check`.

The very last step will be to update tedge-watchdog to send health check requests on the  `te` topics.

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: I think in some places (e.g. health checks) we do actually want to operate on services, and so to make sure a given topic is a service topic, we need to either look at it (if in default MQTT scheme) or call into entity store (if not). IMO it'd be nice not to write a guard every time to make sure that it is a service.
If I'm mistaken about there being more places where a function expects an entity topic id to be a service, then I can remove it for now and do it with plain `ServiceTopicId`s.

crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
Bravo555: You're right, it would be much simpler and it's something that we have to do anyway, so better to do it now.

Bravo555:
Much of what was done here was done in #2276 instead, making this PR obsolete.

Issue #2256, opened 2023-09-13
azure tedge mapper fails to translate health status messages


PradeepKiruvale:
**Describe the bug**

When `tedge-mapper-az` is started and receives the `health status messages`, mapper fails to translate these messages and throws the error below

```
023-09-13T06:23:22.76119982Z ERROR az_mapper_ext::converter: Mapping error: Invalid JSON: invalid type: integer `1694586060`, expected a borrowed string at line 1 column 44: `0}
`    
2023-09-13T06:23:22.76150372Z ERROR az_mapper_ext::converter: Mapping error: Invalid JSON: invalid type: integer `1694581730`, expected a borrowed string at line 1 column 42: `0}
`    
2023-09-13T06:23:22.76159542Z ERROR az_mapper_ext::converter: Mapping error: Invalid JSON: invalid type: integer `1694581730`, expected a borrowed string at line 1 column 42: `0}
```

**To Reproduce**
1. Install thin-edge.io

2. Configure the tedge-mapper-c8y to connect to [azure ](https://github.com/thin-edge/thin-edge.io/blob/main/docs/src/start/connect-azure.md)

3. Publish a message to simulate the mosquitto bridge health message

    ```
    tedge mqtt pub tedge/health/tedge-mapper-az '{"pid":19753,"status":"up","time":1694586060}'
    ```

4. Inspect the tedge-mapper-az logs for any error messages related to mapping/parsing errors

    ```
    journalctl -u tedge-mapper-az -n 50
    ```

**Expected behavior**

The az mapper should properly parse the `health` status message and output on `az/messages/events` topic.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version] : `ubuntu 21`
 - Hardware [incl. revision]: `x86`
 - System-Architecture [e.g. result of "uname -a"]:
 - thin-edge.io version [e.g. 0.1.0]: `0.12.1~237+g7f58cd2`

**Additional context**

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the bug and could not reproduce it anymore.

Issue #2237, opened 2023-09-08
tedge-mapper-aws health message uses two different timestamp formats for up and down messages


reubenmiller:
**Describe the bug**

Inconsistent `time` format between the outgoing `aws/td/health/tedge-mapper-aws` up/down messages.

The status up message uses a unix timestamp:

```
[aws/td/health/tedge-mapper-aws] {"pid":33547,"status":"up","time":1694131076}
```

Where as the down message uses RFC3339 format:

```sh
[aws/td/health/tedge-mapper-aws] {"pid":33394,"status":"down","time":"2023-09-07T23:57:56.732011163Z"}
```

The root cause seems to be due to the up message knows the timestamp upfront and is set in the payload using a unix timestamp, then the aws converted preserves the timestamp in the original format. The down message is sent via the MQTT broker LWT (Last Will and Testament) message and does not include a timestamp as LWT message are static and defined when the process is started, not when the process stops, therefore when the LWT message is processed by the AWS converter, it add the `time` property using the RFC3339 formatted timestamp.

Below shows the difference in payloads which leads to the inconsistency being added by the converter:

```sh
# Up message
[tedge/health/tedge-mapper-aws] {"pid":33394,"status":"up","time":1694130288}

# Down message
[tedge/health/tedge-mapper-aws] {"pid":33394,"status":"down"}
```

Using different formats for different properties makes it harder for other components to reliably subscribe to the data, as they need to cater for different format types which makes parsing the payloads more difficult in some programming languages.

**To Reproduce**

1. Install thin-edge.io
2. Configure the tedge-mapper-c8y to connect to AWS (this [gist](https://gist.github.com/reubenmiller/0aa9dfb6fabbfdbe76156e9c3af4f33f) can help guide you through the setup)
3. Subscribe to the `aws/#` topics on the local broker

    ```sh
    tedge mqtt sub 'aws/#'
    ```

4. In another console, publish a message to simulate the health message messages, both up and down

    ```sh
    tedge mqtt pub -r tedge/health/tedge-mapper-aws '{"pid":33394,"status":"up","time":1694130288}'
    tedge mqtt pub -r tedge/health/tedge-mapper-aws '{"pid":33394,"status":"down"}'
    ```

5. Inspect the subscribed messages from subscription step and view the `time` property. The expectation is that the `time` property are both in the same format

**Expected behavior**

The `time` property should be in the same format (e.g. either unix timestamp or RFC3339) for both the up and down messages.

[...]

PR #2273, opened 2023-09-20
azure tedge mapper fails to translate health status messages

PradeepKiruvale:
## Proposed changes

Issue:
Azure mapper converter was failing to translate the service health message `timestamp`, because it's an integer number.

Proposed solution/fix:
Forward the health status message payload to Azure Cloud without translating.

As a followup tasks
- [ ] Unify the timestamps to `RFC3339` format across thin-edge.io https://github.com/thin-edge/thin-edge.io/issues/2237
- [ ] Send meaningful health status messages to Azure cloud by adding more information about the `service name` in the health status message https://github.com/thin-edge/thin-edge.io/issues/2283

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2256

## Checklist

### DIFF
--- a/crates/extensions/az_mapper_ext/src/converter.rs
+++ b/crates/extensions/az_mapper_ext/src/converter.rs
@@ -48,11 +48,16 @@ impl AzureConverter {
     }
 
     fn try_convert(&mut self, input: &MqttMessage) -> Result<Vec<MqttMessage>, ConversionError> {
+        let default_timestamp = self.add_timestamp.then(|| self.clock.now());
         if is_bridge_health(&input.topic.name) {
             Ok(vec![])
+        } else if input.topic.name.starts_with("tedge/health") {
+            let mut input_msg = input.clone();
+            input_msg.topic = self.mapper_config.out_topic.clone();
+            Ok(vec![input_msg])
         } else {
             self.size_threshold.validate(input)?;
-            let default_timestamp = self.add_timestamp.then(|| self.clock.now());
+
             let mut serializer = ThinEdgeJsonSerializer::new_with_timestamp(default_timestamp);
             tedge_api::parser::parse_str(input.payload_str()?, &mut serializer)?;
 
@@ -259,5 +264,35 @@ mod tests {
         ));
         let res = result.unwrap();
         assert!(res.is_empty());
+    }
+
+    #[test]
+    fn converting_service_health_status_up_message() {
+        let mut converter = AzureConverter::new(false, Box::new(TestClock));
+
+        let input = r#"{"pid":1234,"status":"up","time":1694586060}"#;
+        let result = converter.try_convert(&MqttMessage::new(
+            &Topic::new_unchecked("tedge/health/tedge-mapper-az"),
+            input,
+        ));
+
+        let expected_msg = MqttMessage::new(&Topic::new_unchecked("az/messages/events/"), input);
+        let res = result.unwrap();
+        assert_eq!(res[0], expected_msg);
+    }
+
+    #[test]
+    fn converting_service_health_status_down_message() {
+        let mut converter = AzureConverter::new(false, Box::new(TestClock));
+
+        let input = r#"{"pid":1234,"status":"up"}"#;
+        let result = converter.try_convert(&MqttMessage::new(
+            &Topic::new_unchecked("tedge/health/tedge-mapper-az"),
+            input,
+        ));
+
+        let expected_msg = MqttMessage::new(&Topic::new_unchecked("az/messages/events/"), input);
+        let res = result.unwrap();
+        assert_eq!(res[0], expected_msg);
     }
 }
--- a/crates/extensions/az_mapper_ext/src/error.rs
+++ b/crates/extensions/az_mapper_ext/src/error.rs
@@ -23,4 +23,7 @@ pub enum ConversionError {
 
     #[error(transparent)]
     MqttError(#[from] MqttError),
+
+    #[error(transparent)]
+    FromSerdeJson(#[from] serde_json::Error),
 }


crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -48,11 +50,26 @@ impl AzureConverter {
...
+                let timestamp = timestamp
+                    .format(&time::format_description::well_known::Rfc3339)?
+                    .as_str()
+                    .into();
+                payload_json.entry("time").or_insert(timestamp);
+            }
+            let payload = serde_json::to_string(&payload_json)?;
+            Ok(vec![
+                (MqttMessage::new(&self.mapper_config.out_topic, payload)),
+            ])
```
didier-wenzek, line 69: Why not simply publish the health status message unchanged?

This piece of code will only add a timestamp to `down` status (the only ones without a `time` field).
And this done using a different time format as for the `up` status (as demonstrated by the `converting_service_health_status` test).

An alternative, is to *rewrite* the `time` field if not using the RFC 3339 format.

crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -48,11 +50,26 @@ impl AzureConverter {
     }
 
     fn try_convert(&mut self, input: &MqttMessage) -> Result<Vec<MqttMessage>, ConversionError> {
+        let default_timestamp = self.add_timestamp.then(|| self.clock.now());
         if is_bridge_health(&input.topic.name) {
             Ok(vec![])
+        } else if input.topic.name.starts_with("tedge/health") {
```
rina23q, line 56: This check is too wild. The mapper subscribes to `tedge/health/+` and `tedge/health/+/+`. Therefore, for example, I think this c8y message will be translated by the azure mapper.

```
tedge mqtt pub tedge/health/tedge-mapper-c8y '{"pid":19753,"status":"up","time":1694586060}'
```

crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -48,11 +50,26 @@ impl AzureConverter {
...
+            let mut payload_json: Map<String, Value> =
```
rina23q, line 57: Why not preparing a struct and use `serde` to the struct like this?

```rust
BridgeHeathMessage {
    pid: u32,
    
    status: BridgeStatus,

    #[serde(skip_serializing_if = "Option::is_none")]
    #[serde(with = "time::serde::rfc3339")]
    time: Option<OffsetDateTime>, 
}

#[serde(rename_all = "PascalCase")]
enum BridgeStatus {
    Up,
    Down
}
```

rina23q (commented):
Some comments.

1. Tests are missing! I believe adding them to [converter.rs](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/az_mapper_ext/src/converter.rs#L130) is the right thing.

2. I'm not fully sure if I knew the requirements correctly. So, azure converter should translate these kind of messages? 

From
```
[tedge/health/tedge-mapper-az] {"pid":51367,"status":"down"}
[tedge/health/tedge-mapper-az] {"pid":13280,"status":"up","time":1675330667}
```

To
```
[az/messages/events/] {"pid":51367,"status":"down"}
[az/messages/events/] {"pid":13280,"status":"up","time": 2023-09-20T14:37:49+00:00}
```

crates/extensions/az_mapper_ext/src/converter.rs
reubenmiller: @rina23q Though shouldn't the az mapper forward all service status messages that are published in the json format? It should not care which service is actually behind it, as you can use Azure to monitor the c8y service.

crates/extensions/az_mapper_ext/src/converter.rs
PradeepKiruvale: Yes, we can push the whole `time stamp` without doing any sort of translation.

PradeepKiruvale:
> Some comments.
> 
> 1. Tests are missing! I believe adding them to [converter.rs](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/az_mapper_ext/src/converter.rs#L130) is the right thing.
> 2. I'm not fully sure if I knew the requirements correctly. So, azure converter should translate these kind of messages?
> 
> From
> 
> ```
> [tedge/health/tedge-mapper-az] {"pid":51367,"status":"down"}
> [tedge/health/tedge-mapper-az] {"pid":13280,"status":"up","time":1675330667}
> ```
> 
> To
> 
> ```
> [az/messages/events/] {"pid":51367,"status":"down"}
> [az/messages/events/] {"pid":13280,"status":"up","time": 2023-09-20T14:37:49+00:00}
> ```



> Some comments.
> 
> 1. Tests are missing! I believe adding them to [converter.rs](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/az_mapper_ext/src/converter.rs#L130) is the right thing.
> 2. I'm not fully sure if I knew the requirements correctly. So, azure converter should translate these kind of messages?
> 
> From
> 
> ```
> [tedge/health/tedge-mapper-az] {"pid":51367,"status":"down"}
> [tedge/health/tedge-mapper-az] {"pid":13280,"status":"up","time":1675330667}
> ```
> 
> To
> 
> ```
> [az/messages/events/] {"pid":51367,"status":"down"}
> [az/messages/events/] {"pid":13280,"status":"up","time": 2023-09-20T14:37:49+00:00}
> ```
I already added a test with `time`.I will add it without a time stamp.

crates/extensions/az_mapper_ext/src/converter.rs
PradeepKiruvale: > Why not simply publish the health status message unchanged?
> 
> This piece of code will only add a timestamp to `down` status (the only ones without a `time` field). And this done using a different time format as for the `up` status (as demonstrated by the `converting_service_health_status` test).
> 
> An alternative, is to _rewrite_ the `time` field if not using the RFC 3339 format.

@reubenmiller what's your thought on this? Shall we translate the time to RFC 3339 format or send out the message as it comes in?

crates/extensions/az_mapper_ext/src/converter.rs
rina23q: @reubenmiller Ah okay, but then simple forwarding the payload `{"pid":19753,"status":"up","time":1694586060}` to `az/messages/events/` is not enough. How can user know which service is behind the message?

Honestly speaking, I'm still confused what is preferred conversion output for azure.

crates/extensions/az_mapper_ext/src/converter.rs
reubenmiller: We need to be consistent, otherwise we'll have the same problem as we have in https://github.com/thin-edge/thin-edge.io/issues/2237.

I would align to what the majority is or if you can find best practises in azure about timestamps. RFC3339 is more human readable but less machine readable.

crates/extensions/az_mapper_ext/src/converter.rs
PradeepKiruvale: It may be for another PR, but I feel maybe we can use the `{property-bag}` messages to send extra information. Find more info [here](https://learn.microsoft.com/en-us/azure/iot/iot-mqtt-connect-to-iot-hub#sending-device-to-cloud-messages).

crates/extensions/az_mapper_ext/src/converter.rs
PradeepKiruvale: @didier-wenzek The root cause is the API `health_status_up_message(daemon_name: &str) ` where the timestamp is added as a `Unix` timestamp. So, shall we add in `RFC 3339` format while adding a timestamp to the status message? then it will resolve all the issues and no need to re-write while parsing.

crates/extensions/az_mapper_ext/src/converter.rs
PradeepKiruvale: We are not going to parse the bridge status to the cloud, so it will be filtered out in line #54. The structure that you are suggesting is for `service` health status right?

crates/extensions/az_mapper_ext/src/converter.rs
reubenmiller: Yes we can do the adjustments to how to communicate which service it is in a follow service. Though someone needs to point some thought into how to do this following the best practices of Azure IoT Hub.

crates/extensions/az_mapper_ext/src/converter.rs
reubenmiller: @PradeepKiruvale Technically yes, but you can apply the same suggested idea to the service message struct...though we have to remember that we should always be preserving extra fields in the message format.

PradeepKiruvale: Sure, I will do this as part of another ticket where we need to enhance the `health status` message to have `service` information. As of now I just forward the payload as it comes in, without translating.

rina23q: I understood now, as of this PR's scope, having a struct is over-engineering. Once we decided some nice way to report service to az, depending on the decision, we might need a representation of the message model.

rina23q (approved):
Looks a good bug fix!

crates/extensions/az_mapper_ext/src/error.rs
```diff
@@ -23,4 +23,10 @@ pub enum ConversionError {
 
     #[error(transparent)]
     MqttError(#[from] MqttError),
+
+    #[error(transparent)]
+    FromSerdeJson(#[from] serde_json::Error),
+
+    #[error(transparent)]
+    FromTimeFormatError(#[from] time::error::Format),
```
didier-wenzek, line 31: Is this still an error case?

crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -48,11 +48,16 @@ impl AzureConverter {
     }
 
     fn try_convert(&mut self, input: &MqttMessage) -> Result<Vec<MqttMessage>, ConversionError> {
+        let default_timestamp = self.add_timestamp.then(|| self.clock.now());
         if is_bridge_health(&input.topic.name) {
             Ok(vec![])
+        } else if input.topic.name.starts_with("tedge/health") {
+            Ok(vec![
+                (MqttMessage::new(&self.mapper_config.out_topic, input.payload_str()?)),
```
didier-wenzek, line 56: As the point is to only change the topic, I would avoid to touch the payload.

crates/extensions/az_mapper_ext/src/converter.rs
didier-wenzek: >  So, shall we add in RFC 3339 format while adding a timestamp to the status message?

Indeed an idea to consider.

crates/extensions/az_mapper_ext/src/error.rs
PradeepKiruvale: No, removed it

crates/extensions/az_mapper_ext/src/converter.rs
PradeepKiruvale: refactored to just update the topic in the same payload.

didier-wenzek (approved):
Approved

Issue #2274, opened 2023-09-20
Health check and status for c8y mapper on `te/`


Bravo555:
The `c8y` mapper is changed to use health status from the `te/+/+/+/+/status/health` and health check from `te/+/+/+/+/cmd/health/check` instead of `tedge/health/+` and `tedge/health-check` topics.

### Feature-related

- [x] #2349
- [x] #2348
- [x] [Use new `te/.../cmd/health/check` command](https://github.com/thin-edge/thin-edge.io/issues/2308)

### Cleanup

- [ ] Rewrite sending service startup registration message so there's less boilerplate
- [ ] Clean up `ServiceTopicId`, `EntityTopicId`, `DeviceTopicId`, `Service`, etc. abstractions
- [ ] fill `serviceType` default in mapper instead of the store
- [ ] Simplify passing of ancestors to functions which need to know ancestors of the given entity

didier-wenzek:
See also:
- https://github.com/thin-edge/thin-edge.io/issues/2253
- https://github.com/thin-edge/thin-edge.io/issues/2308

Bravo555:
Because of the upcoming 0.13 release and all feature-related work has been done, this issue can be closed. A new issue(s) will be created for the remaining clean-up tasks.

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/MQTT_health_check/MQTT_health_endpoints.robot
- tests/RobotFramework/tests/MQTT_health_check/health_tedge-mapper-collectd.robot
- tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
- tests/RobotFramework/tests/cumulocity/software_management/software.robot
- tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
- tests/RobotFramework/tests/tedge/call_tedge_config_list.robot
- tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
- tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
- tests/RobotFramework/tests/tedge/call_tedge_config_list.robot
- tests/RobotFramework/tests/tedge_to_te_converter/convert_tedge_topics_to_te_topics.robot
[...]

PR #2276, opened 2023-09-20
Move c8y-mapper to new health topics

Bravo555:
## Proposed changes

This PR updates most of the services to use new health status topics:

```
tedge/health/SERVICE_NAME
-> te/device/main/service/SERVICE_NAME/status/health

tedge/health/CHILD_ID/SERVICE_NAME
-> te/device/CHILD_ID/service/SERVICE_NAME/status/health
```

Health check commands are not part of this PR and will be done in a follow-up PR.

`tedge_to_te_converter` mapping of `tedge/health` to `te/+/+/+/+/health/status` is still active so e.g. cloud bridges can still publish messages to old topics but be converted to new ones.

## Types of changes

## Paste Link to the issue

- #2274: This PR completes half of the scope of this ticket by moving c8y-mapper and other services to new health status message
- #2308: tedge-watchdog was updated to read health status messages from new topics, but still uses old `tedge/health-check` topic for health check commands

## Checklist

### Follow-up tasks

- [ ] Improve construction of startup registration messages


crates/core/tedge_agent/src/agent.rs
```diff
@@ -182,7 +182,13 @@ impl Agent {
         let signal_actor_builder = SignalActor::builder(&runtime.get_handle());
 
         // Health actor
-        let health_actor = HealthMonitorBuilder::new(TEDGE_AGENT, &mut mqtt_actor_builder);
+        let service_topic_id = self.config.mqtt_device_topic_id.to_service_topic_id("tedge-agent")
+            .with_context(|| format!("Device topic id {} currently needs default scheme, e.g: 'device/DEVICE_NAME//'", self.config.mqtt_device_topic_id))?;
```
Bravo555, line 186: This will fail if `mqtt_device_topic_id` doesn't fit default scheme. Is there a better way to handle this?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -340,29 +341,33 @@ impl CumulocityConverter {
...
+        entity: &EntityTopicId,
         message: &Message,
     ) -> Result<Vec<Message>, ConversionError> {
         let mut mqtt_messages: Vec<Message> = Vec::new();
 
         // Send the init messages
-        if check_tedge_agent_status(message)? {
-            create_tedge_agent_supported_ops(self.ops_dir.clone()).await?;
+        // TODO: use entity store to check service statuses
+        if is_message_tedge_agent_up(message)? {
```
Bravo555, line 351: The health status could arguably be a part of an entity store, since it's supposed to know about all entities anyway. I'd be for deferring it to another PR though.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -305,6 +310,75 @@ impl EntityTopicId {
...
+        if let ["device", device_name, "", ""] = self.0.split('/').collect::<Vec<&str>>()[..] {
+            return Some(ServiceTopicId(EntityTopicId(format!(
+                "device/{device_name}/service/{service_name}"
+            ))));
+        }
+        None
+    }
+
+    /// Returns an array of all segments of this entity topic.
+    pub fn segments(&self) -> [&str; ENTITY_ID_SEGMENTS] {
```
didier-wenzek, line 329: I would really avoid to have this method `pub`: from the API a topic Id is atomic.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -305,6 +310,75 @@ impl EntityTopicId {
...
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+}
+
+/// Represents an entity topic identifier known to be a service.
+///
+/// It's most often in a format `device/DEVICE_NAME/service/SERVICE_NAME`.
+#[derive(Debug, Clone, Hash, PartialEq, Eq)]
+pub struct ServiceTopicId(EntityTopicId);
```
didier-wenzek, line 347: I'm not convinced by this type.

Why would we have a specific type to identify services but not devices?

crates/extensions/tedge_health_ext/src/lib.rs
```diff
@@ -37,12 +44,65 @@ impl HealthMonitorBuilder {
...
+        // Update the MQTT config
+        *mqtt.as_mut() = builder.set_init_and_last_will(mqtt.as_mut().clone());
+
+        builder
+    }
+
+    /// Creates a HealthMonitorBuilder that creates a HealthMonitorActor with
+    /// a new topic scheme.
+    pub fn from_service_topic_id(
+        service: ServiceTopicId,
```
didier-wenzek, line 68: It would be far better to pass the`device` topic id along the `service` topic id, instead of  extracting the former from the latter. Indeed, there is no guaranties that the topic scheme will be the default. 

```suggestion
    /// Creates a HealthMonitorBuilder that creates a HealthMonitorActor with
    /// a new topic scheme.
    pub fn from_service_topic_id(
        device: EntityTopicId,
        service: ServiceTopicId,
```

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -305,6 +310,75 @@ impl EntityTopicId {
...
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+
+    pub fn entity(&self) -> &EntityTopicId {
+        &self.0
+    }
+
+    /// If in a default MQTT scheme, returns a device topic id of this service.
+    pub fn to_device_topic_id(&self) -> Option<DeviceTopicId> {
```
didier-wenzek, line 359: We really should avoid these methods that assume a default topic scheme.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -305,6 +310,75 @@ impl EntityTopicId {
         }
         .map(|parent_id| EntityTopicId(format!("device/{parent_id}//")))
     }
+
+    /// If `self` is a device topic id, return a service topic id under this
+    /// device.
+    ///
+    /// The device topic id must be in a format: "device/DEVICE_NAME//"; if not,
+    /// `None` will be returned.
+    pub fn to_service_topic_id(&self, service_name: &str) -> Option<ServiceTopicId> {
```
didier-wenzek, line 319: We really should avoid these methods that assume a default topic scheme.

didier-wenzek (commented):
I still need some time to finalize this review.

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: Would having this be `pub(crate)` be acceptable?

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: Also addressed it [here](https://github.com/thin-edge/thin-edge.io/pull/2271#discussion_r1331351769).

We can have a specific type to identify devices as well, just by the scope of this PR I thought it was a good place to add one for a service. I think as we plan to support custom MQTT schemes, where any identifier can be either a device or a service, it makes sense to save this information by encoding it in a type. Granted, producers of these wrappers are currently not well defined, so maybe this is a premature abstraction, but ultimately, I think these `ServiceTopicId`/`DeviceTopicId` wrapper types could be produced by the entity store (that is, they can be derived from `EntityType` which is a part of `EntityMetadata`), or derived from a default topic scheme.

crates/extensions/tedge_health_ext/src/lib.rs
Bravo555: You're right, it makes no sense to discard device topic id just to derive it later, but instead of adding another argument to the function, it would be better to embed a device topic id inside a `ServiceTopicId` type, so every service contains information about its associated device.

didier-wenzek: Yes, `ServiceTopicId` starts to make sense when embedding the device topic id.

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: Yes, this assumption of the default topic scheme, spread everywhere with no good way to indicate it (i.e. using a comment or a custom `.expect()` message, instead of better utilising the type system) leads to some unreliable code, and I think implementing a solution to this should be a priority. I'll look at the current usages and make a ticket.

crates/core/tedge_agent/src/agent.rs
didier-wenzek: The agent should be given an entity topic identifier for itself (as an argument). `mqtt_device_topic_id.to_service_topic_id("tedge-agent")` should be used only as a default value.

didier-wenzek: I think this is okay-ish for this PR.

crates/extensions/tedge_health_ext/src/lib.rs
```diff
@@ -37,12 +44,67 @@ impl HealthMonitorBuilder {
...
+        builder
+    }
+
+    /// Creates a HealthMonitorBuilder that creates a HealthMonitorActor with
+    /// a new topic scheme.
+    pub fn from_service_topic_id(
+        service: Service,
+        mqtt: &mut (impl ServiceProvider<MqttMessage, MqttMessage, TopicFilter> + AsMut<MqttConfig>),
+        // TODO: pass it less annoying way
+        mqtt_topic_root: Arc<str>,
```
didier-wenzek, line 71: Why not simply pass a `& MqttSchema` ?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -340,29 +341,33 @@ impl CumulocityConverter {
...
-            );
+        if let Some(parent) = entity.default_parent_identifier() {
+            if parent.is_default_child_device() {
+                let child_id = parent.default_device_name().unwrap();
+                add_external_device_registration_message(
+                    child_id.to_string(),
+                    &mut self.children,
+                    &mut mqtt_messages,
+                );
+            }
```
didier-wenzek, line 366: Beware, this logic is being re-visited by https://github.com/thin-edge/thin-edge.io/pull/2266.

crates/core/tedge_api/src/health.rs
```diff
@@ -1,12 +1,97 @@
...
+            topic.split('/').collect::<Vec<&str>>()[..],
+            ["te", _, _, _, _, "status", "health"]
+        )
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    pub async fn send_health_status(&self, responses: &mut impl PubChannel) {
```
didier-wenzek, line 47: I would really avoid to directly use a `PubChannel`.

crates/core/tedge_api/src/health.rs
```diff
@@ -1,12 +1,97 @@
...
 use serde_json::json;
 use time::OffsetDateTime;
 
+/// Encodes a valid health topic.
+///
+/// Health topics are topics on which messages about health status of services are published. To be
+/// able to send health messages, a health topic needs to be constructed for a given entity.
+// TODO: replace `Arc<str>` with `ServiceTopicId` after we're done with transition to new topics
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ServiceHealthTopic(Arc<str>);
```
didier-wenzek, line 18: I'm not convinced by this struct which mixes unrelated things and has overlapping responsibilities with other structs.

- Is this a struct to materialize health topic structures? To build and send health messages?
- Building topics from a root prefix is the responsibility of `mqtt_schema'.
- Sending messages should be done by actors and not using a direct `PubChannel`.

crates/core/tedge_api/src/health.rs
Bravo555: These functions are more or less direct translations of free functions below, just using the `ServiceHealthTopic` type. I'm not sure how to avoid using `PubChannel`, can you give and example of what you had in mind?

crates/core/tedge_api/src/health.rs
Bravo555: > * Is this a struct to materialize health topic structures? To build and send health messages?

Yes, the intent was to have a type wrapper for a health topic so that one has to first produce a valid health topic in order to be able to send health messages on it, in order not to send a health message on something that is not a health topic. The same of course could be done by putting guards on previously used functions to check if input is a health topic before sending the message, but since all the functions take this argument, it made sense to make it a struct and do the check only once when constructing.


> * Building topics from a root prefix is the responsibility of `mqtt_schema'.

Yeah, you're right that `new()` method on this type is probably unnecessary and it's better to use `MqttSchema` to construct it for the "new topic" case. It still needs to handle old topics though, because all the services use the same health monitoring extension and some were not yet updated to new health topics (e.g. one still have to update tests for them), so that's why for now it's just a simple string wrapper, once the transition is complete it can take `ServiceTopicId` so there would be no dependency for knowing the root prefix. The health extension actor could use it like this:

1. take as input `ServiceTopicId` of a service we need a health check for
2. Create a `ServiceHealthTopic` from `ServiceTopicId`, and use `ServiceHealthTopic` to generate health messages and make use of other health-message-related functionality
3. Use `MqttSchema` to create plain MQTT topic for service health status
4. Send MQTT publish message to MQTT actor, to publish on topic generated by `MqttSchema`, with payload generated by `ServiceHealthTopic`

with steps 3/4 perhaps subject to change because it'd be better to just have a single source for generating health messages. Will refactor this struct to work more like this.


> Sending messages should be done by actors and not using a direct `PubChannel`.

That's how it was working previously and I haven't changed it, but I can certainly improve it in this PR. For previous comment i didn't have context, but "should be done by actors instead of `PubChannel`" is very clear, so I'll make necessary modifications.

crates/core/tedge_api/src/health.rs
```diff
@@ -1,12 +1,97 @@
...
+
+impl ServiceHealthTopic {
+    pub fn new(service: ServiceTopicId) -> Self {
+        // XXX: hardcoded MQTT root
+        ServiceHealthTopic(Arc::from(format!("te/{}/status/health", service.as_str())))
+    }
```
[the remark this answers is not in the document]
Bravo555: In terms of the `new` function, the `service` argument is of type `ServiceTopicId` and that type can be any `+/+/+/+` topic, since in the custom MQTT scheme, for a thing to be a service it has to first register as such, just for `device/device-id/service/service-id` registration to entity store happens automatically.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -95,7 +95,8 @@ const INTERNAL_ALARMS_TOPIC: &str = "c8y-internal/alarms/";
 const C8Y_JSON_MQTT_EVENTS_TOPIC: &str = "c8y/event/events/create";
 const TEDGE_AGENT_LOG_DIR: &str = "tedge/agent";
 const CREATE_EVENT_SMARTREST_CODE: u16 = 400;
-const TEDGE_AGENT_HEALTH_TOPIC: &str = "tedge/health/tedge-agent";
+/// XXX: hardcoded topic root
+const TEDGE_AGENT_HEALTH_TOPIC: &str = "te/device/main/service/tedge-agent/status/health";
```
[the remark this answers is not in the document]
Bravo555: > Now `tedge agent` also can run on a child device also right?

Not yet. The cumulocity mapper which runs on the main device only cares about the main device's agent. This constant is used here:

https://github.com/thin-edge/thin-edge.io/pull/2276/files#diff-08a20a8877f5e1e1bce1aac2c6fc0e1a63a36cc9311e7e60cb3aacacf3c8c57cR351-R352

which does filesystem operations on main device, so it can't be done for child devices as it is now.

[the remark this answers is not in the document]
Bravo555: To not hardcode the MQTT root, rewrote this part to use EntityStore and MqttSchema to generate this topic in [5df9538](https://github.com/thin-edge/thin-edge.io/pull/2276/commits/5df95385746516786409c99c7f279cf028e082d2)

crates/core/tedge_api/src/health.rs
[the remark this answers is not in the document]
Bravo555: Changed in 4b365c3

crates/core/tedge_api/src/health.rs
didier-wenzek: Actually, `send_health_status` is not used and can be simply removed.

crates/extensions/tedge_health_ext/src/lib.rs
```diff
@@ -13,18 +13,23 @@ use tedge_actors::RuntimeRequestSink;
...
 
 pub struct HealthMonitorBuilder {
-    service_name: String,
+    service_health_topic: ServiceHealthTopic,
     box_builder: SimpleMessageBoxBuilder<MqttMessage, MqttMessage>,
 }
 
 impl HealthMonitorBuilder {
+    /// Creates a HealthMonitorBuilder that creates a HealthMonitorActor with
+    /// old topic scheme.
```
albinsuresh, line 32: Any specific reason why we decided not to update the producers as part of this PR? Just to keep this PR simpler or were there some reasons that forced us to keep them on the old scheme? I'm fine with updating them all in a separate PR. But, just wanted to know if simplicity was the only reason.

crates/extensions/c8y_mapper_ext/src/converter.rs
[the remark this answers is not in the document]
didier-wenzek: > To not hardcode the MQTT root, rewrote this part to use EntityStore and MqttSchema to generate this topic in [5df9538](https://github.com/thin-edge/thin-edge.io/commit/5df95385746516786409c99c7f279cf028e082d2)

Beware, this commit is the root cause of the failing system tests.

crates/core/tedge_api/src/health.rs
Bravo555: I'm under the impression that health topics are valid only for services, and not for devices. If there were device health status topics, then I'd agree that `ServiceTopicId` doesn't make sense and to just use `EntityTopicId` since for all possible values of `EntityTopicId` we can generate a vaild health topic.
However, if health topics are only a thing for services, then when using an `EntityTopicId` we can't generate valid health topics for all values, because an `EntityTopicId` can denote either a device or a service, so a function that takes `EntityTopicId` and generates a health status topic is partial: it has to either check with an entity store if a given `EntityTopicId` is a registered service and return an error if it isn't (which is a bit of a pain because you need to pass the entity store) OR it can encode this assumption in a type that can only be constructed when the caller has ensured these conditions hold.

This intent is not clear because the "type can only be constructed when the caller has ensured these conditions hold" part is missing from this PR. Because not services are all updated, `ServiceHealthTopic` needs to handle both old and new topic scheme.

Instead of just updating all the usages with as few code changes as possible, I tried to prototype a new API, in order not to put that off for later, possibly forgetting or having to work on more features, leaving this in a suboptimal state.

We talked with Didier and came to the agreement that in feature related PRs, it's better not to add new API but do things the simplest possible way and then do PRs entirely dedicated to changing API, so that new features are not delayed by API discussions. So even If I think this is a good abstraction, I agree that perhaps I introduced it prematurely.

But since it's already done, I can either improve this abstraction, add more documentation hopefully better explaining why I think it has a purpose, or if you're still unconvinced I can remove it and just use `EntityTopicId`.

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: Responded in [this comment](https://github.com/thin-edge/thin-edge.io/pull/2276/files/5df95385746516786409c99c7f279cf028e082d2#r1338516075), if that's your preference I can remove `ServiceTopicId` and use `EntityTopicId`.

crates/core/tedge_api/src/health.rs
albinsuresh: > I'm under the impression that health topics are valid only for services, and not for devices

IIRC, I've heard comments from Reuben that health status for devices isn't quite unreasonable either. For example, when a device is being shutdown for maintenance, it might make sense for it to report that `maintenance` status as a health status. Currently, C8y doesn't support anything like that. But, that the idea itself isn't unreasonable at thin-edge level.

> it has to either check with an entity store if a given EntityTopicId is a registered service and return an error if it isn't (which is a bit of a pain because you need to pass the entity store) OR it can encode this assumption in a type that can only be constructed when the caller has ensured these conditions hold.

Yeah, it's not very unreasonable to encode an already validated info into a concrete type so that the receiving end doesn't need to revalidate it again. The key thing here is that the caller must validate this against the entity store and not just assume things purely based on the topic structure.

> This intent is not clear because the "type can only be constructed when the caller has ensured these conditions hold" part is missing from this PR.

That's probably one of the reasons why I was struggling to fully digest the usefulness of these additional abstractions like `ServiceTopicId` and `DeviceTopicId`. For something like a health status message generator, yes, this type helps in avoiding a further validation. But, in future, we may have to support health endpoints for devices as well, for the reasons mentioned above. Now, if there's some other logic where the device specific `DeviceTopicId` is useful for a receiver, I'm fine to keep it. One use-case that I can think of is the support for commands, which is currently only limited to devices. But, we haven't fully ruled out future commands support for services either.

> So even If I think this is a good abstraction, I agree that perhaps I introduced it prematurely.

Hearing your arguments, I kinda get the value it may have in future. I just don't any use-cases, **yet**. But, I don't really have any strong arguments to justify the removal of these either, especially since you've already done the heavy-lifting of introducing them.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1255,6 +1256,22 @@ impl CumulocityConverter {
             }
         }
     }
+
+    pub fn is_message_tedge_agent_up(&self, message: &Message) -> Result<bool, ConversionError> {
```
didier-wenzek, line 1260: Let's keep this for now. However, when `tedge-agent` will support software list over `te` topics (see https://github.com/thin-edge/thin-edge.io/pull/2235), then will have to revised.

The mapper will have then to react to the `software-list` capability messages sent by the agents running on the main device as well as child devices.

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: This is why I see more values in the `struct Service` which removes the need for this kind of guessing work.

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: This will be replaced by a check of the software-list capability.

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: As of now, I see indeed little value in `ServiceTopicId` and `DeviceTopicId`. However, this is not blocking.

What really add values is the `struct Service` which combines the `EntityTopicId` of a service with the `EntityTopicId` of its parent device removing the need for guessing one from the other.

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: This is why I see more values in the `struct Service` which removes the need for this kind of guessing work.

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: I finally understood what is worrying me. This `struct ServiceTopicId` and even the `struct Service` are halfway between `EntityTopicId` and `EntityMetadata`.

- `EntityTopicId`s are just identifiers of sources & targets of MQTT messages - with no assumption of the nature of these sources and targets.
- `EntityMetadata`s describe these entities - notably providing their type and their parent if any.

So, if a feature needs to know the type of entity or its parent, then the implementation must be given an `EntityMetadata` not a decorated `EntityTopicId`.

Hence, now I do think that it would be better to not introduce neither `ServiceTopicId` nor `DeviceTopicId` and not even `Service`.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -25,47 +26,37 @@ fn default_type() -> String {
...
+    let child_id = if parent.is_default_child_device() {
+        Some(
+            parent
+                .default_device_name()
+                .expect("EntityTopicId should be in default scheme")
+                .to_string(),
+        )
+    } else {
+        None
+    };
```
didier-wenzek, line 54: All these information should not be guessed from the entity topic identifier by given by the entity store.
The entity store might assume a default topic scheme (as done here) but only when appropriate, i.e. when no specific information has been provided by registration messages.

This function must be given an `EntityMetadata` and not the `EntityTopicId` of the source.

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: This can be fully removed. @albinsuresh  can you confirm?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -446,29 +445,32 @@ impl CumulocityConverter {
...
+                add_external_device_registration_message(
+                    child_id.to_string(),
+                    &mut self.children,
+                    &mut mqtt_messages,
+                );
+            }
         }
 
         let mut message = convert_health_status_message(
+            entity,
```
didier-wenzek, line 473: Before, calling `convert_health_status_message()`, the converter must use its `self.entity_store` to get the `EntityMetadata` attached to the entity topic id. This metadata and not the topic identifier have to be passed to `convert_health_status_message()`.

didier-wenzek (dismissed):
Using the `EntityStore` owned by the c8y converter will help to remove the assumption of the default topic schema.

crates/extensions/tedge_health_ext/src/lib.rs
```diff
@@ -53,10 +113,11 @@ impl HealthMonitorBuilder {
     }
 
     fn set_init_and_last_will(&self, config: MqttConfig) -> MqttConfig {
-        let name = self.service_name.to_owned();
+        let name = self.service_health_topic.to_owned();
+        let _name = name.clone();
         config
-            .with_initial_message(move || health_status_up_message(&name))
-            .with_last_will_message(health_status_down_message(&self.service_name))
+            .with_initial_message(move || _name.up_message())
```
albinsuresh, line 119: Surprised that clippy didn't complain here about using a variable with `_` prefix.

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: > So, if a feature needs to know the type of entity or its parent, then the implementation must be given an `EntityMetadata` not a decorated `EntityTopicId`.

Yeah, I agree that `EntityMetadata` is probably the closest to what we eventually want to have. One problem with it though is that this type is designed for flat storage in a vector, i.e. `parent` field being just a string key that one can look in a hashmap - this is far simpler than using references. So currently, if we want to see who is the parent in `EntityMetadata`, we have to call `.get()` again on the entity store.

The solution for this I think would be to have an `Entity` struct, which is models the entity so as to make invalid states unrepresentable and contain any other relevant entities, e.g. a parent, so one wouldn't have to call entity store again. This is similar to `.ancestors()` method, but instead of returning IDs, it could return objects with all necessary information.

For now though, as agreed, I won't be making any sweeping API changes and just do it with `EntityMetadata` directly, in order to have full picture of the pain points before resolving them.

tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
```diff
@@ -174,30 +172,74 @@ Check if a service is down
...
     Custom Test Setup
+
+    ${previous_service_type}=     Execute Command    tedge config get service.type    strip=True
+    Execute Command               tedge config set service.type thinedge
+
     ThinEdgeIO.Restart Service    ${service_name}
-    Device Should Exist                      ${DEVICE_SN}:device:main:service:${service_name}    show_info=False
+    Device Should Exist           ${DEVICE_SN}:device:main:service:${service_name}    show_info=False
+
+    # If service was previously registered, old serviceType should be maintained
```
reubenmiller, line 184: I'm doubtful with this requirement. I think keeping track of what the previously registered type is will be overly complex to maintain and reason with...but I think it would be a good opportunity creating some sequence diagrams showing the interactions between the registration messages and the `/status/health` telemetry data points.

So maybe once we have that, we can think of how to simplify this.

crates/core/c8y_api/src/smartrest/inventory.rs
```diff
@@ -1,7 +1,19 @@
 //! This module provides some helper functions to create SmartREST messages
 //! that can be used to create various managed objects in Cumulocity inventory.
+
+// TODO: Have different SmartREST messages be different types, so we can see
+// where these messages are used, not only created.
+//
+// TODO: both `C8yTopic::smartrest_response_topic(&EntityMetadata)` and
+// `publish_topic_from_ancestors(&[String])` produce C8y MQTT topics on which
+// smartrest messages are sent. There should be one comprehensive API for
+// generating them.
```
didier-wenzek, line 10: This will have to be addressed by https://github.com/thin-edge/thin-edge.io/issues/2316.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -184,7 +189,20 @@ impl Agent {
...
+            service_topic_id,
+            device_topic_id: DeviceTopicId::new(self.config.mqtt_device_topic_id.clone()),
+        };
+        let mqtt_schema = MqttSchema::with_root(self.config.mqtt_topic_root.to_string());
+        let health_actor = HealthMonitorBuilder::from_service_topic_id(
+            service,
+            &mut mqtt_actor_builder,
+            &mqtt_schema,
+            self.config.service_type.clone(),
+        );
```
didier-wenzek, line 205: This is done again and again for all tedge daemons. I would consider to add a specific `HealthMonitorBuilder::from_config()` constructor.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -121,6 +130,8 @@ pub struct EntityStore {
     entity_id_index: HashMap<EntityExternalId, EntityTopicId>,
     external_id_mapper: ExternalIdMapperFn,
     external_id_validator_fn: ExternalIdValidatorFn,
+    // TODO: this is a c8y cloud specific concern and it'd be better to put it somewhere else.
+    default_service_type: String,
```
didier-wenzek, line 134: I don't even get the purpose of default service type.

This should be related to the [inventory](https://github.com/thin-edge/thin-edge.io/issues/2279)

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -314,12 +347,22 @@ impl EntityStore {
                 }
             }
         };
+
+        let JsonValue::Object(mut other) = message.other else {
+            return Err(Error::EntityRegistrationOtherNotMap);
+        };
+
+        // XXX: this is c8y-specific, entity store shouldn't do this!
```
didier-wenzek, line 355: What difference between `XXX` and `FIXME`?

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -502,6 +563,14 @@ impl EntityRegistrationMessage {
     // Serialize/Deserialize.
     #[must_use]
     pub fn new(message: &Message) -> Option<Self> {
+        let topic_id = message
+            .topic
+            .name
+            .strip_prefix(MQTT_ROOT)
+            .and_then(|s| s.strip_prefix('/'))?;
+
+        dbg!(topic_id);
```
didier-wenzek, line 572:
```suggestion
```

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -80,15 +79,22 @@ pub fn service_monitor_client_config(tedge_config: &TEdgeConfig) -> Result<Confi
...
+        "tedge-mapper-c8y",
+        service_type.as_str(),
+        "down",
+        &[],
+    );
+
     let mqtt_config = tedge_config
         .mqtt_config()?
         .with_session_name("last_will_c8y_mapper")
-        .with_last_will_message(service_monitor_status_message(
```
didier-wenzek, line 86: So the call to `service_monitor_status_message` has been replaced by a more generic one .i.e. `service_creation_message`. This seems okay at first (both are 102 smartrest messages), but are you aware that this last will message has a very specific purpose (telling c8y that thin-edge is down) that is not the same as the other health checks (checking that all the components are up).

crates/extensions/tedge_health_ext/src/lib.rs
```diff
@@ -13,50 +13,80 @@ use tedge_actors::RuntimeRequestSink;
...
+            health_topic,
+            registration_message: Some(registration_message),
             box_builder,
         };
 
         // Update the MQTT config
+
+        // XXX: if the same MqttActorBuilder is used in different actors, then
+        // this will override init messages that may have been set by other
+        // actors!
```
didier-wenzek, line 78: Only a single instance of the `HealthMonitor` actor should be built.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -91,8 +93,9 @@ impl MqttSchema {
     ///     "te/device/child001///a/sensors/meta"
     /// );
     /// ```
+    // TODO: add test
     pub fn topic_for(&self, entity: &EntityTopicId, channel: &Channel) -> mqtt_channel::Topic {
-        let topic = format!("{}/{}/{}", self.root, entity, channel);
+        let topic = format!("{}/{}{}", self.root, entity, channel);
```
didier-wenzek, line 98: What's the purpose of this change? It sounds weird to me to force a channel name to include a `/` prefix.

didier-wenzek (commented):
I have no more blocking comments.
There are still a lot of TODOs: which one do you plan to fix in the context of this PR?

To consider for this PR:
- Introduce a new constructor for ` HealthMonitorBuilder` from tedge-config to avoid repeating the same logic in all tedge daemons.

Follow-up tasks:

- Add a user-configurable service topic id to all tedge daemons, instead of building it after the device id.
- Consider to deprecate `ServiceTopicId`, `EntityTopicId` and `Service`, replacing their usage with `EntityMetadata` or intriducing an `Entity` type (see  https://github.com/thin-edge/thin-edge.io/pull/2276#discussion_r1341540315).
- Consider to deprecate this idea of a service type at the level of thin-edge `Entity`. This should be related to the [inventory](https://github.com/thin-edge/thin-edge.io/issues/2279) i.e. data attached by the user to devices and services.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -45,6 +48,7 @@ pub struct AgentConfig {
     pub data_dir: DataDir,
     pub mqtt_device_topic_id: EntityTopicId,
     pub mqtt_topic_root: Arc<str>,
+    pub service_type: String,
```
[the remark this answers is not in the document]
Bravo555: It's in order to use a correct `type` property when sending a registration message on startup. `serviceType` parameter was removed from health status message, so now `serviceType` can only be set via a registration message. And if a registration message doesn't contain a `type` property, the default is filled by the mapper. This default is put into the retained registration message, so the only way for a service to change its `serviceType` is for it to send a registration message itself, which was done as part of this PR. So all the tedge and c8y daemons and plugins have to read `service.type` to put it into the registration message.

crates/core/tedge_api/src/entity_store.rs
Bravo555: Yes, currently the entity store kind of stores data relevant both to the local thin-edge instance but also to the cloud, i.e. entity external id, and I had to add the default service type because the local registration mainly happens inside the entity store code, and the store needs to be a single source of truth about entities and their properties.

[In another comment](https://github.com/thin-edge/thin-edge.io/pull/2310#discussion_r1349992273) I thought about making an entity store generic over additional data related to the cloud, so that we can use different types for data we always expect to be present for an entity when connected to a certain cloud, i.e. an external id.

I was not familiar with inventory proposal, so I'll have to take a look at it to know if we're not duplicating some of the behaviour that we could attach to an entity store.

crates/core/tedge_api/src/entity_store.rs
Bravo555: I think they can be understood to be roughly the same, I personally use `XXX` to highlight code which i think is problematic and should be addressed relatively quickly [after seeing some other people use it this way](https://drewdevault.com/2023/03/09/2023-03-09-Comment-or-no-comment.html). I've also seen both `XXX` and `FIXME` used in thin-edge, so I used `XXX`. If you consider `FIXME` to be more descriptive, then I can use that instead.

Here, as `serviceType` is something only used by c8y, handling it this way works for now, but it will be unsustainable if we keep adding properties, so this should probably be addressed some other way and refactored in the future.

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: [It's because the registration message topic is e.g. `te/device/main/service/servicename`, instead of `te/device/main/service/servicename/`](https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#entity-registration). These are two different topics in MQTT and the entity store would not pick up the registration message when the topic had a trailing slash. I was rushing a bit and didn't add a test case, but I'll note it down as a TODO.

crates/core/tedge_api/src/entity_store.rs
didier-wenzek: I never noticed usage of XXX and I'm used to FIXME. I'm okay with both: the key point is to agree as a team.

crates/extensions/tedge_health_ext/src/lib.rs
Bravo555: I didn't mean another `HealthMonitor`, but potentially another actor. I wanted to highlight that if multiple actors need to communicate with a single MQTT Actor, then exposing this MQTT actor's internal configuration for other actors so that they can call builder methods on MQTT Actor config, they can overwrite each other's i.e. initial messages. I tried to use `mqtt_channel::Config::with_initial_message` in `MqttBuilderActor`, but it was overridden by `.with_initial_message` call in `HealthMonitorBuilder`.

crates/core/tedge_mapper/src/c8y/mapper.rs
Bravo555: > this last will message has a very specific purpose (telling c8y that thin-edge is down) that is not the same as the other health checks (checking that all the components are up).

I was not aware of the distinction, in fact I was thinking of how to remove this custom last will message for the mapper and handle it more like other services.

Is this change incorrect or does it significantly change the behaviour in some way? If so, then I can fix it and perhaps make a follow-up test if some corner case that we don't currently test for has been violated by this change.

tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
Bravo555: The explicit registration was added and the change to the test was reverted.

crates/core/tedge_mapper/src/c8y/mapper.rs
didier-wenzek: So, we need to double check.

@PradeepKiruvale can you check that this PR doesn't break the status reported by thin-edge on c8y?

crates/extensions/tedge_health_ext/src/lib.rs
didier-wenzek: It's true that sharing an MQTT connection with several actors raises questions (session name, last will message). This is why the mapper has two MQTT connection: a main one to send/receive messages and a second one used only for a last will to be forwarded to c8y (see https://github.com/thin-edge/thin-edge.io/pull/2276#discussion_r1360691979).

That said, there is no reasons to forbid independent actors to add their own init messages. For that, `mqtt_channel::Config` must be updated to handle a `Vec<InitMessageFn>` instead of an `Option<InitMessageFn>`.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -21,182 +15,99 @@ fn default_status() -> String {
...
-        );
-
-        mqtt_messages.push(status_message);
+    if entity.topic_id.as_str().contains("bridge") {
+        return mqtt_messages;
     }
 
-    mqtt_messages
-}
+    let payload_str = message.payload_str().expect("payload is valid utf8");
```
albinsuresh, line 31: That `expect` looks dangerous. We can crash the mapper with a malformed payload, unless we transform it to an `unknown` status as done in the old code.

crates/common/mqtt_channel/src/messages.rs
```diff
@@ -16,7 +16,7 @@ pub struct Message {
 }
 
 #[derive(Clone, Eq, PartialEq)]
-pub struct DebugPayload(Payload);
+pub struct DebugPayload(pub Payload);
```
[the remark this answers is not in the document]
Bravo555: I wanted to move out of the payload instead of cloning, but I ended up rewriting that part so this is no longer necessary. Fixed in 000701e.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -412,19 +540,20 @@ impl Display for Channel {
...
-            Channel::Alarm { alarm_type } => write!(f, "a/{alarm_type}"),
-            Channel::AlarmMetadata { alarm_type } => write!(f, "a/{alarm_type}/meta"),
+            Channel::Alarm { alarm_type } => write!(f, "/a/{alarm_type}"),
+            Channel::AlarmMetadata { alarm_type } => write!(f, "/a/{alarm_type}/meta"),
 
-            Channel::Command { operation, cmd_id } => write!(f, "cmd/{operation}/{cmd_id}"),
-            Channel::CommandMetadata { operation } => write!(f, "cmd/{operation}"),
+            Channel::Command { operation, cmd_id } => write!(f, "/cmd/{operation}/{cmd_id}"),
+            Channel::CommandMetadata { operation } => write!(f, "/cmd/{operation}"),
+            Channel::Health => write!(f, "/status/health"),
```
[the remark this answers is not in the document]
Bravo555: Yes, here it was my bad because I wanted to remove the extra trailing slash in `Channel::EntityMetadata` case and didn't think about how `Display` impl makes it so `Channel` can be printed directly. Updated `topic_for` method so it handles the merging and added a test. Fixed in 2485a25.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -80,15 +79,22 @@ pub fn service_monitor_client_config(tedge_config: &TEdgeConfig) -> Result<Confi
...
     let service_type = tedge_config.service.ty.clone();
 
+    // from this level we don't have access to the entity store and registered main device so best
+    // we can do for now is just guess it
+    // TODO: fix this, preferably use a HealthMonitorActor
+    let service_external_id = format!("{device_name}:device:main:service:tedge-mapper-c8y");
+
+    let last_will_message = c8y_api::smartrest::inventory::service_creation_message(
+        service_external_id.as_str(),
+        "tedge-mapper-c8y",
```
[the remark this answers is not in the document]
Bravo555: Fixed in 2f9893a.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -80,15 +79,22 @@ pub fn service_monitor_client_config(tedge_config: &TEdgeConfig) -> Result<Confi
     let device_name = tedge_config.device.id.try_read(tedge_config)?.to_string();
     let service_type = tedge_config.service.ty.clone();
 
+    // from this level we don't have access to the entity store and registered main device so best
+    // we can do for now is just guess it
+    // TODO: fix this, preferably use a HealthMonitorActor
+    let service_external_id = format!("{device_name}:device:main:service:tedge-mapper-c8y");
```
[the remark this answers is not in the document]
Bravo555: Fixed in 9b6c7a615dd3f7b85a1820a600e02bc60802f0f1.

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: Thanks for having updated this. The new version wraps the subtlety inside the `topic_for` function instead of exposing it to all uses of `target.to_string()`.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
Bravo555: Fixed in b0ef0923cf85647526043f916bf99b356ba95532.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -41,6 +46,10 @@ impl AsRef<str> for EntityExternalId {
     }
 }
 
+// XXX: As `EntityExternalId` is used as a part of cloudbound MQTT topic, it
+// can't contain characters invalid in topics, i.e. `+` and `#`. ([MQTT-4.7]).
```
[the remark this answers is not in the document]
didier-wenzek: > But, I get your confusion and may be the solution is that the `ExternalId` should have been just a trait and `C8yExternalId` and `AzureExternalId` structs could have implemented that type and define their own validation rules in their constructors or `From` impls.

This would be a nice follow-up task.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -322,6 +338,116 @@ impl EntityTopicId {
...
+    pub fn as_str(&self) -> &str {
+        self.0.as_str()
+    }
+
+    pub fn entity(&self) -> &EntityTopicId {
+        &self.0
+    }
+
+    /// If in a default MQTT scheme, returns a device topic id of this service.
+    pub fn to_device_topic_id(&self) -> Option<DeviceTopicId> {
```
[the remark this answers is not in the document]
Bravo555: This function was unused so it was removed. Fixed in 3862236def0954d4f3380f2825ec7b5ec94a47a6.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -322,6 +338,116 @@ impl EntityTopicId {
     pub fn is_default_main_device(&self) -> bool {
         self == &Self::default_main_device()
     }
+
+    /// If `self` is a device topic id, return a service topic id under this
+    /// device.
+    ///
+    /// The device topic id must be in a format: "device/DEVICE_NAME//"; if not,
+    /// `None` will be returned.
+    pub fn to_default_service_topic_id(&self, service_name: &str) -> Option<ServiceTopicId> {
```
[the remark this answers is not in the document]
Bravo555: Fixed in 3fbb47024a38c4cb0bf7ae83a3801e708f8abf2a.

crates/core/c8y_api/src/smartrest/inventory.rs
```diff
@@ -35,9 +48,34 @@ pub fn service_creation_message(
...
 }
+
+/// Create a SmartREST message for updating service status.
+///
+/// `service_status` can be any string, but `"up"`, `"down"`, and `"unknown"`
+/// have known meanings and are displayed in the UI in different ways.
+///
+/// https://cumulocity.com/guides/reference/smartrest-two/#104
+pub fn service_status_update_message(entity: &EntityMetadata, service_status: &str) -> Message {
+    let topic = super::topic::C8yTopic::smartrest_response_topic(entity)
```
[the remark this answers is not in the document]
didier-wenzek: @albinsuresh It starts to be urgent to work on https://github.com/thin-edge/thin-edge.io/issues/2316 to avoid such errors

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -401,7 +444,7 @@ impl EntityStore {
                     external_id: Some(service_external_id),
                     r#type: EntityType::Service,
                     parent: Some(parent_device_id),
-                    other: json!({ "name": service_id,  "type": "systemd" }),
+                    other: json!({ "name": service_id, "type": self.default_service_type }),
```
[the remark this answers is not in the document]
Bravo555: Thing is, if the entity store doesn't add this to the payload, then when the mapper is restarted and `service.type` was changed, then `serviceType` on all the services would change because the mapper would apply the default again.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -314,12 +347,22 @@ impl EntityStore {
...
         };
+
+        let JsonValue::Object(mut other) = message.other else {
+            return Err(Error::EntityRegistrationOtherNotMap);
+        };
+
+        // XXX: this is c8y-specific, entity store shouldn't do this!
+        other
+            .entry("type".to_string())
+            .or_insert(JsonValue::String(self.default_service_type.clone()));
```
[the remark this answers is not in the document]
Bravo555: if the entity store doesn't add this to the payload, then when the mapper is restarted and service.type was changed, then serviceType on all the services would change because the mapper would apply the default again.

crates/core/tedge_mapper/src/c8y/mapper.rs
didier-wenzek: > I didn't get the point of contention here. We're still passing the static `down` message here for the mapper service, with the mapper name, as it used to be, just with a different API. The new API doesn't differentiate between `up` or `down` or any status value for that matter. So, the replacement feels the same. Am I missing something subtle here?

My question is not about the up or down status, but about the smartrest message that is sent. Is the same message type used both to declare a new service and to notify a service status?

didier-wenzek: Are these changes the expected ones?
* New way to name a service:  `demo-device-777_tedge-mapper-c8y` -> `demo-device-777:device:main:service:tedge-mapper-c8y`
* New `104` messages on sub topics.

On start (new version):

```
[c8y/s/us] 102,demo-device-777:device:main:service:tedge-mapper-c8y,service,tedge-mapper-c8y,up
[c8y/s/us/demo-device-777:device:main:service:tedge-mapper-c8y] 104,up
```

When c8y-mapper exit (new version):
```
[c8y/s/us] 102,demo-device-777:device:main:service:tedge-mapper-c8y,service,tedge-mapper-c8y,down
```

On start (previous version):
```
[c8y/s/us] 102,demo-device-777_tedge-mapper-c8y,"service",tedge-mapper-c8y,"up"
```

On stop (previous version):
```
[c8y/s/us] 102,demo-device-777_tedge-mapper-c8y,"service",tedge-mapper-c8y,"down"
```

didier-wenzek: Changing service names is a bit annoying on c8y as of now I have two c8y-mapper on my console ;-)

But this is an expected drawback of the new naming scheme.

PradeepKiruvale: I verified that the `last will` message is getting translated to the `service` message and pushed to the c8y cloud by `tedge-mapper-c8y`.

```
[te/device/main/service/tedge-mapper-c8y/status/health] {"pid":404449,"status":"down"}
[c8y/s/us] 102,wsl-ubuntu:device:main:service:tedge-mapper-c8y,service,tedge-mapper-c8y,down

```

Bravo555: > Are these changes the expected ones?
> 
>     * New way to name a service:  `demo-device-777_tedge-mapper-c8y` -> `demo-device-777:device:main:service:tedge-mapper-c8y`

I followed the new C8y device and service topics described in [_Cumulocity Mapper_ documentation page](https://thin-edge.github.io/thin-edge.io/next/references/mappers/c8y-mapper/#child-device-service)



> New `104` messages on sub topics.

Yes, I changed to `104` because it doesn't need a `serviceType`, which simplifies things a bit.

crates/core/tedge_api/src/entity_store.rs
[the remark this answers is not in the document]
albinsuresh: No no, that's not what I meant. The mapper should be deriving the default `service.type` from the tedge_config value. Not apply a hardcoded value. I prefer to keep it with the mapper, because even for child devices, it's the mapper that applies the default: `thin-edge.io.child` if the registration message doesn't have one specified. Since the mapper is filling one default, it's only better if the mapper fills the other default as well. For this, we need to update the signature of `service_creation_message` to accept `Option<&str>` as `service_type`, just like the `child_device_creation_message` API that accepts `device_type: Option<&str>`.

crates/core/tedge_mapper/src/c8y/mapper.rs
didier-wenzek: > However, the behavior of `104` message is buggy.

This is due to the `tedge` to `te` converter. As I'm testing this PR on a machine where old health messages are retained on `tedge/health/+`, these messages are translated by `tedge_to_te_converter` on agent starts messing up the messages sent on `te/+/+/+/+/status/+` by the thin-edge daemons. Only the later ones are correct.

Since, this PR updates *all* thin-edge daemons to use `te/+/+/+/+/status/+` , the `tedge_to_te_converter`  must stop converting the former messages:


```
diff --git a/crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs b/crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
index 1ebb63e21f..d0753079ea 100644
--- a/crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
+++ b/crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
@@ -30,9 +30,6 @@ impl TedgetoTeConverter {
             }
             topic if topic.name.starts_with("tedge/events") => self.convert_event(message),
             topic if topic.name.starts_with("tedge/alarms") => self.convert_alarm(message),
-            topic if topic.name.starts_with("tedge/health") => {
-                self.convert_health_status_message(message)
-            }
             _ => vec![],
         }
     }
(
```

didier-wenzek (changes requested):
The `tedge_to_te_converter`  must stop converting `tedge/health/#` messages.

See https://github.com/thin-edge/thin-edge.io/pull/2276/files#r1362176975

crates/core/tedge_mapper/src/c8y/mapper.rs
Bravo555: > > However, the behavior of `104` message is buggy.
> 
> This is due to the `tedge` to `te` converter.

Indeed this behaviour is buggy and should be corrected. I will remove `tedge_to_te_converter` mapping of health status as part of this PR.

didier-wenzek: Thanks to @PradeepKiruvale and @Bravo555 for checking behavior and expectation on `102` and `104`.

Once `tedge_to_te_converter` fixed, the status are propagated as expected.

crates/core/c8y_api/src/smartrest/inventory.rs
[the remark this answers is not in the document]
albinsuresh: The problem is that `C8yTopic::smartrest_response_topic(entity)` will always create target topics like `c8y/s/us/<entity_external_id>` where the `entity_external_id` is just read from `entity: &EntityMetadata`. That's not enough for a nested child device or a nested child device service. For example, if you have a hierarchy as follows:
```
main -> immediate_child -> nested_child -> grand_nested_child
```

... the real publish topic for `grand_nested_child` is `c8y/s/us/immediate_child/nested_child/grand_nested_child` which is what `publish_topic_from_ancestors` would create. But, `C8yTopic::smartrest_response_topic(entity)` would only create `c8y/s/us/grand_nested_child` implying an immediate child device of `main` with the XID: `grand_nested_child` as per the SmartREST contract.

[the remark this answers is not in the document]
albinsuresh: > @albinsuresh It starts to be urgent to work on #2316 to avoid such errors

@didier-wenzek Yeah I agree. I'll pick it up as the very next task.

crates/core/c8y_api/src/smartrest/inventory.rs
```diff
@@ -35,9 +47,37 @@ pub fn service_creation_message(
...
+///
+/// `service_status` can be any string, but `"up"`, `"down"`, and `"unknown"`
+/// have known meanings and are displayed in the UI in different ways.
+///
+/// `external_ids` differs from what is returned by `ancestors_external_ids` in
+/// that it also contains the external ID of the current entity (the one we want
+/// to set the status of).
+///
+/// https://cumulocity.com/guides/reference/smartrest-two/#104
+pub fn service_status_update_message(external_ids: &[String], service_status: &str) -> Message {
```
albinsuresh, line 68: To avoid any confusion, as described in the comments above, and to keep this API consistent with the other APIs in this module like `service_creation_message`, I'd just take the `service_id` and `ancestors` as separate parameters, the latter without the target `service_id` in it. But while passing it to `publish_topic_from_ancestors`, you'll have to append them.

Bravo555: I'd like do do it properly later as part of a follow-up task.

Bravo555:
> The `tedge_to_te_converter` must stop converting `tedge/health/#` messages.

Should be fixed in 87ac7f8.

crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
```diff
@@ -30,9 +30,6 @@ impl TedgetoTeConverter {
             }
             topic if topic.name.starts_with("tedge/events") => self.convert_event(message),
             topic if topic.name.starts_with("tedge/alarms") => self.convert_alarm(message),
-            topic if topic.name.starts_with("tedge/health") => {
```
albinsuresh, line 33: I understand that this was removed to avoid a certain bug discussed in an earlier comment. But, I'm wondering about the impact of this change on backward compatibility. I was of the impression that these conversions exist to make the transition of third-party component written against the old APIs smoother on our next release where they'll continue publishing to the older topics and we'll internally map them with this `tedge_to_te_converter`. So, removing that conversion support for `tedge/health` will break all thirdy-part components publishing to the old health topics.

But, if `tedge/health` was only an internal API in the earlier releases, meant only to be used by our own components and not by third-party components, then this is okay. I'm just not sure if it was just an internal API.

tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
```diff
@@ -99,26 +99,6 @@ Check health status of tedge-mapper-c8y service on broker restart
 
     Custom Test Teardown
 
-Check health status of child device service
```
albinsuresh, line 102: Why was this test removed and not just adapted to use the new health endpoints? I understand that this test probably was relying on the `tedge_to_te_converter` earlier and that's why you removed. But, we could have just updated it to use the new `te/.../health/status` endpoints directly, unless this is duplicate and a test for child device service health status already exists elsewhere.

crates/core/tedge_agent/src/tedge_to_te_converter/converter.rs
didier-wenzek: That's true that we could only exclude a white-list of tedge daemons. However I don't think this API has been used by many as it was not documented beyond tedge daemons.

didier-wenzek (approved):

Follow-up tasks:
- Introduce a new constructor for ` HealthMonitorBuilder` from tedge-config to avoid repeating the same logic in all tedge daemons.
- Add a user-configurable service topic id to all tedge daemons, instead of building it after the device id.
- Consider to deprecate `ServiceTopicId`, `EntityTopicId` and `Service`, replacing their usage with `EntityMetadata` or intriducing an `Entity` type (see  https://github.com/thin-edge/thin-edge.io/pull/2276#discussion_r1341540315).
- Consider to deprecate this idea of a service type at the level of thin-edge `Entity`. This should be related to the [inventory](https://github.com/thin-edge/thin-edge.io/issues/2279) i.e. data attached by the user to devices and services.

albinsuresh (approved):
Approving, but still review my last comment on an existing test that was removed, which might reduce our test coverage. I'm just hoping that it was removed for genuine reasons.

Also, create a list of planned follow-up tasks in a comment, or even better as issues. Group trivial and related ones together and feel free to raise multiple issues, if some are going to be more time consuming than others, and unrelated to other changes.

tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
Bravo555: I may have made a mistake with that one. We agreed with @reubenmiller that health status messages should not change entity properties, so it follows that they shouldn't also create entities by themselves either.
However I forgot about auto-registration mechanism, in which it is expected that an entity sending any message will be automatically registered in C8y, so the same goes for health status messages.

I'll try to restore and update the test.

Bravo555:
> Also, create a list of planned follow-up tasks in a comment, or even better as issues. Group trivial and related ones together and feel free to raise multiple issues, if some are going to be more time consuming than others, and unrelated to other changes.

Created a list of follow-up tasks in this issue: https://github.com/thin-edge/thin-edge.io/issues/2274

didier-wenzek:
> Created a list of follow-up tasks in this issue: https://github.com/thin-edge/thin-edge.io/issues/2274

Thank you

Issue #2132, opened 2023-08-07
c8y-remote-access cannot start child-process if the bin is located in /usr/bin


toewsar:
**Describe the bug**
The c8y-remote-access-plugin ist called by the mapper according to the command in the `c8y_RemoteAccessConnect` file.
By default the file contains:
```toml
[exec]
command = "/usr/bin/c8y-remote-access-plugin"
topic = "c8y/s/ds"
on_message = "530"
``` 
But if the binary is not located in `/usr/bin` the execution fails, because the plugin tries to start the hardcoded child process `/usr/bin/c8y-remote-access-plugin`.

**To Reproduce**
Shift `/usr/bin/c8y-remote-access-plugin` to e.g. `~/.local/bin` and adopt `/etc/tedge/operations/c8y/c8y_RemoteAccessConnect`.
If you will now try to connect from c8y the `c8y-remote-access-plugin` will fail.

**Expected behavior**
The `c8y-remote-access-plugin` should use internaly the same path as used to call itself.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [e.g. 0.1.0]

**Additional context**

reubenmiller:
@toewsar Yes you're right. I didn't think this was the case, but I can confirm that path to the `c8y-remote-access-plugin` is NOT being read from `/etc/tedge/operations/c8y/c8y_RemoteAccessConnect`.

Here is the offending line: [c8y_remote_access_plugin/src/main.rs](https://github.com/thin-edge/thin-edge.io/blob/ab4e0787daf7344c0d7c99644b117b7085bc4d3e/plugins/c8y_remote_access_plugin/src/main.rs#L99)

```rust
let mut command = tokio::process::Command::new("/usr/bin/c8y-remote-access-plugin")
```

It should be an easy fix though.

didier-wenzek:
Addressed with https://github.com/thin-edge/thin-edge.io/pull/2277

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the bug and could not reproduce it anymore.

PR #2277, opened 2023-09-20
Run remote access plugin with correct executable path in child process

Ruadhri17:
## Proposed changes
This PR changes the executable path in child process from default one: `/usr/bin/c8y-remote-access-plugin` to one that was used to call itself. More in #2132.

## Types of changes

## Paste Link to the issue
* #2132 
<br/>

## Checklist

### DIFF
--- a/plugins/c8y_remote_access_plugin/src/main.rs
+++ b/plugins/c8y_remote_access_plugin/src/main.rs
@@ -1,3 +1,4 @@
+use std::path::PathBuf;
 use std::process::Stdio;
 
 use camino::Utf8Path;
@@ -6,6 +7,7 @@ use clap::Parser;
 use futures::future::try_select;
 use futures::future::Either;
 use input::parse_arguments;
+use miette::miette;
 use miette::Context;
 use miette::IntoDiagnostic;
 use tedge_config::TEdgeConfig;
@@ -14,6 +16,7 @@ use tedge_utils::file::create_directory_with_user_group;
 use tedge_utils::file::create_file_with_user_group;
 use tokio::io::AsyncBufReadExt;
 use tokio::io::BufReader;
+use toml::Table;
 use url::Url;
 
 use crate::auth::Jwt;
@@ -49,7 +52,7 @@ async fn main() -> miette::Result<()> {
             Ok(())
         }
         Command::Connect(command) => proxy(command, tedge_config).await,
-        Command::SpawnChild(command) => spawn_child(command).await,
+        Command::SpawnChild(command) => spawn_child(command, config_dir.tedge_config_root_path()).await,
     }
 }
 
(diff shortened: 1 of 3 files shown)


jarhodes314 (approved):

reubenmiller (changes requested):
Shouldn't we use the path which is set in the actual operation file? This would allow users to run their custom logic if necessary before calling the command.

For example, the file `/etc/tedge/operations/c8y/c8y_RemoteAccessConnect` includes a `exec.command` setting which should technically control which command should be executed when receiving the `530` message (below shows an example of this file):

```toml
/etc/tedge/operations/c8y # cat c8y_RemoteAccessConnect
[exec]
command = "/usr/bin/c8y-remote-access-plugin"
topic = "c8y/s/ds"
on_message = "530"
```

plugins/c8y_remote_access_plugin/src/main.rs
```diff
@@ -184,6 +189,25 @@ fn build_proxy_url(cumulocity_host: &str, key: &str) -> miette::Result<Url> {
         .context("Creating websocket URL")
 }
 
+fn get_executable_path(config_dir: &Utf8Path) -> miette::Result<PathBuf> {
+    let operation_path = supported_operation_path(config_dir);
+
+    let content = std::fs::read_to_string(&operation_path)
+        .into_diagnostic()
+        .context(format!(
```
jarhodes314, line 197:
```suggestion
        .with_context(|| format!(
```

This means that we construct the string lazily when an error occurs rather than every time we call this function.

plugins/c8y_remote_access_plugin/src/main.rs
```diff
@@ -184,6 +189,25 @@ fn build_proxy_url(cumulocity_host: &str, key: &str) -> miette::Result<Url> {
...
+            "The operation file {operation_path} does not exist or is not readable."
+        ))?;
+
+    let operation: Table = content
+        .parse()
+        .into_diagnostic()
+        .context(format!("Failed to parse {operation_path} file"))?;
```
jarhodes314, line 204:
```suggestion
        .with_context(|| format!("Failed to parse {operation_path} file"))?;
```

plugins/c8y_remote_access_plugin/src/main.rs
```diff
@@ -184,6 +189,25 @@ fn build_proxy_url(cumulocity_host: &str, key: &str) -> miette::Result<Url> {
...
+
+    Ok(PathBuf::from(operation["exec"]["command"].as_str().ok_or(
```
jarhodes314, line 206:
```suggestion
    Ok(PathBuf::from(operation["exec"]["command"].as_str().ok_or_else(||
```

plugins/c8y_remote_access_plugin/src/main.rs
```diff
@@ -184,6 +189,25 @@ fn build_proxy_url(cumulocity_host: &str, key: &str) -> miette::Result<Url> {
         .context("Creating websocket URL")
 }
 
+fn get_executable_path(config_dir: &Utf8Path) -> miette::Result<PathBuf> {
+    let operation_path = supported_operation_path(config_dir);
+
+    let content = std::fs::read_to_string(&operation_path)
```
jarhodes314, line 195:
```suggestion
    let content = tokio::fs::read_to_string(&operation_path)
        .await
```

jarhodes314 (changes requested):
A few minor comments

jarhodes314 (approved):

Issue #2279, opened 2023-09-21
New tedge twin (inventory) MQTT topic


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Support new tedge telemetry type called `twin` (in Cumulocity IoT this is called inventory managed object updates) to add json data to the device, child device or service's digital twin in the Cloud.

**Describe the solution you'd like**

New topic following the tedge MQTT V1 interface to allow users to publish additional information to the entity (e.g. device, child device or service).

**Topic name**

```
te/{identity}/twin/<type>
```

**Payload (json, retain=true)**

```json
{
    "family": "Debian",
    "version": 11
}
```

MQTT messages published to `te/{identity}/twin/<type>` should be published with the retained flag to allow other components to more easily observe the information.

**Cloud mapper initialization (to be validated)**

When the tedge-mapper starts up for a given cloud, it should run the following routine to process the initial values after any registration process has finished:

1. Read the inventory.json file, load the values into an in-memory model
2. Subscribe to the data topic using the MQTT topic pattern `{identifier}/twin/+` and merge the results. The values received via MQTT take precedence over any duplicated values from the inventory.json file
3. Publish the inventory changes to the relevant cloud in one message (if possible)
4. Process all subsequent messages received on the `{identifier}/twin/+` MQTT topic as single message to the relevant cloud

**Open questions**

* How to handle inventory.json? Previously the inventory.json file was used to publish static information to the cloud, however it was not that useful as it would only be read on startup and it also required users to have file access to the file (amongst other limitations).
  * Option 1: Deprecate the inventory.json file. Use a feature flag to deactivate/ignore the file.
  * Option 2: Use the inventory.json value as the persistence layer. Any messages published to the `/data/` topics will be merged into the inventory.json before publishing

**Describe alternatives you've considered**

**Additional context**

A [draft PR](https://github.com/thin-edge/thin-edge.io/pull/2280) has already been created to include the following items to assist with the development:

* [Documentation](https://github.com/reubenmiller/thin-edge.io/blob/feat-tedge-inventory-topic/docs/src/references/mappers/c8y-mapper.md#data) about tedge topic to c8y mapping for the new `/twin/` topic
[...]

PR #2280, opened 2023-09-22
feat(mqtt): add docs for tedge data/inventory topic

reubenmiller:
## Proposed changes

PR to communicate the proposed mapping of the `/twin/` topics including integration tests. This PR can be used for the actual implementation.

## Types of changes

## Paste Link to the issue
<br/>

https://github.com/thin-edge/thin-edge.io/issues/2279

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -102,7 +102,7 @@ use tracing::trace;
 const C8Y_CLOUD: &str = "c8y";
 const INVENTORY_FRAGMENTS_FILE_LOCATION: &str = "device/inventory.json";
 const SUPPORTED_OPERATIONS_DIRECTORY: &str = "operations";
-const INVENTORY_MANAGED_OBJECTS_TOPIC: &str = "c8y/inventory/managedObjects/update";
+pub const INVENTORY_MANAGED_OBJECTS_TOPIC: &str = "c8y/inventory/managedObjects/update";
 const INTERNAL_ALARMS_TOPIC: &str = "c8y-internal/alarms/";
 const C8Y_JSON_MQTT_EVENTS_TOPIC: &str = "c8y/event/events/create";
 const TEDGE_AGENT_LOG_DIR: &str = "tedge/agent";
@@ -893,6 +893,10 @@ impl CumulocityConverter {
         }
 
         let mut messages = match &channel {
+            Channel::EntityTwinData { fragment_key } => {
+                self.try_convert_entity_twin_data(&source, message, fragment_key)?
+            }
+
             Channel::Measurement { measurement_type } => {
                 self.try_convert_measurement(&source, message, measurement_type)?
             }
@@ -1451,7 +1455,7 @@ pub struct HealthStatus {
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use super::CumulocityConverter;
     use crate::actor::IdDownloadRequest;
     use crate::actor::IdDownloadResult;
@@ -2458,7 +2462,7 @@ mod tests {
         assert!(!second_registration_message_mapped);
     }
 
-    async fn create_c8y_converter(
+    pub(crate) async fn create_c8y_converter(
         tmp_dir: &TempTedgeDir,
     ) -> (
         CumulocityConverter,
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -2,8 +2,6 @@ use super::alarm_converter::AlarmConverter;
 use super::config::C8yMapperConfig;
 use super::config::MQTT_MESSAGE_SIZE_THRESHOLD;
 use super::error::CumulocityMapperError;
-use super::fragments::C8yAgentFragment;
-use super::fragments::C8yDeviceDataFragment;
 use super::service_monitor;
 use crate::actor::CmdId;
 use crate::actor::IdDownloadRequest;
@@ -55,8 +53,6 @@ use service_monitor::convert_health_status_message;
 use std::collections::HashMap;
 use std::collections::HashSet;
 use std::fs;
-use std::fs::File;
-use std::io::Read;
 use std::path::Path;
 use std::path::PathBuf;
 use tedge_actors::LoggingSender;
@@ -95,14 +91,11 @@ use thiserror::Error;
 use time::format_description::well_known::Rfc3339;
 use tokio::time::Duration;
 use tracing::debug;
-use tracing::info;
 use tracing::log::error;
 use tracing::trace;
 
 const C8Y_CLOUD: &str = "c8y";
-const INVENTORY_FRAGMENTS_FILE_LOCATION: &str = "device/inventory.json";
 const SUPPORTED_OPERATIONS_DIRECTORY: &str = "operations";
-pub const INVENTORY_MANAGED_OBJECTS_TOPIC: &str = "c8y/inventory/managedObjects/update";
 const INTERNAL_ALARMS_TOPIC: &str = "c8y-internal/alarms/";
 const C8Y_JSON_MQTT_EVENTS_TOPIC: &str = "c8y/event/events/create";
 const TEDGE_AGENT_LOG_DIR: &str = "tedge/agent";
@@ -171,7 +164,8 @@ pub struct CumulocityConverter {
     pub config: C8yMapperConfig,
     pub(crate) mapper_config: MapperConfig,
     pub device_name: String,
-    device_type: String,
+    pub(crate) device_topic_id: EntityTopicId,
+    pub(crate) device_type: String,
     alarm_converter: AlarmConverter,
     pub operations: Operations,
     operation_logs: OperationLogs,
@@ -187,6 +181,7 @@ pub struct CumulocityConverter {
     pub auth_proxy: ProxyUrlGenerator,
     pub downloader_sender: LoggingSender<IdDownloadRequest>,
     pub pending_operations: HashMap<CmdId, SmartRestOperationVariant>,
+    pub inventory_model: Value, // Holds a live view of aggregated inventory, derived from various twin data
 }
 
 impl CumulocityConverter {
@@ -198,6 +193,7 @@ impl CumulocityConverter {
         downloader_sender: LoggingSender<IdDownloadRequest>,
     ) -> Result<Self, CumulocityConverterBuildError> {
         let device_id = config.device_id.clone();
+        let device_topic_id = config.device_topic_id.clone();
         let device_type = config.device_type.clone();
         let service_type = config.service_type.clone();
         let c8y_host = config.c8y_host.clone();
@@ -230,11 +226,17 @@ impl CumulocityConverter {
         )
         .unwrap();
 
+        let inventory_model = json!({
+            "name": device_id.clone(),
+            "type": device_type.clone(),
+        });
+
         Ok(CumulocityConverter {
             size_threshold,
             config,
             mapper_config,
             device_name: device_id,
+            device_topic_id,
             device_type,
             alarm_converter,
             operations,
@@ -251,6 +253,7 @@ impl CumulocityConverter {
             auth_proxy,
             downloader_sender,
             pending_operations: HashMap::new(),
+            inventory_model,
         })
     }
 
@@ -1064,31 +1067,24 @@ impl CumulocityConverter {
     }
 
     fn try_init_messages(&mut self) -> Result<Vec<Message>, ConversionError> {
-        let inventory_fragments_message = self.wrap_error(create_inventory_fragments_message(
-            &self.device_name,
-            &self.cfg_dir,
-        ));
+        let mut messages = self.parse_base_inventory_file()?;
 
-        let supported_operations_message = self.wrap_error(
-            self.create_supported_operations(&self.cfg_dir.join("operations").join("c8y")),
-        );
+        let supported_operations_message =
+            self.create_supported_operations(&self.cfg_dir.join("operations").join("c8y"))?;
+
+        let device_data_message = self.inventory_device_type_update_message()?;
+
+        let pending_operations_message = create_get_pending_operations_message()?;
 
         let cloud_child_devices_message = create_request_for_cloud_child_devices();
 
-        let device_data_message = self.wrap_error(create_device_data_fragments(
-            &self.device_name,
-            &self.device_type,
-        ));
-
-        let pending_operations_message = self.wrap_error(create_get_pending_operations_message());
-
-        Ok(vec![
-            inventory_fragments_message,
+        messages.append(&mut vec![
             supported_operations_message,
             device_data_message,
             pending_operations_message,
             cloud_child_devices_message,
-        ])
+        ]);
+        Ok(messages)
     }
 
     fn create_supported_operations(&self, path: &Path) -> Result<Message, ConversionError> {
@@ -1163,17 +1159,6 @@ fn get_child_id(dir_path: &PathBuf) -> Result<String, ConversionError> {
     }
 }
 
-fn create_device_data_fragments(
-    device_name: &str,
-    device_type: &str,
-) -> Result<Message, ConversionError> {
-    let device_data = C8yDeviceDataFragment::from_type(device_type)?;
-    let ops_msg = device_data.to_json()?;
-
-    let topic = Topic::new_unchecked(&format!("{INVENTORY_MANAGED_OBJECTS_TOPIC}/{device_name}",));
-    Ok(Message::new(&topic, ops_msg.to_string()))
-}
-
 fn create_get_software_list_message() -> Result<Message, ConversionError> {
     let request = SoftwareListRequest::default();
     let topic = Topic::new(RequestTopic::SoftwareListRequest.as_str())?;
@@ -1204,17 +1189,6 @@ fn is_child_operation_path(path: &Path) -> bool {
 
 fn create_request_for_cloud_child_devices() -> Message {
     Message::new(&Topic::new_unchecked("c8y/s/us"), "105")
-}
-
-fn create_inventory_fragments_message(
-    device_name: &str,
-    cfg_dir: &Path,
-) -> Result<Message, ConversionError> {
-    let inventory_file_path = format!("{}/{INVENTORY_FRAGMENTS_FILE_LOCATION}", cfg_dir.display());
-    let ops_msg = get_inventory_fragments(&inventory_file_path)?;
-
-    let topic = Topic::new_unchecked(&format!("{INVENTORY_MANAGED_OBJECTS_TOPIC}/{device_name}"));
-    Ok(Message::new(&topic, ops_msg.to_string()))
 }
 
 impl CumulocityConverter {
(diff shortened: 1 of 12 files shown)


docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,305 @@ c8y/alarm/alarms/create
...
+</div>
+
+#### Data - Service on Main Device
+
+<div class="code-indent-left">
+
+**Thin Edge (input)**
+
+```text title="Topic (retain=true)"
+te/device/main/service/tedge-agent/data/runtime_stats
```
albinsuresh, line 712: I understand that this is just an example, but realistically for things like `runtime_stats`, isn't it better to model them as measurements of that service, as it fairly dynamic in nature?

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,305 @@ c8y/alarm/alarms/create
...
 
+### Data
+
+The data telemetry type can be used to store additional information about the device/child/service. Such information could included; operation system name/version, communication statistics, device status, or any other information that is not suited to measurements, events or alarms.
+
+#### Data - Main device
+
+A device's digital twin model can be updated by publishing to a specific topic.
+
+The type part of the topic is used to group the data so it is easier for components to subscribe to relevant parts.
```
albinsuresh, line 579: Do we need to restrict these to a single level? It definitely makes sense to start with that, but technically, we can support as many sub-levels as needed and for the mapping we can map `data/a/b/c` to the following payload structure, if the cloud only supports a single level:

```json
"a": {
    "b": {
        "c": { // payload... }
}
```

But, that should be a mapper restriction and thin-edge can be more open.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,305 @@ c8y/alarm/alarms/create
...
+
+**Cumulocity IoT (output)**
+
+```text title="Topic"
+c8y/inventory/managedObjects/update/<main-device-id>
+```
+
+```json5 title="Payload"
+{
+  "device_OS": {
```
albinsuresh, line 608: Shouldn't we do some mapping of "special keys" in the payload as well? For example, for something like hardware info: a clous-agnostic user might use something like:

```json
"hardware": {
  "model": "BCM2708",
  "revision": "000e",
  "serialNumber": "00000000e2f5ad4d"
}
```

which needs to be mapped the following `c8y_Hardware` fragment, which Cumulocity understands:

```json
"c8y_Hardware": {
  "model": "BCM2708",
  "revision": "000e",
  "serialNumber": "00000000e2f5ad4d"
}
```

We can definitely keep "special payload key mapping" out of scope initially and then add it later if there is interest. In the mean time, users of C8y are free to send the `c8y_Hardware` fragment directly.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
```diff
@@ -70,6 +70,27 @@ Child devices support sending inventory data via c8y topic
...
+    Should Be Equal As Integers    ${mo["device_OS"]["complex"][2]}    3
+    Should Be Equal    ${mo["device_OS"]["object"]["foo"]}    bar
+
+
+Child devices supports sending inventory data via tedge topic to root fragments
+    Execute Command    tedge mqtt pub "te/device/${CHILD_SN}///data/" '{"subtype":"LinuxDeviceA","type":"ShouldBeIgnored"}'
+    Cumulocity.Set Device    ${CHILD_SN}
+    ${mo}=    Device Should Have Fragments    subtype
+    Should Be Equal    ${mo["subtype"]}    LinuxDeviceA
+    Should Be Equal    ${mo["type"]}    thin-edge.io
```
didier-wenzek, line 91: Nice to have tests along the specs ♥

This one makes me understand why a `type` fragment is an issue as already defined by `tedge config get device.type`. But, don't we have to address with a broader view these conflicts with legacy features (`inventory.json` and `tedge config`)?

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,305 @@ c8y/alarm/alarms/create
...
+A device's digital twin model can be updated by publishing to a specific topic.
+
+The type part of the topic is used to group the data so it is easier for components to subscribe to relevant parts.
+
+<div class="code-indent-left">
+
+**Thin Edge (input)**
+
+```text title="Topic (retain=true)"
+te/device/main///data/device_OS
```
didier-wenzek, line 586: If *two* messages are sent for the *same* device, say `te/device/main///data/a` and `te/device/main///data/a`,
then which component is responsible of the combination of these two fragments?

In other words:
- does the mapper have to send a *single*  `{"a": ...,"b": ...}` message to `c8y/inventory/managedObjects/update/<main-device-id>` ?
- or can the mapper send a first `{"a": ...}` message then a second `{"b": ...}` message to C8y?

didier-wenzek (commented):
It would be really nice to use this PR for pushing the implementation along docs and tests.

I see two points that really needs to be clarified:
- How this feature is combined with the inventory fragments extracted from `tedge-config` and `inventory.json`
- The local consolidation of all the fragments. It's a good idea to persist this aggregated view, but I would really avoid to use the former `inventory.json` file for that. We need to be clear what is the source of truth: the persisted view of the inventory must *only* be edited by the mapper after data collected from the config and over MQTT.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,305 @@ c8y/alarm/alarms/create
...
+
+
+#### Data - root
+
+Data can be added on the root level by leaving the type value in topic blank but still using a trailing slash.
+
+Some property names are forbidden and will be ignored if they are included in the payload. This is to avoid setting the property values from two different sources.
+
+* `name`
+* `type`
```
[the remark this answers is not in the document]
albinsuresh: `name` and `type` are already included in the registration message, as these are the kind of metadata that is almost never going to change for a device. All such static data can be sent in the registration message itself.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,305 @@ c8y/alarm/alarms/create
...
+  }
+}
+```
+
+</div>
+
+
+#### Data - root
+
+Data can be added on the root level by leaving the type value in topic blank but still using a trailing slash.
```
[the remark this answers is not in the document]
albinsuresh: > * allow any JSON value (not only object) as  `tedge mqtt pub -r te/device/main///data/sub-type my-custom-type` .

I like that better. This also helps users avoid the error of overwriting all the existing data at that empty topic by publishing something new.

docs/src/references/mappers/c8y-mapper.md
albinsuresh: Fortunately the mapper can send different fragments one after the other and they are aggregated at the cloud.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
albinsuresh: The `inventory.json` contents can be parsed and any direct key-value pairs can be published to the `meta/key` topic, with the `value` as the payload. The same can be extended to `tedge config` as well, where the `value` would be just be the plain value instead of a JSON object, as you suggested.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,305 @@ c8y/alarm/alarms/create
 
 </div>
 
+### Data
+
+The data telemetry type can be used to store additional information about the device/child/service. Such information could included; operation system name/version, communication statistics, device status, or any other information that is not suited to measurements, events or alarms.
```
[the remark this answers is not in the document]
reubenmiller: Using `meta` will cause too much confusion with the `/meta` suffix used on other topics. The `/meta` suffix part is there to describe the data being published to this topic. So having `<identifier>/meta/example/meta` is too confusing.

Plus there is nothing that dictates that only 'meta' information can be stored on it. It is just a generic schema-less data structure.

So the options I see are:

1. `i` - short for information or inventory
2. `data` - (device) data
3. `dd` - device data (same as above just shorter)
4. `state` - device state information
5. `twin` - digital twin data

docs/src/references/mappers/c8y-mapper.md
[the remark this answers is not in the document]
reubenmiller: Ok, we can start with forbidding publishing without a type for now. We can re-evaluate this in the future. It is not ideal as publishing multiple non-object values to the root level requires doing multiple mqtt publishes, however not a deal breaker, and it does avoid the problems introduced by potentially publishing the same data to multiple topics (e.g. to `data/` with .type in the payload and to `/data/type`).

docs/src/references/mappers/c8y-mapper.md
reubenmiller: It depends on the use-case. In Cumulocity IoT publishing some runtime statistics (at long intervals) is more useful as you can then query the devices based on these statistics. For example if you had a runtime statistics which sent the connection stability (as a percentage), then you can create a smart group in Cumulocity for devices which have a higher connection stability to then send the software updates to only these devices (or creates groups based on quota)...Currently Cumulocity IoT devices does not support filtering devices based on an event/measurement or alarm (except for the newish "last known value" feature which just converts the measurements to an inventory managed object fragment).

TLDR; Some use cases are:

* Allow filtering devices based on the published statistics
* Data is not subject to retention rules (otherwise you would loose the data after the retention period)

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
reubenmiller: We could go the layered approach and first parse the inventory.json file and use that as the base model, and then apply any changes from the MQTT topics. The inventory.json is good for being able to hardcode static information into a build, however we could also just phase the file out (behind a feature flag or something).

docs/src/references/mappers/c8y-mapper.md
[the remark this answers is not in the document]
albinsuresh: Agree with your point on `/meta`. Out of the options, I prefer `twin`, `i` and `state` in that order, with `state` being a bit vague. I'm still against `data` for the reasons expressed above and `dd` makes it even more device specific, which doesn't cover services. Lemme just throw a 6th option as well. How about `desc` for entity descriptor, covering both devices and services? Or `ed` if we prefer short names.

Anyway, I'll let you have the final word on this UX matter. Just choose one out of these and we'll stick with that.

docs/src/references/mappers/c8y-mapper.md
reubenmiller: > Fortunately the mapper can send different fragments one after the other and they are aggregated at the cloud. But, if some of the fields within a fragment needs to be updated, then the mapper needs to send the whole fragment with old values as well as new values. That's acceptable IMO.

We can optimize later to group the updates into one message

docs/src/references/mappers/c8y-mapper.md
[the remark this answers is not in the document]
reubenmiller: Ok, let's go with `twin` then.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
albinsuresh: @reubenmiller So, no need to re-publish the contents of the file as `twin` messages to the broker? Without that, the other components relying on the broker for the inventory information will never get that "base model", right? Unless they all parse that file.

didier-wenzek (commented):
I still have to review `inventory.rs`. So far things look really nice.

I'm missing doc comments in  `inventory.rs` though. This makes things a bit harder to understand/check as one has to compare each implementation with its uses.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1060,31 +1067,26 @@ impl CumulocityConverter {
...
+        messages.push(cloud_child_devices_message);
 
-        Ok(vec![
-            inventory_fragments_message,
-            supported_operations_message,
-            device_data_message,
-            pending_operations_message,
-            cloud_child_devices_message,
-        ])
+        Ok(messages)
```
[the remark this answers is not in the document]
albinsuresh: Yeah, I was forced into this change because  `inventory_fragments_message` is not a `Message` anymore but a `Vec<Message>`. To restore the old readability, I'll keep the rest of the messages in this `vec![]` itself and then merge it with `inventory_fragments_messages` right before returning.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -187,6 +181,7 @@ pub struct CumulocityConverter {
     pub auth_proxy: ProxyUrlGenerator,
     pub downloader_sender: LoggingSender<IdDownloadRequest>,
     pub pending_operations: HashMap<CmdId, SmartRestOperationVariant>,
+    pub inventory_model: Value,
```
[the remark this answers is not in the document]
albinsuresh: Sure. This `inventory_model` is a live copy of the aggregated inventory that the mapper builds by aggregating the twin data from all the `twin/xyz` endpoint. We're currently using it to avoid mapping duplicate fragment update messages. In future, this will be persisted on the disk as well, and will even be synced with the cloud on startup.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,301 @@ c8y/alarm/alarms/create
 
 </div>
 
+### Twin
```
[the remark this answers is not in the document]
albinsuresh: Yeah, there's just a one-liner in the `mqtt-api.md` that just mentions this new channel. I'll move some of this content there.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,247 @@ c8y/alarm/alarms/create
...
+  "subtype": "my-custom-type"
+}
+```
+
+</div>
+
+:::warning
+Updating the following properties via the `twin` channel is not supported
+
+* `name`
```
didier-wenzek, line 657: Why is the `name` property forbidden? Looking at the code, `c8y_Agent` is the fragment that must be protected from overriding.

crates/extensions/c8y_mapper_ext/src/inventory.rs
```diff
@@ -0,0 +1,378 @@
...
+    ) -> Result<JsonValue, ConversionError> {
+        let agent_fragment = C8yAgentFragment::new()?;
+        let json_fragment = agent_fragment.to_json()?;
+
+        match Self::read_json_from_file(inventory_file_path) {
+            Ok(mut json) => {
+                json.as_object_mut()
+                    .ok_or(ConversionError::FromOptionError)?
+                    .insert(
+                        "c8y_Agent".to_string(),
```
didier-wenzek, line 141: Why are `c8y_Agent` and `type` treated differently?

- The former is inserted into `self.inventory_model`, the latter is not.
- The former is published over a `te` twin topic, the latter is not.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -568,6 +568,247 @@ c8y/alarm/alarms/create
 
 </div>
 
+### Twin
+
+The `twin` metadata is mapped to [inventory data](https://cumulocity.com/guides/concepts/domain-model/#inventory) in Cumulocity.
```
didier-wenzek, line 573: Two key points are missing:

- support for static fragments with the `inventory.json` file.
- automatic support for `c8y_Agent` and `type` fragments.

crates/extensions/c8y_mapper_ext/src/inventory.rs
```diff
@@ -0,0 +1,378 @@
...
+        let inventory_base = Self::get_inventory_fragments(&inventory_file_path)?;
+        self.inventory_model = inventory_base.clone();
+
+        let message =
+            self.inventory_update_message(&self.device_topic_id, inventory_base.clone())?;
+        messages.push(message);
+
+        if let JsonValue::Object(map) = inventory_base {
+            for (key, value) in map {
+                self.update_inventory_model(key.clone(), value.clone());
```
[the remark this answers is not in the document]
albinsuresh: Actually that `self.inventory_model = inventory_base.clone();` was a mistake as the `self.inventory_model` is already initialized with the name and fragment in the constructor of `CumulocityConverter`. So, here we should have only appended further key-value pairs and not replace it with that assignemnt. This is why I moved those updates into the loop, but forgot to remove the old statement.

crates/extensions/c8y_mapper_ext/src/inventory.rs
```diff
@@ -0,0 +1,378 @@
...
+                );
+                messages.push(mapped_message);
+            }
+        }
+
+        Ok(messages)
+    }
+
+    /// Create an entity twin data message with the provided fragment
+    pub(crate) fn register_twin_data_message(
```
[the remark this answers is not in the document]
albinsuresh: I'll review them once again. Side effects of a transition phase where some of these functions were still in the main `converter` module.

docs/src/references/mappers/c8y-mapper.md
albinsuresh: The `name` and `type` fields are forbidden here because they're already part of the entity registration message. If someone wants to update these, they must send another registration message with the updated values. Allowing them to update the same data via both these endpoints will might be confusing especially in the broker as we may have different `name` and `type` in the device registration topic as well as in the `twin/...` topics.

crates/extensions/c8y_mapper_ext/src/inventory.rs
albinsuresh: > * The former is inserted into `self.inventory_model`, the latter is not

The `name` and `type` are inserted into the inventory in the `CumulocityConverter` constructor. It was just accidentally getting lost with this buggy line: `self.inventory_model = inventory_base.clone();` which I'll revert.

> The former is published over a te twin topic, the latter is not.

`type` and `name` are not published for the same reason why updating them via `twin/` topics are not supported. They must only be present in the entity registration message and we don't want another copy of it at `twin/name` and `twin/type` topics.

docs/src/references/mappers/c8y-mapper.md
didier-wenzek: > The `name` and `type` fields are forbidden here because they're already part of the entity registration message. If someone wants to update these, they must send another registration message with the updated values. Allowing them to update the same data via both these endpoints will might be confusing especially in the broker as we may have different `name` and `type` in the device registration topic as well as in the `twin/...` topics.

Okay

> I didn't really understand why you think `c8y_Agent` must not be overridden?

What is a user publish its own fragment on `te/device/main///twin/c8y_Agent` ?

crates/extensions/c8y_mapper_ext/src/inventory.rs
didier-wenzek: Aren't `name` and `c8y_Agent.name` the same thing? I believe my confusion is coming from that.

albinsuresh: No, `name` here is the device/service name and `c8y_Agent.name` is the name of the agent. When thin-edge connects to C8y, the `c8y_Agent.name` is `thin-edge.io`.

docs/src/references/mappers/c8y-mapper.md
albinsuresh: > What if a user publish its own fragment on te/device/main///twin/c8y_Agent

I don't really have a use-case for someone wanting to update that fragment for the thin-edge device, as we set that. But I don't have any strong arguments to close that option either. So, will need @reubenmiller 's advise on this. For child devices, we probably wouldn't want to prevent them from setting it. For services, this fragment probably doesn't even make any sense.

didier-wenzek: > > What if a user publish its own fragment on te/device/main///twin/c8y_Agent

> I don't really have a use-case for someone wanting to update that fragment for the thin-edge device, as we set that. But I don't have any strong arguments to close that option either. So, will need @reubenmiller 's advise on this. For child devices, we probably wouldn't want to prevent them from setting it. For services, this fragment probably doesn't even make any sense.

Indeed, this would make no sense, but could be done by a user who fails to notice that this is *already* done by thin-edge as for the `type`. I that case there will be inconsistencies.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -808,6 +808,112 @@ c8y/inventory/managedObjects/update/<main-device-id>:device:child01:service:tedg
...
+  },
+  "c8y_Hardware": {
+    "model": "BCM2708",
+    "revision": "000e",
+    "serialNumber": "00000000e2f5ad4d"
+  }
+}
+```
+
+Where `c8y_Agent` fragment auto-generated by thin-edge is appended to the contents of the file before it is published.
```
didier-wenzek, line 858:
```suggestion
Where `c8y_Agent` fragment is auto-generated by thin-edge and appended to the contents of the file before it is published.
```

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -808,6 +808,112 @@ c8y/inventory/managedObjects/update/<main-device-id>:device:child01:service:tedg
...
+
+The fragments in this file are also published to the `te/device/main///twin/<fragment-key>` topics so that
+the local twin metadata is also up-to-date and other components can also consume it.
+For example, the above `inventory.json` would result in the following `twin` messages:
```
didier-wenzek, line 862: I would also add the `c8y_Agent` message.

didier-wenzek (approved):
Approved. Nice feature and implementation!

albinsuresh:
### Test Plan

The new `twin` topics and their mapping contracts are documented in the [mqtt-api.md](https://github.com/thin-edge/thin-edge.io/pull/2280/files#diff-af802a095f72bbf84e172455307666c297802f0376506349420c5c2576aeaa88) and [c8y-mapper.md](https://github.com/thin-edge/thin-edge.io/pull/2280/files#diff-9695a2a09b64f1cfa0ed7f9e9ff632abdc04a7fd3dc4c3849cf5b53a461e4add) docs included in this PR. The basic feature is already covered in the system tests updated in the same PR: [child_device_telemetry.robot](https://github.com/thin-edge/thin-edge.io/pull/2280/files#diff-f462db0457e20ba623dcfd08a09e3533ff6aefad83a039b7c0068bc1d5316f58) and [thin-edge_device_telemetry.robot](https://github.com/thin-edge/thin-edge.io/pull/2280/files#diff-605becc475927687bc7481998a8fd5762373321abf9e8344d2f253f4c20621ea). Tests for clearing inventory fragments will be added in a follow-up PR soon, as some `Cumulocity` library updates are required for it.

gligorisaev:

- tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
- tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot
- tests/RobotFramework/tests/tedge/call_tedge_config_list.robot

PR #2291, opened 2023-09-26
Accept any type as the key of the downloader actor

rina23q:
## Proposed changes
This change is required for c8y-mapper's enhancement to download a file per an operation. For example, if the key is only string, it's difficult to distinguish which operation (software, firmware, config_update) fires the download request.

So, this kind of example usage will be possible with this change, sending `DownloadRequest`:
```rust
struct DownloadKey {
    op_type: OperationType,
    id: String
}

self.message_box
    .download_sender
    .send((DownloadKey{ op_type: OperationType::ConfigUpdate, id: "abcd1234".into() }, download_request))
    .await?;
```

## Types of changes

## Paste Link to the issue
#2292

## Checklist

### DIFF
--- a/crates/extensions/tedge_downloader_ext/src/actor.rs
+++ b/crates/extensions/tedge_downloader_ext/src/actor.rs
@@ -6,6 +6,7 @@ use download::Downloader;
 use log::info;
 use std::path::Path;
 use std::path::PathBuf;
+use tedge_actors::Message;
 use tedge_actors::Sequential;
 use tedge_actors::Server;
 use tedge_actors::ServerActorBuilder;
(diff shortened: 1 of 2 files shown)


didier-wenzek (approved):
Approved

albinsuresh:
Although I've got nothing against making the Downloader accept generic types as request keys, I fail to get the benefits from the mapper side.

The mapper already had to send a unique `operation_id` to trigger a download from this actor and hence the mapper already would have known what kind of `operation_type` it generated that unique id for. The mapper was already caching that ID in the `reqs_pending_download` map and it could have just added the `OperationType` info as well into this map. Instead, here we are sending that info to the downloader, which does nothing with that additional info, essentially using the downloader as that cache, whereas we could have simply extended the `reqs_pending_download` map for the same. Or did I miss something else here?

Issue #2292, opened 2023-09-26
Update c8y-mapper to handle config commands


rina23q:
Subtask for #2098 

Todo:
- [x] Update `tedge_downloader_ext` to accept any type of key for the request/result. #2291
- [x] `tedge-agent` creates `/var/tedge/cache` directory. #2293 
- [x] `tedge-mapper-c8y` handles `config_update` and `config_snapshot` commands.#2303

rina23q:
Verification must be done for #2098 together with all other components, i.e. tedge-mapper-c8y, tedge-configuration-plugin, doc, and robot tests. Therefore I close this sub-task.

PR #2293, opened 2023-09-26
tedge-agent creates /var/tedge/cache directory on start-up

rina23q:
## Proposed changes
This directory was introduced by c8y-firmware-plugin, however, now cloud mappers need to download files from remote. The cache directory must be used also by cloud mappers to avoid multiple file downloads for the same file.

The directory usage will be covered by robot tests during #2292 work.

In addition, to solve this comment https://github.com/thin-edge/thin-edge.io/pull/2293#discussion_r1338429388, I introduced `DataDir`(crates/core/tedge_api/src/path.rs) struct, which is a wrapper of `Utf8PathBuf` with refactoring codes.

## Types of changes

## Paste Link to the issue
#2292 

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/lib.rs
+++ b/crates/core/tedge_api/src/lib.rs
@@ -31,9 +31,6 @@ pub use messages::SoftwareUpdateRequest;
 pub use messages::SoftwareUpdateResponse;
 pub use software::*;
 
-pub const DEFAULT_DATA_PATH: &str = "/var/tedge";
-pub const DEFAULT_FILE_TRANSFER_DIR_NAME: &str = "file-transfer";
-
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/crates/core/tedge_api/src/lib.rs
+++ b/crates/core/tedge_api/src/lib.rs
@@ -10,6 +10,7 @@ pub mod measurement;
 pub mod messages;
 pub mod mqtt_topics;
 pub mod parser;
+pub mod path;
 pub mod serialize;
 mod software;
 pub mod topic;
--- a/crates/core/tedge_api/src/path.rs
+++ b/crates/core/tedge_api/src/path.rs
@@ -6,7 +6,7 @@ use std::path::Path;
 /// Default is /var/tedge.
 /// All directories under the root data directory must be listed here.
 #[derive(Debug, Clone)]
-pub struct DataDir(pub Utf8PathBuf);
+pub struct DataDir(Utf8PathBuf);
 
 impl Default for DataDir {
     fn default() -> Self {
@@ -17,6 +17,12 @@ impl Default for DataDir {
 impl From<Utf8PathBuf> for DataDir {
     fn from(path: Utf8PathBuf) -> Self {
         DataDir(path)
+    }
+}
+
+impl From<DataDir> for Utf8PathBuf {
+    fn from(value: DataDir) -> Self {
+        value.0
     }
 }
 
@@ -44,11 +50,6 @@ impl DataDir {
     /// ```
     pub fn join(&self, path: impl AsRef<Utf8Path>) -> Utf8PathBuf {
         self.0.join(path.as_ref().to_path_buf())
-    }
-
-    /// Return `Utf8PathBuf` of the given `DataDir`.
-    pub fn utf8_path_buf(&self) -> Utf8PathBuf {
-        self.0.clone()
     }
 
     /// Return `Utf8PathBuf` to ThinEdge file transfer repository.
--- a/crates/core/tedge_api/src/path.rs
+++ b/crates/core/tedge_api/src/path.rs
@@ -1,0 +1,95 @@
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+use std::path::Path;
+
+/// Representation of ThinEdge data directory.
+/// Default is /var/tedge.
+/// All directories under the root data directory must be listed here.
+#[derive(Debug, Clone)]
+pub struct DataDir(pub Utf8PathBuf);
+
+impl Default for DataDir {
+    fn default() -> Self {
+        DataDir("/var/tedge".into())
+    }
+}
+
+impl From<Utf8PathBuf> for DataDir {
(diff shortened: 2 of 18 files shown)


crates/core/tedge_api/src/lib.rs
```diff
@@ -32,6 +32,7 @@ pub use software::*;
 
 pub const DEFAULT_DATA_PATH: &str = "/var/tedge";
 pub const DEFAULT_FILE_TRANSFER_DIR_NAME: &str = "file-transfer";
+pub const DEFAULT_FILE_CACHE_DIR_NAME: &str = "cache";
```
didier-wenzek, line 35: I would really avoid to add yet another `pub const`.  

These constants are too low level. Look at the current usages for `DEFAULT_FILE_TRANSFER_DIR_NAME`. All of these are starting with a `data_dir.join()`. i.e. doing the same as `file_transfer_dir_as_pathbuf()` but with some possibly buggy `data_dir`.

What we need are methods providing paths ready to be used.

didier-wenzek (commented):
I think it's better to go a step further by deprecating the 3 `pub const` related to file transfer (namely `DEFAULT_DATA_PATH`, `DEFAULT_FILE_TRANSFER_DIR_NAME` and `DEFAULT_FILE_CACHE_DIR_NAME`).

crates/core/tedge_api/src/path.rs
```diff
@@ -0,0 +1,95 @@
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+use std::path::Path;
+
+/// Representation of ThinEdge data directory.
+/// Default is /var/tedge.
+/// All directories under the root data directory must be listed here.
+#[derive(Debug, Clone)]
+pub struct DataDir(pub Utf8PathBuf);
```
didier-wenzek, line 9: Do we need `self.0` to be `pub`?

didier-wenzek (approved):
Approved. Nice improvement!

crates/extensions/c8y_config_manager/src/config.rs
```diff
@@ -49,7 +49,7 @@ impl ConfigManagerConfig {
         let plugin_config_dir = config_dir.join(DEFAULT_OPERATION_DIR_NAME);
         let plugin_config_path = plugin_config_dir.join(DEFAULT_PLUGIN_CONFIG_FILE_NAME);
 
-        let file_transfer_dir = data_dir.join(DEFAULT_FILE_TRANSFER_DIR_NAME);
+        let file_transfer_dir = data_dir.file_transfer_dir().as_std_path().to_path_buf();
```
albinsuresh, line 52:
```suggestion
        let file_transfer_dir = data_dir.file_transfer_dir().as_std_path().to_path_buf();
```
```suggestion
        let file_transfer_dir = data_dir.file_transfer_dir().into();
```

crates/extensions/c8y_firmware_manager/src/actor.rs
```diff
@@ -253,13 +253,13 @@ impl FirmwareManagerActor {
                 .is_some()
             {
                 if let Ok(token) = self.message_box.jwt_retriever.await_response(()).await? {
-                    DownloadRequest::new(firmware_url, &cache_file_path)
+                    DownloadRequest::new(firmware_url, cache_file_path.as_std_path())
```
albinsuresh, line 256:
```suggestion
                    DownloadRequest::new(firmware_url, cache_file_path.as_ref())
```
Not a request for change but more of an FYI. But, I also know that you prefer explicit conversion functions over the generic ones like `as_ref()` or `into()` 😄

albinsuresh (commented):
The core change looks fine. But, as a result of this introduction of the `DataDir` type, with helper methods that returns `Utf8PathBuf` values for things like `file_transfer_repo_path`, you ended changing a lot of code which was accepting the standard `&Path` or `PathBuf` fields to accept `impl AsRef<Path>` or directly `Utf8PathBuf`. You probably didn't need to change all of that if `DataDir` had all the necessary conversion impls like `impl From<DataDir> for PathBuf`,` impl From<DataDir> for Utf8PathBuf` etc. Seemless conversion of a DataDir into all the standard path types should have been possible with these `From` impls and the `AsRef<Path>` impl that you already had.

That being said, using `Utf8PathBuf` and `Utf8Path` across the board is not such a bad idea either. I'm just saying that you probably didn't **have to** make all these changes just because `DataDir` returns `Utf8PathBuf`s.

rina23q:
> That being said, using Utf8PathBuf and Utf8Path across the board is not such a bad idea either.

Since `tedge_config` API returns `Utf8PathBuf`, using `PathBuf` is a kind of downgrade in most of cases (only config_path from command line option is different, I'm not sure if clap can accept `Utf8Pathbuf`.)

So, I even wish I could change all usage of `PathBuf` to `Utf8PathBuf` in several crates.

crates/core/tedge_api/src/path.rs
rina23q: I'll add it as @albinsuresh suggested.

crates/extensions/c8y_firmware_manager/src/config.rs
```diff
@@ -19,11 +20,11 @@ const FIRMWARE_UPDATE_RESPONSE_TOPICS: &str = "tedge/+/commands/res/firmware_upd
...
-    pub tmp_dir: PathBuf,
-    pub data_dir: PathBuf,
-    pub cache_dir: PathBuf,
-    pub file_transfer_dir: PathBuf,
-    pub firmware_dir: PathBuf,
+    pub tmp_dir: Utf8PathBuf,
+    pub data_dir: DataDir,
+    pub cache_dir: Utf8PathBuf,
+    pub file_transfer_dir: Utf8PathBuf,
+    pub firmware_dir: Utf8PathBuf,
```
didier-wenzek, line 27: This should be only a `DataDir` as all these paths are derived from it.

rina23q: Done

albinsuresh (approved):

Issue #2281, opened 2023-09-22
Phase 1: package thin-edge.io as a multi-call binary to reduce resource requirements


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Consolidate the tedge binaries into one multi-call binary to reduce

* Reduce memory usage by removing duplicated runtime environments (e.g. actor runtime)
* Reduce total disk space required to install all required binaries to offer the same functionality

This has already been requested by some other tickets, and conversations with some users:

* https://github.com/thin-edge/thin-edge.io/issues/1459

**Describe the solution you'd like**

Initial work will have to be done to view the merits of making a multi-call binary and how to package it. The initial goal will be to create a working multi-call binary, and then we will work on how to package it based on the improvements to the install and memory usage.

The main takes of this ticket should include:

1. Create single (multi-call) binary which includes the same functionality as the current individual binaries
    * tedge
    * tedge-mapper
    * tedge-agent
    * tedge-log-plugin
    * c8y-configuration-plugin
    * c8y-firmware-plugin
    * tedge-watchdog

    **Note**

    `tedge-apt-plugin` and `c8y-remote-access-plugin` are considered out of scope at the moment. We will evaluate whether creating a [Multi-Call binary](https://man.archlinux.org/man/busybox.1.en#:~:text=BusyBox%20is%20a%20multi%2Dcall,a%20large%20number%20of%20utilities.) (similar to busybox) is feasible or not

2. Implement a way to enable/disable specific features based on configuration (and possibly flags to facilitate starting a dedicated mapper instance?)

3. Compare disk space used between 0.12.0 and the new multi-call binary setup

4. Compare memory consumption of running multiple services (as of 0.12.0) and running one (or two) services.

5. Assess whether included `tedge-apt-plugin` and `c8y-remote-access-plugin` should also be included in the multi-call binary, and check how much work would need to be done to cover this scenario

**Describe alternatives you've considered**

**Additional context**

The current total size of the thin-edge.io binaries is around 38MB (as of `0.12.1~298+g7e6da86`), however each binary uses very similar dependencies, so there is very significant space saving to be made by making a multi-call binary. Below shows the list of binaries currently included part of thin-edge.io:

```sh
# ls -lh /usr/bin/c8y* /usr/bin/tedge*
-rwxr-xr-x 1 root  root  5.4M Sep 20 00:54 /usr/bin/c8y-configuration-plugin
[...]

PR #2295, opened 2023-09-28
Make tedge a multi-call binary

jarhodes314:
## Proposed changes
This is the first change related to #2281. This converts tedge into a multicall (aka busybox-style) binary, so `tedge` will behave differently depending on the executable name (e.g. creating a symlink to `tedge` called `tedge-mapper` and then executing tedge via this symlink will result in the `tedge-mapper` cli being called, not `tedge`).

The PR includes:
* Creation of symlinks for each of the tedge binaries via the `tedge init` command (e.g.  `/usr/bin/tedge-mapper` is a symlink to `/usr/bin/tedge`) 
* Updated packaging of tar and linux packages to utilize the multi-call `tedge` binary instead of including a binary in each individual package. Now the linux packages only contains some configuration files and service definitions (and some unrelated backwards compatible symlinks)

**NB** the changes will be clearer if you view the diff of the two commits separately as I renamed a bunch of main.rs files to lib.rs in the first commit, and then added a bunch of new main.rs files to preserve backwards compatibility. 

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2281

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/init.rs
+++ b/crates/core/tedge/src/cli/init.rs
@@ -1,6 +1,11 @@
 use crate::command::BuildContext;
 use crate::command::Command;
+use crate::Component;
+use anyhow::bail;
 use anyhow::Context;
+use clap::Subcommand;
+use std::os::unix::fs::MetadataExt;
+use std::path::Path;
 use tedge_utils::file::create_directory;
 use tedge_utils::file::PermissionEntry;
 
@@ -8,19 +13,94 @@ use tedge_utils::file::PermissionEntry;
 pub struct TEdgeInitCmd {
     user: String,
     group: String,
+    relative_links: bool,
     context: BuildContext,
 }
 
 impl TEdgeInitCmd {
-    pub fn new(user: String, group: String, context: BuildContext) -> Self {
+    pub fn new(user: String, group: String, relative_links: bool, context: BuildContext) -> Self {
         Self {
             user,
             group,
+            relative_links,
             context,
         }
     }
 
     fn initialize_tedge(&self) -> anyhow::Result<()> {
+        let executable_name =
+            std::env::current_exe().context("retrieving the current executable name")?;
+        let stat = std::fs::metadata(&executable_name).with_context(|| {
+            format!(
+                "reading metadata for the current executable ({})",
+                executable_name.display()
+            )
+        })?;
+        let Some(executable_dir) = executable_name.parent() else {
+            bail!(
+                "current executable ({}) does not have a parent directory",
+                executable_name.display()
+            )
+        };
+        let Some(executable_file_name) = executable_name.file_name() else {
+            bail!(
+                "current executable ({}) does not have a file name",
+                executable_name.display()
+            )
+        };
+
+        let component_subcommands: Vec<String> =
+            Component::augment_subcommands(clap::Command::new("tedge"))
+                .get_subcommands()
+                .map(|c| c.get_name().to_owned())
+                .chain(["tedge-apt-plugin".to_owned()])
+                .collect();
+
+        for component in &component_subcommands {
+            let link = executable_dir.join(component);
+            match std::fs::symlink_metadata(&link) {
+                Err(e) if e.kind() != std::io::ErrorKind::NotFound => bail!(
+                    "couldn't read metadata for {}. do you need to run with sudo?",
+                    link.display()
+                ),
+                meta => {
+                    let file_exists = meta.is_ok();
+                    if file_exists {
+                        nix::unistd::unlink(&link).with_context(|| {
+                            format!("removing old version of {component} at {}", link.display())
+                        })?;
+                    }
+
+                    let tedge = if self.relative_links {
+                        Path::new(executable_file_name)
+                    } else {
+                        &*executable_name
+                    };
+                    std::os::unix::fs::symlink(tedge, &link).with_context(|| {
+                        format!("creating symlink for {component} to {}", tedge.display())
+                    })?;
+
+                    let res = std::process::Command::new("chown")
+                        .arg("--no-dereference")
+                        .arg(&format!("{}:{}", stat.uid(), stat.gid()))
+                        .arg(&link)
+                        .output()
+                        .with_context(|| {
+                            format!(
+                                "executing chown to change ownership of symlink at {}",
+                                link.display()
+                            )
+                        })?;
+                    anyhow::ensure!(
+                        res.status.success(),
+                        "failed to change ownership of symlink at {}\n\nSTDERR: {}",
+                        link.display(),
+                        String::from_utf8_lossy(&res.stderr),
+                    )
+                }
+            }
+        }
+
         let config_dir = self.context.config_location.tedge_config_root_path.clone();
         create_directory(
             &config_dir,
--- a/crates/core/tedge/src/main.rs
+++ b/crates/core/tedge/src/main.rs
@@ -15,7 +15,7 @@ use tedge_config::system_services::set_log_level;
 fn main() -> anyhow::Result<()> {
     let executable_name = executable_name();
 
-    if executable_name.as_deref() == Some("apt") {
+    if matches!(executable_name.as_deref(), Some("apt" | "tedge-apt-plugin")) {
         let try_opt = AptCli::try_parse();
         tedge_apt_plugin::run_and_exit(try_opt);
     }
--- a/crates/core/tedge/src/main.rs
+++ b/crates/core/tedge/src/main.rs
@@ -17,8 +17,7 @@ fn main() -> anyhow::Result<()> {
 
     if executable_name.as_deref() == Some("apt") {
         let try_opt = AptCli::try_parse();
-        tedge_apt_plugin::run(try_opt);
-        return Ok(());
+        tedge_apt_plugin::run_and_exit(try_opt);
     }
 
     let opt = parse_multicall_if_known(&executable_name);
--- a/crates/core/tedge/src/main.rs
+++ b/crates/core/tedge/src/main.rs
@@ -3,37 +3,87 @@
 
 use anyhow::Context;
 use clap::Parser;
+use std::future::Future;
+use std::path::PathBuf;
 use tedge::command::BuildCommand;
 use tedge::command::BuildContext;
+use tedge::Component;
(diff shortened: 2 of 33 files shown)


didier-wenzek:
This is a nice disk foot-print improvement!

With this PR, `tedge` wraps in a single executable the `tedge` command with `tedge-agent`, `tedge-mapper` and some of the plugins. 

Before:

```
 ls -lh target/release/tedge*
-rwxrwxr-x 2 didier didier 6,4M sept. 28 20:53 target/release/tedge
-rwxrwxr-x 2 didier didier 6,7M sept. 28 20:53 target/release/tedge-agent
-rwxrwxr-x 2 didier didier 3,0M sept. 28 20:53 target/release/tedge-apt-plugin
-rwxrwxr-x 2 didier didier 6,6M sept. 28 20:53 target/release/tedge-configuration-plugin
-rwxrwxr-x 2 didier didier 6,1M sept. 28 20:53 target/release/tedge-log-plugin
-rwxrwxr-x 2 didier didier 7,9M sept. 28 20:53 target/release/tedge-mapper
-rwxrwxr-x 2 didier didier 4,7M sept. 28 20:53 target/release/tedge-watchdog
-rwxrwxr-x 2 didier didier 6,4M sept. 28 20:53 target/release/c8y-firmware-plugin
-rwxrwxr-x 2 didier didier 3,5M sept. 28 20:53 target/release/c8y-remote-access-plugin
```

After:

```
$ ls -lh target/release/tedge
-rwxrwxr-x 2 didier didier 9,7M sept. 28 21:02 target/release/tedge

$ ln -s $PWD/target/release/tedge /tmp/tedge-mapper
$ /tmp/tedge-mapper
tedge-mapper is the mapper that translates thin-edge.io data model to c8y/az data model.

Usage: tedge-mapper [OPTIONS] <COMMAND>

Commands:
  az        
  aws       
  c8y       
  collectd  
  help      Print this message or the help of the given subcommand(s)
```

__Note__ this PR still provides old equivalent executables for `tedge-agent`, `tedge-mapper`, plugins ... This is only a temporary step before packaging is updated.

That said, I see this as first step. There are still too many commands and daemons for the users to install and manage.
One need to go a step further and include in `tedge-agent` all the `tedge-xxx-plugin`s: for log management, as well as for configuration and firmware management (not implemented yet). This can simply be done by integrating the associated actors with those of the `tedge-agent`.

didier-wenzek (commented):
I'm okay to go in that direction and even to merge this PR.

However:
- we must wait an approval from @reubenmiller as this has an important impact on packaging.
- we must update packaging and deprecate the executables which are currently provided for backward compatibilities.
- We still have to extend `tedge-agent` to include all the device management features and to deprecate the `tedge-xxx-plugin`s.

reubenmiller:
> I'm okay to go in that direction and even to merge this PR.
> 
> However:
> 
> * we must wait an approval from @reubenmiller as this has an important impact on packaging.
> * we must update packaging and deprecate the executables which are currently provided for backward compatibilities.
> * We still have to extend `tedge-agent` to include all the device management features and to deprecate the `tedge-xxx-plugin`s.

I had a quick look and it's going in a great direction. I'm happy to merge this even though it will lead to a brief period (about a week) where the total binary size would be ~2.5MB more than before until we switch to the multi call binary.

Additional follow up steps might be able to assist the follow up packaging aspects (in addition to the points that @didier-wenzek already mentioned):

* extend the `tedge init` command to create the symlinks for the required binaries (and also take care of upgrade paths when we consolidate the tedge-xxx-plugin binaries as Didier mentioned)
* Add a new subcommand to remove the initialization (but we can post pone this work until a later point in time)

reubenmiller:
@jarhodes314 This is looking really good now. Can you also add support for `tedge-apt-plugin`? I know it was originally out of scope, however this is the only binary which is not included, and it would simplify the packaging slightly and decrease the overall overhead on Debian systems.

Afterwards, I'll push the packaging changes required to make use of the multi-call binary so that we can have the packaging sorted before merging this PR (and validation from the system tests that everything works as before).

jarhodes314:
> @jarhodes314 This is looking really good now. Can you also add support for `tedge-apt-plugin`? I know it was originally out of scope, however this is the only binary which is not included, and it would simplify the packaging slightly and decrease the overall overhead on Debian systems.
> 
> Afterwards, I'll push the packaging changes required to make use of the multi-call binary so that we can have the packaging sorted before merging this PR (and validation from the system tests that everything works as before).

tedge-apt-plugin is already supported, but it's dealt with directly in main.rs so the parsing can be done separately (as the plugin does some error handling for itself so it can change the default exit codes).

jarhodes314:
> @jarhodes314 This is looking really good now. Can you also add support for `tedge-apt-plugin`? I know it was originally out of scope, however this is the only binary which is not included, and it would simplify the packaging slightly and decrease the overall overhead on Debian systems.
> 
> Afterwards, I'll push the packaging changes required to make use of the multi-call binary so that we can have the packaging sorted before merging this PR (and validation from the system tests that everything works as before).

tedge-apt-plugin is already supported, but it's dealt with directly in main.rs so the parsing can be done separately (as the plugin does some error handling for itself so it can change the default exit codes).

reubenmiller:
> > @jarhodes314 This is looking really good now. Can you also add support for `tedge-apt-plugin`? I know it was originally out of scope, however this is the only binary which is not included, and it would simplify the packaging slightly and decrease the overall overhead on Debian systems.
> > Afterwards, I'll push the packaging changes required to make use of the multi-call binary so that we can have the packaging sorted before merging this PR (and validation from the system tests that everything works as before).
> 
> tedge-apt-plugin is already supported, but it's dealt with directly in main.rs so the parsing can be done separately (as the plugin does some error handling for itself so it can change the default exit codes).

Though I don't think the `tedge init` command created the symlink to `/usr/bin/tedge-apt-plugin` then...

reubenmiller:
> > > @jarhodes314 This is looking really good now. Can you also add support for `tedge-apt-plugin`? I know it was originally out of scope, however this is the only binary which is not included, and it would simplify the packaging slightly and decrease the overall overhead on Debian systems.
> > > Afterwards, I'll push the packaging changes required to make use of the multi-call binary so that we can have the packaging sorted before merging this PR (and validation from the system tests that everything works as before).
> > 
> > 
> > tedge-apt-plugin is already supported, but it's dealt with directly in main.rs so the parsing can be done separately (as the plugin does some error handling for itself so it can change the default exit codes).
> 
> Though I don't think the `tedge init` command created the symlink to `/usr/bin/tedge-apt-plugin` then...

Actually `/usr/bin/tedge-apt-plugin` does exist...something strange is happening...I'll recheck it now.

crates/core/tedge/src/main.rs
```diff
@@ -3,37 +3,86 @@
...
 fn main() -> anyhow::Result<()> {
-    set_log_level(tracing::Level::WARN);
+    let executable_name = executable_name();
 
-    let opt = tedge::cli::Opt::parse();
-
-    if opt.init {
-        warn!("This --init option has been deprecated and will be removed in a future release. Use the `tedge init` command instead");
-        return Ok(());
+    if executable_name.as_deref() == Some("apt") {
```
reubenmiller, line 18: Can you change it to also look for the full name of the binary, `tedge-apt-plugin` as well?

For example (obviously modify the code, but you get the idea):

```
if executable_name.as_deref() == Some("apt") || executable_name.as_deref() == Some("tedge-apt-plugin") {
```

Then for the symlink creation, I would create the symlinks under:
* `/usr/bin/tedge` -> `/usr/bin/tedge-apt-plugin` (we can't use /usr/bin/apt as this is already reserved by the real apt binary)

Then the actual tedge-apt-plugin package can handle the creation of the `/etc/tedge/sm-plugins/apt` symlink when the package is installed.

jarhodes314: Ah sure, I overlooked that

reubenmiller: Looks good now.

reubenmiller (approved):
Great step forward. Package sizes are reduced by approximately 4 times!

Approved.

crates/core/tedge/src/cli/init.rs
```diff
@@ -1,26 +1,102 @@
...
+                        Path::new("./tedge")
+                    } else {
+                        &*executable_name
+                    };
+                    std::os::unix::fs::symlink(tedge, &link).with_context(|| {
+                        format!("creating symlink for {component} to {}", tedge.display())
+                    })?;
+
+                    let res = std::process::Command::new("chown")
+                        .arg("-h")
```
Bravo555, line 80: nit: Surprised that in chown, `-h` is not `--help`. For readability I think it's be better to use long flags, so `--no-dereference` in this case.

crates/core/tedge/src/cli/init.rs
```diff
@@ -1,26 +1,102 @@
...
+                    if file_exists {
+                        nix::unistd::unlink(&link).with_context(|| {
+                            format!("removing old version of {component} at {}", link.display())
+                        })?;
+                    }
+
+                    let tedge = if self.relative_links {
+                        // TODO this could lead to broken links
+                        // TODO is this feature even that useful?
+                        Path::new("./tedge")
```
Bravo555, line 71: question: What are some use cases where using relative links could be useful?

Bravo555 (approved):
One question, but otherwise no issues. This should also simplify Yocto Project recipes quite a lot. Nice work!

crates/core/tedge/src/cli/init.rs
jarhodes314: Honestly, I'm not entirely sure, hence the comment immediately above it. I guess it makes moving the binaries a tad easier, but the fact that `tedge init` sets the links up should solve that problem in any case. I guess if someone wants to store tedge on some non-deterministically named mount point or something that might be a possible use case, but I am leaning towards declaring this feature pointless and removing it to simplify the code and CLI.

jarhodes314: I've slightly refactored this, but I think I'll keep this, as it doesn't add much complexity, and there may be valid use cases. Absolute paths are probably the right default, and what `tedge-apt-plugin` uses (although for that case, a relative path wouldn't make any sense).

PR #2297, opened 2023-09-28
Update index.md

nneuerburg:
Add the docker based demo container to the installation page as the easiest way to explore thin-edge.io


didier-wenzek (commented):
I'm okay to link the demo container from the how to start overview, but I would not from the installation page. The later must focus on installation not training.

PR #2300, opened 2023-09-28
Fix integer underflow in EntityTopicId parsing

Bravo555:
## Proposed changes

In a typical case of a fencepost error, segments were confused with segment separators and thus in topics with 4 slashes erroneously passed the "topic has max of 4 segments" check. This would later cause an underflow when generated missing topic segments, leading to a string of size near `usize::MAX` being allocated.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

As for the error itself, I can identify 3 potential causes:

1. No overflow checks in release mode. This is addressed in this PR by enabling overflow checks in release profile. thin-edge isn't really math-heavy, so I doubt it will impact performance, but it can provide additional safety.
2. Lacking test coverage. The error conditions producing `Result::Error` variants were not sufficiently tested. This is also addressed by this PR.
3. And i think the one we should think a bit about: the decision to automatically pad `EntityTopicId` with slashes if not enough are provided.

   It perhaps unnecessarily complicates the function and it may be used incorrectly by clients who expect `EntityTopicId` to be the same as the input they provided.
   This decision was motivated by [this table](https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#group-identifier) in the documentation which specifies that only the first 2 levels are required. When joining `EntityTopicId` with a channel one has to add slashes if fewer than 4 levels were used, so I decided to do this on construction, but since a user can register entities manually, then registration on topic `devices/main` would result in a `EntityTopicId("devices/main//")` being created, which is a completely different topic.
   So as it stands, current `EntityTopicId` implementation is invalid and either it will have to be changed to support number of levels `2..=4`, or remove the feature of not having to fill in all the levels of the MQTT topic identifier portion.
   If we expect a large discussion on this topic, I can create a new issue for it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -167,3 +167,4 @@ lto = true
 opt-level = "z"
 panic = "abort"
 strip = "symbols"
+overflow-checks = true
--- a/crates/core/tedge_api/src/mqtt_topics.rs
+++ b/crates/core/tedge_api/src/mqtt_topics.rs
@@ -235,12 +235,12 @@ impl FromStr for EntityTopicId {
     fn from_str(entity_id: &str) -> Result<Self, Self::Err> {
         const ENTITY_ID_SEGMENTS: usize = 4;
 
-        let entity_id_segments = entity_id.matches('/').count();
+        let entity_id_segments = entity_id.matches('/').count() + 1;
         if entity_id_segments > ENTITY_ID_SEGMENTS {
             return Err(TopicIdError::TooLong);
         }
 
-        let missing_slashes = ENTITY_ID_SEGMENTS - entity_id_segments - 1;
+        let missing_slashes = ENTITY_ID_SEGMENTS - entity_id_segments;
         let topic_id = format!("{entity_id}{:/<1$}", "", missing_slashes);
         if mqtt_channel::Topic::new(&topic_id).is_err() {
             return Err(TopicIdError::InvalidMqttTopic);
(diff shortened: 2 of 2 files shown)


didier-wenzek (approved):
Approved

Bravo555:
> The user has to understand that a registration on `devices/main` is a shortcut for `devices/main//`. Any telemetry data or command for that device will have to published on a topic with the padded id as in `devices/main///<channel>`.

So do I understand correctly that if a user publishes a registration message on `device/main`, which actually refers to the entity `device/main//`, it's OK because thin-edge components only interact with other devices through the entity store, so they only see `device/main//`? Because then the only tricky part would be for potentially other applications which do not use the entity store but want to see the messages to know where the registration message is. For a device with entity topic id of `device/main//`, the registration message could be published on either `device/main//`, `device/main/` or `device/main`. But registration messages are only thin-edge internal thing and aren't supposed to be interacted with directly by other applications, so it's fine?

didier-wenzek:
> > The user has to understand that a registration on `devices/main` is a shortcut for `devices/main//`. Any telemetry data or command for that device will have to published on a topic with the padded id as in `devices/main///<channel>`.
> 
> So do I understand correctly that if a user publishes a registration message on `device/main`, which actually refers to the entity `device/main//`, it's OK because thin-edge components only interact with other devices through the entity store, so they only see `device/main//`? Because then the only tricky part would be for potentially other applications which do not use the entity store but want to see the messages to know where the registration message is. For a device with entity topic id of `device/main//`, the registration message could be published on either `device/main//`, `device/main/` or `device/main`. But registration messages are only thin-edge internal thing and aren't supposed to be interacted with directly by other applications, so it's fine?

I think it's fine. That said, to avoid any confusion, it would be indeed better to accept registration message only on padded topics.

Issue #2301, opened 2023-09-28
Replace current upload mechanism with the uploader actor in new tedge plugins


Ruadhri17:
**Is your feature request related to a problem? Please describe.**

Currently, both tedge-configuration-plugin and tedge-log-plugin use tedge_http_actor for uploading content to the tedge file repository. That generates two problems:

- Plugins are "blocked" until the upload request is finished.
- There is no retry mechanism when a client-side issue occurs, e.g. temporary network connection lost.

**Describe the solution you'd like**

Add uploader_actor that will work similarly to downloader_actor. It should support a retry mechanism, e.g. using a backoff strategy.

This ticket should consist of two steps:

- [x] Add the upload library and uploader_actor. #2302
- [x] Replace the current upload mechanism with the new uploader_actor. #2302

**Describe alternatives you've considered**

**Additional context**

Ruadhri17:
PR with uploader actor was marged into main branch. Uploader actor is currently used in two plugins:
- tedge-configuration-plugin
- tedge-log-plugin

gligorisaev:
@Ruadhri17  Please include in the ticket more information about the final solution/how to verify the change.

Ruadhri17:
@gligorisaev In both plugins uploader actor is responsible for uploading file to `tedge file transfer repository`. Thanks to that plugins can process next requests without waiting for upload to finish and in case of some network connection issues it will retry the upload and not just return `FAILED` status of the request.

As for testing: I created rust [test](https://github.com/thin-edge/thin-edge.io/blob/57ae569814d8c42b28200018bd66aa191adec7d9/crates/common/upload/src/upload.rs#L240) that simulates server behavior: file is sent to server and it responses with http error code 500 to force upload retry. What need to be thoroughly verifed is that uploader successfully uploads file to `tedge file transfer repository` and correctly reacts to different kind of errors (client side errors, file errors, permission errors etc.).

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the bug and could not reproduce it anymore.

PR #2302, opened 2023-09-28
Create uploader actor

Ruadhri17:
## Proposed changes
This PR adds new uploader actor that works in similar manner to `downloader_actor`. Later on it will be used in both `tedge-log-plugin` and `tedge-configuration-plugin` to take responsibility for uploading content to tedge file repository. More in #2301.

TODO:

- [x] Add test that checks if backoff strategy is working.

## Types of changes

## Paste Link to the issue
* #2301
<br/>

## Checklist


albinsuresh (commented):
The overall code looks fine with the basic retry mechanism. But, unlike the downloader which could resume a download from the point where it was interrupted, I didn't see a similar resumption mechanism here. Is there any plan to add such a multi-part upload functionality to this uploader? It doesn't necessarily have to be in this PR, but would be a good follow-up addition.

BTW, I didn't see the log plugin or config plugin using this actor. Isn't that in scope either?

crates/common/upload/src/upload.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tempfile::tempdir;
+    use tokio::io::AsyncWriteExt;
+
+    #[tokio::test]
+    async fn upload_content_no_auth() {
```
rina23q, line 196: You can also create a test `upload_content_with_auth()` based on this test.
The change will be only below:

```rust
let _mock1 = server
    .mock("PUT", "/some_file.txt")
    .with_status(201)
    .match_header("Authorization", mockito::Matcher::Regex(r"Bearer .*".to_string()))
    .create();

let url = UploadInfo::new(&target_url).with_auth(Auth::Bearer("1234".to_string()));

```

Ruadhri17:
> The overall code looks fine with the basic retry mechanism. But, unlike the downloader which could resume a download from the point where it was interrupted, I didn't see a similar resumption mechanism here. Is there any plan to add such a multi-part upload functionality to this uploader? It doesn't necessarily have to be in this PR, but would be a good follow-up addition.

I would leave the partial upload for next PR as it is more tricky than download. According to the [rfc-editor](https://www.rfc-editor.org/rfc/rfc9110.html#name-partial-put) this is very inconsistent and require some changes on server side. Another thing is that we would have to edit the exponential backoff mechanism because of 400 BAD REQUEST respond which at this point will not trigger retry mechanism.

crates/common/upload/src/upload.rs
```diff
@@ -0,0 +1,348 @@
...
+                    &self.source_filename
+                ))
+                .map_err(backoff::Error::Permanent)?
+                .len();
+
+            let file_body = Body::wrap_stream(FramedRead::new(file, BytesCodec::new()));
+
+            let mut client = reqwest::Client::new()
+                .put(url.url())
+                .header(CONTENT_TYPE, "application/octet-stream")
```
Bravo555, line 110: question: Do we always want to set it to `application/octet-stream` though? I realize that it's a safe default and then it's the problem of who received that data to figure out what MIME is it, but maybe we can use a crate for detecting MIME type and then use that.
But it's not a priority and could be somewhat complex, so if you agree, then it could be done in another PR some other time.

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -151,46 +164,67 @@ impl LogManagerActor {
...
-        log_content: String,
-        url: String,
+        topic: &str,
+        result: UploadResult,
     ) -> Result<(), LogManagementError> {
-        let req_builder = HttpRequestBuilder::put(&url)
-            .header("Content-Type", "text/plain")
-            .body(log_content);
+        if let Some(mut request) = self.pending_operations.remove(topic) {
+            let topic = Topic::new_unchecked(topic);
```
Bravo555, line 201: issue: How do we know topic passed here is always valid? I'd rather we use `Topic::new()` and return the error, or take `topic: &Topic` as an argument.

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -1,45 +1,51 @@
...
 
 use super::error::LogManagementError;
 use super::LogManagerConfig;
 use super::DEFAULT_PLUGIN_CONFIG_FILE_NAME;
-use tedge_api::messages::CommandStatus;
-use tedge_api::messages::LogUploadCmdPayload;
 
-fan_in_message_type!(LogInput[MqttMessage, FsWatchEvent] : Debug);
-fan_in_message_type!(LogOutput[MqttMessage]: Debug);
+type MqttTopic = String;
```
Bravo555, line 34: suggestion: Not all strings are valid MQTT topics, so I'd rather see a type like `Topic` used here if possible.

Bravo555 (changes requested):

crates/extensions/tedge_config_manager/src/tests.rs
```diff
@@ -192,22 +192,22 @@ async fn config_manager_uploads_snapshot() -> Result<(), anyhow::Error> {
...
     );
-    assert_request_eq(actual_request, expected_request);
+    assert_eq!(upload_request.file_path, tempdir.path().join("file_b"));
+
+    assert_eq!(upload_request.auth, None);
 
-    // File transfer responds with 201
-    let response = HttpResponseBuilder::new().status(201).build().unwrap();
-    http.send(Ok(response)).await?;
+    // Simulate downloading a file is completed.
```
albinsuresh, line 208:
```suggestion
    // Simulate upload file completion
```

crates/common/log_manager/src/log_utils.rs
```diff
@@ -1,8 +1,13 @@
+use anyhow::anyhow;
```
albinsuresh, line 1: A library returning `anyhow` errors is highly discouraged. You should be defining concrete error types with the use of `thiserror` instead. Anyhow is usually used at the application level where the errors just need to be reported.

crates/common/log_manager/src/log_utils.rs
```diff
@@ -153,6 +160,30 @@ pub fn filter_logs_path_on_metadata(
     }
 }
 
+fn temp_filename(logfile_path: &Path) -> Result<PathBuf, LogRetrievalError> {
```
albinsuresh, line 163:
```suggestion
fn temp_file_at(path: &Path) -> Result<PathBuf, LogRetrievalError> {
```

crates/common/log_manager/src/log_utils.rs
```diff
@@ -153,6 +160,30 @@ pub fn filter_logs_path_on_metadata(
...
+    let file_name = logfile_path
+        .file_name()
+        .ok_or_else(|| FileError::InvalidFileName {
+            path: logfile_path.to_path_buf(),
+            source: anyhow!("Does not name a valid file"),
```
albinsuresh, line 168: I see that this forced you to use `anyhow` as a dependency. Using `anyhow` in the `source` field of `FileError::InvalidFileName` was a mistake. As a short term solution, I'd suggest that you use `PathsError` here as well. Feel free to add one `InvalidFileName` variant as well.

albinsuresh (commented):
The main changes look fine. Just some minor niggles and I'm happy to approve once those are ironed out.

crates/common/log_manager/src/log_utils.rs
rina23q, line 168: Why dou you need `anyhow!`?

crates/extensions/tedge_log_manager/src/actor.rs
rina23q: The HashMap `self.pending_operations` keep topic names (String) that are waiting for uploads by the upload actor. So, the topic from the hashmap should be always valid.

rina23q (commented):
Except `anyhow` in a library crate, looks good to me

crates/common/log_manager/src/log_utils.rs
Bravo555: IMO there are valid use cases for using `anyhow` in a library crate. What matters the most is the intent, what do we expect users to do. We do not expect code which calls the uploader to behave differently between error variants of e.g. "this does not name a valid file" or "path is not valid unicode", we simply return these errors and show them to the user. We don't have to maintain any Rust API compatibility, we can add or remove error variants at any time.
So I think in this case using an error variant for "file error" and then using `anyhow` for telling the user what exactly went wrong is acceptable.
`anyhow::Error` is also like `Box<dyn Error> + 'static`, but also guarantees `Send` and `Sync`. I originally added it to `InvalidFileName` error variant because i wanted to use either `io::Error` or `Utf8Error` as a source, but using `anyhow!("error details")` directly... I wouldn't say it's an antipattern.

I was recently reading _Zero To Production In Rust_ and there's a free chapter about error handling. There is a dedicated [section about anyhow vs thiserror](https://www.lpalmieri.com/posts/error-handling-rust/#anyhow-or-thiserror), but I'd recommend reading the entire chapter for context.

crates/common/log_manager/src/log_utils.rs
Bravo555: anyhow::Error is also like Box<dyn Error> + 'static, but also guarantees Send and Sync. I originally added it to InvalidFileName error variant because i wanted to use either io::Error or Utf8Error as a source, but I wouldn't say using `anyhow!` here to provide a a description of why is the filename invalid, to the user, is an antipattern. This could've just been a string, but because it's `anyhow::Error`, one can also use any other error as a source.
I addressed creating more enum variants in this comment: https://github.com/thin-edge/thin-edge.io/pull/2302#discussion_r1354377885.

crates/common/log_manager/src/log_utils.rs
albinsuresh: Thanks for the link. Gave me a new perspective. I guess the following is the key point:

```
Library authors cannot (or do not want to) make assumptions on the intent of their users. They steer away from being opinionated (to an extent) - enums give users more control, if they need it.
```

and I completely agree to your following point:

> We do not expect code which calls the uploader to behave differently between error variants of e.g. "this does not name a valid file" or "path is not valid unicode", we simply return these errors and show them to the user.

Here is where I'm wondering what value `anyhow` adds if all we wanted to do was to report the reason to the end-user. Why couldn't it be just a `String` value? Using `anyhow` instead, is just an overkill, right? As the user is less likely to introspect the `source` chain any further, as that was our fundamental assumption when we picked `anyhow` for "opaque" errors in the first place.

crates/common/upload/src/upload.rs
albinsuresh: Good point. @Ruadhri17 could you please confirm that the C8y UI is able to properly render the uploaded files as text files in the UI? Else we have a problem and might need to implement some "file content type detection" before setting this.

Ruadhri17: As there is plan to introduce partial PUT, we can also add MIME type detection then. But I agree: it would be much better to  provide type of data we are sending to the server.

Ruadhri17: > Good point. @Ruadhri17 could you please confirm that the C8y UI is able to properly render the uploaded files as text files in the UI? Else we have a problem and might need to implement some "file content type detection" before setting this.

Sure I will test it and if there will be no issues I will request for a review.

Bravo555: @Ruadhri17 confirmed that there are no issues with rendering uploaded files in Cumulocity UI.

Bravo555 (approved):

albinsuresh (approved):
I still have my reservations against the use of `anyhow` to represent a source in error variants like `InvalidFileName` where that enum itself was self-explanatory and I doubt if the `source` would really add any more context. But, that's a parallel discussion that should not block this PR. Hence approving as the rest of the code looks fine,

gligorisaev:

PR #2306, opened 2023-09-29
Resolve RUSTSEC-2023-0065

Bravo555:
## Proposed changes

Affected tungstenite crate was bumped to the latest version 0.20.1

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,7 +150,7 @@ toml = "0.7"
 tracing = { version = "0.1", features = ["attributes", "log"] }
 tracing-subscriber = { version = "0.3", features = ["time", "env-filter"] }
 try-traits = "0.1"
-tungstenite = "0.17.3"
+tungstenite = "0.20"
 url = "2.3"
 uzers = "0.11"
 walkdir = "2"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -150,7 +150,7 @@ toml = "0.7"
 tracing = { version = "0.1", features = ["attributes", "log"] }
 tracing-subscriber = { version = "0.3", features = ["time", "env-filter"] }
 try-traits = "0.1"
-tungstenite = "0.17.3"
+tungstenite = "0.20.1"
```
PradeepKiruvale, line 153:
```suggestion
tungstenite = "0.20"
```
We follow the norm of using only major versions not subversions

didier-wenzek (approved):
Approved

Cargo.toml
Bravo555: Fixed, but note that `0.20.1` here does not mean _use exactly `0.20.1`_, but _use any version in range `0.20.1 <= x < 0.21`_. For `0.20` it's a range of `0.20.0 <= x < 0.21`, and as version `0.20.0` is still affected, without the lockfile cargo could select to use version `0.20.0`. We do have the lockfile though, but still, the declaration in the `Cargo.toml` file says that version `0.20.0` is fine when in reality it isn't.

More info here: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

So I think this norm can be sometimes incorrect, what is the rationale behind it?

PR #2310, opened 2023-10-03
Do not change XIDs of child devices created via ops dir watcher

albinsuresh:
## Proposed changes

The XIDs (external IDs) of devices created using the operations directory file watcher mechanism were changed in #2266 breaking backward compatibility. So, reverting those changes. As the file watcher mechanism was a lower level API specifically designed for C8y, it makes sense to keep the directory name and XID the same.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -36,6 +36,7 @@ use tedge_timer_ext::SetTimeout;
 use tedge_timer_ext::Timeout;
 use tedge_utils::file::create_directory_with_defaults;
 use tedge_utils::file::FileError;
+use tracing::error;
 
 const SYNC_WINDOW: Duration = Duration::from_secs(3);
 
@@ -127,18 +128,46 @@ impl C8yMapperActor {
         match file_event.clone() {
             FsWatchEvent::DirectoryCreated(path) => {
                 if let Some(directory_name) = path.file_name() {
-                    let child_id = directory_name.to_string_lossy().to_string();
-                    let child_topic_id = EntityTopicId::default_child_device(&child_id).unwrap();
+                    let dir_name = directory_name.to_string_lossy().to_string();
+                    let child_external_id =
+                        match CumulocityConverter::validate_external_id(&dir_name) {
+                            Ok(name) => name,
+                            Err(err) => {
+                                error!(
+                                    "Child device directory: {} ignored due to {}",
+                                    &dir_name, err
+                                );
+                                return Ok(());
+                            }
+                        };
+
+                    let child_name = self
+                        .converter
+                        .default_device_name_from_external_id(&child_external_id);
+                    let child_topic_id = EntityTopicId::default_child_device(&child_name).unwrap();
                     let child_device_reg_msg = EntityRegistrationMessage {
                         topic_id: child_topic_id,
-                        external_id: None,
+                        external_id: Some(child_external_id),
                         r#type: EntityType::ChildDevice,
                         parent: None,
-                        other: json!({ "name": child_id }),
+                        other: json!({ "name": child_name }),
                     };
-                    self.converter
+                    match self
+                        .converter
                         .register_and_convert_entity(&child_device_reg_msg)
-                        .unwrap();
+                    {
+                        Ok(messages) => {
+                            for message in messages {
+                                self.mqtt_publisher.send(message).await?;
+                            }
+                        }
+                        Err(err) => {
+                            error!(
+                                "Processing dynamic child device directory creation failed with {}",
+                                err
+                            );
+                        }
+                    }
                 }
             }
             FsWatchEvent::FileCreated(path)
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -51,6 +51,7 @@ use serde_json::Map;
 use serde_json::Value;
 use service_monitor::convert_health_status_message;
 use std::collections::HashMap;
+use std::collections::HashSet;
 use std::fs;
 use std::fs::File;
 use std::io::Read;
@@ -63,6 +64,7 @@ use tedge_api::entity_store::EntityExternalId;
 use tedge_api::entity_store::EntityRegistrationMessage;
 use tedge_api::entity_store::EntityType;
 use tedge_api::entity_store::Error;
+use tedge_api::entity_store::InvalidExternalIdError;
 use tedge_api::event::error::ThinEdgeJsonDeserializerError;
 use tedge_api::event::ThinEdgeEvent;
 use tedge_api::messages::CommandStatus;
@@ -104,6 +106,7 @@ const TEDGE_AGENT_LOG_DIR: &str = "tedge/agent";
 const CREATE_EVENT_SMARTREST_CODE: u16 = 400;
 const TEDGE_AGENT_HEALTH_TOPIC: &str = "tedge/health/tedge-agent";
 const DEFAULT_EVENT_TYPE: &str = "ThinEdgeEvent";
+const FORBIDDEN_ID_CHARS: [char; 3] = ['/', '+', '#'];
 
 #[derive(Debug)]
 pub struct MapperConfig {
@@ -214,8 +217,12 @@ impl CumulocityConverter {
         };
 
         let main_device = entity_store::EntityRegistrationMessage::main_device(device_id.clone());
-        let entity_store =
-            EntityStore::with_main_device(main_device, Self::map_to_c8y_external_id).unwrap();
+        let entity_store = EntityStore::with_main_device(
+            main_device,
+            Self::map_to_c8y_external_id,
+            Self::validate_external_id,
+        )
+        .unwrap();
 
         Ok(CumulocityConverter {
             size_threshold,
@@ -335,6 +342,34 @@ impl CumulocityConverter {
             )
             .into()
         }
+    }
+
+    /// Returns the `device_name` from the `EntityExternalId`
+    /// if it follows the default naming scheme `MAIN_DEVICE_COMMON_NAME:device:device_name`,
+    /// else returns its `String` representation
+    pub fn default_device_name_from_external_id(&self, external_id: &EntityExternalId) -> String {
+        let _main_device_id = &self.device_name;
+        match external_id.as_ref().split(':').collect::<Vec<&str>>()[..] {
+            [_main_device_id, "device", device_id, "service", _] => device_id.into(),
+            [_main_device_id, "device", device_id] => device_id.into(),
+            _ => external_id.into(),
+        }
+    }
+
+    /// Validates if the provided id contains any invalid characters and
+    /// returns a valid EntityExternalId if the validation passes,
+    /// else returns InvalidExternalIdError
+    pub fn validate_external_id(id: &str) -> Result<EntityExternalId, InvalidExternalIdError> {
+        let forbidden_chars = HashSet::from(FORBIDDEN_ID_CHARS);
+        for c in id.chars() {
+            if forbidden_chars.contains(&c) {
+                return Err(InvalidExternalIdError {
+                    external_id: id.into(),
+                    invalid_char: c,
+                });
+            }
+        }
+        Ok(id.into())
     }
 
     fn try_convert_measurement(
@@ -734,13 +769,24 @@ impl CumulocityConverter {
 
         for child_id in difference {
             // here we register new child devices, sending the 101 code
-            let child_topic_id = EntityTopicId::default_child_device(child_id).unwrap();
+            let child_external_id = match CumulocityConverter::validate_external_id(child_id) {
+                Ok(name) => name,
+                Err(err) => {
+                    error!(
+                        "Child device directory: {} ignored due to {}",
+                        &child_id, err
+                    );
+                    continue;
+                }
+            };
+            let child_topic_id =
+                EntityTopicId::default_child_device(child_external_id.as_ref()).unwrap();
             let child_device_reg_msg = EntityRegistrationMessage {
                 topic_id: child_topic_id,
-                external_id: None,
+                external_id: Some(child_external_id.clone()),
                 r#type: EntityType::ChildDevice,
                 parent: None,
-                other: json!({ "name": child_id }),
+                other: json!({ "name": child_external_id.as_ref() }),
             };
             let mut reg_messages = self
                 .register_and_convert_entity(&child_device_reg_msg)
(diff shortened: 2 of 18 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -752,14 +752,9 @@ impl CumulocityConverter {
...
-            let child_topic_id = EntityTopicId::default_child_device(&child_id).unwrap();
-            let child_external_id =
-                Self::map_to_c8y_external_id(&child_topic_id, &self.device_name.as_str().into());
-
-            self.children
-                .insert(child_external_id.as_ref().into(), ops.clone());
+            self.children.insert(child_id.clone(), ops.clone());
             let ops_msg = ops.create_smartrest_ops_message()?;
-            let topic_str = format!("{SMARTREST_PUBLISH_TOPIC}/{}", child_external_id.as_ref());
+            let topic_str = format!("{SMARTREST_PUBLISH_TOPIC}/{}", child_id);
```
Bravo555, line 757: issue: `child_id` which comes from `local_child_devices`, which in turn is derived from names  of filesystem directories, can also be invalid MQTT topics. To fix this, we need to try to parse it as `EntityTopicId` and if it's incorrect, return `Result::Err` from the function, in effect aborting on first invalid input. This can be improved later, by e.g. successfully processing all valid children and reporting any invalid children.

Bravo555: note: there is a `C8yTopic` type that we should ideally be using for usecases like this, but currently it's a little unwieldy because it takes the entire `EntityMetadata`, so for now this manual `format!` is fine, as modifying `C8yTopic` is outside the scope of this PR.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1025,10 +1020,7 @@ impl CumulocityConverter {
         if is_child_operation_path(path) {
             // operations for child
             let child_id = get_child_id(&path.to_path_buf())?;
-            let child_topic_id = EntityTopicId::default_child_device(&child_id).unwrap();
-            let child_external_id =
-                Self::map_to_c8y_external_id(&child_topic_id, &self.device_name.as_str().into());
-            let stopic = format!("{SMARTREST_PUBLISH_TOPIC}/{}", child_external_id.as_ref());
+            let stopic = format!("{SMARTREST_PUBLISH_TOPIC}/{}", child_id);
```
Bravo555, line 1023: issue: the same problem with input validation as earlier

crates/extensions/c8y_mapper_ext/src/log_upload.rs
```diff
@@ -200,15 +200,13 @@ impl CumulocityConverter {
             }
         })?;
 
+        dbg!("Entity in store", &device);
```
Bravo555, line 203: todo: Remove `dbg!`

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -128,13 +128,19 @@ impl C8yMapperActor {
             FsWatchEvent::DirectoryCreated(path) => {
                 if let Some(directory_name) = path.file_name() {
                     let child_id = directory_name.to_string_lossy().to_string();
-                    let child_topic_id = EntityTopicId::default_child_device(&child_id).unwrap();
+                    let default_child_id_prefix =
+                        format!("{}:{}", self.converter.device_name, "device");
```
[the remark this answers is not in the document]
albinsuresh: No, we never had nested child device support via the file system mechanism. And the plan is to deprecate this API in the near future. So, we won't be adding that support either.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -128,13 +128,19 @@ impl C8yMapperActor {
...
+                    let child_external_id = child_id
+                        .strip_prefix(&default_child_id_prefix)
+                        .unwrap_or(&child_id);
+                    let child_topic_id =
+                        EntityTopicId::default_child_device(child_external_id).unwrap();
```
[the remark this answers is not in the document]
albinsuresh: Resolved by sanitizing the input before unwrap.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1069,6 +1069,51 @@ async fn mapper_publishes_supported_operations_for_child_device() {
...
+async fn mapping_child_device_dirs_with_forbidden_characters() {
+    // The test assures tedge-mapper checks if there is a directory for operations for child devices, then it reads and
+    // correctly publishes supported operations message for that child on to `c8y/s/us/child1`
+    // and verifies that message.
+    let cfg_dir = TempTedgeDir::new();
+    create_thin_edge_child_operations(&cfg_dir, "my#complex+child2", vec!["c8y_ChildTestOp1"]);
+
+    let (mqtt, _http, _fs, _timer) = spawn_c8y_mapper_actor(&cfg_dir, false).await;
+    let mut mqtt = mqtt.with_timeout(TEST_TIMEOUT_MS);
+    mqtt.skip(6).await; //Skip all init messages
```
Bravo555, line 1082: suggestion: IMO doing this results in a test not only asserting that "mapper correctly creates child devices from directory names with invalid characters", but also asserting that "there are always 6 init messages", and adding or removing a message in a future will cause a test to fail with a trivial fix, but the annoying thing is the number of the tests that assume number of init messages, and it will have to be changed in every one of them. IIRC I once had to do this, and it also unnecessarily increases the number of files changed, so I'd prefer if we did not assert things like "number of init messages" in a test which shouldn't really care about that.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1181,30 +1226,131 @@ async fn mapper_dynamically_updates_supported_operations_for_child_device() {
...
+    )
+    .await;
+}
+
+#[tokio::test]
+async fn mapping_dynamically_added_child_device_dir() {
+    let cfg_dir = TempTedgeDir::new();
+    let (mqtt, _http, mut fs, _timer) = spawn_c8y_mapper_actor(&cfg_dir, false).await;
+    let mut mqtt = mqtt.with_timeout(TEST_TIMEOUT_MS);
+    mqtt.skip(6).await; //Skip all init messages
```
Bravo555, line 1261: suggestion: see previous comment

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1069,6 +1069,51 @@ async fn mapper_publishes_supported_operations_for_child_device() {
...
+    let (mqtt, _http, _fs, _timer) = spawn_c8y_mapper_actor(&cfg_dir, false).await;
+    let mut mqtt = mqtt.with_timeout(TEST_TIMEOUT_MS);
+    mqtt.skip(6).await; //Skip all init messages
+
+    mqtt.send(MqttMessage::new(
+        &C8yTopic::downstream_topic(),
+        "106,child-one",
+    ))
+    .await
+    .expect("Send failed");
```
Bravo555, line 1089: question: `106` is smartrest for "get children of device", so as I understand we try to get children of `child-one`, but where is this device defined? We don't refer to `child-one` anywhere else in the test, so I'm a bit confused as to why it's there.

Bravo555 (commented):
One reservation to sanitation, and some questions, suggestions and notes - otherwise good to go.

crates/extensions/c8y_mapper_ext/src/tests.rs
albinsuresh: > so I'd prefer if we did not assert things like "number of init messages" in a test which shouldn't really care about that.

Yeah, the `skip` was a way of avoiding asserting all the init messages in each and every test. But, as you highlighted, if the number of `init` messages change, we'll end up updating every test. So, to avoid this, I can move this skip to a dedicated function and have all the tests use that function. That way, in future if we add more init messages, just that single function needs to be updated.

crates/extensions/c8y_mapper_ext/src/tests.rs
albinsuresh: `105` is the SmartREST message to `get children of device` and the cloud will respond with a `106` message with that list. So here, the idea was to just send a `106` message with a list of child devices that doesn't contain the locally created child device. I could have even sent empty list, to avoid this confusion.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1069,6 +1069,51 @@ async fn mapper_publishes_supported_operations_for_child_device() {
     .await;
 }
 
+#[tokio::test]
+async fn mapping_child_device_dirs_with_forbidden_characters() {
+    // The test assures tedge-mapper checks if there is a directory for operations for child devices, then it reads and
+    // correctly publishes supported operations message for that child on to `c8y/s/us/child1`
+    // and verifies that message.
+    let cfg_dir = TempTedgeDir::new();
+    create_thin_edge_child_operations(&cfg_dir, "my#complex+child2", vec!["c8y_ChildTestOp1"]);
```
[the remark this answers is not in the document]
albinsuresh: Since this is a legacy API which we are planning to deprecate in the near future, I'd also go with just an error message.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -337,6 +343,16 @@ impl CumulocityConverter {
         }
     }
 
+    /// Sanitizes the provided id by replacing any characters not allowed by C8Y
+    /// as an external_id with a safe character like _
+    pub fn sanitize_external_id(id: &str) -> EntityExternalId {
```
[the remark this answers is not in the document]
albinsuresh: > We have multiple EntityExternalIds and we don't have a clear way to tell which have these restrictions and which don't

I didn't fully get that statement. Why do you think that we have multiple `EntityExternalId`s with different rules? A single consumer (c8y-mapper in this case) should only have a single kind of external id with a single ruleset applied to it. A different consumer (aws mapper or azure mapper) is free to have a different set of rules. Did you spot the c8y-mapper or any associated component applying different rules to the `EntityExternalId`?

crates/extensions/c8y_mapper_ext/src/tests.rs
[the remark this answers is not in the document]
albinsuresh: I discussed with @reubenmiller and we've agreed to just reject such ids.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -127,24 +128,49 @@ impl C8yMapperActor {
...
             FsWatchEvent::DirectoryCreated(path) => {
                 if let Some(directory_name) = path.file_name() {
-                    let child_id = directory_name.to_string_lossy().to_string();
-                    let default_child_id_prefix =
-                        format!("{}:{}", self.converter.device_name, "device");
-                    let child_external_id = child_id
-                        .strip_prefix(&default_child_id_prefix)
-                        .unwrap_or(&child_id);
+                    let dir_name = directory_name.to_string_lossy().to_string();
+                    let default_child_id_prefix = format!("{}:device:", self.converter.device_name);
```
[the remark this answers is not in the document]
albinsuresh: I'll move this to a dedicated function in the `converter` module along with the other functions handling ids.

crates/extensions/c8y_mapper_ext/src/converter.rs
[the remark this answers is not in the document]
Bravo555: I don't mean different rules, but that the consumer has to remember to apply these rules itself, so the "satisfies consumer rules" invariant isn't really represented on the type level. I mean, right now, everything looks correct to me, but otherwise it's easy to forget that one is supposed to use a `sanitize_external_id` function to make sure that we reject invalid ids, instead of just calling `EntityExtrenalId::from`. There's no mistake now, but it will be easy to make a mistake in the future.

External id without any restrictions, and external id with no invalid characters, are currently both a single type. This is becuase `EntityExternalId` is used in the entity store which should support all the clouds, so currently they have to be a single type. But if we make it easier for users of `EntityStore` to attach their own data to entities, we could have `c8y::EntityExternalId`, `az::EntityExternalId`, `aws::EntityExternalId`, etc. all of them guaranteeing that the value inside always meets their own rules and requirements.

crates/extensions/c8y_mapper_ext/src/tests.rs
Bravo555: > I can move this skip to a dedicated function and have all the tests use that function. That way, in future if we add more init messages, just that single function needs to be updated.

Extracting `6` to a `INIT_MESSAGES_NUMBER` constant will have the same effect and will be a bit more transparent. What I rather meant was to write the test in a way as to not refer to any other messages and only refer to messages we're interested about.

As in the test we're spinning up all of these actors, and we get input and output channels with messages in them. In the test we check if certain input messages when processed by the actor produce certain output messages, so we can just check that these messages we expect are inside the output channel, without expecting to read any other messages that are not part of the test. Then we don't have to even care how many init messages there are.

I don't know how feasible it is currently, as `assert_received_contains_str` function is probably actively reading from the oldest message and taking messages off the channel, so right now it can stay as it is, just wanted to present my viewpoint and ask if you agree in principle.

crates/extensions/c8y_mapper_ext/src/tests.rs
Bravo555: Ah yes, `106` is a "subscribe template" according to C8y Smartrest documentation, so it makes sense, a comment like "simulate a response to 105 from cumulocity" will suffice.

crates/extensions/c8y_mapper_ext/src/converter.rs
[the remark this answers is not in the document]
albinsuresh: > but otherwise it's easy to forget that one is supposed to use a sanitize_external_id function to make sure that we reject invalid ids, instead of just calling EntityExtrenalId::from

This is the reason why we are storing these sanitization functions as fields of the `EntityStore`. With that design, the consumer defines those rules in a `Fn` once and passes it on to the entity_store. From then on, the consumer can call `update` with any `id` and the entity store will do that sanitization internally. The external user doesn't have to remember to sanitize before passing anything to the store.

> But if we make it easier for users of EntityStore to attach their own data to entities, we could have c8y::EntityExternalId, az::EntityExternalId, aws::EntityExternalId

Yes, if we could come up with an abstraction like that, that'd be perfect, as that'll help us avoid this function business.

crates/extensions/c8y_mapper_ext/src/tests.rs
albinsuresh: > Extracting 6 to a INIT_MESSAGES_NUMBER constant will have the same effect and will be a bit more transparent.

Having a dedicated function that'll just exhaust the init messages will help in doing additional validations as well, rather than just skipping a fixed number of messages.

> so we can just check that these messages we expect are inside the output channel, without expecting to read any other messages that are not part of the test

This is the tricky bit. Sometimes we end up sending the same messages during the `init` as well as the result of a mapping. Examples: child device creation messages(`101,...`), supported operation messages(`118,...`) etc. So, while doing something like you proposed where we just search for the desired message in a stream, we need to make sure that we're not hitting the wrong one.

> as assert_received_contains_str function is probably actively reading from the oldest message and taking messages off the channel, so right now it can stay as it is, just wanted to present my viewpoint and ask if you agree in principle.

Yes, that's the case and if we can write a similar helper that can search for the desired pattern ignoring false positives, that'd be great.

Bravo555 (approved):
LGTM

rina23q (approved):
LGTM

albinsuresh:
### Test Plan

1. Any child devices created in C8y using the file watcher mechanism by creating a dir under `/etc/tedge/operations/c8y` with the child device name should have the same directory name as name and external_id in C8y. This behaviour was accidentally changed in #2266 but the original behaviour has been restored.
2. If a child device directory is created with the default naming scheme such as `MAIN_DEVICE_ID:device:my_child_id`, then this ID is used as the external ID, but the name of the device would be just `my_child_id`, extracted from the long external ID.
3. If the name contains the following forbidden chars: `+`, `/`, or `#`, those directories are ignored with an error logged by the mapper.
4. The above rules apply for dynamically added child directories under `/etc/tedge/operations/c8y` as well.

gligorisaev:

- tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_child_device.robot
- tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_child_device_retry.robot
- tests/RobotFramework/tests/cumulocity/log/log_operation_child.robot
- tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
- tests/RobotFramework/tests/cumulocity/restart/restart_device_child.robot
- tests/RobotFramework/tests/cumulocity/telemetry/c8y_child_alarms_rpi.robot
- tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
- tests/RobotFramework/tests/customizing/data_path_config.robot

Issue #2311, opened 2023-10-03
Cumulocity registration is sent 2 times if an entity is auto-registered


Bravo555:
**Describe the bug**

Because the c8y mapper publishes an entity registration message when auto-registering an entity (so other clients can be aware of a new entity),  a cloud registration message can be sent 2 times. This is because when I was initially working on the entity store, I didn't realise that if an MQTT client subscribes to topic `T` and publishes a message on topic `T`, the broker will echo the message back to the client. This makes it so that when c8y mapper auto-registers an entity it publishes a registration message for other clients, but it will pick up this message again, leading to a 2nd `entity_store.update()` being called and 2nd `self.try_convert_entity_registration()` being called.

**To Reproduce**

1. Publish a message on a topic of entity that wasn't registered, i.e. `te/device/main/service/my_not_registered_service_1/m/my_measurement_series`:
   ```
   $ tedge mqtt pub te/device/main/service/my_not_registered_service_1/m/my_measurement_series '{"temperature": 21.37}'
   ```

2. The [`102 (service creation)`](https://cumulocity.com/guides/reference/smartrest-two/#service-creation-102) smartrest message has been sent twice:

    ```
    [te/device/main/service/my_not_registered_service_2/m/my_measurement_series] {"temperature": 21.37}
    [te/device/main/service/my_not_registered_service_2] {"@id":"devcontainer:device:main:service:my_not_registered_service_2","@parent":"device/main//","@type":"service","name":"my_not_registered_service_2","type":"systemd"}

    **[c8y/s/us] 102,devcontainer:device:main:service:my_not_registered_service_2,systemd,my_not_registered_service_2,up**
    [c8y/measurement/measurements/create] {"temperature":{"temperature":{"value":21.37}},"time":"2023-10-03T17:51:25.679036501Z","type":"my_measurement_series"}

    **[c8y/s/us] 102,devcontainer:device:main:service:my_not_registered_service_2,systemd,my_not_registered_service_2,up**
    ```

**Expected behavior**

Entity auto-registration leads only to a single cloud registration being sent.

**Additional context**

[...]

PR #2312, opened 2023-10-04
Fix 2311: don't publish c8y register message twice

Bravo555:
## Proposed changes

To solve the issue, registration message conversion was changed to only convert if the registration message affected any entities. An update is considered to affect an entity if the entity was previously registered and the update changes any part of entity metadata (i.e. is different from a previous one) or adds or removes children of a device.

As part of the fix, it was necessary to change the `EntityRegistrationMessage::payload` field. It was renamed to `other` and made to not include `@id`, `@type`, `@parent`, etc. fields that are already direct fields of `EntityRegistrationMessage`.

A test `does_not_duplicate_global_registration_2311` was added to ensure the fix had intended effect. The fix was confirmed to turn the test from red to green.

One non-intuitive part is that the registration of a brand new entity is not considered to affect that new entity. Instead, adding or removing any entity always affects the main device, since it's the root parent of all other entities, so the cloud registration message is sent because the main device is updated. In the future, after we update entity store we might instead dynamically parse affected entities returned by the update function to determine which cloud registration messages to send upstream. It would then make sense to revise the decision that registering a new entity is not considered by the entity store to affect that entity.

## Types of changes

## Paste Link to the issue
#2311

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -7,6 +7,7 @@
 
 // TODO: move entity business logic to its own module
 
+use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 
 use crate::entity_store;
@@ -14,6 +15,8 @@ use crate::mqtt_topics::EntityTopicId;
 use crate::mqtt_topics::TopicIdError;
 use mqtt_channel::Message;
 use serde_json::json;
+
+use serde_json::Value as JsonValue;
 
 /// Represents an "Entity topic identifier" portion of the MQTT topic
 ///
@@ -115,7 +118,7 @@ impl EntityStore {
             external_id: entity_id.clone(),
             r#type: main_device.r#type,
             parent: None,
-            other: main_device.payload,
+            other: main_device.other,
         };
 
         Some(EntityStore {
@@ -264,18 +267,26 @@ impl EntityStore {
             r#type: message.r#type,
             external_id: external_id.clone(),
             parent,
-            other: message.payload,
+            other: message.other,
         };
 
         // device is affected if it was previously registered and was updated
-        let previous = self
-            .entities
-            .insert(entity_metadata.topic_id.clone(), entity_metadata);
+        // (i.e. EntityMetadata has changed)
+        let previous = self.entities.entry(topic_id.clone());
+        match previous {
+            Entry::Occupied(mut occupied) => {
+                // if there is no change, no entities were affected
+                if *occupied.get() == entity_metadata {
+                    return Ok(vec![]);
+                }
 
-        if previous.is_some() {
-            affected_entities.push(topic_id);
-        } else {
-            self.entity_id_index.insert(external_id, topic_id);
+                occupied.insert(entity_metadata);
+                affected_entities.push(topic_id);
+            }
+            Entry::Vacant(vacant) => {
+                vacant.insert(entity_metadata);
+                self.entity_id_index.insert(external_id, topic_id);
+            }
         }
 
         Ok(affected_entities)
@@ -320,7 +331,7 @@ impl EntityStore {
                     external_id: Some(device_external_id),
                     r#type: EntityType::ChildDevice,
                     parent: None,
-                    payload: json!({ "name": device_local_id }),
+                    other: json!({ "name": device_local_id }),
                 };
                 register_messages.push(device_register_message.clone());
                 self.update(device_register_message)?;
@@ -336,7 +347,7 @@ impl EntityStore {
                     external_id: Some(service_external_id),
                     r#type: EntityType::Service,
                     parent: Some(parent_device_id),
-                    payload: json!({ "name": service_id,  "type": "systemd" }),
+                    other: json!({ "name": service_id,  "type": "systemd" }),
                 };
                 register_messages.push(service_register_message.clone());
                 self.update(service_register_message)?;
@@ -355,7 +366,7 @@ pub struct EntityMetadata {
     pub parent: Option<EntityTopicId>,
     pub r#type: EntityType,
     pub external_id: EntityExternalId,
-    pub other: serde_json::Value,
+    pub other: JsonValue,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -411,11 +422,16 @@ pub enum Error {
 /// An object representing a valid entity registration message.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct EntityRegistrationMessage {
+    // fields used by thin-edge locally
     pub topic_id: EntityTopicId,
     pub external_id: Option<EntityExternalId>,
     pub r#type: EntityType,
     pub parent: Option<EntityTopicId>,
-    pub payload: serde_json::Value,
+
+    // other properties, usually cloud-specific
+    // TODO: replace with `Map` and use type wrapper that forbids fields `@id`,
+    // `@parent`, etc.
+    pub other: JsonValue,
 }
 
 impl EntityRegistrationMessage {
@@ -424,14 +440,21 @@ impl EntityRegistrationMessage {
     /// MQTT message is an entity registration message if
     /// - published on a prefix of `te/+/+/+/+`
     /// - its payload contains a registration message.
+    // TODO: replace option with proper error handling
+    // TODO: this is basically manual Deserialize implementation, better impl
+    // Serialize/Deserialize.
     #[must_use]
     pub fn new(message: &Message) -> Option<Self> {
         let payload = parse_entity_register_payload(message.payload_bytes())?;
 
-        let r#type = payload
-            .get("@type")
-            .and_then(|t| t.as_str())
-            .map(|t| t.to_owned())?;
+        let JsonValue::Object(mut properties) = payload else {
+            return None;
+        };
+
+        let Some(JsonValue::String(r#type)) = properties.remove("@type") else {
+            return None;
+        };
+
         let r#type = match r#type.as_str() {
             "device" => EntityType::MainDevice,
             "child-device" => EntityType::ChildDevice,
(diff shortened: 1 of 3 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -818,12 +818,17 @@ impl CumulocityConverter {
         match &channel {
             Channel::EntityMetadata => {
                 if let Ok(register_message) = EntityRegistrationMessage::try_from(message) {
-                    if let Err(e) = self.entity_store.update(register_message.clone()) {
-                        error!("Could not update device registration: {e}");
+                    match self.entity_store.update(register_message.clone()) {
+                        Err(e) => {
+                            error!("Could not update device registration: {e}");
```
albinsuresh, line 823:
```suggestion
                            error!("Entity registration failed: {e}");
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2287,6 +2293,50 @@ mod tests {
...
+
+        let mqtt_schema = MqttSchema::new();
+        let measurement_topic = mqtt_schema.topic_for(
+            &"device/main/service/my_measurement_service"
+                .parse()
+                .unwrap(),
+            &tedge_api::mqtt_topics::Channel::Measurement {
+                measurement_type: "my_type".to_string(),
+            },
+        );
```
albinsuresh, line 2309: In tests, feel free to use raw representation of the topics and payloads, rather than complicating it with `MqttSchema` and all. The readability would be much better if we can just see the topics and payloads directly, rather than assuming what the `MqttSchema` schema would have produced.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2287,6 +2293,50 @@ mod tests {
...
+        while let Some(incoming_message) = incoming_messages.pop_front() {
+            let converted_messages = converter.convert(&incoming_message).await;
+            outgoing_messages.extend_from_slice(&converted_messages);
+
+            let converter_subs = &converter.config.topics;
+            converted_messages
+                .into_iter()
+                .filter(|m| converter_subs.accept(m))
+                .for_each(|m| incoming_messages.push_back(m))
+        }
```
albinsuresh, line 2325: Why this while loop when the deque only contains a single message?

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -424,47 +440,75 @@ impl EntityRegistrationMessage {
...
             .topic
             .name
             .strip_prefix(MQTT_ROOT)
             .and_then(|s| s.strip_prefix('/'))?;
 
+        let other = JsonValue::Object(properties);
+
+        assert_eq!(other.get("@id"), None);
+        assert_eq!(other.get("@type"), None);
+        assert_eq!(other.get("@parent"), None);
```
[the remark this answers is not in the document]
Bravo555: These assertions should never fail, as I removed these entries before in a `properties` map, but I'd like to keep them as a final defensive check that the returned object doesn't have those properties. It's also more clear when reading that "`other` should not have these properties" than when it's scattered above in `.remove()` calls.

[Asserts are not only useful for tests](https://tratt.net/laurie/blog/2023/rusts_two_kinds_of_assert_make_for_better_code.html). Here I used them more like a comment that "these things should be true" that also executes and checks that these things are indeed true. Perhaps here it's a little trivial, because I do remove these entries above, so I could've used `debug_assert`, but it's not like this is a hot loop or anything so these asserts should be essentially free.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: In the last line with `for_each` I feed local registration messages back to `incoming_messages`, so both measurement message and auto-generated registration message are processed.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: In the future, I'd like to move away from using plain MQTT topics and contain them only to places which deal with MQTT directly, i.e. the MQTT mapper. I agree though that using `MqttSchema` is a bit of a pain right now, not only here, but everywhere, and it's partly because all the code has to be aware of the MQTT root in order to generate messages. I'll see if this can be simpified.

crates/core/tedge_api/src/entity_store.rs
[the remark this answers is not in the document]
albinsuresh: Thanks for that article. I didn't know about `debug_assert` before, and yes I agree that entity registrations are not that frequent to warrant any performance worries.

albinsuresh (approved):
The core logic looks fine, hence approving. Just added some minor "code style" comments though, which you may address before the merge, if convinced.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2287,6 +2293,50 @@ mod tests {
...
+
+        while let Some(incoming_message) = incoming_messages.pop_front() {
+            let converted_messages = converter.convert(&incoming_message).await;
+            outgoing_messages.extend_from_slice(&converted_messages);
+
+            let converter_subs = &converter.config.topics;
+            converted_messages
+                .into_iter()
+                .filter(|m| converter_subs.accept(m))
+                .for_each(|m| incoming_messages.push_back(m))
```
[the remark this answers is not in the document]
Bravo555: Yes, that's indeed far simpler, so I'll apply these suggestions before merging

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2287,6 +2293,50 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn does_not_duplicate_global_registration_2311() {
```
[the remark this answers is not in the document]
Bravo555: I think by mentioning the issue number you can easily convey more information why the test is there, and why this contract exists in the first place. I could've written a detailed comment about what behaviour this test is testing and why it exists, which would just be restating things from the issue, or just tag the issue number so somebody can look up that issue and hopefully understand why we test that behaviour, since it's not really well documented anywhere else. And usually when closing a bug I think it's good to have a test that verifies the fix, and without tagging the issue you kind of lose context.

So I'd prefer to keep the issue number in, but if you're sure, I can remove it.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -531,7 +575,7 @@ mod tests {
                 external_id: Some("test-device".into()),
                 r#type: EntityType::MainDevice,
                 parent: None,
-                payload: json!({"@type": "device"}),
+                other: json!({"@type": "device"}),
```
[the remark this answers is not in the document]
Bravo555: fixed in 43bfa55

crates/extensions/c8y_mapper_ext/src/converter.rs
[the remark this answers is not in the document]
Bravo555: fixed in 43bfa55

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2287,6 +2293,38 @@ mod tests {
...
+        let tmp_dir = TempTedgeDir::new();
+        let (mut converter, _http_proxy) = create_c8y_converter(&tmp_dir).await;
+
+        let measurement_message = Message::new(
+            &Topic::new_unchecked("te/device/main/service/my_measurement_service/m/my_type"),
+            r#"{"temperature": 21.37}"#,
+        );
+
+        // when auto registrating, local and cloud registration messages should be produced
```
albinsuresh, line 2306:
```suggestion
        // when auto-registered, local and cloud registration messages should be produced
```
Just fix that to avoid this annoying warning from Github actions in future.

crates/extensions/c8y_mapper_ext/src/converter.rs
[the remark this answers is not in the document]
Bravo555: Yeah, now I get what you mean. Fixed in 191309c711.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2287,6 +2293,38 @@ mod tests {
...
+            .iter()
+            .any(|m| m.topic.name == "c8y/s/us" && m.payload_str().unwrap().starts_with("102")));
+
+        // when converting a registration message the same as the previous one, no additional registration messages should be produced
+        let mapped_messages = converter.convert(local_registration_message).await;
+
+        let second_registration_message_mapped = mapped_messages.into_iter().any(|m| {
+            m.topic.name.starts_with("c8y/s/us") && m.payload_str().unwrap().starts_with("102")
+        });
+        assert!(!second_registration_message_mapped);
```
albinsuresh, line 2325:
```suggestion
        assert!(mapped_messages.is_empty());
```
That iteration over `mapped_messages` gives the feeling that some other mapped messages could still be there in it, whereas it shouldn't have any.

albinsuresh (approved):

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: I found that often times when we add new features that result in new messages being published, tests which assume the number of messages published need to get updated. Here I just wanted to assert that the registration message for this particular service was not produced again, with as few additional assumptions as possible.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: Fixed in cd49504

Bravo555:
Everything squashed into cd49504, should be good to merge now.

Issue #2318, opened 2023-10-05
errors when uploading and applying binary configuration files to main and child devices


reubenmiller:
**Describe the bug**

thin-edge.io has the following bugs when dealing with the configuration management on the main and child devices:

* main device
    * Uploading a binary configuration file from the device to the cloud fails
* child device
    * Downloading a binary configuration file from the device to the cloud fails
    * Uploading a binary configuration file 

**To Reproduce**

**Main Device (Uploading from a device to cloud)**

1. Install thin-edge.io and configure to connect to Cumulocity IoT
2. Configure the configuration plugin to access a binary file using the configuration file

    *file: /etc/tedge/c8y/c8y-configuration-plugin.toml*

    ```sh
    # Add the configurations to be managed by c8y-configuration-plugin
    files = [
        { path = '/etc/tedge/tedge.toml.tar.gz' },
    ]
    ```

3. Create the dummy binary file on the thin-edge.io device (so there is something to upload)

    ```sh
    cd /etc/tedge
    tar czvf tedge.toml.tar.gz tedge.toml.tar.gz
    ```

4. In the Cumulocity UI (Device management application), request the binary configuration file with type `/etc/tedge/tedge.toml.tar.gz`

**Child Device**

For downloads, the file is downloaded to the file server, however accessed by a child device, the local http server returns a status code 500 (indicating an unexpected error). Below shows the response shown by the server when access via curl:

```sh
curl -f http://172.21.0.2:8000/tedge/file-transfer/TST_bolt_snow_developer/config_update/config1
curl: (22) The requested URL returned error: 500 Internal Server Error
```

See the linked PR's system tests for the example procedure.

**Expected behavior**

Configuration files should be treated as plain bytes and the contents should not be assumed to be of any specific format (e.g. utf-8 encoded text). This is applicable for both the main and child devices.

**Screenshots**

**Downloading**

Downloading a binary from Cumulocity to the device fails with the failure reason:

```
Could not make a successful request to the remote server
```

```log
Oct 05 20:23:49 pippin c8y-configuration-plugin[449]: 2023-10-05T10:23:49.62405932Z  INFO c8y_config_manager::download: Received c8y_DownloadConfigFile request for config type: /etc/tedge/tedge.toml.tar.gz from device: raspberry-d83add42f121
[...]

PR #2319, opened 2023-10-05
Handle binary uploads/downloads without attempting conversion to UTF-8

jarhodes314:
## Proposed changes
Remove some unnecessary conversions to UTF-8 when handling uploads/downloads that may not be text.

Notes:
* The robotframework-c8y library was updated as a new assertion was added to assist in comparing the md5 checksum of binaries which are attached to a Cumulocity Event.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2318

## Checklist


crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -393,8 +394,7 @@ impl C8YHttpProxyActor {
                 end_point.get_url_for_event_binary_upload(&event_response_id);
             Ok(HttpRequestBuilder::post(&binary_upload_event_url)
                 .header("Accept", "application/json")
-                .header("Content-Type", "text/plain")
```
reubenmiller, line 396: This should be added back in otherwise c8y responds with a 415 status code, though use a Content-Type of `application/octet-stream` instead.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -393,8 +394,7 @@ impl C8YHttpProxyActor {
...
-                .body(file_content.to_string()))
+                .body(file_content.clone()))
```
albinsuresh, line 397: It's unfortunate that we have to clone the whole file contents here again. But, I see how the closure forced you into it. Could we have used an `Arc` over that `file_content` Vector to avoid this?

albinsuresh (approved):
The core changes look fine. Left some minor comments/queries.

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -52,6 +63,15 @@ Get configuration
...
     Should Be Equal    ${target_contents}    {"name":"configuration1"}
 
+Get binary configuration
+    [Tags]    \#2318
+    ${operation}=    Cumulocity.Get Configuration    CONFIG1_BINARY
+    ${operation}=    Operation Should Be SUCCESSFUL    ${operation}    timeout=120
+    ${events}=    Cumulocity.Device Should Have Event/s    minimum=1    maximum=1    type=CONFIG1_BINARY
+    ${uploaded_contents}=    Cumulocity.Event Should Have An Attachment    ${events[0]["id"]}
+    ${target_checksum}=    Execute Command    md5sum /etc/binary-config1.tar.gz | cut -d' ' -f1    strip=${True}
+    Should Be Equal    ${target_checksum}    27e57d7f840592dd683138a609e72fac
```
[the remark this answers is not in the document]
reubenmiller: Good point. Lapse of judgement...It should be comparing the file checksum with the checksum from the ${uploaded_contents} variable

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -27,6 +27,17 @@ Set configuration when file exists
     # Note: File permission will not change if the file already exists
     Should Be Equal    ${FILE_MODE_OWNERSHIP}    644 root:root
 
+Set binary configuration
+    [Tags]    \#2318
+    ${config_url}=    Cumulocity.Create Inventory Binary    binary-config1    CONFIG1_BINARY    file=${CURDIR}/binary-config1.tar.gz
+    ${operation}=    Cumulocity.Set Configuration    CONFIG1_BINARY    url=${config_url}
+    ${operation}=    Operation Should Be SUCCESSFUL    ${operation}    timeout=120
+    ${target_checksum}=    Execute Command    md5sum /etc/binary-config1.tar.gz | cut -d' ' -f1    strip=${True}
+    Should Be Equal    ${target_checksum}    27e57d7f840592dd683138a609e72fac
```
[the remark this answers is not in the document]
reubenmiller: That won't work as we want to get the checksum of a local file (not from the device...though I guess you could do that in the test setup phase...) but I don't think these test files will be changing any time soon.

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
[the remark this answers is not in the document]
reubenmiller: Fixed by dd8e52abf05bf5ac289d34d58b22204eb285edb1. I ended up extending the testing library to support providing the md5 checksum as an optional argument.

crates/extensions/c8y_http_proxy/src/actor.rs
reubenmiller: Fixed by 6d7a38f854b8c99c125503194fa1fd9ee82f710f

crates/extensions/c8y_http_proxy/src/actor.rs
jarhodes314: It is unfortunate, but no different to what was here previously, just a bit more explicit.

albinsuresh: How about reading the file content from inside the closure? I quickly tried it and it seemed to work. The only thing was that I had to switch back to `std::fs::read` from `tokio::fs::read`. Do you see any problem with that?

jarhodes314: I know we were using std::fs::read, but blocking in an async context in tokio is usually a really bad idea. Because thin edge spawns a bunch of small tasks (and therefore tokio runs them on multiple OS threads), it's not that likely to cause an issue, but on thick edge, I've seen issues where async functions completely fail to make any progress, due in part to executing a blocking operation (in that case it was println! that was the culprit)

jarhodes314: That's not to say we can't mitigate this either with spawn_blocking, or by returning a future from the closure, just that naïvely using blocking functions to read files is something that is actually worth avoiding

albinsuresh: Even without the `clone()` problem, we are reading the entire file content to memory. Is there any way to avoid that. I have seen the `reqwest` crate accepting a `File` handle directly into its body. Can we do something similar here? Not necessarily in this PR, if not trivial. Otherwise let's just defer it to a follow-up PR. No point blocking this PR for it.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -388,13 +399,28 @@ impl C8YHttpProxyActor {
...
+
+            async {
+                // TODO: Upload the file as a multi-part stream
+                let file_content = tokio::fs::read(&request.file_path)
+                    .await
+                    .map_err(SMCumulocityMapperError::from)?;
+
+                let (content_type, body) = match String::from_utf8(file_content) {
+                    Ok(text) => ("text/plain", hyper::Body::from(text)),
+                    Err(not_text) => ("application/octet-stream", not_text.into_bytes().into()),
```
albinsuresh, line 414: Much better this way. At least we could avoid that `file_content.clone()`. But, we're still reading the entire file contents to memory(we always were)  and we should explore ways to avoid that, in a separate task of course. I'll create one for that. Passing a `File` handle directly as the `Body` seemed pretty straight-forward with `reqwest`, but not that direct with `hyper`. I still found some samples though, as follows:

```
// Use tokio::io::AsyncRead to stream the file content
    let file_stream = tokio::io::AsyncReadExt::read_to_string(file, String::new()).into_stream();

    // Set the request body to the file content stream
    *request.body_mut() = hyper::Body::wrap_stream(file_stream);
```

May be there are even better ways, but all of that can be explored in a dedicated ticket.

albinsuresh (approved):
Approving again. I'll create a separate task for handling file uploads without reading the entire file content into memory.

crates/extensions/c8y_http_proxy/src/actor.rs
jarhodes314: I would have had a go, but getting the content-type right is the tricky bit (aka a bit with some design decisions that I don't want blocking the rest of the bugfix). I didn't check this myself, but I assume for getting the file to show up directly in the UI without needing to download it, we need to upload it as `text/plain`. At the moment, I get around this by just attempting to parse the entire thing to UTF-8 and setting content-type based on whether that succeeds.

albinsuresh: Aah okay. Yeah externally figuring out that content length is gonna be tricky. In the worst case, we can read the file line-by-line and validating if it's all valid utf-8, instead of reading everything into memory. And then pass the file handle to the `RequestBuilder` with the appropriate content type.

PR #2320, opened 2023-10-05
tedge-mapper-aws health message uses two different timestamp formats for up and down messages

PradeepKiruvale:
## Proposed changes

To unify the timestamps that are used across `thin-edge`, Update the `health_status_up_message` API to use RFC3339 time format rather than `unix` timestamp.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2237

## Checklist

### DIFF
--- a/crates/core/tedge_watchdog/src/systemd_watchdog.rs
+++ b/crates/core/tedge_watchdog/src/systemd_watchdog.rs
@@ -19,6 +19,7 @@ use tedge_api::health::health_status_down_message;
 use tedge_api::health::health_status_up_message;
 use tedge_api::health::send_health_status;
 use tedge_config::TEdgeConfigLocation;
+use time::format_description;
 use time::OffsetDateTime;
 use tracing::debug;
 use tracing::error;
@@ -29,7 +30,7 @@ use tracing::warn;
 pub struct HealthStatus {
     status: String,
     pid: u32,
-    time: i64,
+    time: String,
 }
 
 pub async fn start_watchdog(tedge_config_dir: PathBuf) -> Result<(), anyhow::Error> {
@@ -146,7 +147,11 @@ async fn monitor_tedge_service(
 
         let start = Instant::now();
 
-        let request_timestamp = OffsetDateTime::now_utc().unix_timestamp();
+        let request_timestamp = OffsetDateTime::now_utc();
+        let request_timestamp = request_timestamp
+            .format(&time::format_description::well_known::Rfc3339)?
+            .as_str()
+            .into();
         match tokio::time::timeout(
             tokio::time::Duration::from_secs(interval),
             get_latest_health_status_message(request_timestamp, &mut received),
@@ -154,6 +159,7 @@ async fn monitor_tedge_service(
         .await
         {
             Ok(health_status) => {
+                let health_status = health_status?;
                 debug!(
                     "Sending notification for {} with pid: {}",
                     name, health_status.pid
@@ -174,16 +180,27 @@ async fn monitor_tedge_service(
 }
 
 async fn get_latest_health_status_message(
-    request_timestamp: i64,
+    request_timestamp: String,
     messages: &mut mpsc::UnboundedReceiver<Message>,
-) -> HealthStatus {
+) -> Result<HealthStatus, WatchdogError> {
     loop {
         if let Some(message) = messages.next().await {
             if let Ok(message) = message.payload_str() {
                 debug!("Health response received: {}", message);
                 if let Ok(health_status) = serde_json::from_str::<HealthStatus>(message) {
-                    if health_status.time >= request_timestamp {
-                        return health_status;
+                    let request_timestamp = OffsetDateTime::parse(
+                        &request_timestamp,
+                        &format_description::well_known::Rfc3339,
+                    )?
+                    .unix_timestamp();
+                    let datetime = OffsetDateTime::parse(
+                        &health_status.time,
+                        &format_description::well_known::Rfc3339,
+                    )?
+                    .unix_timestamp();
+
+                    if datetime >= request_timestamp {
+                        return Ok(health_status);
                     } else {
                         debug!(
                             "Ignoring stale health response: {:?} older than request time: {}",
@@ -232,6 +249,8 @@ fn get_watchdog_sec(service_file: &str) -> Result<u64, WatchdogError> {
 
 #[cfg(test)]
 mod tests {
+    use std::time::Duration;
+
     use anyhow::Result;
     use serde_json::json;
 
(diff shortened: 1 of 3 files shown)


albinsuresh (commented):
The `watchdog` also needs to be updated, right? As that's still relying on UNIX timestamps.

reubenmiller:
@PradeepKiruvale You will need to rebase this PR as it is out of date and the failing test is due to the change in the multi-call binary setup introduced in https://github.com/thin-edge/thin-edge.io/pull/2295

PradeepKiruvale:
> The `watchdog` also needs to be updated, right? As that's still relying on UNIX timestamps.

Yes, I have updated.

crates/core/tedge_watchdog/Cargo.toml
```diff
@@ -11,6 +11,7 @@ repository = { workspace = true }
 
 [dependencies]
 anyhow = { workspace = true }
+chrono = { workspace = true }
```
albinsuresh, line 14: Same as above.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -242,27 +251,55 @@ mod tests {
         let (mut sender, mut receiver) = mpsc::unbounded::<Message>();
         let health_topic = Topic::new("tedge/health/test-service").expect("Valid topic");
 
-        for x in 1..5i64 {
+        for _x in 1..5i64 {
             let health_status = json!({
                 "status": "up",
                 "pid": 123u32,
-                "time": x,
+                "time": "2013-06-22T17:03:14.000+02:00",
```
albinsuresh, line 258: The point of the loop was to have a set of health status messages with increasing timestamp values, not a static value like this. So, may be you can generate a few that increases by a second or so.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -242,27 +251,55 @@ mod tests {
...
-        let health_status = get_latest_health_status_message(3, &mut receiver).await;
-        assert_eq!(health_status.time, 3);
+        let health_status =
+            get_latest_health_status_message("2013-06-21T17:03:14.000+02:00".into(), &mut receiver)
+                .await;
+
+        assert_eq!(
+            &health_status.unwrap().time,
+            "2013-06-22T17:03:14.000+02:00"
+        );
```
albinsuresh, line 272: That changed the logic of the test. The test was validating that if it passes the 3rd timestamp, then the response won't have the older timestamps: 1 & 2. This modified just tests the equality aspect.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
PradeepKiruvale: fixed

crates/core/tedge_watchdog/Cargo.toml
PradeepKiruvale: using time crate

Cargo.toml
```diff
@@ -58,6 +58,7 @@ camino = "1.1"
 certificate = { path = "crates/common/certificate" }
 clap = { version = "4.4", features = ["cargo", "derive"] }
 clock = { path = "crates/common/clock" }
+chrono = "0.4"
```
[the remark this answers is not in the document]
PradeepKiruvale: I have updated the code to use `time`

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -241,28 +258,63 @@ mod tests {
...
             let health_message = Message::new(&health_topic, health_status);
             sender.publish(health_message).await?;
         }
 
-        let health_status = get_latest_health_status_message(3, &mut receiver).await;
-        assert_eq!(health_status.time, 3);
+        let health_status =
+            get_latest_health_status_message(base_timestamp_str.clone(), &mut receiver).await;
+
+        assert_eq!(health_status.unwrap().status, "up");
```
albinsuresh, line 284:
```suggestion
        assert_eq!(health_status.time, base_timestamp + Duration::seconds(3));
```
The `status` validation is pointless as it was `up` for every message that was generated.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -241,28 +258,63 @@ mod tests {
...
 
         let timeout_error = tokio::time::timeout(
             tokio::time::Duration::from_secs(1),
-            get_latest_health_status_message(5, &mut receiver),
+            get_latest_health_status_message(
+                "2050-06-22T17:03:14.000+02:00".to_string(),
+                &mut receiver,
+            ),
```
albinsuresh, line 291:
```suggestion
            get_latest_health_status_message(
                base_timestamp + Duration::seconds(5)
                &mut receiver,
            ),
```
Just in case you wanted the old behaviour. This is also fine.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -241,28 +258,63 @@ mod tests {
...
         )
         .await;
         assert!(timeout_error.is_err());
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn test_get_health_status_message() -> Result<()> {
```
albinsuresh, line 300: How is this test different from the previous test?

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -241,28 +258,63 @@ mod tests {
...
+        let (mut sender, mut receiver) = mpsc::unbounded::<Message>();
+
+        let health_message = health_status_up_message("test_service");
+        for _x in 1..5i64 {
+            sender.publish(health_message.clone()).await?;
+        }
+
+        let request_timestamp = OffsetDateTime::now_utc();
```
albinsuresh, line 308: I'm surprised that this works. This timestamp should have been created **before** sending the health messages, right? So that the generated ones would have had higher timestamps than this one and hence the `get_latest_health_status_message` will definitely return something?

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -241,28 +258,63 @@ mod tests {
...
+        let request_timestamp = request_timestamp
+            .format(&time::format_description::well_known::Rfc3339)
+            .unwrap();
+
+        let health_status =
+            get_latest_health_status_message(request_timestamp.to_string(), &mut receiver).await;
+
+        assert_eq!(&health_status.unwrap().status, "up");
```
albinsuresh, line 316: As mentioned earlier, this assertion is pointless as it is always going to be `true` for all the inputs.

albinsuresh (approved):
Approving with these follow-up tasks:

1. Let's move the `HealthStatus` struct into the `tedge_api::health` module as follows:

```
#[derive(Debug, Serialize, Deserialize, PartialEq, Eq, Clone)]
pub struct HealthStatus {
    status: String,
    pid: u32,
    #[serde(with = "time::serde::rfc3339")]
    pub time: OffsetDateTime,
}
```
with on-the-fly transformation of the `time` field.

Then this same struct can be consistently used across all the consumer crates like `converter`, `watchdog` etc.

2. Make `health_status_up_message()` API accept an additional `Option<OffsetDateTime>` parameter, so that the timestamp can be controlled from outside(mainly from tests).

Issue #2326, opened 2023-10-09
tedge-mapper-c8y stops converting message when receiving high volume of messages


reubenmiller:
**Describe the bug**

tedge-mapper stops mapping MQTT messages after high volume (e.g. 1000 single messages within ~210 milliseconds) without any warning. The `tedge-mapper-c8y` needs to be restarted in order to become responsive to mapping measurements again.

Reducing the rate of message (e.g. adding a small delay of ~2 milliseconds in between publishing messages) is enough to avoid this situation.

**To Reproduce**

1. Comment out the `c8y/measurement` bridge topic in `/etc/tedge/mosquitto-conf/c8y-bridge.conf` to prevent the messages from being forwarded to Cumulocity

    ```
    #topic measurement/measurements/create out 2 c8y/ ""
    ```

2. Restart the tedge-mapper-c8y service (in order to use the new edited bridge configuration)

    ```sh
    sudo systemctl restart tedge-mapper-c8y
    ```

2. Copy the [simulate.py](https://gist.github.com/reubenmiller/2461cd875765af24333f4165175a6a17) script to the device under test and install the dependencies

    Install the following dependencies (based on a Debian OS)

    * python3 >=3.9
    * python3-paho-mqtt (via `apt-get install python3-paho-mqtt`)

3. Execute script to simulate high load (e.g. publishing 150 messages sequentially as quick as possible)

    ```sh
    python3 simulate.py  --max-count 1000 --wait 0
    ```

    ```json
        {
        "worker": 0,
        "messages": 1000,
        "idle": 0.0,
        "total_non_idle": 0.210788,
        "total": 0.210788,
        "dropped_messages": 850,
        "qos": 0,
        "total_payload_bytes": 76890,
        "total_payload_bytes_per_message": 76.89,
        "ms_per_message": 0.211
    }
    ```

    In the error scenario, the `dropped_messages` property will be non-zero. This indicates that at some point measurements were not being published to the `c8y/measurement/measurements/create` topic anymore.

    Try increasing the `--max-count` value if you can't reproduce the error scenario.

4. Run the script again but only publishing 1 message

    ```
    python3 simulate.py  --max-count 1 --wait 0
    ```

    ```json
    {
        "worker": 0,
        "messages": 1,
        "idle": 0.0,
        "total_non_idle": 0.000811,
        "total": 0.000811,
        "dropped_messages": 1,
        "qos": 0,
        "total_payload_bytes": 75,
[...]

PR #2327, opened 2023-10-09
Enable concurrent processing of incoming and outgoing MQTT messages

jarhodes314:
## Proposed changes
If tedge-mapper is overwhelmed with messages, this causes a deadlock, preventing any further processing of MQTT messages until the mapper is restarted. The root cause of the deadlock is detailed in https://github.com/thin-edge/thin-edge.io/issues/2326#issuecomment-1753214566, but the pertinent information here is that the MQTT actor blocks itself, waiting on outgoing messages to be processed before it can process incoming messages, but stuck in the middle of processing an incoming message.

This PR spawns a task to process outgoing messages, which allows the mapper to process incoming and outgoing messages concurrently. This avoids the deadlock scenario that is the root cause of the bug.

Running this change locally, I can reliably send 20,000 messages to the mapper running on my laptop using the `simulate.py` script (see the [issue description](https://github.com/thin-edge/thin-edge.io/issues/2326)) without dropping any messages (previously  sending just 500 messages would reliably fail). Sending significantly more (e.g. 100,000 messages) will cause messages to be dropped, but the mapper does continue to run as normal once the pressure is relieved.

I haven't added any tests for this change. Verifying we don't deadlock depends on knowing the capacity of the machine we're running the tests on (and also involves hammering Cumulocity to some extent). But the existing tests should verify that the changes don't break anything (given all that's changed is where existing code is run, I would expect any bugs with the change to be really obvious).

## Types of changes

## Paste Link to the issue
#2326

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -62,7 +62,7 @@ impl Actor for C8yMapperActor {
     async fn run(&mut self) -> Result<(), RuntimeError> {
         let init_messages = self.converter.init_messages();
         for init_message in init_messages.into_iter() {
-            let _ = self.mqtt_publisher.send(init_message).await?;
+            self.mqtt_publisher.send(init_message).await?;
         }
 
         // Start the sync phase
@@ -106,7 +106,7 @@ impl C8yMapperActor {
         let converted_messages = self.converter.convert(&message).await;
 
         for converted_message in converted_messages.into_iter() {
-            let _ = self.mqtt_publisher.send(converted_message).await;
+            self.mqtt_publisher.send(converted_message).await?;
         }
 
         Ok(())
--- a/crates/extensions/tedge_mqtt_ext/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_ext/src/lib.rs
@@ -126,6 +126,14 @@ pub struct MqttMessageBox {
     peer_senders: Vec<(TopicFilter, LoggingSender<MqttMessage>)>,
 }
 
+pub struct FromPeers<'a> {
+    input_receiver: &'a mut LoggingReceiver<MqttMessage>,
+}
+
+pub struct ToPeers<'a> {
+    peer_senders: &'a mut Vec<(TopicFilter, LoggingSender<MqttMessage>)>,
+}
+
 impl MqttMessageBox {
     fn new(
         input_receiver: LoggingReceiver<MqttMessage>,
@@ -135,6 +143,43 @@ impl MqttMessageBox {
             input_receiver,
             peer_senders,
         }
+    }
+
+    fn split(&mut self) -> (FromPeers<'_>, ToPeers<'_>) {
+        (
+            FromPeers {
+                input_receiver: &mut self.input_receiver,
+            },
+            ToPeers {
+                peer_senders: &mut self.peer_senders,
+            },
+        )
+    }
+}
+
+impl<'a> FromPeers<'a> {
+    async fn relay_messages_to(
+        &mut self,
+        outgoing_mqtt: &mut mpsc::UnboundedSender<MqttMessage>,
+    ) -> Result<(), RuntimeError> {
+        loop {
+            match self.try_recv().await {
+                Ok(Some(message)) => outgoing_mqtt.send(message).await.map_err(Box::new)?,
+                Ok(None) | Err(RuntimeRequest::Shutdown) => break Ok(()),
+            }
+        }
+    }
+}
+
+impl<'a> ToPeers<'a> {
+    async fn relay_messages_from(
+        mut self,
+        incoming_mqtt: &mut mpsc::UnboundedReceiver<MqttMessage>,
+    ) -> Result<(), RuntimeError> {
+        while let Some(message) = incoming_mqtt.next().await {
+            self.send(message).await?;
+        }
+        Ok(())
     }
 
     async fn send(&mut self, message: MqttMessage) -> Result<(), ChannelError> {
@@ -148,7 +193,7 @@ impl MqttMessageBox {
 }
 
 #[async_trait]
-impl MessageReceiver<MqttMessage> for MqttMessageBox {
+impl<'a> MessageReceiver<MqttMessage> for FromPeers<'a> {
     async fn try_recv(&mut self) -> Result<Option<MqttMessage>, RuntimeRequest> {
         self.input_receiver.try_recv().await
     }
@@ -187,36 +232,22 @@ impl Actor for MqttActor {
     }
 
     async fn run(&mut self) -> Result<(), RuntimeError> {
+        let (mut from_peers, to_peers) = self.messages.split();
+
         let mut mqtt_client = tokio::select! {
             connection = mqtt_channel::Connection::new(&self.mqtt_config) => {
                 connection.map_err(Box::new)?
             }
-            Some(RuntimeRequest::Shutdown) = self.messages.recv_signal() => {
+            Some(RuntimeRequest::Shutdown) = from_peers.recv_signal() => {
                 // Shutdown requested even before the connection has been established
                 return Ok(())
             }
         };
 
-        loop {
-            tokio::select! {
-                message_or_signal = self.messages.try_recv() => {
-                    match message_or_signal {
-                        Ok(Some(message)) => {
-                            mqtt_client
-                                .published
-                                .send(message)
-                                .await
-                                .map_err(Box::new)?
-                        }
-                        Ok(None) | Err(RuntimeRequest::Shutdown) => break,
-                    }
-                }
-                Some(message) = mqtt_client.received.next() => {
-                    self.messages.send(message).await?
-                },
-                else => break,
-            }
-        }
-        Ok(())
+        tedge_utils::futures::select(
+            from_peers.relay_messages_to(&mut mqtt_client.published),
+            to_peers.relay_messages_from(&mut mqtt_client.received),
+        )
+        .await
     }
 }
--- a/tests/RobotFramework/tests/cumulocity/benchmarks/benchmark.py
+++ b/tests/RobotFramework/tests/cumulocity/benchmarks/benchmark.py
@@ -1,0 +1,675 @@
+#!/usr/bin/env python3
+"""thin-edge.io benchmark"""
+import argparse
+import datetime
+import json
+import logging
+import random
+import re
+import sys
+import shutil
+import subprocess
+import time
+from enum import Enum
+from itertools import cycle
+from multiprocessing import Pool
+import multiprocessing
+from pathlib import Path
+from typing import List, Union
+import paho.mqtt.client as mqtt
+
+# Set sensible logging defaults
+LOG = logging.getLogger()
+LOG.setLevel(logging.WARNING)
+handler = logging.StreamHandler()
+handler.setLevel(logging.WARNING)
+formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
+handler.setFormatter(formatter)
+LOG.addHandler(handler)
+
+
+class TelemetryType(Enum):
+    MEASUREMENT = "measurement"
+    ALARM = "alarm"
+    EVENT = "event"
+
+    def __str__(self) -> str:
+        return str(self.value)
+
+    @classmethod
+    def values(cls) -> List[str]:
+        return [item.value for item in cls]
+
+    @classmethod
+    def from_str(
+        cls, value: str, default_value: "TelemetryType" = None
+    ) -> "TelemetryType":
+        for item in cls:
+            if str(item.value).casefold() == value.casefold():
+                return item
+
+        return default_value
+
+
+class Pub:
+    def __init__(
+        self,
+        topic_id: str = "/",
+        topic_root: str = "te",
+        host: str = "127.0.0.1",
+        port: int = 1883,
+        qos: int = 0,
+        count: int = 5,
+        type_name: str = "",
+        datapoints: int = 3,
+        beats: int = 1,
+        beats_delay: float = 0,
+        period: float = 0,
+        telemetry_type: TelemetryType = TelemetryType.MEASUREMENT,
+    ) -> None:
+        self._topic_root = topic_root
+        self._topic_id = topic_id
+        self.type_name = type_name
+        self.qos = qos
+        self.count = count
+        self.host = host
+        self.port = port
+
+        self.beats = beats
+        self.beats_delay = beats_delay
+        self.period = period
+
+        self.datapoints = datapoints
+        self.telemetry_type = telemetry_type
+
+        self.__ready = False
+        self.__c8y_messages = []
+        self.start_time = None
+
+    @property
+    def topic_prefix(self) -> str:
+        return "/".join([self._topic_root, self._topic_id])
+
+    def get_topic(self, telemetry_type: TelemetryType, type_name: str) -> str:
+        sep = "/"
+        telemetry_topics = {
+            TelemetryType.MEASUREMENT: "m",
+            TelemetryType.EVENT: "e",
+            TelemetryType.ALARM: "a",
+        }
+        return sep.join(
+            [self.topic_prefix, telemetry_topics[self.telemetry_type], type_name]
+        )
+
+    def __on_connect(self, client: mqtt.Client, userdata, flags, rc):
+        if rc != 0:
+            raise Exception("Failed to connect to broker")
+
+        # observe which messages are being sent
+        LOG.info("Subscribing to cloud topic")
+        topics = {
+            TelemetryType.MEASUREMENT: "c8y/measurement/measurements/create",
+            TelemetryType.ALARM: "c8y/alarm/alarms/create",
+            TelemetryType.EVENT: "c8y/event/events/create",
+        }
+        topic = topics[self.telemetry_type]
+        client.subscribe(
+            [
+                (topic, 0),
+            ]
+        )
+        self.__ready = True
+
+    def __on_message(self, _client, _userdata, msg: mqtt.MQTTMessage):
+        try:
+            LOG.debug("Received message: %s", msg.payload)
+            payload = json.loads(msg.payload.decode("utf-8"))
+            if msg.topic == "c8y/measurement/measurements/create":
+                if "msgid" in payload:
+                    self.__c8y_messages.append(payload)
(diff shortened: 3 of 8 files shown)


reubenmiller:
@jarhodes314 Running the same benchmarks with this PR fixes the bug. I'll spend a bit of time to see if I can add reasonable system test to the PR, though as you already mentioned, it might be tricky to "stable" benchmark that will run on every machine.

tests/RobotFramework/tests/cumulocity/benchmarks/benchmark.py
```diff
@@ -0,0 +1,503 @@
...
+    parser.add_argument("--pub-clients", dest="pub_clients", default=1, type=int)
+
+    parser.add_argument("--qos", default=0, type=int)
+    parser.add_argument("--max-datapoints", dest="max_datapoints", default=1, type=int)
+
+    parser.add_argument(
+        "--count",
+        default=[50],
+        type=flag_range,
+        help="Number of messages to send in one iteration",
```
albinsuresh, line 308:
```suggestion
        help="Total number of messages to send in one iteration as multiple bursts",
```

albinsuresh (approved):
The proposal to asynchronously process the MQTT messages make sense. My only concern is about the relevance of the `MqttMessageBox` struct having both the `send` and `recv` aspects together, as the new design is using those aspects separately from two different contexts anway. So, instead an `MqttMessageBox`, the actor could just keep one `SimpleMessageBox` to receive the peer messages and another wrapper over the `peer_senders`, with that specialized `send()` method in the existing `MqttMessageBox` impl, as part of its state. But, that refactoring feels a bit out of scope for this PR, hence approving. I've left some minor suggestions/queries though.

tests/RobotFramework/tests/cumulocity/benchmarks/benchmark.py
reubenmiller: Yes, the api might need some cleaning up, however it is still only an internal benchmarking script so that we have coverage...though the tests seem to be a bit flakey as I don't think we can guarantee the performance on the CI runner.

jarhodes314:
> The proposal to asynchronously process the MQTT messages make sense. My only concern is about the relevance of the `MqttMessageBox` struct having both the `send` and `recv` aspects together, as the new design is using those aspects separately from two different contexts anway. So, instead an `MqttMessageBox`, the actor could just keep one `SimpleMessageBox` to receive the peer messages and another wrapper over the `peer_senders`, with that specialized `send()` method in the existing `MqttMessageBox` impl, as part of its state. But, that refactoring feels a bit out of scope for this PR, hence approving. I've left some minor suggestions/queries though.

I agree that that refactoring away `MqttMessageBox` entirely is too complicated to be worth blocking the fix for this. I do wonder whether instead the internal use of `MqttMessageBox` could be refactored so we can destructure it into incoming and outgoing parts before we start sending and receiving messages as neither part needs to care about the other.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -61,7 +61,7 @@ impl Actor for C8yMapperActor {
     async fn run(&mut self) -> Result<(), RuntimeError> {
         let init_messages = self.converter.init_messages();
         for init_message in init_messages.into_iter() {
-            let _ = self.mqtt_publisher.send(init_message).await?;
+            self.mqtt_publisher.send(init_message).await?;
```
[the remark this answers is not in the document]
jarhodes314: It's a really bad practice actually. If you include `let _ = `, it will stop rust from generating any warnings about the return value (as you're explicitly allowing it to be anything), so if the `?` is removed, you'll never get warned about the non-handling of the error. You should only be using `let _ =` when you want to explicitly ignore an error.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -105,7 +105,7 @@ impl C8yMapperActor {
         let converted_messages = self.converter.convert(&message).await;
 
         for converted_message in converted_messages.into_iter() {
-            let _ = self.mqtt_publisher.send(converted_message).await;
+            self.mqtt_publisher.send(converted_message).await?;
```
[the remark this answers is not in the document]
jarhodes314: See my comment on the other similar example above; this is why you don't want `let _ =` unless you explicitly do need it.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -211,10 +235,7 @@ impl Actor for MqttActor {
                         Ok(None) | Err(RuntimeRequest::Shutdown) => break,
                     }
                 }
-                Some(message) = mqtt_client.received.next() => {
-                    self.messages.send(message).await?
-                },
-                else => break,
```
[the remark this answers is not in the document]
jarhodes314: The else arm has been squashed into the task inside `tokio::spawn` actually, as this only affects the outgoing sender. `else` in `tokio::select` is called when all the other branches are disabled (i.e. we've received values for the awaited futures, but we none of them matched the provided patterns). Given the first pattern is just a variable and will therefore match anything, only `Some(message) = mqtt_client.received.next()` can be disabled, and this occurs when calling `.next()` returns `None`, which happens when `mqtt_client.received` is closed.

Without changing what the original code does (but absolutely changing what happens if we were to add a new variant to `RuntimeRequest`, which would be treated like shutdown without any warning), we could change it to the following to make full use of `else`.

```rust
tokio::select! {
    Ok(Some(message)) = self.messages.try_recv() => {
        mqtt_client
            .published
            .send(message)
            .await
            .map_err(Box::new)?
    }
    Some(message) = mqtt_client.received.next() => {
        self.messages.send(message).await?
    },
    else => break,
}
```

In general though, avoiding pattern matching at the `tokio::select` level is a useful thing to do, as like in this example, there are many cases where it work like match when initially written, but then won't work if changes are made to the code surrounding it. Notably, if we had something like 

```rust
tokio::select! {
    RuntimeRequest::Shutdown => self.signals.next(),
    // more cases
}

enum RuntimeRequest {
    Shutdown,

    // Suppose we then add a future variant like
    AbsolutelyDoNotShutdownJustAcknowledgeIExist,
}
```

`tokio::select` will accept this, but when we receive a `RuntimeRequest::AbsolutelyDoNotShutdownJustAcknowledgeIExist`, the `self.signals.next()` branch has failed to match, which just causes that branch to block forever, and we wouldn't receive any shutdown signals. If however we were doing this in a `match` on `self.signals.next().await`, this absolutely wouldn't compile until we handled both `RuntimeRequest` values (this is the argument in favour of keeping the internal match for the first `tokio::select` branch). When you have a bunch of branches and some of them are failing for this reason, it can be really quite tricky to notice if you're not careful.

In general, I think `tokio::select` is quite a lot riskier than thin-edge gives it credit for (it's also very useful, mind you). It looks similar to a `match`, but where `match` generates a compiler error, `tokio::select` just blocks (one possible code path), which is very likely not the desired behaviour.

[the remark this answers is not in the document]
albinsuresh: Thanks for that detailed explanation.

> which just causes that branch to block forever

That's definitely concerning, as we have such pattern matching `tokio::select!` matching patterns in a lot more places and I'm not sure if we have the `else` fallback everywhere to avoid such "blocking" issues.

> If however we were doing this in a match on self.signals.next().await, this absolutely wouldn't compile until we handled both RuntimeRequest values

I'm assuming that's the recommended way to do it. Just do `tokio:select!` first and inside the processing arm, `match` for sub-variants, rather than pushing those matches to the top level.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -197,6 +197,30 @@ impl Actor for MqttActor {
             }
         };
 
+        let (tx, mut outgoing_messages_finished) = tokio::sync::mpsc::channel(1);
```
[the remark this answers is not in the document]
jarhodes314: They are outgoing messages in my mind because this bit concerns translating an internal actor message into an MQTT message, and the MQTT direction is the one that matters?

crates/extensions/tedge_mqtt_ext/src/lib.rs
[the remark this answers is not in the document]
jarhodes314: > That's definitely concerning, as we have such pattern matching tokio::select! matching patterns in a lot more places and I'm not sure if we have the else fallback everywhere to avoid such "blocking" issues.

`else` will stop you blocking when you cannot feasibly make any more progress, but if there's a branch which can make progress, you can still have issues, and they can be pretty subtle.

> I'm assuming that's the recommended way to do it. Just do `tokio:select!` first and inside the processing arm, `match` for sub-variants, rather than pushing those matches to the top level.

Yeah I think so. I would consider breaking the functionality down into separate async functions/blocks and using [`futures::future::select_all`](https://docs.rs/futures/latest/futures/future/fn.select_all.html), although I think this tends to interact badly with mutable borrows.

jarhodes314:
I've now done an internal refactoring that includes splitting the `MqttMessageBox` into an `InternalInbox` and an `InternalOutbox` allowing the two to be processed completely separately, avoiding the need to spawn a task.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -168,14 +215,14 @@ impl MessageReceiver<MqttMessage> for MqttMessageBox {
 
 pub struct MqttActor {
     mqtt_config: mqtt_channel::Config,
-    messages: MqttMessageBox,
+    messages: Option<MqttMessageBox>,
```
jarhodes314, line 218: It's annoying that this is required, but we need ownership of the MqttMessageBox to split it into chunks, but `run` takes `&mut self`. If `run` is called multiple times, this will cause a panic, although we don't currently ever call run multiple times.

I definitely won't do it in this PR, but it seems reasonable to modify `Actor::run` to take `self` instead of `&mut self` to avoid this issue (and clarify that `run` is terminal, which I think is assumed in a bunch of implementations). From a quick attempt at doing this, it seems that this change causes issues when calling `run` on a `Box<dyn Actor>` (which I'm guessing is the reason it currently takes `&mut self`, but I've found [a mitigation for this problem](https://users.rust-lang.org/t/call-consuming-method-for-dyn-trait-object/69596/7), so I'm planning to create a PR that makes this change).

tests/RobotFramework/tests/cumulocity/benchmarks/benchmark.py
reubenmiller: I've cleaned up the script now. Any further improvements to the script can be covered by another PR.

tests/RobotFramework/tests/cumulocity/benchmarks/benchmark.py
```diff
@@ -0,0 +1,503 @@
...
+    def get_topic(self, sub_path: str) -> str:
+        sep = "/"
+        return sep.join([self.topic_prefix, sub_path.lstrip(sep)])
+
+    def __on_connect(self, client: mqtt.Client, userdata, flags, rc):
+        if rc != 0:
+            raise Exception("Failed to connect to broker")
+
+        # observe which messages are being sent
+        LOG.info("Subscribing to cloud measurements")
```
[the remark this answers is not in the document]
reubenmiller: I've updated the log message to be more generic.

tests/RobotFramework/tests/cumulocity/benchmarks/benchmark.py
```diff
@@ -0,0 +1,503 @@
...
+        self.host = host
+        self.port = port
+        self.timeout = timeout
+
+        self.beats = beats
+        self.beats_delay = beats_delay
+        self.period = period
+
+        self.max_datapoints = max_datapoints
+        self.payload_type = TelemetryType.MEASUREMENT
```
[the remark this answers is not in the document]
reubenmiller: I've added full support for sending either measurements, events or alarms.

reubenmiller (approved):
Approved. I ran the new benchmark system tests 5 times and they seem to be ok.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -187,36 +234,30 @@ impl Actor for MqttActor {
     }
 
     async fn run(&mut self) -> Result<(), RuntimeError> {
+        let (mut inbox, outbox) = self
```
albinsuresh, line 237:
```suggestion
        let (mut peers_to_mqtt, mqtt_to_peers) = self
```
I find the names `inbox` and `outbox` a bit confusing. From this actor's perspective, it receives inputs from 2 different sources: The MQTT broker and peer actors. Similarly, it pushes its output also to both of them. So when you call something an `inbox`, I'm always asking in my head, is it the `inbox` for the MQTT messages from the broker or the the `inbox` for MQTT messages from peers (NOTE: it's the same `MqttMessage` type from/to both sources). Similarly I question `outbox` as well. The problem is that the words "in" and "out" are vague here: "In" from peers or "in" from broker.

Here, the `inbox` reads from the actors and forwards to MQTT and `outbox` reads from MQTT and forwards that to peer actors.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -137,6 +147,43 @@ impl MqttMessageBox {
...
+            },
+            InternalOutbox {
+                peer_senders: self.peer_senders,
+            },
+        )
+    }
+}
+
+impl InternalInbox {
+    async fn forward_to_mqtt(
```
albinsuresh, line 163:
```suggestion
    async fn route_peer_messages_to_mqtt
```

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -187,36 +234,30 @@ impl Actor for MqttActor {
...
-                },
-                else => break,
-            }
+        pin! {
+            let send_mqtt = inbox.forward_to_mqtt(&mut mqtt_client.published);
+            let receive_mqtt = outbox.forward_to_actors(&mut mqtt_client.received);
         }
-        Ok(())
+
+        futures::future::select(send_mqtt, receive_mqtt)
```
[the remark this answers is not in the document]
jarhodes314: Yeah, I think we can get rid of the pinning requirement by creating our own helper function. The reason why select needs the futures to be pinned is because it returns the non completed future, but we don't care in this case.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: I've renamed this, but tried to limit the number of times we mention peers and mqtt (with all of the suggestions, the main usage will involve mentioning both at least twice. Also `peers_to_mqtt` isn't quite true; in one place we use it just to receive the shutdown signal, which isn't "to mqtt". As a result, I've called these `from_peers` and `to _peers`

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -126,6 +126,14 @@ pub struct MqttMessageBox {
     peer_senders: Vec<(TopicFilter, LoggingSender<MqttMessage>)>,
 }
 
+pub struct FromPeers<'a> {
+    input_receiver: &'a mut LoggingReceiver<MqttMessage>,
+}
+
+pub struct ToPeers<'a> {
+    peer_senders: &'a mut Vec<(TopicFilter, LoggingSender<MqttMessage>)>,
```
albinsuresh, line 134: We could have just cloned that `Vec` to avoid these lifetime parameters, but its' okay, as it is consistent with its counterpart.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -137,6 +145,43 @@ impl MqttMessageBox {
...
+            },
+            ToPeers {
+                peer_senders: &mut self.peer_senders,
+            },
+        )
+    }
+}
+
+impl<'a> FromPeers<'a> {
+    async fn relay_messages_to(
```
albinsuresh, line 161:
```suggestion
    async fn relay_peer_messages_to(
```

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -137,6 +145,43 @@ impl MqttMessageBox {
...
+            match self.try_recv().await {
+                Ok(Some(message)) => outgoing_mqtt.send(message).await.map_err(Box::new)?,
+                Ok(None) | Err(RuntimeRequest::Shutdown) => break Ok(()),
+            }
+        }
+    }
+}
+
+impl<'a> ToPeers<'a> {
+    async fn relay_messages_from(
```
albinsuresh, line 175:
```suggestion
    async fn relay_mqtt_messages_from(
```

albinsuresh (approved):

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -187,36 +232,22 @@ impl Actor for MqttActor {
     }
 
     async fn run(&mut self) -> Result<(), RuntimeError> {
+        let (mut from_peers, to_peers) = self.messages.split();
```
albinsuresh, line 235: `from_peers, to_peers`: That's much better. Short and clearer.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: I don't think the lifetimes are that complicated, and I'll modify the `run(self)` PR to remove these once this is merged as they won't be necessary after that.

crates/extensions/c8y_mapper_ext/src/actor.rs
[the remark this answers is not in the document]
didier-wenzek: Unfortunately, we have many of those.

```
thin-edge.io git:(main) ✗ rg 'let _ =' | wc -l           
113
```

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -137,6 +145,43 @@ impl MqttMessageBox {
         }
     }
 
+    fn split(&mut self) -> (FromPeers<'_>, ToPeers<'_>) {
```
didier-wenzek, line 148: This `split` method as well as the whole `struct MqttMessageBox` can be removed, as they are used only by the `MqttActor`.

PR #2329, opened 2023-10-10
Modify Actor trait to take `self` instead of `&mut self`

jarhodes314:
## Proposed changes
In #2327, some changes have been made that assume we only run `MqttActor::run` once. As far as I can tell, we only ever want to run actors once, but some complexity around calling consuming methods (i.e. those which take an owned `self` argument) from `Box<dyn Actor>` prevented us from doing that. I've managed to work around this and modified `Actor::run` to receive `self` instead of `&mut self`.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_actors/src/actors.rs
+++ b/crates/core/tedge_actors/src/actors.rs
@@ -16,7 +16,7 @@ use async_trait::async_trait;
 ///   - sending responses for requests, possibly deferring some responses,
 ///   - acting as a source of messages ...
 #[async_trait]
-pub trait Actor: 'static + Send + Sync {
+pub trait Actor: 'static + Send + Sync + ActorBoxed {
     /// Return the actor instance name
     fn name(&self) -> &str;
 
@@ -25,7 +25,24 @@ pub trait Actor: 'static + Send + Sync {
     /// Processing input messages,
     /// updating internal state,
     /// and sending messages to peers.
-    async fn run(&mut self) -> Result<(), RuntimeError>;
+    async fn run(self) -> Result<(), RuntimeError>;
+}
+
+// The following madness comes from
+// https://users.rust-lang.org/t/call-consuming-method-for-dyn-trait-object/69596/8
+#[async_trait]
+pub trait ActorBoxed {
+    async fn run_boxed(self: Box<Self>) -> Result<(), RuntimeError>;
+}
+
+#[async_trait]
+impl<A> ActorBoxed for A
+where
+    A: Actor,
+{
+    async fn run_boxed(self: Box<Self>) -> Result<(), RuntimeError> {
+        (*self).run().await
+    }
 }
 
 #[cfg(test)]
@@ -48,7 +65,7 @@ pub mod tests {
             "Echo"
         }
 
-        async fn run(&mut self) -> Result<(), RuntimeError> {
+        async fn run(mut self) -> Result<(), RuntimeError> {
             while let Some(message) = self.messages.recv().await {
                 self.messages.send(message).await?
             }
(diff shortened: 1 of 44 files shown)


didier-wenzek:
Nice, to see this move. I can only encourage you to move this forward.

> As far as I can tell, we only ever want to run actors once, but some complexity around calling consuming methods (i.e. those which take an owned self argument) from Box<dyn Actor> prevented us from doing that.

I confirm that we only want to run actors once.

> I've managed to work around this and modified Actor::run to receive self instead of &mut self.

Running an actor as `self` was indeed the original design, but has been changed to `&mut self` for reasons that no more apply as meantime the message box design has been simplified.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -145,19 +145,19 @@ impl MqttMessageBox {
         }
     }
 
-    fn split(&mut self) -> (FromPeers<'_>, ToPeers<'_>) {
+    fn split(self) -> (FromPeers, ToPeers) {
```
didier-wenzek, line 148: Both `fn slip` and `struct MqttMessageBox` can be deprecated as built & split only here.

didier-wenzek (approved):
Approved. This really clarify the intent of `Actor::run()`.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: I don't quite understand what you mean here. I introduced `split` as a mechanism to enable concurrent processing of incoming and outgoing messages, without the need for other actors to know/care about this detail (so outside there is still a single message box for a single actor). `MqttActor` could potentially be split into two actors to deal with this, but that feels like it's just going to increase complexity in all the MQTT interactions for other actors.

didier-wenzek: I meant that `MqttActor` doesn't need to define a message box as a single object. In fact no actor has to. The abstraction with other actors is implemented by the builder with appropriate  `ServiceProvider` / `MessageSource` implementation.

The following is working fine:

```
pub struct MqttActor {
    mqtt_config: mqtt_channel::Config,
    from_peers: FromPeers,
    to_peers: ToPeers,
}
```

didier-wenzek: I pushed a fixup commit: https://github.com/thin-edge/thin-edge.io/pull/2329/commits/cc067c1220c0d696d8d4ceb27dac6ac08e8adaf9

crates/core/tedge_actors/src/actors.rs
```diff
@@ -25,7 +25,24 @@ pub trait Actor: 'static + Send + Sync {
     /// Processing input messages,
     /// updating internal state,
     /// and sending messages to peers.
-    async fn run(&mut self) -> Result<(), RuntimeError>;
+    async fn run(self) -> Result<(), RuntimeError>;
```
albinsuresh, line 28: I was wondering why you didn't go with the simpler fix of `self: Box<Self>` (suggested in that same forum discussion) here instead. I'm assuming you did that to make the `run` method implemented by the actors look more natural and hide the complexity of `Box<Self>` with that blanket impl of `ActorBoxed` below.

albinsuresh (approved):
Approved. Surprised to see that even the changes required in the `runtime` module were so minimal.

crates/core/tedge_actors/src/actors.rs
jarhodes314: The existing API works well for our use case, and I really don't like tests polluted with boxing values simply because the actual code uses dynamic dispatch somewhere.

`Box<Self>` is very much an implementation detail here, and I want to be clear about that, particularly given the main change that's actually being made here is to the communication of what `run` means and what assumptions can be made when implementing it, which up until now have been implicit. Receiving `Box<Self>` doesn't tell me anything useful as an implementor of the trait, and because it's pretty uncommon thing, it's likely to cause some confusion among people.

PR #2331, opened 2023-10-10
Test improvements

jarhodes314:
## Proposed changes
This makes a few test improvements. One of them concerns running the tests under `cargo-nextest`, which fails on some MQTT tests which spin up a server on a random port due to a conflict. The other two are changes to fix the tests on my machine, as they assume I have a group with the same name as my username, and assume a specific error messages for DNS errors (they expect `dns error: Failed to lookup address information`, but my machine gives the error `dns error: Device or resource busy`).

For simplicity, I've assumed there is always a group called `root`, but I think this a more reliable assumption than a group sharing a name with the current username.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/file.rs
+++ b/crates/common/tedge_utils/src/file.rs
@@ -484,17 +484,25 @@ pub fn has_write_access(path: &Path) -> Result<(), FileError> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use once_cell::sync::Lazy;
     use std::os::unix::fs::PermissionsExt;
     use std::path::Path;
     use tempfile::TempDir;
+
+    static USER: Lazy<String> = Lazy::new(whoami::username);
+    static GROUP: Lazy<String> = Lazy::new(|| {
+        uzers::get_current_groupname()
+            .unwrap()
+            .into_string()
+            .unwrap()
+    });
 
     #[test]
     fn create_file_correct_user_group() {
         let temp_dir = TempDir::new().unwrap();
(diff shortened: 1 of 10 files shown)


crates/common/tedge_utils/src/file.rs
```diff
@@ -488,7 +488,7 @@ mod tests {
         let file_path = temp_dir.path().join("file").display().to_string();
 
         let user = whoami::username();
-        create_file_with_user_group(&file_path, &user, &user, 0o644, None).unwrap();
+        create_file_with_user_group(&file_path, &user, "root", 0o644, None).unwrap();
```
Bravo555, line 491: Why not just use the groupname of the user who launched the process?
```suggestion
        let group = uzers::get_current_groupname().unwrap();
        let group = group.to_str().unwrap();
        create_file_with_user_group(&file_path, &user, group, 0o644, None).unwrap();
```

jarhodes314: I didn't realise we already had a dependency that made getting that information easy, I'll change it

justfile
```diff
@@ -62,7 +62,8 @@ release *ARGS:
 
 # Run unit tests
 test:
-    cargo test --no-fail-fast --all-features --all-targets
+    cargo nextest run --no-fail-fast --all-features --all-targets
+    cargo test --doc --no-fail-fast --all-features
```
didier-wenzek, line 66: Indeed, a trap I fall into: `nextest` skips doc tests.

didier-wenzek (approved):
Approved

Issue #1971, opened 2023-05-15
Support publishing service telemetry data


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

If a service health endpoint has been registered to thin-edge.io via the `tedge/health/{service_name}` endpoint, sending measurements to `tedge/measurements/{service_name}` should not result in the creation of a child device in Cumulocity.

**Describe the solution you'd like**

A dedicated MQTT topics which allows services to publish Inventory, Measurements, Events and Alarms. This would allow services to publish additional information about the status of the service, e.g. CPU, RAM usage, additional running meta info (e.g. process ids, run command).

The new topic names could be called.

```sh
tedge/health/{service}/measurements
tedge/health/{service}/events
tedge/health/{service}/alarms
tedge/health/{service}/inventory
```

**Describe alternatives you've considered**

Alternatively, when the `tedge-mapper-c8y` received a message on the `tedge/measurement/{service_name}` topic, it could check if the name has been registered via the `tedge/health/{service_name}` endpoint, and skip register a new child logic (e.g. skip sending the topic=`c8y/s/us` payload=`101,{service_name},{service_name},thin-edge.io-child`

**Additional context**

Cumulocity IoT has two different smart rest messages which involve the registration/creation of a linked entity.

|Type|SmartREST Template ID|
|---|---|
|Create a child device|[101](https://cumulocity.com/guides/reference/smartrest-two/#101)|
|Create a service|[102](https://cumulocity.com/guides/reference/smartrest-two/#service-creation-102)|

Both SmartREST message templates will create a managed object with the given External Identity, however `101` will link the created managed object as a `childDevice`, where as the `102` message will link the device as a `childAddition`. The different types of managed object links control how the managed object is treated in the UI (and other api for that matter).

albinsuresh:
The registration of services has already been done in #2266 and the entity store now has the external ID of the services as well. So, all that is left is to send the telemetry data to that external ID, just like it's done for devices, using their external ID. For example, a measurement for a service with the external ID `myDatabaseService` would be as follows:

```json
{
    "externalSource":{
        "externalId":"myDatabaseService",
        "type":"c8y_Serial"
    },
    "type": "c8y_TemperatureMeasurement",
    "time": "2023-09-29T17:57:00+05:30",
    "c8y_TemperatureMeasurement": {
        "T": {
            "value": 20,
            "unit": "C"
        }
    }
}
```

[...]

PR #2335, opened 2023-10-11
send measurement to main/child device services

PradeepKiruvale:
## Proposed changes

Add a feature to push measurements to the child and main device services using the new `te` topic.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/1971

## Checklist


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1663,6 +1663,87 @@ mod tests {
...
+            vec![
+                expected_smart_rest_message_service,
+                expected_c8y_json_message.clone()
+            ]
+        );
+
+        let out_second_messages = converter.convert(&in_message).await;
+        assert_eq!(out_second_messages, vec![expected_c8y_json_message]);
+    }
+
```
albinsuresh, line 1746: Could you add some tests for events and alarms of services as well? They were working magically already, but we never had test coverage. The existing child device tests for the same could be copied and modified.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1663,6 +1663,87 @@ mod tests {
...
+            &Topic::new_unchecked("c8y/s/us"),
+            "101,test-device:device:child1,child1,thin-edge.io-child",
+        );
+        let expected_smart_rest_message_service = Message::new(
+            &Topic::new_unchecked("c8y/s/us/test-device:device:child1"),
+            "102,test-device:device:child1:service:app1,systemd,app1,up",
+        );
+        let expected_c8y_json_message = Message::new(
+            &Topic::new_unchecked("c8y/measurement/measurements/create"),
+            r#"{"externalSource":{"externalId":"test-device:device:child1:service:app1","type":"c8y_Serial"},"temp":{"temp":{"value":1.0}},"time":"2021-11-16T17:45:40.571760714+01:00","type":"m_type"}"#,
```
albinsuresh, line 1685: I'd strongly recommend the `json!` macro to wrap such complex JSONs. I'd use it even for simpler ones with more than a couple of keys. I hope the macro doesn't change the order of keys so that your assertions work seamlessly, without needing `assert_json_matches!`

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1663,6 +1663,87 @@ mod tests {
...
+        let tmp_dir = TempTedgeDir::new();
+        let (mut converter, _http_proxy) = create_c8y_converter(&tmp_dir).await;
+
+        let in_topic = "te/device/main/service/appm/m/m_type";
+        let in_payload = r#"{"temp": 1, "time": "2021-11-16T17:45:40.571760714+01:00"}"#;
+        let in_message = Message::new(&Topic::new_unchecked(in_topic), in_payload);
+
+        let expected_c8y_json_message = Message::new(
+            &Topic::new_unchecked("c8y/measurement/measurements/create"),
+            r#"{"externalSource":{"externalId":"test-device:device:main:service:appm","type":"c8y_Serial"},"temp":{"temp":{"value":1.0}},"time":"2021-11-16T17:45:40.571760714+01:00","type":"m_type"}"#,
```
albinsuresh, line 1720: Suggest to use `json!`.

crates/extensions/c8y_mapper_ext/src/serializer.rs
```diff
@@ -58,16 +58,16 @@ impl C8yJsonSerializer {
 
         json.write_open_obj();
 
-        if entity.r#type == EntityType::ChildDevice {
-            let child_id = &entity.external_id;
-            // In case the measurement is addressed to a child-device use fragment
-            // "externalSource" to tell c8Y identity API to use child-device
+        if entity.r#type == EntityType::ChildDevice || entity.r#type == EntityType::Service {
+            let entity_id = &entity.external_id;
+            // In case the measurement is addressed to a child-device or for the service use fragment
```
albinsuresh, line 63:
```suggestion
            // In case the measurement is addressed to a child-device or a service, use fragment
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1663,6 +1663,87 @@ mod tests {
...
+                expected_smart_rest_message_service,
+                expected_c8y_json_message.clone()
+            ]
+        );
+
+        // Test the second output messages doesn't contain SmartREST child device creation.
+        let out_second_messages = converter.convert(&in_message).await;
+        assert_eq!(out_second_messages, vec![expected_c8y_json_message]);
+    }
+
```
albinsuresh, line 1708: Could you add one similar test for a service on a nested child device as well? Just to complete the coverage.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
```diff
@@ -79,6 +79,70 @@ Child device supports sending custom child device measurements directly to c8y
...
+
+Send events to a registered child service
+    Execute Command    tedge mqtt pub --retain te/device/${CHILD_SN}/service/app6 '{"@type":"service","@parent":"device/${CHILD_SN}//"}'
+    Cumulocity.Device Should Exist    ${CHILD_SN}
+    Cumulocity.Should Have Services    name=app6    min_count=1    max_count=1
+
+    Cumulocity.Device Should Exist    ${DEVICE_SN}:device:${CHILD_SN}:service:app6
+    Execute Command    tedge mqtt pub te/device/${CHILD_SN}/service/app6/e/event_002 '{"text": "test event"}'
+    Device Should Have Event/s    expected_text=test event    type=event_002    minimum=1    maximum=1
+
```
albinsuresh, line 145: You may want to add similar tests for services on nested child devices as well. The "unregistered nested child service" case is not even needed as nested child devices must have explicit registration.

crates/extensions/c8y_mapper_ext/src/serializer.rs
PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: fixed, using `json!` and `to_string` used, so don't need to use `assert_json_matches`.

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: added in tests.rs

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
PradeepKiruvale: I will do it as part of another PR as there are unit tests are also missing.

reubenmiller: No we should include the tests here as well so that we are sure the code supports it before we merge it.

PradeepKiruvale: Since that is not in the scope of this task, I do not want to bloat this task. The nested child devices are a completely different thing, I would do it as part of another PR because this needs Unit tests and integration tests.

One question: can nested child devices also have the services? they need telemetry support as well right?

reubenmiller: No I would disagree, it is very much in the task as a child device also reference to any child device regardless if it is an immediate child or a nested child.

And generally we should get in the habit of ensuring the code does what it is meant to before it gets merged.

reubenmiller: > One question: can nested child devices also have the services? they need telemetry support as well right?

Yes, services are supported on any type of child device (immediate or nested)

PradeepKiruvale: ok, thanks for the clarification. I will add the tests.

PradeepKiruvale: ok, thanks for the clarification. I will update the tests.

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: added

albinsuresh (commented):
Happy to approve after a little bit of cleanup.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
```diff
@@ -79,6 +77,129 @@ Child device supports sending custom child device measurements directly to c8y
...
+Send events to a registered child service
+    Execute Command    tedge mqtt pub --retain te/device/${CHILD_SN}/service/app6 '{"@type":"service","@parent":"device/${CHILD_SN}//"}'
+    Cumulocity.Device Should Exist    ${CHILD_SN}
+    Cumulocity.Should Have Services    name=app6    min_count=1    max_count=1
+    Cumulocity.Device Should Exist    ${DEVICE_SN}:device:${CHILD_SN}:service:app6
+    Execute Command    tedge mqtt pub te/device/${CHILD_SN}/service/app6/e/event_002 '{"text": "test event"}'
+    Device Should Have Event/s    expected_text=test event    type=event_002    minimum=1    maximum=1
+  
+# Nested child devices
+Nested child devices support sending simple measurements
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2198,6 +2540,244 @@ async fn handle_log_upload_successful_cmd_for_child_device() {
...
+                "type":"ThinEdgeEvent",
+                "time":"2023-10-13T15:00:07.172674353Z",
+                "text":"Temperature high",
+                "externalSource":{"externalId":"nested_service","type":"c8y_Serial"}
+            }),
+        )],
+    )
+    .await;
+}
+
```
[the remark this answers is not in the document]
PradeepKiruvale: Sure, then there were no `assert_messages_matching` APIs, it was tough to assert because of the order issues in the payload. Maybe we can do it as a part of a follow-up task. There are a lot of tests that can be ported.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,224 @@ mod tests {
         assert_eq!(out_second_messages, vec![expected_c8y_json_message]);
     }
 
+    #[tokio::test]
+    async fn convert_measurement_with_nested_child_id() {
```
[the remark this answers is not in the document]
PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,224 @@ mod tests {
...
+            .filter(|m| m.topic.name.starts_with("c8y"))
+            .collect();
+        assert_eq!(
+            out_first_messages,
+            vec![
+                expected_smart_rest_message_child,
+                expected_smart_rest_message_service,
+                expected_c8y_json_message.clone()
+            ]
+        );
```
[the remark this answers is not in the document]
PradeepKiruvale: That can be done in the second phase as a followup task, as there are a lot of tests need to be ported.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,224 @@ mod tests {
...
+            })
+            .to_string(),
+        );
+
+        // Test the first output messages contains SmartREST and C8Y JSON.
+        let out_first_messages: Vec<_> = converter
+            .convert(&in_message)
+            .await
+            .into_iter()
+            .filter(|m| m.topic.name.starts_with("c8y"))
```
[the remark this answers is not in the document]
PradeepKiruvale: If not filtered all the messages will get accumulated. And the assert will fail. For this particular test, the purpose of the tests is to assert only the output messages. So, I feel filtering is OK.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2198,6 +2540,244 @@ async fn handle_log_upload_successful_cmd_for_child_device() {
...
+    mqtt.send(reg_message).await.unwrap();
+
+    mqtt.send(MqttMessage::new(
+        &Topic::new_unchecked("te/device/nested_child///a/"),
+        json!({ "severity": "minor", "text": "Temperature high","time":"2023-10-13T15:00:07.172674353Z" }).to_string(),
+    ))
+    .await
+    .unwrap();
+
+    mqtt.skip(2).await;
```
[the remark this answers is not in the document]
PradeepKiruvale: But the order of messages is fixed for any given operation. So, I feel that skipping is OK here instead of just asserting the topics.

crates/extensions/c8y_mapper_ext/src/converter.rs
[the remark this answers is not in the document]
albinsuresh: > the purpose of the tests is to assert only the output messages

Yes, but all the output messages. But here, with this filter we're only asserting the messages sent to c8y, skipping any messges sent to `te/` or any other topics during the conversion. So, the validation is not complete if we don't assert all the converted messages.

crates/extensions/c8y_mapper_ext/src/tests.rs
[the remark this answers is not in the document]
albinsuresh: > If we do not skip, the assert happens on the wrong topic and tests will fail

This is exactly what we're trying to prevent. When you're skipping 2 messages here, you're not skipping any two messages. You're specifically skipping 2 messages that are probably not relevant for the test. But, still it is important to be sure that those 2 messages are exactly what you expected. Here, for example, you expect to skip the first 2 registration messages. What if one of them failed and sent an error message to `tedge/errors` instead? It'd fail in the next assertion but with some really weird errors than what you expected. So, that's why I'm suggesting a loose assertion on `("c8y/s/us", "101")` topic.

crates/extensions/c8y_mapper_ext/src/converter.rs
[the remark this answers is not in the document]
PradeepKiruvale: Ok, I have updated only to the tests.

crates/extensions/c8y_mapper_ext/src/tests.rs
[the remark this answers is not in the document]
PradeepKiruvale: ok, updated the tests

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,242 @@ mod tests {
...
+                "externalSource":{"externalId":"nested_child","type":"c8y_Serial"},
+                "temp":{"temp":{"value":1.0}},
+                "time":"2021-11-16T17:45:40.571760714+01:00",
+                "type":"ThinEdgeMeasurement"
+            })
+            .to_string(),
+        );
+
+        // Test the first output messages contains SmartREST and C8Y JSON.
+        let out_first_messages: Vec<_> = converter.convert(&in_message).await.into_iter().collect();
```
albinsuresh, line 1731:
```suggestion
        let out_first_messages: Vec<_> = converter.convert(&in_message).await;
```
Why iterate and collect when the output of `convert` is already a vector?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,242 @@ mod tests {
...
+        assert_eq!(out_first_messages, vec![expected_c8y_json_message.clone()]);
+
+        // Test the second output messages doesn't contain SmartREST child device creation.
+        let out_second_messages = converter.convert(&in_message).await;
+        assert_eq!(out_second_messages, vec![expected_c8y_json_message]);
```
albinsuresh, line 1736: The first convert also did not have the SmartREST child creation message, as the nested child device does not support auto-registeration and had to be explicitly registered. So, this assert can be removed.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,242 @@ mod tests {
...
+                "externalSource":{"externalId":"nested_service","type":"c8y_Serial"},
+                "temp":{"temp":{"value":1.0}},
+                "time":"2021-11-16T17:45:40.571760714+01:00",
+                "type":"ThinEdgeMeasurement"
+            })
+            .to_string(),
+        );
+
+        // Test the first output messages contains SmartREST and C8Y JSON.
+        let out_first_messages: Vec<_> = converter.convert(&in_message).await.into_iter().collect();
```
albinsuresh, line 1792: Unnecessary iter and collect.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,242 @@ mod tests {
...
+        assert_eq!(out_first_messages, vec![expected_c8y_json_message.clone()]);
+
+        // Test the second output messages doesn't contain SmartREST child device creation.
+        let out_second_messages = converter.convert(&in_message).await;
+        assert_eq!(out_second_messages, vec![expected_c8y_json_message]);
```
albinsuresh, line 1797: This assertion is also pointless for the same reason mentioned above, unless this test was relying on auto-registration for the service on nested child device. Auto-registration is not supported for nested child devices, but it is supported for services on any child device, if they follow the default topic scheme.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,242 @@ mod tests {
...
+                    "type":"c8y_Serial"
+                },
+                "temp":{"temp":{"value":1.0}},
+                "time":"2021-11-16T17:45:40.571760714+01:00",
+                "type":"m_type"})
+            .to_string(),
+        );
+
+        // Test the first output messages contains SmartREST and C8Y JSON.
+        let out_first_messages: Vec<_> = converter.convert(&in_message).await.into_iter().collect();
```
albinsuresh, line 1854: Unnecessary iter and collect.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1686,6 +1686,242 @@ mod tests {
...
+            .to_string(),
+        );
+
+        let expected_smart_rest_message_service = Message::new(
+            &Topic::new_unchecked("c8y/s/us"),
+            "102,test-device:device:main:service:appm,systemd,appm,up",
+        );
+
+        // Test the first output messages contains SmartREST and C8Y JSON.
+        let out_first_messages: Vec<_> = converter.convert(&in_message).await.into_iter().collect();
```
albinsuresh, line 1911: Unnecessary iter and collect.

PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: fixed

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: yeah, fixed

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: fixed

albinsuresh (approved):

PR #2303, opened 2023-09-29
Add configuration operation support to c8y-mapper

rina23q:
## Proposed changes
Todo:
- [x] Fix why te/../cmd/config.. topic is not picked up by mapper. Check subscription
- [x] Improve error management. Remove all `unwrap`s.
- [x] Add unit tests
- [x] Fix issues found by unit tests
- [x] Delete a file from file-transfer-repo after operation is successful
- [x] Add tedge config like `c8y.enable.config_update` and `c8y.enable.config_snapshot` 

Follow-up:
- [ ] Add timer and timeout if operation takes too long together with log_upload cases
- [ ] Update robot tests (which requires `tedge-configuration-plugin` package)
- [ ] Refactor as suggested in https://github.com/thin-edge/thin-edge.io/pull/2303#discussion_r1356364285
- [ ] Use uploader actor once https://github.com/thin-edge/thin-edge.io/pull/2302 is done

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
#2292 

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2340, opened 2023-10-12
Add try_get() and try_get_by_external_id() to entity_store

rina23q:
## Proposed changes
As commented in https://github.com/thin-edge/thin-edge.io/pull/2303#discussion_r1356364285, we have many patterns using `get()` or `get_by_external_id()` and map them `None` to `Err`. We should stop the same mapping every time.

The new API methods return `Error::UnknownEntity` instead of `None`.

## Types of changes

## Paste Link to the issue
#2303 

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -164,10 +164,25 @@ impl EntityStore {
         self.entities.get(entity_topic_id)
     }
 
-    /// Returns information for an entity under a given device/service id .
+    /// Tries to get information about an entity under a given MQTT entity topic identifier.
+    pub fn try_get(&self, entity_topic_id: &EntityTopicId) -> Result<&EntityMetadata, Error> {
+        self.get(entity_topic_id)
+            .ok_or_else(|| Error::UnknownEntity(entity_topic_id.to_string()))
+    }
+
+    /// Returns information for an entity under a given device/service id.
     pub fn get_by_external_id(&self, external_id: &EntityExternalId) -> Option<&EntityMetadata> {
         let topic_id = self.entity_id_index.get(external_id)?;
         self.get(topic_id)
+    }
+
+    /// Tries to get information for an entity under a given device/service id.
+    pub fn try_get_by_external_id(
+        &self,
+        external_id: &EntityExternalId,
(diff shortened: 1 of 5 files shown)


crates/core/tedge_api/src/entity_store.rs
```diff
@@ -164,12 +164,27 @@ impl EntityStore {
         self.entities.get(entity_topic_id)
     }
 
-    /// Returns information for an entity under a given device/service id .
+    /// Tries to get information about an entity under a given MQTT entity topic identifier.
+    pub fn try_get(&self, entity_topic_id: &EntityTopicId) -> Result<&EntityMetadata, Error> {
```
albinsuresh, line 168: Pedantic comment: Returning `entity_store::Error` here is a bit of an overkill as this API only returns a single type of error. So, it could have been a dedicated error struct.

But, let's be practical and don't go too far with error types. So, let it be.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -164,12 +164,27 @@ impl EntityStore {
         self.entities.get(entity_topic_id)
     }
 
-    /// Returns information for an entity under a given device/service id .
+    /// Tries to get information about an entity under a given MQTT entity topic identifier.
```
albinsuresh, line 167:
```suggestion
    /// Tries to get information about an entity using its `EntityTopicId`,
    /// returning an error if the entity is not registered.
```

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -164,12 +164,27 @@ impl EntityStore {
...
+            .ok_or_else(|| Error::UnknownEntity(entity_topic_id.to_string()))
+    }
+
+    /// Returns information for an entity under a given device/service id.
     pub fn get_by_external_id(&self, external_id: &EntityExternalId) -> Option<&EntityMetadata> {
         let topic_id = self.entity_id_index.get(external_id)?;
         self.get(topic_id)
     }
 
+    /// Tries to get information for an entity under a given device/service id.
```
albinsuresh, line 179:
```suggestion
    /// Tries to get information about an entity using its `EntityExternalId`,
    /// returning an error if the entity is not registered.
```

albinsuresh (approved):

PR #2360, opened 2023-10-20
fix: prevent tedge-agent service stoppage when upgrading from tedge_agent

reubenmiller:
## Proposed changes

Fix an issue discovered during development where updating the tedge-agent using the software management feature would cause the operation to be stuck in executing.

The root cause was the fact that when upgrading from version <=0.8.1, the prerm maintainer script would be executed unexpectidly due to the tedge_agent being listed as a conflict/breaks in the package definition. The prerm script would stop the tedge-agent service thus interrupting the apt package installation during an over the air update.

The only downside to removing the conflict/break fields from the tedge-agent package definition is that the `tedge_agent` deprecated package is that `tedge_agent` will still appear in the software install list as it is technically not replaced. The user can simply remove `tedge_agent` afterwards without any unintended side effects.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/configuration/package_manifests/nfpm.tedge-agent.yaml
+++ b/configuration/package_manifests/nfpm.tedge-agent.yaml
@@ -20,16 +20,18 @@ replaces:
   - tedge_agent
 provides:
   - tedge_agent
-conflicts:
-  - tedge_agent (<= 0.8.1)
+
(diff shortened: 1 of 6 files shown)


didier-wenzek (approved):
Approved

Issue #2351, opened 2023-10-18
Health status on `te/.../status/health` for AWS and Azure mappers


Bravo555:
AWS and Azure mappers still use old health topics. Now most thin-edge services have been updated to publish on new health topics, these mappers won't see health status updates unless their mappers and bridges are also updated.

TODO:

- [x] Add tests for converting health messages
- [x] Implement converting health messages

Issue #2348, opened 2023-10-18
Document health status and health check messages


Bravo555:
This issue encompasses updating all previous references of old health topics (`tedge/health`, `tedge/health-check`), as well as adding new documentation describing health status and health check messages.

- [x] Update documentation pages which still refer to old health topics:
   - [_How to monitor health of tedge services_](https://thin-edge.github.io/thin-edge.io/next/operate/troubleshooting/monitor_tedge_health/)
   - [_How to monitor health of service from Cumulocity IoT_](https://thin-edge.github.io/thin-edge.io/next/operate/c8y/c8y_service_monitoring/)
   - [_Enabling the systemd watchdog feature for a tedge service_](https://thin-edge.github.io/thin-edge.io/next/operate/monitoring/enable_tedge_watchdog_using_systemd/#debugging)
   - [_How to control which MQTT topics the mappers subscribe to_](https://thin-edge.github.io/thin-edge.io/next/operate/configuration/config-mapper-mqtt-topics/#check-the-subscribed-mqtt-topics/)
- [x] _Health status_ section should be added to [_Telemetry_ section of _Cumulocity Mapper_ page](https://thin-edge.github.io/thin-edge.io/next/references/mappers/c8y-mapper/#telemetry)

Bravo555:
Implemented by #2375

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/MQTT_health_check/health_tedge_mapper_c8y.robot
- tests/RobotFramework/tests/aws/aws_telemetry.robot
- tests/RobotFramework/tests/azure/azure_telemetry.robot
- tests/RobotFramework/tests/tedge/call_tedge_config_list.robot
- [x] QA has tested the bug and could not reproduce it anymore.

PR #2364, opened 2023-10-23
Health status on AWS and Azure mappers

Bravo555:
## Proposed changes

This PR completes the migration to new health topics. Besides switching over AWS and Azure mappers to new health status topics, some leftovers of old health check topics were also removed.

`tedge/health/` and `tedge/health-check` topics now no longer appear in the integration test log (besides the test checking upgradability from previous version, which is expected).

## Types of changes

## Paste Link to the issue

- #2351

## Checklist

## Further comments

Next step will be to change or add relevant documentation related to health status/check messages and the health checking mechanism, as described in #2348.

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -420,7 +420,7 @@ define_tedge_config! {
 
         /// Set of MQTT topics the Azure IoT mapper should subscribe to
         #[tedge_config(example = "te/+/+/+/+/a/+,te/+/+/+/+/m/+,te/+/+/+/+/e/+")]
-        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,tedge/health/+,tedge/health/+/+"))]
+        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,te/+/+/+/+/status/health"))]
         topics: TemplatesSet,
     },
 
@@ -444,7 +444,7 @@ define_tedge_config! {
 
         /// Set of MQTT topics the AWS IoT mapper should subscribe to
         #[tedge_config(example = "te/+/+/+/+/a/+,te/+/+/+/+/m/+,te/+/+/+/+/e/+")]
-        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,tedge/health/+,tedge/health/+/+"))]
+        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,te/+/+/+/+/status/health"))]
         topics: TemplatesSet,
     },
 
--- a/crates/core/tedge/src/cli/connect/bridge_config_aws.rs
+++ b/crates/core/tedge/src/cli/connect/bridge_config_aws.rs
@@ -2,6 +2,8 @@ use crate::cli::connect::BridgeConfig;
 use camino::Utf8PathBuf;
 use tedge_config::ConnectUrl;
 
+const MOSQUITTO_BRIDGE_TOPIC: &str = "te/device/main/service/mosquitto-aws-bridge/status/health";
+
 #[derive(Debug, Eq, PartialEq)]
 pub struct BridgeConfigAwsParams {
     pub connect_url: ConnectUrl,
@@ -62,7 +64,7 @@ impl From<BridgeConfigAwsParams> for BridgeConfig {
             local_clean_session: false,
             notifications: true,
             notifications_local_only: true,
-            notification_topic: "tedge/health/mosquitto-aws-bridge".into(),
+            notification_topic: MOSQUITTO_BRIDGE_TOPIC.into(),
             bridge_attempt_unsubscribe: false,
             topics: vec![
                 pub_msg_topic,
@@ -117,7 +119,7 @@ fn test_bridge_config_from_aws_params() -> anyhow::Result<()> {
         local_clean_session: false,
         notifications: true,
         notifications_local_only: true,
-        notification_topic: "tedge/health/mosquitto-aws-bridge".into(),
+        notification_topic: MOSQUITTO_BRIDGE_TOPIC.into(),
         bridge_attempt_unsubscribe: false,
     };
 
--- a/crates/extensions/aws_mapper_ext/src/converter.rs
+++ b/crates/extensions/aws_mapper_ext/src/converter.rs
@@ -4,7 +4,6 @@ use serde_json::Map;
 use serde_json::Value;
 use std::convert::Infallible;
 use tedge_actors::Converter;
-use tedge_api::health::is_bridge_health;
 use tedge_api::mqtt_topics::Channel;
 use tedge_api::mqtt_topics::EntityTopicId;
 use tedge_api::mqtt_topics::MqttSchema;
@@ -15,6 +14,7 @@ use crate::error::ConversionError;
 use crate::size_threshold::SizeThreshold;
 
 const AWS_MQTT_THRESHOLD: usize = 1024 * 255;
+const MOSQUITTO_BRIDGE_TOPIC_ID: &str = "device/main/service/mosquitto-aws-bridge";
 
 pub struct AwsConverter {
     pub(crate) add_timestamp: bool,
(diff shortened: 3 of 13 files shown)


crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -81,12 +66,22 @@ impl AzureConverter {
...
 
         match &channel {
-            Channel::Measurement { .. } | Channel::Event { .. } | Channel::Alarm { .. } => {
+            Channel::Measurement { .. }
+            | Channel::Event { .. }
+            | Channel::Alarm { .. }
+            | Channel::Health => {
                 let payload = self.with_timestamp(input)?;
                 let output = MqttMessage::new(&self.mapper_config.out_topic, payload);
                 self.size_threshold.validate(&output)?;
```
rina23q, line 87: I hope the size validation can be hidden in one-level higher. As of now, this needs to be called everywhere a MQTT message is generated. Easy to forget to call it. The size threshold is determined by the type of cloud (c8y, az, aws..). How about adding `size_threshold.validate()` where `try_convert()` is called?

https://github.com/thin-edge/thin-edge.io/blob/d3ed45f4f1f8909a113693840bfec18b5bbc55c6/crates/extensions/az_mapper_ext/src/converter.rs#L133
```
impl Converter for AzureConverter {
    type Input = MqttMessage;
    type Output = MqttMessage;
    type Error = Infallible;

    fn convert(&mut self, input: &Self::Input) -> Result<Vec<Self::Output>, Self::Error> {
        let messages_or_err = self.try_convert(input);
        // add validation here
        Ok(self.wrap_errors(messages_or_err))
    }
}
```

Then we can remove tehe line 78, `self.size_threshold.validate(input)?;`. This is improvised way to check the size... the logic is here: the output message should be checked, but since input and output messages are almost same, if input message is too large, we presume the output is also too large.

crates/extensions/aws_mapper_ext/src/converter.rs
```diff
@@ -42,60 +42,75 @@ impl AwsConverter {
...
+    fn convert_telemetry_message(
         &mut self,
         input: &MqttMessage,
         source: EntityTopicId,
         telemetry_type: &String,
     ) -> Result<Vec<MqttMessage>, ConversionError> {
         let payload = self.with_time_stamp(input)?;
-        let source = normalize_name(source);
+        let source = normalize_name(&source);
+        // XXX: should match on `Channel` instead
```
rina23q, line 113: Although the channel check is already done by the caller (`try_convert_te_topics()`), we need again to check channel doesn't seem efficient.

rina23q (commented):
Overall the PR looks good already. I think some todo can be done in this PR.

tests/RobotFramework/tests/aws/aws_telemetry.robot
```diff
@@ -55,11 +55,14 @@ Publish child device alarm to te alarm topic with alarm type
...
 
+Publish health status message for main device service
+    Execute Command    tedge mqtt pub te/device/main/service/test-service/status/health '{"status":"up"}'
+    Should Have MQTT Messages    aws/td/device:main:service:test-service/status/health    message_contains="status":"up"
 
 *** Keywords ***
 Custom Setup
     Setup
-    Execute Command    sudo tedge config set aws.topics "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+"
+    # Execute Command    sudo tedge config set aws.topics "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+"
```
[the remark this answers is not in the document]
Bravo555: My bad, commented that out and forgot to remove, now fixed. e39f632c2e

crates/extensions/az_mapper_ext/src/converter.rs
Bravo555: Did the same as i did in AWS converter: e39f632c2e

crates/core/tedge/src/cli/connect/bridge_config_aws.rs
```diff
@@ -2,6 +2,8 @@ use crate::cli::connect::BridgeConfig;
 use camino::Utf8PathBuf;
 use tedge_config::ConnectUrl;
 
+const MOSQUITTO_BRIDGE_TOPIC: &str = "te/device/main/service/mosquitto-aws-bridge/status/health";
```
PradeepKiruvale, line 5: What will happen if the `mosquitto` runs on a separate device? is it still considered a service of the main device?

PradeepKiruvale (commented):
LGTM

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -420,7 +420,7 @@ define_tedge_config! {
 
         /// Set of MQTT topics the Azure IoT mapper should subscribe to
         #[tedge_config(example = "te/+/+/+/+/a/+,te/+/+/+/+/m/+,te/+/+/+/+/e/+")]
-        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,tedge/health/+,tedge/health/+/+"))]
+        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,te/+/+/+/+/status/health"))]
```
albinsuresh, line 423:
```suggestion
        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,te/+/+/+/+/cmd/health/check"))]
```
Surprised to see that these mappers have not subscribed to the `cmd/health/check` topic. How will they respond to health check requests from the watchdog without that?

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -444,7 +444,7 @@ define_tedge_config! {
 
         /// Set of MQTT topics the AWS IoT mapper should subscribe to
         #[tedge_config(example = "te/+/+/+/+/a/+,te/+/+/+/+/m/+,te/+/+/+/+/e/+")]
-        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,tedge/health/+,tedge/health/+/+"))]
+        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,te/+/+/+/+/status/health"))]
```
albinsuresh, line 447: Same comment as above.

crates/core/tedge/src/cli/connect/bridge_config_aws.rs
albinsuresh: Obviously this `tedge connect` logic won't work if mosquitto is running on another device as we don't support updating the bridge config of a remote broker.

But even ignoring that fact, this cloud bridge is something that is associated with the main device only. Even if the main device uses mosquitto running on another device, from the child's perspective, it's the main device's broker.

Now, the child devices may have their own brokers connected/bridged to the main device broker. But, even in that case, this `tedge connect` command is only interested in the main device's broker.

crates/extensions/aws_mapper_ext/src/converter.rs
albinsuresh: I think what he meant is that, it's much better to get that channel passed here via an argument and match on the `channel` object than doing the array matches as done below. That exact same information is available in the channel, making the split/collect/match unnecessary/redundant.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -28,6 +28,8 @@ pub fn convert_health_status_message(
     let mut mqtt_messages: Vec<Message> = Vec::new();
 
     // If not Bridge health status
+    // FIXME: can also match "device/bridge//" or "/device/main/service/my_custom_bridge"
```
albinsuresh, line 31: That looks like a very fragile logic. Creating an issue would be nice, otherwise these FIXMEs might get forgotten/lost, unless you had some plans to fix it immediatley.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
didier-wenzek: These are subscriptions the user can opt-out, say to publish only measurements. In any case, the mappers are subscribing to the topics which are required to well behave as `cmd/health/check`.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
Bravo555: I'll make a separate issue and a PR for this, for visibility.

Bravo555: #2370

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
PradeepKiruvale: But how it will work if the mapper has to be frequently checked for its health by a `watchdog`

albinsuresh: The `HealthMonitorBuilder`, used by all the tedge components, is responsible for subscribing to those `health/check` messages, used by the watchdog. You can't/shouldn't opt out of it, which is why it is not exposed via this config parameter.

albinsuresh (approved):
LGTM

rina23q (approved):
LGTM

Issue #2366, opened 2023-10-24
tedge-configuration-plugin ambiguous failure reason when file not found


reubenmiller:
**Describe the bug**

The new `tedge-configuration-plugin` returns an ambiguous failure reason when trying to retrieve a non-existent configuration file from a device.

In the `c8y-configuration-plugin` the failure reason included the following text:

```
No such file or directory
```

But repeating the same command on the more generic `tedge-configuration-plugin` the error returned is in the form of:

```
Handling of operation failed with Can't open a file /etc/config1.json
```

**To Reproduce**

The change in behaviour was detected in the following [#2368](https://github.com/thin-edge/thin-edge.io/pull/2368) (test: [configuration_operation.robot#L64](https://github.com/thin-edge/thin-edge.io/blob/a2fe33664a4505cd2c80fbac157a04cc161551f0/tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot#L64)):

1. Connect thin-edge.io to Cumulocity

2. Request a configuration file from the device that does not exist

3. Wait for the operation to be set to FAILED and inspect the failure reason.

    Expected: The failure reason should indicate to the user that the file does not exist instead of saying that it could not read the file.

**Expected behavior**

The operation should be set to FAILED and the failure reason should indicate to the user that the file does not exist instead of saying that it could not read the file. Reading a file can also fail due to other reasons such as permissions, so it would be more helpful to debug if the "file does not exist" and "can't read file" error cases were able to be easily differentiated by the user.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Raspbian GNU/Linux 11 (bullseye)`|
|Hardware [incl. revision]|`docker`|
|System-Architecture|`aarch64`|
|thin-edge.io version|`0.12.1~454+gf68652d`|

**Additional context**

reubenmiller:
The failure reason was updated to included the text "No such file or directory".

reubenmiller:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- [x] QA has tested the bug and could not reproduce it anymore.

PR #2368, opened 2023-10-24
Switch to tedge-configuration-plugin

reubenmiller:
## Proposed changes

Switch to using `tedge-configuration-plugin` by default:

* Use tedge-configuration-plugin in multi-call binary (instead of the c8y-configuration-plugin)
* Include tedge-configuration-plugin in default packaging (inc
* Include c8y-configuration-plugin in legacy packaging (to make it easier for some users to build the package themselves)
* Update system tests

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


reubenmiller:
Current test failure is due to https://github.com/thin-edge/thin-edge.io/issues/2366

configuration/package_manifests/nfpm.c8y-configuration-plugin.yaml
```diff
@@ -28,6 +28,10 @@ deb:
     - c8y_configuration_plugin (<= 0.8.1)
 
 contents:
+  # binary
+  - src: .build/c8y-configuration-plugin
+    dst: /usr/bin/
+
```
albinsuresh, line 34: I'm assuming this is the default, as it was working fine without it earlier, and I don't see the same definition in the nfpm definitions of other packages.

configuration/package_manifests/nfpm.tedge-configuration-plugin.yaml
```diff
@@ -0,0 +1,69 @@
...
+maintainer: "thin-edge.io team <info@thin-edge.io>"
+vendor: "thin-edge.io"
+homepage: "https://thin-edge.io"
+license: "Apache-2.0"
+
+depends:
+  - tedge
+replaces:
+  - c8y_configuration_plugin
+  - c8y-configuration-plugin
```
albinsuresh, line 21:
```suggestion
  - c8y_configuration_plugin
  - c8y-configuration-plugin
  - tedge-configuration-plugin
```
Just future-proofing. The same probably applies to the `conflicts` list as well, but I'm not sure.

configuration/package_manifests/nfpm.tedge-configuration-plugin.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+deb:
+  fields:
+    Vcs-Browser: ${CI_PROJECT_URL}
+    Vcs-Git: ${CI_PROJECT_URL}
+  compression: xz
+  breaks:
+    - c8y_configuration_plugin (<= 0.8.1)
+    - c8y-configuration-plugin
```
albinsuresh, line 33: I'm confused with this `breaks` clause. The earlier `replaces` and `conflicts` clauses are instructing to replace the old plugin with the new one and here, we're instructing against it. What am I missing?

crates/core/tedge/Cargo.toml
```diff
@@ -14,7 +14,6 @@ readme = "README.md"
 anstyle = { workspace = true }
 anyhow = { workspace = true }
 base64 = { workspace = true }
-c8y-configuration-plugin = { workspace = true }
```
albinsuresh, line 17: The `packages.json` file still has references to the old plugin. If we remove this from here, they won't be built anymore, right?

tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
```diff
@@ -221,6 +219,13 @@ Custom Setup
     # Parent
     ${parent_sn}=    Setup    skip_bootstrap=False
     Set Suite Variable    $PARENT_SN    ${parent_sn}
+    # Disable new generic handling
+    Execute Command    sudo tedge config set c8y.enable.config_snapshot false
+    Execute Command    sudo tedge config set c8y.enable.config_update false
+
+    # Use deprecated c8y-configuration-plugin over tedge-configuration-plugin
```
albinsuresh, line 226: Wondering why. The new plugin can't be run on child devices yet?

albinsuresh (commented):
I'm assuming the doc updates will be done in a follow-up PR, as this one is already fairly big.

crates/core/tedge/Cargo.toml
reubenmiller: We still need to keep the older c8y-configuration-plugin in place for now as some people might be relying on this old interface to support custom child device implementations (as we [publicly documented the interface](https://thin-edge.github.io/thin-edge.io/references/c8y-configuration-management/))

configuration/package_manifests/nfpm.tedge-configuration-plugin.yaml
reubenmiller: You need both the conflicts/breaks to be defined in order for the tedge-configuration-plugin to automatically replace the c8y-configuration-plugin. This enables a smooth transition from c8y-configuration-plugin to tedge-configuration-plugin.

configuration/package_manifests/nfpm.tedge-configuration-plugin.yaml
reubenmiller: No, a package can't conflict with itself. This package definition is for the `tedge-configuration-plugin` package.

configuration/package_manifests/nfpm.c8y-configuration-plugin.yaml
reubenmiller: The link to the binary was removed a few weeks ago when we made the switch to the multi-binary setup. However this PR replaces the previous c8y-configuration-plugin with tedge-configuration-plugin (inside the tedge binary). And since the c8y-configuration-plugin is no longer included in the multi-call binary, it needs to be manually added to the package.

ci/package_list.sh
```diff
@@ -8,12 +8,20 @@ RELEASE_PACKAGES=(
...
     c8y-remote-access-plugin
     c8y-firmware-plugin
 )
 export RELEASE_PACKAGES
 
+# Deprecated packages are still built but not explicitly tested
+# This allows users to still access the packages if needed, however
+# it is only reserved for packages with a more public facing API
+DEPRECATED_PACKAGES=(
+    c8y-configuration-plugin
```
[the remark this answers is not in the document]
reubenmiller: I opted against this because the log plugin never had a publicly documented public interface (which what I was trying to allude to in the comment above the array definition.

configuration/package_manifests/nfpm.tedge-configuration-plugin.yaml
albinsuresh: Aah okay. I thought this also defines which package(s) to replace during an upgrade of the `tedge-config-plugin` from one version to another.

reubenmiller: The main reason for conflicts/breaks is to handle package renaming (not standard package updates to different versions). You can read more info about it here: https://wiki.debian.org/RenamingPackages

crates/core/tedge/Cargo.toml
albinsuresh: > We still need to keep the older c8y-configuration-plugin in place for now as some people might be relying on this old interface

Yeah, that's what I was expecting as well and got confused by this change, essentially skipping the build of this plugin, which we still want to package for those "special usecases". But then had another look at the `package_list.sh` changes, where this deprecated package is added to the consolidated package list that is finally built. So, it makes sense now.

configuration/package_manifests/nfpm.tedge-configuration-plugin.yaml
albinsuresh: Okay. That automatic replacement is what I was expecting as well. But, the nfpm documentation of the `breaks` key seemed to suggest something totally different:

```
  # Packages which would break if this package would be installed.
  # The installation of this package is blocked if `some-package`
  # is already installed.
  breaks:
    - some-package
```

Which seemed to suggest that attempting to install `tedge-config-plugin` on a system where `c8y-config-plugin` is already installed would be blocked, meaning a smooth upgrade won't be possible. But, I trust you and our system tests more than a possibly misleading doc statement.

crates/extensions/tedge_config_manager/Cargo.toml
```diff
@@ -25,6 +25,7 @@ tedge_mqtt_ext = { workspace = true }
 tedge_uploader_ext = { workspace = true }
 tedge_utils = { workspace = true }
 thiserror = { workspace = true }
+tokio = { workspace = true }
```
rina23q, line 28: I'm surprised to see tokio is added here as a part of this PR. What tool suggested to add this?

tests/RobotFramework/tests/MQTT_health_check/health_c8y-configuration-plugin.robot
rina23q, line 1: The file name must be updated as well.

tests/images/debian-systemd/files/bootstrap.sh
```diff
@@ -543,7 +543,7 @@ install_via_local_files() {
     find_then_install_deb "$INSTALL_SOURCEDIR" "tedge[_-]mapper_*_$ARCH.deb"
     find_then_install_deb "$INSTALL_SOURCEDIR" "tedge[_-]agent_*_$ARCH.deb"
     find_then_install_deb "$INSTALL_SOURCEDIR" "tedge[_-]apt[_-]plugin_*_$ARCH.deb"
-    find_then_install_deb "$INSTALL_SOURCEDIR" "c8y[_-]configuration[_-]plugin_*_$ARCH.deb"
+    find_then_install_deb "$INSTALL_SOURCEDIR" "tedge-configuration-plugin_*_$ARCH.deb"
     find_then_install_deb "$INSTALL_SOURCEDIR" "tedge[_-]log[_-]plugin_*_$ARCH.deb"
```
rina23q, line 547: Out of this PR's scope, but we can simplify this line as well.
```suggestion
    find_then_install_deb "$INSTALL_SOURCEDIR" "tedge-log-plugin_*_$ARCH.deb"
```

tests/images/debian-systemd/files/bootstrap.sh
```diff
@@ -595,8 +595,8 @@ install_tedge() {
 
     # Check if any packages are incompatible, if so remove the previous version first
     # Use new and old packages names, and check each of them one by one
-    packages="tedge tedge-mapper tedge-agent tedge-log-plugin c8y-configuration-plugin c8y-remote-access-plugin"
-    packages="$packages tedge tedge_mapper tedge_agent c8y_log_plugin c8y-log-plugin c8y_configuration_plugin"
+    packages="tedge tedge-mapper tedge-agent tedge-log-plugin tedge-configuration-plugin c8y-remote-access-plugin"
+    packages="$packages tedge tedge_mapper tedge_agent c8y_log_plugin c8y-log-plugin c8y_configuration_plugin c8y-configuration-plugin"
```
rina23q, line 599: Out of scope of this PR, why 'tedge' is added twice?

Cargo.toml
```diff
@@ -138,6 +138,7 @@ tedge_config_macros = { path = "crates/common/tedge_config_macros" }
 tedge_config_macros-impl = { path = "crates/common/tedge_config_macros/impl" }
 tedge_config_manager = { path = "crates/extensions/tedge_config_manager" }
 tedge_downloader_ext = { path = "crates/extensions/tedge_downloader_ext" }
+tedge-configuration-plugin = { path = "plugins/tedge_configuration_plugin" }
```
rina23q, line 141: A bit surprised that 'cargo sort' didn't complain here

rina23q (commented):
Thank you! Only minor comments.

tests/images/debian-systemd/files/bootstrap.sh
reubenmiller: Looks to be an oversight...though it will just be a no-op if the package does not exist anyway, but yes it would be good to clean it up.

crates/extensions/tedge_config_manager/Cargo.toml
reubenmiller: Actually this may not be required anymore. It was previously required due to a patch that @jarhodes314 provided me...but I think the patch was reverted as it was no longer needed, but this change remained. I will revert this change.

Cargo.toml
reubenmiller: There are actually a few dependencies which are not strictly sorted alphabetically due to differences in dashes and underscores:

image

I'm tempted to leave it as is since for now.

crates/extensions/tedge_config_manager/Cargo.toml
reubenmiller: Removed unused dependency e4cf7f761b624857fdcd5c34e88ad7a1a39d1240

Cargo.toml
reubenmiller: Reading the docs on [cargo-sort](https://crates.io/crates/cargo-sort), it looks like the "workspace.dependencies" property is not monitored by cargo-sort...hence why we have some imports out of order.

rina23q (approved):
LGTM! Thanks!

albinsuresh (approved):
LGTM

Issue #2365, opened 2023-10-23
digital twin property not sent to cloud after it was cleared and set again


reubenmiller:
**Describe the bug**

Setting a digital twin property to the same value after it has been cleared does not result in the value being forwarded to Cumulocity IoT.

**To Reproduce**

1. Connect thin-edge.io to Cumulocity

2. Publish a property/fragment to the digital twin

    ```sh
    tedge mqtt pub te/device/main///twin/subtype '"rp1"' -r
    ```

    Verify that the `subtype` property has been set to `rpi` in Cumulocity IoT.

3. Clear the previous fragment from the previous

    ```sh
    tedge mqtt pub te/device/main///twin/subtype '' -r
    ```

    Verify that the `subtype` property has removed in Cumulocity IoT.

4. Publish a property/fragment to the digital twin

    ```sh
    tedge mqtt pub te/device/main///twin/subtype '"rp1"' -r
    ```

    Verify that the `subtype` property has been set to `rpi` in Cumulocity IoT. Currently this step fails. However sending a different value will result in the value being forwarded to Cumulocity IoT.

**Expected behavior**

Setting a previously cleared flag with the original value should write the value in the cloud.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Raspbian GNU/Linux 11 (bullseye)`|
|Hardware [incl. revision]|`Raspberry Pi Model B Rev 2`|
|System-Architecture|`Linux finn 6.1.21+ #1642 Mon Apr  3 17:19:14 BST 2023 armv6l GNU/Linux`|
|thin-edge.io version|`tedge 0.12.1~457+gd3ed45f`|

**Additional context**

Most likely this is a simple bug with the internal model is not cleared when a fragment is cleared which affects the internal diff model evaluation.

reubenmiller:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- [x] QA has tested the bug and could not reproduce it anymore.

PR #2369, opened 2023-10-24
fix: support sending previously cleared twin property value

reubenmiller:
## Proposed changes

The bug was caused by the mapper not removing the cleared fragment from its local inventory cache, when a clear twin data message is received. When the old value existed in the cache, the duplicate message that came after the clear message was getting ignored, because it is the same as the existing value.

Support setting twin property after it has been cleared.

Tasks:

* [x] System test
* [x] Rust implementation

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2365

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -15,6 +15,7 @@ use crate::mqtt_topics::TopicIdError;
 use log::debug;
 use mqtt_channel::Message;
 use serde_json::json;
+use serde_json::Map;
 use serde_json::Value as JsonValue;
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
@@ -180,6 +181,7 @@ impl EntityStore {
             r#type: main_device.r#type,
             parent: None,
             other: main_device.other,
+            twin_data: Map::new(),
         };
 
         Some(EntityStore {
@@ -197,10 +199,25 @@ impl EntityStore {
         self.entities.get(entity_topic_id)
     }
 
+    /// Returns a mutable reference to the `EntityMetadata` for the given `EntityTopicId`.
+    fn get_mut(&mut self, entity_topic_id: &EntityTopicId) -> Option<&mut EntityMetadata> {
+        self.entities.get_mut(entity_topic_id)
+    }
+
     /// Tries to get information about an entity using its `EntityTopicId`,
     /// returning an error if the entity is not registered.
     pub fn try_get(&self, entity_topic_id: &EntityTopicId) -> Result<&EntityMetadata, Error> {
         self.get(entity_topic_id)
+            .ok_or_else(|| Error::UnknownEntity(entity_topic_id.to_string()))
+    }
+
+    /// Tries to get a mutable reference to the `EntityMetadata` for the given `EntityTopicId`,
+    /// returning an error if the entity is not registered.
+    fn try_get_mut(
+        &mut self,
+        entity_topic_id: &EntityTopicId,
+    ) -> Result<&mut EntityMetadata, Error> {
+        self.get_mut(entity_topic_id)
             .ok_or_else(|| Error::UnknownEntity(entity_topic_id.to_string()))
     }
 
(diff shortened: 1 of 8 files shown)


reubenmiller:
@albinsuresh I've added a system test for coverage. Could you please handle the implementation?

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -179,7 +180,7 @@ impl EntityStore {
             external_id: entity_id.clone(),
             r#type: main_device.r#type,
             parent: None,
-            other: main_device.other,
+            twin_data: main_device.other,
```
didier-wenzek, line 183: So `twin_data` stores properties received *either* from registration messages *or* from `twin` message fragments.  Okay.

But, the method `EntityStore::update()` incorrectly overwrite the whole `twin_data` with the `other` field of the  registration message, while `EntityStore::update_twin_data()` correctly updates only the fragment provided by the twin message.

Also, are the data read from a registration message published as twin data to c8y?

didier-wenzek (changes requested):
Sending a registration message erases twin data already sent. Am I missing something?

crates/core/tedge_api/src/entity_store.rs
albinsuresh: This overwrite happens only when the entity is registered. You could even say that, that's when the `twin_data` is initialized for the first time. Further updates to `twin/` channel just updates this initial state.

> Also, are the data read from a registration message published as twin data to c8y?

No, we don't want 2 copies of the same data to exist in two different retained messages, leading to unnecessary consistency issues. That's the reason why we're even rejecting updates to `name` and `type` fields via the `twin/` channel.

albinsuresh: Aah, now I get the issue that you wanted to highlight. Yes, a second registration message for an already registered entity would overwrite the whole twin data with this code. I hadn't considered that. Thought of only the initial registration path. Will fix that.

didier-wenzek: Okay. I would have prefer two distinct `EntityMetadata` fields as the life-cycle and usage of these two json object are so different.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -197,13 +198,28 @@ impl EntityStore {
         self.entities.get(entity_topic_id)
     }
 
+    /// Returns a mutable reference to the `EntityMetadata` for the given `EntityTopicId`.
+    pub fn get_mut(&mut self, entity_topic_id: &EntityTopicId) -> Option<&mut EntityMetadata> {
+        self.entities.get_mut(entity_topic_id)
+    }
```
Bravo555, line 204: issue: Having this method be `pub` is unsafe, it allows the caller to set parent to an entity that doesn't exist, invalid `EntityExternalId`, etc. It also doesn't handle other affected entities: i.e. returning information that when a new entity is created, it affects its parent. This was part of the original specification, but AFAIK it's rarely used, so it's not a problem, but unchecked access to `EntityMetadata` is a problem.

This function is only used once in this module though, so we can just remove the `pub`.

crates/core/tedge_api/src/entity_store.rs
Bravo555: I agree, properties related to thin-edge and ones related to C8y are all together and it's hard to know which ones are which. Some properties inside `twin_data` like `name` and `type` need to be treated differently from the other fields, and it's easy to make a mistake like that. We need to think about more clearly separating them.

Bravo555 (changes requested):
One small issue needs to be fixed

crates/extensions/c8y_mapper_ext/src/inventory.rs
```diff
@@ -50,12 +41,14 @@ impl CumulocityConverter {
 
         if let JsonValue::Object(map) = inventory_base {
             for (key, value) in map {
-                self.update_inventory_model(key.clone(), value.clone());
-                let mapped_message = self.register_twin_data_message(
-                    &EntityTopicId::default_main_device(),
+                let main_device_tid = EntityTopicId::default_main_device();
```
didier-wenzek, line 44: Instead of assuming default naming, it would be better to request the entity store.

```suggestion
                let main_device_tid = self.entity_store.main_device;
```

didier-wenzek (approved):
Approved - with a suggestion.

Issue #2251, opened 2023-09-13
tedge-mapper-c8y clears log_upload commands which did not originate from it


reubenmiller:
**Describe the bug**

The tedge-mapper-c8y is clearing `log_upload` operations which were not requested by it, which results in the chance to give the original requestor confirmation that they have seen the result before the retained operation message is cleared.

**To Reproduce**

1. Install thin-edge.io
2. Make sure the following services are running
    * tedge-mapper-c8y
    * tedge-agent
    * tedge-log-plugin

3. Open a new bash shell and subscribe to the log_upload command types. Leave 

    ```
    tedge mqtt sub 'te/+/+/+/+/cmd/log_upload/+'
    ```

4. Open a new bash shell and paste the following snippet to define a function helper which will be used to create the log file request mqtt messages in the following step

    ```sh
    get_log() {
        MAX_LINES="10000"
        SEARCH_TEXT=""
        LOG_TYPE="example"
        device_id=$(tedge config get device.id)
        cmd_id=$(echo "id-$RANDOM")
        payload=$(printf '{
            "status":"init",
            "tedgeUrl":"http://127.0.0.1:8000/tedge/file-transfer/%s/log_upload/example-%s",
            "type":"%s",
            "dateFrom":"2023-09-12T00:10:27Z",
            "dateTo":"2023-09-13T00:11:27Z",
            "searchText":"%s",
            "lines":%s
        }' "$device_id" "$cmd_id" "$LOG_TYPE" "$SEARCH_TEXT" "$MAX_LINES")
        set -x
        tedge mqtt pub --retain "te/device/main///cmd/log_upload/$cmd_id" "$payload"
        set +x

        echo ""
        echo "You can download the log file from"
        echo "  curl -f http://127.0.0.1:8000/tedge/file-transfer/$device_id/log_upload/example-$cmd_id"
        echo
    }
    ```

    **Note** It assumes that you have configured an "example" log file type in the `/etc/tedge/plugins/tedge-log-plugin.toml` file. If you have trouble setting it up, use the integration test under: `tests/RobotFramework/tests/cumulocity/log/log_operation.robot`

5. Create a log_upload request by using the helper function defined in the previous step

    ```sh
    get_log
    ```

6. Inspect the `tedge mqtt sub` command in the other shell to check which messages have been published.

    The expectation is that you should NOT see the publishing of an empty MQTT payload which will manifest in the form of (where the id will vary as it uses a randomly generated id):

    ```sh
[...]

PR #2381, opened 2023-10-27
Add requester prefix with timestamp to command operations

Ruadhri17:
## Proposed changes
This PR fixes issue of #2251 with clearing log upload commands but not being a requester. It was also applied to config operations as the problem is basically the same. To make command ids more clearer and easier to debug, current `nanoid` was replaced with `RFC3339` timestamp.

## Types of changes

## Paste Link to the issue
#2251 
<br/>

## Checklist


reubenmiller:
@Ruadhri17 Can you please add a system test?

Ruadhri17:
> 1. Can you please add a system test?

Sure

> 2. Does this PR apply to all of the tedge commands? e.g. what happens if a local `restart` command is issued, does the mapper still react to it?

For now only for log and config operations. However, it is possible to apply it to e.g restart though the structure is a little bit different. I will continue working on that.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -628,6 +631,18 @@ pub enum ChannelFilter {
     CommandMetadata(OperationType),
 }
 
+pub fn command_id_requester_with_timestamp(requester: &str) -> Result<String, error::Format> {
```
didier-wenzek, line 634: I would really avoid to return an error here => any formatting error must be handled internally by this function.
It can be using another format e.g. an unix timestamp instead.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -628,6 +631,18 @@ pub enum ChannelFilter {
...
+    Ok(format!(
+        "{}-{}",
+        requester,
+        OffsetDateTime::now_utc().format(&format_description::well_known::Rfc3339)?,
```
didier-wenzek, line 638: Is this date time precise enough to avoid collision if the mapper triggers two operations in a row? Not sure.

Hence, I will add a sequence number. For that you might have to change the API.

```
pub struct IdGenerator {
    prefix: String,
    next_id: usize, 
}

impl IdGenerator {
    pub fn new(prefix: &str) -> Self { ... }
    pub fn new_id(&mut self) -> String {
        // increment self.next_id
        // combine prefix + date + sequence into a unique id
    }
   pub fn is_generator_of(&self, &str) {
       // check the prefix
   }
}
```

didier-wenzek:
>> 2. Does this PR apply to all of the tedge commands? e.g. what happens if a local `restart` command is issued, does the mapper still react to it?

> For now only for log and config operations. However, it is possible to apply it to e.g restart though the structure is a little bit different. I will continue working on that.

Yes, this should not cause specific issues.

crates/core/tedge_api/src/mqtt_topics.rs
Ruadhri17: I am wondering, can't we just `unwrap` this value? It is delivered by `OffsetDateTime::now_utc()` so we should always be able to format it into `RFC3339`.

crates/core/tedge_api/src/mqtt_topics.rs
Ruadhri17: This is exemplary date in that format : `2023-10-30T19:58:26.649479831Z`  so I think for sequence of operations it should be more than enough to provide unique date. However, If we follow the solution above, how do we determine that counter should be incremented? Or by sequence number you mean randomly generated number?

didier-wenzek: > This is exemplary date in that format : `2023-10-30T19:58:26.649479831Z`  so I think for sequence of operations it should be more than enough to provide unique date.

Using such a precision should indeed be enough (even if the nano figures is likely not reliable).

> However, If we follow the solution above, how do we determine that counter should be incremented? Or by sequence number you mean randomly generated number?

I mean a sequence number incremented each time a fresh identifier is requested.

- `c8y-mapper-2023-10-30T19:58:26-1`
- `c8y-mapper-2023-10-30T20:00:00-2`
- ...

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: `unwrap()` is indeed okay.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
```diff
@@ -30,13 +30,26 @@ Request with non-existing log type
...
+    Execute Command    sudo -u tedge mkdir -p /var/tedge/file-transfer/${DEVICE_SN}/log_upload
+    Execute Command    sudo -u tedge touch /var/tedge/file-transfer/${DEVICE_SN}/log_upload/example-1234
+    ${start_timestamp}=    Get Current Date    UTC    -24 hours    result_format=%Y-%m-%dT%H:%M:%SZ
+    ${end_timestamp}=    Get Current Date    UTC    +60 seconds    result_format=%Y-%m-%dT%H:%M:%SZ
+    Execute Command
+    ...    sudo tedge mqtt pub --retain 'te/device/main///cmd/log_upload/example-1234' '{"status":"init","tedgeUrl":"http://127.0.0.1:8000/tedge/file-transfer/${DEVICE_SN}/log_upload/example-1234","type":"example","dateFrom":"${start_timestamp}","dateTo":"${end_timestamp}","searchText":"first","lines":10}'
+    ${messages}=    Should Have MQTT Messages
+    ...    te/device/main///cmd/log_upload/example-1234
+    ...    minimum=3
+    ...    maximum=3
```
albinsuresh, line 43: I'd rather add the assertion for the clear message with `minimum=0` and `maximum=0`. That'd be clearer I feel.

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: To clarify my point of view.

- I have no strong opinion about the precise string format used to generate used to generate command identifiers as long as they are unique and providing some hints about their provenance and age.
- I do think that wrapping the prefix in a `struct IdGenerator` equipped with methods to generate and check identifier with a given prefix is a better design compared to a bunch of functions and string constants. Using a struct leads to better encapsulation (I agree that here there not so much to abstract) and is easier to extend (say by adding later a sequence number).
- In too many places, the thin-edge code base started with constants such as `REQUESTER_NAME` leading to a code base that this is easy to break and hard to change. Instead of this constant, the mapper needs an id generator that encapsulates the prefix and the associated logic.

rina23q: I think nano seconds level is unique enough.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
```diff
@@ -30,13 +30,26 @@ Request with non-existing log type
     ...    failure_reason=.*No such file or directory for log type: example1
     ...    timeout=120
 
+Manual log operation request
```
rina23q, line 33: We also need a similar test for config_update and config_snapshot.

crates/extensions/c8y_mapper_ext/src/config_operations.rs
```diff
@@ -171,22 +173,33 @@ impl CumulocityConverter {
                 };
 
                 let c8y_notification = Message::new(&smartrest_topic, smartrest_operation_status);
-                let clear_local_cmd = Message::new(&message.topic, "")
-                    .with_retain()
-                    .with_qos(QoS::AtLeastOnce);
-                vec![c8y_notification, clear_local_cmd]
+
+                if is_command_requester(cmd_id, REQUESTER_NAME) {
```
rina23q, line 177: Moment. If a command does not belong to the requester, the mapper should not even publish a status update message to c8y??? I think the mapper should just ignore the message.

rina23q (commented):
I need the clarification to the behaviour when the command is not issued by the requester.

I really believe that the mapper that is not the command issuer should ignore the message. It means c8y-mapper should not publish any status change messages (e.g. EXECUTING, FAILED, SUCCESSFUL) to c8y if the command ID doesn't contain `c8y-mapper`, which means the command must be addressed by the other mapper.

didier-wenzek:
> I really believe that the mapper that is not the command issuer should ignore the message. It means c8y-mapper should not publish any status change messages (e.g. EXECUTING, FAILED, SUCCESSFUL) to c8y if the command ID doesn't contain `c8y-mapper`, which means the command must be addressed by the other mapper.

Exactly. The mapper should only emit and process commands with an identifier matching its prefix.

didier-wenzek (commented):
The code is correct and I will be happy to approve once the tests fixed.

The unit tests have to be updated with appropriate prefix:

```
-            &Topic::new_unchecked("te/device/child1///cmd/config_snapshot/1234"),
+            &Topic::new_unchecked("te/device/child1///cmd/config_snapshot/c8y-mapper-1234"),
```

The failing system test is flaky.

didier-wenzek:
I pushed a fix to use the same prefix for all commands forwarded to the mapper.

reubenmiller:
> I pushed a fix to use the same prefix for all commands forwarded to the mapper.

There seems to be other build issues here

didier-wenzek (approved):
Approved

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
reubenmiller: I'll these system tests

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
reubenmiller: Though it is a bit hard to check for the absence of a message when using the mqtt-logger (the historical record of all mqtt messages which are sent, which is used by they keyword). However I'll rewrite the test as follows:

1. Publish local cmd (with status="init")
2. Wait for topic to be updated with a status="successful"
3. Wait 5 seconds (a fixed period which is assumed that if the mapper has not reacted within 5 seconds then it will never react on the message)
4. Check that the last retained message on the topic still exists and status="successful"

crates/extensions/c8y_mapper_ext/src/config_operations.rs
reubenmiller: In 673135cac7ce86cecc4380699234141ead537675, I've also added checks to make sure Cumulocity operation transitions are not sent to Cumulocity for locally triggered operations.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
reubenmiller: Systems tests have been provided in 673135cac7ce86cecc4380699234141ead537675 to cover more commands.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
reubenmiller: Resolved by 673135cac7ce86cecc4380699234141ead537675

PR #2385, opened 2023-10-30
fix #2383: Use "service" if service type missing

### LINKED ISSUE #2383: empty service.type value leads to invalid c8y smart rest message being sent
**Describe the bug**

tedge-mapper-c8y is sending invalid service registration messages to Cumulocity IoT when the `service.type` setting is set to an empty string.

Below shows the Cumulocity smart rest message to register a service (message id `102`), and the `,,` shows that the service type is missing. Cumulocity also publishes an error message on the local `c8y/s/e` topic indicating that the service name and type can not be empty values.

```log
2023-10-28T11:17:01.996203Z      c8y/s/us           102,TST_transform_perpendicular_burrito:device:main:service:tedge-agent,,tedge-agent,up
2023-10-28T11:17:02.138625Z      c8y/s/e            41,102,Service uniqueID, type and name are required
```

**To Reproduce**

The test case is already covered by the following system test, however the test case seems to be flakey.

[Test if all c8y services using default service type when service type configured as empty](https://github.com/thin-edge/thin-edge.io/blob/0e42b3734fd2c2dbe04263524b8565d49f1a0d2a/tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot#L49)
```

But the following manual procedure can be executed to observer the error:

1. Connect thin-edge.io to Cumulocity IoT
2. Configure the service.type

    ```
    tedge config set service.type ""
    ```

3. Restart the mapper

    ```sh
    sudo systemctl restart tedge-mapper-c8y
    ```

4. Open a different console and subscribe to the local cumulocity topic

    ```sh
    tedge mqtt sub 'c8y/#'
    ```

5. Restart the tedge-agent

    ```
    sudo systemctl restart tedge-agent
    ```

6. Observe the output from the `tedge mqtt sub` command in the previous steps

    The bug is visible if there is an error message received on the `c8y/s/e` topic indicating that the service name/type should not be empty.

**Expected behavior**

A non-empty value should always be used in the service type field of the `102` smart rest message, and there should not be any corresponding `c8y/s/e` message received.

An empty value used by `service.type` should be treated the same as an unset value (e.g. using `service` as the default value).

**Screenshots**

 - OS [incl. version]: Debian 11
 - Hardware [incl. revision]: Docker
 - System-Architecture [e.g. result of "uname -a"]: aarch64
 - thin-edge.io version [e.g. 0.1.0]: `0.12.1~501+g7259158`

**Additional context**

### PR DESCRIPTION
**Pull request by Bravo555**
## Proposed changes

Some additions:

1. A C8y converter test was added to make sure that the emitted service creation message is valid if `service.type` is empty.

2. `c8y_api::inventory` functions which created C8y smartrest messages were changed to return `Result`. Currently we only check if values marked as mandatory in [SmartREST reference documentation][1] are not empty, but we could do more validation in the future.

3. Added `ConversionError::UnexpectedError` variant, which can be constructed from an `anyhow::Error`.

   _Unexpected_ part is supposed to denote that the error was not related due to invalid input, or the user doing something wrong, but some unexpected conditions occured that makes it so that the request cannot be completed. Like a `panic!`, it can be used when there is a logic error on part of the programmer, but unlike `panic!` it doesn't crash the entire program. More about the idea [here][2].

[1]: https://cumulocity.com/guides/reference/smartrest-two/#102
[2]: https://www.lpalmieri.com/posts/error-handling-rust/#avoid-ball-of-mud-error-enums

## Types of changes

## Paste Link to the issue

- #2383

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/smartrest/inventory.rs
+++ b/crates/core/c8y_api/src/smartrest/inventory.rs
@@ -22,8 +22,27 @@ pub fn child_device_creation_message(
     device_name: Option<&str>,
     device_type: Option<&str>,
     ancestors: &[String],
-) -> Message {
-    Message::new(
+) -> Result<Message, InvalidValueError> {
+    if child_id.is_empty() {
+        return Err(InvalidValueError {
+            field_name: "child_id".to_string(),
+            value: child_id.to_string(),
+        });
+    }
+    if let Some("") = device_name {
+        return Err(InvalidValueError {
+            field_name: "device_name".to_string(),
+            value: "".to_string(),
+        });
+    }
+    if let Some("") = device_type {
+        return Err(InvalidValueError {
+            field_name: "device_type".to_string(),
+            value: "".to_string(),
+        });
+    }
+
+    Ok(Message::new(
         &publish_topic_from_ancestors(ancestors),
         // XXX: if any arguments contain commas, output will be wrong
         format!(
@@ -32,7 +51,7 @@ pub fn child_device_creation_message(
             device_name.unwrap_or(child_id),
             device_type.unwrap_or("thin-edge.io-child")
         ),
-    )
+    ))
 }
 
 /// Create a SmartREST message for creating a service on device.
@@ -44,15 +63,41 @@ pub fn service_creation_message(
     service_type: &str,
     service_status: &str,
     ancestors: &[String],
-) -> Message {
-    Message::new(
+) -> Result<Message, InvalidValueError> {
+    // TODO: most of this noise can be eliminated by implementing `Serialize`/`Deserialize` for smartrest format
+    if service_id.is_empty() {
+        return Err(InvalidValueError {
+            field_name: "service_id".to_string(),
+            value: service_id.to_string(),
+        });
+    }
+    if service_name.is_empty() {
+        return Err(InvalidValueError {
+            field_name: "service_name".to_string(),
+            value: service_name.to_string(),
+        });
+    }
+    if service_type.is_empty() {
+        return Err(InvalidValueError {
+            field_name: "service_type".to_string(),
+            value: service_type.to_string(),
+        });
+    }
+    if service_status.is_empty() {
+        return Err(InvalidValueError {
+            field_name: "service_status".to_string(),
+            value: service_status.to_string(),
+        });
+    }
+
+    Ok(Message::new(
         &publish_topic_from_ancestors(ancestors),
         // XXX: if any arguments contain commas, output will be wrong
         format!(
             "102,{},{},{},{}",
             service_id, service_type, service_name, service_status
         ),
-    )
+    ))
 }
 
 /// Create a SmartREST message for updating service status.
@@ -80,4 +125,11 @@ pub fn service_status_update_message(external_ids: &[String], service_status: &s
     let payload = format!("104,{service_status}");
 
     Message::new(&topic, payload)
+}
+
+#[derive(thiserror::Error, Debug)]
+#[error("Field `{field_name}` contains invalid value: {value:?}")]
+pub struct InvalidValueError {
+    field_name: String,
+    value: String,
 }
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: crates/extensions/c8y_mapper_ext/src/converter.rs [hunk 1]
```diff
@@ -190,7 +191,13 @@ impl CumulocityConverter {
         let device_id = config.device_id.clone();
         let device_topic_id = config.device_topic_id.clone();
         let device_type = config.device_type.clone();
-        let service_type = config.service_type.clone();
+
+        let service_type = if config.service_type.is_empty() {
+            "service".to_string()
+        } else {
+            config.service_type.clone()
+        };
```
rina23q, line 199: This implementation is defenitely fine as of now, especially with the restricted time limit for the new release.

However, I've got the feeling that the ideal solution is to change the tedge_config API; If user tries to set an empty string, it works the same as "unset". The "service" is already used as the default value of `service.type`. It's a little pity that we need to hard-code the same value here.

crates/core/c8y_api/src/smartrest/inventory.rs
```diff
@@ -44,15 +51,41 @@ pub fn service_creation_message(
     service_type: &str,
     service_status: &str,
     ancestors: &[String],
-) -> Message {
-    Message::new(
+) -> Result<Message, InvalidValueError> {
+    // TODO: most of this noise can be eliminated by implementing `Serialize`/`Deserialize` for smartrest format
```
rina23q, line 55: As you said here, why not implementing the representation struct for SmartREST 101 and 102? We already have them for other message IDs, for operations. 
https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/c8y_api/src/smartrest/smartrest_serializer.rs

rina23q (approved):
It looks good as it is now although there can be improvements.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2708,12 +2718,48 @@ pub(crate) mod tests {
...
+        let mut smartrest_fields = service_creation_message.payload_str().unwrap().split(',');
+
+        assert_eq!(smartrest_fields.next().unwrap(), "102");
+        assert_eq!(
+            smartrest_fields.next().unwrap(),
+            format!("{}:device:main:service:service1", converter.device_name)
+        );
+        assert_eq!(smartrest_fields.next().unwrap(), "service");
+        assert_eq!(smartrest_fields.next().unwrap(), "service1");
+        assert_eq!(smartrest_fields.next().unwrap(), "up");
```
albinsuresh, line 2749: You could use that helper.
```suggestion
        assert_messages_matching(
            &messages,
            [
                (
                    SMARTREST_PUBLISH_TOPIC,
                    "102,test-device:device:main:service:service1,service,service1,up".into(),
                ),
            ],
        );
```

albinsuresh (approved):
LGTM

reubenmiller:
@Bravo555 Looking at the [system test report](https://github.com/thin-edge/thin-edge.io/actions/runs/6707618984/job/18227099720?pr=2385) the error looks to be legitimate. The following error was found in the log output of the test:

```
Oct 31 14:27:58 dcd30d07c97a tedge-mapper[1264]: Error: Field `service_type` contains invalid value: ""
```

Bravo555:
@reubenmiller Yes, I missed one place where empty service type was passed to another actor. It seems to be that handling it on the `tedge_config` level is the only resilient solution, so it's not possible for `service.type` to be empty in the first place. Will add this change to this PR.

didier-wenzek (approved):
Approved, to include this fix in the upcoming release.

However, this will have to be addressed [at the config level](resilient) as suggested by @rina23q and acknowledged by @Bravo555 as a [less fragile solution](https://github.com/thin-edge/thin-edge.io/pull/2385#issuecomment-1787421575)

PR #2397, opened 2023-11-02
Add certificate based authentication to c8y-auth-proxy

jarhodes314:
## Proposed changes
This adds HTTPS support and (opt-in) certificate-based authentication to the c8y auth proxy.

At the moment this is a draft PR as the agent currently doesn't handle the device certificate not being present. It also doesn't yet document the API changes (HTTP -> HTTPS), nor does it touch the integration tests.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

When the only authentication scheme available for software updates was bearer tokens, the mapper simply supplied the token along with the request, which the agent would simply use. As this adds certificate authentication, the mapper requires the agent to also use certificate authentication (if configured). This has the advantage of us being able to delegate authentication largely to a trusted library, but does add complexity in the agent code and adds possible error states, due to the way certificates need to be configured.

The most significant changes to the agent centre around the generic authentication strategy supported by `DownloadInfo`. `DownloadInfo` has been modified to support an incoming request from the mapper (which would simply tell the agent to use a certificate, if required), a request that is being processed (at which point the agent will add the required certificate) and a response (which has the authentication anonymised for security/because we cannot clone `reqwest::Identity`, which represents the certificate/key pair in use).

Due to this complexity, it may be desirable to add an alternative authentication scheme to the auth proxy. This could involve the mapper generating its own bearer token, and the agent consuming this as it does currently. This would ensure backwards compatibility with the existing software update API (if someone is using a custom agent implementation). It would also remove the burden of the agent (possibly) needing to supply certificates, which is difficult to handle errors for proactively before a user attempts to e.g. update software and finds the agent configuration is broken. The difficulty arises from the fact certificates may not be required, either due to not specifying trusted certificates for the auth proxy, or not connecting thin-edge to Cumulocity (in which case the auth proxy isn't running). If the mapper could decide itself how to authenticate a request to itself (e.g. by creating a JWT using the certificate used for the HTTPS connection), this alleviates the need for the agent to care how the request should be authenticated. Certificate-based authentication would still be useful for thin-edge users to connect to the API themselves (if no certificates are provided, the mapper could just disable authentication entirely, both for external users and agents).

### DIFF
--- a/crates/common/download/src/download.rs
+++ b/crates/common/download/src/download.rs
@@ -1,4 +1,5 @@
 mod partial_response;
+
 use crate::error::DownloadError;
 use crate::error::ErrContext;
 use anyhow::anyhow;
@@ -10,6 +11,7 @@ use log::warn;
 use nix::sys::statvfs;
 pub use partial_response::InvalidResponseError;
 use reqwest::header;
+use reqwest::Identity;
 use serde::Deserialize;
 use serde::Serialize;
 use std::fs;
@@ -29,6 +31,8 @@ use tedge_utils::file::PermissionEntry;
 use nix::fcntl::fallocate;
 #[cfg(target_os = "linux")]
 use nix::fcntl::FallocateFlags;
+use reqwest::header::HeaderName;
+use reqwest::header::HeaderValue;
 
 fn default_backoff() -> ExponentialBackoff {
     // Default retry is an exponential retry with a limit of 15 minutes total.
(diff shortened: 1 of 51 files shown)


crates/common/download/src/download.rs
```diff
@@ -79,16 +83,115 @@ impl DownloadInfo {
...
 #[derive(Debug, Clone, Deserialize, PartialEq, Eq, Serialize)]
 #[serde(rename_all = "camelCase")]
 #[serde(deny_unknown_fields)]
-pub enum Auth {
+pub enum RequiredAuth {
     /// HTTP Bearer authentication
     Bearer(String),
+    /// TLS certificate authentication
+    Certificate,
+}
```
didier-wenzek, line 107: I don't understand this change. This is even going in the reverse direction I was expecting.

- In their first version, `tedge-mapper` and `tedge-agent` were exchanging over MQTT the credentials required to download a software package from c8y. This was not only a security issue (the `Bearer` being sent clear-text over MQTT), but also fragile (as the JWT token might expired before being used).
- The introduction of `c8y_auth_proxy` makes the authentication to c8y an internal issue of the mapper. This is less fragile, but still not secure as any local process has full access to the c8y API which is exposed over non-TLS HTTP.
- This PR aims to add HTTPS support with mutual TLS authentication of the mapper and the agent (the agent is possibly running on a different device). Each needs its own certificate (and private key). Each needs to trust the certificate of the other. But *no identity and even no credentials have to be exchanged over MQTT* between the mapper and the agent.
- Hence, *I was expecting the `enum Auth` to be deprecated*, not augmented.

didier-wenzek (commented):
Before going further, it would be good to sketch a user documentation telling how to deploy certificates over misc thin-edge devices for the c8y_auth_proxy to serve only HTTPS requests to authenticated services. Such a documentation would really help to clarify the design and understand the code.

jarhodes314:
This is now entirely superseeded by #2430

PR #2399, opened 2023-11-02
Tedge operation plugins to migrate legacy config file contents

albinsuresh:
## Proposed changes

The `tedge-config-plugin` and `tedge-log-plugin` will migrate their legacy counterparts' configuration files, if found during initialization.

- [x] tedge-config-plugin code update
- [x] tedge-log-plugin code update
- [x] System tests

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_config_manager/src/lib.rs
+++ b/crates/extensions/tedge_config_manager/src/lib.rs
@@ -7,6 +7,7 @@ mod tests;
 
 use actor::*;
 pub use config::*;
+use std::fs::rename;
 use std::path::PathBuf;
 use tedge_actors::futures::channel::mpsc;
 use tedge_actors::Builder;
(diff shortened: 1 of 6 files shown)


reubenmiller:
I've modified the existing system test, and added one for the tedge-log-configuration as well.

crates/extensions/tedge_config_manager/src/lib.rs
```diff
@@ -88,9 +89,22 @@ impl ConfigManagerBuilder {
...
+
         // creating plugin config parent dir
         create_directory_with_defaults(&config.plugin_config_dir)?;
 
+        let legacy_plugin_config = config
+            .config_dir
+            .join("c8y")
+            .join("c8y-configuration-plugin.toml");
+        if legacy_plugin_config.exists() {
+            rename(legacy_plugin_config, &config.plugin_config_path)?;
```
didier-wenzek, line 104: It would have been safer to use a non async version of `tedge_utils::file::move_file()`, but that's okay as it's so unlikely to have /etc/tedge/c8y and /etc/tedge/plugins on different filesystems.

didier-wenzek (approved):
Approved

crates/extensions/tedge_config_manager/src/lib.rs
albinsuresh: Yeah, that's exactly the reason why I went with this, to keep the diff small. I had started with that `move_file` API itself, but there were too many lines that I had to change (less risky ones, and mostly tests) to propagate that async call function call all the way up the call stack.

tests/RobotFramework/tests/cumulocity/configuration/backward_compatibility.robot
```diff
@@ -0,0 +1,25 @@
...
+Library             Cumulocity
+
+Test Teardown      Get Logs    name=${DEVICE_SN}
+
+Test Tags          theme:configuration    theme:installation
+
+*** Test Cases ***
+
+Migrate Legacy Configuration Files
+    ${DEVICE_SN}=    Setup    skip_bootstrap=${True}
```
albinsuresh, line 13: Any reason why you use `${True}` instead of `True` for `skip_bootstrap`? From what I read, `${True}` passes the value as a string where as the `skip_bootstrap` arg is annotated as `boolean` type. Somehow both seems to work. Just wanted to know if this is better for some reason.

PR #2398, opened 2023-11-02
Add a BridgeConfig field to include or not local_cleansession

didier-wenzek:
Add a new `tedge config` setting .

`c8y.include.local_cleansession` can be set either to `true`,  `false` or `auto`:
- [x] `c8y.include.local_cleansession = false` is the default.  No `local_cleansession` property is added the to bridge configuration file which can then be used with any version of mosquiito.
- [x] `c8y.include.local_cleansession = auto` is the recommended setting. With this setting, the `tedge connect c8y` command inserts a  `local_cleansession` setting but only if the version of mosquitto is above 2.0.
- [x] `c8y.include.local_cleansession = true` can be use to force the appropriate setting of `local_cleansession` when `tedge` fails to detect that mosquitto is running a version above 2.0.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2184

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2401, opened 2023-11-02
Add a custom error message for invalid use of `default(value = ...)` in `define_tedge_config!`

jarhodes314:
## Proposed changes
https://github.com/thin-edge/thin-edge.io/pull/2398/files/313cc0efaac938fe47b1267d63a3d89cbb31c312..954da3d0246b6e6db4b5cd3d41421f5864ad9088#r1380338718 highlighted a usability issue with `define_tedge_config` and default values, which can either be provided using `default(value = true)` or `default(variable = "Ipv4Addr::LOCALHOST")`. The distinction between these two formats is important, but subtle.

This PR adds a custom error message for the case which caused confusion, pointing the user towards a suggested solution. See the screenshot below (or the added test) for the error that appears:

image

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config_macros/impl/src/lib.rs
+++ b/crates/common/tedge_config_macros/impl/src/lib.rs
@@ -202,4 +202,18 @@ mod tests {
         })
         .unwrap();
     }
+
+    #[test]
+    fn error_message_suggests_fix_in_case_of_invalid_value() {
+        assert_eq!(generate_configuration(quote! {
+            http: {
+                #[tedge_config(default(value = Ipv4Addr::LOCALHOST))]
+                address: Ipv4Addr,
+            },
+        })
+                       .unwrap_err()
+                       .to_string(),
+                   "Unexpected expression, `default(value = ...)` expects a literal.\n\
+            Perhaps you want to use `#[tedge_config(default(variable = \"Ipv4Addr::LOCALHOST\"))]`?");
+    }
 }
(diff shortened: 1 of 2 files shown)


crates/common/tedge_config_macros/impl/src/lib.rs
```diff
@@ -202,4 +202,18 @@ mod tests {
...
+        assert_eq!(generate_configuration(quote! {
+            http: {
+                #[tedge_config(default(value = Ipv4Addr::LOCALHOST))]
+                address: Ipv4Addr,
+            },
+        })
+                       .unwrap_err()
+                       .to_string(),
+                   "Unexpected expression, `default(value = ...)` expects a literal.\n\
+            Perhaps you want to use `#[tedge_config(default(variable = \"Ipv4Addr::LOCALHOST\"))]`?");
```
didier-wenzek, line 217: Nitpicking.

To me an expression such as `Ipv4Addr::LOCALHOST` is a literal, i.e. an expression which evaluates to itself.
Maybe a lack of knowledge on Rust terminology.

That said the hint `Perhaps you want to use ... variable = \"Ipv4Addr::LOCALHOST\"` is wonderful!

didier-wenzek (approved):
Approved. This adds nice support for using the tedge config macro.

crates/common/tedge_config_macros/impl/src/lib.rs
jarhodes314: The reference explains quite nicely what a literal is: https://doc.rust-lang.org/beta/reference/expressions/literal-expr.html#:~:text=A%20literal%20expression%20is%20an,or%20some%20other%20evaluation%20rule.

> A literal expression is an expression consisting of a single token, rather than a sequence of tokens, that immediately and directly denotes the value it evaluates to, rather than referring to it by name or some other evaluation rule.

didier-wenzek: Indeed, I missed this *single* token constraint.

Issue #2391, opened 2023-10-31
use independent tedge.toml settings to access the http server provided by tedge-agent


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The `http.bind.address` setting is used by two different components but for different purposes,

1. tedge-agent interprets this setting as the bind address where the http server should be made available
2. tedge-mapper-c8y interprets this setting as the target ip address when uploading files to the tedgeUrl.

The different in the interpretation of the two settings makes it difficult to configure the tedge-mapper-c8y address, as it needs to know the ip address of the tedge-agent, and it can't use a hostname (as the setting only accepts an ip address, in the form of `x.x.x.x`). When using thin-edge.io in containers generally the container does not know the ip address of each other upfront, but rather at runtime, that is where using a hostname would be more useful when telling the tedge-mapper-c8y when creating the tedgeUrl. 

This problem is not too dissimilar to the changes which prompted splitting the `mqtt.bind.address` and `mqtt.client.host` settings.

**Describe the solution you'd like**

Create the following new settings which are used by the tedge-mapper-c8y when it needs to communicate with the http server provided by the tedge-agent.

* `http.client.host` - ip or hostname of the http server provided by the tedge-agent
* `http.client.port` - port of the http server provided by the tedge-agent

**Describe alternatives you've considered**

**Additional context**

Bravo555:
# QA Reproduction instructions

1. Create two containers (or physical devices, but I'll call them containers for simplicity):
	1. **container-A**: full thin-edge.io except tedge-agent
	2. **container-B**: only tedge-agent

2. Configure containers to point to each other:
  ```
  # on containter A:
  $ tedge config set mqtt.bind.address 0.0.0.0
  $ tedge config set http.client.host <container-B>

  # on container B:
  $ tedge config set mqtt.client.host <container-A>
  $ tedge config set http.bind.address 0.0.0.0
  ```

3. Verify that all services on `<container-A>` and `<container-B>` correctly start up
4. For operations which contain `tedgeUrl` property, ensure that the hostname is equal to `http.client.host`

gligorisaev:
QA has thoroughly checked the improvement and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/alpine-s6/children/operations.robot
- tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
- tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
[...]

PR #2411, opened 2023-11-03
add http.client.host/port setting options

Bravo555:
## Proposed changes

This commit introduces new `http.client.host` and `http.client.port` setting options. Similarly to the `mqtt.client.*` options, there is a significant difference between an address used to bind, e.g. "0.0.0.0" and an address used to connect.

`http.client.host` is now a string, which makes it possible to use DNS names where previously only IP address could be used.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2391

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -16,6 +16,7 @@ use std::net::IpAddr;
 use std::net::Ipv4Addr;
 use std::num::NonZeroU16;
 use std::path::PathBuf;
+use std::sync::Arc;
 use tedge_config_macros::all_or_nothing;
 use tedge_config_macros::define_tedge_config;
 use tedge_config_macros::struct_field_aliases;
@@ -560,15 +561,26 @@ define_tedge_config! {
 
     http: {
         bind: {
-            /// Http server port used by the File Transfer Service
+            /// The port number of the File Transfer Service HTTP server binds to for internal use
             #[tedge_config(example = "8000", default(value = 8000u16), deprecated_key = "http.port")]
             port: u16,
 
-            /// Http server address used by the File Transfer Service
-            #[tedge_config(default(function = "default_http_address"), deprecated_key = "http.address")]
-            #[tedge_config(example = "127.0.0.1", example = "192.168.1.2")]
+            /// The address of the File Transfer Service HTTP server binds to for internal use
+            #[tedge_config(default(function = "default_http_bind_address"), deprecated_key = "http.address")]
+            #[tedge_config(example = "127.0.0.1", example = "192.168.1.2", example = "0.0.0.0")]
             address: IpAddr,
         },
+
+        client: {
+            /// The port number on the remote host on which the File Transfer Service HTTP server is running
+            #[tedge_config(example = "8000", default(value = 8000u16))]
+            port: u16,
+
+            /// The address of the host on which the File Transfer Service HTTP server is running
+            #[tedge_config(default(value = "127.0.0.1"))]
+            #[tedge_config(example = "127.0.0.1", example = "192.168.1.2", example = "tedge-hostname")]
+            host: Arc<str>,
+        }
     },
 
     software: {
@@ -641,7 +653,7 @@ define_tedge_config! {
     }
 }
 
-fn default_http_address(dto: &TEdgeConfigDto) -> IpAddr {
+fn default_http_bind_address(dto: &TEdgeConfigDto) -> IpAddr {
     let external_address = dto.mqtt.external.bind.address;
     external_address
         .or(dto.mqtt.bind.address)
--- a/crates/extensions/c8y_config_manager/src/config.rs
+++ b/crates/extensions/c8y_config_manager/src/config.rs
@@ -1,9 +1,9 @@
 use super::child_device::ConfigOperationResponseTopic;
 
 use c8y_api::smartrest::topic::C8yTopic;
-use std::net::IpAddr;
 use std::path::Path;
 use std::path::PathBuf;
+use std::sync::Arc;
 use tedge_api::path::DataDir;
 use tedge_config::ReadError;
 use tedge_config::TEdgeConfig;
@@ -22,7 +22,7 @@ pub struct ConfigManagerConfig {
     pub device_id: String,
     pub mqtt_host: String,
     pub mqtt_port: u16,
-    pub tedge_http_host: String,
+    pub tedge_http_host: Arc<str>,
     pub ops_dir: PathBuf,
     pub plugin_config_dir: PathBuf,
     pub plugin_config_path: PathBuf,
@@ -40,10 +40,10 @@ impl ConfigManagerConfig {
         device_id: String,
         mqtt_host: String,
         mqtt_port: u16,
-        tedge_http_address: IpAddr,
+        tedge_http_host: Arc<str>,
         tedge_http_port: u16,
     ) -> Self {
-        let tedge_http_host = format!("{}:{}", tedge_http_address, tedge_http_port);
+        let tedge_http_host = format!("{}:{}", tedge_http_host, tedge_http_port).into();
 
         let ops_dir = config_dir.join("operations/c8y");
         let plugin_config_dir = config_dir.join(DEFAULT_OPERATION_DIR_NAME);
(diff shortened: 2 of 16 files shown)


crates/extensions/c8y_config_manager/src/config.rs
```diff
@@ -84,8 +84,8 @@ impl ConfigManagerConfig {
         let data_dir = tedge_config.data.path.clone().into();
         let mqtt_host = tedge_config.mqtt.client.host.clone();
         let mqtt_port = tedge_config.mqtt.client.port.get();
-        let tedge_http_address = tedge_config.http.bind.address;
-        let tedge_http_port = tedge_config.http.bind.port;
+        let tedge_http_address = tedge_config.http.client.address;
```
reubenmiller, line 87: @Bravo555 We should use `host` instead of `address so that we're consistent with the other parts of the code.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -641,13 +653,20 @@ define_tedge_config! {
...
         .unwrap_or(Ipv4Addr::LOCALHOST.into())
 }
 
+fn default_http_address(dto: &TEdgeConfigDto) -> Arc<str> {
+    let mqtt_client_host = dto.mqtt.client.host.as_ref();
+    mqtt_client_host
+        .map(|s| Arc::from(s.as_str()))
+        .unwrap_or(Arc::from("127.0.0.1"))
+}
+
```
Bravo555, line 669: Would using `mqtt.client.host` as a default make sense here?

The problem is that there is a slight distinction between how `mqtt.client.host` and `http.client.host` is used in practice:
- `mqtt.client.host` should always be set on a child device because it's required to connect to the MQTT broker. There is no other channel where `mqtt.client.host` is passed to the child device by other thin-edge components, because to even begin to communicate with other thin-edge components this setting needs to be set. Setting this on a main device does not have any impact on child device behaviour.
- `http.client.host` setting is read by other thin-edge components, e.g. `c8y-configuration-plugin` and its value is put into MQTT messages directed towards the child device. This setting doesn't have to be set on a child device, but it must be set on the main device, so that the MQTT messages directed to the child device contain the correct address of the File Transfer Service.

This can potentially be very confusing to users, so perhaps some other name ought to be used.

tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
```diff
@@ -74,6 +74,7 @@ Child device config update
 Set external MQTT bind address
     Set Device Context    ${PARENT_SN}
     Execute Command    sudo tedge config set mqtt.external.bind.address ${PARENT_IP}
+    Execute Command    sudo tedge config set http.client.host ${PARENT_IP}
```
Bravo555, line 77: This change is not backwards-compatible, because the previous behaviour of deriving connect address from bind address was invalid in the first place. As such, the tests were changed, and similar change may be required on users installations as well. The old `http.bind.address` is not deprecated and it remains used, so I'm not sure if it's possible to change both in a backwards-compatible and correct manner.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
```diff
@@ -40,6 +41,34 @@ Manual log_upload operation request
...
+
+
+    ${start_timestamp}=    Get Current Date    UTC    -24 hours    result_format=%Y-%m-%dT%H:%M:%S+0000
+    ${end_timestamp}=    Get Current Date    UTC    +60 seconds    result_format=%Y-%m-%dT%H:%M:%S+0000
+    ${operation}=     Cumulocity.Create Operation
+    ...    description=Log file request
+    ...    fragments={"c8y_LogfileRequest":{"dateFrom":"${start_timestamp}","dateTo":"${end_timestamp}","logFile":"example","searchText":"first","maximumLines":10}}
+    Should Have MQTT Messages    te/device/main///cmd/log_upload/+    message_contains=tedgeUrl":"http://${child_ip}
+
+
```
Bravo555, line 71: This test currently checks if correct FTS address is used in the payload, but will be changed in a next PR to check if the operation actually succeeds.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
```diff
@@ -237,6 +237,7 @@ Nested child devices support sending event
  
 # Nested child device services 
 Nested child device service support sending simple measurements
+    Skip    msg=Flaky test (#2416)
```
Bravo555, line 240: Ignoring this test for now because it's flaky.

reubenmiller: Please don't skip the test in a unrelated PR.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
reubenmiller: We can ignore the usage of `c8y-configuration-plugin` as this component has been deprecated.

tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
reubenmiller: What is the default value of `http.bind.address`? I would assume `127.0.0.1`? If so then at least the existing configuration would still work.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
reubenmiller: I don't think we need to explicitly check the `tedgeUrl` address in the as this is just implementation detail. If the an invalid address is used then the operation will just fail anyway.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
reubenmiller: I wouldn't bother linking the `mqtt.client.host` and the `mqtt.bind.address` settings (when setting a default value). Both should use a default value of `127.0.0.1`.

reubenmiller (changes requested):
Added some comments, and main point is to remove the skipping of the system test (we should be doing this in a different PR).

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
reubenmiller: I'll monitor it and retrigger the systems tests if needed (even though it is annoying)

Bravo555: fixed in c0477e5

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
Bravo555: For now I added the `tedgeUrl` check to show that the current code change had an effect, I'll remove `tedgeUrl` assert in the next PR.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
```diff
@@ -40,6 +41,34 @@ Manual log_upload operation request
     ...    payload={"status":"init","tedgeUrl":"http://127.0.0.1:8000/tedge/file-transfer/${DEVICE_SN}/log_upload/example-1234","type":"example","dateFrom":"${start_timestamp}","dateTo":"${end_timestamp}","searchText":"first","lines":10}
     ...    c8y_fragment=c8y_DownloadConfigFile
 
+Log operation uses correct tedgeUrl when file transfer service on diffent host
```
reubenmiller, line 44: Fixing typo
```suggestion
Log operation uses correct tedgeUrl when file transfer service on different host
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -560,15 +561,26 @@ define_tedge_config! {
 
     http: {
         bind: {
-            /// Http server port used by the File Transfer Service
-            #[tedge_config(example = "8000", default(value = 8000u16), deprecated_key = "http.port")]
+            /// The port number of the File Transfer Service HTTP server binds to for internal use
+            #[tedge_config(example = "8000", example = "0", default(value = 8000u16), deprecated_key = "http.port")]
```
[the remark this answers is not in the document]
reubenmiller: Yes I would also avoid using port `0` as well as it will just lead to further questions from users.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
Bravo555: fixed bb43b50b34

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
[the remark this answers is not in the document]
Bravo555: fixed 4ef3acc827

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
Bravo555: fixed d47bfa70eb596b901d90598432c806d058aa112a

reubenmiller (approved):
Approved though please wait for @jarhodes314 to review it again

jarhodes314 (approved):

Issue #2394, opened 2023-10-31
c8y_http_proxy API `create_event` doesn't support child device


rina23q:
**Describe the bug**
While debugging the [test failure](https://github.com/thin-edge/thin-edge.io/actions/runs/6696173895/job/18193851947?pr=2361), I found that the `create_event()` function works only for the main device.

https://github.com/thin-edge/thin-edge.io/blob/928f43e08e95f1bff9a97acaf677cb3d87773f21/crates/extensions/c8y_http_proxy/src/actor.rs#L296-L299

This means, if the size of an event exceeds the MQTT size threshold, creating an event via HTTP for child devices doesn't work. It will be created for the main device instead.

https://github.com/thin-edge/thin-edge.io/blob/928f43e08e95f1bff9a97acaf677cb3d87773f21/crates/extensions/c8y_mapper_ext/src/converter.rs#L451

P.S. we have a [system test for such a large event](https://github.com/thin-edge/thin-edge.io/blob/928f43e08e95f1bff9a97acaf677cb3d87773f21/tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot#L45), but two issues.
1. It was not created for child device.
2. The test didn't test to publish via HTTP.

Here is the output of `tedge mqtt sub "#"` of the test. You can see the event is created via MQTT.

```text
[te/device/main///e/largeEvent] {"text":"Large event","large_text_field":"x"}
[c8y/event/events/create] {"type":"largeEvent","time":"2023-10-31T17:16:15.227539999Z","text":"Large event","large_text_field":"x"}
```

**To Reproduce**
Publish an event for a child device with a large payload (> 16KiB), then the event will be created for its main device.

**Expected behavior**
If an event is supposed for a child device, the event must be created under the child device in c8y.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [0.12.0]

**Additional context**
The root cause is, c8y requires **different** payload to create an event, via REST or MQTT.

#### REST (HTTP)
One must provide the `source` field with `id`. The `id` must be c8y's **internal** device ID. If the event is supposed to belong to a child device, the `id` must be the internal ID of the child device. `externalSource` will be ignored. 

```json
{
    "source": { "id":"4431203" },
    "type": "TestEvent",
    "text": "sensor was triggered",
[...]

PR #2421, opened 2023-11-06
Fix send_event() in c8y_http_proxy to create events under child devices

rina23q:
## Proposed changes
c8y-mapper should create an event for a child device via HTTP when the event size exceeds the MQTT threshold. However, it was created for the main device instead. This PR fixes the issue.

The root cause how to get the `device_id` (= c8y's **internal** device ID) in this part of the code.
https://github.com/thin-edge/thin-edge.io/blob/9dea122c94be8c50ea81f122a3d11548353703ac/crates/extensions/c8y_http_proxy/src/actor.rs#L310-L314

`c8y_event.source` needs to hold an internal ID, however, it's too early to expect to have an internal ID as `source`. If we have to use the current implementation, one has to get the device's internal ID in the mapper's converter level (= the `c8y_event` object in the the below code has to have its internal ID as `source`). Querying an internal ID should be definitely hidden inside c8y_http_proxy API.

https://github.com/thin-edge/thin-edge.io/blob/9dea122c94be8c50ea81f122a3d11548353703ac/crates/extensions/c8y_mapper_ext/src/converter.rs#L468

For this reason, I introduced a new struct `CreateEvent` to parse device's **external** ID to c8y_http_proxy API.

## Types of changes

## Paste Link to the issue
#2394

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -41,6 +41,7 @@ use c8y_api::smartrest::topic::C8yTopic;
 use c8y_api::smartrest::topic::SMARTREST_PUBLISH_TOPIC;
 use c8y_auth_proxy::url::ProxyUrlGenerator;
 use c8y_http_proxy::handle::C8YHttpProxy;
+use c8y_http_proxy::messages::CreateEvent;
 use logged_command::LoggedCommand;
 use plugin_sm::operation_logs::OperationLogs;
 use plugin_sm::operation_logs::OperationLogsError;
@@ -465,7 +466,14 @@ impl CumulocityConverter {
                 messages.push(message);
             } else {
                 // The message must be sent over HTTP
-                let _ = self.http_proxy.send_event(c8y_event).await?;
+                let create_event = CreateEvent {
+                    event_type: c8y_event.event_type,
+                    time: c8y_event.time,
+                    text: c8y_event.text,
+                    extras: c8y_event.extras,
+                    device_id: entity.external_id.clone().into(),
+                };
+                self.http_proxy.send_event(create_event).await?;
                 return Ok(vec![]);
             }
         };
@@ -2271,7 +2279,7 @@ pub(crate) mod tests {
         let tmp_dir = TempTedgeDir::new();
         let (mut converter, mut http_proxy) = create_c8y_converter(&tmp_dir).await;
         tokio::spawn(async move {
-            if let Some(C8YRestRequest::C8yCreateEvent(_)) = http_proxy.recv().await {
+            if let Some(C8YRestRequest::CreateEvent(_)) = http_proxy.recv().await {
                 let _ = http_proxy
                     .send(Ok(c8y_http_proxy::messages::C8YRestResponse::EventId(
                         "event-id".into(),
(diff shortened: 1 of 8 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -465,7 +466,14 @@ impl CumulocityConverter {
                 messages.push(message);
             } else {
                 // The message must be sent over HTTP
-                let _ = self.http_proxy.send_event(c8y_event).await?;
+                let create_event = CreateEvent {
+                    event_type: c8y_event.event_type,
```
albinsuresh, line 470: It looks a bit weird that we first create a `C8yCreateEvent` and then create this `CreateEvent` out of it. I understand the need to create a different representations with `externalSource` for JSON over MQTT and `source` for HTTP. We should only be needing just one struct where that source field can be an enum of either `externalSource` or `source`. If this isn't trivial, you may address that refactoring in a different PR.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
```diff
@@ -48,6 +48,13 @@ Child devices support sending custom events overriding the type
     Log    ${events}
 
 
+Child devices support sending large events
+    Execute Command    tedge mqtt pub te/device/${CHILD_SN}///e/largeEvent "$(printf '{"text":"Large event","large_text_field":"%s"}' "$(yes "x" | head -n 100000 | tr -d '\n')")"
+    ${events}=    Device Should Have Event/s    expected_text=Large event    with_attachment=False    minimum=1    maximum=1    type=largeEvent    fragment=large_text_field
+    Length Should Be    ${events[0]["large_text_field"]}    100000
```
albinsuresh, line 54: Wouldn't it be better to assert the MQTT messages also, to assert that that the converted message is not present?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2288,6 +2296,33 @@ pub(crate) mod tests {
...
+                    .unwrap()
+            }
+        });
+
+        let event_topic = "te/device/child1///e/click_event";
+        let big_event_text = create_packet((16 + 1) * 1024); // Event payload > size_threshold
+        let big_event_payload = json!({ "text": big_event_text }).to_string();
+        let big_event_message = Message::new(&Topic::new_unchecked(event_topic), big_event_payload);
+
+        let mqtt_messages = converter.convert(&big_event_message).await;
```
albinsuresh, line 2319: That looks weird. When the event is big, we shouldn't be getting any MQTT messages, right? It should have been an HTTP message. What am I missing here?

rina23q: Here child device related messages will be received. That's why it checks the outgoing MQTT topic in assert.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
rina23q: Since the robot test should not be affected by c8y's requirement, this test might publish via MQTT. No MQTT message is covered in the Rust test.

reubenmiller: Yes that was my call. The system test does not care that http was used, it more cares that a large message can be sent to c8y from the device.

crates/extensions/c8y_mapper_ext/src/converter.rs
rina23q: I will not address the suggestion in this PR, but created an issue https://github.com/thin-edge/thin-edge.io/issues/2429

albinsuresh (approved):
LGTM

Issue #2409, opened 2023-11-03
registering commands on nested child devices results in duplicate devices when the tedge-mapper-c8y is restarted


reubenmiller:
**Describe the bug**

When known command is registered on a nested child device (e.g. a child of a child) via MQTT, then the definition of the child device changes to an immediate child device, when the tedge-mapper-c8y service is restarted.

**To Reproduce**

1. Register an immediate child device

    ```
    tedge mqtt pub -r te/device/child-level1// '{"@type":"child-device","name":"child-level1"}'
    ```

2. Register a nested child device (from the previous child device)

    ```
    tedge mqtt pub -r te/device/child-level2// '{"@type":"child-device","name":"child-level2","@parent":"device/child-level1//"}'
    ```

3. Register an operation on the nested child device

    ```
    tedge mqtt pub -r te/device/child-level2///cmd/restart '{}'
    ```

4. Verify in Cumulocity the following device hierarchy

    ```mermaid
    graph TD
        main --> child-level1 --> child-level2
    ```

5. Restart the tedge-mapper-c8y

    ```
    sudo systemctl restart tedge-mapper-c8y
    ```

6. Check the thin-edge.io registration message on the `te/+/+/+/+` topic.

    ```
    tedge mqtt sub te/device/child-level2/+/+
    ```

    Notice that the '@parent' property is no longer part of the message for the nested child device.

    ```
    [te/device/child-level2//] {"@id":"ginger:device:child-level2","@type":"child-device","name":"child-level2"}
    ```

**Expected behavior**

The tedge-mapper-c8y should use the registration information on the MQTT broker over any information in the `/etc/tedge/operations/c8y` folder when building the entity store. This will ensure that it does not override existing registration messages with incorrect information.

**Screenshots**

**Environment (please complete the following information):**
 
|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`Raspberry Pi 3 Model B Rev 1.2`|
|System-Architecture|`Linux ginger 6.1.0-rpi4-rpi-v8 #1 SMP PREEMPT Debian 1:6.1.54-1+rpt2 (2023-10-05) aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 0.13.1~12+g5dcb430`|

**Additional context**

The root cause looks to be the mapper's interpretation of the operation files inside the `/etc/tedge/operations/c8y` folder which don't supported nested files.

[...]

PR #2424, opened 2023-11-06
Disable nested child device operation support #2409

albinsuresh:
## Proposed changes

Disable operation support for:
* nested child devices
* services on any device

This is just a temporary fix for #2409 to avoid nested child devices and services from messing up the mapper's `operations` directory, which only supports immediate child devices. A proper fix will be coming in a later PR.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2409

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/config_operations.rs
+++ b/crates/extensions/c8y_mapper_ext/src/config_operations.rs
@@ -425,13 +425,11 @@ impl CumulocityConverter {
                         self.ops_dir.join(target.external_id.as_ref())
                     }
                     _ => {
-                        warn!("config_snapshot and config_update features for nested child devices are currently unsupported");
                         return Ok(vec![]);
                     }
                 }
             }
             EntityType::Service => {
-                warn!("config_snapshot and config_update features for services are currently unsupported");
                 return Ok(vec![]);
             }
         };
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -8,6 +8,7 @@ use crate::actor::IdDownloadRequest;
 use crate::dynamic_discovery::DiscoverOp;
 use crate::error::ConversionError;
 use crate::json;
+use anyhow::anyhow;
 use anyhow::Context;
 use c8y_api::http_proxy::C8yEndPoint;
 use c8y_api::json_c8y::C8yCreateEvent;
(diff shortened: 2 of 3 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1000,6 +1003,29 @@ impl CumulocityConverter {
...
+        op_type: &OperationType,
+        topic_id: &EntityTopicId,
+    ) -> Result<(), ConversionError> {
+        let target = self.entity_store.try_get(topic_id)?;
+
+        match target.r#type {
+            EntityType::MainDevice => Ok(()),
+            EntityType::ChildDevice => match &target.parent {
+                Some(parent) if !parent.is_default_main_device() => {
+                    Err(ConversionError::UnexpectedError(anyhow!(
```
rina23q, line 1017: It's really okayish to say `UnexpectedError` only because it's for this kind of quick temporary fix. However, this error is not a true unexpected error. Creating an operation for nested child devices is defenitetly a standard usecase. Just thin-edge.io is not ready for that yet. For me, it should be UnsupportedFeatureError or whatever.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2766,6 +2792,128 @@ pub(crate) mod tests {
...
+    #[test_case("config_snapshot")]
+    #[test_case("config_update")]
+    #[test_case("custom_op")]
+    #[tokio::test]
+    async fn operations_not_supported_for_services(op_type: &str) {
+        let tmp_dir = TempTedgeDir::new();
+        let (mut converter, _http_proxy) = create_c8y_converter(&tmp_dir).await;
+
+        // Register main device service
+        let _ = converter
```
rina23q, line 2857: I prefer `unwrap()` here instead of using `let _` because unwrap can cause a panic and test can fail here.

rina23q (commented):
The code looks okayish. I would like to decide if I approve or not after @albinsuresh replies to my comments.

crates/extensions/c8y_mapper_ext/src/config_operations.rs
```diff
@@ -425,13 +425,11 @@ impl CumulocityConverter {
                         self.ops_dir.join(target.external_id.as_ref())
                     }
                     _ => {
-                        warn!("config_snapshot and config_update features for nested child devices are currently unsupported");
```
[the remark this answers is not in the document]
albinsuresh: They've been moved to that common logic which is validating all the operations at one place. The `wrap_errors` function that picks up the returned error will be logging it as an `error!("Mapping error")`. So, this would have been redundant. And when this unsupported error path is hit, it wouldn't have reached here anyway, as the new validation  function would return the error, before reaching here.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Yeah, I didn't introduce a new variant for that same reason (add something new that will be removed very soon).

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: I can't `unwrap` here because the `convert` function does not return a Result. It's the `try_convert` function that returns a result.

crates/extensions/c8y_mapper_ext/src/config_operations.rs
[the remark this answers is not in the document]
rina23q: I guessed it's no longer hittable, but then why do we still need to keep the `match` arms?

```rust
match &target.parent {
        Some(parent) if parent.is_default_main_device() => {
            // Support only first level child devices due to the limitation of our file system supported operations scheme.
            self.ops_dir.join(target.external_id.as_ref())
        }
        _ => {
...
```

crates/extensions/c8y_mapper_ext/src/converter.rs
rina23q: Ah okay, I misunderstood what is returned here. So it just ignore the return of the converter.

rina23q (approved):
As a temporary fix, I would approve. I assume that some minor points will be refactored in the following PR.

crates/extensions/c8y_mapper_ext/src/config_operations.rs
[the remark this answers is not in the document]
albinsuresh: To build the op directory path in the normal case for immediate child devices and main device. For the nested child devices and service, it's a no-op, but those 2 specific paths won't be hit. Since match statements need to be exhaustive, I'm forced to keep those match arms with no-ops, even though they'll never be hit.

Issue #2350, opened 2023-10-18
Flaky test: c8y_mapper_ext  custom_operation_timeout_sigkill


didier-wenzek:
The test [` custom_operation_timeout_sigkill`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/tests.rs#L1642) is flaky (see [here](https://github.com/thin-edge/thin-edge.io/actions/runs/6558809614/job/17813142801) where the [second attempt](https://github.com/thin-edge/thin-edge.io/actions/runs/6558809614) is green).

This is not a timing issue, which is concerning.
- Would the check be too impatient (i.e. expecting an outcome before the mapper triggers a sigkill), the error would be `panicked at 'Channel closed while expecting: ...`
- Actually, the test is failing with an assertion failure: i.e. another unexpected message has been received.

didier-wenzek:
Fixed by https://github.com/thin-edge/thin-edge.io/commit/6d6c38dbf891de2c830d1c772b5cd3d26e21561d

PR #2432, opened 2023-11-08
Flaky test: c8y_mapper_ext custom_operation_timeout_sigkill

PradeepKiruvale:
## Proposed changes

There were two issues:
1. Related to handling the signal:
When the operation times out and a sigterm is sent to the `custom operation` (c8y_command) to terminate gracefully, the
command takes some time to process the signal and to call the signal handler. This will make the expected output different and because of this assertion fails.
proposed solution:
Solved asserting part of the expected messages rather than the whole message at once, just to make sure that the signal is handled properly and the operation is terminated.

3.  MQTT Channel getting closed:
In some cases, the MQTT channel gets closed even before the MQTT messages are received. 
So, increased the channel timeout.

Ran the tests for around 100+ iterations, and did not fail. Hope this fixes the issue.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2350

## Checklist


crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2104,25 +2102,14 @@ async fn custom_operation_timeout_sigkill() {
...
-
-stdout <<EOF
-main 1
-sigterm 1
-sigterm 2
-EOF
-
-stderr <<EOF
-operation failed due to timeout: duration=1sEOF
-";
```
didier-wenzek, line 2124: Why this has been removed and replaced by a weaker assertion?

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2074,8 +2073,7 @@ async fn custom_operation_timeout_sigkill() {
     }
     trap handle_term SIGTERM
     for i in {1..50}
-    do
-        echo "main $i"
```
didier-wenzek, line 2078: Why this has been removed?

The point was to check that operation script output is actually captured.

reubenmiller:
@PradeepKiruvale Can you please rebase this PR to the current on main (as that is the source of the system test failures)

crates/extensions/c8y_mapper_ext/src/tests.rs
PradeepKiruvale: In some cases, this log message is printed once, and sometimes it is printed multiple times before the signal is handled and the control is passed to the handler. So, because of this situation, it's hard to assert the whole output, because of the variable number of output logs from the main thread. This is the reason, I removed this log and just slept till the signal arrived and is being processed.

crates/extensions/c8y_mapper_ext/src/tests.rs
PradeepKiruvale: As, mentioned in the other comment, when the `sigterm` is sent for gracefully complete the task, this might be handled and the handler will start printing the `sigterm x` logs. Because still not gracefully stopped, the operation will be killed forcefully by sending the `sigkill`, which will take some time to stop. In some cases, the handler logs 2 times and in some cases 3 times. This is the reason it will be hard to assert with the fixed set of log messages. So, I introduced a weaker assertion. I feel it is enough to assert whether the `sigterm` handled or not and also whether the `sigkill` stopped the operation with the status `unknown` or not.

didier-wenzek:
This has been addressed by https://github.com/thin-edge/thin-edge.io/pull/2479/commits/6d6c38dbf891de2c830d1c772b5cd3d26e21561d, as these tests were systematically failing after a slight change of the timing behavior of operation execution.

The root cause of  flakiness was the signal handlers used by these tests.

Issue #2414, opened 2023-11-05
sporadic failures of log file upload from the tedge-mapper-c8y due to 'No such file or directory'


reubenmiller:
**Describe the bug**

Sporadic failure of the log_upload command when running the tedge-log-plugin on a child device due to the following reason:

```
Upload failed with No such file or directory (os error 2)
```

System test case:

* [log_operation_child.robot](https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/log/log_operation_child.robot)

For a quick inspection it looks like the error is occurring after the local `log_upload` command is set to `successful`, and during the uploads of the binary to Cumulocity in the tedge-mapper-c8y service. Below shows the a reduced set of log entries just relating to the flow of the commands:

```log
Nov 05 14:11:41 9f51399c2809 mqtt-logger[1151]: {"timestamp":1699193501.744442838,"message":{"tst":"2023-11-05T14:11:41.744483Z+0000","topic":"te/device/TST_lateral_watery_planner///cmd/log_upload/c8y-mapper-2023-11-05T14:11:41.742575047Z","qos":0,"retain":0,"payloadlen":308,"payload":"{\"status\":\"init\",\"tedgeUrl\":\"http://172.21.0.6:8000/tedge/file-transfer/TST_insource_weary_fruit:device:TST_lateral_watery_planner/log_upload/example-c8y-mapper-2023-11-05T14:11:41.742575047Z\",\"type\":\"example\",\"dateFrom\":\"2023-11-04T14:11:42Z\",\"dateTo\":\"2023-11-05T14:12:42Z\",\"searchText\":\"first\",\"lines\":10}"},"payload_hex":"7b22737461747573223a22696e6974222c22746564676555726c223a22687474703a2f2f3137322e32312e302e363a383030302f74656467652f66696c652d7472616e736665722f5453545f696e736f757263655f77656172795f66727569743a6465766963653a5453545f6c61746572616c5f7761746572795f706c616e6e65722f6c6f675f75706c6f61642f6578616d706c652d6338792d6d61707065722d323032332d31312d30355431343a31313a34312e3734323537353034375a222c2274797065223a226578616d706c65222c226461746546726f6d223a22323032332d31312d30345431343a31313a34325a222c2264617465546f223a22323032332d31312d30355431343a31323a34325a222c2273656172636854657874223a226669727374222c226c696e6573223a31307d"}
[...]

PR #2435, opened 2023-11-09
tedge-log-manager: wait for upload before marking operation successful

Bravo555:
## Proposed changes

tedge-log-manager previously marked `log_upload` operation as successful without waiting for the actual HTTP file upload operation to finish successfully. This lead to race conditions where after marking the operation as successful the actual logfile may not yet be available, or to bugs where the operation is marked as successful but the upload could not be completed.

Now tedge-log-manager actor waits for the upload to be successful to mark the operation as successful.

## Types of changes

## Paste Link to the issue

#2414

## Checklist

## Further comments

I thought about adding a test, but I don't quite know how to go about it. The scope seems very limited and I don't know if it's even possible to write that test in such a way that it's not flaky. Suggestions welcome.

### DIFF
--- a/crates/extensions/tedge_log_manager/src/actor.rs
+++ b/crates/extensions/tedge_log_manager/src/actor.rs
@@ -137,32 +137,21 @@ impl LogManagerActor {
         topic: &Topic,
         mut request: LogUploadCmdPayload,
     ) -> Result<(), ChannelError> {
-        match self
-            .execute_logfile_request_operation(topic, &request)
-            .await
-        {
-            Ok(()) => {
-                request.successful();
-                self.publish_command_status(topic, &request).await?;
-                info!("Log request processed for log type: {}", request.log_type);
-                Ok(())
-            }
-            Err(error) => {
-                let error_message = format!("Handling of operation failed with {}", error);
-                request.failed(&error_message);
-                self.publish_command_status(topic, &request).await?;
-                error!("{}", error_message);
-                Ok(())
-            }
+        if let Err(error) = self.generate_and_upload_logfile(topic, &request).await {
+            let error_message = format!("Handling of operation failed with {}", error);
+            request.failed(&error_message);
+            self.publish_command_status(topic, &request).await?;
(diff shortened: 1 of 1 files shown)


crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -197,34 +193,37 @@ impl LogManagerActor {
...
-                    info!("Log request processed for log type: {}.", request.log_type);
-
-                    if let Err(err) = std::fs::remove_file(&response.file_path) {
-                        warn!(
-                            "Failed to remove temporary file {}: {}",
-                            response.file_path, err
-                        )
-                    }
+        let Some(mut request) = self.pending_operations.remove(topic) else {
+            return Ok(());
```
albinsuresh, line 197: Wouldn't it be better to log a warning here before the return, so that if at all this unexpected path is hit in future for some reason, we'd have some clues?

albinsuresh (approved):
LGTM

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -137,24 +137,20 @@ impl LogManagerActor {
...
+            error!("{}", error_message);
+            return Ok(());
         }
+
+        self.pending_operations.insert(topic.name.clone(), request);
+
+        Ok(())
     }
 
     /// executes the log file request
```
rina23q, line 156: The comment from L156-L160 is outdated. Could you also fix it? Actually this function does only creating an upload request and sending it to uploader.

rina23q (approved):

crates/extensions/tedge_log_manager/src/actor.rs
Bravo555: Fixed in 2024cd9. New function name is `generate_and_upload_logfile`.

reubenmiller:
Looks good and I'm happy to merge it without system tests as this error was hard to reproduce reliably as it it would rely on controlling the upload speed on localhost.

PR #2434, opened 2023-11-09
misleading log entry when state file does not exist in tedge-agent

PradeepKiruvale:
## Proposed changes

Use proper error messages when the tedge-agent fails to access the state file.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2413

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2437, opened 2023-11-09
Improve error messages for c8y-http-proxy and tedge-agent

Bravo555:
## Proposed changes

While working on #2391, I found details in some of the error messages lacking, so I changed them and included in a separate PR.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/restart_manager/actor.rs
+++ b/crates/core/tedge_agent/src/restart_manager/actor.rs
@@ -170,16 +170,11 @@ impl RestartManagerActor {
             }
             Err(err) => {
                 match err {
-                    StateError::FromIo(e) => match e.kind() {
-                        std::io::ErrorKind::NotFound => {
-                            info!("tedge-agent state file is not present")
-                        }
-                        std::io::ErrorKind::PermissionDenied => {
-                            error!("Failed to read tedge-agent state file due to: PermissionDenied")
-                        }
-                        e => error!("Failed to read tedge-agent state file due to: {e}"),
-                    },
-                    e => error!("Failed to read tedge-agent state file due to: {e}"),
+                    // file missing means we don't have to perform the operation, so just do nothing
+                    StateError::LoadingFromFileFailed { source, .. }
+                        if source.kind() == std::io::ErrorKind::NotFound => {}
+                    // if read failed for some other reason, we should probably log it
+                    _ => error!("{err}"),
                 }
                 None
             }
--- a/crates/core/tedge_agent/src/state_repository/error.rs
+++ b/crates/core/tedge_agent/src/state_repository/error.rs
@@ -1,14 +1,25 @@
+use camino::Utf8Path;
+
 #[derive(Debug, thiserror::Error)]
 #[allow(clippy::enum_variant_names)]
 pub enum StateError {
-    #[error(transparent)]
-    FromTOMLParse(#[from] toml::de::Error),
+    #[error("State file `{path}` contains invalid syntax: {source}")]
+    FromTOMLParse {
+        path: Box<Utf8Path>,
+        source: toml::de::Error,
+    },
 
     #[error(transparent)]
     FromInvalidTOML(#[from] toml::ser::Error),
 
     #[error(transparent)]
     FromIo(#[from] std::io::Error),
+
+    #[error("Could not read state from file `{path}`: {source}")]
+    LoadingFromFileFailed {
+        path: Box<Utf8Path>,
+        source: std::io::Error,
+    },
 
     #[error(transparent)]
     FromUtf8Error(#[from] std::string::FromUtf8Error),
--- a/crates/extensions/c8y_http_proxy/src/messages.rs
+++ b/crates/extensions/c8y_http_proxy/src/messages.rs
@@ -40,6 +40,12 @@ pub enum C8YRestError {
 
     #[error(transparent)]
     FromIOError(#[from] std::io::Error),
+
+    // `Display` impl of `C8yRestError` is used as part of an error message sent to the cloud in a smartrest message.
+    // Using `{anyhow::Error:?}` also prints the lower-level cause, so using it here will result in a more detailed
+    // error message being sent to the cloud
+    #[error("Unexpected error: {0:?}")]
+    Other(#[from] anyhow::Error),
 }
 
 pub type C8YRestResult = Result<C8YRestResponse, C8YRestError>;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -711,6 +711,7 @@ dependencies = [
 name = "c8y_http_proxy"
 version = "0.13.0"
 dependencies = [
+ "anyhow",
  "async-trait",
  "c8y_api",
  "download",
--- a/crates/core/tedge_agent/src/state_repository/state.rs
+++ b/crates/core/tedge_agent/src/state_repository/state.rs
@@ -27,10 +27,19 @@ impl StateRepository for AgentStateRepository {
     type Error = StateError;
 
     async fn load(&self) -> Result<State, StateError> {
-        match fs::read(&self.state_repo_path).await {
-            Ok(bytes) => Ok(toml::from_str::<State>(&String::from_utf8(bytes)?)?),
-            Err(err) => Err(StateError::FromIo(err)),
-        }
+        let text = fs::read_to_string(&self.state_repo_path)
+            .await
+            .map_err(|e| StateError::LoadingFromFileFailed {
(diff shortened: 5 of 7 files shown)


crates/core/tedge_agent/src/state_repository/error.rs
```diff
@@ -10,6 +12,12 @@ pub enum StateError {
     #[error(transparent)]
     FromIo(#[from] std::io::Error),
 
+    #[error("Could not read tedge-agent state from file {repo_path}: {source}")]
```
Bravo555, line 15: When starting a container with only `tedge` and `tedge-agent` installed, the following error message showed on startup:

`Fail to read tedge-agent state: Could not read tedge-agent state from file /etc/tedge/.agent/restart-current-operation: No such file or directory (os error 2)`

crates/extensions/c8y_http_proxy/src/messages.rs
```diff
@@ -40,6 +40,12 @@ pub enum C8YRestError {
 
     #[error(transparent)]
     FromIOError(#[from] std::io::Error),
+
+    // `Display` impl of `C8yRestError` is used as part of an error message sent to the cloud in a smartrest message.
+    // Using `{anyhow::Error:?}` also prints the lower-level cause, so using it here will result in a more detailed
+    // error message being sent to the cloud
+    #[error("Unexpected error: {0:?}")]
+    Unexpected(#[from] anyhow::Error),
```
jarhodes314, line 48: I don't love the name `Unexpected`. This implies the other possible error values in `C8YRestError` are "expected", which I think isn't remotely true for a lot of them. `FromAnyhow` seems like a better fit here.

I know this isn't related to the current changes you've made, but I'm not a fan of using the `Error` suffix in the variant names. This is partially due to the fact the enum name will (almost) always be present when referring to the variant name, and often it's possible to come up with a name that actually implies something has gone wrong (and thus describes itself better). The example I always think of is Java's `NullPointerException`, which IMO should be called `NullPointerDereference`, as that (a) explains what actually went wrong, and (b) like in Rust, it's completely obvious from context that you're dealing with an exception (in Java it's the presence of `throw` or `catch`, in Rust it's probably in a `Result` and probably in an enum called `SomethingError`).

didier-wenzek: I agree with James that the name `Unexpected` is inappropriate and telling the user that the error was unexpected is useless.

About the `Error` suffix on error cases, this also makes sense ... an error should name the failed action. To keep in mind for future error cases!

Bravo555: > I agree with James that the name `Unexpected` is inappropriate and telling the user that the error was unexpected is useless.

True, in this case almost all errors variants are unexpected, but some errors derived from the input, like validation failures, can be _expected_, and in this case _unexpected_ would be something that has no relation to the input or the data provided by the user. _Expected_ errors are never bugs, but unexpected errors can be bugs.

crates/core/tedge_agent/src/state_repository/error.rs
didier-wenzek: Beware that a related change has been merged recently: https://github.com/thin-edge/thin-edge.io/pull/2434

crates/extensions/c8y_http_proxy/src/messages.rs
didier-wenzek: > so would simply `Other` be okay?

I'm okay for that.

crates/core/tedge_agent/src/state_repository/error.rs
Bravo555: I think the problem with this message, before and after #2434, is that it doesn't say the path of the file to which it refers. In the normal case when the file doesn't exist maybe it doesn't matter, but for other cases, like for permissions denied, I think it should absolutely say what file could not be opened.

I changed the offending commit and these are the examples new error messages:

- when file doesn't exist:
  ```
  2023-11-09T19:10:10.934643478Z  INFO tedge_agent::restart_manager::actor: tedge-agent state file `/etc/tedge/.agent/restart-current-operation` is not present

  ```

- when file can't be read due to bad permissions
  ```
  2023-11-09T19:09:36.765052563Z ERROR tedge_agent::restart_manager::actor: Could not read state from file `/etc/tedge/.agent/restart-current-operation`: Permission denied (os error 13)
  ```

- when file can't be correctly parsed
  ```
  2023-11-09T19:07:49.7470124Z ERROR tedge_agent::restart_manager::actor: State file `/etc/tedge/.agent/restart-current-operation` contains invalid syntax: TOML parse error at line 1, column 6
    |
  1 | file content
    |      ^
  expected `.`, `=`
  ```

We'll need to make sure to attach enough context to errors in a more systematic way. When I'll have time after 1.0 I'll create a dedicated issue where I'll talk about sinks for our errors (MQTT `te/errors` topic, logs, smartrest messages), how to present them better with miette, and how to better organise them and attach necessary context.

didier-wenzek: > We'll need to make sure to attach enough context to errors in a more systematic way. When I'll have time after 1.0 I'll create a dedicated issue where I'll talk about sinks for our errors (MQTT te/errors topic, logs, smartrest messages), how to present them better with miette, and how to better organise them and attach necessary context.

Yes, please. Work in pair with @jarhodes314 who has also strong opinion and insight on that matter.
For 1.0, we can have local improvements like this one here and there. But as you said, we have to wait post 1.0 for broader improvements.

didier-wenzek (approved):
Approved

jarhodes314 (approved):

reubenmiller (approved):
Approved

crates/core/tedge_agent/src/state_repository/error.rs
```diff
@@ -1,15 +1,26 @@
+use camino::Utf8Path;
+
 #[derive(Debug, thiserror::Error)]
 #[allow(clippy::enum_variant_names)]
 pub enum StateError {
-    #[error(transparent)]
-    FromTOMLParse(#[from] toml::de::Error),
+    #[error("State file `{path}` contains invalid syntax: {source}")]
+    FromTOMLParse {
+        path: Box<Utf8Path>,
```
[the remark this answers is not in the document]
Bravo555: 8 bytes smaller than `Utf8PathBuf` due to no `capacity` field, little difference in practice

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -170,16 +170,13 @@ impl RestartManagerActor {
...
-                            error!("Failed to read tedge-agent state file due to: PermissionDenied")
-                        }
-                        e => error!("Failed to read tedge-agent state file due to: {e}"),
-                    },
-                    e => error!("Failed to read tedge-agent state file due to: {e}"),
+                    // file missing is not an error
+                    StateError::LoadingFromFileFailed { path, source }
+                        if source.kind() == std::io::ErrorKind::NotFound =>
+                    {
+                        info!("tedge-agent state file `{path}` is not present");
```
albinsuresh, line 177:
```suggestion
                        info!("No pending restart operations found");
```
From the user's perspective, that's what the missing file means. I'm even okay not to log anything for this case, as the `process_pending_sm_operation` doesn't log this case either. It doesn't log the other serious IO errors like `PermissionDenied` either, but those are handled differently.

albinsuresh (approved):
LGTM

crates/core/tedge_agent/src/restart_manager/actor.rs
Bravo555: > From the user's perspective, that's what the missing file means. I'm even okay not to log anything for this case, as the `process_pending_sm_operation` doesn't log this case either.

Okay, now I understand it a bit better, I agree it makes little sense to log it.

> It doesn't log the other serious IO errors like `PermissionDenied` either, but those are handled differently.

Can you expand on that? To me it looks like `process_pending_sm_operation` creates error responses when file could be read, but the state itself is inconsistent, so io-related errors are just discarded.

albinsuresh: > Can you expand on that? To me it looks like process_pending_sm_operation creates error responses when file could be read, but the state itself is inconsistent, so io-related errors are just discarded.

In the `process_pending_sm_operation` function, the `self.state_repository.load()` call is immediately followed by `state_repository.clear()`. So, even though the `load()` returns an IO error which is ignored, the very next `clear()` would also fail and that error would be returned. That's what I meant by "handled differently". Ideally, the error from `load` itself should have been returned.

Issue #2438, opened 2023-11-09
custom operation quoting causes operation results to be parsed incorrectly on the server


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The Cumulocity IoT smart rest message id `503` expects the "results" from the operation to be provide in the format of csv.

Currently the standard output of the custom operation script/binary is processed and converted to a single csv entry, e.g.

```
[c8y/s/us] 503,c8y_Command,"some random output"
```

However for the `c8y_RelayArray` operation (related to the [519](https://cumulocity.com/guides/reference/smartrest-two/#relayarray-519) message id), the double quotes surrounding the result causes the server to interpret the operation's results incorrectly, as the server is expecting the values to be unquoted.

The  following shows the current behaviour (as of 0.13.0) and the expected behavior.

**Current (erroneous) - as of 0.13.0**

```
[c8y/s/ds] 519,ginger,CLOSED,OPEN
[c8y/s/us] 501,c8y_RelayArray
[c8y/s/us] 503,c8y_RelayArray,"CLOSED,OPEN"
```

**Expected**
```
[c8y/s/ds] 519,ginger,CLOSED,OPEN
[c8y/s/us] 501,c8y_RelayArray
[c8y/s/us] 503,c8y_RelayArray,CLOSED,OPEN
```

**Describe the solution you'd like**

An new option that can be used in the custom operation definition file to declare what the format of the result is provided in by the custom operation handler (e.g. the script/binary being called).

The `result_format` (string value) declares one of the following options:
* text - the result should be treated as one value when creating the SmartREST message (this is currently the default functionality, and should remain unchanged)
* csv - the result should be treated as multiple values and no quoting should be applied around the result as this would change how the server would interpret the results.

For example the following is a custom operation definition:

**File: /etc/tedge/operations/c8y/c8y_RelayArray**

```toml
[exec]
topic = "c8y/s/ds"
on_message = "519"
result_format = "csv"    # <== New option to control whether the result is handled as a single or multiple values
command = "/usr/bin/run-control-gpio.sh"
```

**Describe alternatives you've considered**

Previous suggestion:

Use a `quote_result` property to control whether the result is quoted or not. However this approach is not as future proof/flexible as the `result_format` option.

**File: /etc/tedge/operations/c8y/c8y_RelayArray**

```toml
[exec]
topic = "c8y/s/ds"
on_message = "519"
quote_result = false    # <== New property which disables quoting, default = true
[...]

PR #2439, opened 2023-11-09
custom operation option to disable quoting of results to enable c8y_RelayArray compatibility

reubenmiller:
## Proposed changes

Support custom operation option to set the expected formatting of the results returned by the custom operation so that the result can be handled correctly when building the SmartRest message output send via the Cumuocity IoT [503](https://cumulocity.com/guides/reference/smartrest-two/#set-operation-to-successful-503) message.

Available options are:

```toml
# Option 1: as text (default)
result_format = "text"

# Option 2: as csv
result_format = "csv"
```

**Example: c8y_RelayArray operation**

```to l
[exec]
topic = "c8y/s/ds"
on_message = "519"
result_format = "csv"
command = "/etc/tedge/operations/set_relay.sh"
timeout = 10
```

Tasks
* [x] System test
* [x] Rust implementation

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2438

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/smartrest/operations.rs
+++ b/crates/core/c8y_api/src/smartrest/operations.rs
@@ -16,6 +16,13 @@ use std::time::Duration;
 const DEFAULT_GRACEFUL_TIMEOUT: Duration = Duration::from_secs(3600);
 const DEFAULT_FORCEFUL_TIMEOUT: Duration = Duration::from_secs(60);
 
+#[derive(Debug, Clone, Eq, PartialEq, Default)]
+pub enum ResultFormat {
+    #[default]
+    Text,
+    Csv,
+}
+
 /// Operations are derived by reading files subdirectories per cloud /etc/tedge/operations directory
 /// Each operation is a file name in one of the subdirectories
 /// The file name is the operation name
@@ -27,11 +34,26 @@ pub struct OnMessageExec {
     on_message: Option<String>,
     topic: Option<String>,
     user: Option<String>,
+    #[serde(default, deserialize_with = "to_result_format")]
+    result_format: ResultFormat,
     #[serde(rename = "timeout")]
     #[serde(default = "default_graceful_timeout", deserialize_with = "to_duration")]
     pub graceful_timeout: Duration,
     #[serde(default = "default_forceful_timeout", deserialize_with = "to_duration")]
     pub forceful_timeout: Duration,
+}
+
+fn to_result_format<'de, D>(deserializer: D) -> Result<ResultFormat, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let val: String = serde::Deserialize::deserialize(deserializer)?;
+
+    match val.as_str() {
+        "text" => Ok(ResultFormat::Text),
+        "csv" => Ok(ResultFormat::Csv),
+        _ => Err(serde::de::Error::unknown_variant(&val, &["text", "csv"])),
+    }
 }
 
 fn default_graceful_timeout() -> Duration {
@@ -75,6 +97,12 @@ impl Operation {
 
     pub fn topic(&self) -> Option<String> {
         self.exec().and_then(|exec| exec.topic.clone())
+    }
+
+    pub fn result_format(&self) -> ResultFormat {
+        self.exec()
+            .map(|exec| exec.result_format.clone())
+            .unwrap_or_default()
     }
 
     pub fn template(&self) -> Option<String> {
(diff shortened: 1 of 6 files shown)


albinsuresh (commented):
I've suggested an alternate fix as [a comment](https://github.com/thin-edge/thin-edge.io/issues/2438#issuecomment-1805209162) in the source issue.

reubenmiller:
Update: @albinsuresh will be taking over the implementation to change the implementation as discussed in https://github.com/thin-edge/thin-edge.io/issues/2438

reubenmiller (commented):
Approved (after collaboration with @albinsuresh)

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -16,6 +16,27 @@ use std::time::Duration;
...
+impl<'de> Deserialize<'de> for ResultFormat {
+    fn deserialize<D>(deserializer: D) -> Result<ResultFormat, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let val: String = serde::Deserialize::deserialize(deserializer)?;
+
+        match val.as_str() {
+            "text" => Ok(ResultFormat::Text),
+            "csv" => Ok(ResultFormat::CSV),
```
albinsuresh, line 34: Is there an easier way to achieve this deserialization without this `Deserialize` impl? Any tricks with field attributes possible?

Bravo555: You can use [`try_from` and `into` serde container attributes](https://serde.rs/container-attrs.html#try_from) to deserialize via string. It also makes the type itself serializable in other contexts.

```rust

#[derive(Debug, Clone, Eq, PartialEq, Default, Serialize, Deserialize)]
#[serde(try_from = "String", into = "String")]
pub enum ResultFormat {
    #[default]
    Text,
    CSV,
}

#[error("Unknown format {0:?} - expected \"text\" or \"csv\"")]
pub struct UnknownFormatError(String);

impl TryFrom<String> for ResultFormat {
    type Error = UnknownFormatError;

    fn try_from(value: String) -> Result<Self, Self::Error> {
        match value.as_str() {
            "text" => Ok(ResultFormat::Text),
            "csv" => Ok(ResultFormat::CSV),
            _ => Err(UnknownFormatError(value)),
        }
    }
}

impl From<ResultFormat> for String {
    fn from(value: ResultFormat) -> String {
        match value {
            ResultFormat::CSV => "csv".to_string(),
            ResultFormat::Text => "text".to_string(),
        }
    }
}
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -730,8 +733,14 @@ impl CumulocityConverter {
...
-                                let successful_str =
-                                    format!("503,{op_name},\"{sanitized_stdout}\"");
+                                let successful_str = match result_format {
+                                    ResultFormat::Text => {
+                                        format!("503,{op_name},\"{sanitized_stdout}\"")
+                                    }
+                                    ResultFormat::CSV => {
+                                        format!("503,{op_name},{sanitized_stdout}")
+                                    }
+                                };
```
Bravo555, line 743: Is `op_name` confirmed to not have commas?

When I talked with @rina23q, she recommended to use `sanitize_for_smartrest` function on single fields. For now it can stay that way because I think it was already broker, but we'll need to start removing this creating smartrest payload via ad-hoc `format!()` calls and use something that automatically sanitizes so we can't make more mistakes.

Bravo555 (approved):

crates/extensions/c8y_mapper_ext/src/converter.rs
reubenmiller: > Is op_name confirmed to not have commas?

Yes it is safe to say the op_name will never have commas in the name.

> When I talked with @rina23q, she recommended to use sanitize_for_smartrest function on single fields. For now it can stay that way because I think it was already broker, but we'll need to start removing this creating smartrest payload via ad-hoc format!() calls and use something that automatically sanitizes so we can't make more mistakes.

Providing smart rest specific sanitization is actually useful to some degree (e.g. escaping `\n` characters)...so it is true that some unexpected sanitization might happen, however in reality there are not many use-cases for this (apart from the c8y_RelayArray, where the values are either OPEN or CLOSED). So I would say we can skip worrying about unnecessary sanitization for now.

crates/core/c8y_api/src/smartrest/operations.rs
albinsuresh: Those `TryFrom` impls definietly look more natural, but is pretty much the same amount of code as the existing `deserialize_with` impl. So, I'll keep the current impl for now. I was looking for something even more concise that'll do some "magic mapping" from a string to the enum variant with the same text with a single attribute. Looks like there's no much magic available :)

Issue #2390, opened 2023-10-31
decouple tedge-mapper-c8y and tedge-agent when uploading/downloading files from cloud


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The current interaction between the `tedge-mapper-c8y` assumes it has access to the file system (file-transfer folder) managed by the http server provided by the `tedge-agent`.

This prevents running the tedge-agent and tedge-mapper-c8y on independent file systems (e.g. in containers). It also brings in coupling between the two services which is not strictly required.

The errors occurs when the tedge-mapper-c8y processes a successful operations (such as config_snapshot or log_upload) and the tedge-mapper-c8y tries to read the `.path` and upload it to Cumulocity IoT. The operations fail with an error reason similar to the following:

```
No such file or directory (os error 2)
```

Technically speaking the tedge-mapper-c8y could access the file via the `tedgeUrl` property in the operation instead of the `path` and it would decouple the filesystem between the two services.

**Describe the solution you'd like**

The tedge-mapper-c8y service should access files via the `tedgeUrl` and not the "assumed" path.

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
Update on progress of this ticket:

* [x] Support log_upload and config_snapshot. https://github.com/thin-edge/thin-edge.io/pull/2440
* [ ] Support config update

PR #2440, opened 2023-11-10
fix: use FTS for `log_upload` and `config_snapshot` operations

Bravo555:
## TODO

- [x] fix blocking using a downloader actor
- [ ] ~~add a test which confirms that download no longer blocks~~ will not be done because it's too complex for the current architecture @reubenmiller for questions

## Proposed changes

Makes tedge-mapper-c8y use the File Transfer Service (FTS) when processing `log_upload` or `config_snapshot` operations.

Previously tedge-mapper-c8y implicitly assumed that it's installed on the same device as the tedge-agent and tried to access the file-transfer directory directly. Now it correctly downloads the files in the operation's `tedgeUrl` parameter via HTTP.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2390

## Checklist

## Further comments

The 1st commit introduces the FTS download operation by using the `Downloader` directly, fixing the bug.

The 2nd commit in this PR removes the use of `Downloader` inside `log_upload` and `config_snapshot` handlers to use `DownloaderActor` instead.

The advantage of this is that downloading from the FTS no longer blocks and other such operations can be processed in the meantime.

I will argue, however, that this also had negative effects on code complexity and maintainability. Whereas previously the download was in a single place and one could trace the control flow easily, now the state with maintaining the state of the download is spread very thin across many different functions and variables.

The effect of the current architecture is that for every complex operation where multiple steps are required, for every transition `step1 -> step2` (e.g. before the download and after the download) one has to create:

- a piece of state where we await the response from the actor that performs the effect associated with the transition, e.g. `pending_fts_download_operations` for `DownloaderActor`
- a function `step1` which starts the transition to `step2` by sending a message `request`
- a function `step2` which gets the result `response`
- a function which listens to `response` from the actor, and inspects the response to dispatch it to correct function which performs the next step

I believe that Rust's async/await was created to solve, among others, exactly this problem, by transforming every async function into a state machine which automatically handles transistions between `.await` points which block.


crates/extensions/c8y_mapper_ext/src/config_operations.rs
```diff
@@ -151,44 +155,37 @@ impl CumulocityConverter {
                 let event_response_id = self.http_proxy.send_event(create_event).await?;
 
                 // Send a request to the Uploader to upload the file asynchronously.
```
rina23q, line 157: This comment is outdated, hence should be updated.

crates/extensions/c8y_mapper_ext/src/config_operations.rs
```diff
@@ -151,44 +155,37 @@ impl CumulocityConverter {
...
-                    &uploaded_file_path,
-                );
+                let config_filename =
+                    format!("{}-{}", response.config_type.replace('/', ":"), cmd_id);
 
-                self.uploader_sender
-                    .send((cmd_id.into(), upload_request))
-                    .await
-                    .map_err(CumulocityMapperError::ChannelError)?;
+                let uploaded_file_url = format!(
```
rina23q, line 161: Minor. At my first look, I got confused because of the mix usage of "upload" and "downlonad". I think changing the variable name to `tedge_file_url` is better to understand as the url is localhost url.

```suggestion
                let tedge_file_url = format!(
```

crates/extensions/c8y_mapper_ext/src/config_operations.rs
```diff
@@ -151,44 +155,37 @@ impl CumulocityConverter {
...
+                        download_type: FtsDownloadOperationType::ConfigDownload,
                         smartrest_topic,
                         clear_cmd_topic: message.topic.clone(),
-                        c8y_binary_url: binary_upload_event_url.to_string(),
-                        operation: CumulocitySupportedOperations::C8yUploadConfigFile,
+                        event_response_id,
                     },
                 );
 
+                let destination_path = self.config.data_dir.cache_dir().join(config_filename);
```
rina23q, line 178: Using `cache_dir` to store a temporary file! This needs to be discussed.

1. The `cache_dir` aka `/var/tedge/cache` is associated with tedge-agent's file transfer service. Here is tedge-mapper, may be different file system. 
2. The original purpose of `cache_dir` is to prevent multiple downloads for the same file from cloud. For example, firmware_update, you want to apply the same firmware to multiple child devices. The usage of `cache_dir` here doesn't suit the purpose.

For me, a file uploaded from a device per operation is just a temporary file. I think using `/tmp` is better, but I'm opened for more opinions.

crates/extensions/c8y_mapper_ext/src/log_upload.rs
```diff
@@ -131,41 +140,36 @@ impl CumulocityConverter {
...
-                    &uploaded_file_path,
-                )
-                .with_content_type(ContentType::TextPlain);
+                let log_filename = format!("{}-{}", response.log_type, cmd_id);
 
-                self.uploader_sender
-                    .send((cmd_id.into(), upload_request))
-                    .await
-                    .map_err(CumulocityMapperError::ChannelError)?;
+                let uploaded_file_url = format!(
```
rina23q, line 145: Minor. The same comment as config_snapshot.
```suggestion
                let tedge_file_url = format!(
```

rina23q (commented):
Overall looks good. The most important for me is where tedge-mapper-c8y should keep the files by FTS.

crates/extensions/c8y_mapper_ext/src/config_operations.rs
Bravo555: would `fts_file_url` work since we're donwloading from the FTS?

crates/extensions/c8y_mapper_ext/src/config_operations.rs
Bravo555: Agreed, this should definitely be a tmpdir, I got a little lazy and tried to find something similar enough in the `DataDir` object, but cache directory is not it. `TempTedgeDir` is close, but because it uses `TempDir`, it will get removed when it gets out of scope. I'll either use `tmp.path` directly and manually remove the file when finishing the operation handling, or add `TempTedgeDir` to the `pending_fts_download_operations` hashmap, so it gets removed automatically.

didier-wenzek:
> I will argue, however, that this also had negative effects on code complexity and maintainability. Whereas previously the download was in a single place and one could trace the control flow easily, now the state with maintaining the state of the download is spread very thin across many different functions and variables.

I will not argue the contrary. The c8y mapper was already complex and the current design adds complexity.
Not only there is some complexity at the functional level, but also this complexity is made worse - as you highlight well - by this idea to persist the current state of any operation that must be resumed later when a response is received.
To be concrete, for a log upload request, the mapper has to forward the request to the child device, to await its response, to download the child-device log excerpts from the file-transfer service then upload these log excerpts to c8y. Athe code level, this leads to 3 states that have to be materialized, stored, handled, cleaned ("request sent to the child", "fts download request sent" and "c8y upload request sent".

> I believe that Rust's async/await was created to solve, among others, exactly this problem, by transforming every async function into a state machine which automatically handles transistions between `.await` points which block.

This is correct up to the point one needs to share some mutable state. With `async`/`await`, one has a nice view of the sequence of steps (sent a request to a child device, await for its response, download the log excerpt from fts, upload this excerpt to c8y). And tasks can be spawned in the background to handle such sequences concurrently ... as long as there is no mutable shared state.

Do we have such mutable shared state here? Requests are sent over senders, which are `Clone` and hence not an issue.  However to receive responses such a task needs to own its receiver. Why not? No straightforward to quickly implement now given the current state of the c8y mapper, however I see no blocker.

Can code clarity be improved without a deep refactoring? I see some improvements, but I would be happy to merge this PR as is, to meet the 0.13.1 release, and address improvements later.
1. The code is currently scattered. The `pending_fts_download_operations` cache is defined and created in `converter.rs`, populated in `config_operations` and `log_upload.rs` and finally cleared in `actor.rs`.
2. Sending a request to the downloader (or the uploader) works in pair with caching the request. Receiving a response and resuming a pending download (upload) operation also works in pair. Can this be abstracted by some wrapper that cache and send requests as well as received and resume responses with their context?

crates/extensions/c8y_mapper_ext/src/config_operations.rs
rina23q: > would `fts_file_url` work since we're donwloading from the FTS?

Yes, but actually this url is exactly the same as `tedgeUrl` in a command payload.

Bravo555: > Yes, but actually this url is exactly the same as `tedgeUrl` in a command payload.

Ah, I see, it makes sense to call it `tedge_file_url` then.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -167,6 +168,31 @@ pub struct UploadOperationData {
...
+
+    pub smartrest_topic: Topic,
+    pub clear_cmd_topic: Topic,
+    pub event_response_id: String,
+}
+
+/// Used to denote as type of what operation was the file downloaded from the FTS.
+///
+/// Used to dispatch download result to the correct operation handler.
+pub enum FtsDownloadOperationType {
```
albinsuresh, line 191: You could just reuse `CumulocitySupportedOperations` instead of this new struct.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -167,6 +168,31 @@ pub struct UploadOperationData {
...
+///
+/// Functions which download files from the tedge File Transfer Service as part of handling
+/// operations (e.g. when performing `log_upload` or `config_snapshot`, the relevant file is
+/// uploaded into FTS) will use this type for communicating with the Downloader actor.
+pub struct FtsDownloadOperationData {
+    pub download_type: FtsDownloadOperationType,
+    pub url: String,
+
+    // used to automatically remove the temporary file after operation is finished
+    pub destination_dir: tempfile::TempDir,
```
albinsuresh, line 181: May be we could store the `DownloadRequest` itself here.

crates/extensions/c8y_mapper_ext/src/log_upload.rs
```diff
@@ -130,42 +140,41 @@ impl CumulocityConverter {
...
+                    FtsDownloadOperationData {
+                        download_type: FtsDownloadOperationType::LogDownload,
+                        url: tedge_file_url.clone(),
+                        destination_dir,
+
                         smartrest_topic,
                         clear_cmd_topic: message.topic.clone(),
-                        c8y_binary_url: binary_upload_event_url.to_string(),
-                        operation: CumulocitySupportedOperations::C8yLogFileRequest,
+                        event_response_id,
```
albinsuresh, line 165: There's no point creating/sending the event before this download. So, it's better to move that to `handle_fts_log_download` right before the `binary_upload_event_url` is created.

crates/extensions/c8y_mapper_ext/src/log_upload.rs
```diff
@@ -188,6 +197,61 @@ impl CumulocityConverter {
         Ok(messages)
     }
 
+    /// Resumes `log_upload` operation after required file was downloaded from
+    /// the File Transfer Service.
+    pub async fn handle_fts_log_download(
```
albinsuresh, line 202:
```suggestion
    pub async fn handle_fts_log_download_result(
```

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: By using a different enum, I wanted to underline that these are the only operations in which we need to download from the FTS. By having a new enum with only these 2 variants, I don't have to do a catch-all `_ => {}` in matches, so that when a new variant is possibly added, the compiler will automatically highlight all the matches where this variant is not handled.

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -71,6 +72,7 @@ impl C8yMapperConfig {
             device_type,
             service_type,
             ops_dir,
+            tmp_dir: std::env::temp_dir().into(),
```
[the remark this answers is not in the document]
Bravo555: fixed in e1ee595a17

crates/extensions/c8y_mapper_ext/src/log_upload.rs
Bravo555: fixed in 5e1124f0.

crates/extensions/c8y_mapper_ext/src/log_upload.rs
Bravo555: fixed in 5e1124f0

rina23q (approved):

albinsuresh (approved):
LGTM

Issue #2426, opened 2023-11-06
log entry erroneously set to ERROR when handling device/services registered with custom topic schema


reubenmiller:
**Describe the bug**

Registering an entity on a custom schema results in an ERROR log entry even though thin-edge.io does support this use-case.

```sh
Nov 06 20:55:27 ginger tedge-mapper[6266]: 2023-11-06T19:55:27.458503269Z ERROR c8y_mapper_ext::converter: Mapping error: The specified topic id system1/floor1/room1/ does not match the default topic scheme: 'device/<device-id>/service/<service-id>'
Nov 06 20:55:27 ginger tedge-mapper[6266]: 2023-11-06T19:55:27.458620667Z ERROR c8y_mapper_ext::converter: Mapping error: The specified topic id system1/floor1/room1/ does not match the default topic scheme: 'device/<device-id>/service/<service-id>'
Nov 06 20:55:27 ginger tedge-mapper[6266]: 2023-11-06T19:55:27.460218204Z ERROR c8y_mapper_ext::converter: Mapping error: The specified topic id system1/floor1/room1/ does not match the default topic scheme: 'device/<device-id>/service/<service-id>'
Nov 06 20:55:27 ginger tedge-mapper[6266]: 2023-11-06T19:55:27.460583525Z ERROR c8y_mapper_ext::converter: Mapping error: The specified topic id system1/floor1/room1/ does not match the default topic scheme: 'device/<device-id>/service/<service-id>'
Nov 06 20:55:27 ginger tedge-mapper[6266]: 2023-11-06T19:55:27.46073457Z ERROR c8y_mapper_ext::converter: Mapping error: The specified topic id system1/floor1/room1/ does not match the default topic scheme: 'device/<device-id>/service/<service-id>'
```

**To Reproduce**

It is unclear what the exact steps were to reproduce this. But tracing the code should be easy enough to see how to invoke the scenario.

**Expected behavior**

It seems more like a coding error as the `Error::NonDefaultTopicScheme` error should not bubble up to the point that it is logged to the console as an ERROR as it should only be logged on the `DEBUG` level at most.

The link below shows where the error is originating from:

[Err(Error::NonDefaultTopicScheme(entity_topic_id.clone()))](https://github.com/reubenmiller/thin-edge.io/blob/95d5dc36625d329c556e7c6a9368380c964403eb/crates/core/tedge_api/src/entity_store.rs#L486)

**Screenshots**

[...]

Issue #2428, opened 2023-11-07
Ordered processing of entity registration messages before any other messages


albinsuresh:
**Is your feature improvement request related to a problem? Please describe.**

Since MQTT protocol doesn't offer any guarantees about ordering of messages across different topics, even if a device sends a registration message to `te/device/<device-id>//` with an explicit `@id` value (the desired external ID) followed by a telemetry message to `te/device/<device-id>///m/temperature` topic, there is no guarantee that the mapper will receive and process the registration message first, before processing the telemetry message. If the telemetry message is processed first, the auto-registration mechanism would register that entity wrongly with the wrong external ID and other details(`name` and `type` if provided), and when the real device registration message is processed afterwards, it won't be able to undo/correct some of the damage that's already done like registering the entity in the cloud with the wrong auto-derived external ID, instead of the one provided with `@id`.

**Describe the solution you'd like**

1. A mechanism to turn-off auto-registration mechanism, when someone decides to use explicit registration. Having auto-registration turned ON while using explicit registration is forbidden as that'll lead to unnecessary race conditions.
2. When auto-registration is turned OFF,  introduce an ordering mechanism in the mapper that'll wait for an entity registration message, even if it receives the telemetry message first.

**Describe alternatives you've considered**

An alternative thought was to have separate message channels for "entity registration messages" and "other messages" and to always give higher priority to the entity registration message channel, while fetching the next available MQTT message, with the hope of always reading the registration message first, before the other message.

But, this isn't sufficient as the priority mechanism will work only when the both the registration message and the other message are delivered to the MQTT actor, by the broker, so that they're both available in their own respective message channels when we try to read the next message. But, if the broker delivers only the telemetry message and delays the delivery of the registration message, even this "priority read" mechanism will fail, as there is only one message in the "other message channel" and nothing in the entity registration message channel and hence the only available message will be read anyway.
 
If we want to keep both auto-registration and explicit registration turned ON together(which would be ideal, if possible), then we can fall back to auto-registration after waiting for a fixed amount of time for the registration message. But, this solution is highly sensitive to message delivery time and hence not very predictable/deterministic. So, we might still end up with undesired entity metadata whenever there's an unexpected delay in message delivery(when the system is under high load).

albinsuresh:
### Test plan

[...]

PR #2441, opened 2023-11-10
log entry erroneously set to ERROR when handling device/services registered with custom topic schema

PradeepKiruvale:
## Proposed changes

When using a topic that is not covered by auto registration, `topic schema` the `auto_register_entity` must not return an error but it must log the message at the `debug` level.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2426

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -586,7 +586,7 @@ pub enum Error {
     #[error("The specified entity {0} does not exist in the store")]
     UnknownEntity(String),
 
-    #[error("The specified topic id {0} does not match the default topic scheme: 'device/<device-id>/service/<service-id>'")]
+    #[error("Auto registration of the entity with topic id {0} failed as it does not match the default topic scheme: 'device/<device-id>/service/<service-id>'. Try explicit registration instead.")]
     NonDefaultTopicScheme(EntityTopicId),
 
     #[error(transparent)]
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -903,7 +903,20 @@ impl CumulocityConverter {
                 // if device is unregistered register using auto-registration
                 if self.entity_store.get(&source).is_none() {
                     let auto_registration_messages =
-                        self.entity_store.auto_register_entity(&source)?;
+                        match self.entity_store.auto_register_entity(&source) {
+                            Ok(auto_registration_messages) => auto_registration_messages,
+                            Err(e) => match e {
+                                Error::NonDefaultTopicScheme(eid) => {
+                                    debug!("{}", Error::NonDefaultTopicScheme(eid.clone()));
+                                    return Ok(vec![self.new_error_message(
+                                        ConversionError::FromEntityStoreError(
+                                            entity_store::Error::NonDefaultTopicScheme(eid),
+                                        ),
+                                    )]);
+                                }
+                                e => return Err(e.into()),
+                            },
+                        };
                     for auto_registration_message in &auto_registration_messages {
                         registration_messages.append(
                             &mut self.register_and_convert_entity(auto_registration_message)?,
@@ -2948,6 +2961,28 @@ pub(crate) mod tests {
                 )],
             );
         }
+    }
+
+    #[tokio::test]
+    async fn try_auto_registration_on_custom_topic() -> Result<(), anyhow::Error> {
+        use capture_logger::begin_capture;
+        use capture_logger::pop_captured;
+        let tmp_dir = TempTedgeDir::new();
+        let (mut converter, _http_proxy) = create_c8y_converter(&tmp_dir).await;
+
+        let alarm_topic = "te/custom/child2///m/";
+        let alarm_payload = json!({ "text": "" }).to_string();
+        let alarm_message = Message::new(&Topic::new_unchecked(alarm_topic), alarm_payload);
+        begin_capture();
+        let res = converter.try_convert(&alarm_message).await.unwrap();
+        let expected_err_msg = Message::new(&Topic::new_unchecked("te/errors"), "Auto registration of the entity with topic id custom/child2// failed as it does not match the default topic scheme: 'device/<device-id>/service/<service-id>'. Try explicit registration instead.");
+        assert_eq!(res[0], expected_err_msg);
+        let expected_log = "Auto registration of the entity with topic id custom/child2// failed as it does not match the default topic scheme: 'device/<device-id>/service/<service-id>'. Try explicit registration instead.";
+        // skip other log messages
+        pop_captured().unwrap().message();
+        pop_captured().unwrap().message();
+        assert_eq!(pop_captured().unwrap().message(), expected_log);
+        Ok(())
     }
 
     pub(crate) async fn create_c8y_converter(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -766,6 +766,7 @@ dependencies = [
  "c8y_auth_proxy",
  "c8y_http_proxy",
  "camino",
+ "capture-logger",
  "clock",
  "json-writer",
  "logged_command",
@@ -800,6 +801,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c59e92b5a388f549b863a7bea62612c09f24c8393560709a54558a9abdfb3b9c"
 dependencies = [
  "serde",
+]
+
+[[package]]
+name = "capture-logger"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "605a59c5054ce9e87cd329251bcc4c04579d5edbd206b1e4c45fbfe24b13ff82"
+dependencies = [
+ "log",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ certificate = { path = "crates/common/certificate" }
 clap = { version = "4.4", features = ["cargo", "derive"] }
 clock = { path = "crates/common/clock" }
 collectd_ext = { path = "crates/extensions/collectd_ext" }
+capture-logger = "0.1"
 csv = "1.1"
 darling = "0.20"
 doku = "0.21"
--- a/crates/extensions/c8y_mapper_ext/Cargo.toml
+++ b/crates/extensions/c8y_mapper_ext/Cargo.toml
@@ -47,9 +47,11 @@ tracing = { workspace = true }
 [dev-dependencies]
 assert-json-diff = { workspace = true }
 assert_matches = { workspace = true }
+capture-logger = { workspace = true }
 proptest = { workspace = true }
 rand = { workspace = true }
 tedge_actors = { workspace = true, features = ["test-helpers"] }
 tedge_mqtt_ext = { workspace = true, features = ["test-helpers"] }
 tedge_test_utils = { workspace = true }
 test-case = { workspace = true }
+


crates/core/tedge_api/src/entity_store.rs
```diff
@@ -483,7 +483,12 @@ impl EntityStore {
 
             Ok(register_messages)
         } else {
-            Err(Error::NonDefaultTopicScheme(entity_topic_id.clone()))
+            debug!(
+                "Failed to create auto register message due to: {}",
```
reubenmiller, line 487: We should log that something "failed" as we are just saying that the topic being used does not trigger an auto registration messages (which is expected).

Maybe something like this would be more accurate/helpful to the user:

```suggestion
                "Topic does not match auto registration scheme: {}",
```

reubenmiller: I'll leave the review of whether the error should be bubble up to the calling function up to the Rust reviewers.

albinsuresh: This won't be enough. The error still bubbles up all the way up to the `convert` function which will log that error and publish it to `tedge/errors` as well. It's better to catch this error from the caller: `try_convert_te_topics` and do the logging there and avoid bubbling it up any further.

albinsuresh (changes requested):

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -259,6 +263,7 @@ impl CumulocityConverter {
             downloader_sender,
             pending_operations: HashMap::new(),
             command_id,
+            custom_entities: VecDeque::with_capacity(CUSTOM_ENTITY_CACHE_SIZE),
```
albinsuresh, line 266:
```suggestion
            unregistered_entities: VecDeque::with_capacity(CUSTOM_ENTITY_CACHE_SIZE),
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -902,8 +907,39 @@ impl CumulocityConverter {
...
+                        Ok(auto_registration_messages) => auto_registration_messages,
+                        Err(e) => match e {
+                            entity_store::Error::NonDefaultTopicScheme(eid) => {
+                                // if its non supported enity, then cache it, log
+                                // and also send out error message on `errors` topic only at the first time.
+                                if !self.custom_entities.contains(&eid) {
+                                    if self.custom_entities.len() >= CUSTOM_ENTITY_CACHE_SIZE {
+                                        self.custom_entities.pop_front();
+                                    }
+                                    self.custom_entities.push_back(eid.clone());
```
albinsuresh, line 923: To avoid this code complexity at this level, I'd create a `RingBuffer` struct which wraps the `VecDeque` internally and move this logic into its `push` function.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -902,8 +907,39 @@ impl CumulocityConverter {
...
+                                    warn!(
+                                    "Auto registration failed for the custom entity due to: {}, try explicitly registering the entity manually",
```
albinsuresh, line 925: The error message of the `NonDefaultTopicScheme` error itself could have been updated with that message, to keep the log statement and the `te/error` messages consistent. The extra advantage is that we can even just `?` that error from here so that it's logged and sent to the `te/errors` topic automatically from the `convert` function higher up the stack.

The updated error message could be something like this:
```rust
    #[error("Auto registration of the entity with topic id {0} failed as it does not match the default topic scheme: 'device/<device-id>/service/<service-id>'. Try explicit registration instead.")]
    NonDefaultTopicScheme(EntityTopicId),
```
I'd drop the word `manually` as it carries connotations of a manual step to be performed.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -902,8 +907,39 @@ impl CumulocityConverter {
...
-                    let auto_registration_messages =
-                        self.entity_store.auto_register_entity(&source)?;
+                    let auto_registration_messages = match self
+                        .entity_store
+                        .auto_register_entity(&source)
+                    {
+                        Ok(auto_registration_messages) => auto_registration_messages,
+                        Err(e) => match e {
+                            entity_store::Error::NonDefaultTopicScheme(eid) => {
+                                // if its non supported enity, then cache it, log
```
albinsuresh, line 917:
```suggestion
                                // if this unknown entity was not previously seen, then cache it, and then log the error message.
                                // Subsequent errors for the same entity won't be logged repeatedly.
```

albinsuresh (commented):
Some more refactoring will simplify the code a lot more.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: The only side effect of using `?` for the same error is that it will be logged as an error instead of a warning. But I think that's fine since it is only logged once and it genuinely is a user error.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -902,8 +907,39 @@ impl CumulocityConverter {
...
+                                    "Auto registration failed for the custom entity due to: {}, try explicitly registering the entity manually",
+                                    entity_store::Error::NonDefaultTopicScheme(eid.clone())
+                                );
+                                    // send a message over the error topic as well
+                                    registration_messages.push(self.new_error_message(
+                                        ConversionError::FromEntityStoreError(
+                                            entity_store::Error::NonDefaultTopicScheme(eid),
+                                        ),
+                                    ));
+                                    return Ok(registration_messages);
```
didier-wenzek, line 934: This is confusing as the message sent is not a registration message.

```suggestion
                                    // send a message over the error topic
                                    return Ok(self.new_error_message(
                                        ConversionError::FromEntityStoreError(
                                            entity_store::Error::NonDefaultTopicScheme(eid),
                                        ),
                                    ));
```

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek, line 266: I see no value added by this cache of unknown entities and I would simply notify an error for each and every message received for an entity that is not already registered and cannot be automatically registered.

The code of the c8y converter is already is already too complex and doesn't need this extra complexity. Furthermore, this may confuse the user: sending a first message, noticing the error, retrying to understand what is going on, noticing no more errors, wrongly assuming everything is fine.

didier-wenzek (changes requested):
The logic to cache error messages has to be removed.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: I'll let @reubenmiller comment on whether he's okay to remove it. It definitely adds complexity for a minor enhancement, I agree, but it was done on an explicit request to avoid flooding the logs and the `te/error` topics when there are a lot of telemetry messages processed back-to-back with the same unregistered topic id. This will mostly happen during the startup of the mapper, where all persistent telemetry messages queued when the mapper was down, are delivered together.

reubenmiller: Yes we can remove the "unknown entity" store, and in place just log (somewhere) on the DEBUG level so that we don't cause any undue spamming on production systems, but still provide debugging ability in development environments.

albinsuresh: I'm still not fully convinced about logging such a critical user error at DEBUG level, as users typically enable DEBUG level to get more insights into what's going on, when they have already spotted something going wrong. Discovering an error itself from a DEBUG log is quite unusual and finding it from a sea of debug statements is going to be a near impossible task.

Even on the log spamming issue, I really don't see how this can happen in production (once #2428 is fixed, of course). Someone using a custom topic scheme without explicitly registering it sounds like a basic user mistake during the development phase, which would/should be caught before that deployment goes into production. During that development phase, logging user mistakes as ERRORs would be far more beneficial to the user than making it DEBUG. And spamming logs for basic mistakes doesn't sound like a big issue in a dev environment as the same spamming would happen with any other trivial mistakes that they make as well.

But, as you pointed out offline, since the error is also reported to `te/errors`, that log statement missing may not be such a big deal for "error discoverability". So, to avoid stretching this PR further on such a trivial matter, I'm okay with the proposed change.

crates/core/tedge_api/src/entity_store.rs
PradeepKiruvale: Now I have `short circuted` and returns an error message on the `te/errors` message. if the schema is not supported by auto-registration.

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: No more cache of the messages. So, this is not required anymore.

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: Not relevant anymore

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: Updated the `error` message.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -102,6 +104,7 @@ const CREATE_EVENT_SMARTREST_CODE: u16 = 400;
 const DEFAULT_EVENT_TYPE: &str = "ThinEdgeEvent";
 const FORBIDDEN_ID_CHARS: [char; 3] = ['/', '+', '#'];
 const REQUESTER_NAME: &str = "c8y-mapper";
+const CUSTOM_ENTITY_CACHE_SIZE: usize = 16;
```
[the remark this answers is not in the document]
PradeepKiruvale: Not relevant anymore

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: Fixed it

didier-wenzek (approved):
Approved

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -903,7 +903,20 @@ impl CumulocityConverter {
...
+                                Error::NonDefaultTopicScheme(eid) => {
+                                    debug!("{}", Error::NonDefaultTopicScheme(eid.clone()));
+                                    return Ok(vec![self.new_error_message(
+                                        ConversionError::FromEntityStoreError(
+                                            entity_store::Error::NonDefaultTopicScheme(eid),
+                                        ),
+                                    )]);
+                                }
+                                e => return Err(ConversionError::FromEntityStoreError(e)),
+                            },
```
albinsuresh, line 918:
```suggestion
                            Err(e) => match e {
                                Error::NonDefaultTopicScheme(eid) => {
                                    debug!("{}", Error::NonDefaultTopicScheme(eid.clone()));
                                    return Ok(vec![self.new_error_message(
                                        Error::NonDefaultTopicScheme(eid).into(),
                                    )]);
                                }
                                e => return Err(e.into()),
                            },
```

albinsuresh (approved):
LGTM

crates/extensions/c8y_mapper_ext/src/converter.rs
PradeepKiruvale: fixed

Issue #2445, opened 2023-11-13
use independent tedge.toml settings to access the c8y proxy server provided by tedge-mapper-c8y


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The `c8y.proxy.bind.address` setting is used by two different components but for different purposes,

1. tedge-mapper-c8y interprets this setting as the bind address where the c8y proxy server should be made available
2. tedge-mapper-c8y interprets this setting as the address when providing a url for child devices/components to reach the c8y proxy server from (for example, the c8y proxy endpoint is used in Cumulocity IoT software update commands)

The different in the interpretation of the two settings makes it difficult to configure the tedge-mapper-c8y address, as it needs to know the ip address of the tedge-mapper-c8y, and it can't use a hostname (as the setting only accepts an ip address, in the form of `x.x.x.x`). When using thin-edge.io in containers generally the container does not know the ip address of each other upfront, but rather at runtime, that is where using a hostname would be more useful when telling the tedge-mapper-c8y when creating the url which child devices can access the c8y proxy server from. 

This problem is not too dissimilar to the changes which prompted splitting the `mqtt.bind.address` and `mqtt.client.host` settings.

**Describe the solution you'd like**

Create the following new settings which are used by the tedge-mapper-c8y:

* `c8y.proxy.client.host` - ip or hostname of the http server (c8y proxy service) provided by the tedge-mapper-c8y
* `c8y.proxy.client.port` - port of the http server (c8y proxy service) provided by the tedge-mapper-c8y

**Describe alternatives you've considered**

**Additional context**

Bravo555:
# QA reproduction instructions

Following things need to be checked:

- new config options are present
- `c8y.proxy.client.host` can use a hostname
- when C8y proxy server and other components that talk to it run on separate devices (or containers), setting newly added options allow the component running in the container `B` to reach C8y container running on container `A`.

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/tedge/call_tedge_config_list.robot
- [x] QA has tested the function and it's functioning according description.

PR #2446, opened 2023-11-13
fix(#2445): add c8y proxy config settings

Bravo555:
## TODO

- [x] documentation

## Proposed changes

This PR adds `c8y.proxy.client.host` and `c8y.proxy.client.port` used to connect to running C8y HTTP proxy.

## Types of changes

## Paste Link to the issue

- #2445

## Checklist

### DIFF
--- a/docs/src/references/tedge-cumulocity-proxy.md
+++ b/docs/src/references/tedge-cumulocity-proxy.md
@@ -11,9 +11,11 @@ thin-edge device.
 It automatically handles authenticating with a JWT, avoiding the need for clients to support MQTT to retrieve this
 information.
 
-The API can be accessed at `http://{ip}:8001/c8y/{c8y-endpoint}`. `ip` is configured using the `c8y.proxy.bind.address`
-configuration
-(the default value is `127.0.0.1`) and the port can be changed by setting `c8y.proxy.bind.port`.
+The Cumulocity HTTP API can be accessed at `http://{host}:{port}/c8y/{c8y-endpoint}`. Configuration settings
+`c8y.proxy.client.host` and `c8y.proxy.client.port` are used to configure `{host}` and `{port}` parts of the base URL
+which will be used by thin-edge components to make requests to the C8y Proxy. `c8y.proxy.bind.address` and
+`c8y.proxy.bind.port` are read by `tedge-mapper-c8y` and used as bind address and port for the Cumulocity HTTP proxy. In
+both `client` and `bind`, `127.0.0.1` and `8001` are used as defaults for the address and port, respectively.
 
 For example, you can access the current tenant information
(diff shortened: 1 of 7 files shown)


docs/src/references/tedge-cumulocity-proxy.md
```diff
@@ -11,9 +11,11 @@ thin-edge device.
 It automatically handles authenticating with a JWT, avoiding the need for clients to support MQTT to retrieve this
 information.
 
-The API can be accessed at `http://{ip}:8001/c8y/{c8y-endpoint}`. `ip` is configured using the `c8y.proxy.bind.address`
-configuration
-(the default value is `127.0.0.1`) and the port can be changed by setting `c8y.proxy.bind.port`.
+The Cumulocity HTTP API can be accessed at `http://{host}:{port}/c8y/{c8y-endpoint}`. Configuration settings
+`c8y.proxy.client.host` and `c8y.proxy.client.port` are used to configure `{host}` and `{port}` parts of the base URL
+which will be used by thin-edge components to make requests to the C8y Proxy. `c8y.proxy.bind.address` and
```
jarhodes314, line 16: If my understanding is correct, this should stress that the client settings configure the mapper and not the agent (which is what I, as a user, would assume the client setting changes as it is how the equivalent MQTT settings work). I guess the main clue as to this being the case without knowing the implementation details is that the `c8y` settings only affect the mapper as that's the only component that cares about Cumulocity.

reubenmiller: It is defining how clients can reach the endpoint.

The mapper is the only component that uses the setting directly, however the value of the setting will be used within command payloads which are consumed by other clients (e.g. custom device connectors)

jarhodes314: I get that, my point was "there isn't any point setting this on a child device as the value won't be used", which is unlike what happens for the mqtt client settings. It probably doesn't have much of an impact, but the logic behind why it's like that doesn't seem like a thing that's necessarily obvious to a user so might be worth drawing attention to.

didier-wenzek: I agree with @jarhodes314. We must be explicit telling that both `c8y.proxy.bind.address` and `c8y.proxy.client.host` have to be set on the box running the `c8y` mapper. The latter will be forwarded by the c8y mapper to its clients. `c8y.proxy.client.host` can also be directly set on the clients which aim to use the c8y REST API.

Bravo555: My bad, forgot about this ticket. Description tweaked in 0159f91, is it OK now?

didier-wenzek (approved):
Approved

jarhodes314 (approved):

Issue #2451, opened 2023-11-14
Updating `/etc/tedge/plugins/tedge-log-plugin.toml` results in multiple `log_upload` metadata messages being sent


Bravo555:
**Describe the bug**

When saving a modification to `/etc/tedge/plugins/tedge-log-plugin.toml`, multiple `log_upload` messages are created, resulting in multiple `118` Smartrest messages being sent:

```
(`tedge-log-plugin.toml` saved)

[te/device/main///cmd/log_upload] {"types":[]}
[te/device/main///cmd/log_upload] {"types":[]}
[c8y/s/us] 118,
[c8y/s/us] 118,
[te/device/main///cmd/log_upload] {"types":["mosquitto","software-management"]}
[te/device/main///cmd/log_upload] {"types":["mosquitto","software-management"]}
[te/device/main///cmd/log_upload] {"types":["mosquitto","software-management"]}
[te/device/main///cmd/log_upload] {"types":["mosquitto","software-management"]}
[c8y/s/us] 118,mosquitto,software-management
[c8y/s/us] 118,mosquitto,software-management
[c8y/s/us] 118,mosquitto,software-management
[c8y/s/us] 118,mosquitto,software-management
```

Messages with empty `"types"` parameter can appear when saving the file with `vim`. Here's the relevant log output:

```
2023-11-14T16:57:49.406620439Z DEBUG Log Manager: recv Some(FsWatchEvent(Modified("/etc/tedge/plugins/tedge-log-plugin.toml")))    
2023-11-14T16:57:49.40664234Z  INFO log_manager::config: Using the configuration from /etc/tedge/plugins/tedge-log-plugin.toml    
2023-11-14T16:57:49.40667632Z  WARN log_manager::config: The config file /etc/tedge/plugins/tedge-log-plugin.toml does not exist or is not readable.    
```

When this file can't be read, an empty types list is used. It would be better not to send the message in this case.

Additionally, to not produce duplicate messages, a debouncer can be used, as described in [`notify` crate documentation](https://docs.rs/notify/latest/notify/#installation). 

**To Reproduce**

1. Open `/etc/tedge/plugins/tedge-log-plugin.toml` with vim.
2. Modify the file (optional, in my case duplicate messages are produced even if no modifications were introduced)
3. Save the file
4. Duplicate messages are sent to `te/device/main///cmd/log_upload`

**Expected behavior**

At most 1 `log_upload` metadata message should be sent.

[...]

PR #2454, opened 2023-11-15
Stop duplicate messages being sent by `tedge-log-plugin` and `tedge-configuration-plugin` on file reload

Bravo555:
## Proposed changes

This fix consists of 2 parts:

1. `NotifyStream` impl was changed to make use of a `notify-debouncer-full` crate to delay and debounce fs events. This should also impact other consumers of `NotifyStream`.
2. Watch parameters of `tedge-log-plugin` were changed to only listen to `Modified` events, no longer listening to `FileCreated` events.

This fix was confirmed to produce only 1 reload when using following editors:

- nano
- vim
- neovim

More details regarding both parts can be found in their respective commit messages.

## Types of changes

## Paste Link to the issue

- #2451

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/notify.rs
+++ b/crates/common/tedge_utils/src/notify.rs
@@ -1,23 +1,41 @@
+use debouncer::DebouncedEvent;
 use notify::event::AccessKind;
 use notify::event::AccessMode;
 use notify::event::CreateKind;
-use notify::event::ModifyKind;
 use notify::event::RemoveKind;
-use notify::Config;
 use notify::EventKind;
 use notify::INotifyWatcher;
-use notify::RecommendedWatcher;
 use notify::RecursiveMode;
 use notify::Watcher;
+use notify_debouncer_full as debouncer;
 use std::hash::Hash;
 use std::path::Path;
 use std::path::PathBuf;
+use std::time::Duration;
 use strum_macros::Display;
 use tokio::sync::mpsc::channel;
 use tokio::sync::mpsc::Receiver;
 
+/// The time window in which fs events will be subject to debouncing.
+const NOTIFY_STREAM_DEBOUNCE_DURATION: Duration = Duration::from_millis(50);
+
+/// The type of filesystem event that happened.
+///
+/// It needs to be noted that filesystem notifications can be unreliable, and different editors can use different
+/// operations to update the file[1]. Additionally, we support different kinds of filesystem events, but often consumers
+/// want to react only once to when the file is updated, so they're forced between subscribing to everything and
+/// repeating work unnecessarily, or only to some and potentially missing other events. For our purposes, we always emit
+/// [`FsEvent::Modified`] if the content of the file changed for any reason, i.e. due to rename, delete, create, modify,
+/// etc.
+///
+/// [1]: https://github.com/notify-rs/notify/issues/113#issuecomment-281836995
 #[derive(Debug, Display, PartialEq, Eq, Clone, Hash, PartialOrd, Ord, Copy)]
 pub enum FsEvent {
+    /// Returned when the content of the file changes for any reason.
+    ///
+    ///
+    /// Emitted when the file is modified, deleted, created, or renamed. Consumers intrested in responding to changes
+    /// only once, without precise knowledge what caused the change of the content of the file, should use this event.
     Modified,
     FileDeleted,
     FileCreated,
@@ -47,7 +65,7 @@ pub enum NotifyStreamError {
 }
 
 pub struct NotifyStream {
-    watcher: INotifyWatcher,
+    debouncer: debouncer::Debouncer<INotifyWatcher, debouncer::NoCache>,
     pub rx: Receiver<(PathBuf, FsEvent)>,
 }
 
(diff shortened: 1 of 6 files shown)


crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -222,7 +222,12 @@ impl LogManagerActor {
         let path = match event {
             FsWatchEvent::Modified(path) => path,
             FsWatchEvent::FileDeleted(path) => path,
-            FsWatchEvent::FileCreated(path) => path,
+            // if we created the file ealier and now modify it, keep only later modify. some editors
```
reubenmiller, line 225:
```suggestion
            // if we created the file earlier and now modify it, keep only later modify. some editors
```

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -222,7 +222,13 @@ impl LogManagerActor {
...
             FsWatchEvent::Modified(path) => path,
             FsWatchEvent::FileDeleted(path) => path,
-            FsWatchEvent::FileCreated(path) => path,
+            // if we created the file earlier and now modify it, keep only later modify. some
+            // editors (e.g. vim) write to files by creating a temporary file, deleting old, and
+            // renaming temporary to previously deleted file. That way when we update, only 1 event
+            // is emitted.
+            //
+            // https://github.com/notify-rs/notify/issues/113#issuecomment-281836995
+            FsWatchEvent::FileCreated(_) => return Ok(()),
```
didier-wenzek, line 231: There is an issue with that. If the log plugin starts while its configuration is *not* present, then it starts assuming a default configuration. But then, as file creation events are ignored, there is no way to provide a log plugin configuration without restarting the plugin. I agree this is a corner case.

reubenmiller:
> 2. Watch parameters of tedge-log-plugin were changed to only listen to Modified events, no longer listening to FileCreated events.

Are we also sure that this change does not affect replacing existing files using `mv` and `cp`? We have some [system test coverage](https://github.com/reubenmiller/thin-edge.io/blob/1db13f11810929ddddf5b5cacf3c75720e2bb7a4/tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot#L185) for the configuration plugin, but none for the log manager. We should double check that we are not breaking any existing behaviour.

Also, shouldn't we also be doing the same handling for the configuration manager?

crates/extensions/tedge_log_manager/src/actor.rs
Bravo555: `FileCreated` means that a new file was created, but it still needs to be written to, so `FileCreated` is immediately followed by `FileModified` _most_ of the time, which the plugin picks up and correctly reloads.

I tested additional cases which result in successful reload:

- when the file is moved out and into the `/etc/tedge/plugins` directory (regardless if it was present when the plugin started) (the `FileCreated` event is not even emitted, because it's a `Rename` event, but `Modified` is still emitted)
- when plugin starts without the file and new one is created using `neovim` and by the shell (`echo "files = []" > tedge-log-plugin.toml`)
- when file is removed using `rm`

For all these the underlying library emits `EventKind::Modify(_)`.

Bravo555:
> Are we also sure that this change does not affect replacing existing files using `mv` and `cp`?

Moving files and creating files both also emit `Modified` event

> We have some [system test coverage](https://github.com/reubenmiller/thin-edge.io/blob/1db13f11810929ddddf5b5cacf3c75720e2bb7a4/tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot#L185) for the configuration plugin, but none for the log manager. We should double check that we are not breaking any existing behaviour.

This is a good candidate for a Rust integration test, which I'll add for this PR.

> Also, shouldn't we also be doing the same handling for the configuration manager?

We probably should. I'll update the configuration manager as well.

crates/extensions/tedge_log_manager/src/actor.rs
Bravo555: I must've copied this comment on autopilot, because only now I realize that actually it doesn't talk about why `FileCreated` is disabled, but about something else. My bad. Will replace that with some more helpful comment.

crates/common/tedge_utils/src/notify.rs
```diff
@@ -57,59 +61,79 @@ impl NotifyStream {
...
+                    // simplify event type
+                    let mut debounced_events = debounced_events
+                        .into_iter()
+                        .filter_map(|notify_event| {
+                            let event = match notify_event.kind {
+                                EventKind::Access(access_kind) => match access_kind {
+                                    AccessKind::Close(access_mode) => match access_mode {
+                                        AccessMode::Any | AccessMode::Other | AccessMode::Write => {
+                                            Some(FsEvent::Modified)
+                                        }
```
Bravo555, line 82: This is the branch that allows us to only listen to `FsEvent::Modified`. The actual sequence returned by lower-level `notify-debouncer-full` is:

```
[crates/common/tedge_utils/src/notify.rs:77] &notify_event = DebouncedEvent {
    event: Event {
        kind: Create(
            File,
        ),
        paths: [
            "/etc/tedge/plugins/tedge-log-plugin.toml",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
    time: Instant {
        tv_sec: 23846,
        tv_nsec: 112231843,
    },
}
[crates/common/tedge_utils/src/notify.rs:77] &notify_event = DebouncedEvent {
    event: Event {
        kind: Access(
            Close(
                Write,
            ),
        ),
        paths: [
            "/etc/tedge/plugins/tedge-log-plugin.toml",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
    time: Instant {
        tv_sec: 23846,
        tv_nsec: 112263954,
    },
}
[crates/common/tedge_utils/src/notify.rs:77] &notify_event = DebouncedEvent {
    event: Event {
        kind: Remove(
            File,
        ),
        paths: [
            "/etc/tedge/plugins/tedge-log-plugin.toml~",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
    time: Instant {
        tv_sec: 23846,
        tv_nsec: 112322524,
    },
}
```

when saving the file with nvim, or:

```
[crates/common/tedge_utils/src/notify.rs:77] &notify_event = DebouncedEvent {
    event: Event {
        kind: Create(
            File,
        ),
        paths: [
            "/etc/tedge/plugins/tedge-log-plugin.toml",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
    time: Instant {
        tv_sec: 23953,
        tv_nsec: 87955441,
    },
}
[crates/common/tedge_utils/src/notify.rs:77] &notify_event = DebouncedEvent {
    event: Event {
        kind: Access(
            Close(
                Write,
            ),
        ),
        paths: [
            "/etc/tedge/plugins/tedge-log-plugin.toml",
        ],
        attr:tracker: None,
        attr:flag: None,
        attr:info: None,
        attr:source: None,
    },
    time: Instant {
        tv_sec: 23953,
        tv_nsec: 87977201,
    },
}
```

when file is copied. Picking up `AccessKind::Close(AccessMode::Write)` allows us to reload appropriately.

didier-wenzek (approved):
Approved. The code is clear and correct. Using `notify-debouncer-full` was a great idea!

didier-wenzek:
> Are we also sure that this change does not affect replacing existing files using `mv` and `cp`? 

This is working as expected.

albinsuresh (approved):
LGTM.

Bravo555:
> This is a good candidate for a Rust integration test, which I'll add for this PR.

Instead of doing integration tests for both `tedge-log-plugin` and `tedge-configuration-plugin`, I've decided to only add 1 unit test: `notify::tests::modify_emitted_for_move_copy_create_delete` that checks if `FsEvent::Modified` is emitted when the file is modified, deleted, created, or renamed.

To add relevant tests for both plugins, I'd need to rewrite the plugins themselves a little bit, and we're currently more pressed for other stuff, so that should be sufficient for now.

crates/common/tedge_utils/src/notify.rs
```diff
@@ -267,21 +329,109 @@ mod tests {
...
+        let nested_dir = dir.dir("nested");
+
+        let mut stream = fs_notify_stream(&[&nested_dir.path()]).unwrap();
+
+        // Test create
+        let file = nested_dir.file("file");
+        assert_rx_stream(
+            [(
+                file.utf8_path().file_name().unwrap().to_string(),
+                vec![FsEvent::Modified],
```
albinsuresh, line 370: Here I was expecting `FsEvent::FileCreated` event as well in addition to the `FsEvent::Modified` event, as the watcher logic above clearly matches `EventKind::Create` and maps it to `FsEvent::FileCreated`. So, what am I missing here?

crates/common/tedge_utils/src/notify.rs
```diff
@@ -57,59 +76,81 @@ impl NotifyStream {
...
                                     }
-                                    AccessMode::Read | AccessMode::Execute => None,
+                                    CreateKind::Folder => Some(FsEvent::DirectoryCreated),
+                                },
+                                EventKind::Modify(modify_kind) => match modify_kind {
+                                    ModifyKind::Data(_)
+                                    | ModifyKind::Metadata(_)
+                                    | ModifyKind::Name(_)
+                                    | ModifyKind::Any
+                                    | ModifyKind::Other => Some(FsEvent::Modified),
```
albinsuresh, line 113:
```suggestion
                                EventKind::Modify(_) => Some(FsEvent::Modified),
```
Why not just that?

crates/common/tedge_utils/src/notify.rs
```diff
@@ -267,21 +329,109 @@ mod tests {
...
+            &mut stream,
+        )
+        .await;
+
+        // Test remove
+        std::fs::remove_file(file.path()).unwrap();
+        assert_rx_stream(
+            [(
+                file.utf8_path().file_name().unwrap().to_string(),
+                vec![FsEvent::Modified],
```
albinsuresh, line 414: Here also, I was expecting the `FsEvent::FileDeleted` event right before the `FsEvent::Modified` event, as per the logic: `allow consumers to only monitor for `Modified` event if they want` implemented in that second `for` loop.

albinsuresh (commented):
Some queries to better understand the test logic.

crates/common/tedge_utils/src/notify.rs
Bravo555: In this test I wanted to ensure that `Modified` is emitted, testing for if `FileCreated` is emitted is done in other tests. `FileCreated` is emitted before `Modified`, but the `assert_rx_stream` just ignores them, it's only interested in events that are inside the vector.

crates/common/tedge_utils/src/notify.rs
Bravo555: fixed

crates/common/tedge_utils/src/notify.rs
Bravo555: https://github.com/thin-edge/thin-edge.io/pull/2454#discussion_r1397172580

PR #2430, opened 2023-11-07
Add HTTPS/TLS support to the auth proxy

jarhodes314:
## Proposed changes
As #2397 is large and needs some significant changes from the agent's perspective, this PR separates out the changes to the auth proxy, which add the HTTPS support, but not (public-facing) support for client certificates, in an attempt to make some progress in merging the changes. Internally, all the support for certificate authentication exists, and the unit tests use this, but the built mapper will be unchanged apart from using HTTPS to serve the API.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
<br/>

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2461, opened 2023-11-16
Replace routerify in file transfer server with axum

jarhodes314:
## Proposed changes
This is a rewrite of the file transfer server to use axum instead of routerify, which forms the groundwork for adding HTTPS and certificate authentication support once #2430 is merged.

The basic migration to axum was pretty trivial (see 4051213), but once this was complete, I refactored the code considerably to be more idiomatic axum code. The highlights of these changes are:

- **Meaningful handler return values**: the http handlers no longer hand-build responses, they return normal Rust values (e.g. `StatusCode` where all we do is set the HTTP status code, an error enum for handling any unsuccessful response)
- **Local HTTP testing**: the tests no longer listen on a port to test HTTP functionality, the services are now called directly
- **Extractors**: boilerplate called by the handlers to extract path information from the URI and convert that into an absolute path to read from/write to the filesystem is now called before a handler runs and passed in as an argument to the handler
- **General security improvements**: the API no longer exposes absolute paths to files to the user (e.g. when we respond with 404 not found for `/tedge/file-transfer/some/file`, we say `File not found: "some/file"` rather than `File not found: "/var/tedge/file-transfer/some/file"`). It also fixes what appeared to be a (very limited in scope) path traversal attack, where a user could access the data directory directly and not just the `file-transfer` subdirectory of this
- **Removing race conditions**: the previous code checked that files existed before reading them, and handled errors based on that, but there's a file could be deleted between checking it exists and attempting to read it, which would result in a spurious HTTP 500 error which should be a 404.

## Types of changes
Intended to be an improvement, but some minor bugs were fixed along the way, so checking both of those.

## Paste Link to the issue
<br/>

## Checklist


crates/core/tedge_agent/src/file_transfer_server/actor.rs
```diff
@@ -141,31 +144,26 @@ mod tests {
     }
 
     #[tokio::test]
-    #[serial_test::serial]
-    async fn check_server_does_not_panic_when_port_is_in_use() -> Result<(), anyhow::Error> {
+    async fn check_server_does_not_panic_when_port_is_in_use() -> anyhow::Result<()> {
```
jarhodes314, line 147: This refactoring is incidental, I noticed the test had both a hard sleep (so it was comparitively slow) and not terribly deterministic (no guarantee server one would attempt to bind first, no guarantee the port binding would fail within 100ms).

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -1,35 +1,47 @@
...
+use tokio::io::AsyncBufReadExt;
 use tokio::io::AsyncWriteExt;
+use tokio::io::BufReader;
+use tokio_util::io::ReaderStream;
 
+use super::error::FileTransferRequestError as Error;
+use super::request_files::FileTransferPaths;
+
+// TODO is this not just a repeat of code in tedge_config?
 const HTTP_FILE_TRANSFER_PORT: u16 = 8000;
```
jarhodes314, line 31: Probably outside of the scope of these changes, but it feels like we shouldn't duplicate what `tedge_config` knows here. If someone wants the default values, they could surely achieve that with a default config?

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -54,327 +69,370 @@ impl HttpConfig {
...
-fn clean_utf8_path(path: &Utf8Path) -> Utf8PathBuf {
-    Utf8PathBuf::from(path_clean::clean(path.as_str()))
 }
 
-fn separate_path_and_file_name(input: &Utf8Path) -> Option<(Utf8PathBuf, String)> {
-    let (relative_path, file_name) = input.as_str().rsplit_once('/')?;
-
-    let relative_path = Utf8PathBuf::from(relative_path);
-    Some((relative_path, file_name.into()))
+fn separate_path_and_file_name(input: &Utf8Path) -> Option<(&Utf8Path, &str)> {
```
jarhodes314, line 73: The semantics of this function are completely different to what was originally written (and very directly tested (see https://github.com/thin-edge/thin-edge.io/blob/ccff7ee2720fa688e19e0cdd5b4bc7c1c3b3bb79/crates/core/tedge_agent/src/file_transfer_server/http_rest.rs#L250-L261)). However these tests were not testing anything the user could achieve, as the API did path cleanup (which removed trailing slashes) and (I think) the code we call afterwards would have returned an error if we had been able to call this with a trailing slash (as we would have created the entire path as directories and then attempted to create a file with an empty name inside that).

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -54,327 +69,370 @@ impl HttpConfig {
...
     mut request: Request<Body>,
-    file_transfer: &HttpConfig,
-) -> Result<Response<Body>, FileTransferError> {
-    let full_path = file_transfer.local_path_for_uri(request.uri().to_string())?;
-
-    let mut response = Response::new(Body::empty());
-
-    if let Some((relative_path, file_name)) = separate_path_and_file_name(&full_path) {
-        let root_path = file_transfer.data_dir.clone();
-        let directories_path = root_path.join(relative_path);
```
jarhodes314, line 112: This `relative_path` was very definitely an absolute path, so this code was entirely superfluous, as it just returned `relative_path` in any case.

crates/core/tedge_agent/src/file_transfer_server/error.rs
```diff
@@ -24,8 +23,85 @@ pub enum FileTransferError {
...
+                tracing::error!("{self}");
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    "Internal error".to_owned(),
+                )
+            }
+            DeleteIoError { path, .. } => {
+                tracing::error!("{self}");
+                (
+                    StatusCode::FORBIDDEN,
```
jarhodes314, line 92: Do these really want to be `403 Forbidden` (it's what the code did previously)? It feels like an encapsulation failure to recognise for certain paths that we can't delete/upload to them.

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -54,327 +68,370 @@ impl HttpConfig {
...
-            .strip_prefix(&self.file_transfer_uri)
-            .ok_or(FileTransferError::InvalidURI { value: ref_uri })?;
-
-        // This path is relative to the file transfer dir
-        let full_path = self.data_dir.join(path);
-
-        // One must check that once normalized (i.e. any `..` removed)
-        // the path is still under the file transfer dir
-        let clean_path = clean_utf8_path(&full_path);
-        if clean_path.starts_with(&self.data_dir) {
```
jarhodes314, line 81: This was only checking that the file was in e.g. `/var/tedge` since all the manipulation in this function was based around `file-transfer` being part of the URL path (i.e. in the variable `path` in this function), so it was possible for a user to read from/write to `/var/tedge`, not just `/var/tedge/file-transfer`. I'm assuming this wasn't intentional and I've fixed it by storing `file_transfer_dir` inside `HttpConfig` instead.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
```diff
@@ -141,31 +144,26 @@ mod tests {
...
-        // while handle_two is finished.
-        assert!(!handle_one.is_finished());
-        assert!(handle_two.is_finished());
+        tokio::select! {
+            // Ensure we bind server_one first by polling that future first
+            biased;
+            res = server_one => bail!("expected second server to finish first, but first finished with: {res:?}"),
+            res = server_two => ensure!(res.is_err(), "expected server two to fail with port binding error, but no error was found"),
+            _ = tokio::time::sleep(Duration::from_secs(5)) => bail!("timed out waiting for actor to stop running"),
+        }
```
jarhodes314, line 166: This refactoring is incidental, I noticed the test had both a hard sleep (so it was comparitively slow) and not terribly deterministic (no guarantee server one would attempt to bind first, no guarantee the port binding would fail within 100ms).

crates/core/tedge_agent/src/file_transfer_server/error.rs
Bravo555: IMO "could not delete path due to IO error" could happen due e.g. path not existing, in which case we need to return 404, or due to not having permissions, in which case we should return 500, so if I'm not wrong, then even using a single return code for all IO errors is kinda bad. As for `403 Forbidden`, I also feel that it should be used when the client that made the request doesn't have permissions to do something, which doesn't seem to be the case here.

jarhodes314: Yeah, that was roughly what I was thinking. We currently handle the not found case by returning `202 Accepted`, like the successful deletion case (which I assume is to make `DELETE` requests idempotent).  Returning `404 Not Found` may also make sense, or if the file previously existed, `410 Gone`, although a success response tells the user the file definitely no longer exists, and the request definitely succeeded.

I think if we've hit the case handled by this code, it's some error accessing the file system, which in my mind should definitely be a `500 Internal Server Error`.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
```diff
@@ -141,31 +144,26 @@ mod tests {
...
 
     #[tokio::test]
-    #[serial_test::serial]
-    async fn check_server_does_not_panic_when_port_is_in_use() -> Result<(), anyhow::Error> {
+    async fn check_server_does_not_panic_when_port_is_in_use() -> anyhow::Result<()> {
         let ttd = TempTedgeDir::new();
 
         let http_config = HttpConfig::default()
             .with_data_dir(ttd.utf8_path_buf().into())
             .with_port(3746);
```
Bravo555, line 152: suggestion: This will fail if port 3746 is already bound

I know this is previous behaviour, but seeing as how you improved the rest of the test, could you also change this to bind on port 0 and then query which port was actually bound to use later? If it's annoying to query the port after the port was bound, then that's okay, I don't think we use this port anywhere else, so it won't be a problem when running tests in parallel.

crates/core/tedge_agent/src/file_transfer_server/error.rs
```diff
@@ -24,8 +23,85 @@ pub enum FileTransferError {
...
+    fn into_response(self) -> axum::response::Response {
+        use FileTransferRequestError::*;
+        match &self {
+            FromIo(_) | PathRejection(_) => {
+                tracing::error!("{self}");
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    "Internal error".to_owned(),
+                )
+            }
```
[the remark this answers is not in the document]
jarhodes314: It's a bit weird that `Path` returns `400` when it fails; given it's in a URL, I would expect it to be a 404.

Any valid path component can be converted into `Utf8PathBuf`, which is what we're using. The only reason why we would get a `PathRejection` is if we'd failed to include a path component in the URL, which would be an error in the code and thus should be a `500`.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
```diff
@@ -141,31 +144,26 @@ mod tests {
...
+        let server_two = FileTransferServerBuilder::new(config_clone).build().run();
 
-        // to check for the error, we assert that handle_one is still running
-        // while handle_two is finished.
-        assert!(!handle_one.is_finished());
-        assert!(handle_two.is_finished());
+        tokio::select! {
+            // Ensure we bind server_one first by polling that future first
+            biased;
+            res = server_one => bail!("expected second server to finish first, but first finished with: {res:?}"),
```
albinsuresh, line 163:
```suggestion
            res = server_one => bail!("expected second server startup to fail first with port binding error, but first finished with: {res:?}"),
```

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -54,327 +68,370 @@ impl HttpConfig {
...
 
-async fn delete(
-    request: Request<Body>,
-    file_transfer: &HttpConfig,
-) -> Result<Response<Body>, FileTransferError> {
-    let full_path = file_transfer.local_path_for_uri(request.uri().to_string())?;
-
-    let mut response = Response::new(Body::empty());
+// Not a typo, snake_case for: 'err is "is a directory"'
+fn err_is_is_a_directory(e: &std::io::Error) -> bool {
```
albinsuresh, line 131:
```suggestion
fn err_kind_is_a_directory(e: &std::io::Error) -> bool {
```
Minor: If you wanna avoid that naming confusion

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -54,327 +68,370 @@ impl HttpConfig {
...
-                *response.status_mut() = hyper::StatusCode::ACCEPTED;
-                Ok(response)
-            }
-            Err(_err) => {
-                *response.status_mut() = hyper::StatusCode::FORBIDDEN;
-                Ok(response)
-            }
-        }
+async fn delete_file(req: FileTransferPaths) -> Result<StatusCode, Error> {
+    match tokio::fs::remove_file(&req.full).await {
```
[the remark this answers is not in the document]
albinsuresh: I feel a 404 is sufficient, as the API is for deleting a **file** with the given name and if the target path is **not a file**, then it's still an "resource not found" case to me.

crates/core/tedge_agent/src/file_transfer_server/error.rs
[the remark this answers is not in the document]
jarhodes314: I've changed this (still to be pushed) to call `into_response` on the inner value, so `axum` can determine the correct status code.

crates/core/tedge_agent/src/file_transfer_server/request_files.rs
```diff
@@ -0,0 +1,84 @@
...
+use axum::extract::Path;
+use axum::http::request::Parts;
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+
+use super::error::FileTransferRequestError;
+use super::http_rest::HttpConfig;
+
+/// The paths inferred from a request to the File Transfer Service
+pub struct FileTransferPaths {
```
albinsuresh, line 15:
```suggestion
pub struct FileTransferPath {
```
Nitpick(Optional): Although this struct contains two paths, they still point to the same logical entity, right? So, a plural name just sounds weird.

crates/core/tedge_agent/src/file_transfer_server/error.rs
```diff
@@ -24,8 +23,85 @@ pub enum FileTransferError {
...
+        status_code.into_response()
+    }
+}
+
+impl IntoResponse for FileTransferRequestError {
+    fn into_response(self) -> axum::response::Response {
+        use FileTransferRequestError::*;
+        match &self {
+            FromIo(_) | PathRejection(_) => {
+                tracing::error!("{self}");
```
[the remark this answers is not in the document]
jarhodes314: I'm not sure axum's `tracing` is quite what we want here. Here, I want to determine whether I'm hiding information from the user (which is when we have an internal server error), and if so, log that information. This avoids polluting the server logs with errors arising from clients misusing the API, but means we don't expose implementation details to the user of the API. If we rely on some sort of middleware for logging, either it would take an input of `FileTransferRequestError`, in which case we might as well do everything here as we'd still need to log selectively, or we'd take the converted response, by which point we know the status code, so it's easy to log selectively, but we've thrown away the pertinent information (e.g. an IO error).

crates/core/tedge_agent/src/file_transfer_server/error.rs
```diff
@@ -24,8 +23,85 @@ pub enum FileTransferError {
...
+    PathRejection(#[from] PathRejection),
+}
+
 impl From<FileTransferError> for RuntimeError {
     fn from(error: FileTransferError) -> Self {
         RuntimeError::ActorError(Box::new(error))
     }
 }
+
+impl IntoResponse for FileTransferError {
```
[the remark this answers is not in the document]
jarhodes314: Ah, I left this in by mistake when I moved things into a separate error, type. I'll remove it.

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
jarhodes314: That doesn't really avoid the issue, because we're still checking that the kind is "is a directory". Omitting the word kind makes it clearer that the input is a `std::io::Error` and not `std::io::ErrorKind`, and that is definitely a more ergonomically sensible choice here in my mind.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
jarhodes314: It's a good point, and actually means all the complexity can go away, as the test could be simplified to "create a tcp listener" then "spawn a server with a timeout" and check we get an error, which removes the race issue entirely (and the assumptions that make `biased` correct).

jarhodes314:
@albinsuresh 

> I feel a 404 is sufficient, as the API is for deleting a file with the given name and if the target path is not a file, then it's still an "resource not found" case to me.

I agree, I'll implement that.

> 403 doesn't look bad, as [the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) mentions the following: The access is tied to the application logic, such as insufficient rights to a resource. which is the case here, right? I vaguely remember a similar discussion happening when the original author introduced this case in the first place and we settled with 403 for the same reason.

I don't understand what application logic is relevant here. With the current API at least, we have no control over permissions. If there's a permissions error on the file system, that implies some other process is interfering with `/var/tedge/file-transfer`, which is not an application-level error, it's an error from the environment. The question here mostly boils down to whether we want to assume it's some file that should have never been there (in which case encapsulating the error entirely and responding with `404 Not Found` is appropriate, or whether we assume the file was uploaded correctly, but something as changed the internal permissions so we can't access it, in which case a `500 Internal Server Error` is appropriate.

Thinking more broadly, a useful litmus test to apply here would be "if one user receives a 403 for a resource, there must exist some other user that is able to access the resource in question". If not, and the 403 response is global, then I don't believe this can be considered an authorization issue.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
jarhodes314: In response to https://github.com/thin-edge/thin-edge.io/pull/2461#discussion_r1396552608, I've now changed this test again, and as a result the remaining complexity around which port binds first (i.e. what I'm relying on the `biased` directive for) has disappeared entirely.

crates/core/tedge_agent/src/file_transfer_server/error.rs
jarhodes314: I don't understand what application logic is relevant here. With the current API at least, we have no control over permissions. If there's a permissions error on the file system, that implies some other process is interfering with /var/tedge/file-transfer, which is not an application-level error, it's an error from the environment. The question here mostly boils down to whether we want to assume it's some file that should have never been there (in which case encapsulating the error entirely and responding with 404 Not Found is appropriate, or whether we assume the file was uploaded correctly, but something as changed the internal permissions so we can't access it, in which case a 500 Internal Server Error is appropriate.

Thinking more broadly, a useful litmus test to apply here would be "if one user receives a 403 for a resource, there must exist some other user that is able to access the resource in question". If not, and the 403 response is global, then I don't believe this can be considered an authorization issue.

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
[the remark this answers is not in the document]
jarhodes314: I've now added a 404 case for deleting a directory

crates/core/tedge_agent/src/file_transfer_server/error.rs
jarhodes314: @didier-wenzek @reubenmiller do you have opinions on this point?

didier-wenzek: Deleting a resource that doesn't exist is definitely an error. Returning 202 in such a case makes the request idempotent.

Trying to delete a non-empty directory, should be considered as a user error as we don't want to support recursive deletes. So 400 makes sense. I would do the same for empty-directories.

If the error is due to missing access, I would say this is a 500 as the file-transfer service should own all the files under file transfer.

Bravo555: > 403 doesn't look bad, as [the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) mentions the following: `The access is tied to the application logic, such as insufficient rights to a resource.`

From RFC 9110:

> [The 403 (Forbidden) status code indicates that the server understood the request but refuses to fulfill it.](https://www.rfc-editor.org/rfc/rfc9110#section-15.5.4-1)

If we get an IO error due to missing file permissions, I would say that the server tried to execute the request, but failed, due to a misconfiguration (e.g. not having write access to `file-transfer` directory), in which case 500 would be fitting.

jarhodes314: > Trying to delete a non-empty directory, should be considered as a user error as we don't want to support recursive deletes. So 400 makes sense. I would do the same for empty-directories.

I think if it is a directory, 404 makes more sense as it the error is in the URL and the issue is that the resource in question cannot be interacted with (and essentially doesn't exist), not that the request is inherently invalid. I'll make sure the error message clarifies that the file we attempted to delete is a directory.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
jarhodes314: This error message went away after refactoring the test to use a dynamic port

albinsuresh (commented):
Happy to approve once the `DeleteIoError` response is changed from `FORBIDDEN` to `INTERNAL_SERVER_ERROR` as concluded.

jarhodes314:
> Happy to approve once the `DeleteIoError` response is changed from `FORBIDDEN` to `INTERNAL_SERVER_ERROR` as concluded.

Done in https://github.com/thin-edge/thin-edge.io/pull/2461/commits/0056f9c907ed7271c43996d89c4c366583b147fa

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -54,327 +70,382 @@ impl HttpConfig {
...
+            Ok(()) => Ok(StatusCode::CREATED),
+            Err(err) if source_err_is_is_a_directory(&err) => Err(Error::CannotUploadDirectory {
+                path: paths.request,
+            }),
+            Err(err) => Err(internal_error(err, paths.request)),
         }
     } else {
-        *response.status_mut() = hyper::StatusCode::FORBIDDEN;
+        Err(internal_error(
+            anyhow!("cannot retrieve directory name for for {}", paths.full),
```
albinsuresh, line 100:
```suggestion
            anyhow!("cannot retrieve directory name for {}", paths.full),
```

crates/core/tedge_agent/src/file_transfer_server/actor.rs
```diff
@@ -141,31 +144,22 @@ mod tests {
...
 
-        // although the code inside handle_two throws an error it does not panic.
-        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
-
-        // to check for the error, we assert that handle_one is still running
-        // while handle_two is finished.
-        assert!(!handle_one.is_finished());
-        assert!(handle_two.is_finished());
+        tokio::select! {
+            res = server => ensure!(res.is_err(), "expected server startup to fail with port binding error, but actor exited succesfully"),
```
albinsuresh, line 160:
```suggestion
            res = server => ensure!(res.is_err(), "expected server startup to fail with port binding error, but actor exited successfully"),
```
Spellcheck warning

albinsuresh (approved):

didier-wenzek (approved):
Approved.

The next step will be to enable HTTPS support to this file transfer service, introducing `http.cert_path`, `http.key_path` and `http.ca_path` and considering to reuse `http.client.auth.cert_file` and `http.client.auth.cert_file` to authenticate `tedge-agent` when acting as a client (auth proxy) as well as a server (file transfer).

crates/core/tedge_agent/src/file_transfer_server/error.rs
```diff
@@ -24,8 +23,67 @@ pub enum FileTransferError {
...
+    OtherDelete {
+        source: std::io::Error,
+        path: RequestPath,
+    },
+
+    #[error("Request to upload to {path:?} failed: {source:?}")]
+    OtherUpload {
+        source: anyhow::Error,
+        path: RequestPath,
+    },
```
Bravo555, line 47: suggestion: I'd be for keeping old just `Delete` or `Upload` for these variants, I think it's fine if they're just anyhow errors. Current names kinda make one think _"`OtherUpload`? Well, what about the first upload?"_

crates/core/tedge_agent/src/file_transfer_server/error.rs
```diff
@@ -24,8 +23,67 @@ pub enum FileTransferError {
...
+    #[error(transparent)]
+    FromIo(#[from] std::io::Error),
+
+    #[error("Cannot delete: {path:?} is a directory, not a file")]
+    CannotDeleteDirectory { path: RequestPath },
+
+    #[error("Cannot upload: {path:?} is a directory, not a file")]
+    CannotUploadDirectory { path: RequestPath },
+
+    #[error("Request to delete {path:?} failed: {source}")]
```
Bravo555, line 37: note: We need to be careful with including source error message in parent error message.

I think the recommendation is to not include source error messages like this because proper error reporting/diagnostics will themselves call `.source()` to print messages for the entire error chain. However, for code that is not aware of this, and doesn't depend on any error reporting crates, it will most likely just grab outer error message. It depends where the error is used. For the FTS, I think the errors appear only in HTTP responses and logs, but 1) I'm not sure how they are exactly reported by axum 2) it's all internal so it doesn't matter that much.

Bravo555 (approved):
I got answers to my question and most pressing feedback was addressed, so approving.

crates/core/tedge_agent/src/file_transfer_server/error.rs
jarhodes314: Sure, I won't change this right now as I don't want to delay merging further, but I'll make the change when I do the next step to add TLS.

PR #2463, opened 2023-11-16
Give the default failure "reason" when commands failed

rina23q:
## Proposed changes
Without giving the default reason, mappers will parse an error when command status is failed but no "reason" is given. However, "reason" filed should be optional in command payloads.

This bug was found during writing a unit test for firmware_update support.  
```
[te/device/main///cmd/firmware_update/c8y-mapper-2023-11-16T14:02:31.419901262Z] {
   "status":"failed",
   "remoteUrl":"https://speed.hetzner.de/10GB.bin",
   "name":"10GB",
   "version":"2.0.0"
}
[te/errors] missing field `reason` at line 6 column 1
```

The bahaviour is covered by the unit test `handle_log_upload_executing_and_failed_cmd_for_main_device()`.

Maybe worth adding a robot test?

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/messages.rs
+++ b/crates/core/tedge_api/src/messages.rs
@@ -489,8 +489,13 @@ pub enum CommandStatus {
     Executing,
     Successful,
     Failed {
+        #[serde(default = "default_failure_reason")]
         reason: String,
     },
+}
+
+fn default_failure_reason() -> String {
+    "Unknown reason".to_string()
 }
 
 /// TODO: Deprecate OperationStatus
--- a/crates/extensions/c8y_mapper_ext/src/log_upload.rs
+++ b/crates/extensions/c8y_mapper_ext/src/log_upload.rs
@@ -563,8 +563,7 @@ mod tests {
             "dateFrom": "2013-06-22T17:03:14.123+02:00",
             "dateTo": "2013-06-23T18:03:14.123+02:00",
             "searchText": "ERROR",
-            "lines": 1000,
-            "reason": "Something went wrong"
+            "lines": 1000
         })
                 .to_string(),
         ))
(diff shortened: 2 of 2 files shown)


crates/core/tedge_api/src/messages.rs
```diff
@@ -489,10 +489,15 @@ pub enum CommandStatus {
     Executing,
     Successful,
     Failed {
+        #[serde(default = "default_failure_reason")]
         reason: String,
     },
 }
 
+fn default_failure_reason() -> String {
+    "No failure reason is given".to_string()
```
didier-wenzek, line 498: I'm not happy with this message which raises the question *given by who*.

That said I agree, it's hard to come with something sensible. "Unknown error"? In any case, as a user, this is frustrating.

didier-wenzek: A default is indeed required.

albinsuresh: Making the `reason` was indirectly solving that problem, right? It forces the users to provide a reason when there's a failure (as long as it is mandatory only for the `failed` state). When there's a failure, the user has to create a JSON message with the `status` payload anyway. So, why not make him add the `reason` field as well? Why give an option to the user to shoot himself in the foot?

didier-wenzek: > So, why not make him add the reason field as well? Why give an option to the user to shoot himself in the foot?

The issue highlighted by @rina23q is that if by mistake no reason is given then the final error message is more then puzzling ( "missing field reason at line 6 column 1"). A default "Unknown reason" string will be clearer both for the agent developer and the end-user.

```suggestion
    "Unknown reason".to_string()
```

rina23q: > So, why not make him add the reason field as well?
I believe that `reason` should not be mandatory. Highly recommended to add the reason for the failure though.

didier-wenzek (approved):
Approved.

PR #2466, opened 2023-11-17
Ordered processing of entity registration messages before any other messages

albinsuresh:
## Proposed changes

Spec
- [x] Problems statements
- [x] Solution proposals
- [x] Finalized solution

Impl
- [x] Caching of early telemetry messages
- [x] Caching of early twin data messages
- [x] Handling early child entity registration messages
- [ ] Persist entity store: Deferred for later

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -12,6 +12,8 @@ use crate::mqtt_topics::Channel;
 use crate::mqtt_topics::EntityTopicId;
 use crate::mqtt_topics::MqttSchema;
 use crate::mqtt_topics::TopicIdError;
+use crate::pending_entity_store::PendingEntityData;
+use crate::pending_entity_store::PendingEntityStore;
 use log::debug;
 use mqtt_channel::Message;
 use serde_json::json;
@@ -133,6 +135,7 @@ pub struct EntityStore {
     external_id_validator_fn: ExternalIdValidatorFn,
     // TODO: this is a c8y cloud specific concern and it'd be better to put it somewhere else.
     default_service_type: String,
+    pending_entity_store: PendingEntityStore,
 }
 
 impl EntityStore {
@@ -150,19 +153,23 @@ impl EntityStore {
         SF: 'static + Send + Sync,
     {
         Self::with_main_device_and_default_service_type(
+            MqttSchema::default(),
             main_device,
             "service".to_string(),
             external_id_mapper_fn,
             external_id_validator_fn,
+            100,
         )
     }
 
     #[must_use]
     pub fn with_main_device_and_default_service_type<MF, SF>(
+        mqtt_schema: MqttSchema,
         main_device: EntityRegistrationMessage,
         default_service_type: String,
         external_id_mapper_fn: MF,
         external_id_validator_fn: SF,
+        telemetry_cache_size: usize,
     ) -> Option<Self>
     where
         MF: Fn(&EntityTopicId, &EntityExternalId) -> EntityExternalId,
@@ -191,6 +198,7 @@ impl EntityStore {
             external_id_mapper: Box::new(external_id_mapper_fn),
             external_id_validator_fn: Box::new(external_id_validator_fn),
             default_service_type,
+            pending_entity_store: PendingEntityStore::new(mqtt_schema, telemetry_cache_size),
         })
     }
 
@@ -330,17 +338,58 @@ impl EntityStore {
     pub fn update(
         &mut self,
         message: EntityRegistrationMessage,
+    ) -> Result<(Vec<EntityTopicId>, Vec<PendingEntityData>), Error> {
+        match self.register_entity(message.clone()) {
+            Ok(affected_entities) => {
+                if affected_entities.is_empty() {
+                    Ok((vec![], vec![]))
+                } else {
+                    let topic_id = message.topic_id.clone();
+                    let current_entity_data =
+                        self.pending_entity_store.take_cached_entity_data(message);
+                    let mut pending_entities = vec![current_entity_data];
+
+                    let pending_children = self
+                        .pending_entity_store
+                        .take_cached_child_entities_data(&topic_id);
+                    for pending_child in pending_children {
+                        let child_reg_message = pending_child.reg_message.clone();
+                        self.register_entity(child_reg_message)?;
+                        pending_entities.push(pending_child);
+                    }
+
+                    Ok((affected_entities, pending_entities))
+                }
+            }
+            Err(Error::NoParent(_)) => {
+                // When a child device registration message is received before the parent is registered,
+                // cache it in the unregistered entity store to be processed later
+                self.pending_entity_store
+                    .cache_early_registration_message(message);
+                Ok((vec![], vec![]))
+            }
+            Err(err) => Err(err),
+        }
+    }
+
+    fn register_entity(
+        &mut self,
+        message: EntityRegistrationMessage,
     ) -> Result<Vec<EntityTopicId>, Error> {
         debug!("Processing entity registration message, {:?}", message);
-        let topic_id = message.topic_id;
+        let topic_id = message.topic_id.clone();
 
         let mut affected_entities = vec![];
 
         let parent = match message.r#type {
             EntityType::MainDevice => None,
-            EntityType::ChildDevice => message.parent.or_else(|| Some(self.main_device.clone())),
+            EntityType::ChildDevice => message
+                .parent
+                .clone()
+                .or_else(|| Some(self.main_device.clone())),
             EntityType::Service => message
                 .parent
+                .clone()
                 .or_else(|| topic_id.default_parent_identifier())
                 .or_else(|| Some(self.main_device.clone())),
         };
@@ -512,6 +561,10 @@ impl EntityStore {
         }
 
         Ok(true)
+    }
+
+    pub fn cache_early_data_message(&mut self, message: Message) {
+        self.pending_entity_store.cache_early_data_message(message)
     }
 }
 
@@ -691,6 +744,31 @@ impl EntityRegistrationMessage {
         })
     }
 
+    pub fn new_custom(topic_id: EntityTopicId, r#type: EntityType) -> Self {
+        EntityRegistrationMessage {
+            topic_id,
+            r#type,
+            external_id: None,
+            parent: None,
+            other: Map::new(),
+        }
+    }
+
+    pub fn with_parent(mut self, parent: EntityTopicId) -> Self {
+        let _ = self.parent.insert(parent);
+        self
+    }
+
+    pub fn with_external_id(mut self, external_id: EntityExternalId) -> Self {
+        let _ = self.external_id.insert(external_id);
+        self
+    }
+
+    pub fn with_other_fragment(mut self, key: String, value: JsonValue) -> Self {
+        let _ = self.other.insert(key, value);
+        self
+    }
+
     /// Creates a entity registration message for a main device.
     pub fn main_device(main_device_id: String) -> Self {
         Self {
@@ -830,7 +908,7 @@ mod tests {
             )
             .unwrap();
 
-        assert_eq!(updated_entities, ["device/main//"]);
+        assert_eq!(updated_entities.0, ["device/main//"]);
         assert_eq!(
             store.child_devices(&EntityTopicId::default_main_device()),
             ["device/child1//"]
@@ -845,7 +923,7 @@ mod tests {
                 .unwrap(),
             )
             .unwrap();
-        assert_eq!(updated_entities, ["device/main//"]);
+        assert_eq!(updated_entities.0, ["device/main//"]);
         let children = store.child_devices(&EntityTopicId::default_main_device());
         assert!(children.iter().any(|&e| e == "device/child1//"));
         assert!(children.iter().any(|&e| e == "device/child2//"));
@@ -866,7 +944,7 @@ mod tests {
             })
             .unwrap();
 
-        assert_eq!(updated_entities, ["device/main//"]);
+        assert_eq!(updated_entities.0, ["device/main//"]);
         assert_eq!(
             store.services(&EntityTopicId::default_main_device()),
             ["device/main/service/service1"]
@@ -882,7 +960,7 @@ mod tests {
             })
             .unwrap();
 
-        assert_eq!(updated_entities, ["device/main//"]);
+        assert_eq!(updated_entities.0, ["device/main//"]);
         let services = store.services(&EntityTopicId::default_main_device());
         assert!(services
             .iter()
@@ -890,21 +968,6 @@ mod tests {
         assert!(services
             .iter()
             .any(|&e| e == &EntityTopicId::default_main_service("service2").unwrap()));
-    }
-
-    #[test]
-    fn forbids_nonexistent_parents() {
-        let mut store = new_entity_store();
-
-        let res = store.update(EntityRegistrationMessage {
-            topic_id: EntityTopicId::default_main_device(),
-            external_id: None,
-            r#type: EntityType::ChildDevice,
-            parent: Some(EntityTopicId::default_child_device("myawesomeparent").unwrap()),
-            other: Map::new(),
-        });
-
-        assert!(matches!(res, Err(Error::NoParent(_))));
     }
 
     #[test]
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -83,6 +83,7 @@ use tedge_api::mqtt_topics::EntityTopicId;
 use tedge_api::mqtt_topics::IdGenerator;
 use tedge_api::mqtt_topics::MqttSchema;
 use tedge_api::mqtt_topics::OperationType;
+use tedge_api::pending_entity_store::PendingEntityData;
 use tedge_api::DownloadInfo;
 use tedge_api::EntityStore;
 use tedge_config::TEdgeConfigError;
@@ -108,6 +109,7 @@ const CREATE_EVENT_SMARTREST_CODE: u16 = 400;
 const DEFAULT_EVENT_TYPE: &str = "ThinEdgeEvent";
 const FORBIDDEN_ID_CHARS: [char; 3] = ['/', '+', '#'];
 const REQUESTER_NAME: &str = "c8y-mapper";
+const EARLY_MESSAGE_BUFFER_SIZE: usize = 100;
 
 #[derive(Debug)]
 pub struct MapperConfig {
@@ -281,10 +283,12 @@ impl CumulocityConverter {
 
         let main_device = entity_store::EntityRegistrationMessage::main_device(device_id.clone());
         let entity_store = EntityStore::with_main_device_and_default_service_type(
+            mqtt_schema.clone(),
             main_device,
             service_type.clone(),
             Self::map_to_c8y_external_id,
             Self::validate_external_id,
+            EARLY_MESSAGE_BUFFER_SIZE,
         )
         .unwrap();
 
@@ -973,10 +977,22 @@ impl CumulocityConverter {
                         Err(e) => {
                             error!("Entity registration failed: {e}");
                         }
-                        Ok(affected_entities) if !affected_entities.is_empty() => {
-                            let mut c8y_message =
-                                self.try_convert_entity_registration(&register_message)?;
-                            registration_messages.append(&mut c8y_message);
+                        Ok((affected_entities, pending_entities))
+                            if !affected_entities.is_empty() =>
+                        {
+                            for pending_entity in pending_entities {
+                                // Register and convert the entity registration first
+                                let mut c8y_message = self
+                                    .try_convert_entity_registration(&pending_entity.reg_message)?;
+                                registration_messages.append(&mut c8y_message);
+
+                                // Process all the cached data messages for that entity
+                                let mut cached_messages =
+                                    self.process_cached_entity_data(pending_entity).await?;
+                                registration_messages.append(&mut cached_messages);
+                            }
+
+                            return Ok(registration_messages);
                         }
                         Ok(_) => {}
                     }
@@ -1007,57 +1023,72 @@ impl CumulocityConverter {
                             );
                         }
                     } else {
-                        return Err(ConversionError::AutoRegistrationDisabled(
-                            source.to_string(),
-                        ));
+                        // When an entity data message is received before the entity itself is registered,
+                        // cache it in the unregistered entity store to be processed after the entity is registered
+                        self.entity_store.cache_early_data_message(message.clone());
+                        return Ok(vec![]);
                     }
                 }
             }
         }
 
-        let mut messages = match &channel {
+        let mut messages = self
+            .try_convert_data_message(source, channel, message)
+            .await?;
+
+        registration_messages.append(&mut messages);
+        Ok(registration_messages)
+    }
+
+    async fn try_convert_data_message(
+        &mut self,
+        source: EntityTopicId,
+        channel: Channel,
+        message: &Message,
+    ) -> Result<Vec<Message>, ConversionError> {
+        match &channel {
             Channel::EntityTwinData { fragment_key } => {
-                self.try_convert_entity_twin_data(&source, message, fragment_key)?
+                self.try_convert_entity_twin_data(&source, message, fragment_key)
             }
 
             Channel::Measurement { measurement_type } => {
-                self.try_convert_measurement(&source, message, measurement_type)?
+                self.try_convert_measurement(&source, message, measurement_type)
             }
 
             Channel::Event { event_type } => {
-                self.try_convert_event(&source, message, event_type).await?
+                self.try_convert_event(&source, message, event_type).await
             }
 
             Channel::Alarm { alarm_type } => {
-                self.process_alarm_messages(&source, message, alarm_type)?
+                self.process_alarm_messages(&source, message, alarm_type)
             }
 
             Channel::Command { .. } if message.payload_bytes().is_empty() => {
                 // The command has been fully processed
-                vec![]
+                Ok(vec![])
             }
 
             Channel::CommandMetadata { operation } => {
                 self.validate_operation_supported(operation, &source)?;
                 match operation {
-                    OperationType::Restart => self.register_restart_operation(&source).await?,
+                    OperationType::Restart => self.register_restart_operation(&source).await,
                     OperationType::SoftwareList => {
-                        self.register_software_list_operation(&source).await?
+                        self.register_software_list_operation(&source).await
                     }
                     OperationType::SoftwareUpdate => {
-                        self.register_software_update_operation(&source).await?
+                        self.register_software_update_operation(&source).await
                     }
-                    OperationType::LogUpload => self.convert_log_metadata(&source, message)?,
+                    OperationType::LogUpload => self.convert_log_metadata(&source, message),
                     OperationType::ConfigSnapshot => {
-                        self.convert_config_snapshot_metadata(&source, message)?
+                        self.convert_config_snapshot_metadata(&source, message)
                     }
                     OperationType::ConfigUpdate => {
-                        self.convert_config_update_metadata(&source, message)?
+                        self.convert_config_update_metadata(&source, message)
                     }
                     OperationType::FirmwareUpdate => {
-                        self.convert_firmware_metadata(&source, message)?
+                        self.convert_firmware_metadata(&source, message)
                     }
-                    _ => vec![],
+                    _ => Ok(vec![]),
                 }
             }
 
@@ -1066,14 +1097,14 @@ impl CumulocityConverter {
                 cmd_id,
             } if self.command_id.is_generator_of(cmd_id) => {
                 self.publish_restart_operation_status(&source, cmd_id, message)
-                    .await?
+                    .await
             }
 
             Channel::Command {
                 operation: OperationType::SoftwareList,
                 cmd_id,
             } if self.command_id.is_generator_of(cmd_id) => {
-                self.publish_software_list(&source, cmd_id, message).await?
+                self.publish_software_list(&source, cmd_id, message).await
             }
 
             Channel::Command {
@@ -1081,7 +1112,7 @@ impl CumulocityConverter {
                 cmd_id,
             } if self.command_id.is_generator_of(cmd_id) => {
                 self.publish_software_update_status(&source, cmd_id, message)
-                    .await?
+                    .await
             }
 
             Channel::Command {
@@ -1089,7 +1120,7 @@ impl CumulocityConverter {
                 cmd_id,
             } if self.command_id.is_generator_of(cmd_id) => {
                 self.handle_log_upload_state_change(&source, cmd_id, message)
-                    .await?
+                    .await
             }
 
             Channel::Command {
@@ -1097,7 +1128,7 @@ impl CumulocityConverter {
                 cmd_id,
             } if self.command_id.is_generator_of(cmd_id) => {
                 self.handle_config_snapshot_state_change(&source, cmd_id, message)
-                    .await?
+                    .await
             }
 
             Channel::Command {
@@ -1105,7 +1136,7 @@ impl CumulocityConverter {
                 cmd_id,
             } if self.command_id.is_generator_of(cmd_id) => {
                 self.handle_config_update_state_change(&source, cmd_id, message)
-                    .await?
+                    .await
             }
 
             Channel::Command {
@@ -1113,16 +1144,30 @@ impl CumulocityConverter {
                 cmd_id,
             } if self.command_id.is_generator_of(cmd_id) => {
                 self.handle_firmware_update_state_change(&source, message)
-                    .await?
+                    .await
             }
 
-            Channel::Health => self.process_health_status_message(&source, message).await?,
+            Channel::Health => self.process_health_status_message(&source, message).await,
 
-            _ => vec![],
-        };
+            _ => Ok(vec![]),
+        }
+    }
 
-        registration_messages.append(&mut messages);
-        Ok(registration_messages)
+    async fn process_cached_entity_data(
+        &mut self,
+        cached_entity: PendingEntityData,
+    ) -> Result<Vec<Message>, ConversionError> {
+        let mut converted_messages = vec![];
+        for message in cached_entity.data_messages {
+            let (source, channel) = self.mqtt_schema.entity_channel_of(&message.topic).unwrap();
+            converted_messages.append(
+                &mut self
+                    .try_convert_data_message(source, channel, &message)
+                    .await?,
+            );
+        }
+
+        Ok(converted_messages)
     }
 
     fn validate_operation_supported(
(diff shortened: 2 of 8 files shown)


didier-wenzek (commented):
These thoughts make be wonder if using MQTT is the right tool to register which entities are attached to each topic. And even, if this can be dynamic.

I would consider a declarative approach. To opt-out the default topic scheme, a user will have to provide a description of they custom topic scheme. It can be a set of regex rules or a user program returning a registration message for a entity id.

By a set of regex rules, I mean something along these lines:

- `ground/rasp///` -> main device
- `floor[0-9]+/rasp//` -> child device of main device
- `(?<parent> floor[0-9]+)/plc[0-9]+//` -> child device of `${parent}/rasp//`

reubenmiller:
> These thoughts make be wonder if using MQTT is the right tool to register which entities are attached to each topic. And even, if this can be dynamic.
> 
> I would consider a declarative approach. To opt-out the default topic scheme, a user will have to provide a description of they custom topic scheme. It can be a set of regex rules or a user program returning a registration message for a entity id.
> 
> By a set of regex rules, I mean something along these lines:
> 
> * `ground/rasp///` -> main device
> * `floor[0-9]+/rasp//` -> child device of main device
> * `(?<parent> floor[0-9]+)/plc[0-9]+//` -> child device of `${parent}/rasp//`

Yes, very valid points. Some kind of blocking call (e.g. to a HTTP endpoint) might be more appropriate to register devices (or patterns, nice idea).

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,104 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+Here the end-result is even more catastrophic, as the user expects a device with external-id: `desired-xid`,
+but the mapper has created one with the auto-generated external-id, which can't be changed afterwards.
+
+#### Initial solution thoughts
+
+One obvious solution to avoid this problem is to force the users to disable auto-registration,
+when they intend to use explicit registration even for a single entity.
+But the downside with that approach is that we force them to use explicit registration for everything,
+even the entities for which auto-registration would have been perfectly fine,
+just because a few entities required some customizations via explicit registration.
```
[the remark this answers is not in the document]
albinsuresh: I agree too. Anyone opting for a custom topic scheme is more likely to stick to that scheme for his entire fleet and not just a few devices. There was a typo in the above sequence diagram where the 5th message from the broker to the mapper was wrongly written as `te/floor1/plc1//` instead of `te/device/plc1//`, which may have caused the confusion. Corrected now.

But, unfortunately the issue highlighted here is not just limited to custom topic schemes but even for the default topic scheme in cases where an explicit registration is mandatory (nested child devices or when that customer wants to control the `id` of all devices in his fleet, without relying on the auto-generated IDs).

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,121 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+sequenceDiagram
+    participant Child device
+    participant MQTT Broker
+    participant Tedge Mapper
+    
+    Child device ->> MQTT Broker: Child device registration message to `te/device/plc1//` `{ "id": "desired-xid" }`
+    Child device ->> MQTT Broker: Telemetry message to `te/device/plc1///m/temperature`
+    MQTT Broker ->> Tedge Mapper: Telemetry message to `te/device/plc1///m/temperature`
+    Tedge Mapper ->> Tedge Mapper: Auto-register the `plc1` device with auto-generated `external-id`
+    MQTT Broker ->> Tedge Mapper: Child device registration message to `te/floor1/plc1//` `{ "id": "desired-xid" }`
```
albinsuresh, line 1426:
```suggestion
    MQTT Broker ->> Tedge Mapper: Child device registration message to `te/device/plc1//` `{ "id": "desired-xid" }`
```

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,121 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+Even with disk persistence, any indefinite waits will have to be prevented with timeouts,
+but that comes with additional uncertainty: what to do if the registration message gets delivered after the timeout?
+Crash itself (the mapper) with the hope of timely message delivery after the restart?
+
+##### Solution 2
+
+An alternate approach would be to send a feedback for entity registration messages,
+which the clients must wait on, before sending any other data associated to that entity.
+The feedback can be sent on the same topic on which the registration message was received on,
+with a field like `"@registered" : "true"`, in addition to any other enhancements done by tedge,
```
albinsuresh, line 1380: There's a problem with this approach when there are multiple mappers processing the same registration message. Who will add that `@registered` flag? They all can't add the same flag. We can let each mapper add its own cloud-specific flags like `@c8y-registered`, `@az-registered` etc, but then, the device side logic, which needs to wait for the registration complete acknowledgement, will have to wait for all those cloud specific flags separately, pushing cloud specific logic into the device, defeating the cloud-agnosticism that we promise.

Sending these "enriched registration messages" to separate cloud specific topics is even worse, as they suffer from the same limitation of "cloud specific logic pushed to device", with the overhead of multiple message copies per cloud for the same entity.

albinsuresh:
> These thoughts make me wonder if using MQTT is the right tool to register which entities are attached to each topic. And even, if this can be dynamic.

I can't agree more and most "capable devices" probably would prefer such an HTTP API over the clunky MQTT one. But, my understanding was that there would be many limited devices that we'll have to support, that can only handle MQTT and we'll have to provide an MQTT-only solution for them anyway. So, ideally we should have both APIs, but we might be forced to start with MQTT first, as that works for everyone.

> I would consider a declarative approach. To opt-out the default topic scheme, a user will have to provide a description of they custom topic scheme. It can be a set of regex rules or a user program returning a registration message for a entity id.
> 
> By a set of regex rules, I mean something along these lines:
> 
> * `ground/rasp///` -> main device
> * `floor[0-9]+/rasp//` -> child device of main device
> * `(?<parent> floor[0-9]+)/plc[0-9]+//` -> child device of `${parent}/rasp//`

Once we have the registration message feedback mechanism, on which the device-logic will have to wait, do we still need this template mechanism? Such a mechanism would still be useful to the customer, to avoid the explicit registration messages altogether. But this is not necessary in the context of this problem, right?

One other key point to note is that these issues are not just limited to custom topic schemes and the id-generation associated with it. That's definitely one key aspect, but not the only one. Even the default topic scheme suffers from these limitations when explicit registration is mandatory, like in the case of nested child devices. Getting the `id` wrong is not the only problem in this case, but getting the `@parent` wrong as well. I've added another dedicated sequence diagram for nested-child devices case, explaining the error path.

didier-wenzek:
> Once we have the registration message feedback mechanism, on which the device-logic will have to wait, do we still need this template mechanism? Such a mechanism would still be useful to the customer, to avoid the explicit registration messages altogether. But this is not necessary in the context of this problem, right?

We need to choose, sure.
- either a protocol over MQTT or HTTP to dynamically register entities
- or a static topic scheme used to deduce registration messages from entity identifiers.

I do think the latter would simplify things a lot (the topic<->entity relation being computed)
while being flexible enough (a user don't have to list in advance all and every entities - but to detail the logic behind his topic scheme).

> One other key point to note is that these issues are not just limited to custom topic schemes and the id-generation associated with it. That's definitely one key aspect, but not the only one. Even the default topic scheme suffers from these limitations when explicit registration is mandatory, like in the case of nested child devices. Getting the `id` wrong is not the only problem in this case, but getting the `@parent` wrong as well. I've added another dedicated sequence diagram for nested-child devices case, explaining the error path.

Why registration should be mandatory and necessarily done dynamically using registration messages?

A rule such as this one can be used to derive the external id as well as the parent of a grand child device,
assuming a scheme where a bunch of PLC are attached to a Rasp on each floor:

```
(?<floor> floor[0-9]+)/(?<plc> plc[0-9]+)// -> { "@id" -> "${floor}:${plc]", "@parent": "${floor}/rasp//" }
```

albinsuresh:
> A rule such as this one can be used to derive the external id as well as the parent of a grand child device,
> assuming a scheme where a bunch of PLC are attached to a Rasp on each floor:

For the external ID, I completely agree that some kind of correlation would definitely exist between the entity topic ID and its external ID, making that mapping definition fairly straight-forward. But, deriving the parent, I'm not fully sure. Looking at a few examples, like devices deployed in factory floors in a hierarchical way, it seems plausible, but I'm just not sure if we can generalise that assumption.

docs/src/references/mappers/c8y-mapper.md
albinsuresh: On a second thought, allowing all the mappers to enrich the source registration message by publishing to the same topic doesn't look good, as there may be conflicts between them. For example, the `c8y` and `az` mappers may have entirely different external ID generation logic. So, if both published the generated `@id` value to the same entity topic, one will get overwritten. So, either they'll have to make sure that they generate non-conflicting keys or have dedicated topics for each cloud.

tests/RobotFramework/tasks/debug.robot
```diff
@@ -23,18 +23,18 @@ ${DPKG_INSTALL_SCRIPT}   cd /setup \
...
-    Debug
-
-
-Debug device with locally built debian packages (upgrading from last official release)
-    ${DEVICE_SN}=                            Setup
-    Device Should Exist                      ${DEVICE_SN}
-    Install Locally Built Packages
-    Log Device Info
-    Debug
+# Debug device with latest official version
```
albinsuresh, line 26: Accidental commit. Will revert.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,191 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+
+To mitigate the telemetry message drop limitation of the above solution,
+cloud-specific MQTT topics can be used as secondary caches instead, for those "early messages".
+For e.g, when early measurement messages are received on `te/device/child1///m/environment`,
+they can be published back to `c8y-internal/device/child1///m/environment` topic and keep them queued there.
+Once the registration message is eventually processed, the mapper will dynamically subscribe to `c8y-internal/device/child1///+/+` topic
+and process those messages first, before the regular telemetry message on the `te/` topics are processed for it.
+The in-memory caching of non-telemetry data can also be replaced with this topic-based caching, but for minimal little gains.
+
+This solution requires some major changes in the current MQTT mapper to add support for dynamic subscriptions,
```
albinsuresh, line 1512: We can also use an append-only log instead of MQTT topics to avoid the dynamic MQTT subscription, as reading from that log would be much easier than the MQTT topic.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,191 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+#### Phase 1
+
+**Live mapper**
+
+When the mapper is live, if telemetry messages for a new entity is received before its registration message,
+they are simply dropped until that registration message is processed, with the assumption that,
+telemetry data updates will repeat eventually and hence will make up for the lost ones.
+But, all non-telemetry data like twin updates, capability messages etc are cached in-memory,
+and will be processed and registered in the entity store once the registration message eventually arrives.
+Keeping them cached in-memory is not a big deal, as they are eventually going to end-up in the in-memory entity store anyway.
```
albinsuresh, line 1479: This would become a problem only when some rogue client sends a lot of twin update messages without ever registering it.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,191 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+To avoid any conflicts between auto-registration and explicit registration,
+resulting in nested child devices from wrongly getting auto-registered as immediate child devices,
+we can just instruct the users to turn-off auto-registration if they have nested child devices in their deployment.
+Explicit registration must be used even for immediate child devices in such deployments.
+Or to say it differently, use auto-registration only in simple deployments where there are only immediate child devices.
+
+#### Phase 2
+
+To mitigate the telemetry message drop limitation of the above solution,
+cloud-specific MQTT topics can be used as secondary caches instead, for those "early messages".
```
didier-wenzek, line 1505: I'm afraid that this will only push the issue elsewhere. How using mapper private secondary caches would help?  This is only a different kind of cache (in-memory vs files vs delegated to the MQTT broker). The true issues are not specifically addressed and only *mitigated*: message re-ordering, user failing to send registration messages, ever-growing set of pending messages, bad interactions with auto-registration.

I other words, I would only consider phase 1 or a user-defined topic schema.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,191 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+and hence it has to be controlled with a deterministic waiting period.
+Since this timed waiting is during the startup, the impact won't be that visible to external users.
+
+**Auto registration**
+
+To avoid any conflicts between auto-registration and explicit registration,
+resulting in nested child devices from wrongly getting auto-registered as immediate child devices,
+we can just instruct the users to turn-off auto-registration if they have nested child devices in their deployment.
+Explicit registration must be used even for immediate child devices in such deployments.
+Or to say it differently, use auto-registration only in simple deployments where there are only immediate child devices.
```
didier-wenzek, line 1500: I can only agree on that: auto-registration must be explicitly documented as designed only for flat hierarchies where all the child devices are directly attached to the main device.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,191 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+
+**On mapper restart**
+
+Sticking to the same solution of dropping telemetry messages on a mapper restart might result in loosing too much data,
+when the volume of queued telemetry messages are very high on the broker,
+increasing the probability of a lot of those telemetry messages getting delivered to the mapper first,
+before their respective registration messages are.
+
+To avoid so much data loss, on restart the mapper will selectively subscribe to the registration topics first,
+process all the retained registration messages (with a timeout), after which the regular message processing will start.
```
didier-wenzek, line 1489: If I understand correctly, the plan is to gather all the registration messages and to re-order them so parents can be registered before the child entities and to *rebuild* the state reached before the mapper stopped.

What about a persisted representation of the mapper entity store? This store will be updated on the fly (as described above under *live mapper* and persisted on disk after each new registration. On start, the mapper simply uses this restored entity store and updates it on the fly, if there are still new registrations.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,191 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+what if the explicit registration message is received after the auto-registration has already happened after the timeout?
+Crashing itself isn't gonna help here either, as the entity is already auto-registered.
+
+### Final Solution
+
+The ordering problems when the mapper is live and when it is restarted would be handled differently.
+
+#### Phase 1
+
+**Live mapper**
```
didier-wenzek, line 1472: This sounds as a simple mitigation which can be improved in the future without breaking users' configuration files or expectations. I have to admit that my preferred solution (a user-defined regex-based topic scheme) is a commitment for a feature that will have to be maintained.

didier-wenzek (commented):
I'm okay for "Phase 1 - live update" and have a suggestion for an alternative solution to "Phase 1 - on mapper restart".

I see the "Phase 2" proposal as not mature enough. Is this even required?

I also have a question on where to put this content. These are more internal design issues and discussions rather than specifications as implied by the name of the updated file.

docs/src/references/mappers/c8y-mapper.md
albinsuresh: > The true issues are not specifically addressed and only mitigated: message re-ordering, user failing to send registration messages, ever-growing set of pending messages, bad interactions with auto-registration.

The advantage with dedicated mapper-local topics is that it would be easier to manage with a dynamic registration or even a dynamic MQTT client. Once the registration message is received, we can have a temporary MQTT client connecting to that topic and exhaust all the queued messages on that topic, before processing other regular messages.

If the user never sends a registration message, then yes there's the risk of all those messages getting queued on these topics, until the mapper restarts next time.

docs/src/references/mappers/c8y-mapper.md
albinsuresh: > What about a persisted representation of the mapper entity store?

Relying purely on the persisted entity store from the last runtime won't be enough, as there could be registration messages and other messages published when the mapper is down. On a restart, we'd want to avoid any races among those "previously unseen" messages as well. For that, a selective subscription to the registration topics on startup would be needed anyway.

We had considered this idea of "persisted entity store" for a different problem in the past: to distinguish already registered entities from any new ones registered when the mapper was down (to avoid resending all those already registered twin messages on mapper restart). So, this would still be needed, but unfortunately this alone won't solve the ordering problem that we're dealing with.

albinsuresh: The previous comment was made with the assumption that you're proposing it as an alternative to proposal 2 which is trying to avoid any message drops (even telemetry messages). But, since you clarified that it was proposed as an enhancement of proposal 1 itself, with its limitation around dropping some telemetry messages, the proposed enhancement is definitely easier to implement and we can go ahead with that.

> On a restart, we'd want to avoid any races among those "previously unseen" messages as well. For that, a selective subscription to the registration topics on startup would be needed anyway.

This was to minimize dropping as many telemetry messages as possible on restart (even the previously unseen telemetry messages delivered before their corresponding entity registration messages). But, if dropping some of those messages are acceptable (as we're okay dropping some telemetry messages during a live execution anyway), the proposal to use persistent entity store really simplifies the restart logic. It will also help in avoiding the "duplicate entity registration on restart" problem as well.

didier-wenzek: So the proposal is to:

- persist the entity store, so on restart the mapper don't have to process from scratch all the registration messages, but only those published during down-time, limiting ordering issues
- persist in-memory registration messages that cannot be processed till a parent entity is registered
- (possibly) persist on-disk messages for unregistered entities till the registration is received (this would be a bit more involved as this might require a log per unregistered entity).

Is this correct?

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,230 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+
+The message delivery problem gets even worse during a mapper restart,
+as there could be hundreds of retained messages: registration messages, twin update messages and command metadata messages,
+all delivered in quick succession to the mapper.
+Due to the sheer volume of these messages, the chances of out of order delivery is also very high,
+leading to the mapper dropping a lot of critical messages, leaving the entity store in a partial/broken state.
+
+#### Initial solution thoughts
+
+Handling this issue on a restart is easier, by using a special MQTT subscriber just on starup,
```
didier-wenzek, line 1400:
```suggestion
Handling this issue on a restart is easier, by using a special MQTT subscriber just on startup,
```

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,230 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+
+Any nested entity registration messages that arrives before their ancestors are registered
+will also be kept in the same "unregistered entity store" until all their ancestors are registered.
+
+To minimize dropping a lot of telemetry messages, they can also be cached briefly in a bounded queue,
+as in most practical cases, even if the registration messages are late, they will be delivered
+within a fair amount of time and we can avoid dropping messages in such cases.
+Keeping that queue bounded is critical to prevent it from growing indefinitely.
+
+There will also be max time limit for an unregistered entity to stay cached, after which it will be dropped.
```
didier-wenzek, line 1488: Rather than a time limit, I would use a fix-size cache evicting the oldest message when full. This is simpler to implement as no timer is required.

I would do the same for telemetry data. These caches can be small (e.g. 10 pending registrations, 100 pending telemetry data) as the point is to cope with messages that have been sent in order (root registrations < child registrations < telemetry) but received in a different order.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -1325,3 +1325,230 @@ te/device/main///cmd/log_upload/<cmd_id>
...
+
+The entity store could be rebuilt using the retained messages on the MQTT broker as well,
+but it is not easy to distinguish the previously registered entities from any new registration messages
+that may have arrived while the mapper was down.
+This problem doesn't exist while using a persisted entity store, which only contains previously registered entities.
+This can also be used to easily distinguish new registrations from the previously registered ones.
+
+:::note
+Since the real entity store is persisted, the "unregistered entity store" can also be persisted on disk
+to avoid memory pressure caused by rogue clients pounding data without ever registering themselves.
```
didier-wenzek, line 1515: I would not bother with that, focusing on well-behaving clients and dropping the oldest messages if the in-memory cache starts to be full.

didier-wenzek (commented):
The current text clarifies well the problem statement/alternative solutions/pros & cons;
but it's not so clear what's the retained solution among all the alternative ideas and combinations.

- A summary of the retained solution has to be given in the introduction, followed by the discussion (which can be unchanged).
- This content has to be moved under https://github.com/thin-edge/thin-edge.io/tree/main/design/decisions as this is a design discussion and not a reference guide.

docs/src/references/mappers/c8y-mapper.md
albinsuresh: The limitation with size based eviction is that the entries will remain there, consuming resources if that buffer never gets full. For example, if there are 99 large event messages sitting in the queue, with the 100th one never arriving, they'll sit there forever. That's why I was considering a time-based eviction instead, using our timer actor.

didier-wenzek (commented):
I'm okay with the proposed design.

This document can be improved with an *executive summary* of the retained solution.
3 bullet points would be enough:
- The c8y mapper persists its entity store on disk.
- Messages received from entities not registered yet are cached till the registration is received (with an eviction policy to deal with rogue clients).
- Auto-registration has to be disabled if the user opt-in for a custom MQTT scheme.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -232,6 +234,8 @@ pub struct CumulocityConverter {
     pub pending_fts_download_operations: HashMap<CmdId, FtsDownloadOperationData>,
 
     pub command_id: IdGenerator,
+
+    partial_entity_store: PartialEntityStore,
```
didier-wenzek, line 238: Having both an entity store and a partial entity store might be really confusing.

Wouldn't be better to use a facade to the actual entity store:

```rust
struct CumulocityConverter  {
     entity_store: PartialEntityStore,
}

struct PartialEntityStore {
      entity_store: EntityStore,
      cache: CacheOfPendingRegistration,
}
```

crates/core/tedge_api/src/partial_entity_store.rs
```diff
@@ -0,0 +1,150 @@
+use crate::entity_store::EntityRegistrationMessage;
+use crate::mqtt_topics::Channel;
+use crate::mqtt_topics::EntityTopicId;
+use crate::ring_buffer::RingBuffer;
+use mqtt_channel::Message as MqttMessage;
+use std::collections::HashMap;
+
+const EARLY_MESSAGE_BUFFER_SIZE: usize = 5;
+
+pub struct PartialEntityStore {
```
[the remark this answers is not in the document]
albinsuresh: Yeah the doc is missing. Will add. Initially it was `UnregisteredEntityStore` then I renamed it to this for something shorter. Does that sound better? Or any better suggestion.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Yeah, that would be better. Initially I even had the `PartialEntityStore` inside the `EntityStore` itself. But then I moved it out to reduce the impact on existing APIs. But yeah, I agree, some kind of facade would definitely be better.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -613,7 +613,55 @@ pub struct EntityRegistrationMessage {
     pub other: Map<String, JsonValue>,
 }
 
+pub struct EntityRegistrationMessageBuilder {
```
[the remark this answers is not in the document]
albinsuresh: Just to ease the creation of the `EntityRegistrationMessage` in tests, without all the optional fields, unless needed.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -184,14 +179,63 @@ impl EntityStore {
...
+                        };
+
+                        self.update_twin_data(EntityTwinMessage::new(
+                            source,
+                            fragment_key,
+                            fragment_value,
+                        ))?;
+                    }
+                    Channel::CommandMetadata { .. } => {
+                        // Do nothing for now as supported operations are not part of the entity store
```
albinsuresh, line 224: Not doing anything here as the supported operations are persisted separately on the file system. A different mechanism already exists to rebuild the entity store from those files. This arm will be enabled once we migrate all that into the entity store.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -184,14 +179,63 @@ impl EntityStore {
...
-        })
+            message_log,
+        };
+
+        entity_store.load_from_message_log()?;
+
+        Ok(entity_store)
+    }
+
+    pub fn load_from_message_log(&mut self) -> Result<(), InitError> {
```
didier-wenzek, line 200: Have you consider to persist not a log of registration messages but the entity store itself? Or more precisely its main component aka `entities: HashMap<EntityTopicId, EntityMetadata>`. I think this would be simpler.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -492,11 +542,11 @@ impl EntityStore {
     /// If the provided fragment already existed, `false` is returned.
     pub fn update_twin_data(
         &mut self,
-        entity_topic_id: &EntityTopicId,
-        fragment_key: String,
-        fragment_value: JsonValue,
+        twin_message: EntityTwinMessage,
     ) -> Result<bool, entity_store::Error> {
```
didier-wenzek, line 546: I failed to see the point of this `EntityTwinMessage` as they are built only to call the method `update_twin_data()`.

crates/core/tedge_api/src/message_log.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::io::BufWriter;
+use std::io::Write;
+use std::path::Path;
+
+const LOG_FILE_NAME: &str = "entity_store.jsonl";
+
+/// A persistent append-only log of MQTT messages.
+/// Each line is the JSON representation of that MQTT message.
+/// The underlying file is a JSON lines file.
+pub struct MqttMessageLog {
```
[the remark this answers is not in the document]
didier-wenzek: An alternative could be to store the whole `EntityStore` in a single file.

Appending messages in a log is more efficient, sure. However, in practice, there a few registration messages and most of them will arrive just after the device birthday. So atomically writing the whole each time there is a new registration will not be a performance issue. Reading such a persisted store will be simpler as it's no more required to replay the log.

That said, as the `EntityStore` contains an index (`entity_id_index`) which content duplicates the man `entities` map), this requires a  `struct PersistedEntityStore` from which the whole `EntityStore` can be rebuilt including its index.

didier-wenzek (commented):
I didn't try to understand why most of the system tests are failing.

The mappers would benefit from an `EntityStore` taking in charge all the caching issues. Can this be done by two independent implementations: one with auto registration another one with a cache of pending registration.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: My initial plan was to persist the `EntityMetadata` struct itself. But then switched to raw `MqttMessage`s to cover other metadata messages as well, like `CommandMetadata` message which is not part of the entity store at the moment. But then I thought it'd be even better if we decide to persist the telemetry messages (the early message cache) as well in the near future.

crates/common/mqtt_channel/src/messages.rs
```diff
@@ -2,20 +2,43 @@ use crate::errors::MqttError;
...
+use serde::Deserialize;
+use serde::Serialize;
 use std::fmt::Debug;
 use std::fmt::Formatter;
 use std::fmt::Write;
 
 /// A message to be sent to or received from MQTT.
-#[derive(Debug, Clone, Eq, PartialEq)]
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 pub struct Message {
```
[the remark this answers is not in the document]
albinsuresh: Yes, after dropping the plan to serialize `EntityMetadata` struct, I had taken that path as well, making `EntityRegistrationMessage` serializable and even introduced the new `EntityTwinMessage` struct to go along with it as explained in [this comment](https://github.com/thin-edge/thin-edge.io/pull/2466#discussion_r1421979018). But, switched to `MqttMessage` to keep our options open on what we decide to persist in future.

crates/core/tedge_api/src/message_log.rs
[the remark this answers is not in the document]
albinsuresh: Yes, I had taken quite a journey of options starting from persisting the `EntityMetadata` messages, to introducing a new `PersistentMessage` wrapper for all persistable messages to finally persisting raw `MqttMessage`s. The reasons and concerns that finally convinced me to stick with raw `MqttMessage` is described in [this comment](https://github.com/thin-edge/thin-edge.io/pull/2466#discussion_r1421973763). But, some of those points/concerns, like backward compatibility, are worth discussing.

crates/core/tedge_api/src/partial_entity_store.rs
```diff
@@ -0,0 +1,167 @@
...
+
+/// A store for all the entities for which data messages are received before
+/// its registration message itself is received.
+/// It also stores all the child device registration messages received before
+/// their parents themselves are registered, including their data.
+pub struct PartialEntityStore {
+    main_device_tid: EntityTopicId,
+    // This orphans map is keyed by the unregistered parent topic id to their children
+    orphans: HashMap<EntityTopicId, Vec<EntityTopicId>>,
+    entities: HashMap<EntityTopicId, PartialEntityCache>,
```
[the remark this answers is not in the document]
albinsuresh: > Having a cache per unregistered entity makes this partial entity store fragile when facing messages on random entities.

I didn't get that fragility concern.

One reason why I kept it per entity is to make the retrieval of all messages for a given entity easier. I could create a single buffer for all entities as well, despite the need to search the entire buffer to retrieve messages for each entity. But, since these buffers are gonna be small, that's not a big deal. But, this impl with map seemed much easier to handle.

crates/core/tedge_api/src/partial_entity_store.rs
```diff
@@ -0,0 +1,167 @@
...
+use std::collections::HashMap;
+
+const EARLY_MESSAGE_BUFFER_SIZE: usize = 5;
+
+/// A store for all the entities for which data messages are received before
+/// its registration message itself is received.
+/// It also stores all the child device registration messages received before
+/// their parents themselves are registered, including their data.
+pub struct PartialEntityStore {
+    main_device_tid: EntityTopicId,
```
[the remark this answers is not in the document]
albinsuresh: Good point. It was used to derive the default parent when reg messages without an explicit `@parent` info are received. But, a registration message without `@parent` implies an immediate child device and they never need to be cached, as their parent (the main device) is always registered.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -448,6 +448,11 @@ define_tedge_config! {
             /// Enable auto registration feature
             #[tedge_config(example = "true", default(value = true))]
             auto_register: bool,
+
+            /// The directory where the entity store is persisted across restarts
+            #[tedge_config(example = "/data/tedge", default(value = "/data/tedge"))]
+            #[doku(as = "PathBuf")]
+            path: Utf8PathBuf,
```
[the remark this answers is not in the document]
albinsuresh: > Beware that tedge-agent doesn't require this directory to exist and use /etc/tedge/.agent as a fallback. Do you plan something similar for the mapper?

Yes, a default is required. In fact all the test failures were due to that. 

> It might be better to consider a single tedge.db.path for all these use cases where things must be persisted across restarts.

Agree

crates/core/tedge_api/src/partial_entity_store.rs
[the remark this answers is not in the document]
didier-wenzek: >> Having a cache per unregistered entity makes this partial entity store fragile when facing messages on random entities.

> I didn't get that fragility concern.

As there are an unlimited number of buffers, the cache is actually unlimited even if each individual cache is limited.

> One reason why I kept it per entity is to make the retrieval of all messages for a given entity easier. I could create a single buffer for all entities as well, despite the need to search the entire buffer to retrieve messages for each entity. But, since these buffers are gonna be small, that's not a big deal. But, this impl with map seemed much easier to handle.

Having to go through a list of pending messages to filter out those that can be applied after a new registration should not be neither too complex nor inefficient as this cache if for less than one hundred of messages.

crates/core/tedge_api/src/partial_entity_store.rs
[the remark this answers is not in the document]
albinsuresh: Replaced `Partial` with `Pending`. Hope that's better.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: As discussed and agreed offline, we're sticking with the persistence of raw MqttMessages for now. But, this persistence impl won't be included in the current PR anymore. It will be added in a follow-up PR.

crates/core/tedge_api/src/partial_entity_store.rs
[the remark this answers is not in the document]
didier-wenzek: However, the module name is still `partial_entity_store`. It would be better to rename it too.

didier-wenzek (approved):
Approved. The code is clear and far simpler with pending message caching responsibility moved behind the entity store.

> As discussed and agreed offline, we're sticking with the persistence of raw MqttMessages for now. But, this persistence impl won't be included in the current PR anymore. It will be added in a follow-up PR.

Will have then to be addressed the pending comment related to storage.
Notably, we need to find an agreement on [`/data/tedge`](https://github.com/thin-edge/thin-edge.io/pull/2466#discussion_r1420889096)

didier-wenzek (approved):
I confirm my approval

albinsuresh:
# Test Plan

This PR fixes the message race issue where an out-of-order delivery of entity data messages delivered before the entity registrations themselves results in those data messages getting dropped. The fix results in the following behaviour:

1. All data messages: telemetry and other metadata messages like twin data, received before their respective registration messages, are cached in-memory and not converted.
    * Telemetry messages for all entities are cached in a cache with a capacity of only 100 entries. Hence, when the cache is full, older entries are replaced with newer entries.
    * Metadata messages are cached in unbounded buffers as we can't afford dropping such critical data
1. All child device registration messages received before their parents are also cached
1. When the registration message is received, itself and all its cached child devices are registered, and all their cached data are also processed.
1. Auto-registration must be turned off if explicit registration is used for any entity. Keeping it turned on while using explicit registration can sometimes result in undesired behaviours like nested child devices getting registered as immediate child devices.

gligorisaev:

Issue #2456, opened 2023-11-15
tedge-agent: support writing to any file when running as tedge user


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Currently the tedge-configuration-plugin runs as the root user to provide it the permissions required to read and write any file on the system it is running on. However, with the merging of the tedge-configuration-plugin into the tedge-agent (amongst other additional plugins), this presents a problem because the tedge-agent is running under the tedge user. This results in the loss of the ability to change any configuration on the file system.

The problem can be broken into two key functional requirements:

* read any file
* write any file

**Read any file -> Read any file with global read permissions**

I would propose to change the requirement from "Read any file" to "Read and which is permitted to by a non-root user".

Enable the ability to read any file presents a security risk of accidentally publishing data from protected files to the cloud. For example the tedge private certificate (default path: /etc/tedge/device-certs/tedge-private-key.pem), this file should NEVER be exposed (and it is currently configured to only be readable by the mosquitto user):

```
$ ls -l /etc/tedge/device-certs/tedge-private-key.pem
-r-------- 1 mosquitto mosquitto 246 Nov  9 13:50 /etc/tedge/device-certs/tedge-private-key.pem
```

In reality, files which do not contain sensitive/secret information are at least readable by all users (not writable, but readable). And trying to read files under the non-root user (`tedge`) is preferred because it avoids exposing information.

**Write any file**

Writing any file is an important feature that has to be retained as it is important to retain control over the operating system to administer it remotely, and mostly this requires writing to files which are owned or only writable by the root user (e.g. the Debian APT sources list: `/etc/apt/sources.list`).

Security around calling tedge-write can be managed by the `sudoers` mechanism which allows users to either allow or disallow whether the **tedge** user is permitted to call the binary tedge-write or not.

Note: It would be possible to overwrite protected files (such as the private certificate), however the existing file contents (with potentially sensitive information) would be overwritten with the new contents, thus still not exposing the contents (even though the old contents will be now lost, but at least not exposed).

**Describe the solution you'd like**

Create a custom executable called `tedge-write` (name is subject to change) (as part of the multi-call binary) which is used to write a specific file with elevated rights (if enabled under sudo).

**Usage: tedge-write**

```sh
tedge-write <SOURCE_FILE> <DESTINATION_FILE> [[--mode <mode_octal>] [--user <user>] [--group <group>]]
```

[...]

PR #2471, opened 2023-11-20
feat: add tedge-write as new multicall binary

Bravo555:
## Proposed changes

Adds `tedge-write` as a new multicall binary.

## Types of changes

## Paste Link to the issue
- #2456

## Checklist


crates/core/tedge_agent/src/agent.rs
```diff
@@ -82,7 +84,7 @@ impl AgentConfig {
         let tedge_config = config_repository.load()?;
 
         let config_dir = tedge_config_location.tedge_config_root_path.clone();
-        let tmp_dir = tedge_config.tmp.path.clone();
+        let tmp_dir = Arc::from(tedge_config.tmp.path.as_path());
```
jarhodes314, line 87: Do we need to use `.as_path` here, could we not just pass `&tedge_config.tmp.path`?

crates/core/tedge_write/tests/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    // Assert
+    assert_eq!(
+        fs::read_to_string(source_path).unwrap(),
+        fs::read_to_string(destination_path).unwrap(),
+    );
+}
+
+#[test]
+fn changes_file_permissions_if_file_doesnt_exist() {
```
jarhodes314, line 30: If the file doesn't exist, the permissions can't be changed as they don't exist, so it feels like this name needs tweaking.

crates/core/tedge_agent/src/agent.rs
Bravo555: Yes, `&tedge_config.tmp.path` resolves to `&Utf8PathBuf` instead of `&Utf8Path`

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -297,28 +299,90 @@ impl ConfigManagerActor {
...
+        // conditional compilation in tedge-write crate here doesn't work
+        #[cfg(test)]
+        {
+            std::fs::copy(&response.file_path, &file_entry.path)?;
+        }
+        #[cfg(not(test))]
+        {
+            let mode = file_entry.file_permissions.mode;
+            let user = file_entry.file_permissions.user.as_deref();
+            let group = file_entry.file_permissions.group.as_deref();
```
Bravo555, line 337: One very annoying thing: in unit tests, `tedge-write` binary is not present, so instead of spawning a new process, we can just copy the file as usual, but I'd like to do it in `tedge-write` instead of here.

I've tried using `#[cfg(test)]` for changing `copy` function to not spawn a process, but when unit testing, only the current crate is running on a `test` target, so it wouldn't work. Now I'm trying to set up a `testing` feature that only gets used inside tests, but then when doing `cargo check --all-features` it generates warning because the actual implementation is unused.

What would be the best way to provide a testing implementation that doesn't spawn a process?

crates/core/tedge_write/src/bin.rs
```diff
@@ -0,0 +1,160 @@
+// TODO: force `destination_path` to be the first argument in clap
```
didier-wenzek, line 1: What's the motivation for that?

- Is this to avoid security issues with bad sudo configuration?
- Or on the contrary, to give the `tedge-config-plugin` the freedom to use the appropriate user and mod for the files.

e.g:

```
`tedge    ALL = (ALL) NOPASSWD: /usr/bin/tedge-write --user mosquitto /etc/*`
```

crates/core/tedge_write/src/bin.rs
```diff
@@ -0,0 +1,160 @@
...
+        );
+    }
+
+    let file_existed_before_write = target_filepath.is_file();
+
+    write_stdin_to_file_atomic(&target_filepath)?;
+
+    if file_existed_before_write {
+        return Ok(());
+    }
```
didier-wenzek, line 68: That's a good idea to avoid changing file ownership in this case.
This extends the number of cases where `tedge-write` can be used without `sudo`.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -297,28 +299,90 @@ impl ConfigManagerActor {
...
+                );
+
+                request.failed(&error_message);
+                error!("{}", error_message);
+                self.publish_command_status(&topic, &ConfigOperation::Update(request))
+                    .await?;
+                return Ok(());
+            }
+
+            let tedge_write_output = tedge_write_process?;
```
didier-wenzek, line 361: This `?` on a result known to be `Ok` is a bit unusual - at least to me.

crates/extensions/tedge_config_manager/src/actor.rs
didier-wenzek: I would try to reduce the extent of this conditional. What is tested is too different of what is actually used.

Could the `CopyOptions` be built and used in both cases with a single difference in the constructor, a flag to either spawn a process or invoke a method (e.g. `tedge_write::bin::run()`).

didier-wenzek (commented):
Nicely implemented. I only have some questions.

You already addressed a key point (having `tedge_write` consuming its input from `stdin` - so the sudoer file can focus on which files can be overwritten).

crates/common/tedge_utils/src/fs.rs
```diff
@@ -10,6 +10,9 @@ use tokio::io::AsyncWriteExt;
 /// Write file to filesystem atomically using std::fs synchronously.
 pub fn atomically_write_file_sync(dest: impl AsRef<Path>, content: &[u8]) -> std::io::Result<()> {
     let dest_dir = parent_dir(dest.as_ref());
+    // FIXME: `.with_extension` replaces file extension, so if we used this
+    // function to write files `file.txt`, `file.bin`, `file.jpg`, etc.
+    // concurrently, then this will result in an error
     let tempfile = PathBuf::from(dest.as_ref()).with_extension("tmp");
```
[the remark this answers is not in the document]
Bravo555: I was unsure how much code relied on this behaviour, so I'd rather fix it in another PR.

crates/core/tedge_write/src/bin.rs
Bravo555: The main reason is to not grant root permissions when writing outside allowed directories. The set of rules one can express in `sudoers` is very limited, in order to successfully match on the path we write to (e.g. `/etc/*`) it has to be the first argument, as we match on the prefix.

didier-wenzek: Ok the issue is not really that options can be given to `tedge-write` before the destination path,
but that the __tedge-agent__ must always use `tedge-write` giving the destination path first so the `sudoers` file can use wild card for the allowed targets.

In other word, I don't see this TODO as urgent.

crates/extensions/tedge_config_manager/src/actor.rs
didier-wenzek: You pushed a fix along these lines while I was writing this comment. Slightly different, but working fine.

crates/core/tedge_write/src/bin.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut temp_file = fs::OpenOptions::new()
+        .write(true)
+        .create_new(true)
+        .open(temp_filepath.as_std_path())
+        .with_context(|| {
+            format!(
+                "Could not open temporary file `{}` for writing",
+                temp_filepath
+            )
+        })?;
```
jarhodes314, line 114: Minor:
```suggestion
        .with_context(|| format!("Could not open temporary file `{temp_filepath}` for writing"))?;
```

No idea off the top of my head whether it will format to be one line or three, but it would be consistent with all the other format strings here.

crates/core/tedge_write/tests/main.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    command.args(["--mode", "600"]);
+
+    // Act
+    command.assert().success();
+
+    // Assert
+    let dest_mode = destination_path.metadata().unwrap().permissions().mode();
+    // .mode() returns st_mode, we only need to compare a subset
+    assert_eq!(dest_mode & 0o600, 0o600);
```
jarhodes314, line 49: I'm not convinced the bitwise `&` is achieving what you're looking for here. As far as I understand, you're receiving more bits than just the three you care about. I believe that you're trying to test that the 3 least-significant octal digits are `600`, but a quick test in irust shows that's not the case:

```
In: format!("{:o}", 0o704 & 0o600)
Out: "600"
```

I think what you want is:

```suggestion
    assert_eq!(dest_mode & 0o777, 0o600);
```

crates/core/tedge_write/tests/main.rs
```diff
@@ -0,0 +1,84 @@
...
+    command.arg(&destination_path);
+
+    command.args(["--mode", "600"]);
+
+    // Act
+    command.assert().success();
+
+    // Assert
+    let dest_mode = destination_path.metadata().unwrap().permissions().mode();
+    assert_eq!(dest_mode & 0o644, 0o644);
```
jarhodes314, line 72:
```suggestion
    assert_eq!(dest_mode & 0o777, 0o644);
```

as above

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -305,32 +311,98 @@ impl ConfigManagerActor {
...
+        };
+
+        // new config was downloaded into tmpdir, we need to write it into destination using tedge-write
+        let from = Utf8Path::from_path(response.file_path.as_path()).unwrap();
+
+        let deployed_to_path = match self.deploy_config_file(from, &request.config_type) {
+            Ok(path) => path,
+            Err(err) => {
+                let error_message = format!(
+                    "Handling of operation failed with: Could not write to config file: {err}",
```
jarhodes314, line 342:
```suggestion
                    "config-manager failed writing updated configuration file: {err}",
```

(I am changing the one above to say `config-manager` instead of `tedge-configuration-plugin` as the latter is too specific in the case of `tedge-agent-v1`. That case is checked by a robot test, so I'm dealing with that in #2493, but it would be good to avoid adding new ones).

jarhodes314 (changes requested):
I think the tests aren't quite doing what you want them to (or I've misunderstood them), so requesting changes

jarhodes314 (approved):

crates/core/tedge_write/tests/main.rs
Bravo555: You're completely right, that's a very dumb mistake.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -387,6 +352,48 @@ impl ConfigManagerActor {
...
+    ) -> Result<Utf8PathBuf, ConfigManagementError> {
+        let file_entry = self.plugin_config.get_file_entry_from_type(config_type)?;
+
+        let mode = file_entry.file_permissions.mode;
+        let user = file_entry.file_permissions.user.as_deref();
+        let group = file_entry.file_permissions.group.as_deref();
+
+        let to = Utf8PathBuf::from(&file_entry.path);
+
+        match self.config.use_tedge_write {
```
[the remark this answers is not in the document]
reubenmiller: I guess I'm worried that supporting two ways of doing the same thing will be counter productive and confusing. If we expose a new setting which can be used to control whether tedge-write is used or not, then it also means we need to document it. And generally we won't be running the tedge-agent as root in any of out default setups, so I'm not sure what the value is for maintaining two ways of doing the same thing (though I know using the native way is much easier, but we can't avoid the tedge-write binary anyway, so better to use it in both cases imho).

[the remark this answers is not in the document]
reubenmiller: For maybe I'm mixing up topics...if sudo is not installed then sudo should not be used (of course ;))

PR #2457, opened 2023-11-15
feat! switch to tedge-agent (consolidated service)

reubenmiller:
## Proposed changes

Package the new consolidated tedge-agent binary:

* use combined tedge-agent functionaity by default
* remove tedge-log-plugin and tedge-configuration-plugin
* update system tests

**Blocked by**

* [x] Log manager change remove root user requirements - https://github.com/thin-edge/thin-edge.io/issues/2443
* [x] Allow writing to any configuration file - https://github.com/thin-edge/thin-edge.io/issues/2456

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2433

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2474, opened 2023-11-21
Add HTTPS/certificate authentication support to file transfer service

jarhodes314:
## Proposed changes
This adds support for HTTPS and certificate authentication to the file transfer service.

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2475

## Checklist


crates/core/tedge_agent/src/agent.rs
```diff
@@ -97,10 +103,19 @@ impl AgentConfig {
...
-            .with_data_dir(data_dir.clone())
-            .with_port(http_port)
-            .with_ip_address(http_bind_address);
+        let rustls_config = if let Some((cert, key)) = load_certificate_and_key(&tedge_config)? {
+            let ca_path = tedge_config.http.ca_path.or_none();
+            let root_cert_store = ca_path.map(|s| axum_tls::read_trust_store(s)).transpose()?;
+            Some(axum_tls::ssl_config(cert, key, root_cert_store)?)
+        } else {
+            None
+        };
```
Bravo555, line 112: suggestion: I feel like that's not quite the right place to put loading and parsing of certificates and cryptographic keys.

IMO the main reason why we e.g. don't use `TedgeConfig` directly inside the actors, is because then to create actors we'd have to prepare the entire `TedgeConfig` with many config settings that do not affect the actor we're trying to start. It would make the actor harder to use/test in isolation, outside of the context of its primary user (e.g. File Transfer Service HTTP server inside tedge-agent).

I think loading and parsing certificates and private keys in this `TedgeConfig -> ActorConfig` transformation also makes the actor harder to use, because then to create `ActorConfig` directly, you have to manually produce `rustls::ServerConfig`. It wouldn't be that much code, thanks to the helper functions you made, but still, I'd argue that it's a bit wrong level of abstraction to do that here.

Instead, we should pass paths themselves to `FileTransferServerBuilder` itself.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
```diff
@@ -76,91 +92,250 @@ impl Builder<FileTransferServerActor> for FileTransferServerBuilder {
...
+        Ok(())
+    }
+
+    struct Server<Cert> {
+        port: u16,
+        temp_dir: TempTedgeDir,
+        server_cert: Cert,
+    }
+
+    impl Server<()> {
```
Bravo555, line 208: question: Why do we define a separate server for testing reasons instead of testing the actual one? I think adding a comment would make it easier to understand for potential readers.

crates/core/tedge_agent/src/file_transfer_server/error.rs
```diff
@@ -71,7 +71,7 @@ impl IntoResponse for FileTransferRequestError {
                 tracing::error!("{error_message}");
                 err.into_response()
             }
-            FromIo(_) | OtherDelete { .. } | OtherUpload { .. } => {
+            FromIo(_) | Delete { .. } | Upload { .. } => {
```
Bravo555, line 74: note: Also not technically part of this PR, but [`use Enum::*` and matching variant names without `::` can be potentially a footgun if variants are removed](https://youtu.be/8j_FbjiowvE?si=VkIG0HQnCQwEd036&t=98).
Here I think it isn't that much of a problem, because all the variants have attached data, and if any binding here were to match "everything else" then matching individual fields would result in an compiler error, but still, I think doing `use FileTransferRequestError as E` and then matching on `E::...` would be preferable for new code.

crates/core/tedge_agent/src/file_transfer_server/request_files.rs
```diff
@@ -9,7 +9,15 @@ use camino::Utf8Path;
...
 use super::error::FileTransferRequestError;
-use super::http_rest::HttpConfig;
+
+#[derive(Clone)]
+pub(super) struct FileTransferDir(Arc<Utf8Path>);
+
+impl FileTransferDir {
+    pub(super) fn new(file_transfer_dir: Utf8PathBuf) -> Self {
+        Self(Arc::from(file_transfer_dir))
+    }
```
Bravo555, line 19: Good idea to decouple file transfer directories from the rest of the config there.

Bravo555 (commented):
I have only one request - adding some documentation to the new `axum_tls` crate. I think it having doc comments for our crates especially (and sometimes for some complex modules) significantly helps with understanding and why the code is the way it is, and in this case, I think even a short summary would be helpful.

crates/common/axum_tls/src/lib.rs
Bravo555, line 1: suggestion(blocking): I'd like to see some documentation describing the purpose of this crate.

crates/core/tedge_agent/src/file_transfer_server/request_files.rs
jarhodes314: Yeah, as I added more certificate configuration, it became really obvious that the bulk of the tests (which use local, not network-based HTTP testing) needed to provide a bunch of information that the application just didn't care about.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -170,7 +206,7 @@ impl Agent {
     }
 
     #[instrument(skip(self), name = "sm-agent")]
-    pub async fn start(&mut self, v1: bool) -> Result<(), anyhow::Error> {
+    pub async fn start(self, v1: bool) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
jarhodes314: I agree and spent a while trying to understand myself what `true` and `false` mean here. I'm happy to change this, but I don't have the best context on exactly what this means. @albinsuresh @reubenmiller can either of you suggest descriptive names for the variants of this parameter to avoid this confusion? Equally I understand if we decide "this is a feature flag that is destined to be removed before the 1.0 release so we'll just live with it for now" (but the change is likely completely trivial to make if good names spring to mind).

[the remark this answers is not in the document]
albinsuresh: Yeah, it's just a temporary flag that we introduced recently, which would/should be removed with #2457 . So, just ignore that for now.

docs/src/references/tedge-file-transfer-service.md
```diff
@@ -32,3 +32,16 @@ Unique paths must be used in the URL path to avoid such overwrites.
...
+By default, the service is unauthenticated and does not support HTTPS connections.
+HTTPS can be enabled by setting `http.cert_path` and `http.key_path`.
+If the certificates are configured, the agent will automatically host the service via HTTPS, and redirect any
+HTTP requests to the equivalent HTTPS URL.
+If HTTPS is enabled, the configured certificate should be installed in the OS trust store for any connected agents
+in order for them to trust the connection to the mapper.
+
+Once HTTPS is enabled for the file-transfer service, certificate-based authentication can also be enabled.
+The directory containing the certificates that the mapper will trust can be configured using `http.ca_path`,
+and the agent can be configured to use a trusted certificate using the `http.client.auth.cert_file` and `http.client.auth.key_file`
```
albinsuresh, line 46:
```suggestion
and the mapper can be configured to use a trusted certificate using the `http.client.auth.cert_file` and `http.client.auth.key_file`
```

docs/src/references/tedge-file-transfer-service.md
```diff
@@ -32,3 +32,16 @@ Unique paths must be used in the URL path to avoid such overwrites.
 All uploaded files are preserved until they are explicitly deleted with the DELETE API.
 To avoid exhaustion of storage space on the thin-edge device,
 users must be diligent to delete any stored files as soon as their purpose is served.
+
+## HTTPS and authenticated access
+By default, the service is unauthenticated and does not support HTTPS connections.
```
albinsuresh, line 37:
```suggestion
## HTTPS and authenticated access

By default, the service is unauthenticated and does not support HTTPS connections.
```

docs/src/references/tedge-file-transfer-service.md
```diff
@@ -32,3 +32,16 @@ Unique paths must be used in the URL path to avoid such overwrites.
...
+HTTPS can be enabled by setting `http.cert_path` and `http.key_path`.
+If the certificates are configured, the agent will automatically host the service via HTTPS, and redirect any
+HTTP requests to the equivalent HTTPS URL.
+If HTTPS is enabled, the configured certificate should be installed in the OS trust store for any connected agents
+in order for them to trust the connection to the mapper.
+
+Once HTTPS is enabled for the file-transfer service, certificate-based authentication can also be enabled.
+The directory containing the certificates that the mapper will trust can be configured using `http.ca_path`,
```
albinsuresh, line 45:
```suggestion
The directory containing the certificates that the agent will trust can be configured using `http.ca_path`,
```

crates/core/tedge_agent/src/agent.rs
```diff
@@ -46,13 +53,13 @@ use tracing::instrument;
 const TEDGE_AGENT: &str = "tedge-agent";
 
 #[derive(Debug, Clone)]
-pub struct AgentConfig {
+pub(crate) struct AgentConfig {
     pub mqtt_config: MqttConfig,
     pub http_config: HttpConfig,
+    pub http_socket_addr: SocketAddr,
```
albinsuresh, line 59: Why not keep this value also in the `HttpConfig` struct?

crates/core/tedge_agent/src/agent.rs
albinsuresh: Agree. A function creating a config object should be as light-weight as possible.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
jarhodes314: "Server" is maybe not quite the right term here. I wanted some sort of abstraction to allow the tests to talk in terms of only the things that are relevant here, otherwise it's essentially impossible to see the subtle details around whether HTTPS is enabled, whether we expect client certs, whether we provide client certs in the request etc...

docs/src/references/tedge-file-transfer-service.md
jarhodes314: I don't think that's true, but the "mapper" in the sentence above is incorrect (as you can probably tell, I copied and pasted it from the auth proxy docs since they're broadly the same, but forgot to change that bit).

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -622,8 +622,24 @@ define_tedge_config! {
...
                 key_file: Utf8PathBuf,
-            }
-        }
+            },
+        },
+
+        /// The file that will be used as a the server certificate for the File Transfer Service
+        #[tedge_config(example = "/etc/tedge/device-certs/file_transfer_certificate.pem")]
+        #[doku(as = "PathBuf")]
+        cert_path: Utf8PathBuf,
```
didier-wenzek, line 631: I will take the time to write scripts and docs on how to setup all these certificates.

I suspect that the current set of settings could be made simpler. For instance, in practice, the `tedge-agent` running on the main device can use the same certificate to connect the c8y auth proxy and to serve files to child devices. Also, on a child device, a user might wrongly set `http.cert_path` instead of `http.client.auth.cert_file`.
It would be good to have a certificate per service instance and to have config settings that reflect that.
Writing scripts and docs will help me to clarify this.

docs/src/references/tedge-file-transfer-service.md
didier-wenzek, line 46:
```suggestion
and the mapper as well as the child device agents have to be configured to use a trusted certificate using the `http.client.auth.cert_file` and `http.client.auth.key_file`
```

crates/core/tedge_actors/src/errors.rs
```diff
@@ -39,6 +39,9 @@ pub enum RuntimeError {
 
     #[error(transparent)]
     LinkError(#[from] LinkError),
+
+    #[error("Unknown error: {0:?}")]
+    Other(#[from] anyhow::Error),
```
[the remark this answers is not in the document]
jarhodes314: I've just deleted it for now as it wasn't necessary.

crates/core/tedge_agent/src/agent.rs
jarhodes314: In short, `HttpConfig` is used in a separate place to the `SocketAddr`, but I've figured out how I can get them to be used in a single place so that the tests aren't supplying information that isn't used.

docs/src/references/tedge-file-transfer-service.md
albinsuresh: Huh? There seems to be some misunderstanding about this setting then, as I see @didier-wenzek also suggesting a similar change. This is the **`client`** setting, right? Meant to be used by the clients of the agent: which are the mappers and other child devices.

jarhodes314: I've adjusted it to be in line with Didier's comment, as well as correcting the line above, which was also adding to my confusion as that was also wrong.

crates/common/axum_tls/src/lib.rs
jarhodes314: I've now attempted to do this

crates/core/tedge_agent/src/file_transfer_server/actor.rs
jarhodes314: I don't think this is particularly easy/valuable to generalise to other services, so I've instead opted for renaming this to `TestFileTransferService` with a short doc comment justifying its existence. I don't love having lots of ad-hoc utility code to simplify tests, but in this case, without this code it's very difficult to identify what each test is doing differently to the next.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
didier-wenzek: Here is first raw version for a script (where all names are hard-coded!):  https://gist.github.com/didier-wenzek/110b618eda6866814dc02fe4f168ff2d

albinsuresh: > It would be good to have a certificate per service instance and to have config settings that reflect that.

I just hope per-service certificates won't make the setup too complicated for clients. Ideally, a child device should be able to connect to both the `c8y-proxy` and `FTS` using the same client certificate. At least for our docs/demos, having the same root CA sign all the service certificates as well as the client certificates should simplify things I guess.

> Writing scripts and docs will help me to clarify this.

That'd really help. With the agent being a client of the mapper for the proxy and the reverse for the file transfer service, I found the client cert setting twins of both the services really confusing to get my head around and made a lot of errors.

> Here is first raw version for a script (where all names are hard-coded!): https://gist.github.com/didier-wenzek/110b618eda6866814dc02fe4f168ff2d

That really helped and that long list of config settings highlights the complexity as well. It all makes sense, but as a regular user I find it extremely cumbersome.

didier-wenzek:
Something is wrong. Could my certificate settings be the issue?

Triggering a log upload from the cloud, the final outcome is a 404 error: `"Upload failed with Request(reqwest::Error { kind: Status(404), url: Url { scheme: https, cannot_be_a_base: false, username: \", password: None, host: Some(Domain(localhost)), port: Some(8000), path: /tedge/file-transfer/demo-device-777/log_upload/mosquitto-c8y-mapper-2023-11-24T16:45:19.758911928Z, query: None, fragment: None } })`.

The error message could be clearer, but this is not my point. The log upload command is executed as expected up to the point where the c8y mapper has to GET the file from file transfer, leading to a 404 and indeed the file is not here: the log plugin (launched using the all-in-one tedge-agent-v1) pretends that that file have been successfully uploaded - but there is no file nor upload errors.

jarhodes314:
> Something is wrong. Could my certificate settings be the issue?
> 
> Triggering a log upload from the cloud, the final outcome is a 404 error: `"Upload failed with Request(reqwest::Error { kind: Status(404), url: Url { scheme: https, cannot_be_a_base: false, username: \", password: None, host: Some(Domain(localhost)), port: Some(8000), path: /tedge/file-transfer/demo-device-777/log_upload/mosquitto-c8y-mapper-2023-11-24T16:45:19.758911928Z, query: None, fragment: None } })`.
> 
> The error message could be clearer, but this is not my point. The log upload command is executed as expected up to the point where the c8y mapper has to GET the file from file transfer, leading to a 404 and indeed the file is not here: the log plugin (launched using the all-in-one tedge-agent-v1) pretends that that file have been successfully uploaded - but there is no file nor upload errors.

I haven't verified this, but I would imagine the issue is that the agent is not reuploading the file when it receives the 307 redirect response to the HTTPS URL after initially connecting via HTTP. I'll have a look into it on Monday.

Bravo555 (approved):

jarhodes314:
> > Something is wrong. Could my certificate settings be the issue?
> > Triggering a log upload from the cloud, the final outcome is a 404 error: `"Upload failed with Request(reqwest::Error { kind: Status(404), url: Url { scheme: https, cannot_be_a_base: false, username: \", password: None, host: Some(Domain(localhost)), port: Some(8000), path: /tedge/file-transfer/demo-device-777/log_upload/mosquitto-c8y-mapper-2023-11-24T16:45:19.758911928Z, query: None, fragment: None } })`.
> > The error message could be clearer, but this is not my point. The log upload command is executed as expected up to the point where the c8y mapper has to GET the file from file transfer, leading to a 404 and indeed the file is not here: the log plugin (launched using the all-in-one tedge-agent-v1) pretends that that file have been successfully uploaded - but there is no file nor upload errors.
> 
> I haven't verified this, but I would imagine the issue is that the agent is not reuploading the file when it receives the 307 redirect response to the HTTPS URL after initially connecting via HTTP. I'll have a look into it on Monday.

So the thing that's happening here is the PUT request to upload the file cannot be retried, so the response to that request is a 307. As far as reqwest is concerned, this isn't an error, the request has succeeded since it's not a 4xx or 5xx status code. That explains why (a) the upload failed and (b) why the log plugin ignored the issue.

I plan to solve this by making a HEAD request to the URL we intend to upload to, and using the final URL from that request as the target to upload to. This is a bit of a hack at the moment since we don't actually implement the HEAD method in the file transfer service, but it's sufficient to generate the desired 307 response in the case where we are using HTTPS. @didier-wenzek does this sound like a reasonable approach to you? I've implemented it locally, and I'm just refining the error handling so we get clear error messages from the log plugin in the case of a failure.

didier-wenzek:
> > > Something is wrong. Could my certificate settings be the issue?
> > > Triggering a log upload from the cloud, the final outcome is a 404 error: `"Upload failed with Request(reqwest::Error { kind: Status(404), url: Url { scheme: https, cannot_be_a_base: false, username: \", password: None, host: Some(Domain(localhost)), port: Some(8000), path: /tedge/file-transfer/demo-device-777/log_upload/mosquitto-c8y-mapper-2023-11-24T16:45:19.758911928Z, query: None, fragment: None } })`.
> > > The error message could be clearer, but this is not my point. The log upload command is executed as expected up to the point where the c8y mapper has to GET the file from file transfer, leading to a 404 and indeed the file is not here: the log plugin (launched using the all-in-one tedge-agent-v1) pretends that that file have been successfully uploaded - but there is no file nor upload errors.
> > 
> > 
> > I haven't verified this, but I would imagine the issue is that the agent is not reuploading the file when it receives the 307 redirect response to the HTTPS URL after initially connecting via HTTP. I'll have a look into it on Monday.
> 
> So the thing that's happening here is the PUT request to upload the file cannot be retried, so the response to that request is a 307. As far as reqwest is concerned, this isn't an error, the request has succeeded since it's not a 4xx or 5xx status code. That explains why (a) the upload failed and (b) why the log plugin ignored the issue.
> 
> I plan to solve this by making a HEAD request to the URL we intend to upload to, and using the final URL from that request as the target to upload to. This is a bit of a hack at the moment since we don't actually implement the HEAD method in the file transfer service, but it's sufficient to generate the desired 307 response in the case where we are using HTTPS. @didier-wenzek does this sound like a reasonable approach to you? I've implemented it locally, and I'm just refining the error handling so we get clear error messages from the log plugin in the case of a failure.

I'm okay to send a first HEAD request, even if not usual nor ideal. Could the client issue directly an HTTPS request, guessing the scheme from the config?

jarhodes314:
> > > > Something is wrong. Could my certificate settings be the issue?
> > > > Triggering a log upload from the cloud, the final outcome is a 404 error: `"Upload failed with Request(reqwest::Error { kind: Status(404), url: Url { scheme: https, cannot_be_a_base: false, username: \", password: None, host: Some(Domain(localhost)), port: Some(8000), path: /tedge/file-transfer/demo-device-777/log_upload/mosquitto-c8y-mapper-2023-11-24T16:45:19.758911928Z, query: None, fragment: None } })`.
> > > > The error message could be clearer, but this is not my point. The log upload command is executed as expected up to the point where the c8y mapper has to GET the file from file transfer, leading to a 404 and indeed the file is not here: the log plugin (launched using the all-in-one tedge-agent-v1) pretends that that file have been successfully uploaded - but there is no file nor upload errors.
> > > 
> > > 
> > > I haven't verified this, but I would imagine the issue is that the agent is not reuploading the file when it receives the 307 redirect response to the HTTPS URL after initially connecting via HTTP. I'll have a look into it on Monday.
> > 
> > 
> > So the thing that's happening here is the PUT request to upload the file cannot be retried, so the response to that request is a 307. As far as reqwest is concerned, this isn't an error, the request has succeeded since it's not a 4xx or 5xx status code. That explains why (a) the upload failed and (b) why the log plugin ignored the issue.
> > I plan to solve this by making a HEAD request to the URL we intend to upload to, and using the final URL from that request as the target to upload to. This is a bit of a hack at the moment since we don't actually implement the HEAD method in the file transfer service, but it's sufficient to generate the desired 307 response in the case where we are using HTTPS. @didier-wenzek does this sound like a reasonable approach to you? I've implemented it locally, and I'm just refining the error handling so we get clear error messages from the log plugin in the case of a failure.
> 
> I'm okay to send a first HEAD request, even if not usual nor ideal. Could the client issue directly an HTTPS request, guessing the scheme from the config?

It could issue an HTTPS request directly, but...

- in manual testing of the feature, I found the HEAD request necessary to produce a clear error when certificates are required and the client is not providing them
- if client certificates are not being used, we can only infer the values from the configuration on the main device, a child device shouldn't be configured with `http.cert_path` and `http.key_path`, so we would only save effort in certain cases
- the code changes required in the uploader to support this are agnostic to the redirect, they simply perform the HEAD request and use the final URL of the request for the subsequent PUT request. If we were to possibly attempt the request using HTTPS first, this would add a bit of complexity (though this is a minor point and wouldn't dissuade me per se)

I think we're already on the verge of having too many configurations for this feature, I think making this "just work" at the cost of an additional HTTP request is worthwhile compared to adding another configuration.

albinsuresh (commented):
LGTM.

The unit test coverage is definitely good, but shouldn't we have some integration tests as well, which tests a secure setup end-to-end? May be just one test that validates secure access to the file-transfer service with `curl` or something.

But, to really make sure that our own components can work together when security is enabled, the following tests would also be required:

1. `tedge-log-plugin` on child device connecting to secure tedge
2. `tedge-config-plugin` on child device connecting to secure tedge
3. `tedge-agent` on child device connecting to secure tedge

And the variations of the above cases as well: authenticated client and unauthenticated client. These tests can later be used as a reference for our docs on how to configure a secure deployment.

Since adding all those tests would be a fairly big task, it can be done in follow-up tasks and someone else should be able to support you with those tests.

crates/extensions/tedge_uploader_ext/src/actor.rs
```diff
@@ -62,23 +63,28 @@ impl UploadResponse {
...
+#[derive(Debug)]
 pub struct UploaderActor {
     config: ServerConfig,
+    identity: Option<Identity>,
 }
 
 impl UploaderActor {
-    pub fn new() -> Self {
-        UploaderActor::default()
+    pub fn new(identity: Option<Identity>) -> Self {
```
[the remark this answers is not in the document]
jarhodes314: That works fine for "real" optional values, like the back off parameters, where someone may or may not want to configure it. But as far as the code is concerned, this identity is required and comes directly from `tedge_config` (the value happens to be an `Option<Identity>`). If we made this a builder, it would lead to bugs (because people would forget to supply their uploader with an identity) and cause needless pain destructuring the Option value to `Some(identity)` on the caller's side (so they could call the method). This adds a little bit of noise to tests, but `None` isn't a lot of noise (particularly when we consider a test for this has to spin up a server).

crates/extensions/c8y_config_manager/src/plugin_config.rs
```diff
@@ -193,7 +193,9 @@ impl PluginConfig {
 }
 
 #[derive(thiserror::Error, Debug)]
-#[error("The requested config_type {config_type} is not defined in the plugin configuration file.")]
+#[error(
+    "The requested config_type {config_type:?} is not defined in the plugin configuration file."
```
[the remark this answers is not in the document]
jarhodes314: It's nice to delimit values to make the value clear, and debug printing is the safest, and easiest, way to do this (for `String`s). It's a value that's coming from a user, so it's generally a good idea IMO to avoid any log injection risks (even if those are low).

crates/extensions/c8y_mapper_ext/src/config_operations.rs
```diff
@@ -214,8 +214,9 @@ impl CumulocityConverter {
         let download = match download_result {
             Err(err) => {
                 let smartrest_error = SmartRestSetOperationToFailed::new(
-                    CumulocitySupportedOperations::C8yLogFileRequest,
-                    format!("Upload failed with {:?}", err),
+                    // TODO test I'm the right message type
```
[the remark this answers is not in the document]
jarhodes314: I was surprised to see this was how we handle failed operations. We must already know this value is order to execute the operation, so all this function should need to provide is the failure reason, it shouldn't care what the operation is called. This is a particularly critical thing to handle correctly due to Cumulocity's assumption that devices always complete operations, and reproducing this failure is pretty difficult in order to test it (and expensive to test since it requires an integration test if you want more of a guarantee than "hopefully the test author wrote the correct value in their test").

crates/common/log_manager/src/error.rs
```diff
@@ -20,6 +20,6 @@ pub enum LogRetrievalError {
     #[error("Log file has maximum number of lines.")]
     MaxLines,
 
-    #[error("No such file or directory for log type: {log_type}")]
+    #[error("No logs found with type {log_type:?}")]
```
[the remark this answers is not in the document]
jarhodes314: I've changed this to your suggestion now, without the colon since the quotes provide delimiting. At the moment, the code doesn't handle these quotes correctly when sending SmartREST messages, but I'll address that separately since fixing that breaks a lot of tests that rely on needlessly quoted values in CSV files.

crates/extensions/c8y_mapper_ext/src/config_operations.rs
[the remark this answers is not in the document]
albinsuresh: > We must already know this value is order to execute the operation, so all this function should need to provide is the failure reason, it shouldn't care what the operation is called.

It's partly due to the fact that the operation status transitions in SmartREST is [generic](https://cumulocity.com/guides/reference/smartrest-two/#501) and requires the operation type to be sent in that message. That same design was adopted here as well, in generic structures like `SmartRestSetOperationToFailed` which leads to such accidental errors. But, I agree that some sort of "current-operation-aware" warpper should have been used here to avoid such mistakes.

crates/common/log_manager/src/error.rs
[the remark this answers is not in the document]
albinsuresh: > At the moment, the code doesn't handle these quotes correctly when sending SmartREST messages, but I'll address that separately since fixing that breaks a lot of tests that rely on needlessly quoted values in CSV files.

Yeah, blindly quoting all SmartREST values, even for the ones that don't need quoting, has always been a pain while writing tests. I'd be thanking you, if you can fix that.

tests/RobotFramework/tests/cumulocity/log/log_operation.robot
```diff
@@ -30,7 +30,7 @@ Request with non-existing log type
     ...    fragments={"c8y_LogfileRequest":{"dateFrom":"${start_timestamp}","dateTo":"${end_timestamp}","logFile":"example1","searchText":"first","maximumLines":10}}
     Operation Should Be FAILED
     ...    ${operation}
-    ...    failure_reason=.*No such file or directory for log type: example1
+    ...    failure_reason=.*No logs found for log type example1
```
jarhodes314, line 33: I've created (and will likely work on soon if no one objects) https://github.com/thin-edge/thin-edge.io/issues/2487 for this

albinsuresh: No objections. But, in the meantime you can also use single quotes or back quotes to get around that problem.

didier-wenzek:
> I think we're already on the verge of having too many configurations for this feature, I think making this "just work" at the cost of an additional HTTP request is worthwhile compared to adding another configuration.

I confirm that things just work.
- Certificate-based authentication of c8y-auth-proxy, file-transfer and their clients
- Software download and install
- Log upload (using tedge-agent-v1 that includes log upload + TLS support).
- Configuration file upload and download  (using tedge-agent-v1 that includes config management + TLS support).

didier-wenzek (approved):
Approved. Thank you for this quite involved and really useful PR.

jarhodes314:
@albinsuresh @didier-wenzek I've added some integration tests to the configuration plugin. There is currently one case where I believe I have found a genuine bug, where a configuration update is marked as successful in cloud but the file isn't actually updated on disk. This only appears to happen if a child device is configured to use client certificates, but the mapper is not configured to use client certificates. Taking a configuration fails as expected with a helpful error in this case, so it should be obvious that something is wrong.

I think this is a sufficiently niche error in a niche feature that it doesn't justify blocking the merging of this PR. I'll start to investigate the root cause now, but unless anyone objects, I will merge this PR when the tests are passing and @albinsuresh has approved.

albinsuresh (approved):
Great work. The integration test also seems pretty complete. The rest of the variations can be added later on (even by someone else).

Issue #2478, opened 2023-11-22
Finalize Operation Workflow


didier-wenzek:
## For 1.0:
- [x] Use script exit status when there is no JSON output
      See comments: https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1268024883
      and https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1394172319
      and https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1401621879
- [x] Find a better to define a restart action
      See https://github.com/thin-edge/thin-edge.io/pull/2479#discussion_r1405693285
      and https://github.com/thin-edge/thin-edge.io/pull/2479#discussion_r1405695594
      and https://github.com/thin-edge/thin-edge.io/pull/2479#discussion_r1411677398
- [x] Log workflow execution 
- [x] Set timeouts on operation steps
- [x] Trigger a device restart (which is triggered by the workflow script itself)
- [x] Trigger an agent coordinated device restart
- [x] Clarify the purpose of the *owner* setting
      See comments: https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1268026377
      and https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1268207295
- [x] Update the example used in the doc
      See comments: https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1401610417
- [x] Clarify the concurrency/ sequential model

## For 1.0, if useful to support firmware management:
- [x] Clarify how a json excerpt is extracted from script output
      See https://github.com/thin-edge/thin-edge.io/pull/2479#discussion_r1411695470
- [ ] Download action making a cloud file locally available
      See comments: https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1392990394
- [ ] Consider to add the source state causing an error
      See comments: https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1267972097
 
## For later:
- [ ] Consider to combine and override several workflow definitions
      See comments: https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1268016020
      and https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1268042933
- [ ] Consider a way to publish events/alarms when reaching a state
      See comments: https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1266393389
- [ ] Ensure there is one and only one workflow responsible for moving forward a command in a given state
[...]

PR #2479, opened 2023-11-23
Add support for operation workflow triggering device restart

didier-wenzek:
## Proposed changes

The following trigger a device restart, leading to "successful_restart" or "failed_restart" depending on the actual success of the reboot. 

```
[scheduled]
script = "restart"
next = ["restarting", "successful_restart", "failed_restart"]
```

The `next` list of states is used to configure the states to which the workflow move on restart:
- The first state ("restarting" in the example) is the executing state of the command: till the device restart the workflow will stay in that state.
- The second state ("successful_restart" in the example) is where the workflow is moved after a successful restart.
- The third state ("failed_restart" in the example) is where the workflow is moved after a failed restart.

**Note** This file format has to be discussed:
- An alternative could be to have a specific `action` property instead of overloading the `script` property
- For example: `action = "restart --executing restarting --on-success successful_restart --on-error failed_restart"`

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2478

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/software_manager/actor.rs
+++ b/crates/core/tedge_agent/src/software_manager/actor.rs
@@ -125,10 +125,8 @@ impl SoftwareManagerActor {
         config: SoftwareManagerConfig,
         message_box: SimpleMessageBox<SoftwareCommand, SoftwareCommand>,
     ) -> Self {
-        let state_repository = AgentStateRepository::new_with_file_name(
-            config.config_dir.clone(),
-            "software-current-operation",
-        );
+        let state_repository =
+            AgentStateRepository::new(config.config_dir.clone(), "software-current-operation");
 
         let (output_sender, input_receiver) = message_box.into_split();
 
--- a/crates/core/tedge_agent/src/software_manager/actor.rs
+++ b/crates/core/tedge_agent/src/software_manager/actor.rs
@@ -1,15 +1,15 @@
 use crate::software_manager::config::SoftwareManagerConfig;
 use crate::software_manager::error::SoftwareManagerError;
 use crate::software_manager::error::SoftwareManagerError::NoPlugins;
+use crate::state_repository::error::StateError;
 use crate::state_repository::state::AgentStateRepository;
-use crate::state_repository::state::SoftwareOperationVariants;
-use crate::state_repository::state::State;
-use crate::state_repository::state::StateStatus;
 use async_trait::async_trait;
 use plugin_sm::operation_logs::LogKind;
 use plugin_sm::operation_logs::OperationLogs;
 use plugin_sm::plugin_manager::ExternalPlugins;
 use plugin_sm::plugin_manager::Plugins;
+use serde::Deserialize;
+use serde::Serialize;
 use std::path::PathBuf;
 use tedge_actors::fan_in_message_type;
 use tedge_actors::Actor;
@@ -35,7 +35,7 @@ const SUDO: &str = "sudo";
 #[cfg(test)]
 const SUDO: &str = "echo";
 
-fan_in_message_type!(SoftwareCommand[SoftwareUpdateCommand, SoftwareListCommand] : Debug, Eq, PartialEq);
+fan_in_message_type!(SoftwareCommand[SoftwareUpdateCommand, SoftwareListCommand] : Debug, Eq, PartialEq, Deserialize, Serialize);
 
 /// Actor which performs software operations.
 ///
@@ -49,7 +49,7 @@ fan_in_message_type!(SoftwareCommand[SoftwareUpdateCommand, SoftwareListCommand]
 /// operation.
 pub struct SoftwareManagerActor {
     config: SoftwareManagerConfig,
-    state_repository: AgentStateRepository,
+    state_repository: AgentStateRepository<SoftwareCommand>,
 
     // the Option is necessary to be able to concurrently handle a request,
     // which mutably borrows the sender, and listen on signals, which mutably
@@ -165,34 +165,30 @@ impl SoftwareManagerActor {
     }
 
     async fn process_pending_sm_operation(&mut self) -> Result<(), SoftwareManagerError> {
-        let state: Result<State, _> = self.state_repository.load().await;
-
-        if let Ok(State {
-            operation_id: Some(cmd_id),
-            operation: Some(operation),
-        }) = state
-        {
-            match operation {
-                StateStatus::Software(SoftwareOperationVariants::Update) => {
-                    let response = SoftwareUpdateCommand::new(&self.config.device, cmd_id)
-                        .with_error(
-                            "Software Update command cancelled on agent restart".to_string(),
-                        );
-                    self.output_sender.send(response.into()).await?;
-                }
-                StateStatus::Software(SoftwareOperationVariants::List) => {
-                    let response = SoftwareListCommand::new(&self.config.device, cmd_id)
-                        .with_error("Software List request cancelled on agent restart".to_string());
-                    self.output_sender.send(response.into()).await?;
-                }
-                StateStatus::Restart(_) => {
-                    error!("RestartOperation in store.");
-                }
-                StateStatus::UnknownOperation => {
-                    error!("UnknownOperation in store.");
-                }
-            };
-        }
+        match self.state_repository.load().await {
+            Ok(Some(SoftwareCommand::SoftwareUpdateCommand(request))) => {
+                let response = request.with_error(
+                    "Software Update command cancelled due to unexpected agent restart".to_string(),
+                );
+                self.output_sender.send(response.into()).await?;
+            }
+            Ok(Some(SoftwareCommand::SoftwareListCommand(request))) => {
+                let response = request.with_error(
+                    "Software List request cancelled due to unexpected agent restart".to_string(),
+                );
+                self.output_sender.send(response.into()).await?;
+            }
+            Err(StateError::LoadingFromFileFailed { source, .. })
+                if source.kind() == std::io::ErrorKind::NotFound =>
+            {
+                // file missing means the operation has never been performed, so just do nothing
+            }
+            Err(err) => {
+                // if read failed for some other reason, we should probably log it
+                error!("{err}");
+            }
+            Ok(None) => (),
+        };
         self.state_repository.clear().await?;
         Ok(())
     }
@@ -211,12 +207,7 @@ impl SoftwareManagerActor {
         plugins.load()?;
         plugins.update_default(&get_default_plugin(&self.config.config_location)?)?;
 
-        self.state_repository
-            .store(&State {
-                operation_id: Some(request.cmd_id.clone()),
-                operation: Some(StateStatus::Software(SoftwareOperationVariants::Update)),
-            })
-            .await?;
+        self.state_repository.store(&request.clone().into()).await?;
 
         // Send 'executing'
         let executing_response = request.clone().with_status(CommandStatus::Executing);
@@ -250,12 +241,7 @@ impl SoftwareManagerActor {
             return Ok(());
         }
 
-        self.state_repository
-            .store(&State {
-                operation_id: Some(request.cmd_id.clone()),
-                operation: Some(StateStatus::Software(SoftwareOperationVariants::List)),
-            })
-            .await?;
+        self.state_repository.store(&request.clone().into()).await?;
 
         // Send 'executing'
         let executing_response = request.clone().with_status(CommandStatus::Executing);
--- a/crates/core/tedge_agent/src/software_manager/actor.rs
+++ b/crates/core/tedge_agent/src/software_manager/actor.rs
@@ -4,7 +4,6 @@ use crate::software_manager::error::SoftwareManagerError::NoPlugins;
 use crate::state_repository::state::AgentStateRepository;
 use crate::state_repository::state::SoftwareOperationVariants;
 use crate::state_repository::state::State;
-use crate::state_repository::state::StateRepository;
 use crate::state_repository::state::StateStatus;
 use async_trait::async_trait;
 use plugin_sm::operation_logs::LogKind;
@@ -194,7 +193,7 @@ impl SoftwareManagerActor {
                 }
             };
         }
-        let _state: State = self.state_repository.clear().await?;
+        self.state_repository.clear().await?;
         Ok(())
     }
 
@@ -236,8 +235,7 @@ impl SoftwareManagerActor {
         };
         self.output_sender.send(response.into()).await?;
 
-        let _state: State = self.state_repository.clear().await?;
-
+        self.state_repository.clear().await?;
         Ok(())
     }
 
@@ -272,8 +270,7 @@ impl SoftwareManagerActor {
         };
         self.output_sender.send(response.into()).await?;
 
-        let _state: State = self.state_repository.clear().await?;
-
+        self.state_repository.clear().await?;
         Ok(())
     }
 }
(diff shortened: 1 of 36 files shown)


tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
```diff
@@ -0,0 +1,23 @@
+operation = "controled_restart"
+
+[init]
+script = "/etc/tedge/operations/log-restart.sh ${.topic.cmd_id} ${.payload.status}"
+next = ["scheduled"]
+
+[scheduled]
+script = "restart"
```
albinsuresh, line 8: As you highlighted already, I'd also be in favour of introducing a different `action` key instead of overloading the `script` key.

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
```diff
@@ -0,0 +1,23 @@
...
+next = ["restarting", "successful_restart", "failed_restart"]
```
albinsuresh, line 9: I still don't fully get the idea behind the need for that first `restarting` state in this array. Why not just model it as an independent `restarting` state transitioning from the `scheduled` state? That `restarting` state can then have subsequent `successful_restart` and `failed_restart` states?

crates/core/tedge_agent/src/state_repository/state.rs
```diff
@@ -1,186 +1,106 @@
...
-        let mut state_repo_path = state_repo_root.clone();
-        state_repo_path.push(Utf8PathBuf::from_str(file_name).expect("infallible"));
-
-        Self {
-            state_repo_path,
-            state_repo_root,
-        }
+    /// Clear the current operation by clearing the persisted file
+    pub async fn clear(&self) -> Result<(), StateError> {
+        atomically_write_file_async(&self.state_repo_path, "".as_bytes()).await?;
```
albinsuresh, line 58: We should be deleting the file itself here, right? Instead of just clearing the contents? Or the empty file aids with some logic? It's currently harmless as empty content is currently ignored on load. But, still...

crates/core/tedge_agent/src/state_repository/state.rs
```diff
@@ -1,186 +1,106 @@
...
 use serde::Serialize;
-use std::str::FromStr;
+use std::marker::PhantomData;
 use tedge_utils::fs::atomically_write_file_async;
 use tokio::fs;
 
+/// Store the current state of an operation
 #[derive(Debug)]
-pub struct AgentStateRepository {
+pub struct AgentStateRepository<T> {
```
albinsuresh, line 11: Just a remark: Now that this struct is used to store the state at a per operation level and not as a repository of states of multiple operations, the `Repository` in its name sounds a bit misleading. But, we can live with that for now.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -168,35 +165,29 @@ impl SoftwareManagerActor {
...
-                StateStatus::UnknownOperation => {
-                    error!("UnknownOperation in store.");
-                }
-            };
-        }
-        let _state: State = self.state_repository.clear().await?;
+        match self.state_repository.load().await {
+            Ok(Some(SoftwareCommand::SoftwareUpdateCommand(request))) => {
+                let response = request
+                    .with_error("Software Update command cancelled on agent restart".to_string());
```
albinsuresh, line 171:
```suggestion
                    .with_error("Software Update command cancelled due to unexpected agent restart".to_string());
```

crates/core/tedge_api/src/messages.rs
```diff
@@ -457,12 +460,62 @@ impl From<SoftwareError> for Option<SoftwareModuleItem> {
...
 pub struct RestartCommandPayload {
     #[serde(flatten)]
     pub status: CommandStatus,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    context: Option<RestartContext>,
+}
+
+impl RestartCommandPayload {
+    pub fn new(status: CommandStatus) -> Self {
```
albinsuresh, line 513:
```suggestion
    #[cfg(test)]
    pub fn new(status: CommandStatus) -> Self {
```
As this only gonna be used by the tests.

crates/core/tedge_api/src/messages.rs
```diff
@@ -481,6 +534,31 @@ impl CommandPayload for RestartCommandPayload {
...
+        let command = self.command.clone();
+        match error {
+            None => command.move_to(self.on_success.clone()),
+            Some(reason) => {
+                let command = command.update_from_json(json!({ "restartError": reason }));
+                if self.on_error == "failed" {
+                    command.fail_with(reason)
+                } else {
+                    command.move_to(self.on_error.clone())
+                }
```
albinsuresh, line 556:
```suggestion
                if self.on_error == "failed" {
                    command.fail_with(reason)
                } else {
                    let command = command.update_from_json(json!({ "restartError": reason }));
                    command.move_to(self.on_error.clone())
                }
```

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
```diff
@@ -0,0 +1,23 @@
+operation = "controled_restart"
```
albinsuresh, line 1:
```suggestion
operation = "controlled_restart"
```

albinsuresh (commented):
Now that `restart` is an action that can be triggered generically from any workflow, there is the possibility of multiple workflows (for different operations) requesting for a restart simultaneously as well. We'll have to extend the support for that in future so that all those requested operations succeed, even though one of those requests really restarted the device.

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
didier-wenzek: The proposal is to with something along these lines:

```
trigger = {
    action = "restart",
    on_exec = "restarting",
    on_success = "successful_restart",
    on_error = "failed_restart"
}
```

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
didier-wenzek: The shortcut I took I indeed confusing. I should not have hijack the __next__ field and I will fix that.

What you describe is indeed the correct workflow: moving from *scheduled* to *restarting* and then to *successful_restart* or *failed_restart*.

The actual purpose is different: these 3 state names are used as alias to the state names of the *restart* internal workflow. The states of the latter cannot be used unchanged (as the main workflow already has *executing*,  
*successful* and *failed* states) nor hard-coded (as the main workflow might invoke many restarts).

crates/core/tedge_agent/src/state_repository/state.rs
didier-wenzek: I tried to not change the logic of the restart command (which has not been so easy to get right) but to only change the file format to persist a context along the bare minimum info previously stored.

crates/core/tedge_agent/src/state_repository/state.rs
didier-wenzek: This struct is used to store the state at a per operation level since the restart operation has been introduced. However, this were really weird as each operation were ready to have in its store foreign data rejected with unknown operation errors.

crates/core/tedge_agent/src/software_manager/actor.rs
didier-wenzek: fixed

crates/core/tedge_api/src/messages.rs
didier-wenzek: Changed. Indeed better to not duplicate the error message.

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
didier-wenzek: fixed

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
albinsuresh: Why not continue using the `next` field itself for those successful and failed target states instead of this new `on_success` and `on_error` directives? So that it is consistent with the rest of the workflow actions?

That `on_exec` state is still a bit confusing to me. A user looking at the workflow definition to make sense out of the end-to-end control flow will be a bit lost seeing that `restarting` state, as there is no other reference to that state in the workflow file. I guess I'm still missing something about this contract.

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
albinsuresh: Why is it wrong to model the same workflow as follows (note the updated definition of the `scheduled` state):

```toml
operation = "controlled_restart"

[init]
script = "/etc/tedge/operations/log-restart.sh ${.topic.cmd_id} ${.payload.status}"
next = ["scheduled"]

[scheduled]
inbuilt-action = "restart"
next = ["successful_restart", "failed_restart"]

[successful_restart]
script = "/etc/tedge/operations/log-restart.sh ${.topic.cmd_id} ${.payload.status}"
next = ["successful"]

[failed_restart]
script = "/etc/tedge/operations/log-restart.sh ${.topic.cmd_id} ${.payload.status} ${.payload.restartError}"
next = ["failed"]
```

When the `tedge-agent` processes the `inbuilt-action = restart`, it knows that it was called in the context of the `controlled_restart` command, from the `scheduled` state. It can persist all this information before the restart and once the restart completes, it can just retrieve the state from which `restart` was invoked, which is `scheduled` in this case, and then just transition to either of the states defined in its `next` field for that state.

This is simplistic view that I have about handling restart. I'm sure I'm missing something here. I'm just not able to see what it is yet.

didier-wenzek: > When the `tedge-agent` processes the `inbuilt-action = restart`, it knows that it was called in the context of the `controlled_restart` command, from the `scheduled` state. It can persist all this information before the restart and once the restart completes, it can just retrieve the state from which `restart` was invoked, which is `scheduled` in this case, and then just transition to either of the states defined in its `next` field for that state.

With one exception described bellow, what you describe is what is actually implemented. The context of the restart (.i.e. the whole MQTT message describing the state of the main workflow is stored by the restart operation and resumed after the reboot).

> This is simplistic view that I have about handling restarts in the context of other operations by simply adding more states. I'm sure I'm missing something here. I'm just not able to see what it is yet.

Moving from a state triggering a restart to a state waiting for the end of the restart is required to avoid a race condition.

With your example, where there is no *restarting* state, there is a risk for an infinite loop. If the retained message for the *sheduled* state is read before the restart completion is detected (these are two independent threads), then the workflow will erroneously trigger a new restart.

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
albinsuresh: As discussed offline, that `restarting` state being implicit, while the `successful_restart` and `failed_restart` states being explicit was the confusing aspect for me. One way to make things clearer would be to have that `restarting` state also explicitly defined in the workflow as follows:

```toml
operation = "controlled_restart"

# ...

[scheduled]
action = "restart"
next = ["restarting"]

[restarting]
action = "check-restart"
next = ["successful_restart", "failed_restart"]

[successful_restart]
script = "/etc/tedge/operations/log-restart.sh ${.topic.cmd_id} ${.payload.status}"
next = ["successful"]

[failed_restart]
script = "/etc/tedge/operations/log-restart.sh ${.topic.cmd_id} ${.payload.status} ${.payload.restartError}"
next = ["failed"]

# ...
```

The issue with this approach is that we're forcing the user to explicitly define additional states as per the internal contract of the `restart` action. The need to have an explicit `restarting` state may not be that intuitive to a user.

Another option is to completely hide the `restarting` state by allowing the agent to auto-generate its own internal `restarting` states without exposing it to the user in the workflow files. For example, if the `restart` action is called from the `scheduled` state, then the `status` will be updated form `scheduled` to `scheduled_restarting`. If another state `xyz` calls this action, then it will be moved to `xyz_restarting` state. That way, we can avoid state value conflicts as well. The only issue with this approach is that the user will notice these "internal states" that he did not define in his workflow file. But, a minor issue IMO. The user should be expecting it while relying complex inbuilt actions like `restart`.

tests/RobotFramework/tests/tedge_agent/workflows/custom_restart.toml
albinsuresh: > With your example, where there is no restarting state, there is a risk for an infinite loop. If the retained message for the sheduled state is read before the restart completion is detected (these are two independent threads), then the workflow will erroneously trigger a new restart.

Yeah, now the need for that `restarting` state is clear. I've documented some of my thoughts on how to make this flow clearer in this [comment](https://github.com/thin-edge/thin-edge.io/pull/2479#discussion_r1409229947).

docs/src/references/agent/operation-workflow.md
```diff
@@ -255,7 +255,35 @@ Data extracted from the command status topic and message payload can be passed a
 
 The script exit status and output is used to determine the next step for the command.
 - If the script cannot be launched or return a non-zero status, the command request is marked as __failed__.
-- If the script returns a json payload, this payload is injected into the previous message payload
-  (adding new fields, overriding overlapping ones, keeping previous unchanged ones).
-- If the script returns a json payload with a `status` field this status is used as the new status for the command.
+- If the script successfully returns, its standard output is used to update the command state payload.
+  - From this output, only the excerpt between a `:::begin-tedge:::` header and a `:::end-tedge:::` trailer is decoded.
+    This is done to ease script authoring. A script can emit arbitrary output on its stdout,
```
albinsuresh, line 260: Initially I was wondering why we couldn't just parse the last `{` and `}` chars instead of `:::begin-tedge:::` and `:::end-tedge:::`. But, I understand that this simplifies our parsing logic, if there are multiple JSON objects printed by earlier commands. But still, picking up the last JSON object should still be okay, as I can't imagine any script performing more steps after printing this status message, which is supposed to be the last thing that it does.

But, this simplification can be done later as well, as this contract is also acceptable.

albinsuresh (approved):
LGTM.

One confusing thing that I found is the usage of the words `state` and `step` synonymously. I feel that using `state` consistently would make it easier for the users and in the code. I prefer `state` over `step` as that's what each step in the workflow represents at the end of the day.

crates/core/tedge_api/src/workflow.rs
```diff
@@ -54,13 +106,36 @@ pub enum OperationAction {
     /// The command is delegated to a participant identified by its name
     Delegate(String),
 
+    /// Restart the device
+    Restart {
+        restarting: StateName,
```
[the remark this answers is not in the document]
didier-wenzek: Changed.

docs/src/references/agent/operation-workflow.md
```diff
@@ -255,7 +255,35 @@ Data extracted from the command status topic and message payload can be passed a
...
+    (adding new fields, overriding overlapping ones, keeping previous unchanged ones).
+  - If this excerpt is a json payload with a `status` field, then this status is used as the new status for the command.
 - If the script output is empty, then the exit status of the process is used to determine the next step. 
+
+### Restart action
+
+A workflow can trigger a device restart, using the builtin __restart__ action.
+
+This action is controlled by three states:
+- the *on_executing* state to which the workflow moves once the reboot scheduled
```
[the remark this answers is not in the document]
didier-wenzek: changed

gligorisaev:

- tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot

Issue #2487, opened 2023-11-29
Cumulocity operation failureReason field cannot contain quotes


jarhodes314:
**Is your feature improvement request related to a problem? Please describe.**
In #2474, I modified the code to debug print some values in errors that later get turned into failure reasons for operations. I was surprised to discover that the mapper removes quotes from failure reasons when serialising them to SmartREST, so this didn't work as intended.

**Describe the solution you'd like**
Quotes in the error message should be preserved when they are sent to Cumulocity.

**Describe alternatives you've considered**

**Additional context**

rina23q:
Double quotes in failure reason must be escaped by two double quotes like below.
```text
502,c8y_UploadConfigFile,"contains""double""quote"
```

This sanitize function escapes double quotes by two double quotes, so maybe our failure reason generation doesn't use this function?
https://github.com/thin-edge/thin-edge.io/blob/b2162397cfef4c73dd7d5b794fdd3adb6d55b604/crates/core/c8y_api/src/smartrest/message.rs#L37-L46

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
- tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot
- tests/RobotFramework/tests/cumulocity/log/log_operation.robot
- [x] QA has tested the function and it's functioning according description.

PR #2493, opened 2023-11-30
Support quotes in Cumulocity failureReason field

jarhodes314:
## Proposed changes
This PR simplifies the CSV serialization logic for `SmartRestSetOperationToFailed`, allowing quotes inside the `failure_reason` field to be preserved.

It also makes a small tweak to an error message modified as part of #2474 to make it more generic, referring to `config-manager` instead of `tedge-configuration-plugin` as suggested by @albinsuresh. 

## Types of changes

## Paste Link to the issue
#2487 

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/smartrest/message.rs
+++ b/crates/core/c8y_api/src/smartrest/message.rs
@@ -34,17 +34,18 @@ pub fn get_smartrest_template_id(payload: &str) -> String {
 /// - Strip the input according to `max_size`.
 // TODO: make this return Result
 // TODO: make a variant which assumes `max_size = MAX_PAYLOAD_LIMIT_IN_BYTES`
-pub fn sanitize_for_smartrest(input: Vec<u8>, max_size: usize) -> String {
-    String::from_utf8(input)
-        .unwrap_or_else(|err| {
-            error!("The input contains invalid UTF-8: {err}");
-            String::default()
-        })
+pub fn sanitize_bytes_for_smartrest(input: &[u8], max_size: usize) -> String {
+    let input = std::str::from_utf8(input).unwrap_or_else(|err| {
+        error!("The input contains invalid UTF-8: {err}");
+        ""
+    });
+    sanitize_for_smartrest(input, max_size)
+}
+
+pub fn sanitize_for_smartrest(input: &str, max_size: usize) -> String {
+    input
         .chars()
         .filter(|&c| c == '\r' || c == '\n' || c == '\t' || !c.is_control())
-        .collect::<String>()
-        .replace('"', "\"\"")
-        .chars()
         .scan(0, |bytes_count, c| {
             *bytes_count += c.len_utf8();
             Some((*bytes_count, c))
@@ -62,17 +63,16 @@ pub fn sanitize_for_smartrest(input: Vec<u8>, max_size: usize) -> String {
 /// If the input has only one line, returns the line only.
 /// If the input is empty or contains invalid UTF-8, it returns an empty String.
 /// The output is ensured to be SmartREST compatible.
-pub fn get_failure_reason_for_smartrest(input: Vec<u8>, max_size: usize) -> String {
-    let input_string = String::from_utf8(input).unwrap_or_else(|err| {
+pub fn get_failure_reason_for_smartrest(input: &[u8], max_size: usize) -> String {
+    let input_string = std::str::from_utf8(input).unwrap_or_else(|err| {
         error!("The input contains invalid UTF-8: {err}");
-        String::default()
+        ""
     });
-    let last_line = input_string.lines().last().unwrap_or_default();
-    let failure_reason = match input_string.lines().count() {
-        0 | 1 => last_line.to_string(),
-        _ => format!("{}\n\n{}", last_line, input_string.as_str()),
-    };
-    sanitize_for_smartrest(failure_reason.as_bytes().to_vec(), max_size)
+    let last_line = input_string.trim_end().lines().last().unwrap_or_default();
+    match input_string.lines().count() {
+        0 | 1 => sanitize_for_smartrest(last_line, max_size),
+        _ => sanitize_for_smartrest(&format!("{last_line}\n\n{input_string}"), max_size),
+    }
 }
 
 /// Split MQTT message payload to multiple SmartREST messages.
@@ -139,7 +139,6 @@ pub fn collect_smartrest_messages(data: &str) -> Vec<String> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use regex::Regex;
     use test_case::test_case;
 
     #[test_case("512,device_id", Some("device_id"); "valid")]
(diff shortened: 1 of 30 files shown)


rina23q (approved):
The only concern was to address `,` inside failure reason, however, I see that this unit test proves that the `csv` crates adds double quotes if a string contains a comma. So I would approve the PR.

```rust
    #[test]
    fn serialize_smartrest_set_operation_to_failed_with_comma_reason() {
        let smartrest = SmartRestSetOperationToFailed::new(
            CumulocitySupportedOperations::C8ySoftwareUpdate,
            "Failed to install collectd, modbus, and golang.".into(),
        )
        .to_smartrest()
        .unwrap();
        assert_eq!(
            smartrest,
            "502,c8y_SoftwareUpdate,\"Failed to install collectd, modbus, and golang.\"\n"
        );
    }
```

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -1,15 +1,76 @@
...
 
 pub type SmartRest = String;
 
-#[derive(Debug)]
+pub fn request_pending_operations() -> &'static str {
+    "500"
+}
+
+/// Generates a SmartREST message to set the provided operation to executing
+pub fn set_operation_executing(operation: impl C8yOperation) -> String {
```
jarhodes314, line 18: The main additional changes from the last review are things like this, which replace the Serialization structs with functions.

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -32,185 +93,124 @@ impl From<CumulocitySupportedOperations> for &'static str {
...
-            message_id: "502",
-            operation: operation.into(),
-            reason,
+impl Serialize for EmbeddedCsv {
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+        S::Error: serde::ser::Error,
+    {
+        let record = dbg!(parse_single_record(&self.0)?);
```
jarhodes314, line 161: Oops, didn't mean to leave that `dbg!` in. Just leaving this note here so I remember to get rid of it.

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -32,185 +93,124 @@ impl From<CumulocitySupportedOperations> for &'static str {
...
-        }
+impl From<String> for TextOrCsv {
+    fn from(value: String) -> Self {
+        Self::Text(value)
     }
 }
 
-impl<'a> SmartRestSerializer<'a> for SmartRestSetOperationToSuccessful {}
+#[derive(Debug, Eq, PartialEq)]
+pub struct EmbeddedCsv(String);
```
jarhodes314, line 147: Additionally I've added this `EmbeddedCsv` struct for handling embedded CSVs (which come from custom operations with an result type of CSV. This will deserialize and reserialize the CSV when we add the fields to ensure a consistent format, and to ensure we don't generate a broken message (or multiple messages if the output contains newlines).

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -219,76 +219,122 @@ mod tests {
 
     #[test]
```
jarhodes314, line 220: These tests should give a broad overview of the updated functionality. Mostly, it means we no longer have unnecessary quotes and newlines, and generating the messages is simpler.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -68,55 +68,55 @@ mod tests {
         r#"{"pid":"1234","status":"up"}"#,
         "c8y/s/us/test_device:device:main:service:tedge-mapper-c8y",
         r#"104,up"#;
-        "service-monitoring-thin-edge-device"
+        "service monitoring thin-edge device"
```
jarhodes314, line 71: Not related, I just failed to spot (reading the code) that this was a test name and not just more data when it was kebab-case.

crates/extensions/tedge_mqtt_ext/src/test_helpers.rs
```diff
@@ -47,9 +47,7 @@ pub fn assert_message_contains_str(message: &Message, expected: (&str, &str)) {
     let payload = message.payload_str().expect("non UTF-8 payload");
     assert!(
         payload.contains(expected_payload),
-        "Payload assertion failed.\n Actual: {} \n Expected: {}",
-        payload,
-        expected_payload
+        "Payload assertion failed.\n Actual: {payload:?} \n Expected: {expected_payload:?}",
```
jarhodes314, line 50: I added debug printing here because the assertion message was very unclear when it failed due to a newline in the expected payload.

jarhodes314 (commented):
I've added a few comments to guide you through the changes since this was last reviewed. Most of the changes are very monotonous, so the code is lengthy, but they should all stem from the changes in `smartrest_serializer.rs`, and I've highlighted the specific changes there.

crates/core/c8y_api/src/smartrest/message.rs
```diff
@@ -167,63 +166,62 @@ mod tests {
...
-        let input: Vec<u8> = (0x00..0xff).collect();
-        let sanitized = sanitize_for_smartrest(input, MAX_PAYLOAD_LIMIT_IN_BYTES);
-        let re = Regex::new(r"[^\x20-\x7E\xA0-\xFF\t\n\r]").unwrap();
-        assert!(!re.is_match(&sanitized));
+        let input: Vec<u8> = (0x00..=0x7f).collect();
+        let sanitized = sanitize_bytes_for_smartrest(&input, MAX_PAYLOAD_LIMIT_IN_BYTES);
+        assert_eq!(
+            sanitized,
+            "\t\n\r !\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~"
+        );
```
jarhodes314, line 181: I modified the range this test covers and the assertion because the test was originally supplying invalid UTF-8 (so the sanitize function was returning an empty string), and I don't think this was the intention. The Regex matching obfuscated this since the empty string trivially passed the assertion, so I expanded this into a string literal as this guarantees the output is what we want.

albinsuresh (approved):
LGTM

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -204,90 +219,122 @@ mod tests {
...
+        assert_eq!(smartrest, "503,c8y_RelayArray,true,false,true");
+    }
+
+    #[test]
+    fn serialize_smartrest_set_custom_operation_to_successful_with_multi_record_csv_payload() {
+        succeed_operation(
+            "c8y_RelayArray",
+            TextOrCsv::Csv(EmbeddedCsv("true\n1,2,3".to_owned())),
+        )
+        .unwrap_err();
```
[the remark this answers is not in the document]
jarhodes314: I agree that assert is more explicit about what we're doing, but if the assertion fails, the only information you get is that the assertion failed. By comparison, unwrap_err will debug print the Ok value that we tried to extract the error from, which is likely useful.

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
didier-wenzek: Indeed, the intent is better conveyed compared to these SmartRest codes such as 501.

didier-wenzek (approved):

Issue #2489, opened 2023-11-30
New tedge cli command to renew current self-signed certificate


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Renewing the self-signed certificate created by `tedge cert create` requires both the public and private device certificates to be deleted. Also it is annoying for users as the user must get the existing device-id from the public certificate and use it in the `tedge cert create --device-id <name>` command again which makes it very easy for users to misconfigure (e.g. accidentally changing the device-id, or deleting the public cert before reading the device-id thus losing the information.)

**Describe the solution you'd like**

Add a new tedge cli subcommand to handle the recreation of the the public device certificate.

```sh
tedge cert renew
```

The `cert renew` command should:

* Fail (return a non-zero exit code) if an existing public certificate does not exist
* Read the Common Name from the existing certificate (note, reading must be possible even when the certificate is outside the validity period!)
* ~Leave the private cert in-tact (e.g. there is no need to recreate the private key as it does not expire, only the public cert does)~ This requirement will be covered in a different ticket https://github.com/thin-edge/thin-edge.io/issues/2512
* Generate a new public (self-signed) certificate with a new validity period (we can use the existing default of 1 year)
* Print a message on the console that warns the user that they need to upload the public certificate again and that the until this is done the device's connectivity to the cloud will be disrupted

Though it must be noted, that currently `tedge cert renew` is used for self-signed certificates only! The current advice is to still use a PKI service to do certificate signing in production use cases, e.g. Nexus, SmallStep etc.

**Describe alternatives you've considered**

**Additional context**

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tasks/E2E_Tutorials/Connect_to_c8y.robot
- tests/RobotFramework/tests/tedge/tedge_upload_cert.robot
- [x] QA has tested the function and it's functioning according description.
- [x] Documentation existing
- docs/src/operate/security/registration.md

PR #2502, opened 2023-12-05
New tedge cli command to renew current self-signed certificate

PradeepKiruvale:
## Proposed changes

`tedge` command to `renew` the certificate.

```
sudo tedge cert renew

```

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2489

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/certificate/cli.rs
+++ b/crates/core/tedge/src/cli/certificate/cli.rs
@@ -2,6 +2,7 @@ use tedge_config::OptionalConfigError;
 
 use super::create::CreateCertCmd;
 use super::remove::RemoveCertCmd;
+use super::renew::RenewCertCmd;
 use super::show::ShowCertCmd;
 use super::upload::*;
 
@@ -18,6 +19,9 @@ pub enum TEdgeCertCli {
         #[clap(long = "device-id")]
         id: String,
     },
+
+    /// Renew the device certificate
+    Renew,
 
     /// Show the device certificate, if any
     Show,
@@ -70,8 +74,14 @@ impl BuildCommand for TEdgeCertCli {
                 };
                 cmd.into_boxed()
             }
+            TEdgeCertCli::Renew => {
+                let cmd = RenewCertCmd {
+                    cert_path: config.device.cert_path.clone(),
+                    key_path: config.device.key_path.clone(),
+                };
+                cmd.into_boxed()
+            }
         };
-
         Ok(cmd)
     }
 }
--- a/crates/core/tedge/src/cli/certificate/renew.rs
+++ b/crates/core/tedge/src/cli/certificate/renew.rs
@@ -1,0 +1,110 @@
+use super::error::CertError;
+use super::remove::RemoveCertCmd;
+use crate::command::Command;
+use crate::CreateCertCmd;
+use camino::Utf8PathBuf;
+use certificate::NewCertificateConfig;
+use certificate::PemCertificate;
+
+pub struct RenewCertCmd {
+    pub cert_path: Utf8PathBuf,
+    pub key_path: Utf8PathBuf,
+}
+
+impl Command for RenewCertCmd {
+    fn description(&self) -> String {
+        "Renew the self-signed certificate of the device.".into()
+    }
+
+    fn execute(&self) -> anyhow::Result<()> {
+        let config = NewCertificateConfig::default();
+        self.renew_test_certificate(&config)?;
+        eprintln!("Certificate was successfully renewed, for un-interrupted service, the certificate has to be uploaded to the cloud");
+        Ok(())
+    }
+}
+
+impl RenewCertCmd {
+    fn renew_test_certificate(&self, config: &NewCertificateConfig) -> Result<(), CertError> {
+        let id = self.cn_of_self_signed_certificate()?;
+        // Remove existing certificate
+        let rm_cmd = RemoveCertCmd {
+            cert_path: self.cert_path.clone(),
+            key_path: self.key_path.clone(),
+        };
+        rm_cmd.remove_certificate()?;
+
+        // Re-create the certificate, with new validity
+        let create_cmd = CreateCertCmd {
+            id,
+            cert_path: self.cert_path.clone(),
+            key_path: self.key_path.clone(),
+        };
+        create_cmd.create_test_certificate(config)
+    }
+
+    fn cn_of_self_signed_certificate(&self) -> Result<String, CertError> {
+        let pem = PemCertificate::from_pem_file(&self.cert_path).map_err(|err| match err {
+            certificate::CertificateError::IoError(from) => {
+                CertError::IoError(from).cert_context(self.cert_path.clone())
+            }
+            from => CertError::CertificateError(from),
+        })?;
+
+        if pem.issuer()? == pem.subject()? {
+            Ok(pem.subject_common_name()?)
+        } else {
+            Err(CertError::NotASelfSignedCertificate {
+                path: self.cert_path.clone(),
+            })
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use assert_matches::assert_matches;
+    use std::fs;
+    use std::path::Path;
+    use tempfile::*;
+
+    #[test]
+    fn validate_renew_certificate() {
+        let dir = tempdir().unwrap();
+        let cert_path = temp_file_path(&dir, "my-device-cert.pem");
+        let key_path = temp_file_path(&dir, "my-device-key.pem");
+        let id = "my-device-id";
+        let cmd = CreateCertCmd {
+            id: String::from(id),
+            cert_path: cert_path.clone(),
+            key_path: key_path.clone(),
+        };
+
+        assert_matches!(
+            cmd.create_test_certificate(&NewCertificateConfig::default()),
+            Ok(())
+        );
+
+        let cmd = RenewCertCmd {
+            cert_path: cert_path.clone(),
+            key_path: key_path.clone(),
+        };
+
+        assert_matches!(
+            cmd.renew_test_certificate(&NewCertificateConfig::default()),
+            Ok(())
+        );
+        assert_eq!(parse_pem_file(&cert_path).unwrap().tag, "CERTIFICATE");
+        assert_eq!(parse_pem_file(&key_path).unwrap().tag, "PRIVATE KEY");
+    }
+
+    fn temp_file_path(dir: &TempDir, filename: &str) -> Utf8PathBuf {
+        dir.path().join(filename).try_into().unwrap()
+    }
+
+    fn parse_pem_file(path: impl AsRef<Path>) -> Result<pem::Pem, String> {
+        let content = fs::read(path).map_err(|err| err.to_string())?;
+        pem::parse(content).map_err(|err| err.to_string())
+    }
+}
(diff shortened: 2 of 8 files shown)


tests/RobotFramework/tasks/E2E_Tutorials/Connect_to_c8y.robot
```diff
@@ -26,19 +26,8 @@ Set the URL of your Cumulocity IoT tenant
...
-    #You can then check the content of that certificate.
-    ${output}=    Execute Command    sudo tedge cert show    #You can then check the content of that certificate.
-    Should Contain    ${output}    Device certificate: /etc/tedge/device-certs/tedge-certificate.pem
-    Should Contain    ${output}    Subject: CN=${DEVICE_SN}, O=Thin Edge, OU=Test Device
-    Should Contain    ${output}    Issuer: CN=${DEVICE_SN}, O=Thin Edge, OU=Test Device
-    Should Contain    ${output}    Valid from:
-    Should Contain    ${output}    Valid up to:
-    Should Contain    ${output}    Thumbprint:
-
 tedge cert upload c8y command
```
didier-wenzek, line 29: This test has been added here to automate all the steps of the tutorial.

Even, if duplicated elsewhere, it's better to keep it here.

tests/RobotFramework/tests/tedge/tedge_upload_cert.robot
```diff
@@ -6,14 +6,40 @@ Resource    ../../resources/common.resource
...
+    Should Contain    ${output}    Issuer: CN=${DEVICE_SN}, O=Thin Edge, OU=Test Device
+    Should Contain    ${output}    Valid from:
+    Should Contain    ${output}    Valid up to:
+    Should Contain    ${output}    Thumbprint:
+
+Renew the certificate  
+    ${output}=    Execute Command    sudo tedge cert renew    stderr=${True}    stdout=${False}    ignore_exit_code=${True}    
+    Should Contain    ${output}    Certificate was successfully renewed, for un-interrupted service, the certificate has to be uploaded to the cloud
+    #You can then check the content of that certificate.
+    ${output}=    Execute Command    sudo tedge cert show    #You can then check the content of that certificate.
```
didier-wenzek, line 27: This test is easy to fool as a broken renew command that simply emit a message would pass the test.

The test should check that the certificate file has been actually changed.

The test must check that after the certificate uploaded to c8y and and tedge reconnected to c8y, then the device can correctly connect to c8y.

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -0,0 +1,78 @@
...
+        .create(true)
+        .open(path.as_ref())?;
+
+    // Ignore errors - This was the behavior with the now deprecated user manager.
+    // - When `tedge cert create` is not run as root, a certificate is created but owned by the user running the command.
+    // - A better approach could be to remove this `chown` and run the command as mosquitto.
+    let _ = tedge_utils::file::change_user_and_group(path.as_ref(), user, group);
+
+    Ok(file)
+}
```
didier-wenzek, line 78: The function `certificate::create::create_new_file()` must be made public rather than duplicated.

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -0,0 +1,78 @@
...
+        Ok(())
+    }
+}
+
+impl RenewCertCmd {
+    fn renew_test_certificate(&self, config: &NewCertificateConfig) -> Result<(), CertError> {
+        validate_parent_dir_exists(&self.cert_path).map_err(CertError::CertPathError)?;
+
+        // get the ID from the certificate
+        let cert = KeyCertPair::new_selfsigned_certificate(config, &self.device_id()?)?;
```
didier-wenzek, line 37: This doesn't work as this method create a certificate for a *new* private key that is not persisted.

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -0,0 +1,78 @@
...
+use std::fs::OpenOptions;
+use std::io::prelude::*;
+use std::path::Path;
+use tedge_utils::paths::set_permission;
+use tedge_utils::paths::validate_parent_dir_exists;
+
+pub struct RenewCertCmd {
+    /// The path where the device certificate will be stored
+    pub cert_path: Utf8PathBuf,
+}
```
didier-wenzek, line 17: First hint that something is wrong: the command is given no path to the current private key and therefore has no chance to create a self-signed certificate for the *same* private key.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -70,6 +74,12 @@ impl BuildCommand for TEdgeCertCli {
                 };
                 cmd.into_boxed()
             }
+            TEdgeCertCli::Renew => {
+                let cmd = RenewCertCmd {
+                    cert_path: config.device.cert_path.clone(),
+                };
```
didier-wenzek, line 80: Might be simpler to simply recreate also the private key if not required.

```suggestion
                let cmd = CreateCertCmd {
                    id: config.device.id.try_read(&config)?.to_string(),
                    cert_path: config.device.cert_path.clone(),
                    key_path: config.device.key_path.clone(),
                };
```

The error message behind the `?` needs to be improved to make it clear to the user that a certificate can only be renewed if already created.

didier-wenzek (changes requested):
The current proposal doesn't work.

To fix it, the simpler is:
- to re-generate also the private key
- to directly use `CreateCertCmd` instead of creating a  new `RenewCertCmd`.

PradeepKiruvale:
> The current proposal doesn't work.
> 
> To fix it, the simpler is:
> 
> * to re-generate also the private key
Ok, this seems a better solution
> * to directly use `CreateCertCmd` instead of creating a  new `RenewCertCmd`.

 I feel there are a couple of issues here:
- Using the `CreateCertCmd` to create the certificate will not allow us to overwrite the existing one when there is a certificate already. (We can change the API to do that though, is that OK to change).
- This will just tell that the created/renewed certificate needs to be uploaded to the cloud for un-interrupted service

didier-wenzek:
>     * Using the `CreateCertCmd` to create the certificate will not allow us to overwrite the existing one when there is a certificate already. (We can change the API to do that though, is that OK to change).

Correct, `CreateCertCmd` checks first that there is no certificate before proceeding. But, I would really avoid to change the creation logic itself. An option is the device id from the previous one and to delete (or back up in case of errors) the previous certificate before proceeding.

>     * This will just tell that the created/renewed certificate needs to be uploaded to the cloud for un-interrupted service

I don't see what's the issue.

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -0,0 +1,84 @@
+use super::error::CertError;
+use super::remove::RemoveCertCmd;
+use crate::command::Command;
+use crate::CreateCertCmd;
+use camino::Utf8PathBuf;
+use certificate::NewCertificateConfig;
+
+pub struct RenewCertCmd {
+    /// The path where the device certificate will be tored
+    pub id: String,
```
didier-wenzek, line 10: The comment is unrelated to the field.

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -0,0 +1,84 @@
...
+    pub cert_path: Utf8PathBuf,
+    pub key_path: Utf8PathBuf,
+}
+
+impl Command for RenewCertCmd {
+    fn description(&self) -> String {
+        "Renew a self-signed certificate for the device.".into()
```
didier-wenzek, line 17:
```suggestion
        "Renew the self-signed certificate of the device.".into()
```

crates/core/tedge/src/error.rs
```diff
@@ -30,4 +30,7 @@ pub enum TEdgeError {
 
     #[error(transparent)]
     FromConfigNotSet(#[from] tedge_config::ConfigNotSet),
+
+    #[error("Certificate {0} does not exist to renew")]
+    CertificateDoesNotExist(String),
```
didier-wenzek, line 35: There is already an error case defined for that purpose: `CertError::CertificateNotFound`

```suggestion
```

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -0,0 +1,84 @@
...
+        self.renew_test_certificate(&config)?;
+        eprintln!("Certificate was successfully renewed, for un-interrupted service, the certificate has to be uploaded to the cloud");
+        Ok(())
+    }
+}
+
+impl RenewCertCmd {
+    fn renew_test_certificate(&self, config: &NewCertificateConfig) -> Result<(), CertError> {
+        // Remove existing certificate
+        let rm_cmd = RemoveCertCmd {
```
didier-wenzek, line 31: Before removing the certificate, I would add a check that the certificate exists and is actually self-signed.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -70,8 +74,21 @@ impl BuildCommand for TEdgeCertCli {
                 };
                 cmd.into_boxed()
             }
+            TEdgeCertCli::Renew => {
+                if std::fs::metadata(config.device.cert_path.clone()).is_ok() {
```
didier-wenzek, line 78: I would move this check inside `RenewCertCmd::renew_test_certificate()`:

- where the error type can be more appropriate (a specific `CertError` and not a generic `ConfigError`)
- and where the check can be deeper (checking notably that the certificate is actually self-signed)

If you go in that direction, I would also remove the `id` field from `RenewCertCmd` as this id can be extracted from the certificate (this is the subject CN) while checking the certificate is self signed. See `ShowCertCmd::show_certificate()`.

didier-wenzek (commented):
This is working and the code is correct. One point needs to be fixed though (the duplicated error case).

Also, this new command is a bit unsafe and can be misleading: running it on a device with on CA signed certificate will erase that certificate. It might be good to add a check that the certificate to be renew is self-signed (i.e. `pem.subject() == pem.issuer()`).

crates/core/tedge/src/cli/certificate/renew.rs
PradeepKiruvale: fixed

crates/core/tedge/src/error.rs
PradeepKiruvale: removed

crates/core/tedge/src/cli/certificate/renew.rs
PradeepKiruvale: checking if the certificate is `self-signed`

crates/core/tedge/src/cli/certificate/cli.rs
PradeepKiruvale: refactored the code

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -0,0 +1,122 @@
...
+    fn is_self_signed_certificate(&self) -> Result<bool, CertError> {
+        let pem = PemCertificate::from_pem_file(&self.cert_path).map_err(|err| match err {
+            certificate::CertificateError::IoError(from) => {
+                CertError::IoError(from).cert_context(self.cert_path.clone())
+            }
+            from => CertError::CertificateError(from),
+        })?;
+
+        Ok(pem.issuer()? == pem.subject()?)
+    }
```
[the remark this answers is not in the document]
PradeepKiruvale: Yes, fixed it.

crates/core/tedge/src/cli/certificate/cli.rs
reubenmiller: Just for clarification, is the private key recreated or is it only the public certificate which is recreated?

Issue #2494, opened 2023-12-01
don't restart tedge-agent service when using tedge connect or tedge reconnect


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Executing `tedge connect c8y` or `tedge reconnect c8y` rebuilds the bridge configuration and automatically restarts the tedge-agent regardless if it is running. The restarting of the tedge-agent is problematic and no longer required since the decoupling of the tedge-agent to any mapper specific logic.

Being able to update the c8y bridge configuration is useful when performing firmware updates, however if the tedge-agent is restarted, this causes teh firmware update workflow to be interrupted as the tedge-agent is responsible for running the firmware update workflow (e.g. the workflow step would essentially kill the process which is running it.)

**Describe the solution you'd like**

Don't restart `tedge-agent` if the service is already running when running the following commands:
* `tedge reconnect c8y`
* `tedge connect c8y`

Generally service managers already support starting a service in an idempotent way, so it should be a simple case of just using starting the service rather than restarting (you probably don't even need to check if the service is already running yourself)

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
Test included under:
* tests/RobotFramework/tests/tedge_connect/tedge_connect_test.robot

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/tedge_connect/tedge_connect_test.robot
- tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
- tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
- tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot
- tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_child_device.robot
- [x] QA has tested the function and it's functioning according description.
- [x] Documentation existing
- docs/src/references/init-system-config.md

PR #2505, opened 2023-12-05
Don't restart tedge-agent on (re)connect

jarhodes314:
## Proposed changes
This changes `tedge connect` and `tedge reconnect` to merely start `tedge-agent`, and prevents `tedge disconnect c8y` from stopping `tedge-agent` since there is no longer any cloud-specific functionality in tedge-agent.

~In order to make the feature work, a new `start` field in `system.toml`'s `init` section is required, hence this is a breaking change.~ The new `start` field is now optional as per @didier-wenzek's suggestion, so this is no longer a breaking change.

## Types of changes

## Paste Link to the issue
#2494 

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
+++ b/tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
@@ -46,6 +46,7 @@ Prerequisite Parent
     Check for child related content    #Checks if folders that will trigger child device creation are existing
     Set external MQTT bind address    #Setting external MQTT bind address which child will use for communication
     Set external MQTT port    #Setting external MQTT port which child will use for communication Default:1883
+    Restart agent
 
     Sleep    3s
     Execute Command    sudo tedge connect c8y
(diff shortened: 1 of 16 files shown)


tests/RobotFramework/tests/config_management/child_conf_mgmt_plugin.robot
```diff
@@ -228,6 +224,8 @@ Custom Setup
     Execute Command    find . -type f -name "c8y-configuration-plugin_*.deb" -exec dpkg -i {} \\;
 
     ${parent_ip}=    Get IP Address
+    Execute Command    sudo tedge config set http.bind.address ${parent_ip}
+    Restart Service    tedge-agent
```
jarhodes314, line 228: @reubenmiller I'm not quite sure I understand how this test ever worked. It's clearly expecting the child device to connect to the file transfer service on the main device, but the bind address was seemingly never being set?

reubenmiller: Wow, very confusing...but I think I understand it now...These tests were only working because the test cases are executed sequentially, and the first test case `Prerequisite Parent` does a `tedge connect/disconnect` on the parent which was previously restarting the tedge-agent.

jarhodes314: I'll change it to restart the agent where it was previously doing so in that case

didier-wenzek (approved):
Approved

Issue #2337, opened 2023-10-11
Switch from ISO-8601 format to UNIX timestamps in Thin Edge JSON


albinsuresh:
**Is your feature request related to a problem? Please describe.**

Currently, Thin Edge JSON representation of measurements, events and alarms all accept a `time` field in ISO-8601 format. Although this representation is good for readability, it is not that easy to generate one, without using additional libraries. So, when simple child devices need to generate telemetry data, it'd be difficult for them to generate the time in this format, as they'd be forced to use external libraries.

**Describe the solution you'd like**

Using UNIX timestamp representation would be better for the following reasons:

1. Easy to generate natively on many platforms, without requiring any additional libs
2. Easy comparison of two timestamps
3. Parsing is easier (it is just a number)
4. Reduces overhead in messages

**Additional context**

Additional points to consider:

* Should we use seconds or milliseconds since 1970-01-01...I believe seconds might be more "standard" and expected
* Do we allow floats as the timestamps so that we can represent milliseconds and microseconds and how does floating points impact on precision (e.g. rounding errors)

jarhodes314:
In #2508, I have started to implement support for parsing the relevant values. At the moment it parses an `i64` to a value in seconds, which is a pretty standard. In the description, I have summarised the various alternative representations we could use that spring to mind. We could add float support in addition to this for people that want to specify a greater precision than seconds. Equally, we could allow people to provide a string with an integer and then the relevant unit symbol (although this is more complicated and not something I would consider any sort of standard, but is in my mind quite simple) which would avoid rounding errors.

I'm not convinced rounding errors are much of an issue. After a bit of rough calculating, an `f64` has a 52 bit mantissa can store any integer less than or equal to 2^53 with perfect precision. Milliseconds are 1/1000 of a second, which is about 2^-10, so to represent values to millisecond precision* we could do that for values below 2^43 seconds = 278 737.322 years. If someone wants to guarantee their millisecond/microsecond/nanosecond precise values, they could always use RFC3339/ISO8601 instead.

* Ignoring precision errors that could come from converting a decimal value to binary

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the function and it's functioning according description.

PR #2508, opened 2023-12-07
Add support for parsing unix timestamp measurements

jarhodes314:
## Proposed changes
This is an initial proposal for how thin-edge JSON measurements can be augmented to support unix timestamps in addition to ISO-8601 (/RFC-3339 which is what we actually parse and emit). At the moment it does not make any changes to the code generating thin-edge JSON, nor the documentation.

It currently parses a number of seconds (either as an integer or a float). The generated timestamps are `f64`, since this gives us greater than second precision and matches Python's `time.time()`. There are three new configuration values `az.mapper.timestamp_format`, `aws.mapper.timestamp_format` and `service.timestamp_format` (which controls the timestamps we generate for service health messages).

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/issues/2337

## Checklist


crates/core/c8y_api/src/smartrest/alarm.rs
```diff
@@ -8,18 +8,17 @@ pub fn serialize_alarm(c8y_alarm: &C8yAlarm) -> Result<String, time::error::Form
     let smartrest = match c8y_alarm {
         C8yAlarm::Create(alarm) => {
             let smartrest_code = match alarm.severity {
-                AlarmSeverity::Critical => 301,
```
jarhodes314, line 11: As per usual, I bumped into more cases of SmartREST being generated in non-ideal ways, so I tidied that up too in a separate commit

didier-wenzek:
This is a good start.

> It currently parses an integer number of seconds, as this is the easiest format to process with the time crate, although it would be pretty easy to modify this to be something different, which could be:
    integer number of milliseconds/microseconds/nanoseconds
    float number of seconds
    a number with a suffix to represent the unit (e.g. "1701952729000ms") (which could be provided as an alternative to one of the other proposed formats)

I would avoid to accept too numerous alternative formats, as this might add more confusion than help.
- Seconds since Epoch is an usual and sound choice
- If by *float number of seconds* you mean something such as `1701963272.877`, Yes this is a good option. This improves the precision while being compatible with Seconds since Epoch.
- I would avoid Milliseconds since Epoch because not compatible with Seconds since Epoch.

> In short, there's a lot of flexibility here. Parsing something other than seconds or nanoseconds will make the error messages from time a little bit less clear. These errors only occur when times are out of the time crate's supported range, which run from (-9999-01-01 to 9999-12-31), so they're unlikely to occur in practice. So long as we add clear supplemental information about what units we are using, it should be easy to debug in any case 

I will not bother too much on the extent of the accepted date range, as the point is only to represent *now*.

> Not a breaking change at the time of writing this as all it does is add support on the parsing side, but the feature as a whole is definitely a breaking change for people inspecting messages thin-edge sends.

I think the only time-related messages *published* by thin-edge are health check messages. So, the scope of the breaking change is quite small.

All this needs to be discussed with @albinsuresh and @reubenmiller.

jarhodes314:
> I would avoid to accept too numerous alternative formats, as this might add more confusion than help.
> - Seconds since Epoch is an usual and sound choice
> - If by *float number of seconds* you mean something such as `1701963272.877`, Yes this is a good option. This improves the precision while being compatible with Seconds since Epoch.

This is in line with what I would expect as a user. I think the suggestions that aren't just a number of seconds are needlessly confusing for what is meant to be a simpler way of representing timestamps. We can easily accept an integer or floating point number of seconds (your understanding of what I meant by that is correct), I just haven't bothered to do that yet. I expect this to be only marginally less trivial than the integer case though.

jarhodes314:
> > I would avoid to accept too numerous alternative formats, as this might add more confusion than help.
> > 
> > * Seconds since Epoch is an usual and sound choice
> > * If by _float number of seconds_ you mean something such as `1701963272.877`, Yes this is a good option. This improves the precision while being compatible with Seconds since Epoch.
> 
> This is in line with what I would expect as a user. I think the suggestions that aren't just a number of seconds are needlessly confusing for what is meant to be a simpler way of representing timestamps. We can easily accept an integer or floating point number of seconds (your understanding of what I meant by that is correct), I just haven't bothered to do that yet. I expect this to be only marginally less trivial than the integer case though.

The other decision to be made with floats is how much precision do we want to preserve? The way we need to convert the `f64` to an `OffsetDateTime` is by supplying an integer (`i128`) number of nanoseconds. I think at this point I would be tempted to multiply the value by 1000 and then round that to get an integer number of milliseconds before converting to nanoseconds, because much more precision will likely be junk information. If someone really cares about precision, they could always use ISO8601 instead.

reubenmiller:
I agree with all of the suggestions from @didier-wenzek in https://github.com/thin-edge/thin-edge.io/pull/2508#issuecomment-1845625274

> I think the only time-related messages _published_ by thin-edge are health check messages. So, the scope of the breaking change is quite small.
> 

We can make the transition as easy as possible by making the timestamp configuration configurable (on a mapper and tedge level)

* aws setting to control the outgoing message to use RFC3334 or unix timestamp format
* azure setting to control the outgoing message to use RFC3334 or unix timestamp format
* tedge setting to control the local (tedge) message used when internal components of thin-edge.io services use when to post information about when the service health was published locally (e.g. RFC3334 or unix timetsamp)
* Use default settings of unix timestamp (for all of the settings above)

crates/core/tedge_api/src/alarm.rs
```diff
@@ -128,20 +129,40 @@ mod tests {
...
+    source: EntityTopicId::default_main_device(),
+    data: Some(ThinEdgeAlarmData {
+    severity: Some("high".into()),
+    text: Some("I raised it".into()),
+    time: Some(datetime ! (2023 - 12 - 07 13: 00: 00 + 00: 00)),
+    extras: hashmap ! {},
+    }),
+    };
+    "alarm parsing with unix timestamp"
+    )]
```
jarhodes314, line 165: I don't quite know what I messed up here with the indentation, but I'm just leaving a note here so I remember to fix it

crates/core/tedge_api/Cargo.toml
```diff
@@ -19,6 +19,8 @@ mqtt_channel = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 shell-words = { workspace = true }
+# TODO this feels like a bad way to structure things
+tedge_config = { workspace = true }
```
jarhodes314, line 23: This feels like a weird artifact of putting the `TimeFormat` enum in `tedge_config` and it feels like `tedge_api` shouldn't depend on `tedge_config`. @didier-wenzek do you agree/have thoughts on how to resolve this?

didier-wenzek: See https://github.com/thin-edge/thin-edge.io/pull/2508#discussion_r1423944331.

If moving formatting code to `tedge_config` raises other concerns, an alternative is to move the format definition and formatting code to `tedge_utils`, on which  `tedge_config`  already depends.

crates/core/tedge_api/src/parser.rs
```diff
@@ -72,6 +78,119 @@ struct ThinEdgeValueParser<'key, 'vis, T> {
...
+        })
+    }
+
+    pub fn format(&self, time_format: TimeFormat) -> serde_json::Value {
+        match time_format {
+            TimeFormat::Unix => Value::Number(self.0.unix_timestamp().into()),
+            TimeFormat::Rfc3339 => Value::String(self.0.format(&Rfc3339).unwrap()),
+        }
+    }
+}
```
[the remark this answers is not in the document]
jarhodes314: The reformatting can be part of a method on `TimeFormat`, but the `IsoOrUnix` struct exists to support deserialization for the case of Cumulocity values, which are handled very differently to Azure/AWS values. Because we need this deserialization logic in `tedge_api` (e.g. in `alarm.rs`) and also need the same logic to reformat timestamps for Azure and AWS, somewhere we require a crate that knows about both `IsoOrUnix` and `TimeFormat`. At the moment, this is `tedge_api` depending on `tedge_config` for `TimeFormat`, but this could be reversed, or `tedge_config` could implement the logic above and `tedge_api` can just use the implementation there for deserialization as required.

crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -95,18 +104,20 @@ impl AzureConverter {
...
-                .format(&time::format_description::well_known::Rfc3339)?
-                .as_str()
-                .into();
-            payload_json.entry("time").or_insert(timestamp);
+        let default_timestamp = self.add_timestamp.then(|| IsoOrUnix::new(self.clock.now()));
+        let mut payload: Map<String, Value> = serde_json::from_slice(input.payload.as_bytes())?;
+
+        let time = payload
+            .remove("time")
+            .map(|value| IsoOrUnix::reformat_value(value, self.mapper_config.time_format))
```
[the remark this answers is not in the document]
jarhodes314: The actual code has changed to use a match instead now, but worth noting that annoyingly Rust doesn't allow you to specify a bound method as a closure like that, unlike Python, so it would be `.map(|val| self.mapper_config.time_format.format(val))`.

jarhodes314:
I've now moved the timestamp module to `tedge_utils` and re-export it from `tedge_config` (for the convenience of the higher level crates). Annoyingly, it hasn't become much simpler, but now code only needs to interact with either `TimeFormat` (for serialization) or `IsoOrUnix` (for deserialization)

crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -230,15 +260,20 @@ mod tests {
...
+        let mut converter = AzureConverter::new(
+            false,
+            Box::new(TestClock),
+            MqttSchema::default(),
+            TimeFormat::Rfc3339,
+        );
 
         let input = r#"{
-            "time" : "2013-06-22T17:03:14.000+02:00",
+            "time" : "2013-06-22T17:03:14.001+02:00",
```
didier-wenzek, line 271: Why this change has been required? Is this because of a precision-lost issue?

didier-wenzek (commented):
The failing system is likely due to this PR "Watchdog does not kill mapper if it responds"

Also, you still have to fix https://github.com/thin-edge/thin-edge.io/pull/2508/files#r1423786896

didier-wenzek:
> I've now moved the timestamp module to `tedge_utils` and re-export it from `tedge_config` (for the convenience of the higher level crates). Annoyingly, it hasn't become much simpler, but now code only needs to interact with either `TimeFormat` (for serialization) or `IsoOrUnix` (for deserialization)

Yeah, I'm also a bit surprised to see how many dependencies have to be pull down from `tedge_config` to  `tedge_utils`.
But, we will see later how to re-organize the crates.

crates/extensions/az_mapper_ext/src/converter.rs
jarhodes314: I made the change originally because the timestamp was being parsed and then reinserted, which was altering the formatting of the timestamp. With the latest changes, this is no longer the case (we assume given the date is a string rather than a number that it doesn't need reformatting to RFC-3339, so we leave the value unaltered), so I'll undo the change.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -233,47 +231,37 @@ async fn monitor_tedge_service(
...
-
-                    if datetime >= request_timestamp {
-                        return Ok(health_status);
-                    } else {
-                        debug!(
-                            "Ignoring stale health response: {:?} older than request time: {}",
-                            health_status, request_timestamp
-                        );
-                    }
+    while let Some(message) = messages.next().await {
```
jarhodes314, line 245: I figured "loop indefinitely if the channel closes" is likely not the desired behaviour, and could plausibly lead to the actor not shutting down, so I've change the function to return an error when we stop receiving messages.

jarhodes314:
> The failing system is likely due to this PR "Watchdog does not kill mapper if it responds"
> 
> Also, you still have to fix https://github.com/thin-edge/thin-edge.io/pull/2508/files#r1423786896

@didier-wenzek I've now fixed both of these (see https://github.com/thin-edge/thin-edge.io/pull/2508#discussion_r1428044445 about the first point)

crates/extensions/tedge_health_ext/src/lib.rs
```diff
@@ -39,8 +40,10 @@ impl HealthMonitorBuilder {
         mqtt: &mut (impl ServiceProvider<MqttMessage, MqttMessage, TopicFilter> + AsMut<MqttConfig>),
         // TODO: pass it less annoying way
         mqtt_schema: &MqttSchema,
-        mut service_type: String,
+        service_config: &TEdgeConfigReaderService,
```
jarhodes314, line 43: I also modified this to take the relevant chunk of the config rather than two separate arguments for the type and time format, which means a couple of tests now use the new method `TEdgeConfigRepository::from_toml_str` to generate the relevant config.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2547,7 +2548,7 @@ pub(crate) async fn spawn_c8y_mapper_actor(
     let device_name = "test-device".into();
     let device_topic_id = EntityTopicId::default_main_device();
     let device_type = "test-device-type".into();
-    let service_type = "service".into();
+    let config = TEdgeConfigRepository::load_toml_str("service.ty = \"service\"");
```
jarhodes314, line 2551: Ideally we would be able to parse a fake `tedge.toml` string for all of the values here, but we wouldn't be able to control the device ID without generating a certificate. Currently, the need to parse toml is due to use of a tedge config struct to simplify a function signature (see comment in crates/extensions/tedge_health_ext/src/lib.rs for more details). Finding a way to mitigate this would be nice. In general, I would prefer if we could use some mechanism like this more generally as it would be testing the public API, but obviously right now is not the time to deal with such a concern.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -233,47 +231,37 @@ async fn monitor_tedge_service(
...
+        if let Ok(message) = message.payload_str() {
+            debug!("Health response received: {message}");
+            if let Ok(health_status) = serde_json::from_str::<HealthStatus>(message) {
+                let datetime = IsoOrUnix::try_from(&health_status.time)?;
+
+                // Compare to a slightly old timestamp to avoid false negatives from floating-point error in unix timestamps
+                if datetime.into_inner() >= request_timestamp - Duration::from_millis(10) {
```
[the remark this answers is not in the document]
didier-wenzek: >  I am assuming this is only comparing timestamps from the same machine

This is a fair assumption, as the `systemd_watchdog` can only watch processes running on the same box.

> More generally, this doesn't feel like an ideal solution to the problem since the system time is not monotonic and could change between the healthcheck request being sent and the reply being received, although we weren't previously robust to this.

Yeah, an option could be to use `/proc/uptime` for this check. But this is beyond the scope of this PR.

I'm okay with the current proposal.

crates/extensions/c8y_mapper_ext/src/tests.rs
didier-wenzek: >  we wouldn't be able to control the device ID without generating a certificate.

This is also a pain for the users as `tedge config get device.id` is currently meaningful only on the device where the certificate is stored.

didier-wenzek (approved):
Approved.

Issue #2477, opened 2023-11-22
Decouple tedge-mapper-c8y and tedge-agent for config_update 


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
The `config_update` feature requires downloading files from cloud, then storing it to `<data_dir>/cache` directory (e.g. `/var/tedge/cache`). The cache feature is very helpful to save bandwidth because it can prevent downloading the same files multiple times from the internet.

As of now, upon receiving a `config_update` operation from c8y, tedge-mapper-c8y
* checks if a file associated with `remoteUrl` is available in the cache directory
* downloads a file if not available in the cache directory and stores the file to there

The problem is, accessing to `<data_dir>/cache` is the responsibility of tedge-agent. So, tedge-mapper-c8y should not access the directory directly. (For the background of decoupling, refer to #2390)

**Describe the solution you'd like**
The solution is not yet very clear. The must is that tedge-agent downloads a file from cloud.

**Describe alternatives you've considered**

**Additional context**

didier-wenzek:
See https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1392990394

A solution could be to add a cache method in the file-transfer service. A POST request asking the tedge-agent to download a file from the cloud and to serve it locally.

Bravo555:
# QA instructions

The modified test at `tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot` should cover the changes. The gist is that `tedge-agent` should now be fully functional when run on a separate device/container. The new device/container still has to be configured as the main device (via `mqtt.device_topic_id`) and `http.client.host/port` settings should be set only on that physical device on which the agent is running, so that a correct `tedgeUrl` is generated in operation message payloads.

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot
- [x] QA has tested the function and it's functioning according description.

PR #2511, opened 2023-12-08
Move `config_update` file download from tedge-mapper-c8y to tedge-agent

Bravo555:
## TODO

- [x] addressing feedback for unit tests

### Motivation

This PR moves file download from cloud - which needs to happen when handling `config_update` operation - from the mapper to the tedge-agent, in order to simplify the mapper. However, the complexity is not eliminated and is merely moved into tedge-agent, which now needs to have special handling for Cumulocity. However, this is accepted for now, because it enables tedge-agent and tedge-mapper-c8y to be run on different containers, and eliminates the implicit dependency on File Transfer Service (a part of tedge-agent) in tedge-mapper-c8y.

### Summary

The updated operation works like this:

1. The [`Download configuration type with type`](https://cumulocity.com/guides/reference/smartrest-two/#download-configuration-file-with-type-524) Smartrest message is received by the mapper
2. tedge-mapper-c8y converts the smartrest message into the following MQTT message:
    ```
    topic: te/device/device01///cmd/config_update/1234
    payload: {"status": "init", "remoteUrl": "https://example.org/file", "configType": "type"}
    ```
    i.e. `tedgeUrl` property was made optional and in the initial message it isn't set.

3. tedge-configuration-manager (part of tedge-agent) of the given entity, if the config type is supported, marks the operation as executing and waits for the operation to be updated with `tedgeUrl`
4. tedge-agent on the main device, upon seing that `config_update` operation is being executed, but does not have `tedgeUrl`, downloads the file in `remoteUrl`, saves it into FTS, and updates the operation with a valid `tedgeUrl` pointing into FTS
5. tedge-configuration-manager of the given entity, upon seeing that `tedgeUrl` is now present, downloads the file from it and continues processing the operation
6. The workflow continues unchanged from this point.

### Test changes

To test the changes, `configuration_with_file_transfer_https.robot` was modified to start `tedge-agent` on a child device, and `tedge-agent` was removed from the main device.

## Types of changes

## Paste Link to the issue
- https://github.com/thin-edge/thin-edge.io/issues/2477

## Checklist

## Further comments

The current implementation of caching files which are further needed by child devices, is very ad-hoc and as said in https://github.com/thin-edge/thin-edge.io/pull/2071#discussion_r1392990394, can be improved by custom workflow. This is something we'll need to think more about, but for now, it's enough to remove the implicit dependency on the FTS in tedge-mapper-c8y.

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -645,7 +645,7 @@ define_tedge_config! {
         #[doku(as = "PathBuf")]
         key_path: Utf8PathBuf,
 
-        /// Path to a file containing the PEM encoded CA certificates that are
+        /// Path to a directory containing the PEM encoded CA certificates that are
         /// trusted when checking incoming client certificates for the File Transfer Service
         #[tedge_config(example = "/etc/ssl/certs")]
         #[doku(as = "PathBuf")]
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -1,5 +1,6 @@
 use crate::file_transfer_server::actor::FileTransferServerBuilder;
 use crate::file_transfer_server::actor::FileTransferServerConfig;
+use crate::operation_file_cache::FileCacheActorBuilder;
 use crate::restart_manager::builder::RestartManagerBuilder;
 use crate::restart_manager::config::RestartManagerConfig;
 use crate::software_manager::builder::SoftwareManagerBuilder;
@@ -75,6 +76,7 @@ pub(crate) struct AgentConfig {
     pub mqtt_topic_root: Arc<str>,
     pub service_type: String,
     pub identity: Option<Identity>,
+    pub fts_url: Arc<str>,
     pub is_sudo_enabled: bool,
     pub capabilities: Capabilities,
 }
@@ -145,6 +147,11 @@ impl AgentConfig {
             config_snapshot: tedge_config.agent.enable.config_snapshot,
             log_upload: tedge_config.agent.enable.log_upload,
         };
+        let fts_url = format!(
+            "{}:{}",
+            tedge_config.http.client.host, tedge_config.http.client.port
+        )
+        .into();
 
         Ok(Self {
             mqtt_config,
@@ -162,6 +169,7 @@ impl AgentConfig {
             mqtt_device_topic_id,
             service_type: tedge_config.service.ty.clone(),
             identity,
+            fts_url,
             is_sudo_enabled,
             capabilities,
         })
@@ -295,7 +303,7 @@ impl Agent {
             let log_manager_config = LogManagerConfig::from_options(LogManagerOptions {
                 config_dir: self.config.config_dir.clone().into(),
                 tmp_dir: self.config.config_dir.into(),
-                mqtt_schema,
+                mqtt_schema: mqtt_schema.clone(),
                 mqtt_device_topic_id: self.config.mqtt_device_topic_id.clone(),
             })?;
             Some(
@@ -310,6 +318,30 @@ impl Agent {
         } else {
             None
         };
+
+        // TODO: replace with a call to entity store when we stop assuming default MQTT schema
+        let is_main_device =
+            self.config.mqtt_device_topic_id == EntityTopicId::default_main_device();
+        if is_main_device {
+            info!("Running as a main device, starting tedge_to_te_converter and File Transfer Service");
+
+            runtime.spawn(tedge_to_te_converter).await?;
+
+            let file_transfer_server_builder =
+                FileTransferServerBuilder::try_bind(self.config.http_config).await?;
+            runtime.spawn(file_transfer_server_builder).await?;
+
+            let operation_file_cache_builder = FileCacheActorBuilder::new(
+                mqtt_schema,
+                self.config.fts_url.clone(),
+                self.config.data_dir,
+                &mut downloader_actor_builder,
+                &mut mqtt_actor_builder,
+            );
+            runtime.spawn(operation_file_cache_builder).await?;
+        } else {
+            info!("Running as a child device, tedge_to_te_converter and File Transfer Service disabled");
+        }
 
         // Spawn all
         runtime.spawn(signal_actor_builder).await?;
@@ -328,22 +360,6 @@ impl Agent {
         runtime.spawn(script_runner).await?;
         runtime.spawn(converter_actor_builder).await?;
         runtime.spawn(health_actor).await?;
-
-        // TODO: replace with a call to entity store when we stop assuming default MQTT schema
-        let is_main_device =
-            self.config.mqtt_device_topic_id == EntityTopicId::default_main_device();
-        if is_main_device {
-            info!(
-                "Running as a main device, starting tedge_to_te_converter and file transfer actors"
-            );
-
-            let file_transfer_server_builder =
-                FileTransferServerBuilder::try_bind(self.config.http_config).await?;
-            runtime.spawn(tedge_to_te_converter).await?;
-            runtime.spawn(file_transfer_server_builder).await?;
-        } else {
-            info!("Running as a child device, tedge_to_te_converter and file transfer actors disabled");
-        }
 
         runtime.run_to_completion().await?;
 
--- a/crates/core/tedge_agent/src/operation_file_cache/mod.rs
+++ b/crates/core/tedge_agent/src/operation_file_cache/mod.rs
@@ -1,0 +1,360 @@
+//! Inspects incoming operation requests for URLs to files and downloads them into the File Transfer
+//! Service, so that they can be downloaded more quickly by the child devices.
+//!
+//! Payloads of some operations contain `remoteUrl` property which contains a URL from which we
+//! need to download a file. However, child devices may have reduced speed or even may not be able
+//! to reach the remote URL at all. Additionally, multiple child devices may require the same file,
+//! so it makes sense to download it once and place it inside the File Transfer Service, so that
+//! child devices may download the files quickly on the local network.
+//!
+//! This actor, for all child devices, for operations that have `remoteUrl` property, tries to
+//! download the file from this URL, places it in the File Transfer Service, and inserts the URL to
+//! download the file from the FTS in the `tedgeUrl` property.
+
+use async_trait::async_trait;
+use camino::Utf8PathBuf;
+use std::collections::HashMap;
+use std::path::Path;
+use std::sync::Arc;
+use tedge_actors::adapt;
+use tedge_actors::fan_in_message_type;
+use tedge_actors::Actor;
+use tedge_actors::Builder;
+use tedge_actors::DynSender;
+use tedge_actors::LinkError;
+use tedge_actors::LoggingReceiver;
+use tedge_actors::MessageReceiver;
+use tedge_actors::MessageSink;
+use tedge_actors::NoConfig;
+use tedge_actors::RuntimeError;
+use tedge_actors::RuntimeRequest;
+use tedge_actors::RuntimeRequestSink;
+use tedge_actors::Sender;
+use tedge_actors::ServiceProvider;
+use tedge_actors::SimpleMessageBoxBuilder;
+use tedge_api::messages::ConfigUpdateCmdPayload;
+use tedge_api::mqtt_topics::Channel;
+use tedge_api::mqtt_topics::ChannelFilter;
+use tedge_api::mqtt_topics::EntityFilter;
+use tedge_api::mqtt_topics::EntityTopicId;
+use tedge_api::mqtt_topics::MqttSchema;
+use tedge_api::mqtt_topics::OperationType;
+use tedge_api::path::DataDir;
+use tedge_api::CommandStatus;
+use tedge_downloader_ext::DownloadRequest;
+use tedge_downloader_ext::DownloadResult;
+use tedge_mqtt_ext::MqttMessage;
+use tedge_mqtt_ext::Topic;
+use tedge_mqtt_ext::TopicFilter;
+use tracing::error;
+use tracing::info;
+use tracing::warn;
+
+type IdDownloadRequest = (String, DownloadRequest);
+type IdDownloadResult = (String, DownloadResult);
+
+fan_in_message_type!(FileCacheInput[MqttMessage, IdDownloadResult]: Debug);
+
+pub struct FileCacheActor {
+    input_receiver: LoggingReceiver<FileCacheInput>,
+    mqtt_sender: DynSender<MqttMessage>,
+    downloader_sender: DynSender<IdDownloadRequest>,
+
+    tedge_http_host: Arc<str>,
+    mqtt_schema: MqttSchema,
+    data_dir: DataDir,
+
+    pending_operations: HashMap<String, ConfigUpdateCmdPayload>,
+}
+
+#[async_trait]
+impl Actor for FileCacheActor {
+    fn name(&self) -> &str {
+        "FileCacheActor"
+    }
+
+    async fn run(mut self) -> Result<(), RuntimeError> {
+        while let Some(message) = self.input_receiver.recv().await {
+            match message {
+                FileCacheInput::MqttMessage(message) => self.process_mqtt_message(message).await?,
+                FileCacheInput::IdDownloadResult(message) => self.process_download(message).await?,
+            }
+        }
+
+        Ok(())
+    }
+}
+
+impl FileCacheActor {
+    async fn process_mqtt_message(
+        &mut self,
+        mqtt_message: MqttMessage,
+    ) -> Result<(), RuntimeError> {
+        let Ok((entity, channel)) = self.mqtt_schema.entity_channel_of(&mqtt_message.topic) else {
+            return Ok(());
+        };
+
+        let Channel::Command {
+            operation: OperationType::ConfigUpdate,
+            cmd_id,
+        } = channel
+        else {
+            return Ok(());
+        };
+
+        let update_payload =
+            match serde_json::from_slice::<ConfigUpdateCmdPayload>(mqtt_message.payload.as_bytes())
+            {
+                Ok(payload) => payload,
+                Err(err) => {
+                    warn!("Received config update, but payload is malformed: {err}");
+                    return Ok(());
+                }
+            };
+
+        if update_payload.remote_url.is_empty() || update_payload.tedge_url.is_some() {
+            return Ok(());
+        }
+
+        match &update_payload.status {
+            CommandStatus::Executing => {
+                self.download_config_file_to_cache(&mqtt_message.topic, &update_payload)
+                    .await?;
+            }
+            CommandStatus::Successful => self.delete_symlink_for_config_update(
+                &entity,
+                &update_payload.config_type,
+                &cmd_id,
+            )?,
+            CommandStatus::Failed { .. } => self.delete_symlink_for_config_update(
+                &entity,
+                &update_payload.config_type,
+                &cmd_id,
+            )?,
+            _ => {}
+        }
+
+        Ok(())
+    }
+
+    async fn process_download(&mut self, download: IdDownloadResult) -> Result<(), RuntimeError> {
+        let (topic, result) = download;
+
+        let Some(mut operation) = self.pending_operations.remove(&topic) else {
+            return Ok(());
+        };
+
+        let (entity, Channel::Command { cmd_id, .. }) = self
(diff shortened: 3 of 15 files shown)


crates/core/tedge_agent/src/operation_file_cache/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+                self.download_config_file_to_cache(&mqtt_message.topic, &update_payload)
+                    .await?;
+            }
+            CommandStatus::Successful => {
+                self.delete_symlink_for_config_update(
+                    &entity,
+                    &update_payload.config_type,
+                    &cmd_id,
+                )
+                .unwrap();
```
rina23q, line 129: Is it really safe?

crates/core/tedge_agent/src/operation_file_cache/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+            Ok(download) => download,
+        };
+
+        self.create_symlink_for_config_update(
+            &entity,
+            &operation.config_type,
+            &cmd_id,
+            download.file_path,
+        )
+        .unwrap();
```
rina23q, line 174: The same, is it really safe?

rina23q (commented):
The solution to the issue looks good. 

The major problems are

- Removing the access to symlinks in c8y-mapper is forgotten in both code and tests.
- Splitting `config_operations.rs` to two files is nice, but it's a pain to review since you made some changes on `config_update.rs`.

Also some minors.

didier-wenzek:
> - Splitting `config_operations.rs` to two files is nice, but it's a pain to review since you made some changes on `config_update.rs`.

@Bravo555 Can you do the following:

1. Create a new PR, where you introduce only these changes:
     1. Introduce this new `operations` module
     2. `git mv` there `log_update.rs`, `firmware_update.rs`, `config_operations.rs`
     3. split `config_operations.rs` into `config_snapshot.rs` and `config_update.rs`.
2. This reorg PR should be reviewed and merged quickly.
3. Rebase this PR on top of the reorg PR.
4. @rina23q rebase https://github.com/thin-edge/thin-edge.io/pull/2482 on top of this same reorg PR.

This will ease reviewing this PR and merging it with Rina's work.

Bravo555:
Made a separate reorg PR:

https://github.com/thin-edge/thin-edge.io/pull/2517

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -635,6 +635,10 @@ define_tedge_config! {
             },
         },
 
+        /// The entity which should start tedge File Transfer Service
+        #[tedge_config(default(value = "device/main//"), example = "device/main//", example = "device/child//")]
+        entity: Arc<str>,
```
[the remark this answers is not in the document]
Bravo555: No, there is a small difference. This setting is not "what is the entity id of the current device", but "the agent of which device should spawn the FTS?"

First, we don't want to only enable the FTS on the main device, because usually the c8y mapper runs on the main device, so the FTS should run on child devices as well. So I first tried to create `enable.file_transfer_service`, to be set to true on a device when we want to start it, and false when we don't.
But what should be the default value?
1. If it's `false` then by default FTS won't be started on a main device, and we'll introduce a breaking change where it would be necessary to set this config to `true` to enable it back on.
2. If the default is `true`, then it will start on a main device, but also on any child device as well, so we'd be forced to always manually set it to `false` on any child device.
3. If we used a combination of "start if running on a main device, and if not, then use the `enable.file_transfer_service` with default of `false`" then we wouldn't be able to disable the FTS for the main device.

Using an explicit entity topic id solves that problem. By default, it's the main device, so the behaviour is unchanged and everything works. If it's changed to a child device, then only that child device agent starts the FTS and other agents don't.
One problem is that we still need to change configuration not in 1, but 2 places for it to work: the child device on which we plan to run FTS on, and on the main device so it doesn't run there by default. For the rest of the child devices, the default value of `device/main//` will prevent their agents from starting, so we don't have to change anything on them. Since [the plan is to have abstraction over a configuration using MQTT or HTTP API so that all devices don't need to have access to the same filesystem to change or read the configuration](https://github.com/thin-edge/thin-edge.io-roadmap/issues/44), perhaps we could then only set it in 1 place.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+//! proxy to report on their progress. Additionally, while executing operations we often need to
+//! send messages to different actors and wait for their results before continuing.
+//!
+//! The operations are always triggered remotely by Cumulocity, and a triggered operation must
+//! always terminate in a success or failure. This status needs to be reported to Cumulocity.
+//!
+//! This module contains:
+//! - data definitions of various states which are necessary to maintain in the mapper
+//! - status and error handing utilities for reporting operation success/failure in different ways
+//!   (MQTT, Smartrest)
```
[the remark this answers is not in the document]
Bravo555: Right, this should rather say just "MQTT", because that's how we actually report the status of the operations, just publishing an MQTT message on the operation channel. The translation to smartrest is handled elsewhere in the mapper.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -294,11 +294,16 @@ impl ConfigManagerActor {
...
+            return Ok(());
+        };
+
+        let download_request = DownloadRequest::new(tedge_url, temp_path.as_std_path())
             .with_permission(file_entry.file_permissions.to_owned());
 
         info!(
-            "Awaiting download for config type: {} from url: {}",
+            "Awaiting download for config type: {} from url: {:?}",
             request.config_type, request.tedge_url
```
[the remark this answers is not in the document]
jarhodes314: Can't this just be `tedge_url` rather than `request.tedge_url` which would actually guarantee there's a value there?

crates/core/tedge_agent/src/operation_file_cache/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+        };
+
+        let Ok(update_payload) =
+            serde_json::from_slice::<ConfigUpdateCmdPayload>(mqtt_message.payload.as_bytes())
+        else {
+            warn!("Received config update, but payload is malformed");
+            return Ok(());
+        };
+
+        if update_payload.remote_url.is_empty() || update_payload.tedge_url.is_some() {
```
jarhodes314, line 114: I'm hypothesising a lot here, but it feels here like there is some logic that decides from the payload what state we're in that's distributed around this actor's methods. Can we do some sort of conversion from `ConfigUpdateCmdPayload` into an enum that represents what state we're currently in so we process that information in a single place?

jarhodes314 (commented):
I've had a decent look at the agent side, I haven't had a particularly close look at the mapper changes yet

crates/core/tedge_agent/src/operation_file_cache/mod.rs
Bravo555: I may be missing what you mean slightly, but IMO this is much more the case in other operation handlers, which have more states, here we only: 1. process messages which have a `remoteUrl` and don't have `tedgeUrl`, 2. download from the remote URL, put in the file-transfer directory, and put the URL inside `tedgeUrl`. I agree that transitions between states in operation handling is hardly visible, but I thought about improving that in all the operations later, but here I'm not sure if you mean something simpler than that.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
Bravo555: changed it to return `RuntimeError`, any errors when creating symlinks usually arise from bugs, so they are handled by returning and exiting the actor

crates/core/tedge_agent/src/operation_file_cache/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+            .file_transfer_dir()
+            .join(dir_name_of(entity_topic_id))
+            .join("config_update");
+        let symlink_path =
+            symlink_dir_path.join(format!("{}-{cmd_id}", config_type.replace('/', ":")));
+
+        if !symlink_path.is_symlink() {
+            std::fs::create_dir_all(symlink_dir_path)?;
+            std::os::unix::fs::symlink(original, &symlink_path)?;
+        }
```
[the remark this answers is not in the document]
Bravo555: As this symlink is unique and shouldn't be created by anything else, I decided to just return the error if the symlink already exists, but i think in practice there will be little difference.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
jarhodes314: I hadn't deeply considered a solution, it just "felt like" it could do with being more "parse, don't validate", but I'm happy to be overruled here.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -284,6 +298,39 @@ impl Agent {
...
+        } else {
+            info!("Running as a child device, tedge_to_te_converter disabled");
+        }
+
+        if *entity_topic_id
+            .default_parent_identifier()
+            .unwrap()
+            .as_str()
+            == *self.config.fts_entity
+        {
```
didier-wenzek, line 317: I don't understand the need motivating this change.

Having the FTS service running on the main device sounds to be a default that makes.
And even if more flexibility might help in some case, this is independent of this PR.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
```diff
@@ -0,0 +1,364 @@
...
+                    &Topic::new(&topic).unwrap(),
+                    serde_json::to_string(&operation).unwrap(),
+                );
+                self.mqtt_sender.send(message).await?;
+                return Ok(());
+            }
+            Ok(download) => download,
+        };
+
+        self.create_symlink_for_config_update(
```
didier-wenzek, line 171: What's the purpose of this symlink?

crates/core/tedge_agent/src/operation_file_cache/mod.rs
didier-wenzek: My main concern is that the logic to process a config update request is scattered in so many places that it's really difficult to be sure that this is working as expected and really easy to break.

- Instead of what we have today (some dumb payloads in `tedge_api` and processing logic in `c8y_mapper_ext`, `tedge_config_ext` and now `tedge_agent`)
- I would prefer a central place where are defined the misc steps required to process a configuration update and to reduce the role of the mapper, agent, and plugins to trigger actions.

This is how I understand @jarhodes314 call for "parse, don't validate":

```
impl FileCacheActor {
    async fn process_mqtt_message(
        &mut self,
        mqtt_message: MqttMessage,
        match ConfigUpdate::action_for(mqtt_message) {
            Some(DownloadRemote { cloud_url, }) => {
                ...
            } 
        }
}
```

Unfortunately, this is not the best time to fix that.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
Bravo555: I'm not sure if I understand the question. It should be the same as before: we download the file to the data dir cache directory, and then create the symlink to this file in the file transfer directory. After the operation is complete, we delete the symlink.

didier-wenzek: Ok thanks. I though this have been introduced by this PR.

@rina23q do you know the answer?

crates/core/tedge_agent/src/agent.rs
Bravo555: As agreed in the call, this setting was removed and File Transfer Service was reverted to spawn only on the main device.

crates/extensions/c8y_mapper_ext/src/operations/config_update.rs
```diff
@@ -0,0 +1,562 @@
...
+    use tedge_api::mqtt_topics::OperationType;
+    use tedge_mqtt_ext::test_helpers::assert_received_contains_str;
+    use tedge_mqtt_ext::MqttMessage;
+    use tedge_mqtt_ext::Topic;
+    use tedge_test_utils::fs::TempTedgeDir;
+
+    const TEST_TIMEOUT_MS: Duration = Duration::from_millis(5000);
+
+    #[tokio::test]
+    async fn mapper_converts_smartrest_config_download_req_without_new_download_for_child_device() {
```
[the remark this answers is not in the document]
Bravo555: Fixed and previous test for the main device was restored.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -294,12 +294,17 @@ impl ConfigManagerActor {
         // move to destination later
         let temp_path = &self.config.tmp_path.join(&file_entry.config_type);
 
-        let download_request = DownloadRequest::new(&request.tedge_url, temp_path.as_std_path())
+        let Some(tedge_url) = &request.tedge_url else {
+            debug!("tedge_url not present in config update payload, ignoring");
+            return Ok(());
```
albinsuresh, line 299: I guess it's too late now, but it feels like properly using the scheduled state (the agent updating the state from `init` to `schedule` after the download, and then this config manager actor reacting only to that `scheduled` state) would have been clearer, in terms of the operation control flow, IMO. Though it works, reacting to the `executing` state differently at different times looks a bit convoluted.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
```diff
@@ -0,0 +1,360 @@
...
+        let (topic, result) = download;
+
+        let Some(mut operation) = self.pending_operations.remove(&topic) else {
+            return Ok(());
+        };
+
+        let (entity, Channel::Command { cmd_id, .. }) = self
+            .mqtt_schema
+            .entity_channel_of(&Topic::new(&topic).unwrap())
+            .expect("only topics targetting config update command should be inserted")
```
albinsuresh, line 150:
```suggestion
            .expect("only topics targeting config update command should be inserted")
```

crates/core/tedge_agent/src/operation_file_cache/mod.rs
albinsuresh: The file-transfer service only allows you to share files under the `/var/tedge/file-transfer` directory (or at least that was the case earlier). Since the cached files are downloaded to a different location (`/var/tedge/cache`) which doesn't fall under the `file-transfer` dir, we just create symlinks under that directory pointing to the cached file.

Bravo555:
It seems that the integration tests are very flaky, also failing on the main branch:

https://github.com/thin-edge/thin-edge.io/actions/runs/7219433666/job/19670621507

Could it be due to recent changes to registration message handling?

crates/extensions/tedge_config_manager/src/actor.rs
Bravo555: But arguably, downloading the config the file is a normal part of execution of the `config_file` operation, and most of the actual time of the operation execution is spent there, so I don't think downloading in `init` or `schedule` state would be better. For other complex operations there can be multiple steps for which we don't have enough distinct states. I do agree, however, that it's a bit hard to track and I expect that perhaps custom workflows and some other refactorings I plan for operation code would at least alleviate this somewhat.

reubenmiller:
> It seems that the integration tests are very flaky, also failing on the main branch:
> 
> https://github.com/thin-edge/thin-edge.io/actions/runs/7219433666/job/19670621507
> 
> Could it be due to recent changes to registration message handling?

Quickly looking at the failed tests, I think it might be due to the workflow stuff (https://github.com/thin-edge/thin-edge.io/pull/2496)

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -3108,7 +3083,9 @@ pub(crate) mod tests {
         let auth_proxy_protocol = Protocol::Http;
         let mut topics =
             C8yMapperConfig::default_internal_topic_filter(&tmp_dir.to_path_buf()).unwrap();
-        topics.add_all(crate::log_upload::log_upload_topic_filter(&mqtt_schema));
+        topics.add_all(crate::operations::log_upload::log_upload_topic_filter(
+            &mqtt_schema,
+        ));
```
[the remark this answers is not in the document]
Bravo555: fixed in 08944c0a7

didier-wenzek:
> > It seems that the integration tests are very flaky, also failing on the main branch:
> > https://github.com/thin-edge/thin-edge.io/actions/runs/7219433666/job/19670621507
> > Could it be due to recent changes to registration message handling?
> 
> Quickly looking at the failed tests, I think it might be due to the workflow stuff (#2496)

So indeed related to checking that a restart can be triggered using a workflow.

The test is red but for bad reason as the outcome is as expected (except that the message is doubled).

```
Matching messages is greater than maximum. wanted: 1 got: 2 messages: ['{"old-tedge-agent-pid":"MainPID=100","status":"tedge-agent-restarted","tedge-agent-pid":"MainPID=467"}', '{"old-tedge-agent-pid":"MainPID=100","status":"tedge-agent-restarted","tedge-agent-pid":"MainPID=467"}'] 
```

I will push asap a PR to make this test more robust.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
rina23q: What @Bravo555 said is totally correct.

Creating a symlink from cache is more about the security. Since the files under `/var/tedge/file-transfer` can be accessible from all other thin-edge components, that means they can modify the contents by using HTTP requests. Keeping `/var/tedge/cache` out of the file transfer directory ensures the files cannot change by any HTTP requests.

crates/extensions/c8y_mapper_ext/src/operations/config_update.rs
```diff
@@ -237,92 +137,41 @@ impl CumulocityConverter {
...
-            external_id,
-            smartrest.config_type.replace('/', ":"),
-            cmd_id
-        );
+        let proxy_url = self
+            .c8y_endpoint
+            .maybe_tenant_url(&smartrest.url)
+            .map(|cumulocity_url| self.auth_proxy.proxy_url(cumulocity_url).into());
+
+        let remote_url = proxy_url.unwrap_or(smartrest.url.to_string());
```
rina23q, line 146: I would raise a discussion for later. This change will mix up the meaning of `remoteUrl` field.

In config_update, the `remoteUrl` is used only to download a file. So proxy url as `remoteUrl` seems fine.
In firmware_update, the `remoteUrl` is necessary to but it has to be the original url (no proxy url!). Otherwise, in the end, c8y mapper cannot report the current firmware including url.

I think we need two fields, one indicates the original request URL, and the other indicates the url where actually `tedge-agent` downloads the file.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
```diff
@@ -0,0 +1,360 @@
...
+        };
+
+        let download = match result {
+            // if cant download file, operation failed
+            Err(err) => {
+                let error_message = format!("tedge-agent failed downloading a file: {err}");
+                operation.failed(&error_message);
+                error!("{}", error_message);
+                let message = MqttMessage::new(
+                    &Topic::new(&topic).unwrap(),
```
rina23q, line 162: Minor. We have `Topic::new_unchecked()`, which suits here.

rina23q (approved):
LGTM

My only concern is https://github.com/thin-edge/thin-edge.io/pull/2511#discussion_r1427874580. But it's not for this PR.

crates/extensions/tedge_config_manager/src/actor.rs
didier-wenzek: Yeah, indeed too late to fix that, but it would have been better to introduce a specific state - say downloading.

>  I do agree, however, that it's a bit hard to track and I expect that perhaps custom workflows and some other refactorings I plan for operation code would at least alleviate this somewhat.

I expect the same ;-)

didier-wenzek (approved):
Approved.

albinsuresh (approved):
LGTM

We can improve this later with explicit `downloading` and `downloaded` states as proposed in the comments.

crates/core/tedge_agent/src/operation_file_cache/mod.rs
albinsuresh: > Keeping /var/tedge/cache out of the file transfer directory ensures the files cannot change by any HTTP requests.

Good point @rina23q

crates/core/tedge_agent/src/operation_file_cache/mod.rs
Bravo555: fixed

PR #2522, opened 2023-12-14
Make entity store persistent #2428

albinsuresh:
## Proposed changes

Persist the entity store as a JSON lines file.
Every registration message and twin data message is persisted as JSON lines. On startup the in-memory entity store is rebuilt by replaying these messages.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/core/tedge_api/src/entity_store.rs
```diff
@@ -191,15 +186,109 @@ impl EntityStore {
...
+        };
+
+        entity_store.load_from_message_log();
+
+        Ok(entity_store)
+    }
+
+    pub fn load_from_message_log(&mut self) {
+        info!("Loading the entity store from the log");
+        match self.message_log.reader() {
```
didier-wenzek, line 210: I would be better to transform this method into a function taking mutable references to an entity store and a log of messages; populating the store from these messages.

A bit more involved, but I would also try to persist the entity store from outside, i.e with a function taking the store and the message log. This can be done by using an internal vector of non-persisted-yet messages.

Doing so:
- The entity store will stay a pure data structure with no side effects.
- The load and persist methods can be made async.
- The logic would be simpler avoiding this usage of boolean flags to control in depth when something needs to be persisted or not.
- The coupling to the mqtt_schema could also be removed that way.

didier-wenzek:
Persisting the raw MQTT messages is a bit opaque:

```
{"topic":{"name":"te/device/main///twin/c8y_Agent"},"payload":[123,34,110,97,109,101,34,58,34,116,104,105,110,45,101,100,103,101,46,105,111,34,44,34,117,114,108,34,58,34,104,116,116,112,115,58,47,47,116,
104,105,110,45,101,100,103,101,46,105,111,34,44,34,118,101,114,115,105,111,110,34,58,34,48,46,49,51,46,49,34,125],"qos":1,"retain":true}

This is even strange as what matters is cryptic, while what is useless is clear (qos, retain).
```

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -114,20 +119,26 @@ type ExternalIdValidatorFn =
...
 ///     registration_message,
+///     "service".into(),
 ///     |tid, xid| tid.to_string().into(),
 ///     |xid| Ok(xid.into()),
+///     5,
+///     "/tmp"
 /// );
 /// ```
 pub struct EntityStore {
+    mqtt_schema: MqttSchema,
```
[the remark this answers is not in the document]
albinsuresh: Yes, this was one of the options that I had evaluated in the last PR, as described in [this comment](https://github.com/thin-edge/thin-edge.io/pull/2466#discussion_r1421973763). The main reason why I moved away from that `PersistentMessage` enum to raw `MqttMessage` was to cover the possibility of making more message types persistent in future, especially the ones like measurements that don't have a concrete struct representation.

albinsuresh:
> Persisting the raw MQTT messages is a bit opaque:
> 
> ```
> {"topic":{"name":"te/device/main///twin/c8y_Agent"},"payload":[123,34,110,97,109,101,34,58,34,116,104,105,110,45,101,100,103,101,46,105,111,34,44,34,117,114,108,34,58,34,104,116,116,112,115,58,47,47,116,
> 104,105,110,45,101,100,103,101,46,105,111,34,44,34,118,101,114,115,105,111,110,34,58,34,48,46,49,51,46,49,34,125],"qos":1,"retain":true}
> 
> This is even strange as what matters is cryptic (payload), while what is useless is clear (qos, retain).
> ```

Will update the serialization logic of `DebugPayload` to cover this case, once we finalize on [this point](https://github.com/thin-edge/thin-edge.io/pull/2522#discussion_r1428911753) whether to continue persisting raw `MqttMessgae` instances or introduce a dedicated `PersistentMessage` enum for the same.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -727,6 +727,13 @@ define_tedge_config! {
         path: Utf8PathBuf,
     },
 
+    db: {
+        /// The directory used to store any persistent data that survives tedge updates and even firmware updates
+        #[tedge_config(example = "/data/tedge", default(value = "/data/tedge"))]
+        #[doku(as = "PathBuf")]
+        path: Utf8PathBuf,
+    },
```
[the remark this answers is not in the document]
reubenmiller: Yeah I'll have to take some time to think about this one. We might need to have a discussion regarding the nuances between the two behaviours and what are the pros/cons.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: As discussed offline, this would be deferred for later.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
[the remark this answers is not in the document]
reubenmiller: As discussed we have decided to put the entity store persistence under the `/etc/tedge/.tedge-mapper-<name>` folder (or something to that effect).

Having the data under `/etc/tedge` rather than the persisted `/data/tedge` folder makes the upgrade path of the agent across partitions more reliable, as the state under /etc/tedge will be left unmodified allow a rollback to the previous partition more possible (without changing the potentially upgraded state files).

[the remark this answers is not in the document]
albinsuresh: As discussed offline, reverted this config setting and persisting the log file to `/etc/tedge/.tedge-mapper-c8y` directory instead.

crates/core/tedge_api/src/message_log.rs
```diff
@@ -0,0 +1,152 @@
...
+            .open(log_dir.as_ref().join(LOG_FILE_NAME))?;
+
+        // If the file is empty append the version information as a header
+        let metadata = file.metadata()?;
+        let file_is_empty = metadata.len() == 0;
+
+        let mut writer = BufWriter::new(file);
+
+        if file_is_empty {
+            let version = env!("CARGO_PKG_VERSION");
```
albinsuresh, line 86: @reubenmiller suggested that we use an independent version scheme for this persistence store file format rather than using the tedge version itself, which will make the validations easier in future, even when the same file format is used across multiple releases. So, this will be version `1.0` of the schema. And it will only be updated when there's a breaking change to the schema.

crates/core/tedge_api/src/message_log.rs
```diff
@@ -0,0 +1,152 @@
...
+        P: AsRef<Path>,
+    {
+        let file = OpenOptions::new()
+            .create(true)
+            .append(true)
+            .open(log_dir.as_ref().join(LOG_FILE_NAME))?;
+
+        // If the file is empty append the version information as a header
+        let metadata = file.metadata()?;
+        let file_is_empty = metadata.len() == 0;
```
didier-wenzek, line 82: This sounds a bit fragile, at least unusual.

As this is an append-only log of messages - tracking version is a bit more involved. The first messages are not necessarily written by the same version of the software as the following messages. 

What is proposed is okay, but the whole log will have to be rewritten if a new format is introduced.

didier-wenzek (approved):
Approved

crates/core/tedge_api/src/message_log.rs
albinsuresh: My expectation was that it would be better to discard the log and start fresh if there is a major version mismatch between the version expected by the software and the file read from the disk. Even though I've defined a minor version as well, I don't really have a clear plan on when that would change, keeping the formats compatible between those minor versions. TBH I've only thought about breaking major version changes and kept the minor version number just to enable any possibilities in future.

didier-wenzek: > My expectation was that it would be better to discard the log and start fresh if there is a major version mismatch between the version expected by the software and the file read from the disk.

It's why I don't bother to much to have something simple for now - even if a bit fragile.

> Even though I've defined a minor version as well, I don't really have a clear plan on when that would change, keeping the formats compatible between those minor versions. TBH I've only thought about breaking major version changes and kept the minor version number just to enable any possibilities in future.

Yeah, this sounds premature, but this is also harmless. So let's go with that for the rc.

albinsuresh:
# Test Plan

The externally visible effects of this work are the following:

1. All the entities registered and their twin data are now persisted on the disk at `/etc/tedge/.tedge-mapper-c8y/entity_store.jsonl` as a JSON lines file consisting of all the entity registration messages and twin messages. No duplicate messages expected in this file.
2. The entity store of the mapper is rebuilt using the contents of this file on every startup, before it starts processing any MQTT messages. So, when the mapper re-delivers the same retained messages on restart, the mapper will just ignore those, avoiding the duplicate conversion of these messages to Cumulocity as well.
3. Any new messages published while the mapper was down, should still be processed/converted by the mapper after the startup. Only the metadata messages that it had previously seen/registered/updated, before going down, are ignored.

gligorisaev:

- tests/RobotFramework/tests/cumulocity/registration/device_registration.robot

PR #2540, opened 2023-12-20
Fix flaky entity store persistence test

albinsuresh:
## Proposed changes

The tests were failing due to a time precision problem while validating MQTT messages with a `date_from` range and not because there was a duplicate message. The `date_from` value takes a UNIX timestamp without milli-second precision. But in the failing test case, the following was happening (note the time precision):

1. Entity registration message published at `10:10:35.513002` 
2. Converted message received by the logger at `10:10:36.198326`
3. The timestamp to be used in the `date_from` for the assertion was generated using `Get Unix Timestamp` API at `10:10:36.756000`. But, the `Get Unix Timestamp` returns the time: `1703067036` without millisecond precision.

When that timestamp without precision, which is  `10:10:36.000000` is used as the `date_from` in the assertion, it covers the original message as well, and that wrongly detects it as a duplicate message.

To make sure that there is a delay of more than 1 second between the point at which the converted message was emitted and the point of assertion, the managed object existence checks were added in between which should add a significant delay.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
+++ b/tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
@@ -157,7 +157,6 @@ Early data messages cached and processed
 
 
 Entities persisted and restored
-    ${timestamp}=    Get Unix Timestamp
     Execute Command    tedge mqtt pub --retain 'te/factory/shop/plc1/' '{"@type":"child-device","@id":"plc1"}'
     Execute Command    tedge mqtt pub --retain 'te/factory/shop/plc2/' '{"@type":"child-device","@id":"plc2"}'
     Execute Command    tedge mqtt pub --retain 'te/factory/shop/plc1/sensor1' '{"@type":"child-device","@id":"plc1-sensor1","@parent":"factory/shop/plc1/"}'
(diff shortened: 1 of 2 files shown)


tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
```diff
@@ -166,19 +165,26 @@ Entities persisted and restored
...
-    Should Have MQTT Messages    c8y/s/us/plc2    message_contains=101,plc2-sensor1    date_from=${timestamp}    minimum=1    maximum=1
-    Should Have MQTT Messages    c8y/s/us/plc1    message_contains=102,plc1-metrics    date_from=${timestamp}    minimum=1    maximum=1
-    Should Have MQTT Messages    c8y/s/us/plc2    message_contains=102,plc2-metrics    date_from=${timestamp}    minimum=1    maximum=1
+    External Identity Should Exist    plc1
+    External Identity Should Exist    plc2
+    External Identity Should Exist    plc1-sensor1
+    External Identity Should Exist    plc1-sensor2
+    External Identity Should Exist    plc2-sensor1
+    External Identity Should Exist    plc1-metrics
+    External Identity Should Exist    plc2-metrics
```
albinsuresh, line 174: Even a simple 1s sleep here would have solved the issue. But, doing a deterministic assertion here looked better.

reubenmiller (approved):

Issue #2567, opened 2024-01-05
large failure reason text corrupts the 502 message when setting operation status to FAILED


reubenmiller:
**Describe the bug**

When a custom Cumulocity IoT operation fails and there a large standard error output greater than the max MQTT message size, then it results in an ill-formed SmartREST message as the failure reason quote is not terminated:

For example, a `502` SmartREST message is sent where only the opening quote is included in the failure reason (e.g. the 3rd field of the CSV text).

```sh
[c8y/s/us] 502,c8y_NodeRed,"long failure reason....
```

**To Reproduce**

1. Use the c8y-command-plugin (which provides the `c8y_Command` operation handler)
2. Create a Shell Operation via the Device Managment Application which generates at least 17000 characters on the Standard Error stream (stderr)
3. Check the operation status in the UI and view that it is "stuck" in the EXECUTING state

**Expected behavior**

Long failure reasons are generally not that helpful as it causes visualization problems in UIs and is anyway unreadable.

The following are the suggested rules to apply to the failure reason handling:

* Only take the last N lines of the custom operation handler
* If the failure reason exceeds X bytes, then the content should be truncated (from the end), and the open and closing quotes should always be included, e.g. `502,c8y_NodeRed,"some long text..."`

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`unknown`|
|System-Architecture|`Linux f28a6f153954 6.5.0-10-generic #10-Ubuntu SMP PREEMPT_DYNAMIC Fri Oct 13 18:28:22 UTC 2023 aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.0.0-rc.1`|

**Additional context**

Below shows the subscription to the `c8y/#` topic where the error was spotted:

```sh
tedge mqtt sub 'c8y/#'
```

```text
[c8y/s/us] 501,c8y_NodeRed
[c8y/s/us] 502,c8y_NodeRed,"2024-01-05 14:38:14,806 - c8y_NodeRed - ERROR - Aborting due to error: 400 Client Error: Bad Request for url: http://localhost:1880/flow

[...]

PR #2586, opened 2024-01-12
Limit the length of the message for failed operation

Ruadhri17:
## Proposed changes
This PR introduces a limit to the size of the message that is sent when the operation fails. If the failure reason is longer than 200 characters, it is truncated from the end, and the user gets a warning message about truncation. 

## Types of changes

## Paste Link to the issue
* #2567
<br/>

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
+++ b/crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
@@ -7,6 +7,7 @@ use serde::ser::SerializeSeq;
 use serde::Deserialize;
 use serde::Serialize;
 use serde::Serializer;
+use tracing::warn;
 
 pub type SmartRest = String;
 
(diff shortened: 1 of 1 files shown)


crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -21,7 +22,12 @@ pub fn set_operation_executing(operation: impl C8yOperation) -> String {
 
 /// Generates a SmartREST message to set the provided operation to failed with the provided reason
 pub fn fail_operation(operation: impl C8yOperation, reason: &str) -> String {
-    fields_to_csv_string(&["502", operation.name(), reason])
+    if reason.len() <= 200 {
+        fields_to_csv_string(&["502", operation.name(), reason])
+    } else {
+        warn!("Failure reason too long, message trancuated to 200 characters");
```
didier-wenzek, line 28: It might be good to log the whole message.

didier-wenzek (commented):
Two system tests are now failing as the expected error message has been truncated.

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
reubenmiller: Though we should be truncating the failure reason, and not ignoring it completely!

For example, ellipsis (`...`) can be used to represent that the failure reason was truncated:

```
My custom script output failed. check logs fo...
```

reubenmiller: Sorry ignore my comment, I didn't see the line below...

reubenmiller: > It might be good to log the whole message.

I would be careful logging the full message as this could be very large...in the end we also have the full stdout/stderr output in the log file under `/var/log/tedge/agent/*.log`

Ruadhri17: Initial 100 characters were definitely too short so I increased that to 200 but I would go even further and set it to 500.

didier-wenzek (approved):
Approved

rina23q (approved):
LGTM

Issue #2501, opened 2023-12-05
restart tedge-agent after the agent is updated via the cloud


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Currently the tedge-agent service is not restarted after the package is updated, as the service is responsible for orchestrating the software update, so restarting the service during the postinst maintainer script would be destructive and result in a "half-installed" package.

However, not restarting the tedge-agent service results in unexpected behaviour where the older tedge-agent binary is still running but there is no visible signs that the older version is running, and thus debugging any erroneous behaviour very difficult.

**Describe the solution you'd like**

During a software update, if the tedge-agent detects that it's binary has been changed, then the process should be restarted cleanly after all of the packages have been processed.

Due to complications introduced by the [default KillMode=control-group](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillMode=) in Systemd, the restart cannot occur during the linux package management. Instead just "cleanly exiting" might be the best strategy, and rely on the service supervision to restart the service.

* Feature flag to enable/disable this "auto exit" behaviour (this enable users that are not using a process supervisor to opt-out of this behaviour, as it would be catastrophic for them otherwise).
* Detect if the tedge-agent version (e.g. `tedge-agent --version`) after a software installation, and exit cleanly (if enabled)

**Notes**
* Don't assume you are running under a specific service manager
* Don't detect if an agent update is being done by looking at the incoming software update changes, as there is no easy way to determine if the tedge-agent is part of it or not (e.g. if the users are using virtual packages, which pull in dependencies automatically)

**Describe alternatives you've considered**

**Additional context**

didier-wenzek:
The fix cannot be implemented by the **tedge-agent** itself, as this must work even when updating an old version of the agent.

A solution is to trigger a restart of the agent from the mapper when a new version of `tedge-agent` is detected after a software update. Indeed, thanks to the all-in-one `tedge` executable, the mapper and the agent will be simultaneously updated, and, after a successful update of thin-edge from the cloud, the mapper will be in position to restart the agent in case the latter is not running the freshly installed version of `tedge`.

The c8y mapper needs to be extended with the following checks and actions:

1. On any software update request, probe `tedge-agent --version`
2. Forward the software update request to the agent as usual.
3. When the outcome of the software update request is received and is successful,
      1. Probe again `tedge-agent --version` and compare the result with the version returned before the update.
[...]

PR #2587, opened 2024-01-12
Restart tedge-agent after self update

albinsuresh:
## Proposed changes

When tedge-agent detects that the `tedge` binary itself is updated, exit with a non-zero exit code, so that the process manger like systemd restarts the upgraded agent.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2501

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/software_manager/actor.rs
+++ b/crates/core/tedge_agent/src/software_manager/actor.rs
@@ -1,8 +1,10 @@
+use crate::agent::TEDGE_AGENT;
 use crate::software_manager::config::SoftwareManagerConfig;
 use crate::software_manager::error::SoftwareManagerError;
 use crate::software_manager::error::SoftwareManagerError::NoPlugins;
 use crate::state_repository::error::StateError;
 use crate::state_repository::state::AgentStateRepository;
+use anyhow::anyhow;
 use async_trait::async_trait;
 use plugin_sm::operation_logs::LogKind;
 use plugin_sm::operation_logs::OperationLogs;
@@ -11,6 +13,7 @@ use plugin_sm::plugin_manager::Plugins;
 use serde::Deserialize;
 use serde::Serialize;
 use std::path::PathBuf;
+use std::process::Command;
 use tedge_actors::fan_in_message_type;
 use tedge_actors::Actor;
 use tedge_actors::LoggingReceiver;
@@ -101,7 +104,12 @@ impl Actor for SoftwareManagerActor {
 
         while let Some(request) = input_receiver.recv().await {
             tokio::select! {
-                _ = self.handle_request(request, &mut plugins, &operation_logs) => {}
+                _ = self.handle_request(request, &mut plugins, &operation_logs) => {
+                    if let Err(SoftwareManagerError::NotRunningLatestVersion) = Self::detect_self_update() {
+                        error!("Tedge-agent is no more running the latest-version => a restart is required");
+                        return Err(RuntimeError::ActorError(Box::new(SoftwareManagerError::NotRunningLatestVersion)));
+                    }
+                }
 
                 Some(RuntimeRequest::Shutdown) = input_receiver.recv_signal() => {
                     info!("Received shutdown request from the runtime, exiting...");
@@ -147,11 +155,12 @@ impl SoftwareManagerActor {
     ) -> Result<(), SoftwareManagerError> {
         match request {
             SoftwareCommand::SoftwareUpdateCommand(request) => {
-                if let Err(err) = self
+                match self
                     .handle_software_update_operation(request, plugins, operation_logs)
                     .await
                 {
-                    error!("{:?}", err);
+                    Ok(()) => {}
+                    Err(err) => error!("{:?}", err),
                 }
             }
             SoftwareCommand::SoftwareListCommand(request) => {
@@ -229,6 +238,42 @@ impl SoftwareManagerActor {
         self.output_sender.send(response.into()).await?;
 
         self.state_repository.clear().await?;
+        Ok(())
+    }
+
+    fn detect_self_update() -> Result<(), SoftwareManagerError> {
+        info!("Checking if tedge got self updated");
+        let current_running_version = env!("CARGO_PKG_VERSION");
+        info!("Current running version: {}", current_running_version);
+
+        let executable_path = std::env::current_exe()
+            .map_err(|e| anyhow!("Failed to retrieve running executable path due to {}", e))?;
+        let agent_binary_path = executable_path.parent().unwrap().join(TEDGE_AGENT);
+
+        let output = Command::new(agent_binary_path)
+            .args(["--version"])
+            .output()
+            .map_err(|e| anyhow!("Failed to fetch version of installed binary due to {}", e))?;
+        if !output.status.success() {
+            return Err(anyhow!(
+                "Fetching version from installed binary failed with {}",
+                String::from_utf8_lossy(&output.stderr)
+            )
+            .into());
+        }
+
+        let version_output = String::from_utf8_lossy(&output.stdout);
+        let version_output_split: Vec<&str> = version_output.split_whitespace().collect();
+        if let ["tedge-agent", installed_binary_version] = version_output_split.as_slice() {
+            info!("Installed binary version: {}", installed_binary_version);
+            if current_running_version != *installed_binary_version {
+                info!("Self update detected. Requesting shutdown...");
+                return Err(SoftwareManagerError::NotRunningLatestVersion);
+            }
+        } else {
+            return Err(anyhow!("Unexpected version output: {:?}", version_output).into());
+        }
+
         Ok(())
     }
 
--- a/tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
+++ b/tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
@@ -49,7 +49,6 @@ Update tedge version from previous using Cumulocity
     ${OUTPUT}    Execute Command    systemctl is-enabled tedge-mapper-az || exit 1    exp_exit_code=1    strip=True
     Should Be Equal    ${OUTPUT}    disabled    msg=Service should still be disabled
 
-
 Refreshes mosquitto bridge configuration
     ${PREV_VERSION}=    Set Variable    0.10.0
     # Install base version
@@ -74,6 +73,29 @@ Refreshes mosquitto bridge configuration
     # Mosquitto should be restarted with new bridge
     Execute Command    cmd=sh -c '[ $(journalctl -u mosquitto | grep -c "Loading config file /etc/tedge/mosquitto-conf/c8y-bridge.conf") = 2 ]'
 
+Update tedge version from base to current using Cumulocity
+    ${arch}=    Get Debian Architecture
+    Execute Command    mkdir -p /setup/base-version
+    # These base-version packages are built using the ci/build_scripts/build.sh script for all 4 supported architectures
+    # by overriding the version to a lower number using the GIT_SEMVER env variable
+    Transfer To Device    ${CURDIR}/base-version/*${arch}.deb    /setup/base-version/
+
+    # Install base version with self-update capability
+    Install Packages    /setup/base-version
+    Execute Command    cd /setup && test -f ./bootstrap.sh && ./bootstrap.sh --no-install --no-secure
+    Device Should Exist                      ${DEVICE_SN}
+    ${pid_before}=    Execute Command    pgrep tedge-agent    strip=${True}
+
+    # Upgrade to current version
+    Create Local Repository
+    ${OPERATION}=    Install Software    tedge,${NEW_VERSION}    tedge-mapper,${NEW_VERSION}    tedge-agent,${NEW_VERSION}    tedge-watchdog,${NEW_VERSION}    tedge-apt-plugin,${NEW_VERSION}
+
+    Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
+    Device Should Have Installed Software    tedge,${NEW_VERSION_ESCAPED}::apt    tedge-mapper,${NEW_VERSION_ESCAPED}::apt    tedge-agent,${NEW_VERSION_ESCAPED}::apt    tedge-watchdog,${NEW_VERSION_ESCAPED}::apt    tedge-apt-plugin,${NEW_VERSION_ESCAPED}::apt
+
+    ${pid_after}=    Execute Command    pgrep tedge-agent    strip=${True}
+    Should Not Be Equal    ${pid_before}    ${pid_after}
+
 *** Keywords ***
 
 Custom Setup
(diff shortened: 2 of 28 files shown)


crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -101,7 +104,14 @@ impl Actor for SoftwareManagerActor {
 
         while let Some(request) = input_receiver.recv().await {
             tokio::select! {
-                _ = self.handle_request(request, &mut plugins, &operation_logs) => {}
+                res = self.handle_request(request, &mut plugins, &operation_logs) => {
+                    if let Err(SoftwareManagerError::Shutdown) = res {
+                        info!("Received shutdown request while handling the operation. Exiting...");
+                        // Returning an error here instead of break to exit with a non-zero exit code
+                        // return Err(RuntimeError::Shutdown);
+                        panic!("Shutting down...")
```
albinsuresh, line 112: This panic is to be replaced with a proper shutdown request sent to the actor runtime, so that it shuts down all the actors cleanly, before exiting with a non-zero exit code.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -101,7 +104,14 @@ impl Actor for SoftwareManagerActor {
 
         while let Some(request) = input_receiver.recv().await {
             tokio::select! {
-                _ = self.handle_request(request, &mut plugins, &operation_logs) => {}
+                res = self.handle_request(request, &mut plugins, &operation_logs) => {
+                    if let Err(SoftwareManagerError::Shutdown) = res {
+                        info!("Received shutdown request while handling the operation. Exiting...");
+                        // Returning an error here instead of break to exit with a non-zero exit code
+                        // return Err(RuntimeError::Shutdown);
```
albinsuresh, line 111: Just returning from here is not sufficient as it only shuts down the `SoftwareManager` actor and not the other actors. The runtime needs to catch those errors and issue shutdown requests to all the other actors as well.

tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
```diff
@@ -74,6 +73,42 @@ Refreshes mosquitto bridge configuration
     # Mosquitto should be restarted with new bridge
     Execute Command    cmd=sh -c '[ $(journalctl -u mosquitto | grep -c "Loading config file /etc/tedge/mosquitto-conf/c8y-bridge.conf") = 2 ]'
 
+Update tedge version from base to current using Cumulocity
+    Skip    msg="Until the dummy packages are included as test artefacts"
+
+    Execute Command    mkdir -p /setup/old-packages
+    Transfer To Device    /home/old-packages/*.deb    /setup/old-packages/
```
albinsuresh, line 80: Copying test debian packages from arbit locations like `/home/old-packages` is just temporary. These test packages will be checked-in along with the test via Git LFS or something.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -229,6 +243,48 @@ impl SoftwareManagerActor {
...
+                    let version_output_split: Vec<&str> =
+                        version_output.split_whitespace().collect();
+                    if let ["tedge-agent", installed_binary_version] =
+                        version_output_split.as_slice()
+                    {
+                        info!("Installed binary version: {}", installed_binary_version);
+                        if current_running_version != *installed_binary_version {
+                            info!("Self update detected. Requesting shutdown...");
+                            return Err(SoftwareManagerError::Shutdown);
+                        }
```
didier-wenzek, line 268: I would return a `bool`, just telling that a self update has been detected. The fact that a restart is needed is a solution not a diagnostic.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -229,6 +243,48 @@ impl SoftwareManagerActor {
         self.output_sender.send(response.into()).await?;
 
         self.state_repository.clear().await?;
+        Self::detect_self_update()?;
```
didier-wenzek, line 246: I would not do that inside the `handle_software_update_operation()` but at the call site.

didier-wenzek (changes requested):

didier-wenzek:
>   => The runtime has to be fixed to shutdown all the actors when one fails with an error.

I pushed 2 commits for that:
- https://github.com/thin-edge/thin-edge.io/pull/2587/commits/8f14b25995295f7de9742451a9158fc3af749914
- https://github.com/thin-edge/thin-edge.io/pull/2587/commits/ed8f5a33ec82f8faf0faefd82e9c271ecbc9917b

crates/core/tedge_agent/src/software_manager/actor.rs
albinsuresh: > This panic is catched by the runtime and has strictly the same effect as returning an error.

Hmm... But, that's not what I observed while testing. The agent process was never exiting when I was just returning the error. It was just stuck/idle after just shutting down the `SoftwareManagerActor`. But with the panic, the process was always exiting, triggering the restart as follows:

```
Jan 12 15:19:58 d5b8bb2b0f8a tedge-agent[490]: 2024-01-12T15:19:58.47165438Z  INFO tedge_agent::software_manager::actor: Self update detected. Requesting shutdown...
Jan 12 15:19:58 d5b8bb2b0f8a tedge-agent[490]: 2024-01-12T15:19:58.47166938Z  INFO tedge_agent::software_manager::actor: Received shutdown request while handling the operation. Exiting...
Jan 12 15:19:58 d5b8bb2b0f8a tedge-agent[490]: thread 'tokio-runtime-worker' panicked at 'Shutting down...', crates/core/tedge_agent/src/software_manager/actor.rs:112:25
Jan 12 15:19:58 d5b8bb2b0f8a tedge-agent[490]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Jan 12 15:19:58 d5b8bb2b0f8a systemd[1]: tedge-agent.service: Main process exited, code=killed, status=6/ABRT
Jan 12 15:19:58 d5b8bb2b0f8a systemd[1]: tedge-agent.service: Failed with result 'signal'.
Jan 12 15:20:03 d5b8bb2b0f8a systemd[1]: tedge-agent.service: Scheduled restart job, restart counter is at 1.
Jan 12 15:20:03 d5b8bb2b0f8a systemd[1]: Stopped tedge-agent.service - tedge-agent is a thin-edge.io component to support operations..
Jan 12 15:20:03 d5b8bb2b0f8a systemd[1]: Starting tedge-agent.service - tedge-agent is a thin-edge.io component to support operations....
Jan 12 15:20:03 d5b8bb2b0f8a systemd[1]: Started tedge-agent.service - tedge-agent is a thin-edge.io component to support operations..
```

> The runtime has to be fixed to shutdown all the actors when one fails with an error.

Thanks for fixing that.

crates/core/tedge_agent/src/software_manager/actor.rs
albinsuresh: Added this check specifically in this handler to limit it to software **update** requests only and **not** on software **list** requests.

crates/core/tedge_agent/src/software_manager/actor.rs
didier-wenzek: > Hmm... But, that's not what I observed while testing. The agent process was never exiting when I was just returning the error. It was just stuck/idle after just shutting down the `SoftwareManagerActor`.

Yes, this was the expected behavior before https://github.com/thin-edge/thin-edge.io/commit/ed8f5a33ec82f8faf0faefd82e9c271ecbc9917b

> But with the panic, the process was always exiting, triggering the restart as follows:

Here, I don't understand something. First, I observed something different (the panic stopping only the `SoftwareManagerActor`). Second, a test assesses the contrary https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_actors/src/runtime.rs#L523.

crates/core/tedge_agent/src/software_manager/actor.rs
didier-wenzek: > Added this check specifically in this handler to limit it to software **update** requests only and **not** on software **list** requests.

Yeah, I wondered the same. But, I think that having this check is good at the top level two reasons:
- First, this is simpler
- Second, this is even better as this will be helpful if `tedge-agent` is directly updated by-passing the software update feature.

crates/core/tedge_agent/src/software_manager/actor.rs
didier-wenzek: > Here, I don't understand something.

This what I missed: [`panic = "abort"`](https://github.com/thin-edge/thin-edge.io/blob/main/Cargo.toml#L191) for `profile.release`. I was testing the debug release.

We might have to reconsider this.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -232,6 +241,47 @@ impl SoftwareManagerActor {
...
+                        String::from_utf8_lossy(&output.stderr)
+                    )
+                    .into());
+                }
+            } else {
+                return Err(anyhow!("Failed to fetch version of installed binary").into());
+            }
+        } else {
+            return Err(anyhow!("Failed to retrieve running executable path").into());
+        }
```
Bravo555, line 280: suggestion: Rewrite using early return

IMO code nested in such a way is a bit hard to read, looking at any `return Err(...)` case, it's hard to track it to the actual condition which it covers.

Bravo555 (changes requested):

tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
```diff
@@ -74,6 +73,25 @@ Refreshes mosquitto bridge configuration
     # Mosquitto should be restarted with new bridge
     Execute Command    cmd=sh -c '[ $(journalctl -u mosquitto | grep -c "Loading config file /etc/tedge/mosquitto-conf/c8y-bridge.conf") = 2 ]'
 
+Update tedge version from base to current using Cumulocity
+    ${arch}=    Get Debian Architecture
+    Execute Command    mkdir -p /setup/base-version
+    Transfer To Device    ${CURDIR}/base-version/*${arch}.deb    /setup/base-version/
```
didier-wenzek, line 79: It would be good to add a comment on how these packages have been built, notably in case one needs to support another architecture.

didier-wenzek (commented):
Two comments:

- I agree with @Bravo555 that the `detect_self_update` function can be made clearer.
- We have to consider to set `panic = "unwind"` for the release profile. Doing so, the actors will have an opportunity the properly shutdown if one of them panics.

tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
```diff
@@ -74,6 +73,25 @@ Refreshes mosquitto bridge configuration
...
+    ${pid_before}=    Execute Command    pgrep tedge-agent    strip=${True}
+
+    # Upgrade to current version
+    Create Local Repository
+    ${OPERATION}=    Install Software    tedge,${NEW_VERSION}    tedge-mapper,${NEW_VERSION}    tedge-agent,${NEW_VERSION}    tedge-watchdog,${NEW_VERSION}    tedge-apt-plugin,${NEW_VERSION}
+
+    Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
+    ${pid_after}=    Execute Command    pgrep tedge-agent    strip=${True}
+    Should Not Be Equal    ${pid_before}    ${pid_after}
+
```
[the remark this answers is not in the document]
albinsuresh: The other reason why it was omitted was because doing the on-disk binary version check was not sufficient, as this issue itself was about the running version being different from the one on the disk. But, as an upgrade test, it makes sense to keep them both to make the test more complete.

tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
albinsuresh: Will do. It also reminded me that I haven't checked in the dummy packages for other supported platforms.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -232,6 +241,42 @@ impl SoftwareManagerActor {
...
     }
 
+    fn detect_self_update() -> Result<(), SoftwareManagerError> {
+        info!("Checking if tedge got self updated");
+        let current_running_version = env!("CARGO_PKG_VERSION");
+        info!("Current running version: {}", current_running_version);
+
+        let executable_path = std::env::current_exe()
+            .map_err(|e| anyhow!("Failed to retrieve running executable path due to {}", e))?;
+        let agent_binary_path = executable_path.parent().unwrap().join(TEDGE_AGENT);
```
rina23q, line 251: Question. I guess `executable_path` holds the path to the `tedge-agent` executable. Why do we need to go up to parent and join `tedge-agent`? Is it the support to the old executable name with underscore `tedge_agent`?

rina23q (commented):
I basically reviewed the code under tedge_agent and integration test. I got one question.

crates/core/tedge_agent/src/software_manager/actor.rs
albinsuresh: The `executable_path` points to the `tedge` binary and not the `tedge-agent` symlink file (The symlink is resolved). When the `tedge` binary is replaced, after the upgrade is performed, the `executable_path` becomes stale as it is still pointing to the old non-existent binary, even though it points to the same logical entity at the same path. You can't use that `PathBuf` variable anymore as it appears **deleted**. So, I'm rebuilding the path to make sure that it points to the latest binary itself.

tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
```diff
@@ -89,6 +91,8 @@ Update tedge version from base to current using Cumulocity
     ${OPERATION}=    Install Software    tedge,${NEW_VERSION}    tedge-mapper,${NEW_VERSION}    tedge-agent,${NEW_VERSION}    tedge-watchdog,${NEW_VERSION}    tedge-apt-plugin,${NEW_VERSION}
 
     Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
+    Device Should Have Installed Software    tedge,${PREV_VERSION}::apt    tedge_mapper,${PREV_VERSION}::apt    tedge_agent,${PREV_VERSION}::apt    tedge_watchdog,${PREV_VERSION}::apt    c8y_configuration_plugin,${PREV_VERSION}::apt    c8y_log_plugin,${PREV_VERSION}::apt    tedge_apt_plugin,${PREV_VERSION}::apt
```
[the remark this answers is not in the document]
albinsuresh: Sorry, my bad. Silly copy-paste error.

didier-wenzek (approved):
Approved

Bravo555 (approved):

Issue #2591, opened 2024-01-15
Refactor the c8y remote access plugin to use the c8y auth proxy.


didier-wenzek:
Can the c8y remote access plugin be refactored to use  c8y auth proxy instead of requesting JWT tokens? 
Can the remote access feature be built-in the c8y mapper, instead of having a plugin instead of having to configure a plugin ?

reubenmiller:
If websockets were also supported, then it would also allow usage of the following Cumulocity IoT API:

* [realtime](https://cumulocity.com/api/core/#tag/Alarm-notification-API)
* [Notification-2.0-API](https://cumulocity.com/api/core/#tag/Notification-2.0-API)

Other potential applications:
* [opcua-device-gateway](https://github.com/thin-edge/opcua-device-gateway-container) which makes use of the realtime c8y api (and not MQTT)

reubenmiller:
Test Procedure:

1. Configure the Cumulocity IoT remote access feature (e.g. web ssh is the easiest one) and connect to the device. If you can bring up a shell then that should be enough to check that it works as expected.

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the function and it's functioning according description.

PR #2597, opened 2024-01-17
Refactor the c8y remote access plugin to use the c8y auth proxy

jarhodes314:
## Proposed changes
This PR adds support for websockets to the auth proxy, and modifies the c8y remote access plugin to connect via the auth proxy, rather than making a direct connection to c8y.

There are still a few TODOs in the code, but the broad shape of the changes are in place and the happy path works for me testing locally.

## Types of changes

## Paste Link to the issue
- #2591 

## Checklist


crates/extensions/c8y_auth_proxy/Cargo.toml
```diff
@@ -29,14 +29,13 @@ tokio = { workspace = true, features = [
...
 
 [dev-dependencies]
 env_logger = { workspace = true }
 mockito = { workspace = true }
 rcgen = { workspace = true }
+rustls = { workspace = true, features = ["dangerous_configuration"] }
 tedge_http_ext = { workspace = true, features = ["test_helpers"] }
-
-[lints]
-workspace = true
```
reubenmiller, line 42: Is there a good reason that the the `lints.workspace` setting is being removed, as it was added in https://github.com/thin-edge/thin-edge.io/commit/bfe5bbe1214b38f33d7f265f8f259d9b378a66b2

crates/extensions/c8y_auth_proxy/Cargo.toml
```diff
@@ -29,14 +29,13 @@ tokio = { workspace = true, features = [
...
 ] }
+tokio-tungstenite = { version = "0.20.0", features = ["rustls-tls-native-roots"] }
 tracing = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
 env_logger = { workspace = true }
 mockito = { workspace = true }
 rcgen = { workspace = true }
+rustls = { workspace = true, features = ["dangerous_configuration"] }
```
didier-wenzek, line 40: What's this `dangerous_configuration` feature? By the way, `rustls` is not used.

crates/extensions/c8y_auth_proxy/Cargo.toml
jarhodes314: Nope, that was a hit screwup on my part. I'll fix it

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -135,23 +146,127 @@ impl FromRef<AppState> for SharedTokenManager {
...
+    http: Arc<str>,
+    ws: Arc<str>,
+}
+
+fn axum_to_tungstenite(
+    message: axum::extract::ws::Message,
+) -> tokio_tungstenite::tungstenite::Message {
+    use axum::extract::ws::CloseFrame as AxumCf;
+    use axum::extract::ws::Message as AxumM;
+    use tokio_tungstenite::tungstenite::protocol::frame::CloseFrame as TungCf;
```
didier-wenzek, line 159: It would be good to find another prefix as `Tung` confuses spell checking.

crates/extensions/c8y_auth_proxy/Cargo.toml
jarhodes314: `dangerous_configuration` allows me to disable certificate validation for the websocket connection (for testing, hence why it's in `dev-dependencies`). If we don't like this, it can be likely be changed to instead either trust the generated certificate or use a non-secure connection (although that is a little bit tricky without duplicating a bunch of test setup code).

crates/extensions/c8y_auth_proxy/src/server.rs
jarhodes314: I've now renamed this to `InCf`/`OutCf` like the other function that does essentially the same thing.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -135,23 +165,131 @@ impl FromRef<AppState> for SharedTokenManager {
...
+            if let Err(e) = res {
+                error!("Websocket error proxying messages from the client to Cumulocity: {e}");
+            }
+            c8y_to_client.abort();
+        }
+        res = (&mut c8y_to_client) => {
+            if let Err(e) = res {
+                error!("Websocket error proxying messages from Cumulocity to the client: {e}");
+            }
+            c8y_to_client.abort();
```
albinsuresh, line 277:
```suggestion
            client_to_c8y.abort();
```

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -135,23 +165,131 @@ impl FromRef<AppState> for SharedTokenManager {
...
+        }
```
albinsuresh, line 278: Query: Since this whole function doesn't return anything even while detecting one of the above failures, how would the wrapping server react? Continue functioning without the proxying capability? Wouldn't it be better to propagate these errors so that the server can choose to shutdown/panic, with the hope of restarting without that issue?

This might be out of scope for this PR specifically, as it's related to our general strategy on how we handle partial failures of a single component, in a multi-component process.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -227,28 +369,167 @@ async fn respond_to(
...
+                "/my/ws/endpoint",
+                get(|ws: WebSocketUpgrade| async { ws.on_upgrade(|_ws| ready(())) }),
+            )
+            .layer(axum::middleware::from_fn(auth(|token| {
+                token == "correct token"
+            })));
+        let (server, port) = axum_server();
+        tokio::spawn(server.serve(test_app.into_make_service()));
+        let proxy_port = start_server_port(port, vec!["outdated token", "correct token"]);
+        let (mut ws, _) = connect_to_websocket_port(proxy_port).await;
```
albinsuresh, line 451: Looks like a block of code repeated in the previous test as well, that can be moved to a test setup function that takes the `Router` definition as an argument.

crates/extensions/c8y_auth_proxy/src/actor.rs
```diff
@@ -41,8 +41,9 @@ impl C8yAuthProxyBuilder {
         config: &TEdgeConfig,
         jwt: &mut ServerActorBuilder<C8YJwtRetriever, Sequential>,
     ) -> anyhow::Result<Self> {
-        let app_state = AppState {
-            target_host: format!("https://{}", config.c8y.http.or_config_not_set()?).into(),
+        let app_state = AppData {
```
albinsuresh, line 44: Minor:
```suggestion
        let app_data = AppData {
```

albinsuresh (commented):
I'm no expert of either the `axum` or the `tungstenite` crates. But the connection/conversion logic looked sensible. Just spotted one minor bug in the conversion task error handler.

crates/extensions/c8y_auth_proxy/Cargo.toml
didier-wenzek: > If we don't like this

That's okay.

plugins/c8y_remote_access_plugin/src/proxy.rs
```diff
@@ -94,6 +91,7 @@ impl Websocket {
         let socket = async_tungstenite::tokio::connect_async(request)
             .await
             .into_diagnostic()
+            .with_context(|| format!("host {url}"))
```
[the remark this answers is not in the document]
jarhodes314: Each time you call either, it adds a layer to the backtrace:

```rust
In: use anyhow::Context;
In: format!("{:?}", err.with_context(|| format!("Reading {:?}", "not a file")).context("Reading configuration").unwrap_err())
Out: "Reading configuration\n\nCaused by:\n    0: Reading \"not a file\"\n    1: No such file or directory (os error 2)"
```

crates/extensions/c8y_auth_proxy/src/server.rs
jarhodes314: I'm not sure *exactly* what axum does internally, but my belief is that this is a separate task spawned when we call [`ws.on_upgrade`](https://docs.rs/axum/0.6.20/axum/extract/ws/struct.WebSocketUpgrade.html#method.on_upgrade). In any case, it only affects the current connection, nothing else, and I believe this is the most desirable behaviour. If we were talking over HTTP, this is a point at which I would give up and respond with `500 Internal Server Error` to the current request, but I wouldn't want the server to stop running, particularly as the issue at this point is quite likely one with Cumulocity.

crates/extensions/c8y_auth_proxy/src/server.rs
jarhodes314: You're right, I should add unit tests for those.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -135,23 +165,131 @@ impl FromRef<AppState> for SharedTokenManager {
...
+                Err(e) => Err(anyhow::Error::from(e)),
+            }
+        }
+    }
+    .expect("Error connecting to proxied websocket");
+    let (mut to_c8y, mut from_c8y) = c8y.split();
+    let (mut to_client, mut from_client) = ws.split();
+    let mut client_to_c8y = tokio::spawn(async move {
+        while let Some(Ok(msg)) = from_client.next().await {
+            to_c8y.send(axum_to_tungstenite(msg)).await.unwrap();
```
didier-wenzek, line 257: Any error on forwarding a packet should be returned by `client_to_c8y`.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -135,23 +165,131 @@ impl FromRef<AppState> for SharedTokenManager {
...
+        }
+    });
+    let mut c8y_to_client = tokio::spawn(async move {
+        while let Some(Ok(msg)) = from_c8y.next().await {
+            to_client.send(tungstenite_to_axum(msg)).await.unwrap();
```
didier-wenzek, line 262: Any error on forwarding a packet should be returned by `c8y_to_client`.

crates/extensions/c8y_auth_proxy/src/server.rs
jarhodes314: I don't quite understand the value of doing this. It makes failing here more difficult, and it makes handling the error more complicated as we (a) should handle the JoinError (the panic case) anyway and (b) have to remember to handle the inner error additionally to that.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -116,15 +122,39 @@ fn try_bind_with_tls(
...
+    pub is_https: bool,
+    pub host: String,
     pub token_manager: SharedTokenManager,
 }
 
+#[derive(Clone)]
+struct AppState {
+    target_host: TargetHost,
+    token_manager: SharedTokenManager,
+}
```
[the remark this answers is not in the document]
jarhodes314: `AppData` is equivalent to a data struct/dataclass, to effectively give the caller named arguments. It's converted into `AppState` so we only generate the https/wss URLs once.

crates/extensions/c8y_auth_proxy/src/server.rs
jarhodes314: Now done

didier-wenzek (approved):
Approved. Working as expected and nicely done.

Issue #2568, opened 2024-01-05
Allow arguments/subcommands in custom operations


scfx:
**Is your feature improvement request related to a problem? Please describe.**
If my tests were correct, there is currently no possibility to have any whitespaces in the  [custom operation commands](https://thin-edge.github.io/thin-edge.io/operate/c8y/supported_operations/#working-with-custom-operations). So, having custom operations like:

```
[exec]
  topic = "c8y/s/ds"
  on_message = "511"
  command = "python /etc/tedge/operations/command.py"
  timeout = 10
```

will not work, as the thin-edge will try to find end execute the file "python3 /etc/tedge/operations/command.py"

**Describe the solution you'd like**
I would like to see an exec structure similar to what Docker uses. So:

```
[exec]
  topic = "c8y/s/ds"
  on_message = "511"
  command = ["python", "/etc/tedge/operations/command.py"]
  timeout = 10
```
If you try to do this in version (0.13.0), Thin-Edge won't provide any supported operations. I haven't tried it in 1.0.0-rc1, so apologies if that is already fixed.

reubenmiller:
@didier-wenzek Would you be open to supporting this for 1.0.0?

didier-wenzek:
> @didier-wenzek Would you be open to supporting this for 1.0.0?

Yes, this can be done, using [shell-words](https://lib.rs/crates/shell-words), to allow plain commands as in `command = "python /etc/tedge/operations/command.py"`

reubenmiller:
thin-edge.io now supports commands with multiple arguments (in a string form):

```toml
command = "python /etc/tedge/operations/command.py"
```

**System tests**

* [custom_operation_multiple_args/operation_multiple_arguments_in_command.robot](https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_multiple_args/operation_multiple_arguments_in_command.robot)

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_multiple_args/operation_multiple_arguments_in_command.robot
- [x] QA has tested the function and it's functioning according description.

AjayItHub:
[...]

PR #2602, opened 2024-01-18
Add support for commands with multiple arguments in custom operations

Ruadhri17:
## Proposed changes
This PR adds support for arguments/subcommands when running command in custom operation. Command accepts string with arguments separated by whitespaces, e.g:
```
command = "python /etc/tedge/operations/command.py"
```
More in #2568.

## Types of changes

## Paste Link to the issue
* #2568

## Checklist

### DIFF
--- a/crates/common/logged_command/src/logged_command.rs
+++ b/crates/common/logged_command/src/logged_command.rs
@@ -148,23 +148,40 @@ impl std::fmt::Display for LoggedCommand {
 }
 
 impl LoggedCommand {
-    pub fn new(program: impl AsRef<OsStr>) -> LoggedCommand {
+    pub fn new(program: impl AsRef<OsStr>) -> Result<LoggedCommand, std::io::Error> {
         let command_line = match program.as_ref().to_str() {
             None => format!("{:?}", program.as_ref()),
             Some(cmd) => cmd.to_string(),
         };
 
-        let mut command = Command::new(program);
+        let mut args = shell_words::split(&command_line)
+            .map_err(|err| std::io::Error::new(std::io::ErrorKind::InvalidData, err))?;
+
+        let mut command = match args.len() {
+            0 => {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::InvalidData,
+                    "command line is empty.",
+                ))
+            }
+            1 => Command::new(&args[0]),
+            _ => {
+                let mut command = Command::new(args.remove(0));
+                command.args(&args);
+                command
+            }
+        };
+
         command
             .current_dir("/tmp")
             .stdin(Stdio::piped())
             .stdout(Stdio::piped())
             .stderr(Stdio::piped());
 
-        LoggedCommand {
+        Ok(LoggedCommand {
             command_line,
             command,
-        }
+        })
     }
 
     pub fn arg(&mut self, arg: impl AsRef<OsStr>) -> &mut LoggedCommand {
@@ -259,7 +276,7 @@ mod tests {
         let mut logger = BufWriter::new(log_file);
 
         // Prepare a command
-        let mut command = LoggedCommand::new("echo");
+        let mut command = LoggedCommand::new("echo").unwrap();
         command.arg("Hello").arg("World!");
 
         // Execute the command with logging
@@ -292,7 +309,7 @@ EOF
         let mut logger = BufWriter::new(log_file);
 
         // Prepare a command that triggers some content on stderr
-        let mut command = LoggedCommand::new("ls");
+        let mut command = LoggedCommand::new("ls").unwrap();
         command.arg("dummy-file");
 
         // Execute the command with logging
@@ -326,7 +343,7 @@ EOF
         let mut logger = BufWriter::new(log_file);
 
         // Prepare a command that cannot be executed
-        let command = LoggedCommand::new("dummy-command");
+        let command = LoggedCommand::new("dummy-command").unwrap();
 
         // Execute the command with logging
         let _ = command.execute(&mut logger).await;
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -793,7 +793,13 @@ impl CumulocityConverter {
         let command = command.to_owned();
         let payload = payload.to_string();
 
-        let mut logged = LoggedCommand::new(&command);
+        let mut logged =
+            LoggedCommand::new(&command).map_err(|e| CumulocityMapperError::ExecuteFailed {
+                error_message: e.to_string(),
+                command: command.to_string(),
+                operation_name: operation_name.to_string(),
+            })?;
+
         logged.arg(&payload);
 
         let maybe_child_process =
--- a/tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_multiple_args/operation_multiple_arguments_in_command.robot
+++ b/tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_multiple_args/operation_multiple_arguments_in_command.robot
@@ -1,0 +1,26 @@
+*** Settings ***
+Resource            ../../../../resources/common.resource
+Library             Cumulocity
+Library             ThinEdgeIO
+
+Test Setup          Custom Setup
+Test Teardown       Get Logs
+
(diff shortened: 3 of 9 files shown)


tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_multiple_args/operation_multiple_arguments_in_command.robot
```diff
@@ -0,0 +1,26 @@
...
+Test Teardown       Get Logs
+
+Test Tags           theme:c8y    theme:operation    theme:custom
+
+
+*** Test Cases ***
+Run the custom operation with multiple arguments
+    ${operation}=    Cumulocity.Create Operation    description=do something    fragments={"c8y_Command":{"text":""}}
+    Operation Should Be SUCCESSFUL    ${operation}
+    Should Be Equal    ${operation.to_json()["c8y_Command"]["result"]}    command 1\n
```
albinsuresh, line 16:
```suggestion
    Should Be Equal    ${operation}.to_json()["c8y_Command"]["result"]    command 1\n
```
Not really a request for change but I'm just surprised that the current syntax is working.

albinsuresh (approved):
LGTM

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -709,7 +709,19 @@ impl CumulocityConverter {
...
 
-        let mut logged = LoggedCommand::new(&command);
+        let mut args =
+            shell_words::split(&command).map_err(|e| CumulocityMapperError::ExecuteFailed {
+                error_message: e.to_string(),
+                command: command.to_string(),
+                operation_name: operation_name.to_string(),
+            })?;
+
+        let mut logged = LoggedCommand::new(args.remove(0));
```
rina23q, line 719: This line causes a panic if user didn't give anything in the custom command file since `args` is empty vector.

```
[exec]
topic = "c8y/s/ds"
on_message = "511"
command = ""
```

rina23q (changes requested):
Panic must be avoided.

Bravo555 (changes requested):
blocker: update any relevant documentation [(example 1)](docs/src/operate/c8y/supported_operations.md)

Besides the panic that Rina mentioned, I think it would be the best to address the documentation as part of this PR. We need to keep in mind what are the limitations of [`shell_words`](https://docs.rs/shell-words/latest/shell_words/fn.split.html):

> The result is exactly the same as one obtained from Unix shell as long as those unsupported features are not present in input: no operators, no variable assignments, no tilde expansion, no parameter expansion, no command substitution, no arithmetic expansion, no pathname expansion.
> 
> In case those unsupported shell features are present, the syntax that introduce them is interpreted literally.

This can potentially confusing to users, as error message `Operation execution failed: No such file or directory (os error 2). Command: ~/../etc/tedge/operations/command_1.sh. Operation name: c8y_Command` one might think that the file doesn't exist instead of the tilde character not getting expanded. Granted most users shouldn't do that kind of thing anyway, so I'd say error messages are enough, but this behaviour needs to be described in the documentation.

crates/common/logged_command/src/logged_command.rs
```diff
@@ -260,15 +272,15 @@ mod tests {
...
+        command.args(["Hello", "World!"]);
 
         // Execute the command with logging
         let _ = command.execute(&mut logger).await;
 
         let log_content = String::from_utf8(std::fs::read(log_file_path)?)?;
         assert_eq!(
             log_content,
-            r#"----- $ echo "Hello" "World!"
+            r#"----- $ echo ["Hello", "World!"]
```
reubenmiller, line 283: The assert is unexpected, as it is printed invalid shell output (making it harder for users to debug by copying/pasting the output)...

```
$ echo Hello World
```

Alternatively you could not try to replicate a shell output and just use the following format (which does not to mix shell with Rust formatted output).

```
cmd=echo, args=["Hello", "World!"]
```

Though the above example is not so user friendly if people want to copy/paste the command to help with debugging (by manually running the script and checking the output).

reubenmiller: It might be worthwhile checking out the workflow actor does the logging of the commands and using the same convention.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -709,7 +709,19 @@ impl CumulocityConverter {
...
+
+        if !args.is_empty() {
+            logged.args(&args);
+        }
+
         logged.arg(&payload);
```
didier-wenzek, line 725: Beware, that the behavior is not obvious and has to be documented: the smart rest payload is *always* added as the very last argument.

The `command = "python /etc/tedge/operations/command.py"` actually leads to the execution of `python /etc/tedge/operations/command.py $SMART_REST_PAYLOAD`.

didier-wenzek (commented):
The changes can be simpler by moving the logic to parse the command line into the `logged_command` crate.

Instead of introducing an `args` method to mutate a `LoggedCommand` that tries to rebuild the command line from where the arguments have been extracted, it would be far simpler to pass the command line  to `LoggedCommand::new()` and do the split there.

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_multiple_args/operation_multiple_arguments_in_command.robot
reubenmiller: No the current syntax is correct...RobotFramework variables use the syntax `${    }` to declare that it is a variable. Inside the this syntax is straight python code, `operation.to_json()["c8y_Command"]["result"]`

crates/common/logged_command/src/logged_command.rs
```diff
@@ -148,23 +148,34 @@ impl std::fmt::Display for LoggedCommand {
...
         let command_line = match program.as_ref().to_str() {
             None => format!("{:?}", program.as_ref()),
             Some(cmd) => cmd.to_string(),
         };
 
-        let mut command = Command::new(program);
+        let mut args = shell_words::split(&command_line)?;
+
+        let mut command = match args.len() {
+            0 => Command::new(program),
```
didier-wenzek, line 160: This should return an `Err(shell_words::ParseError)`, as this occurs with an empty command line.

Ruadhri17: I was thinking about that, though in previous implementation user was allowed to pass empty command line so I wanted to preserve that behavior.

didier-wenzek (approved):
Approved

crates/common/logged_command/src/logged_command.rs
```diff
@@ -148,23 +148,34 @@ impl std::fmt::Display for LoggedCommand {
...
         let command_line = match program.as_ref().to_str() {
             None => format!("{:?}", program.as_ref()),
             Some(cmd) => cmd.to_string(),
         };
 
-        let mut command = Command::new(program);
+        let mut args = shell_words::split(&command_line)?;
+
+        let mut command = match args.len() {
+            0 => return Err(shell_words::ParseError),
```
Bravo555, line 160: The `Display` impl for `shell_words::ParseError` is `"missing closing quote"` which is very much not the case here. We should use our own error with a correct message.

Bravo555 (changes requested):

crates/common/logged_command/src/logged_command.rs
Bravo555: Also one more thing, need to add `time` feature to `logged_command/Cargo.toml`, because without it's impossible to run tests for the package.

Bravo555 (approved):

crates/common/logged_command/src/logged_command.rs
```diff
@@ -148,23 +148,40 @@ impl std::fmt::Display for LoggedCommand {
 }
 
 impl LoggedCommand {
-    pub fn new(program: impl AsRef<OsStr>) -> LoggedCommand {
+    pub fn new(program: impl AsRef<OsStr>) -> Result<LoggedCommand, std::io::Error> {
```
rina23q, line 151: Very minor: I think `try_new()` is a better name as the function now returns a `Result`.

rina23q (approved):
Approved

Issue #1352, opened 2022-08-19
Use Softwaretype according to new Advanced Software Management in 10.14


mbay-ODW:
**Is your feature improvement request related to a problem? Please describe.**
The current Software plugin concept uses ::{type}} in the version. Since 10.14 the software type does exist which could/should be used.

![image](https://user-images.githubusercontent.com/58465677/185593865-0544923f-b265-4aec-b61e-a0ed1e22a4ff.png)

AndrejS321:
great suggestion, we will review this improvement and decide when to work on this

reubenmiller:
**Proposed changes**

In Cumulocity 10.14, the software management concept was extended to included types, however at the time of the thin-edge.io software management support, this feature did not exist (or was very new at least).

The software management mode should be configurable to control which sets of Cumulocity IoT SmartREST templates are used to either parse the software update operation, or to send the list of installed software to Cumulocity (via MQTT). The following lists the modes.

**Note:** If the software list is larger, then tedge falls back to sending the software list via the Cumulocity REST API instead of MQTT. The json body example is also shown for each mode.

**Mode 1: Software Management templates (Current implementation)**

|Action|SmartREST 2.0 template|
|---|----|
|Set software list (mqtt)|[116](https://cumulocity.com/guides/reference/smartrest-two/#116)|
|Set software list (REST API)|`{"c8y_SoftwareList":[{"name":"a","version":"1.0.0","type":"debian","url":""}]}`|
|Parse Operation|[528](https://cumulocity.com/guides/reference/smartrest-two/#528)|


**Mode 2: Advanced Software Management templates (New)**

|Action|SmartREST 2.0 template|
|---|----|
|Set software list|[140](https://cumulocity.com/guides/reference/smartrest-two/#140)|
|Set software list (REST API)|`{"c8y_SoftwareList":[{"name":"a","version":"1.0.0::debian","url":""}]}`|
|Parse Operation|[529](https://cumulocity.com/guides/reference/smartrest-two/#529)|

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/software_management/software.robot
- [x] QA has tested the function and it's functioning according description.

PR #2613, opened 2024-01-22
Support software type for c8y advanced software management in 10.14

rina23q:
## Proposed changes
Since the version 10.14, c8y starts supporting "type" field for software list and "softwareType" field for software update operation.

Change the Rust code accordingly:
1) Software list
Stop combining version and software type with the connector "::". The module type will be declared as "type" in c8y JSON and the version field holds its pure version only.

2) Software update
Different behaviour if a received operation contains "softwareType" field or not. If yes, softwareType will be used as its module type and version will be used as it is. If no, mapper tries to get the module type from the received version field as we initially implemented.

Todo:

- [x] Rust change
- [x] Robot test change
- [x] Doc change

## Types of changes

## Paste Link to the issue
#1352 

## Checklist


crates/core/c8y_api/src/json_c8y_deserializer.rs
```diff
@@ -277,14 +277,17 @@ impl C8ySoftwareUpdate {
...
-            let version = if self.version.is_empty() || self.version == " " {
-                None
-            } else {
-                Some(self.version.clone())
-            };
-            let module_type = (!module_type.is_empty()).then(|| module_type.clone());
-            return (version, module_type);
+        match &self.software_type {
+            Some(module_type) if !module_type.is_empty() => {
+                let version = if self.version.is_empty() || self.version == " " {
```
reubenmiller, line 282: Wouldn't it be better to trim the whitespace in the version, then check if it is empty?

rina23q: Your suggestion makes the logic smarter. On the other hand, having one space in the version " " comes from our specs.

https://thin-edge.github.io/thin-edge.io/start/software-management/#adding-new-software-into-the-software-repository-of-your-tenant
> thin-edge.io ships a default plugin supporting debian packages from both apt repositories as well as remote locations. If you prefer to use packages from an apt repository, select the Provide a file path option and give an empty space (' ').

So, I think the current implementation makes sense to tell that we address `" "` as special case.

reubenmiller: I think we need to revisit that spec...a magic single space is very fragile and makes absolutely no sense

reubenmiller: In some languages (e.g. c#) there is even a string method to check for a null or whitespace string, `String.IsNullOrWhiteSpace()`

tests/RobotFramework/tests/cumulocity/self-update/tedge_self_update.robot
```diff
@@ -34,14 +35,24 @@ Update tedge version from previous using Cumulocity
     Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=180
 
     # Software list reported by the former agent, which is still running
```
didier-wenzek, line 37: The comment can be extended

```suggestion
    # Software list reported by the former agent, which is still running
    # but formatted with by the c8y-mapper, which has just been installed
```

crates/core/c8y_api/src/json_c8y_deserializer.rs
didier-wenzek: > Wouldn't it be better to trim leading and trailing whitespace in the version, then check if it is empty?

Indeed better to use `self.version.as_ref().trim().is_empty()` here.

Accepting more than one space simply makes things less error prone for the users, without any contradiction with  the docs telling one space is enough as a workaround for fields that are mandatory on the GUI. Accepting `"NA"` for that purpose might be more explicit.

didier-wenzek (commented):
> Doc change (need to check)?

Yes, the doc has to be updated.

The sentence *"thin-edge.io ships a default plugin supporting debian packages from both apt repositories as well as remote locations. If you prefer to use packages from an apt repository, select the Provide a file path option and give an empty space (' ')."* is really unclear.

I would simply tell that if no version or no path are applicable then a string of empty spaces are accepted by thin-edge meaning that this field is non applicable. We can stress that this is the case for the apt plugin when one wants to install the latest version of the package  (.i.e. no version) using the package repository configured on the device (i.e. no path/url).

@reubenmiller does it make sense to accept the string `"NA"` as well as white-spaces to mean that the field is non applicable?

crates/core/c8y_api/src/json_c8y_deserializer.rs
rina23q: To change the code, I need to make sure what is the expected behaviour. Here I just sumarized what we have now in this PR.

a. software version (mandatory in UI)
* (as of now) If user sets to "latest" or an empty space (' '), both get interpreted to "latest" version to be used. ([reference](https://thin-edge.github.io/thin-edge.io/start/software-management/#adding-new-software-into-the-software-repository-of-your-tenant)) 

b. software type (mandatory in UI since 10.14)
* (new in this PR) If software type is not given or having no value (''), falling back to `::` logic. No special care for an empty space.

c. software URL (mandatory in UI)
* (as of now) An empty space (' ') gets interpreted to use the device's default repository.

From the discussion so far, we will trim empty spaces (so it will not matter if it's ONE space or more). But, should we still mention about the empty space stuff in our user guide?

didier-wenzek: > From the discussion so far, we will trim empty spaces (so it will not matter if it's ONE space or more).

Perfect

> But, should we still mention about the empty space stuff in our user guide?

Yes, we should. This is notably useful when updating a package using the package repository configured on the device.

reubenmiller:
> > Doc change (need to check)?
> 
> Yes, the doc has to be updated.
> 
> The sentence _"thin-edge.io ships a default plugin supporting debian packages from both apt repositories as well as remote locations. If you prefer to use packages from an apt repository, select the Provide a file path option and give an empty space (' ')."_ is really unclear.
> 
> I would simply tell that if no version or no path are applicable then a string of empty spaces are accepted by thin-edge meaning that this field is non applicable. We can stress that this is the case for the apt plugin when one wants to install the latest version of the package (.i.e. no version) using the package repository configured on the device (i.e. no path/url).
> 
> @reubenmiller does it make sense to accept the string `"NA"` as well as white-spaces to mean that the field is non applicable?

I don't think it is worth it...from an API view, just leaving out the url is fine. I believe the requirement to use whitespace in the url is only a UI form validation issue...so we could put a change request to remove the UI validation to also allow an option to not treat a missing url as an error.

crates/core/c8y_api/src/json_c8y_deserializer.rs
rina23q: Applied in 4188f534f8a8dad27c5413bab850b0f3d326615a

didier-wenzek (approved):
Approved

Issue #2584, opened 2024-01-11
Services are shown as child devices


mstoffel-sag:
Services are shown as child devices

![image](https://github.com/thin-edge/thin-edge.io/assets/29426188/f1abe63f-ad2e-4620-befa-b9890ce30eda)

It's weird though that the bridge is show correctly as a service:

![image](https://github.com/thin-edge/thin-edge.io/assets/29426188/ea1b33d1-4f9d-4fca-8429-396ed50bbff7)

**To Reproduce**
I used the tedge-rugpi-image repo to create a pi3 image, lashed it to an sd card and followed the instruction to register the device. 

**Expected behavior**
After registration there are child device that should be displayed as services.

**Environment (please complete the following information):**
 - OS rugpi image release [20231221.2255](https://github.com/thin-edge/tedge-rugpi-image/releases/tag/20231221.2255)
 - pi3b
 - arm64
 - thin-edge.io 
 
![image](https://github.com/thin-edge/thin-edge.io/assets/29426188/3e41cbc1-4510-44d4-8f8c-68257847673d)

reubenmiller:
@mstoffel-sag Can you please add the output of the MQTT messages using the following command (wait 5 seconds, then kill the command, then post the console output)

```sh
tedge mqtt sub 'te/+/+/+/+'
```

We suspect that there is something incorrect with the order of messages which is causing the services to be registered a child devices.

mstoffel-sag:
Here's the output:

root@rpi3-b827eb596700 ~ # tedge mqtt sub 'te/+/+/+/+'
INFO: Connected
[te/device/main/service/tedge-mapper-collectd] {"@parent":"device/main//","@type":"service","type":"service"}
[te/device/main/service/tedge-agent] {"@parent":"device/main//","@type":"service","type":"service"}
[te/device/main/service/tedge-mapper-c8y] {"@parent":"device/main//","@type":"service","type":"service"}
[te/device/main/service/mosquitto-c8y-bridge] {"@id":"rpi3-b827eb596700:device:main:service:mosquitto-c8y-bridge","@parent":"device/main//","@type":"service","name":"mosquitto-c8y-bridge","type":"service"}

reubenmiller:
Thanks @mstoffel-sag. The team has been able to reproduce the issue and analyzing the results.

Bravo555:
[...]

PR #2620, opened 2024-01-24
fixed services shown as child devices #2584

### LINKED ISSUE #2606: creating a workflow file creates a warning log message and child device in tedge-mapper-c8y
**Describe the bug**

Creating a workflow file under the `/etc/tedge/operations` folder, results in an erroneous errors message in the tedge-mapper-c8y as the mapper things this should be a customer operation and not a custom workflow definition.

The [workflow definition](https://thin-edge.github.io/thin-edge.io/references/agent/operation-workflow/#user-defined-operation-workflow) describes the process where the workflow definitions are to be kept.

Below shows the warning log message in question:

```
Jan 20 19:58:33 rpi4-d83add90fe56 tedge-mapper[1306]: 2024-01-20T19:58:33.717707703Z  WARN c8y_api::smartrest::operations: Ignoring custom operation definition as the filename uses an invalid character. Only [A-Za-z0-9_] characters are accepted. file=/etc/tedge/operations/firmware_update.toml
```

**To Reproduce**

1. Create a dummy workflow file

    ```sh
    touch /etc/tedge/operations/my_workflow.toml
    ```
2. Check the tedge-mapper-c8y logs

    ```sh
    journalctl -u tedge-mapper-c8y -n 100
    ```

**Expected behavior**

Creating a valid workflow (as per the documentation), should not invoke the custom operation handler logic, and should not log a warning message.

**Screenshots**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`Raspberry Pi 4 Model B Rev 1.5`|
|System-Architecture|`Linux rpi4-d83add90fe56 6.1.0-rpi7-rpi-v8 #1 SMP PREEMPT Debian 1:6.1.63-1+rpt1 (2023-11-24) aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.0.0-rc.2~116+g02d20ea`|

**Additional context**

Bravo555:
## TODO

- [x] add tests

## Proposed changes

Added a workaround for #2584 by changing status update smartrest message to 102.

Also fixed #2606 by properly checking in the fs event handler if the path of the event is actually inside the c8y operations path.

## Types of changes

## Paste Link to the issue
- #2584

## Checklist

## Further comments

Note that changing `104` to `102` in this case is only a workaround, because, fundamentally, Cumulocity maintains a certain state, and depending on this state, we either can or can't send certain messages, e.g. `104` can't be sent before a `102` or it will create a child device, but I assume this is only a single example and there are many more. To be able to use certain messages, we need to actually maintain the state for entities in cumulocity - if they're registered, if they're child devices or services, etc. So ultimately we may need to change our model a bit from "map every local MQTT message to a C8y message" to "every local MQTT message may modify entity store state and if this state is different we need to sync it with C8y".

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/dynamic_discovery.rs
+++ b/crates/extensions/c8y_mapper_ext/src/dynamic_discovery.rs
@@ -28,6 +28,7 @@ pub enum DynamicDiscoverOpsError {
 }
 
 pub fn process_inotify_events(
+    c8y_operations_dir: &Path,
     path: &Path,
     mask: FsWatchEvent,
 ) -> Result<Option<DiscoverOp>, DynamicDiscoverOpsError> {
@@ -39,6 +40,11 @@ pub fn process_inotify_events(
     let parent_dir = path
         .parent()
         .ok_or_else(|| DynamicDiscoverOpsError::NotAnOperation(path.to_path_buf()))?;
+
+    // only files inside `/etc/tedge/operations/c8y` directory are valid c8y operations
+    if !parent_dir.starts_with(c8y_operations_dir) {
+        return Ok(None);
+    }
 
     if is_valid_operation_name(operation_name) {
         match mask {
--- a/crates/extensions/c8y_mapper_ext/src/service_monitor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/service_monitor.rs
@@ -4,6 +4,7 @@ use serde::Serialize;
 use tedge_api::entity_store::EntityMetadata;
 use tedge_api::entity_store::EntityType;
 use tedge_mqtt_ext::Message;
+use tracing::error;
 
 #[derive(Deserialize, Serialize, Debug, Default)]
 pub struct HealthStatus {
@@ -25,13 +26,11 @@ pub fn convert_health_status_message(
         return vec![];
     }
 
-    let mut mqtt_messages: Vec<Message> = Vec::new();
-
     // If not Bridge health status
     // FIXME: can also match "device/bridge//" or "/device/main/service/my_custom_bridge"
     // should match ONLY the single mapper bridge
     if entity.topic_id.as_str().contains("bridge") {
-        return mqtt_messages;
+        return vec![];
     }
 
     let HealthStatus {
@@ -42,16 +41,27 @@ pub fn convert_health_status_message(
         health_status = "unknown".into();
     }
 
-    // FIXME: `ancestors_external_ids` gives external ids starting from the parent, but for health
-    // we need XID of current device as well
-    let mut external_ids = vec![entity.external_id.as_ref().to_string()];
-    external_ids.extend_from_slice(ancestors_external_ids);
-    let status_message =
-        smartrest::inventory::service_status_update_message(&external_ids, &health_status);
+    let display_name = entity
+        .other
+        .get("name")
+        .and_then(|v| v.as_str())
+        .or(entity.topic_id.default_service_name())
+        .unwrap_or(entity.external_id.as_ref());
 
-    mqtt_messages.push(status_message);
+    let display_type = entity
+        .other
+        .get("type")
+        .and_then(|v| v.as_str())
+        .expect("display type should be inserted for every service in the converter");
 
-    mqtt_messages
+    let Ok(status_message) =
+        // smartrest::inventory::service_status_update_message(&external_ids, &health_status);
+        smartrest::inventory::service_creation_message(entity.external_id.as_ref(), display_name, display_type, &health_status, ancestors_external_ids) else {
+            error!("Can't create 102 for service status update");
+            return vec![];
+        };
+
+    vec![status_message]
 }
 
 #[cfg(test)]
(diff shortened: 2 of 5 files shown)


crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -42,16 +41,20 @@ pub fn convert_health_status_message(
...
-    mqtt_messages.push(status_message);
+    let display_name = entity
+        .other
+        .get("name")
+        .and_then(|v| v.as_str())
+        .or(entity.topic_id.default_service_name())
+        .unwrap_or(entity.external_id.as_ref());
+    let Ok(status_message) =
+        // smartrest::inventory::service_status_update_message(&external_ids, &health_status);
+        smartrest::inventory::service_creation_message(entity.external_id.as_ref(), display_name, entity.r#type.as_str(), &health_status, ancestors_external_ids) else {
```
albinsuresh, line 52:
```suggestion
        smartrest::inventory::service_creation_message(entity.external_id.as_ref(), display_name, display_type, &health_status, ancestors_external_ids) else {
```
The service type also needs to be derived from `entity.other.get("type")` like you've done for `display_name`.

crates/extensions/c8y_mapper_ext/src/dynamic_discovery.rs
```diff
@@ -40,6 +42,11 @@ pub fn process_inotify_events(
         .parent()
         .ok_or_else(|| DynamicDiscoverOpsError::NotAnOperation(path.to_path_buf()))?;
 
+    // only files inside `/etc/tedge/operations/c8y` directory are valid c8y operations
```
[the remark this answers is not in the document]
Bravo555: The watcher is watching `/etc/tedge/operations/c8y` directory itself as well, so if there are events related to the directory, we get a `/etc/tedge/operations` parent and `c8y` as filename, which is wrong. It may have changed when I upgraded the fs notification library for the debouncing PR. I think it's good to not rely that fsnotify will give us proper paths and have this guard here just to be sure.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
Bravo555: You're right, I confused the types there.

Bravo555:
Created tests for with and without device autostart, should be ready for review now.

tests/RobotFramework/tests/cumulocity/bootstrap.robot
```diff
@@ -0,0 +1,37 @@
...
+    Sleep    5s
+
+    # Assert that there are no child devices present.
+    Cumulocity.Device Should Not Have Any Child Devices
+
+No unexpected child devices created without service autostart
+    [Tags]    \#2606
+    ${DEVICE_SN}=    Setup
+    Device Should Exist    ${DEVICE_SN}
+
```
didier-wenzek, line 32: I would add the following as this is a way to trigger https://github.com/thin-edge/thin-edge.io/issues/2584

```
    # Touching the operations directories should not create child devices
    Execute Command    touch /etc/tedge/operations
    Execute Command    touch /etc/tedge/operations/c8y
```

didier-wenzek (commented):
This PR is in a good shape to be merged, except that it would be good to explore first [this other alternative(https://github.com/thin-edge/thin-edge.io/pull/2624) to fix the main [issue with unexpected child devices](https://github.com/thin-edge/thin-edge.io/issues/2584).

That said, this PR also fixes https://github.com/thin-edge/thin-edge.io/issues/2606 and cannot be fully replaced by https://github.com/thin-edge/thin-edge.io/pull/2624

Bravo555:
Trying to make sure that the tests fail without the fix, but I'm not able to trigger the fs event `operations` issue from within the test. Attaching to the shell and looking at `journalctl -u tedge-mapper-c8y`, `touch`ing the file, looks like no fs events are emitted inside the test container, but installing `inotify-tools` and `inotifywatch /etc/tedge/operations/c8y/` correctly detects the changes when i `touch` the file. Can anybody confirm similar behaviour?

didier-wenzek (approved):
Approved.

We decided to include this fix combined with a way to re-publish on start all the registration messages.

albinsuresh (approved):

didier-wenzek:
> Trying to make sure that the tests fail without the fix, but I'm not able to trigger the fs event `operations` issue from within the test. Attaching to the shell and looking at `journalctl -u tedge-mapper-c8y`, `touch`ing the file, looks like no fs events are emitted inside the test container, but installing `inotify-tools` and `inotifywatch /etc/tedge/operations/c8y/` correctly detects the changes when i `touch` the file. Can anybody confirm similar behaviour?

Unfortunately, the test doesn't fail without the fix: see https://github.com/thin-edge/thin-edge.io/pull/2627.
Touching the operations directory is not enough to trigger the unexpected child-device creation (in the environment used by the tests - while this is the case on a linux box).

Issue #2623, opened 2024-01-25
flaky test: Update tedge version from base to current using Cumulocity


reubenmiller:
**Describe the bug**

There is a flaky system tests:

* Update tedge version from base to current using Cumulocity

Failed runs are seen here:

* https://github.com/thin-edge/thin-edge.io/actions/runs/7654472811
* https://github.com/thin-edge/thin-edge.io/actions/runs/7651705932
* https://github.com/thin-edge/thin-edge.io/actions/runs/7638473078
* https://github.com/thin-edge/thin-edge.io/actions/runs/7636467771

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [e.g. 0.1.0]

**Additional context**

reubenmiller:
The following contains log entries which show the missing `503` message which results in the software update operation remaining in the *EXECUTING* state after the agent has performed the update.

**Case: Failed**

In the failed case, the `503,c8y_SoftwareUpdate` message is not sent before the **tedge-agent** process exits, and the process does not attempt to send a message when it starts up.

[full failed.log](https://github.com/thin-edge/thin-edge.io/files/14055534/failed.log)

```log
Jan 25 08:33:27 03d75d9284e7 tedge-agent[518]: 2024-01-25T08:33:27.381178295Z  INFO tedge_agent::software_manager::actor: Checking if tedge got self updated
Jan 25 08:33:27 03d75d9284e7 tedge-agent[518]: 2024-01-25T08:33:27.381197381Z  INFO tedge_agent::software_manager::actor: Current running version: 0.0.1
Jan 25 08:33:27 03d75d9284e7 tedge-agent[518]: 2024-01-25T08:33:27.38183023Z  INFO tedge_agent::software_manager::actor: Installed binary version: 1.0.0-rc.2~124+ge534ed3
Jan 25 08:33:27 03d75d9284e7 tedge-agent[518]: 2024-01-25T08:33:27.381843875Z  INFO tedge_agent::software_manager::actor: Self update detected. Requesting shutdown...
Jan 25 08:33:27 03d75d9284e7 tedge-agent[518]: 2024-01-25T08:33:27.381848895Z ERROR tedge_agent::software_manager::actor: Tedge-agent is no more running the latest-version => a restart is required
[...]

PR #2631, opened 2024-01-30
Fix tedge agent restart on self update

didier-wenzek:
## Proposed changes

As highlighted by [flaky tests](https://github.com/thin-edge/thin-edge.io/issues/2623), there is a race condition when a self-update is detected by the agent. While a first thread is publishing the operation status over MQTT, a concurrent thread is triggering a shutdown of the agent, possibly before the status is actually published.

This fix introduces two mitigations:
- The agent shutdown is delayed.
- On shutdown, all the MQTT messages already queued are published.  

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2623

## Checklist


crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -134,12 +134,18 @@ impl FromPeers {
...
-            match self.try_recv().await {
-                Ok(Some(message)) => outgoing_mqtt.send(message).await.map_err(Box::new)?,
-                Ok(None) | Err(RuntimeRequest::Shutdown) => break Ok(()),
-            }
+        while let Ok(Some(message)) = self.try_recv().await {
+            outgoing_mqtt.send(message).await.map_err(Box::new)?;
         }
+
+        // On shutdown, first close input so no new messages can be pushed
+        self.input_receiver.close_input();
```
albinsuresh, line 142: Shouldn't the `incoming_mqtt` channel should also be closed to avoid reading any more messages from the broker as well? For that, these `relay_xxx` methods would need to take the mutable `Connection` itself, instead of the `published` or `received` channels from it.

didier-wenzek: Unfortunately, it would be pointless to close the `incoming_mqtt` channel as the underlying TCP connection to the broker is bi-directional: as we want to publish a latest batch of messages; we have to keep the connection open.

albinsuresh (approved):
LGTM

gligorisaev:
QA has thoroughly checked the bug and here are the results:

- [x] QA has tested the bug and it's not reproducable anymore

Issue #2605, opened 2024-01-19
Configuration flag to bypass duplicate message filtering on mapper startup


albinsuresh:
**Is your feature improvement request related to a problem? Please describe.**

The mapper currently avoids sending duplicate entity registration messages to the cloud on startup, by ignoring the resent registration messages for the entities that are already present in the restored persistent entity store. But, if someone intentionally/accidentally wipes the entities from the cloud, they will never be recreated again by the mapper, due to this filtering mechanism. So, a mechanism to disable this filtering on startup is desired to ease the syncing of the entire entity state on tedge to the cloud.

**Describe the solution you'd like**

Introduce a `c8y.entity_store.sync_on_startup` boolean config flag to enable the above mentioned forced syncing on startup. The default value should be `false`, to keep the current filtering mechanism as the default.

didier-wenzek:
A new tedge config settings has been added: `c8y.entity_store.clean_start`

- On a clean start, the whole state of the device, services and child-devices is resent to the cloud.
- The default for `c8y.entity_store.clean_start` is to be false. Only new services and child-devices, created while the c8y mapper was down, are published to the Cumulocity IoT on mapper restart.

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [ ] Test for ticket exists in the test suite.
- [x] QA has tested the function and it's functioning according description.

PR #2632, opened 2024-01-30
Bypass duplicate message filtering on mapper startup with config flag

Ruadhri17:
## Proposed changes
TODO:

- [x] fix tests
- [x] add test
- [x] replace `sync_on_startup` with better name

## Types of changes

## Paste Link to the issue
#2605 
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -450,6 +450,10 @@ define_tedge_config! {
             /// Enable auto registration feature
             #[tedge_config(example = "true", default(value = true))]
             auto_register: bool,
+
+            /// On a clean start, the whole state of the device, services and child-devices is resent to the cloud
+            #[tedge_config(example = "true", default(value = true))]
+            clean_start: bool,
         },
     },
 
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -135,7 +135,8 @@ type ExternalIdValidatorFn =
 ///     |tid, xid| tid.to_string().into(),
 ///     |xid| Ok(xid.into()),
 ///     5,
-///     "/tmp"
+///     "/tmp",
+///     true
 /// );
 /// ```
 pub struct EntityStore {
@@ -153,6 +154,7 @@ pub struct EntityStore {
 }
 
 impl EntityStore {
+    #[allow(clippy::too_many_arguments)]
     pub fn with_main_device_and_default_service_type<MF, SF, P>(
         mqtt_schema: MqttSchema,
         main_device: EntityRegistrationMessage,
@@ -161,6 +163,7 @@ impl EntityStore {
         external_id_validator_fn: SF,
         telemetry_cache_size: usize,
         log_dir: P,
+        clean_start: bool,
     ) -> Result<Self, InitError>
     where
         MF: Fn(&EntityTopicId, &EntityExternalId) -> EntityExternalId,
@@ -187,11 +190,19 @@ impl EntityStore {
             twin_data: Map::new(),
         };
 
-        let message_log = MessageLogWriter::new(log_dir.as_ref()).map_err(|err| {
-            InitError::Custom(format!(
-                "Loading the entity store log for writes failed with {err}",
-            ))
-        })?;
+        let message_log = if clean_start {
+            MessageLogWriter::new_truncated(log_dir.as_ref()).map_err(|err| {
+                InitError::Custom(format!(
+                    "Loading the entity store log for writes failed with {err}",
+                ))
+            })?
+        } else {
+            MessageLogWriter::new(log_dir.as_ref()).map_err(|err| {
+                InitError::Custom(format!(
+                    "Loading the entity store log for writes failed with {err}",
+                ))
+            })?
+        };
 
         let mut entity_store = EntityStore {
             mqtt_schema: mqtt_schema.clone(),
@@ -1072,7 +1083,7 @@ mod tests {
     #[test]
     fn registers_main_device() {
         let temp_dir = tempfile::tempdir().unwrap();
-        let store = new_entity_store(&temp_dir);
+        let store = new_entity_store(&temp_dir, true);
 
         assert_eq!(store.main_device(), &EntityTopicId::default_main_device());
         assert!(store.get(&EntityTopicId::default_main_device()).is_some());
@@ -1081,7 +1092,7 @@ mod tests {
     #[test]
     fn lists_child_devices() {
         let temp_dir = tempfile::tempdir().unwrap();
-        let mut store = new_entity_store(&temp_dir);
+        let mut store = new_entity_store(&temp_dir, true);
 
         // If the @parent info is not provided, it is assumed to be an immediate
         // child of the main device.
@@ -1119,7 +1130,7 @@ mod tests {
     #[test]
     fn lists_services() {
         let temp_dir = tempfile::tempdir().unwrap();
-        let mut store = new_entity_store(&temp_dir);
+        let mut store = new_entity_store(&temp_dir, true);
 
         // Services are namespaced under devices, so `parent` is not necessary
         let updated_entities = store
@@ -1161,7 +1172,7 @@ mod tests {
     #[test]
     fn list_ancestors() {
         let temp_dir = tempfile::tempdir().unwrap();
-        let mut store = new_entity_store(&temp_dir);
+        let mut store = new_entity_store(&temp_dir, true);
 
         // Assert no ancestors of main device
         assert!(store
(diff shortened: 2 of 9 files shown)


crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -450,6 +450,10 @@ define_tedge_config! {
             /// Enable auto registration feature
             #[tedge_config(example = "true", default(value = true))]
             auto_register: bool,
+
+            /// Determines if entity state on tedge should be sent to cloud on startup
+            #[tedge_config(example = "true", default(value = true))]
+            sync_on_startup: bool,
```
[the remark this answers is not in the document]
didier-wenzek: @albinsuresh why are you no happy with @Ruadhri17 proposal?

I'm okay with `sync_on_startup`, `sync_on_start` and `clean_start`, but with none of the alternatives. These alternatives will only puzzle the users.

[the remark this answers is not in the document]
albinsuresh: That was on a request from @reubenmiller because the word `sync` might imply two-way synchronization. So, we wanted to change it to something that implies unidirectional (device to cloud). I finally settled on `clean_start` as it's short and implies exactly what the code does.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -450,6 +450,10 @@ define_tedge_config! {
             /// Enable auto registration feature
             #[tedge_config(example = "true", default(value = true))]
             auto_register: bool,
+
+            /// Determines if entity state on tedge should be sent to cloud on startup
```
didier-wenzek, line 454: I'm not sure this comment is meaningful for an  end-user reading `tedge config list --doc` 

```suggestion
            /// On a clean start, the whole state of the device, services and child-devices is resent to the cloud
```

crates/core/tedge_api/src/message_log.rs
```diff
@@ -68,15 +68,22 @@ pub struct MessageLogWriter {
 }
 
 impl MessageLogWriter {
-    pub fn new<P>(log_dir: P) -> Result<MessageLogWriter, std::io::Error>
+    pub fn new<P>(log_dir: P, truncate: bool) -> Result<MessageLogWriter, std::io::Error>
```
didier-wenzek, line 71: Instead of adding a `bool` to the constructor, it would be better to add a new constructor.

```
    pub fn new<P>(log_dir: P) -> Result<MessageLogWriter, std::io::Error> {
             // unchanged.
    }

    pub fn new_truncated<P>(log_dir: P) -> Result<MessageLogWriter, std::io::Error> {
             // truncating the file first.
    }

```

crates/core/tedge_api/src/message_log.rs
```diff
@@ -68,15 +68,22 @@ pub struct MessageLogWriter {
...
+                .create(true)
+                .write(true)
+                .truncate(true)
+                .open(log_dir.as_ref().join(LOG_FILE_NAME))?
+        } else {
+            OpenOptions::new()
+                .create(true)
+                .append(true)
+                .open(log_dir.as_ref().join(LOG_FILE_NAME))?
+        };
```
didier-wenzek, line 86:
```suggestion
        let file = OpenOptions::new()
            .create(true)
            .write(true)
            .truncate(truncate)
            .open(log_dir.as_ref().join(LOG_FILE_NAME))?;
```

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -201,7 +204,9 @@ impl EntityStore {
             message_log,
         };
 
-        entity_store.load_from_message_log(log_dir.as_ref());
+        if !clean_start {
+            entity_store.load_from_message_log(log_dir.as_ref());
+        }
```
didier-wenzek, line 209: As the log is wiped on a clean start, it's safe to load it unconditionally.

```suggestion
        entity_store.load_from_message_log(log_dir.as_ref());
```

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -161,6 +163,7 @@ impl EntityStore {
         external_id_validator_fn: SF,
         telemetry_cache_size: usize,
         log_dir: P,
+        clean_start: bool,
```
didier-wenzek, line 166: Passing a `log_dir` and `clean_start` flag telling to truncate or not the log.

I would consider to pass only a `MessageLogWriter`, built from the log using  `MessageLogWriter::new(log_dir)` or  `MessageLogWriter::new_truncated(log_dir)`:

```suggestion
        message_log: MessageLogWriter,
```

For that to work the `MessageLogWriter` has to be improved to build a `MessageLogReader` on the same log dir. So the entity store can use its  `self.message_log` for reads and writes.

It might be better to address this only a follow-up PR after the 1.0 release.

didier-wenzek (commented):
This working as expected. However, the code can be a bit simpler.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: One suggestion is to have a `MessageLog` abstraction, which is passed to the entity store, out of which the `MessageLogWriter` and `MessageLogReader` instances can be retrieved.

tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
```diff
@@ -193,13 +198,55 @@ Entities persisted and restored
         Should Have MQTT Messages    c8y/s/us/plc2    message_contains=102    date_from=${timestamp}    minimum=0    maximum=0
     END
 
+Entities send to cloud on restart
+    Execute Command    tedge mqtt pub --retain 'te/factory/shop/plc1/' '{"@type":"child-device","@id":"plc1"}'
```
albinsuresh, line 202: Could you align the `@id` values used in this test along the lines of https://github.com/thin-edge/thin-edge.io/pull/2657 ?

tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
```diff
@@ -193,13 +198,55 @@ Entities persisted and restored
...
+        Sleep    2s    reason=Make sure that previous iteration was sent to cloud
+    END
 
 *** Keywords ***
 
 Re-enable Auto-registration
     Execute Command    sudo tedge config unset c8y.entity_store.auto_register
     Restart Service    tedge-mapper-c8y
 
+Re-enable Clean start
```
albinsuresh, line 246:
```suggestion
Disable clean start
```

tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
```diff
@@ -193,13 +198,55 @@ Entities persisted and restored
...
+    ${original_last_modified_time}=    Execute Command    date -r /etc/tedge/.tedge-mapper-c8y/entity_store.jsonl
+
+    FOR    ${counter}    IN RANGE    0    5
+        ${timestamp}=    Get Unix Timestamp
+        Restart Service    tedge-mapper-c8y
+        Service Health Status Should Be Up    tedge-mapper-c8y
+
+        # Assert that the file contents changed on restart
+        ${last_modified_time}=    Execute Command    date -r /etc/tedge/.tedge-mapper-c8y/entity_store.jsonl
+        Should Not Be Equal    ${last_modified_time}    ${original_last_modified_time}
```
albinsuresh, line 228: Since the timestamp assertion is already done in the previous test, we can either skip this whole block or at least avoid the looping and limit the check to only once.

tests/RobotFramework/tests/cumulocity/registration/device_registration.robot
```diff
@@ -193,13 +198,55 @@ Entities persisted and restored
...
+        
+        # Assert that entities are sent to cloud again
+        Should Have MQTT Messages    c8y/s/us    message_contains=101,plc    date_from=${timestamp}    minimum=2    maximum=2
+        Should Have MQTT Messages    c8y/s/us/plc1    message_contains=101    date_from=${timestamp}    minimum=2    maximum=2
+        Should Have MQTT Messages    c8y/s/us/plc2    message_contains=101    date_from=${timestamp}    minimum=1    maximum=1
+        Should Have MQTT Messages    c8y/s/us/plc1    message_contains=102    date_from=${timestamp}    minimum=1    maximum=1
+        Should Have MQTT Messages    c8y/s/us/plc2    message_contains=102    date_from=${timestamp}    minimum=1    maximum=1
+
+        Sleep    2s    reason=Make sure that previous iteration was sent to cloud
```
albinsuresh, line 237: Why was this sleep required? That too at the end of this test. The reason isn't clear from the comment.

Ruadhri17: I noticed that sometimes tests just fail because of that test case:
`Should Have MQTT Messages    c8y/s/us    message_contains=101,${prefix}plc    date_from=${timestamp}    minimum=2    maximum=2`. It returns more messages than expected (It can be noticed for example in [this run](https://github.com/thin-edge/thin-edge.io/actions/runs/7734517184/job/21098432434#step:8:282)). The only solution I found to this was adding this sleep, but maybe someone else has better solution to that.

albinsuresh (approved):

Issue #2641, opened 2024-01-31
Documentation punchlist


reubenmiller:
## Documentation updates

* [x] Configuration Tools page - https://github.com/thin-edge/thin-edge.io/pull/2682

    The name of this file does not really match its contents, and it could be possibly moved to the `index.md` of the section rather than a sub-page, as it reads configuration->configuration which is redundant.

    **Page:** /operate/configuration/configuration-tools.md

* [x] Missing Test AWS section - https://github.com/thin-edge/thin-edge.io/pull/2675

    AWS sub section is missing on how to test the connection (only c8y and az are listed)

    **Page:** /operate/troubleshooting/testing-cloud-connection/
* [x] Getting started
    * [x] Remove explicit timestamps in both the events and alarms pages. Using a fixed timestamp is a bad idea as users will not be able to view the event in the UI due to the timestamp being in the past - https://github.com/thin-edge/thin-edge.io/pull/2650

    * [x] Add note about the timestamp being able to be defined as a linux timestamp

* [x] Remove workflow construction sign - https://github.com/thin-edge/thin-edge.io/pull/2648
    * [x] The workflows are out of the proposal phase and should have the warnings removed.

* [x] operate/c8y/software-management/
    * [x] Remove sections about installing thin-edge.io via the Software Management feature and more focus on installing any software only with screenshots. - https://github.com/thin-edge/thin-edge.io/pull/2667

* [x] Legacy firmware management for child devices
    * [x] Two pages for the same topic under the legacy child device topic - https://github.com/thin-edge/thin-edge.io/pull/2665
        * [x] tedge-docs/docs/legacy/child-device-firmware-management.md
        * [x] tedge-docs/docs/legacy/firmware-management.md

* [x] Getting started - Monitoring page
    * [x] Use community package to automatically install collectd rather than going through complicated steps. - https://github.com/thin-edge/thin-edge.io/pull/2674

## Not Applicable

The following items have been removed as they are not mandatory and will only be actioned if required later.

* [ ] ~Remaining Thin Edge references~
    * Renaming `Thin Edge JSON` to `thin-edge.io JSON` looks a bit strange for a title...not sure if there is a better option here, or we just live with it for now.
* [ ] ~Order of Operate Device section~

    The order of the **Operate Device** section needs to be re-thought and maybe simplified.

reubenmiller:
Closing the tracking ticket as all of the PRs have been created and are on their way to being merged.

PR #2650, opened 2024-02-01
docs(getting started): remove explicit timestamps from example commands

reubenmiller:
## Proposed changes

remove explicit timestamps in example commands in the Getting Started section.

Removing the timestamps improves the user experience as it ensures the alarms/events are created at the time of execution and not in the past (which results in the data not being displayed in the c8y UI as it is too far in the past)

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2641

## Checklist

### DIFF
--- a/docs/src/start/getting-started.md
+++ b/docs/src/start/getting-started.md
@@ -208,8 +208,7 @@ A simple event can be represented in %%te%% JSON as follows:
 
 ```json
 {
-  "text": "A door was closed",
-  "time": "2022-06-10T05:30:45+00:00"
+  "text": "A door was closed"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


docs/src/start/getting-started.md
```diff
@@ -222,9 +221,13 @@ te/+/+/+/+/e/+
 So the door open event described above can be sent as follows:
 
 ```sh te2mqtt formats=v1
-tedge mqtt pub te/device/main///e/door '{"text": "A door was closed","time": "2022-06-10T05:30:45+00:00"}'
+tedge mqtt pub te/device/main///e/door '{"text": "A door was closed"}'
 ```
 
+:::info
+The command does not provide the `time` property, so the current timestamp will be injected by the mapper.
```
didier-wenzek, line 228: What adding that a timestamp can  be set along the measurement?

```suggestion
The command does not provide the `time` property, so the current timestamp will be injected by the mapper.
However an explicit time can be given as a Unix timestamp, as in `"time": 1706794400`, or using RFC 3339, as in `"time":"2024-02-01T13:32:19+00:00"`.
```

reubenmiller: Yes good suggestion, mentioning the Unix timestamp was also part of the punchlist, so thanks for addressing that for me :)

didier-wenzek (approved):
Approved

PR #2673, opened 2024-02-04
test: use a more reliable service pid detection

reubenmiller:
## Proposed changes

Due to some flake-finder runs, it was found that the assertions which rely on getting a service's main PID can suffer from race conditions if there is a delay between starting a service and getting the pid via `pgrep`. Instead the `Service Should Be Running` keyword also returns the PID of the running service, so it does not assume that the service is running "now", but will wait if necessary if the service is not already running.

image

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


albinsuresh (approved):

didier-wenzek (approved):
Approved

Issue #2666, opened 2024-02-03
debian software package does not install from url when version contains a colon character


reubenmiller:
**Describe the bug**

The tedge-apt-plugin fails to install a software package from a URL (e.g. a package hosted in Cumulocity IoT) if the software version contains a colon in its name due to `apt-get install <file>` ignoring the `<file>` argument if a colon is present.

apt interprets the colon as the [Epoch separator](https://www.debian.org/doc/debian-policy/ch-controlfields.html#version), so it seems that `apt-get install` silently ignores the given file and does not exit.

The root cause is the fact that `tedge-apt-plugin` downloads the debian file from the given software url, and save the file using the path built from the name and version, e.g. `/tmp/{name}_{version}.deb`, and the filename is not sanitized to be compatible with `apt/apt-get`.

There is a similar issue found in another project:
* https://github.com/camicri/cube-get/issues/10

**Note**

* You can't just rename the `version` field of the software package definition in Cumulocity IoT as `tedge-apt-plugin` does a meta information validation and while throw an error (`MetaDataMismatch`) if the given version from the cloud does not match the package's version

**To Reproduce**

1. Create a software package in Cumulocity for vim

    * Name: `vim`
    * Version: `2:8.2.2434-3+deb11u1`
    * Url: Upload a package that matches your device's CPU architecture from the [debian website](https://packages.debian.org/bookworm/vim)

2. Install vim via the Cumulocity IoT Device Management Software page for the device under test

3. Inspect the software management logs on the device to see if the `apt-get install` command actually tried installing the package or not (see below for an example where it doesn't process the given package)

    ```sh
    ----- $ Downloading: http://127.0.0.1:8001/c8y/inventory/binaries/33714613 to /tmp/vim_2:8.2.2434-3+deb11u1
    ----- $ /usr/bin/sudo "/etc/tedge/sm-plugins/apt" "update-list"
    exit status: 0

    stdout <<EOF
    apt-get install args: ["install", "--quiet", "--yes", "/tmp/vim_2:8.2.2434-3+deb11u1.deb"]
    Reading package lists...
    Building dependency tree...
    Reading state information...
    0 upgraded, 0 newly installed, 0 to remove and 39 not upgraded.
    EOF

    stderr <<EOF
    EOF
    ```

**Expected behavior**

If the `tedge-apt-plugin` downloads the debian package from a given URL, then the temporary file created from the download should sanitize the version by substituting any unsupported characters.

[...]

PR #2683, opened 2024-02-05
Replace special characters in package version

Ruadhri17:
## Proposed changes
This PR provides a quick fix to the tedge-apt-plugin bug, which prevents the package installation if it contains a colon character. More in #2666.

## Types of changes

## Paste Link to the issue
* #2666 
<br/>

## Checklist


crates/core/plugin_sm/src/plugin.rs
```diff
@@ -554,5 +554,17 @@ fn sm_path(name: &str, version: &Option<String>, target_dir_path: impl AsRef<Pat
...
+    target_dir_path.as_ref().join(sanitize_filename(&filename))
+}
+
+fn sanitize_filename(filename: &str) -> String {
+    let mut result = String::new();
+    filename.chars().for_each(|c| {
+        if matches!(c as u8, b'0'..=b'9' | b'A'..=b'Z' | b'a'..=b'z' |  b'-' | b'.' | b'_' | b'~') {
+            result.push(c)
+        } else {
+            result.push_str(&format!("%{:x?}", c as u8))
```
albinsuresh, line 566: I thought the proposal was to replace unexpected chars with `_`. But this is also fine. That being said, shouldn't we have a test?

reubenmiller: The more correct way (at least on linux) is to use the url encoded char (as this is what apt does when it downloads the deb package and stores it under its cache folder).

But yes, bare minimum would be some unit tests.

didier-wenzek:
> Software not installed. errors=[('package-with-epoch', 'MISSING')]

You have to add `package-with-epoch` to the list of packages not filtered out by the `tedge-apt-plugin`.

reubenmiller (approved):
Tests (including generation of test files) looks good.

tests/images/debian-systemd/files/tedge.toml
```diff
@@ -1,3 +1,4 @@
 
 [apt]
 name = "(tedge|c8y|python|wget|vim|curl|apt|mosquitto|ssh|sudo|rolldice).*"
+maintainer = ".*(thin-edge.io).*"
```
didier-wenzek, line 4: > Parent suite setup failed: dpkg -i packages/tedge_*.deb returned an unexpected exit code
TOML parse error at line 4, column 34

The system test is failing because a trailing new line is missing.

reubenmiller: That's unexpected, but good to have a line ending anway...

reubenmiller:
@Ruadhri17 
The reason why the `Apply name filter` test is now failing is because the default maintainer filter was added to the base integration test image, and in this test it is calling the `tedge-apt-plugin list` command manually which is inheriting the maintainer filter from the tedge.toml (which previously did not exist). So we just need to prevent that from happening by using `--maintainer ""` in the test.

Here is the git patch to fix the issue:

```
diff --git a/tests/RobotFramework/tests/plugin_apt/filter_packages_list_output.robot b/tests/RobotFramework/tests/plugin_apt/filter_packages_list_output.robot
index b921944f5e..12e74768ce 100644
--- a/tests/RobotFramework/tests/plugin_apt/filter_packages_list_output.robot
+++ b/tests/RobotFramework/tests/plugin_apt/filter_packages_list_output.robot
@@ -11,7 +11,7 @@ Test Tags           theme:software    theme:plugins
 *** Test Cases ***
 Apply name filter
     ${packages_list}=    Execute Command
-    ...    sudo /etc/tedge/sm-plugins/apt list --name tedge
+    ...    sudo /etc/tedge/sm-plugins/apt list --name tedge --maintainer ""
     ...    exp_exit_code=0
     Should Match Regexp    ${packages_list}    ^tedge\\s+${VERSION}
```

didier-wenzek (approved):
Approved

Issue #2692, opened 2024-02-08
mosquitto unresponsive and using 100% CPU


reubenmiller:
**Describe the bug**

On a device, it was observed that the local mosquitto MQTT broker was unresponsive and the mosquitto process as consuming 100% of the CPU processing time of 1 core.

**mosquitto** was reporting the following error, and after some time no more logs were being written:

```
OpenSSL Error[0]: error:0A000126:SSL routines::unexpected eof while reading
```

The local MQTT broker was non-functional which resulted in all of the thin-edge.io components failing to connect to the broker with the following network connection error:

```
Feb 08 13:13:11 rackfslot1 tedge-mapper[3356145]: 2024-02-08T12:13:11.071637229Z ERROR mqtt_channel::connection: MQTT connection error: Network timeout
Feb 08 13:13:16 rackfslot1 tedge-mapper[3356145]: 2024-02-08T12:13:16.07245946Z ERROR mqtt_channel::connection: MQTT connection error: Network timeout
Feb 08 13:13:21 rackfslot1 tedge-mapper[3356145]: 2024-02-08T12:13:21.073309857Z ERROR mqtt_channel::connection: MQTT connection error: Network timeout
```

Manually subscribing to the local MQTT broker on `localhost:1883` was also met with a Network timeout error.

**mosquitto** was able to be revived only by restarting the service, using:

```
systemctl restart mosquitto
```

Afterwards all of the services started functioning again.

**Secondary symptoms**

The following were some secondary symptoms which were observed when the device was in this state.

* The memory usage of the thin-edge.io components was rising after the MQTT broker was non-functional (possible leakage of resources in this unexpected state?). This memory was not released when the MQTT connect was resumed.

**To Reproduce**

This situation has not been able to be reproduced yet, however there seems to be some correlation between the Cumulocity IoT update occurring and this mosquitto high CPU behaviour.

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`Raspberry Pi 4 Model B Rev 1.5`|
|System-Architecture|`Linux rackfslot1 6.1.0-rpi6-rpi-v8 #1 SMP PREEMPT Debian 1:6.1.58-1+rpt2 (2023-10-27) aarch64 GNU/Linux`|
[...]

PR #2693, opened 2024-02-08
Log memory usage of the agent and the mapper

didier-wenzek:
## Proposed changes

Log memory usage of the agent and the mapper. 
The default interval is of 60 seconds.

This can be changed using the` --log-memory-interval` flag which sets the logging interval in seconds.

```
$ target/debug/tedge-mapper --log-memory-interval 5 c8y
$ sudo target/debug/tedge-agent --log-memory-interval 5
```

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2692#issuecomment-1934466228

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ c8y_http_proxy = { path = "crates/extensions/c8y_http_proxy" }
 c8y_log_manager = { path = "crates/extensions/c8y_log_manager" }
 c8y_mapper_ext = { path = "crates/extensions/c8y_mapper_ext" }
 camino = "1.1"
+cap = "0.1"
 capture-logger = "0.1"
 certificate = { path = "crates/common/certificate" }
 clap = { version = "4.4", features = ["cargo", "derive"] }
--- a/crates/core/tedge/src/main.rs
+++ b/crates/core/tedge/src/main.rs
@@ -2,15 +2,22 @@
 #![deny(clippy::mem_forget)]
 
 use anyhow::Context;
+use cap::Cap;
 use clap::Parser;
+use std::alloc;
 use std::future::Future;
 use std::path::PathBuf;
+use std::time::Duration;
 use tedge::command::BuildCommand;
 use tedge::command::BuildContext;
 use tedge::Component;
 use tedge::TEdgeOptMulticall;
 use tedge_apt_plugin::AptCli;
 use tedge_config::system_services::set_log_level;
+use tracing::log;
+
+#[global_allocator]
+static ALLOCATOR: Cap<alloc::System> = Cap::new(alloc::System, usize::MAX);
 
 fn main() -> anyhow::Result<()> {
     let executable_name = executable_name();
@@ -22,10 +29,20 @@ fn main() -> anyhow::Result<()> {
 
     let opt = parse_multicall_if_known(&executable_name);
     match opt {
-        TEdgeOptMulticall::Component(Component::TedgeMapper(mapper_opt)) => {
-            block_on(tedge_mapper::run(mapper_opt))
+        TEdgeOptMulticall::Component(Component::TedgeMapper(opt)) => {
+            let tedge_config = tedge_config::load_tedge_config(&opt.config_dir)?;
+            block_on_with(
+                tedge_config.run.log_memory_interval.duration(),
+                tedge_mapper::run(opt),
+            )
         }
-        TEdgeOptMulticall::Component(Component::TedgeAgent(opt)) => block_on(tedge_agent::run(opt)),
+        TEdgeOptMulticall::Component(Component::TedgeAgent(opt)) => {
+            let tedge_config = tedge_config::load_tedge_config(&opt.config_dir)?;
+            block_on_with(
+                tedge_config.run.log_memory_interval.duration(),
+                tedge_agent::run(opt),
+            )
+        }
         TEdgeOptMulticall::Component(Component::C8yFirmwarePlugin(fp_opt)) => {
             block_on(c8y_firmware_plugin::run(fp_opt))
         }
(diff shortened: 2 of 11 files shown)


crates/core/tedge_agent/src/lib.rs
```diff
@@ -60,6 +60,10 @@ pub struct AgentOpt {
     /// MQTT root prefix
     #[clap(long)]
     pub mqtt_topic_root: Option<Arc<str>>,
+
+    /// Interval at which is logged memory usage (in second)
```
Bravo555, line 64: nit: sentence could be written better

Bravo555 (approved):

crates/core/tedge_agent/src/lib.rs
reubenmiller:
```suggestion
    /// Interval at which the memory usage is logged (in seconds)
```

reubenmiller: Or
```suggestion
    /// Interval of the memory usage logging (in seconds)
```

crates/core/tedge_agent/src/main.rs
```diff
@@ -1,7 +1,25 @@
...
+    let log_memory_interval = agent_opt.log_memory_interval;
+    if log_memory_interval > 0 {
+        tokio::spawn(async move {
+            let interval = Duration::from_secs(log_memory_interval);
+            loop {
+                log::info!("Allocated memory: {} Bytes", ALLOCATOR.allocated());
+                tokio::time::sleep(interval).await;
+            }
+        });
+    }
```
didier-wenzek, line 22: Oops, I did the fix in the wrong place.

This must be done inside `tedge_agent::run()` and not in the `tedge_main::main()`. Indeed, the latter is only used by the plain executable and not by the sym link pointing to `tedge`.


=> I will fix that tomorrow.

Cargo.toml
```diff
@@ -56,6 +56,7 @@ c8y_http_proxy = { path = "crates/extensions/c8y_http_proxy" }
 c8y_log_manager = { path = "crates/extensions/c8y_log_manager" }
 c8y_mapper_ext = { path = "crates/extensions/c8y_mapper_ext" }
 camino = "1.1"
+cap = "0.1"
```
albinsuresh, line 59: Side commentary: Came across this [sysinfo](https://docs.rs/sysinfo/latest/sysinfo/) crate that covers a lot more parameters like CPU usage. Something we can consider in the future, for advanced monitoring.

crates/core/tedge_agent/src/lib.rs
```diff
@@ -60,6 +60,10 @@ pub struct AgentOpt {
     /// MQTT root prefix
     #[clap(long)]
     pub mqtt_topic_root: Option<Arc<str>>,
+
+    /// Interval at which the memory usage is logged (in seconds)
+    #[clap(long = "log-memory-interval", default_value = "60")]
+    pub log_memory_interval: u64,
```
albinsuresh, line 66: Is this a short term addition or something permanent? If permanent, any specific reason why this is a command-line argument and not tedge config setting (other than convenience and simplicity)? No objections here, but just trying to understand the rationale behind that choice.

crates/core/tedge_agent/src/lib.rs
didier-wenzek: I applied the first proposal.

crates/core/tedge_agent/src/main.rs
didier-wenzek: Fixed with https://github.com/thin-edge/thin-edge.io/pull/2693/commits/3f0fc8902f33e0549b98e737373b288aadc5379a

Cargo.toml
didier-wenzek: The purpose is different. Sysinfo observes the operating system, while here the point is to monitor the amount of  memory actually allocated by tedge on the heap.

crates/core/tedge_agent/src/lib.rs
didier-wenzek: The idea is to set this right when the process starts - ideally before any thin-edge processing.

I see this as a new feature. This can possibly be extended with:
- the number of threads (currently we use the default i.e one thread per core),
- a cap on allocated memory (crashing if the process starts to go wild).

crates/core/tedge_agent/src/main.rs
```diff
@@ -1,7 +1,25 @@
...
 use clap::Parser;
+use std::alloc;
+use std::time::Duration;
+
+#[global_allocator]
+static ALLOCATOR: Cap<alloc::System> = Cap::new(alloc::System, usize::MAX);
 
 #[tokio::main]
-async fn main() -> Result<(), anyhow::Error> {
+async fn main() -> anyhow::Result<()> {
```
didier-wenzek, line 10: We have to consider deprecating this `main` and enforcing the use of a `tedge -agent` as a symlink to `tedge`.

crates/core/tedge_agent/src/lib.rs
reubenmiller: Supporting a config would make it easier to enable on devices under test...as adding a flag is more difficult as we would need to use `systemctl edit ...` to enable this feature.

didier-wenzek: > Supporting a config would make it easier to enable on devices under test...as adding a flag is more difficult as we would need to use `systemctl edit ...` to enable this feature.

Done with https://github.com/thin-edge/thin-edge.io/pull/2693#discussion_r1484348307

The new setting is `run.log_memory_interval` with a default of 5 minutes.

crates/core/tedge_agent/src/main.rs
didier-wenzek: Here is a ticket for that: https://github.com/thin-edge/thin-edge.io/issues/2696

crates/core/tedge/src/main.rs
```diff
@@ -22,10 +29,20 @@ fn main() -> anyhow::Result<()> {
...
     let opt = parse_multicall_if_known(&executable_name);
     match opt {
-        TEdgeOptMulticall::Component(Component::TedgeMapper(mapper_opt)) => {
-            block_on(tedge_mapper::run(mapper_opt))
+        TEdgeOptMulticall::Component(Component::TedgeMapper(opt)) => {
+            let tedge_config = tedge_config::load_tedge_config(&opt.config_dir)?;
+            block_on_with(
+                tedge_config.run.log_memory_interval.duration(),
+                tedge_mapper::run(opt),
+            )
```
didier-wenzek, line 37: I'm not very happy with that, i.e. loading the config without passing it to `tedge_mapper::run()`.

Doing so is not straightforward as each of the actors used by the mapper and the agent have specific way to load the config, with or without a reference to the config-dir and other weirdness including duplicated paths to the same setting. It would be good to simplify all that in a follow up PR.

reubenmiller (approved):
Looks good.

Issue #2689, opened 2024-02-06
supported operations are not being registered after initial bootstrapping of the tedge-rugpi-image


reubenmiller:
**Describe the bug**

**To Reproduce**

1. Build image using [tedge-rugpi-image](https://github.com/thin-edge/tedge-rugpi-image)
2. Onboard devices via

    ```sh
    c8y tedge bootstrap root@mydevice
    ```

3. Open device in Cumulocity IoT Device Management UI and check if the firmware, restart device operations have been registered (e.g. is firmware tab visible, is the restart device button on the info page)

**Workaround**

Restarting the `tedge-mapper-c8y` service results supported operations being sent to Cumulocity.

```sh
systemctl restart tedge-mapper-c8y
```

**Expected behavior**

The supported operations should be sent to Cumulocity IoT after bootstrapping without having to reconnect to c8y.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`Raspberry Pi 5 Model B Rev 1.0`|
|System-Architecture|`Linux rpi5-d83addab8e9f 6.1.0-rpi7-rpi-2712 #1 SMP PREEMPT Debian 1:6.1.63-1+rpt1 (2023-11-24) aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.0.0-rc.2~344+g9cf6904`|

**Additional context**

**logs: tedge-mapper-c8y**

The errors at the beginning of the logs are prior to the bootstrapping, then it starts up after being bootstrapped (e.g. set c8y.url and create certificate)

```log
Feb 06 21:15:18 rpi5-d83addab8e9f tedge-mapper[2046]: Error: A value for 'c8y.url' is missing.
Feb 06 21:15:18 rpi5-d83addab8e9f tedge-mapper[2046]:     A value can be set with `tedge config set c8y.url <value>`
Feb 06 21:15:18 rpi5-d83addab8e9f systemd[1]: tedge-mapper-c8y.service: Main process exited, code=exited, status=1/FAILURE
Feb 06 21:15:18 rpi5-d83addab8e9f systemd[1]: tedge-mapper-c8y.service: Failed with result 'exit-code'.
Feb 06 21:15:23 rpi5-d83addab8e9f systemd[1]: tedge-mapper-c8y.service: Scheduled restart job, restart counter is at 32.
[...]

PR #2695, opened 2024-02-09
Fix missing supported operations #2689

albinsuresh:
## Proposed changes

The bug was caused by sending messages to `c8y/s/us` topic before the c8y-bridge connection is even established, resulting in those messages getting lost. With this fix we make sure that the mapper does not start mapping or even send any init messages to the cloud until the bridge connection is established.

- [x] System test
- [x] Fix

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2689

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/utils.rs
+++ b/crates/core/c8y_api/src/utils.rs
@@ -5,12 +5,23 @@ pub mod bridge {
     // FIXME: doesn't account for custom topic root, use MQTT scheme API here
     pub const C8Y_BRIDGE_HEALTH_TOPIC: &str =
         "te/device/main/service/mosquitto-c8y-bridge/status/health";
-    const C8Y_BRIDGE_UP_PAYLOAD: &str = "1";
+    pub const C8Y_BRIDGE_UP_PAYLOAD: &str = "1";
+    const C8Y_BRIDGE_DOWN_PAYLOAD: &str = "0";
 
     pub fn is_c8y_bridge_up(message: &Message) -> bool {
         match message.payload_str() {
             Ok(payload) => {
                 message.topic.name == C8Y_BRIDGE_HEALTH_TOPIC && payload == C8Y_BRIDGE_UP_PAYLOAD
+            }
+            Err(_err) => false,
+        }
+    }
+
+    pub fn is_c8y_bridge_established(message: &Message) -> bool {
+        match message.payload_str() {
+            Ok(payload) => {
+                message.topic.name == C8Y_BRIDGE_HEALTH_TOPIC
+                    && (payload == C8Y_BRIDGE_UP_PAYLOAD || payload == C8Y_BRIDGE_DOWN_PAYLOAD)
             }
             Err(_err) => false,
         }
@@ -36,6 +47,7 @@ mod tests {
     use mqtt_channel::Topic;
     use test_case::test_case;
 
+    use crate::utils::bridge::is_c8y_bridge_established;
     use crate::utils::bridge::is_c8y_bridge_up;
     use crate::utils::bridge::C8Y_BRIDGE_HEALTH_TOPIC;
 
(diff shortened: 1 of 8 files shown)


tests/RobotFramework/tasks/E2E_Tutorials/Connect_to_c8y.robot
```diff
@@ -39,7 +39,6 @@ Create the certificate
     Should Contain    ${output}    Thumbprint:
 
 tedge cert upload c8y command
-    Execute Command    sudo tedge cert create --device-id ${DEVICE_SN}
```
albinsuresh, line 42: Already done in the previous step.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -51,6 +51,13 @@ impl TEdgeComponent for CumulocityMapper {
         let mut uploader_actor = UploaderActor::new(identity.clone()).builder();
         let mut downloader_actor = DownloaderActor::new(identity).builder();
 
+        // MQTT client dedicated to monitor the c8y-bridge client status and also
+        // set service down status on shutdown, using a last-will message.
+        // A separate MQTT actor/client is required as the last will message of the main MQTT actor
+        // is used to send down status to health topic.
+        let mut service_monitor_actor =
+            MqttActorBuilder::new(service_monitor_client_config(&tedge_config)?);
```
albinsuresh, line 59: This MQTT actor which was previously used just to send the last will message is reused to monitor the bridge health status as well.

crates/core/c8y_api/src/utils.rs
Bravo555, line 1: note: IMO `c8y_api` crate is a bad place for bridge related utilities as name suggests the crate handles cumulocity-specific stuff, e.g. authentication, smartrest, inventory, etc.

This in itself is a low priority, but more important is that checking if the bridge is established should also work for AWS and Azure bridges, but as this change only handles cumulocity, does it mean that this issue can also appear for Azure or AWS? If so, then we should consider fixing it there as well as a follow-up.

tests/images/debian-systemd/files/bootstrap.sh
```diff
@@ -14,6 +14,7 @@ FLAGS
     WORKFLOW FLAGS
     --clean/--no-clean                      Clean the device of any existing tedge installations before installing/connecting. Default False
     --install/--no-install                  Install thin-edge.io. Default True
+    --bootstrap/--no-bootstrap              Bootstrap tedge by setting c8y.url and creating and uploading the device certificate. Default True
```
Bravo555, line 17: todo: delete line 21 which also contains `--bootstrap`

Bravo555 (changes requested):
One small change required in `bootstrap.sh`

crates/core/c8y_api/src/utils.rs
albinsuresh: > note: IMO c8y_api crate is a bad place for bridge related utilities as name suggests the crate handles cumulocity-specific stuff, e.g. authentication, smartrest, inventory, etc.

I agree. I think it was added in this crate for lack of another common crate that could be accessed by all the c8y components like the mapper and other c8y plugins. But, now that the other components are gone, I don't see anything wrong with moving this into the mapper. But, that being said, this bridge itself is going to get replaced with our own bridge component in the near future. So, may not be a high priority item for now.

> This in itself is a low priority, but more important is that checking if the bridge is established should also work for AWS and Azure bridges, but as this change only handles cumulocity, does it mean that this issue can also appear for Azure or AWS? If so, then we should consider fixing it there as well as a follow-up.

The principle definitely applies to other mappers as well, that no mapping should happen until the bridge is established. But, this is less critical in the case of other mappers, in their current state, as they only map telemetry data, and not entity registrations or any other critical metadata messages. It's okay to drop a few telemetry messages before the bridge is established, as updates to them would be coming later on. But, that's not the case with other entity metadata messages which are only sent once and that's why it's far more critical for the c8y mapper.

Bravo555 (approved):

PR #2697, opened 2024-02-11
build: support build/packaging using other tooling

reubenmiller:
## Proposed changes

Preparation for future work to make it easier to reuse the same build scripts to build the linux packages from binaries produced by other tools (such as clang instead of cargo-zigbuild). The changes should not have any effect on the current build process.

The following improvements were made:

* don't install zigbuild when using the --skip-build option (previously zigbuild was installed even when not building the Rust binaries)
* option to skip building the test packages
* option to skip building deprecated packages

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/ci/build_scripts/build.sh
+++ b/ci/build_scripts/build.sh
@@ -81,6 +81,7 @@ TARGET=()
 BUILD_OPTIONS=()
 BUILD=1
 INCLUDE_TEST_PACKAGES=1
+INCLUDE_DEPRECATED_PACKAGES=1
 
 REST_ARGS=()
 while [ $# -gt 0 ]
@@ -92,6 +93,10 @@ do
 
         --skip-test-packages)
             INCLUDE_TEST_PACKAGES=0
+            ;;
+
+        --skip-deprecated-packages)
+            INCLUDE_DEPRECATED_PACKAGES=0
             ;;
 
         -h|--help)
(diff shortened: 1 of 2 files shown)


ci/build_scripts/build.sh
```diff
@@ -80,6 +80,8 @@ ARCH=
 TARGET=()
 BUILD_OPTIONS=()
 BUILD=1
+INCLUDE_TEST_PACKAGES=1
+INCLUDE_DEPRECATED_PACKAGES=1
```
albinsuresh, line 84: How about excluding these packages by default as these are useful only in rare occasions?

ci/build_scripts/build.sh
```diff
@@ -89,6 +91,14 @@ do
             BUILD=0
             ;;
 
+        --skip-test-packages)
+            INCLUDE_TEST_PACKAGES=0
+            ;;
+
+        --skip-deprecated-packages)
```
albinsuresh, line 98:
```suggestion
        --include-deprecated-packages)
```
Only if we decide to exclude these packages by default, as suggested in the previous comment.

ci/build_scripts/build.sh
reubenmiller: I wanted to keep with the current behaviour as the tests required the deprecated packages still. So I would rather remove the deprecated packages and tests in a different PR.

albinsuresh (approved):

PR #2712, opened 2024-02-14
fix: disable memory logging by default to avoid log spamming

reubenmiller:
## Proposed changes

The memory allocation logging was introduced in https://github.com/thin-edge/thin-edge.io/pull/2693, however it was overlooked that the logging was enabled by default.

To avoid spamming the logs on devices, the interval was changed to 0 which disables the logging. Users can activate the logging via the tedge configuration, though it is more meant for troubleshooting deep dives rather than the causual user.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -720,7 +720,7 @@ define_tedge_config! {
         #[tedge_config(example = "true", default(value = true))]
         lock_files: bool,
 
-        /// Interval at which the memory usage is logged (in seconds)
+        /// Interval at which the memory usage is logged (in seconds). Logging is disabled if set to 0
         #[tedge_config(example = "60", default(value = 0_u64))]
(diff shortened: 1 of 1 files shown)


crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -721,7 +721,7 @@ define_tedge_config! {
         lock_files: bool,
 
         /// Interval at which the memory usage is logged (in seconds)
```
rina23q, line 723: Perhaps writing what interval 0 means (like "0 means disabled.") is more helpful in this line, but it's ok as it is

rina23q (approved):
LGTM

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
reubenmiller: resolved by aecec18f6e815aab20f00cb0c1bdecf6ebbd97fc

Issue #2592, opened 2024-01-15
Replace `rumqttc` with a crate that is reliable enough to implement the bridge in the mapper


didier-wenzek:
The bridge,  from the gateway device to the cloud, is the only explicit dependency of thin- edge on mosquitto. Furthermore, this bridge is working as expected only with [a very specific version of mosquitto](https://github.com/thin-edge/thin-edge.io/issues/2485#issuecomment-1837980002) (namely 2.0.18).

It would be good to have the freedom to choose the MQTT broker and to have a less strong dependency on a very specific version of mosquitto. So, __why not let the mapper implement itself the bridge feature?__ Republishing any messages received on a `c8y/#` topic to the corresponding topic on the cloud endpoint (and vice versa).

Unfortunately, the crate currently used (i.e. [`rumqttc`](https://lib.rs/crates/rumqttc)) is not robust enough to implement a bridge that is *reliable*. This has to be confirmed, but this crate provides no way to control acknowledgements for messages with QoS 1 and 2.

One needs an MQTT crate that supports:
- MQTT 3.1
   - at least the features used by [mqtt_channel](https://github.com/thin-edge/thin-edge.io/tree/main/crates/common/mqtt_channel)
- TLS
  -  with certificate-based authentication of the broker and the client.
- Reliable support of QoS 1 and 2.
  - either to be able to acknowledge the reception of a message from one peer only when its copy has been acknowledged by the other peer
  - or to be able to acknowledge the reception of a message only when persisted on disk and ready to be republished till the other peer acknowledges it.
- cross-compilation to various architecture
  - ideally as a static lib and using `cargo zigbuild`.

jarhodes314:
So the situation with the crates appears to be as follows:
## Rumqttc
Pros:
- Easy to enable manual acknowledgements
- Implemented in Rust and an existing dependency, so there won't be unexpected issues with (cross) compilation
Cons:
- Difficult (although I don't think impossible) to establish when a published message has been acked
- No built-in support for persistence, this would have to be dealt with manually

## Paho
Pros:
- Based on a widely used C library, so more "battle tested" than rumqttc
- Simple interface to handle persistence without the need to do so manually
Cons:
[...]

PR #2716, opened 2024-02-15
Add support for built-in bridge to tedge-mapper-c8y

jarhodes314:
## Proposed changes
This adds the bridge connection to the mapper discussed in #2592. It is currently Cumulocity only, and doesn't support all the possible mosquitto connection restrictions, but does function correctly on a basic level.

It also adds support for a configurable topic prefix in the mapper bridge via the configuration `c8y.bridge.topic_prefix`.

In 556a661f8bd3d447ca3130e503cf9bb301cf55bc, I've added a (hidden) configuration ~`c8y.bridge.in_mapper`~ `c8y.bridge.built_in` to control where the mapper runs. When set to `false`, currently the default, this should work like it does now: the mosquitto bridge configuration is generated as before, and the mapper doesn't attempt to run a bridge. When set to `true`, the mapper will run the bridge components, and the mosquitto configuration will be deleted. Upgrading tedge and then running `tedge reconnect c8y` should now work, including the connectivity check and setting the permissions for the certificates. The system tests added in this PR verify this.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+                            error!("MQTT bridge connection error: {err}");
+                        }
+                        last_error = Some(current_error);
+                        continue;
+                    }
+                };
+                if let Event::Incoming(Incoming::Publish(publish)) = &notification {
+                    cloud_client
+                        .publish(
+                            publish.topic.strip_prefix(&topic_prefix).unwrap(),
```
didier-wenzek, line 151: Rather than some mapping hard-coded inside the bridge logic, I would consider to pass a function that convert a source topic to a target topic.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+                            publish.retain,
+                            publish.payload.clone(),
+                        )
+                        .await
+                        .unwrap();
+                    local_client.ack(publish).await.unwrap();
+                }
+            }
+        });
+        let cloud_to_local = tokio::spawn(async move {
```
didier-wenzek, line 127: Instead of two independent loops, it would be better to have a single bridge logic that is used twice, with different mapping parameters.

I even wonder that these two loops might have to be merged with a `select!` over the two end-points. Indeed, one has to listen not only for messages from the source end-point but also for acknowledgements from the target end-point.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
albinsuresh: Shouldn't we consider externalising the mapping rules, where the mapping rules can be defined in an external config file(similar to what Mosquitto has done) and use that for the mapping? That way, we can reuse the same bridge actor with all the cloud mappers using diff config files. It also enables the user to extend the mapper with additional mapping rules for things like custom templates.

reubenmiller: We need to think carefully on this. We don't need to make the mapping too open 

1. Mappings which are critical to the function of the agent (e.g. `c8y/s/ds` and `c8y/s/us` etc.)
2. Optional user specified mapping (e.g. `c8y/dc/<TemplateCollectionID>` etc.)

The first time of mappings should not be influenced by the user as this would lead to a non-functional agent

The second point is already configurable via the tedge.toml file, so we don't need a second level of configuration which converts the tedge.toml to the a mapping file rule. We could still open up to allow users to subscribe/publish additional topics (this will be important for the Cumulocity IoT MQTT Connect feature)...though we can still review the configuration mechanisms for the custom SmartREST templates etc.

albinsuresh: > The first time of mappings should not be influenced by the user as this would lead to a non-functional agent

Yeah definitely, the core and extended mapping rules must be separate to avoid the user messing up the core rules. How about keeping the core mapping rules in a separate file, which should not be directly manipulated by the user for any extensions?

Keeping the mapping rules outside the rust code makes the development of a new mapper easier (without even needing to know Rust), right? The goal was to achieve both "mqtt broker independence" and "generic mapping" with a single design.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+
+#[async_trait]
+impl Actor for MqttBridgeActor {
+    fn name(&self) -> &str {
+        "MQTT-Bridge"
+    }
+
+    async fn run(mut self) -> Result<(), RuntimeError> {
+        Ok(())
+    }
```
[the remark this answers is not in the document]
jarhodes314: If this isn't done, we can't guarantee the MQTT connection is active before other actors start, which means the JWT retriever attempts to retrieve the JWT but fails on the first attempt.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: In my mind, the primary reason for the existence of the mapping rules we have in any case is so tedge-mapper can function correctly. It therefore seems a bit daft to me to store these outside of the mapper code (though they will need relocating to cope with non-Cumulocity clouds).

Additionally, I think the main hinderance to "generic mapping" at the moment is merely the fact it requires additional effort at the moment for someone to update mosquitto configurations, and so long as we don't need to do that, I don't think we're likely to introduce significant barriers to that being a possibility in the future. In general, it seems like a bad idea to try and attack both replacing the existing bridge configuration with something equivalent in the mapper and adding new features to the bridge (beyond trivial matters like automatically subscribing the bridge to new custom smartrest topics).

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: That makes sense. I would be fearful of using `select!` here, it sounds like a recipe for recreating the bug in #2326, but it would make sense to limit the duplication of the logic code.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
albinsuresh: > In my mind, the primary reason for the existence of the mapping rules we have in any case is so tedge-mapper can function correctly. It therefore seems a bit daft to me to store these outside of the mapper code (though they will need relocating to cope with non-Cumulocity clouds).

Externalizing the bridge rules is not needed if we want to keep things in the current form, with dedicated mapper processes for each cloud. But, with the desire to have a "generic mapper" eventually, which I'd assume will cater to all the clouds, an externalized config file seemed like the easiest way, especially to enable no-code development of mappers.

> But we don't need an intermediate file format...we only used that in the past as it was required by mosquitto, but now we are in control if it, then we don't need it.

Good point. This makes me wonder if we should reconsider the current mapping/bridging approach as well. So far, we needed the `c8y/#` and `az/#` like bridge topics, just because the bridge was external. So, the mapper was forced to send the mapped messages to the broker so that the external bridge client can pick them up. But now that we have the converter and the bridge in the same process, the messages between the converter actor and the bridge actor can be exchanged via in-memory channels (message boxes) rather than the broker, eliminating all the serializing/deserializing cost + network hops via the broker.

Obviously, this would be a much bigger change and the approach taken in this PR can be considered as the first step towards that goal.

> We still need to decide what the generic mapping looks like, as I think it is too early to tell if this is configuration, templates or code modules.

Fair point. I was just trying to see if there's a common ground here.

In my head, the generic mapper is something that behaves like(**not directly using**) the mosquitto bridge, but with a more sophisticated mapping rule-set for topic/payload manipulations as well beyond the simple topic-to-topic bridging offered by mosquitto. So, an external config file with just the bridge config aspects felt like a good first step, which can later be enhanced with the relevant mapping rules as well.

But, having this conversation, I get a feeling that we all have different views on how that generic mapper would work, and hence may be it's too early to discuss those aspects under the scope of this ticket.

crates/core/c8y_api/src/smartrest/topic.rs
```diff
@@ -4,64 +4,62 @@ use mqtt_channel::Topic;
...
+const SMARTREST_PUBLISH_TOPIC: &str = "s/us";
+const SMARTREST_SUBSCRIBE_TOPIC: &str = "s/ds";
 
 #[derive(Debug, Clone, Eq, PartialEq)]
 pub enum C8yTopic {
     SmartRestRequest,
     SmartRestResponse,
     ChildSmartRestResponse(String),
-    OperationTopic(String),
+    // OperationTopic(String),
```
jarhodes314, line 17: TODO remove me?

crates/core/c8y_api/src/smartrest/topic.rs
```diff
@@ -4,64 +4,62 @@ use mqtt_channel::Topic;
...
-            Self::SmartRestRequest => SMARTREST_SUBSCRIBE_TOPIC.into(),
-            Self::SmartRestResponse => SMARTREST_PUBLISH_TOPIC.into(),
+            Self::SmartRestRequest => format!("{prefix}/{SMARTREST_SUBSCRIBE_TOPIC}"),
+            Self::SmartRestResponse => format!("{prefix}/{SMARTREST_PUBLISH_TOPIC}"),
             Self::ChildSmartRestResponse(child_id) => {
-                format!("{}/{}", SMARTREST_PUBLISH_TOPIC, child_id)
-            }
-            Self::OperationTopic(name) => name.into(),
+                format!("{prefix}/{SMARTREST_PUBLISH_TOPIC}/{child_id}")
+            } // Self::OperationTopic(name) => name.into(),
```
jarhodes314, line 54: TODO remove me?

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+                    .await
+                    .unwrap();
+            }
+            Event::Outgoing(Outgoing::Publish(pkid)) => {
+                if let Some(msg) = rx_pubs.next().await {
+                    forward_pkid_to_received_msg.insert(pkid, msg);
+                } else {
+                    break;
+                }
+            }
```
didier-wenzek, line 167: Is this really robust, as this requires a perfect synchronization between two event streams? Sure these streams are not independent (the companion loop triggers one event for this `rx_pubs` and publishes a message triggering  the `OutGoing` event. But I would be cautious here. Mainly because of QoS subtleties. Do we have such an outgoing event for QoS 0? If not, then no message should be propagated to the companion loop.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+                target
+                    .publish(
+                        transform_topic(&publish.topic),
+                        publish.qos,
+                        publish.retain,
+                        publish.payload.clone(),
+                    )
+                    .await
+                    .unwrap();
+                tx_pubs.send(publish).await.unwrap();
```
didier-wenzek, line 149: Do we also have to do this for QoS 0?

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+    }
+}
+
+async fn one_way_bridge<F: for<'a> Fn(&'a str) -> Cow<'a, str>>(
+    mut recv_event_loop: EventLoop,
+    target: AsyncClient,
+    transform_topic: F,
+    mut tx_pubs: mpsc::Sender<Publish>,
+    mut rx_pubs: mpsc::Receiver<Publish>,
+) {
```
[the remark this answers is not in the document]
jarhodes314: There was some equivalent to the assertion with a call to `unwrap()` when  calling `forward_pkid_to_received_msg.remove()`. By unwrapping the value, we ensure the packet id is known, and this should only happen reliably if everything is working correctly. If we have accidentally read packet ids for the cloud when we should be reading them for the local connection, one of the directions will quickly encounter an acknowledgement for an unknown packet id. I've currently removed this since I'm confident the loop now works, and I think this particular bug is difficult to achieve due to the required input to the function.

I've also modified the channels to be a single bidirectional channel to avoid exposing states which don't make sense, and (mostly) documented the function (still a little bit to do there, and ascii art will likely help)

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: I replied to this mostly in https://github.com/thin-edge/thin-edge.io/pull/2716#discussion_r1517673228. I think the answer is yes, we are robust as QoS 0 messages do go through the same series of events, just the packet identifiers are all 0.

didier-wenzek: > just the packet identifiers are all 0

Okay then

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: > What I'm less sure about is how, if at all, we may wish to protect ourselves against misbehaving MQTT clients. If we aren't receiving acknowledgements, this could result in the map from packet id to original message growing an unbounded amount.

This can be controlled setting a bound to the [number of in flight messages](https://docs.rs/rumqttc/latest/rumqttc/struct.MqttOptions.html#method.set_inflight)

crates/extensions/tedge_mqtt_bridge/src/lib.rs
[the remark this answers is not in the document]
didier-wenzek: > I've also modified the channels to be a single bidirectional channel to avoid exposing states which don't make sense

This is indeed a good idea.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,305 @@
...
+/// This allows the loop to store the message along with its packet ID when the forwarded message is
+/// published. When an acknowledgement is received for the forwarded message, the packet id is used
+/// to retrieve the original [Publish], which is then passed to [AsyncClient::ack] to complete the
+/// final step of the message flow.
+///
+/// The channel sends [Option<Publish>] rather than [Publish] to allow the bridge to send entirely
+/// novel messages, and not just forwarded ones, as attaching packet IDs relies on pairing every
+/// [Outgoing] publish notification with a message sent by the relevant client. This means the
+///
+/// # Health topics
```
didier-wenzek, line 189: > /// The channel sends [Option<Publish>] rather than [Publish] to allow the bridge to send entirely
/// novel messages, and not just forwarded ones, as attaching packet IDs relies on pairing every
/// [Outgoing] publish notification with a message sent by the relevant client. This means the

I don't get it. What is the point to send a `None` to the half? Or to send *entirely novel* message? Even is useful, how `Option` can help?

> This means the

missing sentence

> # Health topics

missing content

jarhodes314: Hopefully all is clearer now I've finished the paragraph and the next section

crates/extensions/tedge_mqtt_bridge/src/lib.rs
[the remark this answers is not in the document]
didier-wenzek: Here is some ascii art describing how the bridge works https://gist.github.com/didier-wenzek/d37d7cdb8bdbd5b0b572f6fb81d42801

crates/extensions/tedge_mqtt_bridge/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+homepage = { workspace = true }
+repository = { workspace = true }
+
+[features]
+# No features on by default
+default = []
+
+[dependencies]
+async-trait = { workspace = true }
+c8y_api = { workspace = true }
```
didier-wenzek, line 18: It would be really better not to depend on `c8y_api`.

Some dependencies are not used (e.g. `serde_json` and `tedge_utils`). Maybe some more.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,313 @@
+use async_trait::async_trait;
+use c8y_api::utils::bridge::main_device_health_topic;
+use c8y_api::utils::bridge::C8Y_BRIDGE_DOWN_PAYLOAD;
+use c8y_api::utils::bridge::C8Y_BRIDGE_UP_PAYLOAD;
```
didier-wenzek, line 4: These definitions can be moved to `tedge_api` to avoid this dependency on `c8y_api`.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+            Ok(notification) => {
+                if last_err.as_ref().is_some() {
+                    info!("MQTT bridge connected to {name} broker");
+                    last_err = None;
+                    if let Some(health_topic) = &health_topic {
+                        target
+                            .publish(health_topic, QoS::AtLeastOnce, true, C8Y_BRIDGE_UP_PAYLOAD)
+                            .await
+                            .unwrap();
+                        companion_bridge_half.send(None).await.unwrap();
```
didier-wenzek, line 222: This line deserves a comment, as having a de-synchronization of the two bridge halves is the Achille's heel of this algorithm.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -444,7 +451,19 @@ define_tedge_config! {
                 #[tedge_config(note = "If set to 'auto', this cleans the local session accordingly the detected version of mosquitto.")]
                 #[tedge_config(example = "auto", default(variable = "AutoFlag::Auto"))]
                 local_cleansession: AutoFlag,
-            }
+            },
+
+            #[tedge_config(default(value = false))]
+            #[doku(skip)] // Hide the configuration in `tedge config list --doc`
+            in_mapper: bool,
```
didier-wenzek, line 458: Why hide this setting?

didier-wenzek (commented):
This PR can be merged soon. The built-in bridge implementation is convincing enough and the `config.c8y.bridge.in_mapper` setting ensures a smooth transition.

That said, a system test is missing. A test running with `config.c8y.bridge.in_mapper = true` and checking that measurements/alarms/events/operations are forwarded to/from Cumulocity.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -444,7 +451,19 @@ define_tedge_config! {
...
+
+            // TODO validation
+            /// The topic prefix that will be used for the mapper bridge MQTT topic. For instance,
+            /// if this is set to "c8y", then messages published to `c8y/s/us` will be
+            /// forwarded by to Cumulocity on the `s/us` topic
+            #[tedge_config(example = "c8y", default(value = "c8y"))]
+            #[doku(skip)] // Hide the configuration in `tedge config list --doc`
+            topic_prefix: TopicPrefix,
```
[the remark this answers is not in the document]
jarhodes314: Understood. I don't quite know if/why that makes a difference to this, as this is the local prefix. But this will impact the bridge library, which will need to support remote prefixes

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
jarhodes314: I want to get a bit further with the feature before finalising on what configurations we expose, especially while there are rough edges.

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -204,20 +205,20 @@ where
 
 /// Helper to generate a SmartREST operation status message
 pub trait OperationStatusMessage {
-    fn executing() -> Message {
-        Self::create_message(Self::status_executing())
+    fn executing(prefix: &TopicPrefix) -> Message {
+        Self::create_message(Self::status_executing(), prefix)
     }
```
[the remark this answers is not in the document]
jarhodes314: I don't quite understand how that would work. These methods are currently all static, so there's no ability to pass the prefix in via a `self` argument, and we can't include it directly in the code as it's configurable at runtime.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+    async fn update(
+        &mut self,
+        result: &NotificationRes,
+        companion_bridge_half: &mut BidirectionalChannelHalf<Option<Publish>>,
+    ) {
+        let name = self.name;
+        let (err, health_payload) = match result {
+            Ok(_) => {
+                info!("MQTT bridge connected to {name} broker");
```
didier-wenzek, line 351: This leads to several log entries for a regular message ;-)

=> This should be logged only for `Ok(Event::Incoming(Incoming::ConnAck(_)))`.
=> We should also distinguish the case where [ConnAck::code](https://docs.rs/rumqttc/latest/rumqttc/mqttbytes/v4/struct.ConnAck.html) is [not successful](https://docs.rs/rumqttc/latest/rumqttc/mqttbytes/v4/enum.ConnectReturnCode.html)

didier-wenzek:
I noticed that the `tedge connect c8y --test` command is no more working as expected.

```
✗ tedge connect c8y --test                                             
Detected mosquitto version >= 2.0.0
Error: failed to test connection to Cumulocity cloud.

Caused by:
    Device is not connected to Cumulocity cloud
```

While the built-in bridge is properly up and running:

```
✗ tedge mqtt sub 'te/+/+/+/+/status/health'
[te/device/main/service/mosquitto-c8y-bridge/status/health] 0
[te/device/main/service/tedge-mapper-c8y/status/health] {"pid":277988,"status":"up","time":1710251252.6182034}
[te/device/main/service/tedge-mapper-bridge-c8y/status/health] 1
```

And telemetry data actually forwarded to c8y.

jarhodes314:
> I noticed that the `tedge connect c8y --test` command is no more working as expected.
> 
> ```
> ✗ tedge connect c8y --test                                             
> Detected mosquitto version >= 2.0.0
> Error: failed to test connection to Cumulocity cloud.
> 
> Caused by:
>     Device is not connected to Cumulocity cloud
> ```
> 
> While the built-in bridge is properly up and running:
> 
> ```
> ✗ tedge mqtt sub 'te/+/+/+/+/status/health'
> [te/device/main/service/mosquitto-c8y-bridge/status/health] 0
> [te/device/main/service/tedge-mapper-c8y/status/health] {"pid":277988,"status":"up","time":1710251252.6182034}
> [te/device/main/service/tedge-mapper-bridge-c8y/status/health] 1
> ```
> 
> And telemetry data actually forwarded to c8y.

Fixed. I forgot this API existed, but it was checking for the existence of the bridge config file, which obviously doesn't exist with the built-in bridge

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,402 @@
...
+                companion_bridge_half.send(publish).await.unwrap();
+            }
+
+            // Forward acks from event loop to target
+            Event::Incoming(
+                Incoming::PubAck(PubAck { pkid: ack_pkid })
+                | Incoming::PubRec(PubRec { pkid: ack_pkid }),
+            ) => {
+                if let Some(msg) = forward_pkid_to_received_msg.remove(&ack_pkid) {
+                    target.ack(&msg).await.unwrap();
```
albinsuresh, line 303: Shouldn't `target.ack(&msg).await.unwrap()` be done for `Incoming::PubComp` as well? Or is that done automatically if we don't have an explicit message handler here?

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -72,61 +74,62 @@ mod tests {
...
     use test_case::test_case;
+
     #[test_case(
-        "test_device",
-        "te/device/main/service/tedge-mapper-c8y/status/health",
-        r#"{"pid":"1234","status":"up"}"#,
-        "c8y/s/us",
-        r#"102,test_device:device:main:service:tedge-mapper-c8y,service,tedge-mapper-c8y,up"#;
-        "service-monitoring-thin-edge-device"
+    "test_device",
```
albinsuresh, line 79: It was better with the indentation IMO.

albinsuresh (commented):
Despite having very detailed documentation on the message flow, the message flow is still not fully clear to me. So, take my comments with a pinch of salt, as some of it may be just irrelevant, coming out of my broken understanding.

May be just a naive view but I still have this whole bridging imagined as a single `select!` loop over the local and remote event loops and I'm really struggling to move away from that mental model to what's proposed here.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: I've modified it to do both, although https://github.com/bytebeamio/rumqtt/blob/bfc9f9a6932d10ad085354ea1600daf5437399d0/rumqttc/src/eventloop.rs#L467-L473 indicates the latter isn't necessary

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
jarhodes314: I didn't mean to make this change.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+                .host()
+                .to_string(),
+            8883,
+        );
+        cloud_config.set_manual_acks(true);
+
+        cloud_config.set_transport(Transport::tls_with_config(tls_config.into()));
+        let topic_prefix = format!("{prefix}/");
+        let (local_client, local_event_loop) = AsyncClient::new(local_config, 10);
+        let (cloud_client, cloud_event_loop) = AsyncClient::new(cloud_config, 10);
```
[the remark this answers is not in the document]
jarhodes314: It should be. The client won't connect to the cloud until the event loop is polled AFAIK, and that definitely cleanly handles disconnection/reconnection/pinging.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+    async fn update(
+        &mut self,
+        result: &NotificationRes,
+        companion_bridge_half: &mut BidirectionalChannelHalf<Option<Publish>>,
+    ) {
+        let name = self.name;
+        let (err, health_payload) = match result {
+            Ok(_) => {
+                info!("MQTT bridge connected to {name} broker");
+                (None, MQTT_BRIDGE_UP_PAYLOAD)
```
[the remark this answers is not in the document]
albinsuresh: > It would make sense to use a JSON format message (although that sounds like a breaking API change to me)

Since the health endpoint service name itself is new (`tedge-mapper-bridge-c8y` and not `mosquitto-c8y-bridge`), it should be fine.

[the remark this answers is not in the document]
didier-wenzek: >> It would make sense to use a JSON format message (although that sounds like a breaking API change to me)

> Since the health endpoint service name itself is new (`tedge-mapper-bridge-c8y` and not `mosquitto-c8y-bridge`), it should be fine.

Beware that the logic to differentiate a bridge related topic is really weak: 
https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/service_monitor.rs#L29

crates/core/tedge/src/cli/certificate/create.rs
```diff
@@ -57,6 +57,7 @@ impl CreateCertCmd {
 
         let cert = KeyCertPair::new_selfsigned_certificate(config, &self.id, key_kind)?;
 
+        // TODO cope with broker user being tedge
```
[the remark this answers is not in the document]
jarhodes314: The mapper shouldn't need to run as root. The latest changes modify `tedge reconnect c8y` to cope with this case, changing the ownership of existing certificates to `tedge`, but I didn't bother to make the same changes here since it wasn't necessary and is more involved to test (and really does want system testing, so in my mind it's best to do these both in a follow up PR).

tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry_built-in_bridge.robot
```diff
@@ -0,0 +1,280 @@
...
+    Device Should Have Event/s    expected_text=test event    type=event_002    minimum=1    maximum=1
+
+*** Keywords ***
+
+Custom Setup
+    ${DEVICE_SN}=    Setup
+    Set Suite Variable    $DEVICE_SN 
+    Device Should Exist                      ${DEVICE_SN}
+    ThinEdgeIO.Execute Command    tedge config set c8y.bridge.built_in true
+    ThinEdgeIO.Execute Command    tedge config set c8y.bridge.topic_prefix custom-c8y-prefix
```
gligorisaev, line 274: Having in mind that you use variable for the prefix I would use it here as well, like:
``` ThinEdgeIO.Execute Command tedge config set c8y.bridge.topic_prefix ${C8Y_TOPIC_PREFIX} ```
Like this the test can be used in custom prefixes given during runtime
(e.g. ```robot -v C8Y_TOPIC_PREFIX:some prefix thin-edge_device_telemetry_built-in_bridge.robot```)

PR #2741, opened 2024-03-01
feat: compile on macos

reubenmiller:
## Proposed changes

Support developing thin-edge.io on MacOS to ease development. It does not yet included any CI workflow to build and run the tests, this will be done in a follow up PR due it requiring larger changes in the workflow setup.

It is not expected to be able to run a full thin-edge.io instance, however it is possible to run the tests and locally debug the process natively on MacOS.

Most of the changes are related to differences in the OS in the Rust tests, the most notable differences are:

* notify events are different between Linux and MacOS (e.g. Linux will published a `EventKind::Modify` event on new file creation, where as MacOS will use `EventKind::Create`). As a result some tests have been marked as skipped for now.
* MacOS does not have a group called `root`, but instead `staff` can be used, however the id should not be fixed

**Locally run Rust tests**

Below shows the Rust tests run locally on a MacOS M1 (MacOS 14)

```sh
$ uname -a
Darwin SAG-LJWQ4P62TY 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:30:44 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6000 arm64
```

```sh
cargo nextest run --no-fail-fast --all-features --all-targets
# ...
Summary [  28.058s] 1121 tests run: 1121 passed, 6 skipped
```

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/notify.rs
+++ b/crates/common/tedge_utils/src/notify.rs
@@ -222,6 +222,7 @@ mod tests {
         }
     }
 
+    #[cfg_attr(target_os = "macos", ignore)]
     #[tokio::test]
     async fn test_multiple_known_files_watched() {
         let ttd = Arc::new(TempTedgeDir::new());
@@ -247,6 +248,7 @@ mod tests {
         file_handler.await.unwrap();
     }
 
+    #[cfg_attr(target_os = "macos", ignore)]
     #[tokio::test]
     async fn it_works() {
         let ttd = Arc::new(TempTedgeDir::new());
@@ -280,6 +282,7 @@ mod tests {
             .unwrap();
     }
 
+    #[cfg_attr(target_os = "macos", ignore)]
     #[tokio::test]
     async fn test_multiple_unknown_files_watched() {
         let ttd = Arc::new(TempTedgeDir::new());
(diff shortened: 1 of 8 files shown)


reubenmiller:
> This sounds correct. Having no access to a Mac, I can only confirm that things are still working as expected on a Linux box!

@rina23q or @albinsuresh would you be able to build and run tests locally on MacOS (e.g. `cargo build`, and the cargo nextest (see PR description))

In the near future we will have a workflow to cover these points.

rina23q (approved):
Testing on MacOS works 👍 

```
$ uname -a
Darwin SAG-D7WH243Y0T 23.2.0 Darwin Kernel Version 23.2.0: Wed Nov 15 21:53:18 PST 2023; root:xnu-10002.61.3~2/RELEASE_ARM64_T6000 arm64
```

crates/common/tedge_utils/src/notify.rs
```diff
@@ -222,6 +222,7 @@ mod tests {
         }
     }
 
+    #[cfg_attr(target_os = "macos", ignore)]
```
albinsuresh, line 225: Instead of ignoring every test in this module individually, why not just skip the whole test module with a `#[cfg(target_os = "linux")]` annotation? As we'd most likely wanna skip these tests on all non-Linux envs.

reubenmiller: This is more a todo. The notify-fs works fine on MacOS however they events manifest slightly differently here...so ideally we would invest some time in updating the tests to see why some things are different here (but this is less priority...so ignoring just shows that we have todo's rather ignoring them completely which will lead to us forgetting about them).

PR #2744, opened 2024-03-03
refactor: use unix::fs::MetadataExt over linux variant to improve unix compatibility

reubenmiller:
## Proposed changes

Use [std::os::unix::fs::MetadataExt](https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html) over `std::os::linux::fs::MetadataExt` to improve compatibility for unix targets.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/file.rs
+++ b/crates/common/tedge_utils/src/file.rs
@@ -3,7 +3,7 @@ use nix::unistd::*;
 use std::fs;
 use std::io;
 use std::io::Write;
-use std::os::linux::fs::MetadataExt;
+use std::os::unix::fs::MetadataExt;
 use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


crates/common/tedge_utils/src/file.rs
```diff
@@ -3,7 +3,7 @@ use nix::unistd::*;
 use std::fs;
 use std::io;
 use std::io::Write;
-use std::os::linux::fs::MetadataExt;
+use std::os::unix::fs::MetadataExt;
```
albinsuresh, line 6: While we were using the `unix::fs` module for things like `PermissionsExt`, we still decided to use `MetadataExt` from `linux::fs` instead. There must have been something Linux specific that we were using in the past, which probably isn't relevant anymore.

albinsuresh (approved):

crates/common/tedge_utils/src/file.rs
reubenmiller: Though I just checked the difference between the two and the only method provided by the linux variant that is not provided in the unix one is `as_raw_stat` which has been marked as deprecated since `1.8.0`.

* https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html
* https://doc.rust-lang.org/std/os/linux/fs/trait.MetadataExt.html#tymethod.as_raw_stat

Though maybe this has changed over time, since the above doc links are for the latest Rust version, and sometimes it is hard to see which versions the traits were added in.

didier-wenzek: As far as I remember, we are using a linux specific API in a single place. This is the `fallocate` call (see https://github.com/thin-edge/thin-edge.io/pull/2745)

didier-wenzek (approved):
Approved

PR #2751, opened 2024-03-04
refactor: use generic RecommendedWatcher type over platform specific type

reubenmiller:
## Proposed changes

Use the generic [RecommendedWatcher](https://docs.rs/notify/latest/notify/) notify type to remove OS specific types, as `INotifyWatcher` is generally only supported on linux targets.

There should be no functional difference.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/notify.rs
+++ b/crates/common/tedge_utils/src/notify.rs
@@ -4,7 +4,7 @@ use notify::event::AccessMode;
 use notify::event::CreateKind;
 use notify::event::RemoveKind;
 use notify::EventKind;
-use notify::INotifyWatcher;
+use notify::RecommendedWatcher;
 use notify::RecursiveMode;
 use notify::Watcher;
 use notify_debouncer_full as debouncer;
@@ -65,7 +65,7 @@ pub enum NotifyStreamError {
 }
 
 pub struct NotifyStream {
-    debouncer: debouncer::Debouncer<INotifyWatcher, debouncer::NoCache>,
+    debouncer: debouncer::Debouncer<RecommendedWatcher, debouncer::NoCache>,
     pub rx: Receiver<(PathBuf, FsEvent)>,
 }
 


crates/common/tedge_utils/src/notify.rs
```diff
@@ -65,7 +65,7 @@ pub enum NotifyStreamError {
 }
 
 pub struct NotifyStream {
-    debouncer: debouncer::Debouncer<INotifyWatcher, debouncer::NoCache>,
+    debouncer: debouncer::Debouncer<RecommendedWatcher, debouncer::NoCache>,
```
didier-wenzek, line 68: I don't really see how this can help as [the latter is a type alias of the former](https://docs.rs/notify/latest/notify/type.RecommendedWatcher.html), but this is clearer and less dependent on `notify` internals.

didier-wenzek (approved):
Approved

crates/common/tedge_utils/src/notify.rs
reubenmiller: It helps as if you explicitly put the `INotifiyWatcher` then you have to manually use the `#[cfg(target_os = "linux")]` block, and then do your own logic/research to see which notifier is supported on different target os's. The RecommendedWatcher just allows thin-edge.io to take care of the underlying target_os specific types.

e.g..

```
/// The recommended `Watcher` implementation for the current platform
#[cfg(any(target_os = "linux", target_os = "android"))]
pub type RecommendedWatcher = INotifyWatcher;
/// The recommended `Watcher` implementation for the current platform
#[cfg(all(target_os = "macos", not(feature = "macos_kqueue")))]
pub type RecommendedWatcher = FsEventWatcher;
/// The recommended `Watcher` implementation for the current platform
#[cfg(target_os = "windows")]
pub type RecommendedWatcher = ReadDirectoryChangesWatcher;
/// The recommended `Watcher` implementation for the current platform
#[cfg(any(
    target_os = "freebsd",
    target_os = "openbsd",
    target_os = "netbsd",
    target_os = "dragonflybsd",
    all(target_os = "macos", feature = "macos_kqueue")
))]
```

reubenmiller: MacOS and freeBSD do not support iNotify, so compiling on those systems would throw an error.

PR #2752, opened 2024-03-04
tests: use posix shell over bash for improved cross platform compatibility

reubenmiller:
## Proposed changes

Use posix shell, `sh`, in tests which slighlty improves the cross platform compatibility (e.g. between Linux and MacOS).

One of the common problems that can occur is if the locale is not correctly configured in bash (which is unfornately the default MacOS settings), then a warning is printed on stderr when the script is launched. This affects the script output assertion causing it to fail. Using posix shell does not have the same problem.

Notes on bash to posix shell conversion:

* Use `seq N M` over `{N..M}` as the later is bash specific syntax
* Use short form of signal names, e.g. `TERM` instead of `SIGTERM` as the later is not supported by posix shell (though it has the same function)

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


didier-wenzek (approved):
Approved

PR #2753, opened 2024-03-04
feat(watcher): use canonical paths when checking if notify path is a child of the watcher path

reubenmiller:
## Proposed changes

Use canonical paths (e.g. resolving symlinks etc.) when comparing if notify paths are a child of the given watcher path.

On some platforms like MacOS, the watcher path and the paths in the notify messages can differ if the watcher path is a symlink. Most notably this occurs during testing as the default tmp directory is a symlink under MacOS (e.g. the temp folder is given as`/var/folders`, however this is actually a symlink to `/private/var/folders`).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_file_system_ext/src/lib.rs
+++ b/crates/extensions/tedge_file_system_ext/src/lib.rs
@@ -47,7 +47,16 @@ impl FsWatchMessageBox {
         log_message_sent("Inotify", &message);
 
         for (watch_path, sender) in self.watch_dirs.iter_mut() {
-            if path.starts_with(watch_path) {
+            // Try to use canonical paths to avoid false negatives when dealing with symlinks
+            if path
+                .canonicalize()
+                .unwrap_or_else(|_| path.to_path_buf())
+                .starts_with(
+                    watch_path
+                        .canonicalize()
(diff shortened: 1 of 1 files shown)


crates/extensions/tedge_file_system_ext/src/lib.rs
```diff
@@ -47,7 +47,16 @@ impl FsWatchMessageBox {
...
-            if path.starts_with(watch_path) {
+            // Try to use canonical paths to avoid false negatives when dealing with symlinks
+            if path
+                .canonicalize()
+                .unwrap_or_else(|_| path.to_path_buf())
+                .starts_with(
+                    watch_path
+                        .canonicalize()
+                        .unwrap_or_else(|_| watch_path.to_path_buf()),
+                )
```
didier-wenzek, line 58: I would avoid to do this canonicalization for each and every event.

It's better to:
1. canonicalize the `path` outside the loop.
2. canonicalize the `watch_path` when registered.

reubenmiller:
Superseded by https://github.com/thin-edge/thin-edge.io/pull/2758

Issue #2700, opened 2024-02-12
Add docs how to unregister a device/service from the entity store


reubenmiller:
**Is your documentation request related to a problem? Please describe.**

The [documentation](https://thin-edge.github.io/thin-edge.io/references/mappers/c8y-mapper/#registration) inform users how to register new child devices and services, however it does not detail how to remove a device or service (including the information that a device/service is not removed from the cloud, only locally).

**Describe the improvement you'd like**

Add a section detailing how to unregister/remove a child device and service.

For example:

```sh
tedge mqtt pub "te/device/mychild//" "" -r

tedge mqtt pub "te/device/mychild/service/myservice/" "" -r
```

Or if you want to unregister multiple devices using a pattern, then you can use `mosquitto_sub`:

**Example: Unregister all services for a child device**

```sh
mosquitto_sub -t "te/device/mychild/service/+" --remove-retained
```

**Example: Unregister all child devices**

```sh
mosquitto_sub -t "te/device/+//" -T "te/device/main//" --remove-retained
```
*Note:* The above command subscribes to the child devices and but excludes the `te/device/main//` device (if present)

**Note**
* Unregistering a device/child from the local thin-edge.io MQTT broker does not delete the device from the cloud. This is a protection mechanism to ensure the edge device does not delete any historical information from the cloud (e.g. measurements from an older service used for analysis).

**Additional context**

didier-wenzek:
What's the status of this ticket? Do we need still need to improve the documentation to register/unregister devices?

gligorisaev:
> What's the status of this ticket? Do we still need to improve the documentation to register/unregister devices?

The content provides comprehensive guidance on deregistering entities, covering various scenarios and considerations. It effectively balances technical detail with practical instructions, including explanations for MQTT-related concepts.

From my point of view it is OK

PR #2755, opened 2024-03-05
docs: Document commands to de-register entities #2700

albinsuresh:
## Proposed changes

Document commands to de-register entities.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2700

## Checklist

### DIFF
--- a/docs/src/operate/registration/deregister.md
+++ b/docs/src/operate/registration/deregister.md
@@ -1,0 +1,93 @@
+---
+title: Deregister entities
+tags: [Child-Device, Registration]
+sidebar_position: 1
+description: Deregister child devices and services from %%te%%
+---
+
+# Deregister entities
+
+All the entities (devices and services) registered with %%te%% are stored with the MQTT broker as retained messages,
+with their metadata spread across multiple topics.
+For example, for a child device `child01`, the registration message is stored on the `te/device/child01//` topic,
+its twin data is stored across several `te/device/child01///twin/<twin-data-type>` topics,
+its command metadata stored across several `te/device/child01///cmd/<cmd-type>` topics and so on.
+
+On top of that, when a device has services associated with it, or has nested child devices,
+they all have their own respective registration topics and other metadata topics.
+So, deregistering a device involves deregistering itself, its metadata and
+the complete entity hierarchy that is associated with it.
+
+Even though %%te%% doesn't provide a direct API for this yet, this can be easily be done using third-party tools
+like `mosquitto_sub` that supports clearing multiple retained messages together using the `--remove-retained` option.
+
+## Deregister a child device and its services
+
+When the topic ids of the devices and services follow the default topic scheme,
+which maintains the service topic ids (`te/device/<device-id>/service/<service-id>`)
+directly under the device topic id (`te/device/<device-id>//`) in the hierarchy,
+a child device along with all of its services can be deleted as follows:
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/child01/+/+/#"
+```
+
+* The topic filter `te/device/child01/+/+` covers the registration messages of the device and all of its services,
+  and the trailing `#` covers their their metadata messages as well.
+* The `--remove-retained` clears all the retained messages matching that topic filter.
+  Running the same command without the `remove-retained` option shows the list of messages
+  that would be cleared with this command, which can be used for dry runs before the entities are really removed.
+* The `-W 3` option is used to stop the command after a timeout period of 3 seconds, without which the command will not exit.
+  Adjust this timeout value based on the number of entities in your deployment.
+
+:::note
+The entities in the cloud must also be removed separately as deregistrations on the device
+are not propagated to the cloud.
+The Cumulocity mapper must also be restarted if the same entity is to be recreated after deregistration.
+:::
+
+A single service of a device (main or child), can be deregistered as follows:
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/main/service/service01/#"
+```
+
+All the services associated to a device can be deregistered together, as follows:
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/child01/service/+/#"
+```
+
+:::note
+De-registering parent and child entities together with a single command only works
+when the parent and children are hierarchically linked via the topics, which can be queried with a single topic filter,
+as in the case of services while following the default topic scheme.
+While using custom topic schemes, that do not maintain the same topic prefix for a device and its services,
+querying and deregistering them may not be a on-liner as in the examples listed above.
+:::
+
+## Deregister all child devices and services
+
+All the child devices and services can be deregistered using a wildcard topic filter that covers all the entities
+combined with a topic exclusion filter for the main device, as follows:
+
+```sh
+mosquitto_sub -v -t "te/device/+/#" -T "te/device/main/#"
+```
+
+## Unsupported cases
+
+While using the default topic scheme, the following use-cases are not supported as it's not possible to define 
+wildcard topic filters for the same, as the topic scheme does not capture the parent-child relationship of devices:
+
+* Deregister all child devices of a given device
+* Deregister the entire nested child device hierarchy of a device
+
+## Custom topic schemes
+
+Most of the commands listed above are applicable only when the default topic scheme is used.
+The same rules, especially the ones involving multiple entities, do not apply to custom topic schemes,
+as different entities would be linked differently based on the topic scheme that's used.
+The filtering capabilities would also very from scheme to scheme.
+As long as you can define topic filters to select individual entities or a set of entities,
+the same can be used for deregistering as well.
--- a/docs/src/operate/registration/index.md
+++ b/docs/src/operate/registration/index.md
@@ -1,0 +1,11 @@
+---
+title: Registration
+tags: [Registration, Deregistration]
+sidebar_position: 11
+---
+
+import DocCardList from '@theme/DocCardList';
+
+Managing registration and deregistration of child devices and services on %%te%%.
+
+<DocCardList />


docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
+---
+title: Deregister entities
+tags: [Lifecycle]
+sidebar_position: 1
+description: Deregister entities from thin-edge
+---
+
+# Deregister entities
+
+All the entities registered with thin-edge are stored with the MQTT broker as retained messages,
```
reubenmiller, line 10: Can you replace all of the "thin-edge" tags with the placeholder `%%te%%`.

Support for using a placeholder was added in https://github.com/thin-edge/tedge-docs/pull/63, to provide a more consistent product naming in the docs.

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
+---
+title: Deregister entities
+tags: [Lifecycle]
+sidebar_position: 1
+description: Deregister entities from thin-edge
```
reubenmiller, line 5: Use the thin-edge.io placeholder.

```suggestion
description: Deregister entities from %%te%%
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+the complete entity hierarchy that is associated with it.
+
+## Deregister a single entity
+
+De-registering a single entity, excluding its child entities, involves clearing its registration message and
+all the metadata messages associated with it from the MQTT broker.
+Even though thin-edge doesn't provide a direct API for this yet, this can be easily be done using third-party tools
+like `mosquitto_sub` that supports clearing multiple retained messages together using the `--remove-retained` option.
+
+For e.g, a child device with topic id `device/child01//` can be deregistered as follows:
```
reubenmiller, line 28:
```suggestion
For example, a child device with topic id `device/child01//` can be deregistered as follows:
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+Adjust this timeout value based on the number of entities in your deployment.
+
+:::note
+Deregistering the main device using this mechanism is not supported.
+:::
+
+Similarly, a service with topic id `device/main/service/service01` can be deregistered as follows:
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/main/service/service01/#"
```
reubenmiller, line 46: I'm fairly sure we can remove the timeout option, `-W 3` and replace it with `--retained-only` (but double check this, see [docs](https://mosquitto.org/man/mosquitto_sub-1.html)).

```suggestion
mosquitto_sub --remove-retained --retained-only -t "te/device/main/service/service01/#"
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Services of child devices can also be de-registered in a similar manner, using the child device topic id instead.
+
+:::note
+The Cumulocity mapper must be restarted for the de-registrations to take effect.
```
reubenmiller, line 52: > The Cumulocity mapper must be restarted for the de-registrations to take effect.

Are you sure this is true? I was pretty sure that a child device could be deregistered without having to restart the mapper. But if we do need to restart the mapper, then we need to create a ticket and not require a restart.

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/child01/service/+/#"
+```
+
+The device can also be deleted along with all of its services with a minor variation in the topic
+that includes the device as well, as follows:
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/child01/+/+/#"
```
reubenmiller, line 72: Can't the `#` wildcard pattern be used to delete the entity registration and the related telemetry/command topics?

```suggestion
mosquitto_sub --remove-retained -W 3 -t "te/device/child01/#"
```

reubenmiller: Sorry, ignore this comment...I misread the number of slashes...

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Deregister a child device and its services
+
+When the topic ids of the devices and services follow the default topic scheme,
+which maintains the service topic ids (`te/device/<device-id>/service/<service-id>`)
+directly under the device topic id (`te/device/<device-id>//`) in the hierarchy,
+all the services associated to a child device can be deregistered as follows:
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/child01/service/+/#"
```
reubenmiller, line 65: Shouldn't the topic be:

```suggestion
mosquitto_sub --remove-retained -W 3 -t "te/device/child01/service/#"
```

This would cover all of the following topics:

```
te/device/child01/service/myservice
te/device/child01/service/myservice/m/example
te/device/child01/service/myservice/cmd/software_update
te/device/child01/service/myservice/cmd/software_update/c8y-1234
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+Deregistering the main device using this mechanism is not supported.
+:::
+
+Similarly, a service with topic id `device/main/service/service01` can be deregistered as follows:
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/main/service/service01/#"
+```
+
+Services of child devices can also be de-registered in a similar manner, using the child device topic id instead.
```
reubenmiller, line 49:
```suggestion
Services of child devices can also be deregistered in a similar manner, using the child device topic id instead.
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+Deregistering parent and child entities together with a single command only works when
+the parent and children are hierarchically linked via the topics, as in the case of services
+while following the default topic scheme, which can be queried with a single topic filter.
+While using custom topic schemes, that do not maintain the same topic prefix for a device and its services,
+querying and deregistering them may not be a on-liner as in the previous example.
+:::
+
+## Deregister all child devices and services
+
+All the child devices and services can be de-registered using a wildcard topic filter that covers all the entities
```
reubenmiller, line 85:
```suggestion
All the child devices and services can be deregistered using a wildcard topic filter that covers all the entities
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+* Deregister the entire nested child device hierarchy of a device
+
+## Custom topic schemes
+
+Most of the commands listed above are applicable only when the default topic scheme is used.
+The same rules, especially the ones involving multiple entities, do not apply to custom topic schemes,
+as different entities would be linked differently based on the topic scheme that's used.
+The filtering capabilities would also very from scheme to scheme.
+As long as you can define topic filters to select individual entities or a set of entities,
+the same can be used for de-registering as well.
```
reubenmiller, line 106:
```suggestion
the same can be used for deregistering as well.
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+
+All the entities registered with thin-edge are stored with the MQTT broker as retained messages,
+with their metadata spread across multiple topics.
+For e.g, for a child device `child01`, the registration message is stored on the `te/device/child01//` topic,
+its twin data is stored across several `te/device/child01///twin/<twin-data-type>` topics,
+its command metadata stored across several `te/device/child01///cmd/<cmd-type>` topics and so on.
+
+On top of that, when a device has services associated with it, or has nested child devices,
+they all have their own respective registration topics and other metadata topics.
+So, de-registering a device involves de-registering itself, its metadata and
```
reubenmiller, line 18:
```suggestion
So, deregistering a device involves deregistering itself, its metadata and
```

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+
+```sh
+mosquitto_sub --remove-retained -W 3 -t "te/device/main/service/service01/#"
+```
+
+Services of child devices can also be de-registered in a similar manner, using the child device topic id instead.
+
+:::note
+The Cumulocity mapper must be restarted for the de-registrations to take effect.
+The entities in the cloud must also be removed separately as de-registrations on the device
```
reubenmiller, line 53:
```suggestion
The entities in the cloud must also be removed separately as deregistration on the device
```

docs/src/operate/lifecycle/index.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Lifecycle
+tags: [Lifecycle]
+sidebar_position: 11
+---
+
+import DocCardList from '@theme/DocCardList';
+
+Managing lifecycle of entities on thin-edge.
```
reubenmiller, line 9: Use placeholder.

```suggestion
Managing lifecycle of entities on %%te%%.
```

docs/src/operate/lifecycle/index.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Lifecycle
```
reubenmiller, line 2: I'm not convinced with the name, `Entity Registration` might be better, and it would align with the reference doc usage, http://localhost:3000/thin-edge.io/next/references/mqtt-api/#entity-registration.

didier-wenzek: And this *deregistration* section is a call for a new *registration* page.

docs/src/operate/lifecycle/deregister.md
rina23q, line 72: I think this command is the most demanded one from users as it deletes all things related to a single child device (except the alarm topic, but it's out of the scope of this doc). So, why not moving this section in the beginning of this doc?

docs/src/operate/lifecycle/index.md
albinsuresh: Yeah, currently most of it is in the references doc, which definitely isn't the first thing an end-user would look for.

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,106 @@
...
+its command metadata stored across several `te/device/child01///cmd/<cmd-type>` topics and so on.
+
+On top of that, when a device has services associated with it, or has nested child devices,
+they all have their own respective registration topics and other metadata topics.
+So, de-registering a device involves de-registering itself, its metadata and
+the complete entity hierarchy that is associated with it.
+
+## Deregister a single entity
+
+De-registering a single entity, excluding its child entities, involves clearing its registration message and
```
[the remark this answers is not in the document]
albinsuresh: Yes, and this is something we should prevent when we provide our own APIs. But at the MQTT level, the user is free to do this and mess things up. On the next restart of the mapper, the same child devices will get auto-created, if auto-registration is ON. Else the services messages will be ignored for lacking the parent.

docs/src/operate/lifecycle/deregister.md
albinsuresh: Yes. The mapper just ignores empty registration messages (clear messages). An empty message could have been interpreted as a "deregister request" resulting in the removal of that entity from the entity store, but it wasn't implemented that way due to the complexities in clearing other associated metadata messages as well from the broker. BTW, this is one of the API options that I'm considering for our "Deregistration API"

docs/src/operate/lifecycle/deregister.md
reubenmiller: yeah "de-registering" does seem to be the correct usage as a verb (at least according to [the search results](https://www.google.com/search?q=de-registering&oq=de-registering&gs_lcrp=EgZjaHJvbWUyCQgAEEUYORiABDIHCAEQABiABDIHCAIQABiABDIHCAMQABiABDIHCAQQABiABDIGCAUQABgeMgYIBhAAGB4yBggHEAAYHjIGCAgQABgeMgYICRAAGB7SAQgxOTA4ajBqN6gCALACAA&sourceid=chrome&ie=UTF-8))

docs/src/operate/lifecycle/deregister.md
albinsuresh: Even though the doc of `--retained-only` states that "Remove all retained messages on the server, assuming we have access to do so, **and then exit**", I don't see that happening in practice, even after explicitly sending a non-retained message. Anyway, it would only work when there is some non-retained message on the given topic filter, which isn't always guaranteed, right?

docs/src/operate/lifecycle/deregister.md
albinsuresh: That's not true for `#`. `te/device/child01/service/+/#` covers `te/device/child01/service/myservice` as well, in addition to `te/device/child01/service/xyz` and ``te/device/child01/service/`.

But, `te/device/child01/service/#` unnecessarily includes `te/device/child01/service` as well, which we don't want.

docs/src/operate/lifecycle/deregister.md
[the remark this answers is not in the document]
albinsuresh: Removed this error-prone example of deleting a single child device and replaced it with the one that deletes the device and its services.

docs/src/operate/lifecycle/deregister.md
albinsuresh: Moved as the first command as suggested.

docs/src/operate/lifecycle/deregister.md
albinsuresh: Keeping `de-registering` to satisfy the typo checker even though `deregistering` also seems generally valid.

docs/src/operate/lifecycle/deregister.md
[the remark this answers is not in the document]
albinsuresh: Keeping `de-registering` to satisfy the typo checks even though `deregistering` also seems generally valid.

docs/src/operate/lifecycle/deregister.md
rina23q: I'm curious, `de-registering` requires `-` whilst `deregistrations` doesn't?

docs/src/operate/lifecycle/deregister.md
reubenmiller: I learnt something new then. Thanks

docs/src/operate/lifecycle/deregister.md
albinsuresh: Yeah, it's weird. When I google the following variants without the `-` seemed valid:

1. deregister
2. deregistered
3. deregistering
4. deregistration

But the spellchecker plugin in our devcontainer was specifically unhappy with `deregistering` and `deregistration` while the other two were fine. But, it looks like the spell checker in our workflow is happy with the non-hyphen versions. So, I'll just use those everywhere to be consistent.

docs/src/operate/lifecycle/deregister.md
```diff
@@ -0,0 +1,93 @@
+---
+title: Deregister entities
+tags: [Deregister]
```
didier-wenzek, line 3: A better tag would be `Registration`. I also notice we have no tags related to chid-devices.

```suggestion
tags: [Child-Device, Registration]
```

didier-wenzek (approved):
Approved with a comment.

PR #2757, opened 2024-03-05
refactor: support building watchdog on non linux os targets

reubenmiller:
## Proposed changes

Update the dummy watchdog and errors to support building on non-linux os targets.

* Update the dummy start_watchdog function (as the body no longer matched the function signature)
* Use target os specific error enum declaration to clearing separate which errors are used in different os targets (to avoid littering specific errors with os specific targets, given that linux is the only target which supports the full systemd watchdog)

With this change it should now be possible to build on MacOS targets (though the tests will still fail due to various differences in os targets).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/core/tedge_watchdog/src/error.rs
```diff
@@ -1,17 +1,24 @@
...
 use mqtt_channel::MqttError;
-
+#[cfg(target_os = "linux")]
 use tedge_config::CertificateError;
+#[cfg(target_os = "linux")]
 use tedge_config::ConfigSettingError;
+#[cfg(target_os = "linux")]
 use tedge_config::TEdgeConfigError;
+#[cfg(target_os = "linux")]
+use time::error::Parse;
```
didier-wenzek, line 10: Instead of adding a target to all these lines, I would keep the error module linux specific and move the non-unix case in `dummy_watchdog.rs`.

reubenmiller: Thanks for pushing the fix for this.

crates/core/tedge_watchdog/src/dummy_watchdog.rs
```diff
@@ -1,3 +1,7 @@
+use std::path::PathBuf;
+
 pub async fn start_watchdog(_config_dir: PathBuf) -> Result<(), anyhow::Error> {
-    anyhow::Error::from(crate::error::WatchdogError::WatchdogNotAvailable)
+    Err(anyhow::Error::from(
+        crate::error::WatchdogError::WatchdogNotAvailable,
+    ))
```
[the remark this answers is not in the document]
reubenmiller: Yes, in the near future we will have a workflow to at least give a gentle reminder that we have unexpectedly broken support.

didier-wenzek (approved):
Approved

PR #2758, opened 2024-03-05
feat(watcher): use canonical path when watching a directory

didier-wenzek:
## Proposed changes

Use canonical path (e.g. resolving symlinks etc.) when watching a directory.

On some platforms like MacOS, the watcher path and the paths in the notify messages can differ if the watcher path is a symlink. Most notably this occurs during testing as the default tmp directory is a symlink under MacOS (e.g. the temp folder is given as`/var/folders`, however this is actually a symlink to `/private/var/folders`).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/tests/tedge_test_utils/src/fs.rs
+++ b/crates/tests/tedge_test_utils/src/fs.rs
@@ -60,6 +60,22 @@ impl TempTedgeDir {
         TempTedgeFile { file_path: path }
     }
 
+    pub fn link_dir(&self, original: &TempTedgeDir, link_name: &str) -> TempTedgeDir {
+        let root = Utf8Path::from_path(self.temp_dir.path()).unwrap();
+        let path = root.join(&self.current_file_path).join(link_name);
+
+        #[cfg(unix)]
+        std::os::unix::fs::symlink(original.path(), &path).unwrap();
+
+        #[cfg(windows)]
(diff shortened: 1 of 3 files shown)


crates/tests/tedge_test_utils/src/fs.rs
```diff
@@ -60,6 +60,32 @@ impl TempTedgeDir {
...
+            current_file_path: path,
+        }
+    }
+
+    #[cfg(windows)]
+    pub fn link_dir(&self, original: &TempTedgeDir, link_name: &str) -> TempTedgeDir {
+        let root = Utf8Path::from_path(self.temp_dir.path()).unwrap();
+        let path = root.join(&self.current_file_path).join(link_name);
+
+        std::os::windows::fs::symlink_dir(original.path(), &path).unwrap();
```
reubenmiller, line 81: This test in windows might be problematic as on IT managed windows laptops, generally the default seems to be to disallow creation of symlinks (without admin rights)...Is there someway we can mark this test as optional, so it does not run by default when running on windows?

reubenmiller: We've decided to include the test by default rather than hiding it behind an opt-in feature (as this would lead to the test being too easily forgotten).

Running the tests on public Github runner should not be affected by the admin rights issue.

rina23q (approved):
LGTM

PR #2770, opened 2024-03-08
refactor: Deprecate service provider and consumer traits

didier-wenzek:
## Proposed changes

- [x] Deprecate the ServiceProvider and ServiceConsumer trait
- [x] Simplify the MessageSource and MessageSink traits
     - [x] Deprecate MessageSink::get_config 
     - [x] Remove Config type parameter from MessageSink
- [x] Fix ~~or deprecate~~ the Probe test helper (used nowhere, but demonstrating well actor ease of connectivity)

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_actors/src/builders.rs
+++ b/crates/core/tedge_actors/src/builders.rs
@@ -102,9 +102,6 @@ pub struct NoConfig;
 /// For that purpose each source implementation defines a `Config` type parameter,
 /// and the sink has to provide the configuration value specific to its needs.
 pub trait MessageSink<M: Message, Config> {
-    /// Return the config used by this actor to connect the message source
-    fn get_config(&self) -> Config;
-
     /// Return the sender that can be used by peers to send messages to this actor
     fn get_sender(&self) -> DynSender<M>;
 
@@ -299,9 +296,6 @@ pub trait RuntimeRequestSink {
 /// #    }
 /// # }
 /// # impl MessageSink<MyActorInput, NoConfig> for MyActorBuilder {
-/// #    fn get_config(&self) -> NoConfig {
-/// #        NoConfig
-/// #    }
 /// #    fn get_sender(&self) -> DynSender<MyActorInput> {
 /// #        self.messages.get_sender()
 /// #    }
@@ -413,10 +407,6 @@ impl<I: Message, O: Message, C> MessageSource<O, C> for SimpleMessageBoxBuilder<
 
 /// A `SimpleMessageBoxBuilder<Input,Output>` is a [MessageSink] of `Input` messages with no specific config.
 impl<I: Message, O: Message> MessageSink<I, NoConfig> for SimpleMessageBoxBuilder<I, O> {
-    fn get_config(&self) -> NoConfig {
-        NoConfig
-    }
-
     fn get_sender(&self) -> DynSender<I> {
         self.input_sender.sender_clone()
     }
--- a/crates/core/tedge_actors/src/builders.rs
+++ b/crates/core/tedge_actors/src/builders.rs
@@ -187,8 +187,8 @@ pub trait MessageSource<M: Message, Config> {
     fn register_peer(&mut self, config: Config, sender: DynSender<M>);
 
     /// Connect a peer actor that will consume the message produced by this actor
-    fn add_sink(&mut self, peer: &impl MessageSink<M, Config>) {
-        self.register_peer(peer.get_config(), peer.get_sender());
+    fn add_sink<C>(&mut self, config: Config, peer: &impl MessageSink<M, C>) {
+        self.register_peer(config, peer.get_sender());
     }
 }
 
--- a/crates/core/tedge_actors/src/builders.rs
+++ b/crates/core/tedge_actors/src/builders.rs
@@ -109,12 +109,12 @@ pub trait MessageSink<M: Message, Config> {
     fn get_sender(&self) -> DynSender<M>;
 
     /// Add a source of messages to the actor under construction
-    fn add_input<N>(&mut self, source: &mut impl MessageSource<N, Config>)
+    fn add_input<N, C>(&mut self, config: C, source: &mut impl MessageSource<N, C>)
     where
         N: Message,
         M: From<N>,
     {
-        source.register_peer(self.get_config(), self.get_sender().sender_clone())
+        source.register_peer(config, self.get_sender().sender_clone())
     }
 
     /// Add a source of messages to the actor under construction, the messages being translated on the fly.
@@ -129,6 +129,7 @@ pub trait MessageSink<M: Message, Config> {
     /// # use tedge_actors::MessageReceiver;
     /// # use tedge_actors::MessageSink;
     /// # use tedge_actors::NoMessage;
+    /// # use tedge_actors::NoConfig;
     /// # use tedge_actors::Sender;
     /// # use tedge_actors::SimpleMessageBox;
     /// # use tedge_actors::SimpleMessageBoxBuilder;
@@ -142,7 +143,7 @@ pub trait MessageSink<M: Message, Config> {
     /// let mut sender_builder = SimpleMessageBoxBuilder::new("Send", 16);
     ///
     /// // Convert the `&str` sent by the source into an iterator of `char` as expected by the receiver.
-    /// receiver_builder.add_mapped_input(&mut sender_builder, |str: &'static str| str.chars() );
+    /// receiver_builder.add_mapped_input(NoConfig, &mut sender_builder, |str: &'static str| str.chars() );
     ///
     /// let mut sender: SimpleMessageBox<NoMessage, &'static str>= sender_builder.build();
     /// let receiver: SimpleMessageBox<char, NoMessage> = receiver_builder.build();
@@ -161,9 +162,10 @@ pub trait MessageSink<M: Message, Config> {
     /// # Ok(())
     /// # }
     /// ```
-    fn add_mapped_input<N, MS, MessageMapper>(
+    fn add_mapped_input<N, C, MS, MessageMapper>(
         &mut self,
-        source: &mut impl MessageSource<N, Config>,
+        config: C,
+        source: &mut impl MessageSource<N, C>,
         cast: MessageMapper,
     ) where
         N: Message,
@@ -172,7 +174,7 @@ pub trait MessageSink<M: Message, Config> {
         MessageMapper: 'static + Send + Sync,
     {
         let sender = MappingSender::new(self.get_sender(), cast);
-        source.register_peer(self.get_config(), sender.into())
+        source.register_peer(config, sender.into())
     }
 }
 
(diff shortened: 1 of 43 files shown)


crates/core/tedge_actors/src/servers/builders.rs
```diff
@@ -201,16 +201,6 @@ impl<S: Server + Clone> Builder<ConcurrentServerActor<S>> for ServerActorBuilder
...
 
-impl<S: Server, K> ServiceProvider<S::Request, S::Response, NoConfig> for ServerActorBuilder<S, K> {
-    fn connect_consumer(
-        &mut self,
-        config: NoConfig,
-        response_sender: DynSender<S::Response>,
-    ) -> DynSender<S::Request> {
-        self.box_builder.connect_consumer(config, response_sender)
-    }
-}
```
didier-wenzek, line 212: Removing this implementation is the first step to deprecate the service provider and consumer traits.

Server actors have been the main motivation for the `ServiceProvider` trait. With the now deprecated original design, registering a client actor used to require an exchange of two senders (one given by the client and a second one built specifically for the client and returned by the server to the client). Now that clients send a reply-to address along their requests, the `MessageSink` trait is enough to register a client to a server.

crates/core/tedge_actors/src/builders.rs
```diff
@@ -213,82 +197,6 @@ pub trait RuntimeRequestSink {
...
 
-/// A trait that defines that an actor provides a service
-/// by accepting `Request` messages and sending `Response` from/to its peers.
-///
-/// In order to connect to a to `ServiceProvider<Req, Res, Conf>` and avail its services,
-/// the peer must be a `ServiceConsumer<Req, Res, Conf>`.
-///
-/// The config parameter is typically used by the `ServiceConsumer`
-/// to register any message filtering criteria to the `ServiceProvider`.
-pub trait ServiceProvider<Request: Message, Response: Message, Config> {
```
didier-wenzek, line 224: Deprecating this trait and its companion, the `ServiceConsumer` trait, is the heart of this PR.

There are two cases to consider when updating an actor which was using this trait.

### First case

When there is no request/response relationship between the message received and those sent by the provider actor. This is for instance the case for the MQTT messages received from and published to an MQTT actor.  An actor builder that needs to connect such an MQTT actor, will no more require a service provider:

~`mqtt: &mut impl ServiceProvider<MqttMessage, MqttMessage, TopicFilter>`~

but connect an actor that is a source of MQTT messages (received from the broker) as well as a sink of MQTT messages (to be published to the broker).


```
mqtt: &mut (impl MessageSource<MqttMessage, TopicFilter> + MessageSink<MqttMessage, NoConfig>),
```

### Second case

When there is a request/response relationship between the message received and those sent by the provider actor. This is for instance the case for the SetTimeout requests to the timer actor and its Timeout responses.

The expectation on the actor to connect to is no more a

~`timer: &mut impl ServiceProvider<SyncStart, SyncComplete, NoConfig>`~

but:

```
timer: &mut impl Service<SyncStart, SyncComplete>,
```

crates/core/tedge_actors/src/builders.rs
```diff
@@ -213,82 +197,6 @@ pub trait RuntimeRequestSink {
...
-    fn set_connection(
-        &mut self,
-        service: &mut impl ServiceProvider<Request, Response, Config>,
-    ) -> &mut Self
-    where
-        Self: Sized,
-    {
-        service.add_peer(self);
-        self
-    }
```
didier-wenzek, line 274: As the ServiceProvider and ServiceConsumer traits are deprecated, this convenient method can no more be used.

One has now to be explicit.

A connection to MQTT is bidirectional:

```
        let mqtt_publisher = mqtt.get_sender(); 
        mqtt.register_peer(mqtt_subscriptions, message_box.get_sender());
```

To connect a server, one can either use a `ClienbtMessageBox` to await the response of each request:

```
        let download_sender = ClientMessageBox::new(downloader_actor);
```

or to redirect the responses to another message box:

```
        let timer_sender = timer.add_requester(box_builder.get_sender());
```

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -99,7 +99,9 @@ async fn child_device_registration_mapping() {
     let cfg_dir = TempTedgeDir::new();
     let (mqtt, _http, _fs, mut timer, _ul, _dl) = spawn_c8y_mapper_actor(&cfg_dir, true).await;
 
-    timer.send(Timeout::new(())).await.unwrap(); // Complete sync phase so that alarm mapping starts
+    // Complete sync phase so that alarm mapping starts
+    trigger_timeout(&mut timer).await;
```
didier-wenzek, line 103: This change has been required because the test are now using a `FakeServerBox` that handles message envelopes behind the scene but can no more send a response before a request has been actually sent.

crates/core/tedge_actors/src/builders.rs
```diff
@@ -97,15 +97,14 @@ pub struct NoConfig;
 ///
 /// An actor whose builder is a `MessageSink<M, C>` can be connected to any other actor
 /// whose builder is a `MessageSource<M, C>` so that the sink can receive messages from that source.
-///
-/// A sink might be interested only in a subset of the messages emitted by the source.
-/// For that purpose each source implementation defines a `Config` type parameter,
-/// and the sink has to provide the configuration value specific to its needs.
-pub trait MessageSink<M: Message, Config> {
+pub trait MessageSink<M: Message> {
```
didier-wenzek, line 100: The `Config` type parameter is now useless (all the implementations were using `NoConfig`.

crates/core/tedge_actors/src/builders.rs
```diff
@@ -102,9 +102,6 @@ pub struct NoConfig;
 /// For that purpose each source implementation defines a `Config` type parameter,
 /// and the sink has to provide the configuration value specific to its needs.
 pub trait MessageSink<M: Message, Config> {
-    /// Return the config used by this actor to connect the message source
-    fn get_config(&self) -> Config;
-
```
didier-wenzek, line 123: All the implementations were either returning `NoConfig` or delegating some inner implementation.

crates/core/tedge_actors/src/builders.rs
```diff
@@ -113,24 +97,20 @@ pub struct NoConfig;
...
+pub trait MessageSink<M: Message> {
     /// Return the sender that can be used by peers to send messages to this actor
     fn get_sender(&self) -> DynSender<M>;
 
     /// Add a source of messages to the actor under construction
-    fn add_input<N>(&mut self, source: &mut impl MessageSource<N, Config>)
+    ///
+    /// A sink might be interested only in a subset of the messages emitted by the source.
+    /// This subset is defined by the config parameter.
+    fn add_input<N, C>(&mut self, config: C, source: &mut impl MessageSource<N, C>)
```
albinsuresh, line 108: Something for a follow-up PR: I personally find this `add_input` helper that takes a config counter-intuitive. In my head, the config is something meant for the source which is rightly captured in `MessageSource::register_peer`. So, I'd rather have the users use the source's `register_peer` API rather than this confusing indirection.

And if we're keeping it, I'd rename it to `add_source` or `connect_source` to be in sync with its `add_sink` counterpart.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -433,10 +434,10 @@ pub fn create_tedge_to_te_converter(
 
     // Tedge to Te converter
     let mut tedge_converter_actor =
-        ConvertingActor::builder("TedgetoTeConverter", tedge_to_te_converter, subscriptions);
+        ConvertingActor::builder("TedgetoTeConverter", tedge_to_te_converter);
 
-    tedge_converter_actor.add_input(mqtt_actor_builder);
-    tedge_converter_actor.add_sink(mqtt_actor_builder);
+    tedge_converter_actor.add_input(subscriptions, mqtt_actor_builder);
```
albinsuresh, line 439:
```suggestion
    mqtt_actor_builder.add_sink(subscriptions, &tedge_converter_actor);
```
This is the counter-intuitive bit that I was talking about earlier. This variant is clearer to me than the reversed version. But, this is just my personal view.

albinsuresh (approved):
Limiting the number of connection abstractions to just `MessageSource` and `MessageSink` was a great feat. As a next step, we can try to streamline the remaining connection abstractions as well. Here are a few points that I feel can be improved:

1. Improve the API names in API names in the `MessageSource` and `MessageSink`: For e.g, use symmetric names like `connect_source`, `connect_sink` so that it's more explicit at the user site where actors are connected.
2. Provide more guidance in the docs to users on why we have so many connection abstractions/indirections and when it is best to use which.

crates/core/tedge_actors/src/builders.rs
didier-wenzek: I'm fully aligned. And as even been tempted to make this change in the same PR. I decided otherwise - postponing this change to a follow-up PR focused on naming + docs.

crates/core/tedge_agent/src/agent.rs
didier-wenzek: Indeed, a personal view ;-) I prefer the reverse, .i.e. adding a source.

crates/core/tedge_agent/src/tedge_to_te_converter/tests.rs
```diff
@@ -327,12 +329,12 @@ async fn spawn_tedge_to_te_converter(
...
-        ConvertingActor::builder("TedgetoTeConverter", tedge_to_te_converter, subscriptions);
+        ConvertingActor::builder("TedgetoTeConverter", tedge_to_te_converter);
 
-    let mqtt_box = SimpleMessageBoxBuilder::new("MQTT", 5)
-        .with_connection(&mut tedge_converter_actor)
-        .build()
-        .with_timeout(Duration::from_millis(100));
+    let mut mqtt_box = SimpleMessageBoxBuilder::new("MQTT", 5);
+    mqtt_box.register_peer(subscriptions, tedge_converter_actor.get_input_sender());
+    tedge_converter_actor.register_peer(NoConfig, mqtt_box.get_sender());
```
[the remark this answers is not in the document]
didier-wenzek: Yeah, the `set_connection()` method is convenient. However, we need to avoid the previous issues conflating the two types of Config and too specialized on a mutual-connection (MQTT being used both as a source and a sink) ignoring the pipeline case (see for instance the batching actor used for the collectd mapper).

This can be discussed along renaming of the Sink and Source methods.

crates/core/tedge_actors/src/lib.rs
```diff
@@ -204,44 +196,33 @@
...
+//!     /// possibly using the configuration `config` to filter messages.
+//!     fn register_peer(&mut self, config: SomeConfig, sender: DynSender<SomeMessage>) {
 //!         todo!()
 //!     }
+//! }
 //!
-//!     /// Tell the service where to send its output messages to this actor
-//!     fn get_response_sender(&self) -> DynSender<SomeOutput> {
+//! /// An actor builder also declares that it consumes messages
+//! /// by implementing the `MessageSink` trait for the appropriate inputand config types.
```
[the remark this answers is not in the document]
didier-wenzek: fixed.

PR #2772, opened 2024-03-11
Add deregistering entities test

gligorisaev:
## Proposed changes

I have created the test case:

/tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot

but because there is no explicit checking if the de-registration is done, I consider this test as not complete.

I have tried several approches how to check the deregistration, but on all of them there was at least one step that could not be automated, needed to check on c8y GUI, which we avoid.

I am open for suggestions for changing this approach, also I think that after implementing all "deregistration tasks" that we intent to do we can create a propper and in detail test automation

## Types of changes

## Paste Link to the issue
https://github.com/thin-edge/thin-edge.io/pull/2755

## Checklist


tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,103 @@
...
+*** Test Cases ***
+
+Main device registration
+    ${mo}=    Device Should Exist              ${DEVICE_SN}
+    ${mo}=    Cumulocity.Device Should Have Fragment Values    name\=${DEVICE_SN}
+    Should Be Equal    ${mo["owner"]}    device_${DEVICE_SN}
+    Should Be Equal    ${mo["name"]}    ${DEVICE_SN}
+
+
+Child device registration
```
albinsuresh, line 20:
```suggestion
Child device deregistration
```
Instead of duplicating these tests from the `device_registration.robot`, how about adding the de-registration steps to the existing tests themselves?

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,103 @@
...
+Library    ThinEdgeIO
+
+Test Tags    theme:c8y    theme:registration
+Suite Setup    Custom Setup
+Test Setup    Test Setup
+Test Teardown    Get Logs    ${DEVICE_SN}
+
+*** Test Cases ***
+
+Main device registration
```
albinsuresh, line 13: Since main device deregistration is not supported, this test can be skipped.

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,103 @@
...
+    Should Be Equal    ${child_mo["owner"]}    device_${DEVICE_SN}    # The parent is the owner of the child
+    Should Be Equal    ${child_mo["name"]}     ${CHILD_SN}
+
+    # Check child device relationship
+    Cumulocity.Set Device    ${DEVICE_SN}
+    Cumulocity.Should Be A Child Device Of Device    ${CHILD_SN}
+
+    #Deregister Child device
+    Run Keyword And Ignore Error    Execute Command    mosquitto_sub --remove-retained -W 3 -t 'te/device/${CHILD_SN}//'
+
```
albinsuresh, line 35: One way to validate the de-registration is to deregister the device as done here, restart the mapper and then try re-registering it again with the same registration payload and validate if the corresponding 101 SmartREST message is published by the mapper. If you wanna do it completely via C8y, then you can delete the child device from C8y as well, after the deregistration step and then validate if the child device is recreated after the re-registration.

Once we implement the proper deregistration handling in the mapper (which removes the entries from the in-memory entity store on receipt of the clear messages), then the mapper restart won't be needed.

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,103 @@
...
+    # Check child device relationship
+    Cumulocity.Set Device    ${DEVICE_SN}
+    Cumulocity.Should Be A Child Device Of Device    ${CHILD_SN}
+
+    #Deregister Child device
+    Run Keyword And Ignore Error    Execute Command    mosquitto_sub --remove-retained -W 3 -t 'te/device/${CHILD_SN}//'
+    
+
+Register child device with defaults via MQTT
```
albinsuresh, line 37: This test can also be skipped, as the "defaults" (applicable only during registration) don't really influence the deregistration in anyway.

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,103 @@
...
+    # Check child registration
+    Check Child Device    parent_sn=${DEVICE_SN}    child_sn=${CHILD_XID}    child_name=${CHILD_SN}    child_type=linux-device-Aböut
+
+    # Check service registration
+    Check Service    child_sn=${CHILD_XID}    service_sn=${CHILD_XID}:service:custom-app    service_name=custom-app    service_type=custom-type    service_status=up
+
+    #Deregister service on a child device
+    Run Keyword And Ignore Error    Execute Command    mosquitto_sub --remove-retained -W 3 -t 'te/device/${CHILD_SN}/service/custom-app/#'
+
+
```
albinsuresh, line 57: Deregistration of services can also be tested the same way as described above for devices, by deleting the service and recreating it, which should succeed.

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
gligorisaev: I did this at the beginning and would preffer that approach, my intention was: as soon we have good solution for automating the deregistration to merge the test steps to one test.

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
gligorisaev: What about the main device service, isn't it needed for that?

albinsuresh: Yeah, deregistering main device services would be supported.

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
albinsuresh: Yeah, the failure is a result of using timeouts (the `-W` option) to end the command. But, `mosquitto_sub` exits with exit code 27 when it is a timeout. So, you can selectively ignore exit code 27, instead of ignoring everything to be more confident that the command did work.

reubenmiller: So this  could be refactored to:

```
Execute Command    mosquitto_sub --remove-retained -W 3 -t 'te/device/${CHILD_SN}//'    exp_exit_code=27
```

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,89 @@
...
+Enable clean start and collect logs
+    [Teardown]    Get Logs    ${DEVICE_SN}
+    Execute Command    sudo tedge config set c8y.entity_store.clean_start true
+    Restart Service    tedge-mapper-c8y
+    Service Health Status Should Be Up    tedge-mapper-c8y
+
+Check Child Device
+    [Arguments]    ${parent_sn}    ${child_sn}    ${child_name}    ${child_type}
+    ${child_mo}=    Device Should Exist        ${child_sn}
```
reubenmiller, line 63: We don't need to assign back to a variable if we're not going to use it.

```suggestion
    Device Should Exist        ${child_sn}
```

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,89 @@
...
+    [Teardown]    Get Logs    ${DEVICE_SN}
+    Execute Command    sudo tedge config unset c8y.entity_store.auto_register
+    Restart Service    tedge-mapper-c8y
+    Service Health Status Should Be Up    tedge-mapper-c8y
+
+Enable clean start and collect logs
+    [Teardown]    Get Logs    ${DEVICE_SN}
+    Execute Command    sudo tedge config set c8y.entity_store.clean_start true
+    Restart Service    tedge-mapper-c8y
+    Service Health Status Should Be Up    tedge-mapper-c8y
```
reubenmiller, line 59: These keywords don't seem to be used. Or if they do intend being used, then it might be better to put the PR is the "draft" state.

tests/RobotFramework/tests/cumulocity/registration/device_deregistration.robot
```diff
@@ -0,0 +1,89 @@
...
+
+Check Child Device
+    [Arguments]    ${parent_sn}    ${child_sn}    ${child_name}    ${child_type}
+    ${child_mo}=    Device Should Exist        ${child_sn}
+
+    ${child_mo}=    Cumulocity.Device Should Have Fragment Values    name\=${child_name}
+    Should Be Equal    ${child_mo["owner"]}    device_${DEVICE_SN}    # The parent is the owner of the child
```
reubenmiller, line 66: I would adjust the comment slightly as it could be misleading for nested child devices. The owner is the main device, in this case the parent device is the main device, however for nested child devices, then the "main device" would be the owner, and not the immediate parent.

```suggestion
    Should Be Equal    ${child_mo["owner"]}    device_${DEVICE_SN}    # The main device is the owner
```

gligorisaev:
Will create new PR

PR #2768, opened 2024-03-07
Stop returning an error when clearing command capabilities in c8y-mapper

rina23q:
## Proposed changes

Stop publishing an error on the `te/errors` topic when tedge-mapper-c8y receives command capability clearing messages.

P.S. I think it would be a nice to create a Robot test to cover device deregistration scenario.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
#2739

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2774, opened 2024-03-11
Add deregister test steps to register testcase

gligorisaev:
## Types of changes

## Paste Link to the issue
Covers:
https://github.com/thin-edge/thin-edge.io/pull/2755
https://github.com/thin-edge/thin-edge.io/pull/2768

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
+++ b/tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
@@ -3,7 +3,7 @@ Resource    ../../../resources/common.resource
 Library    Cumulocity
 Library    ThinEdgeIO
 
-Test Tags    theme:c8y    theme:registration
+Test Tags    theme:c8y    theme:registration    theme:deregistration
 Suite Setup    Custom Setup
 Test Setup    Test Setup
 Test Teardown    Get Logs    ${DEVICE_SN}
@@ -29,6 +29,18 @@ Child device registration
     # Check child device relationship
     Cumulocity.Set Device    ${DEVICE_SN}
     Cumulocity.Should Be A Child Device Of Device    ${CHILD_SN}
+
+    # Deregister Child device
+    Execute Command    mosquitto_sub --remove-retained -W 3 -t "te/device/${CHILD_SN}/+/+/#"    exp_exit_code=27
+    
+    # Check if deregistration was successful
+    Sleep    1s    reason=Allowing components to process messages
+    Should Have Retained Message Count    te/device/${CHILD_SN}/+/+/#    0     
+
+    # Checking if child device will be recreated after mapper restart
+    Restart Service    tedge-mapper-c8y    
+    Sleep    5s    reason=Allowing startup to complete
+    Should Have Retained Message Count    te/device/${CHILD_SN}/+/+/#    0  
 
 Register child device with defaults via MQTT
     Execute Command    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device"}'
@@ -73,6 +85,12 @@ Register service on a child device via MQTT
     # Check service registration
     Check Service    child_sn=${CHILD_XID}    service_sn=${CHILD_XID}:service:custom-app    service_name=custom-app    service_type=custom-type    service_status=up
 
+    # Deregister service on a child device
+    Execute Command    mosquitto_sub --remove-retained -W 3 -t 'te/device/${CHILD_SN}/service/custom-app/#'    exp_exit_code=27
+
+    # Check if deregistration was successful
+    Sleep    1s    reason=Allowing components to process messages
+    Should Have Retained Message Count    te/device/${CHILD_SN}/service/custom-app/#    0   
 
 Register devices using custom MQTT schema
     [Documentation]    Complex example showing how to use custom MQTT topics to register devices/services using
@@ -234,6 +252,11 @@ Entities send to cloud on restart
 
 *** Keywords ***
 
+Should Have Retained Message Count  
+    [Arguments]    ${topic}    ${exp_count}  
+    ${output}=    Execute Command    mosquitto_sub --retained-only -W 3 -t "${topic}" -v    exp_exit_code=27    return_stdout=True  
+    Length Should Be    ${output.splitlines()}    ${exp_count}  
+
 Re-enable auto-registration and collect logs
     [Teardown]    Get Logs    ${DEVICE_SN}
     Execute Command    sudo tedge config unset c8y.entity_store.auto_register
@@ -251,7 +274,7 @@ Check Child Device
     ${child_mo}=    Device Should Exist        ${child_sn}
 
     ${child_mo}=    Cumulocity.Device Should Have Fragment Values    name\=${child_name}
-    Should Be Equal    ${child_mo["owner"]}    device_${DEVICE_SN}    # The parent is the owner of the child
+    Should Be Equal    ${child_mo["owner"]}    device_${DEVICE_SN} 
     Should Be Equal    ${child_mo["name"]}     ${child_name}
     Should Be Equal    ${child_mo["type"]}     ${child_type}
 


tests/RobotFramework/tests/cumulocity/registration/registration_livecycle.robot
```diff
@@ -3,7 +3,7 @@ Resource    ../../../resources/common.resource
 Library    Cumulocity
 Library    ThinEdgeIO
 
-Test Tags    theme:c8y    theme:registration
+Test Tags    theme:c8y    theme:registration    theme:deregistrationgit
```
reubenmiller, line 6:
```suggestion
Test Tags    theme:c8y    theme:registration    theme:deregistration
```

reubenmiller (changes requested):
Aren't we missing the step on how to verify if the deregistration worked?

I understand that thin-edge.io does not yet support full removing a device from the cloud (which makes it hard to verify), however we should be doing at least another local check on the MQTT broker if the topic exists or not.

Or alternatively, we could manually delete the child device/service from the cloud (using the local c8y proxy), and then verify that restarting the mapper does not recreate the device in the cloud...once thin-edge.io supports deleting the device from the cloud, then all we have to do is remove the manual api call in the test.

tests/RobotFramework/tests/cumulocity/registration/registration_livecycle.robot
```diff
@@ -30,6 +30,19 @@ Child device registration
...
+    
+    # Check if deregistration was successful
+    ${retained_message_count}=    Get Retained Message Count    te/device/${CHILD_SN}/+/+/#
+    Should Be Equal As Numbers    ${retained_message_count}    0    
+
+    # Checking if child device will be recreated after mapper restart
+    Restart Service    tedge-mapper-c8y    
+    Sleep    5s    reason=Allowing startup to complete
+    ${retained_message_count}=    Get Retained Message Count    te/device/${CHILD_SN}/+/+/#
+    Should Be Equal As Numbers    ${retained_message_count}    0
```
albinsuresh, line 44: A better validation would be to try and recreate the device by publishing the registration message again and validate that the mapper emits the mapped 101 message.

tests/RobotFramework/tests/cumulocity/registration/registration_livecycle.robot
```diff
@@ -73,6 +86,13 @@ Register service on a child device via MQTT
     # Check service registration
     Check Service    child_sn=${CHILD_XID}    service_sn=${CHILD_XID}:service:custom-app    service_name=custom-app    service_type=custom-type    service_status=up
 
+    # Deregister service on a child device
+    Execute Command    mosquitto_sub --remove-retained -W 3 -t 'te/device/${CHILD_SN}/+/+/#'    exp_exit_code=27
```
albinsuresh, line 90:
```suggestion
    Execute Command    mosquitto_sub --remove-retained -W 3 -t 'te/device/${CHILD_SN}/service/custom-app/#'    exp_exit_code=27
```
That's the topic filter to target the given service. That topic filter `te/device/${CHILD_SN}/+/+/#` covers the device as well along with all of its services as well. If the intent was to deregister all the services of that given child device, then the topic filter should have been `te/device/${CHILD_SN}/service/+/#`.

albinsuresh (commented):
A few more corrections required.

tests/RobotFramework/tests/cumulocity/registration/registration_livecycle.robot
gligorisaev: it was @reubenmiller choice to go this way, if he agrees....

albinsuresh: This validation makes sure that the mapper doesn't re-publish the deleted entity back to the broker, which it was never doing. But it's still fine as an external backbox test.

But the key thing that we want to make sure is the fact that the mapper doesn't have any stale references to the deleted entity in its in-memory entity store after the restart. The only available way to check that currently is to send another registration message and make sure that the mapper doesn't ignore it as duplicate, but really maps that message.

reubenmiller: Once we have an official deregister api, then we can improve the assertions to check if the entities are registered in the cloud or not after deletion. We didn't want to invest too much into the assertions yet as it would of involved manually deleting the cloud entity via Cumulocity API (the part not yet implemented by thin-edge.io).

albinsuresh: Technically, we don't need to delete the device from the cloud for this specific assertion, but just make sure that the mapper emits the 101 message for the device that is recreated. Obviously, that 101 message would result in a "device already exists" error on C8y, but that can be ignored. That 101 message getting generated is all that we care about.
 
That being said, since the assertion might be different with the official deregister API (although the above assertion would still be valid), we can skip that if this effort is going to be wasted.

reubenmiller: Though I would prefer if the tests are decoupled from the implementation (e.g. it does not have to know anything about the c8y SmartREST messages), so we'll just skip it for now.

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -234,6 +254,12 @@ Entities send to cloud on restart
 
 *** Keywords ***
 
+Get Retained Message Count
+    [Arguments]    ${topic}
+    ${output}=    Execute Command    mosquitto_sub --retained-only -W 3 -t "${topic}" -v    exp_exit_code=27    return_stdout=True
+    ${message_count}=    Get Length    ${output.splitlines()}
+    [Return]    ${message_count}
```
reubenmiller, line 261: Should be using `RETURN` over `[Return]` as the latter is marked as deprecated (see https://github.com/robotframework/robotframework/issues/4078).

```suggestion
    RETURN    ${message_count}
```

reubenmiller: Though the more I look at the `Get Retained Message Count` the more it would suggest doing the assertion within the keyword, as all references do a follow up assertion on the count.

And it could be simplified by using the [Length Should Be](https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Length%20Should%20Be) built-in assertion.

The keyword could be refactored to the following:

```
Should Have Retained Message Count
    [Arguments]    ${topic}    ${exp_count}
    ${output}=    Execute Command    mosquitto_sub --retained-only -W 3 -t "${topic}" -v    exp_exit_code=27    return_stdout=True
    Length Should Be    ${output.splitlines()}    ${exp_count}
```

And calls to the keyword would just require a single call, and you don't have to deal with the returned value:

```
Should Have Retained Message Count    te/device/${CHILD_SN}/+/+/#    0
```

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -30,6 +30,19 @@ Child device registration
     Cumulocity.Set Device    ${DEVICE_SN}
     Cumulocity.Should Be A Child Device Of Device    ${CHILD_SN}
 
+    # Deregister Child device
+    Execute Command    mosquitto_sub --remove-retained -W 3 -t "te/device/${CHILD_SN}/+/+/#"    exp_exit_code=27
+    
+    # Check if deregistration was successful
+    ${retained_message_count}=    Get Retained Message Count    te/device/${CHILD_SN}/+/+/#
```
reubenmiller, line 37: An explicit sleep is used in the deregister service case, so by the same logic it should be used here as well.

For example
```suggestion
    Sleep    1s    reason=Allowing components to process messages
    ${retained_message_count}=    Get Retained Message Count    te/device/${CHILD_SN}/+/+/#
```

gligorisaev:
> Just some small refactoring for consistency and some improvements on the custom keyword to simplify the code.

Done

reubenmiller (approved):
Approved

Issue #2776, opened 2024-03-13
Installing tedge after uninstall is failing


gligorisaev:
**Describe the bug**

Running test which is uninstalling previously installed tedge and than installing again is failing with unmet dependencies error

**To Reproduce**

1. Raspberry PI 3 with installed tedge (used the 192.168.193.16 and 192.168.193.144 from our environment)
2. Uninstalling tedge 
    ```
    curl -sSL https://raw.githubusercontent.com/thin-edge/thin-edge.io/main/uninstall-thin-edge_io.sh | sudo sh -s purge
    ```
4. Installing tedge

    ```
    curl -fsSL https://thin-edge.io/install.sh | sh -s
    ```

Results in packages have unmet dependencies error:

```
pi@raspberrypi:~ $ curl -fsSL https://thin-edge.io/install.sh | sh -s
Thank you for trying thin-edge.io!

Package management dependencies met
Detected package manager: apt
Hit:1 http://deb.debian.org/debian bullseye InRelease
Hit:2 http://deb.debian.org/debian bullseye-updates InRelease                                                                              
Hit:3 http://security.debian.org/debian-security bullseye-security InRelease                                                               
Hit:4 http://archive.raspberrypi.org/debian bullseye InRelease                                                                             
Hit:5 http://download.zerotier.com/debian/bullseye bullseye InRelease                                                                      
Get:6 https://dl.cloudsmith.io/public/thinedge/community/deb/debian bullseye InRelease [5187 B]                                            
Get:7 https://dl.cloudsmith.io/public/thinedge/tedge-main/deb/raspbian bullseye InRelease [4469 B]
Get:8 https://dl.cloudsmith.io/public/thinedge/tedge-release/deb/debian bullseye InRelease [4436 B]
Fetched 14.1 kB in 4s (3892 B/s)   
Reading package lists... Done
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
ca-certificates is already the newest version (20210119).
0 upgraded, 0 newly installed, 0 to remove and 41 not upgraded.

Downloading https://dl.cloudsmith.io/public/thinedge/tedge-release/setup.deb.sh...OK
Using repository arch: aarch64
Executing the  setup script for the 'thinedge/tedge-release' repository ...

   OK: Checking for required executable 'curl' ... 
   OK: Checking for required executable 'apt-get' ... 
   OK: Detecting your OS distribution and release using system methods ...
[...]

PR #2777, opened 2024-03-13
build: fix building of virtual packages

reubenmiller:
## Proposed changes

Due to a build bug instruced by [#2741](https://github.com/thin-edge/thin-edge.io/pull/2741), no virtual packages were being built or published to cloudsmith. This resulted in the install script failing as there was a mismatch in between the last published virtual package and the newer published tedge packages (non-virtual packages).

The following improvements were made:
* Add missing target case `all` which is used when building the virtual packages (this is the part was the root cause)
* Fail the build (e.g. non zero exit code) if no virtual packages are created - this should at least avoid future silent errors when no virtual packages are produced.

## Types of changes

## Paste Link to the issue

* https://github.com/thin-edge/thin-edge.io/pull/2741
* https://github.com/thin-edge/thin-edge.io/issues/2776

## Checklist

### DIFF
--- a/ci/build_scripts/package.sh
+++ b/ci/build_scripts/package.sh
@@ -119,7 +119,7 @@ fi
 
 if [ -z "$PACKAGE_TYPES" ]; then
     case "$TARGET" in
-        *linux*) PACKAGE_TYPES="deb,apk,rpm,tarball" ;;
+        *linux*|all) PACKAGE_TYPES="deb,apk,rpm,tarball" ;;
         *apple*) PACKAGE_TYPES="tarball" ;;
         *) PACKAGE_TYPES="tarball" ;;
     esac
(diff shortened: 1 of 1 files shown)


ci/build_scripts/package.sh
```diff
@@ -194,6 +194,14 @@ build_virtual_package() {
     if [[ "$PACKAGE_TYPES" =~ apk ]]; then
         env GIT_SEMVER="${APK_VERSION:-$GIT_SEMVER}" RELEASE="r0" nfpm "${COMMON_ARGS[@]}" --packager apk
     fi
+
+    # Expect to build at least 1 virtual package to prevent silent errors
+    # when no virtual packages were created
+    VIRTUAL_PACKAGE_COUNT=$(find target/virtual-packages -type f | wc -l | xargs)
```
didier-wenzek, line 200: What's the purpose of the extra `| xargs` with no command using the expended arguments?

reubenmiller: It just strips whitespace

reubenmiller: An alternative way of doing this would be to use `tr`, however the command is much longer and always hard to remember the correct whitespace name:

```
echo "   some string    " | tr -d '[:blank:]'
```

didier-wenzek (approved):
Approved

PR #2781, opened 2024-03-14
feat(ci): build thin-edge.io for additional targets

reubenmiller:
## Proposed changes

Build thin-edge.io for additional targets

* arm-unknown-linux-musleabi
* i686-unknown-linux-musl
* riscv64gc-unknown-linux-gnu
* aarch64-apple-darwin
* x86_64-apple-darwin
* armv5te-unknown-linux-musleabi (though it is only built, it is not published to cloudsmith due to overlap of the armel type in debian)

To make this possible the following changes were made:

* use clang to build the targets (the setup used by brainsmith/ring)
* minor fixes to some test cases which were failing on some of the different targets

Notes:
* developers can still build locally using cargo-zigbuild, however official builds will use clang

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/common/download/src/download.rs
```diff
@@ -559,7 +559,7 @@ mod tests {
     async fn downloader_download_with_content_length_larger_than_usable_disk_space() {
         use nix::sys::statvfs;
         let tmpstats = statvfs::statvfs("/tmp").unwrap();
-        let usable_disk_space = tmpstats.blocks_free() * tmpstats.block_size();
+        let usable_disk_space = (tmpstats.blocks_free() as u64) * (tmpstats.block_size() as u64);
```
reubenmiller, line 562: This test was failing to compile for some 32 bit targets as it seems that the two types returned from the function calls would depend on the target.

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -605,6 +607,8 @@ mod tests {
 
         assert_eq!(actual_c8y_event.event_type, "empty_event".to_string());
         assert_eq!(actual_c8y_event.text, "empty_event".to_string());
+        // delay before checking the time again to ensure an observable time difference
+        sleep(Duration::from_millis(10));
```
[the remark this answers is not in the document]
didier-wenzek: The check can also be less strict.

```
assert!(actual_c8y_event.time <= OffsetDateTime::now_utc());
```

crates/common/download/src/download.rs
didier-wenzek: As the type of the value returned by `tmpstats.blocks_free()` is target dependent it's better to use `u64::from()` instead of `as 64`, even if this is more than unlikely that a target would use a type larger than `u64`.

```suggestion
        let usable_disk_space = u64::from(tmpstats.blocks_free()) * u64::from(tmpstats.block_size());
```

didier-wenzek (commented):
An `mk/README.md` file is missing. Telling that:

- These files are taken from https://github.com/briansmith/ring/tree/main/mk
- That the copyright notice must be kept
- What has been added / changed.

crates/common/download/src/download.rs
reubenmiller: Though the suggested change invoke the following clippy warning:
image

crates/core/c8y_api/src/json_c8y.rs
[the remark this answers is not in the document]
reubenmiller: > The check can also be less strict.
> 
> ```
> assert!(actual_c8y_event.time <= OffsetDateTime::now_utc());
> ```

Good idea, I will go with this approach.

crates/common/download/src/download.rs
didier-wenzek: So let's go using `as u64`.

reubenmiller:
> An `mk/README.md` file is missing. Telling that:
> 
> * These files are taken from https://github.com/briansmith/ring/tree/main/mk
> * That the copyright notice must be kept
> * What has been added / changed.

Resolved by 022103ca6c31c2db6d0a29d335df74aef23315eb. I also improved the diffs somewhat to make it it easier to merge new updates.

didier-wenzek (approved):
Approved

Issue #2760, opened 2024-03-05
tedge cert command to create a certificate signing request


reubenmiller:
**Is your feature request related to a problem? Please describe.**

If the user wants to use a device certificate which is signed by a Certificate Authority (CA), then they have to use other tooling to generate the Certificate Signing Request (CSR) which is used by the CA to generate the device certificate (in most instances this is openssl). The reliance of additional tooling means that users have to have this tooling installed, and also know how to use it which can be daunting for the less experienced users.

**Describe the solution you'd like**

A new `tedge cert` command to generate a Certificate Signing Request file which the user can then use to generate the public device certificate (via their CA).

For example, the command could look something like:

```sh
# Create a new private key (if it does not already exist)
tedge cert create-csr --device-id mydevice001 > mydevice.csr

# Re-use the existing device-id from an already existing certificate
tedge cert create-csr > mydevice.csr
```

Where:
* A private key will be created automatically if it does not already exist
* If a public key already exists, then re-use the device-id from the CommonName field of the public certificate (similar to what the `tedge cert renew` command does)

**Describe alternatives you've considered**

**Additional context**

Ruadhri17:
Resolved by #2783

Ticket introduces new command `tedge cert create-csr` that create a Certificate Signing Request under directory that can be set with `tedge config set device.csr_path` (by default it is `/etc/tedge/device-certs/tedge.csr`). User can also provide output path as command argument with `--output-path`. 

More detailed description, how a CSR is created, can be found [here](https://github.com/thin-edge/thin-edge.io/blob/main/docs/src/operate/security/certificate-signing-request.md). 

System test was created and can be found [here](https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/tedge/certificate_signing_request.robot).

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test for ticket exists in the test suite.
- tests/RobotFramework/tests/tedge/certificate_signing_request.robot
- [x] QA has tested the function and it's functioning according

PR #2783, opened 2024-03-15
feat(cert): Create a Certificate signing request using `tedge cert create-csr`

Ruadhri17:
## Proposed changes
This PR is introducing new `tedge cert` command: `create-csr` along with new config option `device.csr_path`. This allow to create a certificate signing request which later can be used to create device certificate using CA. The Syntax for the command is following:
```sh
# Create a new private key (if it does not already exist)
tedge cert create-csr --device-id mydevice001

# Re-use the existing device-id from an already existing certificate
tedge cert create-csr
```   
Command will reuse existing public certificate to use its `CommonName`. 

Moreover, the command will reuse private key if it already exists. Same goes for `tedge cert create`. 

## Types of changes

## Paste Link to the issue
* #2760 

## Checklist

### DIFF
--- a/crates/common/certificate/src/lib.rs
+++ b/crates/common/certificate/src/lib.rs
@@ -113,15 +113,19 @@ impl KeyCertPair {
     ) -> Result<KeyCertPair, CertificateError> {
         let today = OffsetDateTime::now_utc();
         let not_before = today - Duration::days(1); // Ensure the certificate is valid today
-        KeyCertPair::new_selfsigned_certificate_at(config, id, not_before, key_kind)
+        let params =
+            Self::create_selfsigned_certificate_parameters(config, id, key_kind, not_before)?;
+        Ok(KeyCertPair {
+            certificate: Zeroizing::new(Certificate::from_params(params)?),
+        })
     }
 
-    fn new_selfsigned_certificate_at(
+    fn create_selfsigned_certificate_parameters(
         config: &NewCertificateConfig,
         id: &str,
+        key_kind: &KeyKind,
         not_before: OffsetDateTime,
-        cert_kind: &KeyKind,
-    ) -> Result<KeyCertPair, CertificateError> {
+    ) -> Result<CertificateParams, CertificateError> {
         KeyCertPair::check_identifier(id, config.max_cn_size)?;
         let mut distinguished_name = rcgen::DistinguishedName::new();
         distinguished_name.push(rcgen::DnType::CommonName, id);
@@ -131,21 +135,62 @@ impl KeyCertPair {
             &config.organizational_unit_name,
         );
 
-        let not_after = not_before + Duration::days(config.validity_period_days.into());
-
         let mut params = CertificateParams::default();
+
         params.distinguished_name = distinguished_name;
+
+        let not_after = not_before + Duration::days(config.validity_period_days.into());
         params.not_before = not_before;
         params.not_after = not_after;
-        params.alg = &rcgen::PKCS_ECDSA_P256_SHA256; // ECDSA signing using the P-256 curves and SHA-256 hashing as per RFC 5758
+
         params.is_ca = rcgen::IsCa::Ca(rcgen::BasicConstraints::Unconstrained); // IsCa::SelfSignedOnly is rejected by C8Y
-        if let KeyKind::Reuse { keypair_pem } = cert_kind {
+
+        params.alg = &rcgen::PKCS_ECDSA_P256_SHA256; // ECDSA signing using the P-256 curves and SHA-256 hashing as per RFC 5758
+
+        if let KeyKind::Reuse { keypair_pem } = key_kind {
             params.key_pair = Some(KeyPair::from_pem(keypair_pem)?);
         }
 
+        Ok(params)
+    }
+
+    // Create Certificate without `not_before` and `not_after` fields
+    // as rcgen library will not parse it for certificate signing request
+    pub fn new_certificate_sign_request(
+        config: &NewCertificateConfig,
+        id: &str,
+        key_kind: &KeyKind,
+    ) -> Result<KeyCertPair, CertificateError> {
+        let params = Self::create_certificate_sign_request_parameters(config, id, key_kind)?;
         Ok(KeyCertPair {
             certificate: Zeroizing::new(Certificate::from_params(params)?),
         })
+    }
+
+    fn create_certificate_sign_request_parameters(
+        config: &NewCertificateConfig,
+        id: &str,
+        key_kind: &KeyKind,
+    ) -> Result<CertificateParams, CertificateError> {
+        KeyCertPair::check_identifier(id, config.max_cn_size)?;
+        let mut distinguished_name = rcgen::DistinguishedName::new();
+        distinguished_name.push(rcgen::DnType::CommonName, id);
+        distinguished_name.push(rcgen::DnType::OrganizationName, &config.organization_name);
+        distinguished_name.push(
+            rcgen::DnType::OrganizationalUnitName,
+            &config.organizational_unit_name,
+        );
+
+        let mut params = CertificateParams::default();
+        params.distinguished_name = distinguished_name;
+
+        params.alg = &rcgen::PKCS_ECDSA_P256_SHA256; // ECDSA signing using the P-256 curves and SHA-256 hashing as per RFC 5758
+
+        if let KeyKind::Reuse { keypair_pem } = key_kind {
+            params.key_pair = Some(KeyPair::from_pem(keypair_pem)?);
+        }
+
+        Ok(params)
     }
 
     pub fn certificate_pem_string(&self) -> Result<String, CertificateError> {
@@ -154,6 +199,10 @@ impl KeyCertPair {
 
     pub fn private_key_pem_string(&self) -> Result<Zeroizing<String>, CertificateError> {
         Ok(Zeroizing::new(self.certificate.serialize_private_key_pem()))
+    }
+
+    pub fn certificate_signing_request_string(&self) -> Result<String, CertificateError> {
+        Ok(self.certificate.serialize_request_pem()?)
     }
 
     fn check_identifier(id: &str, max_cn_size: usize) -> Result<(), CertificateError> {
@@ -243,6 +292,7 @@ mod tests {
     use super::*;
     use std::error::Error;
     use time::macros::datetime;
+    use x509_parser::der_parser::asn1_rs::FromDer;
 
     impl KeyCertPair {
         fn new_selfsigned_certificate_with_new_key(
@@ -258,6 +308,24 @@ mod tests {
             .certificate_pem_string()
             .expect("Fail to read the certificate PEM");
         PemCertificate::from_pem_string(&pem_string).expect("Fail to decode the certificate PEM")
+    }
+
+    fn subject_of_csr(keypair: &KeyCertPair) -> String {
+        let csr = keypair
+            .certificate_signing_request_string()
+            .expect("Failed to read the CSR string");
+
+        let pem = x509_parser::pem::Pem::iter_from_buffer(csr.as_bytes())
+            .next()
+            .unwrap()
+            .expect("Reading PEM block failed");
+
+        x509_parser::certification_request::X509CertificationRequest::from_der(&pem.contents)
+            .unwrap()
+            .1
+            .certification_request_info
+            .subject
+            .to_string()
     }
 
     #[test]
@@ -326,9 +394,19 @@ mod tests {
         let id = "some-id";
         let birthdate = datetime!(2021-03-31 16:39:57 +01:00);
 
-        let keypair =
-            KeyCertPair::new_selfsigned_certificate_at(&config, id, birthdate, &KeyKind::New)
-                .expect("Fail to create a certificate");
+        let params = KeyCertPair::create_selfsigned_certificate_parameters(
+            &config,
+            id,
+            &KeyKind::New,
+            birthdate,
+        )
+        .expect("Fail to get a certificate parameters");
+
+        let keypair = KeyCertPair {
+            certificate: Zeroizing::new(
+                Certificate::from_params(params).expect("Fail to create a certificate"),
+            ),
+        };
 
         // Check the not_before date
         let pem = pem_of_keypair(&keypair);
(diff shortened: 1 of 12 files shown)


reubenmiller:
@didier-wenzek Do you think it would make sense if there was also an option to override where to write the csr file to? Reasoning is that the CSR is generally something that has to be triggered by another system (e.g. service which is doing the request to a PKI API and the service needs to know how to access the resulting csr file...and having to do a `tedge config get csr.path` involves having to know too much about tedge)

For example:

```sh
tedge cert csr --output-file ./mycsr
```

Or even writing the csr to stdout?

```sh
tedge cert csr > ./mycsr
```

reubenmiller (changes requested):
Given that this will be used by users, can you please add a system test?

didier-wenzek:
> @didier-wenzek Do you think it would make sense if there was also an option to override where to write the csr file to? Reasoning is that the CSR is generally something that has to be triggered by another system (e.g. service which is doing the request to a PKI API and the service needs to know how to access the resulting csr file...and having to do a `tedge config get csr.path` involves having to know too much about tedge)
>
> For example:
> 
> ```shell
> tedge cert csr --output-file ./mycsr
> ```
> 
> Or even writing the csr to stdout?
> 
> ```shell
> tedge cert csr > ./mycsr
> ```

Why not.

That said, to interact with a CA what we provide now might not be enough to get a device certificate and is definitely not enough to get certificates for local entities. Indeed, we will have to provide not only a CN, but also critical infos (Subject, Subject Alternative Name, Key Usage, Extended Key Usage).

So we will also need to provide a CSR template on the command line as well as template examples for each kind of entity (the main device, the MQTT broker,  tedge_agent, tedge_mapper, child devices ...).

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,78 @@
...
+
+Now you should have a CSR in the `/etc/tedge/device-certs/` directory.
+
+```sh
+ls -l /etc/tedge/device-certs/
+```
+
+```text title="Output"
+total 8
+-r--r--r-- 1 mosquitto mosquitto 664 May 31 09:26 tedge-csr.csr
```
didier-wenzek, line 43: Unless we provide a `tedge` command to show the CSR, it would be good to provide the user a hint on how to check the CSR.

```
$ openssl req -in /etc/tedge/device-certs/tedge-csr.csr -noout -text
```

didier-wenzek (commented):
This is working fine as described. However, can you share the on-disk writing logic between the commands to create a self-signed certificate and a CSR. Indeed, this logic contains several security-critical steps that is would be better keep in one place.

crates/core/tedge/src/cli/certificate/create.rs
```diff
@@ -62,20 +79,42 @@ impl CreateCertCmd {
...
-        // Prevent the certificate to be overwritten
-        set_permission(&cert_file, 0o444)?;
+        let cert = match &self.csr_path {
+            Some(csr_path) => {
+                validate_parent_dir_exists(csr_path).map_err(CertError::CsrPathError)?;
+
+                let cert = KeyCertPair::new_certificate_sign_request(config, &self.id, key_kind)?;
+                let cert_csr = cert.certificate_signing_request_string()?;
+
+                create_file_with_user_group(
```
reubenmiller, line 89: I don't think it makes sense having this file owned by the broker user/group as a CSR has nothing to do with the broker. Instead the owner should be the component which is calling the tedge cli (e.g. don't explicitly change the ownership)...CSR are not secrets, so they can be shared.

crates/core/tedge/src/cli/certificate/create.rs
```diff
@@ -47,6 +58,19 @@ impl CreateCertCmd {
...
+        &self,
+        config: &NewCertificateConfig,
+    ) -> Result<(), CertError> {
+        // Reuse private key if it already exists
+        let key_kind = match std::fs::read_to_string(&self.key_path) {
+            Ok(keypair_pem) => KeyKind::Reuse { keypair_pem },
+            Err(err) if err.kind() == ErrorKind::NotFound => KeyKind::New,
+            Err(err) => return Err(CertError::IoError(err).cert_context(self.cert_path.clone())),
+        };
+        self.create_test_certificate_for(config, &key_kind)
```
didier-wenzek, line 71: Okayish. The method name is a bit confusing. It's not obvious that this will create a CSR because the config provides some path to store it.

crates/core/tedge/src/cli/certificate/create.rs
```diff
@@ -55,20 +79,42 @@ impl CreateCertCmd {
...
+                cert
+            }
+            None => {
+                let cert = KeyCertPair::new_selfsigned_certificate(config, &self.id, key_kind)?;
+
+                // TODO cope with broker user being tedge
+                // Creating files with permission 644 owned by the MQTT broker
+                let mut cert_file =
+                    create_new_file(&self.cert_path, crate::BROKER_USER, crate::BROKER_GROUP)
+                        .map_err(|err| err.cert_context(self.cert_path.clone()))?;
```
didier-wenzek, line 106: This TODO is in the wrong place. It must be related to the private key, which owner matters.

crates/common/certificate/src/lib.rs
```diff
@@ -105,23 +105,16 @@ pub struct KeyCertPair {
...
-        id: &str,
-        key_kind: &KeyKind,
-    ) -> Result<KeyCertPair, CertificateError> {
-        let today = OffsetDateTime::now_utc();
-        let not_before = today - Duration::days(1); // Ensure the certificate is valid today
-        KeyCertPair::new_selfsigned_certificate_at(config, id, not_before, key_kind)
-    }
+pub struct CertParameters {
+    params: CertificateParams,
+}
```
didier-wenzek, line 110: [Minor] I would avoid to add this type.

What you need are simple functions to create and populate `rcgen::CertificateParams` for different usages.

crates/core/tedge/src/cli/certificate/create.rs
Ruadhri17: Yeah, I just resolved the conflict and placed this comment in the same spot as [here](https://github.com/thin-edge/thin-edge.io/blob/01f1cea1d7459e177c89f715c1a54b54d6c7034c/crates/core/tedge/src/cli/certificate/create.rs#L60)

tests/RobotFramework/tests/tedge/certificate_signing_request.robot
```diff
@@ -0,0 +1,57 @@
...
+Resource            ../../resources/common.resource
+Library             ThinEdgeIO
+
+Test Teardown       Get Logs
+
+Test Tags           theme:cli
+
+
+*** Test Cases ***
+Generate CSR using the device-id from an existing certificate and private key
```
didier-wenzek, line 11: This test is already nicely complete. To make it comprehensive, you can also check that `tedge cert create-csr` doesn't alter neither the private key nor the certificate, when existing.

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,114 @@
...
+```sh
+sudo tedge cert create-csr --device-id alpha
+```
+
+or
+
+```sh
+sudo tedge cert create-csr
+```
+if public key already exists and you want to reuse device name.
```
didier-wenzek, line 22:
```suggestion
if a self-signed certificate already exists and you want to reuse device name.
```

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+```text title="Output"
+Certificate Signing Request was successfully created.
+```
+
+:::note
+`tedge cert` requires `sudo` privilege. This command provides no output on success.
```
didier-wenzek, line 30:
```suggestion
`tedge cert` requires `sudo` privilege as creating a private key owned by the MQTT broker. This command provides no output on success.
```

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,114 @@
...
+```
+
+[`sudo tedge cert create-csr`](../../references/cli/tedge-cert.md) creates the certificate signing request in a default location (`/etc/tedge/device-certs/`). To use a custom location, refer to [`tedge config`](../../references/cli/tedge-config.md) or provide absolute path as a command argument:
+
+```sh
+sudo tedge cert create-csr --device-id alpha --output-path /custom/path/mycsr.csr
+```
+
+:::note
+`tedge cert create-csr` will reuse private key if it was already created, e.g by the `tedge cert create` command
```
didier-wenzek, line 52:
```suggestion
`tedge cert create-csr` will reuse the private key if already created, e.g by the `tedge cert create` command
```

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,114 @@
...
+            (none)
+            Requested Extensions:
+    Signature Algorithm: ecdsa-with-SHA256
+    Signature Value:
+        30:46:02:21:00:81:28:11:28:9b:92:cb:b8:d9:d2:1c:3c:8d:
+        00:1f:4e:44:ae:ba:61:7f:ca:17:75:d9:d4:11:04:fa:11:e8:
+        a2:02:21:00:f2:f4:11:77:5c:32:c8:d5:86:66:29:d5:ae:27:
+        3f:64:31:be:f8:4a:89:29:bf:e0:01:b4:f2:63:1f:f0:f0:fb
+```
+## Errors
```
didier-wenzek, line 87: Add a separation line

```suggestion

## Errors
```

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,114 @@
...
+
+Caused by:
+    0: DeviceID Error
+    1: The string '"+"' contains characters which cannot be used in a name [use only A-Z, a-z, 0-9, ' = ( ) , - . ? % * _ ! @]
+```
+
+
+### Certificate Signing Request already exists in the given location
+
+If the CSR file already exists you may see following error:
```
didier-wenzek, line 104: Be more assertive.

```suggestion
If the CSR file already exists the command fails with:
```

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,114 @@
...
+
+```text
+Error: failed to Generate the Certificate Signing Request.
+
+Caused by:
+    A certificate already exists and would be overwritten.
+            Existing file: "/etc/tedge/device-certs/tedge-csr.csr"
+```
+
+Remove the CSR file manually and try executing [`tedge cert create-csr`](../../references/cli/tedge-cert.md) again.
```
didier-wenzek, line 114: Be more assertive

```suggestion
Remove the CSR file manually and re-execute [`tedge cert create-csr`](../../references/cli/tedge-cert.md) again.
```

crates/core/tedge/src/cli/certificate/create_csr.rs
```diff
@@ -0,0 +1,165 @@
...
+use crate::CreateCertCmd;
+use camino::Utf8PathBuf;
+use certificate::NewCertificateConfig;
+
+pub struct CreateCsrCmd {
+    pub id: Option<String>,
+    pub cert_path: Utf8PathBuf,
+    pub key_path: Utf8PathBuf,
+    pub csr_path: Utf8PathBuf,
+    pub output_path: Option<Utf8PathBuf>,
```
didier-wenzek, line 13: We should have only one parameter here as both have the same role.

```suggestion
    pub csr_path: Utf8PathBuf,
```

crates/core/tedge/src/cli/certificate/create_csr.rs
```diff
@@ -0,0 +1,165 @@
...
+        let id = match &self.id {
+            Some(id) => id.clone(),
+            None => cn_of_self_signed_certificate(&self.cert_path)?,
+        };
+
+        // Use output file instead of csr_path from tedge config if provided
+        let csr_path = match &self.output_path {
+            Some(output_path) => output_path,
+            None => &self.csr_path,
+        };
```
didier-wenzek, line 44: This should be done when creating the `CreateCsrCmd` struct, not when using it.

```suggestion
```

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -44,6 +57,18 @@ impl BuildCommand for TEdgeCertCli {
...
+                cmd.into_boxed()
+            }
+
+            TEdgeCertCli::CreateCsr { id, output_path } => {
+                let cmd = CreateCsrCmd {
+                    id,
+                    cert_path: config.device.cert_path.clone(),
+                    key_path: config.device.key_path.clone(),
+                    csr_path: config.device.csr_path.clone(),
+                    output_path,
```
didier-wenzek, line 71: This is where the configured csr path as a default when no output path is provided on the cli.


```suggestion
                    // Use output file instead of csr_path from tedge config if provided
                    csr_path: output_path.unwrap_or_else(|| config.device.csr_path.clone()),
```

didier-wenzek (changes requested):
The `output_path` field has to be removed from `CreateCsrCmd `.

didier-wenzek (approved):
Approved. Thank you.

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,115 @@
...
+```sh
+sudo tedge cert create-csr --device-id alpha
+```
+
+or
+
+```sh
+sudo tedge cert create-csr
+```
+if a self-signed certificate already exists and you want to reuse device name.
```
reubenmiller, line 22: Technically it does not have to be a self-signed certificate, just any existing device certificate.

```suggestion
if a device certificate already exists and you want to reuse device name.
```

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,115 @@
...
+
+
+```text title="Output"
+Certificate Signing Request was successfully created.
+```
+
+:::note
+`tedge cert` requires `sudo` privilege as creating a private key owned by the MQTT broker. This command provides no output on success.
```
reubenmiller, line 30: Though the command only creates a private key if one does not already exist right?

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,115 @@
...
+
+Now you should have a CSR in the `/etc/tedge/device-certs/` directory:
+
+```sh
+ls -l /etc/tedge/device-certs/
+```
+
+```text title="Output"
+total 8
+-r--r--r-- 1 mosquitto mosquitto 664 May 31 09:26 tedge-csr.csr
```
reubenmiller, line 41: The double usage of `csr` seems unnecessary. We should change the default csr file name just to `tedge.csr`

docs/src/operate/security/certificate-signing-request.md
```diff
@@ -0,0 +1,115 @@
...
+
+If the CSR file already exists the command fails with:  
+
+```text
+Error: failed to Generate the Certificate Signing Request.
+
+Caused by:
+    A certificate already exists and would be overwritten.
+            Existing file: "/etc/tedge/device-certs/tedge-csr.csr"
+```
```
reubenmiller, line 113: I'm not sure what the benefit would be to prevent overwriting an already created CSR. It is more likely that the user is going to forget to delete the CSR after getting the public certificate, so this just makes it more difficult to managed, as the user has to read the output, delete the file, then re-run the command.

tests/RobotFramework/tests/tedge/certificate_signing_request.robot
```diff
@@ -0,0 +1,66 @@
+*** Settings ***
+Resource            ../../resources/common.resource
+Library             ThinEdgeIO
+Library             ../../.venv/lib/python3.11/site-packages/robot/libraries/OperatingSystem.py
```
reubenmiller, line 4: This is the wrong import statement.

```suggestion
Library             OperatingSystem
```

reubenmiller: Though looking at the test, are you sure we need this Library?

reubenmiller (changes requested):
Still a few minor points to address (mostly usability things)

docs/src/operate/security/certificate-signing-request.md
reubenmiller: Actually, we can leave it as is, as you need sudo access as well when creating the csr in the default location (under `/etc/tedge/device-certs`).

didier-wenzek: Things are subtle here - but I don't think this needs to be documented.
- root privileges are required for `sudo tedge cert create-cs --device-id xyz` because a new private key is generated
- no privilege is required for `sudo tedge cert create-cs` because the private key is reused.

I think a better approach would have been to `chown` no files inside the command, and to ask the user to run `sudo -u mosquitto tedge cert create  --device-id xyz`  using the appropriate owner for the certificate.

reubenmiller (approved):
Approved. Nice addition

Issue #2787, opened 2024-03-17
Custom MQTT port is not respected when connecting using tedge connect c8y


reubenmiller:
**Describe the bug**

The `tedge connect c8y` does not use the port given via the `c8y.mqtt` configuration setting (feature introduced in https://github.com/thin-edge/thin-edge.io/issues/1931).

This affects cases where Cumulocity IoT is configured to use a non-default port of 8883.

The code currently incorrectly uses a static value of `8883` (see below):

[let mut mqtt_options = MqttOptions::new(bridge_config.remote_clientid.clone(), host[0], 8883);](https://github.com/thin-edge/thin-edge.io/blob/819568187a483d14615de2c1046b516d875e6b30/crates/core/tedge/src/cli/connect/c8y_direct_connection.rs#L30)

However it must be noted that the `c8y-bridge.conf` uses the configured `c8y.mqtt` host/port.

**To Reproduce**

The easiest way to check this is to set an incorrect port which should cause thin-edge.io not to connect (due to the misconfigured port).

1. Set the `c8y.url`

    ```sh
    tedge config c8y.url thin-edge-io.eu-latest.cumulocity.com
    ```

2. Set a custom `c8y.mqtt` value which uses an invalid port `1234`

    ```sh
    tedge config set c8y.mqtt thin-edge-io.eu-latest.cumulocity.com:1234
    ```

3. It is expected that the `tedge connect c8y` should fail (due to the incorrect port)

    ```sh
    tedge connect c8y
    ```

**Expected behavior**

When the user specifies a custom MQTT port using the `c8y.mqtt` configuration.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `Any`
 - Hardware [incl. revision]: `Any`
 - System-Architecture [e.g. result of "uname -a"]: `Any`
 - thin-edge.io version [e.g. 0.1.0]: `1.0.1`

**Additional context**

PR #2789, opened 2024-03-18
fix: Honor c8y.mqtt setting on tedge connect

albinsuresh:
## Proposed changes

- [x] Use the port specified in `c8y.mqtt` setting while establishing direct connection to c8y on `tedge connect c8y`
- [x] Refactor `BridgeConfig` to use `HostPort` instead of separate host and port settings

## Types of changes

## Paste Link to the issue

* https://github.com/thin-edge/thin-edge.io/issues/2787

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/http_proxy.rs
+++ b/crates/core/c8y_api/src/http_proxy.rs
@@ -101,12 +101,19 @@ impl C8yEndPoint {
         // * <tenant_id>.<domain> eg: t12345.c8y.io
         // These URLs may be both equivalent and point to the same tenant.
         // We are going to remove that and only check if the domain is the same.
-        let tenant_uri = &self.c8y_host;
+        let (tenant_host, _port) = self
+            .c8y_host
+            .split_once(':')
+            .unwrap_or((&self.c8y_host, ""));
         let url = Url::parse(url).ok()?;
         let url_host = url.domain()?;
 
         let (_, host) = url_host.split_once('.').unwrap_or((url_host, ""));
-        let (_, c8y_host) = tenant_uri.split_once('.').unwrap();
+        let (_, c8y_host) = tenant_host.split_once('.').unwrap_or((tenant_host, ""));
+
+        // The configured `c8y.http` setting may have a port value specified,
+        // but the incoming URL is less likely to have any port specified.
+        // Hence just matching the host prefix.
         (host == c8y_host).then_some(url)
     }
 }
@@ -237,9 +244,26 @@ mod tests {
     #[test_case("http://test.co.te")]
     #[test_case("http://test.com:123456")]
     #[test_case("http://test.com::12345")]
+    #[test_case("http://localhost")]
+    #[test_case("http://abc.com")]
     fn url_is_my_tenant_incorrect_urls(url: &str) {
         let c8y = C8yEndPoint::new("test.test.com", "test_device");
         assert!(c8y.maybe_tenant_url(url).is_none());
+    }
+
+    #[test]
+    fn url_is_my_tenant_with_hostname_without_commas() {
+        let c8y = C8yEndPoint::new("custom-domain", "test_device");
+        let url = "http://custom-domain/path";
+        assert_eq!(c8y.maybe_tenant_url(url), Some(url.parse().unwrap()));
+    }
+
+    #[ignore = "Until #2804 is fixed"]
+    #[test]
+    fn url_is_my_tenant_check_not_too_broad() {
+        let c8y = C8yEndPoint::new("abc.com", "test_device");
+        dbg!(c8y.maybe_tenant_url("http://xyz.com"));
+        assert!(c8y.maybe_tenant_url("http://xyz.com").is_none());
     }
 
     #[test]
--- a/crates/core/tedge/src/bridge/aws.rs
+++ b/crates/core/tedge/src/bridge/aws.rs
@@ -1,14 +1,14 @@
 use super::BridgeConfig;
 use crate::bridge::config::BridgeLocation;
 use camino::Utf8PathBuf;
-use tedge_config::ConnectUrl;
+use tedge_config::HostPort;
+use tedge_config::MQTT_TLS_PORT;
 
 const MOSQUITTO_BRIDGE_TOPIC: &str = "te/device/main/service/mosquitto-aws-bridge/status/health";
 
 #[derive(Debug, Eq, PartialEq)]
 pub struct BridgeConfigAwsParams {
-    pub connect_url: ConnectUrl,
-    pub mqtt_tls_port: u16,
+    pub mqtt_host: HostPort<MQTT_TLS_PORT>,
     pub config_file: String,
     pub remote_clientid: String,
     pub bridge_root_cert_path: Utf8PathBuf,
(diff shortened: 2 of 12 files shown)


crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -200,8 +202,10 @@ pub fn bridge_config(
         }
         Cloud::Aws => {
             let params = BridgeConfigAwsParams {
-                connect_url: config.aws.url.or_config_not_set()?.clone(),
-                mqtt_tls_port: MQTT_TLS_PORT,
+                mqtt_host: HostPort::<MQTT_TLS_PORT>::try_from(
+                    config.aws.url.or_config_not_set()?.as_str(),
+                )
+                .map_err(TEdgeConfigError::from)?,
```
didier-wenzek, line 208: I'm surprised to see a change for AWS and the same for Azure, but nothing for Cumulocity. Can this be the root cause of all the system tests failing to connect Cumulocity?

crates/core/tedge/src/bridge/aws.rs
```diff
@@ -101,7 +98,7 @@ fn test_bridge_config_from_aws_params() -> anyhow::Result<()> {
         cloud_name: "aws".into(),
         config_file: "aws-bridge.conf".to_string(),
         connection: "edge_to_aws".into(),
-        address: "test.test.io:8883".into(),
+        address: HostPort::<MQTT_TLS_PORT>::try_from("test.test.io")?,
```
reubenmiller, line 101: Since the HostPort is doing some additional logic (e.g. selecting a default port etc.), it might make sense to have this test check the value against the expected string output (e.g. `test.test.io:8883` rather than just comparing the HostPort struct)

tests/RobotFramework/README.md
```diff
@@ -234,8 +234,8 @@ The following dependencies are requires for the instructions:
 3. Add the `CA_KEY` and `CA_PUB` environment variables to your `.env` file used by the tests.
 
     ```sh
-    echo CA_KEY=\""$(cat ~/tedge-ca.key | base64)"\" >> .env
-    echo CA_PUB=\""$(cat ~/tedge-ca.crt | base64)"\" >> .env
+    echo CA_KEY=\""$(cat ~/tedge-ca.key | base64 -w 0)"\" >> .env
+    echo CA_PUB=\""$(cat ~/tedge-ca.crt | base64 -w 0)"\" >> .env
```
reubenmiller, line 238: This won't work on MacOS (freebsd) as it does not use the `-w` option but it is called `-b` instead.

A more cross platform way to do it would be:

```suggestion
    echo CA_KEY=\""$(cat ~/tedge-ca.key | base64 | tr -d '\n')"\" >> .env
    echo CA_PUB=\""$(cat ~/tedge-ca.crt | base64 | tr -d '\n')"\" >> .env
```

crates/core/tedge/src/bridge/aws.rs
rina23q: I got the exactly same thought, also on other cloud mappers, az and c8y.

albinsuresh: Added unit tests at the source level itself for `HostPort.to_string()`, to avoid duplication of the same for all clouds.

crates/core/tedge/src/cli/connect/command.rs
albinsuresh: The problem was with the string representation of the `HostPort` type, which I've fixed now to include the `port` value as well.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -107,7 +107,11 @@ impl C8yEndPoint {
 
         let (_, host) = url_host.split_once('.').unwrap_or((url_host, ""));
         let (_, c8y_host) = tenant_uri.split_once('.').unwrap();
-        (host == c8y_host).then_some(url)
+
+        // The configured `c8y.http` setting may have a port value specified,
+        // but the incoming URL is less likely to have any port specified.
+        // Hence just matching the host prefix.
```
albinsuresh, line 113: The other option was to change the `C8yEndPoint::c8y_host` field into a `HostPort` and just validate the `host` part of it against the incoming host. But, that would be have been a huge change needing the generic port parameter `<P>` to be added to `C8yEndPoint` as well, since `HostPort<P>` is gene]
ric.

didier-wenzek (commented):
The system test "tedge-agent should terminate on SIGINT while downloading file" is failing, because the url `http://localhost/speedlimit/10MB` is erroneously rewritten `http://127.0.0.1:8001/c8y/speedlimit/10MB`, preventing the download to start.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -107,7 +107,11 @@ impl C8yEndPoint {
...
+        (c8y_host.starts_with(host)).then_some(url)
```
Bravo555, line 114: question: Wouldn't this condition also be satisfied by an empty string, which is probably not what we want?

`host` will be empty if the URL doesn't contain any `.` characters, e.g. `http://localhost`, and the function will return `Some(_)`, when it probably should return `None`. If I'm correct, then we should also add such case to the unit tests below.

Bravo555 (commented):
Found one thing I'm unsure of, otherwise LGTM

crates/core/c8y_api/src/http_proxy.rs
didier-wenzek: Good catch!

This is why the system test "tedge-agent should terminate on SIGINT while downloading file" is failing!

The url `http://localhost/speedlimit/10MB` is erroneously taken as an url to the tenant and rewritten `http://127.0.0.1:8001/c8y/speedlimit/10MB`, preventing the download to start.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -101,12 +101,19 @@ impl C8yEndPoint {
...
+        let (tenant_host, _port) = self
+            .c8y_host
+            .split_once(':')
+            .unwrap_or((&self.c8y_host, ""));
         let url = Url::parse(url).ok()?;
         let url_host = url.domain()?;
 
         let (_, host) = url_host.split_once('.').unwrap_or((url_host, ""));
-        let (_, c8y_host) = tenant_uri.split_once('.').unwrap();
+        let (_, c8y_host) = tenant_host.split_once('.').unwrap();
```
didier-wenzek, line 112: Not sure this `unwrap` is safe. What's about a C8Y Edge tenant known only by its host name?

albinsuresh: Yeah, this is something that @reubenmiller and I were discussing in the morning and the conclusion was that this whole subdomain based check is fragile and must be replaced with a proper check that retrieves the real tenant URL behind any custom domain and validates the incoming URL against that. But, since that work is a bit more involved, we decided to create a separate issue for that, and just resolve the port aspect in this PR (keeping the existing fragile mechanism as-is for).

didier-wenzek: So, let's go with this.

didier-wenzek (approved):
Approved

crates/core/c8y_api/src/http_proxy.rs
albinsuresh: Here is ticket for the custom domain issue: https://github.com/thin-edge/thin-edge.io/issues/2804

In the mean time, I've just updated the unwrap logic to at least avoid a panic.

Issue #2749, opened 2024-03-04
sudo.enable config is not respected everywhere


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The `sudo.enable` setting is not respected by all components. This makes it more difficult to run thin-edge.io under and environment that either does not have sudo, or the user does not want to use sudo.

The following source files do not respect the `sudo.enable` setting:

* crates/core/tedge_agent/src/restart_manager/actor.rs
* crates/core/tedge_agent/src/software_manager/actor.rs

**Describe the solution you'd like**

The usage of sudo in all thin-edge.io components should only be used if:
* `sudo.enable` is set to true
* and `sudo` is available to be used

**Describe alternatives you've considered**

**Additional context**

Bravo555:
No functional changes, only a refactor, so there's nothing to verify by QA.

reubenmiller:
I'll test with my original setup which was the reason for creating the ticket

reubenmiller:
I've verified that `sudo.enabled` is now respected by both the software and reboot functionality.

PR #2791, opened 2024-03-19
fix(#2749): Check sudo.enable in software and restart managers

Bravo555:
## Proposed changes

Check `sudo.enable` config setting before using sudo in software and restart managers. If `sudo.enable` is `true` but isn't found in `$PATH`, a warning is emitted.

## Types of changes

## Paste Link to the issue

- #2749 

## Checklist

## Further comments

I'll also probably add a test, once I figure out where to put it.

### DIFF
--- a/crates/common/logged_command/src/logged_command.rs
+++ b/crates/common/logged_command/src/logged_command.rs
@@ -131,30 +131,26 @@ fn send_signal_to_stop_child(child: Option<u32>, signal_type: CmdStatus) {
 /// A command which execution is logged.
 ///
 /// This struct wraps the main command with a nice representation of that command.
-/// This `command_line` field is only required because the
-/// [`Command::get_program()`](https://doc.rust-lang.org/std/process/struct.Command.html#method.get_program)
-/// and
-/// [`Command::get_args()`](https://doc.rust-lang.org/std/process/struct.Command.html#method.get_args)
-/// are nightly-only experimental APIs.
 pub struct LoggedCommand {
-    command_line: String,
     command: Command,
 }
 
 impl std::fmt::Display for LoggedCommand {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.command_line.fmt(f)
+        let command = self.command.as_std();
+
+        command.get_program().to_string_lossy().fmt(f)?;
+        for arg in command.get_args() {
+            // The arguments are displayed as debug, to be properly quoted and distinguished from each other.
+            write!(f, " {:?}", arg.to_string_lossy())?;
+        }
+        Ok(())
     }
 }
 
 impl LoggedCommand {
     pub fn new(program: impl AsRef<OsStr>) -> Result<LoggedCommand, std::io::Error> {
-        let command_line = match program.as_ref().to_str() {
-            None => format!("{:?}", program.as_ref()),
-            Some(cmd) => cmd.to_string(),
-        };
-
-        let mut args = shell_words::split(&command_line)
+        let mut args = shell_words::split(&program.as_ref().to_string_lossy())
             .map_err(|err| std::io::Error::new(std::io::ErrorKind::InvalidData, err))?;
 
         let mut command = match args.len() {
@@ -178,16 +174,10 @@ impl LoggedCommand {
             .stdout(Stdio::piped())
             .stderr(Stdio::piped());
 
-        Ok(LoggedCommand {
-            command_line,
-            command,
-        })
+        Ok(LoggedCommand { command })
     }
 
     pub fn arg(&mut self, arg: impl AsRef<OsStr>) -> &mut LoggedCommand {
-        use std::fmt::Write;
-        // The arguments are displayed as debug, to be properly quoted and distinguished from each other.
-        let _ = write!(self.command_line, " {:?}", arg.as_ref());
         self.command.arg(arg);
         self
     }
@@ -202,7 +192,7 @@ impl LoggedCommand {
     pub async fn execute(mut self, logger: &mut BufWriter<File>) -> Result<Output, std::io::Error> {
         let outcome = self.command.output().await;
 
-        if let Err(err) = LoggedCommand::log_outcome(&self.command_line, &outcome, logger).await {
+        if let Err(err) = LoggedCommand::log_outcome(&self.to_string(), &outcome, logger).await {
             error!("Fail to log the command execution: {}", err);
         }
 
@@ -212,7 +202,7 @@ impl LoggedCommand {
     pub fn spawn(&mut self) -> Result<LoggingChild, std::io::Error> {
         let child = self.command.spawn()?;
         Ok(LoggingChild {
-            command_line: self.command_line.clone(),
+            command_line: self.to_string(),
             inner_child: child,
         })
     }
--- a/crates/common/logged_command/src/logged_command.rs
+++ b/crates/common/logged_command/src/logged_command.rs
@@ -149,6 +149,11 @@ impl std::fmt::Display for LoggedCommand {
 }
 
 impl LoggedCommand {
+    /// Creates a new `LoggedCommand`.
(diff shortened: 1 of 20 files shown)


rina23q (approved):
LGTM

Just I'm not sure how we will test an environment without sudo.

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -244,8 +245,13 @@ impl RestartManagerActor {
...
         // Check if sudo command is available
-        match which(SUDO) {
-            Ok(sudo) => {
+        let sudo = which(SUDO).ok();
+        if self.config.is_sudo_enabled && sudo.is_none() {
+            warn!("`sudo.enable` is `true` but sudo wasn't found in $PATH");
+        }
+
+        match sudo {
+            Some(sudo) if self.config.is_sudo_enabled => {
```
didier-wenzek, line 254: This pattern combining `which` and `config.is_sudo_enabled` to get an `Option<Path>` itself pattern-matched to run a command with or without `sudo` is used in too many places with the risk to be non consistent (as proven by the need for this PR).

A first step to fix that is to stop passing boolean values telling to use `sudo` or not.

Instead the `restart_manager` (as well as the `software_manager` and also `tedge_writer` and `plugin_sm`) should be given a capability object, a struct enabling the capability to run a command with root privileges (*possibly* using `sudo` under the hood).

Something along these lines:

```rust
pub struct Sudo {
      sudo: Option<PathBuf>
}

impl Sudo {
     pub fn new(config: &TEdgeConfig) -> Sudo { ... }
     pub fn new_command<S: AsRef<OsStr>>(&self program: S) -> Command { ... }
}
```

Where to define this capability object? in `tedge_utils`? `tedge_config`? I would go with the latter.

didier-wenzek (changes requested):

crates/common/logged_command/src/logged_command.rs
```diff
@@ -131,30 +131,31 @@ fn send_signal_to_stop_child(child: Option<u32>, signal_type: CmdStatus) {
...
+        Ok(())
     }
 }
 
 impl LoggedCommand {
+    /// Creates a new `LoggedCommand`.
+    ///
+    /// In contrast to [`std::process::Command`], `program` can contain space-separated arguments,
+    /// which will be properly parsed, split, and passed into `.args()` call for the underlying
+    /// command.
```
Bravo555, line 156: note: Since `LoggedCommand` mirrors stdlib's `Command` API a bit, wanted to underline that passing a `program` with spaces to this constructor is fine. Now that we have `impl From<StdCommand>`, it may be tricky to remember the behaviour.

crates/common/logged_command/src/logged_command.rs
```diff
@@ -259,6 +254,20 @@ impl LoggedCommand {
...
+    }
+}
+
+impl From<std::process::Command> for LoggedCommand {
+    fn from(command: std::process::Command) -> Self {
+        Self {
+            command: tokio::process::Command::from(command),
+        }
+    }
+}
```
Bravo555, line 269: note: Had to add this impl because `SudoCommandBuilder` creates `StdCommand`s, and I didn't feel like playing with generics.

crates/core/tedge_agent/src/restart_manager/config.rs
```diff
@@ -1,12 +1,14 @@
...
 use tedge_api::mqtt_topics::EntityTopicId;
+use tedge_config::SudoCommandBuilder;
 
 #[derive(Debug, Clone)]
 pub struct RestartManagerConfig {
     pub device_topic_id: EntityTopicId,
     pub tmp_dir: Utf8PathBuf,
     pub config_dir: Utf8PathBuf,
     pub state_dir: Utf8PathBuf,
+    pub sudo_wrapper: SudoCommandBuilder,
```
didier-wenzek, line 11: [Minor] I have a preference for the name `sudo` as used by the plugin manager.

 ```suggestion
     pub sudo: SudoCommandBuilder,
```

File: crates/core/tedge_agent/src/software_manager/config.rs [hunk 5]
```diff
@@ -11,6 +11,7 @@ pub struct SoftwareManagerConfig {
     pub log_dir: Utf8PathBuf,
     pub default_plugin_type: Option<String>,
     pub config_location: TEdgeConfigLocation,
+    pub sudo_command_builder: SudoCommandBuilder,
```
didier-wenzek, line 14: [minor] Here we have a 3rd name for the same thing.

```suggestion
    pub sudo: SudoCommandBuilder,
```

crates/common/tedge_config/src/sudo.rs
```diff
@@ -0,0 +1,53 @@
...
+use std::process::Command;
+use std::sync::Arc;
+
+use tracing::warn;
+
+use crate::TEdgeConfig;
+
+const SUDO: &str = "sudo";
+
+/// An object used to remember the user's sudo preference.
```
didier-wenzek, line 12: The point is not only to remember user's preference (a bool would do), but also to enable root privileges.

crates/common/tedge_config/src/sudo.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct SudoCommandBuilder {
+    path: Option<Arc<Path>>,
+    enabled: bool,
+}
```
didier-wenzek, line 17: I would remove the `enable` field, as it's redundant with the `None` case.

```suggestion
#[derive(Debug, PartialEq, Eq, Clone)]
pub struct SudoCommandBuilder {
    path: Option<Arc<Path>>,
}
```

crates/common/tedge_config/src/sudo.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+
+        match &self.path {
+            Some(sudo) => {
+                let mut c = Command::new(sudo.as_ref());
+                c.arg(program);
+                c
+            }
+            None => {
+                warn!("`sudo.enable` set to `true`, but sudo not found in $PATH");
```
didier-wenzek, line 48: I would move this warning into `SudoCommandBuilder::enabled` to remove the `enabled` field, which is redundant.

didier-wenzek (approved):
Approved. Nice improvement.

The `struct SudoCommandBuilder` can be simplified, though.

crates/common/tedge_config/src/sudo.rs
Bravo555: I think it would be easier for users if we emit any sudo-related warnings when the command actually gets created and the process is about to be spawned, though. One warning at the beginning is easy to miss, amongst all the other warnings that can appear when the agent is started.

And for that, we need to actually distinguish the case whether or not `sudo` is enabled by the user, but wasn't found in `$PATH`. If we wanted to save space, couldn't we just run `which` every time instead of saving the result? I'd think it's not that expensive and in exchange we'd eliminate the behaviour where currently, if `sudo` is in path when agent is started, but during runtime is removed (a bit of an edge case, I know), the agent would still try to run the old sudo path, which seems like an oversight.

For now I've kept the current behaviour because I wanted to avoid adding too much to the PR, but I think it would be an improvement.

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -33,12 +32,6 @@ use tedge_config::TEdgeConfigError;
 use tracing::error;
 use tracing::info;
 use tracing::warn;
-use which::which;
-
-#[cfg(not(test))]
-const SUDO: &str = "sudo";
-#[cfg(test)]
-const SUDO: &str = "echo";
```
didier-wenzek, line 35: This was actually used by some tests. Notably `tedge-agent` `test_new_software_update_operation`.

didier-wenzek (approved):
I confirm my approval

Issue #2790, opened 2024-03-19
Invoking a sub-command from an operation workflow


didier-wenzek:
An operation workflow uses a combination of scripts and built-in actions to define the steps moving forward an operation request. Having, the __ability to trigger a command defined by another workflow__  would not only add flexibility but also help to be consistent, enforcing a systematic process to handle operations.

For instance, to support [device profile](https://github.com/thin-edge/thin-edge.io-roadmap/issues/39), a profile update workflow will have to combine configuration updates with software and firmware updates.  Having the ability to delegate steps from this main workflow to other workflows will ease the implementation of consistent processes - where all individual updates are following the same rules.

**Proposed Solution**

A command defined by another workflow can be used to handle an operation step.

- A step on which a `operation = <operation-name>` has been attached, is processed using the workflow for the given `<operation>`.
- The workflow definition include all the parameters to be passed to the operation.
- Data extracted from the calling workflow can be injected into the parameters for the inner workflow - using the same conventions as those used for the script handlers. 
- The agent handles the sub command from its initialization steps up to its completion.
   - Notably, the agent assigns a command  identifier to the sub-command instance,
   - populates the init state of the sub-command with the data extracted form the calling workflow
   - monitors the sub-command till its completion
   - awake the calling workflow when a success, a failure or a timeout is reach
   - finally, clean-up MQTT topic used for the sub-command instance
- The calling workflow defines how to handle successes, failures or timeouts using the same conventions as those used for the script handlers.

**Alternatives**

A sub-command could be triggered from a workflow using a script. It would be a bit tricky, though, to monitor the sub-command and to block the script till the sub-command reaches completion.

**Additional context**

See [device profile](https://github.com/thin-edge/thin-edge.io-roadmap/issues/39)

gligorisaev:
Documentation:
- The documentation provides a clear explanation of how to trigger sub-operations from within a workflow and how to handle their execution.
- The explanation of providing operation-specific arguments to the sub-operation instance is thorough and easy to understand.

Test Case:
- The test ensures that the tedge-agent service is restarted properly and verifies the success of the operation through MQTT messages and PID comparison using a sub-command trigger

From my point of view the feature is correctly documented and the step(s) in the Test Case:
[...]

PR #2792, opened 2024-03-19
feat: Invoking a sub-command from an operation workflow

didier-wenzek:
## Proposed changes

- [x] Reference documentation for workflow sub-command execution
- [x] Trigger a sub-command and await its completion
- [x] Pass arguments to a sub-command
- [x] Inject sub-command outcome in the invoking command state
- [x] Give the ability to use a script to compute the sub-command init state.
- [x] Logging a sub-command in the same log file as the invoking command

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/2790

## Checklist

## Follow-up tasks

- Handle sub-command execution timeouts
- Improve workflow log output to ease debugging
- Add syntactic sugar so users don't have to add explicit `await_operation_completion` steps.
- Deprecate built-in restart action
- Consider alternative designs to track the invoking command of a sub-command. See https://github.com/thin-edge/thin-edge.io/pull/2792#discussion_r1553095838
- Improve test coverage https://github.com/thin-edge/thin-edge.io/pull/2792#discussion_r1541216742
- Finalize naming of workflow definition items, see https://github.com/thin-edge/thin-edge.io/pull/2792#discussion_r1577304892
- Fix the behavior of sub-sub-commands. See https://github.com/thin-edge/thin-edge.io/pull/2792#issuecomment-2073641045
- Consider to publish a failed state on state parsing error (see https://github.com/thin-edge/thin-edge.io/pull/2792#discussion_r1577395454)


docs/src/references/agent/operation-workflow.md
```diff
@@ -164,8 +164,12 @@ on_error = "failed"
 
 [init]
   script = "/usr/bin/firmware_handler.sh plan"
-  on_success = "install"
+  on_success = "executing"
   on_error = { status = "failed", reason = "not timely" }
+  
+[executing]
+  action = "proceed"
+  on_success = "install"
```
didier-wenzek, line 172: This is unrelated - but as the executing step was missing the example was not working with Cumulocity.

docs/src/references/agent/operation-workflow.md
```diff
@@ -328,6 +332,56 @@ Note that:
...
+
+```toml
+[trigger_config_update]
+command = "config_update"
+cmd_arg.tedgeUrl = "http://127.0.0.1:8000/tedge/file-transfer/example/config_update/mosquitto-1234",
+cmd_arg.type = "mosquitto"
+on_exec = "waiting_for_config_update"
+
+[waiting_for_config_update]
+action = "await-command-completion"
```
didier-wenzek, line 378: Having two states and a new `await-command-completion` is required from a technical point of view. However, this might seem a bit contrived  from a user point of view. What I propose is to get first a working solution and only then to add some syntactic sugar to let the user define a single state with all the command related handlers and to let the system introduce the waiting state.

docs/src/references/agent/operation-workflow.md
```diff
@@ -379,6 +433,8 @@ For some action, notably a device `restart`, the handlers are limited to one:
 Currently, here are the available actions:
 
 - `restart` triggers a reboot of the device
```
didier-wenzek, line 435: With the ability to trigger sub commands, the `restart` action can be deprecated.

docs/src/references/agent/operation-workflow.md
```diff
@@ -328,6 +332,56 @@ Note that:
...
+  adds an `url` property to the init message of the sub-command.
+- Similarly, `cmd_arg.url = "${.payload.config_update.url}"`
+  adds an `url` property to the init message of the sub-command, but extracting this url from the calling command state.
+- Several such arguments can be added to the sub-command initial state.
+  - All these individual arguments are combined into a JSON object value published on the sub-command MQTT topic.
+  - The sub-command initial state can be provided as a whole with `cmd_arg = ${.payload.path.to.sub-command.init-status}`
+  - The calling command excerpts injected into this initial state are defined using the same path conventions as for script handlers.
+  - If a `cmd_arg.status` value is provided, this value will be ignored and replaced by `"init"`
+    as this is the only valid value for the initial state of a command.
+  - If no value at all is provided for `cmd_arg`, then the default value is `{"status" = "init"}`.
```
reubenmiller, line 357: Providing the arguments properties in the current state works for trivial payloads. I see a few different options to handle no-trival arg conversions when calling a sub-command:

**Option 1:** Any prior conversion must be done in the previous state

```toml
[prepare_software]
command = "/usr/bin/format_software_list.sh"
success = "update_software"

[update_software]
command = "/usr/bin/start_software_update.sh"
```

**Option 2:** Allow users to provide a command which returns the requested sub-command with arguments.

Users can provide a command which will return the data to be used to call the sub-command. This would at least remove the need to have a "prepare_xxx" command before the state which actually does the request (though it would make it a little harder to track).

```toml
[update_software]
command = "/usr/bin/start_software_update.sh"
```

And the command would return some well formed command request structure (similar to the begin/end blocks for the state modifications)

```sh
#!/bin/sh
echo --- tedge command begin ---
echo command=software_update
echo args.value1=$(som_external_func)
echo args.value2.other={"foo":"bar"}
echo --- tedge command end ---
```

**Option 3:** Support more complex template language that can be used to build the command args.

didier-wenzek: > Providing the arguments properties in the current state works for trivial payloads. I see a few different options to handle no-trival arg conversions when calling a sub-command:

It seems that I failed to convey the idea. In `cmd_arg.x = "${.payload.x}"` the right-end-side expression can be any JSON value, not only a scalar but also an array or an object or a combination of these. 
 
> **Option 1:** Any prior conversion must be done in the previous state
> 
> ```toml
> [prepare_software]
> command = "/usr/bin/format_software_list.sh"
> success = "update_software"
> 
> [update_software]
> command = "/usr/bin/start_software_update.sh"
> ```

This is exactly the plan, assigning to `cmd_arg` pre-computed values, stored in the main operation state.

```
[update_software]
command = "update_software"
cmd_args = "${.payload.software_list}"
```

> **Option 2:** Allow users to provide a command which returns the requested sub-command with arguments.
> 
> Users can provide a command which will return the data to be used to call the sub-command. This would at least remove the need to have a "prepare_xxx" command before the state which actually does the request (though it would make it a little harder to track).
> 
> ```toml
> [update_software]
> command = "/usr/bin/start_software_update.sh"
> ```
> 
> And the command would return some well formed command request structure (similar to the begin/end blocks for the state modifications)
> 
> ```shell
> #!/bin/sh
> echo --- tedge command begin ---
> echo command=software_update
> echo args.value1=$(som_external_func)
> echo args.value2.other={"foo":"bar"}
> echo --- tedge command end ---
> ```

Why not. This will avoid to clutter the state of the calling operation with arguments for the invoked operation. However, I will keep the same conventions as for scripts (i.e. tedge begin / end) the name of the command being defined in the main operation workflow.

```
command = "software_update"
cmd_args_from = "/script/extracting/software_update/init_state/from/current_state.sh"
``` 
 
> **Option 3:** Support more complex template language that can be used to build the command args.

We need first more examples. If it happens that most of the `/script/extracting/software_update/init_state/from/current_state.sh` are using `jq`, then we can make `jq` first citizen.

tests/RobotFramework/tests/tedge_agent/workflows/restart_sub_command.toml
```diff
@@ -0,0 +1,20 @@
+operation = "restart_sub_command"
+
+[init]
+action = "proceed"
+on_success = "executing"
+
+# Reboot the device using a restart command
+[executing]
+command = "restart"
+on_exec = "awaiting_reboot"
```
didier-wenzek, line 10: This test demonstrate that the `restart` builtin action can be now deprecated.

docs/src/references/agent/operation-workflow.md
```diff
@@ -370,8 +369,8 @@ For example, a configuration update can be triggered with:
 ```toml
 [trigger_config_update]
 command = "config_update"
-cmd_arg.tedgeUrl = "http://127.0.0.1:8000/tedge/file-transfer/example/config_update/mosquitto-1234"
-cmd_arg.type = "mosquitto"
+input.tedgeUrl = "http://127.0.0.1:8000/tedge/file-transfer/example/config_update/mosquitto-1234"
+input.type = "mosquitto"
```
didier-wenzek: I changed my mind regarding the name of the properties used to push/gather data to/from a sub-command.

- I prefer `input` rather than `cmd_arg` and `output` rather than `payload`.
- An alternative is to use `cmd_input` and `cmd_output`.

My motivation is two folds.
- First, using `payload` was confusing even if technically this is the payload of the invoking command that is updated. Using `output` makes clear that the focus is on extracting data from the sub-command output.
- Second, if we decide to [merge the `command` and `await-command-completion`](https://github.com/thin-edge/thin-edge.io/pull/2792#discussion_r1530410208) in a single step definition, then the two aspects (pushing data to the sub-command init state and extracting data from its final state) will be grouped together. `input` and `output` sounds then better compared to `cmd_arg` and `payload`.

docs/src/references/agent/operation-workflow.md
```diff
@@ -500,24 +499,24 @@ This waiting state specifies, with `on_success`, `on_error` and `on_timeout` han
...
+- The invoking command payload cannot be overridden as a whole. Subfields can be added and updated but not removed.
 - Specific path expressions can be used to extract specific excerpts from the sub-command final state.
-- `${.command}` is a json for the whole message including the `topic` and the `payload`.
-- `${.command.topic}` is the sub-command request topic
-- `${.command.topic.target}` is the sub command target identity
-- `${.command.topic.operation}` is the command operation
-- `${.command.topic.cmd_id}` is the command request unique identifier
-- `${.command.payload}` is the whole sub command json payload
-- `${.command.payload.status}` is the sub command final status
-- `${.command.payload.x.y.z}` is the json value extracted from the payload following the given `x.y.z` path if any.
```
didier-wenzek: I also decided to remove the `.command` prefix, even if the data source is the sub-command final state.

- In no case, data will be extracted from several data sources.
- The command init state, i.e. its input, is populated with data extracted from the invoking command state.
- The command final state, i.e. its output, is injected back into the  invoking command state.

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -281,6 +374,104 @@ fn json_as_string(value: &Value) -> String {
     }
 }
 
+/// A set of values to be injected/extracted into/from a [GenericCommandState]
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
+#[serde(try_from = "Option<Value>")]
+pub enum StateExcerpt {
```
albinsuresh, line 380: Some unit tests for the conversions to/from this struct would be good.

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -281,6 +374,104 @@ fn json_as_string(value: &Value) -> String {
...
+            }
+        }
+    }
+}
+
+impl From<Value> for StateExcerpt {
+    fn from(value: Value) -> Self {
+        match value {
+            Value::Null => StateExcerpt::Literal(Value::Null),
+            Value::Bool(b) => StateExcerpt::Literal(Value::Bool(b)),
```
albinsuresh, line 457: I'm a bit confused with this inconsistency in this conversion impl compared to the impl of `From<Option<Value>>`. Literal values like boolean, number etc were error cases in that impl, but valid here? Something is not obvious.

docs/src/references/agent/operation-workflow.md
```diff
@@ -437,6 +492,37 @@ on_timeout = "timeout_restart"
...
+- `${.topic.cmd_id}` is the sub command request unique identifier
+- `${.payload}` is the whole sub command json payload
+- `${.payload.status}` is the sub command final status
+- `${.payload.x.y.z}` is the json value extracted from the payload following the given `x.y.z` path if any.
+- If given `${.some.unknown.path}`, the subfield is assigned the unchanged expression.
+
+```toml
+[waiting_for_config_update]
+action = "await-command-completion"
+output.some.new.command.field = "${.payload.of.the.sub.command}"
```
albinsuresh, line 519: I found `output` to be a bit confusing. It depends on the perspective I guess. Since it is the `input` to the next state derived from the output/payload of the subcommand, I'd have just used `input` here as well. Or to be explcit, `on_success.input` with the value having references to the `payload`.

docs/src/references/agent/operation-workflow.md
```diff
@@ -328,6 +332,55 @@ Note that:
 - No `on_exit` nor `on_kill` status can be provided, as the script is not monitored.
 - If the script cannot be launched, the workflow will be moved to the final `"failed"` state.
 
+### Sub-Command Execution
+
+An operation workflow can trigger a command defined by another workflow.
+
+```toml
+["<state-name>"]
+command = "<operation-name>"
```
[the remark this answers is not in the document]
didier-wenzek: Indeed a good suggestion

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
```diff
@@ -278,6 +286,80 @@ impl TedgeOperationConverterActor {
...
                 Ok(())
             }
+            OperationAction::Command(sub_operation, input_excerpt, handlers) => {
+                let next_state = &handlers.on_exec.status;
+                info!(
+                    "Triggering {sub_operation} command, and moving {operation} operation to {next_state} state"
+                );
+
+                // Create the sub-command init state with a reference to its invoking command
+                let sub_cmd_id = self.id_generator.new_id();
```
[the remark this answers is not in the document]
didier-wenzek: > Why not reuse the invoking command id itself so that it is easier to connect/correlate the entire command chain?

That's true that we can avoid to generate a truly new command id.

However, we cannot reuse directly the same command id, because this might confuse the c8y mapper, as it uses the prefix of the command ids to determine if it owns a command instance.

So, I will remove the use an id generator and simply prepend  an agent specific prefix to the cmd id generated by the mapper.

> Of course it can be done using the combination of the `__invoking_command__` and `__sub_command__` values in the payloads as well. But, I feel that a common id in the topic itself would have been just easier compared to those values in cluttered payloads across multiple messages, unless there's no other issue with that.

Reusing for the invoked command the command id of the invoking command is not enough. We also need the operation types. The invoked command must be able to resume the invoking command. And the invoking command needs to retrieve the final state of the invoked command.

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: The difference is about the level in the tree structure of the command state payload.

- `input = true`  is rejected because the command payload as a whole must be a JSON object.
- `input.x = true` is accepted and an `x` field is added to the  command init state.

docs/src/references/agent/operation-workflow.md
didier-wenzek: We have to discuss this naming.

I think we have to take the (non yet implemented) view where the input and output of a command are given in a single state (the awaiting state being implied). As in:

```toml
operation = "sub-operation"
input.x = "${.excerpt.of.the.invoking.command.to.use.as.init.state}"
input.y = "${.another.excerpt.for.the.init.state}"
output.a = "${.exceprt.of.the.sub.command.final.state.to.inject.into.the.command.state}"
output.b = "${.another.excerpt.of.the.sub.command.final.state}"
```

albinsuresh: Just wondering if, instead of defining the `output` at the `sub-operation` level, we can define the same as `input` of the next `on_success` state.

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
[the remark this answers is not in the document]
didier-wenzek: Thinking about these extra `__invoking_command__` and `__sub_command__` fields injected into the command and sub-command payloads. There is a way to get rid of them. Instead of storing the invocation relationships in the command payload, these relationships can be persisted in `/etc/tedge/.agent/workflows` along the current command states, without polluting the state payload. I will give a try to this idea.

docs/src/references/agent/operation-workflow.md
didier-wenzek: > Just wondering if, instead of defining the `output` at the `sub-operation` level, we can define the same as `input` of the next `on_success` state.

Yeah, I see what you mean. However, there are perhaps several states with a transition to the `on_success` state and this state payload might have to be computed differently in each case.

didier-wenzek: Noticing that I always refer to *init* and *final* state of the sub-command, I see as an alternative the following:

```toml
operation = "sub-operation"
init.x = "${.excerpt.of.the.invoking.command.to.use.as.init.state}"
init.y = "${.another.excerpt.for.the.init.state}"
final.a = "${.exceprt.of.the.sub.command.final.state.to.inject.into.the.command.state}"
final.b = "${.another.excerpt.of.the.sub.command.final.state}"
```

docs/src/references/agent/operation-workflow.md
didier-wenzek: > Option 2: Allow users to provide a command which returns the requested sub-command with arguments.

> Users can provide a command which will return the data to be used to call the sub-command. This would at least remove the need to have a "prepare_xxx" command before the state which actually does the request (though it would make it a little harder to track).

This has been implemented  by  https://github.com/thin-edge/thin-edge.io/pull/2792/commits/b3dae720e140fc8938766642bd1ee9c36f2965c5

docs/src/references/agent/operation-workflow.md
```diff
@@ -328,6 +332,67 @@ Note that:
...
+  - If both a script and explicit input properties are provided, the script output is applied first and the explicit properties second.
+- There is no way to assign a command identifier for the new sub-command instance.
+  - This identifier is generated by %%te%%, which creates, monitors and clears the MQTT topic for the sub-command.
+
+From an execution perspective, triggering a sub-command is similar to running a background script.
+
+- The calling command workflow is paused while the sub-command is processed.
+- The calling workflow move to an `on_exec` state from where to sub-command execution is watched till success, failure or timeout.
+- A specific `"await-command-completion"` action is to be used for this `on_exec` state where the sub-command completion is awaited.
+- The steps following the sub-command completion are defined on the `on_exec` state with `on_success`, `on_error` and `on_timeout` handlers.
```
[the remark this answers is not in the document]
didier-wenzek: > question: why is sub-command similar to a background script?

In both case, the main workflow is moved into an intermediate state where the sub-command has been launched but not finalized yet". By contrast, with a regular script the main workflow is paused during the execution of the script, *but* stays in the same state. If for some reason the agent stops before the script ends, then this script will be relaunched again on agent restart. This is not be the case for sub-commands and back-ground scripts which are not relaunched if interrupted.

> By _calling command workflow is paused while the sub-command is processed_ I would assume it's more like a foreground script, where we await the result of the execution before continuing.

Yeah, *background* is perhaps not the best wording.

> After reading a bit more, I assume the reason is that workflows have to ultimately be resume-able (restart being a chief example, because after the restart we want to resume in a different state that the one which caused the restart, or else we'd just run the restart command again). Is my understanding correct?

Your understanding is correct. We need to persist the progress from "command to be launched" -> "command launched but not returned yet" -> "command terminated - successfully or not".

docs/src/references/agent/operation-workflow.md
```diff
@@ -328,6 +332,67 @@ Note that:
 - No `on_exit` nor `on_kill` status can be provided, as the script is not monitored.
 - If the script cannot be launched, the workflow will be moved to the final `"failed"` state.
 
+### Sub-Command Execution
+
+An operation workflow can trigger a command defined by another workflow.
+
+```toml
+["<state-name>"]
+command = "<operation-name>"
```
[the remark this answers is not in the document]
didier-wenzek: > question: are we sure we don't want to explicitly state the filename of the sub-operation we're using?

This can be discussed. However, using the operation name has a main benefit. This can be used to trigger an operation with no user-defined workflow (e.g. a restart).

> As the top-level `operation = "..."` key inside the operation `.toml` file can have other name than the filename, it can be a bit annoying to find the sub-operation, because you have to search file contents.

And in practice, I fooled myself with such errors.

> But there's no concern of collisions, because multiple workflow files with the same name inside should cause the `tedge-agent` to quit with an error, is this right?

Currently, the agent doesn't quit on such errors but rejects with a warning such collisions.

tests/RobotFramework/tests/tedge_agent/workflows/super_command.toml
```diff
@@ -0,0 +1,30 @@
...
+
+[dump_payload]
+script = "/etc/tedge/operations/write-file.sh ${.payload.output_file} ${.}"
+on_success = "successful"
+
+[successful]
+action = "cleanup"
+
+[failed]
+action = "cleanup"
```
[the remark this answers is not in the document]
didier-wenzek: > question: So all the outputs from all the states are added to the `payload` field of the initial command request JSON object?

Yes all these fields are accumulated over command execution. This helps to propagate parameters to the following steps.

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -65,6 +72,30 @@ Trigger native-reboot within workflow (on_error) - missing sudoers entry for reb
...
 
+Invoke sub-command from a super-command operation
+    Execute Command    tedge mqtt pub --retain te/device/main///cmd/super_command/test-42 '{"status":"init", "output_file":"/tmp/test-42.json"}'
+    ${cmd_messages}    Should Have MQTT Messages    te/device/main///cmd/super_command/test-42    message_pattern=.*successful.*   maximum=1
+    Execute Command    tedge mqtt pub --retain te/device/main///cmd/super_command/test-42 ''
+    ${actual_log}      Execute Command    cat /tmp/test-42.json
+    ${expected_log}    Get File    ${CURDIR}/super-command-expected.log
+    Should Be Equal    ${actual_log}    ${expected_log}
+    # Remove all dates from the workflow log
+    ${workflow_log}=  Execute Command    sed -e 's/....-..-..T..:..:............Z//g' /var/log/tedge/agent/workflow-super_command-test-42.log
```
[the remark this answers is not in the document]
didier-wenzek: Removed with https://github.com/thin-edge/thin-edge.io/pull/2792/commits/90519e579ed05a23f001833ff18e9019e6ac7f7f

docs/src/references/agent/operation-workflow.md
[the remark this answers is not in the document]
didier-wenzek: Done with https://github.com/thin-edge/thin-edge.io/pull/2792/commits/acf48da654e7a6ac0ad5013ad194d3c06c078ee3

reubenmiller (commented):
Everything seems to be work as expected, though just a few minor changes to address a few usability, debugging aspects:

**Note:** The plans to simplify the calling and awaiting of a sub-operation are not mentioned below as they are out of scope for this ticket.

#### 1. Workflows called from other workflows still produce log files

Should a workflows called by another workflow result in individual log files?

**Advantages**

* Insight into the contents of a high level workflow

**Disadvantages**

* Duplicate log information
* Obfuscates the high level workflow which contains all of the information


#### 2. input_scripts output is not logged

This will make it harder to debug if users generate some erroneous output, e.g. invalid json syntax (trailing commas etc.)

Ideally the command being used to generate the transition and the output of that command should be included in the log output somehow.


#### 3. Naming consistency: renaming of log entry: command -> operation

Rename the following log entry

```log
lite_device_profile operation awaiting_step_1_completion waiting for sub-command completion
```

To

```log
lite_device_profile operation awaiting_step_1_completion waiting for sub-operation completion
```

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -176,39 +239,84 @@ impl GenericCommandState {
...
+    }
+}
+
+/// Build a sub command identifier from its invoking command identifier
+///
+/// Using such a structure command id for sub commands is key
+/// to retrieve the invoking command of a sub-command from its state using [extract_invoking_command_id].
+fn sub_command_id(operation: &str, cmd_id: &str) -> String {
+    format!("sub:{operation}:{cmd_id}")
+}
```
didier-wenzek, line 273: I'm not happy with this solution to track the invoking command of a sub-command.

Beyond the fact that this doesn't work with sub-sub-commands (this can be fixed), this is polluting the command topic. Somehow using a hidden field in the sub-command state was cleaner.

didier-wenzek:
> Workflows called from other workflows still produce log files

fixed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/8508868958aa82ab10e91fa146aee18e16269fc1

> Naming consistency: renaming of log entry: command -> operation

fixed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/4522483be302345bbb900911953454006c60a1fc

> input_scripts output is not logged

fixed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/ca7ef188af2fee315bfe066cda7d01b23ff5dc03

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -23,7 +32,41 @@ pub struct GenericStateUpdate {
...
+            &Channel::Command {
+                operation: OperationType::Custom(sub_operation),
+                cmd_id: sub_cmd_id,
+            },
+        );
+        let status = INIT.to_string();
+        let payload = json!({
+            STATUS: status
+        });
+
```
Bravo555, line 62: question: What's the difference from just doing this?


```suggestion
        let payload = json!({
            STATUS: status,
            INVOKING_COMMAND: invoking_command_topic.name
        });
```

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -176,39 +239,86 @@ impl GenericCommandState {
...
+        matches!(self.status.as_str(), FAILED)
+    }
+}
+
+/// Return the invoking command topic name, if any
+pub fn invoking_command(sub_command: &TopicName) -> Option<TopicName> {
+    match sub_command.split('/').collect::<Vec<&str>>()[..] {
+        [pre, t1, t2, t3, t4, "cmd", _, sub_id] => extract_invoking_command_id(sub_id)
+            .map(|(op, id)| format!("{pre}/{t1}/{t2}/{t3}/{t4}/cmd/{op}/{id}")),
+        _ => None,
```
Bravo555, line 263: todo: Manual topic parsing without checking topic root

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -176,39 +239,86 @@ impl GenericCommandState {
...
 
-    pub fn is_terminal(&self) -> bool {
-        matches!(self.status.as_str(), "successful" | "failed")
+    pub fn is_successful(&self) -> bool {
+        matches!(self.status.as_str(), SUCCESSFUL)
+    }
+
+    pub fn is_failed(&self) -> bool {
+        matches!(self.status.as_str(), FAILED)
+    }
```
Bravo555, line 255: thought: Lots of methods attached to this type, could some potentially be extracted or placed elsewhere?

crates/core/tedge_api/src/workflow/error.rs
```diff
@@ -33,6 +36,16 @@ pub enum ScriptDefinitionError {
...
 
+/// Error related to state excerpt definitions
+#[derive(thiserror::Error, Debug, Eq, PartialEq)]
+pub enum StateExcerptError {
+    #[error("Only records can be used for command input/output, not a {kind} as {value}")]
+    NotAnObject {
+        kind: String,
+        value: serde_json::Value,
+    },
+}
```
Bravo555, line 47: note: IMO instead of putting all error types into `error` module, it would be better to place them together with the thing that can potentially return them, so in this case, `StateExcerpt`

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -176,39 +239,86 @@ impl GenericCommandState {
...
+        _ => None,
+    }
+}
+
+pub fn extract_command_identifier(topic: &str) -> Option<(String, String)> {
+    match topic.split('/').collect::<Vec<&str>>()[..] {
+        [_, _, _, _, _, "cmd", operation, cmd_id] => {
+            Some((operation.to_string(), cmd_id.to_string()))
+        }
+        _ => None,
```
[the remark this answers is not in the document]
didier-wenzek: Yeah, this topic root is a true burden ;-)

crates/core/tedge_api/src/workflow/script.rs
```diff
@@ -265,6 +246,41 @@ impl ExitHandlers {
...
+            )),
+            Some(excerpt) => match serde_json::from_str(&excerpt) {
+                Ok(json) => Ok(json),
+                Err(err) => Err(format!(
+                    "{program} returned non JSON content on stdout: {err}"
+                )),
+            },
+        },
+    }
+}
```
Bravo555, line 282: suggestion: Here, `program` is only passed to append it to the error message. Instead, as the error is opaque, we could easily use `anyhow::Error` and attach `.context()` with program name in calling code.

Also, I think nesting matches makes it a bit hard to read due to Ok and Err arms being in arbitrary order; would you agree that it's more readable when written as series of steps with early return?

```rust
/// Extract the json output of a script outcome
pub fn extract_json_output(
    outcome: std::io::Result<std::process::Output>,
) -> anyhow::Result<Value> {
    let output = outcome.context("Program cannot be launched")?;

    let exit_code = output
        .status
        .code()
        .with_context(|| format!("Killed by SIG{}", output.status.signal().unwrap_or(0) as u8))?;

    anyhow::ensure!(exit_code != 0, "Failed with exit code {exit_code}");

    let stdout = String::from_utf8(output.stdout).context("Stdout not UTF8")?;

    let excerpt = extract_script_output(stdout)
        .context("Program returned no :::tedge::: content on stdout")?;

    let json =
        serde_json::from_str(&excerpt).context("Program returned non JSON content on stdout")?;

    Ok(json)
}
```

And program name can be attached as context by the caller:

```rust
match extract_json_output(&script.command, output).with_context(|| format!("Could not extract output of program {}", &script.command)) {
    Ok(init_state) => init_state,
    Err(reason) => {
        let err_state = state.update(GenericStateUpdate::failed(reason));
        self.publish_command_state(err_state).await?;
        return Ok(());
    }
}
```

Of course, one would first have to add `anyhow` here, and this probably goes into `reason` field, which is a single string, so we'd have to loop through `source()` of `anyhow::Error` appending it to a single string, so maybe it's not workable. But still, I'd argue removing needless matches makes it more readable.

didier-wenzek: > Also, I think nesting matches makes it a bit hard to read due to Ok and Err arms being in arbitrary order; would you agree that it's more readable when written as series of steps with early return?

This is a matter of taste. With early returns is easy to miss one of these returns especially when hidden being a `?`. However, I tried to follow your suggestion.

See: https://github.com/thin-edge/thin-edge.io/pull/2792/commits/10affc17b54d784c91c6caea6b6a2abae97b6484

> suggestion: Here, `program` is only passed to append it to the error message. Instead, as the error is opaque, we could easily use `anyhow::Error` and attach `.context()` with program name in calling code. Of course, one would first have to add `anyhow` here, and this probably goes into `reason` field, which is a single string, so we'd have to loop through `source()` of `anyhow::Error` appending it to a single string, so maybe it's not workable. But still, I'd argue removing needless matches makes it more readable.

Indeed the `program` parameter is better handled with a context. Using `anyhow` being not adapted here (for the reason you cited), I introduced a poor-man context specialized on `String`. I also used this context only on private functions.

See: https://github.com/thin-edge/thin-edge.io/pull/2792/commits/10affc17b54d784c91c6caea6b6a2abae97b6484

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: Removed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/da71c7ee5b70f8f21babacc33a1df4eb02efab97

crates/core/tedge_api/src/workflow/state.rs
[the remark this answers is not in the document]
didier-wenzek: This has been partially fixed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/36eb228bf375cd3d840fb10a56787ba4023afd05 by moving these functions under `MqttSchema`, which defines the topic structures.

Partially fixed, because these functions are still not checking the topic root. This is done on purpose, as the point of these functions is to extract data from a topic, *even* if not actually part of subscription.

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: `GenericCommandState` sounds to be a good place for these methods.

didier-wenzek:
The commit https://github.com/thin-edge/thin-edge.io/pull/2792/commits/da71c7ee5b70f8f21babacc33a1df4eb02efab97 broke "Trigger Device Restart Using A Sub-Command".

The issue is that the invoking command is lost by the restart command. The successful restart state is `{"status":"successful"}` while one expects `{"status":"successful", "__invoking_command__":"<topic-of-the-invoking-command>" }`.

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: Unfortunately, reverted by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/c1450fa94663faa77290690e7bb95645703b5eb4 to fix the regression on "rigger Device Restart Using A Sub-Command"

=> I still have to improve this topic parsing logic

didier-wenzek:
> The commit [da71c7e](https://github.com/thin-edge/thin-edge.io/commit/da71c7ee5b70f8f21babacc33a1df4eb02efab97) broke "Trigger Device Restart Using A Sub-Command".
> 
> The issue is that the invoking command is lost by the restart command. The successful restart state is `{"status":"successful"}` while one expects `{"status":"successful", "__invoking_command__":"<topic-of-the-invoking-command>" }`.

Fixed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/c1450fa94663faa77290690e7bb95645703b5eb4

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: Finally addressed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/8ac28aaec09b9b2d23e92b2b3e80ddaa1f645069

albinsuresh:
Nesting of sub-commands beyond a single level seems to have some issues with the unwinding of the chain. I just tried a 2-level nesting by updating the `inner_command.toml` workflow adding yet another `nested_command.toml` workflow as follows:

**inner_command.toml:**

```toml 
operation = "sub_command"

[init]
action = "proceed"
on_success = "executing"

[executing]
script = "/etc/tedge/operations/echo-as-json.sh x_bis ${.payload.x} y_bis ${.payload.y}  "
on_success = "nested_command"

[nested_command]
operation = "nested_command"
input.x = "some x value"
input.y = "${.payload.y}"
on_exec = "awaiting_completion"

[awaiting_completion]
action = "await-operation-completion"
on_success = "successful"

[successful]
action = "cleanup"

[failed]
action = "cleanup"

```

**nested_command.toml:**

```toml
operation = "nested_command"

[init]
action = "proceed"
on_success = "executing"

[executing]
action = "proceed"
on_success = "successful"

[successful]
action = "cleanup"

[failed]
action = "cleanup"
```

With this setup, even though the `nested_command` workflow got executed to its `successful` state, it was never cleared and the calling `sub_command` was stuck in the `awaiting_completion` state as can be seen in the following message output:

```
te/device/main///cmd/super_command/test-42 {"status":"init", "output_file":"/tmp/test-42.json"}
te/device/main///cmd/super_command/test-42 {"output_file":"/tmp/test-42.json","status":"executing","y":"some y value"}
te/device/main///cmd/super_command/test-42 {"output_file":"/tmp/test-42.json","status":"awaiting_completion","y":"some y value"}
te/device/main///cmd/sub_command/sub:super_command:test-42 {"status":"init","x":"some x value","y":"some y value"}
te/device/main///cmd/sub_command/sub:super_command:test-42 {"status":"executing","x":"some x value","y":"some y value"}
te/device/main///cmd/sub_command/sub:super_command:test-42 {"status":"nested_command","x":"some x value","x_bis":"some x value","y":"some y value","y_bis":"some y value"}
te/device/main///cmd/sub_command/sub:super_command:test-42 {"status":"awaiting_completion","x":"some x value","x_bis":"some x value","y":"some y value","y_bis":"some y value"}
te/device/main///cmd/nested_command/sub:sub_command:sub:super_command:test-42 {"status":"init","x":"some x value","y":"some y value"}
te/device/main///cmd/nested_command/sub:sub_command:sub:super_command:test-42 {"status":"executing","x":"some x value","y":"some y value"}
te/device/main///cmd/nested_command/sub:sub_command:sub:super_command:test-42 {"status":"successful","x":"some x value","y":"some y value"}
```

Instead of a custom workflow like `nested_command`, I tried the inbuilt `restart` workflow at the nested level as well and the result was the same.

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
```diff
@@ -119,37 +124,35 @@ impl TedgeOperationConverterActor {
...
+            log::error!("Unknown command channel: {}", &message.topic.name);
+            return Ok(());
+        };
 
-            _ => {
-                log::error!("Unknown command channel: {}", message.topic.name);
-                return Ok(());
-            }
+        let Ok(state) = GenericCommandState::from_command_message(&message) else {
+            log::error!("Invalid command payload: {}", &message.topic.name);
```
albinsuresh, line 134: Is the log enough or should we have moved the command to the `failed` state?

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
```diff
@@ -119,37 +124,35 @@ impl TedgeOperationConverterActor {
...
-                if message.payload_bytes().is_empty() {
-                    log_file
-                        .log_step("", "The command has been fully processed")
-                        .await;
-                    self.persist_command_board().await?;
-                }
-            }
-            Ok(Some(state)) => {
+        let mut log_file =
+            self.open_command_log(state.invoking_command_topic(), &operation, &cmd_id);
```
albinsuresh, line 140:
```suggestion
            self.open_command_log(state.root_command_topic(), &operation, &cmd_id);
```
Not that it'll work now, but that would be required for nested ops support.

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -90,34 +164,42 @@ impl GenericCommandState {
...
+        GenericCommandState::inject_text_property(&mut self.payload, REASON, &reason);
 
         GenericCommandState {
             status: status.to_owned(),
             ..self
         }
     }
 
+    /// Mark the command as terminated
+    pub fn terminate(self) -> Self {
```
albinsuresh, line 185: Optional:
```suggestion
    /// Clear the command as complete
    pub fn clear(self) -> Self {
```
Since `terminate` has a negative connotation to it, implying a failed path.

tests/RobotFramework/tests/tedge_agent/workflows/inner_command.toml
```diff
@@ -0,0 +1,15 @@
+operation = "sub_command"
```
albinsuresh, line 1: Minor:
```suggestion
operation = "inner_command"
```
Or rename the file to `sub_command.toml ` just for consistency while referring to this workflow.

crates/core/tedge_api/src/workflow/state.rs
albinsuresh: You seem to have missed this. Can be added in a follow-up PR as well, to avoid blocking this one.

albinsuresh (commented):
The code looks fine and the feature also working fine with single level nesting of subcommands. Further nesting seems to have issues, as highlighted in the previous comment. Happy to approve once some minor things are modified/clarified.

reubenmiller:
I found a flaky test (failed 4 times from 25 runs). I'll continue investigating the test to find out why (if it needs a test change, or a code change).

Test name: `Trigger Device Restart Using A Sub-Command`

Below are the results from the test run (where the above test was the reason for each failed run).

```sh
$ invoke flake-finder --iterations 25 --suite "Custom Operation" --clean
Overall: FAILED
Results: 25 iterations, 21 passed, 4 failed
Elapsed time: 1:34:17.920294
Failed iterations: [1, 7, 20, 22]
```

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -34,6 +34,14 @@ Trigger Device Restart
     ${expected_log}     Get File    ${CURDIR}/restart-command-expected.log
     Should Be Equal     ${actual_log}    ${expected_log}
 
+Trigger Device Restart Using A Sub-Command
+    [Documentation]    To detect if the device has been removed, a marker file is created in the /run directory
```
didier-wenzek, line 38:
```suggestion
    [Documentation]    To detect if the device has been rebooted, a marker file is created in the /run directory
```

reubenmiller: Good catch, sorry sloppy comment. Resolved by 3d352e7cf003e1ba4c11dc72ed82360c1ddbcf78

docs/src/references/agent/operation-workflow.md
didier-wenzek: Added as a follow-up task.

reubenmiller (approved):
Approved. Thanks for the sustained effort here, a large change but it lays the foundation for supporting device profile operations in the future.

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
```diff
@@ -396,35 +508,78 @@ impl TedgeOperationConverterActor {
 
         Ok(())
     }
+
+    fn extract_command_identifiers(
```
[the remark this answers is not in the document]
didier-wenzek: Maybe a bit too specialized.

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -249,6 +259,9 @@ impl OperationAction {
             OperationAction::Script(script, handlers) => {
                 OperationAction::Script(script, handlers.with_default(default))
             }
+            OperationAction::AwaitingAgentRestart(handlers) => {
```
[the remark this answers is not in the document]
didier-wenzek: Indeed missing.

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
didier-wenzek: Oops! However the fix is more subtle as the root command topic cannot be inferred directly from the command state.

didier-wenzek: fixed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/d6aef9a517a9112ae229509bfc1f118eb35dc97a

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: renamed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/c560c87d4213dee7ca4d505481ae032c6e35f026

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -141,76 +223,143 @@ impl GenericCommandState {
...
-        matches!(self.status.as_str(), "successful" | "failed")
+    pub fn is_successful(&self) -> bool {
+        matches!(self.status.as_str(), SUCCESSFUL)
+    }
+
+    pub fn is_failed(&self) -> bool {
+        matches!(self.status.as_str(), FAILED)
+    }
+
+    pub fn is_term(&self) -> bool {
```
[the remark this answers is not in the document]
didier-wenzek: renamed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/c560c87d4213dee7ca4d505481ae032c6e35f026

crates/core/tedge_api/src/workflow/mod.rs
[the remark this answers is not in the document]
didier-wenzek: Added by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/36907cbc678a11aa0b289172013ebbb338ca4050

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -109,10 +129,20 @@ impl Display for OperationAction {
...
             OperationAction::BuiltIn => "builtin".to_string(),
-            OperationAction::AwaitingAgentRestart { .. } => "awaiting agent restart".to_string(),
+            OperationAction::AwaitingAgentRestart { .. } => "await agent restart".to_string(),
             OperationAction::Restart { .. } => "trigger device restart".to_string(),
             OperationAction::Script(script, _) => script.to_string(),
             OperationAction::BgScript(script, _) => script.to_string(),
+            OperationAction::Operation(operation, maybe_script, _, _) => match maybe_script {
+                None => format!("execute {operation} sub-operation"),
+                Some(script) => format!(
+                    "execute {operation} sub-operation, initialized with: {}",
```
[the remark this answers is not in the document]
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/7b95167c80253525a2d8bf79dacbf1cfd0fddbfa

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -100,6 +101,25 @@ pub enum OperationAction {
     /// ```
     BgScript(ShellScript, BgExitHandlers),
 
+    /// Trigger an operation and move to the next state from where the outcome of the operation will be awaited
```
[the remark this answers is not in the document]
didier-wenzek: Added by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/78c9a45108488a7a5b26873cec09d6ab3f937b14

tests/RobotFramework/tests/tedge_agent/workflows/extract_updates.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+set -e
+
```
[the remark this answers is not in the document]
didier-wenzek: Added by https://github.com/thin-edge/thin-edge.io/pull/2792/commits/e3bbeaca963ac9c7156e9fa7d23480d3ca75eb21

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -61,18 +61,19 @@ pub enum OperationAction {
     /// ```
     BuiltIn,
 
-    /// The command is delegated to a participant identified by its name
+    /// Await a restart
```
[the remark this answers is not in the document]
didier-wenzek: Fixed

albinsuresh (approved):
Approved.

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -34,6 +34,14 @@ Trigger Device Restart
     ${expected_log}     Get File    ${CURDIR}/restart-command-expected.log
     Should Be Equal     ${actual_log}    ${expected_log}
 
+Trigger Device Restart Using A Sub-Command
+    [Documentation]    To detect if the device has been rebooted, a marker file is created in the /run directory
+        ...            which should be deleted when the device is restarted
+    Execute Command    sudo touch /run/boot1
```
albinsuresh, line 40: Seeing this, I'm wondering if we can update our `await-agent-restart` logic to do something similar (or introduce an `await-device-restart`)

Issue #2794, opened 2024-03-25
Follow up tasks for built-in bridge


jarhodes314:
In #2716, support was added to the mapper for a built-in Cumulocity bridge, to replace the mosquitto-based bridge.

Here is a list of follow up tasks that need completing before the feature is fully complete and ready to be enabled for users:

- [ ] Don't drop messages published when the mapper is down (see https://github.com/thin-edge/thin-edge.io/pull/2716#discussion_r1526243685 for a test case)
- [x] Support non-c8y clouds
- [x] Ensure custom templates work
This appears to just work. I added a custom template ID to tedge config, restarted the mapper, and created the template in the Cumulocity UI and the mapper was forwarding messages received to `c8y/s/uc/my-template-id`. There were no issues with subscribing to the template topic before the template existed, as I hoped would be the case. Deleting the template in the UI didn't stop Cumulocity from sending the template messages, although this is clearly a Cumulocity behaviour, and not within our control, even if it does seem a bit weird.
- [x] Make mapper bridge health message look like other messages
- [x] Fix connectivity check in `tedge connect c8y`
- [x] Ensure status is reported as "up" after restarting local broker
- [x] Fix bridge health topic filtering (https://github.com/thin-edge/thin-edge.io/pull/2810#discussion_r1549914042)
- [x] Use the custom topic prefix in the `tedge-mapper-c8y` service name
- [x] Custom topic prefix validation
- [x] Various TODOs left in code in the original PR (to be expanded as notable ones are encountered)
- [ ] Documentation
- [x] Make the associated configurations visible, both in `tedge config list` and `tedge config list --doc`
- [ ] Remove the feature flag and enable the feature by default

jarhodes314:
I've been looking at making the logic work for `tedge cert create` and have hit a bit of an issue. In the case of the built-in bridge, we want the device certificate to be owned by `tedge`, but in the mosquitto bridge, the device certificate should instead be owned by `mosquitto`. At the moment, the configuration for the bridge is `c8y.bridge.built_in`, implying that this setting is per-cloud, but since only one of mosquitto or the mapper should be able to read the cert/key at a time, it isn't actually possible to have this setting work differently for different clouds.

@didier-wenzek @reubenmiller do you think it's reasonable to change the setting to `bridge.built_in` (or some other name), or do you have other suggestions as to how to resolve this?

didier-wenzek:
[...]

PR #2810, opened 2024-04-03
Initial follow up tasks for built-in bridge

jarhodes314:
## Proposed changes
These are some of the changes discussed in #2794, specifically:
- Ensure custom templates work (this was just a check I performed, no code changes were made)
- Modifying the bridge healthcheck message to use thin-edge json, rather than 0/1 like mosquitto
- Fix connectivity check in `tedge connect c8y` so new devices connect successfully

## Types of changes

## Paste Link to the issue
- #2794

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/utils.rs
+++ b/crates/core/c8y_api/src/utils.rs
@@ -4,16 +4,6 @@ pub mod bridge {
     use tedge_api::MQTT_BRIDGE_DOWN_PAYLOAD;
     use tedge_api::MQTT_BRIDGE_UP_PAYLOAD;
 
-    pub fn is_c8y_bridge_up(message: &Message, service: &str) -> bool {
-        let c8y_bridge_health_topic = main_device_health_topic(service);
-        match message.payload_str() {
-            Ok(payload) => {
-                message.topic.name == c8y_bridge_health_topic && payload == MQTT_BRIDGE_UP_PAYLOAD
-            }
-            Err(_err) => false,
-        }
-    }
-
     pub fn is_c8y_bridge_established(message: &Message, service: &str) -> bool {
         let c8y_bridge_health_topic = main_device_health_topic(service);
         match message.payload_str() {
@@ -46,22 +36,9 @@ mod tests {
     use test_case::test_case;
 
     use crate::utils::bridge::is_c8y_bridge_established;
-    use crate::utils::bridge::is_c8y_bridge_up;
 
     const C8Y_BRIDGE_HEALTH_TOPIC: &str =
         "te/device/main/service/tedge-mapper-bridge-c8y/status/health";
-
-    #[test_case(C8Y_BRIDGE_HEALTH_TOPIC, "1", true)]
-    #[test_case(C8Y_BRIDGE_HEALTH_TOPIC, "0", false)]
-    #[test_case("tedge/not/health/topic", "1", false)]
-    #[test_case("tedge/not/health/topic", "0", false)]
-    fn test_bridge_is_up(topic: &str, payload: &str, expected: bool) {
-        let topic = Topic::new(topic).unwrap();
-        let message = Message::new(&topic, payload);
-
-        let actual = is_c8y_bridge_up(&message, "tedge-mapper-bridge-c8y");
-        assert_eq!(actual, expected);
-    }
 
     #[test_case(C8Y_BRIDGE_HEALTH_TOPIC, "1", true)]
     #[test_case(C8Y_BRIDGE_HEALTH_TOPIC, "0", true)]
--- a/crates/core/c8y_api/src/utils.rs
+++ b/crates/core/c8y_api/src/utils.rs
@@ -9,10 +9,22 @@ pub mod bridge {
         match message.payload_str() {
             Ok(payload) => {
                 message.topic.name == c8y_bridge_health_topic
-                    && (payload == MQTT_BRIDGE_UP_PAYLOAD || payload == MQTT_BRIDGE_DOWN_PAYLOAD)
+                    && (payload == MQTT_BRIDGE_UP_PAYLOAD
+                        || payload == MQTT_BRIDGE_DOWN_PAYLOAD
+                        || is_valid_status_payload(payload))
             }
             Err(_err) => false,
         }
+    }
+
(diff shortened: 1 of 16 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -341,6 +341,10 @@ impl CumulocityConverter {
                 let ancestors_external_ids =
                     self.entity_store.ancestors_external_ids(entity_topic_id)?;
 
+                if entity_topic_id.is_bridge_health_topic() {
```
jarhodes314, line 344: I'm very confused as to why I needed to make this change twice to actually get the bridge to not appear in c8y. Both `converter.rs` and `service_monitor.rs` appear to publish status messages for services, with `converter.rs` doing it relatively cleanly, while `service_monitor.rs` is publishing both `down` & `up` messages when a service starts.

crates/core/c8y_api/src/utils.rs
```diff
@@ -4,26 +4,28 @@ pub mod bridge {
...
+
+    #[derive(serde::Deserialize)]
+    struct HealthStatus<'a> {
+        status: &'a str,
+    }
+
+    fn is_valid_status_payload(payload: &str) -> bool {
+        serde_json::from_str::<HealthStatus>(payload)
+            .map_or(false, |h| h.status == "up" || h.status == "down")
+    }
```
didier-wenzek, line 28: [minor] I would move into `tedge_api` this logic to accept `0` | `1` | `status = up` | `status = down`.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -77,13 +78,24 @@ impl Actor for C8yMapperActor {
...
         // Wait till the c8y bridge is established
+        info!("Waiting for Cumulocity bridge to be established before mapper starts");
         while let Some(message) = self.bridge_status_messages.recv().await {
             if is_c8y_bridge_established(&message, &self.c8y_bridge_service_name) {
                 break;
             }
         }
 
+        log_if_bridge_not_detected.abort();
+        info!("Cumulocity bridge is established, mapper is starting");
```
didier-wenzek, line 97: Having a background error printing task, that is aborted on the happy path, is more than unusual!

But that's not my main point.
I wonder, why introducing a timeout that was not present before?

Actually:
- with the mosquitto bridge, the c8y mapper *must* wait for the bridge to be established (up or down it doesn't matter) otherwise all the init messages sent on the `c8y/#` topics will be lost (as there is no persisted subscriber).
- with the built-in bridge, the c8y mapper *can* directly send the init messages, since there is one subscriber to the `c8y/#` topics (the mapper itself).

To make things simple, I would only take care of the case where awaiting the bridge is required.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -390,6 +390,12 @@ impl EntityTopicId {
     pub fn as_str(&self) -> &str {
         self.0.as_str()
     }
+
+    // FIXME: can also match "device/bridge//" or "/device/main/service/my_custom_bridge"
+    // should match ONLY the single mapper bridge
+    pub fn is_bridge_health_topic(&self) -> bool {
+        self.as_str().contains("bridge")
+    }
```
didier-wenzek, line 398: This PR might be the opportunity to resolve this FIXME.

didier-wenzek (commented):
I'm okay with these changes except for the added timeout causing the mapper to start even if the bridge is not established.

crates/core/c8y_api/src/utils.rs
jarhodes314: If I'm removing the check for bridge health, then this more complicated logic isn't required, so unless we need it for some other reason you know of, I'll undo this change

crates/extensions/c8y_mapper_ext/src/actor.rs
jarhodes314: The timeout doesn't change the behaviour it exists purely so the mapper can log that it's totally not functional because the bridge doesn't exist (and I couldn't think of a better way to notify than just "if nothing has happened after a certain amount of time, something is broken"). I agree that this isn't necessary though, and I should just skip it if `c8y.bridge.built_in` is `true`.

crates/core/c8y_api/src/utils.rs
didier-wenzek: Checking that the bridge has been initialized before sending registration messages to c8y is required for a very new thin-edge device. If the c8y mapper starts before the bridge and even before the c8y.url is provided (as this is the case for a generic image of thin-edge), then the mapper must wait for the bridge to be initialized - otherwise all the registration messages are lost.

didier-wenzek: So this logic is required - but can be simplified by simply checking that at least one message has been received on this topic - whatever the payload.

crates/extensions/c8y_mapper_ext/src/actor.rs
didier-wenzek: > I didn't consider it too deeply as I was mostly concerned with "this bit of code that's critically breaking the mapper didn't tell me it existed" and that taking a while to debug. 

That's true that some logging is required as the mapper is doing nothing this the bride is connected.

>  I should just skip it if `c8y.bridge.built_in` is `true`.

Yes, if this is actually simpler.

crates/core/c8y_api/src/utils.rs
didier-wenzek: I just saw https://github.com/thin-edge/thin-edge.io/pull/2810/commits/4ef0fcd5aa235612f682fb725cd8c031073ba4a6 and I'm okay with it.

crates/extensions/c8y_mapper_ext/src/actor.rs
didier-wenzek: I just saw https://github.com/thin-edge/thin-edge.io/pull/2810/commits/4ef0fcd5aa235612f682fb725cd8c031073ba4a6 and I'm okay with it.

didier-wenzek (approved):
Approved

Issue #2763, opened 2024-03-06
add filename when uploading files to Cumulocity IoT


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When files are uploaded to Cumulocity from thin-edge.io, the files do not included meta information like the filename. This results in a suboptimal user experience when the user downloads the file from the UI, as the downloaded file will automatically get the filename set to the event id (as the filename was not provided when the file was uploaded).

The file upload is used in the following thin-edge.io features:
* Log file request
* Configuration snapshow request

Here is a screenshot showing that when the user clicks on the "download" button on the events page, that the downloaded filename is set to the event id (which is not helpful to the user)

<img width="729" alt="image" src="https://github.com/thin-edge/thin-edge.io/assets/3029781/c41ea4b6-33dc-48f1-a3af-99b7fa11002a">

In comparison, when the filename is set, then it is used when downloading from the UI (the file was uploaded using go-c8y-cli):

<img width="710" alt="image" src="https://github.com/thin-edge/thin-edge.io/assets/3029781/9c9e8098-69f5-4b12-98e5-b7827710c713">

**Describe the solution you'd like**

Set the `filename` when uploading a file to the [Cumulocity Event Attachment API](https://cumulocity.com/api/core/10.18.0/#operation/postEventBinaryResource). 

This can be achieved by using a `multipart/form-data` request, where the `Content-Disposition` header includes meta information about the filename that is being uploaded.

For example, the `data.txt` file is uploaded (attached) to the Cumulocity IoT event id `36200` using the following go-c8y-cli command (which shows the multipart/form-data request as a HTTP dump):

```sh
$ c8y events createBinary --id 36200 --file $HOME/tmp/data.txt --dry --dryFormat dump
POST /event/events/36200/binaries HTTP/1.1
Host: example.c8y.io
Accept: application/json
Authorization: Bearer <token>
Content-Type: multipart/form-data; boundary=1159266905bce6f9491acf392d3df53b87c05b35dd797fdf289ecf62043d

--1159266905bce6f9491acf392d3df53b87c05b35dd797fdf289ecf62043d
Content-Disposition: form-data; name="file"; filename="data.txt"
Content-Type: application/octet-stream

foo
bar

--1159266905bce6f9491acf392d3df53b87c05b35dd797fdf289ecf62043d--
```

**Filename convention**

The filename should also include some information about the device it is coming from (to make it easier for users to  compare files from multiple devices). The following filename format is proposed:

[...]

PR #2811, opened 2024-04-03
Uploader supports multipart/form-data and attaches filename

rina23q:
## Proposed changes

With this change, now you can download log files with file name such as `rina0005_software-management-c8y-mapper-5369.txt`.

Technical changes:
- Use async client to support multipart/form-data
- Expand Uploader to support both PUT and POST method
- Add MIME auto-detection based on file extension

### Takeaways

I added a half working content-type auto detection mechanism to this PR. Stories behind:
* It is nice to have auto-detection of `Content-Type` header. 
* The `reqwest` feature `multipart` adds a dependency on the `mime_guess` crate. So, using `mime_guess` crate shouldn't give any new dependency to thin-edge.
* However, their function gives a guess based on a file extension without file access. Read [this](https://docs.rs/mime_guess/latest/mime_guess/struct.MimeGuess.html#method.from_path).

For example, in our real scenario, uploading `tedge.toml` to the configuration snapshot,
1. `PUT` to the file transfer service with content-type `text/x-toml` since the target file has `.toml` extension.
2. `GET` the file from the file transfer service, but at this point, we lose the file extension since it is stored as `tedge-configuration-plugin-c8y-mapper-5375`.
3. `POST` the file to the c8y-proxy endpoint. However, content-type is now set to `applications/octet-stream` since the file is missing the extension.

This is the log. 
Screenshot from 2024-04-03 19-45-58

So probably as the next step, we can choose either 1) reading the file content and guess the mime or 2) giving a file extension when we store the file in the file transfer service.

## Types of changes

## Paste Link to the issue
#2763 

## Checklist

### DIFF
--- a/crates/common/upload/src/upload.rs
+++ b/crates/common/upload/src/upload.rs
@@ -5,12 +5,13 @@ use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use log::info;
 use log::warn;
+use mime::Mime;
+use mime_guess::MimeGuess;
 use reqwest::header::CONTENT_LENGTH;
 use reqwest::header::CONTENT_TYPE;
+use reqwest::multipart;
 use reqwest::Body;
 use reqwest::Identity;
-use std::fmt::Display;
-use std::fmt::Formatter;
 use std::time::Duration;
 use tokio::fs::File;
 use tokio_util::codec::BytesCodec;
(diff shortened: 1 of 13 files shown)


crates/common/upload/src/upload.rs
```diff
@@ -27,26 +29,69 @@ fn default_backoff() -> ExponentialBackoff {
...
+    Auto,
+    Custom(String),
+    FormData(FormData),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub enum Mime {
     TextPlain,
     ApplicationOctetStream,
 }
```
didier-wenzek, line 46: Why not using [`mine::Mine`](https://docs.rs/mime/latest/mime/#mime)?

didier-wenzek (commented):
Yes, it would be better to add a system test ;-)

crates/common/upload/src/upload.rs
rina23q: I was hesitated to add `mime` crate to `Cargo.toml`. However, as you suggested, we can replace it and thin-edge itself has already dependency on the `mime` crate because of `reqwest`.

rina23q: Done with 7b511cd60680348c0fda66e26b5979cf05fb50df

crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs
```diff
@@ -207,21 +209,33 @@ impl CumulocityConverter {
...
+            &file_path,
         )
-        .with_content_type(ContentType::TextPlain);
+        .post()
+        .with_content_type(ContentType::FormData(
+            FormData::new(format!(
+                "{xid}_{filename}",
+                filename = file_path.file_name().unwrap_or("filename")
+            ))
+            .set_mime("text/plain".parse().unwrap()), // safe
```
didier-wenzek, line 237:
```suggestion
            .set_mime(mime::TEXT_PLAIN),
```

didier-wenzek (approved):
Approved

crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs
rina23q: Done in [62e808f](https://github.com/thin-edge/thin-edge.io/pull/2811/commits/62e808ff60b33c3c2f136d25c9dd8ff50d9b02be)

didier-wenzek:
@rina23q you only have to rebase your PR to fix the system test.

rina23q:
@didier-wenzek 

> @rina23q you only have to rebase your PR to fix the system test.

I had to add more system tests other than log_upload for main device (done in [082e884](https://github.com/thin-edge/thin-edge.io/pull/2811/commits/082e884b80482a05d1a5d5569424140a2523f819))

In [c83a3b6](https://github.com/thin-edge/thin-edge.io/pull/2811/commits/c83a3b67863aa1f961fd88151ec6afa793cae5e5), I've updated 
- [log_upload] match regex for main device
- [log_upload] newly adding content and filename check for child device  
- [config_snapshot] newly adding file name check for both main and child device

I'll fixup all commits after @reubenmiller checks the system tests.

reubenmiller (approved):
Approved.

A good step in the right direction. We'll need to follow up this PR with another ticket where we set the extension (if we have that info), and to also add some custom meta information to it, for example log files would be helpful to have the dateFrom/dateTo filter in the filename.

rina23q:
> Approved.
> 
> A good step in the right direction. We'll need to follow up this PR with another ticket where we set the extension (if we have that info), and to also add some custom meta information to it, for example log files would be helpful to have the dateFrom/dateTo filter in the filename.

Ticket created #2823

PR #2818, opened 2024-04-09
Deprecate tedge topic parsing method

didier-wenzek:
## Proposed changes

Deprecate `get_child_id_from_child_topic`, a method parsing the now deprecated `tedge` command topic.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/c8y_firmware_manager/src/message.rs
```diff
@@ -98,11 +97,17 @@ impl TryFrom<&MqttMessage> for FirmwareOperationResponse {
...
+            // This plugin is still listening to the deprecated `tedge` command topic
+            ["tedge", child_id, "commands", "res", "firmware_update"] if !child_id.is_empty() => {
+                child_id.to_string()
+            }
+            _ => {
+                return Err(FirmwareManagementError::InvalidTopicFromChildOperation {
+                    topic: topic.into(),
+                })
+            }
+        };
```
Bravo555, line 110: note: I think this not only removes the deprecated function, but also fixes the wrong behaviour

Previously `get_child_id_from_child_topic` just extracted child id, but the topic itself wasn't checked, so it could be anything.

Bravo555 (approved):

crates/extensions/c8y_firmware_manager/src/message.rs
didier-wenzek: Indeed, this also fixes the behavior and I failed to missed that trying at first to use one of the `tedge_api::MqttSchema` method!

PR #2821, opened 2024-04-12
Support different clouds in built-in bridge

jarhodes314:
## Proposed changes
This adds support for AWS and Azure to the built-in bridge, as well as some clean session configuration to ensure we don't lose messages.

Previously the rules were Cumulocity rules, where all we needed to do was remove and add local prefixes, so a large section of the changes are around supporting more complex mapping logic, so a `BridgeRule` struct has been added to support the same configurations as mosquitto.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,7 +158,7 @@ tempfile = "3.9"
 test-case = "3.2"
 thiserror = "1.0"
 time = "0.3"
-tokio = { version = "1.23", default-features = false }
+tokio = { version = "1.37", default-features = false }
 tokio-rustls = "0.24.1"
 tokio-tungstenite = { version = "0.20.0" }
 tokio-util = { version = "0.7", features = ["codec"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,6 +94,7 @@ mockall = "0.11"
 mockito = "1.1.0"
 mqtt_channel = { path = "crates/common/mqtt_channel" }
 mqtt_tests = { path = "crates/tests/mqtt_tests" }
+mutants = "0.0.3"
 nanoid = "0.4"
 nix = "0.26"
 notify = { version = "6.1.1", default-features = false }
--- a/crates/core/tedge_mapper/Cargo.toml
+++ b/crates/core/tedge_mapper/Cargo.toml
@@ -25,7 +25,6 @@ collectd_ext = { workspace = true }
 flockfile = { workspace = true }
 mqtt_channel = { workspace = true }
 reqwest = { workspace = true }
-rumqttc = { workspace = true }
 tedge_actors = { workspace = true }
 tedge_api = { workspace = true }
 tedge_config = { workspace = true }
--- a/crates/core/tedge_mapper/Cargo.toml
+++ b/crates/core/tedge_mapper/Cargo.toml
@@ -25,6 +25,7 @@ collectd_ext = { workspace = true }
 flockfile = { workspace = true }
 mqtt_channel = { workspace = true }
 reqwest = { workspace = true }
+rumqttc = { workspace = true }
 tedge_actors = { workspace = true }
 tedge_api = { workspace = true }
 tedge_config = { workspace = true }
--- a/crates/core/tedge_mapper/src/aws/mapper.rs
+++ b/crates/core/tedge_mapper/src/aws/mapper.rs
@@ -10,6 +10,7 @@ use tedge_actors::MessageSource;
 use tedge_actors::NoConfig;
 use tedge_api::mqtt_topics::MqttSchema;
 use tedge_config::TEdgeConfig;
+use tedge_mqtt_bridge::use_key_and_cert;
 use tedge_mqtt_bridge::BridgeConfig;
 use tedge_mqtt_bridge::MqttBridgeActorBuilder;
 use tracing::warn;
@@ -34,11 +35,23 @@ impl TEdgeComponent for AwsMapper {
         if tedge_config.mqtt.bridge.built_in {
             let device_id = tedge_config.device.id.try_read(&tedge_config)?;
             let rules = built_in_bridge_rules(device_id)?;
+
+            let mut cloud_config = rumqttc::MqttOptions::new(
+                tedge_config.device.id.try_read(&tedge_config)?,
+                tedge_config.aws.url.or_config_not_set()?.to_string(),
+                8883,
+            );
+            cloud_config.set_clean_session(false);
+            use_key_and_cert(
+                &mut cloud_config,
+                &tedge_config.aws.root_cert_path,
+                &tedge_config,
+            )?;
             let bridge_actor = MqttBridgeActorBuilder::new(
                 &tedge_config,
-                "tedge-mapper-bridge-az".to_owned(),
+                "tedge-mapper-bridge-aws".to_owned(),
                 rules,
-                &tedge_config.az.root_cert_path,
+                cloud_config,
             )
             .await;
             runtime.spawn(bridge_actor).await?;
@@ -85,8 +98,7 @@ fn built_in_bridge_rules(remote_client_id: &str) -> Result<BridgeConfig, anyhow:
     bridge.forward_from_remote("cmd/#", local_prefix, device_id_prefix)?;
 
     // topic to interact with the shadow of the device
-    bridge.forward_from_local("shadow/#", local_prefix, things_prefix.clone())?;
-    bridge.forward_from_remote("shadow/#", local_prefix, things_prefix)?;
+    bridge.forward_bidirectionally("shadow/#", local_prefix, things_prefix.clone())?;
 
     // echo topic mapping to check the connection
     bridge.forward_from_local("", "aws/test-connection", conn_check.clone())?;
--- a/crates/core/tedge_mapper/src/aws/mapper.rs
+++ b/crates/core/tedge_mapper/src/aws/mapper.rs
@@ -36,7 +36,7 @@ impl TEdgeComponent for AwsMapper {
             let device_id = tedge_config.device.id.try_read(&tedge_config)?;
             let rules = built_in_bridge_rules(device_id)?;
 
-            let mut cloud_config = rumqttc::MqttOptions::new(
+            let mut cloud_config = tedge_mqtt_bridge::MqttOptions::new(
                 tedge_config.device.id.try_read(&tedge_config)?,
                 tedge_config.aws.url.or_config_not_set()?.to_string(),
                 8883,
--- a/crates/core/tedge_mapper/src/aws/mapper.rs
+++ b/crates/core/tedge_mapper/src/aws/mapper.rs
@@ -10,6 +10,8 @@ use tedge_actors::MessageSource;
 use tedge_actors::NoConfig;
 use tedge_api::mqtt_topics::MqttSchema;
 use tedge_config::TEdgeConfig;
+use tedge_mqtt_bridge::BridgeConfig;
+use tedge_mqtt_bridge::MqttBridgeActorBuilder;
 use tracing::warn;
 
 const AWS_MAPPER_NAME: &str = "tedge-mapper-aws";
@@ -29,6 +31,18 @@ impl TEdgeComponent for AwsMapper {
     ) -> Result<(), anyhow::Error> {
         let (mut runtime, mut mqtt_actor) =
             start_basic_actors(self.session_name(), &tedge_config).await?;
+        if tedge_config.mqtt.bridge.built_in {
+            let device_id = tedge_config.device.id.try_read(&tedge_config)?;
+            let rules = built_in_bridge_rules(device_id)?;
+            let bridge_actor = MqttBridgeActorBuilder::new(
+                &tedge_config,
+                "tedge-mapper-bridge-az".to_owned(),
+                rules,
+                &tedge_config.az.root_cert_path,
+            )
+            .await;
+            runtime.spawn(bridge_actor).await?;
+        }
         let clock = Box::new(WallClock);
         let mqtt_schema = MqttSchema::with_root(tedge_config.mqtt.topic_root.clone());
         let aws_converter = AwsConverter::new(
@@ -57,4 +71,31 @@ fn get_topic_filter(tedge_config: &TEdgeConfig) -> TopicFilter {
         }
     }
     topics
+}
+
+fn built_in_bridge_rules(remote_client_id: &str) -> Result<BridgeConfig, anyhow::Error> {
+    let local_prefix = "aws/";
+    let device_id_prefix = format!("thinedge/{remote_client_id}/");
+    let things_prefix = format!("$aws/things/{remote_client_id}/");
+    let conn_check = format!("thinedge/devices/{remote_client_id}/test-connection");
+    let mut bridge = BridgeConfig::new();
+
+    // telemetry/command topics for use by the user
+    bridge.forward_from_local("td/#", local_prefix, device_id_prefix.clone())?;
+    bridge.forward_from_remote("cmd/#", local_prefix, device_id_prefix)?;
+
+    // topic to interact with the shadow of the device
+    bridge.forward_from_local("shadow/#", local_prefix, things_prefix.clone())?;
+    bridge.forward_from_remote("shadow/#", local_prefix, things_prefix)?;
+
+    // echo topic mapping to check the connection
+    bridge.forward_from_local("", "aws/test-connection", conn_check.clone())?;
+    bridge.forward_from_remote("", "aws/connection-success", conn_check)?;
+
+    Ok(bridge)
+}
+
+#[test]
+fn bridge_rules_are_valid() {
+    built_in_bridge_rules("test-device-id").unwrap();
 }
--- a/crates/core/tedge_mapper/src/az/mapper.rs
+++ b/crates/core/tedge_mapper/src/az/mapper.rs
@@ -10,6 +10,8 @@ use tedge_actors::MessageSource;
 use tedge_actors::NoConfig;
 use tedge_api::mqtt_topics::MqttSchema;
 use tedge_config::TEdgeConfig;
+use tedge_mqtt_bridge::BridgeConfig;
+use tedge_mqtt_bridge::MqttBridgeActorBuilder;
 use tracing::warn;
 
 const AZURE_MAPPER_NAME: &str = "tedge-mapper-az";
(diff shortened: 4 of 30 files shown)


crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -40,22 +42,325 @@ use tedge_api::main_device_health_topic;
...
+
+impl SingleConverter {
+    fn convert_topic<'a>(&'a self, topic: &'a str) -> Cow<'a, str> {
+        self.0.iter().find_map(|rule| rule.apply(topic)).unwrap()
+    }
+}
+
+fn prepend<'a>(value: &'a str, prefix: &str) -> Cow<'a, str> {
+    match prefix {
+        "" => Cow::Borrowed(value),
```
albinsuresh, line 343: I'm assuming that this is the reason why you are using raw `&str` instead of `Topic` in all the bridge rule APIs, to avoid that extra string allocation. Just wondering if returning `Topic` would have been better as the returned `Cow` is going to be converted to a `Topic` eventually, for the target messages to be created.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -99,36 +405,39 @@ impl MqttBridgeActorBuilder {
             8883,
         );
         cloud_config.set_manual_acks(true);
+        cloud_config.set_clean_session(false);
```
albinsuresh, line 408: Is this required? Because I don't think that'll work for C8y as it expects the clean session to be true as documented [here](https://cumulocity.com/docs/device-integration/mqtt/#mqtt-clean-session)

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -40,22 +42,325 @@ use tedge_api::main_device_health_topic;
...
+            let converter = SingleConverter(vec![
+                BridgeRule::new("topic", "x/", "b/"),
+                BridgeRule::new("#", "a/", "c/"),
+            ]);
+            assert_eq!(converter.convert_topic("a/topic"), "c/topic");
+        }
+    }
+}
+
+struct SingleConverter(Vec<BridgeRule>);
```
albinsuresh, line 333:
```suggestion
struct TopicConverter(Vec<BridgeRule>);
```
In terms of naming, this feels more like the topic converter, and the existing struct named `TopicConverter` looks more like a `BridgeConfig` itself.

albinsuresh (commented):
I didn't fully get the benefit of using raw `&str` form of `Topic`s and `TopicFilter`s in the newly added APIs. It felt like an eventual conversion to a `Topic` is unavoidable and hence the benefits of using `Cow<&str>` might not be significant.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -40,22 +42,325 @@ use tedge_api::main_device_health_topic;
 use tedge_config::MqttAuthConfig;
 use tedge_config::TEdgeConfig;
 
+pub trait MqttBridgeRule {
```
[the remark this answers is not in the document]
jarhodes314: Nope, I'll remove it

crates/extensions/tedge_mqtt_bridge/src/lib.rs
reubenmiller: Yes we need to make sure the mqtt connection to the cloud uses a clean session, whilst using a non-clean session when connecting to the local broker.

didier-wenzek: > Yes we need to make sure the mqtt connection to the cloud uses a clean session, whilst using a non-clean session when connecting to the local broker.

Correct, but unfortunate.
Using a [clean session](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Ref362965194) on each and every connection defeats the purpose of QoS 1 and 2.

crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -30,6 +32,18 @@ impl TEdgeComponent for AwsMapper {
...
         let (mut runtime, mut mqtt_actor) =
             start_basic_actors(self.session_name(), &tedge_config).await?;
+        if tedge_config.mqtt.bridge.built_in {
+            let device_id = tedge_config.device.id.try_read(&tedge_config)?;
+            let rules = built_in_bridge_rules(device_id)?;
+            let bridge_actor = MqttBridgeActorBuilder::new(
+                &tedge_config,
+                "tedge-mapper-bridge-az".to_owned(),
+                rules,
+                &tedge_config.az.root_cert_path,
```
didier-wenzek, line 42:
```suggestion
                &tedge_config.aws.root_cert_path,
```

crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -59,3 +73,30 @@ fn get_topic_filter(tedge_config: &TEdgeConfig) -> TopicFilter {
...
+    let conn_check = format!("thinedge/devices/{remote_client_id}/test-connection");
+    let mut bridge = BridgeConfig::new();
+
+    // telemetry/command topics for use by the user
+    bridge.forward_from_local("td/#", local_prefix, device_id_prefix.clone())?;
+    bridge.forward_from_remote("cmd/#", local_prefix, device_id_prefix)?;
+
+    // topic to interact with the shadow of the device
+    bridge.forward_from_local("shadow/#", local_prefix, things_prefix.clone())?;
+    bridge.forward_from_remote("shadow/#", local_prefix, things_prefix)?;
```
didier-wenzek, line 90: How do we ensure this doesn't lead to an infinite loop?

crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -59,3 +73,30 @@ fn get_topic_filter(tedge_config: &TEdgeConfig) -> TopicFilter {
     }
     topics
 }
+
+fn built_in_bridge_rules(remote_client_id: &str) -> Result<BridgeConfig, anyhow::Error> {
```
didier-wenzek, line 77: Could this `BridgeConfig` also used to generate the [mosquitto bridge config](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/bridge/aws.rs#L19)? This would avoid to repeat ourselves.

crates/core/tedge_mapper/src/az/mapper.rs
```diff
@@ -57,3 +72,27 @@ fn get_topic_filter(tedge_config: &TEdgeConfig) -> TopicFilter {
     }
     topics
 }
+
+fn built_in_bridge_rules(remote_clientid: &str) -> anyhow::Result<BridgeConfig> {
```
didier-wenzek, line 76: Same comment as for the AWS bridge. It would be good to avoid repeating ourselves as the same config is already described [here](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/bridge/azure.rs#L63).

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -35,27 +39,190 @@ pub use mqtt_channel::MqttError;
...
+pub struct BridgeConfig {
+    local_to_remote: Vec<BridgeRule>,
+    remote_to_local: Vec<BridgeRule>,
+}
+
+#[derive(Debug, Clone)]
+pub struct BridgeRule {
+    topic_filter: Cow<'static, str>,
+    prefix_to_remove: Cow<'static, str>,
+    prefix_to_add: Cow<'static, str>,
```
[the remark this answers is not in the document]
didier-wenzek: I prefer the original names compared to those proposed, even if not perfect.

A `BridgeRule` can be ` local_to_remote` or `remote_to_local`, so the incoming and outgoing terms are confusing. The add and remove internal usage hints are okay as these names are only used privately.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -35,27 +39,190 @@ pub use mqtt_channel::MqttError;
...
 use tedge_config::TEdgeConfig;
 
+#[derive(Default, Debug, Clone)]
+pub struct BridgeConfig {
+    local_to_remote: Vec<BridgeRule>,
+    remote_to_local: Vec<BridgeRule>,
+}
+
+#[derive(Debug, Clone)]
+pub struct BridgeRule {
```
didier-wenzek, line 53:
```suggestion
struct BridgeRule {
```

crates/extensions/tedge_mqtt_bridge/src/lib.rs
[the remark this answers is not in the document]
albinsuresh: Yeah, I also intentionally avoided the names `local_prefix` and `remote_prefix` for the same reason. Suggested `incoming ` and `outgoing` more from the bridge client's perspective, more like `input_prefix` and `output_prefix`. But yeah, it was just an optional suggestion and can be ignored.

crates/core/tedge_mapper/src/aws/mapper.rs
jarhodes314: We only have one client connected to the local broker, and we shouldn't receive messages that we originally sent. I'll attempt to add a test for this.

crates/core/tedge_mapper/Cargo.toml
```diff
@@ -25,6 +25,7 @@ collectd_ext = { workspace = true }
 flockfile = { workspace = true }
 mqtt_channel = { workspace = true }
 reqwest = { workspace = true }
+rumqttc = { workspace = true }
```
didier-wenzek, line 28: I would avoid to add this direct dependency to `rumqttc`.

As this is done only to build an `rumqttc::MqttOptions` struct passed to `MqttBridgeActorBuilder`, can this dependency hidden behind a light facade by `tedge_mqtt_bridge`?

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -45,24 +48,53 @@ impl TEdgeComponent for CumulocityMapper {
...
+
+            let mut cloud_config = rumqttc::MqttOptions::new(
+                tedge_config.device.id.try_read(&tedge_config)?,
+                tedge_config
+                    .c8y
+                    .mqtt
+                    .or_config_not_set()?
+                    .host()
+                    .to_string(),
+                8883,
```
didier-wenzek, line 82: Should be `tedge_config.c8y.mqtt.port()`.

crates/core/tedge_mapper/src/az/mapper.rs
```diff
@@ -30,6 +33,38 @@ impl TEdgeComponent for AzureMapper {
...
         let (mut runtime, mut mqtt_actor) =
             start_basic_actors(self.session_name(), &tedge_config).await?;
+        if tedge_config.mqtt.bridge.built_in {
+            let remote_clientid = tedge_config.device.id.try_read(&tedge_config)?;
+            let rules = built_in_bridge_rules(remote_clientid)?;
+
+            let mut cloud_config = rumqttc::MqttOptions::new(
+                remote_clientid,
+                tedge_config.az.url.or_config_not_set()?.to_string(),
+                8883,
```
didier-wenzek, line 43: Something is wrong here. I would expect call to `host()` and `port()` on `tedge_config.az.url`.

crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -30,6 +33,30 @@ impl TEdgeComponent for AwsMapper {
...
         let (mut runtime, mut mqtt_actor) =
             start_basic_actors(self.session_name(), &tedge_config).await?;
+        if tedge_config.mqtt.bridge.built_in {
+            let device_id = tedge_config.device.id.try_read(&tedge_config)?;
+            let rules = built_in_bridge_rules(device_id)?;
+
+            let mut cloud_config = rumqttc::MqttOptions::new(
+                tedge_config.device.id.try_read(&tedge_config)?,
+                tedge_config.aws.url.or_config_not_set()?.to_string(),
+                8883,
```
didier-wenzek, line 43: Something is wrong here. I would expect call to `host()` and `port()` on `tedge_config.aws.url`.

crates/extensions/tedge_mqtt_bridge/src/topics.rs
```diff
@@ -0,0 +1,34 @@
...
+
+pub struct TopicConverter(pub Vec<BridgeRule>);
+
+impl TopicConverter {
+    pub fn convert_topic<'a>(&'a self, topic: &'a str) -> Cow<'a, str> {
+        self.0
+            .iter()
+            .find_map(|rule| rule.apply(topic))
+            // TODO should this be an error
+            .unwrap_or_else(|| panic!("Failed to convert {topic:?} with {:?}", self.0))
```
didier-wenzek, line 32: As long as the bridge rules are hard-coded in the mapper, it's okay to panic here.

didier-wenzek (commented):
I still have to review the bulk of the changes, notably in `tests/bridge.rs`.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -83,55 +88,53 @@ impl MqttBridgeActorBuilder {
...
-            8883,
-        );
+        local_config.set_clean_session(false);
+
         cloud_config.set_manual_acks(true);
 
-        cloud_config.set_transport(Transport::tls_with_config(tls_config.into()));
-        let topic_prefix = format!("{prefix}/");
         let (local_client, local_event_loop) = AsyncClient::new(local_config, 10);
         let (cloud_client, cloud_event_loop) = AsyncClient::new(cloud_config, 10);
```
didier-wenzek, line 100: > `sudo tedge connect c8y --test` returned an unexpected exit code stdout: Sending packets to check connection.

I observe on my laptop the same behavior as on the failing system tests: no JWT token is received when requested on `c8y/s/uat`:

- The message published on `c8y/s/uat` is not even processed by the c8y built-in bridge.
- The built-in bridge is even stuck: processing no more messages after a first batch of 11 messages.
- Changing the `cap` parameter from 10 to some  crazy number 10000 unblocks the built-in bridge
  (this change has to be for the `local_config`, the `cloud_config` and the `bidirectional_channel`).

Cargo.toml
```diff
@@ -187,6 +187,11 @@ panic = "unwind"
 strip = "symbols"
 overflow-checks = true
 
+[profile.dev]
+opt-level = "z"
+panic = "unwind"
+strip = "symbols"
+overflow-checks = true
```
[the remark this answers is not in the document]
jarhodes314: No, I need to work out how to make a local configuration for this so I don't keep accidentally committing it, but need it to test on a raspberry pi in seconds rather than minutes

crates/core/tedge_mapper/src/az/mapper.rs
jarhodes314: `tedge_config.az.url` is a `ConnectUrl`, which explicitly doesn't support a port. We know the port is 8883 because that is what Azure uses for MQTT connections.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -35,27 +39,190 @@ pub use mqtt_channel::MqttError;
...
+    topic_filter: Cow<'static, str>,
+    prefix_to_remove: Cow<'static, str>,
+    prefix_to_add: Cow<'static, str>,
+}
+
+fn prepend<'a>(target: Cow<'a, str>, prefix: &Cow<'a, str>) -> Cow<'a, str> {
+    match (prefix, target) {
+        (prefix, target) if prefix.is_empty() => target,
+        (prefix, target) if target.is_empty() => prefix.clone(),
+        (prefix, Cow::Borrowed(target)) => format!("{prefix}{target}").into(),
```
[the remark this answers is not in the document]
jarhodes314: This does make a difference (albeit not a particularly notable one). Most of the Cumulocity rules make use of this, for instance. They are `&'static str` to begin with as the topics don't change, and the prefix is empty.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: I've now understood and fixed the issue here. The issue was the change to the subscriptions. Previously we were subscribing before connecting, but with this PR, we subscribe upon receiving a `ConnAck`, which means we resubscribe after a dropped connection.

The issue with this is the cloud connection takes a reasonable period of time to come up after starting the mapper, and if during that time 10 messages are sent, this fills up the channel for the cloud event loop. When the cloud then successfully connects, it attempts to subscribe with `recv_client.subscribe_many(&topics).await`. This attempts to send a message to the event loop to process the subscription request, but the event loop is already full, so this results in a deadlock.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -35,27 +39,190 @@ pub use mqtt_channel::MqttError;
...
+}
+
+#[derive(Debug, Clone)]
+pub struct BridgeRule {
+    topic_filter: Cow<'static, str>,
+    prefix_to_remove: Cow<'static, str>,
+    prefix_to_add: Cow<'static, str>,
+}
+
+fn prepend<'a>(target: Cow<'a, str>, prefix: &Cow<'a, str>) -> Cow<'a, str> {
```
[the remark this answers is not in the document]
jarhodes314: I'm going to replace this with the `+` operator as it's unambiguous and we clearly have opposing opinions here.

crates/core/tedge_mapper/Cargo.toml
jarhodes314: Changed to a `pub use` in `tedge_mqtt_bridge` and then importing from there

crates/core/tedge_mapper/src/aws/mapper.rs
jarhodes314: Again, this is a `ConnectUrl`

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -414,6 +505,7 @@ impl RuntimeRequestSink for MqttBridgeActorBuilder {
 pub struct MqttBridgeActor {}
 
 #[async_trait]
+#[mutants::skip]
 impl Actor for MqttBridgeActor {
     fn name(&self) -> &str {
```
didier-wenzek, line 510:
```suggestion
impl Actor for MqttBridgeActor {
   #[mutants::skip]
    fn name(&self) -> &str {
```

didier-wenzek (approved):
Approved.

I really like the [use of `cargo mutants`](https://github.com/thin-edge/thin-edge.io/pull/2821/commits/1e4d2224f3f4d4590176608812ce69b135d64fff) to improve code coverage. The result gives strong confidence on this tricky piece of code.

```
$ cargo mutants --test-tool=nextest -p tedge_mqtt_bridge
Found 92 mutants to test
ok       Unmutated baseline in 39.6s build + 0.4s test
 INFO Auto-set test timeout to 20s
92 mutants tested in 6m 48s: 46 caught, 46 unviable
```

Issue #2761, opened 2024-03-05
Upload logs automatically after a command has been completed


reubenmiller:
**Is your feature request related to a problem? Please describe.**

It is difficult and slow to debug an command (operation) which failed, as it is difficult for the user to retrieve the related log file to an command instance as there are generally many different log commands, and finding the correct log file to the latest operation is non-trivial.

**Describe the solution you'd like**

An option to enable the automatic uploading of a command's associated log file when a command finishes (e.g. the status is set to `failed`). The file should be uploaded to the cloud where the operation was created from.

It might make sense to have this setting configurable to support the following settings:
* upload on failure
* upload on successful and failure
* don't upload

**Describe alternatives you've considered**

**Additional context**

albinsuresh:
### Test Plan

The feature has been implemented as described, where the workflow log file of the operation is automatically uploaded on completion of the operations, if configured to do so. But the log upload is currently limited to software update operations only.

Auto log upload is disabled by default, which can be enabled by using the config setting : `tc8y.operations.auto_log_upload` which supports values `on-failure`, `always` and `never`(default option).

E.g: `tedge config set c8y.operations.auto_log_upload always`

Three tests have been added to `software.robot` test to test the effect of all the config variants.

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test changes has been reviewed.
        tests/RobotFramework/tests/cumulocity/software_management/software.robot
- [x] QA has tested the function and it's functioning according description.

Issue #1526, opened 2022-10-25
Empty directory for a child ops not creating the child device in c8y


PradeepKiruvale:
**Describe the bug**
When a directory created in `/etc/tedge/operations/c8y/child_test`, its not creating the child device `child_test` in c8y.

Creates the child device in c8y only when at least one operation is inside the `child operation` directory.

**To Reproduce**
`sudo tedge connect c8y`
`sudo mkdir /etc/tedge/operations/c8y/child_test`

**Expected behavior**
Expect a child device to appear in c8y cloud under the `tedge` device.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [e.g. 0.1.0]

**Additional context**
Maybe we need to handle this error 
[tedge/errors] Is a directory (os error 21)

rina23q:
Deliver in 0.8.1.

PR #2833, opened 2024-04-22
Auto upload operation log on completion #2761

albinsuresh:
## Proposed changes

Auto-upload the operation log on software update execution. It's the plugin execution log file that's uploaded and not the workflow log file. Since the other operations don't have any such logs to be uploaded, the feature currently works only for software update operation.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


reubenmiller:
> Since the other operations don't have any such logs to be uploaded, the feature currently works only for software update operation.

Though can't we still add this as a generic feature, and just for a non-zero file size before uploading it, skipping the upload if the file size is zero? That the feature will work for all operation once they actually log something.

albinsuresh:
After an offline discussion with @reubenmiller, we have decided to do this PR slightly differently, by uploading the workflow log files for all the operations instead of operation specific log files like the one done in this PR. But currently there are many inconsistencies in how the logging happens for different operations as raised in #2835 and #2836. Those will have to be fixed as well for this feature to be useful/complete. But, the switch to uploading workflow logs can be done independently.

albinsuresh:
@reubenmiller @didier-wenzek Have updated the PR with a common log upload routine for all ops, that doesn't require explicit invocation per operation. Streamlining the command payloads types of all the ops to implement the `CommandPayload` trait, which enabled the conversion to the `GenericCommandPayload`  struct was the key to this approach.

So, the approach can be reviewed now. It's still marked draft because the parsing logic of ops like config, firmware and log are still not fully using the generic `Command` representation of them yet, which I'll add next, as it wasn't really needed for this feature to work. But, @didier-wenzek  I have a feeling that you might be repeating the same for #1526 as part of integrating log and config actors into the workflow subsystem.

crates/core/tedge_api/src/commands.rs
```diff
@@ -169,6 +171,10 @@ pub trait CommandPayload {
     fn set_error(&mut self, reason: String) {
         self.set_status(CommandStatus::Failed { reason });
     }
+
+    fn log_path(&self) -> Option<&Path>;
+
+    fn set_log_path(&mut self, path: &Path);
```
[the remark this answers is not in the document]
albinsuresh: Added.

crates/core/tedge_agent/src/tedge_operation_converter/actor.rs
```diff
@@ -143,7 +143,10 @@ impl TedgeOperationConverterActor {
             Ok(Some(new_state)) => {
                 self.persist_command_board().await?;
                 if new_state.is_init() {
-                    self.process_command_state_update(new_state).await?;
+                    self.process_command_state_update(
+                        new_state.update_with_key_value("logPath", log_file.path.as_str()),
+                    )
```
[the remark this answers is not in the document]
albinsuresh: Couldn't use `CommandPayload::set_log_path()` here as we're dealing with `GenericCommandState` representation of the commands here. I could parse `GenericCommandState::payload` into a `CommandPayload` and use that function, but a secondary parsing felt like an overkill just for updating that single field.

Instead, I added a `set_log_path` function to `GenericCommandState`, as we already have a desire to replace the newly introduced `GenericCommandPayload` with `GenericCommandState` itself in a follow-up tasks. So, currently, the log path is set using `GenericCommandState::set_log_path` by the `tedge-agent` is consumed using `CommandPayload::log_path() by the converter.

crates/core/tedge_api/src/commands.rs
```diff
@@ -697,6 +781,30 @@ impl LogUploadCmdPayload {
     }
 }
 
+impl CommandPayload for LogUploadCmdPayload {
```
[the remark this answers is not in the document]
albinsuresh: Resolved by #2846.

gligorisaev (commented):
Test **software.robot**

Steps:
- Operation log uploaded automatically with default auto_log_upload setting as onFailure
    * This test ensures that the operation log uploading behavior aligns with the configured setting (auto_log_upload).
    * It covers both successful and failed operation scenarios, validating the logging behavior accordingly.

- Operation log uploaded automatically with default auto_log_upload setting as always
    * These tests ensures that even successful operations trigger operation log uploads when auto_log_upload is set to "always".
    * It provides a robust validation of the behavior under different operation outcomes.

- Operation log uploaded automatically with default auto_log_upload setting as never
    * These test ensures that operation log uploads are suppressed when auto_log_upload is set to "never".
    * It confirms the expected behavior of no log uploads, regardless of operation outcome.

Test **custom_operation.robot**
- The use of libraries like JSONLibrary facilitates effective verification of JSON messages and data

crates/core/tedge_agent/src/software_manager/tests.rs
```diff
@@ -97,6 +97,7 @@ async fn test_new_software_update_operation() -> Result<(), DynError> {
             status: CommandStatus::Scheduled,
             update_list: vec![debian_list],
             failures: vec![],
+            log_path: None,
```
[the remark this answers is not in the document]
didier-wenzek: > But, I',m not sure if avoiding that at the cost of a secondary parsing is worth it.

Sure, but having to add a `log_path` field to all concrete command types (`LogUploadCmdPayload`, `ConfigUpdateCmdPayload`, `FirmwareUpdateCmdPayload` ...) while the command itself doesn't use this field is something I'm not really happy with.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1540,38 +1554,68 @@ impl CumulocityConverter {
...
-                ])
+                ]
+            }
+        };
+
+        Ok((messages, Some(command.into())))
+    }
+
+    async fn upload_operation_log(&mut self, command: GenericCommand) {
+        if command.target == *self.entity_store.main_device()
```
didier-wenzek, line 1593: Don't we have to also upload the log file for child devices?

crates/extensions/c8y_mapper_ext/src/operations/config_update.rs
```diff
@@ -91,7 +105,7 @@ impl CumulocityConverter {
             }
         };
 
-        Ok(messages)
+        Ok((messages, Some(command.into())))
```
didier-wenzek, line 108: All the command-processing method are built along the same pattern:
- The method is given an MQTT `message`
- This `message` is parsed into a `command` of the appropriate type
- The `command` is processed leading a vector of MQTT `messages`
- The `command` is then converted into a `GenericCommand` to gather all log related data.
- The intermediate `command` acts a vehicle the `log_path` field without needing it.

So why not:

```suggestion
        Ok((messages, message.into()))
```

with an `impl From<MqttMessage> for Option<GenericCommand>` ?

I agree that this implies the message payload to be parsed twice, but this also keeps the command structs free of any unused `log_path`.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Yes, I'll revert that. This guard was added in the initial version, where we were uploading the software plugin log, instead of the workflow log, which was only available for the main device. But now that we've switched to uploading the workflow logs, this guard is not needed anymore.

crates/extensions/c8y_mapper_ext/src/operations/config_update.rs
albinsuresh: > message payload to be parsed twice

This was the very first solution that came to my mind as well, but felt like an overkill and tried to avoid, as it's adding runtime cost to the operation handling just to parse an extra field. Of course, the impact is minimal, considering the overall cost of an operation handling.

> this also keeps the command structs free of any unused `log_path`

Now I think I understand your perspective. Yes, that `log_path` field would not make any sense to someone purely looking at the command payload struct APIs without any context of the mapper or the agent. So yes, it is clearly an example of the impl details creeping into the API, which should be avoided

One alternate solution which might be less intrusive and symmetric would be to add an additional `Value` field to all the `CommandPayload` implementors to capture any additional fields in the payload other than the essential ones.  We can argue that it also represents the open-ness of the payload. Then, we can implement an `impl<T: CommandPayload> From<Command<T>> for GenericCommandState` and eliminate the newly introduced `GenericCommand` struct. That way, we get symmetry as well, where the agent uses `GenericCommandState::set_log_path` and the converter uses the corresponding `GenericCommandState::get_log_path`, as you had suggested in an earlier comment. WDYT?

didier-wenzek: > One alternate solution which might be less intrusive and symmetric would be to add an additional `Value` field to all the `CommandPayload` implementors to capture any additional fields in the payload other than the essential ones.  We can argue that it also represents the open-ness of the payload. Then, we can implement an `impl<T: CommandPayload> From<Command<T>> for GenericCommandState` and eliminate the newly introduced `GenericCommand` struct. That way, we get symmetry as well, where the agent uses `GenericCommandState::set_log_path` and the converter uses the corresponding `GenericCommandState::get_log_path`, as you had suggested in an earlier comment. WDYT?

Indeed, it's a good idea to add a `Value` field that holds the whole message payload with all the extra fields such as the `log_path` which are not used by the command itself but are required for bookkeeping.

The next question is where to store this `Value`? In all the `CommandPayload` implementations (as you suggest), or directly in the `Command<T>` struct along the decoded payload of type `T`? The latter might be less intrusive.

albinsuresh: > The next question is where to store this `Value`? In all the `CommandPayload` implementations (as you suggest), or directly in the `Command<T>` struct along the decoded payload of type `T`? The latter might be less intrusive.

I was thinking of keeping them in all the impls as that better represents the flexibility in the palyoad. Keeping another copy of `Value` with all the payload fields in the `Command` would have been easier to manage from the users' perspective, but I'm afraid that it'd again involves a 2-step parsing of the payload, once into `Value` and then into the concrete payload structs, unless there's some parsing trick that can deserialize the same byte stream into both the representations simultaneously.

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -3,6 +3,7 @@ Resource            ../../../resources/common.resource
 Library             ThinEdgeIO
 Library             Cumulocity
 Library             OperatingSystem
+Library             JSONLibrary
```
rina23q, line 6: This file has no other change, so I guess this can be removed.

crates/common/tedge_config/src/tedge_config_cli/models/mod.rs
```diff
@@ -25,3 +26,35 @@ pub use self::ipaddress::*;
...
+        match input {
+            "never" => Ok(AutoLogUpload::Never),
+            "always" => Ok(AutoLogUpload::Always),
+            "onFailure" => Ok(AutoLogUpload::OnFailure),
+            _ => Err(InvalidAutoLogUploadConfig {
+                input: input.to_string(),
+            }),
+        }
+    }
+}
```
rina23q, line 60: I think it's better to create a new file under `tegge_config/cli/models/` and move the code there. All other enum or struct that are used in tedge config are stored in that directory.

albinsuresh: Yeah, I noticed that pattern being followed for other config enums like `apt_conifg.rs` and `c8y_software_management.rs`. If that's an accepted convention and not mere accident, I can move this one also to a separate module. But, to be honest, a module per enum/struct looked like an overkill to me, which is why I consciously kept this one at the top module level itself. I usually create dedicated modules when there are multiple related structs/enums that are to be kept together.

crates/extensions/c8y_mapper_ext/src/operations/config_update.rs
albinsuresh: Unfortunately, the plan to have an `additional_fields: HashMap<String, Value>` field in all payloads to capture all non-essential fields did not quite work due to a conflict with the `CommandStatus` definition that has `serde(flatten)` attribute defined on it (primarily to capture the failed `status` and `reason` fields into the `Failed` same enum variant). When both `status` and `additional_fields` has `serde(flatten)` attribute defined on them, it was leading to a conflict where the `status` and `reason` fields get duplicated inside the `additional_fields` map as well. There was no easy way to fix this other than writing a custom deserialization logic, which would have complicated things for little value.

So, I was out of other options but to switch to the simpler solution of just parsing the `MqttMessage` into `GenericCommandState`, accepting the cost of double parsing.

crates/common/tedge_config/src/tedge_config_cli/models/mod.rs
rina23q: I understand your point, but having only one enum in the top level and all others in the separated files doesn't look well organized. All enums for config should be treated in some clear rule, either all in the top or all in the separated files, or some light-weight enums in the top and heavy-weight in the separated files. We need a clear rule.

didier-wenzek (commented):
I'm not fully convinced by these `extra_fields`. However, this is not a blocker, as things are working as expected.

crates/common/tedge_config/src/tedge_config_cli/models/mod.rs
albinsuresh: > some light-weight enums in the top and heavy-weight in the separated files. We need a clear rule.

That's what I'd follow too. The other lightweight enums can also be moved into the top level in a follow-up PR.

crates/core/tedge_agent/src/restart_manager/actor.rs
```diff
@@ -116,7 +116,7 @@ impl RestartManagerActor {
     }
 
     async fn process_pending_restart_operation(&mut self) -> Option<RestartCommand> {
-        match self.state_repository.load().await {
+        match self.load_state_repository().await {
```
didier-wenzek, line 119: What's the motivation for this change?

I saw the command is *sanitized* by `load_state_repository()`, but wonder why this is required. Touching a critical piece of code sounds inadequate for this log-related PR.

crates/core/tedge_api/src/commands.rs
```diff
@@ -142,6 +157,12 @@ where
         }
     }
 
+    pub fn payload_from(bytes: &'a [u8]) -> Result<Payload, serde_json::Error> {
+        let mut payload = Payload::from_slice(bytes)?;
+        payload.sanitize();
```
didier-wenzek, line 162: I don't get why calling this `sanitize` method so important?

This is about two fields (status and reason) in operation-related messages received at low frequency.

albinsuresh: If we don't sanitize the payload (esp, removing that duplicate `status` field), then when that payload gets serialized and persisted into the state repository, it'll have two `status` fields. When the same is deserialized on the next startup, it'll cause the parsing to fail.

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -285,9 +286,7 @@ fn request_from_message(
     if message.payload_bytes().is_empty() {
         Ok(None)
     } else {
-        Ok(Some(LogUploadCmdPayload::from_json(
-            message.payload_str()?,
-        )?))
+        Ok(Some(LogUploadCmd::payload_from(message.payload_bytes())?))
```
didier-wenzek, line 289: Beware that this usage of `payload_from` will be made useless by https://github.com/thin-edge/thin-edge.io/pull/2844 which purpose is to use the same pattern for Log&Config management as for Software management: i.e. to have the log manager MQTT independent. The `request_from_message` method will be removed.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -548,11 +550,11 @@ impl ConfigOperation {
             Ok(None)
         } else if config.config_snapshot_topic.accept(message) {
             Ok(Some(ConfigOperation::Snapshot(
-                ConfigSnapshotCmdPayload::from_json(message.payload_str()?)?,
+                ConfigSnapshotCmd::payload_from(message.payload_bytes())?,
             )))
         } else if config.config_update_topic.accept(message) {
             Ok(Some(ConfigOperation::Update(
-                ConfigUpdateCmdPayload::from_json(message.payload_str()?)?,
+                ConfigUpdateCmd::payload_from(message.payload_bytes())?,
```
didier-wenzek, line 557: Beware that this usage of `payload_from` will be made useless by https://github.com/thin-edge/thin-edge.io/pull/2844 which purpose is to use the same pattern for Log&Config management as for Software management: i.e. to have the log manager MQTT independent. The `request_from_message` method will be removed.

reubenmiller:
I just tried out the feature

1. We should add a more descriptive text to the uploaded log, as for a failed "software_update" operation, the uploaded log file just says "software_update" without any context that it is log file
2. Can we add the filename to the meta information (this feature was added recently in https://github.com/thin-edge/thin-edge.io/issues/2763). If we don't add a useful filename, then when a file is downloaded by the user, they lose most of the human readable context about the original file (e.g. which device did it come from, what log is it from etc.)

**Example of point 1**
image

**Example of point 2**

image

crates/extensions/c8y_mapper_ext/src/operations/config_update.rs
didier-wenzek: We decided to go with an explicit `log_path` added to all command specific payload. There are opportunities for simplifications, however not before https://github.com/thin-edge/thin-edge.io/pull/2844 is merged: one needs first to use the same representation for all commands. Currently, there are some small but blocking differences for Log & Config management commands.

didier-wenzek (commented):
After many attempts and discussions, we agreed with @albinsuresh to use the straightforward approach with a `log_path` field added to all command types despite this field is only indirectly used on the json representation of these commands. 

This PR can be merged once addressed [this comment](https://github.com/thin-edge/thin-edge.io/pull/2833#issuecomment-2097743163)

crates/core/tedge_agent/src/restart_manager/actor.rs
albinsuresh: Obsolete, as we switched to a different solution.

reubenmiller:
Just some really minor points (after having another look at other aspects of the UX).

1. The allowed values for `c8y.operations.auto_log_upload` should be mentioned when using the `--doc` flag on the configuration list. Currently it only shows one example:

    ```sh
    $ tedge config list --doc
            c8y.operations.auto_log_upload  Auto-upload the operation log once it finishes. 
                                        Example: always
    ```


    Though what is nice is that when the user gives a wrong value, that they are also presented with the supported values, however some users won't even try entering a values without first knowning the supported values.

    ```
    # tedge config set c8y.operations.auto_log_upload onfailure
    Error: failed to set the configuration key: 'c8y.operations.auto_log_upload' with value: onfailure.

    Caused by:
    0: Failed to parse input
    1: Failed to parse flag: onfailure. Supported values are: 'never', 'always' or 'onFailure'
    ```

2. Use `on-failure` instead of `onFailure` (more preference to avoid upper/lower case variants)

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1530,74 +1559,137 @@ impl CumulocityConverter {
...
+            }
+        };
+
+        Ok((
+            messages,
+            Some(command.into_generic_command(&self.mqtt_schema)),
+        ))
+    }
+
+    pub async fn upload_operation_log(
```
didier-wenzek, line 1600: The code of this method is close to both [`CumulocityConverter::handle_fts_log_download_result`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs#L180) and [`CumulocityConverter::handle_fts_config_download_result`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs#L181) which are already really close to each other.

A single method must be used to upload logs, log excerpts and config snapshots. This would avoid user-confusing small differences in upload content description.

didier-wenzek (changes requested):

reubenmiller:
@albinsuresh Since the log consolidation work probably won't be finished before the next release (1.1.0),  can you please change the default config value to "never". Reason is that we want to provide a consistent feature, and uploading only partially useful log files will detract too much from the usefulness of this feature.

didier-wenzek (approved):

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -69,6 +69,44 @@ Manual software_update operation request
     ...    expected_status=failed
     ...    c8y_fragment=c8y_SoftwareUpdate
 
+Operation log uploaded automatically with auto_log_upload setting as on-failure
+    Execute Command    tedge config set c8y.operations.auto_log_upload on-failure
+    Restart Service    tedge-mapper-c8y
+
+    # Validate that the operation log is NOT uploaded for a successful operation
+    ${OPERATION}=    Install Software        c8y-remote-access-plugin
```
gligorisaev, line 77: I would suggest Parametrizing test data, it can enhance test coverage and maintainability by allowing the same test case to be executed with different inputs. for example:
```
*** Variables *** 
${SUCCESS_PACKAGE}    c8y-remote-access-plugin 
${FAILED_PACKAGE}     non-existent-package

*** Test Cases ***
Operation log uploaded automatically with default auto_log_upload setting as onFailure
    ${OPERATION}=    Install Software        ${SUCCESS_PACKAGE}
    ...

Operation log uploaded automatically with default auto_log_upload setting as always
    ${OPERATION}=    Install Software        ${SUCCESS_PACKAGE}
    ...

Operation log uploaded automatically with default auto_log_upload setting as never
    ${OPERATION}=    Install Software        ${SUCCESS_PACKAGE}
```

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -69,6 +69,44 @@ Manual software_update operation request
...
+    Operation Should Be SUCCESSFUL           ${OPERATION}    timeout=60
```
gligorisaev, line 78: I would parametrize the timeouts as well, If timeouts are common across multiple test cases, it make sence to parameterize them to ensure consistency and easy modification, for example:
```
*** Variables ***
${TIMEOUT}    60

*** Test Cases ***
Operation log uploaded automatically with default auto_log_upload setting as onFailure
    Operation Should Be SUCCESSFUL           ${OPERATION}    timeout=${TIMEOUT}
    ...

Operation log uploaded automatically with default auto_log_upload setting as always
    Operation Should Be SUCCESSFUL           ${OPERATION}    timeout=${TIMEOUT}
```

Issue #2355, opened 2023-10-19
tedge-watchdog: Using low values for `WatchdogSec` leads to tedge-watchdog sending health check messages too fast


Bravo555:
**Describe the bug**

When trying to use low values for `WatchdogSec` attribute, like `2` or `3`, tedge-watchdog sends messages with an interval of ~0.001s:

```text
Oct 19 15:35:26 e17ab0835ebd tedge-watchdog[318793]: 2023-10-19T15:35:26.39360965Z  WARN tedge_watchdog::systemd_watchdog: No health check response received from device/main/service/tedge-mapper-c8y in time
Oct 19 15:35:26 e17ab0835ebd tedge-watchdog[318793]: 2023-10-19T15:35:26.39486311Z  WARN tedge_watchdog::systemd_watchdog: No health check response received from device/main/service/tedge-mapper-c8y in time
Oct 19 15:35:26 e17ab0835ebd tedge-watchdog[318793]: 2023-10-19T15:35:26.39611997Z  WARN tedge_watchdog::systemd_watchdog: No health check response received from device/main/service/tedge-mapper-c8y in time
Oct 19 15:35:26 e17ab0835ebd tedge-watchdog[318793]: 2023-10-19T15:35:26.39736915Z  WARN tedge_watchdog::systemd_watchdog: No health check response received from device/main/service/tedge-mapper-c8y in time
Oct 19 15:35:26 e17ab0835ebd tedge-watchdog[318793]: 2023-10-19T15:35:26.39863253Z  WARN tedge_watchdog::systemd_watchdog: No health check response received from device/main/service/tedge-mapper-c8y in time
Oct 19 15:35:26 e17ab0835ebd tedge-watchdog[318793]: 2023-10-19T15:35:26.399971749Z  WARN tedge_watchdog::systemd_watchdog: No health check response received from device/main/service/tedge-mapper-c8y in time
```

**To Reproduce**

1. Connect to c8y
2. Stop `tedge-watchdog` and `tedge-mapper-c8y` services
3. Add `WatchdogSec=3` to `/lib/systemd/system/tedge-mapper-c8y.service` under `[Service]` section
4. Start `tedge-watchdog`
5. `journalctl -u tedge-watchdog -f` should show it sending messages very fast, like in the snippet in the section above

**Expected behavior**

[...]

PR #2834, opened 2024-04-23
fix(#2355): cap tedge-watchdog interval to 1s minimum

Bravo555:
## Proposed changes

Cap tedge-watchdog interval in which we expect to receive health check response to a minimum of 1s.

I also had questions of "why is the interval one-fourth of what we read from `WatchdogSec`?", so decided to make it a constant and document it a little more clearly. Found that this ratio was changed [here](https://github.com/thin-edge/thin-edge.io/pull/1060) and [here](https://github.com/thin-edge/thin-edge.io/pull/1170), but the PRs don't mention why these specific values were chosen (and links to issues are missing).

## Types of changes

## Paste Link to the issue

- #2355 

## Checklist

## Further comments

### DIFF
--- a/crates/core/tedge_watchdog/src/systemd_watchdog.rs
+++ b/crates/core/tedge_watchdog/src/systemd_watchdog.rs
@@ -33,6 +33,13 @@ use tracing::warn;
 
 const SERVICE_NAME: &str = "tedge-watchdog";
 
+/// How many times more often do we send notify to systemd watchdog, than is necessary from the
+/// `WatchdogSec` value.
+///
+/// Notifications are sent more often to make sure we don't miss the notify interval due to
+/// a timing misalignment.
+const NOTIFY_SEND_FREQ_RATIO: u64 = 4;
+
 // TODO: extract to common module
 #[derive(Debug, Serialize, Deserialize)]
 pub struct HealthStatus {
@@ -61,7 +68,10 @@ async fn start_watchdog_for_self() -> Result<(), WatchdogError> {
                     let _ = notify_systemd(process::id(), "WATCHDOG=1").map_err(|e| {
                         eprintln!("Notifying systemd failed with {}", e);
                     });
-                    tokio::time::sleep(tokio::time::Duration::from_secs(interval / 4)).await;
+                    let delay = tokio::time::Duration::from_secs(
+                        (interval / NOTIFY_SEND_FREQ_RATIO).max(1),
+                    );
+                    tokio::time::sleep(delay).await;
                 }
             });
             Ok(())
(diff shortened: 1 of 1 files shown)


crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -61,7 +68,10 @@ async fn start_watchdog_for_self() -> Result<(), WatchdogError> {
                     let _ = notify_systemd(process::id(), "WATCHDOG=1").map_err(|e| {
                         eprintln!("Notifying systemd failed with {}", e);
                     });
-                    tokio::time::sleep(tokio::time::Duration::from_secs(interval / 4)).await;
+                    let delay = tokio::time::Duration::from_secs(
+                        (interval / NOTIFY_SEND_FREQ_RATIO).max(1),
```
didier-wenzek, line 72: Don't you mean:

```suggestion
                        (interval / NOTIFY_SEND_FREQ_RATIO).min(1),
```

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -129,12 +138,14 @@ async fn start_watchdog_for_tedge_services(tedge_config_dir: PathBuf) {
 
                 let tedge_config_location = tedge_config_location.clone();
                 watchdog_tasks.push(tokio::spawn(async move {
+                    //
+                    let interval = Duration::from_secs((interval / NOTIFY_SEND_FREQ_RATIO).max(1));
```
didier-wenzek, line 142:
```suggestion
                    let interval = Duration::from_secs((interval / NOTIFY_SEND_FREQ_RATIO).min(1));
```

crates/core/tedge_watchdog/src/systemd_watchdog.rs
didier-wenzek: Oops, you are correct !

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -33,6 +33,13 @@ use tracing::warn;
 
 const SERVICE_NAME: &str = "tedge-watchdog";
 
+/// How many times more often do we send notify to systemd watchdog, than is necessary from the
+/// `WatchdogSec` value.
```
albinsuresh, line 37: It couldn't be `WatchdogSec` value itself, as there's the risk of missing notifications due to the additional delay between the health check request and response. For e.g, assume a `WatchdogSec` value of `30s` and the systemd watchdog timer starts at `T`. If our tedge_watchdog wakes up from its sleep at `T+29` and sends the health check request, the response might be slightly delayed and received at `T+31` leading to the watchdog timer expiring leading to a restart.

That's why we started with half of `WatchdogSec` which gives sufficient buffer for the responses to arrive. Theoretically that should have worked, but in practice, even that interval seemed insufficient and we still encountered occasional notification misses, which lead to the quarter value.

BTW, even smaller fractions won't work if the `WatchdogSec` value itself is too small, like a couple of seconds, as the resultant computed delay of 1 sec might also not work, if the request-response delay is higher.

albinsuresh (approved):

Issue #2370, opened 2023-10-25
Health status message mapping doesn't work when topic contains `bridge`


Bravo555:
**Describe the bug**

When any part of the topic contains the word `bridge`, message mapping is disabled, due to this check:

https://github.com/thin-edge/thin-edge.io/blob/d3ed45f4f1f8909a113693840bfec18b5bbc55c6/crates/extensions/c8y_mapper_ext/src/service_monitor.rs#L30-L33

**To Reproduce**

1. Publish a health status message for an entity with a topic id that **doesn't** contain `bridge`:
   `tedge mqtt pub te/device/main/service/my-custom-service/status/health '{"status": "up"}'`

2. The following **3** messages are published by the mapper (local registration message, cloud service registration message, service health status cloud update message):
   ```
   [te/device/main/service/my-custom-service] {"@id":"marcel_dev_container:device:main:service:my-custom-service","@parent":"device/main//","@type":"service","name":"my-custom-service","type":"service"}
   [c8y/s/us] 102,marcel_dev_container:device:main:service:my-custom-service,service,my-custom-service,up
   [c8y/s/us/marcel_dev_container:device:main:service:my-custom-service] 104,up
   ```

3. Publish a health status message for an entity with a topic id that **does** contain `bridge` but is not an actual mosquitto bridge notification topic:
   `tedge mqtt pub te/device/main/service/my-custom-bridge/status/health '{"status": "up"}'`

4. The following **2** message are published by the mapper (just local registration and cloud service creation):
   ```
   [te/device/main/service/my-custom-bridge] {"@id":"marcel_dev_container:device:main:service:my-custom-bridge","@parent":"device/main//","@type":"service","name":"my-custom-bridge","type":"service"}
   [c8y/s/us] 102,marcel_dev_container:device:main:service:my-custom-bridge,service,my-custom-bridge,up
   ```
5. Successive health status messages for this entity are not mapped (no 104 smartrest messages are published by the mapper):
   ```
   $ tedge mqtt pub te/device/main/service/my-custom-bridge/status/health '{"status": "down"}
   [te/device/main/service/my-custom-bridge/status/health] {"status": "down"}
   $ tedge mqtt pub te/device/main/service/my-custom-bridge/status/health '{"status": "up"}'
   [te/device/main/service/my-custom-bridge/status/health] {"status": "up"}
   ```

**Expected behavior**

Health status should be correctly mapped for services which contain `bridge` in their topic ids.

**Additional context**

[...]

PR #2837, opened 2024-04-23
fix(#2370): use bridge topic id in service_monitor

Bravo555:
## Proposed changes

Use the correct bridge topic id when converting health status message.

## Types of changes

## Paste Link to the issue

- #2370

## Checklist


crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -28,7 +28,8 @@ pub fn convert_health_status_message(
         return vec![];
     }
 
-    if entity.topic_id.is_bridge_health_topic() {
+    // the bridge itself is not registered as a service, only the mapper
+    if entity.topic_id == crate::C8Y_BRIDGE_TOPIC_ID {
```
didier-wenzek, line 32: With this change, the `az` bridge is reported as a service on `c8y`. This could make sense, but is this what we expect?

crates/extensions/c8y_mapper_ext/src/lib.rs
```diff
@@ -14,6 +14,8 @@ pub mod service_monitor;
 #[cfg(test)]
 mod tests;
 
+pub(crate) const C8Y_BRIDGE_TOPIC_ID: &str = "device/main/service/mosquitto-c8y-bridge";
```
albinsuresh, line 17: We can't assume this hardcoded `device/main` topic id even for the main device since we support overriding `device_topic_id` via the config. That being said, I'm not fully sure if all tedge components properly respect that config setting yet either. I'm fairly sure that it is still hardcoded in many places.

Issue #2836, opened 2024-04-23
Empty workflow log files generated for config and log management operations


albinsuresh:
**Describe the bug**

The workflow log files created on execution of configuration and log management operations are empty as seen here:

```
$ ls -l /var/log/tedge/agent/
total 24
-rw-r--r-- 1 tedge tedge    0 Apr 23 13:44 workflow-config_snapshot-c8y-mapper-35841214.log
-rw-r--r-- 1 tedge tedge    0 Apr 23 13:32 workflow-log_upload-c8y-mapper-35588951.log
```

**To Reproduce**

Execute any config management or log management operation and check the corresponding workflow log files at `/var/log/tedge/agent/`

**Expected behavior**

The workflow log files should properly capture the state transitions of these operations, just like it does for software management or device restart operations

**Additional context**

This is a side effect of the config management and log management actors being integrated with `tedge-agent` as independent actors and not via the `tedge_operation_converter` as it is done for software manager and restart manager actors.

didier-wenzek:
Indeed, something to do. However, this is unfortunately far beyond just "fix logging".

- The config and log management operations are currently not managed by the workflow sub-system.
- Till this is fixed:
  - Users cannot provide enhanced version of the log and config operations
  - A workflow cannot trigger a log or config update as a sub operation
  - The latter point makes this issue *high priority* to support device profile. 
- The fact the logs are empty is actually a bug:
  - the log file is created on reception of an MQTT message for a command, 
  - but never populated as the workflow sub-system is not even aware of those operations.

didier-wenzek:
Thanks to https://github.com/thin-edge/thin-edge.io/pull/2844, the log and config operations are now plainly supported by the workflow feature:

- Users can provide enhanced version of the log and config operations
- A user-defined workflow can trigger a log or config update as a sub operation
- Workflow log files are generated for log and config commands as for the other commands.

gligorisaev:
This bug was reproducable before the PR's that are fixing it were merged.
I confirm that this bug is NOT reproducable anymore.

Review done for following tests:
`tests/RobotFramework/tests/cumulocity/software_management/software.robot`
`tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot`
`tests/RobotFramework/tests/cumulocity/log/log_operation.robot`

#### software.robot

In this Test Suite following Test Steps are added:

`Operation log uploaded automatically with default auto_log_upload setting as onFailure`
[...]

PR #2844, opened 2024-04-26
Feat: Config and log management workflows

didier-wenzek:
## Proposed changes

Update the log and config operations so those can be used by the workflow system,
enabling the ability to trigger a log or config update as a sub operation.

- [x] Make log and config message types similar to those used for the workflows
- [ ] Make the LogManagerActor acting on LogUploadCmd instead of MqttMessage
- [ ] Make the ConfigManagerActor acting on ConfigSnapshotCmd  & ConfigUpdateCmd instead of MqttMessage
- [ ] Enable log operation workflows
- [ ] Enable config operation workflows

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2836

## Checklist


crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -103,73 +105,55 @@ impl ConfigManagerActor {
...
-        match ConfigOperation::request_from_message(&self.config, &message) {
-            Ok(Some(ConfigOperation::Snapshot(request))) => match request.status {
-                CommandStatus::Init => {
+    async fn process_operation_request(
+        &mut self,
+        request: ConfigOperation,
+    ) -> Result<(), ChannelError> {
+        match request {
+            ConfigOperation::Snapshot(topic, request) => match request.status {
+                CommandStatus::Init | CommandStatus::Scheduled => {
```
didier-wenzek, line 120: A command has to be executed either on `init` or the `scheduled` state.

- Init is the trigger when the actor is directly exposed to MQTT (in the unit tests notably).
- When the actor is running behind the workflow actor, the trigger is the scheduled state.

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -79,87 +92,70 @@ impl LogManagerActor {
...
-                Ok(None) => {}
-                Err(err) => {
-                    error!("Incorrect log request payload: {}", err);
-                }
+    pub async fn process_logfile_request(
+        &mut self,
+        request: LogUploadCmd,
+    ) -> Result<(), ChannelError> {
+        match request.status() {
+            CommandStatus::Init | CommandStatus::Scheduled => {
```
didier-wenzek, line 112: A command has to be executed either on `init` or the `scheduled` state.

- Init is the trigger when the actor is directly exposed to MQTT (in the unit tests notably).
- When the actor is running behind the workflow actor, the trigger is the scheduled state.

crates/extensions/tedge_config_manager/src/lib.rs
```diff
@@ -147,27 +155,111 @@ impl ConfigManagerBuilder {
...
+                        operation,
+                        payload: json!( {
+                            "types": types
+                        }),
+                    }))
+                }
+            },
+        )
+    }
+}
```
didier-wenzek, line 226: I not happy with this connection logic.

This is unfortunate but each operation actor is using is own representation for command metadata and states. Here Payload are used. The log actor is using Command<Payload>. I will try to make this more consistent once the main feature properly working.

albinsuresh (commented):
A great step forward. Have sought some minor clarifications though.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -511,61 +496,80 @@ impl ConfigManagerActor {
...
+        self.output_sender.send(state).await
     }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum ConfigOperation {
-    Snapshot(ConfigSnapshotCmdPayload),
-    Update(ConfigUpdateCmdPayload),
+    Snapshot(Topic, ConfigSnapshotCmdPayload),
+    Update(Topic, ConfigUpdateCmdPayload),
```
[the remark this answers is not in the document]
didier-wenzek: What you propose is the way to go and I started by implementing this change. However, this proved to be more involved than expected. Not difficult, but touching most of the logic of configuration manager. Passing a topic along a payload is what was already done by all configuration related methods.

I will add this change as a follow-up task.

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -79,87 +92,70 @@ impl LogManagerActor {
...
             let error_message = format!("Failed to initiate log file upload: {error}");
             request.failed(&error_message);
-            self.publish_command_status(topic, &request).await?;
+            self.publish_command_status(request).await?;
             error!("{}", error_message);
             return Ok(());
         }
 
-        self.pending_operations.insert(topic.name.clone(), request);
+        let topic = request.topic(&self.config.mqtt_schema).as_ref().to_string();
```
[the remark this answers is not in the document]
didier-wenzek: Good idea. Something I will do in a follow up PR.

crates/extensions/tedge_config_manager/src/lib.rs
```diff
@@ -147,27 +155,111 @@ impl ConfigManagerBuilder {
...
+
+impl IntoIterator for &ConfigManagerBuilder {
```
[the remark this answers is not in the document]
didier-wenzek: Agree that this is not fully satisfying. I considered a specific trait to connect actors that handle operations; but failed to find the right place for this definition. Should this be `tedge_actors`, `tedge_api,` `tedge-agent`?

- Technically this is only used by actors designed to be connected to the agent, but having a dependency from an operation manager to the agent sounds wrong.
- Can this be `tedge_actors`? No, as this is too thin-edge specific.
- `tedge_api` could be a fit, but would introduce a new dependency on `tedge_actors`.

I think the better would to find the right abstraction similar to what has been done the configuration of message sources and to have it defined by  `tedge_actors`. Hence, this first try with concrete examples.

didier-wenzek:
I finally understood why all the system tests related to config update are failing.
The `ConfigManagerActor` is not autonomous: one of the step of a config update command is handled by the `FileCacheActor`.
The interactions between these two actors are done over MQTT, bypassing the workflow actor that uses in-memory channels.

1. On reception of a config update command, the `ConfigManagerActor` move this request to the executing state.
2. The `FileCacheActor` reacts to config update commands in the executing state, downloads the remote file and makes it locally available.
3. When the file is ready for local download, the `FileCacheActor` re-publishes the config update state message with an extra `tedgeUrl` field.
4. The `ConfigManagerActor` then proceeds using the `tedgeUrl` field instead of the former `remoteUrl`.

Three points in the current implementation prevent the workflow system to handle config update operations:
1. When the `FileCacheActor` re-publishes the config update state, the status is *unchanged* leading to two *executing* messages in a row.
   - The first *executing* message is published by `ConfigManagerActor`
   - The second is published by the `FileCacheActor` with an extra `tedgeUrl` field.
   - The `ConfigManagerActor` is aware of that trick
     and [ignores the first message](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/tedge_config_manager/src/actor.rs#L362).
   - But the actor workflow cannot behave differently on these two messages as the action is solely determined by the command status.
   - => One needs to introduce a specific state for a config update command which config file has been locally downloaded.
2. Currently, the workflow system assumes that a *single* actor handles all the built-in steps of an operation.
   - This means that the state changes sent by the built-in actor over an in-memory channel are published over MQTT by the workflow system,
     but the state changes notified over MQTT are not pushed to the built-in actor
     (that is supposed to be the source and not a consumer of these messages).
   - This is what breaks all the system tests related to config update:
     *the messages with the extra `tedgeUrl` field are never transmitted to the `ConfigManagerActor`*.
3. Using MQTT for the interactions between the `ConfigManagerActor` and the `FileCacheActor` is mandatory.
   - Indeed, the `ConfigManagerActor` might be running on a child device, while the `FileCacheActor` is running on the main device.
   - => One can improve the workflow system with an action telling to await for an update over MQTT.
   - Here:
     1. the `ConfigManagerActor` moves the command to *executing* with an action `wait-mqtt-update`,
     2. the `FileCacheActor` downloads the file and moves the command to *downloaded*,
     3. the workflow actor (being in a `wait-mqtt-update` for that command) processes the `FileCacheActor`s response.
     4. and the action for the *downloaded* state being configured to *built-in*,
        the workflow actor forward the `FileCacheActor`s response to the `ConfigManagerActor`.

That said, I wonder to have a `FileCacheActor` that is too config-update specific. I will take some time to explore alternative solutions.

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -108,6 +108,30 @@ Manual config_snapshot operation request
...
+    ${snapshot}    Execute Command    curl http://${PARENT_IP}:8000/tedge/file-transfer/${PARENT_SN}/sub_config_snapshot/local-1111
+    ${config}      Get File    ${CURDIR}/tedge-configuration-plugin.toml
+    Should Be Equal    ${snapshot}     ${config}
+
+Trigger custom config_snapshot operation
+    [Teardown]    Restore config operations
+    Customize config operations
+    Set Device Context    ${PARENT_SN}
+    Publish and Verify Local Command
+    ...    topic=te/device/main///cmd/sub_config_snapshot/local-1111
```
albinsuresh, line 127:
```suggestion
    ...    topic=te/device/main///cmd/config_snapshot/local-1111
```
Right?

albinsuresh (commented):
Everything looks good. Happy to approve once the test failures are resolved.

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -227,6 +226,9 @@ impl OperationWorkflow {
                         .with_qos(QoS::AtLeastOnce),
                 )
             }
+            // Builtin operations dynamically publish their capability message,
+            // notably to include a list of supported types.
```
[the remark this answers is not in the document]
didier-wenzek: > Another contract that could be captured by an abstraction/trait that we introduce in future for the operation actors that that are connected to the agent.

Good point

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
didier-wenzek: You are correct

didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/2844/commits/2457d5bec9a6eb0441ebc97f9279fa010ea2b06b

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -134,6 +158,34 @@ Manual config_update operation request
...
 
+Trigger config_update operation from another workflow
+    Set Device Context    ${PARENT_SN}
+
+    Execute Command     curl -X PUT --data-binary 'new content for CONFIG1' "http://${PARENT_IP}:8000/tedge/file-transfer/${PARENT_SN}/sub_config_update/sub-2222"
+    Publish and Verify Local Command
+    ...    topic=te/device/main///cmd/sub_config_update/sub-2222
+    ...    payload={"status":"init","tedgeUrl":"http://${PARENT_IP}:8000/tedge/file-transfer/${PARENT_SN}/sub_config_update/sub-2222","remoteUrl":"","type":"CONFIG1"}
+    ...    expected_status=successful
+    ...    c8y_fragment=c8y_UploadConfigFile
```
albinsuresh, line 169:
```suggestion
    ...    c8y_fragment=c8y_DownloadConfigFile
```
Since the test is validating a negation, it was passing, but I'm assuming that you really meant this.

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -134,6 +158,34 @@ Manual config_update operation request
...
+    [Teardown]    Restore config operations
+    Set Device Context    ${PARENT_SN}
+    Customize config operations
+
+    Execute Command     curl -X PUT --data-binary 'updated config' "http://${PARENT_IP}:8000/tedge/file-transfer/${PARENT_SN}/config_update/custom-2222"
+    Publish and Verify Local Command
+    ...    topic=te/device/main///cmd/config_update/custom-2222
+    ...    payload={"status":"init","tedgeUrl":"http://${PARENT_IP}:8000/tedge/file-transfer/${PARENT_SN}/config_update/custom-2222","remoteUrl":"","type":"/tmp/config_update_target"}
+    ...    expected_status=successful
+    ...    c8y_fragment=c8y_UploadConfigFile
```
albinsuresh, line 184:
```suggestion
    ...    c8y_fragment=c8y_DownloadConfigFile
```
Same comment as above.

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/2844/commits/ca0f4cde902c638786c6f57a25b7df6781030a1f

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/2844/commits/ca0f4cde902c638786c6f57a25b7df6781030a1f

albinsuresh (approved):
Approved.

reubenmiller (approved):
Approved.

Great job (especially for the sustained effort!)

tests/RobotFramework/tests/cumulocity/log/sub_log_upload.toml
```diff
@@ -0,0 +1,26 @@
...
+on_success = "executing"
+
+[executing]
+operation = "log_upload"
+input.tedgeUrl = "${.payload.tedgeUrl}"
+input.type = "${.payload.type}"
+input.dateFrom = "${.payload.dateFrom}"
+input.dateTo = "${.payload.dateTo}"
+input.searchText = "${.payload.searchText}"
+input.lines = "${.payload.lines}"
```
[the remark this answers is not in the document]
didier-wenzek: Yes, this has still to be implemented along other improvements such as full [string interpolation](https://github.com/thin-edge/thin-edge.io/issues/2847).

Issue #2848, opened 2024-04-30
tedge-apt-plugin should support exclude patterns based on name and maintainer


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

It is difficult to filter out uninteresting apt packages from the software list as the current filters only support known names. This makes it difficult to filter out packages such as `libc.*`, `iptables-module-.*`, `kernel-.*` (packages which are common in Yocto builds when including the apt package manager). In Yocto builds this means that from the ~2000 packages which part of the base image, that only 200 packages are interesting for the user (e.g. not kernel modules, iptables module etc.)

Below shows the currently support apt package filters (which are only "include" filters).

```toml
[apt]
name = ".+"
maintainer = ".+"
```

The Rust standard regex create does not support Regex negative look behinds, so it makes virtually impossible to filter out known packages.

**Describe the solution you'd like**

Extend the `tedge-apt-plugin` filtering possibility to also support exclude patterns for both the package name and maintainer.

```toml
[apt]
exclude.name = ""
exclude.maintainer = ""
```

The additional settings should behave as follows:

* The new properties would also support regular expressions
* If `apt.exclude.name` is empty, then the filter will be ignored
* If `apt.exclude.maintainer` is empty, then the filter will be ignored
* The `apt.include.*` patterns take precedence over the exclude logic, meaning that if the include patterns match, then the package SHOULD appear in the software list.

**Describe alternatives you've considered**

**Additional context**

Ruadhri17:
This logic was moved from `apt` plugin to `sm-plugin` module. Two new configs were added: `software.plugin.include` and `software.plugin.exclude`. Following rules are applied to filtering depending on what option is used:

- if only `software.plugin.include` is provided, keep a package if its name matches
- if only `software.plugin.exclude` is provided, keep a package if its name doesn't match
- if both are provided, the logic is to keep a package if its name matches the `software.plugin.include` regex or doesn't match the `software.plugin.exclude`. 

New robot framework tests were added [here](https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/software_management/sm-plugin.robot).

New doc entry exists [here](https://github.com/thin-edge/thin-edge.io/blob/main/docs/src/references/agent/software-management.md).

gligorisaev:
QA has thoroughly checked the feature and here are the results:

- [x] Test Suite has been reviewed.
[...]

PR #2869, opened 2024-05-09
feat(sm-plugin): filter packages output list with exclude/include pattern

Ruadhri17:
## Proposed changes
This PR introduce two new tedge configs: `software.plugin.include` and `software.plugin.exclude`. This allows to filter packages list output in `sm-plugin` using include pattern and exclude pattern.  When both patterns are provided in config, include pattern takes precedence over exclude pattern. 

## Types of changes

## Paste Link to the issue
#2848 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -745,7 +745,15 @@ define_tedge_config! {
 
             /// The maximum number of software packages reported for each type of software package
             #[tedge_config(example = "1000", default(value = 1000u32))]
-            max_packages: u32
+            max_packages: u32,
+
+            /// The filtering criterion, in form of regex, that is used to filter packages list output
+            #[tedge_config(example = "^(tedge|c8y).*")]
+            include: String,
+
+            /// The filtering criterion, in form of regex, that is used to filter out packages from the output list
+            #[tedge_config(example = "^(glibc|lib|kernel-|iptables-module).*")]
+            exclude: String,
         }
     },
 
@@ -1134,6 +1142,8 @@ mod tests {
     #[test_case::test_case("mqtt.external.certfile")]
     #[test_case::test_case("mqtt.external.keyfile")]
     #[test_case::test_case("software.plugin.default")]
+    #[test_case::test_case("software.plugin.exclude")]
+    #[test_case::test_case("software.plugin.include")]
     #[test_case::test_case("software.plugin.max_packages")]
     #[test_case::test_case("tmp.path")]
     #[test_case::test_case("logs.path")]
--- a/crates/core/plugin_sm/src/plugin.rs
+++ b/crates/core/plugin_sm/src/plugin.rs
@@ -2,6 +2,7 @@ use async_trait::async_trait;
 use csv::ReaderBuilder;
 use download::Downloader;
 use logged_command::LoggedCommand;
+use regex::Regex;
 use reqwest::Identity;
 use serde::Deserialize;
 use std::error::Error;
@@ -239,6 +240,8 @@ pub struct ExternalPluginCommand {
     pub path: PathBuf,
     pub sudo: SudoCommandBuilder,
     pub max_packages: u32,
+    exclude: Option<String>,
+    include: Option<String>,
     identity: Option<Identity>,
 }
 
@@ -248,6 +251,8 @@ impl ExternalPluginCommand {
         path: impl Into<PathBuf>,
         sudo: SudoCommandBuilder,
         max_packages: u32,
+        exclude: Option<String>,
+        include: Option<String>,
         identity: Option<Identity>,
     ) -> ExternalPluginCommand {
         ExternalPluginCommand {
@@ -255,6 +260,8 @@ impl ExternalPluginCommand {
             path: path.into(),
             sudo,
             max_packages,
+            exclude,
+            include,
             identity,
         }
     }
@@ -461,13 +468,37 @@ impl Plugin for ExternalPluginCommand {
         let command = self.command(LIST, None)?;
         let output = self.execute(command, logger).await?;
         if output.status.success() {
+            let filtered_output = match (&self.exclude, &self.include) {
+                (None, None) => output.stdout,
+                _ => {
+                    // If no exclude pattern is given, exclude everything (except what matches the include pattern)
+                    let exclude_filter =
+                        Regex::new(self.exclude.as_ref().unwrap_or(&r".*".to_string()))?;
+                    // If no include pattern is given, include nothing (except what doesn't match the exclude pattern)
+                    let include_filter =
+                        Regex::new(self.include.as_ref().unwrap_or(&r"^$".to_string()))?;
+
+                    output
+                        .stdout
+                        .split_inclusive(|c| *c == b'\n')
+                        .filter_map(|line| std::str::from_utf8(line).ok())
+                        .filter(|line| {
+                            line.split_once('\t').is_some_and(|(name, _)| {
+                                include_filter.is_match(name) || !exclude_filter.is_match(name)
+                            })
+                        })
+                        .flat_map(|line| line.as_bytes().to_vec())
+                        .collect()
+                }
+            };
+
             // If max_packages is set to an invalid value, use 0 which represents all
             // of the content, don't bother filtering the content when all of it will
             // be included anyway
             let max_packages = usize::try_from(self.max_packages).unwrap_or(0);
             let last_char = match max_packages {
                 0 => 0,
-                _ => String::from_utf8(output.stdout.as_slice().to_vec())
+                _ => String::from_utf8(filtered_output.as_slice().to_vec())
                     .unwrap_or_default()
                     .char_indices()
                     .filter(|(_, c)| *c == '\n')
@@ -479,8 +510,8 @@ impl Plugin for ExternalPluginCommand {
             Ok(deserialize_module_info(
                 self.name.clone(),
                 match last_char {
-                    0 => &output.stdout[..],
-                    _ => &output.stdout[..=last_char],
+                    0 => &filtered_output[..],
+                    _ => &filtered_output[..=last_char],
                 },
             )?)
         } else {
--- a/tests/RobotFramework/tests/cumulocity/software_management/sm-plugin.robot
+++ b/tests/RobotFramework/tests/cumulocity/software_management/sm-plugin.robot
@@ -74,6 +74,51 @@ sm-plugins download files from Cumulocity
     ${downloaded}=    Execute Command    cat /tmp/dummy3/installed_dummy-software
     Should Be Equal    ${downloaded}    Testing a thing
 
+Filter packages list using include pattern
+    Execute Command    sudo tedge config set software.plugin.include "^dummy1-[0-1]00[2-4]$"
+    Execute Command    sudo tedge config set software.plugin.max_packages 0
+    Connect Mapper    c8y
+    Device Should Exist    ${DEVICE_SN}
+    ${software}=    Device Should Have Installed Software
+    ...    {"name": "dummy1-0002", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-0003", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-0004", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-1002", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-1003", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-1004", "version": "1.0.0", "softwareType": "dummy1"}
+    Length Should Be    ${software}    6
+
+Filter packages list using exclude pattern
+    Execute Command    sudo tedge config set software.plugin.exclude "^dummy[1-2]-\\d+(0|2|4|6|8)$"
+    Execute Command    sudo tedge config set software.plugin.max_packages 0
+    Connect Mapper    c8y
+    Device Should Exist    ${DEVICE_SN}
+    ${software}=    Device Should Have Installed Software
+    ...    {"name": "dummy1-0001", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-0003", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-0005", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-1499", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy2-0001", "version": "1.0.0", "softwareType": "dummy2"}
+    ...    {"name": "dummy2-0003", "version": "1.0.0", "softwareType": "dummy2"}
+    ...    {"name": "dummy2-0005", "version": "1.0.0", "softwareType": "dummy2"}
+    ...    {"name": "dummy2-1499", "version": "1.0.0", "softwareType": "dummy2"}
+    Length Should Be    ${software}    1500
+
+Filter packages list using both patterns
+    Execute Command    sudo tedge config set software.plugin.exclude "^(dummy1.*)"
+    Execute Command    sudo tedge config set software.plugin.include "^(dummy1-\\d(|3|6|9)00)$"
+    Execute Command    sudo tedge config set software.plugin.max_packages 0
+    Connect Mapper    c8y
+    Device Should Exist    ${DEVICE_SN}
+    ${software}=    Device Should Have Installed Software
+    ...    {"name": "dummy1-0300", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-0600", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-0900", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy1-1300", "version": "1.0.0", "softwareType": "dummy1"}
+    ...    {"name": "dummy2-0001", "version": "1.0.0", "softwareType": "dummy2"}
+    ...    {"name": "dummy2-1500", "version": "1.0.0", "softwareType": "dummy2"}
+    Length Should Be    ${software}    1504
+
 *** Keywords ***
 Custom Setup
     ${DEVICE_SN}=    Setup    skip_bootstrap=${True}
--- a/crates/core/plugin_sm/src/plugin_manager.rs
+++ b/crates/core/plugin_sm/src/plugin_manager.rs
@@ -207,6 +207,8 @@ impl ExternalPlugins {
                             &path,
                             self.sudo.clone(),
                             config.software.plugin.max_packages,
+                            config.software.plugin.exclude.or_none().cloned(),
+                            config.software.plugin.include.or_none().cloned(),
                             identity,
                         );
                         self.plugin_map.insert(plugin_name.into(), plugin);
--- a/crates/core/plugin_sm/tests/plugin.rs
+++ b/crates/core/plugin_sm/tests/plugin.rs
@@ -67,6 +67,8 @@ mod tests {
             &dummy_plugin_path,
             SudoCommandBuilder::enabled(false),
             config.software.plugin.max_packages,
+            None,
+            None,
             config.http.client.auth.identity()?,
         );
         assert_eq!(plugin.name, "test");
@@ -85,6 +87,8 @@ mod tests {
             dummy_plugin_path,
             SudoCommandBuilder::enabled(false),
             100,
+            None,
+            None,
             None,
         );
 
@@ -115,6 +119,8 @@ mod tests {
             dummy_plugin_path,
             SudoCommandBuilder::enabled(false),
             100,
+            None,
+            None,
             None,
         );
 
@@ -151,6 +157,8 @@ mod tests {
             dummy_plugin_path,
             SudoCommandBuilder::enabled(false),
             100,
+            None,
+            None,
             None,
         );
 
--- a/crates/core/tedge_api/Cargo.toml
+++ b/crates/core/tedge_api/Cargo.toml
@@ -16,6 +16,7 @@ download = { workspace = true }
 json-writer = { workspace = true }
 log = { workspace = true }
 mqtt_channel = { workspace = true }
+regex = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 shell-words = { workspace = true }
@@ -35,7 +36,6 @@ assert_matches = { workspace = true }
 clock = { workspace = true }
 maplit = { workspace = true }
 mockall = { workspace = true }
-regex = { workspace = true }
 tempfile = { workspace = true }
 test-case = { workspace = true }
 time = { workspace = true, features = ["macros"] }
--- a/crates/core/tedge_api/src/error.rs
+++ b/crates/core/tedge_api/src/error.rs
@@ -100,6 +100,9 @@ pub enum SoftwareError {
 
     #[error("CSV error: {reason:?}")]
     FromCSV { reason: String },
+
+    #[error("Regex error: {reason:?}")]
+    RegexError { reason: String },
 }
 
 impl From<serde_json::Error> for SoftwareError {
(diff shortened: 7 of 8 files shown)


reubenmiller (changes requested):
We'll have to do the settings in a non breaking way as the old apt.name and apt.maintainer (inclusive) filters still need to be accepted (otherwise it will break everyone's existing config, and it will be in violation of the semver versioning).

We can introduce new settings but we need to then map the old settings to the new ones.

Also I'm not sure I fully understand the decision around the tedge-apt-plugin flag choice, as it stands the user can't use inclusive and exclusive filters via the cli...I would have expected something like "--inclusive-name" and "--exclusive-name" flags. The flag renaming is less important here as this was only ever used for adhoc validation an shouldn't break anyone's existing setup

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -808,13 +808,6 @@ define_tedge_config! {
     },
 
     apt: {
-        /// The filtering criterion that is used to filter packages list output by name
-        #[tedge_config(example = "tedge.*")]
-        name: String,
-        /// The filtering criterion that is used to filter packages list output by maintainer
-        #[tedge_config(example = "thin-edge.io team.*")]
-        maintainer: String,
-
```
didier-wenzek, line 817: As noted by @reubenmiller:

> The old apt.name and apt.maintainer (inclusive) filters still need to be accepted

For that purpose, the `tedge_config` macro features `deprecated_key`s.

See for example, how [`mqtt.external.port` has been deprecated in favor of `mqtt.external.bind.port`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs#L622)

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -825,6 +818,24 @@ define_tedge_config! {
...
+            #[tedge_config(example = "tedge.*")]
+            name: String,
+            /// The filtering criterion that is used to filter out packages from the output list by maintainer
+            #[tedge_config(example = "thin-edge.io team.*")]
+            maintainer: String,
+        },
+
+        include: {
+            /// The filtering criterion that is used to filter packages list output by name
+            #[tedge_config(example = "tedge.*")]
```
didier-wenzek, line 833:
```suggestion
            #[tedge_config(example = "tedge.*", deprecated_key = "apt.name")]
```

plugins/tedge_apt_plugin/src/lib.rs
```diff
@@ -34,14 +34,24 @@ pub struct AptCli {
...
         maintainer: Option<String>,
+
+        /// Filtering mode
+        #[clap(
+            long = "filter-mode",
+            requires = "filter_target",
+            default_value_t,
+            value_enum
+        )]
+        filter_mode: FilterMode,
```
didier-wenzek, line 54: I don't understand the purpose of this `--filter-mode` option. Aren't `apt.include.name` and `apt.exclude.name` enough do define filtering expectations?

reubenmiller: Yes this is the same point I raised in my review. There is no point having the `--filter-mode` flag. The config values should just be mapped to flags, this would allow the user to test both inclusive and exclusive filters at the same time (which is closer to what actual behaviour when run by the tedge-agent).

gligorisaev (commented):
Test 
- filter_packages_list_output_with_exclude_pattern.robot
Appears to be well-structured, covering a range of scenarios, and designed with maintainability and reliability in mind. Provides solid foundation for testing package filtering functionality
- tests/RobotFramework/tests/plugin_apt/filter_packages_list_output_with_include_pattern.robot
Changes should be acurate enought to check the functionality properly 
- tests/RobotFramework/tests/tedge/call_tedge_config_list.robot
Effectively tests the setting and unsetting of configuration values for apt name and maintainer.

plugins/tedge_apt_plugin/src/lib.rs
```diff
@@ -103,29 +117,80 @@ fn run_op(apt: AptCli) -> Result<ExitStatus, InternalError> {
...
+                }
+                (Some(inclusive_filter), Some(exclusive_filter)) => {
+                    for line in stdout.trim_end().lines() {
+                        if installed_filter.is_match(line)
+                            && (inclusive_filter.is_match(line) || !exclusive_filter.is_match(line))
+                        {
+                            let (name, version) = get_name_and_version(line);
+                            println!("{name}\t{version}");
+                        }
+                    }
```
didier-wenzek, line 185: An alternative to avoid to repeat ourselves for the 4 cases (with and without include or exclude pattern), is to take as a default

- a pattern that always matches for the include regex (i.e. `".*"`)
- a pattern that never matches for the exclude regex (i.e. `"^$"`)

plugins/tedge_apt_plugin/src/lib.rs
```diff
@@ -103,29 +117,67 @@ fn run_op(apt: AptCli) -> Result<ExitStatus, InternalError> {
...
+                        .unwrap_or_else(|| Regex::new(r"install ok installed").unwrap());
+                    let exclude_filter =
+                        exclude_filter.unwrap_or_else(|| Regex::new(r"^$").unwrap());
+                    for line in stdout.trim_end().lines() {
+                        if include_filter.is_match(line) && !exclude_filter.is_match(line) {
+                            let (name, version) = get_name_and_version(line);
+                            println!("{name}\t{version}");
+                        }
+                    }
                 }
```
didier-wenzek, line 181: What if an *include* filter is given but no *exclude* filter? Are only the *installed* packages correctly listed?
Removing case matching would really help to be confident.

```rust
let installed_filter = Regex::new(r"install ok installed").unwrap();
let include_filter = include_filter.unwrap_or(Regex::new(r".*").unwrap());
let exclude_filter = exclude_filterunwrap_or(Regex::new(r^$").unwrap());
for line in stdout.trim_end().lines() {
       if installed_filter.is_match(line)
             && (include_filter.is_match(line) || !exclude_filter.is_match(line))
       {
               let (name, version) = get_name_and_version(line);
               println!("{name}\t{version}");
        }
}
```

If you dislike this suggestion, I would prefer the very first proposal you published: i.e. with the 4 cases being explicit. At least, it's easy to check correctness, while this is not the case with this intermediate solution.

Ruadhri17: I still think it is still better than four cases. I forgot that check for installed packages should also be added in include_filter: 
```rust
Regex::new(format!(r"(^{}\t.*|^\S+\t\S+\t{}\s+.*installed ok install)", include_name.unwrap_or_default(), include_maintainer.unwrap_or_default())
```
Anyway, I like the proposed solution more, it is indeed easier to verify.

plugins/tedge_apt_plugin/src/lib.rs
```diff
@@ -103,29 +117,70 @@ fn run_op(apt: AptCli) -> Result<ExitStatus, InternalError> {
...
+                    for line in stdout.trim_end().lines() {
+                        if installed_filter.is_match(line)
+                            && (include_filter.is_match(line) && !exclude_filter.is_match(line))
+                        {
+                            let (name, version) = get_name_and_version(line);
+                            println!("{name}\t{version}");
+                        }
+                    }
                 }
             }
```
didier-wenzek, line 185: The `match` is useless as the `_` case simply unwraps the given filters.

```suggestion
                    let include_filter =
                        include_filter.unwrap_or_else(|| Regex::new(r".*").unwrap());
                    let exclude_filter =
                        exclude_filter.unwrap_or_else(|| Regex::new(r"^$").unwrap());
                    for line in stdout.trim_end().lines() {
                        if installed_filter.is_match(line)
                            && (include_filter.is_match(line) && !exclude_filter.is_match(line))
                        {
                            let (name, version) = get_name_and_version(line);
                            println!("{name}\t{version}");
                        }
                    }
```

didier-wenzek (approved):
Approved.

reubenmiller (commented):
I'll remove by blocker (as it has been addressed) and have another look at it tomorrow

plugins/tedge_apt_plugin/src/lib.rs
didier-wenzek: > a pattern that always matches for the include regex (i.e. `".*"`)

I was wrong. This would defeat the purpose of the exclude filter when combined with the requirement that the include filter takes precedence over the exclude filter.

plugins/tedge_apt_plugin/src/lib.rs
didier-wenzek: > The `match` is useless as the `_` case simply unwraps the given filters.

Again I was wrong. I missed the use of a disjonction (`||`) vs a conjonction (`&&`).

didier-wenzek: That said, is this really what we expect?

##  Current behavior of this PR

With `include_filter.is_match(line) || !exclude_filter.is_match(line)`, a setting that includes `.*python.*` and excludes `.*lib*.`, would, as expected, emits `python` and `python-httplib` and rejects `libcurl`. But it would also emit `curl` and anything that doesn't match the exclude filter.

## How I understand the priority requirement

To meet the requirement that the `apt.include.*` patterns take precedence over the exclude logic, a better solution would be to ignore all the  `apt.exclude.*` patterns when the user provides an `apt.include.*` pattern. With `include_filter.is_match(line) `, a setting that includes `.*python.*` and excludes `.*lib*.`, would emit `python` as well as `python-httplib`, rejecting `curl` and `libcurl`.

## What I would prefer

I do think the better is to always use the two filters provided by the user with `include_filter.is_match(line) && !exclude_filter.is_match(line)`. A setting that includes `.*python.*` and excludes `.*lib*.`, would emit `python` and reject `python-httplib` as well as `curl` and `libcurl`. To list both  `python` and `python-httplib`, the user has to provide only an include pattern for `.*python.*`. This is what is suggested by https://github.com/thin-edge/thin-edge.io/pull/2869#discussion_r1600560753

reubenmiller: The main use-case would reflect a real-world scenario:

1. Create a filter to exclude all "lib" values
2. Someone then finds that there is an exception to the "generic exclusion filter", where `lib-curl` needs to be included.

How would this be made possible?

reubenmiller: I see the following cases:

### Case 1: Explicit exclude pattern

```
NOT (NAME MATCHES EXCLUDE_PATTERN)
```

### Case 2: Include with exclusion exceptions

```
NAME MATCHES INCLUDE_PATTERN
AND NOT (NAME MATCHES EXCLUDE_PATTERN)

```

### Case 3: Exclude with include exceptions

```
NAME MATCHES INCLUDE_PATTERN
OR NOT (NAME MATCHES EXCLUDE_PATTERN
```

didier-wenzek:
After discussion with @reubenmiller, we concluded that:

- the current behavior of this PR is what we expect (i.e. list a package if explicitly included *or* not explicitly excluded)
- __except this should be done for all the software plugins and not just for apt__
- hence the filters apply only to the package names and not to the maintainers (the latter being only available at the apt plugin level).

Hence:
- the current work as to be reverted as no changes are expected for `tedge-apt-plugin` :-(
- two new `tedge config` setting have to be added: `software.plugin.include` and  `software.plugin.exclude`
- these two optional regex patterns are to be used by [ExternalPluginCommand::list()](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/plugin_sm/src/plugin.rs#L457)
- the patterns apply to the package names
- if both are provided, the logic is to keep a package if its name matches the `software.plugin.include` regex *or* doesn't match the `software.plugin.exclude`. The logic is that the primary usage will be to *exclude* packages (say all libs and kernel modules) and only then add some exceptions (say to include the `libc`).
- if only `software.plugin.include` is provided, keep a package if its name matches
- if only `software.plugin.exclude` is provided, keep a package if its name doesn't match
- by the way, @Ruadhri17 , you were correct: it might be simpler to keep separate each case
- finally this has to combined with the current logic for keeping at most ` software.plugin.max_packages` packages.

gligorisaev (commented):
Tests:
1. tests/RobotFramework/tests/plugin_apt/filter_packages_list_output_with_exclude_pattern.robot
	- 	The test suite is well-structured with clear test cases and keywords.
	- Test scenarios cover various filter combinations ensuring thorough testing of the functionality.
	- Assertions are appropriately used to verify the expected behavior of the commands.
	- Variable usage is consistent and meaningful.
	- The suite appears comprehensive and should effectively verify the functionality described.

2. tests/RobotFramework/tests/plugin_apt/filter_packages_list_output_with_include_pattern.robot
	- Demonstrates a commendable level of organization with clear delineation of test cases and keywords.
	- The suite appears comprehensive and should effectively verify the functionality described.

3. tests/RobotFramework/tests/tedge/call_tedge_config_list.robot
	- The test suite is relatively easy to understand and maintain.
	- It tests both setting and unsetting of configurations, ensuring proper functionality in both cases.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -745,7 +745,15 @@ define_tedge_config! {
...
-            max_packages: u32
+            max_packages: u32,
+
+            /// The filtering criterion, in form of regex, that is used to filter packages list output
+            #[tedge_config(example = "tedge.*")]
+            include: String,
+
+            /// The filtering criterion, in form of regex, that is used to filter out packages from the output list
+            #[tedge_config(example = "tedge.*")]
+            exclude: String,
```
Bravo555, line 756: suggestion: for examples add packages mentioned in the issue, i.e. `libc.*, iptables-module-.*, kernel-.*`

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -461,13 +468,54 @@ impl Plugin for ExternalPluginCommand {
...
+                    let include_filter = if let Some(include) = self.include.as_ref() {
+                        Regex::new(format!(r"^{}\t.*", include).as_str())?
+                    } else {
+                        Regex::new(r".*")?
+                    };
+
+                    output
+                        .stdout
+                        .split_inclusive(|c| *c == b'\n')
+                        .map(|line| std::str::from_utf8(line).unwrap_or_default())
```
Bravo555, line 503: todo: unwrap here will crash, should return error instead, unless this is intended

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -461,13 +468,54 @@ impl Plugin for ExternalPluginCommand {
...
+            let filtered_output = match (&self.exclude, &self.include) {
+                (None, None) => output.stdout,
+                (Some(exclude), Some(include)) => {
+                    let exclude_filter = Regex::new(format!(r"^{}\t.*", exclude).as_str())?;
+                    let include_filter = Regex::new(format!(r"^{}\t.*", include).as_str())?;
+
+                    output
+                        .stdout
+                        .split_inclusive(|c| *c == b'\n')
+                        .map(|line| std::str::from_utf8(line).unwrap_or_default())
```
Bravo555, line 480: todo: unwrap here will crash, should return error instead, unless this is intended

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -461,13 +468,54 @@ impl Plugin for ExternalPluginCommand {
...
+                        .filter(|line| {
+                            include_filter.is_match(line) && !exclude_filter.is_match(line)
+                        })
+                        .flat_map(|line| line.as_bytes().to_vec())
+                        .collect()
+                }
+            };
```
Bravo555, line 510: suggestion: Is there a specific reason why we do "both filters present" and "single filter present" cases slightly differently? If there's not, I think writing it in terms of a single, general case, would be easier to follow, as right now it's a bit confusing, understanding why we use fallbacks that we do and why the filter condition is OR first and AND later.

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -461,13 +468,54 @@ impl Plugin for ExternalPluginCommand {
         let command = self.command(LIST, None)?;
         let output = self.execute(command, logger).await?;
         if output.status.success() {
+            let filtered_output = match (&self.exclude, &self.include) {
+                (None, None) => output.stdout,
+                (Some(exclude), Some(include)) => {
+                    let exclude_filter = Regex::new(format!(r"^{}\t.*", exclude).as_str())?;
```
Bravo555, line 474: suggestion: We tell the user that these options are regexes, will the code correctly handle when the user includes `^`, `$`, `\t` or `.*` as part of their regex? I'd like to see it in the test case to make sure.

tests/RobotFramework/tests/cumulocity/software_management/sm-plugin.robot
```diff
@@ -95,3 +132,4 @@ Set Service User
     Execute Command    mkdir -p /etc/systemd/system/${SERVICE_NAME}.service.d/
     Execute Command    cmd=printf "[Service]\nUser = ${SERVICE_USER}" | sudo tee /etc/systemd/system/${SERVICE_NAME}.service.d/10-user.conf
     Execute Command    systemctl daemon-reload
+
```
Bravo555, line 135: suggestion: you can enable `files.insertFinalNewline` in vscode 😛

Bravo555 (changes requested):
Two unwraps to remove, and some suggestions regarding matching logic

crates/core/plugin_sm/src/plugin.rs
reubenmiller: No, we fell into the trap before. When both inclusive and exclusive filters are applied, then it is an OR condition. For example if a users has an exclude filter, but they want to have an exception to the exclude, then they can use a single "include" pattern to specify the exceptions.

Ruadhri17: Yes, when two filters are present, the behavior is different (thus implementation is slightly different) as described in Didier's comment above. Basically, include filter takes precedence over exclude one in that case.

crates/core/plugin_sm/src/plugin.rs
didier-wenzek: As this is an unwrap or *default*, this will not crash but ignore lines which are not utf8. This is a reasonable behavior.

Bravo555: I am dumb, this is `unwrap_or_default`, there is no problem.

crates/core/plugin_sm/src/plugin.rs
Bravo555: I am dumb, this is `unwrap_or_default`, there is no problem.

crates/core/plugin_sm/src/plugin.rs
reubenmiller: As a user I would also find the underlying behaviour unexpected, as I can't could do a "starts-with" filter using `^python3-`.

I might make more sense to run the filter on the **package name** field only (e.g. split the string first on `\t` and run the regex on the first element). Then the user can use the full regex syntax, e.g. `^(foo|bar)$`

crates/core/plugin_sm/src/plugin.rs
didier-wenzek: By the way, the *and clause* in  the `_` case is not really an *and* as there is only a single regex that is user provided.

@Ruadhri17 I think you can fallback to your very fist proposal with 4 explicit cases: this mix of `||` and `&&` with puzzle any new comers.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
reubenmiller: Yes, example makes sense...I'm almost tempted to set the default to the following, as this would mean out of the box thin-edge.io will send back something sensible (reducing the network bandwidth)

```sh
^(glibc|libc|kernel-|iptables-module|).*
```

@didier-wenzek Or would this default "do too much", and be unexpected for users?

crates/core/plugin_sm/src/plugin.rs
Bravo555: from [Didier's comment](https://github.com/thin-edge/thin-edge.io/pull/2869#issuecomment-2112298883) (emphasis mine):

- if both are provided, the logic is to keep a package if its name matches the software.plugin.include regex or doesn't match the software.plugin.exclude. The logic is that the primary usage will be to exclude packages (say all libs and kernel modules) and only then add some exceptions (say to include the libc).
- if only software.plugin.include is provided, keep a package if its name matches **(and deny everything else)**
- if only software.plugin.exclude is provided, keep a package if its name doesn't match

I have trouble understanding why "if only `include` is provided then we only keep things it matches" doesn't directly map into "both `include` and `exclude` are provided but `exclude` denies everything" case.

> When both inclusive and exclusive filters are applied, then it is an OR condition. For example if a users has an exclude filter, but they want to have an exception to the exclude, then they can use a single "include" pattern to specify the exceptions.

Yeah, I'm with you on that one. But I don't understand why do we need the second AND case. If we only have `include`, and only want to keep items matched by `include`, I think we can use a fallback value of "deny everything" for `exclude` and reuse the OR condition. Alternatively, if we only have `exclude` given by the user, then for include we can use a regex that includes nothing (so it doesn't add any exceptions) and again we would be able to use the OR condition.

So if I'm wrong, and I'm stepping into the same trap than you did (because maybe there's some hidden assumption that doesn't follow from these 3 rules), then I think we should be extra careful how this would be documented.

crates/core/plugin_sm/src/plugin.rs
Ruadhri17: Added extra split so now only package name is filtered.

crates/core/plugin_sm/src/plugin.rs
Ruadhri17: After quick discussion with @Bravo555 we came to agreement that when you switch default filters, only two cases are needed:

1.  When both filters are None - we just pass the output list and skip unnecessary. Switched filters would not produce desired output anyway.
2. All other cases can be covered by OR condition.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
didier-wenzek: As an example, this makes sense but not as the default. It's better to let the user hide packages himself.

reubenmiller: ok, agreed. let's leave the default empty. Though adding the following regex in the --exclude example will be helpful.

```
^(glibc|libc|kernel-|iptables-module).*
```

crates/core/plugin_sm/src/plugin.rs
```diff
@@ -461,13 +468,35 @@ impl Plugin for ExternalPluginCommand {
         let command = self.command(LIST, None)?;
         let output = self.execute(command, logger).await?;
         if output.status.success() {
+            let filtered_output = match (&self.exclude, &self.include) {
+                (None, None) => output.stdout,
+                _ => {
+                    let exclude_filter =
+                        Regex::new(self.exclude.as_ref().unwrap_or(&r".*".to_string()))?;
+                    let include_filter =
+                        Regex::new(self.include.as_ref().unwrap_or(&r"^$".to_string()))?;
```
didier-wenzek, line 477: That's correct, but deserves comments: this PR got too many comments/questions on that point.

```suggestion
                    // If no exclude pattern is given, exclude everything (except what matches the include pattern)
                    let exclude_filter =
                        Regex::new(self.exclude.as_ref().unwrap_or(&r".*".to_string()))?;
                    // If no include pattern is given, include nothing (except what doesn't match the exclude pattern)   
                    let include_filter =
                        Regex::new(self.include.as_ref().unwrap_or(&r"^$".to_string()))?;
```

didier-wenzek (approved):
Approved - with a suggestion

Bravo555 (approved):
Assuming the test cases are sufficient and correct, the implementation is correct.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -745,7 +745,15 @@ define_tedge_config! {
...
             #[tedge_config(example = "1000", default(value = 1000u32))]
-            max_packages: u32
+            max_packages: u32,
+
+            /// The filtering criterion, in form of regex, that is used to filter packages list output
+            #[tedge_config(example = "^(tedge|c8y).*")]
+            include: String,
+
+            /// The filtering criterion, in form of regex, that is used to filter out packages from the output list
+            #[tedge_config(example = "^(glibc|libc|kernel-|iptables-module).*")]
```
reubenmiller, line 755: Sorry one minor correction (as I just tested the example on yocto):

```suggestion
            #[tedge_config(example = "^(glibc|lib|kernel-|iptables-module).*")]
```

Issue #2873, opened 2024-05-13
tedge init: only create multi-call binary symlinks if they do not exist


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Forcefully recreating the thin-edge.io executable symlinks (for the multi-call binary setup) even though the symlinks already exist causes problems when running on a read-only root filesystem (namely Fedora IoT).

`tedge init` is called on the systemd `ExecStartPre` property to ensure that the required folders exist prior to the service starts (e.g. https://github.com/thin-edge/thin-edge.io/blob/main/configuration/init/systemd/tedge-agent.service#L8). Whilst that a non-zero exit in the `ExecStartPre` phase does not prevent the service from starring, it does introduce noise for the user as they might think something is wrong.

The code below shows the current logic which blindly deletes the symlinks if they exist, then recreates them:

[let file_exists = meta.is_ok();](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/cli/init.rs#L68-L96)

**Describe the solution you'd like**

Symlinks should only be created if they do not already exist. This would avoid ExecStartPre errors when running services on a read-only root filesystem.

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
There is sufficient test coverage provided by the unit tests as part of the linked PR.

gligorisaev:
I am confirming that with the unit tests, part of the PR, are covering this feature, no need to duplicate with sstem test

PR #2876, opened 2024-05-14
Avoid modifying files if they are up to date in `tedge init`

jarhodes314:
## Proposed changes
`tedge init` was blindly recreating symlinks for the tedge components, even if those symlinks existed. This was causing problems running thin-edge on read-only file-systems. This PR adds checks to establish if the symlinks need to change, and avoids changing them if not. It also makes a similar change to some permission changes further through the process.

Testing on my local machine, I can now run `tedge init` successfully with the default settings without using `sudo`, so it doesn't appear to be attempting to modify any files following these changes.

## Types of changes

## Paste Link to the issue
- #2873

## Checklist


crates/core/tedge/src/cli/init.rs
```diff
@@ -57,140 +62,43 @@ impl TEdgeInitCmd {
...
-            config_dir.join(".tedge-mapper-c8y"),
-            PermissionEntry::new(
-                Some(self.user.clone()),
-                Some(self.group.clone()),
-                Some(0o775),
-            ),
-        )?;
+        create_directory(&config.logs.path, permissions(0o775))?;
+        create_directory(&config.data.path, permissions(0o775))?;
+        create_directory(config_dir.join(".tedge-mapper-c8y"), permissions(0o775))?;
```
jarhodes314, line 101: None of these changes should change the behaviour from what it was previously, it just felt like a lot of code that could easily be tidied up.

I was however surprised to see that the mode bits vary (mostly `775`, but occassionally `755`), but couldn't find any explanation in the git history as to why that is. Is it intentional that some directories can only be written to by the `tedge` user, and not the `tedge` group, or are they all meant to be the same?

crates/common/tedge_utils/src/file.rs
```diff
@@ -380,7 +383,7 @@ pub fn change_user_and_group(file: &Path, user: &str, group: &str) -> Result<(),
     let gid = get_metadata(Path::new(file))?.gid();
 
     // if user and group are same as existing, then do not change
-    if (ud != uid) && (gd != gid) {
+    if (ud != uid) || (gd != gid) {
```
jarhodes314, line 386: This looked like a bug as I think we want to `chown` if there is any change, not just if both the `uid` and `gid` have changed.

didier-wenzek (approved):
Approved

reubenmiller (approved):
Looks good. I also ran tedge init with `/etc/tedge` mounted as read-only and it exited with a 0 status code.

```sh
root@afdc02e108df:~# mount --bind -o ro /etc/tedge /etc/tedge
root@afdc02e108df:~# touch /etc/tedge/testfile
touch: cannot touch '/etc/tedge/testfile': Read-only file system
root@afdc02e108df:~# tedge init
root@afdc02e108df:~# echo $?
0
```

PR #2878, opened 2024-05-15
Create tasks that will follow the Documentation instructions

gligorisaev:
## Proposed changes
For documentation checking purposes I am creating collection of End to End tasks which will follow the descriptions/instructions from thinedge documentation. 
These tasks are running on an Raspberry PI.
The idea behind is that as soon as some task fail that would mean that changes in the documentations are needed or some regression mailfunction has been detected.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,47 @@
...
+    
+*** Variables ***
+${REMOTE_FILE}    output.txt
+${LOCAL_FILE}    output1.txt
+
+*** Tasks ***
+
+Install/Update of thinedge curl
+   ${log}   Execute Command    curl -fsSL https://thin-edge.io/install.sh | sh -s
+
```
didier-wenzek, line 19: Basic checks have to be added. e.g. running `tedge --help` or checking the mapper and the agent are installed.

gligorisaev:
For now no additional test are planned for this PR

gligorisaev:
> Some questions:
> 
> * I see that in RobotFramework besides tests one can create tasks. How are they different and why are we doing it as tasks and not tests in this case?

- **Tasks** are used for automating processes that are not directly related to testing but rather to operations, administrative tasks, or other kinds of workflows.
- **Tests** are designed specifically for validating the functionality, performance, and reliability of software applications. They include assertions to verify that the application behaves as expected.

> * As far as I can see these tasks are not part of the normal test suite (in the `RobotFramework/tests/` directory) and they're not run using a docker adapter, but physical devices. In this case will running these tasks be automated to happen on PRs, as current integration test suite is, or will it only be run manually?

- The purpose of this tasks is to keep the integrity and correctness of our documentation under controll, meaning, if task fails than we need to adjust the documentation

> * Trying to run this task suite on my end, I am getting some errors because tasks assume a home directory of `/home/pi`. Is the goal of these tasks to run them under a controlled environment so these assumptions ultimately don't matter, or should we try to keep it portable? IMO it would be nice to run them on the developer machines, but perhaps that's not possible due to some blockers?

- Running these task understand to adjust your .env file like:
#
# SSH - Device Target
#
SSH_CONFIG_HOSTNAME=192.168.193.56
SSH_CONFIG_USERNAME=pi
SSH_CONFIG_PASSWORD=thinedge

just add the information of your developer machine (or like in the example use one of the devices available on ZeroTier)

reubenmiller:
> * Trying to run this task suite on my end, I am getting some errors because tasks assume a home directory of `/home/pi`. Is the goal of these tasks to run them under a controlled environment so these assumptions ultimately don't matter, or should we try to keep it portable? IMO it would be nice to run them on the developer machines, but perhaps that's not possible due to some blockers?

Though we should avoid using the `/home/pi` directory explicitly as this is not a standard linux user, and would only work on devices using a raspberry pi built OS image.

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,102 @@
...
+   Execute Command    ./uninstall-thin-edge_io.sh purge
+   Log    Succesfull uninstalled with purge
+
+#Verify ThinEdgeIO is uninstalled
+   ${OUTPUT}    Execute Command    tedge --help    ignore_exit_code=True
+   Should Not Contain    ${OUTPUT}    tedge is the cli tool for thin-edge.io
+
+Check repository creation
+   ${OUTPUT}    Execute Command    ls /etc/apt/sources.list.d/
+   Should Contain    ${OUTPUT}    thinedge-tedge-release.list
```
reubenmiller, line 87:
```suggestion
```

I would exclude the check for an explicit `*.list` file as the filename is not part of the public interface. The `apt-cache search tedge` is enough to validate if the thin-edge.io apt source list was added successfully.

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,102 @@
...
+   Should Contain    ${OUTPUT}    tedge - CLI tool use to control and configure thin-edge.io
+   Should Contain    ${OUTPUT}    tedge-agent - thin-edge.io interacts with a Cloud Mapper and one or more Software Plugins
+   Should Contain    ${OUTPUT}    tedge-apt-plugin - thin-edge.io plugin for software management using apt
+   Should Contain    ${OUTPUT}    tedge-configuration-plugin - thin-edge.io configuration management
+   Should Contain    ${OUTPUT}    tedge-dummy-plugin - TEST ONLY: thin-edge dummy software management plugin
+   Should Contain    ${OUTPUT}    tedge-full - thin-edge.io virtual package to automatically install all tedge packages
+   Should Contain    ${OUTPUT}    tedge-log-plugin - thin-edge.io log file retriever
+   Should Contain    ${OUTPUT}    tedge-mapper - thin-edge.io mapper that translates thin-edge.io data model to c8y/az data model.
+   Should Contain    ${OUTPUT}    tedge-minimal - thin-edge.io virtual package to install a minimal set of tedge components
+   Should Contain    ${OUTPUT}    tedge-watchdog - thin-edge.io component which checks the health of all the thin-edge.io components/services.
```
reubenmiller, line 98: We don't need to check for every package here as there are a lot of old or test packages in this list (which is not part of the public interface).

```suggestion
   Should Contain    ${OUTPUT}    tedge - CLI tool use to control and configure thin-edge.io
   Should Contain    ${OUTPUT}    tedge-agent - thin-edge.io interacts with a Cloud Mapper and one or more Software Plugins
   Should Contain    ${OUTPUT}    tedge-apt-plugin - thin-edge.io plugin for software management using apt
   Should Contain    ${OUTPUT}    tedge-full - thin-edge.io virtual package to automatically install all tedge packages
   Should Contain    ${OUTPUT}    tedge-mapper - thin-edge.io mapper that translates thin-edge.io data model to c8y/az data model.
   Should Contain    ${OUTPUT}    tedge-watchdog - thin-edge.io component which checks the health of all the thin-edge.io components/services.
```

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,102 @@
...
+Verify ThinEdgeIO is installed
+   ${OUTPUT}    Execute Command    tedge --help
+   Should Contain    ${OUTPUT}    tedge is the cli tool for thin-edge.io
+   Log    ThinEdgeIO was successfully installed
+
+Uninstall ThinEdgeIO
+   Transfer To Device    ${CURDIR}/uninstall-thin-edge_io.sh    /home/pi/uninstall-thin-edge_io.sh
+   Execute Command    chmod a+x uninstall-thin-edge_io.sh
+   Execute Command    ./uninstall-thin-edge_io.sh purge
+   Log    Succesfull uninstalled with purge
```
reubenmiller, line 79: Avoid using `/home/pi` as it assumes that the `pi` user exists (which is not always the case).

```suggestion
   Transfer To Device    ${CURDIR}/uninstall-thin-edge_io.sh    /var/local/share/uninstall-thin-edge_io.sh
   Execute Command    chmod a+x /var/local/share/uninstall-thin-edge_io.sh
   Execute Command    /var/local/share/uninstall-thin-edge_io.sh purge
   Log    Succesful uninstalled with purge
```

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,173 @@
...
+    Should Contain    ${output}    Connection check is successful.
+    Should Contain    ${output}    Checking if tedge-mapper is installed.
+    Should Contain    ${output}    Starting tedge-mapper-c8y service.
+    Should Contain    ${output}    Persisting tedge-mapper-c8y on reboot.
+    Should Contain    ${output}    tedge-mapper-c8y service successfully started and enabled!
+    Should Contain    ${output}    Enabling software management.
+    Should Contain    ${output}    Checking if tedge-agent is installed.
+    Should Contain    ${output}    Starting tedge-agent service.
+    Should Contain    ${output}    Persisting tedge-agent on reboot.
+    Should Contain    ${output}    tedge-agent service successfully started and enabled!
```
reubenmiller, line 114: I don't think it is worth explicitly checking every message here...as generally there is no explicit contract between console output and the user (except for "conveying useful information")...The official contract is that the command should exit with an exit code of `0` if everything worked as expected.

You could add an additional connection test, `tedge connect c8y --test` to verify that the connection is working if you wanted to verify it (but I'm pretty sure we already have system tests for that part).

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,173 @@
...
+
+Send Child Device Temperature Measurement
+    [Documentation]    Send a temperature measurement to a child device.
+    Execute Command    tedge mqtt pub 'te/device/child1///m/environment' '{"temperature": 25}'
+    Log    Sent temperature measurement to child device: 25°C
+
+Verify Measurement In Cumulocity
+    [Arguments]    ${type}    ${value}
+    [Documentation]    Verify the measurement in Cumulocity.
+    # Add implementation for checking the measurement in Cumulocity
```
reubenmiller, line 140: This is a TODO, so it might make sense to add it in the PR (since we already have an assertion for this included in the `Cumulocity` RobotFramework library).

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,173 @@
...
+    # Add implementation for checking the child device measurement in Cumulocity
+    Log    Verified ${type} measurement for child device ${child_id} with value ${value} in Cumulocity
+
+Custom Setup
+    [Documentation]    Initializes the device environment. 
+    ...                Sets up the device, transfers necessary packages, 
+    ...                installs them, and configures Cumulocity for connectivity.
+    ${DEVICE_SN}=    Setup    skip_bootstrap=True
+    Set Suite Variable    ${DEVICE_SN}
+    ${log}    Transfer To Device    target/aarch64-unknown-linux-musl/packages/*.deb    /home/pi/
```
reubenmiller, line 160: This assumes you are running on an arm64/aarch64 system. We should be more dynamic and allow detecting the target architecture and then selecting the correct package to install.

tests/RobotFramework/tasks/E2E_Tutorials/raising_alarms.robot
```diff
@@ -0,0 +1,96 @@
...
+
+*** Keywords ***
+
+Raise Alarm Keyword
+    [Documentation]    Publishes an alarm message with the specified type, text, 
+    ...                and severity to the device using ThinEdgeIO MQTT. 
+    ...                Verifies that the alarm is present on the device.
+    [Arguments]    ${alarm_type}    ${text}    ${severity}
+    ${payload}=    Set Variable    "text": "${text}", "severity": "${severity}"
+    Execute Command    tedge mqtt pub te/device/main///a/${alarm_type} '{${payload}}' -r -q 2
```
reubenmiller, line 44: Slight refactoring to make it more readable (and it can be applied to all of the other keywords which use a similar setup)

```suggestion
    ${payload}=    Set Variable    {"text": "${text}", "severity": "${severity}"}
    Execute Command    tedge mqtt pub te/device/main///a/${alarm_type} '${payload}' -r -q 2
```

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
gligorisaev: Please have in mind that the main purpose of the tasks is not to test the feature because we do have tests that are doing that, with the tasks I am checking the documentation correctness and in the documentation (https://thin-edge.github.io/thin-edge.io/start/connect-c8y/) this content is stated as it should exist

reubenmiller: Yes, but we have to remember that the "accurate documentation" does not mean that every single console output needs to be 100% accurate, as the exact output is not important, it is more important that the "commands" and the outcome of the commands are accurate.

If we do verbatim comparisons of the console output, it will just introduce more problems to keep everything in sync rather than actually providing value to the user.

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,97 @@
...
+   Execute Command    chmod a+x /var/local/share/uninstall-thin-edge_io.sh
+   Execute Command    /var/local/share/uninstall-thin-edge_io.sh purge
+
+#Verify ThinEdgeIO is uninstalled
+   ${OUTPUT}    Execute Command    command -V tedge    exp_exit_code=!0 
+   ${OUTPUT}    Execute Command    command -V tedge    exp_exit_code=!0
+
+Check repository creation
+   ${OUTPUT}    Execute Command    ls /etc/apt/sources.list.d/
+   Should Contain    ${OUTPUT}    *.list
```
reubenmiller, line 86: I would leave this assertion out, as it does not provide any value.

```suggestion
```

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,97 @@
...
+   ${OUTPUT}    Execute Command    apt-cache search tedge
+   Should Contain    ${OUTPUT}    tedge - CLI tool use to control and configure thin-edge.io
+   Should Contain    ${OUTPUT}    tedge - CLI tool use to control and configure thin-edge.io
```
reubenmiller, line 89: Duplicated line.

```suggestion
   Should Contain    ${OUTPUT}    tedge - CLI tool use to control and configure thin-edge.io
```

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
gligorisaev: I would leave it as it is because I do use this taks exclusively only with raspberry's

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
gligorisaev: at least I will get alerted that something has been changed, which is my aim

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
gligorisaev: Is this task the right place for it?

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
gligorisaev: as stated in the other Task, I do use this exclusively with Raspberry

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -82,13 +82,9 @@ Uninstall ThinEdgeIO
    ${OUTPUT}    Execute Command    command -V tedge    exp_exit_code=!0
 
 Check repository creation
-   ${OUTPUT}    Execute Command    ls /etc/apt/sources.list.d/
-   Should Contain    ${OUTPUT}    *.list
    ${OUTPUT}    Execute Command    apt-cache search tedge
    Should Contain    ${OUTPUT}    tedge - CLI tool use to control and configure thin-edge.io
    Should Contain    ${OUTPUT}    tedge - CLI tool use to control and configure thin-edge.io
-   Should Contain    ${OUTPUT}    tedge-agent - thin-edge.io interacts with a Cloud Mapper and one or more Software Plugins
-   Should Contain    ${OUTPUT}    tedge-apt-plugin - thin-edge.io plugin for software management using apt
```
reubenmiller: Something went wrong here as the wrong lines have been deleted

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
reubenmiller: Yes it makes sense to add it since we are taking about a new file, so we shouldn't be adding active TODOs to new files. Either remove the step completely or add the check...But it should be a simple `Device Should Have Measurements    type=environment    minimum=1    maximum=1` (updating to fit the measurements being created).

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
reubenmiller: Though the idea is to have something which can be applied to also different devices, and doesn't fail when using on a Raspberry Pi armv7 device...

Or alternatively, clearly state at the top of the task that this script is only intended to be run on Raspberry Pi devices with an aarch64/arm64 architecture.

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
reubenmiller: ok, though I don't think it provides much value.

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,201 @@
...
+# Custom Setup
+#     [Documentation]    Initializes the device environment. 
+#     ...                Sets up the device, transfers necessary packages, 
+#     ...                installs them, and configures Cumulocity for connectivity.
+#     ${DEVICE_SN}=    Setup    skip_bootstrap=True
+#     Set Suite Variable    ${DEVICE_SN}
+#     ${log}    Transfer To Device    target/aarch64-unknown-linux-musl/packages/*.deb    /home/pi/
+#     Execute Command    sudo dpkg -i *.deb
+#     Log    Installed new packages on device
+
```
reubenmiller, line 191: This should be deleted

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,201 @@
...
+    Log    Sent combined measurement
+
+Send Child Device Temperature Measurement
+    [Documentation]    Send a temperature measurement to a child device.
+    Execute Command    tedge mqtt pub 'te/device/child1///m/environment' '{"temperature": 25}'
+    Log    Sent temperature measurement to child device: 25°C
+
+Verify Measurement In Cumulocity
+    Device Should Have Measurements    type=environment    minimum=1    maximum=1
+
```
reubenmiller, line 139: It would make more sense to make this a Keyword (function) that takes an argument for the type, then you can reuse it for the other measurements.

```
Verify Measurement In Cumulocity
    [Arguments]        ${type}
    Device Should Have Measurements    type=${type}    minimum=1    maximum=1

```

Then the usage is just:

```
Verify Measurement In Cumulocity    type=environment
```

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,92 @@
...
+   Uninstall ThinEdgeIO
+
+Update using a package manager
+   ${log}    Execute Command    sudo apt-get update && yes | sudo apt-get install tedge-full
+   Verify ThinEdgeIO is installed
+   Uninstall ThinEdgeIO
+
+Optional: Linux distributions without systemd curl
+   ${OUTPUT}    Execute Command    curl -fsSL https://thin-edge.io/install-services.sh | sh -s    ignore_exit_code=True
+   #Not verifiing this test step because the test running in container already exists: tests/RobotFramework/tests/installation/install_on_linux.robot
```
reubenmiller, line 34:
```suggestion
   #Skip verification step as system tests already exist: tests/RobotFramework/tests/installation/install_on_linux.robot
```

tests/RobotFramework/tasks/E2E_Tutorials/Install.robot
```diff
@@ -0,0 +1,92 @@
...
+   #Checking only that the link is correct
+   Should Contain    ${OUTPUT}    Welcome to the thin-edge.io community!
+
+Optional: Linux distributions without systemd wget
+   ${OUTPUT}    Execute Command    wget -O - https://thin-edge.io/install-services.sh | sh -s
+   #Not verifiing this test step because the test running in container already exists: tests/RobotFramework/tests/installation/install_on_linux.robot
```
reubenmiller, line 40:
```suggestion
   #Skip verification step as system tests already exist: tests/RobotFramework/tests/installation/install_on_linux.robot
```

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,192 @@
...
+    Log    Sent temperature measurement to child device: 25°C
+
+Verify Measurement In Cumulocity
+    [Arguments]        ${type} 
+    Device Should Have Measurements    type=${type}    minimum=1    maximum=1
+
+Verify Combined Measurement In Cumulocity
+    [Documentation]    Verify the combined measurement in Cumulocity.
+    # Add implementation for checking the combined measurement in Cumulocity
+    Log    Verified combined measurement in Cumulocity
```
reubenmiller, line 144: Now you can use the same `Verify Measurement In Cumulocity` keyword to verify that this message is also published.

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,192 @@
...
+    ...    ELSE IF    '${arch}' == 'armv7l'    Transfer Armv7l Packages
+    ...    ELSE    Log    Unsupported architecture: ${arch}
+
+    # Install packages
+    Execute Command    sudo dpkg -i *.deb
+    Log    Installed new packages on device
+
+Transfer Aarch64 Packages
+    [Documentation]    Transfers Aarch64 architecture packages to the device.
+    ${log}    Transfer To Device    target/aarch64-unknown-linux-musl/packages/*.deb    /home/pi/
```
reubenmiller, line 175: We shouldn't be using `/home/pi` path...use some other path like:

```suggestion
    ${log}    Transfer To Device    target/aarch64-unknown-linux-musl/packages/*.deb    /var/local/share/
```

tests/RobotFramework/tasks/E2E_Tutorials/connect_to_c8y_send_measurements.robot
```diff
@@ -0,0 +1,192 @@
...
+
+Transfer Armv7l Packages
+    [Documentation]    Transfers ARMv7l architecture packages to the device.
+    ${log}    Transfer To Device    target/armv7-unknown-linux-musleabihf/packages/*.deb    /home/pi/
+    Log    Transferred ARMv7l packages to device
+
+Custom Teardown
+    [Documentation]    Cleans up the device environment. 
+    ...                Uninstalls ThinEdgeIO, removes packages and scripts, and retrieves logs.
+    Transfer To Device    ${CURDIR}/uninstall-thin-edge_io.sh    /home/pi/uninstall-thin-edge_io.sh
```
reubenmiller, line 186:
```suggestion
    Transfer To Device    ${CURDIR}/uninstall-thin-edge_io.sh    /var/local/share/uninstall-thin-edge_io.sh
```

reubenmiller: Though there are other places which also use the `/home/pi` folder...best to avoid this all together.

tests/RobotFramework/tasks/E2E_Tutorials/uninstall-thin-edge_io.sh
```diff
@@ -0,0 +1,84 @@
...
+            sudo apt --assume-yes "$1" "$package"
+        fi
+    done
+}
+
+remove_installed_files() {
+    for path in $installation_paths; do
+        if [ -d "$path" ]; then
+            sudo rm -rf "$path/tedge"
+            sudo rm -rf "$path/tedge_*"
```
reubenmiller, line 65: This won't work as expected, as the shell wildcard is within the double quotes, meaning it will not be expanded to folders with `tedge_*` in the name.

```suggestion
            sudo rm -rf "$path/tedge_"*
```

reubenmiller:
@gligorisaev What is the state of this PR, it seems to be still a work-in-progress. In that case it would be better to set it to the Draft state. It is unclear which comments have been addressed or not as there aren't any links to the commits which address the comments.

didier-wenzek:
Closing as the value of having tasks vs tests is not clear.

Issue #2880, opened 2024-05-15
Simplify the operation handling code by replacing fragmented control flow with regular async/await


Bravo555:
**Is your refactoring request related to a problem? Please describe.**

As originally described in the [Further Comments section in the PR about using the File Transfer Service for log and config operations](https://github.com/thin-edge/thin-edge.io/pull/2440), the current architecture of the C8y mapper actor adds unnecessary complexity to the operation handling code, by:

- spreading the execution of the operation between many different functions, making it harder to locally reason about the code
- needing additional variables to keep the state necessary for handling of the operations, read by the different functions that run at different stages of the operation lifecycle

This:

- introduces additional friction when making changes to operation handling, because one needs to update multiple functions that each handles a part of the operation's lifecycle, and the signature of the data being passed between these functions, frequently adding additional fields
- mixes levels of abstraction: the actor needs to be aware of the different operations so that it can invoke correct functions to continue handling them

It concerns mainly the interaction between downloader and uploader actors, where every time we need to download/upload from within the operation, we need to send the request and return, in order not to block handling of other things.

**Describe the solution you'd like**

The preferred solution would be to handle the entirety of an operation in a module dedicated to that operation, without returning, saving state, and relying on other parts of the code to continue the execution. We should be able to run the future that handles the operation concurrently, as that's what async/await should allow.

This was already done for `c8y_firmware_manager` (https://github.com/thin-edge/thin-edge.io/pull/2600), now the aim would be to do something similar with `c8y_mapper_ext` and its operation handling.

**Describe alternatives you've considered**

**Additional context**

Bravo555:
#2909 eliminated the biggest problems with operation handling, but there are still some big improvements to do:

Follow-up tasks:
- [ ] add tests confirming that new requests are still processed and mapped while file downloads/uploads are ongoing (https://github.com/thin-edge/thin-edge.io/pull/2904#discussion_r1620205761)
- [ ]stop relying on MQTT echoing back operation clearing message to remove it from `OperationConverter` (https://github.com/thin-edge/thin-edge.io/pull/2904#discussion_r1673714165)
- [ ] remove `IdGenerator` from `software_update.rs`
- [ ] clean up Operation API
	- generate topic clearing messages once for all operations
	- introduce an "Operation" abstraction layer that decouples it from our MQTT API
- [ ] split off more Operation-related things from `CumulocityConverter`
	- `operations`
	- `children`
	- `operation_logs`
	- `active_commands`
[...]

PR #2881, opened 2024-05-15
Simplify the operation handling code by replacing fragmented control flow with regular async/await

Bravo555:
## Proposed changes

This PR should make the operation handle simpler to reason about and change, by allowing operation handlers to use `.await` without blocking. The scope of the PR is to show the idea working on the `config_snapshot` and `config_update` operations.

Below i describe the current state, which is subject to change.

1. In 6b90d51e4420dbfe5cd7906bcf57e48494e9ae4e, the main C8y mapper actor is prepared for its messages to be handled concurrently in separate tasks. For now, we use `Mutex`es to guard the `MqttPublisher` and `CumulocityConverter`:
   - using wrapper `struct MqttPublisher(Arc<Mutex<LoggingSender<MqttMessage>>>)` with `fn send(&self)` is ok, as we lock the mutex to use the sender, and unlock it immediately after we finished sending, so it shouldn't block. The alternative would be to clone the sender for each worker, and I will explore it later.
   - with `Mutex<CumulocityConverter>`, it ensures that only a single worker can use `CumulocityConverter` at any given time, which can lead to deadlocks if we're not careful, and currently results in the same behaviour as before, i.e. blocking in the operation handling code will block processing of other messages, because the lock will be open across await points.
      This is temporary, and will have to be fixed, but is used right now so I could show how the full operation handling function would look like, without the current fragmentation. Also, the test demonstrating that the converter doesn't block should be made.
2. In 72dc9d09bb644758c4684253ee8f276d53168666, I export `Identity` type from the `download` crate, so I can use it in the next commit.
3. In 0e8656851b2fb1d9be90021ef0feab33d9d7d818, `Downloader` is used directly in the operation handler, to show how we can use it directly without going through the actor: if we only want to download a file via HTTP, and this download shouldn't influence anything else, we can use it directly.
4.  In 892ecbec71a21a9e9bc93252f27abdac4057390d we use `UploaderActor` via `ClientMessageBox`, where we receive the response directly in the same place where we `send()`, instead of going to the top-level receiver in the top-level actor, which then needs to decide to call an operation handling function again. This way we still use an uploader actor, and have messages sent between them, but don't need to fragment our control flow. This was done to show that we can still send and receive messages from other actors from within the proposed new operation handlers.

### Next steps

- [x] rebase newest changes
- [x] ~~replace lock on mqtt sender with cloned senders~~ (can be done, but will require `&mut self`)
- [ ] add more tests to see which operations are blocked/executed concurrently
- [ ] make necessary changes to `CumulocityConverter` so operation handling doesn't block
- [ ] clean up operation handling code

## Types of changes

## Paste Link to the issue
- #2880 

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -14,11 +14,14 @@ use c8y_auth_proxy::url::ProxyUrlGenerator;
 use c8y_http_proxy::handle::C8YHttpProxy;
 use c8y_http_proxy::messages::C8YRestRequest;
 use c8y_http_proxy::messages::C8YRestResult;
+use camino::Utf8Path;
 use std::path::PathBuf;
+use std::sync::Arc;
 use std::time::Duration;
 use tedge_actors::fan_in_message_type;
 use tedge_actors::Actor;
 use tedge_actors::Builder;
+use tedge_actors::ChannelError;
 use tedge_actors::CloneSender;
 use tedge_actors::DynSender;
 use tedge_actors::LoggingSender;
@@ -45,6 +48,7 @@ use tedge_uploader_ext::UploadRequest;
 use tedge_uploader_ext::UploadResult;
 use tedge_utils::file::create_directory_with_defaults;
 use tedge_utils::file::FileError;
+use tokio::sync::Mutex;
 use tracing::error;
 use tracing::warn;
 
@@ -62,13 +66,34 @@ pub(crate) type IdDownloadRequest = (CmdId, DownloadRequest);
 fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, IdUploadResult, IdDownloadResult] : Debug);
 type C8yMapperOutput = MqttMessage;
 
+#[derive(Clone)]
+struct MqttPublisher(Arc<Mutex<LoggingSender<MqttMessage>>>);
+
+impl MqttPublisher {
+    pub fn new(mqtt_publisher: LoggingSender<MqttMessage>) -> Self {
+        Self(Arc::new(Mutex::new(mqtt_publisher)))
+    }
+
+    pub async fn send(&self, message: MqttMessage) -> Result<(), ChannelError> {
+        self.0.lock().await.send(message).await
+    }
+}
+
 pub struct C8yMapperActor {
     converter: CumulocityConverter,
     messages: SimpleMessageBox<C8yMapperInput, C8yMapperOutput>,
-    mqtt_publisher: LoggingSender<MqttMessage>,
+    mqtt_publisher: MqttPublisher,
     timer_sender: LoggingSender<SyncStart>,
     bridge_status_messages: SimpleMessageBox<MqttMessage, MqttMessage>,
     c8y_bridge_service_name: String,
+}
+
+pub struct C8yMapperWorker {
+    // converter methods that handle download and upload state changes are mutable, so need to
+    // synchronize them with a mutex
+    converter: Mutex<CumulocityConverter>,
+    mqtt_publisher: MqttPublisher,
+    ops_dir: Arc<Utf8Path>,
 }
 
 #[async_trait]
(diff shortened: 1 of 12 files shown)


crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -60,15 +64,36 @@ pub(crate) type IdDownloadRequest = (CmdId, DownloadRequest);
 fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, IdUploadResult, IdDownloadResult] : Debug);
 type C8yMapperOutput = MqttMessage;
 
+#[derive(Clone)]
+struct MqttPublisher(Arc<Mutex<LoggingSender<MqttMessage>>>);
```
albinsuresh, line 63: Any specific reason you couldn't just freely clone the senders and use them independently? I see that you're already planning on exploring that option, but just wondering why you chose this path first. Faced any issues with cloned senders?

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -95,25 +120,46 @@ impl Actor for C8yMapperActor {
...
+        let worker = Arc::new(C8yMapperWorker {
+            converter: Mutex::new(self.converter),
+            mqtt_publisher: self.mqtt_publisher,
+            ops_dir,
+        });
+
+        while let Some(event) = messages.recv().await {
+            let worker = Arc::clone(&worker);
+
+            tokio::spawn(async move {
```
albinsuresh, line 130: I had a different model in mind where we have independent workers for each operation handler with minimal coordination with the top level mapper actor/worker, eliminating the need to share and lock the top level converter. This mapper actor would still coordinate the MQTT messages and convert them to typed operation messages and dispatch that to those respective operation workers. Once the worker gets an operation message, they should be able to freely exchange messages between the other actors like upload/download actors. Extending the usage of the `RequestEnvelope` beyond the HTTP actors would be key to enabling this.

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
```diff
@@ -125,37 +123,93 @@ impl CumulocityConverter {
...
                 let destination_path = destination_dir.path().join(config_filename);
 
-                self.pending_fts_download_operations.insert(
-                    cmd_id.into(),
-                    FtsDownloadOperationData {
-                        download_type: FtsDownloadOperationType::ConfigDownload,
-                        url: tedge_file_url.clone(),
-                        file_dir: destination_dir,
+                let downloader =
+                    Downloader::new(destination_path.clone(), self.config.identity.clone());
```
albinsuresh, line 141: I believe we can continue using the downloader actor here by using a `ClientMessageBox` to it so that we can await on the response receiver inline. The `c8y_firmware_manager` is already using the same.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -375,7 +309,7 @@ impl C8yMapperBuilder {
         mqtt.connect_sink(config.topics.clone(), &box_builder.get_sender());
         let http_proxy = C8YHttpProxy::new(http);
         let timer_sender = timer.connect_client(box_builder.get_sender().sender_clone());
-        let upload_sender = uploader.connect_client(box_builder.get_sender().sender_clone());
+        let uploader = ClientMessageBox::new(uploader);
```
albinsuresh, line 312: We should be able to do the same for the downloader as well.

crates/extensions/c8y_mapper_ext/src/actor.rs
Bravo555: I did that first because I knew that I'll need `Mutex`es for synchronization of other parts, and to be honest, I forgot that our senders are `Clone`, so I thought `Mutex` is the only option, but it's not the case.

crates/extensions/c8y_mapper_ext/src/actor.rs
Bravo555: > I had a different model in mind where we have independent workers (even actors, may be) for each operation handler with minimal coordination with the top level mapper actor/worker, eliminating the need to share and lock the top level converter.

Yes, the idea would be to decouple as much as possible from `CumulocityConverter` wrt. the operation handling, so it can be run separately as a new task, or maybe as a new actor, without locking the rest of the converter. Right now, basically all the functionality is implemented in `CumulocityConverter` in `&mut self` methods, so it's impossible to run stuff concurrently without locking.

Ideally the each operation handler should be as independent as possible, and we should be able to easily reason about each in isolation, without the control flow jumping between `actor.rs`, `converter.rs`, and e.g. `log_upload.rs`.

Right now they're just tokio tasks, because my current goal is to just move the boundary on which we block, but the handlers could definitely be extracted out of the main actor.

> This mapper actor would still coordinate the MQTT messages and convert them to typed operation messages and dispatch that to those respective operation workers. Once the worker gets an operation message, they should be able to freely exchange messages between the other actors like upload/download actors. Extending the usage of the RequestEnvelope beyond the HTTP actors would be key to enabling this.

That's right, the actor which owns the MQTT message box should route these operation state change message to the operation handlers, which should then be able to do do blocking stuff, like wait for the upload, without synchronizing with the main actor. But I'd say that we overuse messages a bit. It's fine to have actors that manage shared mutable state, but for downloading/uploading, what shared state is there that needs to be managed between different downloads/uploads? Why shouldn't we use the downloader/uploader directly instead of going via the actor?

And I believe that sometimes we use the actor only so that we can mock it in tests. That instead of having an interface that we can mock, we effectively mock the interface by simulating messages being sent in the message boxes. And I think this way of testing is a bit tedious, because you need to handle and ignore unrelated messages, and sometimes when messages are not being send by actor under test, you need to do timeouts, and so on. I think there's room for improvement there.

But that's a discussion for another time, and I expect we'll discuss it more in the future. For the purposes of this PR, the important part is allowing handlers to await either tasks they themselves spawn, or requests sent to other actors via `RequestEnvelope`, one of these two.

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
Bravo555: That's right, we could use the `ClientMessageBox` here, but I also wanted to explore if we could just use the downloader directly, without relying on an outside actor to do it. I'd say it has an advantage of decreasing coupling. The disadvantage is that as the messages are no longer visible, but does other code need them to be visible, other than testing code? If so, then maybe we don't need to go via the actor.

But the main way I did it like this, was to just demonstrate that after the refactor it will be possible, whereas currently this would block. But now we can easily swap this with an `EnvelopeSender` if we want to go via the actor.

crates/extensions/c8y_mapper_ext/src/actor.rs
Bravo555: We can, explanation why it was not done for now in comment: https://github.com/thin-edge/thin-edge.io/pull/2881#discussion_r1603267939

crates/extensions/c8y_mapper_ext/src/actor.rs
albinsuresh: > Why shouldn't we use the downloader/uploader directly instead of going via the actor?

One additional reason, other than message visibility and testability, is to control the bandwidth usage as well. We don't want to let too many uploads/downloads happening parallelly, to reduce the load on the network and the disk. Routing those via a centralised downloader/uploader will let us better control the maximum parallelism that can be allowed.

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
albinsuresh: And additional benefit of going via the downloader actor is described in this comment: https://github.com/thin-edge/thin-edge.io/pull/2881#discussion_r1603970946

Bravo555:
Closing in favour of #2904, which is a more promising and feasible approach.

PR #2897, opened 2024-05-24
Repair test that is failing when running in collection of tests

gligorisaev:
When running as part of bigger collection the `RobotFramework/tests/cumulocity/log/log_operation.robot` is failing because the expected result in the Test Case 

```Default plugin configuration```

```Start Service    tedge-agent```

```Service Should Be Running    tedge-agent```

is returning `Activating` which means that the activation is still going on.

Note that this error accures always when this test is part of a bigger test execution. It can be reproduced if you run:
`robot /<path to the fork>thin-edge.io/tests/RobotFramework/tests/cumulocity`
which will trigger running all tests that in the folder and the subfolders of cumulocity 

## Paste Link to the issue
<br/>

## Checklist

### CODE REVIEW

File: tests/RobotFramework/tests/cumulocity/log/log_operation.robot [hunk 1]
```diff
@@ -101,6 +101,7 @@ Default plugin configuration
     Stop Service    tedge-agent
     ${timestamp}=        Get Unix Timestamp
     Start Service    tedge-agent
+    Sleep    5s    reason=Added waiting time because this test sometimes fails when running as part of workflow
```
reubenmiller, line 104: I'm not sure how this helps exactly, as the "activating" state is something part of systemd, and the service status is sent by the tedge-agent process.

So if this is flaky for other reasons, then it might be better to check that the service status is "down? prior to start the service again.

gligorisaev: Actually it is not "flaky", it is failing always when this test is part of "bigger run" but never failing when run as only one test or when part of not so big (don't know the exact boundry number) run. After adding the sleep time it never failed again

reubenmiller: But then why does the CI pass?

reubenmiller: But the sleep is not logical for me (despite the effect)...It would be more correct to establish that service status is down (after doing `Stop Service`), then proceeding to start the service...in that case you then don't need an explicit sleep.

gligorisaev: added check if sevice is down solved the issue about test fail.

didier-wenzek (approved):
Approved.

Issue #2902, opened 2024-05-24
c8y-mapper ignores any bridge status updates after registration and always reports "up" at registration


rina23q:
**Describe the bug**
There are two issues.
1. c8y-mapper ignores any bridge status updates after the initial registration for the services of bridge status is done.
2. c8y-mapper reports the status "up" although the reported bridge status is `0` at registration.

thin-edge supports multi clouds. If both c8y and az bridges are configured, but c8y-bridge is up and az-bridge is down, still c8y-mapper should report the az-bridge service status "down" in the c8y tenant. However, due to the issue 1), after the initial `102` message for the az-bridge service, even though the az-bridge status changed, the change status won't be reported to c8y tenant.  

**To Reproduce**
[in the versions 1.0.0 and 1.0.1]
1. Connect thin-edge to c8y
2. Publish this message.
```sh
tedge mqtt pub -r te/device/main/service/mosquitto-az-bridge/status/health 0
```
Then, you can see these messages on the console running `tedge mqtt sub "#"`. These messages are the proof that the new service `mosquitto-az-bridge` is registered in the entity store and the service is created in c8y.
```
[te/device/main/service/mosquitto-az-bridge/status/health] 0
[te/device/main/service/mosquitto-az-bridge] {"@id":"TST_tune_concave_molding:device:main:service:mosquitto-az-bridge","@parent":"device/main//","@type":"service","name":"mosquitto-az-bridge","type":"service"}
[c8y/s/us] 102,TST_tune_concave_molding:device:main:service:mosquitto-az-bridge,service,mosquitto-az-bridge,up
```
The last message should be `down`. This is the issue 2).
3. Publish the status up message.
```
tedge mqtt pub -r te/device/main/service/mosquitto-az-bridge/status/health 1
```
Then you see nothing in the cosole `tedge mqtt sub "#"`. This is the issue 1). 102 message should be created for `mosquitto-az-bridge`.

**Expected behavior**
1) At least other bridges (az, aws, ...) status updates messages from mosquitto should be converted to either SmartREST 102 or 104, and reported to c8y. 
2) If the payload of the first bridge status message arrived in c8y-mapper is `0`, it should be translated to "down". So, `102` service creation message should have the status `down`.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
[...]

PR #2903, opened 2024-05-24
fix(mqtt): Map mosquitto bridge service health status updates

albinsuresh:
## Proposed changes

- [x] Map health status updates from mosquitto bridge services
- [x] Avoid redundant service creation messages while mapping the health status updates
- [x] Ensure that the initial health status is propagated as-is without assuming it to be `up`

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2902

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -359,11 +359,6 @@ impl CumulocityConverter {
                 let ancestors_external_ids =
                     self.entity_store.ancestors_external_ids(entity_topic_id)?;
 
-                if self.config.bridge_in_mapper && entity_topic_id.is_bridge_health_topic() {
-                    // Skip service creation for the mapper-inbuilt bridge, as it is part of the mapper service itself
-                    return Ok(vec![]);
-                }
-
                 let service_creation_message = service_creation_message(
                     external_id.as_ref(),
                     display_name.unwrap_or_else(|| {
@@ -1007,8 +1002,8 @@ impl CumulocityConverter {
                     }
 
                     let auto_registered_entities = self.try_auto_register_entity(&source)?;
-                    converted_messages =
-                        self.try_convert_auto_registered_entities(auto_registered_entities)?;
+                    converted_messages = self
+                        .try_convert_auto_registered_entities(auto_registered_entities, &channel)?;
                 }
 
                 let result = self
@@ -1075,11 +1070,13 @@ impl CumulocityConverter {
     fn try_convert_auto_registered_entities(
         &mut self,
         entities: Vec<EntityRegistrationMessage>,
+        channel: &Channel,
     ) -> Result<Vec<MqttMessage>, ConversionError> {
         let mut converted_messages: Vec<MqttMessage> = vec![];
         for entity in entities {
             // Append the entity registration message itself and its converted c8y form
-            converted_messages.append(&mut self.try_convert_auto_registered_entity(&entity)?);
+            converted_messages
+                .append(&mut self.try_convert_auto_registered_entity(&entity, channel)?);
         }
         Ok(converted_messages)
     }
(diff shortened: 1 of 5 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -359,11 +359,6 @@ impl CumulocityConverter {
                 let ancestors_external_ids =
                     self.entity_store.ancestors_external_ids(entity_topic_id)?;
 
-                if self.config.bridge_in_mapper && entity_topic_id.is_bridge_health_topic() {
-                    // Skip service creation for the mapper-inbuilt bridge, as it is part of the mapper service itself
-                    return Ok(vec![]);
-                }
-
```
albinsuresh, line 366: Replaced this with a better logic higher up the stack on the auto-registration path, which was the only path where this logic was relevant.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -17,6 +26,28 @@ fn default_status() -> String {
     "unknown".to_string()
 }
 
+impl HealthStatus {
+    fn from_mosquitto_bridge_payload_str(payload: &str) -> Self {
+        let status = match payload {
+            MQTT_BRIDGE_UP_PAYLOAD => UP_STATUS,
+            MQTT_BRIDGE_DOWN_PAYLOAD => DOWN_STATUS,
+            _ => UNKNOWN_STATUS,
+        };
```
didier-wenzek, line 35: This logic should be moved into `tedge_api`. It would even be better to have a function taking both the topic and the payload. Indeed, with the current API (a bunch of constants), this is up to the caller to figure out that this is a bridge health topic and that the payload has to be decoded differently).

didier-wenzek (commented):
The new system tests related to bridge health status are clear and extensive enough to provide the confidence that this PR fix the recently-discovered issues.

However, the code related to health status is still fragile because unstructured.
- There are two similar `HealthStatus` structs defined by c8y related crates (`c8y_api` and `c8y_mapper_ext`). 
- The respective roles of `tedge_api` and `c8y_api` is not clear.  Several methods should be moved into  `tedge_api` to encapsulate all the tedge conventions. 
- This PR introduces `UP_STATUS` and `DOWN_STATUS` constants, but `"up"` and `"down"` literals are used everywhere.

Hence, I'm okay to merge this PR if the point if to have an urgent fix. But to avoid other future regressions, it would be really better to take the time to clean-up health status handling.

rina23q (approved):
I confirmed that this PR fixes the reported issues. Thanks Albin. 
I approve this state since it fixes the issues and robot tests coverage is good, although I agree with the comments from Didier about `c8y_api` vs `c8y_mapper_ext`.

albinsuresh:
> However, the code related to health status is still fragile because unstructured.
> 
> * There are two similar `HealthStatus` structs defined by c8y related crates (`c8y_api` and `c8y_mapper_ext`).
> * The respective roles of `tedge_api` and `c8y_api` is not clear.  Several methods should be moved into  `tedge_api` to encapsulate all the tedge conventions.
> * This PR introduces `UP_STATUS` and `DOWN_STATUS` constants, but `"up"` and `"down"` literals are used everywhere.

I completely agree and this was the cleanup that I wanted to do as well. But, kept the minimal changes in this PR because of the urgency in fixing the bug. Will definitely do a follow-up PR with all that cleanup.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
albinsuresh: Yes, a comprehensive cleanup of the currently fragmented health monitoring code will be done a follow-up PR as mentioned in this [comment](https://github.com/thin-edge/thin-edge.io/pull/2903#issuecomment-2133431005)

PR #2904, opened 2024-05-25
Simplify the operation handling code by replacing fragmented control flow with regular async/await - attempt #2

Bravo555:
## TODO

- [x] ~~add more tests to see which operations are blocked/executed concurrently~~ added `log_upload`, no other operations downloading from the FTS (from the perspective of the mapper)
- [x] ~~remove remaining mutexes/locks~~ replaced with ordinary sender clones
- [ ] ~~clean up operation handling code~~ will be done in next PR

## Proposed changes

This PR is a 2nd attempt at simplifying operation control flow. I found that the top-down approach in #2881 of spawning a task for every main actor message would be too tedious to get working, because to remove the lock on the converter would require to change `convert` method to `&self`, which at this moment is infeasible because of how many things it does/how many other functions it calls.

Instead, the approach taken here was a bottom-up one:
1. It was identified that the operation handling and its properties do not fit into `convert` method's assumptions:
    - `convert()` needs to convert MQTT messages in both directions without blocking
    - but operation handling can block, and it can take some time before MQTT response is ready
2. As such, some operations (not all) were pulled out of `CumulocityConverter` and placed in a newly created `OperationHandler`. The operation handler was later made to `tokio::spawn` the operation handlers. This separates the scope of operation handling from the converter, and makes it so the converter can immediately return.
3. `pending_upload_operations` and `pending_fts_download_operations` hashmaps were removed.

In contrast to the previous attempt, the stated aims, i.e.: simplifying the control flow without impacting the ability to execute multiple operations concurrently, appear to have been achieved.

cont. in _Further comments_

## Types of changes

## Paste Link to the issue

- #2880

## Checklist

#### Testing and correctness 

There was a test created, but it only tests `config_snapshot` and `log_upload` command, and it only tests if it's concurrent by checking if the download requests were created. Ideally we'd spawn operations of all types and check if they're progressing equally, but the only view into their execution is messages they send to other actors, but some don't do this at all.

https://github.com/Bravo555/thin-edge.io/blob/4f53e990e160924aec0931ba1afe771f9301195f/crates/extensions/c8y_mapper_ext/src/tests.rs#L2421-L2469

### DIFF
--- a/crates/core/tedge_api/src/mqtt_topics.rs
+++ b/crates/core/tedge_api/src/mqtt_topics.rs
@@ -771,6 +771,7 @@ impl From<&Channel> for ChannelFilter {
     }
 }
 
+#[derive(Debug, Clone)]
 pub struct IdGenerator {
     prefix: String,
 }
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -3,8 +3,6 @@ use super::converter::CumulocityConverter;
 use super::dynamic_discovery::process_inotify_events;
 use crate::converter::UploadContext;
 use crate::converter::UploadOperationLog;
-use crate::operations::FtsDownloadOperationData;
-use crate::operations::FtsDownloadOperationType;
 use crate::service_monitor::is_c8y_bridge_established;
 use async_trait::async_trait;
 use c8y_api::smartrest::smartrest_serializer::fail_operation;
@@ -22,6 +20,7 @@ use std::time::Duration;
 use tedge_actors::fan_in_message_type;
 use tedge_actors::Actor;
 use tedge_actors::Builder;
+use tedge_actors::ClientMessageBox;
 use tedge_actors::CloneSender;
 use tedge_actors::DynSender;
 use tedge_actors::LoggingSender;
@@ -81,7 +80,7 @@ impl From<PublishMessage> for MqttMessage {
     }
 }
 
-fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, IdUploadResult, IdDownloadResult, PublishMessage] : Debug);
+fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, IdUploadResult, PublishMessage] : Debug);
 type C8yMapperOutput = MqttMessage;
 
 pub struct C8yMapperActor {
@@ -92,7 +91,6 @@ pub struct C8yMapperActor {
     bridge_status_messages: SimpleMessageBox<MqttMessage, MqttMessage>,
     message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
 
-    pending_fts_download_operations: Arc<Mutex<HashMap<CmdId, FtsDownloadOperationData>>>,
     pending_upload_operations: Arc<Mutex<HashMap<CmdId, UploadContext>>>,
 }
 
@@ -140,9 +138,6 @@ impl Actor for C8yMapperActor {
                 C8yMapperInput::IdUploadResult((cmd_id, result)) => {
                     self.process_upload_result(cmd_id, result).await?;
                 }
-                C8yMapperInput::IdDownloadResult((cmd_id, result)) => {
-                    self.process_download_result(cmd_id, result).await?;
-                }
                 C8yMapperInput::PublishMessage(message) => {
                     self.mqtt_publisher.send(message.0).await?;
                 }
@@ -160,7 +155,6 @@ impl C8yMapperActor {
         timer_sender: LoggingSender<SyncStart>,
         bridge_status_messages: SimpleMessageBox<MqttMessage, MqttMessage>,
         message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
-        pending_fts_download_operations: Arc<Mutex<HashMap<CmdId, FtsDownloadOperationData>>>,
         pending_upload_operations: Arc<Mutex<HashMap<CmdId, UploadContext>>>,
     ) -> Self {
         Self {
@@ -170,7 +164,6 @@ impl C8yMapperActor {
             timer_sender,
             bridge_status_messages,
             message_handlers,
-            pending_fts_download_operations,
             pending_upload_operations,
         }
     }
@@ -412,51 +405,6 @@ impl C8yMapperActor {
             Err(err) => fail_operation(operation, &format!("Upload failed with {err}")),
         }
     }
-
-    async fn process_download_result(
-        &mut self,
-        cmd_id: CmdId,
-        result: DownloadResult,
-    ) -> Result<(), RuntimeError> {
-        // download not from c8y_proxy, check if it was from FTS
-        let fts_download = self
-            .pending_fts_download_operations
-            .lock()
-            .await
-            .remove(&cmd_id);
-
-        let operation_result = if let Some(fts_download) = fts_download {
-            let cmd_id = cmd_id.clone();
-            match fts_download.download_type {
-                FtsDownloadOperationType::ConfigDownload => {
-                    self.converter
-                        .handle_fts_config_download_result(cmd_id, result, fts_download)
-                        .await
-                }
-                FtsDownloadOperationType::LogDownload => {
-                    self.converter
-                        .handle_fts_log_download_result(cmd_id, result, fts_download)
-                        .await
-                }
-            }
-        } else {
-            error!("Received a download result for the unknown command ID: {cmd_id}");
-            return Ok(());
-        };
-
-        match operation_result {
-            Ok(converted_messages) => {
-                for converted_message in converted_messages.into_iter() {
-                    self.mqtt_publisher.send(converted_message).await?
-                }
-            }
-            Err(err) => {
-                error!("Error occurred while processing a download result. {err}")
-            }
-        }
-
-        Ok(())
-    }
 }
 
 pub struct C8yMapperBuilder {
@@ -466,7 +414,7 @@ pub struct C8yMapperBuilder {
     http_proxy: C8YHttpProxy,
     timer_sender: DynSender<SyncStart>,
     upload_sender: DynSender<IdUploadRequest>,
-    download_sender: DynSender<IdDownloadRequest>,
+    downloader: ClientMessageBox<IdDownloadRequest, IdDownloadResult>,
     auth_proxy: ProxyUrlGenerator,
     bridge_monitor_builder: SimpleMessageBoxBuilder<MqttMessage, MqttMessage>,
     message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
@@ -493,8 +441,11 @@ impl C8yMapperBuilder {
         mqtt.connect_sink(config.topics.clone(), &box_builder.get_sender());
         let http_proxy = C8YHttpProxy::new(http);
         let timer_sender = timer.connect_client(box_builder.get_sender().sender_clone());
+
         let upload_sender = uploader.connect_client(box_builder.get_sender().sender_clone());
-        let download_sender = downloader.connect_client(box_builder.get_sender().sender_clone());
+
+        let downloader = ClientMessageBox::new(downloader);
+
         fs_watcher.connect_sink(
             config.ops_dir.as_std_path().to_path_buf(),
             &box_builder.get_sender(),
@@ -522,7 +473,7 @@ impl C8yMapperBuilder {
             http_proxy,
             timer_sender,
             upload_sender,
-            download_sender,
+            downloader,
             auth_proxy,
             bridge_monitor_builder,
             message_handlers,
@@ -572,10 +523,7 @@ impl Builder<C8yMapperActor> for C8yMapperBuilder {
         let timer_sender = LoggingSender::new("C8yMapper => Timer".into(), self.timer_sender);
         let uploader_sender =
             LoggingSender::new("C8yMapper => Uploader".into(), self.upload_sender);
-        let downloader_sender =
-            LoggingSender::new("C8yMapper => Downloader".into(), self.download_sender);
 
-        let pending_fts_download_operations = Arc::new(Mutex::new(HashMap::new()));
         let pending_upload_operations = Arc::new(Mutex::new(HashMap::new()));
 
         let converter = CumulocityConverter::new(
@@ -584,8 +532,7 @@ impl Builder<C8yMapperActor> for C8yMapperBuilder {
             self.http_proxy,
             self.auth_proxy,
             uploader_sender.clone(),
-            downloader_sender.clone(),
-            Arc::clone(&pending_fts_download_operations),
+            self.downloader,
             Arc::clone(&pending_upload_operations),
         )
         .map_err(|err| RuntimeError::ActorError(Box::new(err)))?;
@@ -600,7 +547,6 @@ impl Builder<C8yMapperActor> for C8yMapperBuilder {
             timer_sender,
             bridge_monitor_box,
             self.message_handlers,
-            pending_fts_download_operations,
             pending_upload_operations,
         ))
     }
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -70,6 +70,7 @@ impl From<PublishMessage> for MqttMessage {
 }
 
 fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, PublishMessage] : Debug);
+
 type C8yMapperOutput = MqttMessage;
 
 pub struct C8yMapperActor {
@@ -331,7 +332,7 @@ pub struct C8yMapperBuilder {
     http_proxy: C8YHttpProxy,
     timer_sender: DynSender<SyncStart>,
     downloader: ClientMessageBox<IdDownloadRequest, IdDownloadResult>,
-    uploader: ClientMessageBox<(String, UploadRequest), (String, UploadResult)>,
+    uploader: ClientMessageBox<IdUploadRequest, IdUploadResult>,
     auth_proxy: ProxyUrlGenerator,
     bridge_monitor_builder: SimpleMessageBoxBuilder<MqttMessage, MqttMessage>,
     message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -1,21 +1,14 @@
 use super::config::C8yMapperConfig;
 use super::converter::CumulocityConverter;
 use super::dynamic_discovery::process_inotify_events;
-use crate::converter::UploadContext;
-use crate::converter::UploadOperationLog;
 use crate::service_monitor::is_c8y_bridge_established;
 use async_trait::async_trait;
-use c8y_api::smartrest::smartrest_serializer::fail_operation;
-use c8y_api::smartrest::smartrest_serializer::succeed_static_operation;
-use c8y_api::smartrest::smartrest_serializer::CumulocitySupportedOperations;
-use c8y_api::smartrest::smartrest_serializer::SmartRest;
 use c8y_auth_proxy::url::ProxyUrlGenerator;
 use c8y_http_proxy::handle::C8YHttpProxy;
 use c8y_http_proxy::messages::C8YRestRequest;
 use c8y_http_proxy::messages::C8YRestResult;
 use std::collections::HashMap;
 use std::path::PathBuf;
-use std::sync::Arc;
 use std::time::Duration;
 use tedge_actors::fan_in_message_type;
 use tedge_actors::Actor;
@@ -42,7 +35,6 @@ use tedge_downloader_ext::DownloadRequest;
 use tedge_downloader_ext::DownloadResult;
 use tedge_file_system_ext::FsWatchEvent;
 use tedge_mqtt_ext::MqttMessage;
-use tedge_mqtt_ext::QoS;
 use tedge_mqtt_ext::TopicFilter;
 use tedge_timer_ext::SetTimeout;
 use tedge_timer_ext::Timeout;
@@ -50,9 +42,6 @@ use tedge_uploader_ext::UploadRequest;
 use tedge_uploader_ext::UploadResult;
 use tedge_utils::file::create_directory_with_defaults;
 use tedge_utils::file::FileError;
-use tokio::sync::Mutex;
-use tracing::error;
-use tracing::warn;
 
 const SYNC_WINDOW: Duration = Duration::from_secs(3);
 
@@ -80,7 +69,7 @@ impl From<PublishMessage> for MqttMessage {
     }
 }
 
-fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, IdUploadResult, PublishMessage] : Debug);
+fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, PublishMessage] : Debug);
 type C8yMapperOutput = MqttMessage;
 
 pub struct C8yMapperActor {
@@ -90,8 +79,6 @@ pub struct C8yMapperActor {
     timer_sender: LoggingSender<SyncStart>,
     bridge_status_messages: SimpleMessageBox<MqttMessage, MqttMessage>,
     message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
-
-    pending_upload_operations: Arc<Mutex<HashMap<CmdId, UploadContext>>>,
 }
 
 #[async_trait]
@@ -135,9 +122,6 @@ impl Actor for C8yMapperActor {
                 C8yMapperInput::SyncComplete(_) => {
                     self.process_sync_timeout().await?;
                 }
-                C8yMapperInput::IdUploadResult((cmd_id, result)) => {
-                    self.process_upload_result(cmd_id, result).await?;
-                }
                 C8yMapperInput::PublishMessage(message) => {
                     self.mqtt_publisher.send(message.0).await?;
                 }
@@ -155,7 +139,6 @@ impl C8yMapperActor {
         timer_sender: LoggingSender<SyncStart>,
         bridge_status_messages: SimpleMessageBox<MqttMessage, MqttMessage>,
         message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
-        pending_upload_operations: Arc<Mutex<HashMap<CmdId, UploadContext>>>,
     ) -> Self {
         Self {
             converter,
@@ -164,7 +147,6 @@ impl C8yMapperActor {
             timer_sender,
             bridge_status_messages,
             message_handlers,
-            pending_upload_operations,
         }
     }
 
@@ -340,71 +322,6 @@ impl C8yMapperActor {
 
         Ok(())
     }
-
-    async fn process_upload_result(
-        &mut self,
-        cmd_id: CmdId,
-        upload_result: UploadResult,
-    ) -> Result<(), RuntimeError> {
-        let pending_upload = self.pending_upload_operations.lock().await.remove(&cmd_id);
-        match pending_upload {
-            Some(UploadContext::OperationData(queued_data)) => {
-                let payload = match queued_data.operation {
-                    CumulocitySupportedOperations::C8yLogFileRequest
-                    | CumulocitySupportedOperations::C8yUploadConfigFile => self
-                        .get_smartrest_response_for_upload_result(
-                            upload_result,
-                            &queued_data.c8y_binary_url,
-                            queued_data.operation,
-                        ),
-                    other_type => {
-                        warn!("Received unsupported operation {other_type:?} for uploading a file");
-                        return Ok(());
-                    }
-                };
-
-                let c8y_notification = MqttMessage::new(&queued_data.smartrest_topic, payload);
-                let clear_local_cmd = MqttMessage::new(&queued_data.clear_cmd_topic, "")
-                    .with_retain()
-                    .with_qos(QoS::AtLeastOnce);
-
-                let messages = self
-                    .converter
-                    .upload_operation_log(
-                        &queued_data.topic_id,
-                        &cmd_id,
-                        &queued_data.operation.into(),
-                        queued_data.command,
-                        vec![c8y_notification, clear_local_cmd],
-                    )
-                    .await;
-                for message in messages {
-                    self.mqtt_publisher.send(message).await?
-                }
-            }
-            Some(UploadContext::OperationLog(UploadOperationLog { final_messages })) => {
-                for message in final_messages {
-                    self.mqtt_publisher.send(message).await?
-                }
-                return Ok(());
-            }
-            None => error!("Received an upload result for the unknown command ID: {cmd_id}"),
-        }
-
-        Ok(())
-    }
-
-    fn get_smartrest_response_for_upload_result(
-        &self,
-        upload_result: UploadResult,
-        binary_url: &str,
-        operation: CumulocitySupportedOperations,
-    ) -> SmartRest {
-        match upload_result {
-            Ok(_) => succeed_static_operation(operation, Some(binary_url)),
-            Err(err) => fail_operation(operation, &format!("Upload failed with {err}")),
-        }
-    }
 }
 
 pub struct C8yMapperBuilder {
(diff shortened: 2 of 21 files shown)


crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
```diff
@@ -135,39 +78,88 @@ impl CumulocityConverter {
...
+                    Some(ref tedge_file_url) => Cow::Borrowed(tedge_file_url),
+                    None => {
+                        let tedge_file_url = format!(
+                        "http://{}/tedge/file-transfer/{external_id}/config_snapshot/{config_filename}",
+                        &self.tedge_http_host,
+                        external_id = target.external_id.as_ref()
+                    );
+                        Cow::Owned(tedge_file_url)
+                    }
+                };
```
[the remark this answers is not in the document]
Bravo555: I probably shouldn't be doing any functional changes in a refactoring PR regardless, so I will remove this before merging, but I first wanted to ask if this is something that should be fixed.

Bravo555:
After rerunning the tests locally, I can observe that the mapper correctly sends `502` and `503` smartrest messages, but operations status asserts are hitting timeouts nonetheless. Could this be a problem with Cumulocity tenant?

didier-wenzek:
> After rerunning the tests locally, I can observe that the mapper correctly sends `502` and `503` smartrest messages, but operations status asserts are hitting timeouts nonetheless. Could this be a problem with Cumulocity tenant?

I would be surprised by that. The tests are green on other PRs and the errors are related to this PR's changes.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2418,6 +2418,72 @@ async fn c8y_mapper_nested_child_service_event_mapping_to_smartrest() {
...
+            .await.unwrap();
+    }
+
+    for _ in 0..(num_operations * 2) {
+        dl.recv()
+            .await
+            .expect("there should be one DownloadRequest per operation");
+    }
+}
+
```
albinsuresh, line 2486: To validate the effectiveness of this work, one critical test that we really need to add, is to validate that a new request is still processed and mapped while a file upload, in response to a terminal status update of another client, is in-progress.

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
```diff
@@ -189,60 +181,62 @@ impl CumulocityConverter {
...
-        let response = &ConfigSnapshotCmdPayload::from_json(payload)?;
-
-        let download = match download_result {
-            Err(err) => {
-                let smartrest_error =
-                    fail_operation(
-                        CumulocitySupportedOperations::C8yUploadConfigFile,
-                        &format!("tedge-mapper-c8y failed to download configuration snapshot from file-transfer service: {err}"),
-                    );
+pub fn topic_filter(mqtt_schema: &MqttSchema) -> TopicFilter {
```
albinsuresh, line 186: Minor: It would have been better to leave this function and the next `impl CumulocityConverter` blocks in their former location (at the top) would have reduced the diff volume and better highlighted the changed bits. The **request** mapping logic appearing first looks more intuitive as well.

crates/extensions/c8y_mapper_ext/src/operations/firmware_update.rs
```diff
@@ -162,6 +116,56 @@ impl CumulocityConverter {
             Some(command.into_generic_command(&self.mqtt_schema)),
         ))
     }
+}
+
+pub fn firmware_update_topic_filter(mqtt_schema: &MqttSchema) -> TopicFilter {
```
albinsuresh, line 121: Minor: Same comment as in `config_snaphot.rs`. Better to restore this function and the next impl block to its former location to reduce the diff size, and to better highlight the changed bits.

crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs
```diff
@@ -184,69 +170,49 @@ impl CumulocityConverter {
...
-
-        let download_response = match download_result {
-            Err(err) => {
-                let smartrest_error = fail_operation(
-                    CumulocitySupportedOperations::C8yLogFileRequest,
-                    &format!(
-                        "tedge-mapper-c8y failed to download log from file transfer service: {err}",
-                    ),
-                );
+impl CumulocityConverter {
```
albinsuresh, line 175: Minor: Same comment as in config_snaphot.rs. Better to restore this function and the next impl block to its former location to reduce the diff size, and to better highlight the changed bits.

albinsuresh (commented):
Great effort. It was nice to see the `OperationHandler` abstraction with only the stuff that really needs to be shared for operation handling. And I really liked that subtle trick you used with the `Entity` abstraction, eliminating the need for sharing the entity_store as well.

I'm still a bit confused to see the locking required around the proxy. I was under the impression that it wouldn't be needed anymore. But still, it's good to see that the locking is limited to the upload path and it doesn't influence the processing of any other parallel requests.

crates/extensions/c8y_mapper_ext/src/operations/upload.rs
```diff
@@ -0,0 +1,105 @@
...
+    ) -> Result<(Url, UploadResult), ConversionError> {
+        let create_event = CreateEvent {
+            event_type: event_type.clone(),
+            time: OffsetDateTime::now_utc(),
+            text: event_text.unwrap_or(event_type),
+            extras: HashMap::new(),
+            device_id: external_id.into(),
+        };
+
+        let mut c8y_http_proxy = self.http_proxy.lock().await.clone();
```
didier-wenzek, line 40: As `C8YHttpProxy` the use of `Arc<Mutex<C8YHttpProxy>>` can be simply removed.

```suggestion
        let mut c8y_http_proxy = self.http_proxy.clone();
```

didier-wenzek (commented):
I will be really happy to merge this PR!

However, I see this work as half-done:

- *All* the operations must be handled by the `OperationHandler`.
- This is an opportunity to move the code for `Restart`, `SoftwareList` and `SoftwareUpdate` into their own modules under `operations`.

crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs
```diff
@@ -129,40 +80,75 @@ impl CumulocityConverter {
                 // Send a request to the Downloader to download the file asynchronously from FTS
                 let log_filename = format!("{}-{}", command.payload.log_type, cmd_id);
 
-                let tedge_file_url = format!(
-                    "http://{}/tedge/file-transfer/{external_id}/log_upload/{log_filename}",
-                    &self.config.tedge_http_host,
-                    external_id = target.external_id.as_ref()
-                );
+                let tedge_file_url = &command.payload.tedge_url;
```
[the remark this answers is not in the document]
Bravo555: You mean do the fallback if the string is empty? In the other operation it's an `Option<String>`, would it be a better fit here?

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -13,47 +13,152 @@
...
 use tracing::error;
 
 pub mod config_snapshot;
 pub mod config_update;
 pub mod firmware_update;
 pub mod log_upload;
+mod upload;
 
-/// Represents a pending download performed by the downloader from the FTS.
+/// Handles non-generic operations.
```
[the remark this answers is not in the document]
Bravo555: Admittedly I may have been a bit confused when I wrote that. I meant handling concrete operations like `config_snapshot`, `log_upload`, etc. I may have thought User-defined Operation Workflows to be like _generic operations_, but they really aren't, once I think about it the second time. Also, I may have been a bit confused by `GenericCommandState` and `into_generic_command()`, but operations are subsets of commands which invoke some extra behaviour, so also not really applicable here.

So I think I'll just change that `OperationHandler` doc summary line to "Handles operations." for the ultimate "you don't say?" reading experience.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -310,8 +190,8 @@ pub struct C8yMapperBuilder {
     mqtt_publisher: DynSender<MqttMessage>,
     http_proxy: C8YHttpProxy,
     timer_sender: DynSender<SyncStart>,
-    upload_sender: DynSender<IdUploadRequest>,
-    download_sender: DynSender<IdDownloadRequest>,
+    downloader: ClientMessageBox<IdDownloadRequest, IdDownloadResult>,
+    uploader: ClientMessageBox<(String, UploadRequest), (String, UploadResult)>,
```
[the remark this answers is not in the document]
Bravo555: fixed in f72e3c4f

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -108,12 +99,8 @@ impl Actor for C8yMapperActor {
...
                     self.process_sync_timeout().await?;
                 }
-                C8yMapperInput::IdUploadResult((cmd_id, result)) => {
-                    self.process_upload_result(cmd_id, result).await?;
-                }
-                C8yMapperInput::IdDownloadResult((cmd_id, result)) => {
-                    self.process_download_result(cmd_id, result).await?;
-                }
+                C8yMapperInput::IdUploadResult(_) => {}
+                C8yMapperInput::IdDownloadResult(_) => {}
```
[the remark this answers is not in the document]
Bravo555: fixed in f72e3c4f

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -298,6 +298,9 @@ pub enum C8yMapperConfigBuildError {
 
     #[error(transparent)]
     FromTopicIdError(#[from] TopicIdError),
+
+    #[error(transparent)]
+    OtherError(#[from] anyhow::Error),
```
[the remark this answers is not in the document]
Bravo555: Turns out I needed it earlier but later out it cancelled out with other changes, so I was able to remove it.
fixed in f72e3c4f

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -297,6 +289,23 @@ impl CumulocityConverter {
...
+            mqtt_schema: mqtt_schema.clone(),
+            c8y_prefix: prefix.clone(),
+            mqtt_publisher: mqtt_publisher.clone(),
+
+            downloader,
+            uploader,
+
+            c8y_endpoint: c8y_endpoint.clone(),
+            auth_proxy: auth_proxy.clone(),
+            http_proxy: Arc::new(Mutex::new(http_proxy.clone())),
```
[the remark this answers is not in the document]
Bravo555: fixed in f72e3c4f

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
```diff
@@ -629,6 +623,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[ignore = "TODO: replace with mock http"]
```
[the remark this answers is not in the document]
Bravo555: My bad, this was a leftover I forgot to remove from an earlier approach. After enabling it back it also showed that I indeed broke the log upload, which is probably why integration tests were faling 🤦. Should be fixed now.
fixed in 7dc2f4e

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
```diff
@@ -135,39 +78,88 @@ impl CumulocityConverter {
...
+                        return Ok((vec![c8y_notification, clean_operation], None));
+                    }
+                    Ok(download) => download,
+                };
+
+                let file_path =
+                    Utf8PathBuf::try_from(destination_path).map_err(|e| e.into_io_error())?;
+                let event_type = command.payload.config_type.clone();
+
+                // Create an event in c8y
```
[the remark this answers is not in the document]
Bravo555: fixed in f72e3c4f

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -13,47 +13,152 @@
...
+#[derive(Clone)]
+pub struct OperationHandler {
+    pub capabilities: Capabilities,
+    pub auto_log_upload: AutoLogUpload,
+    pub tedge_http_host: Arc<str>,
+    pub tmp_dir: Arc<Utf8Path>,
+    pub mqtt_schema: MqttSchema,
+    pub c8y_prefix: tedge_config::TopicPrefix,
+
+    pub http_proxy: Arc<Mutex<C8YHttpProxy>>,
```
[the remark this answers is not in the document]
Bravo555: fixed in f72e3c4f

crates/extensions/c8y_mapper_ext/src/operations/upload.rs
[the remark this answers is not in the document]
Bravo555: My bad, missed that one.
fixed in f72e3c4f

Bravo555: fixed in f72e3c4f

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
Bravo555: Indeed, the diff looks much easier to review. I'll remember not to shuffle the code next time for easier review.
fixed in b8bd269

crates/extensions/c8y_mapper_ext/src/operations/firmware_update.rs
Bravo555: Indeed, the diff looks much easier to review. I'll remember not to shuffle the code next time for easier review.
fixed in b8bd269

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -13,47 +13,152 @@
...
-    // the message that triggered the operation
-    pub message: MqttMessage,
-
-    pub entity_topic_id: EntityTopicId,
+/// Handling an operation usually consists of 3 steps:
+///
+/// 1. Receive a smartrest message which is an operation request, convert it to thin-edge message,
+///    and publish on local MQTT (done by the converter).
+/// 2. Various local thin-edge components/services execute the operation, and when they're done,
+///    they publish an MQTT message with 'status: successful/error'
```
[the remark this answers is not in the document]
Bravo555: fixed in f72e3c4f1133849f128439c3b33a8e91291209fb

crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs
Bravo555: Indeed, the diff looks much easier to review. I'll remember not to shuffle the code next time for easier review.
fixed in b8bd269

Bravo555:
Because proposed changes currently spawn a new task for every operation MQTT message, and because tasks can run and finish in a different order, we can get a situation where we receive MQTT messages 1. "executing" 2. "failed", but tasks can output 1. "502" (failed) 2. "501" (executing) which is fatally wrong.

This would have to be solved by either:
1. spawning a task that manages the entire lifecycle of a single operation, additional MQTT messages are routed to it, the task consumes it one message at a time preserving order
2. having current task-per-message tasks derive the current state of the operation not from a message they just received, but from some piece of state, that requires sequential access to update operation state.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -13,47 +13,270 @@
...
+                let handler = self.clone();
+                let running_operation = RunningOperation::spawn(message, handler);
+
+                lock.insert(topic, running_operation);
+                None
+            }
+        };
+
+        if let Some(running_operation) = running_operation {
+            let join_result = running_operation.handle.await;
```
albinsuresh, line 148: This could be my misunderstanding of the `await` contract on a tokio `JoinHandle`, but I'm wondering how that `handle.await` does not block the converter until the spawned task is complete.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -13,47 +13,270 @@
...
+
+    pub http_proxy: C8YHttpProxy,
+    pub c8y_endpoint: C8yEndPoint,
+    pub auth_proxy: ProxyUrlGenerator,
+
+    pub downloader: ClientMessageBox<IdDownloadRequest, IdDownloadResult>,
+    pub uploader: ClientMessageBox<IdUploadRequest, IdUploadResult>,
+    pub mqtt_publisher: LoggingSender<MqttMessage>,
+
+    pub running_operations: Arc<Mutex<HashMap<Arc<str>, RunningOperation>>>,
```
albinsuresh, line 91: Why was this map required at this `OperationHandler` level? The `RunningOperation` has no use of it. From what I understand, this was only required at the converter level for book-keeping the spawned tasks and the message routing. I feel that even the message routing to these tasks could be moved into the converter level so that we can avoid mutable state here and avoid concurrent accesses to it as well.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -13,47 +13,270 @@
...
+/// A subset of entity-related information necessary to handle an operation.
 ///
-/// Used to dispatch download result to the correct operation handler.
-pub enum FtsDownloadOperationType {
-    LogDownload,
-    ConfigDownload,
+/// Because the operation may take time and other operations may run concurrently, we don't want to
+/// query the entity store.
+#[derive(Clone, Debug)]
+pub struct Entity {
```
albinsuresh, line 276:
```suggestion
pub struct EntityTarget {
```
or `EntityInfo` or anything similar to avoid any confusion with the general concept of entity.

albinsuresh (commented):
The overall refactoring looks really good. I think we are very close to getting this merged. I just have some concerns regarding the usage of that shared mutable map of the spawned operation tasks.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
Bravo555: > Why was this map required at this OperationHandler level? The RunningOperation has no use of it.

`RunningOperation` doesn't use it, but it uses other fields of `OperationHandler`. They should ideally be in a separate object, but passing all the properties from `CumulocityConverter` is a pain. Before finalizing the PR, I'll try one approach I had in mind to simplify this.

> From what I understand, this was only required at the converter level for book-keeping the spawned tasks and the message routing.

Yes, the idea is that we have to route messages received via MQTT into the task that handle an operation with a given id. But I don't understand what you mean when you say "it was required at the converter level". Do you mean it was required in the converter before my PR, or do you mean something else?

> I feel that even the message routing to these tasks could be moved into the converter level so that we can avoid mutable concurrent accesses to it as well.

I wanted to hide from the converter how exactly the messages get passed into tasks, and only expose a simple interface - the converter only calls `handle()` and the rest of operation handling happens inside this module. So I'm not really sure what you mean by "move routing into the converter level" and what other levels we're talking about.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -13,47 +13,270 @@
...
+                lock.insert(topic, running_operation);
+                None
+            }
+        };
+
+        if let Some(running_operation) = running_operation {
+            let join_result = running_operation.handle.await;
+            error!("handle: {join_result:?}");
+        }
+    }
```
didier-wenzek, line 174: The name `running_operation` is confusing. Yes, there was a running operation, but this is operation is no more consuming messages (`running_operation.tx.send(message).is_err()`).

```suggestion
        if let Some(terminated_operation) = terminated_operation {
            let join_result = terminated_operation.handle.await;
            error!("handle: {join_result:?}");
        }
    }
```

By the way, will the entry in `self.running_operations` for a topic be cleared, if no more message is received for that topic?

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
Bravo555: > RunningOperation doesn't use it, but it uses other fields of OperationHandler. They should ideally be in a separate object, but passing all the properties from CumulocityConverter is a pain. Before finalizing the PR, I'll try one approach I had in mind to simplify this.

The handler was cleaned up a lot in c4e27504de95d47aa8a3ac598eeec2397e4d74b5, mutex on `running_operations` was removed.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -12,48 +12,313 @@
...
+    handle: tokio::task::JoinHandle<()>,
+    tx: tokio::sync::mpsc::UnboundedSender<OperationMessage>,
+}
+
+impl RunningOperation {
+    /// Spawns a task that handles the operation.
+    ///
+    /// The task handles a single operation with a given command id, and via a channel it receives
+    /// operation state changes (if any) to drive an operation to completion.
+    fn spawn(message: OperationMessage, context: Arc<OperationContext>) -> Self {
```
didier-wenzek, line 187: Okay this design is correct. I'm not fully happy with it, though. That said, I acknowledge that the complexity is nicely grouped in one place and that all the operation *managers* (i.e. the misc `context.handle_XXX`) are fully unaware of it.

=> I will take more time to clarify my thoughts on that.

didier-wenzek (commented):
Nice work. I still have some details to look closer. Will do asap.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
albinsuresh: > I wanted to hide from the converter how exactly the messages get passed into tasks, and only expose a simple interface - the converter only calls handle() and the rest of operation handling happens inside this module. So I'm not really sure what you mean by "move routing into the converter level" and what other levels we're talking about.

Based on that description of how you wanted to abstract the `OperationHandler` without overloading the converter, I see 3 different levels here:
1. Converter: Just converts messages and passes it to operation handler
2. Operation handler: Maintains the map of running operations and routes messages to them
3. The configs and handles (the proxy, message boxes, senders etc) required by the `RunningOperation` to handle individual operations

I think you've combined both 2 and 3 into that single `OperationHandler` struct which is causing this confusion. If they were separated out, then the container can just have a single `OperationHandler` instance in it, with no `Arc<Mutex>>` guard required over it and the `RunningOperation` instances can just use clones of the rest of that struct, without no shared mutable access required between them either.

> RunningOperation doesn't use it, but it uses other fields of `OperationHandler`. They should ideally be in a separate object, but passing all the properties from CumulocityConverter is a pain. Before finalizing the PR, I'll try one approach I had in mind to simplify this.

Yes, I think that's the way to go.

> But I don't understand what you mean when you say "it was required at the converter level". Do you mean it was required in the converter before my PR, or do you mean something else?

I was proposing to move that `running_operations` maintenance responsibility into the converter, but your proposal to keep those with the `OperationHandler` struct sounds better, as long as the rest of the shared fields are moved out of it.

albinsuresh: > The handler was cleaned up a lot in https://github.com/thin-edge/thin-edge.io/commit/c4e27504de95d47aa8a3ac598eeec2397e4d74b5, mutex on running_operations was removed.

Ignore my previous comment. I guess I was replying to an older stale view of your PR, and I missed this comment. From a quick look, extracting the `OperationContext` abstraction from `OperationHandler` was exactly what was required and that precisely addresses my concerns.

didier-wenzek:
The test "Publish Connection Closed after publish and no error message with TLS"" is failing but for reasons unrelated to this PR. See https://github.com/thin-edge/thin-edge.io/pull/2987 where the same test fails with no code changes.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
Bravo555: > By the way, will the entry in self.running_operations for a topic be cleared, if no more message is received for that topic?

No. I realized that currently, after the operation has terminated, the converter has to call `handle()` with one additional message, and in practice the only message that can be used is the topic clearing message, because it's published by the operation on MQTT broker, and the broker echoes clearing message back to us because we're subscribed, at which point converter can call `handle()` and only then is this task removed from the map.

So the current design is not unlike polling, where the caller (converter in this case) needs to call `.handle()` for every operation-related message for the operation to be successfully handled and freed.

This has some problems, the biggest being that we can't remove the operation immediately after it terminates, and have to rely on MQTT broker to echo back the clearing message, which shouldn't be necessary, but I think this could be addressed in the following PR.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -12,48 +12,315 @@
...
+                auto_log_upload: c8y_mapper_config.auto_log_upload,
+                tedge_http_host: c8y_mapper_config.tedge_http_host.clone(),
+                tmp_dir: c8y_mapper_config.tmp_dir.clone(),
+                mqtt_schema: c8y_mapper_config.mqtt_schema.clone(),
+                mqtt_publisher: mqtt_publisher.clone(),
+                software_management_api: c8y_mapper_config.software_management_api,
+
+                // TODO(marcel): would be good not to generate new ids from running operations, see if
+                // we can remove it somehow
+                command_id: IdGenerator::new(crate::converter::REQUESTER_NAME),
```
didier-wenzek, line 103: Agree that it would be good to have *no* `IdGenerator` at all outside the `CumulucityConverter`.

This generator is used by the `OperationContext` only in one place: to trigger a software list request after software updates. One should be able to move the call to `request_software_list` from `OperationContext::publish_software_update_status` to the `CumulucityConverter` where the outcome of the software update is forwarded to c8y.

crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs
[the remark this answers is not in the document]
albinsuresh: Yes, since `tedgeUrl` really is an optional field, representing it using an `Option` seems appropriate.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -12,48 +12,315 @@
...
+                        }
+
+                        for message in messages {
+                            // if task publishes MQTT clearing message, an operation is considered finished, so we can
+                            // terminate the task as well
+                            if message.retain
+                                && message.payload_bytes().is_empty()
+                                && message.topic == command_topic
+                            {
+                                rx.close();
```
albinsuresh, line 271: Minor: Just like the `upload_operation_log` is done above, at one place, for all operations, generating and publishing the clear messages also could be done from here, without expecting all the different operation handlers to do the same (resulting in duplicated code). But this can be deferred for a follow-up PR.

albinsuresh (commented):
A giant leap in terms of streamlining the bloated responsibilities of the converter. The code looks really good and consider this my stamp of approval of the refactoring.

Yes, there are still some minor niggles in how the clear messages are a handled etc. But, to avoid blocking this PR for much longer,  I'm happy to approve this PR in its current state, as the improvements made far outweighs such minor niggles. Will give the "formal" approval once Didier's comments are also addressed.

crates/extensions/c8y_mapper_ext/src/operations/config_snapshot.rs
```diff
@@ -135,39 +78,88 @@ impl CumulocityConverter {
...
+                    Err(err) => {
+                        println!("{err}");
+                        let smartrest_error =
+                    fail_operation(
+                        CumulocitySupportedOperations::C8yUploadConfigFile,
+                        &format!("tedge-mapper-c8y failed to download configuration snapshot from file-transfer service: {err}"),
+                    );
+
+                        let c8y_notification = MqttMessage::new(&smartrest_topic, smartrest_error);
+                        let clean_operation = MqttMessage::new(&message.topic, "")
```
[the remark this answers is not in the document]
Bravo555: will do in a follow-up: https://github.com/thin-edge/thin-edge.io/issues/2880#issuecomment-2223213567

crates/extensions/c8y_mapper_ext/src/tests.rs
Bravo555: will be done in a follow-up: https://github.com/thin-edge/thin-edge.io/issues/2880#issuecomment-2223213567

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -771,6 +771,7 @@ impl From<&Channel> for ChannelFilter {
     }
 }
 
+#[derive(Debug, Clone)]
 pub struct IdGenerator {
     prefix: String,
 }
```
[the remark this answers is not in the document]
Bravo555: will be done in a follow-up: https://github.com/thin-edge/thin-edge.io/issues/2880#issuecomment-2223213567

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
Bravo555: will be done in a follow-up: https://github.com/thin-edge/thin-edge.io/issues/2880#issuecomment-2223213567

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -12,48 +12,315 @@
...
+    /// `entity` needs to be the same entity as in `message`.
+    ///
+    /// When a message with a new id is handled, a task will be spawned that processes this
+    /// operation. For the operation to be completed, subsequent messages with the same command id
+    /// need to be handled as well.
+    ///
+    /// When an operation terminates (successfully or unsuccessfully), an MQTT operation clearing
+    /// message will be published, but it also needs to be handled when an MQTT broker echoes this
+    /// message back to us, so that `OperationHandler` can free the data associated with the
+    /// operation.
```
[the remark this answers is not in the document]
Bravo555: comment itself was changed, the other problem will be tackled in a follow-up: https://github.com/thin-edge/thin-edge.io/issues/2880#issuecomment-2223213567

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
Bravo555: will be done in a follow-up: https://github.com/thin-edge/thin-edge.io/issues/2880#issuecomment-2223213567

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
didier-wenzek: > => I will take more time to clarify my thoughts on that.

Here is my point of view about this design:

Pros:

- The complexity is located in one place. Two to be precise:
  - the `RunningOperation::spawn()` function that spawn a running operation for a specific request
  - the `OperationHandler::handle()` method that manages the running operations and dispatches status messages up to completion.
- The operation handlers are focused on the task to be done
  - the sequence of steps to handle the outcome of an operation is written in plain async/await code
  - there is no more state cache between steps
  - see `OperationContext::handle_config_snapshot_state_change` for an example
  - this  alone is a reason to merge this PR

Cons:

- The `RunningOperation::spawn()` function  mixes low level aspects with high level ones.
  - tasks, join handles and channels on one side 
  - tedge operations and log uploads on the other side
  - plus a specific behavior on command clearing messages
- The `RunningOperation::spawn()` function is akin to an `Actor::run()` method
  - consuming messages in a loop up to completion,
  - but mixed with message delivery aspects 
- Each `OperationContext::handle_*()` method is also akin to an `Actor` message handler

didier-wenzek:
> A giant leap in terms of streamlining the bloated responsibilities of the converter. The code looks really good and consider this my stamp of approval of the refactoring.

Agree. The benefits are great and I will be happy to merge this PR despite being not fully happy with the design. See https://github.com/thin-edge/thin-edge.io/pull/2904#discussion_r1674217880.

> Yes, there are still some minor niggles in how the clear messages are a handled etc. But, to avoid blocking this PR for much longer, I'm happy to approve this PR in its current state, as the improvements made far outweighs such minor niggles. Will give the "formal" approval once Didier's comments are also addressed.

I'm afraid my comments are not straightforward to address:

- As noted by @Bravo555 about the [comment related to the duplicated generator](https://github.com/thin-edge/thin-edge.io/pull/2904#discussion_r1673772928): currently the c8y converter is not responsible for sending the final outcome of a command. This is done by each operation handler. Hence, it will be a quite involve step to remove all the finalization steps out of the operation handlers.
- And to fix the [asymmetry of topic initialization and clearing](https://github.com/thin-edge/thin-edge.io/pull/2904#discussion_r1673772928), we also need to move the finalization steps out of the operation handlers.

Hence, I'm happy to address these pending comments in a following PR.

didier-wenzek (approved):
Approved. Thank you for this sustained effort to untangle the c8y converter. A great leap forward.

albinsuresh (approved):
There's some test failure, but looks unrelated.

PR #2907, opened 2024-05-29
add missing timeout to service monitoring asserts

Bravo555:
## Proposed changes

Before the change, sometimes the test would get "stuck" when running locally, because some asserts did not have a timeout. This commit adds missing timeout so that this test suite no longer gets stuck.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
+++ b/tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
@@ -76,7 +76,7 @@ Test if all c8y services using default service type when service type configured
 
 Check health status of tedge-mapper-c8y service on broker stop start
     Device Should Exist                      ${DEVICE_SN}:device:main:service:tedge-mapper-c8y    show_info=False
-    ${SERVICE}=    Cumulocity.Device Should Have Fragment Values    status\=up
+    ${SERVICE}=    Cumulocity.Device Should Have Fragment Values    status\=up    timeout=${TIMEOUT}
     Should Be Equal    ${SERVICE["name"]}    tedge-mapper-c8y
     Should Be Equal    ${SERVICE["status"]}    up
 
@@ -84,7 +84,7 @@ Check health status of tedge-mapper-c8y service on broker stop start
     ThinEdgeIO.Service Should Be Stopped  mosquitto.service
 
     Device Should Exist                      ${DEVICE_SN}:device:main:service:tedge-mapper-c8y    show_info=False
-    ${SERVICE}=    Cumulocity.Device Should Have Fragment Values    status\=down
+    ${SERVICE}=    Cumulocity.Device Should Have Fragment Values    status\=down    timeout=${TIMEOUT}
     Should Be Equal    ${SERVICE["name"]}    tedge-mapper-c8y
     Should Be Equal    ${SERVICE["status"]}    down
 
@@ -126,7 +126,7 @@ Check health status of child device service
 
     # Check created service entries
     Device Should Exist                      ${child_sn}:service:childservice    show_info=False
-    ${SERVICE}=    Device Should Have Fragment Values    status\=unknown
+    ${SERVICE}=    Device Should Have Fragment Values    status\=unknown    timeout=${TIMEOUT}
     Should Be Equal    ${SERVICE["name"]}    childservice
     Should Be Equal    ${SERVICE["serviceType"]}    service
     Should Be Equal    ${SERVICE["status"]}    unknown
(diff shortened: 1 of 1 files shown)


reubenmiller:
> Before the change, sometimes the test would get "stuck" when running locally, because some asserts did not have a timeout. This commit adds missing timeout so that this test suite no longer gets stuck.

All "smart assertions" in the testing library have a default timeout defined (30 seconds). The user can either change this globally or locally...so I'm curious if the test was really "stuck" or the timeout was just set too high (maybe left over from another test)?

albinsuresh (approved):

gligorisaev (approved):
I have run 50 test runns with the default value of the timeout, which is set to 30 seconds for all smart assertions as it is defined in the CumulocityLibrary (Cumulocity.py)


```
   # Default parameter settings
    DEFAULT_TIMEOUT = 30

    # Default HTTP request timeout
    DEFAULT_REQUEST_TIMEOUT = 30.0
```

and 30 runs with the custom value of 90 seconds


Have observed following results:

**Default Timeout (30 seconds)**

- Averege time needed for test suite to complete: **3.05**
- only once the test run failed 

The Error message:

```
KEYWORD Check if a service using configured service type   tedge-agent
Start / End / Elapsed:	20240603 10:19:10.836 / 20240603 10:19:59.131 / 00:00:48.295
00:00:00.052KEYWORD ThinEdgeIO . Execute Command   tedge config set service.type thinedge
00:00:00.035KEYWORD ThinEdgeIO . Restart Service   ${service_name}
00:00:00.249KEYWORD Cumulocity . Device Should Exist   ${DEVICE_SN}:device:main:service:${service_name}    show_info=False
Documentation:	
Deprecated: Please use "External Identity Should Exist" instead Assert that a device exists by checking its external identity

Start / End / Elapsed:	20240603 10:19:10.926 / 20240603 10:19:11.175 / 00:00:00.249
00:00:47.952KEYWORD ${SERVICE} =   Cumulocity . Device Should Have Fragment Values   status\=up    serviceType\=thinedge    timeout=${TIMEOUT}
Documentation:	
Deprecated: Use "Managed Object Should Have Fragment Values" instead Assert that a managed object contains specific fragment values.

Start / End / Elapsed:	20240603 10:19:11.177 / 20240603 10:19:59.129 / 00:00:47.952
10:19:42.856	FAIL	Managed object does not contain fragment values
  wanted={'status': 'up', 'serviceType': 'thinedge'}
  got={'type': 'c8y_Service', 'name': 'tedge-agent', 'owner': 'device_TST_cue_steel_mascarpone', 'serviceType': 'service', 'status': 'up'}
```

I have run 50 test runns with the custom value of the timeout, which is set to 90 seconds

**Custom Timeout (90 seconds)**
- Average time needed for test suite to complete: **2.26**
- No test failures observed

reubenmiller (changes requested):
Sorry I'm going to have to block this as I still don't understand why adding an explicit timeout of 30 seconds helps here (considering the default timeout is 30 seconds)

gligorisaev:
> Sorry I'm going to have to block this as I still don't understand why adding an explicit timeout of 30 seconds helps here (considering the default timeout is 30 seconds)

Just to mention that he is actually increasing the Timeout to 90 Seconds 
`${TIMEOUT}= ${90}`

but I also agree that it should not be the reason for failing

reubenmiller:
> > Sorry I'm going to have to block this as I still don't understand why adding an explicit timeout of 30 seconds helps here (considering the default timeout is 30 seconds)
> 
> Just to mention that he is actually increasing the Timeout to 90 Seconds `${TIMEOUT}= ${90}`
> 
> but I also agree that it should not be the reason for failing

Though we should analyse why this is the case before doing any changes (or at least have a theory that could explain the behaviour)

Bravo555:
> All "smart assertions" in the testing library have a default timeout defined (30 seconds).

Sometimes when running this test suite, I would get logs e.g. `still running Tests.Cumulocity.Service Monitoring.Service Monitoring after 1500 seconds`. I have assumed that it was a lack of timeout on the asserts that caused this, and is struck me as odd that some asserts have the timeouts, and some don't, hence this PR. I wasn't aware of the 30 seconds default timeout. 

So there are 2 possibilities:
- either the default timeout isn't having an effect
- or there's something else that sometimes causes the tests to not finish

I'll add that I sometimes get this behaviour when I run all the tests locally, and almost never when I run this single suite, so on my machine the test seems to be a bit flaky.

reubenmiller:
Closing PR as the changes are not needed.

Issue #2899, opened 2024-05-24
backoff strategy for the in-built bridge


reubenmiller:
**Is your feature request related to a problem? Please describe.**

thin-edge.io should be a first good citizen when it comes to connecting to an external MQTT broker. It should implement reasonable backoff algorithms so that if the external MQTT becomes overloaded, it does not cause a "thundering herd" problem but quickly retrying to connect to the broker.

The new in-built MQTT bridge functionality should support a backoff strategy which will avoid such issues.

**Describe the solution you'd like**

The following backoff strategy is proposed:

* On each retry, the wait interval should be doubled (with jitter), but cap the max result a configurable value (e.g. 10mins)
* A minimum wait interval should be enforced (and configurable) to ensure that clients do not try to reconnect to quickly after a failed attempt. Default value should be 30 seconds
* The wait interval should only be reset to the minimum wait interval after the MQTT connection has been connected for a minimum period (e.g. 5 mins)

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
@didier-wenzek @albinsuresh @jarhodes314, we can discuss the specifics in the ticket.

didier-wenzek:
The proposed strategy is not directly aligned to what is proposed by the [backoff](https://docs.rs/backoff/latest/backoff/) crate, but one should be able to make it.

reubenmiller:
> The proposed strategy is not directly aligned to what is proposed by the [backoff](https://docs.rs/backoff/latest/backoff/) crate, but one should be able to make it.

The linked crate looks good, and I'm sure we can wrap it to utilize the reset logic regarding the "stable connection".

reubenmiller:
See [these comments](https://github.com/thin-edge/thin-edge.io/issues/2935#issuecomment-2160726488) for information about the backoff algorithm (plus jitter).

gligorisaev:
QA has througly checked the feature and is pleased to confirm that the described is functioning.

PR #2915, opened 2024-06-03
feat: Add reconnection backoff to built-in bridge

jarhodes314:
## Proposed changes
This adds some backoff logic to the MQTT bridge reconnection based on the [backoff crate](https://docs.rs/backoff).  It also fixes a minor issue with the tedge_config macro I encountered in the process of making the changes.

Still TODO before merging:
- [x] Ensure it is safe not to poll the rumqttc event loop after a connection error (I believe it is, but if there is an active connection, not polling the event loop will cause issues)
- [x] Deal with the exposed configurations for the timeout parameters, either by exposing the configuration to enable the built-in bridge, or by hiding the new configurations

## Types of changes

## Paste Link to the issue
- #2899 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -4,7 +4,7 @@ use crate::AutoFlag;
 use crate::AutoLogUpload;
 use crate::ConnectUrl;
 use crate::HostPort;
-use crate::Seconds;
+use crate::SecondsOrHumanTime;
 use crate::SoftwareManagementApiFlag;
 use crate::TEdgeConfigLocation;
 use crate::TemplatesSet;
@@ -666,16 +666,16 @@ define_tedge_config! {
 
             reconnect_policy: {
                 /// The minimum time the built-in bridge will wait before reconnecting
-                #[tedge_config(example = "30", default(function = "|| Seconds(30)"))]
-                initial_interval: Seconds,
+                #[tedge_config(example = "30s", default(from_str = "30s"))]
+                initial_interval: SecondsOrHumanTime,
 
                 /// The maximum time the built-in bridge will wait before reconnecting
-                #[tedge_config(example = "600", default(function = "|| Seconds(600)"))]
-                maximum_interval: Seconds,
+                #[tedge_config(example = "10m", default(from_str = "10m"))]
+                maximum_interval: SecondsOrHumanTime,
 
                 /// How long to wait after successful reconnection before resetting the reconnect timeout
-                #[tedge_config(example = "300", default(function = "|| Seconds(300)"))]
-                reset_window: Seconds,
+                #[tedge_config(example = "5m", default(from_str = "5m"))]
+                reset_window: SecondsOrHumanTime,
             },
         },
     },
@@ -787,9 +787,9 @@ define_tedge_config! {
         #[tedge_config(example = "true", default(value = true))]
         lock_files: bool,
 
-        /// Interval at which the memory usage is logged (in seconds). Logging is disabled if set to 0
-        #[tedge_config(example = "60", default(value = 0_u64))]
-        log_memory_interval: Seconds,
+        /// Interval at which the memory usage is logged (in seconds if no unit is provided). Logging is disabled if set to 0
+        #[tedge_config(example = "60s", default(from_str = "0"))]
+        log_memory_interval: SecondsOrHumanTime,
     },
 
     logs: {
@@ -817,8 +817,8 @@ define_tedge_config! {
         child: {
             update: {
                 /// The timeout limit in seconds for firmware update operations on child devices
-                #[tedge_config(example = "3600", default(value = 3600_u64))]
-                timeout: Seconds,
+                #[tedge_config(example = "1h", default(from_str = "1h"))]
+                timeout: SecondsOrHumanTime,
             }
         }
     },
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -663,6 +663,20 @@ define_tedge_config! {
             #[tedge_config(default(value = false))]
             #[doku(skip)] // Hide the configuration in `tedge config list --doc`
             built_in: bool,
+
+            reconnect_policy: {
+                /// The minimum time the built-in bridge will wait before reconnecting
+                #[tedge_config(example = "30", default(function = "|| Seconds(30)"))]
+                initial_interval: Seconds,
+
+                /// The maximum time the built-in bridge will wait before reconnecting
+                #[tedge_config(example = "600", default(function = "|| Seconds(600)"))]
+                maximum_interval: Seconds,
+
+                /// How long to wait after successful reconnection before resetting the reconnect timeout
+                #[tedge_config(example = "300", default(function = "|| Seconds(300)"))]
+                reset_window: Seconds,
+            },
         },
     },
 
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -465,7 +465,6 @@ define_tedge_config! {
             /// if this is set to "c8y", then messages published to `c8y/s/us` will be
             /// forwarded by to Cumulocity on the `s/us` topic
             #[tedge_config(example = "c8y", default(function = "c8y_topic_prefix"))]
-            #[doku(skip)] // Hide the configuration in `tedge config list --doc`
             topic_prefix: TopicPrefix,
         },
 
@@ -661,7 +660,6 @@ define_tedge_config! {
 
         bridge: {
             #[tedge_config(default(value = false))]
-            #[doku(skip)] // Hide the configuration in `tedge config list --doc`
             built_in: bool,
 
             reconnect_policy: {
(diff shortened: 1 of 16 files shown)


crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -663,6 +663,20 @@ define_tedge_config! {
             #[tedge_config(default(value = false))]
             #[doku(skip)] // Hide the configuration in `tedge config list --doc`
             built_in: bool,
+
+            reconnect_policy: {
+                /// The minimum time the built-in bridge will wait before reconnecting
+                #[tedge_config(example = "30", default(function = "|| Seconds(30)"))]
+                initial_interval: Seconds,
```
jarhodes314, line 670: I've used `Seconds` here, so we take an integer and use that as the number of seconds as that is consistent with other configurations. However, personally I think it would make more sense to use [humantime](https://docs.rs/humantime/latest/humantime/fn.parse_duration.html) to parse a duration value from the configuration.

This would allow a user both more granular precision if needed (`500ms` may make sense for this configuration), and clearer configuration of longer values (e.g. `5m` for 5 minutes instead of `300`).

crates/extensions/tedge_mqtt_bridge/src/backoff.rs
```diff
@@ -0,0 +1,194 @@
...
+
+        BackoffHandle {
+            duration: self.eb.next_backoff().unwrap(),
+            inner: self,
+        }
+    }
+
+    pub fn mark_success(&mut self) {
+        self.last_state = State::Success;
+    }
```
didier-wenzek, line 68: There is something I don't understand.

I would expect this method to keep track of when occurred the first successful connection event, so we don't reset the timer if an error occurs before a minimum period with an effective connection. Note also, that I would track the time of the last success and not of the last failure.

```suggestion
    pub fn mark_success(&mut self) {
        if self.last_state == State::Failure {
            self.okay_since = now();
            self.last_state = State::Success;
        }
    }
```

What I'm missing?

jarhodes314: I think the answer is that you're missing nothing. I initially started by not relying on the consumer of this module manually "resetting" the backoff, but that's both needlessly complicated and also flawed (someone may reasonably want to disable backoff through setting `inital_interval` to 0).

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
jarhodes314: @reubenmiller @didier-wenzek what do you think is the best way to integrate this change with the existing values. Do we:
- Just change these configurations to support a human-time value
- Support humantime everywhere we currently use `Seconds`, but in a backwards-compatible manner (either by having a warning when converting an integer to a value of seconds, or by automatically rewriting `tedge.toml` to use the new format, like we did for changing some of the key names)

reubenmiller: > Just change these configurations to support a human-time value

Would could use "human-time" just for the settings related to this PR, and then look at using humantime by default for all intervals/durations in the config.

> Support humantime everywhere we currently use Seconds, but in a backwards-compatible manner (either by having a warning when converting an integer to a value of seconds, or by automatically rewriting tedge.toml to use the new format, like we did for changing some of the key names)

If no units are provided, I would just assume seconds (I'm not sure if adding a warning would be worthwhile). I would avoid rewriting the `tedge.toml`.

didier-wenzek: Ideally, I would accept both:

- `initial_interval = 2` assuming this is a number of seconds
- `initial_interval = "5m"` where the human-time value is given as a string.

As this is backward compatible, why not apply this to all settings in `Seconds`.

crates/extensions/tedge_mqtt_bridge/src/backoff.rs
didier-wenzek: This commit implements what I have in mind https://github.com/thin-edge/thin-edge.io/pull/2915/commits/9e6ae7374a1e1a5f1cfc321ff0acd10f72a5ceef. Thanks

albinsuresh (approved):
I was really struggling to follow the first iteration, but the last commit really simplified the code. Thanks for that.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -663,6 +663,20 @@ define_tedge_config! {
...
+
+                /// The maximum time the built-in bridge will wait before reconnecting
+                #[tedge_config(example = "600", default(function = "|| Seconds(600)"))]
+                maximum_interval: Seconds,
+
+                /// How long to wait after successful reconnection before resetting the reconnect timeout
+                #[tedge_config(example = "300", default(function = "|| Seconds(300)"))]
```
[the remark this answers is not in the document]
reubenmiller: Actually the intent is not to avoid eager reconnect when the network is unreliable, it is more to protect the server in cases where massive load results in clients being dropped, otherwise the reconnects will cause even more load on the server.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
jarhodes314: I've now modified the type to accept seconds or humantime formats. It preserves the input, so it won't rewrite seconds to humantime values, nor will it change the format of humantime values (so if someone prefers `1 hour` instead of `1h`, they can set that and it won't get clobbered when updating tedge config). I'm wondering:

1. Do we want to place limits on what humantime values we accept? For instance, there are units of "week", "month" and "year", all of which are likely longer than would ever be sensible for our existing configurations (and someone could plausibly write `1M` thinking it's a minute and actually it's a month). If these are sensible for some future configuration, I would argue accepting seconds in place of that is daft.
2. Do we want to convert existing seconds values to humantime values when people update other `tedge config set` commands? Aside from setting a value to `0`, it doesn't make a lot of sense in my mind for anyone to be using values that are implicitly seconds if they have a choice, and the format is (to me at least) intuitive. Equally, it might be confusing to have values change unexpectedly.

I've set the example and default values to use a short humantime format, e.g. `30s`/`5m`.

crates/common/tedge_config_macros/impl/src/lib.rs
```diff
@@ -61,6 +62,33 @@ pub fn generate_configuration(tokens: TokenStream) -> Result<TokenStream, syn::E
         })
         .collect::<Vec<_>>();
 
+    let fromstr_default_tests = fields_with_keys
```
jarhodes314, line 65: The changes in `define_tedge_config` add a new `default(from_str = "...")` attribute, which parses the given str to the default value. This section of code adds a test that explicitly checks the default value can be parsed successfully.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
reubenmiller: > Do we want to place limits on what humantime values we accept? For instance, there are units of "week", "month" and "year", all of which are likely longer than would ever be sensible for our existing configurations (and someone could plausibly write 1M thinking it's a minute and actually it's a month). If these are sensible for some future configuration, I would argue accepting seconds in place of that is daft.

The limits are very config setting specific, so I would leave it up to the individual settings to define their own limits.

> Do we want to convert existing seconds values to humantime values when people update other tedge config set commands? Aside from setting a value to 0, it doesn't make a lot of sense in my mind for anyone to be using values that are implicitly seconds if they have a choice, and the format is (to me at least) intuitive. Equally, it might be confusing to have values change unexpectedly.

Can't we decouple the value in the tedge.toml from the stored value? If the value does not change, then why should the serialisation process change the value?

didier-wenzek (commented):
The new config type for human-time values is clear despite some complications to handle two formats and be backward compatible. The user benefits really worth the extra-complexity.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
didier-wenzek: > Do we want to place limits on what humantime values we accept?

I would not ask for `month` or `year` units if missing, but I see no value in adding code to remove these units. 

> Do we want to convert existing seconds values to humantime values when people update other tedge config set commands? 

I would not. As you said, it might be confusing to have values change unexpectedly.

gligorisaev (approved):
Prioritizing readability with 1h is the better approach than using 3600

jarhodes314:
> @jarhodes314 Is there any insight provided to the user what the current backoff period is (e.g. a log entry)? This would be helpful to debug things when checking when the next connection attempt will be made.

There isn't but I'll add it

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
jarhodes314: > Can't we decouple the value in the tedge.toml from the stored value? If the value does not change, then why should the serialisation process change the value?

Not easily, because the current process is:
1. Deserialize the entire `tedge.toml` into a `TedgeConfig` struct
2. Update the relevant configuration
3. Serialize the `TedgeConfig` struct to update the `tedge.toml`

We could plausibly modify things in the future to use something like [toml_edit](https://docs.rs/toml_edit/latest/toml_edit/) to do minimal, comment-preserving updates to the toml file, only affecting the key that is being changed, but right now this isn't possible.

Even without that though, the input to `tedge config set` isn't a TOML value. So in this case the approach that makes most sense is to parse the input to the config value's type and then (as a separate step) serialize it to a TOML value. As a result we would still need to store the input if we want to store the user provided values verbatim.

Issue #2847, opened 2024-04-30
feat(workflow): support combining fixed strings with variable expansions


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The variable expansion supported by the tedge-agent workflow syntax is too restrictive and does not support combining fixed strings with a variable

For example the following `script` will not be expanded as the `tedge-mapper-`

```toml
script = "sudo systemctl restart tedge-mapper-${.payload.mapperType}"
```

Because of that, if the user wants to build a script argument by combining a fixed string with a variable reference, they would then need to create a wrapper script which will do the string interpolation.

For example the above would have to be write a wrapper script, e.g. `/usr/bin/my_wrapper.sh`:

```sh
#!/bin/sh
sudo systemctl restart "tedge-mapper-$1"
```

And then call the wrapper script from the workflow state:

```toml
script = "/usr/bin/my_wrapper.sh ${.payload.mapperType}"
```

Writing a wrapper script adds too much overhead and is harder to maintain as it is easy for the wrapper script to get out of sync with the workflow definition.

**Describe the solution you'd like**

Support expanding variable references (e.g. `${<dotnotation>}`) even when a static prefix/suffix is provided (or even multiple variable references in the same argument?)

```toml
operation = "restart_mapper"

[init]
action = "proceed"
on_success = "executing"

[executing]
script = "sudo systemctl restart tedge-mapper-${.payload.mapperType}"
on_success = "successful"

[successful]
action = "cleanup"

[failed]
action = "cleanup"
```

**Describe alternatives you've considered**

**Additional context**

gligorisaev:
#### Test Cases:

**1. /tests/cumulocity/configuration/configuration_operation.robot**

##### New or Changed Sections
`Trigger config_snapshot operation from another operation`

- Ensures a configuration snapshot can be triggered as a part of another operation, validating the integration of operations.

`Trigger custom config_snapshot operation`

- Custom configurations for snapshot operations are tested, ensuring flexibility.

`Trigger config_update operation from another workflow`

- Similar to config_snapshot, but for update operations, verifying the integration and triggering from other workflows.

`Trigger custom config_update operation`

- Validates the handling of custom configuration updates.

`Customize config operations & Restore config operations`

- Ensures custom configurations can be applied and restored correctly, maintaining system integrity.

`Customize Operation Workflows`

- Tests custom operation workflows, ensuring they work as intended without affecting the main workflows.

[...]

PR #2918, opened 2024-06-04
feat: workflow support combining fixed strings with variable expansions

didier-wenzek:
## Proposed changes

Full string interpolation for workflow script parameters.

- in a context where the operation instance payload is  `{"x":"X", "y":"Y"}`
- the template `"prefix-${.payload.x}-separator-${.payload.y}-suffix"`
- is replaced by `"prefix-X-separator-Y-suffix"`

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/2847

## Checklist


crates/core/tedge_api/src/workflow/script.rs
```diff
@@ -241,7 +250,7 @@ impl ExitHandlers {
 
     pub fn forceful_timeout_extension(&self) -> Option<Duration> {
         self.timeout.map(|timeout| {
-            let extra = min(60, timeout.as_secs() / 20);
+            let extra = max(60, timeout.as_secs() / 20);
```
didier-wenzek, line 253: Yeah this was buggy. We want to wait at least 1 minute before killing the process.

docs/src/references/agent/operation-workflow.md
```diff
@@ -236,6 +236,19 @@ Data extracted from the command status topic and message payload can be passed a
...
+    - is replaced by `"prefix-X-separator-Y-suffix"`.
+- These substitutions are applied using the following rules:
+  - The script command is first tokenized using shell escaping rules.
+    *e.g.* `"sudo systemctl restart tedge-mapper-${.payload.mapperType}"`
+    is interpreted as the `sudo` command with 3 parameters.
+  - Then all the path expressions of each parameter are expended using the command current topic and payload.
+    *e.g.* `"tedge-mapper-${.payload.mapperType}"` is interpreted as `"tedge-mapper-c8y"`
+    (assuming the JSON payload contains a `mapperType` field with `"c8y"` as string value).
+  - These substitutions apply to the command itself as well as all its arguments.
+  - Any ill-formed path expression is let unchanged.
```
gligorisaev, line 251: The guide effectively explains the concept of argument substitution using path expressions and literals. It is well-structured and provides clear examples that illustrate the rules.

gligorisaev (approved):

docs/src/references/agent/operation-workflow.md
reubenmiller, line 251:
```suggestion
  - Any ill-formed path expression is left unchanged.
```

docs/src/references/agent/operation-workflow.md
```diff
@@ -236,6 +236,19 @@ Data extracted from the command status topic and message payload can be passed a
   - `${.payload.status}` is the command current status (*e.g.* `"init"`)
   - `${.payload.x.y.z}` is the json value extracted from the payload following the given `x.y.z` path if any.
   - If given `${.some.unknown.path}`, the argument is passed unchanged to the script.
+  - An argument can contain a combination of path expressions and literal:
```
reubenmiller, line 239:
```suggestion
  - An argument can contain a combination of path expressions and literals:
```

docs/src/references/agent/operation-workflow.md
reubenmiller: Though do we class non-existent paths as "ill-formed"?

For example, if a non-existent payload property is referenced, then it makes it more difficult for the binary to determine if the value is an expanded or unexpanded value:

```toml
script = "/usr/bin/mybinary ${.payload.doesNotExist}"
```

Checking for unexpanded values in a bash script is not so trivial, it would involve the following regex comparison:

```sh
#!/usr/bin/env bash

# check if value is expanded
if [[ "$1" =~ ^\\$\{.+\}$ ]]; then
    echo "Value is not expanded" >&2
    exit 1
fi
```

However if we just expanded the value to an empty string, then it becomes easier to detect (and generally a script does not care about the template variable language, just the value at the end)

```sh
#!/usr/bin/env bash

# check if value is empty
if [ -z "$1" ]; then
    echo "Value is empty" >&2
    exit 1
fi
```

reubenmiller: The distinction between what is ill-formed and what is non-existent is more applicable to the `${.payload}` object. Since the payload object represents a free-form value, then any nested paths should be allowed (regardless if the nested path exists under the payload or not).

```sh
# ok
${.payload.<path>}

# not ok
${.payld.<path>}
${.tpc}
```

didier-wenzek: Okay. I will change the behavior for unknown path (as in `${.payload.doesNotExist}`) to replace the path expression by an empty string. The intent of letting the expression unchanged in such a case was to help debugging by stressing something unexpected. I agree there are pros and cons and that an empty string can be more convenient in practice. Notably, if the user want to use message payloads with unset optional fields.

However, making a difference between `${.payload.<unknown-path>}` and `${.unknow-root.<path>}` might be more subtle. I will see what can be done.

docs/src/references/agent/operation-workflow.md
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/2918/commits/b06cc792fb601aa4c4f9d30f7ff7b1a70056ddac

docs/src/references/agent/operation-workflow.md
didier-wenzek: > The distinction between what is ill-formed and what is non-existent is more applicable to the `${.payload}` object. Since the payload object represents a free-form value, then any nested paths should be allowed (regardless if the nested path exists under the payload or not).
> 
> ```shell
> # ok
> ${.payload.<path>}
> 
> # not ok
> ${.payld.<path>}
> ${.tpc}
> ```

This has been addressed by https://github.com/thin-edge/thin-edge.io/pull/2918/commits/ccd8d2677f297b978c6ef075cfa242d1ab099cda

reubenmiller: great work

reubenmiller (approved):
Approved (from a functional perspective)

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -296,10 +321,14 @@ impl GenericCommandState {
             ".topic.operation" => self.operation().map(|s| s.into()),
             ".topic.cmd_id" => self.cmd_id().map(|s| s.into()),
             ".payload" => Some(self.payload.clone()),
-            path => path
-                .strip_prefix(".payload.")
-                .and_then(|path| json_excerpt(&self.payload, path))
-                .cloned(),
+            path if path.contains(['[', ']']) => None,
```
jarhodes314, line 324: I'm slightly confused as to what case this is covering? Is it explicitly filtering out TOML table headings? If so, why isn't `strip_prefix` covering this case?

jarhodes314 (approved):
One question, but I'm happy with the Rust code apart from that

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: The point is to [exclude JSON path where  a bracket notation is used to access an item below an array](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/workflow/state.rs#L756). Such a path is currently considered as ill-formed (the path expression is left unchanged), by contrast with paths leading to an unknown value  (the path expression being then substituted for the empty string). See https://github.com/thin-edge/thin-edge.io/pull/2918#discussion_r1627076669.

albinsuresh: Why is accessing an array value considered invalid? Because array indexes positions are non-deterministic? Or to avoid out-of-bound index risks? If it was the latter, I'd have used the same logic for that as unknown path in an existing root: returning empty.

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -270,24 +270,33 @@ impl OperationAction {
...
                     handlers.clone(),
                 )
             }
             _ => self.clone(),
         }
     }
+
+    fn inject_values_into_script(state: &GenericCommandState, script: &ShellScript) -> ShellScript {
+        ShellScript {
+            command: state.inject_values_into_template(&script.command),
```
albinsuresh, line 298: Jus wondering if we should add some additional restrictions to the variable substitution in script and operation names, limiting the substitution to primitive types only (Strings, numbers etc) and explicitly prevent complex values like arrays and objects as that's most likely an error when used in a script/operation name? It would be better fail during the validation phase itself rather than during execution.

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -258,21 +258,44 @@ impl GenericCommandState {
...
 
-    /// Inject values extracted from the message payload into a script argument
+    /// Inject values extracted from the message payload into a template string
+    ///
+    /// - Search the template string for path patterns `${...}`
+    /// - Replace all these paths by the value extracted from self using the paths
+    ///
+    /// `"prefix-${.payload.x}-separator-${.payload.y}-suffix"` is replaced by
+    /// `"prefix-X-separator-Y-suffix"` in a context where the payload is `{"x":"X", "y":"Y"}`
+    pub fn inject_values_into_template(&self, script_parameter: &str) -> String {
```
albinsuresh, line 274: Minor(since this function is more generic now):
```suggestion
    pub fn inject_values_into_template(&self, target: &str) -> String {
```

albinsuresh (approved):
LGTM.

crates/core/tedge_api/src/workflow/mod.rs
didier-wenzek: Yeah, why not.

crates/core/tedge_api/src/workflow/state.rs
didier-wenzek: Renamed

crates/core/tedge_api/src/workflow/mod.rs
reubenmiller: Though this shouldn't prevent passing complex objects as arguments to the script right?

For example:

```toml
[executing]
script = "/usr/bin/do_something ${.payload}"
```

didier-wenzek: Sure. I think @albinsuresh is speaking only about operation names.

PR #2920, opened 2024-06-05
feat: support device profile operation

reubenmiller:
## Proposed changes

An initial proof of concept to demonstrate one way of processing a device profile operations on the tedge `/cmd/` topic. This PR can be taken over by someone else to work on Rust implementation.

The current example device profile workflow uses a complex set of actions (to push the limits of this feature), the following highlights the nested workflows which are called through-out the process:

* device_profile
    * firmware_update
        * restart (device)
        * software_update (to update thin-edge.io components)
    * software_update (install other packages)
    * config_update

## Types of changes

## Paste Link to the issue

* https://github.com/thin-edge/thin-edge.io/issues/2921

## Checklist


tests/RobotFramework/tests/tedge_agent/device_profile/device_profile.toml
```diff
@@ -0,0 +1,69 @@
...
+# End states
+#
+[successful]
+action = "cleanup"
+
+[failed]
+action = "cleanup"
+
+# Points
+# * How to do conditional actions, e.g. only try installing firmware if there is firmware to be installed
```
didier-wenzek, line 68: Isn't this the role of `device_profile.sh` script?

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile.toml
```diff
@@ -0,0 +1,69 @@
...
+# * Allow variable expansion on the "operation" state property
```
didier-wenzek, line 69: This can easily be added. Will do in https://github.com/thin-edge/thin-edge.io/pull/2918

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile.toml
reubenmiller: Yes, though the initial approach was to have a sequential workflow without having a script doing the conditional logic.

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile.toml
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/2918/commits/81d22b3c41c627488006545f7a2abb577fea2e45

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,213 @@
...
+
+A device profile defines any desired combination of a firmware, software and associated configurations to be installed on a device.
+Device profiles are used to get a fleet of devices into a consistent and homogenous state by having the same set of firmware,
+software and configurations installed on all of them.
+
+The `tedge-agent` handles `device_profile` operations as follows:
+
+* Declares `device_profile` support by sending the capability message to `<root>/<device-topic-id>/cmd/device_profile` topics
+* Subscribes to `<root>/<device-topic-id>/cmd/device_profile/+` to receive `device_profile` commands.
+* The `device_profile` command payload is an ordered list of modules representing any firmware, software and configuration combination.
```
didier-wenzek, line 18: I'm not fully sure of what you define as a module?

The term *combination* confuses me.
- Is a module homogeneous: either a firmware or a set of software package or a set of configuration files?
- Or is a module a combination of those?
- Or is a module a single firmware/software/configuration sub-operation?

Reading the specs further, I can see you have the third case in mind.
Nevertheless this sentence is confusing.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,213 @@
...
+On such devices, pushing the entire firmware binary for each iterative change would considerably increase the binary size overhead.
+
+# Device profile capability
+
+A device that supports device profile operation must declare that capability by publishing the following message:
+
+```sh te2mqtt formats=v1
+tedge mqtt pub -r 'te/device/main///cmd/device_profile' '{
+  "name": "prod-profile",
+  "version: "v1"
```
didier-wenzek, line 45: From where are coming these name and version?

I guess from a previous successful device profile update. But what if meantime, *some* configuration files have been updated?

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+The profile definition in the payload is an array of modules.
+Each module could be a firmware, software or configuration.
+The modules are installed/applied in the order in which they are defined in the profile definition.
+There is no restriction on the order of modules in a profile and hence can be defined in any preferred order.
+For example, additional software can be installed or configurations updated before the firmware is updated.
+
+The `"include"` field is optional and the value is true, by default.
+It can be used to exclude already installed modules when the same profile is reapplied on a device after a partial failure.
```
didier-wenzek, line 111: You mean by manually excluding modules that have been fully installed by a previous partial run?

This might be cumbersome. Couldn't this be a responsibility of the system?
E.g to persist a watermark of what has been done by the partial run, and to resume from there a new attempt.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,213 @@
...
+
+When the profile is applied, `tedge-agent` is free to group modules of the same kind that are listed sequentially,
+to optimize the execution of that operation.
+For example, if 3 software modules are defined sequentially, the agent could group them into an `updateList`
+so that they can be installed in one-go using the `update-list` API of software plugins.
+
+## Tedge agent handling device profile commands
+
+The `tedge-agent` handless `device_profile` commands using the workflow definition at `/etc/tedge/operations/device_profile.toml`.
```
didier-wenzek, line 120:
```suggestion
The `tedge-agent` handles `device_profile` commands using the workflow definition at `/etc/tedge/operations/device_profile.toml`.
```

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,213 @@
...
+
+### On failure
+
+On failure, the `device_profile` operation is aborted at the module that caused the failure.
+The remaining modules in the profile are not installed and no attempt is made to rollback the already installed modules either.
+If the sub-operations support rollbacks at the sub-operation level, it is performed for the failed module.
+
+For example, if a profile includes a firmware, 2 software and 1 configuration updates in that sequence,
+if the failure happens during the second software update, no rollback is performed at the overall `device_profile` operation level,
+or even for that failed software update, unless the `software_update` workflow for that software type supports rollbacks.
```
didier-wenzek, line 156: >  unless the `software_update` workflow for that software type supports rollbacks.

What's the responsibility of the device profile workflow here?
Who triggers the rollback?

- Is this done by the sub-operation before reporting the failure?
- Or is this done by the device profile workflow? How then?

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,213 @@
...
+
+```sh te2mqtt formats=v1
+tedge mqtt pub -r 'te/device/main///cmd/device_profile/1234' '{
+  "status": "failed",
+  "reason": "Installation of software module: `jq` failed. Refer to operation logs for further details."
+  ... // Other input fields
+}'
+```
+
+The same profile can be applied again, by excluding all the modules(`"include": false`) before the failed `jq` module as follows:
```
didier-wenzek, line 169: Does this have to be the *same* profile? I would expect the user to tune something, say changing some config settings or a software version.

docs/src/references/agent/device-profile.md
albinsuresh: I used the term `module` as a generic term to refer to a single firmware, software or configuration item. And `profile` is any ordered combination of multiple modules of any kind.

docs/src/references/agent/device-profile.md
albinsuresh: Yes, the last applied device profile or the device profile with which the device was initially flashed in the factory. If such a factory profile is unavailable, that's when the agent can declare the empty JSON payload.

docs/src/references/agent/device-profile.md
albinsuresh: Yes, we can consider that. Since we've traditionally not stored/maintained any state about previously executed (esp. failed) operations, this was the only other alternative that I could think of, other than reapplying the same modules again and expect them to be idempotent.

docs/src/references/agent/device-profile.md
albinsuresh: In my mind, the `device_profile` workflow has no role in rollbacks. If it had, it should have done it at the entire profile level, which I don't think is feasible. Since rollbacks are typically handled at the individual operation workflow level (e.g: firmware update), it is left to that suboperation to rollback the failed module, if it can handle that.

The solution in my mind to the partially applied profile issue is to retry the same profile from the point of failure. In the current proposal it was left to the user to exclude all the already applied modules and resume from the failed module. But, as you suggested, we can make the agent maintain that state and automatically resume from the failed module without any user intervention.

docs/src/references/agent/device-profile.md
albinsuresh: It doesn't **have to be** the same profile. What I meant is that the user doesn't always have to create a new profile to fix the issue, as the underlying issue can even be fixed externally (directly or via remote access plugin, may be), if the failure was limited to a small number of devices in the whole fleet. And then the same profile can be reapplied again. But, if he wishes to push the fix also via a device profile, then he can definitely do that, by having a newer version of the profile with the fix.

Obviously, if the failure was affecting the majority of the fleet, then the user would definitely want to push the fix also using a device profile.

docs/src/references/agent/device-profile.md
reubenmiller: Though we already use `module` in the software operation data structure, so I'm not sure if that would introduce some confusion there.

We don't necessary have to give it a name, we could just say:

```
* The `device_profile` command payload is an ordered list of items, where an item is either firmware, software and configuration.
```

docs/src/references/agent/device-profile.md
reubenmiller: > I guess from a previous successful device profile update. But what if meantime, some configuration files have been updated?

The question of "has anything changed since applying the device profile" is a tricky one, as I don't think there is one-way here, as there are multiple ways that the device profile state could get "out of sync" (e.g. sending individual firmware/software/config operations, user running manual commands via ssh etc.)

Though I don't think that device profiles are always used to control the whole on-device setup (as this usually leads to have an unwieldy amount of device profiles to manage, especially if a single device only needs 1 extra software package or configuration)

Below are some options (not exhaustive) that thin-edge.io could do:

**Option 1:** Do nothing

thin-edge.io doesn't do anything, and it is up to the device maintainer to control how the device is managed

**Option 2:** Flag profile as "dirty" after other operations

We introduce a "dirty" flag to the device profile current state (e.g. in the /twin setting), which is set anytime a single firmware/software/configuration operation is sent after a device profile is applied. The "dirty" flag would work similar to git, where it just indicates that something has changed (but does not say what).

There could also be a possible option made to even disable individual cloud operations, e.g. only accept cloud device profile operations, and reject cloud based firmware/software/configuration operations.

**Option 3:** Add interface which can be executed to check if the device profile is still 

Similar to the sm-plugin "list" command, a device profile handler could be used to indicate whether a device profile is still in the desired state (e.g. has not been affected by individual firmware/software/configuration operations). This would give a lot of control over the user deciding what is correct to do here (which could be nice).

docs/src/references/agent/device-profile.md
reubenmiller: We should let each firmware/software/configuration handler be responsible for the idempotency. The device profile should just execute the same workflow again, and each individual handler can work out if it should really do something or not.

docs/src/references/agent/device-profile.md
didier-wenzek: >  The `device_profile` command payload is an ordered list of items

Definitely clearer.

docs/src/references/agent/device-profile.md
didier-wenzek: >  The device profile should just execute the same workflow again, and each individual handler can work out if it should really do something or not.

This will also let the user fix one step or another.

docs/src/references/agent/device-profile.md
didier-wenzek: What is proposed by @reubenmiller in this [comment](https://github.com/thin-edge/thin-edge.io/pull/2920#discussion_r1648507189) seems simpler and more convenient for the users:

> We should let each firmware/software/configuration handler be responsible for the idempotency. The device profile should just execute the same workflow again, and each individual handler can work out if it should really do something or not.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,213 @@
...
+      "version": "20240430.1139"
+    },
+    {
+      "module": "software",
+      "include": true,
+      "name": "collectd",
+      "version": "latest",
+      "type": "apt",
+      "action": "install"
+    },
```
reubenmiller, line 84: 1. I don't think it is a good idea mixing the the `module` property with the other properties, as it makes it more difficult to avoid name clashes between the device profile data structure and the component processing the item (e.g. software_update).
2. Updating of multiple software packages should still be possible when calling the `software_update` operation. It is probably better to just allow the payload within the array of items rather than flattening it out.
3. Using the software operation `software_update` instead of `software`, otherwise we have to maintain a mapping between the device profile component, and the translation to the different sub operations doing the work.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+                "name": "collectd",
+                "version": "latest",
+                "type": "apt",
+                "action": "install"
+              }
+            ]
+          }
+        ]
+      }
+    },
```
albinsuresh, line 107: Based on the feedback from last round, I've updated the input format moving away from the previous flat structure and kept it in-line with what @reubenmiller had originally proposed. The definitely helps with payload consistency although I'm not a bit unhappy with the level of nesting that's there.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+[scheduled]
+script = "/etc/tedge/operations/device_profile.sh ${.payload.status} ${.payload}"
+on_success = "executing"
+on_error = { status = "failed", reason = "fail to sort the profile list"}
+
+[executing]
+operations = [
+    { operation = "firmware_update", target_operation="firmware_update" }
+    { operation = "software_update", target_operation="custom_software_update" }
+    { operation = "config_update" }
```
albinsuresh, line 178:
```suggestion
    { operation = "firmware_update", target_operation="firmware_update" },
    { operation = "software_update", target_operation="custom_software_update" },
    { operation = "config_update" }
```

albinsuresh: Added a `target_operation` in addition to the `operation` so that the user can use a workflow different from the standard one that is used for standalone operations.

docs/src/references/agent/device-profile.md
albinsuresh: Added the firmware example as well.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+**Q2: Why not model each device profile update as a firmware update that includes the desired software and configurations as well**
+
+Although this is a more robust approach, it is not feasible on all kinds of devices,
+especially the ones that does not support delta firmware updates.
+On such devices, pushing the entire firmware binary for each iterative change would considerably increase the binary size overhead.
+
+# Requirements
+
+* Ability to control the order of execution of the operations defined in the command input
+* Ability to apply
```
didier-wenzek, line 41: Missing sentence end

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+* Ability to apply 
+
+# Device profile capability
+
+A device that supports device profile operation must declare that capability by publishing the following message:
+
+```sh te2mqtt formats=v1
+tedge mqtt pub -r 'te/device/main///cmd/device_profile' '{
+  "name": "prod-profile",
+  "version: "v1"
```
didier-wenzek, line 50: Why is this firmware version published as a capability? It rather sounds to be related to a device state such as `twin/firmware`

docs/src/references/agent/device-profile.md
didier-wenzek: >  I'm not a bit unhappy with the level of nesting that's there.

You mean *"I'm a bit unhappy "*? Don't you?

Level of nesting doesn't matter as this is read by a machine.

docs/src/references/agent/device-profile.md
didier-wenzek: I prefer the new proposal to `skip` a step.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+
+* The workflow just proceeds to the `scheduled` state from the `init` state
+* The order of operation execution is finalized in the `scheduled` state.
+  If the `builtin` action is specified in this state, the default order as defined in the input is used.
+  This order can be modified using a `script` action, if desired.
+  The updated list is captured into a `updated_profile` field with the starting `current_index` value.
+  The `current_index` value specifies the next item in the list to be processed by the `executing` state.
+  By default, this `current_index` value starts with `0`.
+  This index value is incremented as the operation execution proceeds.
+  This index can also be used to skip the first `n` items in the list, especially when the same profile is reapplied.
```
didier-wenzek, line 207: This raises many questions:

- Who increments this index value? The agent? A user provided script?
- When is this index value incremented?
- How is done the branching depending this index reached the end or not?

I guess that the intent is to free the user from writing the looping logic. However, as far I understand the proposal, the looping logic is scattered over different places making it hard both to implement and to understand. For instance the `"await-operation-completion"` action has to handle a looping logic even if a regular context. Furthermore, having a `current_operation` or `next_operation` is more convenient and meaningful than a loop index.

docs/src/references/agent/device-profile.md
didier-wenzek, line 178: A device profile operation might also have to trigger a restart or any user defined operation. Hence, I would not restrict the list of sub-operations here.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+```
+
+The same profile can be applied again after fixing the issues that caused the failure,
+and it is left to the individual sub-operations to determine whether
+the same operation that was successfully applied in the last attempt must be reapplied or not.
+The user can also manually skip any operation using the `skip` field.
+
+# Cumulocity operation mapping
+
+TBD
```
didier-wenzek, line 270: The main question to address is "how to define the order in which the operations have to be executed, while c8y just provides an unordered set of actions".

An option is to let the cloud user include in the device profile set a script with a well-know name that compute the desire order. The device profile workflow would look for that script in the device profile and would update its own config first thing if found.

didier-wenzek (commented):
I'm not happy with the current proposal, even if I appreciate the intent.

Yes, it would be good to liberate the user from defining the looping logic, but not at the price of a fragile and complicated logic scattered over all the states.

I would address the problem starting from the `tests/device_profile.sh` script and `tests/device_profile.toml` workflow. Can the script be built-in, producing the next operation at each invocation? The user would have then only to provide a preparation script that groups and sequences the operations.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+  ]
+}'
+```
+
+The profile definition in the payload is an array of operations.
+Each operation could be a `firmware_update`, `software_update` or `config_update`.
+The operations are executed in the order in which they are defined in the profile definition, by default.
+There is no restriction on the order of modules in a profile and hence can be defined in any preferred order.
+For example, additional software can be installed or configurations updated before the firmware is updated.
+This default execution order can also be overridden in the workflow definition, by updating the order in the `scheduled` state.
```
reubenmiller, line 144: Do we need to restrict this to a specific state? Can't we just assume that the sorting has happened prior to starting to process the items?

docs/src/references/agent/device-profile.md
reubenmiller: Yeah I would agree with with @didier-wenzek, I don't think this is necessary to add more mapping (as it is also unexpected). For example the "software_update" operation should behave exactly the same regardless of being run through a device_profile operation or a software_update operation

docs/src/references/agent/device-profile.md
reubenmiller: The Cumulocity IoT mapping shouldn't matter so much as the sorting belongs in the workflow scheduled step. So we can just take the list as it stands and initially sort it firmware, software, configuration (and keeping any sorting of individual sections)

docs/src/references/agent/device-profile.md
reubenmiller: Yeah I'm in favour of being as simple as possible, and possibly moving all as much of the logic outside of rust as it allows it to be easily modified if someone isn't happy with how we've handled it.

So we could even leave out all of the internal details about the `current_index` and it is just implementation details about how our specific script handler keeps track of which operation should be assigned to the `current_operation` fragment.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+  The updated list is captured into a `updated_profile` field with the starting `current_index` value.
+  The `current_index` value specifies the next item in the list to be processed by the `executing` state.
+  By default, this `current_index` value starts with `0`.
+  This index value is incremented as the operation execution proceeds.
+  This index can also be used to skip the first `n` items in the list, especially when the same profile is reapplied.
+* The `target_operation` to be invoked for each item in the input is defined in the `executing` state.
+  If a `target_operation` is not explicitly specified, then the default workflow for the `operation` value itself is used.
+  The sub-operation execution starts from the item corresponding to the `current_index` value.
+  As soon as the sub-operation is triggered, the workflow moves to the `awaiting_sub_operation` state defined as the `on_exec` target.
+* In the `awaiting_sub_operation` state, workflow just waits monitoring the state of the sub-operation completion.
```
reubenmiller, line 212: I won't bother going into too much detail about the `awaiting_sub_operation` as we wanted to remove this explicit step anyway as it seems unnecessary from the user's perspective (e.g. could be solved via syntactic sugar)

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,270 @@
...
+  Once the sub-operation is successful, the workflow moves back to the `executing` state with the index value incremented,
```
reubenmiller, line 213: It might be better to separate the `executing` state from the "processing current item", as the `executing` state is already linked to transitioning the operation state in the cloud, so calling the executing state multiple times might cause some unexpected issues.

docs/src/references/agent/device-profile.md
reubenmiller: Yes, the firmware version is published to the `twin/firmware` topic, below is an example:

```
[te/device/main///twin/firmware] {"name":"tedge_rugpi_tryboot","version":"20240410.1917","updated":"Tue  9 Jul 20:15:19 BST 2024"}
```

docs/src/references/agent/device-profile.md
albinsuresh: Yes. Although not mentioned explicitly, it was conceived to be an open table of any set of operations as we already anticipate the device profile to support more kinds of operations like `shell_command` (to execute random shell commands in between other operations) in future.

docs/src/references/agent/device-profile.md
albinsuresh: Yeah, I agree that something like `next_operation` would be better than overloading the `await-operation-completion` action, as it is more explicit.

I used indexes instead of the individual operations themselves, for their simplicity/ease of manipulation and also to avoid the confusion with identical operation entries in the device profile list (e.g: @reubenmiller has highlighted a use-case in the past where the same `tedge` version had to be installed twice as part of a device profile: once before firmware update and once after the firmware update, if the firmware includes an older version). In such cases, I thought an index would be more deterministic in highlighting the current position on the list of operations executed.

docs/src/references/agent/device-profile.md
albinsuresh: > For example the "software_update" operation should behave exactly the same regardless of being run through a device_profile operation or a software_update operation

If that's the case, then yeah it's not required. I assumed that sometimes the user may want the same operation to behave differently in the context of a device profile, as overriding the operation workflow itself overrides the individual operation execution as well.

docs/src/references/agent/device-profile.md
albinsuresh: Yeah, that's true. It just has to be any state prior to the `executing` state. I was just trying to keep the example as minimal as possible reusing the existing states as much as possible without introducing any additional non-essential states, as that might confuse the user that those states are also mandatory for device profiles.

But yeah, I'll rephrase the sentence to clarify that the sorting can be done in this state or any other subsequent state before `executing`.

docs/src/references/agent/device-profile.md
```diff
@@ -172,19 +172,24 @@ on_success = "executing"
...
+action = "builtin"
+on_success = "next_operation"
+
+[next_operation]
+action = "builtin"
+on_success = "apply_operation"
+on_error = { status = "failed", reason = "Failed to pick the next operation to be executed" }
+
+[apply_sub_operation]
+operations = [ "firmware_update", "software_update", "config_update" ]
```
didier-wenzek, line 184: I guess the meaning of this list of operations. "Apply the current operation among those". Isn't it?

But what is this current operation? Can a user-provided script set it? What are the rules?

I would try to do something along the lines of what we already have:

```
operation = "${.payload.next.operation}"
```

This might require some improvements. Notably it's currently uneasy to set the initial payload of an operation in a generic way. We could extend the system with:

```
operation = "${.payload.next.operation}"
input = "${.payload.next.input}"
```

docs/src/references/agent/device-profile.md
albinsuresh: Resolved by pushing the idempotency requirement to the sub-operation and not mandating the user to skip the already applied operations.

docs/src/references/agent/device-profile.md
albinsuresh: Resolved as suggested, by using the same payload format as used by the standalone operations.

docs/src/references/agent/device-profile.md
albinsuresh: Does not apply anymore as the operation name itself is completely dynamic now.

docs/src/references/agent/device-profile.md
albinsuresh: Moved the next operation computation logic completely into a built-in `iterator` action, which can be overridden using a script.

docs/src/references/agent/device-profile.md
albinsuresh: Rephrased the sentence as suggested.

docs/src/references/agent/device-profile.md
albinsuresh: Split this responsibility into a lightweight `executing` state and a `next_operation` state.

docs/src/references/agent/device-profile.md
albinsuresh: Addressed as suggested.

docs/src/references/agent/device-profile.md
albinsuresh: Updated as suggested, keeping it consistent with the `firmware` twin data.

docs/src/references/agent/device-profile.md
albinsuresh: Fixed in 0b3e2017

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,327 @@
...
+script = "/etc/tedge/operations/device_profile.sh ${.payload.status} ${.payload}"
+on_success = "executing"
+on_error = { status = "failed", reason = "fail to sort the profile list"}
+
+[executing]
+action = "builtin"
+on_success = "next_operation"
+
+[next_operation]
+action = "iterator"
```
albinsuresh, line 190: I went with this new `iterator` action instead of `builtin` just to provide a clear `on_next` handle.

didier-wenzek: > I went with this new `iterator` action instead of `builtin` just to provide a clear `on_next` handle.

This is clearer for the user and also more consistent with the current implementation. Using a `builtin` action here would be confusing as it would mean "delegates the step (here `next_operation`) to a built-in implementation of the workflow (here `device-profile`)".

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,327 @@
...
+on_next = "apply_operation"
+on_success = "successful"
+on_error = { status = "failed", reason = "Failed to compute the next operation to be executed" }
```
didier-wenzek, line 193: The current proposal is far better but still a bit too adhoc.

- The iterator action implicitly works on a list named `operations` yielding `operation` in turn. Why not iterating over any list yielding the items of this list in turn. For example `action = "iterate @next_operation ${.operations}"` where the user can choose for the list to iterate over as well as the current item name.
- Introducing a specific `on_next` handler makes impossible to substitute a script for iterate action.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,327 @@
...
+
+  If the `@next_operation` field is not present in the input payload.
+  one is added with an initial `index` value of `0` and the corresponding `operation` and `payload` values.
+  If the field already exists, the `index` value is incremented along with its `operation` and `payload` values.
+  Once the next operation is successfully computed, the workflow moves to the `on_next` target.
+  Once the `operations` list is exhausted (`index` value higher than its size),
+  the profile application is deemed complete and the workflow proceeds to the `on_success` target.
+  If the operation computation fails for some reason, then the workflow moves to the `on_error` target.
+  This builtin iteration logic can be overridden using a `script` action which can manipulate the order in any manner, dynamically.
+* The `apply_sub_operation` state executes the sub-operation defined in the `@next_operation` field in the payload.
```
reubenmiller, line 257:
```suggestion
* The `apply_operation` state executes the sub-operation defined in the `@next_operation` field in the payload.
```

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,327 @@
...
+  The `input` to the sub-operation is also extracted from the `payload` field of the `@next_operation`.
+  As soon as the sub-operation is triggered, the workflow moves to the `awaiting_sub_operation` state defined as the `on_exec` target.
+* In the `awaiting_sub_operation` state, workflow just waits monitoring the state of the sub-operation completion.
```
reubenmiller, line 260: Rename the `awaiting_sub_operation` state to `awaiting_operation` (for consistency) - Also in the example workflow toml.

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,327 @@
+---
+title: Configuration Management
+tags: [Reference, Device Profile, Firmware Management, Software Management, Configuration Management]
+sidebar_position: 6
+description: Device profile API proposal
+---
+
+# Device profile
+
+A device profile defines any desired combination of a firmware, software and associated configurations to be installed on a device.
```
reubenmiller, line 10:
```suggestion
A device profile defines any desired combination of firmware, software and associated configurations to be installed on a device.
```

docs/src/references/agent/device-profile.md
albinsuresh: > The iterator action implicitly works on a list named operations yielding operation in turn. Why not iterating over any list yielding the items of this list in turn. For example action = "iterate @next_operation ${.operations}" where the user can choose for the list to iterate over as well as the current item name.

Yeah, I can make this a generic iterator that takes any array as `input`, with the `@next` value generated on each iteration.

> Introducing a specific on_next handler makes impossible to substitute a script for iterate action.

I didn't fully get that point. If a user overrides this action with a script, then they don't need to use the `on_next` handle at all, but can use any number of target states using either the `on_stdout` directive or `on_exit.n` directives, right? So, I didn't really understand how this special `on_next` handle exclusive to the `iterator` action can hinder any overriding possibilities.

reubenmiller: Though the idea of reusing the `on_exit.n` properties is to also reduce the API, so we're not adding yet another property (which is only used in a very specific use-case).

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -249,8 +268,8 @@ impl OperationWorkflow {
 }
 
 impl OperationAction {
-    pub fn with_default(self, default: &DefaultHandlers) -> Self {
-        match self {
+    pub fn with_default(self, default: &DefaultHandlers) -> Result<Self, WorkflowDefinitionError> {
```
didier-wenzek, line 271: I would not return errors here. We need to distinguish "definition consistency checking" from "default values insertion". 

```suggestion
    pub fn with_default(self, default: &DefaultHandlers) -> Self
```


I think there are sensible defaults in most cases. 
For instance, the default "on_success" step for the iterate action can simply be the "successful" state.

That said, I agree that the way default values are managed can be improved. There are too many places where the workflow actor has to forge default steps. => I will address that in a follow-up PR.

crates/core/tedge_api/src/workflow/script.rs
```diff
@@ -347,6 +348,45 @@ impl AwaitHandlers {
...
+    ) -> Result<Self, WorkflowDefinitionError> {
+        if self.on_error.is_none() {
+            self.on_error.clone_from(&default.on_error)
+        }
+
+        if self.on_error.is_none() {
+            return Err(WorkflowDefinitionError::MissingState {
+                state: "on_success".to_string(),
+            });
+        }
```
didier-wenzek, line 384: There is a typo:

```suggestion
        if self.on_success.is_none() {
            return Err(WorkflowDefinitionError::MissingState {
                state: "on_success".to_string(),
            });
        }
```

Better is to provide a sensible default, instead of raising an error:

```suggestion
        if self.on_success.is_none() {
           self.on_success = Some(GenericStateUpdate::successful())
        }
```

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -465,6 +482,71 @@ impl WorkflowActor {
...
+
+    fn process_iterate(
+        &mut self,
+        state: GenericCommandState,
+        target_json_path: String,
+        handlers: IterateHandlers,
+    ) -> Result<GenericCommandState, IterationError> {
+        let Some(json_path) = GenericCommandState::extract_path(&target_json_path) else {
+            return Err(IterationError::InvalidPathExpression(target_json_path));
+        };
```
didier-wenzek, line 494: This should be done while parsing the workflow definition in `impl TryFrom<TomlExitHandlers> for IterateHandlers`

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -465,6 +482,71 @@ impl WorkflowActor {
             _ => Err(CommandTopicError::InvalidCommandTopic),
         }
     }
+
+    fn process_iterate(
+        &mut self,
+        state: GenericCommandState,
+        target_json_path: String,
+        handlers: IterateHandlers,
+    ) -> Result<GenericCommandState, IterationError> {
```
didier-wenzek, line 491: I would try to move this logic to `tedge_api`, the goal being to keep a clear separation between workflow definition and workflow execution.

didier-wenzek (commented):
Nice to see the progress! This is really close to the goal.

crates/core/tedge_agent/src/device_profile_manager/mod.rs
```diff
@@ -1,14 +1,23 @@
...
+use tedge_utils::file::create_file_with_defaults;
+use tedge_utils::file::FileError;
+
+pub struct DeviceProfileManagerBuilder {}
+
+impl DeviceProfileManagerBuilder {
+    pub fn try_new(ops_dir: &Utf8PathBuf) -> Result<Self, FileError> {
+        let workflow_file = ops_dir.join("device_profile.toml");
+        if !workflow_file.exists() {
+            let workflow_definition = r#"
```
didier-wenzek, line 11: You can also use [include_str!](https://doc.rust-lang.org/std/macro.include_str.html):

```suggestion
            let workflow_definition = include_str!("device_profile.toml");
```

crates/core/tedge_agent/src/device_profile_manager/mod.rs
```diff
@@ -40,3 +49,10 @@ action = "cleanup"
 
 [failed]
 action = "cleanup"
+"#;
+
+            create_file_with_defaults(workflow_file, Some(workflow_definition))?;
```
didier-wenzek, line 54: Creating a toml file for a builtin workflow is an idea to explore. Indeed, this would simplify the logic (with all the workflow definitions being file based) and this would give a pattern for the advanced user willing to tune an operation workflow.

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile.toml
```diff
@@ -0,0 +1,42 @@
...
+
+[init]
+action = "proceed"
+on_success = "scheduled"
+
+# Sort the inputs as desired
+[scheduled]
+action = "proceed"
+on_success = "executing"
+on_error = { status = "failed", reason = "fail to sort the profile list"}
```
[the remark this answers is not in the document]
albinsuresh: Remnants of the POC when this state was overridden by a sorting script. Will remove as it is not relevant for the inbuilt definition that just "proceeds" from this state.

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile_operation.robot
```diff
@@ -0,0 +1,177 @@
...
+    ...      "operations": [
+    # ...        {
+    # ...          "operation": "firmware_update",
+    # ...          "skip": false,
+    # ...          "payload": {
+    # ...            "name": "core-image-tedge-rauc",
+    # ...            "remoteUrl": "https://abc.com/some/firmware/url",
+    # ...            "version": "20240430.1139"
+    # ...          }
+    # ...        },
```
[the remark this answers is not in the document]
albinsuresh: Just realised that I missed the "skipping" logic in the iterator impl. Will do.

crates/core/tedge_api/src/workflow/mod.rs
albinsuresh: I was under the impression that the default values can be used only when default handlers are explicitly defined in the workflow. An implicit default of `failed` for an `on_error` target might be fine, but is it okay to implicitly default to the `successful` state when it is not specified explicitly? The iterate action is most likely an intermediate state in a workflow, and even after the iteration completes, there could be additional states that the workflow needs to go through, before deeming the execution as successful. So, if we implicitly default to the successful state right after the iteration (when one was not specified accidentally), critical steps in the workflow might get skipped. So, isn't it better to force the user to be explicit with the `on_success` target?

didier-wenzek: > So, isn't it better to force the user to be explicit with the `on_success` target?

This  can be discussed. From my point of view, the `successful` state makes sense as a default `on_success` target.

However, the main point of my comment is not here. Rather than on the UX, I was pointing out code organization.
The `with_default` method should only inject default values and not check correctness.

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile.toml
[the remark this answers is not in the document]
albinsuresh: Resolved by 70bf946048194d01f59af3782a9b0821b8f21a98

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,452 @@
...
+on_error = { status = "failed", reason = "fail to sort the profile list"}
+
+[executing]
+action = "builtin"
+on_success = "next_operation"
+
+[next_operation]
+action = "iterator"
+iterator.next_item = "next_operation"
+iterator.items = "operations"
```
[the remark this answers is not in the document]
albinsuresh: Resolved by 5c01f79272322f808484b5b6b67fc1ae4c65a1bd

crates/core/tedge_api/src/workflow/mod.rs
albinsuresh: Resolved by 1012f41c0069b302fb54aa249149fdee9ab20071

crates/core/tedge_api/src/workflow/script.rs
albinsuresh: Resolved by 1012f41c0069b302fb54aa249149fdee9ab20071

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: Resolved by 1012f41c0069b302fb54aa249149fdee9ab20071

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: Resolved by 1012f41c0069b302fb54aa249149fdee9ab20071

crates/core/tedge_agent/src/device_profile_manager/mod.rs
albinsuresh: Resolved by 4fedbabd8760b6c00565c3d4af6ccc145cb34d63

gligorisaev (commented):
LGTM

docs/src/references/agent/device-profile.md
```diff
@@ -0,0 +1,327 @@
...
+}'
+```
+
+### On failure
+
+On failure, the `device_profile` operation is aborted at the operation that caused the failure.
+The remaining operations in the profile are executed and no attempt is made to rollback the already completed operations either.
+If the sub-operations support rollbacks at the sub-operation level, it is performed for the failed operation.
+
+For example, if a profile includes a firmware, 2 software and 1 configuration updates in that sequence,
```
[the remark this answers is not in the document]
didier-wenzek: > For example, if a profile includes a firmware, 2 software and 1 configuration updates in that sequence,

This sentence has still a pending comment.

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile_operation.robot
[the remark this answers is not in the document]
didier-wenzek: > Just realised that I missed the "skipping" logic in the iterator impl. Will do.

This can be done is a follow-up PR.

didier-wenzek (approved):
Approved. Great achievement!

Issue #2923, opened 2024-06-05
Add `@id` to entity metadata when c8y-mapper generates external ID


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
When a new entity is created, c8y-mapper generates its external ID.

In this example, `TST_burn_sunny_confirmation:device:child-1234` is the external ID generated by c8y-mapper. This ID is currently stored in the entity store of c8y-mapper's converter. As long as one component has an access to the entity store (= must be a part of c8y-mapper's converter), it can consume the external ID.

```
[te/device/child-1234//] {"@type":"child-device"}
[c8y/s/us] 101,TST_burn_sunny_confirmation:device:child-1234,TST_burn_sunny_confirmation:device:child-1234,thin-edge.io-child
```

However, for other components, getting the external ID is fairly not possible, although such external ID is necessary to publish  JSON over MQTT (for both main and child devices) and SmartREST (for child devices) messages as part of the topic name.

I want a solution to get to know the entity's external ID in an easy way.

**Describe the solution you'd like**
During the entity registration, c8y-mapper re-publishes the metadata message with `@id` field when it's generated by mapper. By using the example above again, the last message should be added so that all components can know the generated external ID.

```
[te/device/child-1234//] {"@type":"child-device"} # user send
[c8y/s/us] 101,TST_burn_sunny_confirmation:device:child-1234,TST_burn_sunny_confirmation:device:child-1234,thin-edge.io-child # c8y registration
[te/device/child-1234//] {"@type":"child-device", "@id":"TST_burn_sunny_confirmation:device:child-1234"} # mapper re-publish
```

If `@id` is provided by user in the initial registration message, c8y-mapper will not do anything extra.

**Describe alternatives you've considered**

**Additional context**
While I am working in availability monitoring actor, I got a problem how to know devices' external IDs on Cumulocity.

rina23q:
Resolved by #2925

gligorisaev:
Feature checked with:

1.1.1-113-g82732a9e21

Test Case: tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot

The added test steps are covering the implemented feature

PR #2925, opened 2024-06-06
refactor: c8y mapper republishes metadata message with `@id`

rina23q:
## Proposed changes
Append `@id` if not given in entity registration messages.

Remark:
I haven't decided how to support the main device **before** any metadata messages for the main device are sent. Should mapper publish as part of init messages?
=> I gave it up. Read my comment. https://github.com/thin-edge/thin-edge.io/pull/2925#issuecomment-2155552747

## Types of changes

## Paste Link to the issue
#2923

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -985,7 +985,7 @@ impl CumulocityConverter {
             Channel::EntityMetadata => {
                 if let Ok(register_message) = EntityRegistrationMessage::try_from(message) {
                     return self
-                        .try_register_entity_with_pending_children(register_message)
+                        .try_register_entity_with_pending_children(&register_message)
                         .await;
                 }
                 Err(anyhow!(
@@ -1026,7 +1026,7 @@ impl CumulocityConverter {
 
     async fn try_register_entity_with_pending_children(
         &mut self,
-        register_message: EntityRegistrationMessage,
+        register_message: &EntityRegistrationMessage,
     ) -> Result<Vec<MqttMessage>, ConversionError> {
         let mut mapped_messages = vec![];
         match self.entity_store.update(register_message.clone()) {
@@ -1040,6 +1040,11 @@ impl CumulocityConverter {
                         self.try_convert_entity_registration(&pending_entity.reg_message)?;
                     mapped_messages.append(&mut c8y_message);
 
+                    // Republish the metadata message with @id if it's not given
+                    let mut updated_metadata_message =
+                        self.append_id_if_not_given(&pending_entity.reg_message);
+                    mapped_messages.append(&mut updated_metadata_message);
+
                     // Process all the cached data messages for that entity
                     let mut cached_messages =
                         self.process_cached_entity_data(pending_entity).await?;
@@ -1048,7 +1053,11 @@ impl CumulocityConverter {
 
                 return Ok(mapped_messages);
             }
-            Ok(_) => {}
+            Ok(_) => {
+                // Handle the case where @id is missing but there are no other changes to the payload
+                let mut updated_metadata_message = self.append_id_if_not_given(register_message);
+                mapped_messages.append(&mut updated_metadata_message);
+            }
         }
         Ok(mapped_messages)
     }
@@ -1086,6 +1095,25 @@ impl CumulocityConverter {
                 .append(&mut self.try_convert_auto_registered_entity(&entity, channel)?);
         }
         Ok(converted_messages)
+    }
+
+    fn append_id_if_not_given(
+        &mut self,
+        register_message: &EntityRegistrationMessage,
+    ) -> Vec<MqttMessage> {
+        let source = &register_message.topic_id;
+
+        if register_message.external_id.is_none() {
+            if let Some(metadata) = self.entity_store.get(source) {
+                let register_message_with_xid = register_message
+                    .clone()
+                    .with_external_id(metadata.external_id.clone());
+                let updated_metadata_message =
+                    register_message_with_xid.to_mqtt_message(&self.mqtt_schema);
+                return vec![updated_metadata_message];
+            }
+        }
+        vec![]
     }
 
     async fn try_convert_data_message(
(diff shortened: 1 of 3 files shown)


albinsuresh:
> I haven't decided how to support the main device before any metadata messages for the main device are sent. Should mapper publish as part of init messages?

Yes, that seems like as sensisble place for it.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -984,9 +984,16 @@ impl CumulocityConverter {
...
-                        .try_register_entity_with_pending_children(register_message)
-                        .await;
+                    let mut converted_messages = self
+                        .try_register_entity_with_pending_children(&register_message)
+                        .await?;
+
+                    // Republish the metadata message with @id if it's not given
+                    let mut updated_metadata_message =
+                        self.append_id_if_not_given(register_message, &source);
+                    converted_messages.append(&mut updated_metadata_message);
```
albinsuresh, line 994: You may have to push this logic into the `try_register_entity_with_pending_children` function so that the same logic is applied to the pending children as well. They should also have their generated `@id`s published. Once that's done, add some test coverage as well for the pending entities. Refer to `registration_lifecycle.robot::Early data messages cached and processed` for some samples.

tests/RobotFramework/tests/cumulocity/registration/mqtt_representation.robot
```diff
@@ -0,0 +1,27 @@
...
+Library    Cumulocity
+Library    ThinEdgeIO
+
+Test Tags    theme:c8y    theme:registration
+Suite Setup    Custom Setup
+Test Teardown    Get Logs
+
+*** Test Cases ***
+
+Mapper republishes entity registration messages with @id
```
albinsuresh, line 12: An entire test suite just for this test looks like an overkill. I'd have just added this test to `registration_lifecycle.robot` itself or even modified some of those existing tests to validate this additional message as well.

crates/extensions/c8y_mapper_ext/src/converter.rs
rina23q: Addressed in 0ef0227031078fd9e74e6073437fde2ea113bd72

rina23q:
> > I haven't decided how to support the main device before any metadata messages for the main device are sent. Should mapper publish as part of init messages?
> 
> Yes, that seems like as sensisble place for it.

I found this approach has a problem. If `te/device/main//` registration message is already retained, init function will overwrite the existing metadata message. It is significantly difficult to know **if no message is already retained on the topic**. There will be the same problem even if we move publishing `te/device/main//` registration message to `tedge connect c8y`.

Now I understood why we didn't publish auto generated `te/device/main//` metadata... I'm even considering to give it up, since it's not as critical as child devices. In any case, we can know the main device's external ID from tedge config.

rina23q:
By the way I'll modify the commit messages later. I didn't know that using `@` causes to mention someone who I didn't intend at all :(

gligorisaev (approved):
Reviewed the registration_lifecycle.robot

All test steps passed, Execution time 00:09:02

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1088,6 +1100,24 @@ impl CumulocityConverter {
         Ok(converted_messages)
     }
 
+    fn append_id_if_not_given(
+        &mut self,
+        register_message: &EntityRegistrationMessage,
+        source: &EntityTopicId,
+    ) -> Vec<MqttMessage> {
```
didier-wenzek, line 1107: On all the sites this method is called with a source that is `register_message.topic_id`.

```suggestion
    fn append_id_if_not_given(
        &mut self,
        register_message: &EntityRegistrationMessage,
    ) -> Vec<MqttMessage> {
    let source = &register_message.topic_id;
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1048,7 +1055,12 @@ impl CumulocityConverter {
 
                 return Ok(mapped_messages);
             }
-            Ok(_) => {}
+            Ok(_) => {
+                // Handle the case where @id is missing but there are no other changes to the payload
+                let mut updated_metadata_message =
+                    self.append_id_if_not_given(register_message, &register_message.topic_id);
+                mapped_messages.append(&mut updated_metadata_message);
```
didier-wenzek, line 1062: So this is the case a registration message forged by some external software (running on behalf of a child device).

- Is this okay for C8Y to have an external id that is not derived from the main device name (as done here)?
- Is this compatible with auto-registration (say a measurement is received from this child device *before* its registration)?
- If auto-registration is disabled and no @id is provided by the registration message, what would be the external id for the device?

rina23q: This is the case where a registration message is arrived, but it's judged that the message affects no entities...

> Is this okay for C8Y to have an external id that is not derived from the main device name (as done here)?

Should be ok.

> Is this compatible with auto-registration (say a measurement is received from this child device before its registration)?

If some entities are registered during the auto-registration process, the new registration messages will be published. Then, these messages are arrived in this condition arm exactly, and republishes (=overwrite) the retained messages with `@id`.

> If auto-registration is disabled and no @id is provided by the registration message, what would be the external id for the device?

I think the message goes to "pending entities". @albinsuresh can you confirm?

crates/extensions/c8y_mapper_ext/src/converter.rs
rina23q: done in [02caf34](https://github.com/thin-edge/thin-edge.io/pull/2925/commits/02caf346e9135d3deb2fe09feff5a1a23987a807)

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: > If some entities are registered during the auto-registration process, the new registration messages will be published. Then, these messages are arrived in this condition arm exactly, and republishes (=overwrite) the retained messages with @id.

For auto-registered entities, their generated registration messages that are published should already have the `@id` field in it. So, it shouldn't need another republish.

>  think the message goes to "pending entities". @albinsuresh can you confirm?

That's correct. They are kept cached until the registration message arrives (with or without the `@id`).

albinsuresh (approved):
LGTM

didier-wenzek (approved):
Approved

Issue #2922, opened 2024-06-05
tedge mqtt cli commands do not disconnect cleanly from MQTT broker


reubenmiller:
**Describe the bug**

When the local mosquitto MQTT broker instance is configured to use certificates, then publishing an MQTT message via `tedge mqtt pub`, an `OpenSSL` log error is visible in the mosquitto logs (related to the publishing of the message), below shows an example of this error:

```log
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: OpenSSL Error[0]: error:0A000126:SSL routines::unexpected eof while reading
```

Publishing the same message with mosquitto_sub does not result in an OpenSSL error (see the "additional context" section for an example).

The logs related to the entire MQTT publish call is shown below (including the OpenSSL error).

```sh
tedge mqtt pub foo bar
```

```log
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: New connection from 127.0.0.1:45954 on port 8883.
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: New client connected from 127.0.0.1:45954 as tedge-pub-1335 (p2, c1, k60).
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: No will message specified.
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: Sending CONNACK to tedge-pub-1335 (0, 0)
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: Received PUBLISH from tedge-pub-1335 (d0, q0, r0, m0, 'foo', ... (3 bytes))
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: Sending PUBLISH to auto-3AC5450B-1A23-EAFA-561A-34EBB76A5A49 (d0, q0, r0, m0, 'foo', ... (3 bytes))
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: OpenSSL Error[0]: error:0A000126:SSL routines::unexpected eof while reading
Jun 05 14:56:19 fe9669c87b15 mosquitto[382]: 1717599379: Client tedge-pub-1335 closed its connection.
```

Most likely the MQTT connection is not being correctly stopped, and instead just being abruptly terminated without sending an MQTT disconnect message.

[...]

PR #2936, opened 2024-06-12
fix: Ensure the MQTT connection is properly closed

didier-wenzek:
## Proposed changes

Ensure the MQTT connection is properly closed

- by `tedge mqtt pub <topic> <message>`
- by `tedge mqtt sub <topic-filter>` when interrupted.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2922

## Checklist


crates/core/tedge/src/cli/mqtt/publish.rs
```diff
@@ -135,6 +135,13 @@ fn publish(cmd: &MqttPublishCommand) -> Result<(), MqttError> {
     }
 
     client.disconnect()?;
+    for event in connection.iter() {
```
albinsuresh, line 138: Initially I found the placement of this loop here, after the disconnect call, very confusing. But, I'm assuming that the disconnection request is also not processed until we explicitly poll the event loop as per the rumqtt design. Never expected this for a disconnect call as I thought that the `Outgoing::Disconnect` event is received by the event loop **after** the disconnect message is already sent to the broker, just as an ack to the client which doesn't need to be actively polled. But I guess they went for consistency in the processing model for disconnect as well.

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -108,6 +110,10 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), MqttError> {
                 eprintln!("INFO: Disconnected");
                 break;
             }
+            Ok(Event::Outgoing(Outgoing::Disconnect)) => {
+                eprintln!("INFO: Interrupted");
```
albinsuresh, line 114:
```suggestion
                eprintln!("ERROR: Interrupted");
```

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -82,6 +83,7 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), MqttError> {
     }
 
     let (mut client, mut connection) = Client::new(options, DEFAULT_QUEUE_CAPACITY);
+    disconnect_if_interrupted(client.clone());
```
albinsuresh, line 86: Why not do this for the publisher as well, to handle cases like interrupting the publisher that's stuck in the connect phase itself?

albinsuresh (commented):
LGTM. Happy to approve once the query about not having a signal handler for the publisher is answered.

crates/core/tedge/src/cli/mqtt/subscribe.rs
reubenmiller: I wouldn't class this log message as an ERROR as it is normal behaviour for the user to want to stop the subscriber and the expected way is to use Ctrl-c

didier-wenzek: I even wonder to get rid of this INFO message.

I tried to be consistent with the messages already emitted by this command (`Connected`, `Disconnect`, and now `Interrupted`). But are these really useful?

crates/core/tedge/src/cli/mqtt/publish.rs
didier-wenzek: This was the root cause of the issue. Triggering a `disconnect()` is not enough.

However, the point of polling the rumqtt event loop for an `Outgoing::Disconnect` event  is:
- not to perform the disconnection (this is done by rumqtt in a background thread)
- but to be sure the disconnection actually happened before the `tedge mqtt pub` command dies.

crates/core/tedge/src/cli/mqtt/subscribe.rs
didier-wenzek: Why not. A publisher will unlikely be stuck (`tedge mqtt pub` disconnects on error and awaits nothing more than an acknowledgement) but the connection could be not responsive.

=> I will add it for consistency

crates/core/tedge/src/cli/mqtt/publish.rs
albinsuresh: > but to be sure the disconnection actually happened before the tedge mqtt pub command dies.

In that case, should we wait for the `Event::Incoming(Incoming::Disconnect(_))` ack from the broker as well? To make sure that the broker received it as well?

didier-wenzek: > In that case, should we wait for the `Event::Incoming(Incoming::Disconnect(_))` ack from the broker as well? To make sure that the broker received it as well?

No, this event is sent when the broker itself triggers the disconnection.

crates/core/tedge/src/cli/mqtt/subscribe.rs
didier-wenzek: Added by https://github.com/thin-edge/thin-edge.io/pull/2936/commits/531411e2ba94302723dc084b31fbc9b3cfbac523

crates/core/tedge/src/cli/mqtt/subscribe.rs
rina23q: > I even wonder to get rid of this INFO message.

I vote +1 for it.

reubenmiller: Yeah, let's get rid of the log message. I just checked what `mosquitto_sub` does, and it does not print out a log message about being killed.

didier-wenzek: Removed by https://github.com/thin-edge/thin-edge.io/pull/2936/commits/53d65369d5544b0e320a978b80863cd6a2d9b8dc

albinsuresh (approved):
LGTM.

reubenmiller (approved):
Functions nicely now, and clean disconnects for both `tedge mqtt pub` and `tedge mqtt sub` commands.

Issue #2842, opened 2024-04-24
Support Cumulocity Device Availability feature


rina23q:
**Is your feature request related to a problem? Please describe.**
Support the Cumulocity IoT availability monitoring out of the box to improve the usability by showing the thin-edge.io agent.

Cumulocity References:
* https://cumulocity.com/docs/device-integration/fragment-library/#device-availability

**Example (from Cumulocity IoT Device Management Application):**
<img width="246" alt="image" src="https://github.com/thin-edge/thin-edge.io-roadmap/assets/3029781/e0fc5d39-9b40-4236-aeb4-699724175fbd">
**Requirements**
* Send "heartbeat" periodically (e.g. this is generally an empty managed object update request, or sending of an event, measurement or alarm)
* Configurable heartbeat period at which the heartbeat signal is sent (e.g. default to 30mins)

**Describe the solution you'd like**
1. Add two new keys to tedge config as below. These are the default values. (Thanks to @reubenmiller)
```toml
[c8y.availability]
period = 60     # Required interval, and period of the heartbeat (if enabled = true)
enable = true   # disable this feature all together, don't send a 117 message, and disable the heartbeat
```
2. `tedge-mapper-c8y` reads the keys on startup, and send a SmartREST [`117` ](https://cumulocity.com/docs/smartrest/mqtt-static-templates/#117) message per device, to both main and child devices. Note that this SmartREST message doesn't update the value if it's already set as described in the c8y user guide. 
3. If a new child device is registered in runtime, `tedge-mapper-c8y` publishes a `117` message for the device.
4. `tedge-mapper-c8y` starts an internal timer for each device. Timer for the main device starts on startup. Timer for a child device starts when it is registered to the entity store.
5. When timer fires, `tedge-mapper-c8y` checks the service status of the device's `tedge-agent` by default. If the status is "up", `tedge-mapper-c8y` publishes an empty inventory update message.
6. If the device declares a different lead service in its metadata as `@health`, `tedge-mapper-c8y` uses the lead service's status to determine the device's availability. For the details of the mechanism , see the description of the [PR](https://github.com/thin-edge/thin-edge.io/pull/2866).

**Describe alternatives you've considered**
Since sending alarms, events, measurements, inventory updates are considered an availability update message, when such message has been sent by tedge-mapper-c8y, tedge-mapper-c8y resets the timer of sending heartbeat messages.

**Additional context**
**Note: Updating service status is NOT considered an availability update event**
[...]

PR #2901, opened 2024-05-24
fix(mqtt): Publish mosquitto bridge status to c8y when enabled

albinsuresh:
## Proposed changes

If the mosquitto bridge is used for C8y connectivity, then the bridge service status must also be published to C8y.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2898

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #2940, opened 2024-06-16
feat: support c8y device availability by AvailabilityActor

rina23q:
## Proposed changes
This PR is the replacement of the PR #2866. 

Now `@health` is limited to `EntityTopicId` scheme, the 4-segments identifier. 
If user wants to monitor a custom service, they need to provide `@health` in the registration message payload like below.
```
{
  "@health": "any/four/segments/topic",
  "@type": "child-device"
}
```
Then, the status of `te/any/four/segments/topic/status/health` will be respected.

Todo:

- [x] Implement AvailabilityActor
- [x] Add robot tests
- [x] Add unit tests => done a49ab99c7110258eb9e20030d4cd2171613c03d0
- [x] Change the actor Input and Output as suggested in https://github.com/thin-edge/thin-edge.io/pull/2940#discussion_r1643007345 => done ca3027605560c40fc0bfddb1585db7ee08199bf3
- [x] Add doc => 544315dee26e1037a040c346bfc00197283bc709

## Types of changes

## Paste Link to the issue
#2842
## Checklist


crates/extensions/c8y_mapper_ext/src/availability/builder.rs
```diff
@@ -0,0 +1,85 @@
...
+    config: AvailabilityConfig,
+    box_builder: SimpleMessageBoxBuilder<AvailabilityInput, AvailabilityOutput>,
+    mqtt_publisher: DynSender<MqttMessage>,
+    timer_sender: DynSender<TimerStart>,
+}
+
+impl AvailabilityBuilder {
+    pub fn new(
+        config: AvailabilityConfig,
+        mqtt: &mut (impl MessageSource<MqttMessage, TopicFilter> + MessageSink<MqttMessage>),
```
albinsuresh, line 33: In the context of the message race that you highlighted between the 117 message emitted by this actor and the 101 message emitted by the mapper actor, this might be the root of it: Both the actors connected directly to the broker and acting on them independently without any coordination. The previous iteration of this PR was getting around that coordination by making the mapper actor do everything. This is not an issue limited to this actor alone, but in future when we extract the operation managers also from the mapper actor as independent actors, we'll face the exact same issues in terms of race between operation capability messages and the device registration messages.

One solution to this problem is to have some coordination between these actors and the mapper actor. So, in this case, instead of this actor processing MQTT messages directly from the broker, let the mapper be the central actor that deals with the conversion and despatch of MQTT messages. And then these actors would only receive proper typed messages from the mapper actor. Similarly, instead of these actors publishing MQTT messages directly, let them route those via the mapper actor.

That way, we can avoid the possibility of any message races, as the same mapper actor can make sure that it publishes the registration message before the converted message is handed over to the availability actor and similarly the converted availability messages sent by this actor are also published by the same mapper actor ensuring that it always follows the registration message that was previously sent.

These in-memory channels(message boxes) between these actors and the mapper actor also enables them to query the main entity store maintained by the mapper, instead of each actor maintaining their own "lightweight" entity stores.

rina23q: Your proposal is definitely a proper way to address the issue. However, it will block the availability feature again. Can't we just put a delay (e.g. sleep 1s) before MqttMessage to the MQTT actor? Or having proper solution is the highest priority? I need a guide what to do next

didier-wenzek: > One solution to this problem is to have some coordination between these actors and the mapper actor. So, in this case, instead of this actor processing MQTT messages directly from the broker, let the mapper be the central actor that deals with the conversion and despatch of MQTT messages. And then these actors would only receive proper typed messages from the mapper actor. Similarly, instead of these actors publishing MQTT messages directly, let them route those via the mapper actor.

@albinsuresh Your proposal is definitely the way to go.

> Your proposal is definitely a proper way to address the issue. However, it will block the availability feature again. Can't we just put a delay (e.g. sleep 1s) before MqttMessage to the MQTT actor? Or having proper solution is the highest priority? I need a guide what to do next

@rina23q I appreciate your concerns and I agree that deferring again the availability feature is not desirable.
However, we must avoid this PR adding more work in the near future.

So I propose to change the input and output types of the `AvailabilityActor` to be fully unrelated to MQTT.
i.e. to directly consume `EntityRegistrationMessage` and `tedge_api::HealthStatus`
and to produce `c8y_api::HealthStatus`

```
fan_in_message_type!(AvailabilityInput[EntityRegistrationMessage, tedge_ap::HealthStatus] : Debug);
fan_in_message_type!(AvailabilityOutput[[MqttMessage](c8y_api::HealthStatus)] : Debug);
```

MQTT message parsing, which is currently done in `AvailabilityActor::process_mqtt_message()`,
will have to be move in `AvailabilityBuilder::new()` using something along the lines of [`MessageSink::connect_mapped_source`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_actors/src/builders.rs#L161) or [`MessageSource::connect_mapped_sink()``](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_actors/src/builders.rs#L202).

This will prepare for the refactoring proposed by @albinsuresh where the mapping actor will be the interface between MQTT and in-memory channels forwarding `tedge_api::*` messages to the actual actors and receiving back `c8y_api::*` messages.

didier-wenzek: See https://github.com/thin-edge/thin-edge.io/commit/4e1d3d58be403f12d5674c0575067bc7e191d97d for an example

didier-wenzek: > This solution is not 100% fool-proof because there is a corner case where the `@id`, `@health` and `@name` are all specified in the registration message, the 117 message might create the child with the wrong "name".

I don't get that. Why an explicit  `@name` might lead to a message with the wrong name?

rina23q: @albinsuresh 
> Instead of a delay, I'd just make the availability actor react only to registration messages with an `@id` in it

It's already like that. Otherwise there is no way to know `external ID` for child devices :D Also, for the main device, the device registration is done during `tedge connect c8y`, so it's out of concern.

> This solution is not 100% fool-proof because there is a corner case where the `@id`, `@health` and `@name` are all specified in the registration message, the 117 message might create the child with the wrong "name".

Both `@name` and `@health` are unrelated to 117 messages. Just a corner case is where the `@id` is provided in the initial registration message for child devices, in the case, tedge-mapper-c8y doesn't republishes the message since already `@id` is there.

rina23q: ca3027605560c40fc0bfddb1585db7ee08199bf3

> See [4e1d3d5](https://github.com/thin-edge/thin-edge.io/commit/4e1d3d58be403f12d5674c0575067bc7e191d97d) for an example

ca3027605560c40fc0bfddb1585db7ee08199bf3 changed the code like the example above. The actor does no longer consume `MqttMessage` as input and output ;)

tests/RobotFramework/tests/cumulocity/availability/c8y_required_availability.robot
```diff
@@ -0,0 +1,64 @@
...
+
+    ${CHILD_SN}=    Get Random Name
+    Set Test Variable    $CHILD_SN
+    Set Test Variable    $CHILD_XID    ${DEVICE_SN}:device:${CHILD_SN}
+
+Register child
+    [Arguments]
+    Execute Command    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device"}'
+    Set Device    ${CHILD_XID}
+    Device Should Exist    ${CHILD_XID}
```
gligorisaev, line 64: Empty line on the end is missing

tests/RobotFramework/tests/cumulocity/availability/heartbeat.robot
```diff
@@ -0,0 +1,72 @@
...
+    ${CHILD_SN}=    Get Random Name
+    Set Test Variable    $CHILD_SN
+    Set Test Variable    $CHILD_XID    ${DEVICE_SN}:device:${CHILD_SN}
+
+    # Set tedge config value before connecting
+    Execute Command    ./bootstrap.sh --no-bootstrap --no-connect
+    Execute Command    sudo tedge config set c8y.availability.interval 1
+    Execute Command    ./bootstrap.sh --no-install
+
+    Device Should Exist    ${DEVICE_SN}
```
gligorisaev, line 72: Here as well empty line on the end is missing

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -728,6 +734,7 @@ pub enum ChannelFilter {
     AlarmMetadata,
     AnyCommandMetadata,
     CommandMetadata(OperationType),
+    Health,
```
didier-wenzek, line 737: This comes as a surprise to see that `status/health` was not already listed among all the MQTT channels!

crates/core/c8y_api/src/smartrest/inventory.rs
```diff
@@ -116,6 +117,23 @@ pub fn service_creation_message_payload(
...
 
+/// Create a SmartREST message to set a response interval for c8y_RequiredAvailability.
+///
+/// The interval can be <=0, which means the device is in maintenance mode in the c8y context.
+/// Details: https://cumulocity.com/docs/device-integration/fragment-library/#device-availability
+pub fn set_required_availability_message(
+    c8y_topic: C8yTopic,
+    interval: i16,
+    prefix: &TopicPrefix,
+) -> MqttMessage {
```
didier-wenzek, line 128: Rather than such an adhoc function, it would be better to `impl Into<MqttMessage> for C8ySmartRestSetInterval117`.

crates/extensions/c8y_mapper_ext/src/availability/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+mod tests;
+
+pub type TimerStart = SetTimeout<TimerPayload>;
+pub type TimerComplete = Timeout<TimerPayload>;
+pub type SourceHealthStatus = (ServiceTopicId, HealthStatus);
+
+fan_in_message_type!(AvailabilityInput[EntityRegistrationMessage, SourceHealthStatus, TimerComplete] : Debug);
+fan_in_message_type!(AvailabilityOutput[C8ySmartRestSetInterval117, C8yJsonInventoryUpdate] : Debug);
+
+// TODO! Make it generic and move to c8y_api crate while refactoring c8y-mapper
```
didier-wenzek, line 27: Better to do this right now.

This crate is already using `c8y_api::smartrest::inventory::set_required_availability_message` so moving `C8ySmartRestSetInterval117` there should be easy!

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
```diff
@@ -0,0 +1,151 @@
...
+                    Channel::EntityMetadata => {
+                        if let Ok(registration_message) =
+                            EntityRegistrationMessage::try_from(&message)
+                        {
+                            return Some(registration_message.into());
+                        }
+                    }
+                    Channel::Health => {
+                        // Support only the default service schema: "device/+/service/+/status/health"
+                        if source.is_default_service() {
```
didier-wenzek, line 90: Why this restriction? It seems to be unused, at least here.

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
```diff
@@ -0,0 +1,295 @@
...
+                        }
+                    }
+                }
+            }
+            Err(err) => RegistrationResult::Error(format!("'@health' contains invalid value in {source}. Details: {err}")),
+        }
+    }
+
+    /// Set a new timer for heartbeat
+    /// Caution: the heartbeat interval from config is defined in MINUTES, not seconds
```
didier-wenzek, line 209: Using `SecondsOrHumanTime` would help to remove the risk for such unit mismatches.

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
```diff
@@ -0,0 +1,295 @@
...
+            None => {
+                Ok(registration_message.topic_id.to_default_service_topic_id("tedge-agent").unwrap())
+            }
+            Some(raw_value) => {
+                match raw_value.as_str() {
+                    None => Err(format!("'@health' must hold a string value. Given: {raw_value:?}")),
+                    Some(maybe_service_topic_id) => {
+                        EntityTopicId::from_str(maybe_service_topic_id)
+                            .map(|id| id.into())
+                            .map_err(|_| format!("'@health' must be the default service topic schema 'device/DEVICE_NAME/service/SERVICE_NAME'. Given: {maybe_service_topic_id}"))
```
didier-wenzek, line 185: This error message is not aligned with what is checked: `maybe_service_topic_id` is expected to match  `EntityTopicId` and not necessarily a service id of the default MQTT scheme.

didier-wenzek (commented):
This PR is in a good shape. Some points can still be improved, though.

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
rina23q: This is actually used.
- `EntityTopicId` can be anything as long as it has 4 segments. `custom/foo/bar/baz` is a valid `EntityTopicId`.
- `mqtt_schema.entity_channel_of(&topic)` checks if the topic is `EntityTopicId` + `/status/health`.

So, if `source.is_default_servie()` gets rid of something like the `source` = `custom/foo/bar/baz/status/health`.

Or is your question why it doesn't accept `custom/foo/bar/baz/status/health`?

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -492,7 +492,17 @@ define_tedge_config! {
...
+        },
+
+        availability: {
+            /// Enable sending heartbeat to Cumulocity periodically. If set to false, c8y_RequiredAvailability won't be sent
+            #[tedge_config(example = "true", default(value = true))]
+            enable: bool,
+
+            /// Heartbeat interval in minutes and the value to be sent to Cumulocity as c8y_RequiredAvailability
+            #[tedge_config(example = "60", default(value = 60_i16))]
+            interval: i16,
```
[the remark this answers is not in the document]
rina23q: Nice! This is exactly what I wished to do!

tests/RobotFramework/tests/cumulocity/availability/c8y_required_availability.robot
rina23q: Done c68e00a94a4d5ba0eb4aeeb4f2cfff2ac4bed64d

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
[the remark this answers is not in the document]
didier-wenzek: Cumulocity gives the same meaning to *all* values <= 0  (the device is considered in maintenance mode). Hence, a duration of 0 can be used for that purpose.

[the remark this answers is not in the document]
rina23q: Changed in ac55af84304812888b895086bbc67df0cd59cc0f

crates/core/c8y_api/src/smartrest/inventory.rs
```diff
@@ -116,6 +117,23 @@ pub fn service_creation_message_payload(
     ]))
 }
 
+/// Create a SmartREST message to set a response interval for c8y_RequiredAvailability.
+///
+/// The interval can be <=0, which means the device is in maintenance mode in the c8y context.
+/// Details: https://cumulocity.com/docs/device-integration/fragment-library/#device-availability
```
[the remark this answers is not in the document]
rina23q: Changed in ac55af84304812888b895086bbc67df0cd59cc0f

crates/extensions/c8y_mapper_ext/src/availability/mod.rs
rina23q: Done in 2e05dc2b36fc1d10b9c6bcc28bd3e492e5520b05

rina23q: Done in 2e05dc2b36fc1d10b9c6bcc28bd3e492e5520b05

crates/core/c8y_api/src/smartrest/inventory.rs
rina23q: Done in 2e05dc2b36fc1d10b9c6bcc28bd3e492e5520b05

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
rina23q: I'm now thinking about the design.

Indeed, in terms of implementation level, it doesn't matter if the `@health` has either `EntityTopicId` or `ServiceTopicId`. `EntityTopicId` is just four segments, can be `custom/any/four/segs`. On the contrary, `ServiceTopicId` limits to the default scheme,`device/DEVICE_NAME/service/SERVICE_NAME`.

If `EntityTopicId`, the custom health endpoint can be `te/custom/any/four/segs/status/health` as well as the default service scheme.
If `ServiceTopicId`, the custom health endpoint can be only the default scheme, `te/device/DEVICE_NAME/service/SERVICE_NAME/status/health`.

Now I re-read the discussion in the initial PR, from the @reubenmiller 's [comment,](https://github.com/thin-edge/thin-edge.io/pull/2866#discussion_r1612930218), he said just entities, so I think I misunderstood that it needs to limit to `ServiceTopicId`. 

Any thoughts?

crates/core/c8y_api/src/smartrest/inventory.rs
```diff
@@ -116,6 +117,28 @@ pub fn service_creation_message_payload(
...
 
+/// Create a SmartREST message to set a response interval for c8y_RequiredAvailability.
+///
+/// In the SmartREST 117 message, the interval must be in MINUTES, and can be <=0,
+/// which means the device is in maintenance mode in the c8y context.
+/// Details: https://cumulocity.com/docs/device-integration/fragment-library/#device-availability
+#[derive(Debug)]
+pub struct C8ySmartRestSetInterval117 {
+    pub c8y_topic: C8yTopic,
+    pub interval: u64,
```
didier-wenzek, line 128:
```suggestion
    pub interval: Duration,
```

Using a duration here would keep the knowledge, that c8y expects here a number of minutes, in a single place: the `impl From<C8ySmartRestSetInterval117> for MqttMessage `.

For instance, this would avoid such fragile code and comment:

```
                interval: self.config.interval.as_secs() / 60, // convert to MINUTES
```

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
didier-wenzek: > Or is your question why it doesn't accept `custom/foo/bar/baz/status/health`?

Yes, why `te/custom/foo/bar/baz/status/health` should be rejected?
Why not accept any `te/+/+/+/+/status/health` messages?

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
didier-wenzek: I would accept any `te/+/+/+/+/status/health` messages.

If a registration message declares its `@heath` entity being `custom/any/four/segs`,
then simply uses it!

rina23q: Done in cb41d2df39dd6dfb81f47ee784af77e08368e2f6

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
rina23q: Done in cb41d2df39dd6dfb81f47ee784af77e08368e2f6

crates/core/c8y_api/src/smartrest/inventory.rs
rina23q: Done in cb41d2df39dd6dfb81f47ee784af77e08368e2f6

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
rina23q: Actually [robot tests failures](https://github.com/thin-edge/thin-edge.io/actions/runs/9585880674/job/26456761734) proves that the race condition of `101` and `117` really occurs :/ I'd try to add a small delay and how the result will be.

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
```diff
@@ -0,0 +1,287 @@
...
+
+    /// Set a new timer for heartbeat if the given interval is positive value
+    async fn start_heartbeat_timer_if_interval_is_positive(
+        &mut self,
+        source: &EntityTopicId,
+    ) -> Result<(), RuntimeError> {
+        if !self.config.interval.is_zero() {
+            self.timer_sender
+                .send(SetTimeout::new(
+                    self.config.interval,
```
albinsuresh, line 207:
```suggestion
                    self.config.interval / 2,
```
Reason explained in [this comment](https://github.com/thin-edge/thin-edge.io/pull/2866#discussion_r1611209242) from the old PR.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -492,7 +492,19 @@ define_tedge_config! {
...
-        }
+        },
+
+        availability: {
+            /// Enable sending heartbeat to Cumulocity periodically. If set to false, c8y_RequiredAvailability won't be sent
+            #[tedge_config(example = "true", default(value = true))]
+            enable: bool,
+
+            /// Heartbeat interval to be sent to Cumulocity as c8y_RequiredAvailability.
+            /// If it is set to 0, the device is considered in maintenance mode in the Cumulocity context.
```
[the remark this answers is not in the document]
rina23q: Done 4fea7a84dd83af120f026fef0771eea262013c17

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
rina23q: Done 4fea7a84dd83af120f026fef0771eea262013c17

rina23q:
Now all robot tests passed after increasing sleep time from 500ms to 1000ms and fixed a bug! However, it was the 3rd try... 
these two tests failed before that. Already flaky? or this actor made them flaky?

- Trigger Agent Restart in `Custom Operation`. [In Attempt#1](https://github.com/thin-edge/thin-edge.io/actions/runs/9619001997/attempts/1?pr=2940)
- Test non-mosquitto-bridge service status mapping in `Service Monitoring`. [In Attempt#2](https://github.com/thin-edge/thin-edge.io/actions/runs/9619001997/attempts/2?pr=2940)

rina23q:
Regarding timeout,

If it's 5s, some tests in `Registration Lifecycle`. See the two runs:
- https://github.com/thin-edge/thin-edge.io/actions/runs/9643131755
- https://github.com/thin-edge/thin-edge.io/actions/runs/9643414016

If it's 1s, all tests passed. I don't know why 🤯

tests/RobotFramework/tests/cumulocity/availability/c8y_required_availability.robot
```diff
@@ -0,0 +1,64 @@
...
+Library    Cumulocity
+Library    ThinEdgeIO
+
+Test Tags    theme:c8y
+Test Setup    Test Setup
+Test Teardown    Get Logs
+
+*** Test Cases ***
+
+c8y_RequiredAvailability is set by default
```
didier-wenzek, line 12:
```suggestion
c8y_RequiredAvailability is set by default to an houur
```

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
```diff
@@ -0,0 +1,289 @@
...
+                        }
+                    }
+                }
+            }
+            Err(err) => RegistrationResult::Error(err),
+        }
+    }
+
+    /// Set a new timer for heartbeat if the given interval is positive value
+    async fn start_heartbeat_timer_if_interval_is_positive(
```
didier-wenzek, line 201: It's not a good idea to name a function after a internal details

```suggestion
    async fn start_heartbeat_timer(
```

didier-wenzek (commented):
I'm okay to merge this PR despite the weak solution used to prevent a 117 to be sent before a 101.

But before, it would be good to add a banner in the doc.

```
:::info
🚧 This feature is not finalized yer and might not work as expected on all devices.
:::
```

rina23q:
> I'm okay to merge this PR despite the weak solution used to prevent a 117 to be sent before a 101.
> 
> But before, it would be good to add a banner in the doc.
> 
> ```
> :::info
> 🚧 This feature is not finalized yer and might not work as expected on all devices.
> :::
> ```

If we want to mark like this, I think we should also disable the feature by default. (`c8y.availability.enable` = `false`)

tests/RobotFramework/tests/cumulocity/availability/c8y_required_availability.robot
rina23q: done in d1c4656

crates/extensions/c8y_mapper_ext/src/availability/actor.rs
rina23q: done in d1c4656

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -67,6 +84,7 @@ pub struct C8yMapperActor {
     mqtt_publisher: LoggingSender<MqttMessage>,
     timer_sender: LoggingSender<SyncStart>,
     bridge_status_messages: SimpleMessageBox<MqttMessage, MqttMessage>,
+    message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
```
didier-wenzek, line 87: Nice!

I really like how this message source can be used to dispatch MQTT messages to subscribers, including the option for a given subscriber to receive message from several channels.

didier-wenzek (approved):
Approved.

Thank you @rina23q for your perseverance.
Thank you @albinsuresh for the icing on the cake!

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1280,37 +1301,37 @@ impl CumulocityConverter {
         Ok(registration_messages)
     }
 
-    fn convert_entity_registration_message(
-        &self,
-        value: &EntityRegistrationMessage,
-    ) -> MqttMessage {
-        let entity_topic_id = value.topic_id.clone();
+    // fn convert_entity_registration_message(
```
albinsuresh, line 1304: To be removed. Using `EntityRegistrationMessage::to_mqtt_message()` instead.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -984,86 +1034,65 @@ impl CumulocityConverter {
...
-                    return self
-                        .try_register_entity_with_pending_children(&register_message)
-                        .await;
-                }
-                Err(anyhow!(
-                    "Invalid entity registration message received on topic: {}",
-                    message.topic.name
-                )
-                .into())
+                // No conversion done here as entity data messages must be converted using pending_entities_from_incoming_message
```
albinsuresh, line 1037:
```suggestion
                // No conversion done here as entity data messages must be converted using try_register_source_entities
```
I'm not fully happy with this special case for registration messages. I was forced to do this because of the differences in the way we process a user published registration message vs auto-generated registration message, where the auto-registered entity registration message created by a `health` message is skipped to avoid duplicate status messages which was the cause of #2902 and fixed in #2901.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -325,6 +324,57 @@ impl CumulocityConverter {
         })
     }
 
+    pub async fn try_register_source_entities(
```
didier-wenzek, line 327: This method deserves a doc comment.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -325,6 +324,57 @@ impl CumulocityConverter {
...
+                        "Invalid entity registration message received on topic: {}",
+                        message.topic.name
+                    )
+                    .into())
+                }
+                _ => {
+                    // if the target entity is unregistered, try to register it first using auto-registration
+                    if self.entity_store.get(&source).is_none()
+                        && self.config.enable_auto_register
+                        && source.matches_default_topic_scheme()
```
didier-wenzek, line 349: I would be better to make these checks inside `self.try_auto_register_entity(&source)`.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -325,6 +324,57 @@ impl CumulocityConverter {
...
+                    {
+                        let auto_registered_entities = self.try_auto_register_entity(&source)?;
+                        Ok(auto_registered_entities
+                            .into_iter()
+                            .map(|reg_msg| reg_msg.into())
+                            .collect())
+                    } else {
+                        Ok(vec![])
```
didier-wenzek, line 357: As the message is related to some unknown entity, I would expect an error here.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -129,26 +153,125 @@ impl C8yMapperActor {
...
+                self.process_data_message(message.clone()).await?;
+            }
+        } else {
+            self.convert_and_publish(&message).await?;
+        }
+
+        Ok(())
+    }
+
+    async fn process_registered_entities(
```
didier-wenzek, line 194: It's not easy to follow the progression of a message moving all these methods:

- `process_mqtt_message`
- `process_registered_entities`
- `process_registration_message`
- `process_data_message`
- `convert_and_publish`
- `publish_message_to_subscribed_handles`
- `publish_message`

Doc comments would help to grasp what is the happy path for a message.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -135,12 +148,21 @@ async fn child_device_registration_mapping() {
...
             "c8y/s/us/test-device:device:child1",
             "101,test-device:device:child2,test-device:device:child2,thin-edge.io-child",
-        ), (
+        )],
+    )
+    .await;
+
+    // while let Some(msg) = availability.recv().await {
+    //     dbg!(msg);
+    // }
```
didier-wenzek, line 157: To be removed

tests/RobotFramework/tests/cumulocity/availability/c8y_required_availability.robot
```diff
@@ -0,0 +1,64 @@
...
+Test Teardown    Get Logs
+
+*** Test Cases ***
+
+c8y_RequiredAvailability is set by default to an hour
+    Execute Command    ./bootstrap.sh
+
+    # Main
+    Device Should Exist    ${DEVICE_SN}
+    # Device Should Have Fragment Values    c8y_RequiredAvailability.responseInterval\=60
```
didier-wenzek, line 17: Why has this check been commented out?

didier-wenzek (commented):
I saw nothing obviously wrong. However, despite great efforts to clarify the code,  I would have a hard time to explain the registration rules by reading the code. Too many aspects are tangled: tedge registration, c8y registration, entity store, MQTT inputs, MQTT outputs. I'm afraid this cannot be fixed in this PR,. This has to done short term, though. Indeed, this over complicated state is one of the reason this PR is taking so long.

crates/extensions/c8y_mapper_ext/src/inventory.rs
```diff
@@ -467,29 +467,19 @@ mod tests {
             r#"{"name":"firmware", "version":"1.0"}"#,
         );
 
+        converter
+            .try_register_source_entities(&twin_message)
+            .await
+            .unwrap();
```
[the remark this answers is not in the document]
albinsuresh: Instead of repeating the validation of the auto-registered messages in the conversion tests for all data types: measurements, events, twin updates etc, I just wrote two new tests dedicated for auto-registration named `auto_register_child_device` and `auto_register_child_device_service` where the registration messages are validated. One additional tweak that I can make to those tests is to make it `test_case` based and add variants for all the data types as additional test cases.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: I didn't get that. This `else` case is for the case where the target topic id is **present** in the entity store. So, there is nothing left to be registered and hence returning empty vector.

didier-wenzek: > This `else` case is for the case where the target topic id is **present** in the entity store.

Not only. This `else` case also covers the case where the target topic id  is unknown *but* auto-registration is off or the topic id is not along the default scheme (`device/<name>//`).

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: In a follow up refactoring, I think it would be good to remove such inconsistencies between auto vs explicit registration. I also think that all the auto-registration logic should be encapsulated inside the entity store making the c8y mapper fully unaware of it.

didier-wenzek (commented):
I'm okay to merge this work. However, I need some clarification about the system tests: what's the current status after all these back and forth changes?

albinsuresh:
> I'm okay to merge this work. However, I need some clarification about the system tests: what's the current status after all these back and forth changes?

The earlier tests which were all sharing the same broker were trying to isolate things using random prefixes in topics, ids etc. Now that we've switched to isolated test setup, that randomness that was added in an earlier commit in this PR was reverted. The randomness in the external IDs is still required as the test are still sharing the same C8y tenant.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: > > This `else` case is for the case where the target topic id is **present** in the entity store.
> 
> Not only. This `else` case also covers the case where the target topic id is unknown _but_ auto-registration is off or the topic id is not along the default scheme (`device/<name>//`).

Correct. In those cases as well, we're supposed to just cache that message in the pending entity store to be processed later. That caching logic was in the `try_convert_data_message` earlier. I've now moved it here, with detailed commentary, to avoid this confusion. Hope the logic is clearer now.

didier-wenzek: Clearer

didier-wenzek (approved):
Approved. Thank you, @rina23q and @albinsuresh, for this sustained effort.

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
albinsuresh: Resolved by routing messages to the availability actor through the mapper actor.

Issue #2941, opened 2024-06-17
thin-edge.io installaltion with deb files need getent


bjoernsauer:
**Describe the bug**
The installation of thin-edge.io with the Debian Packages need `getent`. This command is not available on busybox. 

**To Reproduce**
Try to install tedge.deb on a target that runs with busybox.

**Expected behavior**
Avoid getent use grep for example to check if a group or user exists.

**Environment (please complete the following information):**
 - OS
    Custom Yocto Linux 
 - Hardware [incl. revision]
   Phoenix Contact Controller AXC F 2152
 - System-Architecture [e.g. result of "uname -a"]
   Linux axcf2152 5.4.257-rt87-pxc SMP PREEMPT_RT Thu Nov 16 06:49:45 UTC 2023 armv7l armv7l armv7l GNU/Linux
 - thin-edge.io version
   v1.1.1

**Additional context**

```
root@axcf2152# dpkg -i -D=2 /opt/plcnext/apps/60002172000584/images/tedge/arm/tedge.deb 
(Reading database ... 32 files and directories currently installed.)
Preparing to unpack .../images/tedge/arm/tedge.deb ...
/var/lib/dpkg/tmp.ci/preinst: line 9: getent: command not found
groupadd: group 'tedge' already exists
dpkg: error processing archive /opt/plcnext/apps/60002172000584/images/tedge/arm/tedge.deb (--install):
 new tedge:armhf package pre-installation script subprocess returned error exit status 9
Errors were encountered while processing:
 /opt/plcnext/apps/60002172000584/images/tedge/arm/tedge.deb
```

reubenmiller:
We would have to avoid doing a manual grep on `/etc/passwd` and `/etc/group` as this doesn't work if the OS uses custom user/groups controlled by [NSS](https://man7.org/linux/man-pages/man5/nss.5.html), namely Fedora IoT does not store the values under aforementioned files during installation due to its usage of rpm-ostree (to enable rolling back of packages etc.).

@bjoernsauer I think thin-edge.io can switch to use checks with `id` (for both user and group), and this should work in busybox as [id](https://man7.org/linux/man-pages/man1/id.1.html) is included in the [busy-box list of commands](https://boxmatrix.info/wiki/BusyBox-Commands)

bjoernsauer:
Hi @reubenmiller, yes `id` would be available. Maybe check for `getent` and prefere to use it if it is available and fall back to a alternative implementation with `id` or something else.

[...]

PR #2942, opened 2024-06-17
fix: improve compatibility of user/group existence check in package maintainer scripts

reubenmiller:
## Proposed changes

Improve the compatibility across different linux/bsd distributions when checking for the existence of users or groups during the installation and remove of thin-edge.io.

The new logic will checks for the existence of the dependent tool before using it, and if not found will continue to the next tool, the order is as follows:
1. `id`
2. `getent`
3. `grep /etc/passwd` (user) or `grep /etc/group` (groups) - Though this is only a last resort and will not work reliably on distributions that use some exotic NSS setup, like Fedora IoT, however in the majority of cases, `id` SHOULD be installed.

There is another related PR regarding the thin-edge.io/install.sh script which also used `getent` by default:
* https://github.com/thin-edge/thin-edge.io-website/pull/59

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2941

## Checklist

### DIFF
--- a/configuration/package_scripts/_generated/tedge/apk/postrm
+++ b/configuration/package_scripts/_generated/tedge/apk/postrm
@@ -5,8 +5,32 @@ command_exists() {
     command -V "$1" >/dev/null 2>&1
 }
 
+group_exists() {
+    name="$1"
+    if command_exists id; then
+        id -g "$name" >/dev/null 2>&1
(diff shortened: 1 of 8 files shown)


didier-wenzek (approved):
Approved

Issue #2864, opened 2024-05-08
easily add/remove values to existing tedge config properties of an array type


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Editing tedge config values which are arrays is difficult to managed for users as it involves checking the initial value, and adding the new value only if it is not already included in the list. Most of the time the "thing" adding the new value does not care about the existing values, and just wants to make sure the value it cares about is added to the list without breaking any other components.

There are currently the following tedge config settings which expect an array:

```sh
$ tedge --config-dir /opt/homebrew/etc/tedge config list --all | grep '\['
c8y.smartrest.templates=["collection1", "collection2"]
c8y.topics=["te/+/+/+/+", "te/+/+/+/+/twin/+", "te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+/+/+/+/status/health"]
az.topics=["te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+/+/+/+/status/health"]
aws.topics=["te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+/+/+/+/status/health"]
```

The problem can be demonstrated via  the tedge config `c8y.smartrest.templates` setting.

1. Package 1 includes an operation handler which requires subscribing to a SmartREST template called `collection1`

    ```sh
    tedge config set c8y.smartrest.templates "collection1"
    ```

    After adding the `collection1` template, the configuration get be retrieved using:

    ```sh
    $ tedge config get c8y.smartrest.templates
    ["collection1"]
    ```

2. Package 2 includes another operation handlers which requires subscribing to a different SmartREST template called `collection2`

    The package can't just run the following command, as it would overwrite the existing value from the previous step, so it requires more complex logic using jq to splice in a new value (below adds the new values, and then removes any duplicated template names)

    ```sh
    SMARTREST_TEMPLATE_NAME=collection2

    NEW_VALUE=$(tedge config get c8y.smartrest.templates | jq -r ". += [\"$SMARTREST_TEMPLATE_NAME\"]| unique | @tsv" | tr '\t' ',')
    tedge config set c8y.smartrest.templates "$NEW_VALUE"
    ```

    **Note** If the user would not have `jq` then the logic was be more complex (and naive):

    ```sh
    SMARTREST_TEMPLATE_NAME=collection2

[...]

PR #2943, opened 2024-06-17
feat(config): add new subcommand for appending/removing values in an array-type tedge configs

Ruadhri17:
## Proposed changes
This PR introduces two new tedge config commands: `add` and `remove`. User can append array type tedge config using the `tedge config add <key> <value>`. If this command is used on non array type config, it will work just like `tedge config set` and replace the value. The `tedge config remove <key> <value>` will remove one or more elements from the array type if they match the <value>. If it is used on non array type config, it will unset the config, provided `<value>` match it.
 

## Types of changes

## Paste Link to the issue
* #2864

<br/>

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,6 +82,7 @@ hyper-rustls = { version = "0.24", default_features = false, features = [
     "http1",
     "http2",
 ] }
+itertools = "0.13"
 json-writer = { path = "crates/common/json_writer" }
 lazy_static = "1.4"
 log = "0.4"
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -318,6 +318,88 @@ impl TEdgeTomlVersion {
     }
 }
 
+pub trait AppendRemoveItem {
+    type Item;
+
+    fn append(current_value: Option<Self::Item>, new_value: Self::Item) -> Option<Self::Item>;
+
+    fn remove(current_value: Option<Self::Item>, remove_value: Self::Item) -> Option<Self::Item>;
+}
+
+macro_rules! impl_append_remove_for_single_value {
+    ($($type:ty),*) => {
+        $(
+            impl AppendRemoveItem for $type {
+                type Item = $type;
+
+                fn append(_current_value: Option<Self::Item>, new_value: Self::Item) -> Option<Self::Item> {
+                    Some(new_value)
+                }
+
+                fn remove(current_value: Option<Self::Item>, remove_value: Self::Item) -> Option<Self::Item> {
+                    match current_value {
+                        Some(current) if current == remove_value => None,
+                        _ => current_value
+                    }
+                }
+            }
+        )*
+    }
+}
+
+impl_append_remove_for_single_value!(
+    Utf8PathBuf,
+    String,
+    ConnectUrl,
+    HostPort<HTTPS_PORT>,
+    HostPort<MQTT_TLS_PORT>,
+    bool,
+    IpAddr,
+    u16,
+    Arc<str>,
+    AutoFlag,
+    TopicPrefix,
+    SoftwareManagementApiFlag,
+    AutoLogUpload,
+    TimeFormat,
+    NonZeroU16,
+    SecondsOrHumanTime,
+    u32,
+    AptConfig
+);
+
+impl AppendRemoveItem for TemplatesSet {
+    type Item = TemplatesSet;
+
+    fn append(current_value: Option<Self::Item>, new_value: Self::Item) -> Option<Self::Item> {
+        if let Some(current_value) = current_value {
+            Some(TemplatesSet(
+                current_value
+                    .0
+                    .into_iter()
+                    .chain(new_value.0)
+                    .collect::<std::collections::BTreeSet<String>>()
+                    .into_iter()
+                    .collect(),
+            ))
+        } else {
+            Some(new_value)
+        }
+    }
+
+    fn remove(current_value: Option<Self::Item>, remove_value: Self::Item) -> Option<Self::Item> {
+        let mut current_value = current_value;
+
+        if let Some(ref mut current_value) = current_value {
+            let to_remove = std::collections::BTreeSet::from_iter(remove_value.0);
+
+            current_value.0.retain(|value| !to_remove.contains(value));
+        }
+
+        current_value
+    }
+}
+
 define_tedge_config! {
     #[tedge_config(reader(skip))]
     config: {
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -318,6 +318,10 @@ impl TEdgeTomlVersion {
     }
 }
 
+#[diagnostic::on_unimplemented(
+    message = "To use `{Self}` as a tedge config type, it must implement the `AppendRemoveItem` trait",
+    note = "This can be done using impl_append_remove_for_single_value! macro"
+)]
 pub trait AppendRemoveItem {
     type Item;
 
--- a/crates/common/tedge_config_macros/examples/macro.rs
+++ b/crates/common/tedge_config_macros/examples/macro.rs
@@ -15,6 +15,26 @@ pub enum ReadError {
     GenericError(String),
 }
 
+pub trait AppendRemoveItem {
+    type Item;
+
+    fn append(current_value: Option<Self::Item>, new_value: Self::Item) -> Option<Self::Item>;
+
+    fn remove(current_value: Option<Self::Item>, remove_value: Self::Item) -> Option<Self::Item>;
+}
+
+impl<T> AppendRemoveItem for T {
+    type Item = T;
+
+    fn append(_current_value: Option<Self::Item>, _new_value: Self::Item) -> Option<Self::Item> {
+        unimplemented!()
+    }
+
+    fn remove(_current_value: Option<Self::Item>, _remove_value: Self::Item) -> Option<Self::Item> {
+        unimplemented!()
+    }
+}
+
 define_tedge_config! {
     #[tedge_config(reader(skip))]
     config: {
--- a/crates/common/tedge_config_macros/examples/renaming_keys.rs
+++ b/crates/common/tedge_config_macros/examples/renaming_keys.rs
@@ -10,6 +10,26 @@ pub enum ReadError {
     GenericError(String),
 }
 
+pub trait AppendRemoveItem {
+    type Item;
+
+    fn append(current_value: Option<Self::Item>, new_value: Self::Item) -> Option<Self::Item>;
+
+    fn remove(current_value: Option<Self::Item>, remove_value: Self::Item) -> Option<Self::Item>;
+}
+
+impl<T> AppendRemoveItem for T {
+    type Item = T;
+
+    fn append(_current_value: Option<Self::Item>, _new_value: Self::Item) -> Option<Self::Item> {
+        unimplemented!()
+    }
+
+    fn remove(_current_value: Option<Self::Item>, _remove_value: Self::Item) -> Option<Self::Item> {
+        unimplemented!()
+    }
+}
+
 define_tedge_config! {
     device: {
         // Fields and groups can be renamed using #[tedge_config(rename)]
--- a/crates/common/tedge_config_macros/impl/src/query.rs
+++ b/crates/common/tedge_config_macros/impl/src/query.rs
@@ -2,6 +2,7 @@ use crate::error::extract_type_from_result;
 use crate::input::ConfigurableField;
 use crate::input::FieldOrGroup;
 use heck::ToUpperCamelCase;
+use itertools::Itertools;
 use proc_macro2::TokenStream;
 use quote::quote;
 use quote::quote_spanned;
@@ -337,7 +338,12 @@ fn generate_string_readers(paths: &[VecDeque<&FieldOrGroup>]) -> TokenStream {
 
 fn generate_string_writers(paths: &[VecDeque<&FieldOrGroup>]) -> TokenStream {
     let variant_names = paths.iter().map(variant_name);
-    let (update_arms, unset_arms): (Vec<syn::Arm>, Vec<syn::Arm>) = paths
+    let (update_arms, unset_arms, append_arms, remove_arms): (
+        Vec<syn::Arm>,
+        Vec<syn::Arm>,
+        Vec<syn::Arm>,
+        Vec<syn::Arm>,
+    ) = paths
         .iter()
         .zip(variant_names)
         .map(|(path, variant_name)| {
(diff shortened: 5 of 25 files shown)


crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -318,6 +318,78 @@ impl TEdgeTomlVersion {
     }
 }
 
+pub trait AppendRemoveItem {
```
jarhodes314, line 321: I'm wondering whether we actually do want to implement this (and the accompanying macro for implementing it) in `tedge_config_macros/src/...` as it does need to be used in so many examples.

In order to help people find the relevant place to implement the trait (as well as the macro that likely does it for them), we could use a [custom diagnostic attribute](https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#diagnostic-attributes).

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -318,6 +318,78 @@ impl TEdgeTomlVersion {
...
+    type Item;
+
+    fn append(current_value: &mut Option<Self::Item>, new_value: Self::Item);
+
+    fn remove(current_value: &mut Option<Self::Item>, remove_value: Self::Item);
+}
+
+macro_rules! append_remove_for_single_value {
```
jarhodes314, line 329:
```suggestion
macro_rules! impl_append_remove_for_single_value {
```

crates/common/tedge_config_macros/examples/macro.rs
```diff
@@ -15,6 +15,45 @@ pub enum ReadError {
...
+append_remove_for_single_value!(
+    u32,
+    String,
+    Utf8PathBuf,
+    ConnectUrl,
+    bool,
+    IpAddr,
+    NonZeroU16,
+    u16
+);
```
jarhodes314, line 55: As an alternative to my previous suggestion of moving the trait into the `tedge_config_macros` crate, we could avoid the duplication by providing a stub implementation here as the relevant code will never be called.

```suggestion
impl<T> AppendRemoveItem for T {
    type Item = T;
    
    fn append(_current_value: &mut Option<Self::Item>, _new_value: Self::Item) {
        unimplemented!()
    }
    
    fn remove(_current_value: &mut Option<Self::Item>, _remove_value: Self::Item) {
        unimplemented!()
    }
}
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
jarhodes314: It would also be useful to add some documentation of what this is for (i.e. making `tedge config add/remove` work)

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -318,6 +319,57 @@ impl TEdgeTomlVersion {
...
+    type Item = T;
+
+    fn append(current_value: &mut Option<Self::Item>, new_value: Self::Item) {
+        if let Some(current_value) = current_value {
+            let ty_curr = current_value as &mut dyn Any;
+            let ty_new = &new_value as &dyn Any;
+
+            if let (Some(current), Some(new)) = (
+                ty_curr.downcast_mut::<TemplatesSet>(),
+                ty_new.downcast_ref::<TemplatesSet>(),
```
jarhodes314, line 340: I'm not convinced by using reflection here, partially because it means if we add more array-like configurations, we also have to remember this exists and cope appropriately. I'm also a bit wary of using reflection as it's based on the exact type used (so if the input was a `TemplatesSet` in the form of some dynamic trait object, we wouldn't be able to downcast to `TemplatesSet`).

There isn't any reason I can see why we need to have a generic implementation here, so this may have been a misunderstanding of my previous comments?

tests/RobotFramework/tests/tedge/tedge_config_add_remove.robot
```diff
@@ -0,0 +1,59 @@
...
+tedge config add/remove value in non-empty array type config
+    Execute Command    sudo tedge config set c8y.smartrest.templates 1
+    ${set}    Execute Command    tedge config list
+    Should Contain    ${set}    c8y.smartrest.templates=["1"]
+
+    Execute Command    sudo tedge config add c8y.smartrest.templates 2,3
+    ${add}    Execute Command    tedge config list
+    Should Contain    ${add}    c8y.smartrest.templates=["1", "2", "3"]
+
+    Execute Command    sudo tedge config remove c8y.smartrest.templates 1,2,3
```
didier-wenzek, line 30: Before removing all the values, I test the case where we just remove one value, say "2".

crates/common/tedge_config_macros/examples/renaming_keys.rs
```diff
@@ -10,6 +10,26 @@ pub enum ReadError {
...
+    type Item = T;
+
+    fn append(_current_value: &mut Option<Self::Item>, _new_value: Self::Item) {
+        unimplemented!()
+    }
+
+    fn remove(_current_value: &mut Option<Self::Item>, _remove_value: Self::Item) {
+        unimplemented!()
+    }
+}
```
didier-wenzek, line 31: I don't get the point of this local unimplemented trait?

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
didier-wenzek: The implementation cannot be specific to `c8y.smartrest.templates` and the `TemplateSet` type.

One should be able to use that for any setting that is a list, e.g. ` az.topics`

```
$ tedge config get az.topics                             
["te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+/+/+/+/status/health"]

$ tedge config add az.topics azure

$ tedge config get az.topics           
["azure"]
```

I was expecting `"te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+/+/+/+/status/health","azure"]` and not `["azure"]`

Ruadhri17: > I'm not convinced by using reflection here, partially because it means if we add more array-like configurations, we also have to remember this exists and cope appropriately. I'm also a bit wary of using reflection as it's based on the exact type used (so if the input was a `TemplatesSet` in the form of some dynamic trait object, we wouldn't be able to downcast to `TemplatesSet`). 
> 
> There isn't any reason I can see why we need to have a generic implementation here, so this may have been a misunderstanding of my previous comments?

I thought it would be easier to maintain the generic implementation. In first implementation you had to fill macro with all the types (and then remember to append it if new type was added). But you are right, I haven't thought about the dynamic trait object. With additional array-like configurations you could modify current solution to check all array types with `is::<TemplateSet>()` in if..else blocks but I see it can get pretty clunky over time. I will reverse current solution to previous one with macro.

> The implementation cannot be specific to `c8y.smartrest.templates` and the `TemplateSet` type.
> 
> One should be able to use that for any setting that is a list, e.g. ` az.topics`

AFAIK, the `TemplateSet` is the only array type field in `tedge config` (All topic fields use `TemplateSet`). Apparently, there is an issue with default value that it is not present when I am checking current value of the fields, therefore instead of appending, it replace the value. I did not notice that with `c8y.smartrest.templates` because it is empty by default. I need to investigate how exactly default value is assigned.

crates/common/tedge_config_macros/examples/renaming_keys.rs
Ruadhri17: Macro calls either `append` or `remove` from `AppendRemoveItem` trait like [here](https://github.com/thin-edge/thin-edge.io/pull/2943/commits/c8f74617fd55bfdf882cfd1ae2c0d150d736f217#diff-cd95e9aea2dc9e842e4cd738eaf06c1e284ccab0754a684defe0851b188a7faeR373). We can't implement it inside the macro as it does not know about the types that are provided. Instead, local implementation of that trait is needed but we have to do that in every module that has `tedge_config` defined as sharing it between different modules is not possible. And we don't actually need to implement `append` or `remove` in those test modules because It will never be called there (but macro needs to be aware of them) thus we used `unimplemented!()`.

tests/RobotFramework/tests/tedge/tedge_config_add_remove.robot
```diff
@@ -0,0 +1,59 @@
...
+*** Test Cases ***
+tedge config add/remove value in empty array type config
+    Execute Command    sudo tedge config add c8y.smartrest.templates 1
+    ${add}    Execute Command    tedge config list
+    Should Contain    ${add}    c8y.smartrest.templates=["1"]
+
+    Execute Command    sudo tedge config remove c8y.smartrest.templates 1
+    ${remove}    Execute Command    tedge config list
+    Should Contain    ${remove}    c8y.smartrest.templates=[]
+
```
gligorisaev, line 20: I would suggest checking the initial state in order to ensure consistency and to validate that the setup is correct, something like:
```
tedge config add/remove value in empty array type config
    # Verify initial state
    ${initial}    Execute Command    tedge config list
    Should Not Contain    ${initial}    c8y.smartrest.templates

    # Add value to empty array
    Execute Command    sudo tedge config add c8y.smartrest.templates 1
    ${add}    Execute Command    tedge config list
    Should Contain    ${add}    c8y.smartrest.templates=["1"]

    # Remove value
    Execute Command    sudo tedge config remove c8y.smartrest.templates 1
    ${remove}    Execute Command    tedge config list
    Should Contain    ${remove}    c8y.smartrest.templates=[]

File: crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs (hunk 4 above)
Comment (didier-wenzek) [reply crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs:340]: >> The implementation cannot be specific to `c8y.smartrest.templates` and the `TemplateSet` type.
>> 
>> One should be able to use that for any setting that is a list, e.g. ` az.topics`

> AFAIK, the `TemplateSet` is the only array type field in `tedge config` (All topic fields use `TemplateSet`). Apparently, there is an issue with default value that it is not present when I am checking current value of the fields, therefore instead of appending, it replace the value. I did not notice that with `c8y.smartrest.templates` because it is empty by default. I need to investigate how exactly default value is assigned.

You are correct. There is an issue *only* when the list has not been explicitly set by the user, but provided as a default. This is working fine once a first list has been set by the user.

```
$ sudo tedge config set az.topics 'foo,bar'
$ tedge config get az.topics
["foo", "bar"]
$ sudo tedge config add az.topics 'azure'  
$ tedge config get az.topics             
["foo", "bar", "azure"]
```

I have been also positively surprised to see that *append* is working too:

```
$ sudo tedge config add az.topics 'xyz,abc'
$ tedge config get az.topics               
["foo", "bar", "azure", "xyz", "abc"]
```

File: crates/common/tedge_config_macros/impl/src/query.rs [hunk 8]
```diff
@@ -363,9 +381,26 @@ fn generate_string_writers(paths: &[VecDeque<&FieldOrGroup>]) -> TokenStream {
...
+                    WritableKey::#variant_name => self.#(#segments).* = <#ty as AppendRemoveItem>::append(
+                        self.#(#segments).*.to_owned().or_else(|| #default_value),
+                        value
+                        .#parse
+                        .#convert_to_field_ty
+                        .map_err(|e| WriteError::ParseValue(Box::new(e)))?),
+                },
+                parse_quote_spanned! {ty.span()=>
+                    WritableKey::#variant_name => self.#(#segments).* = <#ty as AppendRemoveItem>::remove(
+                        self.#(#segments).*.to_owned().or_else(|| #default_value),
```
jarhodes314, line 394: Is there a reason why we try reading from the DTO in the first instance and then the reader if that's `None` rather than just using the value from the reader in any case?

Ruadhri17: No, we can actually directly provide values from reader to append/remove functions. This is a leftover from previous implementation where I wanted to edit the dto value if it was `None` and provide it as `&mut` argument. However, in the mean time I changed the implementation and forgot to simplify it.

crates/common/tedge_config_macros/impl/src/query.rs
```diff
@@ -363,9 +381,26 @@ fn generate_string_writers(paths: &[VecDeque<&FieldOrGroup>]) -> TokenStream {
...
+                    WritableKey::#variant_name => self.#(#segments).* = <#ty as AppendRemoveItem>::append(
+                        #default_value,
+                        value
+                        .#parse
+                        .#convert_to_field_ty
+                        .map_err(|e| WriteError::ParseValue(Box::new(e)))?),
+                },
+                parse_quote_spanned! {ty.span()=>
+                    WritableKey::#variant_name => self.#(#segments).* = <#ty as AppendRemoveItem>::remove(
+                        #default_value,
```
jarhodes314, line 394: Now we're using it unconditionally, we should call it something like `reader_value` or `current_value`

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -318,6 +318,88 @@ impl TEdgeTomlVersion {
...
+                        Some(current) if current == remove_value => None,
+                        _ => current_value
+                    }
+                }
+            }
+        )*
+    }
+}
+
+impl_append_remove_for_single_value!(
```
didier-wenzek, line 350: As suggested by @jarhodes314, it would be really good to add a [custom diagnostic](https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#diagnostic-attributes) to the `AppendRemoveItem` telling the type of a config setting should listed here.

To have something like that:

```
error[E0277]: To be used as a tedge config type, connect_url::ConnectUrl must implement the trait `AppendRemoveItem`
   --> crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs:617:14
    |
617 |         url: ConnectUrl,
    |              ^^^^^^^^^^ the trait `AppendRemoveItem` is not implemented for `connect_url::ConnectUrl`
    |
    = note: This can be done using impl_append_remove_for_single_value!
```

instead of:

```
error[E0277]: the trait bound `connect_url::ConnectUrl: AppendRemoveItem` is not satisfied
   --> crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs:614:14
    |
614 |         url: ConnectUrl,
    |              ^^^^^^^^^^ the trait `AppendRemoveItem` is not implemented for `connect_url::ConnectUrl`

```

didier-wenzek (commented):
I would be happy to see this PR merged. This is working as expected. However, as suggested by @jarhodes314 it to be good to:

- document the purpose of the `AppendRemoveItem` trait
- add a custom diagnostic when an implementation is missing

Cargo.toml
```diff
@@ -14,7 +14,7 @@ resolver = "2"
 version = "1.1.1"
 authors = ["thin-edge.io team <info@thin-edge.io>"]
 edition = "2021"
-rust-version = "1.70"
+rust-version = "1.78"
```
didier-wenzek, line 17: You also have to update [the version used by CI](https://github.com/thin-edge/thin-edge.io/blob/main/.github/workflows/build-workflow.yml#L62)

didier-wenzek: Sorry, I noticed a bit that this is what you propose with https://github.com/thin-edge/thin-edge.io/pull/2965

Ruadhri17: Yes, I prefer to do MSRV update in separate PR so I will reverse that change of `rust-version` here.

didier-wenzek: Yes, perfect

didier-wenzek (approved):
Approved. Thank you. Upgrading the tedge config macro has not been a simple task, but well done and the outcome is really handy for the users.

PR #2944, opened 2024-06-17
test: ensure the MQTT connection is properly closed

gligorisaev:
Adding test:

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2922

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
+++ b/tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
@@ -1,0 +1,65 @@
+*** Settings ***
+Resource        ../../resources/common.resource
+Library         ThinEdgeIO
+Library         OperatingSystem
+Library         String
+
+Test Setup     Setup    skip_bootstrap=True
+Suite Teardown  Get Logs
+
+Test Tags      theme:mqtt
+
+
+*** Test Cases ***
+
+Publish Connection Closed after publish
+    [Documentation]    Tests that the connection to the MQTT broker is closed after publishing.
+    Execute Command    /setup/bootstrap.sh
+    Execute Command    tedge mqtt pub test/topic Hello
+    ${MQTT_PUB}    Execute Command    journalctl -u mosquitto -n 30
+    Should Contain    ${MQTT_PUB}    Received DISCONNECT from tedge-pub
+
+Publish Connection Closed after publish and no error message with TLS
+    [Documentation]    same as previous test but with TLS
+    Execute Command    /setup/bootstrap.sh --no-connect --no-secure
+    Set up broker with server and client authentication
+    tedge configure MQTT server authentication
+    tedge configure MQTT client authentication
+    Execute Command    tedge mqtt pub test/topic Hello
+    ${MQTT_PUB}    Execute Command    journalctl -u mosquitto -n 30
+    Should Contain    ${MQTT_PUB}    Received DISCONNECT from tedge-pub
+    Should Not Contain    ${MQTT_PUB}    Error
+
+Subscribe Connection Closed On Interruption
+    [Documentation]    Tests that a subscribe connection to MQTT broker closes upon interruption.
+    Execute Command    /setup/bootstrap.sh
+    Execute Command    timeout 2 tedge mqtt sub test/topic    ignore_exit_code=True
+    ${MQTT_SUB}    Execute Command    journalctl -u mosquitto -n 30
+    Should Contain    ${MQTT_SUB}    Received DISCONNECT from tedge-sub
+
+Stop subscription on SIGINT even when broker is not available
+    Execute Command    /setup/bootstrap.sh
+    Execute Command    sudo systemctl stop mosquitto
+    ${output}=     Execute Command    cmd=timeout --signal=SIGINT 2 tedge mqtt sub '#'    timeout=5    stderr=${True}    stdout=${False}    ignore_exit_code=${True}
+    Should Contain    ${output}    Connection refused
+
+
+
+*** Keywords ***
+
+Set up broker with server and client authentication
+    Transfer To Device    ${CURDIR}/mosquitto-client-auth.conf    /etc/tedge/mosquitto-conf/mosquitto-client-auth.conf
+    Execute Command
+    ...    mv /etc/tedge/mosquitto-conf/mosquitto-client-auth.conf /etc/tedge/mosquitto-conf/tedge-mosquitto.conf
+    Transfer To Device    ${CURDIR}/gen_certs.sh    /setup/gen_certs.sh
+    Execute Command    chmod u+x /setup/gen_certs.sh
+    Execute Command    /setup/gen_certs.sh
+    Restart Service    mosquitto
+
+tedge configure MQTT server authentication
+    Execute Command    tedge config set mqtt.client.port 8883
+    Execute Command    tedge config set mqtt.client.auth.ca_file /etc/mosquitto/ca_certificates/ca.crt
+
+tedge configure MQTT client authentication
+    Execute Command    tedge config set mqtt.client.auth.cert_file client.crt
+    Execute Command    tedge config set mqtt.client.auth.key_file client.key


rina23q (approved):
LGTM

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,71 @@
...
+
+*** Keywords ***
+Custom Setup
+    Setup    skip_bootstrap=True
+    Execute Command    /setup/bootstrap.sh --no-connect --no-secure
+
+Set up broker with server and client authentication
+    Transfer To Device    ${CURDIR}/mosquitto-client-auth.conf    /etc/tedge/mosquitto-conf/mosquitto-client-auth.conf
+    Execute Command
+    ...    mv /etc/tedge/mosquitto-conf/mosquitto-client-auth.conf /etc/tedge/mosquitto-conf/tedge-mosquitto.conf
```
didier-wenzek, line 60: Why not directly move the file to its destination ?

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,71 @@
...
+
+*** Test Cases ***
+
+Validate no Error message in journalctl
+    [Documentation]    This Test step is inspired by 'tedge mqtt cli commands do not disconnect cleanly from MQTT broker #2922' bug
+    ...                extended to check for any possible error that may appear in the journalctl
+    Set up broker with server and client authentication
+    Restart Service    mosquitto
+    tedge configure MQTT server authentication
+    tedge configure MQTT client authentication
```
didier-wenzek, line 26: I would move this setup out into a Suite Setup used by all the tests  of this suite.

didier-wenzek (approved):
Approved

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,71 @@
...
+    Sleep    2s    reason=Allow some time for connection
+    Kill Process    ${pub_process}
+    Process Should Not Be Running    ${pub_process}
+    Log    MQTT publish connection was closed upon interruption.
+
+Subscribe Connection Closed On Interruption
+    [Documentation]    Tests that a subscribe connection to MQTT broker closes upon interruption.
+    ${sub_process}=    Execute Command    tedge mqtt sub ${TOPIC} & echo $!   return_stdout=True
+    Sleep    2s    reason=Allow some time for connection
+    Kill Process    ${sub_process}
```
reubenmiller, line 47: It would be easier just to use the `timeout` command for this,

For example:

```
Execute Command    timeout 2 tedge mqtt sub ${TOPIC}    ignore_exit_code=${True}
```

Though the test case is not to check if the process is killed (as this is a given), but if a clean MQTT disconnect is sent before the process exit.s

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,71 @@
...
+Publish Connection Closed On Interruption
+    [Documentation]    Tests that a publish connection to MQTT broker closes upon interruption.
+    Set up broker with server and client authentication
+    Restart Service    mosquitto
+    tedge configure MQTT server authentication
+    tedge configure MQTT client authentication
+    ${pub_process}=    Execute Command    tedge mqtt pub ${TOPIC} ${MESSAGE} & echo $!    return_stdout=True
+    Sleep    2s    reason=Allow some time for connection
+    Kill Process    ${pub_process}
+    Process Should Not Be Running    ${pub_process}
```
reubenmiller, line 40: I'm pretty sure the process will never take longer than 2 seconds to publish to the MQTT broker (unless if the connection to the broker is purposfully misconfigured?)

The goal of the test should be to check if a clean MQTT disconnect is sent (can be checked by observing the log messages for error), so I would expect the assertion to include:

```
    ${log}    Execute Command    journalctl -u mosquitto -n 30
    Should Not Contain    ${log}    Error
```

reubenmiller (changes requested):
The tests can be simplified a bit, and some of the assertions should be changed (see comments).

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
reubenmiller: Is this file actually needed? The bootstrapping of the test should take care of the mosquitto setup, so I don't think we need the extract mosquitto-client-auth.conf setup. Port `1883` should be fine for these tests

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,56 @@
...
+    Execute Command    tedge mqtt pub foo bar
+    ${log}    Execute Command    journalctl -u mosquitto -n 30
+    Should Not Contain    ${log}    Error
+
+Publish Connection Closed On Interruption
+    [Documentation]    Tests that a publish connection to MQTT broker closes upon interruption.
+    Execute Command    tedge mqtt pub ${TOPIC} ${MESSAGE}    ignore_exit_code=True
+    ${MQTT_PUB}    Execute Command    journalctl -u mosquitto -n 30
+    Should Contain    ${MQTT_PUB}    Received DISCONNECT from tedge-pub
+    Log    MQTT publish connection was closed upon interruption.
```
reubenmiller, line 33: These log statements aren't required as they don't provide any value which is not already provided by the test case title.

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,56 @@
...
+
+Subscribe Connection Closed On Interruption
+    [Documentation]    Tests that a subscribe connection to MQTT broker closes upon interruption.
+    Execute Command    timeout 2 tedge mqtt sub ${TOPIC}    ignore_exit_code=True
+    ${MQTT_SUB}    Execute Command    journalctl -u mosquitto -n 30
+    Should Contain    ${MQTT_SUB}    Received DISCONNECT from tedge-sub
+    Log    MQTT subscribe connection was closed upon interruption.
```
reubenmiller, line 40: These log statements aren't required as they don't provide any value which is not already provided by the test case title.

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,56 @@
...
+
+Validate no Error message in journalctl
+    [Documentation]    This Test step is inspired by 'tedge mqtt cli commands do not disconnect cleanly from MQTT broker #2922' bug
+    ...                extended to check for any possible error that may appear in the journalctl
+    Execute Command    tedge mqtt pub foo bar
+    ${log}    Execute Command    journalctl -u mosquitto -n 30
+    Should Not Contain    ${log}    Error
+
+Publish Connection Closed On Interruption
+    [Documentation]    Tests that a publish connection to MQTT broker closes upon interruption.
```
reubenmiller, line 29: The test case title needs to change as we aren't technically interrupting the subscription (and it won't be feasible to do), just rephrase it to something like:

```
Publish Connection Closed after publish
```

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,56 @@
...
+    Execute Command    tedge mqtt pub ${TOPIC} ${MESSAGE}    ignore_exit_code=True
```
reubenmiller, line 30: We don't need the ignore_exit_code flag for this since weren't run using `timeout` to kill the process.

```suggestion
    Execute Command    tedge mqtt pub ${TOPIC} ${MESSAGE}
```

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,56 @@
...
+    ${MQTT_PUB}    Execute Command    journalctl -u mosquitto -n 30
+    Should Contain    ${MQTT_PUB}    Received DISCONNECT from tedge-pub
+    Log    MQTT publish connection was closed upon interruption.
+
+Subscribe Connection Closed On Interruption
+    [Documentation]    Tests that a subscribe connection to MQTT broker closes upon interruption.
+    Execute Command    timeout 2 tedge mqtt sub ${TOPIC}    ignore_exit_code=True
```
reubenmiller, line 37: Using a variable for `${TOPIC}` isn't really useful, and it just makes the test harder to read.

```suggestion
    Execute Command    timeout 2 tedge mqtt sub test/topic    ignore_exit_code=True
```

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,32 @@
...
+
+Validate no Error message in journalctl
+    [Documentation]    This Test step is inspired by 'tedge mqtt cli commands do not disconnect cleanly from MQTT broker #2922' bug
+    ...                extended to check for any possible error that may appear in the journalctl
+    Execute Command    tedge mqtt pub foo bar
+    ${log}    Execute Command    journalctl -u mosquitto -n 30
+    Should Not Contain    ${log}    Error
+
+Publish Connection Closed after publish  
+    [Documentation]    Tests that a publish connection to MQTT broker closes upon interruption.
```
reubenmiller, line 23: The doc string is inaccurate

```suggestion
    [Documentation]    Tests that the connection to the MQTT broker is closed after publishing.
```

reubenmiller (changes requested):
Things are looking good (just one minor comment left).

Though it might make sense adding the new test case given that https://github.com/thin-edge/thin-edge.io/issues/2949 is very similar topic.

For example you could add the following test case:

```
Stop subscription on SIGINT even when broker is not available
    Execute Command    sudo systemctl stop mosquitto
    Execute Command    cmd=timeout --signal=SIGINT 2 tedge mqtt sub '#'    timeout=5
```

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,56 @@
...
+    tedge configure MQTT server authentication
+    tedge configure MQTT client authentication
+
+tedge configure MQTT server authentication
+    Execute Command    tedge config set mqtt.client.port 8883
+    Execute Command    tedge config set mqtt.client.auth.ca_file /etc/mosquitto/ca_certificates/ca.crt
+
+tedge configure MQTT client authentication
+    Execute Command    tedge config set mqtt.client.auth.cert_file client.crt
+    Execute Command    tedge config set mqtt.client.auth.key_file client.key
```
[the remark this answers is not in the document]
didier-wenzek: > I would be surprised if these settings weren't actually already set by the bootstrap script...and we don't need to test TLS communication in these test cases.

Actually, [errors was only observed with TLS enabled](https://github.com/thin-edge/thin-edge.io/issues/2922).

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,32 @@
...
+
+
+*** Test Cases ***
+
+Validate no Error message in journalctl
+    [Documentation]    This Test step is inspired by 'tedge mqtt cli commands do not disconnect cleanly from MQTT broker #2922' bug
+    ...                extended to check for any possible error that may appear in the journalctl
+    Execute Command    tedge mqtt pub foo bar
+    ${log}    Execute Command    journalctl -u mosquitto -n 30
+    Should Not Contain    ${log}    Error
```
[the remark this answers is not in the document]
reubenmiller: Setting up TLS in not so straight forward and the bootstrap.sh script (used in the setup of the device under test) takes care of setting up TLS (if activated)...so doing this in a specific test to either opt-out or opt-in to TLS will definitely complicate the test considerably.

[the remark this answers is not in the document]
didier-wenzek: > I created the original ticket with TLS included just because I was using TLS on my device. I didn't explicitly test it without TLS, but I assume it would have no bearing on the outcome.

I tested without TLS when fixing the issue. There were no errors reported in mosquitto logs.

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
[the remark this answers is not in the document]
reubenmiller: This comment was from me, and what I meant by that is I only tested it with TLS, so I didn't want to assume that a non-TLS was also affected.

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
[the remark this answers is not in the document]
reubenmiller: Ok, then sorry for the confusion. I guess I was just thinking about the assertion which checks for the `DISCONNECT` message and not the absence of the `Error` messages.

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,65 @@
...
+
+*** Test Cases ***
+
+Publish Connection Closed after publish and no error message 
+    [Documentation]    Tests that the connection to the MQTT broker is closed after publishing.
+    Execute Command    /setup/bootstrap.sh
+    Execute Command    tedge mqtt pub test/topic Hello
+    ${MQTT_PUB}    Execute Command    journalctl -u mosquitto -n 30
+    Should Contain    ${MQTT_PUB}    Received DISCONNECT from tedge-pub
+    Should Not Contain    ${MQTT_PUB}    Error
```
reubenmiller, line 21: I'm pretty sure the bootstrap.sh script will use TLS if the global default is to use TLS (which I believe is the default).

And the assertion confirms this, as if a non-TLS connection was being used, then we wouldn't get the "Error" line in the log (as @didier-wenzek confirmed this earlier in a different discussion).

```
Should Not Contain ${MQTT_PUB} Error
```

So it might just be easier to remove the "Error" assertion altogether, or force using non-TLS setting when calling the bootstrap.sh

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,65 @@
...
+
+*** Keywords ***
+
+Set up broker with server and client authentication
+    Transfer To Device    ${CURDIR}/mosquitto-client-auth.conf    /etc/tedge/mosquitto-conf/mosquitto-client-auth.conf
+    Execute Command
+    ...    mv /etc/tedge/mosquitto-conf/mosquitto-client-auth.conf /etc/tedge/mosquitto-conf/tedge-mosquitto.conf
+    Transfer To Device    ${CURDIR}/gen_certs.sh    /root/gen_certs.sh
+    Execute Command    chmod u+x /root/gen_certs.sh
+    Execute Command    /root/gen_certs.sh
```
reubenmiller, line 57: A more portable solution would be to copy the file to the `/tmp` directory, as you're assuming the test is always running as root.

```suggestion
    Execute Command    chmod u+x /tmp/gen_certs.sh
    Execute Command    /tmp/gen_certs.sh
```

gligorisaev: instead of `/tmp` I am using `/setup` because, from some reason that I could not find, `/tmp` is not writtable

reubenmiller:
FYI: I found out why you can't transfer files directly to `/tmp` (which is mounted as a tmpfs), it is because copying files to a tmpfs is not supported, see the docker docs: https://docs.docker.com/reference/cli/docker/container/cp/#corner-cases

It is very unfortunate that from the docker side there is absolutely no indication that it didn't work, everything registers as being successful even though it did not copy any files to the target directory.

tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot
```diff
@@ -0,0 +1,65 @@
...
+
+Test Setup     Setup    skip_bootstrap=True
+Suite Teardown  Get Logs
+
+Test Tags      theme:mqtt
+
+
+*** Test Cases ***
+
+Publish Connection Closed after publish and no error message
```
reubenmiller, line 15: Test title is misaligned to the content

```suggestion
Publish Connection Closed after publish
```

reubenmiller (approved):
Approved, everything is looking in good shape

PR #2946, opened 2024-06-17
fix: Remove unwanted host header before forwarding request to Cumulocity

jarhodes314:
## Proposed changes
~This should fix a customer issue we have observed seemingly due to duplicate host headers from the c8y-auth-proxy. Currently it just removes the headers, although I am currently trying to reproduce the issue and write a test to verify it is now actually fixed.~

This fixes a bug in `c8y_auth_proxy` where the original headers were sent unmodified, meaning the Host header that was sent with the request originated from the request to the proxy, not the request the proxy makes to Cumulocity. This meant the request had a host of `127.0.0.1:8001` rather than the host that we were proxying to (e.g. `xyz.cumulocity.com`).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_auth_proxy/src/server.rs
+++ b/crates/extensions/c8y_auth_proxy/src/server.rs
@@ -221,13 +221,12 @@ async fn connect_to_websocket(
 ) -> Result<WebSocketStream<MaybeTlsStream<TcpStream>>, tokio_tungstenite::tungstenite::Error> {
     let mut req = Request::builder();
     for (name, value) in headers {
-        if name != HOST {
-            req = req.header(name.as_str(), value);
-        }
+        req = req.header(name.as_str(), value);
     }
     req = req.header("Authorization", format!("Bearer {token}"));
     let req = req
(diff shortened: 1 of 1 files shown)


crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -225,6 +226,7 @@ async fn connect_to_websocket(
     req = req.header("Authorization", format!("Bearer {token}"));
     let req = req
         .uri(uri)
+        .header(HOST, url::Url::parse(uri).expect("URI must have valid host to reach c8y proxy").host_str().unwrap())
```
jarhodes314, line 229: There is likely a tidier way to achieve this, I'll work on it while I try and add some unit tests. From what I can tell, `tokio_tungstenite` requires us to add the host header in manually, which was why the unit tests were failing after removing the incorrect host header copying as we then had no host header for outgoing websocket requests.

didier-wenzek (approved):
Approved

Issue #2949, opened 2024-06-18
tedge mqtt sub ignores ctrl+c if MQTT broker is not available


reubenmiller:
**Describe the bug**

`tedge mqtt sub` does not react to `ctrl+c` (`SIGINT`) when it is in a reconnect loop when the MQTT broker is not available.

```sh
# tedge mqtt sub '#'
ERROR: I/O: Address not available (os error 99)
ERROR: I/O: Address not available (os error 99)
ERROR: I/O: Address not available (os error 99)
ERROR: I/O: Address not available (os error 99)
^CERROR: I/O: Address not available (os error 99)
ERROR: I/O: Address not available (os error 99)
^C^C^CERROR: I/O: Address not available (os error 99)
ERROR: I/O: Address not available (os error 99)
```

**To Reproduce**

1. Stop the MQTT broker

    ```sh
    # if using systemctl and mosquitto
    sudo systemctl stop mosquitto
    ```

3. Subscribe to any topic

    ```sh
    tedge mqtt sub '#'
    ```

4. Try and stop the process using ctrl+c

The robot test would look something like this:

```
Stops the subscription on SIGINT even when broker does not exist
    Execute Command    sudo systemctl stop mosquitto
    Execute Command    cmd=timeout --signal=SIGINT 2 tedge mqtt sub '#'    timeout=5
```

**Expected behavior**

* `tedge mqtt sub` should stop on Ctrl+C (SIGINT).

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`unknown`|
|System-Architecture|`Linux e7f905341c36 6.5.0-41-generic #41-Ubuntu SMP PREEMPT_DYNAMIC Mon May 20 22:58:57 UTC 2024 aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.1.2~124+g8502b28`|

**Additional context**

gligorisaev:
Bug is not reproducable anymore
Test Step is added as part of:
https://github.com/thin-edge/thin-edge.io/pull/2944

PR #2950, opened 2024-06-18
fix: tedge mqtt ignoring ctrl-c when connection fails

didier-wenzek:
## Proposed changes

As rumqttc doesn't trigger a client disconnect when the connection is not established,
one has to combine a client disconnect with the use of an `AtomicBool` to break the event loop
if for some reason the connection is not established when the user hits ctrl-c. 

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2949

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/mqtt/subscribe.rs
+++ b/crates/core/tedge/src/cli/mqtt/subscribe.rs
@@ -11,6 +11,7 @@ use rumqttc::MqttOptions;
 use rumqttc::Outgoing;
 use rumqttc::Packet;
 use rumqttc::QoS;
+use std::sync::atomic::Ordering;
 use tedge_config::MqttAuthClientConfig;
 
 const DEFAULT_QUEUE_CAPACITY: usize = 10;
@@ -83,7 +84,7 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), MqttError> {
     }
 
     let (mut client, mut connection) = Client::new(options, DEFAULT_QUEUE_CAPACITY);
-    super::disconnect_if_interrupted(client.clone());
+    let interrupted = super::disconnect_if_interrupted(client.clone());
 
     for event in connection.iter() {
         match event {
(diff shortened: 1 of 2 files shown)


crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -117,12 +118,11 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), MqttError> {
                 eprintln!("INFO: Connected");
                 client.subscribe(cmd.topic.as_str(), cmd.qos).unwrap();
             }
-            // TODO: should we keep trying to reconnect for all errors, or just
-            // if the broker isn't up and abort when e.g. we receive connection
-            // refused?
             Err(err) => {
+                if interrupted.load(Ordering::Relaxed) {
```
albinsuresh, line 122: Since the performance of these utilities is not at all a concern, but immediate response to the `Ctrl + C` command is desired, isn't it better to go with a stronger ordering guarantee using `Ordering::Acquire` and `Ordering::Release` so that the program closes as soon as `Ctrl + C` is pressed?

albinsuresh (approved):

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -117,12 +118,11 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), MqttError> {
...
+                    break;
```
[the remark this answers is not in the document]
didier-wenzek: Yes, the point is to handle long running subscribers ignoring connection blips unless the user requests to break.

crates/core/tedge/src/cli/mqtt/subscribe.rs
didier-wenzek: Yeah. A user will not notice the difference. Besides, the main delay is introduced by the event loop. The break will occur only after a connection error is reported.

reubenmiller: The delay (of up to 1 second) between pressing `ctrl+c` is on there when the MQTT broker is not reachable, so I don't think the user will run into it very often, so it's ok from UX perspective.

reubenmiller (approved):
Approved

PR #2952, opened 2024-06-19
feat: use loopback ip address instead of alias to improve compatibility

reubenmiller:
## Proposed changes

Some Linux distributions installations where the `/etc/hosts` file is not defined, the loopback address (`127.0.0.1`), the `localhost` alias does not get resolved to `127.0.0.1`. And devices which have a read-only rootfs, can't even add a manual entry to the list, so the only option is to change the `tedge.toml` values (which is not ideal for such basic settings).

Changing the defaults from `localhost` to `127.0.0.1` improves the out of the box compatibility.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/configuration/contrib/collectd/collectd.conf
+++ b/configuration/contrib/collectd/collectd.conf
@@ -276,7 +276,7 @@ LoadPlugin unixsock
 
 <Plugin mqtt>
 	<Publish "name">
-		Host "localhost"
+		Host "127.0.0.1"
 		Port 1883
 		ClientId "tedge-collectd"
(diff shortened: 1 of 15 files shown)


jarhodes314 (approved):

reubenmiller:
Ok, seems like the mosquitto TLS cert generation might need to change to also support TLS connections when using the loopback address, currently it only supports the hostname, or `localhost`.

didier-wenzek (approved):
Approved

PR #2961, opened 2024-06-25
feat: tedge cert upload usability improvements

reubenmiller:
## Proposed changes

Small usability improvements around the `tedge cert upload c8y` command:

* Support reading the username from the env variable via `C8Y_USER` (if provided)
* If user is not provided by the `--user` flag, or the `C8Y_USER`, then the user will be prompted to enter their username => **Reason:** Users generally don't know that they have to provide the `--user` flag, so it is more intuitive to prompt them for the missing information rather than throwing an error about the missing `--user` flag
* Password can also be set via the more common env variable, `C8Y_PASSWORD`, though `C8YPASS` is still supported, but it has been deprecated and a warning is printed to the console => **Reason:** other Cumulocity tooling (like go-c8y-cli) uses C8Y_PASSWORD which means the tools work well with each other

The new changes makes the following usage possible:

```sh
# Prompt for both username and password
tedge cert upload c8y

# Reuse go-c8y-cli session variables (which provide the username/password)
export C8Y_USER="foo"
export C8Y_PASSWORD="bar"
tedge cert upload c8y

# Use the existing method (the user is prompted for the password if the C8Y_PASSWORD env var is not defined)
tedge cert upload c8y --user "foo"

# Force prompting of the password (regardless if C8Y_PASSWORD env var is defined)
tedge cert upload c8y --user "foo" --password ""
```

Since the new code is using clap to handle the reading of the values from either the flags, or env variables, the default --help text is much more helpful now and the each flag has the linked env variable documented:

```sh
Upload root certificate to Cumulocity

The command will upload root certificate to Cumulocity.

Usage: tedge cert upload c8y [OPTIONS]

Options:
      --user <USERNAME>
          Provided username should be a Cumulocity IoT user with tenant management permissions. You will be prompted for input if the value is not provided or is empty
          
          [env: C8Y_USER]

      --password <PASSWORD>
          Cumulocity IoT Password. You will be prompted for input if the value is not provided or is empty
          
          Notes: `C8YPASS` is deprecated. Please use the `C8Y_PASSWORD` env variable instead
          
          [env: C8Y_PASSWORD]

      --config-dir <CONFIG_DIR>
          [default: /etc/tedge]

  -h, --help
          Print help (see a summary with '-h')
```

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/certificate/upload.rs
+++ b/crates/core/tedge/src/cli/certificate/upload.rs
@@ -44,6 +44,21 @@ impl Command for UploadCertCmd {
 
 impl UploadCertCmd {
     fn upload_certificate(&self) -> Result<(), CertError> {
+        // Prompt for password if not already set
+        let username = if self.username.is_empty() {
+            print!("Enter username: ");
+            std::io::stdout().flush()?;
+            let mut input = String::new();
+            std::io::stdin()
+                .read_line(&mut input)
+                .expect("Invalid username");
+            input
+                .trim_end_matches(|c| c == '\n' || c == '\r')
+                .to_string()
+        } else {
+            self.username.clone()
+        };
+
         // Read the password from /dev/tty
         // Unless a password is provided using the `C8YPASS` env var.
         let password = match std::env::var("C8YPASS") {
@@ -83,16 +98,17 @@ impl UploadCertCmd {
         let tenant_id = get_tenant_id_blocking(
             &client,
             build_get_tenant_id_url(&self.host.to_string())?,
-            &self.username,
+            &username,
             &password,
         )?;
-        self.post_certificate(&client, &tenant_id, &password)
+        self.post_certificate(&client, &tenant_id, &username, &password)
     }
 
     fn post_certificate(
         &self,
         client: &reqwest::blocking::Client,
         tenant_id: &str,
+        username: &str,
         password: &str,
     ) -> Result<(), CertError> {
         let post_url = build_upload_certificate_url(&self.host.to_string(), tenant_id)?;
@@ -107,7 +123,7 @@ impl UploadCertCmd {
         let res = client
             .post(post_url)
             .json(&post_body)
-            .basic_auth(&self.username, Some(password))
+            .basic_auth(username, Some(password))
             .send()
             .map_err(get_webpki_error_from_reqwest)?;
 
--- a/crates/core/tedge/src/cli/certificate/upload.rs
+++ b/crates/core/tedge/src/cli/certificate/upload.rs
@@ -30,6 +30,7 @@ pub struct UploadCertCmd {
     pub path: Utf8PathBuf,
     pub host: HostPort<HTTPS_PORT>,
     pub username: String,
+    pub password: String,
 }
 
 impl Command for UploadCertCmd {
@@ -44,7 +45,11 @@ impl Command for UploadCertCmd {
 
 impl UploadCertCmd {
     fn upload_certificate(&self) -> Result<(), CertError> {
-        // Prompt for password if not already set
+        if std::env::var("C8YPASS").is_ok() {
+            eprintln!("WARN: Detected use of a deprecated env variable, C8YPASS. Please use C8Y_PASSWORD instead\n");
+        }
+
+        // Prompt if not already set
         let username = if self.username.is_empty() {
             print!("Enter username: ");
             std::io::stdout().flush()?;
@@ -60,10 +65,10 @@ impl UploadCertCmd {
         };
 
         // Read the password from /dev/tty
-        // Unless a password is provided using the `C8YPASS` env var.
-        let password = match std::env::var("C8YPASS") {
-            Ok(password) => password,
-            Err(_) => rpassword::read_password_from_tty(Some("Enter password: "))?,
+        let password = if self.password.is_empty() {
+            rpassword::read_password_from_tty(Some("Enter password: "))?
+        } else {
+            self.password.clone()
         };
 
         let config = TEdgeConfig::try_new(TEdgeConfigLocation::default())?;
--- a/crates/core/tedge/Cargo.toml
+++ b/crates/core/tedge/Cargo.toml
@@ -22,6 +22,7 @@ certificate = { workspace = true }
 clap = { workspace = true, features = [
     "cargo",
     "derive",
+    "env",
     "string",
     "unstable-styles",
 ] }
--- a/crates/core/tedge/src/cli/certificate/cli.rs
+++ b/crates/core/tedge/src/cli/certificate/cli.rs
@@ -127,8 +127,9 @@ pub enum UploadCertCli {
     /// The command will upload root certificate to Cumulocity.
     C8y {
         #[clap(long = "user")]
+        #[arg(env = "C8Y_USER", default_value = "")]
         /// Provided username should be a Cumulocity user with tenant management permissions.
-        /// The password is requested on /dev/tty, unless the $C8YPASS env var is set to the user password.
+        /// The username is requested on stdin unless provided by the --user flag, or the C8Y_USER env var.
         username: String,
     },
 }
--- a/crates/core/tedge/src/cli/certificate/cli.rs
+++ b/crates/core/tedge/src/cli/certificate/cli.rs
@@ -98,11 +98,12 @@ impl BuildCommand for TEdgeCertCli {
 
             TEdgeCertCli::Upload(cmd) => {
                 let cmd = match cmd {
-                    UploadCertCli::C8y { username } => UploadCertCmd {
+                    UploadCertCli::C8y { username, password } => UploadCertCmd {
                         device_id: config.device.id.try_read(&config)?.clone(),
                         path: config.device.cert_path.clone(),
                         host: config.c8y.http.or_err()?.to_owned(),
                         username,
+                        password,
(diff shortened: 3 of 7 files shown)


didier-wenzek:
I have no strong objection.

However, the following will be confusing, as the wrong password will be used.

```
# In the context of  c8y-go-cli session
export C8Y_USER="bar"
export C8Y_PASSWORD="bar"

# But trying to use a different user
tedge cert upload c8y --user "foo"
```

If the user name is provided both on the command line and as an environment variable,
then the user must be prompt for the password.

reubenmiller:
> I have no strong objection.
> 
> However, the following will be confusing, as the wrong password will be used.
> 
> ```
> # In the context of  c8y-go-cli session
> export C8Y_USER="bar"
> export C8Y_PASSWORD="bar"
> 
> # But trying to use a different user
> tedge cert upload c8y --user "foo"
> ```
> 
> If the user name is provided both on the command line and as an environment variable, then the user must be prompt for the password.

Yeah good point...I'll see if I can come up with a consistent way to handle things (given the user is currently set transparently by clap from either the arg or the `C8Y_USER` env variable.

Though generally if someone is already using a go-c8y-cli session, then want to use it, otherwise they can clear the env variables.

albinsuresh:
> If the user name is provided both on the command line and as an environment variable, then the user must be prompt for the password.

In that case, isn't it better to use the value provided in the `--user` argument, as we can establish a clear order of precedence here? And it helps users to easily override the env value as well.

albinsuresh (commented):
LGTM. Can be merged once we agree on the solution for conflicting usernames provided both on env variable and the `--user` option.

One follow-up improvement would be to replicate the same in `tedge cert create` command as well for the `--device-id` option.

reubenmiller:
> > If the user name is provided both on the command line and as an environment variable, then the user must be prompt for the password.
> 
> 
> 
> In that case, isn't it better to use the value provided in the `--user` argument (as is the current behaviour), as we can establish a clear order of precedence here? And it helps users to easily override the env value as well.
> 
> 

What you describe already works, but didier was pointing out that if the user provides a user via a flag the command will still try to read the password from the given environment variable (even though there could be a mismatch between the username and password pair)

reubenmiller:
> I have no strong objection.
> 
> However, the following will be confusing, as the wrong password will be used.
> 
> ```
> # In the context of  c8y-go-cli session
> export C8Y_USER="bar"
> export C8Y_PASSWORD="bar"
> 
> # But trying to use a different user
> tedge cert upload c8y --user "foo"
> ```
> 
> If the user name is provided both on the command line and as an environment variable, then the user must be prompt for the password.

@didier-wenzek I'm struggling to find a nice rule which is consistent and low maintenance.

What about an alternative to comparing if the `C8Y_PASSWORD` and the `--user` argument is provided, that the user could force the prompting of the password

```sh
export C8Y_USER=foo
export C8Y_PASSWORD=bar

# Option 1: Option to force prompting
tedge cert upload c8y --user other --prompt

# Option 2: Option to ignore env variables (if the user is having problems controlling
tedge cert upload c8y --user other --ignore-env

# Option 3: Enable, where the password will be 
tedge cert upload c8y --user other --password ""
```

Or a new strategy would be to use the given username/password (automatically inherited from the env or flags), and prompt the user for the password if upload receives a 401 response, and then try again.

didier-wenzek:
> @didier-wenzek I'm struggling to find a nice rule which is consistent and low maintenance.
> 
> What about an alternative to comparing if the `C8Y_PASSWORD` and the `--user` argument is provided, that the user could force the prompting of the password
> 
> ```shell
> export C8Y_USER=foo
> export C8Y_PASSWORD=bar
> 
> # Option 1: Option to force prompting
> tedge cert upload c8y --user other --prompt
> 
> # Option 2: Option to ignore env variables (if the user is having problems controlling
> tedge cert upload c8y --user other --ignore-env
> 
> # Option 3: Enable, where the password will be 
> tedge cert upload c8y --user other --password ""
> ```
> 
> Or a new strategy would be to use the given username/password (automatically inherited from the env or flags), and prompt the user for the password if upload receives a 401 response, and then try again.

I don't think an explicit `--prompt` would help. A user who is unaware of the clash between `--user` and `C8Y_USER` is unlikely to request a `--prompt`.

What about prompting for a password when both `--user` and `C8Y_USER` are provided, unless equal and `C8Y_PASSWORD` set?

reubenmiller:
> What about prompting for a password when both `--user` and `C8Y_USER` are provided, unless equal and `C8Y_PASSWORD` set?

Though I find that behaviour very unexpected and it results in very custom handling of cli arguments, as you can't just let clap handle the setting of the arguments (e.g. flag -> env -> default). Normally for CLI tools you want to argument parsing to be very consistent, and not have to manually check "was it set by the flag, or env, or not at all"...that work is best suited to be done by the cli parser lib (in this case "clap"), as the cli parsing library can use the information to auto generate nicely documented flags which states which env variables are used to control which flags, and the default values etc.

reubenmiller:
> > What about prompting for a password when both `--user` and `C8Y_USER` are provided, unless equal and `C8Y_PASSWORD` set?
> 
> Though I find that behaviour very unexpected and it results in very custom handling of cli arguments, as you can't just let clap handle the setting of the arguments (e.g. flag -> env -> default). Normally for CLI tools you want to argument parsing to be very consistent, and not have to manually check "was it set by the flag, or env, or not at all"...that work is best suited to be done by the cli parser lib (in this case "clap"), as the cli parsing library can use the information to auto generate nicely documented flags which states which env variables are used to control which flags, and the default values etc.

The PR has been updated to introduce the `--password` flag to add a consistent handling of env variables, and flags (and to simplify the code).

To force the prompting of the password when the `C8Y_PASSWORD` env variable is set, you can do the following:

```sh
export C8Y_USER=foo
export C8Y_PASSWORD=bar

tedge cert upload c8y --user peter --password ""
```

Whilst, a little unconventional, I don't think this use-case will occur too often to matter (and the code is much cleaner as env variable is handled by clap directly).

crates/core/tedge/src/cli/certificate/upload.rs
```diff
@@ -44,11 +45,30 @@ impl Command for UploadCertCmd {
 
 impl UploadCertCmd {
     fn upload_certificate(&self) -> Result<(), CertError> {
+        if std::env::var("C8YPASS").is_ok() {
+            eprintln!("WARN: Detected use of a deprecated env variable, C8YPASS. Please use C8Y_PASSWORD instead\n");
+        }
+
+        // Prompt for password if not already set
```
didier-wenzek, line 52:
```suggestion
        // Prompt for user if not already set
```

didier-wenzek (approved):
Approved

crates/core/tedge/src/cli/certificate/upload.rs
reubenmiller: Resolved by 9355e9a59e6de2054323c4dfa82d307032fa5b46

gligorisaev:
I have reviewed the test case and it is covering the described feature, only part missing is the check for the warning message. 

> Password can also be set via the more common env variable, C8Y_PASSWORD, though C8YPASS is still supported, but it has been deprecated and a warning is printed to the console => Reason: other Cumulocity tooling (like go-c8y-cli) uses C8Y_PASSWORD which means the tools work well with each other

I am not sure if there is a need to have it in test automation because the warning messages usually are removed after some time, so it would mean maintaining of the test case, but if needed it can look like:
```
Deprecated C8YPASS env variable prints warning
    [Setup]    Setup With Self-Signed Certificate
    Execute Command    sudo tedge disconnect c8y
    ${output}=    Execute Command    sudo env C8YPASS='${C8Y_CONFIG.password}' tedge cert upload c8y --user ${C8Y_CONFIG.username} stderr=True stdout=False ignore_exit_code=True
    Should Contain    ${output}    WARN: Detected use of a deprecated env variable, C8YPASS. Please use C8Y_PASSWORD instead
    ${output}=    Execute Command    sudo tedge connect c8y
    Should Contain    ${output}    Connection check is successful.

reubenmiller:
> I have reviewed the test case and it is covering the described feature, only part missing is the check for the warning message.
> 
> > Password can also be set via the more common env variable, C8Y_PASSWORD, though C8YPASS is still supported, but it has been deprecated and a warning is printed to the console => Reason: other Cumulocity tooling (like go-c8y-cli) uses C8Y_PASSWORD which means the tools work well with each other
> 
> I am not sure if there is a need to have it in test automation because the warning messages usually are removed after some time, so it would mean maintaining of the test case, but if needed it can look like:
> 
> ```
> Deprecated C8YPASS env variable prints warning
>     [Setup]    Setup With Self-Signed Certificate
>     Execute Command    sudo tedge disconnect c8y
>     ${output}=    Execute Command    sudo env C8YPASS='${C8Y_CONFIG.password}' tedge cert upload c8y --user ${C8Y_CONFIG.username} stderr=True stdout=False ignore_exit_code=True
>     Should Contain    ${output}    WARN: Detected use of a deprecated env variable, C8YPASS. Please use C8Y_PASSWORD instead
>     ${output}=    Execute Command    sudo tedge connect c8y
>     Should Contain    ${output}    Connection check is successful.
> ```

Yeah I don't think it is worth explicitly checking the warning message as it is not part of a formal contract.

Issue #2954, opened 2024-06-19
too many software_list commands are created on tedge-mapper c8y startup when using nanomq


reubenmiller:
**Describe the bug**

**Update** this behaviour is only observed when using the nanomq MQTT broker. The behaviour is not observed when using mosquitto.

In a highly customized setup (details of the setup are shown later), Starting the `tedge-mapper c8y` (for the first time) triggers multiple `te/device/main///cmd/software_list` commands on startup, 29 **software_list** commands were created within 1 second.

For example, the following topics were created:

```sh
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.893595279Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.894731237Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.895986612Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.897403279Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.898662695Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.899841445Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.900913403Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.901975486Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.903152694Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.904377861Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.905543861Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.906903694Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.90813136Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.909227068Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.910274943Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.911272818Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.912283985Z
te/device/main///cmd/software_list/c8y-mapper-2024-06-19T18:06:24.913332985Z
[...]

PR #2964, opened 2024-06-26
fix: too many software list requests

didier-wenzek:
## Proposed changes

One of the cause of [too many software list requests](https://github.com/thin-edge/thin-edge.io/issues/2954), is that NanoMQ has not the same behavior as Mosquitto when a connection has overlapping subscriptions. See https://github.com/thin-edge/thin-edge.io/issues/2954#issuecomment-2188244048

- [x] Improve how MQTT messages are logged
- [x] Remove overlapping subscriptions on connect
- [ ] Is there other differences in how NanoMQ behaves compared to Mosquitto?
 

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2954

## Checklist

### DIFF
--- a/crates/extensions/tedge_mqtt_ext/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_ext/src/lib.rs
@@ -11,9 +11,8 @@ use tedge_actors::futures::channel::mpsc;
 use tedge_actors::Actor;
 use tedge_actors::Builder;
 use tedge_actors::ChannelError;
+use tedge_actors::CombinedReceiver;
 use tedge_actors::DynSender;
-use tedge_actors::LoggingReceiver;
-use tedge_actors::LoggingSender;
 use tedge_actors::MessageReceiver;
 use tedge_actors::MessageSink;
 use tedge_actors::MessageSource;
@@ -32,9 +31,9 @@ pub use mqtt_channel::TopicFilter;
 
 pub struct MqttActorBuilder {
     mqtt_config: mqtt_channel::Config,
-    input_receiver: LoggingReceiver<MqttMessage>,
+    input_receiver: CombinedReceiver<MqttMessage>,
     publish_sender: mpsc::Sender<MqttMessage>,
-    pub subscriber_addresses: Vec<(TopicFilter, LoggingSender<MqttMessage>)>,
+    pub subscriber_addresses: Vec<(TopicFilter, DynSender<MqttMessage>)>,
     signal_sender: mpsc::Sender<RuntimeRequest>,
 }
 
@@ -42,7 +41,7 @@ impl MqttActorBuilder {
     pub fn new(config: mqtt_channel::Config) -> Self {
         let (publish_sender, publish_receiver) = mpsc::channel(10);
         let (signal_sender, signal_receiver) = mpsc::channel(10);
-        let input_receiver = LoggingReceiver::new("MQTT".into(), publish_receiver, signal_receiver);
+        let input_receiver = CombinedReceiver::new(publish_receiver, signal_receiver);
 
         MqttActorBuilder {
             mqtt_config: config,
@@ -56,6 +55,9 @@ impl MqttActorBuilder {
     pub(crate) fn build_actor(self) -> MqttActor {
         let mut combined_topic_filter = TopicFilter::empty();
         for (topic_filter, _) in self.subscriber_addresses.iter() {
+            for pattern in topic_filter.patterns() {
+                tracing::info!(target: "MQTT sub", "{pattern}");
+            }
             combined_topic_filter.add_all(topic_filter.to_owned());
         }
         let mqtt_config = self.mqtt_config.with_subscriptions(combined_topic_filter);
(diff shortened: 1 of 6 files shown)


crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -58,8 +57,16 @@ impl MqttActorBuilder {
         for (topic_filter, _) in self.subscriber_addresses.iter() {
             combined_topic_filter.add_all(topic_filter.to_owned());
         }
-        let mqtt_config = self.mqtt_config.with_subscriptions(combined_topic_filter);
 
+        let removed = combined_topic_filter.remove_overlapping_patterns();
+        for pattern in combined_topic_filter.patterns() {
+            tracing::info!(target: "MQTT sub", "{pattern}");
```
didier-wenzek, line 63: Now all the subscriptions are logged when the agent or the mapper starts.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -115,6 +122,7 @@ impl FromPeers {
         outgoing_mqtt: &mut mpsc::UnboundedSender<MqttMessage>,
     ) -> Result<(), RuntimeError> {
         while let Ok(Some(message)) = self.try_recv().await {
+            tracing::info!(target: "MQTT pub", "[{}] {}", &message.topic.name, message.payload_str().unwrap_or_default());
```
didier-wenzek, line 125: Now a message that is published is logged as:

```
2024-06-26T15:35:33.01872567Z  INFO MQTT pub: [te/device/main///cmd/software_list/c8y-mapper-2024-06-26T15:35:33.010468866Z] {"status":"init"}
```

This has to be compared to the current log entries for the same:

```
2024-06-20T17:50:27.811717059Z DEBUG MQTT: recv Ok(Some(MqttMessage { topic: Topic { name: "te/device/main///cmd/software_list/c8y-mapper-2024-06-20T17:50:27.811587059Z" }, payload: "{"status":"init"}", qos: AtLeastOnce, retain: true })) 
```

Which is confusing. This has to be read as "the MQTT actor receive a request to publish a message".
With the new proposal, it's clear: the message is published over MQTT.

That said, the following has to be discussed:
- Is INFO the right level?  This is really helpful for commands, but might be overwhelming with telemetry data.
- One can configure the level depending on the topic. Is it worth it?
- Do we have to log QoS and Retain flags?

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -139,6 +147,7 @@ impl ToPeers {
         incoming_mqtt: &mut mpsc::UnboundedReceiver<MqttMessage>,
     ) -> Result<(), RuntimeError> {
         while let Some(message) = incoming_mqtt.next().await {
+            tracing::info!(target: "MQTT recv", "[{}] {}", &message.topic.name, message.payload_str().unwrap_or_default());
```
didier-wenzek, line 150: This gives :

```
2024-06-26T15:35:33.230788346Z  INFO MQTT recv: [te/device/main/service/tedge-mapper-bridge-c8y/status/health] {"status":"up"}
```

to be compared with the current log entries:

```
2024-06-20T17:50:28.806953202Z DEBUG MQTT: send MqttMessage { topic: Topic { name: "te/device/main/service/tedge-mapper-bridge-c8y/status/health" }, payload: "{"status":"up"}", qos: AtLeastOnce, retain: true } 
```

Where we have to understand that the MQTT actor is sending to the other actors a message that has been just received from its connection to the broker.

Same questions as for the published messages: what is the appropriate log level?

reubenmiller: I think the log level depends on how often are these messages being sent around...for periodic data like the measurements/events/alarms, `DEBUG` is more appropriate, but maybe for services having the status updates shown on `INFO` level would be more hopeful to understanding the interactions between the services in the logs, though since we don't really want to go down the conditional log level (based on topics), is there a "health"status actor that can write the `INFO` log messages (rather than the mqtt actor)?

crates/extensions/tedge_mqtt_ext/src/lib.rs
reubenmiller: > Is INFO the right level? This is really helpful for commands, but might be overwhelming with telemetry data.
I would default to `DEBUG` level for MQTT message logging. We can have additional command handling logging at `INFO` level as the message is handled in context of the workflows (except for child devices with external agents doing their own custom command processing).

> One can configure the level depending on the topic. Is it worth it?
No I don't think so either, though as long as the first comment above can be addressed.

> Do we have to log QoS and Retain flags?
Yes, QoS and Retain flags would be useful, though if possible I would like to avoid logging of Rust related stuff e.g. `Ok(Some(MqttMessage` seems unnecessary to me.

didier-wenzek: Improved by https://github.com/thin-edge/thin-edge.io/pull/2964/commits/0b497e6d238ae8f1229420c2fe20f09cc42d9a76

reubenmiller:
@didier-wenzek I just re-ran the original setup using the tedge build `1.1.2~144+g0b497e6`, and there are still multiple `software_list` operation being created by the tedge-mapper-c8y (see logs below):

https://github.com/thin-edge/debug-tedge-readonly/tree/main/logs/run3-g0b497e6

reubenmiller (approved):
Approved. I've tested the new application against nanomq 0.21 and it works as expected and the new logging is very helpful

rina23q (approved):
The code looks good

PR #2970, opened 2024-07-01
docs: add instructions for Cumulocity IoT custom domains

reubenmiller:
## Proposed changes

Add instructions for Cumulocity IoT users which are using the [custom domain feature](https://cumulocity.com/docs/enterprise-tenant/customization/#domain-name).

Add both instructions to a dedicated section (which is easier to link to), and add the error message in the "common errors" section.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/docs/src/start/connect-c8y.md
+++ b/docs/src/start/connect-c8y.md
@@ -50,6 +50,25 @@ This will set the root certificate path of the Cumulocity IoT.
 In most of the Linux flavors, the certificate will be present in `/etc/ssl/certs`.
 If not found download it from [here](https://www.identrust.com/dst-root-ca-x3).
 
+### Custom domain
+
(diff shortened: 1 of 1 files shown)


docs/src/start/connect-c8y.md
```diff
@@ -50,6 +50,25 @@ This will set the root certificate path of the Cumulocity IoT.
...
+
+For example, below shows setting the HTTP and MQTT endpoints:
+
+```sh
+tedge config set c8y.http "custom.example.com"
+tedge config set c8y.mqtt "t12345.cumulocity.com:8883"
+```
+
+:::tip
+If you have `curl` and `jq` installed, then you can check if you are using a custom domain by executing the following command. If the response includes a different domain than the url used in the curl command, then you are using a custom domain, and you should use the output to set the `c8y.mqtt` thin-edge.io setting instead.
```
didier-wenzek, line 65: This tip suggests that thin-edge could do the same to infer `c8y.mqtt` from `c8y.http`.

didier-wenzek (approved):
Approved

docs/src/start/connect-c8y.md
reubenmiller: Potentially, though we'd have to do some more testing to make sure it works in all cases (where as a doc statement is a lot more forgiving).

Issue #2972, opened 2024-07-01
Incorrect assertion in tedge-write test "doesnt_change_permissions_if_file_exists"


renaudcalmont:
Hi,
the result seems OK because the expectation is 644 and the root user umask on the testing system happens to be set at 022.
So 644 is the maximum file mode automatically applied but it isn't a proof that the code didn't change it.

https://github.com/thin-edge/thin-edge.io/blob/cf109f329b8513bc56b70e228fcc812a9cf94b05/crates/core/tedge_write/tests/main.rs#L76

Whithout any code change this test will fail immediately on a system where the root user umask is 077.
I strongly suggest to use 666 as the expected value for this test.

Moreover, it checks only the file mode but not the owner and group.
They happen to be reset to root:root, in contradiction with the code at https://github.com/thin-edge/thin-edge.io/blob/cf109f329b8513bc56b70e228fcc812a9cf94b05/crates/core/tedge_write/src/bin.rs#L66 which is supposed to prevent both mode and owner/group changes.

Bravo555:
Hi, so in total following changes were made:

1. Most important, permissions and owner information were being reset if writing to a file that exists.
    This was fixed: `tedge-write` should now not overwrite original file owner/permissions.
    Also `tedge-write --help` was edited to clarify what should happen when file exists or doesn't exist.
2. Test were failing due to assumptions about default umask.
    We now set an umask as part of the test, so a case where `tedge-write` changes permissions to defaults should no longer be missed.
3. `tedge-write` can now take user/group id instead of only a name.

Some `tedge-write` tests were moved to Robot Framework, under `tests/RobotFramework/tests/tedge_write.robot`, you can run them with `just integration-test -s tedge_write`. Once it's merged you'll be able to run the test suite and verify if your issue is solved.

gligorisaev:
QA can verify that this bug in test is fixed

PR #2974, opened 2024-07-03
fix: make tedge-write not reset owner/group

Bravo555:
## Proposed changes

Fixed two closely related problems in tedge-write:

- When destination file existed, its owner/group was reset because when creating a temporary file for atomic write we forgot to set the same permissions as the destination file. Now we copy destination file permissions to the temporary, ensuring that after the rename permissions stay the same.
- Changed some permissions used in the tests from 644 to 666, so the tests don't fail when default umask is different.

Also added new functionality: tedge-write can now use plain uid/gid numbers to change owner instead of needing a name. This was done so we can more easily test file owner changes, but is probably a user improvement in its own right.

## Types of changes

## Paste Link to the issue
- #2972

## Checklist


crates/core/tedge_write/tests/main.rs
```diff
@@ -31,49 +33,59 @@ fn creates_dest_file_if_doesnt_exist() {
 }
 
 #[test]
-fn sets_file_permissions_if_file_doesnt_exist() {
+fn changes_permissions_if_destination_doesnt_exist() {
     // Arrange
     let (temp_dir, source_path) = setup_source_file();
-    fs::set_permissions(&source_path, Permissions::from_mode(0o644)).unwrap();
+    chown(&source_path, Some(2137), Some(2137)).unwrap();
```
didier-wenzek, line 39: Such a test can only be successful if run by a privileged user allowed to `chown` a file.

didier-wenzek (commented):
Adding an option to use plain uid/gid numbers is a good idea. And the fix seems correct. But, you have to come with another way to test it without assuming chown privileges.

crates/core/tedge_write/tests/main.rs
Bravo555: Rust integration tests were a bad environment for such a test, they were moved to RobotFramework in 43f5705.

didier-wenzek (commented):
Indeed replacing system tests by integration tests is more appropriate. The code being correct, I will approve this PR once rebased with the system-tests fix.

didier-wenzek (approved):
Approved

Issue #2959, opened 2024-06-22
tedge-agent fails to startup due to SoftwareManagerActor failing due to an existing file


reubenmiller:
**Describe the bug**

**tedge-agent** exits shortly after starting due to the following error (see the Additional context for the full log)

```log
2024-06-22T17:04:13.206357574Z ERROR Runtime: Actor SoftwareManagerActor-11 has finished unsuccessfully: ActorError(FromState(FromIo(Os { code: 17, kind: AlreadyExists, message: "File exists" })))
```

The error message is not very helpful as it does not show which file already "exists", so the user can't fix the issue easily.

**To Reproduce**

It is unclear how to reproduce the error as it was encountered when using a custom config-dir location on a very resource constrained device.

**Expected behavior**

The error message should list the path of the file which already exists, so the user has a chance to potentially "fix" the current situation.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`unknown`|
|Hardware [incl. revision]|`unknown`|
|System-Architecture|`Linux mdm9607 3.18.48 #1 PREEMPT Mon May 27 09:38:30 CEST 2024 armv7l GNU/Linux`|
|thin-edge.io version|`tedge 1.1.2~118+g476783c`|

**Additional context**

The full logs from the tedge-agent are shown below:

```log
2024-06-22T17:04:13.147691793Z  INFO tedge_agent::agent: tedge-agent starting
2024-06-22T17:04:13.148309293Z  INFO sm-agent: tedge_agent::agent: Starting tedge-agent v1.1.2~118+g476783c
2024-06-22T17:04:13.148817366Z  INFO Runtime: Started
2024-06-22T17:04:13.150195595Z  INFO sm-agent: tedge_config_manager::config: Using the configuration from /data/thin-edge.io/etc/tedge/plugins/tedge-configuration-plugin.toml
2024-06-22T17:04:13.151657991Z  INFO sm-agent: log_manager::config: Using the configuration from /data/thin-edge.io/etc/tedge/plugins/tedge-log-plugin.toml
2024-06-22T17:04:13.152286689Z  INFO sm-agent: tedge_agent::agent: Running as a main device, starting tedge_to_te_converter and File Transfer Service
2024-06-22T17:04:13.152505282Z  INFO Runtime: Running TedgetoTeConverter-0
[...]

PR #2975, opened 2024-07-04
feat(fs): add extra context to errors when atomically writing to file

Ruadhri17:
## Proposed changes
This PR adds extra info to error messages when atomically writing to file.

## Types of changes

## Paste Link to the issue

#2959

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/fs.rs
+++ b/crates/common/tedge_utils/src/fs.rs
@@ -8,41 +8,111 @@ use std::path::PathBuf;
 use tokio::fs as tokio_fs;
 use tokio::io::AsyncWriteExt;
 
+#[derive(Debug, thiserror::Error)]
+pub enum AtomFileError {
+    #[error("Writing the content to the file {file:?} failed: {context:?}. source={source:?}")]
+    WriteError {
+        file: Box<Path>,
+        context: String,
+        source: std::io::Error,
+    },
+}
+
+pub trait ErrContext<T> {
+    fn with_context(
+        self,
+        context: impl Fn() -> String,
+        file: impl AsRef<Path>,
+    ) -> Result<T, AtomFileError>;
+}
+
+impl<T, E: Into<std::io::Error>> ErrContext<T> for Result<T, E> {
+    fn with_context(
+        self,
+        context: impl Fn() -> String,
+        file: impl AsRef<Path>,
+    ) -> Result<T, AtomFileError> {
+        self.map_err(|err| AtomFileError::WriteError {
+            file: Box::from(file.as_ref()),
+            context: context(),
+            source: err.into(),
+        })
+    }
+}
+
+pub struct TempFile(PathBuf);
+
+impl Drop for TempFile {
+    fn drop(&mut self) {
+        let _ = std_fs::remove_file(&self.0);
+    }
+}
+
 /// Write file to filesystem atomically using std::fs synchronously.
 pub fn atomically_write_file_sync(
     dest: impl AsRef<Path>,
     mut reader: impl Read,
-) -> std::io::Result<()> {
+) -> Result<(), AtomFileError> {
     let dest_dir = parent_dir(dest.as_ref());
     // FIXME: `.with_extension` replaces file extension, so if we used this
     // function to write files `file.txt`, `file.bin`, `file.jpg`, etc.
     // concurrently, then this will result in an error
-    let tempfile = PathBuf::from(dest.as_ref()).with_extension("tmp");
+    let tempfile = TempFile(PathBuf::from(dest.as_ref()).with_extension("tmp"));
 
     // Write the content on a temp file
     let mut file = std_fs::OpenOptions::new()
         .write(true)
         .create_new(true)
-        .open(&tempfile)?;
+        .open(&tempfile.0)
+        .with_context(
+            || format!("could not create the temporary file {:?}", tempfile.0),
+            &dest,
+        )?;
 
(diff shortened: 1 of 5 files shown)


crates/common/tedge_utils/src/fs.rs
```diff
@@ -8,11 +9,47 @@ use std::path::PathBuf;
...
+
+    #[error("Writing the content to the file {file:?} failed: {context:?}.")]
+    FailedToWrite {
+        file: PathBuf,
+        context: std::io::Error,
+    },
+
+    #[error(transparent)]
+    FromIoError(#[from] std::io::Error),
+}
```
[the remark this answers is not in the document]
didier-wenzek: I'm aligned with @Bravo555 here. I would not that far as using `anyhow::Error`, though. A specific error type with a single error case will be enough to catch the context (source file, io error and possibly the step (create/copy/rename)

crates/common/tedge_utils/src/fs.rs
```diff
@@ -23,26 +47,55 @@ pub fn atomically_write_file_sync(
...
+            &dest,
+        )?;
+
+    if let Err(err) = std::io::copy(&mut reader, &mut file).context(
+        format!("could not copy the content to the temporary file {tempfile:?}",),
+        &dest,
+    ) {
+        let _ = std_fs::remove_file(&tempfile);
         return Err(err);
     }
```
didier-wenzek, line 62: As this is done several time, I would try to wrap `tempfile` with a struct that remove the file on drop so we can write:

```suggestion
    std::io::copy(&mut reader, &mut file).context(
        format!("could not copy the content to the temporary file {tempfile:?}",),
        &dest,
    ) ?;
```

That said this would not work in the `async` case.

crates/common/tedge_utils/src/fs.rs
```diff
@@ -23,26 +47,55 @@ pub fn atomically_write_file_sync(
...
 
     // Ensure the content reach the disk
-    file.flush()?;
-    file.sync_all()?;
+    if let Err(err) = file.flush().context(
+        format!("could not flush the content to the temporary file {tempfile:?}",),
```
didier-wenzek, line 66:
```suggestion
        format!("could not flush the content of the temporary file {tempfile:?}",),
```

crates/common/tedge_utils/src/fs.rs
```diff
@@ -23,26 +47,55 @@ pub fn atomically_write_file_sync(
...
+    ) {
+        let _ = std_fs::remove_file(&tempfile);
         return Err(err);
     }
 
     // Ensure the new name reach the disk
-    let dir = std::fs::File::open(dest_dir)?;
-    dir.sync_all()?;
+    let dir =
+        std::fs::File::open(dest_dir).context("could not open the file".to_string(), &dest)?;
```
didier-wenzek, line 95:
```suggestion
        std::fs::File::open(dest_dir).context("could not open the directory".to_string(), &dest)?;
```

didier-wenzek (approved):
Approved. With some suggestions.

crates/common/tedge_utils/src/fs.rs
```diff
@@ -8,41 +8,97 @@ use std::path::PathBuf;
...
+    )?;
 
     // Ensure the new name reach the disk
-    let dir = std::fs::File::open(dest_dir)?;
-    dir.sync_all()?;
+    let dir =
+        std::fs::File::open(dest_dir).context("could not open the directory".to_string(), &dest)?;
+
+    dir.sync_all()
+        .context("could not save the file to disk".to_string(), &dest)?;
```
Bravo555, line 101: issue: because argument to `.context()` is a `String`, and not `impl Fn() -> String`, we're always allocating strings for all possible errors, even if they don't happen!

We should instead use a closure, so we're only allocating context when it's necessary:

```rust
pub trait ErrContext<T> {
    fn with_context(
        self,
        context: impl Fn() -> String,
        file: impl AsRef<Path>,
    ) -> Result<T, AtomFileError>;
}

impl<T, E: Into<std::io::Error>> ErrContext<T> for Result<T, E> {
    fn with_context(
        self,
        context: impl Fn() -> String,
        file: impl AsRef<Path>,
    ) -> Result<T, AtomFileError> {
        self.map_err(|err| AtomFileError::WriteError {
            file: Box::from(file.as_ref()),
            context: context(),
            source: err.into(),
        })
    }
}
```

In anyhow, `.context()` can take in a `&str`, which is statically allocated, but for `String` you should use `.with_context()`, which similarly takes in a closure. If we give up formatting the context string, and only use a string slice, we could use `&'static str` instead:

```rust
pub enum AtomFileError {
    #[error("Writing the content to the file {file:?} failed: {context:?}. source={source:?}")]
    WriteError {
        file: Box<Path>,
        context: &'static str,
        source: std::io::Error,
    },
}

pub trait ErrContext<T> {
    fn context(
        self,
        context: &'static str,
        file: impl AsRef<Path>,
    ) -> Result<T, AtomFileError>;
}

impl<T, E: Into<std::io::Error>> ErrContext<T> for Result<T, E> {
    fn context(
        self,
        context: &'static str,
        file: impl AsRef<Path>,
    ) -> Result<T, AtomFileError> {
        self.map_err(|err| AtomFileError::WriteError {
            file: Box::from(file.as_ref()),
            context,
            source: err.into(),
        })
    }
}
```

Bravo555 (changes requested):

Bravo555 (approved):

Issue #2969, opened 2024-06-27
Cumulocity IoT urls fail to download when using custom domain name


reubenmiller:
**Describe the bug**

When using a custom Cumulocity IoT domain (a enterprise tenant feature), installing software packages with binaries uploaded to Cumulocity IoT via the UI fail with a `401` error as the tedge-agent only checks the `c8y.http` value to see if the local c8y proxy service should be used or not.

```sh
Jun 27 15:22:26 b9cdad7c25fe tedge-agent[1017]: 2024-06-27T15:22:26.603900183Z  INFO tedge_agent::tedge_operation_converter::actor: Processing software_update operation executing step
Jun 27 15:22:27 b9cdad7c25fe tedge-agent[1017]: 2024-06-27T15:22:27.568984292Z ERROR plugin_sm::plugin: Download error: DownloadError {
Jun 27 15:22:27 b9cdad7c25fe tedge-agent[1017]:     reason: "HTTP status client error (401 Unauthorized) for url (https://t123456.cumulocity.com/inventory/binaries/12520)",
Jun 27 15:22:27 b9cdad7c25fe tedge-agent[1017]:     url: "https://t123456.cumulocity.com/inventory/binaries/12520",
Jun 27 15:22:27 b9cdad7c25fe tedge-agent[1017]:     source_err: "",
Jun 27 15:22:27 b9cdad7c25fe tedge-agent[1017]: }
Jun 27 15:22:27 b9cdad7c25fe sudo[2397]:    tedge : PWD=/tmp ; USER=root ; COMMAND=/etc/tedge/sm-plugins/apt finalize
```

This is caused because when binaries uploaded to Cumulocity IoT use the underlying tenant url, and not the custom domain to store a reference to itself, and this URL is sent to thin-edge.io via the Software Update operation.

**To Reproduce**

This requires an enterprise tenant and a custom domain.

1. Configure thin-edge.io with different `c8y.http` and `c8y.mqtt` urls

    ```sh
    tedge config set c8y.http dashboard.customer.com:443
    tedge config set c8y.mqtt t123456.cumulocity.com:8883
    ```

2. Upload a binary to the Software repository (any binary will do)

3. Install the software via the Cumulocity IoT Software tab

Inspect the `tedge-agent` logs to see if the file was successfully downloaded or not.

**Expected behavior**

[...]

Issue #2804, opened 2024-03-28
C8y tenant validation broken while using custom domains


albinsuresh:
**Describe the bug**

When the c8y-mapper receives a configuration update request with the URL of the artefact to be downloaded, it checks if the incoming URL belongs to the same tenant domain that the mapper is configured with, using the `http_proxy::C8yEndPoint::maybe_tenant_url` function. But, this function relies on common domain suffixes for this check so that aliases like `xyz.c8y.io` and `t123.c8y.io` pass the check as they both share the common `c8y.io` domain suffix.

But, while using custom domains feature of C8y, the custom domain URL and the URL received in the incoming request may not have a common suffix. For example, a customer might be using a custom domain `dashboard.hennypenny.com` to connect to C8y, but the URLs in incoming requests would contain the real tenant URL behind the custom domain, in the form `t517788845.us.cumulocity.com`. Since they don't share a common suffix, the mapper will wrongly flag that URL as an external URL and try to download it without any authentication, resulting in a failure.

The current check is broken even in cases where `abc.com` and `xyz.com` will be qualified as part of the same domain, just because they have the matching `.com` suffix, which is clearly wrong.

**To Reproduce**

To reproduce the first case, you need to have a C8y instance with a custom domain name that does not have a common suffix with the real tenant URL (If using the custom domain feature is not feasible, you can also simulate this behaviour by creating a local DNS entry on the tedge device with the custom domain URL pointing to the real tenant URL). With such a setup, configuration updates won't work as the mapper will fail to download the updated config artefacts.

For the second case, the following unit test would suffice, which fails with the existing code:

```rust
    #[test]
    fn url_is_my_tenant_not_too_broad() {
        let c8y = C8yEndPoint::new("abc.com", "test_device");
        assert!(c8y.maybe_tenant_url("http://xyz.com").is_none());
    }
```

**Expected behavior**

Instead of doing a simple suffix match of the URLs, thin-edge should retrieve the real tenant URL behind the provided custom domain URL and use that for any further matching. The real URL can be retrieved by querying the `loginOptions` endpoint (e.g: `https://your.custom.domain/tenant/loginOptions`) and extracting the value of the `self` key from the JSON response. A sample response would be as follows:

```json
{
    "self": "https://t517788845.us.cumulocity.com/tenant/loginOptions/",
    "loginOptions": []
}
```

[...]

PR #2977, opened 2024-07-05
fix: use MQTT as well as HTTP host to determine possible tenant url

jarhodes314:
## Proposed changes
Determine whether URLs are Cumulocity URLs based on both the HTTP and MQTT host, not just the HTTP host.

## Types of changes

## Paste Link to the issue
- #2969 

## Checklist


albinsuresh (commented):
The logic is pretty straightforward for those having separate `c8y.http` and `c8y.mqtt` settings, and implemented as described in the ticket.

But, my understanding was that the end user sometimes may not even be aware of the internal URLs and they are only exposed to the external custom domains. So, they probably wouldn't even know that they need to configure both settings separately. This is the reason why querying the `https://your.custom.domain/tenant/loginOptions` endpoint was proposed as a better fix approach in #2804.

But, if it is guaranteed that custom domain users would configure their `tedge` like this with both the settings, then I'm happy to approve this PR.

reubenmiller:
> The logic is pretty straightforward for those having separate `c8y.http` and `c8y.mqtt` settings, and implemented as described in the ticket.
> 
> But, my understanding was that the end user sometimes may not even be aware of the internal URLs and they are only exposed to the external custom domains. So, they probably wouldn't even know that they need to configure both settings separately. This is the reason why querying the `https://your.custom.domain/tenant/loginOptions` endpoint was proposed as a better fix approach in #2804.
> 
> But, if it is guaranteed that custom domain users would configure their `tedge` like this with both the settings, then I'm happy to approve this PR once the tests pass (a rebase required).

We can address the "easy of configuration" aspect to a separate PR. This PR is purely focusing on the case where the user knows that they have to use two separate URLs for the MQTT and HTTP traffic, however thin-edge.io may receive binaries either using the internal URL OR the custom domain name (as the user can manually use the custom domain when uploading binaries to their tenant).

didier-wenzek (approved):
Approved

didier-wenzek:
There is a flaky test (unrelated to this PR) failing the merge queue :

```
test tests::test_cli_pub_basic::none_expects ... FAILED

failures:

---- tests::test_cli_pub_basic::none_expects stdout ----
thread 'tests::test_cli_pub_basic::none_expects' panicked at /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/assert_cmd-2.0.13/src/output.rs:65:25:
command="/home/runner/work/thin-edge.io/thin-edge.io/target/llvm-cov-target/debug/tedge" "--config-dir" "/tmp/.tmp5EPpVm" "mqtt" "pub" "topic" "message"
code=1
stdout=""
stderr=
ERROR: the message has not been published
Error: failed to publish the message \"message\" on the topic \"topic\" with QoS \"AtMostOnce\".

Caused by:
    0: I/O: connection closed by peer
    1: connection closed by peer
```

PR #2979, opened 2024-07-07
tests: don't remove systemd-tmpfiles-setup service which is responsible for creating tmp folders

reubenmiller:
## Proposed changes

Fix an issue where system tests were failing due to missing the `/run/lock` folder. The change no longer deletes the `systemd-tmpfiles-setup.service` which is the mechanism systemd uses to create temporary files such as `/run/lock`. Using `systemd-tmpfiles-setup.service` is preferable over using the docker tmpfs mechansim as the former behaves more true to a device installation (rather than a docker container).

In addition, mosquitto is now installed from the debian "bookworm-backports" rather than "sid", as it seems the "sid" channel is installing a much newer version of systemd (v256 vs v253) and there might be some unexpected changes in the new version (which are still to be investigated in a follow up PR).

Note: It is still unclear when the exact change was introduced, possibly due to the change in the `debian:12-slim` image recently.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


gligorisaev (approved):

didier-wenzek (approved):
Approved

PR #2981, opened 2024-07-08
tests: improve debugging of bootstrap script and include systemd/cgroup compat check

reubenmiller:
## Proposed changes

The root cause to the problem that prompted the change in https://github.com/thin-edge/thin-edge.io/pull/2979 has been identified. Previously the debian `sid` repository was being used to install mosquitto 2.0.18, then around 2024-07-05, the usage of the `sid` repository also resulted in the systemd version being upgraded to 256 where support for cgroup v1 has been deprecated (see the [systemd news](https://github.com/systemd/systemd/blob/main/NEWS)). The Github runner ubuntu-20.04 was being used to run the system tests which uses cgroup v1 resulting in some system test failures due to some systemctl commands not functioning as expected. Using the Github runner `ubuntu-22.04` fixes the issue as the OS uses cgroup v2 by default.

In addition to using switching to the Github runner ubuntu-22.04, the following changes were made:
* Add an explicit systemd/cgroup check in the bootstrap.sh script to fail loudly if the incompatibility is detected
* Add additional logging to indicate which sections completed successfully and if key parts failed (e.g. the restarting of the mosquitto service)

To verify if the ubuntu-22.04 runner would make the tests pass, the build-workflow was run on a fork directly from the PR branch (as changing the runner OS does not activate for the initial PR workflow run, only after the PR has been approved). The following test runs shows that the system tests are passing with the systemd 256 (installed indirectly via the debian sid repository), but the PR will not include the debian sid repo as the bookworm-backports is the more desired approach as it is deemed as "stable" incomparison with the "sid" repo.
* https://github.com/reubenmiller/thin-edge.io/actions/runs/9870745687

## Types of changes

## Paste Link to the issue

The PR which prompted this investigation and changes:
* https://github.com/thin-edge/thin-edge.io/pull/2979

## Checklist

### DIFF
--- a/.github/workflows/build-workflow.yml
+++ b/.github/workflows/build-workflow.yml
@@ -184,7 +184,7 @@ jobs:
     needs: [approve, check-build]
     environment:
       name: Test Auto
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         job:
--- a/tests/images/debian-systemd/files/bootstrap.sh
+++ b/tests/images/debian-systemd/files/bootstrap.sh
@@ -735,7 +735,37 @@ EOF
     tedge config set mqtt.client.auth.cert_file /setup/client.crt
     tedge config set mqtt.client.auth.key_file /setup/client.key
 
-    systemctl restart mosquitto
+    if ! sudo systemctl restart mosquitto.service; then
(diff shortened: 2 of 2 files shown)


didier-wenzek (approved):
Approved. Thank you for taking the time to precisely identify the root cause.

PR #2989, opened 2024-07-11
test: remove OpenSSL Error check because of flakiness

gligorisaev:
The step:
`Should Not Contain    ${MQTT_PUB}    Error` 
from the test:
`tests/RobotFramework/tests/mqtt/disconnect_from_broker.robot`
has to be commented out because this validation is causing unstable test runs, if needed can be included again when (and if) we find the reason for it

## Paste Link to the issue
<br/>

## Checklist


didier-wenzek (commented):
The PR has to be improved. For instance:

SSL errors are observed in mosquitto log, but these errors are related to connections while the point of this test was to check error on disconnect. As this introduces flakyness, the simpler is to remove this check.

We can see in a second step what is  going on.

didier-wenzek (approved):
Approved.

Issue #2966, opened 2024-06-27
log file request does not return fixed log file path if the file's last modified date does not match date filter


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Log file request operations have some unexpected behaviour (from the user's perspective) when using fixed paths.

For instance, when given the following `tedge-log-plugin.toml` configuration:

*file: /etc/tedge/plugins/tedge-log-plugin.toml*

```toml
files = [
    { type = "tedge-bridge-monitor", path = '/var/log/tedge-bridge-monitor/tedge-bridge-monitor.log' },
]
```

When the user requests a log file for the `tedge-bridge-monitor` type, the user is met with the following failureReason in the operation if the file has not been modified within the given dateFrom/dateTo range (of the operation request):

```sh
Failed to initiate log file upload: No logs found for log type "tedge-bridge-monitor"
```

The above error message is unexpected from the user's perspective because they have given an explicit path to a log file, and just want to see the contents, however the result is dependent on the last modified timestamp of the file (which the user has no idea what it would be), where the failureReason suggests that the file does not exist.

* The failureReason isn't very helpful to realize that the file did not match because of the dateFrom/dateTo filters and not because it does not exist
* 

I understand that the limitation exists what since the log request handler can't know the contents of the file to apply dateFrom/dateTo range filtering from the contents of the file, however applying the date range filtering on a single file is very unexpected, and is annoying for users, as they then have to change the dateFrom/dateTo range to some very wide range (e.g. 1970 -> now) if they have no idea when the file was last written to.

**Describe the solution you'd like**

For file based log requests:

* don't filter the file for last modified dates when a static file path is used (e.g. when the path does not contain any wildcards like `*`)

**Describe alternatives you've considered**

Alternatively, the error message could be changed to something indicates that the a time range was applied:

```
Failed to initiate log file upload: No logs found for log type "tedge-bridge-monitor" for the given time range
```

**Additional context**

gligorisaev:
The review was done in the PR:
https://github.com/thin-edge/thin-edge.io/pull/2991

PR #2991, opened 2024-07-11
feat(log manager): skip filtering by date when a static path is used

Ruadhri17:
## Proposed changes
Currently the log manager is filtering logs by date no matter whether provided path is static or containing wildcard. With this PR log manager will return log path if it is fixed. Explanation why this feature is needed can be found in #2966.

## Types of changes

## Paste Link to the issue
* #2966 
<br/>

## Checklist

### DIFF
--- a/crates/common/log_manager/src/log_utils.rs
+++ b/crates/common/log_manager/src/log_utils.rs
@@ -2,6 +2,8 @@ use super::config::FileEntry;
 use super::error::LogRetrievalError;
 use easy_reader::EasyReader;
 use glob::glob;
+use regex::Regex;
+use std::cmp::Reverse;
 use std::collections::VecDeque;
 use std::fs::File;
 use std::io::Write;
@@ -11,16 +13,15 @@ use time::OffsetDateTime;
 
 /// read any log file coming from `obj.log.log_type`
 pub fn new_read_logs(
-    files: &Vec<FileEntry>,
+    files: &[FileEntry],
     log_type: &str,
     date_from: OffsetDateTime,
     lines: usize,
     search_text: &Option<String>,
     tmp_dir: &Path,
 ) -> Result<PathBuf, LogRetrievalError> {
-    // first filter logs on type
-    let mut logfiles_to_read = filter_logs_on_type(files, log_type)?;
-    logfiles_to_read = filter_logs_path_on_metadata(log_type, date_from, logfiles_to_read)?;
+    //filter logs on type and date
+    let logfiles_to_read = filter_logs(files, log_type, date_from)?;
 
     let temp_path = tmp_dir.join(format!("{log_type}-{}", rand::random::<u128>()));
     let mut temp_file = File::create(&temp_path)?;
@@ -44,7 +45,7 @@ pub fn new_read_logs(
     Ok(temp_path)
 }
 
-pub fn read_log_content(
+fn read_log_content(
     logfile: &Path,
     mut line_counter: usize,
     max_lines: usize,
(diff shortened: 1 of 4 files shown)


crates/common/log_manager/src/log_utils.rs
```diff
@@ -93,68 +93,65 @@ pub fn read_log_content(
     }
 }
 
-pub fn filter_logs_on_type(
+pub fn filter_logs(
```
didier-wenzek, line 96: This function does what is expected by [#2966](https://github.com/thin-edge/thin-edge.io/issues/2966), but is a bit convoluted.

I would try to use a pipeline style:
- generate a list of `(path, modification_date, wildcard_match)`
- filter out paths out of date range (unless not being the result of a wildcard match)
- sort by modification date
- map to keep only the paths

crates/common/log_manager/src/log_utils.rs
```diff
@@ -93,42 +93,73 @@ pub fn read_log_content(
...
-) -> Result<Vec<PathBuf>, LogRetrievalError> {
-    let mut out = vec![];
+) -> Vec<PathBuf> {
+    // include log file with static path no matter whether or not it is in date range
+    files
+        .into_iter()
+        .filter(|(_, modification_time, wildcard_match)| {
+            !wildcard_match || *modification_time >= date_from
+        })
+        .map(|(path, _, _)| path)
```
didier-wenzek, line 157: I would do the sort by modification date here, before discarding the modification date that has already been extracted.

didier-wenzek:
> Question: when we are filtering logs by date, why only lower limit is taken into account (`date_from`) and we don't check upper limit (`date_to`)?

I think thin-edge uses these two time-related filters in an unexpected way.  The intent is to filter the log files using the timestamps of the log entries. However, thin-edge doesn't even try to decode these timestamps and only guesses that some entries are more recent that the `date_from` using the file modification date. Using `date_to` the same way would be really confusing (a whole log being discarded simply because of a recent entry).

rina23q:
> > Question: when we are filtering logs by date, why only lower limit is taken into account (`date_from`) and we don't check upper limit (`date_to`)?
> 
> I think thin-edge uses these two time-related filters in an unexpected way. The intent is to filter the log files using the timestamps of the log entries. However, thin-edge doesn't even try to decode these timestamps and only guesses that some entries are more recent that the `date_from` using the file modification date. Using `date_to` the same way would be really confusing (a whole log being discarded simply because of a recent entry).

It was a difficult decision how to "interpret" the c8y's keys `DateTo` and `DateFrom` to thin-edge. For some log types like syslog, there are timestamps in each line. If all logs were like that, the design would have been different (filter `inside` log files)

On the other hand, thin-edge needs a generic solution. Then, we are using `DateFrom` to filter by the file's modification time. I think it's an oversight that we don't use `DateTo` to filter that. However, I don't see the value to be very strict there since our solution is anyway not very ideal as I mentioned above.

crates/common/log_manager/src/log_utils.rs
```diff
@@ -172,95 +185,114 @@ mod tests {
...
+                config_type: "type_one".to_string(),
+            },
+            FileEntry {
+                path: format!("{tempdir_path}/file_c_two"),
+                config_type: "type_two".to_string(),
+            },
+            FileEntry {
+                path: format!("{tempdir_path}/file_d_one"),
+                config_type: "type_one".to_string(),
+            },
```
rina23q, line 285: I'm actually surprised that user can define multiple entries with the same `config_type`. But it makes sense. Good test case.

rina23q (commented):
I think it would be better to modify one system test.

Just giving out of range timestamp to `${start_timestamp}` in this test. Then the test will cover your change. Probably renaming test name makes also sense so that it tells that we ignore `dateFrom` and `dateTo` for a static path.

https://github.com/thin-edge/thin-edge.io/blob/4fbf29f913643c0df8f2609caa6da749f54d0d2d/tests/RobotFramework/tests/cumulocity/log/log_operation.robot#L16-L23

For the wildcard case, the `Log file request supports date/time filters and can search across multiple log files` test already covered enough, so no action needed.

gligorisaev (commented):
The PR addresses the issues highlighted in #2966

### Testing Performed

#### Integration Tests:

- Conducted integration tests to validate the end-to-end functionality of the error handling improvements.
- Simulated various MQTT connection error scenarios to verify the effectiveness of the new logic.

#### Manual Testing:

- Manually triggered MQTT connection errors to observe the system's behavior.
- Verified the accuracy and helpfulness of the error messages produced.

### Test Results

#### Manual Testing: 

- Manual tests corroborated the automated test results, demonstrating clear and actionable error messages during connection issues.

#### Observations
- The improved error handling significantly enhances the reliability of the tedge-mapper.
- The addition of retry logic helps mitigate transient connection issues without manual intervention.

**The implementation is robust, thoroughly tested, and ready for deployment.**

rina23q (approved):
LGTM

didier-wenzek (approved):
Approved

Issue #2990, opened 2024-07-11
Return exit code 1 if tedge config get doesn't find a value


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
Currently, `tedge config get` returns `0` when the provided key is not set.

```sh
root@0d735c3815ef:/setup# tedge config get aws.url
The provided config key: 'aws.url' is not set
root@0d735c3815ef:/setup# echo $?
0
```

It is inconvenient for writing a third party plugin. See Additional Context.

**Describe the solution you'd like**
Return `1` instead of `0` if the key doesn't have value set.
If the key has a default value, returns 0 since it's anyway set.

**Describe alternatives you've considered**

**Additional context**
While writing a configuration file for [tedge-monit-setup](https://github.com/thin-edge/tedge-monit-setup), this check "if c8y.url is not set, unmonitor c8y" is required. However, since the exit code is anyway `0` regardless whether it's set or not, the rule needed a kind of workaround.

```yml
check program c8y-enabled with path "/bin/sh -c '[ -n \0x22$(tedge config get c8y.url)\0x22 ]'"
    with timeout 5 seconds
    every 2 cycles
    if status != 0 then unmonitor
    group c8y
```

This could have been re-written more simply if the status is non-zero like below.

```yml
check program c8y-enabled with path "/usr/bin/tedge config get c8y.url"
    with timeout 5 seconds
    every 2 cycles
    if status != 0 then unmonitor
    group c8y
```

reubenmiller:
As reference, `git` also returns a non-zero exit code (1) if a specific configuration item is not set. This makes it possible to use a simple one-liner to do an action if something isn't set:

```sh
git config user.name || echo "user.name is not set"
```

rina23q:
Resolved by #2995

gligorisaev:
#### QA Report for Ticket #2990

**PR:** https://github.com/thin-edge/thin-edge.io/pull/2995



**Summary**

The purpose of this PR is to address and resolve the issues outlined in this ticket.


**Testing Performed**

Unit Tests:	

`crates/core/tedge/tests/main.rs`


**Integration Tests:**

Conducted integration tests to verify end-to-end functionality.
Tested various configuration update scenarios to confirm proper handling.

**Manual Testing:**

Manually updated configurations and observed system behavior.
Verified log entries for accuracy and completeness.

**Automated Testing**

`tests/RobotFramework/tests/tedge/tedge_config_get.robot`



[...]

PR #2995, opened 2024-07-12
fix(tedge): `tedge config get` returns 1 when a key is not set

rina23q:
## Proposed changes
Return `1` if `tedge config get` is called for unset key.

## Types of changes

## Paste Link to the issue
#2990 

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/config/commands/get.rs
+++ b/crates/core/tedge/src/cli/config/commands/get.rs
@@ -19,9 +19,11 @@ impl Command for GetConfigCommand {
             }
             Err(tedge_config::ReadError::ConfigNotSet { .. }) => {
                 eprintln!("The provided config key: '{}' is not set", self.key);
+                std::process::exit(1)
             }
             Err(tedge_config::ReadError::ReadOnlyNotFound { message, key }) => {
                 eprintln!("The provided config key: '{key}' is not configured: {message}",);
+                std::process::exit(1)
             }
             Err(err) => return Err(err.into()),
         }
--- a/crates/core/tedge/tests/main.rs
+++ b/crates/core/tedge/tests/main.rs
@@ -88,7 +88,7 @@ mod tests {
         // We start we no certificate, hence no device id
         get_device_id_cmd
             .assert()
-            .success()
+            .failure()
             .stderr(predicate::str::contains("'device.id' is not configured"));
 
         // The create command created a certificate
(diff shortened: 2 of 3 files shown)


Bravo555 (approved):

crates/core/tedge/src/cli/config/commands/get.rs
```diff
@@ -19,6 +19,7 @@ impl Command for GetConfigCommand {
             }
             Err(tedge_config::ReadError::ConfigNotSet { .. }) => {
                 eprintln!("The provided config key: '{}' is not set", self.key);
+                std::process::exit(1)
             }
             Err(tedge_config::ReadError::ReadOnlyNotFound { message, key }) => {
                 eprintln!("The provided config key: '{key}' is not configured: {message}",);
```
didier-wenzek, line 25:
```suggestion
                eprintln!("The provided config key: '{key}' is not configured: {message}",);
                std::process::exit(1)
```

Even better: simply remove the case for `Err(tedge_config::ReadError::ReadOnlyNotFound {})` and let it handled by the catch all case .i.e. `Err(err) => return Err(err.into())`.

rina23q: The default of `clap` returns `2` for the error case. Your suggestion will change the exit code

didier-wenzek: I'm speaking about the second case `ReadOnlyNotFound`, not the one you fixed `ConfigNotSet`.

`ReadOnlyNotFound` must exit 1 or 2 but not 0 as today.

```
$ tedge config unset device.cert_path
$ tedge config get device.id                                                   
The provided config key: 'device.id' is not configured: The device id is read from the device certificate.
To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.
$ echo $?         
0
```

rina23q: This `execute()` is called here. 

https://github.com/thin-edge/thin-edge.io/blob/e77a5dfb64e7dd9da45b7ba2429af2d56101ba52/crates/core/tedge/src/main.rs#L65-L66

> Even better: simply remove the case for Err(tedge_config::ReadError::ReadOnlyNotFound {}) and let it handled by the catch all case .i.e. Err(err) => return Err(err.into()).

I prefer your original code suggestion, just to add `std::process::exit(1)` to `ReadOnlyNotFound` case. The reason is, if it returns `Err`, `.with_context()` will be called, that adds extra stderr output compared to now. I feel it's unnecessary change to user.

rina23q: Done in 0458c17271a50a6b6efb9dd9167dc5a68eeb4083

didier-wenzek (approved):
Approved

PR #2997, opened 2024-07-15
Improve c8y_mapper_ext operation handler API

Bravo555:
## TODO

- [x] ~~Stop parsing the operation message payload twice. First parse it as a generic operation, and then only parse the rest of the payload~~ A bit complicated to do right now, would require `Command<Payload>` and `GenericCommandState` to be refactored, better to do it in another PR.
- [x] add a test to verify the behaviour of the `handle` function (https://github.com/thin-edge/thin-edge.io/pull/2997#discussion_r1684226156)

## Proposed changes

- Different operation functions themselves no longer send a clearing message, now done by the main handler
- All error paths now send a proper c8y operation fail message, where previously some error paths (e.g. failure to convert path to utf-8, `ChannelError` when trying to send a message through a channel) would not
-  moved tests of some operations into these operations' modules
- Generating `EXECUTING` and `FAILED` messages out to main handler.

## Types of changes

## Paste Link to the issue
- #2880 

## Checklist

## Further comments

This PR's focus is mainly removing unnecessary assumptions about MQTT API from operation functions and centralizing them in the main handler. This makes the code simpler to understand as it provides a more rigid structure and a simpler interface, and serves as a stepping stone to eventual extraction of clearing messages to the same place where we're initiating local operations. Once operation initialization and clearing messages will be in the same place, we can move the state that keeps track of which operations are running there and stop relying on MQTT broker echo of command id messages back to the mapper for it to handle it.

Also, the PR contains one new test that checks some paths in the `OperationHandler::handle` method. This test does not utilize sending raw MQTT messages through the actor message boxes. I plan to add more tests like this for individual operations, that more thoroughly test the possible branches.

Finally, the PR tackles _clean up operation API_ section of https://github.com/thin-edge/thin-edge.io/issues/2880#issuecomment-2223213567, but doesn't fully close the task. I'd like to handle the `SUCCESSFUL` messages in another PR, to keep this one relatively small.


crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -82,6 +81,7 @@ pub fn succeed_operation(
 
 #[derive(Debug, Copy, Clone)]
 pub enum CumulocitySupportedOperations {
+    C8ySoftwareList,
```
rina23q, line 84: `c8y_SoftwareList` is a very old operation replaced by `c8y_SoftwareUpdate` long time ago. If we need this, something is wrong (likely confusion of the inventory fragment name `c8y_SoftwareList`).

Bravo555: Ah, I didn't know that, it seemed weird that it was missing, now i know the reason :) I'll replace it with a proper usage of `c8y_SoftwareUpdate`

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -374,185 +512,104 @@ fn get_smartrest_response_for_upload_result(
...
+    use tedge_actors::SimpleMessageBoxBuilder;
     use tedge_test_utils::fs::TempTedgeDir;
 
-    const TEST_TIMEOUT_MS: Duration = Duration::from_millis(3000);
+    use crate::tests::test_mapper_config;
+
+    use super::*;
 
     #[tokio::test]
-    async fn mapper_converts_config_metadata_to_supported_op_and_types_for_main_device() {
```
didier-wenzek, line 391: This test has been moved to `tests.rs`, while other tests has been moved out of `tests.rs` and dispatched over misc `xxx.rs`.

We should really avoid to do that on a refactoring PR. Reorganize either the code or the tests, not both simultaneously. This makes things harder to understand & check for little benefits.

crates/extensions/c8y_mapper_ext/src/operations/software_update.rs
```diff
@@ -1,79 +1,73 @@
...
 
-                vec![
-                    MqttMessage::new(&topic, smartrest_set_operation),
-                    command.clearing_message(&self.mqtt_schema),
-                    self.request_software_list(&target.topic_id),
-                ]
+                Ok(OperationResult::Finished {
+                    messages: vec![
+                        MqttMessage::new(topic, smartrest_set_operation),
+                        self.request_software_list(&target.topic_id),
```
didier-wenzek, line 52: Thanks to this refactoring PR, there is now an opportunity to remove from here the software list request, as command finalization in now done in one place

Unfortunately, this is not enough to remove the need for an id generator clone (currently both the converter and the running operation for the software update need to generate id). An option could be to delegate the `OperationHandler`  the generation of operation ids.

didier-wenzek (approved):
Approved. This is a really nice simplification of the operation handlers.

crates/extensions/c8y_mapper_ext/src/operations/software_update.rs
Bravo555: > Thanks to this refactoring PR, there is now an opportunity to remove from here the software list request, as command finalization in now done in one place

I'm not sure if removing `software_list` publish from `software_update` is practical:

1. Local `software_update` operation is published on local MQTT broker in `c8y_mapper_ext/src/converter.rs:770`, in function `forward_software_request`
2. This local `software_update` operation is picked up by the `OperationHandler` which runs `c8y_mapper_ext::operations::software_update::OperationContext::publish_software_update_status`
3. `publish_software_update_status` returns "succeed operation" message with no payload, and invokes a local `software_list` operation after `software_update` operation is succeeded or failed.
4. `publish_software_list`, when it's successful, sends the software list via HTTP proxy or smartrest.

What I'm seeing is that `publish_software_list` isn't an operation, and it can be invoked anytime, just after `software_update` software list changes, so it means it has to be updated. `software_list` isn't necessary to finalize the `software_update` operation (we send together success/failure of `software_update` and invocation of `software_list`), and `software_update` is the only operation that does it, so it makes little sense to extract it.

The only thing that's problematic is sharing this id generator, because generating the ids is necessary for creating local operations. Right now, creating local operations is also spread among methods in `CumulocityConverter`: `forward_software_request`, `forward_restart_request`, `forward_operation_request`, but also `convert_log_upload_request`, `convert_config_snapshot_request` etc. in `operations` module. They all do the same thing (converting c8y operation messages to local thin-edge operation messages), so they should be put together in a single place, either in `OperationHandler`, or in some other object that will have smaller scope. It would then contain the `IdGenerator` and provide methods to start new local operations by publishing on the local operation topics. `software_update` would then simply call this to create a new `software_list` operation instead of doing it manually.

didier-wenzek: > What I'm seeing is that `publish_software_list` isn't an operation, and it can be invoked anytime, just after `software_update` software list changes, so it means it has to be updated. `software_list` isn't necessary to finalize the `software_update` operation (we send together success/failure of `software_update` and invocation of `software_list`), and `software_update` is the only operation that does it, so it makes little sense to extract it.

Yes, sending a software list after a software update is not strictly required by c8y and is always done at the device initiative. This is done in two cases: when the  software update capability is added and after each software update request.

Technically, I see no reason to have this done inside the software update handler or elsewhere.
My concern is that one should have *only* one component that trigger operations on behalf of c8y.

> The only thing that's problematic is sharing this id generator, because generating the ids is necessary for creating local operations.

We are fully aligned here.

> Right now, creating local operations is also spread among methods in `CumulocityConverter`: `forward_software_request`, `forward_restart_request`, `forward_operation_request`, but also `convert_log_upload_request`, `convert_config_snapshot_request` etc. in `operations` module. They all do the same thing (converting c8y operation messages to local thin-edge operation messages), so they should be put together in a single place, either in `OperationHandler`, or in some other object that will have smaller scope. It would then contain the `IdGenerator` and provide methods to start new local operations by publishing on the local operation topics. `software_update` would then simply call this to create a new `software_list` operation instead of doing it manually.

As you say: *they should be put together in a single place, either in `OperationHandler` or in some other object.*
I'm looking for something symmetric:
- the object creating an operation request should also be the object that clears that operation request when done
- the `OperationHandler` seems to be a nice candidate for that
- when clearing a software update this object could also possibly trigger a following software list request 
- in case you wonder: I'm not requesting to do that in this PR.

didier-wenzek:
Observed for the second time since yesterday on merge queue:

```
test bridge_reconnects_successfully_after_cloud_connection_interrupted ... ok
test bidirectional_forwarding_avoids_infinite_loop ... ok
test bridge_many_messages ... ok
test bridge_reconnects_successfully_after_local_connection_interrupted ... FAILED

test result: FAILED. 3 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 5.09s
```

albinsuresh (approved):
LGTM. I'm still not able to make up my mind on what the right way to handle channel errors emitted from these spawned tasks are. But since the current mechanism also works, approving.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -210,74 +222,123 @@ impl RunningOperation {
...
+                    OperationResult::Executing => {
+                        let c8y_state_executing_payload = set_operation_executing(c8y_operation);
+                        let c8y_state_executing_message = MqttMessage::new(
+                            &entity.smartrest_publish_topic,
+                            c8y_state_executing_payload,
+                        );
+                        mqtt_publisher
+                            .send(c8y_state_executing_message)
+                            .await
+                            .unwrap();
```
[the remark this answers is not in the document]
Bravo555: > And these terminated tasks will be removed from `running_operations` when the next message for that operation type is handled.

Not next message for that operation type, but next message for that particular operation, with that particular id.

But I think there's a problem: in practice, when a panic in a child task happens, `.handle()` will never get called for that operation again. Operation task itself sends the messages, so if they can't be sent over MQTT because of a panic, they won't be handled, won't be joined, so we will not see that an exception has occurred.

For now, in this PR I will add a test to verify the behaviour of the `handle` function - when does it panic and when does it actually join the tasks.

But what I realize is that `OperationHandler` needs its own background task that joins the operation tasks (i.e. polls next future in `FuturesUnordered`), which will have to be added, or I'll have to make `OperationHandler` an actor, which I probably should. So far I've avoided it because I wanted `OperationHandler` to be a small object used and managed by the converter, not another actor, and also because our actors tend to have a lot of boilerplate, but the time for that is probably now and it's what I'll have to do in a follow-up PR.

So AFAIK right now we don't actually join the panicked tasks and don't know that a panic has occurred. This is the first thing that I should fix. Then, I'll need to make sure tasks are joined automatically so `.handle()` doesn't have to be called to know whether or not there's been a panic. I leave to you the decision of whether these changes should be made in a follow-up PR, or in this one.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2610,6 +2351,275 @@ async fn mapper_processes_operations_concurrently() {
...
+        json!({
+            "status": "successful",
+        })
+        .to_string(),
+    ))
+    .await
+    .unwrap();
+
+    assert_received_contains_str(&mut mqtt, [("c8y/s/us", "503,c8y_Restart")]).await;
+}
```
Bravo555, line 2452: @albinsuresh My bad, by mistake I included the test you were proposing (https://github.com/thin-edge/thin-edge.io/pull/2904#discussion_r1620205761) in this PR, but since it's already here, you can take a look.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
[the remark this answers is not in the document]
albinsuresh: > I'll have to make OperationHandler an actor, which I probably should

Yeah, that's pretty much aligned with the model that I had in mind too, esp after the introduction of the availability actor.

> but the time for that is probably now and it's what I'll have to do in a follow-up PR.

That's fine. One step at a time.

crates/extensions/c8y_mapper_ext/src/tests.rs
albinsuresh: Looks fine.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -375,185 +530,280 @@ fn get_smartrest_response_for_upload_result(
...
+        }
+    }
+
+    struct TestHandle {
+        operation_handler: OperationHandler,
+        mqtt: SimpleMessageBox<MqttMessage, MqttMessage>,
+        c8y_proxy: FakeServerBox<C8YRestRequest, C8YRestResult>,
+        uploader: FakeServerBox<IdUploadRequest, IdUploadResult>,
+        downloader: FakeServerBox<IdDownloadRequest, IdDownloadResult>,
+        _ttd: TempTedgeDir,
```
albinsuresh, line 807:
```suggestion
        let mqtt = mqtt_builder.build();
        let downloader = downloader_builder.build();
        let uploader = uploader_builder.build();
        let c8y_proxy = c8y_proxy_builder.build();

        TestHandle {
            mqtt,
            downloader,
            uploader,
            c8y_proxy,
            operation_handler,
            ttd,
        }
    }

    struct TestHandle {
        operation_handler: OperationHandler,
        mqtt: SimpleMessageBox<MqttMessage, MqttMessage>,
        c8y_proxy: FakeServerBox<C8YRestRequest, C8YRestResult>,
        uploader: FakeServerBox<IdUploadRequest, IdUploadResult>,
        downloader: FakeServerBox<IdDownloadRequest, IdDownloadResult>,
        ttd: TempTedgeDir,
```

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -375,185 +530,280 @@ fn get_smartrest_response_for_upload_result(
...
-
-        // Assert that the updated config type list does not trigger a duplicate supported ops message
-        assert_received_contains_str(
-            &mut mqtt,
-            [(
-                "c8y/s/us/test-device:device:child1",
-                "119,typeB,typeC,typeD",
-            )],
+        // normally clearing message would be sent by operation task, but in this case we send it
+        // manually
```
albinsuresh, line 745:
```suggestion
        // normally clearing message would be sent by operation task.
        // Using it here just as a dummy, to call `handle` with the same cmd-id, so that it panics
```

albinsuresh (approved):
LGTM. Just some minor cosmetic comments.

Issue #2859, opened 2024-05-06
use socket activated service to launch c8y-remote-access-plugin when using systemd


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Using the Cumulocity IoT Remote Access feature to connect to a device via SSH is limiting as the `c8y-remote-access-plugin` process is spawned by the `tedge-mapper-c8y` process, meaning that any command used the result in restarting the tedge-mapper-c8y service will disconnect the ssh session.

Due to this limitation, the following actions are not possible:

* Updating tedge-mapper package
* Restarting tedge-mapper-c8y service

**Describe the solution you'd like**

On devices which use SystemD, use a [socket-activated service](https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html) to launch the c8y-remote-access-plugin process which is independent from the tedge-mapper-c8y service.

In order to get full independence, the following changes would have to be done:

* Change installation script to install the socket handling logic only on systems where SystemD is installed
* Use a direct connection to the Cumulocity IoT URL instead of using the local c8y proxy service (as the service is maintained by the tedge-mapper-c8y service) - (technically this will be reverting the change where we switched to using the local c8y proxy)
* Optional: Improve the c8y-remote-access-plugin handshake so that the when the process is launched via a socket, that the client publishing the message can easily verify if the c8y-remote-access-plugin was able to successfully connect to c8y or not
* Support sending a message to either a TCP or UNIX socket via the cli (possibly in the c8y-remote-access-plugin, tedge, or tedge-mapper so other potential operation handlers can take advantage of it?)

**Important Notes about socket activated services**

* SystemD is responsible for socket creation and listening to incoming client connections (which means very minimal overhead)
* The c8y-remote-access-plugin does not have to be running prior to any incoming client connections. SystemD will launch the process when a client tries to connect/send a message to the socket

**Describe alternatives you've considered**

Alternatively, the SystemD [KillMode](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html) of the **tedge-mapper-c8y** service could be changed to `process` (from the default `control-group`), however this is a best practice as it could result in orphaned processes, and it would mean that the tedge-mapper-service would have to managed any child process launched by itself so it could prevent said orphaned processes (this is currently managed by SystemD).

**Additional context**

A detailed guide to how a socket activated service would work to launch an independent c8y-remote-access-plugin process:

[...]

PR #3006, opened 2024-07-18
feat: use direct c8y connection for c8y-remote-access-plugin

rina23q:
## Proposed changes

Remove the c8y-remote-access-plugin connection dependency on the c8y local proxy which is launched by the **tedge-mapper-c8y** service. This is part one of the large ticket (#2859) which looks at creating the remote access connection which is fully independent to the **tedge-mapper-c8y** service, thus allowing the user to execute `tedge reconnect c8y` without the remote access connection from being dropped.

**Note**

Prior to the c8y local proxy feature, the c8y-remote-access-plugin did use a direct connection to Cumulocity IoT, however the implementation was switch to use the local proxy which adds an additional connection dependency on the tedge-mapper-c8y (e.g. if you restart the tedge-mapper-c8y, the local proxy will also restart thus stopping the connection). This PR does not change the fact that there is still a process dependency between the spawned c8y-remote-access-plugin instance and the **tedge-mapper-c8y** service, however this will be addressed via #3007 

The PR is based on reverting the c8y local proxy commit: https://github.com/thin-edge/thin-edge.io/commit/845480e066769b51a8d4114ffe2fc79c1ced8fa9

## Types of changes

## Paste Link to the issue
#2859

> Use a direct connection to the Cumulocity IoT URL instead of using the local c8y proxy service (as the service is maintained by the tedge-mapper-c8y service) - (technically this will be reverting the change where we switched to using the local c8y proxy)

## Checklist

### DIFF
--- a/plugins/c8y_remote_access_plugin/src/auth.rs
+++ b/plugins/c8y_remote_access_plugin/src/auth.rs
@@ -1,0 +1,22 @@
+use c8y_api::http_proxy::C8yMqttJwtTokenRetriever;
+use miette::IntoDiagnostic;
+use tedge_config::TEdgeConfig;
+
+pub struct Jwt(String);
+
+impl Jwt {
+    pub fn authorization_header(&self) -> String {
+        format!("Bearer {}", self.0)
+    }
+
+    pub async fn retrieve(config: &TEdgeConfig) -> miette::Result<Jwt> {
+        let mut retriever =
+            C8yMqttJwtTokenRetriever::from_tedge_config(config).into_diagnostic()?;
+
+        retriever
+            .get_jwt_token()
+            .await
+            .map(|resp| Jwt(resp.token()))
+            .into_diagnostic()
+    }
+}
--- a/plugins/c8y_remote_access_plugin/src/lib.rs
+++ b/plugins/c8y_remote_access_plugin/src/lib.rs
@@ -1,7 +1,3 @@
-use std::fmt::Display;
-use std::fmt::Formatter;
-use std::process::Stdio;
-
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use futures::future::try_select;
@@ -10,6 +6,7 @@ use input::parse_arguments;
 use miette::miette;
 use miette::Context;
 use miette::IntoDiagnostic;
+use std::process::Stdio;
 use tedge_config::TEdgeConfig;
 use tedge_utils::file::create_directory_with_user_group;
 use tedge_utils::file::create_file_with_user_group;
@@ -17,11 +14,13 @@ use tokio::io::AsyncBufReadExt;
 use tokio::io::BufReader;
 use url::Url;
 
+use crate::auth::Jwt;
 pub use crate::input::C8yRemoteAccessPluginOpt;
 use crate::input::Command;
 use crate::input::RemoteAccessConnect;
 use crate::proxy::WebsocketSocketProxy;
 
+mod auth;
 mod csv;
 mod input;
 mod proxy;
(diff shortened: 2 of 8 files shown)


didier-wenzek:
@jarhodes314 Can you please review this PR. Thanks

jarhodes314 (approved):

reubenmiller:
I've pushed some system tests, but don't be surprised if there are a few failures as the remote access has a few dependencies (e.g. generate ssh keypair, install go-c8y-cli, calling go-c8y-cli to create the passthrough configuration, and then run the ssh command).

plugins/c8y_remote_access_plugin/src/lib.rs
```diff
@@ -155,38 +154,19 @@ async fn spawn_child(command: String, config_dir: &Utf8Path) -> miette::Result<(
...
+        .into_diagnostic()?
+        .to_string();
+    let url = build_proxy_url(host.as_str(), command.key())?;
+    let jwt = Jwt::retrieve(&config)
+        .await
+        .context("Failed when requesting JWT from Cumulocity")?;
 
-    let proxy =
-        WebsocketSocketProxy::connect(&url, command.target_address(), client_config).await?;
+    let proxy = WebsocketSocketProxy::connect(&url, command.target_address(), jwt).await?;
```
reubenmiller, line 169: @jarhodes314  Do we know which ca certificates are used here (e.g. system or the custom c8y.root_cert_path setting?). Or would this handling be covered by your PR, https://github.com/thin-edge/thin-edge.io/pull/2953

jarhodes314: Good point. It'll almost certainly not be using `root_cert_path`, but I'll deal with it in my PR.

gligorisaev (commented):
The added Test Step is OK, however having in mind that this test has additional test steps comming from https://github.com/thin-edge/thin-edge.io/pull/3007 I will do the more detail check in that PR

reubenmiller:
> The added Test Step is OK, however having in mind that this test has additional test steps coming from #3007 I will do the more detail check in that PR

Though we can't yet test the independence of the connection as a socket activated service is required for this (which is provided in #3007)

plugins/c8y_remote_access_plugin/src/auth.rs
```diff
@@ -0,0 +1,22 @@
...
+        let mut retriever =
+            C8yMqttJwtTokenRetriever::from_tedge_config(config).into_diagnostic()?;
+
+        retriever
+            .get_jwt_token()
+            .await
+            .map(|resp| Jwt(resp.token()))
+            .into_diagnostic()
+    }
+}
```
reubenmiller, line 22: Do we have any built-in retries if the initial retrieval of the JWT fails?

jarhodes314: Yep, this is using the same logic as the rest of the mapper for retries (although I suspect in general the 20s wait will be longer than useful here). That said, I expect the retries to generally be useful when starting things up for the first time, where we might not have a c8y connection, rather than in this case where we must have a connection as we were triggered from an operation.

reubenmiller: Ok, that sounds fine. And yes the 20s wait would cause a timeout on the server side (in the remote access microservice), but we can ignore this use-case...and once Cumulocity IoT HTTP requests support certificate authentication, then the JWT might become less relevant.

Bravo555 (approved):
I'm not very familiar with the OG commit that's being reverted, but I don't see anything wrong.

gligorisaev:
LGTM

PR #3012, opened 2024-07-23
refactor: Reorganise c8y_mapper_ext::operations module

Bravo555:
## Proposed changes

No functional changes were made.

`operations` module was reorganised into submodules to better separate different concerns:
- plain conversions of operations-related messages
- topic filters
- spawning and joining of operation tasks
- handling different types of operations

This prepares the `OperationHandler` to more neatly be converted into an actor, as the lower-level handling of different operation types should remain unaffected.

## Types of changes

## Paste Link to the issue
- #2880 

## Checklist


crates/extensions/c8y_mapper_ext/src/operations/implementation/config_update.rs
```diff
@@ -94,70 +64,7 @@ impl OperationContext {
...
-            config_type: config_download_request.config_type.clone(),
-            path: None,
-            log_path: None,
-        };
-
-        // Command messages must be retained
-        vec![MqttMessage::new(&topic, request.to_json()).with_retain()]
-    }
-}
+impl CumulocityConverter {}
```
rina23q, line 67: Should have been removed, right?

rina23q (commented):
The directory name `implementation` can be improved since all codes in repo are a kind of "implementation". Maybe `contexts` (as all files are inheriting `OperationContext`) or `commands` (as all files are specific operation type but the name `operation` is already taken).

crates/extensions/c8y_mapper_ext/src/operations/topic_filter.rs
```diff
@@ -0,0 +1,43 @@
+//! Operations' topic filters.
```
albinsuresh, line 1: Not quite happy with this module with all the operation specific filters. I see that you wanted to separate the conversion aspects from the command handling aspects. In that case I get a feeling that we need two separate operation specific modules for each operation/command type: one with just the conversion aspects and other with the command processing aspects as done now.

albinsuresh (commented):
As @rina23q already raised, I'm not quite happy with the module named `implementation` either. How about calling that wrapper module `handlers` since those modules are extensions of the common handler?

crates/extensions/c8y_mapper_ext/src/operations/topic_filter.rs
Bravo555: Yeah, I wanted to separate these because reacting to operation status changes and MQTT topics are really different concerns, and I went on a bit of an autopilot separating modules, but looking at these functions, they're completely superfluous, the implementation is the same, just with different arguments.

> I get a feeling that we need two separate operation specific modules for each operation/command type: one with just the conversion aspects and other with the command processing aspects as done now.

I think that wrt. the conversion aspect, there are really two things here: the topic filters (`topic_filter` module) and the conversions of incoming cumulocity messages that initiate thin-edge.io operations (`convert` module).

## On the topic fiilters:

I'm not sure if we need a conversion module per operation type. Topic filters are necessary to create an MQTT subscription, and in this case I think the pattern is more like:
- `OperationHandler` handles some set of operations
- if `CumulocityConverter` gets an operation of type that can be handled by `OperationHandler`, it should pass the message to the handler
- so `OperationHandler` needs to expose the set of operations it supports, so that `CumulocityConverter` makes a proper MQTT subscription
- but it doesn't have to be per operation type. I think what's preferable is that `OperationHandler` produces only one topic filter, that covers all the operations that are enabled by the current configuration.

Also this will hopefully clear up more once we've made `OperationHandler` an actor: the topic filters will fit neatly into the message types of the new operations actor.

## On the conversion functions in the `convert` module:

These are impl blocks on `CumulocityConverter`, which I've pulled under `operations` module because other than converting between Cumulocity messages and thin-edge.io messages, they perform some side effects like querying the entity store or calling into c8y proxy to obtain a URL where the operation has to upload a file.

This _auxiliary communication with C8y when we receive an operation_ is very much symmetric to _auxiliary communication with C8y when we terminate the operation (support or failure)_, so the long term plan is definitely to move these side-effects into `OperationHandler`, but right now it's very much coupled to `CumulocityConverter` which uses it primarily to convert between MQTT messages, so I think refactoring this will be best left as a follow-up.

## Conclusion

- as for the topic filters, instead of having different topic filters for different operation types, there should be `OperationHandler::topic_filter` that just creates a filter for all operation types the handler is interested in, and `CumulocityConverter` should just call it and don't care about different operation types.
- as for the `convert` module functions, I think this should be left as a follow-up.

How does that sound?

crates/extensions/c8y_mapper_ext/src/operations/implementation/config_update.rs
Bravo555: Resolved in 80e6962

crates/extensions/c8y_mapper_ext/src/operations/topic_filter.rs
Bravo555: `implementation` renamed in 80e6962, `OperationHandler::topic_filter` introduced in 6c7d4ae

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2779,13 +2780,20 @@ pub(crate) fn test_mapper_config(tmp_dir: &TempTedgeDir) -> C8yMapperConfig {
...
-        &mqtt_schema,
-    ));
-    topics.add_all(crate::operations::config_update::topic_filter(&mqtt_schema));
+
+    let capabilities = Capabilities {
+        log_upload: true,
+        config_snapshot: true,
+        config_update: true,
+        firmware_update: true,
+    };
```
rina23q, line 2789: This one is equal to `Capapilities::default()`.

crates/extensions/c8y_mapper_ext/src/operations/mod.rs
```diff
@@ -1,809 +1,29 @@
...
 //! This module contains:
-//! - data definitions of various states which are necessary to maintain in the mapper
-//! - status and error handing utilities for reporting operation success/failure in different ways
-//!   (MQTT, Smartrest)
-//! - implementations of operations
+//! - operation handler, which handles thin-edge operation MQTT messages by spawning tasks that
+//!   handle different operations ([`handler`])
+//! - operation topic filters ([`topic_filter`])
+//! - conversion from C8y operation messages into thin-edge operation messages ([`convert`])
+//! - implementations of operations ([`implementation`])
```
rina23q, line 15: These lines need to be updated accordingly to your recent changes.

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -198,24 +199,11 @@ impl C8yMapperConfig {
...
-        }
-        if capabilities.firmware_update {
-            topics.add_all(
-                crate::operations::firmware_update::firmware_update_topic_filter(&mqtt_schema),
-            );
-        }
+        let operation_topics = OperationHandler::topic_filter(&capabilities)
+            .into_iter()
+            .map(|(e, c)| mqtt_schema.topics(e, c))
+            .collect();
```
rina23q, line 205: Nice simplification!

rina23q (approved):
My comments are not related to the main code, so I would approve now.

albinsuresh (approved):
LGTM

Issue #2960, opened 2024-06-24
built-in bridge connection not sending or receiving messages from the cloud


reubenmiller:
**Describe the bug**

The tedge-mapper-c8y process stops sending data to Cumulocity IoT when using the built-in bridge.

The device where the problem was seen includes a special MQTT subscription which looks at the `tedge-mapper-bridge-c8y` service status, and the log shows that the status was toggling between `up` and `down`, though it is currently showing as `up` even though the device is not responding to operations or sending measurements.

```json
{"tst":"2024-06-20T17:33:43.322236+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":1,"payloadlen":15,"payload":"{\"status\":\"up\"}"}
{"tst":"2024-06-20T17:34:26.586193+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":1,"payloadlen":15,"payload":"{\"status\":\"up\"}"}
{"tst":"2024-06-20T21:03:53.619944+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":0,"payloadlen":17,"payload":"{\"status\":\"down\"}"}
{"tst":"2024-06-20T21:04:22.153423+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":0,"payloadlen":15,"payload":"{\"status\":\"up\"}"}
{"tst":"2024-06-20T23:47:14.195201+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":0,"payloadlen":17,"payload":"{\"status\":\"down\"}"}
{"tst":"2024-06-20T23:47:42.685712+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":0,"payloadlen":15,"payload":"{\"status\":\"up\"}"}
{"tst":"2024-06-20T23:57:14.234556+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":0,"payloadlen":17,"payload":"{\"status\":\"down\"}"}
{"tst":"2024-06-20T23:57:45.771221+0200","topic":"te/device/main/service/tedge-mapper-bridge-c8y/status/health","qos":0,"retain":0,"payloadlen":15,"payload":"{\"status\":\"up\"}"}
[...]

PR #3018, opened 2024-07-26
fix: Handle messages republished by MQTT bridge following a disconnection event

jarhodes314:
## Proposed changes
The OSADL device running the built in bridge was occasionally disconnecting from cloud while sending messages and then failing to recover upon reconnection. The root cause was that unacknowledged messages were being republished by `rumqttc` upon reconnection, but the code was always waiting for a corresponding event from the other half of the bridge (as it implicitly assumed each message was only published once). Since there was no such corresponding event, this caused the bridge to block.

This PR ignores the duplicate publishes in the section of code that was blocking to fix this bug.

Note: The PR includes a workaround for the detected [flaky test](https://github.com/thin-edge/thin-edge.io/issues/3021)

## Types of changes

## Paste Link to the issue
- #2960

## Checklist


crates/extensions/tedge_mqtt_bridge/tests/bridge.rs
```diff
@@ -116,6 +116,75 @@ async fn bridge_many_messages() {
...
+                "c8y/s/us",
+                QoS::AtMostOnce,
+                false,
+                "a,fake,smartrest,message",
+            )
+            .await
+            .unwrap();
+    }
+
+    // tokio::time::sleep(Duration::from_secs(5)).await;
```
reubenmiller, line 178: Remove commented out code

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -36,6 +37,7 @@ use tedge_actors::RuntimeError;
 use tedge_actors::RuntimeRequest;
 use tedge_actors::RuntimeRequestSink;
 use tracing::info;
+use tracing::log::debug;
```
rina23q, line 40: Trivial

```suggestion
use tracing::debug;
```

crates/extensions/tedge_mqtt_bridge/tests/bridge.rs
```diff
@@ -398,7 +467,7 @@ impl Proxy {
                         };
 
                         write_socket.shutdown().await.unwrap();
-                        write_conn.shutdown().await.unwrap();
+                        let _ = write_conn.shutdown().await; //.await.unwrap();
```
rina23q, line 470: Why accepting any return value? with `.await.unwrap()` looks good enough although it's commented out.

jarhodes314: We want to shut the connection down cleanly, but it may have already closed. If it is already closed, we should just ignore the error we get.

jarhodes314:
I've now removed the commented out code and tidied up the import.

@reubenmiller suggested that I should also modify the JWT retrieving logic in the auth proxy so that it doesn't panic when we don't retrieve a JWT, although that isn't trivial and I think trying to get that change rushed in is more likely to do harm than good. I'm confident that the panicking JWT handler isn't the root cause (if the auth proxy panicking was causing the bridge to go down, I would expect it to crash the entire mapper, and the same error appeared multiple times, so the panic is limited to just the single request to the auth proxy).

rina23q (dismissed):
LGTM

rina23q (approved):
Re-approved

Issue #3024, opened 2024-07-30
tedge connect c8y does not cleanly close the local MQTT connection


reubenmiller:
**Describe the bug**

The following MQTT client instances do not send an MQTT `DISCONNECT` to official close the connection.

* check_connection_c8y
* get_jwt_token_c8y

Below are the mosquitto log entries showing the abrupt disconnects (which seem to be visible via the `OpenSSL Error` message):

**client id: check_connection_c8y**

```log
Jul 30 07:09:10 fc0518bf190f mosquitto[418]: 1722323350: OpenSSL Error[0]: error:0A000126:SSL routines::unexpected eof while reading
Jul 30 07:09:10 fc0518bf190f mosquitto[418]: 1722323350: Client check_connection_c8y closed its connection.
```

**client id: get_jwt_token_c8y**

```log
Jul 30 07:33:55 7cf39fd378c8 mosquitto[418]: 1722324835: OpenSSL Error[0]: error:0A000126:SSL routines::unexpected eof while reading
Jul 30 07:33:55 7cf39fd378c8 mosquitto[418]: 1722324835: Client get_jwt_token_c8y closed its connection.
```

**To Reproduce**

When using TLS settings on the local MQTT mosquitto broker, the following commands seem to be enough to invoke the error:

```sh
tedge connect c8y
tedge connect c8y --test
```

The following ticket is a similar ticket which has already been solved but was another instance of an MQTT client not closing the connection cleanly:

* https://github.com/thin-edge/thin-edge.io/issues/2922

**Expected behavior**

In normal operation, an MQTT client should close the MQTT connection cleanly by sending a `DISCONNECT` packet to the broker.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `All`
 - Hardware [incl. revision]: `All`
 - System-Architecture [e.g. result of "uname -a"]: `All`
 - thin-edge.io version [e.g. 0.1.0]: `1.1.2~320+g48a9843`

**Additional context**

reubenmiller:
The linked PR includes a system test: `Check absence of OpenSSL Error messages #3024`

gligorisaev:
The bug is not reproducable anymore.
Review done in PR: https://github.com/thin-edge/thin-edge.io/pull/3025
Test Case: tests/RobotFramework/tests/tedge_connect/tedge_connect_test.robot

PR #3025, opened 2024-07-30
fix: disconnect cleanly from local MQTT broker in tedge connect

reubenmiller:
## Proposed changes

Send MQTT `DISCONNECT` on local MQTT connections (in normal operation) in the `tedge connect` and `tedge reconnect` commands.

## Types of changes

## Paste Link to the issue

* https://github.com/thin-edge/thin-edge.io/issues/3024

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/connect/command.rs
+++ b/crates/core/tedge/src/cli/connect/command.rs
@@ -289,6 +289,7 @@ fn check_device_status_c8y(tedge_config: &TEdgeConfig) -> Result<DeviceStatus, C
         .network_options
         .set_connection_timeout(CONNECTION_TIMEOUT.as_secs());
     let mut acknowledged = false;
+    let mut exists = false;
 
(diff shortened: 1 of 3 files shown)


reubenmiller:
First commit is to add the system test which should fail, and the second commit will be to fix the bug.

gligorisaev (commented):
The added test step is doing what is intended for

gligorisaev (commented):
The added test was reviewed and it is passing

Bravo555 (approved):
Confirming that this patch fixes the problem. 
Simple and sufficient for now, but another example of `rumqttc` API being a bit too verbose for our use case here. For a more elegant fix in the future, we should either use our `mqtt-channel` or create a wrapper over `rumqttc::Client` that at minimum provides a disconnect on drop.

rina23q (approved):

PR #3032, opened 2024-07-31
refactor: add OperationHandlerActor

Bravo555:
## Proposed changes

A preliminary OperationHandlerActor which still uses OperationHandler underneath, but receives command messages using the actor runtime.

This initial implementation shows the limitations of the entity store, which is now a part of CumulocityConverter, and other actors don't have any access to it. Because OperationHandlerActor needs some information about a registered entity (smartrest publish topic), I have added another `MessageSource<(MqttMessage, EntityMetadata)>` impl to the `C8yMapperActor`, but this obviously very hacky.

When moving more things out of CumulocityConverter into other actors, we'll need some organized way to access entity metadata (and entity store) from other actors. First thing that comes to mind is making an `EntityStoreActor` that manages the entity store and allows queries and updates to the store via incoming messages.

But we probably shouldn't allow all actors to modify the entity store freely, and we'll need to make sure what to do when entities are deleted or updated.

## Types of changes

## Paste Link to the issue
- #2880 

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -780,7 +780,6 @@ dependencies = [
  "clock",
  "json-writer",
  "mime",
- "mockito",
  "plugin_sm",
  "proptest",
  "rand",
--- a/crates/core/tedge_mapper/src/c8y/mapper.rs
+++ b/crates/core/tedge_mapper/src/c8y/mapper.rs
@@ -212,17 +212,24 @@ impl TEdgeComponent for CumulocityMapper {
         )?);
 
         let mut c8y_mapper_actor = C8yMapperBuilder::try_new(
-            c8y_mapper_config,
+            c8y_mapper_config.clone(),
             &mut mqtt_actor,
             &mut c8y_http_proxy_actor,
             &mut timer_actor,
-            &mut uploader_actor,
-            &mut downloader_actor,
             &mut fs_watch_actor,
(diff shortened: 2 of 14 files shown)


rina23q:
Honestly speaking, I'm pretty much lost in the design level, what the roles of `OperationHandlerActor` and converter for the operations. What is the benefit to split off the handler to one actor.

From the code, I see still converter is responsible for accepting c8y's operation MQTT messages (JSON over MQTT topic) and converting and publishing the thin-edge command format message (`te/../cmd/<cmd>/<id>`). If parse error occurs, converter publishes the SmartREST c8y's executing and failed messages.
This means that the possible source of SmartREST operation status update messages are distributed in converter and `OperationHandlerActor`.

I was actually expecting that one actor is addressing all the operation related things, including the conversion from c8y JSON over MQTT  / SmartREST (only for custom operations) to thin-edge commands. 

But the core issue is not this level. I believe we need a clear picture, how many components we will have, which component is connected to another components with directions. @albinsuresh made a proposal before, but it was a proposal. We don't have a design conclusion somewhere as reference of what you're doing (remark: "conclusion" not a stone, changeable during implementation, but overview of the refactoring). Without it, it's difficult to "review"...

Bravo555:
This refactor missed the mark and is now outdated, so closing.

PR #3053, opened 2024-08-06
don't process operation status duplicates

Bravo555:
## Proposed changes

If c8y mapper receives operation message with the same status again, don't send smartrest operation status message again.

It's still not clear what should be done if operation is cleared while processing or if some component publishes a previous status (e.g. c8y mapper received operation message with status SUCCESSFUL, but later received operation with the same ID with status EXECUTING).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/operations/handler.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/handler.rs
@@ -1,7 +1,6 @@
 use super::handlers::EntityTarget;
 use super::handlers::OperationContext;
 use super::handlers::OperationMessage;
-use super::handlers::UpdateStatus;
 use crate::actor::IdDownloadRequest;
 use crate::actor::IdDownloadResult;
 use crate::actor::IdUploadRequest;
@@ -11,6 +10,7 @@ use crate::Capabilities;
 use c8y_api::http_proxy::C8yEndPoint;
 use c8y_auth_proxy::url::ProxyUrlGenerator;
 use c8y_http_proxy::handle::C8YHttpProxy;
+use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::sync::Arc;
 use tedge_actors::ClientMessageBox;
@@ -19,8 +19,11 @@ use tedge_api::mqtt_topics::Channel;
 use tedge_api::mqtt_topics::ChannelFilter;
 use tedge_api::mqtt_topics::EntityFilter;
 use tedge_api::mqtt_topics::IdGenerator;
+use tedge_api::workflow::GenericCommandState;
 use tedge_mqtt_ext::MqttMessage;
-use tokio::task::JoinError;
+use tracing::debug;
+use tracing::error;
+use tracing::warn;
 
 /// Handles operations.
 ///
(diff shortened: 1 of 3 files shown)


rina23q:
> It's still not clear what should be done if operation is cleared while processing or if some component publishes a previous status (e.g. c8y mapper received operation message with status SUCCESSFUL, but later received operation with the same ID with status EXECUTING).

Once the operation is marked final state (either SUCCESSFUL or FAILED), SmartREST cannot change the state (doable via REST API though). However, I believe we should not update the operation state in the case as how SmartREST behave. In short, ignore further operation status (ideally logging warning for this case).

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -138,23 +141,70 @@ impl OperationHandler {
...
+                // handling a message? E.g. if we're doing a download to handle SUCCESSFUL status, do
+                // we abort or keep going?
+                // - what if we get a message that moves back status, i.e. SUCCESSFUL -> EXECUTING?
+
+                // if handling clearing message, just join task and remove
+                let Some(status) = status else {
+                    operation
+                        .handle
+                        .await
+                        .expect("operation task should not panic");
```
albinsuresh, line 195: Although not done in this PR, I'm having second thoughts about that `expect` call. Isn't it safer to just log the operation handler failure, if at all one happens, rather than crashing the whole mapper?

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -138,23 +141,70 @@ impl OperationHandler {
...
+                    return;
+                };
+
+                // we got a new status, await we handled the previous one and handle the new one
+                let context = Arc::clone(&self.context);
+                let handle = tokio::spawn(async move {
```
albinsuresh, line 201: I didn't funny understand why you moved to this model where new tasks are spawned for each state transition as opposed to the previous model of single task per operation.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -138,23 +141,70 @@ impl OperationHandler {
...
+                        "already handling operation message with this topic and status, ignoring"
+                    );
+                    return;
+                }
+
+                let (key, operation) = entry.remove_entry();
+
+                // TODO(marcel):
+                //- what happens when clearing message is posted by other client while we're
+                // handling a message? E.g. if we're doing a download to handle SUCCESSFUL status, do
```
albinsuresh, line 186: I'd process it only after the previous state transition is fully processed.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -138,23 +141,70 @@ impl OperationHandler {
...
+                // we abort or keep going?
+                // - what if we get a message that moves back status, i.e. SUCCESSFUL -> EXECUTING?
```
albinsuresh, line 188: That shouldn't be allowed. Log a warning and ignore.

albinsuresh (commented):
Added some queries. Also noticed that the codecov coverage has gone down. Something to look at.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
Bravo555: Because we need to extract status from every message to prevent illegal state transitions, it makes sense to do it once in the handle method of the handler for all operation types. And for that we need to know what is the current state of the operation.
It also achieves separation of concerns: operation handler keeps state and handles the lifecycle of the operation, and `context.update()` is just a function that sends a smartrest response when given an operation message.
Ultimately, we only spawn a task so we that we can eventually generate a smartrest response (possibly after downloads and uploads) for that operation message. The logic to determine if a state transition is valid should be lifted up.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
Bravo555: `context.update()` only panics when it can't send a message to the MQTT publisher via a channel. The MQTT sender it uses is a cloned from C8yMapper, so if this happens we can't send any MQTT messages. This is an unrecoverable failure, so it should be a panic. The actor runtime should restart the entrire actor.

The fact that we only panic on MQTT channel failure should become more clear after it's extracted.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -138,23 +141,70 @@ impl OperationHandler {
             message,
         };
 
-        let topic: Arc<str> = message.message.topic.name.clone().into();
+        let topic = Arc::from(message.message.topic.name.as_str());
 
-        let running_operation = self.running_operations.remove(&topic);
+        let status = match GenericCommandState::from_command_message(&message.message) {
+            Ok(command) if command.status.is_empty() && is_clearing_message(&message.message) => {
```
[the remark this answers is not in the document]
Bravo555: Fixed in 0a185301

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
Bravo555: Already was happening, TODO removed and clarification comment in 0a185301

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
Bravo555: Fixed in 0a185301

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -254,8 +258,15 @@ struct RunningOperation {
...
 
-fn is_clearing_message(message: &MqttMessage) -> bool {
-    message.payload().is_empty() && message.retain && message.qos == QoS::AtLeastOnce
+// TODO: logic of which status transitions are valid should be defined in tedge_api and be
+// considered together with custom statuses of custom workflows
+fn is_operation_status_transition_valid(previous: &str, next: &str) -> bool {
+    #[allow(clippy::match_like_matches_macro)]
+    match (previous, next) {
+        ("successful", "executing") => false,
+        ("failed", "executing") => false,
```
rina23q, line 267: [successful -> failed] and [failed -> successful] cases should be also forbidden. To be precise, any state transitions from the final state (successful/failed) is forbidden. So, something like below is more accurate. (not sure if this is valid Rust, however, hope you understand what I wanted to say)

```rust
        ("successful", _) => false,
        ("failed", _) => false,
```

Bravo555: Fixed in 0ebe371d15

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -206,51 +260,27 @@ impl OperationHandler {
...
-                }
-            }
-        });
+// TODO: logic of which status transitions are valid should be defined in tedge_api and be
+// considered together with custom statuses of custom workflows
+fn is_operation_status_transition_valid(previous: &str, next: &str) -> bool {
+    #[allow(clippy::match_like_matches_macro)]
+    match (previous, next) {
+        // not really a transition but true for simplicity
+        (prev, next) if prev == next => true,
```
albinsuresh, line 274:
```suggestion
        (prev, next) if prev == next => false,
```
Allowing duplicate state transitions is problematic when we use SmartREST. For example, if the mapper processes two `executing` responses back-to-back, the first mapped `501` message will properly change the status of the original request. But if we send another `501` message for the duplicated `executing` response, then it might wrongly change the status of the next operation of the same type, if multiple requests of the same type are queued together on the cloud. This is a side effect of the id-less design of SmartREST that does not target a specific operation id, but always tries to change that status of the operations in the order in which they were triggered. So, we should just skip mapping duplicate state transitions. Fortunately, the check in line no. 174 is already doing that, which bypasses this case here. But, it's still better to keep it consistent here as well.

albinsuresh (commented):
The changes look fine, except for a minor issue regarding duplicate state tranistions.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
Bravo555: Yeah, it depends what we mean by "transition". Here I meant more like an identity transition of the status on the operation itself, not really related to reporting of the transition to Cumulocity. As the TODO says, if it were declared in tedge_api it would be a bit more clear, but here it looks like the logic is duplicated, which I agree with.
Still, for consistency's sake, changed the case to false in 46f1f2a090.

albinsuresh (approved):
LGTM.

PR #3028, opened 2024-07-31
fix: built-in MQTT bridge handling of QoS 0 messages published to the cloud topic

reubenmiller:
## Proposed changes

Messages with packet id 0 should be handled differently as they won't receive a PubAck or PubRec control messages hence it will never be removed from the hashmap and future messages with packet id 0 are ignored.

This should resolve the [last incident](https://github.com/thin-edge/thin-edge.io/issues/2960#issuecomment-2249874311) seen on the OSADL slot 2 device, as the tedge mapper went unresponsive after requesting a file from the device, which most likely invoked a JWT request, which never was answered. Note, that the problem is limited to publishing messages directly to the cloud topic with QoS 0, as publishing a message to the `te/#` topics will be republished to the `c8y/#` topics with QoS > 0 (something automatically done by the bridge, to ensure reliable cloud communication).

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue

* https://github.com/thin-edge/thin-edge.io/issues/2960

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #3059, opened 2024-08-08
fix: increase max packet size for built-in bridge

jarhodes314:
## Proposed changes

- Increase the max packet size of the built-in bridge to the same value as is used by `tedge mqtt pub`
- Document the feature flag for the built-in bridge
- Refactor the logic from #3028 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_mqtt_bridge/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_bridge/src/lib.rs
@@ -57,6 +57,8 @@ use crate::topics::matches_ignore_dollar_prefix;
 use crate::topics::TopicConverter;
 pub use config::*;
 
+const MAX_PACKET_SIZE: usize = 10 * 1024 * 1024;
+
 pub struct MqttBridgeActorBuilder {}
 
 impl MqttBridgeActorBuilder {
@@ -90,6 +92,7 @@ impl MqttBridgeActorBuilder {
         if let Some(tls_config) = local_tls_config {
             local_config.set_transport(Transport::tls_with_config(tls_config.into()));
         }
+        local_config.set_max_packet_size(MAX_PACKET_SIZE, MAX_PACKET_SIZE);
         local_config.set_manual_acks(true);
         local_config.set_last_will(LastWill::new(
             &health_topic.name,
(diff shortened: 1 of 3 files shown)


crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -57,6 +57,8 @@ use crate::topics::matches_ignore_dollar_prefix;
 use crate::topics::TopicConverter;
 pub use config::*;
 
+const MAX_PACKET_SIZE: usize = 10 * 1024 * 1024;
```
albinsuresh, line 60: Looks like we have some inconsistency with packet sizes used across the project.
1. In `mqtt_channel::config::Config`: 1024 * 1024
2. In `bridge::common_mosquitto_config::CommonMosquittoConfig`: 256 * 1024 * 1024 - 1
3. In `tedge::cli::mqtt::MAX_PACKET_SIZE`: 10 * 1024 * 1024

Any reason why you chose the 3rd option and not the second one, keeping it consistent with the existing bridge config?

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -377,28 +381,16 @@ async fn half_bridge(
...
                     match companion_bridge_half.recv().await {
                         // A message was forwarded by the other bridge half, note the packet id
                         Some(Some((topic, msg))) => {
                             loop_breaker.forward_on_topic(topic, &msg);
-                            e.insert(msg);
+                            if pkid != 0 {
+                                // Messages with pkid 0 (meaning QoS=0) should not be added to the hashmap
+                                // as multiple messages with the pkid=0 can be received
+                                e.insert(msg);
+                            }
```
albinsuresh, line 393: Can this `pkid = 0` case be covered with a unit test?

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: I didn't realise that mosquitto used a completely different name for the configuration, so didn't notice it. I'll change this to be consistent with the existing bridge.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: Not easily. In an ideal world, we would have some deterministic tests where we control the broker we connect to directly, so we can send QoS 0 messages. This would also shore up the existing integration tests, which have to flood the bridge with thousands of messages before disconnecting to reproduce the desired behaviour.

didier-wenzek: > Can this `pkid = 0` case be covered with a unit test?

This is tested by the system test "Support publishing QoS 0 messages to c8y topic #2960".

Sure, this test is not deterministic (as messages with QoS 0 might not be delivered), but I think that is enough.

albinsuresh (dismissed):
LGTM

didier-wenzek (approved):
Approved

PR #3069, opened 2024-08-19
config-manager: Only use tedge-write after normal copy fails due to permissions and improve tedge-write logging

Bravo555:
## Proposed changes

- Only use `tedge-write` for privilege elevation after normal copy fails due to `PermissionDenied` error and tedge-write is enabled
- Put more details into warn and error logs when deploying configuration files using `tedge-write`:
    
    ```
    before: "WARN: `sudo.enable` set to `true`, but sudo not found in $PATH"
    after:  "WARN: `sudo.enable` set to `true`, but sudo not found in $PATH, using tedge-write directly"
    
    before: "ERROR: config-manager failed writing updated configuration file: Starting tedge-write process failed"
    after:  "ERROR: config-manager failed writing updated configuration file: failed to start process '/usr/bin/tedge-write': No such file or directory (os error 2)"
    
    before: "ERROR: config-manager failed writing updated configuration file: sudo: /usr/bin/tedge-write: command not found"
    after:  "ERROR: config-manager failed writing updated configuration file: process returned non-zero exit code (1); stderr="sudo: /usr/bin/tedge-write: command not found""
    ```

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/sudo.rs
+++ b/crates/common/tedge_config/src/sudo.rs
@@ -46,6 +46,7 @@ impl SudoCommandBuilder {
     /// prepare a [`Command`](std::process::Command) using sudo. Otherwise,
     /// prepares a Command that starts `program` directly.
     pub fn command<S: AsRef<OsStr>>(&self, program: S) -> Command {
+        let program = program.as_ref();
         if !self.enabled {
             return Command::new(program);
         }
@@ -57,7 +58,7 @@ impl SudoCommandBuilder {
                 c
             }
             Err(_) => {
-                warn!("`sudo.enable` set to `true`, but sudo not found in $PATH");
+                warn!("`sudo.enable` set to `true`, but sudo not found in $PATH, invoking '{}' directly", program.to_string_lossy());
                 Command::new(program)
             }
         }
--- a/crates/core/tedge_write/src/api.rs
+++ b/crates/core/tedge_write/src/api.rs
@@ -33,24 +33,34 @@ impl<'a> CopyOptions<'a> {
     ///
     /// Stdin and Stdout are UTF-8.
     pub fn copy(self) -> anyhow::Result<()> {
-        let output = self
-            .command()?
-            .output()
-            .context("Starting tedge-write process failed")?;
+        let mut command = self.command()?;
+
+        let output = command.output();
+
+        let program = command.get_program().to_string_lossy();
+        let output = output.with_context(|| format!("failed to start process '{program}'"))?;
 
         if !output.status.success() {
-            return Err(anyhow!(
-                String::from_utf8(output.stderr).expect("output should be utf-8")
-            ));
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            let stderr = stderr.trim();
+            let err = match output.status.code() {
+                Some(exit_code) => anyhow!(
+                    "process '{program}' returned non-zero exit code ({exit_code}); stderr=\"{stderr}\""
+                ),
+                None => anyhow!("process '{program}' was terminated; stderr=\"{stderr}\""),
+            };
+
+            return Err(err);
         }
 
         Ok(())
     }
 
-    fn command(&self) -> std::io::Result<Command> {
+    fn command(&self) -> anyhow::Result<Command> {
         let mut command = self.sudo.command(crate::TEDGE_WRITE_PATH);
 
-        let from_reader = std::fs::File::open(self.from)?;
+        let from_reader = std::fs::File::open(self.from)
+            .with_context(|| format!("could not open file for reading '{}'", self.from))?;
         command.stdin(from_reader).arg(self.to);
 
         if let Some(mode) = self.mode {
--- a/crates/extensions/tedge_config_manager/src/actor.rs
+++ b/crates/extensions/tedge_config_manager/src/actor.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use async_trait::async_trait;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
@@ -6,6 +7,10 @@ use log::error;
 use log::info;
 use serde_json::json;
 use std::collections::HashMap;
+use std::io::ErrorKind;
+use std::os::unix::fs::fchown;
+use std::os::unix::fs::MetadataExt;
+use std::os::unix::fs::PermissionsExt;
 use tedge_actors::fan_in_message_type;
 use tedge_actors::Actor;
 use tedge_actors::ChannelError;
@@ -32,6 +37,7 @@ use tedge_uploader_ext::UploadRequest;
 use tedge_uploader_ext::UploadResult;
 use tedge_write::CopyOptions;
 
+use crate::FileEntry;
 use crate::TedgeWriteStatus;
 
 use super::config::PluginConfig;
@@ -379,7 +385,9 @@ impl ConfigManagerActor {
         let response = match result {
             Ok(response) => response,
             Err(err) => {
-                let error_message = format!("config-manager failed downloading a file: {err}",);
+                let err =
+                    anyhow::Error::from(err).context("config-manager failed downloading a file");
+                let error_message = format!("{err:#}");
                 request.failed(&error_message);
                 error!("{}", error_message);
                 self.publish_command_status(ConfigOperation::Update(topic, request))
@@ -401,6 +409,10 @@ impl ConfigManagerActor {
                 error!("{}", error_message);
                 self.publish_command_status(ConfigOperation::Update(topic, request))
                     .await?;
+
+                // TODO: source temporary file should be cleaned up automatically
+                let _ = std::fs::remove_file(from);
+
                 return Ok(());
             }
         };
@@ -413,16 +425,24 @@ impl ConfigManagerActor {
         self.publish_command_status(ConfigOperation::Update(topic, request))
             .await?;
 
+        // TODO: source temporary file should be cleaned up automatically
+        let _ = std::fs::remove_file(from);
+
         Ok(())
     }
 
     /// Deploys the new version of the configuration file and returns the path under which it was
     /// deployed.
     ///
-    /// This function ensures that the configuration file under `dest` is overwritten by a new
-    /// version currently stored in a temporary directory under `src`. Depending on if
-    /// `use_tedge_write` is used, either a new `tedge-write` process is spawned, or a file is
-    /// copied directly.
+    /// Ensures that the configuration file under `dest` is overwritten atomically by a new version
+    /// currently stored in a temporary directory.
+    ///
+    /// If the configuration file doesn't already exist, a new file with target permissions is
+    /// created. If the configuration file already exists, its content is overwritten, but owner and
+    /// mode remains unchanged.
+    ///
+    /// If `use_tedge_write` is enabled, a `tedge-write` process is spawned when privilege elevation
+    /// is required.
     fn deploy_config_file(
         &self,
         from: &Utf8Path,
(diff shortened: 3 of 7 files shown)


crates/common/tedge_config/src/sudo.rs
```diff
@@ -57,7 +57,7 @@ impl SudoCommandBuilder {
                 c
             }
             Err(_) => {
-                warn!("`sudo.enable` set to `true`, but sudo not found in $PATH");
+                warn!("`sudo.enable` set to `true`, but sudo not found in $PATH, using tedge-write directly");
```
didier-wenzek, line 60: `tedge_config::sudo::command()` is not only used by `tedge-write`.

```suggestion
                warn!("`sudo.enable` set to `true`, but sudo not found in $PATH, invoking {program} directly");
```

crates/core/tedge_write/src/api.rs
```diff
@@ -33,24 +33,30 @@ impl<'a> CopyOptions<'a> {
...
-            .context("Starting tedge-write process failed")?;
+        let mut command = self.command()?;
+
+        let output = command.output();
+
+        let program = command.get_program().to_string_lossy();
+        let output = output.with_context(|| format!("failed to start process '{program}'"))?;
 
         if !output.status.success() {
+            let stderr = String::from_utf8(output.stderr).expect("output should be utf-8");
```
albinsuresh, line 44: I'm assuming that we're using `expect` here as we know that the output of `tedge-write` process can't be non-UTF8.

crates/core/tedge_write/src/api.rs
```diff
@@ -33,24 +33,30 @@ impl<'a> CopyOptions<'a> {
...
+            let stderr = stderr.trim();
+            let exit_code = output.status.code().unwrap();
```
albinsuresh, line 46: Here, isn't it better to be defenesive and account for the `None` status code case as well, as there is the theoretical possibility of the `tedge-write` process getting killed?

albinsuresh (approved):
LGTM, except for one minor unwarp issues.

didier-wenzek (commented):
First, as this as been an issue observed by a user, it would be good to have a issue giving the context. Notably, it is not obvious reading this, that the root cause is that configuration management is not working on a device where `/usr/bin` is readonly the path for `tedge_write` being hard coded.

Second, instead of an option to disable `tedge_write`, it would be better to have a setting telling where `tedge_write` has been installed.

didier-wenzek (changes requested):
We need to discuss what is the best option : `tedge_write.disable`or `tedge_write.path`.

crates/core/tedge_write/src/api.rs
Bravo555: Hmm, I think that `tedge-write` process being terminated by a signal is very unlikely, as it is spawned and joined by tedge-agent, but you're right, it's better for that edge case to be properly handled as well.

Bravo555:
During the call with @albinsuresh and @didier-wenzek we decided the following:

1. A ticket with details about original user issue and their environment which required changes for `tedge-write` to be made will be created. It should explain the interaction between `tedge-write` and sudoers which makes it tricky.
2. In the above mentioned user environment, `tedge-agent` is run as root. If `tedge-agent` runs as root, there is no reason for privilege elevation, so `tedge-agent` should never spawn `tedge-write` and just do a normal copy. This way we solve the original issue more elegantly, without adding yet more configuration options.
3. **Ability to customise the path of `tedge-write`**: if a user has a custom environment where they install thin-edge.io in non-standard locations, they should have an ability to customise paths of all thin-edge components, including `tedge-write`. However, for `tedge-write` to work as intended, there needs to be an entry in `/etc/sudoers` with a full path to `tedge-write`, e.g. `/usr/bin/tedge-write` and that full path also needs to be provided when spawning the process. If we allow users to use an alternative location for `tedge-write`, we need to make sure to let the user know they need to change the appropriate `/etc/sudoers` entry for privilege elevation to work.
    Thus, instead of hardcoding the path we can use another flow:
    1. use `which` to find full path of `tedge-write`, e.g. `/opt/thinedge/bin/tedge-write`
    2. spawn a sudo process using that path, i.e. `sudo /opt/thinedge/bin/tedge-write`
    3. Check if we're running as sudo/have necessary write permissions/copying file returns a success. If not, then we **need** to print a hint telling the user to edit their `/etc/sudoers` configuration.

didier-wenzek:
Side remark. It's a pity that the unit tests fail for a reason already highlighted by clippy: unused import: `crate::FileEntry`. We should not run the unit tests with `RUSTFLAGS -D warnings`

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -528,6 +567,92 @@ impl ConfigManagerActor {
     }
 }
 
+/// Writes a file atomically and optionally sets its permissions.
+///
+/// Setting permissions (owner and group) of a file is a privileged operation so it needs to be run
+/// as root. If the any of the filesystem operations fail due to not having permissions, the
+/// function will return an error.
```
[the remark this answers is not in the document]
Bravo555: Addressed in 93c3eba.

tedge-agent will attempt to deploy configuration by copying directly, and then if it fails because of permission denied, will repeat the entire process using tedge-write. Because we're using `tempfile`, if the process fails tempfile will be deleted automatically. I needed to use tempfile directly to set permissions before the final rename. 

1. Tempfile with mode 600 but owned by tedge.
2. The file is already downloaded and resides in the temporary directory (`/tmp`), so this is only a copy, but we try to do it as tedge first.
3. As requested
4. As requested.

As for the `/etc/tedge/tedge.toml`, TODO in store function was fixed, so we can adjust default permissions of the tempfile in `atomically_write_file_*` functions in another PR.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -401,6 +409,10 @@ impl ConfigManagerActor {
                 error!("{}", error_message);
                 self.publish_command_status(ConfigOperation::Update(topic, request))
                     .await?;
+
+                // TODO: source temporary file should be cleaned up automatically
+                let _ = std::fs::remove_file(from);
```
didier-wenzek, line 414: This has definitely to be done.

But not necessarily here. This file has been created and populated outside this method. It should be automatically cleaned by the caller. Not easy to implement with the current design, though. Can this be done once merged [this refactoring](https://github.com/thin-edge/thin-edge.io/pull/3086)

didier-wenzek (approved):
Approved

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -516,6 +548,142 @@ impl ConfigManagerActor {
...
+/// will also change its uid/gid/mode, if writing process euid/egid is different from file's
+/// uid/gid. To keep uid/gid the same, after the write we need to do `chown`, and to do it we need
+/// sudo.
+///
+/// # Errors
+///
+/// - `src` doesn't exist
+/// - user or group doesn't exist
+/// - we have no write/execute permissions to the directory
+fn move_file_atomic_set_permissions_if_doesnt_exist(
```
albinsuresh, line 578:
```suggestion
fn move_file_preserving_target_permissions(
```
Can we also move this API into `tedge_utils` so that it can be reused? After removing the dependency on `FileEntry`, of course. It could take the expected permissions instead.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -516,6 +548,142 @@ impl ConfigManagerActor {
...
+    let mut src_file = std::fs::File::open(src)
+        .with_context(|| format!("failed to open temporary source file '{src}'"))?;
+
+    // TODO: create tests to ensure writes we expect are atomic
+    let mut tempfile = tempfile::Builder::new()
+        .permissions(std::fs::Permissions::from_mode(0o600))
+        .tempfile_in(dest.parent().context("invalid path")?)
+        .with_context(|| format!("could not create temporary file at '{dest}'"))?;
+
+    std::io::copy(&mut src_file, &mut tempfile).context("failed to copy")?;
```
albinsuresh, line 596: The function name implies move but we're doing a copy here. Can we attempt a file move at least when the source and target files are under the same parent directory?

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -516,6 +548,142 @@ impl ConfigManagerActor {
...
+        .as_file()
+        .set_permissions(std::fs::Permissions::from_mode(target_permissions.mode))
+        .context("failed to set mode on the destination file")?;
+
+    fchown(
+        tempfile.as_file(),
+        Some(target_permissions.uid),
+        Some(target_permissions.gid),
+    )
+    .context("failed to change ownership of the destination file")?;
```
albinsuresh, line 608: Could have been done with `tedge_utils::file::PermissionEntry::apply()` as well, right?

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -516,6 +548,142 @@ impl ConfigManagerActor {
...
+    let mode = current_file_permissions
+        .as_ref()
+        .map(|p| p.mode())
+        .or(entry_mode)
+        .unwrap_or(0o644);
+
+    Ok(Permissions { uid, gid, mode })
+}
+
+struct Permissions {
```
albinsuresh, line 681: Why not reuse `tedge_utils::file::PermissionEntry`?

crates/extensions/tedge_config_manager/src/actor.rs
Bravo555: Yes, I have this done for a follow-up PR

crates/extensions/tedge_config_manager/src/actor.rs
Bravo555: Yeah, move is probably a little misleading for now, but this function will be used by both config-manager directly and by tedge-write, which reads from stdin, so the function will be renamed to `write_...` in a follow-up PR

crates/extensions/tedge_config_manager/src/actor.rs
Bravo555: AFAIK `PermissionEntry::apply()` uses a `chown`, but since we have a file descriptor open and want all these operations finished before the final rename, I decided to use `fchown`, which should complete before the file descriptor is closed.

There is also one problem with `PermissionEntry::apply()` when used in this context:

Since we're doing an atomic write as either `tedge` or `root`, the temporary file will be owned by `tedge` or `root` but the target configuration file will likely be owned by some other user, so most likely we will always have to change file owner. We want the atomic write to look like a normal write so it has to preserve original ownership no matter what it is.

But it is technically possible for a file to have uid/gid of a user that does not exist on the system. Perhaps the file was created by a program coming from a package that set up its own user, and the package was since uninstalled and the user was removed. Still in this case we still want to preserve original permissions, so we have to use original uid/gid directly. But `PermissionEntry::apply()` uses a name, so if we wanted to use it here we'd need to look up the name of the owner of the file, which can fail.

So not using `PermissionEntry::apply()` avoids an extra failure mode.

crates/extensions/tedge_config_manager/src/actor.rs
Bravo555: https://github.com/thin-edge/thin-edge.io/pull/3069#discussion_r1736017001

We want to use uid/gid directly instead of names, since doing so avoids an extra failure mode.

gligorisaev (commented):
Checked the test and it is doing the expected verification

Issue #3014, opened 2024-07-24
allow users to add workflow states after a built-in executing action


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When overriding the `config_update` workflow which re-uses the existing thin-edge.io `config_update` functionality and only adds a pre and post state. The `postprocess` state is not currently not possible to add as the built-in `executing` state transitions automatically to `successful` ignoring whatever `on_success` is defined.

An example workflow is show below, where `preprocess` and `postprocess` have been added to perform some custom actions before letting the `tedge-agent` execute the built-in functionality.

```toml
#:schema https://gist.githubusercontent.com/reubenmiller/4e28e8403fe0c54b7461ac7d1d6838c2/raw/4ad7e3bc3ce2e3a7542a5b4873a0196cbe91f35d/tedge.workflow.json

operation = "config_update"
on_error = "failed"

[init]
  action = "proceed"
  on_success = "preprocess"

[preprocess]
  script = "/usr/bin/config_update.sh preprocess ${.payload.type}"
  on_success = "scheduled"
  # Proceed on any unexpected errors to ensure the operations status changed are handled
  on_error = "scheduled"

[scheduled]
  action = "builtin"
  on_success = "executing"

[executing]
  action = "builtin"
  on_success = "postprocess"

[postprocess]
  script = "/usr/bin/config_update.sh postprocess ${.payload.type}"
  on_success = "successful"

[successful]
 action = "cleanup"

[failed]
  action = "cleanup"
```

Below shows the workflow log after an execution, notice how the `postprocess` state is not present in the log (but the `preprocess` state is).

```log
==================================================================
Triggered config_update workflow
==================================================================

topic:     te/device/main///cmd/config_update/c8y-mapper-38494494
operation: config_update
cmd_id:    c8y-mapper-38494494
time:      2024-07-24T13:43:33.22652579Z

==================================================================

----------------------[ config_update @ init | time=2024-07-24T13:43:33.228707959Z ]----------------------

State:    {"logPath":"/var/log/tedge/agent/workflow-config_update-c8y-mapper-38494494.log","remoteUrl":"http://127.0.0.1:8001/c8y/inventory/binaries/38462323","status":"init","type":"custom1"}

Action:   move to preprocess state

=> moving to config_update @ preprocess

----------------------[ config_update @ preprocess | time=2024-07-24T13:43:33.243623851Z ]----------------------

[...]

PR #3070, opened 2024-08-20
feat: workflow builtin actions honoring on_success handlers

didier-wenzek:
## Proposed changes

Improve the workflow builtin actions with explicit operation names and the ability to define specific state on success and on error, as well as for the executing state.

```toml
[scheduled]
  action = "software_update"
  on_exec = "executing"

[executing]
  action = "await_software_update"
  on_success = "postprocess"
  on_error = "rollback"
```

- [x] Distinguish `BuiltIn` from `AwaitBuiltIn` actions
- [x] Make the name of the operation explicit for `BuiltIn` and `AwaitBuiltIn` actions
- [x] Equip the `BuiltIn` action with an `on_exec` handler.
- [x] Equip the `AwaitBuiltIn` action with `on_success` and `on_error` handlers.
- [x] Let the user provides its own steps for the *executing*, *successful* and *failed* state
- [x] Update the documentation - deprecating the  `builtin` keyword
- [ ] Make sure the deprecated  `builtin` keyword can still be used for backward compatibility
- [ ] Add a system test checking that a `builtin:<op>` can be triggered on any state.

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/3014

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/workflow/mod.rs
+++ b/crates/core/tedge_api/src/workflow/mod.rs
@@ -68,7 +68,7 @@ pub enum OperationAction {
     /// on_success = "<state>"
     /// on_error = "<state>"
     /// ```
-    BuiltIn(BgExitHandlers, AwaitHandlers),
+    BuiltIn(ExecHandlers, AwaitHandlers),
 
     /// Trigger a built-in operation
     ///
@@ -76,7 +76,7 @@ pub enum OperationAction {
     /// action = "<builtin-operation-name>"
     /// on_exec = "<state>"
     /// ```
-    BuiltInAction(OperationName, BgExitHandlers),
+    BuiltInAction(OperationName, ExecHandlers),
 
     /// Await the outcome of a built-in operation
     ///
@@ -111,7 +111,7 @@ pub enum OperationAction {
     /// background_script = "sudo systemctl restart tedge-agent"
     /// on_exec = "<state>"
     /// ```
-    BgScript(ShellScript, BgExitHandlers),
+    BgScript(ShellScript, ExecHandlers),
 
     /// Trigger an operation and move to the next state from where the outcome of the operation will be awaited
     ///
@@ -125,7 +125,7 @@ pub enum OperationAction {
         OperationName,
         Option<ShellScript>,
         StateExcerpt,
-        BgExitHandlers,
+        ExecHandlers,
     ),
 
     /// Await the completion of a sub-operation
@@ -237,7 +237,7 @@ impl OperationWorkflow {
     /// Create a built-in operation workflow
     pub fn built_in(operation: OperationType) -> Self {
         let operation_name = operation.to_string();
-        let exec_handler = BgExitHandlers::builtin_default();
+        let exec_handler = ExecHandlers::builtin_default();
         let await_handler = AwaitHandlers::builtin_default();
         let states = [
             ("init", OperationAction::MoveTo("scheduled".into())),
--- a/crates/core/tedge_api/src/workflow/mod.rs
+++ b/crates/core/tedge_api/src/workflow/mod.rs
@@ -417,6 +417,63 @@ impl OperationAction {
 
         Ok(new_state)
     }
+
+    /// Rewrite a command state before pushing it to a builtin operation actor
+    ///
+    /// Depending the action is to trigger or await the operation,
+    /// set the status to schedule or executing.
+    ///
+    /// Return the command state unchanged if there is no appropriate substitute.
+    pub fn adapt_builtin_request(&self, command_state: GenericCommandState) -> GenericCommandState {
+        match self {
+            OperationAction::BuiltInAction(_, _) => {
+                command_state.update(GenericStateUpdate::scheduled())
+            }
+            OperationAction::AwaitBuiltInAction(_, _) => {
+                command_state.update(GenericStateUpdate::executing())
+            }
+            _ => command_state,
+        }
+    }
+
+    /// Rewrite the command state returned by a builtin operation actor
+    ///
+    /// Depending the operation is executing, successful or failed,
+    /// set the new state using the user provided handlers
+    ///
+    /// Return the command state unchanged if there is no appropriate handlers.
+    pub fn adapt_builtin_response(
+        &self,
+        command_state: GenericCommandState,
+    ) -> GenericCommandState {
+        match self {
+            OperationAction::BuiltIn(exec_handlers, _)
+            | OperationAction::BuiltInAction(_, exec_handlers)
+                if command_state.is_executing() =>
+            {
+                command_state.update(exec_handlers.on_exec.clone())
+            }
+            OperationAction::BuiltIn(_, await_handlers)
+            | OperationAction::AwaitBuiltInAction(_, await_handlers)
+                if command_state.is_successful() =>
+            {
+                command_state.update(await_handlers.on_success.clone())
+            }
+            OperationAction::BuiltIn(_, await_handlers)
+            | OperationAction::AwaitBuiltInAction(_, await_handlers)
+                if command_state.is_failed() =>
+            {
+                let mut on_error = await_handlers.on_error.clone();
+                if on_error.reason.is_none() {
+                    if let Some(builtin_reason) = command_state.failure_reason() {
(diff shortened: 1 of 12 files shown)


gligorisaev (commented):
Have checked and runned flake-finder, looks ok

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -57,13 +57,35 @@ pub enum OperationAction {
...
+    /// ```
+    BuiltIn(ExecHandlers, AwaitHandlers),
+
+    /// Trigger a built-in operation
+    ///
+    /// ```toml
+    /// action = "<builtin-operation-name>"
+    /// on_exec = "<state>"
+    /// ```
+    BuiltInAction(OperationName, ExecHandlers),
```
didier-wenzek, line 79: Afterthought, I see that as over engineering.

The motivation was to be explicit:

- With `action = "builtin"` both the operation and the step are given by the context.
   - the operation is the operation of the workflow itself
   - the step is derived from the current state of the command (i.e scheduled or executing)
- With `action = "<builtin-operation-name>"` and `action = "await <builtin-operation-name>" both the operation and the step are explicit
   -  `action = "<builtin-operation-name>"` triggers the scheduled step of the given operation
   - `action = "await <builtin-operation-name>" triggers the execution step

But this raises other issues:
- It might then be confusing to be able to invoke an operation either directly (with `action = software_update`) or using a sub-workflow (with `operation = software_update`). These are meaningful and even useful differences, but this comes with complications notably to guide users.
- What if the user try to invoke the builtin behavior of operation A in a workflow for B? This could work but would requires extra work to handle the command payload, making even greater the confusion between builtin actions and sub operations.
- => In practice, __having the operation implied by the workflow is a good thing__,  simple and effective.

Hence, I'm considering to revert this change. i.e. to only keep the `action = "builtin"` case.

Another appealing alternative is the following:
- Keep the `action = "builtin"` as of now (honoring the user-provided handlers)
- Introduce two slight variants, to make the action clearer as it's not obvious of what is done by the `builtin` action depends on the state name (`scheduled` or `executing`).
- I propose: `action = "execute"` for the `scheduled` step (which can then be named differently).
- and `action = "await"` for the `executing` step.

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -26,6 +26,10 @@ Override Built-In Operation
     Should Contain      ${software_list[0]}    "currentSoftwareList"
     Should Contain      ${software_list[0]}    "mosquitto"
     Should Contain      ${software_list[0]}    "tedge"
+    Should Contain      ${software_list[0]}    "preprocess"
+    Should Contain      ${software_list[0]}    "ready"
```
didier-wenzek, line 30: This check fails despite the preprocess step being actually executed. Indeed, the final payload `'{"currentSoftwareList":...,"postprocess":"done","status":"successful"}'` does not contain the `"preprocess"` key. However, as witnessed by the `"postprocess":"done"`  and `"status":"successful"` values, the command went successfully over all the steps in sequence (init -> scheduled -> preprocess -> executing -> postprocess -> successful).

What is going on? The issue is that the builtin actions remove any unknown field. This has to be fixed.

crates/core/tedge_api/src/workflow/mod.rs
didier-wenzek: Here a new proposal https://github.com/thin-edge/thin-edge.io/issues/3014#issuecomment-2317687695.

I will address it in a different PR so we will be in position to compare and choose.

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -384,6 +417,63 @@ impl OperationAction {
...
+        match self {
+            OperationAction::BuiltInAction(_, _) => {
+                command_state.update(GenericStateUpdate::scheduled())
+            }
+            OperationAction::AwaitBuiltInAction(_, _) => {
+                command_state.update(GenericStateUpdate::executing())
+            }
+            _ => command_state,
+        }
+    }
```
Bravo555, line 437: thought: it's nonobvious how setting a scheduled state leads to an operation being spawned, execution state leads to it being awaited, and whether or not the state being set is for current command or a different one (sub-command)

At the beginning I was a bit confused why `<builtin-operation-name>` moves to scheduled and `await-<builtin-operation-name>` moves to scheduling, but, as the comment says, if idea is for `builtin-*` to trigger an operation and `await-builtin-*` to await the result, then it's not clear to me how this is eventually achieved.

And particularly, when we're doing a sub-workflow execution, e.g. an example from the documentation:

```toml
[trigger_config_update]
operation = "config_update"
input.tedgeUrl = "http://127.0.0.1:8000/tedge/file-transfer/example/config_update/mosquitto-1234"
input.type = "mosquitto"
on_exec = "waiting_for_config_update"

[waiting_for_config_update]
action = "await-operation-completion"
timeout_second = 600
on_timeout = "timeout_config_update"
on_success = "successful_config_update"
on_error = { status = "failed", reason = "fail to update the config"}
```

it's relatively clear that `action = "await-...` below is waiting for the entire config update to complete, but when applied to the lower level of states, I have trouble understanding why that's necessary for the executing state.

Bravo555 (commented):
A bit out of my depth on the design front (I should probably keep track of custom workflow developments more closely), but the Rust side overall looks alright, I just have one question regarding the interaction between `await-...` and executing status.

crates/core/tedge_api/src/workflow/mod.rs
didier-wenzek: > thought: it's nonobvious how setting a scheduled state leads to an operation being spawned, execution state leads to it being awaited, and whether or not the state being set is for current command or a different one (sub-command)

In both case, the workflow engine [delegates](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_agent/src/operation_workflows/message_box.rs#L19) the action to the [actor registered for that builtin operation](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_agent/src/operation_workflows/message_box.rs#L34).

The response is sent back by the operation specific actor to the workflow actor [for further processing](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_agent/src/operation_workflows/actor.rs#L79).
[This commit](https://github.com/thin-edge/thin-edge.io/pull/3070/commits/73dd8fd079ececa9edaa4f99feba9867d4e57c4c) improves method name as they were not really clear,

> It seems like there is an implicit assumption that `scheduling` state is handled by other agents trivially and immediately, so we model it as moving immediately to `executing` state, and we also assume that `executing` is a potentially long-running process that we need to await the completion of, but the question is, is this distinction necessary.

There is indeed an assumption that all the builtin operator actors follow the same workflow, transitioning from scheduled to executing and then either to successful or failed. However, there is no assumption on the timing. Technically, in both cases (scheduled and executing), the workflow engine awaits for the transition to occur.

Is this distinction necessary? From a workflow perspective, yes, as this makes explicit the transitions (todo - doing - done) and clarifies error management. This is notably a key point for the restart operation, where the device reboots between the two states. Internally, for the builtin actors, no. This is a legacy of when these actors where directly exposed to MQTT notifying themselves all the state transitions.

> In any case, I have now a bit better idea of the flow after looking at `adapt_builtin_request` and `adapt_builtin_response` methods:

> ```toml
> [scheduled]
>  action = "software_update"
>  on_exec = "executing"
>
>[executing]
>  action = "await_software_update"
>  on_success = "postprocess"
>  on_error = "rollback"
>```

> 1. workflow is in `scheduled` state, builtin operation actor is notified of that
> 2. builtin operation actor moves to `scheduled` state
> 3. builtin operation actor moves to `executing` state, tells workflow to run `on_exec` handler
> 4. workflow moves to `executing` state, builtin operation actor is notified of that
> 5. builtin operation actor moves to `successful` or `failed` state, tells workflow to run `on_success` or `on_error` handlers
> 6. workflow moes to `successful` or `failed` state
> 7. workflow is terminated

Your understanding is correct.

didier-wenzek:
Closing as superseded by https://github.com/thin-edge/thin-edge.io/pull/3105

Issue #2616, opened 2024-01-23
Use c8y operation ID for operation status update


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
JSON over MQTT input (#1718) enabled to use c8y's operation ID as thin-edge commend ID. Reporting c8y's operation status with operation ID has benefits over ID-less operation status update via SmartREST. If there are concurrent operations with the same operation type, without operation ID may cause an issue, that is, updating unintended operation. 

**Describe the solution you'd like**
However, it's still not clear what is the best solution. We need investigation.

1. JSON over MQTT
As long as I read the [c8y's user guide](https://cumulocity.com/guides/10.18.0/reference/smartrest-two/#supported-endpoint), the endpoint seems not available for updating operations.
2. REST API
We already use REST API to report device's software list. REST API has flexibility, however, some operations require us to know which field name must be used. For example, `c8y_Command` operation requires `text` field to keep the command output. 
3. New SmartREST message with ID
Since c8y team is planning to introduce the new SmartREST messages to update operation with ID, if we wait for long time (years?), such SmartREST will be available.

**Describe alternatives you've considered**

**Additional context**

d-timotijevic:
We've encountered a similar issue in our projects using the thin-edge.io configuration plugin. The lack of operation ID in the SmartREST template for acknowledgement sometimes leads to incorrect operation being acknowledged, especially when there are concurrent operations of the same type.

This issue aligns with the problem described here, where an operation ID-less update via SmartREST could potentially cause the update of an unintended operation. It's evident that a more precise method of updating operation status is required.

Is there any possibility of accelerating the development of this new SmartREST template with operation ID?

Additionally, we've noticed that the nomenclature used in the [Cumulocity documentation](https://cumulocity.com/guides/10.18.0/reference/smartrest-two/#set-operation-to-successful-506) might be a bit confusing. There seems to be a template (506) that sets an operation as successful, including what is termed an 'operation ID,' but in reality, this 'ID' refers to the operation type (c8y_Restart). In other places in the documentation ID is typically understood to refer to a unique attribute, for instance a managed object ID, event ID, etc. 

![image](https://github.com/thin-edge/thin-edge.io/assets/153067759/e0ab9209-b449-4401-957a-cdb945ab2a42)

reubenmiller:
[...]

PR #3076, opened 2024-08-21
feat: add an option to use c8y operation ID to update its status

rina23q:
## Proposed changes
Cumulocity 10.18 version added [504-506](https://cumulocity.com/docs/smartrest/mqtt-static-templates/#504) SmartREST static templates, which enables users to update operation status with its operation ID. This PR adds a new tedge config to switch either 501-503 or 504-506 for operation status updates.

One can change the setting by `tedge config`. `true` (504-506), `false` (501-503). Ideally making `auto` option, however, this is out of scope of this PR.
```sh
tedge config set c8y.smartrest.use_operation_id true
```

Note: for the naming decision, see this [thread](https://github.com/thin-edge/thin-edge.io/pull/3076#discussion_r1732377605).

This change **excludes** the supports for below since they react to SmartREST operation messages, not JSON over MQTT. This means it's not possible to take its operation ID. Also, we don't generate thin-edge commands for these operations.
- custom operation plugin
- legacy c8y-firmware-plugin 

## Types of changes

## Paste Link to the issue
#2616

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -469,6 +469,10 @@ define_tedge_config! {
             /// Set of SmartREST template IDs the device should subscribe to
             #[tedge_config(example = "templateId1,templateId2", default(function = "TemplatesSet::default"))]
             templates: TemplatesSet,
+
+            /// Switch using 501-503 (without operation ID) or 504-506 (with operation ID) SmartREST messages for operation status update
+            #[tedge_config(example = "true", default(value = true))]
+            use_operation_id: bool,
         },
 
 
--- a/crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
+++ b/crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
@@ -1,14 +1,11 @@
 use crate::smartrest::csv::fields_to_csv_string;
 use crate::smartrest::error::SmartRestSerializerError;
-use crate::smartrest::topic::C8yTopic;
 use csv::StringRecord;
-use mqtt_channel::MqttMessage;
 use serde::ser::SerializeSeq;
 use serde::Serialize;
 use serde::Serializer;
 use tedge_api::SoftwareListCommand;
 use tedge_api::SoftwareModule;
-use tedge_config::TopicPrefix;
 use tracing::warn;
 
 pub type SmartRest = String;
@@ -18,36 +15,70 @@ pub fn request_pending_operations() -> &'static str {
 }
 
 /// Generates a SmartREST message to set the provided operation to executing
-pub fn set_operation_executing(operation: impl C8yOperation) -> String {
+pub fn set_operation_executing_with_name(operation: impl C8yOperation) -> String {
     fields_to_csv_string(&["501", operation.name()])
 }
 
+/// Generates a SmartREST message to set the provided operation ID to executing
+pub fn set_operation_executing_with_id(op_id: &str) -> String {
+    fields_to_csv_string(&["504", op_id])
+}
+
 /// Generates a SmartREST message to set the provided operation to failed with the provided reason
-pub fn fail_operation(operation: impl C8yOperation, reason: &str) -> String {
+pub fn fail_operation_with_name(operation: impl C8yOperation, reason: &str) -> String {
+    fail_operation("502", operation.name(), reason)
+}
+
+/// Generates a SmartREST message to set the provided operation ID to failed with the provided reason
+pub fn fail_operation_with_id(op_id: &str, reason: &str) -> String {
+    fail_operation("505", op_id, reason)
+}
+
+fn fail_operation(template_id: &str, operation: &str, reason: &str) -> String {
     // If the failure reason exceeds 500 bytes, trancuate it
     if reason.len() <= 500 {
-        fields_to_csv_string(&["502", operation.name(), reason])
+        fields_to_csv_string(&[template_id, operation, reason])
     } else {
         warn!("Failure reason too long, message trancuated to 500 bytes");
-        fields_to_csv_string(&["502", operation.name(), &reason[..500]])
+        fields_to_csv_string(&[template_id, operation, &reason[..500]])
     }
 }
 
 /// Generates a SmartREST message to set the provided operation to successful without a payload
-pub fn succeed_operation_no_payload(operation: CumulocitySupportedOperations) -> String {
-    succeed_static_operation(operation, None::<&str>)
+pub fn succeed_operation_with_name_no_parameters(
+    operation: CumulocitySupportedOperations,
+) -> String {
+    succeed_static_operation_with_name(operation, None::<&str>)
 }
 
 /// Generates a SmartREST message to set the provided operation to successful with an optional payload
-pub fn succeed_static_operation(
+pub fn succeed_static_operation_with_name(
     operation: CumulocitySupportedOperations,
+    payload: Option<impl AsRef<str>>,
+) -> String {
+    succeed_static_operation("503", operation.name(), payload)
+}
+
+/// Generates a SmartREST message to set the provided operation ID to successful without a payload
+pub fn succeed_operation_with_id_no_parameters(op_id: &str) -> String {
+    succeed_static_operation_with_id(op_id, None::<&str>)
+}
+
+/// Generates a SmartREST message to set the provided operation ID to successful with an optional payload
+pub fn succeed_static_operation_with_id(op_id: &str, payload: Option<impl AsRef<str>>) -> String {
+    succeed_static_operation("506", op_id, payload)
+}
+
+fn succeed_static_operation(
+    template_id: &str,
+    operation: &str,
     payload: Option<impl AsRef<str>>,
 ) -> String {
     let mut wtr = csv::Writer::from_writer(vec![]);
     // Serialization will never fail for text
     match payload {
-        Some(payload) => wtr.serialize(("503", operation.name(), payload.as_ref())),
-        None => wtr.serialize(("503", operation.name())),
+        Some(payload) => wtr.serialize((template_id, operation, payload.as_ref())),
+        None => wtr.serialize((template_id, operation)),
     }
     .unwrap();
     let mut output = wtr.into_inner().unwrap();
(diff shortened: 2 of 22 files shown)


crates/common/tedge_config/src/tedge_config_cli/models/c8y.rs
```diff
@@ -0,0 +1,73 @@
...
+/// A flag that switches using the SmartREST static template 501-503 or 504-506 to update operation status.
+/// Can be set to auto in the future, see #2616.
+#[derive(
+    Debug, Clone, Copy, serde::Serialize, serde::Deserialize, Eq, PartialEq, doku::Document, Display,
+)]
+#[strum(serialize_all = "camelCase")]
+pub enum SmartRestWithOperationId {
+    Yes,
+    No,
+}
```
didier-wenzek, line 44: I would not introduce a new type here and simply use `bool`.

- To be consistent with the other settings (e.g  `tedge config c8y.enable.device_profile true`)
- To be backward compatible with the [`Auto` type](https://github.com/thin-edge/thin-edge.io/blob/main/crates/common/tedge_config/src/tedge_config_cli/models/auto.rs#L8), in case we add support for `auto` in the future.

rina23q: Done in [ad57206](https://github.com/thin-edge/thin-edge.io/pull/3076/commits/ad57206d021cfbde1005a326544727697d781896)

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -469,6 +469,10 @@ define_tedge_config! {
             /// Set of SmartREST template IDs the device should subscribe to
             #[tedge_config(example = "templateId1,templateId2", default(function = "TemplatesSet::default"))]
             templates: TemplatesSet,
+
+            /// Switch using 501-503 (without operation ID) or 504-506 (with operation ID) SmartREST messages for operation status update
+            #[tedge_config(example = "true", default(value = true))]
+            with_operation_id: bool,
```
didier-wenzek, line 475: What will be the behavior if operation ids are enabled on the device but not available on the c8y instance? Would things work? If not how the user will be notified? By error messages or stuck operations?

didier-wenzek (commented):
I will be really happy to merge this PR. I just have one question.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -125,6 +126,33 @@ Workflow log includes plugin output
...
+    Device Should Have Installed Software    c8y-remote-access-plugin
+    Should Have MQTT Messages    c8y/s/us    message_pattern=^(501|502|503),c8y_SoftwareUpdate.*    minimum=1
+
+Operation gets updated regardless of the order of the creation time
+    Stop Service    tedge-agent
+    ${OPERATION1}=    Install Software        non-existent-package1
+    ${OPERATION2}=    Install Software        non-existent-package2
+    ${op1dict}=    Operation Should Be PENDING    ${OPERATION1}
+    ${op2dict}=    Operation Should Be PENDING    ${OPERATION2}
+    ${op_id1}=    Get From Dictionary    ${op1dict}    id
```
gligorisaev, line 144: The Keyword Get From Dictionary is expecting the optional value. If optional default value is given, it will be returned instead of failing.
so the step should look like:

${op_id1}=    Get From Dictionary    ${op1dict}    id    default

the step is passing even if this is not defined, difference is that the keyword is expecting it and underlined red as notcomplete/faulty

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -125,6 +126,33 @@ Workflow log includes plugin output
...
+    ${op_id2}=    Get From Dictionary    ${op2dict}    id
```
gligorisaev, line 145: The Keyword Get From Dictionary is expecting the optional value. If optional default value is given, it will be returned instead of failing.
so the step should look like:

${op_id1}= Get From Dictionary ${op12dict} id default

the step is passing even if this is not defined, difference is that the keyword is expecting it and underlined red as notcomplete/faulty

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
albinsuresh, line 475:
```suggestion
            direct_operation_update: bool,
```
Not fully happy about this suggestion either. Accept only if you're convinced.

albinsuresh (approved):
LGTM.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
rina23q: If `op_id` is not found here, it is something wrong. The test should not proceed further. So it should be good as it is.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
rina23q: see below my comment

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
rina23q: > What will be the behavior if operation ids are enabled on the device but not available on the c8y instance? Would things work?

No, it won't work (that's we want to improve with `auto` option)

> If not how the user will be notified? By error messages or stuck operations?

I can't really test it since all tenants available for us is already 10.18 or above. User will definitely receive an error message on `c8y/s/e` for unknown static template like below.  

```
[c8y/s/us] 900
[c8y/s/e] 40,900,No static template for this message id
```

In this case, operation will be executed and successful/failed in thin-edge, but the status will stay `PENDING` in c8y.

rina23q: The key will look `c8y.smartrest.direct_operation_update`. I don't think what "direct" means is so clear. At least, not for me. How about something in between my proposal and yours, `c8y.smartrest.operation_id_update`?

Bravo555: My bikeshed take would be something like `c8y.smartrest.operation_explicit_id` or `operation_template_explicit_id`

didier-wenzek: I see no reason to change anything here: `with_operation_id` is simple and clear.

didier-wenzek (approved):
Approved

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
rina23q: I vote +1 for `use_operation_id`

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -18,36 +15,70 @@ pub fn request_pending_operations() -> &'static str {
...
+pub fn fail_operation_with_name(operation: impl C8yOperation, reason: &str) -> String {
+    fail_operation("502", operation.name(), reason)
+}
+
+/// Generates a SmartREST message to set the provided operation ID to failed with the provided reason
+pub fn fail_operation_with_id(op_id: &str, reason: &str) -> String {
+    fail_operation("505", op_id, reason)
+}
+
+fn fail_operation(number: &str, operation: &str, reason: &str) -> String {
```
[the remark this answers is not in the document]
rina23q: Addressed all your renaming suggestion in this commit: 2e159505b321d0f1ef26b81fc909fb4fc3c8c43a

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
rina23q: > What's the plan to implement auto detection? Can we probe for errors on c8y/s/e when sending 504-506 messages and then fallback to the previous API?

Definitely it's one of the doable options. We have similar issue for Advanced Software Management (requirement: 10.14> and dedicated Microservice). Want to have a consistent solution for this and Advanced Software Management. Probably we need to discuss Device Management Team to choose the best solution.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -803,6 +803,14 @@ impl IdGenerator {
     pub fn is_generator_of(&self, cmd_id: &str) -> bool {
         cmd_id.contains(&self.prefix)
     }
+
+    /// Extract c8y's operation ID from the provided thin-edge command ID.
+    pub fn get_c8y_operation_id(&self, cmd_id: &str) -> Option<String> {
```
Bravo555, line 808: issue(blocking): this won't always be c8y operation id, consider this test where it would return `1111`:

https://github.com/thin-edge/thin-edge.io/blob/e48ffd7518720aaef774fb9fc086714d6eca45a1/tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot#L115-L124

As it's only relevant to c8y mapper, it should probably be moved there and we should make sure that it's run only on commands spawned by the mapper that indeed contain c8y operation id.

Bravo555 (changes requested):

crates/core/tedge_api/src/mqtt_topics.rs
Bravo555: thought: overall I think we're starting to put a lot of information as prefixes or suffixes into command ids:

- if they're run as part of sub-workflow (`sub:` prefix)
- what c8y operation id they're spawned in response to (`c8y_mapper_ext/src/converter.rs:628`)
- the mapper that spawned them (`prefix` field in `IdGenerator`)

And this is sort of in between two approaches:
- opaque identifiers that are created by some thing and then handled only by the same thing
- structured identifiers that contain multiple fields

I think 1st approach is what we should target and this current mix a result of c8y operations being decoded and commands being spawned in two totally different places, so it's hard to save this state except in a command id.

As in the 2nd paragraph of [this comment](https://github.com/thin-edge/thin-edge.io/pull/3032#issuecomment-2299191772) the situation should improve once handling incoming c8y operation messages is moved to a dedicated operation handling actor. Now I have bit better picture for this refactoring.

rina23q: I'll change this function to generic and move the check for the number to c8y-mapper.

```rust
    pub fn get_value(&self, cmd_id: &str) -> Option<String> {
        cmd_id
            .strip_prefix(&format!("{}-", self.prefix))
            .map(|s| s.to_string())
    }
```

Bravo555: Then one more improvement would be to avoid allocations and just return a slice:

```rust
pub fn get_value<'a>(&self, cmd_id: &'a str) -> Option<&'a str> {
    cmd_id
        .strip_prefix(&self.prefix)
        .and_then(|s| s.strip_prefix('-'))
}
```

crates/extensions/c8y_firmware_manager/src/worker.rs
```diff
@@ -371,10 +373,8 @@ impl FirmwareManagerWorker {
 
         let c8y_child_topic = C8yTopic::ChildSmartRestResponse(child_id.to_string())
             .to_topic(&self.config.c8y_prefix)?;
-        let executing_msg = MqttMessage::new(
-            &c8y_child_topic,
-            DownloadFirmwareStatusMessage::status_executing(),
-        );
+        let payload = set_operation_executing_with_name(CumulocitySupportedOperations::C8yFirmware);
```
Bravo555, line 376: thought: wasn't aware that firmware manager doesn't go through the mapper but publishes c8y messages directly
Are there any plans to have it handled by the mapper or does the firmware manager benefit from having a direct line to c8y without going through the mapper?

crates/core/tedge_api/src/mqtt_topics.rs
rina23q: Done in https://github.com/thin-edge/thin-edge.io/pull/3076/commits/0136b9ebe68a38f28380210b0cbb53e34f69beed

crates/extensions/c8y_firmware_manager/src/worker.rs
rina23q: This `c8y_firmware_manager` is used only for the legacy `c8y_firmware_plugin`. All legacy operation plugins have direct MQTT connection without mapper in order to delivery the feature faster.

Bravo555 (approved):

crates/extensions/c8y_firmware_manager/src/worker.rs
rina23q: Technically, we don't have to update such legacy code. However, I did here because I wanted to get rid of `OperationStatusMessage`. To avoid someone will use the duplicated old API to update operation status.

PR #3081, opened 2024-08-23
fix issues in atomically_write_file_* functions

Bravo555:
## Proposed changes

- no longer fail when `.tmp` file already exists, new tempfile are created with 6 random bytes and names are regenerated if such name already exists
- ensure async tempfile is cleaned up by wrapping it in a `tempfile::NamedTempFile` that provides a drop implementation

`tempfile` was also bumped because `tempfile::Builder::permissions` was not available on the current version.

Running `configuration_operation.robot` suite, it was flaking for me, and I have found the following error that caused `tedge-agent` to never start:

```
ERROR sm-agent: tedge_actors::runtime: Aborted due to Writing the
content to the file "/etc/tedge/.agent/software-current-operation"
failed: "could not create the temporary file \"/etc/tedge/.agent/software-current-operation.tmp\"".
source=Os { code: 17, kind: AlreadyExists, message: "File exists" }`
```

If a `software-current-operation` already had a temporary file, the atomic copy would fail. As it used async version of the atomic write function, which was trying to delete tempfiles manually on errors, it was nevertheless possible for this function to be terminated while running if runtime finished, leading to tempfile being not cleaned up on Drop.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/fs.rs
+++ b/crates/common/tedge_utils/src/fs.rs
@@ -2,9 +2,9 @@ use nix::NixPath;
 use std::fs as std_fs;
 use std::io::Read;
 use std::io::Write;
+use std::os::unix::fs::PermissionsExt;
 use std::path::Path;
 use std::path::PathBuf;
-
 use tokio::fs as tokio_fs;
 use tokio::io::AsyncWriteExt;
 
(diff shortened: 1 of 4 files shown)


crates/common/tedge_utils/src/fs.rs
```diff
@@ -40,40 +40,36 @@ impl<T, E: Into<std::io::Error>> ErrContext<T> for Result<T, E> {
...
+        .rand_bytes(6)
+        // FIXME: setting tempfile permissions to 644 is required not to break config
+        //
+        // normally tempfile creates a temporary file with mode 600, which is good for tmpfiles but
+        // this function is used for saving /etc/tedge/tedge.toml, which doesnt explicitly set a
+        // mode, so it assumes file created by this function be with default umask and that default
+        // umask is 022; but in the test suite `bootstrap.sh` is run as root, which makes
+        // /etc/tedge/tedge.toml owned by root, but later when `tedge-mapper-c8y` runs as `tedge` it
+        // can no longer read the config file, rendering thin-edge completely unusable. the fix is
+        // not trivial, because we can't set permissions on the config calling side if
```
albinsuresh, line 61: I didn't get that. Why couldn't the config file be created with the desired mode (644)?

didier-wenzek: Same for me. I don't understand the comment.

didier-wenzek (commented):
I will be happy to merge this PR. I just need a better understanding of why the default tempfile permissions break tedge config.

crates/common/tedge_utils/src/fs.rs
Bravo555: The comment was prompted by 2 things:

1. this `TODO` comment:
    https://github.com/thin-edge/thin-edge.io/blob/a9f2fcfcd211230cf926521db5f2802beab1fcce/crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs#L145-L156
    When in the test suite, `bootstrap.sh` is run as root, then this function is run as root and `/etc/tedge/tedge.toml` ends up being owned by root, so it needs `a+r` permissions to be readable. As the `TODO` says, we should probably instead set ownership and mode in this function. In e090dc0 this is fixed by setting permissions and warning on errors. Is this acceptable?
2. Other code that calls these functions implicitly assumed that resulting destination files would have permissions 644, but had the umask been different the permissions would be different and this assumption would be violated.
    Now the function manually sets 644 permissions and additional doc comment was written to clarify what happens with file permissions (e090dc0).

Perhaps I initially overestimated how hard it would be to handle setting the permissions in the `store` function because tedge-agent is run as either root or tedge-agent, but we can just ignore the error. As such `FIXME` comment was removed.

didier-wenzek (approved):
Approved

PR #3091, opened 2024-08-29
Extract atomic write function into a module

Bravo555:
## Proposed changes

Make tedge-write and config-manager use the same atomic write function

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/atomic.rs
+++ b/crates/common/tedge_utils/src/atomic.rs
@@ -1,0 +1,125 @@
+//! Utilities for atomic writes.
+//!
+//! For deployment of configuration files, we need to create a file atomically because certain
+//! programs might watch configuration file for changes, so if it's not written atomically, then
+//! file might be only partially written and a program trying to read it may crash.
+//!
+//! Atomic write of a file consists of creating a temporary file in the same directory, filling it
+//! with correct content and permissions, and only then renaming the temporary into the destination
+//! filename. Because we're never actually writing into the file, we don't need to write permissions
+//! for the destination file, even if it exists. Instead we need only write/execute permissions to
+//! the directory file is located in unless the directory has a sticky bit set. Overwriting a file
+//! will also change its uid/gid/mode, if writing process euid/egid is different from file's
+//! uid/gid. To keep uid/gid the same, after the write we need to do `chown`, and to do it we need
+//! sudo.
+
+use std::io::ErrorKind;
+use std::io::Read;
+use std::os::unix::fs::fchown;
+use std::os::unix::fs::MetadataExt;
+use std::os::unix::fs::PermissionsExt;
+use std::path::Path;
+
+use anyhow::Context;
+
+/// Writes a file atomically and optionally sets its permissions.
+///
+/// Setting ownership of a file is a privileged operation so it needs to be run as root. If any of
+/// the filesystem operations fail due to not having permissions, the function will return an error.
+///
+/// If the file already exists, its content will be overwritten but its permissions will remain
+/// unchanged.
+pub fn write_file_atomic_set_permissions_if_doesnt_exist(
+    mut src: impl Read,
+    dest: impl AsRef<Path>,
+    permissions: &MaybePermissions,
+) -> anyhow::Result<()> {
+    let dest = dest.as_ref();
+
+    let target_permissions = target_permissions(dest, permissions)
+        .context("failed to compute target permissions of the file")?;
+
+    // TODO: create tests to ensure writes we expect are atomic
+    let mut tempfile = tempfile::Builder::new()
+        .permissions(std::fs::Permissions::from_mode(0o600))
+        .tempfile_in(dest.parent().context("invalid path")?)
+        .with_context(|| {
+            format!(
+                "could not create temporary file at '{}'",
+                dest.to_string_lossy()
+            )
+        })?;
+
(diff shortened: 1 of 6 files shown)


crates/common/tedge_utils/src/atomic.rs
```diff
@@ -0,0 +1,125 @@
...
+use anyhow::Context;
+
+/// Writes a file atomically and optionally sets its permissions.
+///
+/// Setting ownership of a file is a privileged operation so it needs to be run as root. If any of
+/// the filesystem operations fail due to not having permissions, the function will return an error.
+///
+/// If the file already exists, its content will be overwritten but its permissions will remain
+/// unchanged.
+pub fn write_file_atomic_set_permissions_if_doesnt_exist(
```
didier-wenzek, line 32: There is already a variant of this new function: [`fs::atomically_write_file_sync`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/common/tedge_utils/src/fs.rs#L47).

Can one be written using the other?

Bravo555: The function in `fs` does not allow setting file ownership and permissions before renaming into a destination, and it has some other callers that I didn't want to edit as that's probably fine for them.

The biggest difference is that this function will attempt to `chown` the file before renaming it, which if it fails the file will not be persisted. This is necessary behaviour for configuration management, where we don't want the file to be persisted if we don't have permissions to `chown` (doing so would destroy original permissions on the file) so that we can retry with tedge-write.

By contrast callers of `fs::atomically_write_file_sync` only read and write files created by `tedge` so they don't need to chown.

Indeed it would be good to unify these atomic write functions as it may be confusing why there are multiple versions of them currently, but we use them in many different contexts and for now I wanted to ensure `config-manager` and `tedge-write` are doing the same thing.

didier-wenzek: > for now I wanted to ensure `config-manager` and `tedge-write` are doing the same thing.

I fully appreciate that.

> By contrast callers of `fs::atomically_write_file_sync` only read and write files created by `tedge` so they don't need to chown.

Sure, but `atomic::write_file_atomic_set_permissions_if_doesnt_exist()` takes a `MaybePermissions` argument which could be slightly modified to hold the case where `chown` is not required.

Where I understand your point of view, is that `fs::atomically_write_file_sync` is used in many critical places. Better to be cautious. By the way, this is because the `Trigger Agent Restart` system test is failing that I started to wonder if there were any interactions with  `fs::atomically_write_file_sync`  which is used before rebooting the device. For device restart, it's critical that the data is properly flushed to disk: this is a key difference between `fs::atomically_write_file_sync` and `atomic::write_file_atomic_set_permissions_if_doesnt_exist`.

Bravo555: > For device restart, it's critical that the data is properly flushed to disk: this is a key difference between fs::atomically_write_file_sync and atomic::write_file_atomic_set_permissions_if_doesnt_exist.

`atomic::write_file_atomic_set_permissions_if_doesnt_exist` does a `sync_all()` call on the file, which is supposed to flush all the in-memory data to the filesystem before renaming, so I'm not sure what's the difference between the two in that respect.

albinsuresh, line 32:
```suggestion
pub fn write_file_atomic_with_permissions_if_new(
```

crates/core/tedge_write/src/bin.rs
```diff
@@ -62,133 +56,34 @@ pub fn run(args: Args) -> anyhow::Result<()> {
...
-                Some(
-                    uzers::get_group_by_name(g)
-                        .with_context(|| format!("Group `{g}` does not exist"))?
-                        .gid(),
-                )
-            }
-        }
-        None => None,
-    };
+    let src = std::io::stdin().lock();
```
albinsuresh, line 79: Not a remark on this PR, but a query on the initial design itself. Why was this plugin designed this way to read the input file content via stdin? If you could just share a link to some discussions on this point on the older issue/PR, that'd suffice.

albinsuresh (approved):
LGTM.

crates/common/tedge_utils/src/atomic.rs
didier-wenzek: > `atomic::write_file_atomic_set_permissions_if_doesnt_exist` does a `sync_all()` call on the file,

Indeed, I failed to notice that. Then, I so no reason preventing to merge the two functions.

There are only two small differences:
- `fs::atomically_write_file_sync` calls `flush_all()` on the directory to persist file renaming (as this is not done by [`tempfile::persist`](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#method.persist)).
- `fs::atomically_write_file_sync` doesn't change file ownership.

crates/core/tedge_write/src/bin.rs
Bravo555: https://github.com/thin-edge/thin-edge.io/issues/2456

There was a requirement that tedge-write should not be able to read files that are not globally readable, e.g. private keys. This is naturally achieved when the calling side needs to open the file to pipe it into stdin.

crates/common/tedge_utils/src/atomic.rs
didier-wenzek: I don't think a function name has to tell the whole story about that function ;-)

Issue #3073, opened 2024-08-21
configuration update fails due to tedge-write when thin-edge.io is installed in a non-standard location


Bravo555:
This issue happened on a custom user environment where thin-edge.io was installed on a non-standard location due to the standard `/usr/bin` directory being non-writable, as such thin-edge was installed at `/data/tedge`. This is not achievable via normal installation methods and required a lot of tweaking, so it's not technically a bug. Nonetheless, we should allow users to customise their thin-edge.io installations if their environment requires it.

**Describe the bug**

When thin-edge.io is installed in a non-standard location and `/usr/bin/tedge-write` is not present, configuration update operation doesn't work:

```
2024-08-19T10:45:36.87635586Z ERROR tedge_config_manager::actor: config-manager failed writing updated configuration file: Starting tedge-write process failed
```

This happens because currently `ConfigManagerActor` always tries to use `tedge-write` when deploying configuration files.

tedge-write (#2456) is a mechanism to elevate tedge-agent's permissions to write to configuration files that tedge-agent doesn't normally have write permissions for because they're not owned by `tedge`, but perhaps e.g. `root` or `mosquitto`. tedge-agent spawns a `sudo tedge-write SRC DST --user USER --group GROUP --mode MODE` process, the contents of the source file are written into a destination file, and `sudo` allows `tedge-write` to write to a file that's normally not writable by `tedge`. For sudo to run `tedge-write` as root, there needs to be an entry in `/etc/sudoers`. By default the installation script generates the following entry which allows user `tedge` to execute `/usr/bin/tedge-write` as root without password when 1st argument matches `/etc/*`:

```
tedge    ALL = (ALL) NOPASSWD: /usr/bin/tedge-write /etc/*
```

`sudo` matches its command line arguments with the pattern in the `sudoers` entry, so for that reason it is crucial that a full path `/usr/bin/tedge-write` is used, or `sudo` won't run `tedge-write` as root. For the same reason the destination file needs to be the first argument, despite `tedge-write` command line parser allowing arguments to be provided in any order.

Because full path needs to be used, it is tricky to customise the path of `tedge-write` because the full path in `sudoers` entry needs to stay in sync as well. If a user changes location of `tedge-write`, they need to update their sudoers entry.

**To Reproduce**

1. Change the path of `tedge-write` to another location that is in `tedge-agent` process' `$PATH`, e.g.:
    ```
    mv /usr/bin/tedge-write /data/bin/tedge-write
    ```
[...]

PR #3099, opened 2024-09-03
fix: tedge write remove hardcoded path

Bravo555:
## Proposed changes

API for spawning a tedge-write process no longer uses a hardcoded `/usr/bin/tedge-write` path.

The path was hardcoded because of the wrong assumption of how `sudo` works: when provided with a command without the full path, e.g. `sudo tedge-write /etc/config1`, `sudo` will internally look for `tedge-write` in $PATH and then compare path with its sudoers entry, so it's not necessary to use full path for binaries used with `sudo`.

## Types of changes

## Paste Link to the issue

- #3073

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/sudo.rs
+++ b/crates/common/tedge_config/src/sudo.rs
@@ -54,6 +54,8 @@ impl SudoCommandBuilder {
         match which::which(self.sudo_program.as_ref()) {
             Ok(sudo) => {
                 let mut c = Command::new(sudo);
+                // non-interactive
+                c.arg("-n");
                 c.arg(program);
                 c
             }
--- a/crates/core/tedge_write/src/api.rs
+++ b/crates/core/tedge_write/src/api.rs
@@ -6,6 +6,8 @@ use camino::Utf8Path;
 use std::process::Command;
 use tedge_config::SudoCommandBuilder;
 
+use crate::TEDGE_WRITE_BINARY;
+
 /// Options for copying files using a `tedge-write` process.
 #[derive(Debug, PartialEq)]
 pub struct CopyOptions<'a> {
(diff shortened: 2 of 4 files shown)


crates/common/tedge_config/src/sudo.rs
```diff
@@ -54,6 +54,7 @@ impl SudoCommandBuilder {
         match which::which(self.sudo_program.as_ref()) {
             Ok(sudo) => {
                 let mut c = Command::new(sudo);
+                c.arg("--non-interactive");
```
reubenmiller, line 57: I'd recommend using the short form option `-n` here (even though it is less readable) because `-n` is more cross compatible, e.g. [freebsd](https://man.freebsd.org/cgi/man.cgi?query=sudo&apropos=0&sektion=0&manpath=FreeBSD+8.1-RELEASE+and+Ports&format=html) only has the `-n` option and not the long form.

didier-wenzek (approved):
Approved

gligorisaev (commented):
Have checked the added Test step and keyword, working as expected

crates/common/tedge_config/src/sudo.rs
Bravo555: fixed

albinsuresh (approved):
So, we're doing an explicit `which` lookup from the `tedge-agent` process to find `tedge-write` in its customized `$PATH` because `sudo`, by default, ignores the `$PATH` env of the parent process, right?

LGTM and I don't want to block this PR but I'm just wondering if running all sudo commands issued by `tedge-agent` with the `--preserve-env` option would be a better fix, as that'll make sure that not just `tedge-write` but any other process spawned by `tedge-agent`, like the software management plugins as well, would benefit from the same fix. I'm only concerned if it's some sort of "bad security practice" invoking `sudo` preserving the calling process's env.

crates/core/tedge_write/src/api.rs
```diff
@@ -57,7 +59,11 @@ impl<'a> CopyOptions<'a> {
     }
 
     fn command(&self) -> anyhow::Result<Command> {
-        let mut command = self.sudo.command(crate::TEDGE_WRITE_PATH);
+        // if tedge-write is in PATH of tedge process, use it, if not, defer PATH lookup to sudo
+        let tedge_write_binary =
+            which::which_global(TEDGE_WRITE_BINARY).unwrap_or(TEDGE_WRITE_BINARY.into());
+
+        let mut command = self.sudo.command(tedge_write_binary);
```
didier-wenzek, line 66: Oops, I have not been fast enough to answer @albinsuresh question before this PR being merged by @Bravo555 !

> I'm just wondering if running all sudo commands issued by `tedge-agent` with the `--preserve-env` option would be a better fix, as that'll make sure that not just `tedge-write` but any other process spawned by `tedge-agent`, like the software management plugins as well, would benefit from the same fix. I'm only concerned if it's some sort of "bad security practice" invoking `sudo` preserving the calling process's env.

One simple way,  to have all the components benefiting from the same fix, is to move the call to `which::which_global()`  *inside* the method `sudo.command()`.

Issue #1794, opened 2023-03-07
Allow the configuration directory to be specified globally via environment variables


jarhodes314:
**Is your feature improvement request related to a problem? Please describe.**
It's currently only possible to specify a custom config directory using a command line argument to `tedge`/`tedge-mapper`/`tedge-agent`. This means the core functionality of tedge can be configured using a `tedge.toml` located anywhere, but this breaks with plugins like the remote access plugin. Since the remote access plugin is a separate process, it has no clue about arguments passed to the mapper so it blindly assumes the configuration is stored in `/etc/tedge`. 

**Describe the solution you'd like**
`tedge_config` could determine the location of the config dir using an environment variable (e.g. `TEDGE_CONFIG_DIR`). This can then be passed from the mapper or agent to child processes. If the default behaviour of `TEdgeConfigRepository` is to read from `$TEDGE_CONFIG_DIR` (or `/etc/tedge` if that's not set), the remote access plugin (and other affected Rust-based plugins) would work without any changes (other than rebuilding them).

The interface to e.g. `tedge-mapper` doesn't necessarily need to change. It could have the configuration directory specified by a command line argument as it is now and then create the relevant environment variable from this when spawning a plugin process.

**Describe alternatives you've considered**
We could support this configuration as something a child process can query via MQTT/HTTP, but this would probably require deliberate support from the plugins and would be a lot more complicated. This also adds an extra requirement to plugins to be able to connect to tedge.

**Additional context**
If we use a `TEDGE_` prefixed environment variable for this, we will need to ignore it when reading the standard configuration which could be done [here](https://github.com/thin-edge/thin-edge.io/blob/2ee912f34d80641b3af3a5d4f47eb36b0b73e81d/crates/common/tedge_config/src/tedge_config_cli/figment.rs#L175) using [`Env::ignore`](https://docs.rs/figment/latest/figment/providers/struct.Env.html#method.ignore) from figment.

reubenmiller:
Yes this would be very helpful as it makes it easier to control the custom components when using a custom config-dir location, and removes the need to use the `--config-dir` option when starting any of the components.

With the change the following should be possible:

```sh
export TEDGE_CONFIG_DIR=$HOME/tedge
tedge init
tedge-agent &
tedge mqtt sub '#'
```

reubenmiller:
Implemented as described, the config-dir can now be controlled by setting the environment variable, `TEDGE_CONFIG_DIR`.

[...]

PR #3113, opened 2024-09-06
feat: support setting config dir via env variable

reubenmiller:
## Proposed changes

Support setting the tedge `--config-dir` flag via an environment variable, `TEDGE_CONFIG_DIR`

This makes it much easier for users which are installing thin-edge.io in custom locations, as they can just set an environment variable and all of the binaries will respect the config dir path without having to add `--config-dir` everywhere.

**Implementation Note:** Clap's `.env(key)` function as it does not ignore empty environment variables, e.g. if an empty env variable is defined`export TEDGE_CONFIG_DIR=''`, then clap's logic will set the value to an empty string instead of falling back to the default value. Hence why the `default_value` logic was used instead, but that has a downside that the doc string needs to be manually provided (since the .env and .default_value functions are not used).

**Note:** @jarhodes314 Already add the `TEDGE_CONFIG_DIR` exception in the configuration parsing logic in 716e330 (`ignore(&["CONFIG_DIR"])`), so that the variable does not register as an "unknown configuration property".

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/1794

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
@@ -13,13 +13,28 @@ use crate::TEdgeConfigReader;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use serde::Serialize;
+use std::path::PathBuf;
 use tedge_utils::file::PermissionEntry;
 use tedge_utils::fs::atomically_write_file_sync;
 use tracing::debug;
 use tracing::warn;
 
-pub const DEFAULT_TEDGE_CONFIG_PATH: &str = "/etc/tedge";
+const DEFAULT_TEDGE_CONFIG_PATH: &str = "/etc/tedge";
+const ENV_TEDGE_CONFIG_DIR: &str = "TEDGE_CONFIG_DIR";
 const TEDGE_CONFIG_FILE: &str = "tedge.toml";
+
+/// Get the location of the configuration directory
+///
+/// Check if the TEDGE_CONFIG_DIR env variable is set and only
+/// use the value if it is not empty, otherwise use the default
+/// location, /etc/tedge
+pub fn get_config_dir() -> PathBuf {
+    match std::env::var(ENV_TEDGE_CONFIG_DIR) {
+        Ok(s) if !s.is_empty() => PathBuf::from(s),
+        _ => PathBuf::from(DEFAULT_TEDGE_CONFIG_PATH),
+    }
+}
+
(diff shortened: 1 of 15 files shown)


crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
```diff
@@ -13,13 +13,30 @@ use crate::TEdgeConfigReader;
...
+/// Check if the TEDGE_CONFIG_DIR env variable is set and only
+/// use the value if it is not empty, otherwise use the default
+/// location, /etc/tedge
+pub fn get_config_dir() -> PathBuf {
+    let maybe_string = std::env::var(ENV_TEDGE_CONFIG_DIR)
+        .unwrap_or_else(|_| DEFAULT_TEDGE_CONFIG_PATH.to_string());
+    match Some(maybe_string) {
+        Some(s) if !s.is_empty() => PathBuf::from(s),
+        _ => PathBuf::from(DEFAULT_TEDGE_CONFIG_PATH),
+    }
```
didier-wenzek, line 37:
```suggestion
    match std::env::var(ENV_TEDGE_CONFIG_DIR) {
        Ok(s) if !s.is_empty() => PathBuf::from(s),
        _ => PathBuf::from(DEFAULT_TEDGE_CONFIG_PATH),
    }
```

crates/core/tedge/src/cli/mod.rs
```diff
@@ -40,7 +40,14 @@ pub enum TEdgeOptMulticall {
         #[clap(subcommand)]
         cmd: TEdgeOpt,
 
-        #[clap(long = "config-dir", default_value = DEFAULT_TEDGE_CONFIG_PATH, global = true)]
+        // [env: TEDGE_CONFIG_DIR, default: /etc/tedge]
```
didier-wenzek, line 43: You have to use a doc comment, otherwise `clap` doesn't use the comment for `--help`.

```suggestion
        /// [env: TEDGE_CONFIG_DIR, default: /etc/tedge]
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
didier-wenzek: fixed by https://github.com/thin-edge/thin-edge.io/pull/3113/commits/ca7ee0508e8d89eb82f06dccd4cff38f4f1bea59

didier-wenzek (approved):
Approved

Issue #3083, opened 2024-08-26
mosquitto stops to persisting messages when rust bridge disconnects


sagIoTPower:
Hi,
 
we have the behaviour that mosquitto stops to persist messages when the connection tedge_mqtt_bridge to the cloud is interrupted: (see screenshot tedge_mqtt_bridge_disconnect):
 
`2024-08-23T06:42:30.112741403Z ERROR tedge_mqtt_bridge::health: MQTT bridge failed to connect to cloud broker: Mqtt state: Last pingreg isn't acked`
 
The persistence-store continues to grow, as expected, until: `Time 6:48:35 AM 08/23/2024`.

See screenshots ![persistence_store_grow_start_github](https://github.com/user-attachments/assets/aac18804-bddb-4155-927e-ac2c52725686) and
![persistence_store_grow_end_github](https://github.com/user-attachments/assets/81eec3e5-0ddf-44c7-8039-7b8e76c5e49e)

But then is stops to grow. Which is strange.
 
We use the rust bridge and thin edge 1.2

This error does not happen, when we switch to the mosquitto bridge.
Is it a bug or do we have to configure something?
 
When the connection is finally restored the before persisted messages are successfully transmitted. But these are the old messages, messages received after persistence_store_grow_end discarded.

sagIoTPower:
Hi,

we set `max_queued_messages` and `max_queued_bytes` to 0 ( no limitation).

With the Mosquitto Bridge, persistence worked perfectly. After about a quarter of an hour and about 300 KB of persisted data, we stopped the test.

![Bild](https://github.com/user-attachments/assets/aeb6d061-96c6-4a98-923e-a7d7fdd2001e)
 
With the Rust Bridge, persistence stopped after approx. 5 minutes and approx. 215 KB of persisted data. At the same time, we received the following output in the MQTT Broker log
 
       `Client tedge-mapper-bridge-c8y has exceeded timeout, disconnecting.`
 
The persisted messages are not rolling, i.e. when reconnecting, only the data directly after the disconnect is sent and not the data shortly before the reconnect.

didier-wenzek:
There are two independent issues here:

- If the too many messages are queued by mosquitto for the builtin bridge during a network interruption,
  then messages are silently dropped by mosquitto and definitely lost.
  - An entry is added in the mosquitto log telling that `Outgoing messages are being dropped for client tedge-mapper-bridge-c8y`.
[...]

PR #3122, opened 2024-09-12
fix: builtin bridge disconnected when under heavy load

didier-wenzek:
## Proposed changes

When a large amount of messages has to be published from the device to the cloud (say after a network outage),
the builtin bridge fails to cope with the load and reaches a point where no more progress is made. After a while, the local  MQTT broker closes the builtin bridge connection, with an error:

```
1725524939: Client tedge-mapper-bridge-c8y has exceeded timeout, disconnecting.
```

The root cause is a dead lock inside the builtin bridge. The two main tasks (one publishing local messages to the remote MQTT endpoint, the other forwarding the remote message acknowledgements to the local broker) are blocked waiting for the other to make progress.

What makes the issue not so easy to fix is that the MQTT library used by thin-edge (`rumqttc`) provides no way to reduce the pressure. When a message is received from the local MQTT connection and delivered to the builtin bridge, then the bridge can only acknowledge the message (and this should be done only when properly published and acknowledged by the cloud endpoint) or drop it (by pretending it has been processed). The bridge has no way to tell `rumqttc` that a message cannot be processed right now. Until the message is properly acknowledged, the broker keeps sending the same message again and again to `rumqttc` but those retries are not forwarded to the bridge.

So, what is proposed here is hold in memory all the messages until they are properly processed (received from local, published to cloud, acknowledged by cloud, acknowledged to local). This is done using `mpsc::unbounded()`.

Notes:
- If the mapper is shutdown with pending messages, all those will be resent by the MQTT broker.
- Not only a unbounded channel has been introduced, but also a new background task is spawned.
  This is to ensure that the rumqtt channel and the half-bridge companion channel are in sync.  

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3083

## Checklist

### DIFF
--- a/crates/extensions/tedge_mqtt_bridge/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_bridge/src/lib.rs
@@ -130,18 +130,19 @@ impl MqttBridgeActorBuilder {
             .map(|t| SubscribeFilter::new(t.to_owned(), QoS::AtLeastOnce))
             .collect();
 
-        let [msgs_local, msgs_cloud] =
+        let [cloud_target, local_target] =
             bidirectional_channel(cloud_client.clone(), local_client.clone(), in_flight.into());
         let [(convert_local, bidir_local), (convert_cloud, bidir_cloud)] =
             rules.converters_and_bidirectional_topic_filters();
-        let (tx_status, monitor) = BridgeHealthMonitor::new(health_topic.name.clone(), &msgs_cloud);
+        let (tx_status, monitor) =
+            BridgeHealthMonitor::new(health_topic.name.clone(), &local_target);
         tokio::spawn(monitor.monitor());
         tokio::spawn(half_bridge(
             local_event_loop,
             local_client.clone(),
+            cloud_target,
             convert_local,
             bidir_local,
-            msgs_local,
             tx_status.clone(),
             "local",
             local_topics,
@@ -150,9 +151,9 @@ impl MqttBridgeActorBuilder {
         tokio::spawn(half_bridge(
             cloud_event_loop,
             cloud_client.clone(),
+            local_target,
             convert_cloud,
             bidir_cloud,
-            msgs_cloud,
             tx_status.clone(),
             "cloud",
             cloud_topics,
@@ -171,12 +172,12 @@ fn bidirectional_channel(
     cloud_client: AsyncClient,
     local_client: AsyncClient,
     buffer: usize,
-) -> [BidirectionalChannelHalf; 2] {
+) -> [BridgeAsyncClient; 2] {
     let (tx_first, rx_first) = mpsc::channel(buffer);
     let (tx_second, rx_second) = mpsc::channel(buffer);
     [
-        BidirectionalChannelHalf::new(cloud_client, tx_first, rx_second),
-        BidirectionalChannelHalf::new(local_client, tx_second, rx_first),
+        BridgeAsyncClient::new(cloud_client, tx_first, rx_second),
+        BridgeAsyncClient::new(local_client, tx_second, rx_first),
     ]
 }
 
@@ -200,11 +201,18 @@ enum BridgeMessage {
     Pub { publish: Publish },
 }
 
-struct BidirectionalChannelHalf {
+/// Wraps the target of an half bridge with a channel to its half bridge companion.
+///
+/// So when a message is received and published by this half,
+/// the companion will await for that message to be acknowledged by the target
+/// before acknowledging to the source.
+struct BridgeAsyncClient {
     /// MQTT target for the messages
     target: AsyncClient,
+
     /// Receives messages from the companion half bridge
     rx: mpsc::Receiver<Option<(String, Publish)>>,
+
     /// Sends to a background task that forwards the messages to the target and companion
     ///
     /// (None, message) => {
@@ -219,7 +227,7 @@ struct BidirectionalChannelHalf {
     unbounded_tx: mpsc::UnboundedSender<BridgeMessage>,
 }
 
-impl BidirectionalChannelHalf {
+impl BridgeAsyncClient {
     pub fn recv(&mut self) -> futures::stream::Next<mpsc::Receiver<Option<(String, Publish)>>> {
         self.rx.next()
     }
@@ -234,7 +242,7 @@ impl BidirectionalChannelHalf {
         rx: mpsc::Receiver<Option<(String, Publish)>>,
     ) -> Self {
         let (unbounded_tx, unbounded_rx) = mpsc::unbounded();
-        let companion_bridge_half = BidirectionalChannelHalf {
+        let companion_bridge_half = BridgeAsyncClient {
             target,
             rx,
             unbounded_tx,
@@ -401,9 +409,9 @@ impl BidirectionalChannelHalf {
 async fn half_bridge(
     mut recv_event_loop: EventLoop,
     recv_client: AsyncClient,
+    mut target: BridgeAsyncClient,
     transformer: TopicConverter,
     bidirectional_topic_filters: Vec<Cow<'static, str>>,
-    mut companion_bridge_half: BidirectionalChannelHalf,
     tx_health: mpsc::Sender<(&'static str, Status)>,
     name: &'static str,
     topics: Vec<SubscribeFilter>,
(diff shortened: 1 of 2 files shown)


crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -349,21 +401,15 @@ async fn half_bridge(
...
-                                publish.qos,
-                                publish.retain,
-                                publish.payload.clone(),
-                            )
-                            .await
-                            .unwrap();
+                        let duplicate = (topic.to_string(), publish.clone());
                         companion_bridge_half
-                            .send(Some((topic.into_owned(), publish)))
+                            .send((topic.to_string(), publish), Some(duplicate))
```
jarhodes314, line 409: It's a bit confusing seeing us send the same data twice to the same method. Now the bidirectional channel is no longer so general, maybe we should replace `send` with two methods, `forward_message` and `publish_novel_message` (or something along those lines). This is especially useful in this case as the reasoning behind the weird API is only apparent when looking at the healthcheck code, not the main loop.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -162,33 +172,63 @@ fn bidirectional_channel<T>(buffer: usize) -> [BidirectionalChannelHalf<T>; 2] {
...
-        BidirectionalChannelHalf {
-            tx: tx_second,
-            rx: rx_first,
-        },
+        BidirectionalChannelHalf::new(tx_first, rx_second),
+        BidirectionalChannelHalf::new(tx_second, rx_first),
     ]
 }
 
 struct BidirectionalChannelHalf<T> {
```
jarhodes314, line 180: Following looking at the way we now send, it feels like `BidirectionalChannelHalf` is a bit too generic. If you used it in the obvious way, just calling `send` and `recv`, it wouldn't do anything. It could do with renaming to reflect this change.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -162,33 +172,63 @@ fn bidirectional_channel<T>(buffer: usize) -> [BidirectionalChannelHalf<T>; 2] {
...
 impl<'a, T> BidirectionalChannelHalf<T> {
-    pub fn send(&'a mut self, item: T) -> futures::sink::Send<'a, mpsc::Sender<T>, T> {
-        self.tx.send(item)
+    fn new(tx: mpsc::Sender<Option<T>>, rx: mpsc::Receiver<Option<T>>) -> Self {
+        let (unbounded_tx, unbounded_rx) = mpsc::unbounded::<(T, Option<T>)>();
+        BidirectionalChannelHalf {
+            tx,
+            rx,
+            unbounded_tx,
+            unbounded_rx: Some(unbounded_rx),
```
jarhodes314, line 198: This feels like a bit of a hack. If `new` instead took the `AsyncClient` and spawned the publisher loop, we wouldn't need this `Option` (and we can remove an argument in the function to spawn the bridge halves).

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -400,9 +453,23 @@ async fn half_bridge(
...
+                info!("Bridge {name} connection still waiting ack for pkid={pkid}");
+            }
+
+            Event::Incoming(Incoming::Disconnect) => {
+                info!("Bridge {name} connection closed by peer");
+            }
+
+            Event::Outgoing(Outgoing::PingReq) => {
+                let waiting = forward_pkid_to_received_msg.len();
+                info!("Bridge {name} connection: received={received} forwarded={forwarded} published={published} waiting={waiting} acknowledged={acknowledged} finalized={finalized} ignored={ignored}");
```
jarhodes314, line 471: Is this meant to be at INFO level?

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -336,10 +385,13 @@ async fn half_bridge(
             }
         };
         debug!("Received notification ({name}) {notification:?}");
+        debug!("Bridge {name} connection: received={received} forwarded={forwarded} published={published} waiting={} acknowledged={acknowledged} finalized={finalized} ignored={ignored}",
```
jarhodes314, line 388: This is a really nice way to observe what's going on without resorting to logging every event and trying to manually reconstruct what's going on.

jarhodes314 (commented):
Given how problematic this issue seemed at first, this is a surprisingly elegant solution. I've left some comments about improving the bidirectional channel API now it's less general-purpose.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: I can only agree that this is a really weird API!

Now that things are working, I will give a try to your proposal. An alternative is to send more specific structs.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: It is. Except that I wonder if this appropriate to log this progress every minute. Every hours might be useful.

My only concern here is that this is not logged when the builtin bridge is blocked! i.e. when this would be the more insightful :-(

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: Suggestion applied: https://github.com/thin-edge/thin-edge.io/pull/3122/commits/5783f15acfbf034e51a15af51082c5aca25c8c9c

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: The duplicate is now created underneath by the publisher loop when required.

https://github.com/thin-edge/thin-edge.io/pull/3122/commits/3b2235f4d6ac4b369855047721b36b998b86b4eb

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: I will remove this not so usefull info! message.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -373,14 +443,21 @@ async fn half_bridge(
                 Incoming::PubAck(PubAck { pkid: ack_pkid })
                 | Incoming::PubRec(PubRec { pkid: ack_pkid }),
             ) => {
+                acknowledged += 1;
                 if let Some(msg) = forward_pkid_to_received_msg.remove(&ack_pkid) {
-                    let target = target.clone();
-                    tokio::spawn(async move { target.ack(&msg).await.unwrap() });
+                    if let Err(err) = target.ack(&msg).await {
```
jarhodes314, line 448: Can `ack` not block in the same way `publish` does?

didier-wenzek: I previously observed blocked `ack` when the channel cycle was only made of bounded channels. Now that one of the channel of this cycle is unbounded, I don't think `ack` might block. However, to be sure, I'm missing a better understanding of `rumttc` internals.

That said, now that [BidirectionalChannelHalf owns the target](https://github.com/thin-edge/thin-edge.io/pull/3122/commits/5783f15acfbf034e51a15af51082c5aca25c8c9c), we can do the same for `ack` as for `pub`.
=> I will do that.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: Removed: https://github.com/thin-edge/thin-edge.io/pull/3122/commits/4cadd6dd24a5ece70b21a84dca9a0cef681c110e

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -351,19 +506,12 @@ async fn half_bridge(
...
-                        companion_bridge_half
-                            .send(Some((topic.into_owned(), publish)))
-                            .await
-                            .unwrap();
+                        received += 1;
+                        target.publish(topic.to_string(), publish).await;
+                    } else {
+                        // Being not forwarded to this bridge target
+                        // The message has to be acknowledged
+                        recv_client.ack(&publish).await.unwrap()
```
jarhodes314, line 514: Good point. This shouldn't ever be the case, but it is the correct way to handle it. We should be able to convert any incoming message as the subscription is based on the mapping rules.

jarhodes314 (approved):
This looks really good, and the new API now makes things largely simpler. The improved debug logging is also really useful. :)

PR #3125, opened 2024-09-20
Adding Benchmark section to our documentation page

gligorisaev:
## Proposed changes

This PR provides performance benchmarks for ThinEdge.io, including detailed CPU and memory usage for key processes. The metrics are  measured on a device *rackfslot1* running in the OSADL QA Farm.

- Included are also the Munin plugin files which are located on the OSADL QA Farm at `/etc/munin/plugins/`

- The `tedgecpuprocent` plugin is up and running, the `tedgemonitorcpu` is still in draft and under work because I am trying to trigger the needed condition to raise an alarm, commiting it here only because somebody can maybe see some problem, improvement, etc. to the code.

- Prepared script for downloading the needed .svg files which are showing the graphs on the documents page.

- Included is also the workflow file which should trigger automatic download of the files once per month on month beggining 

**TODO**
- [ ] Decision needed for the final location of the downloading script
- [ ] adding the munin username and password to GIT Secrets
- [ ] Decision needed on which repo the workflow will be located and automaticaly trigerred 
- [ ] Decision needed what should be the workload we want to measure and show in the benchmark

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


scripts/download_svgs.py
didier-wenzek, line 1: This file has to be under `ci/build_scripts` and not under yet another top level directory.

docs/src/benchmarks/index.md
```diff
@@ -0,0 +1,101 @@
+---
+title: Benchmarks
+sidebar_position: 10
+---
+
+# Benchmarks
```
didier-wenzek, line 6: A key point is missing: what is the workload used for these benchmarks?

docs/src/benchmarks/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+In this section, we monitor the memory consumption of key system components to ensure efficient performance. The memory plugin tracks several metrics to provide a detailed view of how the system memory is being utilized.
+
+### Graph of Memory Usage
+
+Below is a visual representation of the memory consumption on the device over the past month.
+
+![Memory Usage](./memory-month.svg)
+
+### Monitored Metrics:
```
didier-wenzek, line 66: I don't see these metrics as useful for giving an idea of the memory footprint of thin-edge,  since there are not directly related.

scripts/download_svgs.py
gligorisaev: The script was at that location, but if I understand right @reubenmiller, he proposed not to be there so thats why I have it in the To Do section as needed to be decided

docs/src/benchmarks/index.md
gligorisaev: Agree fully, I put it in the ToDo section as decision needed what we want to be the workload

scripts/download_svgs.py
reubenmiller: Though I'm pretty sure we can reduce some dependencies (e.g. `requests` which is not available by default in python) and just use a simple shell script which just depends on `wget`...

```sh
wget --user xxxx --password xxxx https://munin.osadl.org/munin/osadl.org/rackfslot1.osadl.org/tedgecpuprocent-month.svg
wget --user xxxx --password xxxx https://munin.osadl.org/munin/osadl.org/rackfslot1.osadl.org/memory-month.svg
```

docs/src/benchmarks/index.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Memory Usage
+
+In this section, we monitor the memory consumption of key system components to ensure efficient performance. The memory plugin tracks several metrics to provide a detailed view of how the system memory is being utilized.
+
+### Graph of Memory Usage
+
+Below is a visual representation of the memory consumption on the device over the past month.
+
+![Memory Usage](./memory-month.svg)
```
reubenmiller, line 64: The memory script should only show the thin-edge.io applications and not the whole OS memory usage etc.

docs/src/benchmarks/index.md
reubenmiller: Yes, there are other metrics which just look at the memory for each of the thin-edge.io related memory usage, e.g. https://munin.osadl.org/munin/osadl.org/rackfslot1.osadl.org/tedgemem.html

scripts/download_svgs.py
gligorisaev: @reubenmiller  you are right, I changed the downloading script according you comment

PR #3126, opened 2024-09-20
refactor: add support to `define_tedge_config` for (optionally) multi-value fields

jarhodes314:
## Proposed changes
In order to add support multiple Cumulocity connections, we need the data model defined for `tedge.toml` to support multiple cloud connections. This PR adds support in `define_tedge_config!` for a `#[tedge_config(multi)]` attribute, which allows the field to store either the underlying struct directly or a hashmap of strings to structs (which provides us with multiple named configurations for a single key).

More concretely, this will allow `tedge_config` to support configurations like:
```toml
[c8y.edge]
url = "https://edge.example.com"
proxy.bind.port = 8001
bridge.topic_prefix = "c8y-edge"

[c8y.cloud]
url = "https://cloud.example.com"
proxy.bind.port = 8002
bridge.topic_prefix = "c8y-cloud"

[mqtt]
bridge.built_in = true
```

while still also supporting as it does currently:

```toml
[c8y]
url = "https://c8y.example.com"
```

This PR doesn't make any changes to the existing `tedge_config` definitions and therefore doesn't change the behaviour of thin-edge, hence the refactoring classification.

Still TODO:
- [x] Update key parsing logic
- [x] Update `tedge config list` logic
- [x] Rename unused fields in `match` blocks to `_`
- [x] Other TODOs left in code

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/common/tedge_config_macros/examples/multi.rs
```diff
@@ -0,0 +1,66 @@
+use camino::Utf8PathBuf;
+// use std::net::IpAddr;
```
jarhodes314, line 2: I need to remove this commented out code.

This will also become a more fleshed-out example in the future, this currently just checks that the new `multi` attribute doesn't cause compilation errors.

crates/common/tedge_config_macros/examples/multi.rs
```diff
@@ -0,0 +1,66 @@
...
+define_tedge_config! {
+    #[tedge_config(multi)]
+    c8y: {
+        #[tedge_config(reader(private))]
+        url: String,
+    },
+}
+
+fn main() {
+    // let dto = TEdgeConfigDto::default();
```
jarhodes314, line 48: Same as above, I should replace this with something that actually demonstrates the functionality.

crates/common/tedge_config_macros/impl/src/dto.rs
```diff
@@ -46,6 +46,21 @@ pub fn generate(
                     });
                 }
             }
+            FieldOrGroup::Multi(group) => {
+                if !group.dto.skip {
+                    let sub_dto_name = prefixed_type_name(&name, group);
+                    // TODO check if skip serializing if is actually any use, I deleted it because it was complicated to implement
```
jarhodes314, line 52: Just delete this, it's done now

crates/common/tedge_config_macros/impl/src/query.rs
```diff
@@ -292,42 +351,72 @@ fn keys_enum(
     }
 }
 
+#[derive(Debug, Default)]
+struct NumericIdGenerator {
```
jarhodes314, line 355:
```suggestion
struct SequentialIdGenerator {
```

crates/common/tedge_config_macros/impl/src/query.rs
```diff
@@ -292,42 +351,72 @@ fn keys_enum(
...
+    count: u32,
+}
+
+pub trait IdGenerator: Default {
+    fn next_id(&mut self, span: Span) -> syn::Ident;
+}
+
+impl IdGenerator for NumericIdGenerator {
```
jarhodes314, line 363: I will likely adopt this approach more widely, since the PR currently has a bunch of ways of assigning these in various places.

I should also document what this is doing.

crates/common/tedge_config_macros/impl/src/query.rs
```diff
@@ -346,8 +435,31 @@ fn generate_string_writers(paths: &[VecDeque<&FieldOrGroup>]) -> TokenStream {
...
+            }).collect::<Vec<_>>();
+            let mut id_gen = NumericIdGenerator::default();
+            let write_segments = path.iter().map(|&thing| {
+                let ident = thing.ident();
+                match thing {
+                    FieldOrGroup::Field(_) => quote!(#ident),
+                    FieldOrGroup::Group(_) => quote!(#ident),
+                    FieldOrGroup::Multi(_) => {
+                        let field = id_gen.next_id(ident.span());
+                        quote_spanned!(ident.span()=> #ident.get_mut(#field.as_deref())?)
```
jarhodes314, line 459: This `get_mut` is the only difference between the two cases. This duplication should be removed

crates/common/tedge_config_macros/impl/src/query.rs
```diff
@@ -401,29 +513,31 @@ fn generate_string_writers(paths: &[VecDeque<&FieldOrGroup>]) -> TokenStream {
...
-            pub fn try_update_str(&mut self, key: WritableKey, value: &str) -> Result<(), WriteError> {
+            pub fn try_update_str(&mut self, key: &WritableKey, value: &str) -> Result<(), WriteError> {
                 match key {
                     #(#update_arms)*
                 };
                 Ok(())
             }
 
-            pub fn unset_key(&mut self, key: WritableKey) {
+            pub fn try_unset_key(&mut self, key: &WritableKey) -> Result<(), WriteError> {
```
jarhodes314, line 526: This can now fail as `WritableKey` is no longer guaranteed to point to a key that definitely exists (when named `c8y`s are enabled).

crates/common/tedge_config_macros/impl/src/reader.rs
```diff
@@ -67,12 +71,38 @@ fn generate_structs(
                     false => parse_quote!(pub),
                 });
             }
+            // TODO skipping?
+            FieldOrGroup::Multi(group) => {
+                // TODO is this right?
```
jarhodes314, line 76: Probably, it seems to compile

crates/common/tedge_config_macros/impl/src/reader.rs
```diff
@@ -67,12 +71,38 @@ fn generate_structs(
...
+                let sub_reader_name = prefixed_type_name(name, group);
+                idents.push(&group.ident);
+                tys.push(parse_quote_spanned!(group.ident.span()=> ::tedge_config_macros::Multi<#sub_reader_name>));
+                let mut parents = parents.clone();
+                parents.push(PathItem::Static(group.ident.clone()));
```
jarhodes314, line 81: I should remove `PathItem` as it's a less flexible alternative to what we are using `IdGenerator` for earlier on.

crates/common/tedge_config_macros/src/multi.rs
```diff
@@ -0,0 +1,66 @@
...
+    SingleNotMulti,
+    #[error("You need a name for this field")]
+    MultiNotSingle,
+    #[error("You need a name for this field")]
+    MultiKeyNotFound,
+}
+
+impl<T> Multi<T> {
+    // TODO rename this to something more rusty
+    pub fn get(&self, key: Option<&str>) -> Result<&T, MultiError> {
```
jarhodes314, line 34:
```suggestion
    pub fn try_get(&self, key: Option<&str>) -> Result<&T, MultiError> { 
```

crates/common/tedge_config_macros/src/multi.rs
```diff
@@ -0,0 +1,66 @@
...
+        match (self, key) {
+            (Self::Single(val), None) => Ok(val),
+            (Self::Multi(map), Some(key)) => map.get(key).ok_or(MultiError::MultiKeyNotFound),
+            (Self::Multi(_), None) => Err(MultiError::SingleNotMulti),
+            (Self::Single(_), Some(_key)) => Err(MultiError::MultiNotSingle),
+        }
+    }
+
+    pub fn get_mut(&mut self, key: Option<&str>) -> Result<&mut T, MultiError> {
```
jarhodes314, line 43:
```suggestion
    pub fn try_get_mut(&mut self, key: Option<&str>) -> Result<&mut T, MultiError> {
```

crates/common/tedge_config_macros/impl/src/reader.rs
jarhodes314: Didn't remove it, but have used `SequentialIdGenerator` to do the ID generation.

didier-wenzek (commented):
> This PR doesn't make any changes to the existing `tedge_config` definitions and therefore doesn't change the behaviour of thin-edge, hence the refactoring classification.

It makes sense. However, testing the multi feature is not easy.

I tried to add the following in `tedge_config.rs`:

```rust
    #[tedge_config(multi)]
    foo: {
        /// Endpoint URL of foo IoT tenant
        #[tedge_config(example = "foo.net")]
        url: ConnectUrl,

        /// Set of MQTT topics the foo IoT mapper should subscribe to
        #[tedge_config(example = "te/+/+/+/+/a/+,te/+/+/+/+/m/+,te/+/+/+/+/e/+")]
        #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,te/+/+/+/+/status/health"))]
        topics: TemplatesSet,
    },
```

But got a weird error I failed to solve:

```
the trait `From<tedge_config_macros::MultiError>` is not implemented for `ReadError`, which is required by `Result<std::string::String, ReadError>: FromResidual<Result<Infallible, tedge_config_macros::MultiError>>`
```

jarhodes314:
> > This PR doesn't make any changes to the existing `tedge_config` definitions and therefore doesn't change the behaviour of thin-edge, hence the refactoring classification.
> 
> It makes sense. However, testing the multi feature is not easy.
> 
> I tried to add the following in `tedge_config.rs`:
> 
> ```rust
>     #[tedge_config(multi)]
>     foo: {
>         /// Endpoint URL of foo IoT tenant
>         #[tedge_config(example = "foo.net")]
>         url: ConnectUrl,
> 
>         /// Set of MQTT topics the foo IoT mapper should subscribe to
>         #[tedge_config(example = "te/+/+/+/+/a/+,te/+/+/+/+/m/+,te/+/+/+/+/e/+")]
>         #[tedge_config(default(value = "te/+/+/+/+/m/+,te/+/+/+/+/e/+,te/+/+/+/+/a/+,te/+/+/+/+/status/health"))]
>         topics: TemplatesSet,
>     },
> ```
> 
> But got a weird error I failed to solve:
> 
> ```
> the trait `From<tedge_config_macros::MultiError>` is not implemented for `ReadError`, which is required by `Result<std::string::String, ReadError>: FromResidual<Result<Infallible, tedge_config_macros::MultiError>>`
> ```

The macro requires that the `ReadError` defined by your crate/example needs to implement both `From<ConfigNotSet>` but also `From<tedge_config_macros::MultiError>`. The example `multi.rs` includes this.

Cargo.toml
```diff
@@ -107,6 +107,8 @@ pem = "1.0"
 pin-project = { version = "1.1.3", features = [] }
 plugin_sm = { path = "crates/core/plugin_sm" }
 predicates = "2.1"
+pretty_assertions = "1.4.1"
```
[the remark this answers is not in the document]
jarhodes314: Yeah, it would be essentially impossible to assert eq for large blocks of code without a line-by-line comparison.

didier-wenzek:
> The macro requires that the `ReadError` defined by your crate/example needs to implement both `From<ConfigNotSet>` but also `From<tedge_config_macros::MultiError>`. The example `multi.rs` includes this.

That's work. Thanks.

crates/common/tedge_config_macros/impl/src/namegen.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Debug, Default)]
+pub struct SequentialIdGenerator {
+    pub count: u32,
+}
+
+impl SequentialIdGenerator {
+    pub fn replay(&self, span: Span) -> syn::Ident {
+        let i = self.count - 1;
+        syn::Ident::new(&format!("key{i}"), span)
+    }
```
didier-wenzek, line 65: This seems a bit tricky to use correctly.

- Why not simply cloning the `ident` returned by the former call to `next_id()`?

crates/common/tedge_config_macros/impl/src/reader.rs
```diff
@@ -336,13 +427,37 @@ fn generate_conversions(
...
+                    .chain(once(&new_arg))
+                    .filter_map(|path_item| match path_item {
+                        PathItem::Static(_) => None,
+                        PathItem::Dynamic(span) => Some(id_gen.next_id(*span)),
+                    })
+                    .collect();
+                let mut parents = parents.clone();
+                parents.push(PathItem::Static(group.ident.clone()));
+                let read_path = read_field(&parents);
+                let new_arg2 = id_gen.replay(group.ident.span());
```
didier-wenzek, line 453: Why not the following:

```suggestion
                let new_arg2 = extra_call_args.get(extra_call_args.len() -1).unwrap().clone();
```

This makes explicit the assumptions hidden by the call to `replay`. I.e. that at least one id must have been generated and that we are looking for the last one.

didier-wenzek (commented):
I tested get/set/add with this [test config](https://github.com/thin-edge/thin-edge.io/pull/3126#pullrequestreview-2336693880). This is working as expected and I will be happy to see this PR merged.

However, we need to improve the error messages.

```
$ tedge config get foo.url       
Error: failed to get the configuration value for key: 'foo.url'

Caused by:
    You need a name for this field
```

Ideally, I would expect something along the lines of "A dimension is required for this property as in foo.<dimension>.url".

Similarly: 

```
$ tedge config get foo.unset_dim.url
Error: failed to get the configuration value for key: 'foo.unset_dim.url'

Caused by:
    Key not found in multi-value group
```

Ideally, I would expect something like "Unknown dimension : foo.unset_dim".

crates/common/tedge_config_macros/src/multi.rs
```diff
@@ -0,0 +1,149 @@
...
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum MultiError {
+    #[error("You are trying to access a named field, but the fields are not named")]
+    SingleNotMulti,
+    #[error("You need a name for this field")]
+    MultiNotSingle,
+    #[error("Key not found in multi-value group")]
+    MultiKeyNotFound,
```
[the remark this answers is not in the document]
reubenmiller: I would call this a "profile" or "instance"

crates/common/tedge_config_macros/src/multi.rs
```diff
@@ -0,0 +1,209 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum MultiError {
+    #[error("You are trying to access a named field ({1}) of {0}, but the fields are not named")]
+    SingleNotMulti(String, String),
+    #[error("You need a name for the field {0}")]
+    MultiNotSingle(String),
+    #[error("Key {0}.{1} not found in multi-value group")]
+    MultiKeyNotFound(String, String),
+}
```
didier-wenzek, line 53: Thanks for adding a context to these error messages.

I would go one step further, improving the wording after [Reuben's suggestion to use the term *profile*](https://github.com/thin-edge/thin-edge.io/pull/3126#discussion_r1781215203).

```suggestion
#[derive(Debug, thiserror::Error)]
pub enum MultiError {
    #[error("Unexpected profile {1} for the non multi-profile property {0}")]
    SingleNotMulti(String, String),
    #[error("A profile is required for the multi-profile property {0}")]
    MultiNotSingle(String),
    #[error("Unknown profile {1} for the multi-profile property {0}")]
    MultiKeyNotFound(String, String),
}
```

By the way, I don't see how a user can get the `SingleNotMulti` error case. If a property is not multi, then the error will be that the key is unknown. Isn't it?

crates/common/tedge_config_macros/impl/src/reader.rs
```diff
@@ -334,14 +340,50 @@ fn reader_value_for_field<'a>(
                 }
                 FieldDefault::FromKey(default_key) | FieldDefault::FromOptionalKey(default_key) => {
                     observed_keys.push(default_key);
+                    // TODO error if the field touches an unrelated multi?
```
jarhodes314, line 343: Oops, this is done now, and tested, although the error message https://github.com/thin-edge/thin-edge.io/pull/3126/commits/85b30f82e42672999fdaa6389f2bccbd882a8bba#diff-407a6572ceebfdf137f7e0369291214688fa368a92f6381eb18689aa5d8eb982R374  isn't very informative

crates/common/tedge_config_macros/impl/src/reader.rs
```diff
@@ -314,15 +399,101 @@ fn reader_value_for_field<'a>(
...
+    for (i, field) in key.iter().take(key.len() - 1).enumerate() {
+        new_parents.push(PathItem::Static(field.to_owned()));
+        if let Some(PathItem::Static(s)) = parents.next() {
+            if field == s {
+                if let Some(PathItem::Dynamic(span)) = parents.peek() {
+                    parents.next();
+                    new_parents.push(PathItem::Dynamic(*span));
+                }
+            } else {
+                // This key has diverged from the currrent key's parents, so empty the list
```
didier-wenzek, line 445:
```suggestion
                // This key has diverged from the current key's parents, so empty the list
```

crates/common/tedge_config_macros/src/multi.rs
```diff
@@ -44,7 +44,9 @@ impl<T: Default + PartialEq> MultiDto<T> {
 
 #[derive(Debug, thiserror::Error)]
 pub enum MultiError {
-    #[error("Unexpected profile {1} for the non multi-profile property {0}")]
+    #[error(
+        "You are trying to access a profile `{1}` of {0}, but profiles are not enabled for {0}"
+    )]
```
didier-wenzek: I have a preference for the former error message.

didier-wenzek (approved):
Approved. Thank you for this *tour de force*.

Despite its underneath complexity, tedge config is simple to use both by end users and developers  (despite or thanks to this complexity?). However, we reached a point where things are really complicated and I would be very careful about adding additional features.

Issue #3124, opened 2024-09-19
Cumulocity mapper stuck in a disconnect-reconnect loop when it receives a message more than 1 MiB in size


albinsuresh:
**Describe the bug**

When a large payload > 1 MiB is published to the Cumulocity mapper, it gets stuck in a disconnect-reconnect loop at the MQTT client level. This happens because maximum message size configured by the mapper for its connection is just 1 MiB and when it receives any messages exceeding that limit, the embedded mqtt client disconnects itself from the broker and tries to reconnect, only for the same message to be re-delivered which triggers the next disconnection. And this cycle goes on forever.

**To Reproduce**

Publish a very large event message with a JSON fragment that exceeds 1 MiB in size. Then you'll notice the mapper repeatedly disconnecting and reconnecting in its logs as follows:

```
Sep 19 11:44:45 6ae4ab4508f7 tedge-mapper[2395]: 2024-09-19T11:44:45.16040775Z ERROR mqtt_channel::connection: MQTT connection error: Mqtt state: Mqtt serialization/deserialization error: payload size limit exceeded: 1048615
Sep 19 11:44:46 6ae4ab4508f7 tedge-mapper[2395]: 2024-09-19T11:44:46.164981578Z  INFO mqtt_channel::connection: MQTT connection re-established
Sep 19 11:44:46 6ae4ab4508f7 tedge-mapper[2395]: 2024-09-19T11:44:46.165699133Z ERROR mqtt_channel::connection: MQTT connection error: Mqtt state: Mqtt serialization/deserialization error: payload size limit exceeded: 1048615
Sep 19 11:44:47 6ae4ab4508f7 tedge-mapper[2395]: 2024-09-19T11:44:47.169036827Z  INFO mqtt_channel::connection: MQTT connection re-established
Sep 19 11:44:47 6ae4ab4508f7 tedge-mapper[2395]: 2024-09-19T11:44:47.169553586Z ERROR mqtt_channel::connection: MQTT connection error: Mqtt state: Mqtt serialization/deserialization error: payload size limit exceeded: 1048615
```

**Expected behavior**

The mapper should receive the message and drop it, if it can't be forwarded to the cloud, with an appropriate error message logged.

**Additional context**

The same limitation message size limitation exists for the other cloud mappers as well and even the agent, but with a higher message size limit of 10 MiB. The fix must cover them all.

albinsuresh:
Fixed by #3130 

### Test Plan

[...]

PR #3130, opened 2024-09-23
Increase default MQTT payload size to 256 MiB

albinsuresh:
## Proposed changes

* Increase default MQTT payload size to 256 MiB
* Make MQTT cloud payload size configurable

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3124

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/config.rs
+++ b/crates/common/mqtt_channel/src/config.rs
@@ -13,6 +13,8 @@ use std::path::Path;
 use std::sync::Arc;
 use zeroize::Zeroizing;
 
+pub const MAX_PACKET_SIZE: usize = 268435455;
+
 /// Configuration of an MQTT connection
 #[derive(Debug, Clone)]
 pub struct Config {
@@ -45,7 +47,7 @@ pub struct Config {
 
     /// Maximum size for a message payload
     ///
-    /// Default: `1024 * 1024`.
+    /// Default: `16777216` (16 MB).
     pub max_packet_size: usize,
 
     /// LastWill message for a mqtt client
@@ -164,7 +166,7 @@ impl Default for Config {
             subscriptions: TopicFilter::empty(),
             clean_session: false,
             queue_capacity: 1024,
-            max_packet_size: 1024 * 1024,
+            max_packet_size: 16 * 1024 * 1024,
             last_will_message: None,
             initial_message: None,
         }
@@ -351,7 +353,7 @@ impl Config {
             mqtt_options.set_transport(rumqttc::Transport::tls_with_config(tls_config.into()));
         }
 
-        mqtt_options.set_max_packet_size(self.max_packet_size, self.max_packet_size);
+        mqtt_options.set_max_packet_size(MAX_PACKET_SIZE, MAX_PACKET_SIZE);
 
         if let Some(lwp) = &self.last_will_message {
             let last_will_message = LastWill {
--- a/crates/common/tedge_config/src/tedge_config_cli/models/mod.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/models/mod.rs
@@ -9,6 +9,10 @@ pub mod port;
 pub mod seconds;
 pub mod templates_set;
 
+use doku::Document;
+use serde::Deserialize;
+use serde::Serialize;
+use std::fmt::Display;
 use std::str::FromStr;
 use strum::Display;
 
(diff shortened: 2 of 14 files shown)


crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -72,6 +72,7 @@ impl TEdgeComponent for AwsMapper {
             mqtt_schema,
             tedge_config.aws.mapper.timestamp_format,
             prefix.clone(),
+            tedge_config.az.mapper.mqtt.max_payload_size,
```
reubenmiller, line 75: Copy/paste typo

```suggestion
            tedge_config.aws.mapper.mqtt.max_payload_size,
```

albinsuresh: Fixed.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -164,7 +166,7 @@ impl Default for Config {
             subscriptions: TopicFilter::empty(),
             clean_session: false,
             queue_capacity: 1024,
-            max_packet_size: 1024 * 1024,
+            max_packet_size: MAX_PACKET_SIZE,
```
didier-wenzek, line 169: Looking closer to root cause of an infinite loop caused by a message that is too large, it appears that the issue is not rooted in the MQTT protocol per se, but is due to a bad combination of rumqttc and thin-edge.

- when a message that is too large is received, rumqttc raises a `PayloadSizeLimitExceeded` error
- when this error is received thin-edge do nothing beyond logging and adding some delay (which is useless) 
- rumqttc also cuts the connection and then reconnects.
- the message is then resent by the broker that is still expecting an acknowledgement.

Unfortunately, the rumqttc API provides no means to skip that poison pill (the error structs does not contain the message pkid).
 
=> Removing the limit as done here fixes the issue. However, we can consider to add our own check with no limit on what is received from rumqttc but with a filter implemented by `mqtt_channel` to prune messages that are too large. Without this limit a rogue client might overwhelmed the mapper or the agent.

didier-wenzek (commented):
The issue is addressed by the proposed fix. However, this introduces a weakness, and I would consider to limit the size of the messages actually processed by thin-edge, ignoring those too large.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -513,6 +517,14 @@ define_tedge_config! {
             device_profile: bool,
         },
 
+        mapper: {
+            mqtt: {
+                /// The maximum message payload size that can be mapped to the cloud
+                #[tedge_config(example = "8192", default(variable = "C8Y_MQTT_PAYLOAD_LIMIT"))]
+                max_payload_size: u32,
```
rina23q, line 524: The maximum size of `u32` is much bigger than 256MiB. (`4294967295` > `268435455`). I think we need protection for such big input from user.

crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -72,6 +72,7 @@ impl TEdgeComponent for AwsMapper {
             mqtt_schema,
             tedge_config.aws.mapper.timestamp_format,
             prefix.clone(),
+            tedge_config.aws.mapper.mqtt.max_payload_size,
```
rina23q, line 75: Related to my comment above, for example, here we can add the size check for `aws.mapper.mqtt.max_payload_size`

rina23q (commented):
One suggestion

crates/core/tedge/src/cli/mqtt/mod.rs
```diff
@@ -11,7 +11,7 @@ mod error;
 mod publish;
 mod subscribe;
 
-const MAX_PACKET_SIZE: usize = 10 * 1024 * 1024;
+const MAX_PACKET_SIZE: usize = 268435455; // 256 MB
```
didier-wenzek, line 14: I would not change the default, now that the correct protection is implemented.

```suggestion
    /// Default: `1024 * 1024`.
```

didier-wenzek (commented):
The code is correct. I would prefer to revert the default max size to the previous values.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -45,7 +47,7 @@ pub struct Config {
 
     /// Maximum size for a message payload
     ///
-    /// Default: `1024 * 1024`.
+    /// Default: `268435455` (256 MB).
```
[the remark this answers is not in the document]
albinsuresh: Resolved by f43495ebc367685bb9cfc8fafdc9d445a8eea263

crates/core/tedge/src/cli/mqtt/mod.rs
albinsuresh: Resolved by f43495ebc367685bb9cfc8fafdc9d445a8eea263

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
albinsuresh: Resolved by f32c7099318420e693a89e9afe2823e9cc71b5a7

crates/core/tedge_mapper/src/aws/mapper.rs
albinsuresh: Implemented the validation at the `tedge_config` level itself, so that the user gets the feedback much earlier.

crates/common/mqtt_channel/src/config.rs
albinsuresh: Resolved a4a6af90e047bae0e82be752512648b90ec71583

crates/common/tedge_config/src/tedge_config_cli/models/mod.rs
```diff
@@ -59,3 +63,67 @@ impl FromStr for AutoLogUpload {
...
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        let limit = value
+            .parse::<u32>()
+            .map_err(|_| InvalidMqttPayloadLimit(value.to_string()))?;
+        limit.try_into()
+    }
+}
+
+impl TryFrom<u32> for MqttPayloadLimit {
```
didier-wenzek, line 87: I wouldn't have gone so far. But since  this has been done, we can keep it.

didier-wenzek (approved):
Approved

Issue #3128, opened 2024-09-22
status field in services messages is ignored if pid field is not an integer


reubenmiller:
**Describe the bug**

In https://github.com/thin-edge/thin-edge.io/pull/2919, the parsing of the service status messages added the `pid` field (as a `Option<u32>`), however adding this field made the parsing of the message had the side effect from rejecting the whole message if the `pid` field was not in the expected format. In previous thin-edge.io versions, the `pid` field could of been any data type, where as thin-edge.io 1.3.0 it MUST be left out or an uint.

This chang causes a breaking change for service status payloads that, which results in the `status` property being ignored and defaulting to `unknown` (due to the parsing error on the `pid` field). From a logic point of view, the mapper should not care about the `pid` field as it is not using the the information in the mapping, so it should not affect the parsing of the service `status` field.

For example, the publishing the following message with the `pid` set as a string, will result in the outgoing message to Cumulocity IoT, will use the status=`unknown` on the outgoing message.

```sh
tedge mqtt pub te/device/main/service/my-service/status/health '{"status": "up", "pid": "21037"}'
```

**Message published to Cumulocity IoT**

```sh
[c8y/s/us] 102,rmi-macos:device:main:service:my-service,service,my-service,unknown
```

**To Reproduce**

1. On the device, open a new console and subscribe to the outgoing c8y cloud messages

    ```sh
    tedge mqtt sub 'c8y/s/us'
    ```

2. Open a new console, publish the following message

    ```sh
    tedge mqtt pub te/device/main/service/my-service/status/health '{"status": "up", "pid": "21037"}'
    ```

3. Observe the outgoing cloud message in the first console, it will show the `unknown` value (below is an example of the message)

    ```sh
    [c8y/s/us] 102,rmi-macos:device:main:service:my-service,service,my-service,unknown
    ```

**Expected behavior**

The tedge-mapper should only try to parse the information that is required for the mapping of the service message. At this point in time the `pid` field is not used by the mapper, so the value of it should not affect the parsing of the `status` field.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `any`
 - Hardware [incl. revision]: `any`
 - System-Architecture [e.g. result of "uname -a"]: `any`
 - thin-edge.io version [e.g. 0.1.0]: `1.3.0`

**Additional context**

reubenmiller:
**Update** The initial conclusion I posted in the ticket was incorrect and I've updated the ticket title and description accordingly

[...]

PR #3132, opened 2024-09-24
fix: Ignore extra HealthStatus fields in service monitor

Bravo555:
## Proposed changes

tedge-mapper will now not only try to parse `status` field and will ignore other extra fields.

## Types of changes

## Paste Link to the issue

- #3128

## Checklist

### DIFF
--- a/crates/core/tedge_watchdog/src/systemd_watchdog.rs
+++ b/crates/core/tedge_watchdog/src/systemd_watchdog.rs
@@ -1,4 +1,3 @@
-use crate::error::WatchdogError;
 use anyhow::Context;
 use freedesktop_entry_parser::parse_entry;
 use futures::channel::mpsc;
@@ -8,6 +7,9 @@ use futures::StreamExt;
 use mqtt_channel::MqttMessage;
 use mqtt_channel::PubChannel;
 use mqtt_channel::Topic;
+use serde::Deserialize;
+use serde::Serialize;
+use serde_json::Value as JsonValue;
 use std::path::PathBuf;
 use std::process;
 use std::process::Command;
@@ -20,7 +22,6 @@ use tedge_api::mqtt_topics::Channel;
 use tedge_api::mqtt_topics::EntityTopicId;
 use tedge_api::mqtt_topics::MqttSchema;
 use tedge_api::mqtt_topics::OperationType;
-use tedge_api::HealthStatus;
 use tedge_config::TEdgeConfigLocation;
 use tedge_utils::timestamp::IsoOrUnix;
 use time::OffsetDateTime;
@@ -28,6 +29,8 @@ use tracing::debug;
 use tracing::error;
 use tracing::info;
 use tracing::warn;
+
+use crate::error::WatchdogError;
 
 const SERVICE_NAME: &str = "tedge-watchdog";
 
@@ -37,6 +40,16 @@ const SERVICE_NAME: &str = "tedge-watchdog";
 /// Notifications are sent more often to make sure we don't miss the notify interval due to
 /// a timing misalignment.
 const NOTIFY_SEND_FREQ_RATIO: u64 = 4;
+
+/// A subset of fields of health status payload required by the watchdog.
+///
+/// https://thin-edge.github.io/thin-edge.io/operate/troubleshooting/monitoring-service-health/
+#[derive(Debug, Serialize, Deserialize)]
+struct HealthStatusExt {
+    /// Used for tracking service restarts
+    pub pid: Option<u32>,
+    pub time: Option<JsonValue>,
+}
 
 pub async fn start_watchdog(tedge_config_dir: PathBuf) -> Result<(), anyhow::Error> {
     // Send ready notification to systemd.
@@ -128,7 +141,6 @@ async fn start_watchdog_for_tedge_services(tedge_config_dir: PathBuf) {
 
                 let tedge_config_location = tedge_config_location.clone();
                 watchdog_tasks.push(tokio::spawn(async move {
-                    //
                     let interval = Duration::from_secs((interval / NOTIFY_SEND_FREQ_RATIO).max(1));
                     monitor_tedge_service(
                         tedge_config_location,
@@ -243,11 +255,11 @@ async fn monitor_tedge_service(
 async fn get_latest_health_status_message(
     request_timestamp: OffsetDateTime,
     messages: &mut mpsc::UnboundedReceiver<MqttMessage>,
-) -> Result<HealthStatus, WatchdogError> {
+) -> Result<HealthStatusExt, WatchdogError> {
     while let Some(message) = messages.next().await {
         if let Ok(message) = message.payload_str() {
             debug!("Health response received: {message}");
-            if let Ok(health_status) = serde_json::from_str::<HealthStatus>(message) {
+            if let Ok(health_status) = serde_json::from_str::<HealthStatusExt>(message) {
                 if health_status.time.is_none() {
                     error!("Ignoring invalid health response: {health_status:?} without a `time` field in it");
                     continue;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,13 @@ c8y_mapper_ext = { path = "crates/extensions/c8y_mapper_ext" }
 camino = "1.1"
 cap = "0.1"
 certificate = { path = "crates/common/certificate" }
-clap = { version = "4.4", features = ["cargo", "derive"] }
+clap = { version = "4.4", features = [
+    "cargo",
+    "derive",
+    "string",
+    "env",
+    "unstable-styles",
+] }
 clock = { path = "crates/common/clock" }
 collectd_ext = { path = "crates/extensions/collectd_ext" }
 csv = "1.1"
--- a/crates/core/tedge/Cargo.toml
+++ b/crates/core/tedge/Cargo.toml
@@ -19,13 +19,7 @@ c8y-remote-access-plugin = { workspace = true }
 camino = { workspace = true }
 cap = { workspace = true }
 certificate = { workspace = true, features = ["reqwest-blocking"] }
-clap = { workspace = true, features = [
-    "cargo",
-    "derive",
-    "env",
-    "string",
-    "unstable-styles",
-] }
+clap = { workspace = true }
 doku = { workspace = true }
 hyper = { workspace = true, default-features = false }
 nix = { workspace = true }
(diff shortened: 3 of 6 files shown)


crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -38,6 +41,15 @@ const SERVICE_NAME: &str = "tedge-watchdog";
 /// a timing misalignment.
 const NOTIFY_SEND_FREQ_RATIO: u64 = 4;
 
+#[derive(Debug, Serialize, Deserialize)]
+pub struct HealthStatusExt {
```
didier-wenzek, line 45: I would consider to move this struct in `tedge_api`.

Also, I'm not really convinced by having two public structs with similar purpose. Wouldn't be better to keep the `HealthStatus` struct unchanged with a deserialize method that fallbacks on a minimal private struct in case some fields cannot be parsed?

Bravo555: I did talk about this with @albinsuresh because I started out with the same approach but concluded that having separate structs with the one in `tedge_watchdog` extending the one in `tedge_api` would be better.

> Also, I'm not really convinced by having two **public** structs with similar purpose.

My bad, the struct wasn't supposed to be public and the aim was for it to be very much private to the `tedge_watchdog` crate. `pub` was a leftover from when I had it in a nested module, which is now removed. 

> Wouldn't be better to keep the HealthStatus struct unchanged with a deserialize method that fallbacks on a minimal private struct in case some fields cannot be parsed?

I understand that pros of the "one struct" approach to be the following:
- simpler interface (only one type of `HealthStatus`) at the cost of a more complex implementation (custom `Deserialize` impl that tries to deserialize fields to an expected format but falls back to some other representation when it's not possible)
- all possible fields in one place so one has to only look at `tedge_api` to see that `HealthStatus` can have these fields only

But IMHO it would make sense to keep `HealthStatus` in `tedge_api` as small as possible and not to put in fields that are not really required and erect more structure than necessary:

My view is that the payload under `te/device/device001/service/service001/status/health` is from an open set: it needs to have  `status` property for its status to be synced with `c8y`, but it may also have `pid` and `time` when it's using `tedge_watchdog`. Perhaps some other component would require other fields to be present, but not care about `status`, `pid`, or `time`.

So we can't really deserialize to a struct that has all these fields defined because then if there's a problem with one of the fields that we're not using, we can't use other fields that we need and that are fine.

The easiest would be to just deserialize to `HashMap<String, JsonValue>`, but then the callers have to look up the fields they need and validate their types.

So we can do one additional step - make a struct for these properties so that `serde` will fetch the fields and validate their types and values for us. Because serde by default omits unknown fields, the struct doesn't say "Health status payload has **only** these fields", it says "it has **at least** these fields"

> deserialize method that fallbacks on a minimal private struct in case some fields cannot be parsed

This is the somewhat problematic part. Deserialization either succeeds or fails, so it's not clear to me how to handle failures on individual fields while keeping other fields in a good way. If we want to deserialize e.g. `pid: Option<u32>`, how to handle `"pid": "invalid"` case? If we put `None`, we lose the error. If we deserialize to `Result<u32, DeserializeError>` that's just clumsy. Maybe we can put it into `extra: HashMap<String, JsonValue>`, but the outcome is not much different from just skipping the complex deserialize impl and using the `extra` hashmap directly, or even better defining another struct that has fields we require so that we don't have to extract from the hashmap manually.

IMO the default serde behaviour of "I want these fields to be present and valid and I want to ignore other fields" is conducive to the approach of  defining multiple structs, even for things that are only slightly different.

**TL;DR**: IMO health status payload is a dynamic type, we shouldn't try to enumerate all possible fields for a deserialize impl, and coping with different components having different failure modes when some of the fields are missing. Instead it makes sense to only require fields we require in each caller, but be upfront in the documentation that the JSON payload may also contain other fields.

didier-wenzek: > Perhaps some other component would require other fields to be present, but not care about `status`, `pid`, or `time`.

Here, you are going a bit too far. If there no relationship with an  `HealthStatus`, what's the point to use that struct?

> The easiest would be to just deserialize to `HashMap<String, JsonValue>`, but then the callers have to look up the fields they need and validate their types.

This definitely not what I'm asking for.

> Deserialization either succeeds or fails, so it's not clear to me how to handle failures on individual fields while keeping other fields in a good way. If we want to deserialize e.g. `pid: Option<u32>`, how to handle `"pid": "invalid"` case? If we put `None`, we lose the error. If we deserialize to `Result<u32, DeserializeError>` that's just clumsy.

I would put `None` in *that* case (I agree this is not always appropriate), meaning there is no `pid` information that can be used.

> IMO the default serde behaviour of "I want these fields to be present and valid and I want to ignore other fields" is conducive to the approach of  defining multiple structs, even for things that are only slightly different.

It can be indeed appropriate to define multiple *independent* structs. But here you propose something a bit complicated for nothing: why an extension of `tedge_api::HealthStatus` and not a fully independent `tedge_watchdog::HealthStatus`.

This is even a surprise to me that `tedge_watchdog` doesn't use the `status` field! Did I miss something?

Bravo555: > It can be indeed appropriate to define multiple independent structs. But here you propose something a bit complicated for nothing: why an extension of tedge_api::HealthStatus and not a fully independent tedge_watchdog::HealthStatus.

Right, if there were more fields in base `HealthStatus` that `tedge_watchdog` would be using then it would make sense to extend, right now they can be made independent with not much difference.

> This is even a surprise to me that tedge_watchdog doesn't use the status field! Did I miss something?

You're correct, `tedge_watchdog` seems to not inspect `status` field before sending notifications to systemd, but it probably should, right?

In any case, I'll provide an alternative implementation more in line with your requirements - I had it working at one point but decided the current approach to be simpler, so I'll just find it in my `git reflog` and we can compare.

didier-wenzek: > In any case, I'll provide an alternative implementation more in line with your requirements - I had it working at one point but decided the current approach to be simpler, so I'll just find it in my `git reflog` and we can compare. 

As `tedge_watchdog` is not using the `status` field, the simpler is to go your way but with two fully independent structs.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -38,6 +41,18 @@ const SERVICE_NAME: &str = "tedge-watchdog";
 /// a timing misalignment.
 const NOTIFY_SEND_FREQ_RATIO: u64 = 4;
 
+/// A subset of fields of health status payload required by the watchdog.
+///
+/// https://thin-edge.github.io/thin-edge.io/operate/troubleshooting/monitoring-service-health/
+#[derive(Debug, Serialize, Deserialize)]
+struct HealthStatusExt {
+    #[serde(flatten)]
+    pub health_status: tedge_api::HealthStatus,
```
didier-wenzek, line 50: Oddly enough, it's no use!

```suggestion
```

Bravo555: fixed 1753f4696c

crates/core/tedge_watchdog/src/systemd_watchdog.rs
Bravo555: Alternative approach here: #3133

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -243,11 +255,11 @@ async fn monitor_tedge_service(
 async fn get_latest_health_status_message(
     request_timestamp: OffsetDateTime,
     messages: &mut mpsc::UnboundedReceiver<MqttMessage>,
-) -> Result<HealthStatus, WatchdogError> {
+) -> Result<HealthStatusExt, WatchdogError> {
     while let Some(message) = messages.next().await {
         if let Ok(message) = message.payload_str() {
             debug!("Health response received: {message}");
-            if let Ok(health_status) = serde_json::from_str::<HealthStatus>(message) {
+            if let Ok(health_status) = serde_json::from_str::<HealthStatusExt>(message) {
```
rina23q, line 262: I feel a bit odd that the watchdog doesn't seem caring if the `status` is `up` or not. As long as `pid` and `time` are included in the health status message, the health status is considered "up"?

rina23q (commented):
Question

crates/core/tedge_watchdog/src/systemd_watchdog.rs
Bravo555: Yes, if it contains `time` and `pid` it sends systemd-notify notification without checking `status`, and this definitely seems like an overlook to me. I would expect notifications to be sent only when `status` is `up`.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
```diff
@@ -38,6 +41,16 @@ const SERVICE_NAME: &str = "tedge-watchdog";
 /// a timing misalignment.
 const NOTIFY_SEND_FREQ_RATIO: u64 = 4;
 
+/// A subset of fields of health status payload required by the watchdog.
+///
+/// https://thin-edge.github.io/thin-edge.io/operate/troubleshooting/monitoring-service-health/
+#[derive(Debug, Serialize, Deserialize)]
+struct HealthStatusExt {
```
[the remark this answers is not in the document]
didier-wenzek: > I was expecting this struct to wrap the `HealthStatus` struct as a flattened field, hence the name.

Yeah, but the wrapped value is not used.

crates/core/tedge_watchdog/src/systemd_watchdog.rs
rina23q: > Yes, if it contains `time` and `pid` it sends systemd-notify notification without checking `status`, and this definitely seems like an overlook to me. I would expect notifications to be sent only when `status` is `up`.

I also expect in this way. However, it's out of scope to fix it in this PR?

crates/core/tedge_watchdog/src/systemd_watchdog.rs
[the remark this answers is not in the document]
rina23q: > Yeah, but the wrapped value is not used.

I think the value should have used. It's the implementation oversight that we are not using the `status` in watchdog. See [this thread](https://github.com/thin-edge/thin-edge.io/pull/3132/files#r1773564626).

rina23q (approved):
Approved the code as of now. I'm looking forward the follow-up PR to fix the watchdog issue! 👍

PR #3133, opened 2024-09-24
fix: Ignore extra HealthStatus fields in service monitor (maintain single `HealthStatus`)

Bravo555:
## Proposed changes

This is an alternative fix to #3132. It contains the same test, but differs in the approach to how we should deserialize `HealthStatus`.

In #3132, we use separate structs in `tedge_api` and `tedge_watchdog`. This allows each component to deserialize only mapper and watchdog to deserialize only fields they require, but could be argued duplicates structs and makes it harder to see what are possible fields for this type of message because they're spread out.

In this PR, an alternative approach is taken where we use a single `tedge_api::HealthStatus` struct which just fills in a fallback `None` on deserialization errors for other fields. Here we keep all the fields in a single place but ignore errors, which is not desirable. I thought it's going to be more complex but it turned out simpler than I anticipated.

## Types of changes

## Paste Link to the issue
- #3128

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,13 @@ c8y_mapper_ext = { path = "crates/extensions/c8y_mapper_ext" }
 camino = "1.1"
 cap = "0.1"
 certificate = { path = "crates/common/certificate" }
-clap = { version = "4.4", features = ["cargo", "derive"] }
+clap = { version = "4.4", features = [
+    "cargo",
+    "derive",
+    "string",
+    "env",
+    "unstable-styles",
+] }
 clock = { path = "crates/common/clock" }
 collectd_ext = { path = "crates/extensions/collectd_ext" }
 csv = "1.1"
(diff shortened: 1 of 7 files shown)


didier-wenzek (commented):
Thank you for this alternative implementation for https://github.com/thin-edge/thin-edge.io/pull/3132.

Pros (of this proposal):
- A single `HealthStatus` struct
- More natural if the watchdog is fixed to use the `status` field.

Cons:
- More complex, even if simpler than what you were expecting.
- It's really not obvious that the mapper uses only the `status` field and the watchdog the `pid` and `time` fields.

For the latter point, I finally prefer the first proposal i.e. https://github.com/thin-edge/thin-edge.io/pull/3132.

@Bravo555 What's your point of view?

Bravo555:
> @Bravo555 What's your point of view?

If the team prefers the first proposal we can go with it, there's not much code built on top so we're free to change it later if it's no longer fit for purpose.

As for addressing the `status` field in `tedge_watchdog`, can be done in a follow-up PR.

didier-wenzek:
> If the team prefers the first proposal we can go with it, there's not much code built on top so we're free to change it later if it's no longer fit for purpose.

I'm okay first the first proposal. Approving it.

Issue #2986, opened 2024-07-09
c8y mapper panics with malformed registration messages


rina23q:
**Describe the bug**
c8y mapper panics with this error.

```
thread 'tokio-runtime-worker' panicked at crates/core/tedge_api/src/store/pending_entity_store.rs:162:49:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2024-07-09T16:23:02.121644225Z ERROR Runtime: Actor CumulocityMapper-9 has finished unsuccessfully: JoinError(JoinError::Panic(Id(28), ...))    
```

**To Reproduce**
Run these commands in this order.

```sh
tedge mqtt pub -r 'te/device/child02//' '{"@type":"child-device"}'
tedge mqtt pub -r 'te/device/child02//' '{"@type":"service"}'
tedge mqtt pub -r 'te/device/child02//' '{"@type":"child-device"}'
```

**Expected behavior**
No panic. This scenario, changing `@type` should be treated as an error.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [main branch (after 1.1.1 release)]  https://github.com/thin-edge/thin-edge.io/commit/6fc5699c7f027a6bd1e3dd1f3cfd178fe66ab200

**Additional context**

didier-wenzek:
This issue is quite serious. Trying to reproduce the error, I ended in a state where the mapper fails to start, panicking at `crates/core/tedge_api/src/store/pending_entity_store.rs:162:49`.

1. Sending a first `tedge mqtt pub -r 'te/device/child02//' '{"@type":"child-device"}'` is okay.
2. When sending an unexpected type `tedge mqtt pub -r 'te/device/child02//' '{"@type":"child-device"}'` makes the mapper crazy, taking 100% cpu.
3. If killed, the mapper fails then to restart.
4. The c8y mapper successfully restarts only after the child type is erased with `tedge mqtt pub -r 'te/device/child02//' ''`.

Note: for some unknown reason, there is no `/etc/tedge/device/inventory.json` on my device.

rina23q:
@didier-wenzek 
Thanks for reproducing and creating a PR. The behaviour that you described is exactly what I found.

> Note: for some unknown reason, there is no /etc/tedge/device/inventory.json on my device.

[...]

PR #3134, opened 2024-09-24
fix: c8y mapper panics on unexpected registration message

didier-wenzek:
The mapper panics when receiving a registration message such as `'te/device/child02//' '{"@type":"service"}'`.

The issue is that:

- the entity is seen as a child device i.e. matching `device/+//`
- but its registration message pretends that this is a service.
- the mapper tries then to register the entity as its own parent.

## Proposed changes

When the type of entity is explicitly set to be a service, but the topic id is attached to a child device (according to the default scheme), then this entity must be considered as non-following the default scheme.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2986

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -496,7 +496,7 @@ impl EntityStore {
             EntityType::Service => message
                 .parent
                 .clone()
-                .or_else(|| topic_id.default_parent_identifier())
+                .or_else(|| topic_id.default_service_parent_identifier())
                 .or_else(|| Some(self.main_device.clone())),
         };
 
@@ -610,7 +610,7 @@ impl EntityStore {
             let mut register_messages = vec![];
 
             let parent_device_id = entity_topic_id
-                .default_parent_identifier()
+                .default_source_device_identifier()
                 .expect("device id must be present as the topic id follows the default scheme");
 
             if !parent_device_id.is_default_main_device() && self.get(&parent_device_id).is_none() {
--- a/tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
+++ b/tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
@@ -484,6 +484,18 @@ Entities send to cloud on restart
     ...    minimum=1
     ...    maximum=1
 
+Unexpected message doesn't cause a panic #3134
+    # Register a service on a device topic
+    Execute Command    tedge mqtt pub --retain 'te/device/child1//' '{"@type":"service"}'
+    External Identity Should Exist    ${DEVICE_SN}:device:child1    show_info=False
+    Cumulocity.Managed Object Should Have Fragment Values    status\=up
+    Service Health Status Should Be Up    tedge-mapper-c8y
+
+    # Register a child device on a service topic
+    Execute Command    tedge mqtt pub --retain 'te/device/child2/service/foo' '{"@type":"child-device"}'
+    Device Should Exist    ${DEVICE_SN}:device:child2:service:foo
+    Service Health Status Should Be Up    tedge-mapper-c8y
+
 
 *** Keywords ***
 Should Have Retained Message Count
--- a/crates/core/tedge_api/src/mqtt_topics.rs
+++ b/crates/core/tedge_api/src/mqtt_topics.rs
@@ -395,14 +395,26 @@ impl EntityTopicId {
         }
     }
 
-    /// Returns the topic identifier of the parent of a service,
-    /// assuming `self` is the topic identifier of a device `device/+//
-    /// or a service `device/+/service/+`
+    /// Returns the topic identifier of the source device of an entity,
+    /// - for a service, this is the parent entity
+    /// - for a device, this is the device itself
     ///
-    /// Returns None if this is not a service or if the pattern doesn't apply.
-    pub fn default_parent_identifier(&self) -> Option<Self> {
+    /// Returns None if the pattern doesn't apply.
+    pub fn default_source_device_identifier(&self) -> Option<Self> {
         match self.0.split('/').collect::<Vec<&str>>()[..] {
             ["device", parent_id, "", ""] => Some(parent_id),
+            ["device", parent_id, "service", _] => Some(parent_id),
+            _ => None,
+        }
+        .map(|parent_id| EntityTopicId(format!("device/{parent_id}//")))
+    }
+
+    /// Returns the topic identifier of the parent of a service,
+    /// assuming `self` is the topic identifier of a service `device/+/service/+`
+    ///
+    /// Returns None if this is not a service or if the pattern doesn't apply.
+    pub fn default_service_parent_identifier(&self) -> Option<Self> {
+        match self.0.split('/').collect::<Vec<&str>>()[..] {
             ["device", parent_id, "service", _] => Some(parent_id),
             _ => None,
         }


crates/core/tedge_api/src/entity_store.rs
```diff
@@ -496,7 +496,7 @@ impl EntityStore {
             EntityType::Service => message
                 .parent
                 .clone()
-                .or_else(|| topic_id.default_parent_identifier())
+                .or_else(|| topic_id.default_service_parent_identifier())
```
didier-wenzek, line 499: My first attempt as been to rewrite [TopicId::default_parent_identifier](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/mqtt_topics.rs#L403) along the following lines:

```rust
    /// Returns the topic identifier of the parent of an entity
    pub fn default_parent_identifier(&self) -> Option<Self> {
        match self.0.split('/').collect::<Vec<&str>>()[..] {
            ["device", "main", "", ""] => None,
            ["device", _, "", ""] => Some("main"),
            ["device", parent_id, "service", _] => Some(parent_id),
            _ => None,
        }
        .map(|parent_id| EntityTopicId(format!("device/{parent_id}//")))
    }
```

Indeed the current behavior of this function is weird (the parent of a child device is the child itself) and doesn't match is doc comment.

However, my proposal matches the doc comment but makes fail 35 unit tests!

@albinsuresh any idea on why this is not the correct way to fix this issue?

albinsuresh: Yeah, the current impl of `default_parent_identifier` seems wrong as it implies that the parent of `device/child1//` is `child1` which doesn't make any sense. That logic only applies to `service` entities following the default topic scheme. So, the new function that you've added is correct and named appropriately. I'd drop the old one in favour of the new one.

albinsuresh: Regarding the test failures (I looked at some, not all), it's coming from the auto-registration logic that tries to register `device/child1//` entity when a message like `te/device/child1///m/temp` is received. That auto-registration logic relies on the same `default_parent_identifier` function to extract the target entity topic id from the incoming topic. In this case, theyre' not looking for the parent, but the target entity itself. So, the current `default_parent_identifier` function can be renamed to `default_target_entity_identifier` to better reflect that logic, and the newly added function can be used to detect the "parent" of a "service".

didier-wenzek: >  In this case, they're not looking for the "parent", but the "source" entity itself. 

That's really confusing but is aligned with the code (returning self for child devices and the parent for a service).

=> I will fix the name and the doc comment.

didier-wenzek: > => I will fix the name and the doc comment.

Done: https://github.com/thin-edge/thin-edge.io/pull/3134/commits/c24e633d49047801a3d0e7af77a52ee7573978d4

rina23q (commented):
Looks good, but test seems missing. I'm not sure which level of test is suitable for this though. Maybe adding a system test to [registration_lifecycle](https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot)?
I can add a test there if you like.

rina23q (approved):

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -484,6 +484,18 @@ Entities send to cloud on restart
     ...    minimum=1
     ...    maximum=1
 
+Unexpected message doesn't cause a panic #3134
+    # Registered as service
+    Execute Command    tedge mqtt pub --retain 'te/device/child1//' '{"@type":"service"}'
```
didier-wenzek, line 489:
```suggestion
    # Register a device topic as a service
    Execute Command    tedge mqtt pub --retain 'te/device/child1//' '{"@type":"service"}'
```

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -484,6 +484,18 @@ Entities send to cloud on restart
...
+    External Identity Should Exist    ${DEVICE_SN}:device:child1    show_info=False
+    Cumulocity.Managed Object Should Have Fragment Values    status\=up
+    Service Health Status Should Be Up    tedge-mapper-c8y
+
+    # Registered as child device
+    Execute Command    tedge mqtt pub --retain 'te/device/child2/service/foo' '{"@type":"child-device"}'
```
didier-wenzek, line 495:
```suggestion
    # Register a service topic as a child device
    Execute Command    tedge mqtt pub --retain 'te/device/child2/service/foo' '{"@type":"child-device"}'
```

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -484,6 +484,18 @@ Entities send to cloud on restart
...
+    Device Should Exist    ${DEVICE_SN}:device:child2:service:foo
```
didier-wenzek, line 496: Here I do think the test should fail ... but it doesn't.

Indeed, ` 'te/device/child2/service/foo' '{"@type":"child-device"}'` registers a child device on a service topic.

@albinsuresh Is this expected or a side effect of auto-registration?

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
didier-wenzek: changed https://github.com/thin-edge/thin-edge.io/pull/3134/commits/515cb7bba8f1034cbaaf4bb0ec464f188cb4864b

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
didier-wenzek: Changed https://github.com/thin-edge/thin-edge.io/pull/3134/commits/515cb7bba8f1034cbaaf4bb0ec464f188cb4864b

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
reubenmiller: Though shouldn't auto registration only kick in if you are not registration something...here it is explicitly registering a given type, so there is no need to try to auto detect anything.

didier-wenzek: Indeed, this is what happens: no auto registration kicks in. That's a just bit confusing because applied on a topic matching auto registration requirements.

albinsuresh: Yes @reubenmiller is right. Auto-registration kicks in only when the source entity for a message isn't explicitly registered up-front. But, since there is an explicit registration happening here, it just doesn't kick in. In fact we advise the users to disable auto-registration if they decide to use explicit registration for any entities.

We don't prevent the users from re-using the default topic scheme while doing explicit registrations either, but yeah it can result in such unexpected/uncommon usages as well. Not sure if we should completely prevent it, as the default topic scheme still provides a starting point/guidance.

albinsuresh (approved):
LGTM.

PR #3138, opened 2024-09-25
test: Create a tedge-watchdog test suite

Bravo555:
## Proposed changes

Created a test suite checking if tedge-watchdog correctly prevents services from being restarted if they're healthy and makes systemd restart them when they're unhealthy.

We do have tests for other services that utilise tedge-watchdog which check if these services react to watchdog's health check requests and otherwise integrate with it, but they test these other services, not the watchdog itself. As a separate component, tedge-watchdog very much needs a test suite to itself validating its behaviour, particularly because it doesn't get much attention and isn't extensively documented.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

This test suite was motivated by the issue that was found in tedge-watchdog where it didn't look at status before notifying systemd that service is healthy. See: https://github.com/thin-edge/thin-edge.io/pull/3132#discussion_r1774825722

But as explained in the comment in the first test, despite this issue, the services for which `"status":"down"` is published are being restarted as we'd expect, either because of some intended trickery in the implementation or of some other issue. 

Now I'll try to establish why it fails, so the implementation might end up being correct, but still I think it would be good to have a dedicated test suite for tedge-watchdog.


albinsuresh:
> But as explained in the comment in the first test, despite this issue, the services for which "status":"down" is published are being restarted as we'd expect, either because of some intended trickery in the implementation or of some other issue.

I finally remember why it was written without an explicit `status` value check. If a service responds to the health check request, the watchdog implicitly assumes that the service is healthy, as it received a response. If the service was genuinely down, it wouldn't have gotten any response, not even the `down` response, as the `down` status message is sent only once by the MQTT broker when that service disconnects. But, `tedge-watchdog` does not rely on those retained messages, but actively sends a health check and awaits its response. That's why the whole scheme is working even without any explicit `status` field check. But, that logic was written with only `up` and `down` statuses considered, and not account for newly added ones like `unknown`. So, the logic needs to be updated to account for that.

didier-wenzek:
> But, that logic was written with only `up` and `down` statuses considered, and not account for newly added ones like `unknown`.

What's the purpose of this `unknown` status?

albinsuresh:
> > But, that logic was written with only `up` and `down` statuses considered, and not account for newly added ones like `unknown`.
> 
> What's the purpose of this `unknown` status?

To cover buggy clients that are reporting "some" health status messages on those specific topics, but not in the desired format. Interpreting such erroneous messages as `down` would be wrong, as something is still there responding to health requests, meaning it's really not "down". So, we want to differentiate such components from those that are genuinely down. This might help the user to fix the erroneous reporting of their component's health status.

Bravo555:
> But as explained in the comment in the first test, despite this issue, the services for which "status":"down" is published are being restarted as we'd expect, either because of some intended trickery in the implementation or of some other issue.

Found the problem that caused `tedge-watchdog` to disconnect. It wasn't due to `"status": "down"`, but due to `"time"` field. This is payload sent in the test:

```
Execute Command
...    tedge mqtt pub 'te/device/main/service/${SERVICE_NAME}/status/health' '{"status": "down", "pid": ${pid}, "time": "${before_restart_timestamp}"}'
```

`time` is a string, so we're expecting it to be in ISO format, but it's actually a UNIX timestamp instead. This returns an error from `get_latest_health_status_message` function and exits the watchdog for the service.

Changing `time` from string to float in the test (i.e. removing quotes) makes the test fail as expected.

---

> If a service responds to the health check request, the watchdog implicitly assumes that the service is healthy, as it received a response. If the service was genuinely down, it wouldn't have gotten any response, not even the `down` response

> But, that logic was written with only `up` and `down` statuses considered, and not account for newly added ones like `unknown`. So, the logic needs to be updated to account for that.

So it seems to me there are 2 TODOs at this moment:

- [ ] change deserialization logic so watchdog doesn't exit if it fails to deserialize time
- [ ] update logic so that we notify only on `"status": "up"` and `down/unknown` are treated as unhealthy for which we don't send notifications

tests/RobotFramework/tests/tedge_watchdog.robot
```diff
@@ -0,0 +1,115 @@
...
+    Start Service    ${SERVICE_NAME}
+    Start Service    tedge-watchdog
+
+    # wait for service to start
+    ${pid} =    Service Should Be Running    ${SERVICE_NAME}
+
+    # Simulate service signalling that it's unhealthy
+    ${before_restart_timestamp} =    Get Unix Timestamp
+    Execute Command
+    ...    tedge mqtt pub 'te/device/main/service/${SERVICE_NAME}/status/health' '{"status": "down", "pid": ${pid}, "time": "${before_restart_timestamp}"}'
```
albinsuresh, line 39: This may not be fully deterministic as the watchdog would be waiting for a response **after** it sends the health check (with a later timestamp) and just ignores other random health messages that it receives with a lower timestamp. So, we need to make sure that this response reaches the watchdog after the health check request is received, with a higher timestamp, and also ensure that it reaches before the real running `tedge-agent` responds with its `up` status.

gligorisaev (commented):
**Finding**

1. Non-deterministic Behavior:

• The watchdog sends a health check request and expects a response with a higher timestamp (indicating a more recent status). However, since the watchdog may receive multiple health messages (both valid and outdated), it must be able to discern between these messages.

• The risk is that the watchdog might ignore random health messages with lower timestamps that arrive after its health check, potentially leading to inconsistencies in how it processes the health state of the service.

2. Ensuring the Right Order of Events:

• To maintain determinism, the test should ensure that the watchdog receives the response to its health check request in the correct sequence:
	
• First, the watchdog sends a health check.
• Second, it receives a response with a timestamp that is newer than the health check request.
• The watchdog should not accept outdated health messages with lower timestamps that may incorrectly reflect the state of the service.

3. Race Condition Risk:
	
• There’s a race condition where the real, healthy tedge-agent might send an “up” status before the watchdog processes the simulated unhealthy response. If the healthy message arrives first, the test might falsely pass, even though it hasn’t correctly handled the simulated failure scenario.
	
• To avoid this, the test needs to ensure that the simulated unhealthy message is processed before the actual healthy status message from the real tedge-agent.

**Possible Issues in Testing:**

• Flaky Tests: Current test setup may produce inconsistent results, as the timing of messages isn’t guaranteed to always happen in the correct order.

**Solution Suggestions:**

• The test might need to introduce delays or synchronization mechanisms to ensure the correct sequence of events. For example, introducing a slight delay before the real tedge-agent sends its healthy status could help.

• Alternatively, adding more robust checks or timestamps in the test would allow it to explicitly confirm the order of events and prevent accepting out-of-order messages.

**Impact on Test Reliability:**

• Race Conditions: If the race condition isn’t handled properly, the test could pass incorrectly because it might receive the healthy message too soon.

• Deterministic Timing: For the test to be reliable, it must ensure that the simulated failure is properly processed and reflected in the results before the actual healthy state is acknowledged.

**Recommendation:**

To improve the reliability of the test:

• It should be ensured that the simulated unhealthy message (with a higher timestamp) is processed by the watchdog before the real service has a chance to send its healthy status.

• Should be Introduced specific synchronization points in the test where the test waits for confirmation that the health check has been processed before allowing further messages to be sent.

• Should be onsidered adding explicit logging or timestamps to better track when each message is sent and processed, allowing the test to fail more meaningfully when the sequence is incorrect.

Bravo555:
> 5. Race Condition Risk:
> 
> - There’s a race condition where the real, healthy tedge-agent might send an “up” status before the watchdog processes the simulated unhealthy response. If the healthy message arrives first, the test might falsely pass, even though it hasn’t correctly handled the simulated failure scenario.
> - To avoid this, the test needs to ensure that the simulated unhealthy message is processed before the actual healthy status message from the real tedge-agent.
> 
> **Possible Issues in Testing:**
> 
> - Flaky Tests: Current test setup may produce inconsistent results, as the timing of messages isn’t guaranteed to always happen in the correct order.

Indeed the current approach of letting a real service stay alive and send messages while the test tries to scoop in some of its own messages is not very reliable. To fix this, the solution would be to change the service unit definition so that it doesn't start a real service, but a mock, that the script can control. I've started to work on this, but got distracted by some other tickets.
I'll use this approach and implement the recommendations from the quoted comment once I've completed other more pressing work

tests/RobotFramework/tests/tedge_watchdog.robot
didier-wenzek: An way to *disconnect* tedge-watchdog from the MQTT topics used by the agent and the mapper (for testing purposes) is to configure tedge-watchdog with a fake topic root (say `test-te`).

reubenmiller:
FYI: I've created a [PR to track the decision](https://github.com/thin-edge/thin-edge.io/pull/3193) about the **tedge-watchdog** feature

tests/RobotFramework/tests/tedge_watchdog/health_check_respond.sh
```diff
@@ -0,0 +1,27 @@
...
+# SERVICE_NAME=$1
+
+# # e.g. '{"status":"up", "pid":2137, "time":1727439518}'
+
+SERVICE_NAME=${SERVICE_NAME:-"tedge-agent"}
+
+HEALTH_CHECK_TOPIC="te/device/main/service/$SERVICE_NAME/cmd/health/check"
+HEALTH_STATUS_TOPIC="te/device/main/service/$SERVICE_NAME/status/health"
+
+echo "Response: $RESPONSE"
```
albinsuresh, line 17:
```suggestion
# The  $RESPONSE flag is used to control whether this service responds to healthcheck requests or not
echo "Response: $RESPONSE"
```
Just some adding docs. I'd have called it `RESPOND` though.

tests/RobotFramework/tests/tedge_watchdog/tedge_watchdog.robot
```diff
@@ -0,0 +1,135 @@
...
+    Should Have MQTT Messages    te/device/main/service/${SERVICE_NAME}/cmd/health/check
+
+    Sleep    ${WATCHDOG_SEC}s
+
+    # Verify that service got restarted by systemd and it's now healthy
+    ${pid1} =    Service Should Be Running    ${SERVICE_NAME}
+    Should Not Be Equal    ${pid}    ${pid1}
+
+Watchdog doesn't fail on unexpected time format
+    Set Service health check response    response=1
```
albinsuresh, line 55:
```suggestion
    Set Service health check response    response=0
```
When you're simulating the health responses on behalf of the service from this test, the service should have be configured not to respond to the health checks by itself, right?

tests/RobotFramework/tests/tedge_watchdog/tedge_watchdog.robot
```diff
@@ -0,0 +1,135 @@
...
+    ${before_restart_timestamp} =    Get Unix Timestamp
+    Execute Command
+    ...    tedge mqtt pub 'te/device/main/service/${SERVICE_NAME}/status/health' '{"status": "down", "pid": ${pid}, "time": "${before_restart_timestamp}"}'
+
+    # Wait until service is restarted due to tedge-watchdog not sending out the notification
+    Sleep    ${WATCHDOG_SEC}s
+
+    # Verify that tedge-watchdog is still running
+    ${pid1} =    Service Should Be Running    tedge-watchdog
+    Should Be Equal    ${pid}    ${pid1}
```
albinsuresh, line 81: Shouldn't we also validate that the actual service itself is also restarted, when it keeps sending bad responses?

tests/RobotFramework/tests/tedge_watchdog/tedge_watchdog.robot
Bravo555: You're right, as we're sending messages from inside the test, the service should not respond.

tests/RobotFramework/tests/tedge_watchdog/tedge_watchdog.robot
```diff
@@ -0,0 +1,135 @@
...
+    Execute Command    ./bootstrap.sh --no-bootstrap --no-connect
+
+    Execute Command    mv /lib/systemd/system/tedge-agent.service /lib/systemd/system/tedge-agent.service.bak
+    Transfer To Device    ${CURDIR}/tedge-agent.service    /lib/systemd/system/
+
+    Execute Command    cmd=sed -i '13iWatchdogSec=${WATCHDOG_SEC}' /lib/systemd/system/${service_name}.service
+
+    Transfer To Device    ${CURDIR}/health_check_respond.sh    /setup/
+
+    # Without this line mqtt-logger can't connect to listener at 1883, but with it it successfully connects to listener
```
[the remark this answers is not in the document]
Bravo555: Fixed cdca9120e4

tests/RobotFramework/tests/tedge_watchdog/health_check_respond.sh
Bravo555: fixed cdca9120e4

tests/RobotFramework/tests/tedge_watchdog/tedge_watchdog.robot
Bravo555: Indeed, apart from checking that tedge-watchdog doesn't crash, we can also check if it still functions correctly.
Also the test was buggy: tedge-watchdog PID was used in a message from a service, which was fixed.
fixed cdca9120e4

albinsuresh (approved):
LGTM

PR #3139, opened 2024-09-25
Refactor: Simplify entity auto-registration

didier-wenzek:
## Proposed changes

 Introduce a new function `mqtt_topic::default_topic_schema::parse()` taking an entity topic id and returning the registration messages matching the auto-registration behavior.
    
Using this `parse` function simplifies the auto-registration logic which has no more to parse again and again the entity topic id to extract registration info.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -8,6 +8,7 @@
 // TODO: move entity business logic to its own module
 
 use crate::entity_store;
+use crate::mqtt_topics::default_topic_schema;
 use crate::mqtt_topics::Channel;
 use crate::mqtt_topics::EntityTopicId;
 use crate::mqtt_topics::MqttSchema;
@@ -21,7 +22,6 @@ use log::error;
 use log::info;
 use log::warn;
 use mqtt_channel::MqttMessage;
-use serde_json::json;
 use serde_json::Map;
 use serde_json::Value as JsonValue;
 use std::collections::hash_map::Entry;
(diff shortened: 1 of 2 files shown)


crates/core/tedge_api/src/entity_store.rs
```diff
@@ -1401,7 +1378,7 @@ mod tests {
                     topic_id: EntityTopicId::from_str("device/child1//").unwrap(),
                     r#type: EntityType::ChildDevice,
                     external_id: Some("device:child1".into()),
-                    parent: None,
+                    parent: Some(EntityTopicId::from_str("device/main//").unwrap()),
```
didier-wenzek, line 1381: I was surprised by this. Why the parent of child device was expected to be `None`? Because no registration is required for the main device?

Bravo555 (approved):
Seems a small simplification with no functional changes, approving with a small comment regarding the changed unit test.

But I'm reminded how entity store and entity topic id logic could be made simpler, as currently it seems it has a very large surface with lots of methods that are ultimately a bit shallow.

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -456,6 +430,56 @@ impl EntityTopicId {
...
+                // Then the service can be registered
+                registrations.push(EntityRegistrationMessage {
+                    topic_id: topic_id.clone(),
+                    external_id: None,
+                    r#type: EntityType::Service,
+                    parent: Some(EntityTopicId::default_child_device(child).unwrap()),
+                    other: json!({ "name": service }).as_object().unwrap().to_owned(),
+                });
+                registrations
+            }
```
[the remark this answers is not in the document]
albinsuresh: Or even a dedicated arm for the `main` device service case, as the parent (`main` device) need not be registered in that case.

albinsuresh (commented):
The refactoring looks fine to me. Yet there are some test failures.

crates/core/tedge_api/src/entity_store.rs
didier-wenzek: > and the parent would be resolved by the entity store when registering the child.

This is what I would like to fix with this PR. Indeed, for auto registration, the entity store starts with a topic matching the default topic scheme (starting with a validation) and then extracts sub data (type, names, relationship) using misc functions assuming each the default topic scheme.

With this PR,  the entity store can parse an entity topic id into all the info required for auto-registration is applicable.

This also addresses what [you noted](https://github.com/thin-edge/thin-edge.io/pull/3139#pullrequestreview-2328805664) : *currently it seems it has a very large surface with lots of methods that are ultimately a bit shallow*

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -602,59 +602,36 @@ impl EntityStore {
...
-                return Ok(vec![]); // Do nothing as the main device is always pre-registered
-            }
-
-            let mut register_messages = vec![];
-
-            let parent_device_id = entity_topic_id
-                .default_source_device_identifier()
-                .expect("device id must be present as the topic id follows the default scheme");
-
-            if !parent_device_id.is_default_main_device() && self.get(&parent_device_id).is_none() {
```
didier-wenzek, line 616: Oops, I removed this check that the device should not already be registered.

```
self.get(&parent_device_id).is_none()
```

This explains why the system tests are failing.

For instance for []():

Before, only the service where registered:

```
c8y/s/us/foo_child1   102,foo:device:foo_child1:service:app1,service,app1,up
```

But this check missing service is also registered, while already known by c8y:

```
c8y/s/us                       101,foo:device:foo_child1,foo_child1,thin-edge.io-child
c8y/s/us/foo_child1   102,foo:device:foo_child1:service:app1,service,app1,up
```

didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/3139/commits/eaf5a8f765f9e6c64631f74b39835ba17a74698a (with a following simplification: https://github.com/thin-edge/thin-edge.io/pull/3139/commits/1d3f7070af0c8faeb7f339c59e9ec73ffda0c1ea)

crates/core/tedge_api/src/mqtt_topics.rs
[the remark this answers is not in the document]
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/3139/commits/1d3f7070af0c8faeb7f339c59e9ec73ffda0c1ea

crates/core/tedge_api/src/entity_store.rs
didier-wenzek: > So I think it'd be better to leave it as None.

However, I will opt for the other way, i.e. setting the parent. Indeed, the point of the `parse(entity_topic_id)`
is to provide an *explicit representation* of all what is implied by an entity topic of the default topic scheme.

albinsuresh (approved):
LGTM.

PR #3142, opened 2024-09-27
fix: Support skipping operations in device profile

albinsuresh:
## Proposed changes

Support skipping operations in device profile

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/workflow/mod.rs
+++ b/crates/core/tedge_api/src/workflow/mod.rs
@@ -401,10 +401,21 @@ impl OperationAction {
                 return Err(IterationError::IndexOutOfBounds(index));
             }
 
-            let next_index = index + 1;
-            if next_index >= items.len() {
-                info!("Iteration finished");
-                return Ok(state.update(handlers.on_success));
+            let mut next_index = index + 1;
+            loop {
+                if next_index >= items.len() {
+                    info!("Iteration finished");
+                    return Ok(state.update(handlers.on_success));
+                }
+
+                let next = &items[next_index];
+                let skipped = next.get("@skip").and_then(|v| v.as_bool()).unwrap_or(false);
+                if skipped {
+                    next_index += 1;
+                    continue;
+                } else {
+                    break;
+                }
             }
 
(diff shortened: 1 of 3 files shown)


reubenmiller:
@albinsuresh I don't think it makes sense to use `@skip` over `skip`. We previously used the `@` prefix to denote special thin-edge.io property (e.g. in the registration syntax), however in the same payload we still use `operation` and `payload`...so the `@skip` does not look consistent (imho).

I'd be happy to hear other opinions on this. @didier-wenzek @jarhodes314

didier-wenzek (approved):
Approved.

reubenmiller:
> > @albinsuresh I don't think it makes sense to use `@skip` over `skip`. We previously used the `@` prefix to denote special thin-edge.io property (e.g. in the registration syntax), however in the same payload we still use `operation` and `payload`...so the `@skip` does not look consistent (imho).
> > I'd be happy to hear other opinions on this. @didier-wenzek @jarhodes314
> 
> In my view, the `@skip` field is different from the `operation` and `payload` fields because the `iterator` doesn't really care that the `operation` and `payload` fields are named as such. They could have been named anything like `my_operation` and `my_payload` as well. The user explicitly refers to those fields in the `on_next` target of the `iterator` as follows:
> 
> ```
> [next_operation]
> iterate = "${.payload.operations}"
> on_next = "apply_operation"
> on_success = "twin_update"
> on_error = "rollback"
> 
> [apply_operation]
> operation = "${.payload.@next.item.my_operation}"
> input = "${.payload.@next.item.my_payload}"
> on_exec = "awaiting_operation"
> ```
> 
> But, the `@skip` field gets a special treatment by the `iterator` action during the iteration, just like the `@next` field is. And `@next` having that special `@` prefix is what that prompted me to add that same prefix to `@skip` as well, so that all our internal fields have that consistency.

With this explanation, then the `@skip` would be consistent then. Thanks for the thorough explanation.

reubenmiller (approved):

PR #3146, opened 2024-09-30
fix: Capability message for firmware_update workflow

albinsuresh:
## Proposed changes

Tedge-agent to publish empty capability message when a  workflow is defined for `firmware_update ` operation.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_workflow.robot
reubenmiller, line 1: I think we should just replace the existing system tests in this file: `tests/RobotFramework/tests/cumulocity/firmware/firmware_operation.robot`, as the old test file was created before thin-edge.io had official support for c8y_Firmware, and now the existing test add no value.

albinsuresh: Okay. I left it as-is since it's still testing the C8Y operation overriding mechanism with the `c8y_Firmware ` operation definition file. If we don't really need that anymore, then yeah I can replace that with this one.

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_workflow.robot
```diff
@@ -0,0 +1,34 @@
...
+Custom Test Setup
+    Execute Command
+    ...    cmd=echo 'tedge ALL = (ALL) NOPASSWD: /usr/bin/tedge, /usr/bin/systemctl, /etc/tedge/sm-plugins/[a-zA-Z0-9]*, /bin/sync, /sbin/init, /sbin/shutdown, /usr/bin/on_shutdown.sh, /usr/bin/tedge-write /etc/*' > /etc/sudoers.d/tedge
+
+Custom Setup
+    ${DEVICE_SN}=    Setup
+    Set Suite Variable    $DEVICE_SN
+    Device Should Exist    ${DEVICE_SN}
+
+    Execute Command    tedge config set c8y.enable.firmware_update true
```
albinsuresh, line 30: Wondering why we've disabled this by default at the mapper level. Unless someone explicitly creates a C8Y operation file or defines a workflow, the supported operation won't be declared by the mapper. So, I'm not sure why we want to disable the whole mapping itself, by default. @rina23q do you remember why? Can we switch to back to `true`?

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_workflow.robot
reubenmiller: Since we have workflows, we don't need the older operation declaration, especially for the c8y_Firmware operation as the official recommended extension point is to create a workflow.

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_workflow.robot
reubenmiller: I believe it was done to ensure we don't break systems which are using `c8y-firmware-plugin`...though we do need to think about a cleaner upgrade path (though correct me if I'm wrong here).

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_workflow.robot
```diff
@@ -0,0 +1,34 @@
...
+Suite Setup         Custom Setup
+Test Setup          Custom Test Setup
+Test Teardown       Get Logs
+
+Test Tags           theme:tedge_agent
+
+
+*** Test Cases ***
+Send firmware update operation from Cumulocity IoT
+    Cumulocity.Should Contain Supported Operations    c8y_Firmware
```
reubenmiller, line 15: It might also be worthwhile to add a check for the local capability message as well, e.g. `te/device/main///cmd/firmware_update`, to ensure that the functionality on the main device is discoverable to other clients listening to the same MQTT broker.

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_workflow.robot
rina23q: I think first mapper implementation was done while agent side was not completed. That's why it was marked `false` as a temporary solution, however, it's forgotten to be `true`, even after firmware implementation in agent side was done...

albinsuresh: Resolved by bdaabc5

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation_workflow.robot
albinsuresh: Resolved by bdaabc5

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation.robot
```diff
@@ -1,34 +1,28 @@
 *** Settings ***
 Resource            ../../../resources/common.resource
-Library             Cumulocity
 Library             ThinEdgeIO
+Library             Cumulocity
 
-Test Setup          Custom Setup
+Suite Setup         Custom Setup
```
reubenmiller, line 6: Why do we need to change to a `Suite Setup`, it would be better to leave it untouched as `Test Suite` is less prone to unexpected side effects from preceding tests.

```suggestion
Test Setup         Custom Setup
```

tests/RobotFramework/tests/cumulocity/firmware/firmware_operation.robot
```diff
@@ -1,34 +1,28 @@
 *** Settings ***
 Resource            ../../../resources/common.resource
-Library             Cumulocity
 Library             ThinEdgeIO
+Library             Cumulocity
```
reubenmiller, line 4: Revert the Libary import sequence (not change is necessary)

```
```

albinsuresh: Resolved by d2c2cff

reubenmiller (approved):
Looks good (Tests and change of default firmware_update setting)

rina23q (approved):

Issue #3143, opened 2024-09-27
File Ownership Issue in 'send collectd configuration to device'


ButKor:
**Describe the bug**
I am experiencing an unexpected behaviour when updating a certain configuration file (/etc/collectd/collectd.conf) via thin-edge from Cumulocity. The same operation succeeds only every second try, the others in between failing with:
```
config-manager failed downloading a file: Error while performing a file operation: Could not move file from "/tmp/collectd.conf.tmp" to "/tmp/collectd.conf": Failed to change owner: "/tmp/collectd.conf". 
```
See reproduction steps to make the problem clear. 

**To Reproduce**
1. Have a Cumulocity tenant and bootstrap a thin-edge device via https://github.com/thin-edge/tedge-demo-container
2. Go to your Device in Cumulocity and check current configuration of the tedge-configuration-plugin:
```
files = [
    { path = '/etc/tedge/tedge.toml', type = 'tedge.toml', user = 'tedge', group = 'tedge', mode = 0o444 },
    { path = '/etc/tedge/plugins/tedge-log-plugin.toml', type = 'tedge-log-plugin', user = 'tedge', group = 'tedge', mode = 0o644 },
    { path = '/etc/tedge/system.toml', type = 'system.toml', user = 'tedge', group = 'tedge', mode = 0o444 },
    { path = '/etc/mosquitto/conf.d/mosquitto.conf', type = 'mosquitto.conf', user = 'root', group = 'root', mode = 0o644 },
    { path = '/etc/collectd/collectd.conf', type = 'collectd.conf', user = 'root', group = 'root', mode = 0o644 },
    { path = '/etc/c8y-command-plugin/env', type = 'shell', user = 'tedge', group = 'tedge', mode = 0o644 },
]
```
3. Within Configuration tab select `collectd` and get snapshot from Device. Click on save to repository once it's there.
4. Stay on Configuration tab and send the collectd file (which you've just saved to repo) to the Device:
- First try: it fails with `Failed to change owner: "/tmp/collectd.conf"`
- Second try (on same device and same file): Now the operation succeeds.
- Third try: it fails with `Failed to change owner: "/tmp/collectd.conf"`
- Fourth try: succeeds
...and so on. 

**Environment (please complete the following information):**
 - thin-edge.io version 1.3.0
 - Environment: reproducible in tedge-demo-container, also reproducible on a Pi5

Bravo555:
First try fails because:
[...]

PR #3148, opened 2024-10-01
fix: Stop chowning in tedge-agent

Bravo555:
## Proposed changes

Stop setting file ownership in tedge-agent, which runs as `tedge` user. Also added a missing test for a case when a file is owned by root.

Setting file ownership is a privileged operation, which should be done by tedge-write. It's also unnecessary to change permissions on a temporary file which is atomically moved into destination by config-manager anyway.

## Types of changes

## Paste Link to the issue

- #3143

## Checklist

## Further comments

This change should resolve the issue and is small, so it'd be good to merge it ASAP. It resolves first 2 points of https://github.com/thin-edge/thin-edge.io/issues/3143#issuecomment-2383833471, i.e. adds missing test case and removes the usage of `chown` while running as `tedge`.

We still unnecessarily leave temporary files, but that will be done in a follow-up PR.

### DIFF
--- a/crates/extensions/tedge_downloader_ext/src/actor.rs
+++ b/crates/extensions/tedge_downloader_ext/src/actor.rs
@@ -40,13 +40,6 @@ impl DownloadRequest {
             ..self
         }
     }
-
-    pub fn with_permission(self, permission: PermissionEntry) -> Self {
-        Self {
-            permission: Some(permission),
-            ..self
-        }
-    }
 }
 
 pub type DownloadResult = Result<DownloadResponse, DownloadError>;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4023,7 +4023,6 @@ dependencies = [
  "tedge_utils",
  "tokio",
  "uzers",
- "whoami",
 ]
 
 [[package]]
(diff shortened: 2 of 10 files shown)


crates/extensions/tedge_downloader_ext/src/actor.rs
```diff
@@ -41,6 +41,7 @@ impl DownloadRequest {
         }
     }
 
+    #[deprecated = "Setting ownership is a privileged operation, only tedge-write which runs as root should do this"]
     pub fn with_permission(self, permission: PermissionEntry) -> Self {
```
didier-wenzek, line 45: Why not simply removing this method? This crate is only internally used.

Bravo555: Amended the commit to remove the method.

didier-wenzek (approved):
Approved

Bravo555:
Tried to merge 2 times, 3 tests failed, but they don't repeat across both tries, seems they're flakes? Going to test them with flake-finder:

- `Tests.Tedge Connect.Tedge Connect Offline.Check offline bootstrapping`
- ` Tests.Tedge Agent.Workflows.Custom Operation`
   - `Command steps should not be executed twice`
   - `Placeholder workflow created for ill-defined operations`

PR #3150, opened 2024-10-03
fix: check if log exists before trying to upload it

reubenmiller:
## Proposed changes

The tedge-mapper-c8y feature controlled by `c8y.operations.auto_log_upload=on-failure`, uploads an operation log based on the given condition. This PR just adds an additional check to only do the upload if the log file exists to avoid scenarios where an event is created with no attachment...this typically occurs when running the tedge-agent on a child device, and the mapper does not yet have access to the related operation log.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/operations/upload.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/upload.rs
@@ -79,7 +79,9 @@ impl OperationContext {
         command: &GenericCommandState,
     ) -> Result<(), ConversionError> {
         if command.is_finished()
-            && command.get_log_path().is_some()
+            && command
+                .get_log_path()
+                .is_some_and(|log_path| log_path.exists())
             && (self.auto_log_upload == AutoLogUpload::Always
                 || (self.auto_log_upload == AutoLogUpload::OnFailure && command.is_failed()))
         {


crates/extensions/c8y_mapper_ext/src/operations/upload.rs
```diff
@@ -79,7 +79,9 @@ impl OperationContext {
         command: &GenericCommandState,
     ) -> Result<(), ConversionError> {
         if command.is_finished()
-            && command.get_log_path().is_some()
+            && command
+                .get_log_path()
+                .is_some_and(|log_path| log_path.try_exists().is_ok())
```
didier-wenzek, line 84: I would simply use [`exists`](https://docs.rs/camino/latest/camino/struct.Utf8PathBuf.html#method.exists) as `try_exists().is_ok()` returns true is a symlink is broken.

reubenmiller: I read the docs of `.exists()` too fast as those docs were recommending to use `try_exists()` instead...but I didn't go into the reasons too much.

reubenmiller: Fixed. Since this is a one-liner change, I didn't bother with a fix up commit and just force pushed the change directly.

didier-wenzek (approved):
Approved

PR #3151, opened 2024-10-03
fix: add mosquitto as a weak dependency to tedge

reubenmiller:
## Proposed changes

Fix the tedge's linux package dependency to mosquitto by making it a weak dependency so that mosquitto is not installed by default when installing `tedge` directly (and not via the `tedge-full` or `tedge-minimal` virtual linux packages).

This will allow users to install tedge and any other package which depends on tedge (like tedge-agent) to be installed without mosquitto. The virtual packages will have a hard dependency to mosquitto as this is a "typical" setup and mosquitti is required on the main device.

The official `thin-edge.io/install.sh` script uses the virtual packages, so this does not break and existing behaviour.

FYI: Debian has [good documentation](https://www.debian.org/doc/debian-policy/ch-relationships.html) which go into details about the package dependency types.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/configuration/package_manifests/virtual/nfpm.tedge-minimal.yaml
+++ b/configuration/package_manifests/virtual/nfpm.tedge-minimal.yaml
@@ -26,16 +26,19 @@ overrides:
   # slightly different formats
   apk:
     depends:
+        - mosquitto
         - tedge = ${APK_VERSION}
         - tedge-mapper = ${APK_VERSION}
   rpm:
     depends:
         # FIXME: Work out a better way to reference the full package specific version which includes the release number (-1) suffix
         # without having to manually add it
+        - mosquitto
         - tedge = ${RPM_VERSION}-1
         - tedge-mapper = ${RPM_VERSION}-1
   deb:
     depends:
+        - mosquitto
         - tedge (= ${DEB_VERSION})
         - tedge-mapper (= ${DEB_VERSION})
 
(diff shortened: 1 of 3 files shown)


configuration/package_manifests/virtual/nfpm.tedge-minimal.yaml
```diff
@@ -26,16 +26,19 @@ overrides:
   # slightly different formats
   apk:
     depends:
+        - mosquitto
```
albinsuresh, line 29: I didn't quite get why we want to add this additional dependency to the `minimal` package. It was supposed to be `minimal`, right? It looks weird that `tedge-minimal` pulls in an extra dependency that regular `tedge` doesn't have, as it's supposed to be minimal than `tedge` itself.

reubenmiller: This isn't an additional dependency as previously when`tedge-minimal` was being installed it would also install `mosquitto` (as `mosquitto` was a dependency of `tedge`). So in order to not break things for people already using the `tedge-minimal`, it `mosquitto` was added to the `tedge-minimal` dependencies.

reubenmiller: Though I must admit the `tedge-minimal` is not so practical in my opinion...so we might eventually phase it out.

albinsuresh (approved):

PR #3155, opened 2024-10-04
Entity registration API v2 proposal

albinsuresh:
## Proposed changes

- [x] HTTP APIs for entity registration, de-registration and querying

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/design/decisions/0005-entity-registration-api.md
+++ b/design/decisions/0005-entity-registration-api.md
@@ -1,0 +1,461 @@
+# Entity registration API
+
+* Date: __2024-10-01__
+* Status: __Approved__
+
+## Background
+
+The existing MQTT entity registration API `v1` which expects entities to publish their registration messages
+as retained messages to their respective entity topic ids has some limitations.
+This design expects entities not to conflict with each other in terms of their topic ids or `@id` values.
+And since there is no direct feedback to these messages, the publishing entities need to just hope that it just works.
+This expectation might be okay in a very controlled environment where the person provisioning the devices in the field
+has complete control over all the devices to avoid any conflicts and even avoid any mistakes (conflicting ids).
+But, in environments where that is not feasible, the current model without feedback can have unpleasant side effects.
+
+For example, if an entity registers itself with topic id: `te/device/child01//` and `@id`: `child01`,
+a device twin with external id `child01` is created in the cloud and all of its data is routed to this twin.
+If another device accidentally registers itself using the same topic id, but with `@id`: `child02`,
+a second device twin is created in the cloud with the external id: `child02`
+and all the messages published to that topic by `child02` and even `child01` are routed to the new device twin.
+The `child01` twin would remain orphaned in the cloud, without the user noticing anything immediately, for lack of feedback.
+
+To avoid such issues, the registering entities must be made aware of any conflicts/errors immediately
+so that they don't proceed publishing any data to the wrong entity assuming that it succeeded.
+In this case, %%te&& must have rejected the second registration message.
+Any updates to an existing entity must have been done using dedicated entity update APIs.
+
+## Challenges
+
+* In the current architecture, the mapper handles all the registration messages and maintains the entity store.
+  So, in a deployment with multiple mappers, there would be multiple entity stores maintained by each mapper.
+  This results in unnecessary duplication of data and makes it error-prone with multiple sources of truth.
+  A central component like `tedge-agent` would be a better fit to store and maintain the entity store.
+* The current entity store also maintains the external id of devices, which is a cloud-specific aspect.
+  When the entity store is moved to the `tedge-agent`, this aspect must be detached and left with the mappers,
+  without the mappers having to maintain a duplicate entity store.
+* When the `tedge-agent` handles the registration of entities, it can't consult with the mappers to make sure that
+  the entities that it accepted locally are valid in the cloud as well.
+  If the local registration succeeds, but the cloud registration fails, the user must get that feedback as well.
+* A mapper might get connected much later, after a lot of local registrations are already processed.
+  Even in such cases, the mapper should be able to fetch all the entities that were registered till then,
+  and replicate that in the cloud.
+  Even in that case, any failed registrations must be conveyed to the user asynchronously.
+
+## Exclusions
+
+* Bulk updation or deletion of entities.
+  The same can be achieved with a query followed by performing individual update or delete on the result set.
+* Pagination of entity query results as the query results are typically small.
+  Even for a fleet of 1000s of devices, the results won't exceed a few 100 KBs.
+
+## Solution Proposal: REST APIs for registration
+
+* The `tedge-agent` maintains the entity store and provides REST APIs for registration, and querying.
+  The mappers and other components can query entity data from it when needed.
+* On every successful registration, `tedge-agent` publishes those registration messages to their respective topics,
+  so that other components likes mappers subscribed to those topics are notified of those new entities.
+* The `tedge-agent` can persist the entity store on disk, as it is done today, for faster recovery on restart.
+  It also helps in differentiating already seen entities from newly published ones.
+* The mappers no longer maintains the entire entity store but only maintains a simple mapping
+  between topic ids and their cloud external ids, so that mapped messages can be published to the same.
+* The mappers can rebuild their external id map on every restart, or even persist the same on disk for faster recovery.
+* When the registration of a local entity fails in the cloud for reasons like conflicting external ids,
+  the corresponding error message from the cloud is forwarded to `te/errors` topic so that the user is aware.
+
+### Create new entity
+
+**Endpoint**
+
+```
+POST /v1/entities
+```
+
+**Payload**
+
+```json
+{
+    "@topic-id": "device/child01//",
+    "@type": "child-device",
+    "@id": "child01",
+    "name": "child01",
+    "extra-fragment": {
+        "extra-key": "extra-value"
+    }
+}
+```
+
+**Example**
+
+```bash
+curl http://localhost:8000/tedge/entity-store/v1/entities \
+  -X POST \
+  -H "Content-Type: application/json" \
+  -d '{
+    "@topic-id": "device/child01//",
+    "@type": "child-device",
+    "name": "child01",
+    "extra-fragment": {
+        "extra-key": "extra-value"
+    }
+  }'
+```
+
+**Responses**
+
+* 201: Created
+  ```json
+  {
+      "@topic-id": "device/child01//"
+  }
+  ```
+* 409: Conflict
+  ```json
+  {
+      "error": "Entity with topic-id: 'device/child01//' already exists."
+  }
+  ```
+
+### Update existing entity
+
+Partial updates to an exiting entity, adding new fragments or removing existing fragments:
+
+**Endpoint**
+
+```
+PATCH /v1/entities/{topic-id}
+```
+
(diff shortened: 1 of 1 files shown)


design/decisions/0005-entity-registration-api.md
```diff
@@ -0,0 +1,266 @@
...
+      "error": "Entity with topic-id: 'device/child01//' already exists."
+  }
+  ```
+
+#### Update existing entity
+
+**Endpoint**
+
+```
+PUT /v1/entity/{topic-id}
```
reubenmiller, line 103: Since the `{topic-id}` includes forward slashes, this *may* cause some implementation problems (though it depends on the path rules on your HTTP handler from whatever library will be used). So it might be worth checking if that is possible first.

The alternative would be to force users to url encode the `{topic-id}`, e.g. `%2F` but that would be annoying to use.

albinsuresh: Yeah, I was expecting URL encoded values only. Isn't that widely used in query parameters and hence an accepted convention?

jarhodes314: URL encoding would be a _bit_ annoying to use, although I think this is a relatively simple case where `.replace("/", "%2F")` would generally suffice. I'm fairly sure allowing people to use non-escaped entity IDs would be technically possible, but that would prevent us from doing anything of the form `/v1/entity/{topic-id}/something-else` (or at least I would have reservations about doing that, even if it might be technically feasible).

design/decisions/0005-entity-registration-api.md
```diff
@@ -0,0 +1,266 @@
...
+a device twin with external id `child01` is created in the cloud and all of its data is routed to this twin.
+If another device accidentally registers itself using the same topic id, but with `@id`: `child02`,
+a second device twin is created in the cloud with the external id: `child02`
+and all the messages published to that topic by `child02` and even `child01` are routed to the new device twin.
+The `child01` twin would remain orphaned in the cloud, without the user noticing anything immediately, for lack of feedback.
+
+To avoid such issues, the registering entities must be made aware of any conflicts/errors immediately
+so that they don't proceed publishing any data to the wrong entity assuming that it succeeded.
+In this case, %%te&& must have rejected the second registration message,
+unless it was explicitly done to correct a mistake in the previous registration.
```
didier-wenzek, line 26: I would be explicit and add that "the API must distinguish requests creating an entity registration from those updating an entity registration".

design/decisions/0005-entity-registration-api.md
reubenmiller: The pain of having to enforce url encoding of the `{topic-id}` could be reduced via a tedge cli command, `tedge register device device/child01//`...then it can do the encoding

didier-wenzek, line 103: > PUT /v1/entity/{topic-id}

Technically, this should be a `POST` not a `PUT`. Indeed, looking at this example and the following for `GET`, it seems that the intent is to inject new fragments into an existing json resource. A `PUT` implies a full replacement of the resource content.

reubenmiller: Doing any url encoding (in the path) is super painful if you have to do it in bash...

```
rawurlencode() {
  local string="${1}"
  local strlen=${#string}
  local encoded=""
  local pos c o

  for (( pos=0 ; pos<strlen ; pos++ )); do
     c=${string:$pos:1}
     case "$c" in
        [-_.~a-zA-Z0-9] ) o="${c}" ;;
        * )               printf -v o '%%%02x' "'$c"
     esac
     encoded+="${o}"
  done
  echo "${encoded}"    # You can either set a return variable (FASTER) 
  REPLY="${encoded}"   #+or echo the result (EASIER)... or both... :p
}
```

reubenmiller: Or a slightly better one (taken from the same link as above):

```sh
ENCODED_TOPIC_ID=$(curl -s -o /dev/null -w %{url_effective} --get --data-urlencode "device/main//" localhost:1883 | cut -d'?' -f2-)

```

albinsuresh: If URL encoding is not a preferred option, then we can reconsider using entity id directly in the path.

> I'm fairly sure allowing people to use non-escaped entity IDs would be technically possible

It should be possible as we're doing the same in the file transfer service where we allow users to POST and GET file names of arbitrary length

> but that would prevent us from doing anything of the form /v1/entity/{topic-id}/something-else

Although I can't think of any API requirements that would add further path segments after the entity id, at least for the current requirements, I don't want to rule out future evolutions of the API that would require it. But, even in that case we have a way around it, as the entity topic IDs are fixed length (at most 4 path segments). So, even if we introduce such an API in future, we can take advantage of this fixed length. The only niggle there would be to use empty path segments for things like `device/main///` which may not be a commonly accepted convention.

I wasn't planning to enforce 4 segment entity IDs for the proposed APIs, and let users just provide just `device/main` instead of `device/main///` in the path, and the server logic to append any missing trailing `/`s to compute the entity topic id, to keep it simple. We can enforce the 4-segment length only when a future API enhancement absolutely requires it.

albinsuresh: Used `PATCH` instead of `POST` which seemed to be the convention for partial updates to a resource. Kept `PUT` as well for a complete update.

design/decisions/0005-entity-registration-api.md
albinsuresh: Resolved by c33517187d6c49c3e6de72f041ba53a73df0dc88

design/decisions/0005-entity-registration-api.md
albinsuresh: Resolved by 34a1e54269dcfaf12b3af4fb99f159b55206825a

design/decisions/0005-entity-registration-api.md
```diff
@@ -0,0 +1,398 @@
...
+      }
+  ]
+  ```
+
+#### Delete all child entities of an entity
+
+Delete all children of `device/child01//`:
+
+```
+DELETE /v1/entities?%40parent=device/child01//
```
albinsuresh, line 365: This is slightly unconventional, to `DELETE` a resource using query parameters. The conventional way would have been to use sub paths of the target resource:

```
DELETE /v1/entities/{topic-id}/children
```

But, I chose to avoid sub-paths due to the empty `/` characters in the `topic-id` as follows:

```
DELETE /v1/entities/device/child01///children
```

...which is not an acceptable convention either.

Another alternative would have been to provide a dedicated path with the target entity `topic-id` provided as a query parameter as follows:

```
DELETE /v1/entities/children?%40topic-id=device/child01//
```

But, I don't find it to be much of an improvement from the existing version either.

didier-wenzek: Do we really need a way to `DELETE` in one operation all the child-devices of given device?

Isn't a  child-devices query and a series one-by-one `DELETE`s enough?

design/decisions/0005-entity-registration-api.md
albinsuresh: > Although I can't think of any API requirements that would add further path segments after the entity id

That statement didn't age well :wink:. Already came across this requirement for bulk deletion of child entities of a given entity `topic-id`.

design/decisions/0005-entity-registration-api.md
reubenmiller: > Do we really need a way to `DELETE` in one operation all the child-devices of given device?
> 
> Isn't a child-devices query and a series of one-by-one `DELETE`s enough?

I would agree with @didier-wenzek. We don't need an explicit "delete all children" API command as we can do a query and then do individual api calls (we could also wrap the calls in a special tedge cli command to still do the same function but not require a specific API call).

design/decisions/0005-entity-registration-api.md
didier-wenzek: >> Although I can't think of any API requirements that would add further path segments after the entity id

> That statement didn't age well :wink:. Already came across this requirement for bulk deletion of child entities of a given entity `topic-id`.

I see things the other way around. Since its more convenient to have the entity topic identifier used as a suffix (to re-conciliate MQTT and HTTP naming rues), we can take this as a starting point for URL design, e.g. `/v1/entity/{topic-id}` This is also aligned with the REST principles where an URL identifies an entity and not some action.

design/decisions/0005-entity-registration-api.md
```diff
@@ -0,0 +1,398 @@
...
+  Even in that case, any failed registrations must be conveyed to the user asynchronously.
+
+## Solution Proposal
+
+### Proposal 1: REST APIs for registration
+
+* The `tedge-agent` maintains the entity store and provides REST APIs for registration, and querying.
+  The mappers and other components can query entity data from it when needed.
+* On every successful registration, `tedge-agent` publishes those registration messages to their respective topics,
+  so that other components likes mappers subscribed to those topics are notified of those new entities.
```
didier-wenzek, line 52: Does the agent also updates the entity store when a registration message is published by another component?

I would prefer not (to force the clients use the HTTP API), but this breaks backward compatibility.

If the agent also reacts on registration updates over MQTT, one must ensure the agent doesn't react on its own updates.

didier-wenzek (commented):
I'm aligned with what is proposed here:
- Move the responsibility of the entity store to the agent (to ensure consistency independently of the cloud mappers).
- Use HTTP instead of MQTT to update the entity store (to return status and error along a request/response protocol)
- Use MQTT to notify all the components of any entity registration update.

I vote against bulk updates and deletes. There is also no need for pagination.This API should focus on the content and its primary clients are scripts not users.

design/decisions/0005-entity-registration-api.md
albinsuresh: > Does the agent also updates the entity store when a registration message is published by another component?

We can take a stance that those using the HTTP APIs for registration must stick to it for all registrations and not combine both HTTP and MQTT based explicit registrations. I still expect many users continuing to use the auto-registration mechanism over MQTT. But, I'm assuming that a very small subset of users are currently using explicit registrations and we can ask them to migrate to the HTTP APIs instead.

reubenmiller: I don't think doing a breaking change will be acceptable here. Some plugins are already doing this (e.g. [tedge-container-plugin](https://github.com/thin-edge/tedge-container-plugin)). And I don't see a "nice" transition phase.

didier-wenzek: > I don't think doing a breaking change will be acceptable here.

So, the mapper has to:
- handle registration messages received over MQTT as `PATCH` requests,
- publish registration state over MQTT only after successful updates introducing an actual change.

design/decisions/0005-entity-registration-api.md
albinsuresh: Addressed by eb0eba88aa10427f4eb7ad70a0c04677202395b5

design/decisions/0005-entity-registration-api.md
albinsuresh: Addressed by eb0eba88aa10427f4eb7ad70a0c04677202395b5

design/decisions/0005-entity-registration-api.md
```diff
@@ -0,0 +1,461 @@
+# Entity registration API
+
+* Date: __2024-10-01__
+* Status: __New__
```
didier-wenzek, line 4:
```suggestion
* Status: __Approved__
```

didier-wenzek (approved):
Approved

PR #3174, opened 2024-10-08
feat: add profile support to tedge-mapper

jarhodes314:
## Proposed changes
This adds support to `tedge-mapper c8y` for different Cumulocity profiles, using the `#[tedge_config(multi)]` attribute added in #3126. The profile can be selected using either the `--profile` argument, or by setting the `C8Y_PROFILE` environment variable. Once it is configured with a profile, `tedge-mapper c8y` will set the `C8Y_PROFILE` variable for itself, ensuring that any child processes from custom operations etc. will use that c8y profile.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


didier-wenzek (commented):
What has been updated is correct, but the system tests highlight a missing fix for health checks. I will be happy to approve when fixed.

jarhodes314:
So I've now understood the healthcheck issue. The problem is that in the system test, `c8y.availability.interval` is set, but no other `c8y` keys are set. At the moment, this is erroneously causing the `c8y` config field to be deserialised as a named Cumulocity profile with the unknown property `c8y.interval` set, rather than as an unnamed Cumulocity with `c8y.availability.interval` set.

When we then run `tedge config set c8y.url`, this reads the config, decides that `c8y` is set to the default value and sets `c8y.url` before saving `tedge.toml`. Since the availability interval is not encoded in the DTO due to the incorrect deserialisation, this value is simply discarded.

So I'll try and come up with a resolution for this broken deserialisation logic. I think this issue also highlights an issue with the availability tests. We don't currently verify the `requiredInterval` property is actually set to the correct value for the device, and this made debugging the issue tricky (admittedly mostly due to my lack of familiarity with the feature not causing me to investigate that property sooner).

jarhodes314:
> So I've now understood the healthcheck issue. The problem is that in the system test, `c8y.availability.interval` is set, but no other `c8y` keys are set. At the moment, this is erroneously causing the `c8y` config field to be deserialised as a named Cumulocity profile with the unknown property `c8y.interval` set, rather than as an unnamed Cumulocity with `c8y.availability.interval` set.

I can think of two possible solutions to this issue:
1. Add `#[serde(deny_unknown_fields)]` to `TEdgeConfigDtoC8y`, `TEdgeConfigDtoAz`, etc. This would prevent us from coping with certain deprecated configurations, although we have automatically updated the relevant `tedge.toml` files since 0.12, although the toml should be preserved just fine for someone to manually fix it.
2. Give profile names some distinguishing feature (such as a prefix of `@`), so it is completely unambiguous whether we are attempting to use profiles are not.
3. Change the enum tagging in `MultiDto` so that profiled representations are externally tagged.

Examples of what the keys look like in these cases:
1. `c8y.cloud.url`, `c8y.cloud.availability.interval` (unchanged from right now)
2. `c8y.@cloud.url`, `c8y.@cloud.availability.interval` (`@` is just a suggestion, anything that isn't a valid rust identifier should work just fine for this method
3. `c8y.profiles.cloud.url`, `c8y.profiles.cloud.availability.interval`

As far as error handling is concerned (specifically in the case of deserialising a toml file, this shouldn't affect the `tedge config` CLI), I think both of 2 & 3 are easily compatible with the behaviour we have currently, warning on an unrecognised key rather than having a hard error, and reading pre 0.12 keys just fine. I'll do some more experiments tomorrow to try and more concretely prove this is the case. With option 1, the error messages are a lot less helpful, something along the lines of `data did not match any variant of untagged enum MultiDto`.

tests/RobotFramework/tests/cumulocity/availability/heartbeat.robot
```diff
@@ -22,6 +22,10 @@ Heartbeat is sent
     ...    The tests therefore relies on the backend availability service which then sets the c8y_Availability fragment
     ...    based on the last received telemetry data.
     Device Should Have Fragment Values
+    ...    c8y_RequiredAvailability.responseInterval\=1m
```
rina23q, line 25:
```suggestion
    ...    c8y_RequiredAvailability.responseInterval\=60
```

c8y accepts only seconds. At least, this caused the `Heartbeat is sent` test failure.

rina23q: This is wrong. Sorry! Corrected in https://github.com/thin-edge/thin-edge.io/pull/3174#discussion_r1796915753

didier-wenzek (approved):

crates/common/tedge_config_macros/src/multi.rs
```diff
@@ -120,11 +187,21 @@ impl<T> MultiReader<T> {
...
+                itertools::Either::Right(map.keys().map(|k| k.0.as_str()).map(Some))
+            }
+        }
+    }
+
+    pub fn keys(&self) -> impl Iterator<Item = Option<&ProfileName>> {
+        match self {
+            Self::Single { .. } => itertools::Either::Left(std::iter::once(None)),
+            Self::Multi { map, .. } => {
+                // itertools::Either::Right(map.keys().map(|k| k.0.as_str()).map(Some))
```
jarhodes314, line 203: Remove this

Issue #3171, opened 2024-10-07
custom operation output can exceed Cumulocity's max allowed payload size


reubenmiller:
**Describe the bug**

Custom Cumulocity operations which produce a large result (output) and makes heavy usage of double quotes (`"`) will result in publishing an MQTT message to Cumulocity IoT where the payload size exceeds the 16K limit.

This results in the operation being stuck in `EXECUTING` as the message is rejected by Cumulocity IoT and ultimately causes MQTT connection issues.

After reviewing the code, whilst the output of the custom operation is being truncated to `16000` bytes, the truncation is occurring before the CSV serialization. During the CSV serialization, the payload size can increase a single double quote (`"`) is expanded to two double quotes (`""`).

The following shows the sequence in which the custom operation output is processed:

1. Sanitize output by removing unsupported control characters
2. Truncate output to the first 16000 bytes
3. Convert to CSV <-- Problem step! Size increases from 16000 to `16000 + total number of double quotes`

**To Reproduce**

1. Install the `c8y-command-plugin` community plugin

2. Send a shell operation where more than ~600 double quotes are used (so that the total payload is greater than size limit (at least ~17K to be safe))

    ```sh
    yes 'hello"' | head -n 100000
    ```

    The above command will result in trying to publish an MQTT message to `c8y/s/us` where the payload is about `18303` bytes, which is above the Cumulocity IoT default MQTT payload size limit of ~16K.

3. Wait for the operation to be set to SUCCESSFUL. The bug is here, as the operation will be stuck in the `EXECUTING` state.

**Fixing after the test**

Since the test is destructive, to restore the system to normal function you have to do the following:

```sh
tedge disconnect c8y
systemctl stop mosquitto
rm /var/lib/mosquitto/mosquitto.db
tedge reconnect c8y
```

Then cancel the operation in Cumulocity IoT by using go-c8y-cli, where `{mydevice}` is the name or id of the device:

```sh
c8y operations list --device "{mydevice}" --fragmentType c8y_Command --status EXECUTING \
| c8y operations cancel
```

**Expected behavior**

The total payload of an operation should be truncated after the CSV serialization has occurred.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `any`
 - Hardware [incl. revision]: `any`
 - System-Architecture [e.g. result of "uname -a"]: `any`
 - thin-edge.io version [e.g. 0.1.0]: `1.3.1`

**Additional context**

[...]

PR #3179, opened 2024-10-09
fix: Trim `reason` field in "succeed custom operation" smartrest message

Bravo555:
## TODO
- [x] add test
- [x] consider how to trim CSV output
- [x] remove unwraps
- [x] Rust unit test to improve coverage

## Proposed changes

Trims `reason` field for sent "set operation to SUCCESSFUL" smartrest messages.

Additionally, `...<trimmed>` indicator is inserted at the end to let the user know that the output was trimmed.

## Types of changes

## Paste Link to the issue
- https://github.com/thin-edge/thin-edge.io/issues/3171

## Checklist

## Further comments

This PR solves the issue for operation success message for `c8y_Command`, but may still possibly fail for other types of messages.
For a follow-up we should do a clean up and use types to verify that message is smaller than max limit.

### DIFF
--- a/crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
+++ b/crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
@@ -96,6 +96,9 @@ fn succeed_static_operation(
 /// rules will be applied. For example, `a,field "with" quotes` will be converted to
 /// `a,"field ""with"" quotes"` before being appended to the output of this function.
 ///
+/// If the message size turns out to be bigger than maximum cumulocity message size, it will be truncated to fit in the
+/// limit.
+///
 /// # Errors
 /// This will return an error if the payload is a CSV with multiple records, or an empty CSV.
 pub fn succeed_operation(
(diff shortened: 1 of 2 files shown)


reubenmiller:
@Bravo555 It is probably enough just to serialize the result field and apply the constraints to that field before constructing the full csv response.

Bravo555:
> It is probably enough just to serialize the result field and apply the constraints to that field before constructing the full csv response.

Will use that approach, but is/should there be a way for the user to know that the response is truncated as opposed to not truncated but maximum allowed size?

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -96,19 +96,67 @@ fn succeed_static_operation(
...
+        let reason = reason.strip_prefix('"').unwrap_or(reason);
+        let reason = reason.strip_suffix('"').unwrap_or(reason);
+
+        // if we'd cut across an escaped " character, move trim point 1 char back to omit it
+        if &reason[max_result_limit - 1..=max_result_limit] == r#""""# {
+            max_result_limit -= 1;
+        }
+        let trimmed_reason = &reason[..max_result_limit];
+
+        format!("{prefix},\"{trimmed_reason}\"")
```
didier-wenzek, line 154: It would be good to add a hint that the reason has been truncated as suggested by @reubenmiller: https://github.com/thin-edge/thin-edge.io/pull/3179#issuecomment-2413316148

```suggestion
        format!("{prefix},\"{trimmed_reason} ...<truncated>\"")
```

didier-wenzek (commented):
The code is correct and thoroughly tested. I will be happy to approve once added the `...<truncated>` suffix.

tests/RobotFramework/tests/cumulocity/shell/shell_operation.robot
```diff
@@ -29,6 +29,11 @@ Failed shell command
     ${operation}=    Cumulocity.Execute Shell Command    exit 1
     Operation Should Be FAILED    ${operation}
 
+Shell command succeeds if output is too large
+    [Documentation]    Output should be trimmed by c8y mapper.
+    ${operation}=    Cumulocity.Execute Shell Command    yes 'hello"' | head -n 100000
+    Operation Should Be SUCCESSFUL    ${operation}
+
```
gligorisaev, line 37: After the operation succeeds, I would add a validation step to check the size of the output

Bravo555: I think the actual size is an implementation detail. From the user's perspective, it doesn't matter what the size of the message is, if they e.g. launched some command that generated massive output, but they care that the operation is marked SUCCESSFUL, even if the output is massive.
But as said in https://github.com/thin-edge/thin-edge.io/pull/3179#pullrequestreview-2373065149, a user might want to know if the message was truncated, so I'm happy to add an assert for that case.

gligorisaev: Something like 
```
${output_length}=    Get Length of Output    ${operation}
Should Be Less Than    ${output_length}    50000    # Assuming trimming occurs above 50,000 characters.
```

reubenmiller: > Something like
> 
> ```
> ${output_length}=    Get Length of Output    ${operation}
> Should Be Less Than    ${output_length}    50000    # Assuming trimming occurs above 50,000 characters.
> ```

This won't be possible as this is checking the operation in the json format will will be different to the length of the SmartREST 2.0 MQTT message. So doing this check won't yield any useful results.

The fact that operation is set to SUCCESSFUL is proof that the message is under Cumulocity's max MQTT payload size.

didier-wenzek (approved):
Approved.

reubenmiller (approved):
Approved

Issue #3156, opened 2024-10-04
support reloading workflows at runtime


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

With the upcoming feature to support mapping customer Cumulocity IoT operation to local thin-edge.io commands (see https://github.com/thin-edge/thin-edge.io/issues/3095), this will encourage more users to deploy new workflows to the devices. The typical way to package a new custom operation handlers, would be to package it has a linux package where the customer operation definition which typically consists of:

* custom cloud operation handler definition (this will be a new type) , .e.g. `/etc/tedge/operations/c8y/c8y_Command.template`
* workflow definition for the new functionality, .e.g. `/etc/tedge/operations/execute_shell.toml`
* Some custom script/binary which is called from the workflow, e.g. `/usr/bin/execute_shell.sh`

Deploying a package would be typically done at runtime using thin-edg.io's Software Management plugin (e.g. `tedge-apt-plugin`). However currently if a new workflow is created, the `tedge-agent` service must be restarted before it detects the new workflow that was installed by the linux package. Restarting the `tedge-agent` from within the Software Management plugin by way of [Maintainer Scripts](https://wiki.debian.org/MaintainerScripts) is not feasible as it would kill the script's parent process, which would result in the maintainer script also being killed.

**Describe the solution you'd like**

**tedge-agent** should support a way for users to tell the service to reload the workflows from disk.

A typical linux way to do this would be to support "tell" an application of new configuration would be to reload the configuration when receiving a `SIGHUP` signal. This way the user could control when the workflows are reloaded/re-read disk.

For example, the following command could be used to trigger a reload:

```sh
kill -s HUP <pid>
```

And by supporting `SIGHUP`, then it would enable the `tedge-agent` systemd service definition to support the `systemctl reload tedge-agent` syntax by adding the following field:

```ini
[Service]
ExecReload=/usr/bin/kill -HUP $MAINIP
```

**Open Questions**

* How should the current workflow react if it is changed whilst processing the current operation? Should an in-memory model be preserved for the "in-progress" workflow? And how would that effect the device profile, as it might be useful for users to install a new workflow and then use that workflow in the same device profile operation (as long as the new workflow is only affecting a subworkflow and not the device_profile itself)

**Describe alternatives you've considered**

[...]

Issue #2488, opened 2023-11-30
Enable users to configure location of tedge-agent current operation state folder (.agent)


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Doing firmware updates requires certain tedge-agent state to be persisted across partition swaps in order for the agent to know that it was in the middle of processing a firmware update operation (even before any kind of restore backup state can be executed).

This "current operation" state is stored in the `/etc/tedge/.agent` folder.

**Describe the solution you'd like**

Make the `/etc/tedge/.agent` folder configurable via the `tedge.toml`.

Below shows an example of the desired tedge.toml setting.

*file: /etc/tedge/tedge.toml*

```toml
agent.state.path = "/data/tedge/.agent"
```

Another slight variant of the above value would be to have a configurable fallback settings, which 

```toml
agent.state.persistent.path = "/data/tedge/.agent"
```

On startup, the tedge-agent would check if the persistent folder exists and if it does use that folder, otherwise use the default `/etc/tedge/.agent` (e.g. the path controlled via the `--config-dir` flag of the tedge-agent)

```
STATE_DIR=/etc/tedge/.agent
IF PATH_EXISTS(get_value(agent.state.persistent.path)) THEN
    STATE_DIR = get_value(agent.state.persistent.path)
END
```

The later would have the advantage that this setting would be baked into the defaults of the thin-edge.io agent making it slightly less open to misconfiguration, as the value has a sensible default value, where the behaviour can be modified by just creating the folder at /data/tedge/.agent (for example). Other misconfigurations would be very easy as the user could be changing some other settings in the tedge.toml file during the build process, and this would result in the firmware operation not resuming after a partition swap (with no possibility of a rollback happening without the assistance of an independent rollback service)

**Notes**

* For this to work the `agent.state.path` value needs to be set both before and after the firmware update. In practice this value will part of the initial firmware (OS) image and included in all subsequent firmware (OS) images, otherwise it will not work.

**Describe alternatives you've considered**

The current workaround is to configure the `/etc/tedge/.agent` folder to be a symlink to `/data/tedge/.agent` (where the `/data` is mounted to a persisted partition which is available from both A/B partitions). A symlink is hard to maintain as it also has to be based up in order to preserve it across firmware updates, and this is even more complicated to manage since the folder is mixed in with other important settings under the `/etc/tedge/` folder (e.g. operation definitions, child device info, sm-plugins etc.)

[...]

PR #3180, opened 2024-10-09
feat: support reloading workflows at runtime

didier-wenzek:
## Proposed changes

In order to support dynamic reloading of workflow, without breaking a running workflow,
the proposal is to:

- use a hash of the workflow definition file to distinguish workflow versions
- persist a copy of each version for a given operation
- make the copy when a command instance is triggered
- use reference-counting to remove copies when no more in-use.

Plan:

- [x] Attach a version to each workflow, this version being a hash of the definition file.
- [x] Attach a version to each operation instance, taking the version of the workflow definition in use.
- [x] Refactor the agent to group the workflow loading logic in a single place.
- [x] When a new operation instance is created, copy its definition in a persisted directory of workflows in-use. 
- [x] On-start load not only the user-provided workflow definition, but also those of the workflows in-use.
- [x] Reload workflow definitions on file change using inotify. 
- [x] On file change, reload the workflow definition (without erasing any definition still in-use).
- [x] When a user-defined workflow is removed, clear the associated capability (letting any still running operation run to completion).
- [x] When a builtin workflow specialization is removed, restore the builtin version if any.
- [x] When a workflow definition is removed, keep a copy of that definition if still in-use.  

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3156

## Checklist


crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -0,0 +1,114 @@
...
+        }
+    }
+}
+
+async fn read_operation_workflow(
+    path: &Path,
+) -> Result<(OperationWorkflow, WorkflowVersion), anyhow::Error> {
+    let bytes = tokio::fs::read(path).await.context("Fail to read file")?;
+    let input = std::str::from_utf8(&bytes).context("Fail to extract UTF8 content")?;
+    let version = sha256::digest(input);
```
didier-wenzek, line 101: Using `md5` should be enough here - as there is no crypto concerns.

gligorisaev (commented):
The Test is testing the described feature
# 📊 **Consolidated Test Summary for 10 Runs**


### 🛠️ **Test Results Summary - Run 1**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:33:18  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 37.56 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 2**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:33:55  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 34.62 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 3**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:34:31  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 34.70 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 4**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:35:08  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 35.52 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 5**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:35:43  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 32.91 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 6**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:36:19  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 34.90 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 7**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:36:55  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 34.91 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 8**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:37:32  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 34.53 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 9**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:38:10  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 36.98 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

### 🛠️ **Test Results Summary - Run 10**

**Test Suite**: Dynamic Workflow Reloading  
**Date**: 2024-10-17 14:38:49  
**Total Tests**: 7  
**Passed**: 7  
**Failed**: 0  
**Duration**: 37.12 seconds  
**Success Rate**: 100.00%

#### 📄 **Test Details**:
- Test: Create User-Defined Operation - Status: PASSED
- Test: Update User-Defined Operation - Status: PASSED
- Test: Remove User-Defined Operation - Status: PASSED
- Test: Updating A Workflow Twice Before Using It - Status: PASSED
- Test: Override Builtin Operation - Status: PASSED
- Test: Recover Builtin Operation - Status: PASSED
- Test: Trigger Workflow Update From A Main Workflow - Status: PASSED

crates/core/tedge_api/src/workflow/supervisor.rs
```diff
@@ -229,6 +298,152 @@ impl WorkflowSupervisor {
...
+/// - The current version, i.e. the version to be used for new operation instances.
+/// - One in-use version, that was used by some operation instance when the current version has been updated.
+/// - The builtin version provided by the agent and which can be overridden by the users.
+///
+/// None of these versions are mandatory. For instance, an operation can have no builtin version,
+/// no current version (because the user removed the definition file), but a version still in-use
+/// (that has been started before the user deprecated the operation).
+struct WorkflowVersions {
+    operation: OperationName,
+    current: Option<WorkflowVersion>,
```
albinsuresh, line 314: While reviewing the code, I was often getting `current` and `in-use` confused and mixed up, esp with `current` as it could be wrongly interpreted as the "current one being used". Calling it something like `latest` might help.

crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -0,0 +1,405 @@
...
+                .register_builtin_workflow(capability.as_str().into())
+            {
+                error!("Fail to register built-in workflow for {capability} operation: {err}");
+            }
+        }
+    }
+
+    /// Update the workflow definitions after some on-disk changes
+    ///
+    /// Return the operation name when the operation has been deprecated.
```
albinsuresh, line 162:
```suggestion
    /// Return the operation capability deregistration message when the operation has been deprecated.
```

reubenmiller:
> The Test is testing the described feature
> 
> Tested on RPI 4:
> 
> # 📊 **Consolidated Test Summary for 10 Runs**


@gligorisaev it would be more efficient to use the flakefinder task, and just post the summary from the console output (just the last 3 lines is fine). Posting the results of all the tests is unnecessary (duplicate information) and just clouds the ticket. In the future we could even look at creating a CI workflow which can do on-demand flake finder runs on specific suites or tagged tests.

By using flakefinder, it is easier for other people to run to check that any newly added suite is (hopefully) flake-free. If there is a detected "flake" then the developer will also need to know how to re-run the same tests to help them with debugging.

In this case, running the following command did the same thing as you did (though this was run inside a container):

```sh
invoke flake-finder --suite "dynamic_workflow_reloading" --iterations 10 --outputdir output_ff --clean
```

**Output (only the last 3 lines)**

```sh
Overall: PASSED
Results: 10 iterations, 10 passed, 0 failed
Elapsed time: 0:05:46.354385
```

albinsuresh (commented):
LGTM. Just a few minor suggestions on possible simplifications (if feasible).

crates/core/tedge_agent/src/state_repository/state.rs
```diff
@@ -17,30 +17,41 @@ pub struct AgentStateRepository<T> {
...
+    // Check that the given directory is actually writable, by creating an empty test file
+    let test_file = state_dir.join(state_dir.join(".--test--"));
+    match File::create(test_file.clone()).and_then(|mut file| file.write_all(b"")) {
+        Ok(_) => {
+            let _ = std::fs::remove_file(test_file);
+            state_dir
+        }
+        Err(err) => {
+            warn!("Cannot use {state_dir:?} to store tedge-agent state: {err}");
+            agent_default_state_dir(tedge_root)
```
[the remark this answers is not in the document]
didier-wenzek: This directory is critical for A/B firmware update.

See the rational here: https://github.com/thin-edge/thin-edge.io/issues/2488

crates/core/tedge_api/src/workflow/supervisor.rs
didier-wenzek: Unfortunately, `latest` is used for the on-disk version.

- `current` is the version currently loaded by the workflow engine.
- `in-use` is a copy that is kept for a running workflow in case the agent restarts while the operation is still under execution. On restart that version must be used and not a new one.

crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -0,0 +1,405 @@
...
+    state_dir: Utf8PathBuf,
+
+    // Map each workflow version to its workflow file
+    //
+    // For a fresh new workflow definition, this points to the user-defined file
+    // When the workflow definition is in use, this points to a copy in the state directory.
+    definitions: HashMap<WorkflowVersion, (OperationName, Utf8PathBuf)>,
+    in_use_copies: HashMap<OperationName, WorkflowVersion>,
+
+    // The in-memory representation of all the workflows (builtin, user-defined, i,n-use).
```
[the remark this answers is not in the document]
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3180/commits/f0e154c13fcc71a4b63c7c48cb7cca15c13481e8

crates/core/tedge_agent/src/operation_workflows/persist.rs
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3180/commits/3ecdaa6bc4fe405ae93ef4871eeda5a1a17dc23e

crates/core/tedge_api/src/workflow/supervisor.rs
```diff
@@ -229,6 +298,152 @@ impl WorkflowSupervisor {
...
+    builtin: Option<WorkflowVersion>,
+    in_use: Option<WorkflowVersion>,
+    versions: HashMap<WorkflowVersion, OperationWorkflow>,
```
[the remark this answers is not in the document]
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3180/commits/e512dd77a77bd45566ca0b755c6abebb0b559e8a

Thank @albinsuresh, this was a really neat suggestion.

crates/core/tedge_api/src/workflow/supervisor.rs
```diff
@@ -229,6 +298,116 @@ impl WorkflowSupervisor {
     }
 }
 
+/// A stack of known versions for a workflow
+///
+/// In practice, one might have 3 concurrent versions:
+///
+/// - The current version, i.e. the version to be used for new operation instances.
+/// - One in-use version, that was used by some operation instance when the current version has been updated.
+/// - The builtin version provided by the agent and which can be overridden by the users.
```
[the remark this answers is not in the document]
albinsuresh: Since we don't allow multiple operations of the same type to run concurrently, I don't think this can happen.

albinsuresh (approved):
LGTM.

I still have a feeling that the `@version` field and the associated lookup logic could still be simplified. As far as I understood, the SHA256 version value doesn't really have any relevance. The value `builtin` implies built-in operation. The absence of value implies the latest version on disk. So, any other value implies a cached `in-use` copy at `/etc/tedge/.agent`. So, I'm wondering why not just use some static value like `in-use` or `cached` to represent the same? That simplifies the lookup into the `WorkflowVersions` struct as well. WDYT?

crates/core/tedge_api/src/workflow/supervisor.rs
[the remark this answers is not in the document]
didier-wenzek: > we don't allow multiple operations of the same type to run concurrently

Yes, this is the assumption on which this work is based upon.

But, I'm afraid that @Bravo555 raised a valid concern.

- For the builtin operations, the commands are processed one after the other and the assumption is correct.
- But for user-defined operations, there is no mechanism to block a new command till the previous is done.
- A command instance topic cannot be re-used, but the two instances can run concurrently for the same command.
- => I will add a system test for @Bravo555 scenario.
- => I will have to fix the workflow engine to prevent concurrent runs for a given command.

didier-wenzek:
There are two ways to solve the issue of concurrent in-use versions for an operation workflow:
1. Enforce that the operation instances for a given operation run sequentially
2. Revisit this PR to cope with several versions concurrently in use for the same operation.

We will go with the first option.

Indeed, in a near future, the plan is to give the users a way to control the concurrency of running operations.
The idea is to have too kind of operations (*update* and *request*);
so at a given time at most one *update* operation could be running
and with no limit on the number of *request* operations.

Hence, the simpler is to revisit this PR to cope with several versions concurrently in use for the same operation
Concurrency control will be then added on top of that to prevent concurrent update operations.

reubenmiller:
> There are two ways to solve the issue of concurrent in-use versions for an operation workflow:
> 
> 1. Enforce that the operation instances for a given operation run sequentially
> 2. Revisit this PR to cope with several versions concurrently in use for the same operation.
> 
> We will go with the first option.
> 
> Indeed, in a near future, the plan is to give the users a way to control the concurrency of running operations. The idea is to have too kind of operations (_update_ and _request_); so at a given time at most one _update_ operation could be running and with no limit on the number of _request_ operations.
> 
> Hence, the simpler is to revisit this PR to cope with several versions concurrently in use for the same operation Concurrency control will be then added on top of that to prevent concurrent update operations.

Here is the link to the roadmap item to control the execution of the operations (e.g. either serial or concurrent): https://github.com/thin-edge/thin-edge.io-roadmap/issues/69

crates/core/tedge_api/src/workflow/supervisor.rs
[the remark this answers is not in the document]
didier-wenzek: Finally, I removed this commit.

To maintain multiple concurrent versions, yet another design was required.

See https://github.com/thin-edge/thin-edge.io/pull/3180/commits/d0cd0ebdfcdc46f960f81fc64b3da629ffb272b6

didier-wenzek:
Here is my proposal to handle concurrent commands running the same operation but with different workflow versions.

https://github.com/thin-edge/thin-edge.io/pull/3180/commits/d0cd0ebdfcdc46f960f81fc64b3da629ffb272b6

I still have to:
- remove cached versions when no more in use
- improve the tests: there is still a race condition there

crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -264,18 +263,15 @@ impl WorkflowRepository {
...
-        if let Some(in_use_version) = self.in_use_copies.get(operation) {
-            if in_use_version == version {
-                return;
-            }
+        if self.in_use_copies.contains(version) {
+            return;
         };
         if let Some((_, source)) = self.definitions.get(operation) {
-            let target = self.state_dir.join(operation).with_extension("toml");
+            let target = self.state_dir.join(version).with_extension("toml");
```
albinsuresh, line 270: How about using a combination of `operation` and `version` in the file name so that we're not looking at a random hash number during debugging?

tests/RobotFramework/tests/tedge_agent/workflows/long-running-command-v1.toml
```diff
@@ -7,7 +7,7 @@ on_success = "executing"
 [executing]
 script = "sleep ${.payload.duration}"
 on_success = "logging"
-on_error = "logging"
+on_kill = "executing"
```
albinsuresh, line 10: The intent of looping back to `executing` state is to resume the operation from here after the agent restart?

crates/core/tedge_agent/src/operation_workflows/persist.rs
didier-wenzek: A `grep` should make it, but why not.

tests/RobotFramework/tests/tedge_agent/workflows/long-running-command-v1.toml
didier-wenzek: The intent is to be sure the command is not finalized *before* the agent stops.

Currently, the actors are stopped in a random order, and I observed a run where the script was killed before the workflow engine, the latter being fast enough to move the operation forward before being stopped.

crates/core/tedge_api/src/workflow/supervisor.rs
```diff
@@ -185,13 +186,16 @@ impl WorkflowSupervisor {
             });
         };
 
-        let version = command_state.workflow_version();
+        let Some(version) = &command_state.workflow_version() else {
+            return Err(WorkflowExecutionError::MissingVersion);
+        };
```
didier-wenzek, line 191: I have to revert this change. Indeed, this might be a source of issue if the agent is updated whilst there is a pending operation. If the previous agent was not featuring command versions, then the resumed operation should not be rejected by the new agent: the current version of the workflow must be used.

crates/core/tedge_agent/src/operation_workflows/persist.rs
didier-wenzek: Done along https://github.com/thin-edge/thin-edge.io/pull/3180/commits/ae56d97652643352a7e94e2b9ee73b93da3e6df6

crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -290,6 +297,22 @@ impl WorkflowRepository {
         }
     }
 
+    async fn release_in_use_copy(&mut self, operation: &OperationName, version: &WorkflowVersion) {
+        if let Some(count) = self.in_use_copies.get_mut(version) {
+            *count -= 1;
+            if *count == 0 {
+                return;
+            }
```
didier-wenzek: Oops! Fixed by https://github.com/thin-edge/thin-edge.io/pull/3180/commits/431fa64457901d5242e3729303ba4fefc979caf1

crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -74,7 +71,7 @@ impl WorkflowRepository {
         // First, all the user-defined workflows are loaded
         let dir_path = &self.custom_workflows_dir.clone();
         if let Err(err) = self
-            .load_operation_workflows(WorkflowSource::UserDefined, dir_path)
+            .load_operation_workflows(WorkflowSource::UserDefined(dir_path))
```
albinsuresh, line 74: Opinion: Although I understand that `WorkflowSource` being a generic type allows such generic usages, but this one with the directory path seems slightly overloaded. I agree that we can still defend it since the directory roots are also linked to the source type.

crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -289,6 +297,22 @@ impl WorkflowRepository {
         }
     }
 
+    async fn release_in_use_copy(&mut self, operation: &OperationName, version: &WorkflowVersion) {
+        if let Some(count) = self.in_use_copies.get_mut(version) {
+            *count -= 1;
+            if *count == 0 {
```
albinsuresh:
```suggestion
            if *count != 0 {
```
We should proceed with the removal only when the count become `0`, right?

tests/RobotFramework/tests/tedge_agent/workflows/long-running-command-v1.toml
```diff
@@ -2,12 +2,16 @@ operation = "long-running-command"
 
 [init]
 action = "proceed"
-on_success = "executing"
+on_success = "scheduled"
+
+[scheduled]
+operation = "sleep"
```
albinsuresh, line 8:
```suggestion
operation = "sleep"     # Using a sub-operation for sleep to allow concurrent execution of multiple versions of this workflow. Using sleep command directly would have blocked each workflow, resulting in serial execution.
```

albinsuresh (approved):
LGTM.

tests/RobotFramework/tests/tedge_agent/workflows/long-running-command-v1.toml
didier-wenzek: fixed by https://github.com/thin-edge/thin-edge.io/pull/3180/commits/c546f423b4220fcebcd8aad11b99c3658d5c584e

tests/RobotFramework/tests/tedge_agent/workflows/dynamic_workflow_reloading.robot
```diff
@@ -0,0 +1,250 @@
...
+    Should Have MQTT Messages
+    ...    te/device/main///cmd/software_list/dyn-test-4
+    ...    message_pattern=.*successful.*
+    ${workflow_log}    Execute Command    cat /var/log/tedge/agent/workflow-software_list-dyn-test-4.log
+    Should Contain
+    ...    ${workflow_log}
+    ...    item="@version":"76e9afe834b4a7cadc9029670ba76745fcda73784f9e78c09f0c0416f7f58ad2"
+
+Recover Builtin Operation
+    ThinEdgeIO.File Should Exist    /etc/tedge/operations/software_list.toml
```
Bravo555, line 88: suggestion: the test fails if run by itself and not part of the test suite, because this file is created by the previous test case. Could we instead use Transfer to Device everywhere so that there are no dependencies between the test cases? I'd expect Transfer to Device to overwrite the file if it already exists, so it should be okay to use that?

crates/core/tedge_api/src/workflow/supervisor.rs
```diff
@@ -229,6 +298,138 @@ impl WorkflowSupervisor {
     }
 }
 
+/// The set of in-use workflow versions for an operation
+///
+/// - The current version is the version that will be used for a new command instance.
+/// - The current version might be none. This is the case when the command has been deprecated.
+/// - When a new command instance is initialized, the current version is stored as being in use.
+/// - When all the commands using a given version are finalized, these copies are removed (TODO).
```
Bravo555, line 306: TODO: does ae56d97652 implement removal mentioned here? If so, `TODO` can be removed

crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -0,0 +1,478 @@
+use anyhow::Context;
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+use log::error;
```
Bravo555, line 4:
```suggestion
use tracing::error;
```

crates/core/tedge_agent/src/operation_workflows/persist.rs
Bravo555, line 1: thought: this module has quite a bit of functionality but doesn't have any unit tests - codecov reports 210 missed lines and 34.2% patch coverage (from other workflow-related tests)

Bravo555 (approved):
Some nits, but LGTM overall.

crates/core/tedge_api/src/workflow/supervisor.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3180/commits/3b7d30eb8d3a38f0efad50b2a650816078a66b0f

crates/core/tedge_agent/src/operation_workflows/persist.rs
didier-wenzek: Fixed https://github.com/thin-edge/thin-edge.io/pull/3180/commits/41f4208c93e8d51f8e43ee6dd95994d64472143d

crates/core/tedge_agent/src/operation_workflows/persist.rs
didier-wenzek: I can only acknowledge that the *unit* test coverage is poor. However, this code is quit extensively tested by the system test suite added by this PR (Dynamic Workflow Reloading). I opted for system-tests instead of unit-tests because the features introduced by this module are heavily related to the file system and inotify as well as sequence of user actions (adding/updating/removing files while the agent is running/restarted. One place where unit tests can be improved is the `tedge_api::workflow::supervisor` module which provide the in-memory representation of the uploaded workflow definitions.

tests/RobotFramework/tests/tedge_agent/workflows/dynamic_workflow_reloading.robot
didier-wenzek: I have a mix feeling here. On one side, you are correct, it would be handy to have independent tests. But, on the other side, this test suite represents well a scenario where a user creates and iterate updating a workflow file.

Concretely, replacing`File Should Exist` assertion by a `Transfer to Device` command would lead to a *different* test while running the suite vs the isolated test. Indeed, in the suite case, one checks that a user can update a workflow while, in the isolated case, one checks that the user can create a workflow (i.e. `Update User-Defined Operation` doing the same test as `Create User-Defined Operation`).

Issue #3185, opened 2024-10-11
mosquitto bug causes mqtt session persistence to be non-functional


reubenmiller:
**Describe the bug**

Due to this [bug in mosquitto](https://github.com/eclipse/mosquitto/issues/2618), persistent MQTT sessions don't work when using mosquitto `> 2.0.11 <= 2.0.18` and the `per_listener_settings true` setting (which is the default in thin-edge.io).

For thin-edge.io, it means that publishing messages to the local MQTT broker whilst the `tedge-mapper-c8y` is down will be lost, as the mapper does not receive these messages when it connects to the broker.

Related mosquitto bugs:

* https://github.com/eclipse/mosquitto/issues/2618
* https://github.com/eclipse/mosquitto/issues/2526

**To Reproduce**

The behaviour is easy to reproduce

1. Install any mosquitto version from 2.0.12 - 2.0.18 (inclusive)

2. Configure thin-edge.io and connect it to Cumulocity (as this will add the problematic `per_listener_settings true` mosquitto setting)

3. On the device execute the following script to 

    ```sh
    systemctl stop tedge-mapper-c8y
    tedge mqtt pub -q 1 te/device/main///e/test "{\"text\":\"hello $(date +%s)\"}"
    systemctl start tedge-mapper-c8y
    ```

    Alternatively you can reproduce the exact mosquitto MQTT broker issue by using `mosquitto_sub` and `mosquitto_pub` directly. The following one-liner will print "PASSED" or "FAILED" at the end to show if the test was successful or not.

    ```sh
    mosquitto_sub -h 127.0.0.1 -c -i subscriber1 -t topic1 --qos 1 -W 2; mosquitto_pub -h 127.0.0.1 -t topic1 -m test --qos 1; sleep 1; mosquitto_sub -h 127.0.0.1 -c -i subscriber1 -t topic1 --qos 1 -C 1 -W 2 && echo PASSED || echo FAILED;
    ```

**Expected behavior**

thin-edge.io MQTT messages published with QoS 1 or 2 published when the mapper is not active should be published to the cloud once it starts (and the bridge is online).

Note: this is an upstream bug in mosquitto, so we still have to see how to address this from thin-edge.io.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`unknown`|
[...]

PR #3187, opened 2024-10-15
test: check publishing of telemetry data whilst mapper is down

gligorisaev:
This PR adds a new test case to validate the behavior of ThinEdge.io when the tedge-mapper-c8y service is down, and events are published during the downtime. The test ensures that events sent via MQTT while the mapper is offline are published to Cumulocity when the service is restored.

Test Flow:

1.	Stop the tedge-mapper-c8y service.
2.	Publish an MQTT event with a random event type and a test message.
3.	Restart the tedge-mapper-c8y service.
4.	Assert that the event is correctly received and processed by Cumulocity, verifying both the event type and message content.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3185
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/mqtt/offline_publishing.robot
+++ b/tests/RobotFramework/tests/mqtt/offline_publishing.robot
@@ -1,0 +1,24 @@
+*** Settings ***
+Resource            ../../resources/common.resource
+Library             ThinEdgeIO
+Library             Cumulocity
+
+Test Setup          Setup Test
+Test Teardown       Get Logs
+
+
+*** Test Cases ***
+Publish event whilst mapper is down
+    [Documentation]    The mapper should publish the event to the cloud when it comes back online
+    ${event_type}=    Get Random Name
+    Stop Service    tedge-mapper-c8y
+    Execute Command    tedge mqtt pub -q 1 te/device/main///e/${event_type} '{"text":"test message"}'
+    Start Service    tedge-mapper-c8y
+    Cumulocity.Device Should Have Event/s    type=${event_type}    expected_text=test message
+
+
+*** Keywords ***
+Setup Test
+    ${DEVICE_SN}=    Setup
+    Set Suite Variable    $DEVICE_SN
+    Device Should Exist    ${DEVICE_SN}
--- a/tests/images/debian-systemd/debian-systemd.dockerfile
+++ b/tests/images/debian-systemd/debian-systemd.dockerfile
@@ -20,17 +20,24 @@ RUN apt-get -y update \
     net-tools \
     # json tools (for tests)
     jq \
-    jo
+    jo \
+    # mosquitto (default version used by Debian, see below for more details)
+    mosquitto \
+    mosquitto-clients
 
+# Note: Avoid using mosquitto 2.0.18 due to a session persistence bug when using `per_listener_settings true
+# Only comment out the custom install logic to make it easier to re-enable once the bug is resolved
+# See https://github.com/thin-edge/thin-edge.io/issues/3185 for more details
 # Install more recent version of mosquitto >= 2.0.18 from debian backports to avoid mosquitto following bugs:
 # The mosquitto repo can't be used as it does not included builds for arm64/aarch64 (only amd64 and armhf)
 # * https://github.com/eclipse/mosquitto/issues/2604 (2.0.11)
 # * https://github.com/eclipse/mosquitto/issues/2634 (2.0.15)
-RUN sh -c "echo 'deb [signed-by=/usr/share/keyrings/debian-archive-keyring.gpg] http://deb.debian.org/debian bookworm-backports main' > /etc/apt/sources.list.d/debian-bookworm-backports.list" \
-    && apt-get update \
-    && DEBIAN_FRONTEND=noninteractive apt-get -y --no-install-recommends install -t bookworm-backports \
-        mosquitto \
-        mosquitto-clients
+# * https://github.com/eclipse/mosquitto/issues/2618 (2.0.18)
+#RUN sh -c "echo 'deb [signed-by=/usr/share/keyrings/debian-archive-keyring.gpg] http://deb.debian.org/debian bookworm-backports main' > /etc/apt/sources.list.d/debian-bookworm-backports.list" \
+#    && apt-get update \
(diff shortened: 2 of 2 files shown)


tests/RobotFramework/tests/mqtt/mosquitto_bug_check.robot
```diff
@@ -0,0 +1,38 @@
...
+Publish an MQTT message with QoS 1
+    Execute Command    tedge mqtt pub -q 1 te/device/main///e/test "{\\"text\\":\\"hello $(date +%s)\\"}"
+
+Start tedge-mapper-c8y
+    Execute Command    sudo systemctl start tedge-mapper-c8y.service
+
+Check result
+    ${result} =    Execute Command    mosquitto_sub -h 127.0.0.1 -c -i subscriber1 -t topic1 --qos 1 -W 2; mosquitto_pub -h 127.0.0.1 -t topic1 -m test --qos 1; sleep 1; mosquitto_sub -h 127.0.0.1 -c -i subscriber1 -t topic1 --qos 1 -C 1 -W 2 && echo PASSED || echo FAILED;
+    ${trimmed_result} =    Remove Extra Spaces And Newlines    ${result}
+    Log Result    ${trimmed_result}
```
reubenmiller, line 23: Just create 1 test case rather the splitting the test setup into individual test cases, and don't bother testing for the local mqtt persistence, just check if the message is delivered to the cloud or not (as this is the user contract).

So the (single) test case should do:

1. Stop the tedge-mapper-c8y
2. Publish an event
3. Start the tedge-mapper-c8y
4. Check that the event is visible in the cloud

tests/RobotFramework/tests/mqtt/mosquitto_bug_check.robot
didier-wenzek, line 1: This test suite has to be removed. The test is green but for the wrong reason: the twin event that exists on c8y is not the event that has just been published.

=> One way to avoid such name clashes is to come with fresh new names for each test case. E.g. '{ "text": "Event sent when stopped" }'

reubenmiller:
@gligorisaev some general comments/improvements:
* Please use **fixup** commits when addressing comments explicitly related changes/problems to previously committed code (this helps keep the git commit history clean, as your second commit does not add any valuable information)

* The naming of the test suite and the filename does not suit the contract that we want to provide to users. e.g. we shouldn't be explicitly checking for the mosquitto bug, we should be checking that thin-edge.io does not lose data if a client publishes qos1 or 2 messages whilst the mapper is down (notice the commit that I pushed https://github.com/thin-edge/thin-edge.io/pull/3187/commits/54b7b0c425babf3b8ffffb36bc94a29c60f5322f)
* Prefer using **Test Setup** over **Suite Setup** unless you have a really good reason to use **Suite Setup**
* Use unique identifiers when publishing test data. This helps reduce unexpected mistakes/behaviour if whatever assertion that you're doing looks beyond the expected scope (e.g. it looks at all events in the platform instead of events for the given device)...this is the reason why your test is passing (when it shouldn't), and my test is failing (as it should)
* Use existing keywords where possible, e.g. prefer using `Stop Service` instead of manually running `sudo systemctl stop ...`. The existing library keywords help the tests to be abstracted so that we can support other service managers using the same tests (e.g. SysVinit etc.)
* Prefix the library on specific keywords where the meaning matters, for example `Cumulocity.Device Should Have Event/s` is more clear where the events are meant to be located (e.g. in Cumulocity and not on the local MQTT broker)
* Keep the PR description up to date to reflect the purpose of the changes. The purpose of the ticket is to add coverage for offline publishing (not to test a mosquitto bug)...the mosquitto bug aspect just was the motivation to add such a test...if we had had the system test prior to upgrading to mosquitto 2.0.18, the test would of shown us that 2.0.18 has an issue (ahead of time before we used it in all of our images; system test image, rugpi image, yocto image etc.)

reubenmiller:
@gligorisaev 
* Always put the linked ticket under the **Paste Link to the issue** section. You can also mention the ticket link in the description as well (if provides additional context), however having the links all in the same place makes it easier to read all the PRs and the content is normalized.

reubenmiller (changes requested):
@gligorisaev Can you remove the following commits from the PR as they have been superseded by 54b7b0c425babf3b8ffffb36bc94a29c60f5322f:

* 84bca7153ba02ff67b85add3f539ee4613856095
* 820755e0bb25346dafe7913850d23a2946d2a3b6

And can you please update the PR description to suite (as per comments):

* https://github.com/thin-edge/thin-edge.io/pull/3187#issuecomment-2418765325
* https://github.com/thin-edge/thin-edge.io/pull/3187#issuecomment-2418776651

reubenmiller (dismissed):
@gligorisaev Thanks for dropping the two other commits, but now the commit I pushed yesterday to switch back to using mosquitto 2.0.11 is missing which means the test is failing (as the system tests are still using the bugging mosquitto 2.0.18 version)

gligorisaev:
> @gligorisaev Thanks for dropping the two other commits, but now the commit I pushed yesterday to switch back to using mosquitto 2.0.11 is missing which means the test is failing (as the system tests are still using the bugging mosquitto 2.0.18 version)

I did rebase and choose drop of my commits only (made also screenshot of it)

tests/RobotFramework/tests/mqtt/mosquitto_bug_check.robot
reubenmiller: The test has been refactored to use a unique event type, 4d9741de6d59d498e1300521407ccd3d1098b970

reubenmiller (commented):
I've removed my review from this PR as the commits are authored by myself, so I shouldn't be the one approving it.

reubenmiller:
> > @gligorisaev Thanks for dropping the two other commits, but now the commit I pushed yesterday to switch back to using mosquitto 2.0.11 is missing which means the test is failing (as the system tests are still using the bugging mosquitto 2.0.18 version)
> 
> I did rebase and choose drop of my commits only (made also screenshot of it)

Then something went very wrong (maybe you didn't do a fetch and pull prior to rebasing). But anyway I've forced pushed the changes that I still had locally.

didier-wenzek (approved):
Approved

Issue #3095, opened 2024-09-02
Use workflows for custom operation handlers


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
The custom operation is currently limited to SmartREST input and main device. Since we have workflow feature, mapper can focus on generating a thin-edge command, and move the actual execution of the operation to tedge-agent. 

**Describe the solution you'd like**
Mapper provides a way for users how to map the c8y JSON to thin-edge JSON.

Let's use this custom operation **`com_cumulocity_model_WebCamDevice`** as an example.

### C8Y JSON input
All operations including custom operations will be delivered to the JSON over MQTT topic `c8y/devicecontrol/notifications`.

```json
{
   "delivery":{
      "log":[
         
      ],
      "time":"2024-09-02T19:56:01.342Z",
      "status":"PENDING"
   },
   "agentId":"201802315",
   "creationTime":"2024-09-02T19:56:01.313Z",
   "deviceId":"201802315",
   "id":"1800380",
   "status":"PENDING",
   "com_cumulocity_model_WebCamDevice":{
      "name":"take picture",
      "parameters":{
         "duration":"5s",
         "quality":"HD"
      }
   },
   "description":"WebCam picture",
   "externalSource":{
      "externalId":"TST_haul_searing_set",
      "type":"c8y_Serial"
   }
}
```

### Drop 1
`/etc/tedge/operations/c8y/com_cumulocity_model_WebCamDevice` (for main device)

The mapper spawns a child process (= same mechanism as the existing custom operation handler) from the JSON over MQTT input.
The filename is used for declaring supported operation. The key`on_fragment` is used to check if the mapping file matches the input payload. If `topic` is not provided, the device control topic is used by default unless `on_message` is defined. `skip_status_update` is optional. If `true`, mapper doesn't send operation status update messages (501-503/504-506). 

```toml
[exec]
topic = "c8y/devicecontrol/notifications"
on_fragment = "com_cumulocity_model_WebCamDevice"
command = "/usr/bin/do_something.sh ${.payload.com_cumulocity_model_WebCamDevice.parameters.duration}"
skip_status_update = true
```

#### FYI: existing custom operation handler definition

```toml
[exec]
topic = "c8y/s/ds"
on_message = "511"
command = "/usr/bin/c8y-command"
```

#### Mapping definition by user
The mapper converts the incoming JSON over MQTT message to thin-edge command message as per the definition file.

[...]

PR #3189, opened 2024-10-15
feat: support user topics when handling custom operation

Ruadhri17:
## Proposed changes
This PR adds support for user topics when handling custom operation. Custom topic can be provided in custom operation in following way (templates supported):

```
[exec]
topic = "${.bridge.topic_prefix}/custom/topic/one"
on_fragment = "com_cumulocity_model_WebCamDevice"
command = "/usr/bin/do_something.sh ${.payload.com_cumulocity_model_WebCamDevice.parameters.duration}"
```

Mapper replace template with mapper prefix (if needed) and subscribe to the topic. If template is different than `${.bridge.topic_prefix}` then this topic will be skipped. 

TODO:

- [x] Robot framework test

## Types of changes

## Paste Link to the issue
* #3095
<br/>

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/smartrest/operations.rs
+++ b/crates/core/c8y_api/src/smartrest/operations.rs
@@ -1,5 +1,6 @@
 use crate::smartrest::error::OperationsError;
 use crate::smartrest::smartrest_serializer::declare_supported_operations;
+use mqtt_channel::TopicFilter;
 use serde::Deserialize;
 use serde::Deserializer;
 use std::collections::HashMap;
@@ -8,6 +9,7 @@ use std::fs;
 use std::path::Path;
 use std::path::PathBuf;
 use std::time::Duration;
+use tedge_api::substitution::Record;
 use tracing::warn;
 
 use super::payload::SmartrestPayload;
@@ -28,8 +30,11 @@ impl Operations {
         self.operations.push(operation);
     }
 
-    pub fn try_new(dir: impl AsRef<Path>) -> Result<Self, OperationsError> {
-        get_operations(dir.as_ref())
+    pub fn try_new(
+        dir: impl AsRef<Path>,
+        bridge_config: &impl Record,
+    ) -> Result<Self, OperationsError> {
+        get_operations(dir.as_ref(), bridge_config)
     }
 
     pub fn get_operations_list(&self) -> Vec<String> {
@@ -78,6 +83,25 @@ impl Operations {
         ops.sort();
         let ops = ops.iter().map(|op| op.as_str()).collect::<Vec<_>>();
         declare_supported_operations(&ops)
+    }
+    pub fn get_json_custom_operation_topics(&self) -> Result<TopicFilter, OperationsError> {
+        Ok(self
+            .operations
+            .iter()
+            .filter(|operation| operation.on_fragment().is_some())
+            .filter_map(|operation| operation.topic())
+            .collect::<HashSet<String>>()
+            .try_into()?)
+    }
+
+    pub fn get_smartrest_custom_operation_topics(&self) -> Result<TopicFilter, OperationsError> {
+        Ok(self
+            .operations
+            .iter()
+            .filter(|operation| operation.on_message().is_some())
+            .filter_map(|operation| operation.topic())
+            .collect::<HashSet<String>>()
+            .try_into()?)
     }
 }
 
@@ -235,7 +259,10 @@ where
     }
 }
 
-pub fn get_operations(dir: impl AsRef<Path>) -> Result<Operations, OperationsError> {
+pub fn get_operations(
+    dir: impl AsRef<Path>,
+    bridge_config: &impl Record,
+) -> Result<Operations, OperationsError> {
     let mut operations = Operations::default();
     let dir_entries = fs::read_dir(&dir)
         .map_err(|_| OperationsError::ReadDirError {
@@ -262,6 +289,11 @@ pub fn get_operations(dir: impl AsRef<Path>) -> Result<Operations, OperationsErr
                 Err(err) => return Err(OperationsError::FromIo(err)),
             };
 
+            if let Some(ref mut exec) = details.exec {
+                if let Some(ref topic) = exec.topic {
+                    exec.topic = Some(bridge_config.inject_values_into_template(topic))
+                }
+            }
             path.file_name()
                 .and_then(|filename| filename.to_str())
                 .ok_or_else(|| OperationsError::InvalidOperationName(path.to_owned()))?
@@ -277,6 +309,7 @@ pub fn get_operations(dir: impl AsRef<Path>) -> Result<Operations, OperationsErr
 
 pub fn get_child_ops(
     ops_dir: impl AsRef<Path>,
+    bridge_config: &impl Record,
 ) -> Result<HashMap<String, Operations>, OperationsError> {
     let mut child_ops: HashMap<String, Operations> = HashMap::new();
     let child_entries = fs::read_dir(&ops_dir)
@@ -289,7 +322,7 @@ pub fn get_child_ops(
         .filter(|path| path.is_dir())
         .collect::<Vec<PathBuf>>();
     for cdir in child_entries {
-        let ops = Operations::try_new(&cdir)?;
+        let ops = Operations::try_new(&cdir, bridge_config)?;
         if let Some(id) = cdir.file_name() {
             if let Some(id_str) = id.to_str() {
                 child_ops.insert(id_str.to_string(), ops);
@@ -363,6 +396,7 @@ mod tests {
     use std::io::Write;
 
     use super::*;
+    use tedge_config::TopicPrefix;
     use test_case::test_case;
 
     // Structs for state change with the builder pattern
@@ -448,13 +482,32 @@ mod tests {
         }
     }
 
+    struct TestBridgeConfig {
+        c8y_prefix: TopicPrefix,
+    }
+
+    impl Record for TestBridgeConfig {
+        fn extract_value(&self, path: &str) -> Option<serde_json::Value> {
+            match path {
+                ".bridge.c8y_prefix" => Some(self.c8y_prefix.as_str().into()),
+                _ => None,
+            }
+        }
+    }
+
     #[test_case(0)]
     #[test_case(1)]
     #[test_case(5)]
     fn get_operations_all(ops_count: usize) {
         let test_operations = TestOperations::builder().with_operations(ops_count).build();
 
-        let operations = get_operations(test_operations.temp_dir()).unwrap();
+        let operations = get_operations(
+            test_operations.temp_dir(),
+            &TestBridgeConfig {
+                c8y_prefix: TopicPrefix::try_from("c8y").unwrap(),
+            },
+        )
+        .unwrap();
 
         assert_eq!(operations.operations.len(), ops_count);
     }
--- a/crates/extensions/c8y_mapper_ext/src/config.rs
+++ b/crates/extensions/c8y_mapper_ext/src/config.rs
@@ -6,6 +6,7 @@ use c8y_api::smartrest::operations::Operations;
 use c8y_api::smartrest::topic::C8yTopic;
 use c8y_auth_proxy::url::Protocol;
 use camino::Utf8Path;
+use serde_json::Value;
 use std::path::Path;
 use std::str::FromStr;
 use std::sync::Arc;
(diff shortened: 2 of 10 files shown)


gligorisaev (commented):
Checked the whole test, The added test step is doing what it was intended for

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_command/c8y_Command_3
```diff
@@ -0,0 +1,4 @@
+[exec]
+topic = "${.topic.root_prefix}/custom/topic/one"
```
reubenmiller, line 2: What does `${.topic.root_prefix}` refer to?

It would be more common to be able to expand the mapper prefix (e.g. `c8y` or whether the cloud prefix etc.), and to avoid confusion with the `topic.root_topic` value (which is currently the local thin-edge.io topic prefix, e.g. `te`).

reubenmiller: Sorry I just read the other code, and could answer the questions.

We should rename the variable as follows:

`${.topic.root_prefix}` => `${.bridge.topic_prefix}`

The `.bridge.topic_prefix` then aligns with the tedge.toml value `c8y.bridge.topic_prefix`, where the leading `c8y.` prefix is not required as the mapper already knows what context it is (as the mapper is started with `tedge-mapper c8y`)

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -50,23 +51,30 @@ impl Operations {
         None
     }
 
-    pub fn filter_by_topic(&self, topic_name: &str) -> Vec<(String, Operation)> {
+    pub fn filter_by_topics(
+        &self,
+        topics: Vec<String>,
+        state: &GenericCommandState,
```
rina23q, line 57: As per this [comment](https://github.com/thin-edge/thin-edge.io/pull/3189#discussion_r1804756584),  this function doesn't have to do anything with workflow's `GenericCommandState` for the topic. Instead, you need to parse prefix here.

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -258,27 +271,69 @@ impl C8yMapperConfig {
...
+                            } else {
+                                warn!("The custom topic '{topic}' is invalid and ignored.");
+                                s
+                            }
+                        } else {
+                            s
+                        }
+                    })
+                    .collect::<String>();
+
```
rina23q, line 327: The logic here can be much simpler. You just need to replace `${.bridge.topic_prefix}` by `C8yMapperConfig::c8y_prefix`.

rina23q (commented):
Still need to work and investigate the test failure, why 504&506 are sent twice.

crates/extensions/c8y_mapper_ext/src/config.rs
Ruadhri17: resolved by c1311aaf198b4b93fd96ab439ab77f816945e7e3

crates/core/c8y_api/src/smartrest/operations.rs
Ruadhri17: resolved by c1311aaf198b4b93fd96ab439ab77f816945e7e3

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_command/c8y_Command_3
Ruadhri17: resolved by c1311aaf198b4b93fd96ab439ab77f816945e7e3

didier-wenzek: `${.bridge.topic_prefix}` sounds really ad-hoc and not extensible in the future.

I would be explicit with `${.config.c8y.bridge.topic_prefix}` and more general with the ability to inject any config setting. Surely more useful for the script rather than the topic on which commands are expected, I agree. However,  making the context explicit avoid consistency and extensibility issues.

Injecting any config value might be a bit too involved for this PR. Hence, it's okay to start with a simple string replacement (as currently implemented) but with an explicit config context (i.e. `${.config.c8y.bridge.topic_prefix}`).

reubenmiller: Though the `${.bridge.topic_prefix}` value is actually the prefix is used by the cloud  "profile" used by the mapper. And since the custom operation templates are under the context of the mapper's cloud already, the value does not need to "c8y" context helper.

didier-wenzek: > the value does not need to "c8y" context helper.

Yes. My concern is that this looks like as a path templating system while it is not, just variable substitution using `${.bridge.topic_prefix}` as a variable name.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -50,23 +52,31 @@ impl Operations {
...
             }
         }
         vec
     }
-
+    /// Get topics of operations that are delivered via JSON over MQTT
     pub fn topics_for_operations(&self) -> HashSet<String> {
         self.operations
             .iter()
+            .filter(|operation| operation.on_fragment().is_some())
```
rina23q, line 79: Oh don't limit it to the ones with `on_fragment`.. This function is used to extract topics also for SmartREST custom operation handlers.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -50,23 +52,31 @@ impl Operations {
         None
     }
 
-    pub fn filter_by_topic(&self, topic_name: &str) -> Vec<(String, Operation)> {
+    pub fn filter_by_topics(
```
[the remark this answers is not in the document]
rina23q: Originally, since `Operations` is a collection of custom operation handlers, the function is intended to extract the matched custom operation handlers by topic. However, it looks repurposed now.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -77,6 +87,13 @@ impl Operations {
         let ops = ops.iter().map(|op| op.as_str()).collect::<Vec<_>>();
         declare_supported_operations(&ops)
     }
+
```
rina23q, line 90: So, as we discussed, you can add two functions here to extract the topics, one for custom JSON over MQTT handlers, and the other for SmartREST custom handlers.
This will allow you to get rid of `device_control_topics` that is added in `C8yMapperConfig`.

```rust
    pub fn get_json_custom_operation_topics(&self) -> TopicFilter {
        let mut topic_filter = TopicFilter::empty();
        for op in self.operations {
            if op.on_fragment().is_some() {
                topic_filter.add(op.topic())
            }
        }
        topic_filter
    }

    pub fn get_smartrest_custom_operation_topics(&self) -> TopicFilter {todo!()}
```

rina23q (commented):
Tests are accurate. Code need to be improved as we discussed.

rina23q (approved):
Now the Rust code looks great. Left one question.

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_command/c8y_Command_3
reubenmiller: Though doesn't the the property name `topic` suggest otherwise?

tests/RobotFramework/tests/cumulocity/shell/c8y_Command
```diff
@@ -1,4 +1,4 @@
 [exec]
-topic = "c8y/s/ds"
+topic ="${.bridge.topic_prefix}/s/ds"
```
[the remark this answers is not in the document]
Ruadhri17: > Why is this change needed?

The tests were failing because of the change we implemented with "get_smartrest_custom_operation_topics()" but it is not exactly the fault of that function but rather we were matching the topic wrongly. This tests are using custom prefix `c8y-custom` but operation file used `c8y` prefix as it can be seen in your comment above. So this file was telling us that message will come on topic `c8y/s/ds` but in reality it was `c8y-custom/s/ds`. Previously it was passing because we just checking incoming topic (if it has mapper prefix or not) and not what is actually in the file.

didier-wenzek (approved):
Approved

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_command/c8y_Command_3
reubenmiller: Sorry I misunderstood your point. I've opened a new discussion with the code in question. https://github.com/thin-edge/thin-edge.io/pull/3189/files#r1813004355

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -260,6 +289,21 @@ pub fn get_operations(dir: impl AsRef<Path>) -> Result<Operations, OperationsErr
...
+                    .and_then(|topic| topic.find(BRIDGE_PREFIX_TEMPLATE))
+                {
+                    // safe to unwrap as we checked that topic exists when looking for bridge prefix template
+                    exec.topic
+                        .as_mut()
+                        .unwrap()
+                        .replace_range(index..index + BRIDGE_PREFIX_TEMPLATE.len(), prefix);
+                }
+            }
+
```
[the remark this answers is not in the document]
reubenmiller: The workflow code that does the variable expansion can be found here: https://github.com/thin-edge/thin-edge.io/blob/be82810edc0ba5d4d68712617d627aa3018d30d4/crates/core/tedge_api/src/workflow/state.rs#L294

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_command/c8y_Command_3
didier-wenzek: @reubenmiller And I misunderstood yours ;-)

After an offline discussion, I agree that it makes sense to use the *bridge* config and not the *tedge* config as root of the substitution in `${.bridge.topic_prefix}`.

For the implementation, what is done is working and can be taken as fist step - even if not extensible.

crates/core/c8y_api/src/smartrest/operations.rs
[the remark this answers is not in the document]
didier-wenzek: Indeed, it would be better to use a similar template expansion. However, the workflow implementation cannot directly be reused here. I will take the time to see how this code can be shared.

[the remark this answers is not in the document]
didier-wenzek: The following commit will help to fix that: https://github.com/thin-edge/thin-edge.io/pull/3204/commits/0bd5274eff1122ed38c3ca5c8472af470472f45a

By implementing the `Record` trait for the bridge config (a single method returning a Json value for a given), this config can then be used to inject value into a string or a script.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -55,12 +54,18 @@ impl Operations {
...
+        topic_name: &str,
+        record: &impl Record,
+    ) -> Vec<(String, Operation)> {
         let mut vec: Vec<(String, Operation)> = Vec::new();
         for op in self.operations.iter() {
             match (op.topic(), op.on_fragment()) {
                 (None, Some(on_fragment)) => vec.push((on_fragment, op.clone())),
-                (Some(topic), Some(on_fragment)) if topic == topic_name => {
+                (Some(ref topic), Some(on_fragment))
+                    if record.inject_values_into_template(topic) == topic_name =>
```
didier-wenzek, line 67: It would be better to inject any `${.bridge...}` config when the operations are loaded. The bridge config will not change after the mapper starts.

```suggestion
                    if topic == topic_name =>
```

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -448,13 +482,32 @@ mod tests {
         }
     }
 
+    struct TestBridgeConfig {
+        c8y_prefix: TopicPrefix,
+    }
```
didier-wenzek, line 487: Why another struct `TestBridgeConfig` that is exactly a copy of `BridgeConfig`?

Ruadhri17: `BridgeConfig` is part of `C8yMapperConfig` so cannot provide it to operations but one is needed to run tests

didier-wenzek (approved):
Approved.

Issue #3036, opened 2024-07-31
Support Cumulocity IoT basic device registration to allow SmartREST 1.0 usage for older clients


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Support Cumulocity IoT basic auth registration process to provide an upgrade path for existing customers whom use (legacy) SmartREST 1.0 (which is not supported when using cert based authentication)

**Background**

Existing customers with existing agents would like to take advantage of thin-edge.io and slowing migrate custom components. However these agents sometimes use SmartREST 1.0 which is incompatible with the cert-based device credentials (the default authentication used by thin-edge.io).

**Describe the solution you'd like**

Add support for basic auth device credentials for the Cumulocity IoT mapper so that devices can be registered following the [Device Registration](https://cumulocity.com/docs/device-management-application/registering-devices/) mechanism.

**Implementation details**

**When using basic auth device credentials**

* Do not store Cumulocity device credentials in the *tedge.toml* as these should be protected like a certificate (and not mixed with other non-sensitive configuration)
* c8y proxy must use a Basic Auth ,e.g. `Authorization: Basic <base64(username:password)>`
* Don't subscribe to `s/uat` and `s/dat` topics as they are only supported when using cert-based device credentials and errors will be posted to `s/e` if they are subscribed to

*Open questions*

* How to test an MQTT connection as the JWT topics are not available - Should a very basic SmartREST 1.0 template be used instead?

**Implementation plan**

The feature implementation can be broken into the following parts:

1. Support new configuration sets the registration mode (e.g. use basic auth, or device credentials/jwt)
    * Activate the basic mode via `tedge config set c8y.registration_mode "basic"`
    * Allow a username and password to be set via the command line and store in a new file (owned by `tedge` and with `600` permissions)
2. Update components to support using either Bearer or Basic auth for communication with Cumulocity IoT REST API (see below for a list of components)
3. Support registration of SmartREST 1.0 templates via a new tedge config `tedge config set c8y.smartrest1.templates "template1,template2"`
4. Support the Cumulocity IoT device registration process using the [Cumulocity IoT bootstrap credentials](https://cumulocity.com/api/core/#operation/postDeviceCredentialsCollectionResource) when calling `tedge connect` and the "basic" registration mode is active (there are some UX aspects here to consider).

**Describe alternatives you've considered**

**Additional context**

Check list for components that require an update of the Authorization header mechanisms:

[...]

PR #3190, opened 2024-10-16
refactor: flexible Authorization header type of HTTP request

rina23q:
## Proposed changes
Pure refactoring PR to make a progress of #3036 easier.
- Now `C8YJwtRetriever` returns `HeaderMap`, previously it was just JWT token (`String`).
- Deprecate `Auth` enum in downloader so that downloader can accept any headers easily
- Remove unused functions in 897c1022a28c8593578788055b422b6f048bd966

Why this refactoring is needed?
#3036 needs to introduce Basic authentication instead of JWT token authentication. However, the current implementation, especially HTTP APIs, are designed to use Authorization header together with JWT token. Since the JWT actor is actually a server trait, for basic authentication, we can just implement a server trait.

Question:
The JWT token inplementation in `c8y_remote_access_plugin` seems duplicated to `C8YJwtRetriever` implementation... 
- https://github.com/thin-edge/thin-edge.io/blob/main/plugins/c8y_remote_access_plugin/src/auth.rs 
- https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_http_proxy/src/credentials.rs
=> I understand they are different level. 

## Types of changes

## Paste Link to the issue
#3036

## Checklist

### DIFF
--- a/crates/common/download/src/download.rs
+++ b/crates/common/download/src/download.rs
@@ -50,7 +50,7 @@ fn default_backoff() -> ExponentialBackoff {
 pub struct DownloadInfo {
     pub url: String,
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub auth: Option<Auth>,
+    pub auth: Option<String>,
 }
 
 impl From<&str> for DownloadInfo {
@@ -69,9 +69,9 @@ impl DownloadInfo {
     }
 
     /// Creates new [`DownloadInfo`] from a URL with authentication.
-    pub fn with_auth(self, auth: Auth) -> Self {
+    pub fn with_auth(self, auth: &str) -> Self {
         Self {
-            auth: Some(auth),
+            auth: Some(auth.into()),
             ..self
         }
     }
@@ -82,21 +82,6 @@ impl DownloadInfo {
 
     pub fn is_empty(&self) -> bool {
         self.url.trim().is_empty()
-    }
-}
-
-/// Possible authentication schemes
-#[derive(Debug, Clone, Deserialize, PartialEq, Eq, Serialize)]
-#[serde(rename_all = "camelCase")]
-#[serde(deny_unknown_fields)]
-pub enum Auth {
-    /// HTTP Bearer authentication
-    Bearer(String),
-}
-
-impl Auth {
-    pub fn new_bearer(token: &str) -> Self {
-        Self::Bearer(token.into())
     }
 }
 
@@ -384,8 +369,8 @@ impl Downloader {
 
         let operation = || async {
             let mut request = self.client.get(url.url());
-            if let Some(Auth::Bearer(token)) = &url.auth {
-                request = request.bearer_auth(token)
+            if let Some(header_value) = &url.auth {
+                request = request.header("Authorization", header_value)
             }
 
             if range_start != 0 {
@@ -926,7 +911,7 @@ mod tests {
         // applying token if `with_token` = true
         let url = {
             if with_token {
-                url.with_auth(Auth::Bearer(String::from("token")))
+                url.with_auth("Bearer token")
             } else {
                 url
             }
--- a/crates/common/download/src/download.rs
+++ b/crates/common/download/src/download.rs
@@ -11,6 +11,7 @@ use log::warn;
 use nix::sys::statvfs;
 pub use partial_response::InvalidResponseError;
 use reqwest::header;
+use reqwest::header::HeaderMap;
 use reqwest::Client;
 use reqwest::Identity;
 use serde::Deserialize;
@@ -20,8 +21,6 @@ use std::fs::File;
 use std::io::Seek;
 use std::io::SeekFrom;
 use std::io::Write;
-#[cfg(target_os = "linux")]
-use std::os::unix::prelude::AsRawFd;
 use std::path::Path;
 use std::path::PathBuf;
 use std::time::Duration;
@@ -31,6 +30,8 @@ use tedge_utils::file::FileError;
 use nix::fcntl::fallocate;
 #[cfg(target_os = "linux")]
 use nix::fcntl::FallocateFlags;
+#[cfg(target_os = "linux")]
+use std::os::unix::prelude::AsRawFd;
 
 fn default_backoff() -> ExponentialBackoff {
     // Default retry is an exponential retry with a limit of 15 minutes total.
@@ -49,8 +50,8 @@ fn default_backoff() -> ExponentialBackoff {
 #[serde(deny_unknown_fields)]
 pub struct DownloadInfo {
     pub url: String,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub auth: Option<String>,
+    #[serde(skip)]
+    pub headers: HeaderMap,
 }
 
 impl From<&str> for DownloadInfo {
@@ -64,14 +65,14 @@ impl DownloadInfo {
     pub fn new(url: &str) -> Self {
         Self {
             url: url.into(),
-            auth: None,
+            headers: HeaderMap::new(),
         }
     }
 
     /// Creates new [`DownloadInfo`] from a URL with authentication.
-    pub fn with_auth(self, auth: &str) -> Self {
+    pub fn with_headers(self, header_map: HeaderMap) -> Self {
         Self {
-            auth: Some(auth.into()),
+            headers: header_map,
             ..self
         }
     }
@@ -369,9 +370,7 @@ impl Downloader {
 
         let operation = || async {
             let mut request = self.client.get(url.url());
-            if let Some(header_value) = &url.auth {
-                request = request.header("Authorization", header_value)
-            }
+            request = request.headers(url.headers.clone());
 
             if range_start != 0 {
                 request = request.header("Range", format!("bytes={range_start}-"));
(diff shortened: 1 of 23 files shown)


reubenmiller:
@rina23q Wouldn't it make more sense to have a generic "Add N Headers"...that way we could write providers to inject any required headers and not assume it was just be the "Authorization" header?

crates/extensions/c8y_http_proxy/src/credentials.rs
```diff
@@ -32,16 +32,17 @@ impl C8YJwtRetriever {
 
 #[async_trait]
 impl Server for C8YJwtRetriever {
```
didier-wenzek, line 34: Adding the new `C8YBasicAuthProvider` in this PR would really help to validate the refactoring, even if not use yet.

crates/common/download/src/download.rs
```diff
@@ -85,21 +85,6 @@ impl DownloadInfo {
...
 }
 
-/// Possible authentication schemes
-#[derive(Debug, Clone, Deserialize, PartialEq, Eq, Serialize)]
-#[serde(rename_all = "camelCase")]
-#[serde(deny_unknown_fields)]
-pub enum Auth {
-    /// HTTP Bearer authentication
-    Bearer(String),
-}
```
didier-wenzek, line 95: Have you consider to keep this type, i.e. to add a `Basic` case?

Using strings for auth headers is convenient but also error prone. This is why I would not throw away this type too quickly. The main issue I see is related to the current crate organization we have: we don't want to have all HTTP related crate depending on this download crate.

didier-wenzek (commented):
> Question:
> The JWT token implementation in `c8y_remote_access_plugin` seems duplicated to `C8YJwtRetriever` implementation... 
> - https://github.com/thin-edge/thin-edge.io/blob/main/plugins/c8y_remote_access_plugin/src/auth.rs 
> - https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_http_proxy/src/credentials.rs

This is not a duplicate, just a difference on the API level which is used. The `c8y_remote_access_plugin` directly invokes `C8yMqttJwtTokenRetriever::get_jwt_token()` instead as wrapping this call behind an actor as done by `C8YJwtRetriever`.

=> The implication for that refactoring PR is that the difference between the misc auth method has to be addressed in the `c8y_api` with a facade over the ``C8yMqttJwtTokenRetriever` and a `BasicAuthRetriver`. Both returning an authentication header ready to be used.

crates/common/download/src/download.rs
rina23q: > Have you consider to keep this type, i.e. to add a Basic case?

Yes, but there are two mixed feelings. Then I decided to discard it.
- One side, in the `C8YJwtRetriever` level, keep returning `Result<String, JwtError>` will make less changes in existing code. 
- Another side, in the `downloader` level, since it's using `reqwest` directly (on the contrary `c8y_http_proxy` is using a wrapper of `hyper`), this `Auth` enum helps because `reqwest` has a dedicated API to add auth like `request.bearer_auth(token)`.

> The main issue I see is related to the current crate organization we have: we don't want to have all HTTP related crate depending on this download crate.

Exactly. Another issue. Then question is where is appropriate... maybe create a new file at `crates/common/tedge_utils/http.rs`? 

If we keep `enum`,  the return type of `C8YJwtRetriever` and `C8YBasicAuthProvider` to `Result<Auth, JwtError>` can be changed. Then implement functions on `Auth` to return string with `Bearer` or `Basic` prefix.

```rust
pub enum Auth {
    /// HTTP Bearer authentication
    Bearer(String),
    Basic(String),
}

impl Auth {
    pub fn to_header_value(&self) -> String {
        match self {
            Auth::Bearer(value) => format!("Bearer {value}"),
            Auth::Basic(value) => format("Basic {value}"),
        }
    }
}
```

But again, mixed feeling for efforts vs beauty. :(

However, I'm not sure how this `enum` will be helpful compared to the changes that is required due to the return type change. The point is, we use different HTTP crates for downloader (`reqwest`) and http_proxy_api (`hyper`).

crates/common/tedge_config_macros/src/multi.rs
```diff
@@ -39,7 +39,7 @@ impl AsRef<OsStr> for ProfileName {
 }
 
 fn validate_profile_name(value: &str) -> Result<(), anyhow::Error> {
-    ensure!(value.starts_with("@"), "Profile names must start with `@`");
+    ensure!(value.starts_with('@'), "Profile names must start with `@`");
```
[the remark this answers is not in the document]
rina23q: My local clippy complained it.. but not complained in our pipeline.

rina23q:
> @rina23q @didier-wenzek @albinsuresh Wouldn't it make more sense to have a generic "Add N Headers"...that way we could write providers to inject any required headers and not assume it was just be the "Authorization" header?

Ah like returning `({header_key}, {header_value})`, e.g. `("Authorization", "Bearer XXX")`? So it means `C8YJwtRetriever` returns `Result<(String, String), HeaderError>` instead of `Result<String, AuthError>`.

crates/common/tedge_config_macros/src/multi.rs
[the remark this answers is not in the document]
didier-wenzek: Create a PR just for that then.

crates/common/download/src/download.rs
didier-wenzek: > But again, mixed feeling for efforts vs usability. :(

My concern is not related to usability but maintainability. Using `String` values is convenient until you have to refactor your code: you have then no support from the compiler for which a bearer string or a basic-auth string make no difference.

> Probably the best way is that `C8YBasicAuthProvider` returns `Result<(HeaderName, HeaderValue), Error)`. See my [comment](https://github.com/thin-edge/thin-edge.io/pull/3190#issuecomment-2419148275).

Why not. This makes the intent clearer.

reubenmiller:
> > @rina23q @didier-wenzek @albinsuresh Wouldn't it make more sense to have a generic "Add N Headers"...that way we could write providers to inject any required headers and not assume it was just be the "Authorization" header?
> 
> Ah like returning `({header_key}, {header_value})`, e.g. `("Authorization", "Bearer XXX")`?
> 
> So it means `C8YJwtRetriever` returns `Result<(String, String), HeaderError>` instead of `Result<String, AuthError>`.
> 
> Or even getting a help from the http crate's [`HeaderName`](https://docs.rs/http/latest/http/header/struct.HeaderName.html) and [`HeaderValue`](https://docs.rs/http/latest/http/header/struct.HeaderValue.html), it can be `Result<(HeaderName, HeaderValue), Error)`.

Or even better, it results a vector of header name/value 's)...As some more complicated authenticated mechanisms need to set multiple header properties, for example [AWS S3 API](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html) allows users to also add `x-*` headers to add some additional options.

rina23q:
> Or even better, it results a vector of header name/value 's)...As some more complicated authenticated mechanisms need to set multiple header properties, for example [AWS S3 API](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html) allows users to also add `x-*` headers to add some additional options.

Then this [`HeaderMap`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html) is the way to go 👍

rina23q:
> Then this [`HeaderMap`](https://docs.rs/http/latest/http/header/struct.HeaderMap.html) is the way to go 👍

This change is done in e7b87248c836c63df9be25b6afb91827e7d55037.

Also, I removed unused code in 897c1022a28c8593578788055b422b6f048bd966. If they should stay in our code, please inform me. Then I can revert in that case.

crates/common/download/src/download.rs
```diff
@@ -384,8 +370,8 @@ impl Downloader {
 
         let operation = || async {
             let mut request = self.client.get(url.url());
-            if let Some(Auth::Bearer(token)) = &url.auth {
-                request = request.bearer_auth(token)
+            for (key, value) in &url.headers {
+                request = request.header(key, value)
             }
```
didier-wenzek, line 375: Simply use [headers](https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.headers)

crates/extensions/c8y_http_proxy/src/credentials.rs
didier-wenzek: Sorry, I should have remove this comment which is conflicting with my other comment https://github.com/thin-edge/thin-edge.io/pull/3190#pullrequestreview-2374370308, telling that:

> the difference between the misc auth method has to be addressed in the c8y_api with a facade over the `C8yMqttJwtTokenRetriever` and a `BasicAuthRetriver`. Both returning an authentication header ready to be used.

Anyway. This can be done in the next step.

didier-wenzek (approved):
Approved. This refactoring PR is a nice progress forward.

> Also, I removed unused code in 897c1022a28c8593578788055b422b6f048bd966. If they should stay in our code, please inform me. Then I can revert in that case.

Removing dead code is always a good thing.

crates/common/download/src/download.rs
didier-wenzek: The use of `HeaderMap` is far better than trying to keep this adhoc `Auth` type.

PR #3193, opened 2024-10-18
docs: add tedge-watchdog decision page

reubenmiller:
## Proposed changes

Add a new decision page after a decision about the future of the **tedge-watchdog** component and rationale on why shifting focus to independent solutions such as `monit` would lead to a more reliable/holistic approach to moniting components and performing corrective actions to bring a system back to a healthy state.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/design/decisions/0005-tedge-watchdog.md
+++ b/design/decisions/0005-tedge-watchdog.md
@@ -1,0 +1,50 @@
+# tedge-watchdog
+
+* Date: __2024-10-18__
+* Status: __New__
+
+## Background
+
+The **tedge-watchdog** component is to provide an interface to the [systemd watchdog functionality](https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html) by using the thin-edge.io MQTT health status topic to publish messages to the systemd watchdog mechanism (which uses [sd_notify](https://www.freedesktop.org/software/systemd/man/latest/sd_notify.html#)).
+
+The **tedge-watchdog** service, will send out periodic messages on the the `te/<service-topic-id>/cmd/health/check` topic, and wait for a response on the `te/<service-topic-id>/status/health` topic. If an updated response IS received, then the **tedge-watchdog** will send an `sd_notify` event on behalf of the service so that systemd knows that the service is still "ok/healthy".
+
+Whilst the utilizing the systemd watchdog functionality makes for a low-resource usage solution, it has the following disadvantages:
+
+* It can only monitor systemd services
+* It relies on the `health/check` status being representative of being "healthy"
+* It only works for systemd services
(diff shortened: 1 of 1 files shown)


Bravo555 (approved):

didier-wenzek (commented):
I'm okay to go in the direction proposed here. i.e. to stop further development of the **tedge-watchdog** component (at to deprecate it) and to improve the `tedge-monit-setup` for thin-edge.

reubenmiller:
> I'm okay to go in the direction proposed here. i.e. to stop further development of the **tedge-watchdog** component (at to deprecate it) and to improve the `tedge-monit-setup` for thin-edge.

I'm still in two minds about whether to fully deprecate the plugin or not. I'm leaning towards at least keeping it in the code based, so the other users can continue development (if they do prefer the systemd based approach).

@didier-wenzek What do you think?

didier-wenzek:
> I'm still in two minds about whether to fully deprecate the plugin or not.

Do we have known users? Not sure. We have few or no questions about it.

I would keep the code for the `tedge_watchdog` but remove it from the documentation as we do for the `c8y_firmware_plugin`.

reubenmiller:
> > I'm still in two minds about whether to fully deprecate the plugin or not.
> 
> Do we have known users? Not sure. We have few or no questions about it.
> 

We don't have any exact data to support this, however since the tedge-watchdog always needed some manual configuration to enable it, I really doubt anyone has been using it, however I would use caution and expect there to be at least a few users.


> I would keep the code for the `tedge_watchdog` but remove it from the documentation as we do for the `c8y_firmware_plugin`.

The tedge watchdog docs has already been marked as "unlisted" since a while now (when the original limitation was found see https://github.com/thin-edge/thin-edge.io/pull/2635). Unlisted pages don't turn up in the doc searches, but can still be accessed if someone knows the url (e.g. they have saved it in a bookmark) - [here is a link to the unlisted page](https://thin-edge.github.io/thin-edge.io/operate/monitoring/systemd-watchdog/)

didier-wenzek:
> The tedge watchdog docs has already been marked as "unlisted"

So, let's leave things as they are now.

design/decisions/0005-tedge-watchdog.md
```diff
@@ -0,0 +1,50 @@
+# tedge-watchdog
+
+* Date: __2024-10-18__
+* Status: __New__
```
didier-wenzek, line 4:
```suggestion
* Status: __Approved__
```

didier-wenzek (approved):
Approved

albinsuresh (approved):

design/decisions/0005-tedge-watchdog.md
reubenmiller: Resolved by 79a9ab175aef11dea1cb1446707ff2cb56c768c6

PR #3196, opened 2024-10-21
feat: support SmartREST1.0 for Cumulocity

rina23q:
## Proposed changes
To support SmartREST1.0, this PR includes:

- basic auth support for HTTP including `c8y-remote-access-plugin`
- credentials file implementation
- username/password support for MQTT bridge
- username/password authentication for MQTT connections
- SmartREST1.0 templates support in `tedge config` and MQTT bridges.
- system tests, thanks to @reubenmiller

---

How to use basic auth instead of device certs and SmartREST1.0?

1. Set the mode to "basic"
```
tedge config set c8y.auth_mode "basic"
```

2. Give username and password to the credentials file. The location is `/etc/tedge/credentials` by default. Configurable by `c8y.credentials_path`.

```toml
[c8y]
username = "t5678/octocat"
password = "abcd1234"
```

3. Give the SmartREST1.0 templates names to tedge config.
```
tedge config set c8y.smartrest1.templates "{template_xid}"
```

4. Run `tedge connect c8y`! The appropriate bridge for Basic Auth and SmartREST1.0 will be created.

Note: if a device is already registered via certificate, you cannot switch to basic auth for the device.

---

Current limitations:
- `device.id` is still referring the value from the device certificate.
- the device will not be registered during `tedge connect c8y`. It expects to use bulk operation API to get credentials (username/password) beforehand.
- the connection check and the tenant URL check in `tedge connect c8y` is disabled since it uses JWT token. However, the connection check is actually done by sending `100` message, and supposed to receive `71, device already exists`.

## Types of changes

## Paste Link to the issue
#3036

## Checklist


reubenmiller:
@rina23q What about changing the default credential location to the following, keeping in mind that it would allow us to add additional credentials to it (maybe in the future adding the c8y device bootstrapping credentials might be necessary).

Below is an example of using a single credentials file with multiple secrets, where keys are used to classify the secrets:

*File*
```sh
/etc/tedge/credentials
```

*Contents*
```toml
[c8y]
username = "octocat"
password = "abcd1234"
```

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -131,53 +135,120 @@ impl C8yEndPoint {
...
+    }
 
-        C8yMqttJwtTokenRetriever {
-            mqtt_config,
-            topic_prefix,
-        }
+    async fn get_basic_auth(&self) -> Result<BasicCredentials, C8yAuthRetrieverError> {
+        let C8yAuthType::Basic { credentials_path } = &self.auth else {
+            panic!("This method should not be called if not intended to use Basic auth (username/password).");
+        };
```
didier-wenzek, line 235: The fact that you have to add this `panic!` is a sign that something is wrong in the design of the `C8yAuthRetriever`.

Instead of `struct` using an `enum` field used to distinguish the cases, `C8yAuthRetriever` should directly be an `enum`:

```rust
pub enum C8yAuthRetriever {
    Basic { credentials_path: Utf8PathBuf },
    Jwt { mqtt_config: mqtt_channel::Config,  topic_prefix: TopicPrefix },
}

impl C8yAuthRetriever {
    pub async fn get_auth_header_value(&self) -> Result<HeaderValue, C8yAuthRetrieverError> {
        match self {
            C8yAuthRetriever::Jwt { mqtt_config, topic_prefix } => {
                let jwt_token = Self::get_jwt_token(mqtt_config, topic_prefix).await?;
                Ok(format!("Bearer {}", jwt_token))
            }
            C8yAuthRetriever::Basic { credentials_path } => {
                let basic_auth = Self::get_basic_auth(credentials_path).await?;
                Ok(format!("Basic {}", basic_auth))
            }
        };
    }
    
    async fn get_jwt_token(mqtt_config: mqtt_channel::Config,  topic_prefix: TopicPrefix) -> Result<String, C8yAuthRetrieverError> {
        ...
    }
    
    async fn get_basic_auth(credentials_path: Utf8PathBuf) -> Result<String, C8yAuthRetrieverError> {
        ...
    }
}
```

Note that, `get_basic_auth` is no more a method but a function which is given the expected parameters.

didier-wenzek (commented):
The switch between JWT versus BasicAuth is implemented in the correct place. i.e. under `c8y_api` which is used both by the mapper and the remote access plugin.

However, the `C8yAuthRetriever` can be simplified.

crates/core/c8y_api/src/http_proxy.rs
rina23q: Done in [0eeb299](https://github.com/thin-edge/thin-edge.io/pull/3196/commits/0eeb299ce6a4c7715f78839d222a82de44da0e78)

rina23q:
> fca45c5

Done in [fca45c5](https://github.com/thin-edge/thin-edge.io/pull/3196/commits/fca45c504f6a8eaa1257f29b69a57a8a8d9f1764)

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -470,6 +470,17 @@ define_tedge_config! {
         #[doku(as = "PathBuf")]
         root_cert_path: Utf8PathBuf,
 
+        // TODO: Enum or bool? If enum, what are the other items?
+        /// Use basic authentication (username/password) instead of device certificate based authentication
+        #[tedge_config(example = "true", default(value = false))]
+        use_basic_auth: bool,
```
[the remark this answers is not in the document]
rina23q: > If device credentials are configure for c8y, then uses HTTP basic-auth to connect c8y. Otherwise, use a JWT token.

@reubenmiller 
Can you give me UX perspective here? For me, both (credentials are configured or enum) sound good...

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -131,54 +136,103 @@ impl C8yEndPoint {
...
+    Jwt {
+        mqtt_config: Box<mqtt_channel::Config>,
+        topic_prefix: TopicPrefix,
+    },
+}
+
+/// The credential file representation. e.g.:
+/// ```toml
+/// [c8y]
+/// username = "octocat"
```
rina23q, line 152:
```suggestion
/// username = "t1234/octocat"
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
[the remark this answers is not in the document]
reubenmiller: @rina23q I would go with an Enum as there is a case if someone wants to transition from username/password to token based auth (via certificates), then you'd change the mode to JWT, verify that the new connection works, then delete the old username/password file.

But in some cases I do like the auto detection stuff, so maybe we can do both, but having another enum value:

* certificate (e.g. jwt)
* basic
* auto (will use basic if username/password are provided)

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
```diff
@@ -909,6 +909,7 @@ def get_bridge_service_name(self, cloud: str) -> str:
         # Legacy mosquitto bridge
         return f"mosquitto-{cloud}-bridge"
 
+<<<<<<< HEAD
```
rina23q, line 912:
```suggestion
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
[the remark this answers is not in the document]
rina23q: Implemented in 4e41c4efd320286ba796c87d57122dbe69077d13

crates/core/tedge/src/cli/connect/c8y_direct_connection.rs
```diff
@@ -33,11 +35,26 @@ pub fn create_device_with_direct_connection(
...
+            bridge_config
+                .remote_username
+                .clone()
+                .expect("username must be set to use basic auth"),
+            bridge_config
+                .remote_password
+                .clone()
+                .expect("password must be set to use basic auth"),
+        );
+        dbg!(&mqtt_options);
```
rina23q, line 49:
```suggestion
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -470,6 +472,16 @@ define_tedge_config! {
         #[doku(as = "PathBuf")]
         root_cert_path: Utf8PathBuf,
 
+        /// The type of authentication method
+        #[tedge_config(example = "certificate", example = "basic", example = "auto", default(variable = RegistrationMode::Certificate))]
+        registration_mode: RegistrationMode,
```
didier-wenzek, line 477: The term `registration_mode` is not crystal clear.

```suggestion
        /// The authentication method used to connect Cumulocity
        #[tedge_config(note = "In the auto mode, basic auth is used if c8y.credentials_path is set")]
        #[tedge_config(example = "certificate", example = "basic", example = "auto", default(variable = RegistrationMode::Certificate))]
        auth_method: RegistrationMode,
```

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -131,54 +137,107 @@ impl C8yEndPoint {
...
-            .with_subscriptions(topic);
-
-        C8yMqttJwtTokenRetriever {
-            mqtt_config,
-            topic_prefix,
-        }
+    pub async fn get_auth_header_value(&mut self) -> Result<HeaderValue, C8yAuthRetrieverError> {
+        let header_value = match &self {
+            Self::Basic { credentials_path } => {
+                let (username, password) = read_c8y_credentials(credentials_path)?;
```
didier-wenzek, line 221: Is this a conscious choice to read the password file on each and every HTTP request?

This has the benefit that a wrong password can be corrected on the fly. But this updated password will not be used by the MQTT bridge. So in any case a `tedge reconnect c8y` will be required after a password update.

crates/core/tedge/src/cli/connect/c8y_direct_connection.rs
```diff
@@ -18,6 +19,7 @@ use rumqttc::Transport;
 
 // Connect directly to the c8y cloud over mqtt and publish device create message.
 pub fn create_device_with_direct_connection(
+    use_basic_auth: bool,
```
didier-wenzek, line 22: To we need to create the device on Cumulocity when the device is authenticated using a password? Somehow the device should already been known by the tenant.

crates/core/c8y_api/src/http_proxy.rs
rina23q: I didn't really think about it deeply. Since we get JWT token in every new HTTP requests, it follows that way. Do you prefer to change it?

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
rina23q: Done in [6d8a324](https://github.com/thin-edge/thin-edge.io/pull/3196/commits/6d8a3241a80521d15e9215e1358b8930314215a0)

didier-wenzek (commented):
LGTM. I will be happy to approve once the system tests fixed.

crates/core/tedge/src/bridge/c8y.rs
```diff
@@ -87,6 +99,34 @@ impl From<BridgeConfigC8yParams> for BridgeConfig {
             .collect::<Vec<String>>();
         topics.extend(templates_set);
 
+        // SmartRest1 (to support customers with existing solutions based on SmartRest 1)
+        // Only add the topics if at least 1 template is defined
+        if !smartrest_one_templates.0.is_empty() {
+            topics.extend([
+                format!(r#"s/ul/# out 2 {topic_prefix}/ """#),
```
albinsuresh, line 106:
```suggestion
                format!(r#"s/ul/# out 1 {topic_prefix}/ """#),
```
In a recent PR, we switched all those QoS values to 1, as Cumulocity doesn't support 2. So, you may wanna change all the following ones as well to `1`.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -48,28 +50,44 @@ impl TEdgeComponent for CumulocityMapper {
...
             let cloud_topics = [
-                "s/dt",
-                "s/dat",
-                "s/ds",
-                "s/e",
-                "devicecontrol/notifications",
-                "error",
+                ("s/dt", true),
+                ("s/ds", true),
+                ("s/dat", use_certificate),
```
albinsuresh, line 76: Not a request for change but just a remark: I'd have just kept the older array (or a vec) and just added this `s/dat` topic to that vector based on the `is_certificate` value, to keep the logic simpler.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -87,6 +105,12 @@ impl TEdgeComponent for CumulocityMapper {
             tc.forward_from_local("q/us/#", local_prefix.clone(), "")?;
             tc.forward_from_local("c/us/#", local_prefix.clone(), "")?;
 
+            // SmartREST1
+            tc.forward_from_local("s/ul/#", local_prefix.clone(), "")?;
+            tc.forward_from_local("t/ul/#", local_prefix.clone(), "")?;
+            tc.forward_from_local("q/ul/#", local_prefix.clone(), "")?;
+            tc.forward_from_local("c/ul/#", local_prefix.clone(), "")?;
```
albinsuresh, line 112: Why not guard this block with an `if !use_certificate`, as there's no point adding these rules when using certificates.

didier-wenzek (approved):
Approved.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -185,14 +224,14 @@ impl TEdgeComponent for CumulocityMapper {
                 )
                 .await?;
         }
-        let mut jwt_actor =
-            C8YJwtRetriever::builder(mqtt_config.clone(), c8y_config.bridge.topic_prefix.clone());
+
+        let mut header_actor = C8YHeaderRetriever::try_builder(&tedge_config, c8y_profile)?;
```
albinsuresh, line 228:
```suggestion
        let mut auth_actor = C8YHeaderRetriever::try_builder(&tedge_config, c8y_profile)?;
```

plugins/c8y_firmware_plugin/src/lib.rs
```diff
@@ -94,15 +94,7 @@ async fn run_with(
...
-    let mut jwt_actor = C8YJwtRetriever::builder(
-        mqtt_config.clone(),
-        tedge_config
-            .c8y
-            .try_get(c8y_profile)?
-            .bridge
-            .topic_prefix
-            .clone(),
-    );
+    let mut jwt_actor = C8YHeaderRetriever::try_builder(&tedge_config, c8y_profile)?;
```
albinsuresh, line 97:
```suggestion
    let mut auth_actor = C8YHeaderRetriever::try_builder(&tedge_config, c8y_profile)?;
```

tests/RobotFramework/tests/cumulocity/smartrest_one/smartrest_one.robot
```diff
@@ -0,0 +1,78 @@
...
+
+*** Test Cases ***
+Supports SmartREST 1.0 Templates
+    [Template]    Register and Use SmartREST 1.0. Templates
+    use_builtin_bridge=true
+    use_builtin_bridge=false
+
+
+*** Keywords ***
+Register and Use SmartREST 1.0. Templates
```
albinsuresh, line 28: The registration part can be extracted as a keyword into our `robotframework-c8y` in a separate PR along with the`Delete SmartREST 1.0 Template` that was also added separately. The registration keyword can also auto cleanup of the registered  templates as part of test teardown.

tests/RobotFramework/tests/cumulocity/smartrest_one/smartrest_one.robot
```diff
@@ -0,0 +1,78 @@
...
+
+    Execute Command    cmd=tedge mqtt pub te/device/main///a/test '{"text":"test alarm","severity":"major"}' -r
+    Device Should Have Alarm/s    type=test    expected_text=test alarm
+
+    # Get alarms
+    Execute Command    cmd=tedge mqtt pub c8y/s/ul/${TEMPLATE_XID} '311,${mo["id"]},ACTIVE'
+    Should Have MQTT Messages    c8y/s/dl/${TEMPLATE_XID}    message_pattern=^808,\\d+,\\d+,test    timeout=10
+
+    # Operations
+    ${OPERATION}=    Get Configuration    tedge-configuration-plugin
```
albinsuresh, line 55: What's the point of this? Just making sure that other features are also working fine?

crates/core/tedge/src/bridge/config.rs
```diff
@@ -91,17 +101,19 @@ impl BridgeConfig {
...
+            if !self.bridge_certfile.exists() {
+                return Err(ConnectError::Certificate);
+            }
 
-        if !self.bridge_keyfile.exists() {
-            return Err(ConnectError::Certificate);
+            if !self.bridge_keyfile.exists() {
+                return Err(ConnectError::Certificate);
+            }
         }
```
albinsuresh, line 117: No validation required when `use_basic_auth == true`? Like whether the `credentials` file exists?

This validation is done during the `Validating the bridge certificates.` phase of `tedge connect`. We may wanna change that message as well, to better align with the new option.

albinsuresh (commented):
The changes are looking good. Pretty clean. Just a few minor comments.

Also, I noticed that the device is getting created automatically as part of `tedge connect`, without having to explicitly register the device on the cloud first: a side effect of the direct device creation step during `tedge connect`. I see nothing wrong with that. In fact, quite consistent with behaviour when certs are used. But, if that isn't the expected behaviour, then we may wanna bypass the device creation step while using basic auth.

crates/core/tedge/src/bridge/c8y.rs
rina23q: This PR and #3201 will make a conflict to each other. The one gets merged faster will fix ;)

crates/core/tedge_mapper/src/c8y/mapper.rs
rina23q: Actually `header_actor` is more accurate. It is returning only authorization header as of now, but in the future, one can modify it to return any kinds of header(s).

tests/RobotFramework/tests/cumulocity/smartrest_one/smartrest_one.robot
rina23q: Exactly. Testing if receiving operation from c8y works or not.

crates/core/tedge_mapper/src/c8y/mapper.rs
rina23q: quickly done in [452943b](https://github.com/thin-edge/thin-edge.io/pull/3196/commits/452943bf7b9d8933c5d20513d813259e2389116d)

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -67,7 +67,7 @@ impl TEdgeComponent for CumulocityMapper {
                 .iter()
                 .map(|id| Cow::Owned(format!("s/dc/{id}")));
 
-            let use_certificate = c8y_config
+                let use_certificate = c8y_config
```
albinsuresh, line 70: That tripped the formatter.

albinsuresh (approved):
LGTM.

Another niggle that I observed during `tedge connect` is the following message:

```
Failed to get the connected tenant URL from Cumulocity.
```

Is that endpoint not supported while using basic auth?

rina23q:
The SmartREST1.0 system test seems freaky :/

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -470,6 +472,17 @@ define_tedge_config! {
         #[doku(as = "PathBuf")]
         root_cert_path: Utf8PathBuf,
 
+        /// The authentication method used to connect Cumulocity
+        #[tedge_config(note = "In the auto mode, basic auth is used if c8y.credentials_path is set")]
+        #[tedge_config(example = "certificate", example = "basic", example = "auto", default(variable = AuthMethod::Certificate))]
```
albinsuresh, line 477: Keeping `certificate` as the default value makes the `auto` variant irrelevant IMO as I can't imagine it ever being used by anyone. When the default is already set to certificate, a user might either keep it or change it to basic. I can't imagine why someone would explicitly change that to `auto` and embrace uncertainty.

tests/RobotFramework/tests/cumulocity/smartrest_one/smartrest_one.robot
reubenmiller: More specifically, it tests the thin-edge.io components that rely on c8y API also work (e.g. uploading a file to Cumulocity).

plugins/c8y_firmware_plugin/src/lib.rs
rina23q: I'll not change code in `c8y_firmware_plugin` unless it's critical since it's a legacy feature.

rina23q:
> LGTM.
> 
> A few other niggles that I observed during `tedge connect`:
> 
> ```
> Local MQTT publish has timed out.
> 
> Make sure mosquitto is running.
> Failed to get the connected tenant URL from Cumulocity.
> ```
> 
> Users might get alarmed with such error messages. So, we'll have to hide or update some of these messages to account for basic auth based connections. BTW, since fetching the tenant URL from C8y failed, is that endpoint not supported while using basic auth?

This reported issue is resolved in b1ace6761ea49393fc25716d9d7c34fa090d9884.

Issue #2089, opened 2023-07-24
Control command's log level via a flag


PradeepKiruvale:
**Is your feature improvement request related to a problem? Please describe.**

The `tedge` CLI framework is used in `thin-edge` devices to create/update configs, create certificates, and connect to the cloud. It's also used to subscribe and publish MQTT messages. If something goes wrong while the CLI framework is being used, there is no way to debug, it can’t print more information about what is happening.

For example it is useful to set the log level to debug when things aren't working as expected, e.g. when trying to connect thin-edge.io to Cumulocity and you want more details like certificate info etc.

```sh
tedge connect c8y --log-level trace
```

**Describe the solution you'd like**

Enable the debug feature to `tedge` the CLI framework,

* Add a `--log-level <level>` flag to all tedge binaries (e.g. `tedge`, tedge-agent`, `tedge-mapper`

**Describe alternatives you've considered**

- Add `--debug` flag. When debug flag is enabled, this will print the `info, warn, debug, trace` log messages. 
- Add log messages in the `tedge` code at appropriate places to make it easy to debug

**Additional context**

reubenmiller:
For those who are interested, you can enable detailed logging by setting the `RUST_LOG` variable before calling the command, e.g.

```
sudo RUST_LOG=debug tedge connect c8y
```

```
The system config file '/etc/tedge/system.toml' doesn't exist. Use '/bin/systemctl' as a service manager.

2023-11-07T09:54:13.508234073Z DEBUG tedge_config::tedge_config_cli::tedge_config_repository: Loading configuration from "/etc/tedge/tedge.toml"
Detected mosquitto version >= 2.0.0
Checking if systemd is available.

Checking if configuration for requested bridge already exists.

Validating the bridge certificates.

Creating the device in Cumulocity cloud.

2023-11-07T09:54:13.678628567Z DEBUG rustls::client::hs: No cached session for DnsName("example.c8y.io")
2023-11-07T09:54:13.678969297Z DEBUG rustls::client::hs: Not resuming any session
2023-11-07T09:54:13.719520593Z DEBUG rustls::client::hs: Using ciphersuite TLS13_AES_256_GCM_SHA384
2023-11-07T09:54:13.719630333Z DEBUG rustls::client::tls13: Not resuming
[...]

PR #3205, opened 2024-10-24
Add --level flag

Bravo555:
## Proposed changes

Add a `--level` flag that sets the verbosity of reported logs.

In contrast to `--debug`, which enables `debug` logs in addition to the default `error`, `warn`, and `info` logs, `--level` sets the maximum verbosity, i.e. when set to `error` print only errors, if set to `warn` print warnings and errors, etc. and when set to `trace`, print all the logs.

If `--level` is selected, this value is used and takes precedence over `--debug`. `RUST_LOG` env variable still overrides the flag. 

## Types of changes

## Paste Link to the issue
- #2089 

## Checklist

## Further comments

There were 4 places where an identical change was necessary, perhaps we should consider how to extract all the common CLI flags into a shared component.


crates/core/tedge_mapper/src/lib.rs
```diff
@@ -122,14 +128,19 @@ pub async fn run(mapper_opt: MapperOpt) -> anyhow::Result<()> {
...
+    // If `debug` is `false` then only `error!`, `warn!` and `info!` are reported.
+    // If `debug` is `true` then also `debug!` is reported.
+    // If neither was provided, use a log level from a config file.
+    let log_level = mapper_opt
+        .log_level
+        .or(mapper_opt.debug.then_some(tracing::Level::DEBUG));
+
+    let log_level = match log_level {
+        Some(log_level) => log_level,
+        None => get_log_level("tedge-agent", &tedge_config_location.tedge_config_root_path)?,
```
albinsuresh, line 141:
```suggestion
        None => get_log_level("tedge-mapper", &tedge_config_location.tedge_config_root_path)?,
```

crates/core/tedge_watchdog/src/lib.rs
```diff
@@ -46,13 +52,17 @@ pub async fn run(watchdog_opt: WatchdogOpt) -> Result<(), anyhow::Error> {
...
+    // If `debug` is `false` then only `error!`, `warn!` and `info!` are reported.
+    // If `debug` is `true` then also `debug!` is reported.
+    // If neither was provided, use a log level from a config file.
+    let log_level = watchdog_opt
+        .log_level
+        .or(watchdog_opt.debug.then_some(tracing::Level::DEBUG));
+
+    let log_level = match log_level {
+        Some(log_level) => log_level,
+        None => get_log_level("tedge-agent", &tedge_config_location.tedge_config_root_path)?,
```
albinsuresh, line 65:
```suggestion
        None => get_log_level("tedge-watchdog", &tedge_config_location.tedge_config_root_path)?,
```

plugins/c8y_firmware_plugin/src/lib.rs
```diff
@@ -67,10 +73,18 @@ pub async fn run(firmware_plugin_opt: FirmwarePluginOpt) -> Result<(), anyhow::E
...
+    // If `debug` is `false` then only `error!`, `warn!` and `info!` are reported.
+    // If `debug` is `true` then also `debug!` is reported.
+    // If neither was provided, use a log level from a config file.
+    let log_level = firmware_plugin_opt
+        .log_level
+        .or(firmware_plugin_opt.debug.then_some(tracing::Level::DEBUG));
+
+    let log_level = match log_level {
+        Some(log_level) => log_level,
+        None => get_log_level("tedge-agent", &tedge_config_location.tedge_config_root_path)?,
```
albinsuresh, line 87:
```suggestion
        None => get_log_level("c8y-firmware-plugin", &tedge_config_location.tedge_config_root_path)?,
```

albinsuresh (commented):
A few copy paste issues.

didier-wenzek (commented):
> There were 4 places where an identical change was necessary, perhaps we should consider how to extract all the common CLI flags into a shared component.

Could this be implemented in `tedge::main.rs`, the multi-call entry for the agent and the mapper?

reubenmiller:
> If --log-level is selected, this value is used and takes precedence over --debug. RUST_LOG env variable still overrides the flag.

Flags should always take precedence over environment variable values...generally flags are often used by users to override the environment which they usually have little control over

Bravo555:
> Flags should always take precedence over environment variable values...generally flags are often used by users to override the environment which they usually have little control over.
> 
> The first non-empty value should be taken (in order):
> 
>     1. Flag
> 
>     2. Environment variable
> 
>     3. Configuration file

Made a dedicated PR that fixes that mistake: #3208


> Could this be implemented in `tedge::main.rs`, the multi-call entry for the agent and the mapper?

That would probably require some more changes because for every component we use a separate CLI parser, but for now #3208 eliminates some duplication.

Bravo555:
Although, would it be a better user experience to only allow using either `--debug` OR `--log-level`, but not both? It might eliminate some confusion of which overrides the other.

didier-wenzek (approved):
Approved

didier-wenzek:
> Although, would it be a better user experience to only allow using either `--debug` OR `--log-level`, but not both? It might eliminate some confusion of which overrides the other.

Indeed these two flags overlap. But, I don't see that as a big deal.

PR #3210, opened 2024-10-28
fix: agent leaves interrupted operations in non-final state 

didier-wenzek:
## Proposed changes

- [x] Reproduce https://github.com/thin-edge/thin-edge.io/issues/3149
- [x] Correctly emit over MQTT the state of  commands interrupted by a restart.

## Types of changes

## Paste Link to the issue

Reproduce https://github.com/thin-edge/thin-edge.io/issues/3149

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/operation_workflows/actor.rs
+++ b/crates/core/tedge_agent/src/operation_workflows/actor.rs
@@ -514,6 +514,10 @@ impl WorkflowActor {
                     .workflow_repository
                     .load_pending_commands(pending_commands)
                 {
+                    // Make sure the latest state is visible over MQTT
+                    self.mqtt_publisher
+                        .send(command.clone().into_message())
+                        .await?;
                     self.process_command_update(command.clone()).await?;
                 }
             }
--- a/tests/RobotFramework/tests/tedge_agent/workflows/dynamic_workflow_reloading.robot
+++ b/tests/RobotFramework/tests/tedge_agent/workflows/dynamic_workflow_reloading.robot
@@ -204,6 +204,26 @@ Resume On Restart A Pending Operation Which Workflow Is Deprecated
     ...    timeout=60
     Should Contain    ${messages[0]}    item="what a long sleep"
 
+Resume On Restart A Pending Operation
+    # Trigger a long running operation
+    Update Workflow    ${CURDIR}/sleep-command.toml    sleep
+    Execute Command
+    ...    tedge mqtt pub --retain te/device/main///cmd/sleep/dyn-test-12 '{"status":"init", "duration":120}'
+
+    # Restart the agent, once sure the command is executing
+    Should Have MQTT Messages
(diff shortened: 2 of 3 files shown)


didier-wenzek:
The test "Update tedge version from base to current using Cumulocity" failed for unrelated reasons. Not obviously flaky though:

```
$ invoke flake-finder --test-name "Update tedge version from base to current using Cumulocity" --iterations 10 --outputdir output_ff --clean
Overall: PASSED
Results: 10 iterations, 10 passed, 0 failed
Elapsed time: 0:10:42.849008
```

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -514,6 +514,10 @@ impl WorkflowActor {
                     .workflow_repository
                     .load_pending_commands(pending_commands)
                 {
+                    // Make sure the latest state is visible over MQTT
+                    self.mqtt_publisher
```
albinsuresh, line 518: Should we limit this republishing only for commands in their terminal state? If not, there's the risk of the even the intermediate command states like `scheduled` or `executing` getting resent, risking duplicate execution of the same logic by any external component listening on them.

reubenmiller: Though there isn't any official contract that says these messages will only be published exactly once. But generally such listeners are not doing control (as control logic should be in the workflow itself), so triggering twice on a status change should not be a problem.

reubenmiller: Just using qos 1 is enough to receive a duplicate message (without publishing the status change twice).

didier-wenzek: > Should we limit this republishing only for commands in their terminal state?

This would introduce the same issue on intermediate states (.e.g the "executing" state being never published).

> If not, there's the risk of the even the intermediate command states like `scheduled` or `executing` getting resent, risking duplicate execution of the same logic by any external component listening on them.

The agent protects itself from reacting to messages of which it is the publisher.

Other components, such as the mappers, should be prepared to receive a status twice.

> Just using qos 1 is enough to receive a duplicate message (without publishing the status change twice).

It's also enough to restart, the messages being retained. If the c8y mapper stop after sending an EXECUTING 501 message and restart before the command is successful, a second EXECUTING 501 will be sent.

albinsuresh (approved):
LGTM

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: > Just using qos 1 is enough to receive a duplicate message (without publishing the status change twice).

I have always assumed that users would send critical messages like commands only with QoS 2, as they might expect once-and-only-once execution for such things. But yeah, if the users are already expected to be resilient to duplicate delivery, esp since most cloud platforms don't support QoS 2 either, I guess this is okay.

PR #3217, opened 2024-10-30
Attach topic to c8y message conversion error

Bravo555:
## Proposed changes

Improve error reporting in c8y mapper by printing the topic when there was an error converting a message on that topic.

Example: when publishing an invalid command metadata message: `$ tedge mqtt pub -r te/device/main///cmd/config_snapshot 'a'`

before:
`ERROR: Mapping error: expected value at line 1 column 1`

after:
`ERROR: Mapping error: Failed to convert a message on topic 'te/device/main///cmd/config_snapshot': expected value at line 1 column 1`

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -7,6 +7,7 @@ use crate::actor::IdDownloadRequest;
 use crate::actor::IdDownloadResult;
 use crate::dynamic_discovery::DiscoverOp;
 use crate::error::ConversionError;
+use crate::error::MessageConversionError;
 use crate::json;
 use crate::operations;
 use crate::operations::OperationHandler;
@@ -121,7 +122,7 @@ pub struct MapperConfig {
 impl CumulocityConverter {
     pub async fn convert(&mut self, input: &MqttMessage) -> Vec<MqttMessage> {
         let messages_or_err = self.try_convert(input).await;
-        self.wrap_errors(messages_or_err)
+        self.wrap_errors_with_input(messages_or_err, input)
     }
 
     pub fn wrap_errors(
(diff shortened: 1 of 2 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -131,8 +132,26 @@ impl CumulocityConverter {
...
+    ) -> Vec<MqttMessage> {
+        let error = match messages_or_err {
+            Ok(messages) => return messages,
+            Err(error) => error,
+        };
+
+        let error = MessageConversionError {
+            error,
+            topic: input.topic.name.clone(),
+        };
```
albinsuresh, line 148: How about moving this `MessageConversionError` creation logic into the `convert` function and propagate the same type into `wrap_errors` and `new_error_message` functions as well, since the `new_error_message` function already has the logic below that creates the message on the error topic. That

Bravo555: `wrap_errors` is also used in other places where we don't have a message we're converting, for example auto entity registration. Then using the topic of a message that causes auto-registration (for example some measurement) next to any errors about registration could cause confusion

albinsuresh: Okay, makes sense.

In that case, we can just change the signature of the `new_error_message` function to accept a `std::error::Error` instead of `ConversionError` so that this whole function can be shortened as:

```
message_or_err
    .map_err(
        |error| MessageConversionError {
            error,
            topic: input.topic.name.clone(),
        })
    .unwrap_or_else(|error| self.new_error_message(error))
```

albinsuresh (approved):

Issue #2938, opened 2024-06-13
Creating an event should use the local c8y proxy


rina23q:
**Describe the bug**
When tedge-mapper-c8y tries to upload a file for a log operation, it needs to create two requests
1. Create an event so that tedge-mapper-c8y can get an event ID
2. Upload a file by attaching to the event created by the step 1.

However, I found that the step 1 is done in the old way, using `bearer` auth and actual http host and port (like `https://example.cumulocity.com:443/event/events`). The step 2 is already using the local c8y proxy endpoint.

Here is the proof from our code:

At the end of one `log_upload` command, c8y-mapper calls this function.
https://github.com/thin-edge/thin-edge.io/blob/1c21e8756f46a79bed26c7dac927bcdb73d32462/crates/extensions/c8y_mapper_ext/src/operations/log_upload.rs#L225-L235

Then, first it tries to get the `event_response_id`.
https://github.com/thin-edge/thin-edge.io/blob/1c21e8756f46a79bed26c7dac927bcdb73d32462/crates/extensions/c8y_mapper_ext/src/converter.rs#L1744

This one is called as part of `http_proxy`, which is a part of `C8yHttpProxyActor`. The function `send_event()` calls the function `send_event_internal()` eventually. You see now the `create_event_url` holds the actual URL.  
https://github.com/thin-edge/thin-edge.io/blob/1c21e8756f46a79bed26c7dac927bcdb73d32462/crates/extensions/c8y_http_proxy/src/actor.rs#L508-L536

And if we go further, the `execute` function uses `bearer` authentication. It's a proof that the `http_proxy.send_event()` doesn't use the local c8y proxy.

https://github.com/thin-edge/thin-edge.io/blob/1c21e8756f46a79bed26c7dac927bcdb73d32462/crates/extensions/c8y_http_proxy/src/actor.rs#L258

**To Reproduce**
Actually I found this issue while reading the code. If you sniff the HTTP requests, you can confirm it.

**Expected behavior**
To get an event, we should also use local c8y proxy endpoint. Probably, we can stop using the `C8yHttpProxyActor` completely. 

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [1.1.1]

**Additional context**

PR #3224, opened 2024-11-05
refactor: simplify c8y_http_proxy

didier-wenzek:
## Proposed changes

Can `c8y_http_proxy` be fully deprecated and replaced by `c8y_auth_proxy` ?

As noted [here](https://github.com/thin-edge/thin-edge.io/issues/2938) the c8y mapper is already bypassing `c8y_http_proxy`  in favor of  `c8y_auth_proxy` but only partially: to upload the file attached to an event but not to create that event.

Actually, `c8y_mapper_ext` (and also `c8y_firmware_manager`) can be independent of `c8y_http_proxy`  and  `c8y_auth_proxy` altogether. Indeed, the key point is related to url rewriting.

1. To use the local auth proxy to Cumulocity, the mapper has simply to rewrite the urls provided by `c8y_api::http_proxy::C8yEndPoint` using the `c8y_auth_proxy::url::ProxyUrlGenerator`.  There is no need for a direct dependency on the `c8y_auth_proxy`  crate beyond that url generator.
2. All the c8y urls being rewritten to go through the local auth proxy, there is no more needs for the mapper to fetch JWT tokens. Hence, the code of the  `c8y_http_proxy`  can be simplified to focus only one URL generation.

Tasks:

- [x] Remove unused method C8YHttpProxy::upload_log_binary
- [x] Move ProxyUrlGenerator from c8y_auth_proxy to c8y_api
- [x] Combine C8yEndPoint with ProxyUrlGenerator: enforce the use of the auth proxy for c8y urls
- [x] Remove JWT fetching from  `c8y_http_proxy`
- [x] Remove  the dependencies on  `c8y_http_proxy` from `c8y_auth_proxy`, `c8y_firmware_plugin`, and `c8y_firmware_manager` (this is only to get some struct definitions).
- [ ] Fully remove  `c8y_http_proxy`.

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/2938

## Checklist


crates/extensions/c8y_mapper_ext/src/operations/handlers/config_update.rs
```diff
@@ -120,7 +120,7 @@ mod tests {
                 "te/device/main///cmd/config_update/c8y-mapper-123456",
                 json!({
                     "status": "init",
-                    "remoteUrl": "http://www.my.url",
+                    "remoteUrl": "http://www.my.url/",
```
didier-wenzek, line 123: It would be better to understand why the new method `local_proxy_url` is adding a trailing `/`.

Bravo555: In accordance with the [URL spec](https://github.com/servo/rust-url/issues/808), `Url::parse` by default adds a `/` character after the host if the path is empty.
In our case we parse the URL and use a `.to_string()` here:

https://github.com/thin-edge/thin-edge.io/pull/3224/files#diff-6806117b091373774a361e1c5eaf2ba17e029724012bc70656e85fe99b2c87cdR271

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -34,33 +36,52 @@ pub enum C8yEndPointError {
...
+            .to_string();
+        let c8y_mqtt_host = tedge_config
+            .c8y
+            .try_get(c8y_profile)?
+            .mqtt
+            .or_config_not_set()?
+            .to_string();
+
+        let c8y_config = tedge_config.c8y.try_get(c8y_profile)?;
+        let auth_proxy_addr = c8y_config.proxy.client.host.clone();
```
albinsuresh, line 61:
```suggestion
        let c8y_config = tedge_config.c8y.try_get(c8y_profile)?;
        let c8y_host = c8y_config
            .http
            .or_config_not_set()?
            .to_string();
        let c8y_mqtt_host = c8y_config
            .mqtt
            .or_config_not_set()?
            .to_string();

        let auth_proxy_addr = c8y_config.proxy.client.host.clone();
```

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -1,445 +1,84 @@
...
-            let endpoint = request.uri().path().to_owned();
-            let method = request.method().to_owned();
-
-            match self.peers.http.await_response(request).await? {
-                Ok(response) => {
-                    match response.status() {
-                        StatusCode::OK => {
-                            let internal_id_response: InternalIdResponse = response.json().await?;
-                            let internal_id = internal_id_response.id();
+    pub async fn try_get_internal_id(&mut self, device_id: &str) -> Result<String, C8YRestError> {
```
albinsuresh, line 32: I'm assuming that even the remaining functions in this "non-actor" would be moved to `handle` in the next step.

albinsuresh (commented):
Everything looks fine, except for dropping the `get_internal_id` retry logic which may have other side effects. Happy to approve once that comment is resolved.

crates/extensions/c8y_http_proxy/src/actor.rs
```diff
@@ -29,59 +24,20 @@ pub struct C8YHttpProxyActor {
...
-                },
-
-                Err(e) => return Err(C8YRestError::FromHttpError(e)),
-            }
-        }
+        let http_result = self.http.await_response(request).await?;
+        let http_response = http_result.error_for_status()?;
+        let internal_id_response: InternalIdResponse = http_response.json().await?;
+        let internal_id = internal_id_response.id();
+        Ok(internal_id)
```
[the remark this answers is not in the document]
albinsuresh: Ignore that comment. That internal id lookup was valid only when we were eagerly fetching the internal id of the main device during the startup itself (in the `C8yHttpActor::init()`). Now that we don't have that `init` logic and internal id fetched always on-demand, this shouldn't be an issue.

Bravo555 (approved):
Happy to approve as it is and defer the full removal of `c8y_http_proxy` to another PR, as this one is already quite big.

crates/extensions/c8y_http_proxy/src/actor.rs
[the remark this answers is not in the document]
didier-wenzek: Yes, this was one of the intent, however the removed piece of code was useless in case of internet disruptions: TCP and TLS errors breaking the retry loop.

The only place where a retry might is useful is when publishing the software list on start. But this should be done at a different level - in the c8y mapper and not blindly at the HTTP level.

crates/core/c8y_api/src/http_proxy.rs
didier-wenzek: Done

PR #3225, opened 2024-11-05
feat: add workflow support for custom operation (drop 2)

Ruadhri17:
## Proposed changes
TODO:

- [x] thin-edge command conversion
- [x] subscribe to command topics for custom operation
- [x] state change handler
- [x] main device system test
- [x] child device system test
- [x] fix bug with `externalID` 
- [x] fix codecov (maybe)

## Types of changes

## Paste Link to the issue
* #3095 
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/file.rs
+++ b/crates/common/tedge_utils/src/file.rs
@@ -2,6 +2,7 @@ use futures::TryFutureExt;
 use nix::unistd::*;
 use std::fs;
 use std::io;
+use std::io::Error;
 use std::io::Write;
 use std::os::unix::fs::MetadataExt;
 use std::os::unix::fs::PermissionsExt;
@@ -64,6 +65,12 @@ pub enum FileError {
 
     #[error(transparent)]
     FileMove(#[from] FileMoveError),
+
+    #[error("Failed to create a symlink {link:?}: {source:?}.")]
+    CreateSymlinkFailed {
+        link: PathBuf,
+        source: std::io::Error,
+    },
 }
 
 pub fn create_directory<P: AsRef<Path>>(
@@ -505,6 +512,29 @@ pub fn has_write_access(path: &Path) -> Result<(), FileError> {
     }
 }
 
+pub fn create_symlink(original: impl AsRef<Path>, link: impl AsRef<Path>) -> Result<(), FileError> {
+    match std::os::unix::fs::symlink(&original, &link) {
+        Ok(_) => Ok(()),
+        Err(e) if e.kind() == io::ErrorKind::AlreadyExists => match fs::read_link(&link) {
+            Ok(path) if path == original.as_ref() => Ok(()),
+            Ok(_) => Err(FileError::CreateSymlinkFailed {
+                link: link.as_ref().to_path_buf(),
+                source: Error::other(format!(
+                    "symlink exists but does not point to {:?}",
+                    original.as_ref()
+                )),
+            }),
+            Err(e) => Err(FileError::CreateSymlinkFailed {
+                link: link.as_ref().to_path_buf(),
+                source: e,
+            }),
+        },
+        Err(e) => Err(FileError::CreateSymlinkFailed {
+            link: link.as_ref().to_path_buf(),
+            source: e,
+        }),
+    }
+}
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -674,6 +704,41 @@ mod tests {
         assert_ne!(get_gid_by_name("staff").unwrap(), 0);
         let err = get_gid_by_name("nonexistent_group").unwrap_err();
         assert!(err.to_string().contains("Group not found"));
+    }
+
+    #[test]
+    fn create_new_symlink() {
+        let temp_dir = TempDir::new().unwrap();
+        let source_path = temp_dir.path().join("source_file").display().to_string();
+        let another_source_path = temp_dir
+            .path()
+            .join("another_source_file")
+            .display()
+            .to_string();
+        let invalid_source_path = temp_dir.path().join("invalid_file").display().to_string();
+        let dest_path = temp_dir.path().join("dest_file").display().to_string();
+
+        // create symlink
+        create_file_with_user_group(&source_path, &USER, &GROUP, 0o644, None).unwrap();
+        assert!(Path::new(source_path.as_str()).exists());
+        create_symlink(&source_path, &dest_path).unwrap();
+        assert!(Path::new(dest_path.as_str()).exists());
+
+        // creating symlink again should not return error if source is the same
+        assert!(create_symlink(&source_path, &dest_path).is_ok());
+
+        // creating symlink again should  return error if source is different
+        create_file_with_user_group(&another_source_path, &USER, &GROUP, 0o644, None).unwrap();
+        assert!(Path::new(another_source_path.as_str()).exists());
+        let err = create_symlink(&another_source_path, &dest_path).unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("symlink exists but does not point to")
+                && err.to_string().contains(another_source_path.as_str())
+        );
+
+        // creating symlink should not be possible to file that does not exists
+        assert!(create_symlink(invalid_source_path, dest_path).is_err());
     }
 
     #[tokio::test]
--- a/crates/core/c8y_api/src/smartrest/operations.rs
+++ b/crates/core/c8y_api/src/smartrest/operations.rs
@@ -26,6 +26,7 @@ const DEFAULT_FORCEFUL_TIMEOUT: Duration = Duration::from_secs(60);
 #[derive(Debug, Default, Clone, PartialEq, Eq)]
 pub struct Operations {
     operations: Vec<Operation>,
+    templates: Vec<Operation>,
 }
 
 impl Operations {
@@ -52,6 +53,10 @@ impl Operations {
         bridge_config: &impl Record,
     ) -> Result<Self, OperationsError> {
         get_operations(dir.as_ref(), bridge_config)
+    }
+
+    pub fn add_template(&mut self, template: Operation) {
+        self.templates.push(template);
     }
 
     pub fn get_operations_list(&self) -> Vec<String> {
@@ -127,6 +132,41 @@ impl Operations {
             .collect::<HashSet<String>>()
             .try_into()?)
     }
+
+    /// Return operation name if `workflow.operation` matches
+    pub fn get_operation_name_by_workflow_operation(&self, command_name: &str) -> Option<String> {
+        let matching_templates: Vec<&Operation> = self
+            .templates
+            .iter()
+            .filter(|template| {
+                template
+                    .workflow_operation()
+                    .is_some_and(|operation| operation.eq(command_name))
+            })
+            .collect();
+
+        if matching_templates.len() > 1 {
+            warn!(
+                "Found more than one template with the same `workflow.operation` field. Picking {}",
+                matching_templates.first().unwrap().name
+            );
+        }
+
+        matching_templates
+            .first()
+            .and_then(|template| template.on_fragment())
+    }
+
+    pub fn get_template_name_by_operation_name(&self, operation_name: &str) -> Option<&str> {
+        self.templates
+            .iter()
+            .find(|template| {
+                template
+                    .on_fragment()
+                    .is_some_and(|name| name.eq(operation_name))
+            })
+            .map(|template| template.name.as_ref())
+    }
 }
 
 #[derive(Debug, Clone, Deserialize, Eq, PartialEq)]
@@ -144,6 +184,22 @@ impl Operation {
 
     pub fn command(&self) -> Option<String> {
         self.exec().and_then(|exec| exec.command.clone())
+    }
+
+    pub fn workflow_operation(&self) -> Option<&str> {
+        self.exec().and_then(|exec| {
+            exec.workflow
+                .as_ref()
+                .and_then(|workflow| workflow.operation.as_deref())
+        })
+    }
+
+    pub fn workflow_input(&self) -> Option<&str> {
+        self.exec().and_then(|exec| {
+            exec.workflow
+                .as_ref()
+                .and_then(|workflow| workflow.input.as_deref())
+        })
     }
 
     pub fn topic(&self) -> Option<String> {
@@ -230,8 +286,13 @@ impl Operation {
             if self.on_fragment().is_none() {
                 return Err(InvalidCustomOperationHandler::MissingOnFragment);
             }
+
             if self.command().is_none() {
-                return Err(InvalidCustomOperationHandler::MissingCommand);
+                if self.workflow_operation().is_none() {
+                    return Err(InvalidCustomOperationHandler::MissingCommand);
+                }
+            } else if self.workflow_operation().is_some() || self.workflow_input().is_some() {
+                return Err(InvalidCustomOperationHandler::CommandExists);
             }
         }
         Ok(())
@@ -246,6 +307,7 @@ pub struct OnMessageExec {
     on_fragment: Option<String>,
     topic: Option<String>,
     user: Option<String>,
+    workflow: Option<ExecWorkflow>,
     #[serde(default)]
     skip_status_update: bool,
     #[serde(default, deserialize_with = "to_result_format")]
@@ -283,7 +345,14 @@ where
     }
 }
 
-pub fn get_operations(
+#[derive(Debug, Clone, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "snake_case")]
+struct ExecWorkflow {
+    operation: Option<String>,
+    input: Option<String>,
+}
+
+fn get_operations(
     dir: impl AsRef<Path>,
     bridge_config: &impl Record,
 ) -> Result<Operations, OperationsError> {
@@ -295,7 +364,7 @@ pub fn get_operations(
         .map(|entry| entry.map(|e| e.path()))
         .collect::<Result<Vec<PathBuf>, _>>()?
         .into_iter()
-        .filter(|path| path.is_file())
+        .filter(|path| path.is_file() || path.is_symlink())
         .collect::<Vec<PathBuf>>();
 
     for path in dir_entries {
@@ -303,7 +372,7 @@ pub fn get_operations(
             if !is_valid_operation_name(file_name) {
                 // Warn user about invalid operation names, otherwise the
                 // user does not know that the operation is being ignored
-                warn!("Ignoring custom operation definition as the filename uses an invalid character. Only [A-Za-z0-9_] characters are accepted. file={}", path.display());
+                warn!("Ignoring custom operation definition as the filename uses an invalid character. Only [A-Za-z0-9_.] characters are accepted. file={}", path.display());
                 continue;
             }
 
@@ -320,10 +389,17 @@ pub fn get_operations(
             };
 
             if details.is_valid_operation_handler() || details.is_supported_operation_file() {
-                operations.add_operation(details);
+                match path.extension() {
+                    None => operations.add_operation(details),
+                    Some(extension) if extension.eq("template") => operations.add_template(details),
+                    Some(_) => {
+                        return Err(OperationsError::InvalidOperationName(path.to_owned()));
+                    }
+                };
             }
         }
     }
+
     Ok(operations)
 }
 
@@ -380,9 +456,15 @@ pub fn get_operation(
 /// this `operation_name_is_valid` fn will ensure that only files that do not contain
 /// any special characters are allowed.
 pub fn is_valid_operation_name(operation: &str) -> bool {
-    operation
-        .chars()
-        .all(|c| c.is_ascii_alphabetic() || c.is_numeric() || c.eq(&'_'))
+    match operation.split_once('.') {
+        Some((file_name, extension)) if extension.eq("template") => file_name
+            .chars()
+            .all(|c| c.is_ascii_alphanumeric() || c.eq(&'_')),
+        None => operation
+            .chars()
+            .all(|c| c.is_ascii_alphanumeric() || c.eq(&'_')),
+        _ => false,
+    }
 }
 
 fn default_graceful_timeout() -> Duration {
@@ -413,8 +495,11 @@ pub enum InvalidCustomOperationHandler {
     #[error("'on_fragment' is missing")]
     MissingOnFragment,
 
-    #[error("'command' is missing")]
+    #[error("'command' or 'workflow.operation' is missing")]
     MissingCommand,
+
+    #[error("'command' should not be provided with 'workflow.operation' or 'workflow.input'")]
+    CommandExists,
 }
 
 #[cfg(test)]
@@ -566,7 +651,7 @@ mod tests {
     #[test_case("~file_b", false)]
     #[test_case("c8y_Command", true)]
     #[test_case("c8y_CommandA~", false)]
-    #[test_case(".c8y_CommandB", false)]
+    #[test_case("c8y_CommandB.template", true)]
     #[test_case("c8y_CommandD?", false)]
     #[test_case("c8y_CommandE?!£$%^&*(", false)]
     #[test_case("?!£$%^&*(c8y_CommandF?!£$%^&*(", false)]
@@ -666,6 +751,7 @@ mod tests {
 
         let ops = Operations {
             operations: vec![operation1.clone(), operation2.clone()],
+            ..Default::default()
         };
 
         let prefix = TopicPrefix::from_str("c8y").unwrap();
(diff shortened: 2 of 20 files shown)


crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -218,6 +257,8 @@ impl Operation {
 #[serde(rename_all = "snake_case")]
 pub struct OnMessageExec {
     command: Option<String>,
+    command_name: Option<String>,
+    command_payload: Option<String>,
```
Bravo555, line 261: question: these fields are meant to be for `.template` files, but right now it's technically possible for them to be in non-template files as well, right?

Bravo555 (changes requested):
Blockers:
- function that creates symlinks ignores some errors
- memory leak in `From<CumulocitySupportedOperations> for &str`
- typo

Other than that, some redundant branches could use some more terseness. Modified modules could also use some general cleanup, but that can be defered to another PR.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -271,19 +312,24 @@ pub fn get_operations(
...
-                warn!("Ignoring custom operation definition as the filename uses an invalid character. Only [A-Za-z0-9_] characters are accepted. file={}", path.display());
+                warn!("Ignoring custom operation definition as the filename uses an invalid character. Only [A-Za-z0-9_.] characters are accepted. file={}", path.display());
                 continue;
             }
 
-            let mut details = match fs::read(&path) {
+            let original_path = match fs::read_link(&path) {
+                Ok(original_path) => original_path,
+                Err(_) => path.clone(),
+            };
```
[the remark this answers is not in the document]
Ruadhri17: Yes, that is leftover after previous solution. I will remove it in follow-up fixup commit

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_workflow/custom_operation_workflow.robot
```diff
@@ -0,0 +1,59 @@
...
+
+Test Tags           theme:c8y    theme:troubleshooting    theme:plugins
+
+
+*** Test Cases ***
+Run custom operation with workflow execution
+    # Capability/Supported Operations check
+    Execute Command    tedge mqtt pub -r te/device/main///cmd/take_picture '{}'
+
+    # Better to create and replace with "Symlink Should Exist"
```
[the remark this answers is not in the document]
rina23q: It's done. Now `Symlink Should Exist` keyword is available.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1543,9 +1592,19 @@ impl CumulocityConverter {
...
-        let ops_file = ops_dir.join(c8y_operation_name);
+
         create_directory_with_defaults(&*ops_dir)?;
-        create_file_with_defaults(ops_file, None)?;
+        if let Some((template_name, Some(operation_name))) = self
+            .operations
+            .match_template_with_operation_by_command_name(c8y_operation_name)
+        {
+            let template_file = self.config.ops_dir.join(&template_name);
+            let ops_file = ops_dir.join(&operation_name);
```
rina23q, line 1602: This `operation_name` is equal to `c8y_operation_name`.

```suggestion
            let ops_file = ops_dir.join(&c8y_operation_name);
```

It means we can simplfy `match_template_with_operation_by_command_name()`. It can return only single value.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1543,9 +1592,19 @@ impl CumulocityConverter {
                 return Ok(vec![]);
             }
         };
-        let ops_file = ops_dir.join(c8y_operation_name);
+
         create_directory_with_defaults(&*ops_dir)?;
-        create_file_with_defaults(ops_file, None)?;
+        if let Some((template_name, Some(operation_name))) = self
+            .operations
+            .match_template_with_operation_by_command_name(c8y_operation_name)
```
rina23q, line 1599: It took a lot of time to understand why this function works. It seems that the argument (`c8y_operation_name`) is the c8y's operation name (e.g. `c8y_Command`), however, the function checks if the template's command name (e.g. `command`) equals to the c8y operation name.

```rust
    /// Return operation and template name if `command_name` match incoming command
    pub fn match_template_with_operation_by_command_name(
        &self,
        command: &str,
    ) -> Option<(String, Option<String>)> {
        self.templates
            .iter()
            .find(|x| {
                x.command_name()
                    .is_some_and(|command_name| command_name.eq(command))
            })
            .map(|template| (template.name.clone(), template.on_fragment()))
    }
```

But then why it works? I figured it out that the problem deeper, is in `try_convert_data_message()`.

```rust
        match &channel {
            ...
                Channel::CommandMetadata { operation } => {
                    ...
                    OperationType::Custom(c8y_op_name) => {
                        dbg!(&c8y_op_name);
                        self.register_custom_operation(&source, c8y_op_name)
                    }
                }
            ...
        }
```

`OperationType::Custom(c8y_op_name)` holds actually **command** name! Not **c8y_op_name**!
Indeed I added as above, and it prints out `command`.
```
[crates/extensions/c8y_mapper_ext/src/converter.rs:1361:25] c8y_op_name = "command"
```

This needs to be fixed. Here `c8y_op_name` should hold the right value. Then, unfortunately, the function `match_template_with_operation_by_command_name()` needs to be rewrite to check if `template.on_fragment()` is equals to the c8y operation.

rina23q (changes requested):
Unfortunately, there is mistake in converter for custom operation. It needs to be fixed first, then the implementation in this PR should be modified.

crates/core/c8y_api/src/smartrest/operations.rs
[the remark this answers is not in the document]
Ruadhri17: fixed by 451bb9146abda49a6cbe646643bf0006349628db

crates/common/tedge_utils/src/file.rs
```diff
@@ -505,6 +505,13 @@ pub fn has_write_access(path: &Path) -> Result<(), FileError> {
     }
 }
 
+pub fn create_symlink(original: impl AsRef<Path>, link: impl AsRef<Path>) -> Result<(), FileError> {
+    if !link.as_ref().is_symlink() {
+        std::os::unix::fs::symlink(original, link)?;
+    }
```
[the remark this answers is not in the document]
Ruadhri17: fixed by 451bb9146abda49a6cbe646643bf0006349628db

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -103,6 +115,20 @@ impl Operations {
...
+
+    /// Return operation and template name if `command_name` match incoming command
+    pub fn match_template_with_operation_by_command_name(
+        &self,
+        command: &str,
+    ) -> Option<(String, Option<String>)> {
+        self.templates
+            .iter()
+            .find(|x| {
+                x.command_name()
```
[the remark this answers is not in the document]
Ruadhri17: fixed by 451bb9146abda49a6cbe646643bf0006349628db

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -196,6 +197,7 @@ impl From<CumulocitySupportedOperations> for &'static str {
             CumulocitySupportedOperations::C8yDownloadConfigFile => "c8y_DownloadConfigFile",
             CumulocitySupportedOperations::C8yFirmware => "c8y_Firmware",
             CumulocitySupportedOperations::C8yDeviceProfile => "c8y_DeviceProfile",
+            CumulocitySupportedOperations::C8yCustom(operation) => operation.leak(),
```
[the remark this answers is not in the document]
Ruadhri17: fixed by 24f3203ab59882b6db7b682feb137f017ec74f75

crates/core/c8y_api/src/smartrest/smartrest_serializer.rs
```diff
@@ -303,7 +305,7 @@ pub trait C8yOperation {
 
 impl C8yOperation for CumulocitySupportedOperations {
     fn name(&self) -> &str {
-        (*self).into()
+        self.clone().into()
```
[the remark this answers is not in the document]
Ruadhri17: fixed by 24f3203ab59882b6db7b682feb137f017ec74f75

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -753,27 +766,63 @@ impl CumulocityConverter {
...
+                    break;
+                }
             }
         }
 
         // MQTT messages are sent during the operation execution
         Ok(vec![])
     }
 
+    fn get_opeartion_handlers(
```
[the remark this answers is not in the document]
Ruadhri17: fixed by 2bedc99af4cc8593b01f143a8d38dab5483954b2

crates/extensions/c8y_mapper_ext/src/converter.rs
Ruadhri17: After discussion with @rina23q and @albinsuresh, the fix was implemented in 451bb9146abda49a6cbe646643bf0006349628db

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -735,14 +741,21 @@ impl CumulocityConverter {
...
 
         let result = self
-            .process_json_custom_operation(operation.op_id.clone(), &operation.extras, message)
+            .process_json_custom_operation(
+                operation.op_id.clone(),
+                cmd_id,
+                device_xid,
+                &operation.extras,
+                message,
+            )
```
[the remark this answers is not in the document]
Ruadhri17: Unfortunately, this function is called from two places: `parse_json_custom_operation_topic()` and `process_json_over_mqtt()` as both `C8yDeviceControlTopic` and custom topics can handle this kind of operation. In the latter one all that parameters are extracted so in order to not create two functions for those case or two instances of those parameters I decided to just pass those three you mentioned.

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
```diff
@@ -0,0 +1,253 @@
...
+    use serde_json::json;
+    use std::time::Duration;
+    use tedge_actors::test_helpers::MessageReceiverExt;
+    use tedge_actors::Sender;
+    use tedge_mqtt_ext::MqttMessage;
+    use tedge_mqtt_ext::Topic;
+
+    use tedge_mqtt_ext::test_helpers::assert_received_contains_str;
+
+    use tedge_test_utils::fs::TempTedgeDir;
```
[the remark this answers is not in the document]
Ruadhri17: fixed by 30c84b15e37d73357842338844ed8fe2fee64498

crates/common/tedge_utils/src/file.rs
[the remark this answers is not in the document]
Bravo555: > should return [EEXIST](https://www.man7.org/linux/man-pages/man2/symlink.2.html#ERRORS) if symlink already exists, and if so, we can just ignore it.

I may have mislead you a little there. After some more checking, `EEXIST` will also be returned if a symlink exists but points to another file entirely, so if we just continue, we may end up with a wrong path.
At this point we can either return an error or overwrite, depending on if we can recover ourselves and not screw up any user data, which given that only we create these symlinks, may be the case. Or you can just return the `EEXIST` error, ultimately I'm also not sure how to handle this optimally.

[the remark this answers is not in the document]
Ruadhri17: In that case we can use `read_link()` to check if those pathes are matching. If not we can return error to override symlink.

rina23q (commented):
A bit more work required

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -127,6 +140,29 @@ impl Operations {
             .collect::<HashSet<String>>()
             .try_into()?)
     }
+
+    /// Return operation and template name if `command_name` matches
+    pub fn match_template_with_operation_by_command_name(&self, command: &str) -> Option<String> {
```
albinsuresh, line 145:
```suggestion
    pub fn get_operation_template_by_command_name(&self, command: &str) -> Option<String> {
```

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -127,6 +140,29 @@ impl Operations {
...
+        self.templates
+            .iter()
+            .filter(|template| {
+                template
+                    .command_name()
+                    .is_some_and(|command_name| command_name.eq(command))
+            })
+            .find_map(|template| template.on_fragment())
```
albinsuresh, line 153: This makes me wonder if the `on_fragment` filed is redundant, when the name of the operation file should have been be the same. Are there any cases in C8y where the operation name and this fragment name could be different?

crates/core/c8y_api/src/smartrest/operations.rs
Ruadhri17: I would go with `get_operation_name_by_command_name` or `get_operation_fragment_by_command_name`

Ruadhri17: resolved by 1f913a33c1d9e6e87664d5e184192e6f512b67ce

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -382,7 +440,7 @@ pub fn get_operation(
 pub fn is_valid_operation_name(operation: &str) -> bool {
     operation
         .chars()
-        .all(|c| c.is_ascii_alphabetic() || c.is_numeric() || c.eq(&'_'))
+        .all(|c| c.is_ascii_alphanumeric() || c.eq(&'_') || c.eq(&'.'))
```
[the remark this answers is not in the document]
Ruadhri17: resolved by 1f913a33c1d9e6e87664d5e184192e6f512b67ce

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2148,6 +2166,159 @@ async fn json_custom_operation_status_update_with_custom_topic() {
...
+            r#"[exec]
+            topic = "c8y/devicecontrol/notifications"
+            on_fragment = "c8y_Command"
+            command_name = "command"
+            command_payload = "${.payload.c8y_Command}"
+            "#,
+        );
+
+    let config = C8yMapperConfig {
+        smartrest_use_operation_id: false,
```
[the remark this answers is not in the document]
Ruadhri17: resolved by 3d322e43ad1493e0f71bec602c69f787c3d1741a

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2148,6 +2166,159 @@ async fn json_custom_operation_status_update_with_custom_topic() {
...
```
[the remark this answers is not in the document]
Ruadhri17: resolved by 3d322e43ad1493e0f71bec602c69f787c3d1741a

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -101,6 +106,14 @@ impl Operations {
             .collect::<HashSet<String>>()
     }
 
+    pub fn command_names_for_templates(&self) -> HashSet<String> {
+        self.templates
+            .iter()
+            .filter_map(|template| template.command_name())
+            .map(|command_name| command_name.to_string())
+            .collect::<HashSet<String>>()
+    }
```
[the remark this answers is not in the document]
Ruadhri17: resolved by 1f913a33c1d9e6e87664d5e184192e6f512b67ce

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -242,6 +291,8 @@ impl Operation {
 #[serde(rename_all = "snake_case")]
 pub struct OnMessageExec {
     command: Option<String>,
+    command_name: Option<String>,
+    command_payload: Option<String>,
```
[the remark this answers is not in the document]
Ruadhri17: resolved

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1554,10 +1554,28 @@ async fn mapper_dynamically_updates_supported_operations_for_nested_child_device
...
+            .to_path_buf(),
+    ))
+    .await
+    .expect("Send failed");
+
+    // Assert that the creation of the operation file alone doesn't trigger the supported operations update
+    assert!(
+        mqtt.recv().await.is_none(),
+        "No messages expected on operation file creation event"
+    );
```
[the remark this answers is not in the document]
albinsuresh: As discussed offline, automatically publishing the 114 message on any updates to the operations directory was limited only to main device addressing https://github.com/thin-edge/thin-edge.io/issues/2409

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_workflow/custom_operation_workflow.robot
```diff
@@ -0,0 +1,116 @@
...
+
+Run custom operation with workflow execution on child device
+    Child Setup
+    Set Test Variable    $CHILD_XID    ${PARENT_SN}:device:${CHILD_SN}
+
+    Set Device Context    ${PARENT_SN}
+    Symlink Should Exist    /etc/tedge/operations/c8y/${CHILD_XID}/c8y_TakePicture
+
+    Cumulocity.Device Should Exist    ${CHILD_XID}
+    Cumulocity.Should Contain Supported Operations    c8y_TakePicture
```
Bravo555, line 53: question: Is my understanding correct that this line checks if `c8y_TakePicture` is supported by the parent device, not the child device, so it will always be true because we register `c8y_TakePicture` in the suite setup?

I see that we send `114,...,c8y_TakePicture` on MQTT topic `c8y/s/us/TST_augment_connected_sesame:device:TST_bark_randomized_exit`, so we should be checking that it's the child device that supports the operation, right?

Given that both devices support that operation, I don't understand how it ends up on the child device and not the main device.

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_workflow/custom_operation_workflow.robot
```diff
@@ -0,0 +1,116 @@
...
+    ...    minimum=1
+    ...    maximum=1
+    ...    message_contains="status":"executing"
+    Should Have MQTT Messages
+    ...    te/device/${device}///cmd/${cmd}/#
+    ...    minimum=1
+    ...    maximum=1
+    ...    message_contains="status":"successful"
+
+Trasfer Configuration Files
```
Bravo555, line 77: nit: typo

Ruadhri17: fixed by a531772c004475ae768f6b978b8b37a2a35d8de0

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_workflow/custom_operation_workflow.robot
Bravo555: Had a call about this with @Ruadhri17

> Given that both devices support that operation, why does incoming message on c8y/devicecontrol/notifications has externalSource of child device and it ends up being executed on the child device and not the main device?

`Cumulocity.Create Operation` seems to set `externalSource` to the most recent child device that supports the operation, but ultimately which device receives an operation can be decided by the user, using e.g. `go-c8y-cli`. So my concern was just due to my unfamiliarity with this Cumulocity feature and not anything related to thin-edge.

> I see that we send 114,...,c8y_TakePicture on MQTT topic c8y/s/us/TST_augment_connected_sesame:device:TST_bark_randomized_exit, so we should be checking that it's the child device that supports the operation, right?

At the end of the test we verify that the command is executed by the child device, so these previous checks that device exists and device supports the operation may be redundant, but most importantly they are slightly misleading because wrong device is checked.

These checks should be fixed to use a correct device, but they could also just be removed since they're redundant.

albinsuresh: > Cumulocity.Create Operation seems to set externalSource to the most recent child device that supports the operation, but ultimately which device receives an operation can be decided by the user, using e.g. go-c8y-cli.

Actually, it's the `Cumulocity.Device Should Exist    ${CHILD_XID}` call that changes the context to that device and then all subsequent C8y operations would be done on that device. Yes, implicitly changing the context like that was a bit confusing for me also, when I initially started using this library and have gotten used to it over time. I agree that it would have been better if it was clearer from the API name itself.

albinsuresh (commented):
Happy to approve once https://github.com/thin-edge/thin-edge.io/pull/3225#discussion_r1848879538 is also resolved. If that `on_fragment` key is redundant (always same as the operation name itself), we can get rid of it.

crates/core/c8y_api/src/smartrest/operations.rs
reubenmiller: I wouldn't that the fragment and the name of the operation are the same. These are custom operations which means that the customer is free to configure them however they want, so being more explicit would be safer here with the downside of a bit of boilerplate.

albinsuresh: @Ruadhri17 if you're also wondering why any user would want to declare one name in the supported operations list (the operation file name) and use an entirely different name for his operation fragment, here is the use-case: The declared supported operation name is typically used to enable/disable UI components in C8y. The developer of that UI component dictates the expected operation name. And it is a possibility that the custom operation developer may not be able to (or don't want to) reuse the same name for his operation as well. In such cases, this design would help.

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_workflow/custom_operation_workflow.robot
Bravo555: > Actually, it's the Cumulocity.Device Should Exist ${CHILD_XID} call that changes the context to that device and then all subsequent C8y operations would be done on that device

Ah, that explains it, and as there's no mention of it in the documentation of the keyword, the implicit change of the context is quite unexpected indeed.

Bravo555 (approved):
https://github.com/thin-edge/thin-edge.io/pull/3225/files#r1851457754

Thanks to Albin's comment I now know that the checks are indeed correct, so I don't have any other blockers. 

The code and the test suite could use some more documentation comments, to make it easier why we do the things we do without retracing the original tickets, but I don't want to block this PR anymore.

crates/extensions/c8y_mapper_ext/src/operations/convert.rs
```diff
@@ -387,4 +390,42 @@ impl CumulocityConverter {
...
+                err_msg: format!("Invalid JSON message, {e}. Message: {message:?}"),
+            }
+        })?;
+
+        let payload: Value = if let Some(tedge_cmd_value) = custom_handler.tedge_cmd_payload() {
+            let payload = state.inject_values_into_template(tedge_cmd_value);
+            serde_json::from_str(&payload)?
+        } else {
+            serde_json::Value::Null
+        };
```
didier-wenzek, line 425: This can be improved using the same logic as for spawning a sub-operation using [`StateExcerpt`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/workflow/state.rs#L540)

```
let excerpt = StateExcerpt::PathExpr(custom_handler.tedge_cmd_payload() );
let payload = excerpt.extract_value_from(state);
```

The main benefit is that this will help to compatible with workflow definition and enable feature like building the tedge operation payload as a *combination* of several fragments of the c8y operation payload (as done [here](https://thin-edge.github.io/thin-edge.io/references/agent/operation-workflow/#sub-operation-execution):

```
command_payload.x = "some static value"
command_payload.y = "${.payload.y}"
command_payload.z = "${.payload.z}"
```

didier-wenzek: I'm okay to merge this PR without this refinement.

However, we need to agree on the property names as discussed [here](https://github.com/thin-edge/thin-edge.io/issues/3095#issuecomment-2473876874) and [here](https://github.com/thin-edge/thin-edge.io/issues/3095#issuecomment-2490426004). @reubenmiller what's your point of view of these property names?

reubenmiller: Ok, I've added my suggestion which is merging the two concepts. https://github.com/thin-edge/thin-edge.io/issues/3095#issuecomment-2493486901

didier-wenzek (approved):
Approved. Thank you for taking the time to address all the comments.

albinsuresh (approved):
LGTM

Issue #3212, opened 2024-10-29
sm-plugin fails to download c8y resource when local proxy is using cert based authentication


reubenmiller:
**Describe the bug**

The thin-edge.io software management plugin interface fails to download files from the Cumulocity Local Proxy service if local certificate-based authentication is being used.

The error appears when thin-edge.io tries to download the artifact from the Cumulocity Local Proxy (if the artifact's url is hosted in Cumulocity), before calling the sm-plugin `install` subcommand. Below shows the error in the tedge-agent logs:

```log
Oct 29 10:16:24 rpi5-d83add9f145a tedge-agent[871]: 2024-10-29T10:16:24.153856163Z ERROR plugin_sm::plugin: Download error: DownloadError {
Oct 29 10:16:24 rpi5-d83add9f145a tedge-agent[871]:     reason: "error sending request for url (https://rpi5-d83add9f145a:8001/c8y/inventory/binaries/40663602): error trying to connect: received corrupt message of type InvalidContentType",
Oct 29 10:16:24 rpi5-d83add9f145a tedge-agent[871]:     url: "https://rpi5-d83add9f145a:8001/c8y/inventory/binaries/40663602",
Oct 29 10:16:24 rpi5-d83add9f145a tedge-agent[871]:     source_err: "error trying to connect: received corrupt message of type InvalidContentType",
Oct 29 10:16:24 rpi5-d83add9f145a tedge-agent[871]: }
```

Looking at the code, it seems that the tedge.toml setting; `c8y.proxy.key_path`, `c8y.proxy.cert_path` and `c8y.proxy.ca_path` are ignored, as shown in the following code (though please verify if this link is correct).

https://github.com/thin-edge/thin-edge.io/blob/94b992d7f6d92bc4d167b2008bddac8eba984c2a/crates/common/download/src/download.rs#L105-L106

**To Reproduce**

1. Configure thin-edge.io to cert-based authentication for local services such as
2. Create a software and version in the Cumulocity Software repository
3. Install software with uses the `url` field (to download the artifact from the platform)

**Expected behavior**

* If the Cumulocity Local Proxy is configured to use cert-based authentication (for local connections), then any thin-edge.io component which uses it should use the configured client certificates

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `any`
 - Hardware [incl. revision]: `any`
 - System-Architecture [e.g. result of "uname -a"]: `any`
[...]

Issue #2321, opened 2023-10-06
c8y proxy fails with sporadic dns lookup address errors


reubenmiller:
**Describe the bug**

Requests to thin-edge.io Cumulocity IoT proxy (localhost:8001) returns sporadically with a 502 Bad Gateway status code.

The `tedge-mapper-c8y` which is providing the local Cumulocity IoT proxy logs the following error on a failed proxy attempt:

```log
Oct 06 11:32:26 pippin tedge-mapper[459]: 2023-10-06T01:32:26.117459217Z ERROR c8y_auth_proxy::server:   × making proxied request to https://test-ci-runner01.latest.stage.c8y.io/
Oct 06 11:32:26 pippin tedge-mapper[459]:   │ event/events/3767200/binaries
Oct 06 11:32:26 pippin tedge-mapper[459]:   ├─▶ error sending request for url (https://test-ci-
Oct 06 11:32:26 pippin tedge-mapper[459]:   │   runner01.latest.stage.c8y.io/event/events/3767200/binaries): error
Oct 06 11:32:26 pippin tedge-mapper[459]:   │   trying to connect: dns error: failed to lookup address information:
Oct 06 11:32:26 pippin tedge-mapper[459]:   │   Try again
Oct 06 11:32:26 pippin tedge-mapper[459]:   ├─▶ error trying to connect: dns error: failed to lookup address
Oct 06 11:32:26 pippin tedge-mapper[459]:   │   information: Try again
Oct 06 11:32:26 pippin tedge-mapper[459]:   ├─▶ dns error: failed to lookup address information: Try again
Oct 06 11:32:26 pippin tedge-mapper[459]:   ╰─▶ failed to lookup address information: Try again
```

However using curl to communicate directly with the Cumulocity IoT tenant is able to successfully communicate with the tenant. A script was written to compare requests against the local proxy and direct communication with the Cumulocity IoT Tenant:

In combination with the duration of the failed requests and the error message printed in the log, there is a strong correlation with the duration of the request and the failure rate. It seems that if the DNS resolver takes longer than 5 seconds, then proxied request fails.

```
$ ./check_proxy.sh
------- Attempt 1 (12:09:47) -------
proxy	PASSED	2s
curl	PASSED	1s
------- Attempt 2 (12:09:55) -------
proxy	PASSED	2s
curl	PASSED	6s
------- Attempt 3 (12:10:08) -------
proxy	PASSED	2s
curl	PASSED	1s
[...]

PR #3229, opened 2024-11-06
Change which download errors are retryable

Bravo555:
## Proposed changes

Whereas previously we considered only some errors to be permanent and the rest to be retryable, that logic was inverted so we check if an error is one of a few that is transient, and the rest are considered permanent.

Biggest changes:
- connect errors are considered transient
- timeouts are considered transient

## Types of changes

## Paste Link to the issue

- #3212

## Checklist

### DIFF
--- a/crates/common/download/src/download.rs
+++ b/crates/common/download/src/download.rs
@@ -927,7 +927,7 @@ mod tests {
         let target_path = target_dir_path.path().join("test_download");
         let mut downloader = Downloader::new(target_path, None, CloudRootCerts::from([]));
         downloader.set_backoff(ExponentialBackoff {
-            current_interval: Duration::ZERO,
+            max_elapsed_time: Some(Duration::ZERO),
             ..Default::default()
         });
         match downloader.download(&url).await {
--- a/crates/common/download/src/download.rs
+++ b/crates/common/download/src/download.rs
@@ -379,25 +379,8 @@ impl Downloader {
             request
                 .send()
                 .await
-                .map_err(|err| {
-                    // rustls errors are caused by e.g. CertificateRequired
-                    // If this is the case, retrying won't help us
-                    if err.is_builder()
-                        || err.is_connect()
-                        || axum_tls::rustls_error_from_reqwest(&err).is_some()
-                    {
-                        backoff::Error::Permanent(err)
-                    } else {
-                        backoff::Error::transient(err)
-                    }
-                })?
-                .error_for_status()
-                .map_err(|err| match err.status() {
-                    Some(status_error) if status_error.is_client_error() => {
-                        backoff::Error::Permanent(err)
-                    }
-                    _ => backoff::Error::transient(err),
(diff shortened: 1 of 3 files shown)


crates/common/download/src/download.rs
```diff
@@ -408,6 +391,19 @@ impl Downloader {
     }
 }
 
+/// Decides whether HTTP request error is retryable.
+fn reqwest_err_to_backoff(err: reqwest::Error) -> backoff::Error<reqwest::Error> {
+    if err.is_timeout() || err.is_connect() {
+        return backoff::Error::transient(err);
+    }
+    if let Some(status) = err.status() {
+        if status.is_server_error() {
```
albinsuresh, line 400: What about 404? In most cases it might be non-retryable, but aren't there cases where a resource not existing also could be transient (delays in getting a newly uploaded resource synced across all regions or something like that)?

Bravo555: Yeah, currently 404 is considered permanent. And perhaps indeed we could retry on 404 for such scenarios, but then we add delay in cases where the URL is simply wrong.
I have no preference either way. @reubenmiller what do you think?

didier-wenzek: I would consider 404 as a permanent error.

To handle the case where a resource is not existing *yet* but which creation is pending,  the solution will depends on the server but should not rely on 404, but using a combination of 202 Accepted (the creation of the resource has been accepted) and 400 Bad Request (possibly with a hint to retry in a sec).

crates/common/download/src/download.rs
```diff
@@ -931,7 +927,7 @@ mod tests {
         let target_path = target_dir_path.path().join("test_download");
         let mut downloader = Downloader::new(target_path, None, CloudRootCerts::from([]));
         downloader.set_backoff(ExponentialBackoff {
-            current_interval: Duration::ZERO,
+            max_elapsed_time: Some(Duration::ZERO),
```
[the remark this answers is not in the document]
Bravo555: Yes. As now the connect errors are considered transient, it causes DNS resolution failure to be considered transient, which caused the test to hang instead of immediately returning.
I think the intention behind setting `current_interval` to zero would be to disable retries in the test, but it wasn't actually doing anything, and only `max_elapsed_time` set to zero actually disables the retries.

But another question is should DNS resolution failure be considered transient. On one hand we can timeout during the query because the network is down, then we should retry, but if the network is up and the domain does not exist, then we should return ASAP.

I'll check if we can extract information from `reqwest::Error` so that we can distinguish between these 2 cases.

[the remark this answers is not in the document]
reubenmiller: > But another question is should DNS resolution failure be considered transient.

Yes DNS resolution errors are transient and should be retried, see https://github.com/thin-edge/thin-edge.io/issues/2321

albinsuresh (approved):

PR #3230, opened 2024-11-06
feat: Entity registration HTTP API

albinsuresh:
## Proposed changes

- [x] Refactor file-transfer logic into its own module 
- [x] Add entity store endpoints to existing tedge HTTP server
- [x] Integrate the HTTP endpoints with the entity store
- [x] Refactor file transfer API logic to its own module
- [x] Publish entities registered via HTTP APIs to the MQTT broker
- [x] Accept registration messages via MQTT as well
- [x] Remove `entity_store` usage from the mappers, extracting only the external ID aspects from it
- [x] Handle auto-registration
- [x] Rename existing `file_transfer_server` module into `tedge_server`

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/core/tedge_agent/src/file_transfer_server/mod.rs
```diff
@@ -1,4 +1,6 @@
+//TODO Rename this module to tedge_server
```
didier-wenzek, line 1: I would prefer `http_server`, stressing this is about HTTP while tedge is obvious in the context of `tedge-agent`.

crates/core/tedge_agent/src/file_transfer_server/mod.rs
```diff
@@ -1,4 +1,6 @@
...
 pub mod actor;
 pub mod error;
+mod file_transfer_controller;
```
didier-wenzek, line 5: Why not simply `file_transfer`?

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -26,7 +26,7 @@ impl AgentState {
     }
 }
 
-pub(crate) fn http_file_transfer_server(
+pub(crate) fn tedge_server(
```
didier-wenzek, line 42: `http_server` would be more specific

crates/core/tedge_agent/src/file_transfer_server/mod.rs
albinsuresh: Will be renamed as suggested, right before merging this PR.

crates/core/tedge_agent/src/file_transfer_server/mod.rs
albinsuresh: It actually started with `entity_store_controller`, as I didn't want to name it `entity_store` as the actual entity store module is also named that way. And I wanted to differentiate this from that and hence ended up adding the `controller` suffix. I've renamed it as you suggested.

But, now I'm wondering if naming them `file_transfer_service` and `entity_store_service` would have been clearer/better.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -352,8 +370,34 @@ impl Agent {
...
+                main_device,
+                self.config.service.ty.clone(),
+                Self::dummy_external_id_mapper,
+                Self::dummy_external_id_validator,
+                EARLY_MESSAGE_BUFFER_SIZE,
+                state_dir,
+                clean_start,
+            )
+            .unwrap();
+            let entity_store = Arc::new(Mutex::new(entity_store));
```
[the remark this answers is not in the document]
albinsuresh: Addressed in 20f71f4

crates/core/tedge_agent/src/entity_manager/actor.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn name(&self) -> &str {
+        "EntityManagerActor"
+    }
+
+    async fn run(mut self) -> Result<(), RuntimeError> {
+        while let Some(input) = self.input_receiver.recv().await {
+            match EntityRegistrationMessage::try_from(&input) {
+                Ok(register_message) => {
+                    info!("Hippo: Received reg message: {:?}", register_message);
+                    if let Err(e) = self.entity_store.lock().unwrap().update(register_message) {
```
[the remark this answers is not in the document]
albinsuresh: Addressed in 20f71f464d2546

crates/core/tedge_agent/src/agent.rs
```diff
@@ -390,6 +434,31 @@ impl Agent {
 
         Ok(())
     }
+
+    // TODO: Remove these dummy impls once external ID aspects are removed from entity store
```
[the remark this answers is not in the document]
albinsuresh: To be done in the next step.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -73,7 +82,9 @@ pub(crate) struct AgentConfig {
     pub log_dir: Utf8PathBuf,
     pub agent_log_dir: Utf8PathBuf,
     pub data_dir: DataDir,
+    pub state_dir: Utf8PathBuf,
     pub operations_dir: Utf8PathBuf,
+    pub device_id: Option<String>, //Some for main device, None for child devices
```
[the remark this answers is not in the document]
albinsuresh: This is just an intermediary measure. This dependency on `device.id` would be gone once the external id aspects are moved from the entity store to the mapper.

crates/extensions/c8y_mapper_ext/src/alarm_converter.rs
```diff
@@ -36,9 +37,10 @@ impl AlarmConverter {
     pub(crate) fn try_convert_alarm(
         &mut self,
         source: &EntityTopicId,
+        external_id: &EntityExternalId,
+        entity_type: &EntityType,
         input_message: &MqttMessage,
         alarm_type: &str,
-        entity_store: &EntityStore,
```
didier-wenzek, line 41: Oh, the *whole* entity_store was passed to convert a specific alarm !? Good to have that fixed.

crates/extensions/c8y_mapper_ext/src/error.rs
```diff
@@ -149,6 +152,9 @@ pub enum CumulocityMapperError {
     #[error(transparent)]
     FromEntityStore(#[from] tedge_api::entity_store::Error),
 
+    #[error(transparent)]
+    FromEntityCacheError(#[from] crate::entity_cache::Error),
+
```
didier-wenzek, line 157: Why do we have to maintain two error types with the same error cases?

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -0,0 +1,229 @@
...
+    UnknownEntity(String),
+
+    #[error(transparent)]
+    InvalidExternalId(#[from] InvalidExternalIdError),
+
+    #[error("Auto registration of the entity with topic id {0} failed as it does not match the default topic scheme: 'device/<device-id>/service/<service-id>'. Try explicit registration instead.")]
+    NonDefaultTopicScheme(EntityTopicId),
+}
+
+pub(crate) struct EntityCache {
```
didier-wenzek, line 39: A comment telling the purpose of this cache would be really helpful.

- What the difference of this cache compared to the entity store?
- How both cooperate?
- What's the content of the pending entities?

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -23,7 +24,7 @@ pub fn is_c8y_bridge_established(
 pub fn convert_health_status_message(
     mqtt_schema: &MqttSchema,
     entity: &EntityMetadata,
-    ancestors_external_ids: &[String],
+    parent_external_id: Option<&EntityExternalId>,
```
didier-wenzek, line 27: Why is the id of the parent now enough while the whole ancestor list was previously required?

crates/core/tedge_agent/src/agent.rs
```diff
@@ -352,9 +371,39 @@ impl Agent {
...
+                        reply_to: Box::new(NullSender),
+                    })
+                },
+            );
+
+            let file_transfer_server_builder = FileTransferServerBuilder::try_bind(
+                self.config.http_config,
+                &mut entity_store_actor_builder,
+                &mut mqtt_actor_builder,
+            )
```
didier-wenzek, line 405: I'm surprised to see the `file_transfer_server` listening on both HTTP and MQTT.

I would expect the `entity_store_actor` to listen to MQTT plus a channel of requests forwarded by the `file_transfer_server`. 

- What has been your drivers for this design?
- Is this because on the HTTP side a response is expected?

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
albinsuresh: Maintaining the entire ancestors list was not really required to create or publish the data to a nested entity. We didn't know that earlier (when child device support was added). We later realised that, once a child device is created at any level with an external id (XID), its nested child device can be created by simply publishing to `s/us/<xid>` directly without the entire ancestry in the topic. This is possible because XIDs are unique for a tenant and hence there can't be 2 different entities with the same XID at different levels in the entity hierarchy.

With the introduction of the entity cache, I just used this opportunity to clean that up, as that helped me simplify and minimize the metadata maintained by the entity cache.

crates/core/tedge_agent/src/agent.rs
albinsuresh: > What has been your drivers for this design?

It might be just a naming confusing during this transition period. Even though the entire HTTP code is currently kept under the module named `file_transfer_server`, the plan is to rename this module into `http_server` and associated structs into `HttpServerBuilder`,`HttpServerActor`, etc. Then this `HttpActor` would delegate HTTP requests to `file_transfer_service` and `entity_store_service` respectively. When that happens, the `file_transfer_service` will no have any MQTT aspects in it.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -378,26 +292,26 @@ impl EntityStore {
...
-                    .unwrap()
-            })
-            .collect();
-
-        Ok(mapped_ancestors)
-    }
+    // /// Returns an ordered list of ancestors' external ids of the given entity
+    // /// starting from the immediate parent all the way till the root main device.
+    // /// The last parent in the list for any entity would always be the main device id.
+    // pub fn ancestors_external_ids(
```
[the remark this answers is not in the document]
albinsuresh: This will be removed for the reason mentioned in https://github.com/thin-edge/thin-edge.io/pull/3230#discussion_r1877749556

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
albinsuresh: Yes, I'll be adding a detailed comment for the `EntityCache` and the newly introduced `EntityMetadata` structs. Here is the quick summary though:

> What the difference of this cache compared to the entity store?

While `EntityStore`, owned by the agent, maintains **all** the metadata associated with an entity like its registration message, device twin data, supported operations etc, the `EntityCache` was envisaged as a lightweight map maintained by the mappers to maintain the external id mapping and a minimal set of additional metadata required to do the appropriate mapping to the cloud twins. 

> How both cooperate?

The entity cache is updated every time a new entity is successfully registered with the entity store(when an entity registration message is emitted either directly by the client or by the agent in response to a registrartion attempt over HTTP). When the mapper receives an entity registration message either way, it assumes that it was successfully registered and adds it to its entity cache.

> What's the content of the pending entities?

When the mapper receives an entity registration message with a parent, and if that parent is not registered yet at that point, it is added to the pending entity store and such pending entities are added to the entity cache only when their parent is registered. So, whenever a new entity is registered, the mapper checks its entity cache to find any of its children that arrived earlier, waiting to be registered and registers them along with the parent.

crates/extensions/c8y_mapper_ext/src/error.rs
albinsuresh: The errors emitted by the entity cache would be similar, but a much smaller set compared to the entity store, as it maintains far less metadata. That's why I didn't want them to share the same error type. Eventually I see the entity store structure moving from `tedge_api` to `tedge_agent` and the entity cache being mainained by the `tedge_mapper`. So, avoiding any direct dependency between them would be better for that move.

crates/core/tedge_api/src/entity.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl From<&EntityExternalId> for String {
+    fn from(value: &EntityExternalId) -> Self {
+        value.0.clone()
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct EntityMetadata {
```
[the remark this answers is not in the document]
albinsuresh: As mentioned in https://github.com/thin-edge/thin-edge.io/pull/3230#discussion_r1878052707, the entity store and entity cache needs to maintain their own versions of entity metadata as the latter would be much lighter with minimal metadata maintained at the mapper level. Ideally the `EntityMetadata` maintained by the mapper should have been as minimal as the XID. But, I was forced to store a bit more info like the entity type, display name and display type just because of the way we map service health status messages by publishing the service creation message itself(which requires all that metadata) for status updates as well. One way to eliminate caching this additional metadata at the mapper level is by making it query the entity store contents from the agent, whenever it requires all that metadata. But that extra HTTP call during the mapping complicates the conversion logic and I just chose caching extra metadata instead of that added complexity. But, open to revisit that decision in a follow-up PR dedicated for that change as the unit test impact would be quite big.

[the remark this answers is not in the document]
didier-wenzek: > But that extra HTTP call during the mapping complicates the conversion logic and I just chose caching extra metadata instead of that added complexity. 

Agree, caching registration messages is simpler than late-requests to the entity store.

But could we share the type of cached data? Say raw registration messages? Even if some fields are not currently used in the mapper.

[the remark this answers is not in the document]
albinsuresh: > But could we share the type of cached data? Say raw registration messages? Even if some fields are not currently used in the mapper.

Yeah, that is an option. I introduced a different struct to clearly represent the minimal set of things that a mapper would need to do the mapping. And also, there were minor things like the external id (`@id`) being an `Option` in the actual `EntityRegistrationMessage`, but the same being a mandatory field in the entity cache. But yeah, if we want to avoid having too many "metadata representations" across the crates, I can reuse some of the existing ones with implicit assumptions about fields like `@id`.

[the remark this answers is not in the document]
didier-wenzek: We really have to reduce the number of similar but different structs.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -0,0 +1,157 @@
...
+            EntityStoreRequest::Create(entity) => {
+                let res = self.entity_store.update(entity);
+                EntityStoreResponse::Create(res)
+            }
+            EntityStoreRequest::Delete(topic_id) => {
+                let deleted_entities = self.entity_store.deregister_entity(&topic_id);
+                EntityStoreResponse::Delete(deleted_entities)
+            }
+            EntityStoreRequest::MqttMessage(mqtt_message) => {
+                self.process_mqtt_message(mqtt_message).await;
```
didier-wenzek, line 83: I'm surprised that there is a lot more to do when a message is received from MQTT vs HTTP.

Why that?

albinsuresh:
> * actor naming and branching in the agent

Yes, we can discuss the details. Renaming some of the modules are already planned as described [here](https://github.com/thin-edge/thin-edge.io/pull/3230#discussion_r1878030806) 

> * the two definitions of `struct EntityMetadata`

The `EntityMetadata` struct used by the `entity_cache` was envisaged to be a lot simpler than what is it today. Now that both the structs are awfully similar barring some minor differences, I'll just re-use the existing one itself.

> * the two similar error types `c8y_mapper_ext::CumulocityMapperError` and `c8y_mapper_ext::ConversionError`

Yeah, this has been a legacy issue that can be fixed. But, I'd prefer to do that in a separate PR as this PR is already huge and resolving merge conflicts is a pain every time someone else touches the mapper.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: The difference is in the "pending entities" caching aspect. When the registration is done over MQTT, due to the lack of feedback to the client, we are forced to cache any early child reg messages, until the parents are also registered. But, with the HTTP, when a client tries to register before its parent, we can always immediately send the response back that the parent is not available so that the client can retry later. We don't **need to** do caching in this case. But, we **can** if we want to.

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
```diff
@@ -0,0 +1,409 @@
...
+        .await_response(EntityStoreRequest::Create(entity.clone()))
+        .await?;
+    let EntityStoreResponse::Create(result) = response else {
+        return Err(Error::InvalidEntityStoreResponse);
+    };
+
+    if !result?.0.is_empty() {
+        let message = entity.to_mqtt_message(&state.mqtt_schema);
+        state.mqtt_publisher.clone().send(message).await?;
+    }
```
didier-wenzek, line 92: I would delegate to `EntityStoreServer` the responsibility to publish over MQTT the updated entity metadata.

-  `EntityStoreServer`  is already MQTT dependent and publishing entity metadata.
- This HTTP handler would be free of any MQTT dependencies.

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
```diff
@@ -0,0 +1,409 @@
...
+    for topic_id in deleted {
+        let topic = state
+            .mqtt_schema
+            .topic_for(&topic_id, &Channel::EntityMetadata);
+        let clear_entity_msg = MqttMessage::new(&topic, "")
+            .with_retain()
+            .with_qos(QoS::AtLeastOnce);
+
+        state.mqtt_publisher.clone().send(clear_entity_msg).await?;
+    }
```
didier-wenzek, line 144: I would delegate to `EntityStoreServer` the responsibility to publish over MQTT the updated entity metadata.

-  `EntityStoreServer`  is already MQTT dependent and publishing entity metadata.
- This HTTP handler would be free of any MQTT dependencies.

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
```diff
@@ -1,144 +1,51 @@
...
-    error
-        .downcast_ref()
-        .map(|e| err_is_is_a_directory(e, path))
-        .unwrap_or(false)
+#[derive(Clone)]
+pub(crate) struct AgentState {
+    pub(crate) file_transfer_dir: Utf8PathBuf,
+    pub(crate) entity_store_handle: ClientMessageBox<EntityStoreRequest, EntityStoreResponse>,
+    pub(crate) mqtt_schema: MqttSchema,
+    pub(crate) mqtt_publisher: LoggingSender<MqttMessage>,
```
didier-wenzek, line 23: I would delegate to `EntityStoreServer` the responsibility to publish over MQTT the updated entity metadata.

-  `EntityStoreServer`  is already MQTT dependent and publishing entity metadata.
- This HTTP handler would be free of any MQTT dependencies.

didier-wenzek (commented):
This PR is really close to be merged. I only have a change request:
to move the responsibility to publish over MQTT the updated entity metadata from the `entity_store_router` HTTP handler to the `EntityStoreServer` actor.

crates/core/tedge_agent/src/agent.rs
didier-wenzek: I understand now where we are not aligned:

- The responsibility to publish over MQTT entity state should be moved  from the `entity_store_router` HTTP handler to the `EntityStoreServer` actor.
- See https://github.com/thin-edge/thin-edge.io/pull/3230#discussion_r1900976673

```suggestion
            let file_transfer_server_builder = FileTransferServerBuilder::try_bind(
                self.config.http_config,
                &mut entity_store_actor_builder,
            )
```

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    fn process_entity_registration(&mut self, message: MqttMessage) {
+        match EntityRegistrationMessage::try_from(&message) {
+            Ok(entity) => match self.entity_store.update(entity.clone()) {
+                Ok((_, pending_entities)) => {
+                    for pending_entity in pending_entities {
+                        if let Err(err) =
+                            self.entity_store.update(pending_entity.reg_message.clone())
```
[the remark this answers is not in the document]
albinsuresh: > Why do we have to register the pending entities returned by self.entity_store.update?
> Why is this not directly done by the first call to self.entity_store.update?

Yeah, this was unnecessary. Any pending entities that can be registered are already registered as part of `EntityStore:update`.

> What is pending entity if not an entity that cannot be registered yet because of missing ancestors?

When an attempt is made to register an entity without all of its ancestors registered upfront, it is just cached in the pending entity store. Once that missing ancestor that connects the child entity tree is registered, all of its pending children are also registered along with it. The returned `Vec<PendingEntityData>` indicates all the entities that got registered as part of that `update` call. This `Vec` would be empty on an attempt to register an "early entity" that is only cached.

tests/RobotFramework/tests/cumulocity/mapper_recovery/recover_and_publish_software_update_message.robot
```diff
@@ -35,6 +35,7 @@ Mapper recovers and processes output of ongoing software update request
     Device Should Have Installed Software    rolldice
 
 Recovery from corrupt entity store file
+    Skip    Persistence is not a mapper responsibility anymore
```
rina23q, line 38: Don't we need to re-write this test case for `tedge-agent`?

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -306,6 +306,7 @@ Early data messages cached and processed
     END
 
 Entities persisted and restored
+    Skip    Persistence is not a mapper responsibility anymore
     Execute Command    sudo tedge config set c8y.entity_store.clean_start false
     Restart Service    tedge-mapper-c8y
     Service Health Status Should Be Up    tedge-mapper-c8y
```
rina23q, line 312: Since it's registration_lifecycle test, I think the purpose of this test is still valid. Instead of skipping it, can't we replace tedge-mapper-c8y with tedge-agent?

```suggestion
    Execute Command    sudo tedge config set agent.entity_store.clean_start false
    Restart Service    tedge-agent
    Service Health Status Should Be Up    tedge-agent
```

crates/core/tedge_agent/src/entity_manager/server.rs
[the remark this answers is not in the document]
Bravo555: So the plan is to change that code to have only one `update()` call?

Bravo555 (commented):
Most pressing questions I had were also raised already by Didier and Rina, so my feedback is mostly a few questions and a suggestion for some documentation comments in some places.
With the PR being so large, it is tricky to know and remember the reason behind so many changes, so I think it's crucial to have a high level summary of the changes in the PR description before it's merged, probably also referring to [0005-entity-registration-api.md](https://github.com/thin-edge/thin-edge.io/blob/07bf931fa7e13352e21acf174bebb149bac4c7e0/design/decisions/0005-entity-registration-api.md) design document, and making clear if this PR addresses it in its entirety, or if there's additional work to be done.

crates/extensions/c8y_mapper_ext/src/operations/handlers/mod.rs
```diff
@@ -37,7 +37,7 @@ use std::sync::Arc;
 use tedge_actors::ClientMessageBox;
 use tedge_actors::LoggingSender;
 use tedge_actors::Sender;
-use tedge_api::entity_store::EntityExternalId;
+use tedge_api::entity::EntityExternalId;
```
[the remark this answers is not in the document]
albinsuresh: This change is part of a specific commit that moved the external id aspects from the entity store to the entity cache. It wouldn't have made much sense as an independent commit without that context. But, to reduce the size of the PR, I'll consider doing that before the merge.

tests/RobotFramework/tests/cumulocity/mapper_recovery/recover_and_publish_software_update_message.robot
albinsuresh: Yes. Skipping was just a temporary measure which got forgotten after a while.

crates/core/tedge_api/Cargo.toml
```diff
@@ -20,6 +20,7 @@ regex = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 shell-words = { workspace = true }
+strum = { workspace = true }
```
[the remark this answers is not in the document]
albinsuresh: Leftover from an earlier experiement. Will remove.

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
```diff
@@ -0,0 +1,409 @@
...
+use tedge_api::mqtt_topics::Channel;
+use tedge_api::mqtt_topics::EntityTopicId;
+use tedge_api::mqtt_topics::TopicIdError;
+use tedge_mqtt_ext::MqttMessage;
+use tedge_mqtt_ext::QoS;
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) struct EntityDef {
+    topic_id: String,
+}
```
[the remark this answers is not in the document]
albinsuresh: Yeah, Definition. But, I'll be removing this struct as it wasn't fully implemented as the agreement following [this](https://github.com/thin-edge/thin-edge.io/pull/3230#discussion_r1872949767) conversation was to reduce the number of entity metadata structs in the code and reuse the existing ones as much as possible.

crates/core/tedge_agent/src/file_transfer_server/actor.rs
```diff
@@ -37,6 +49,7 @@ pub(crate) struct FileTransferServerConfig<CertKeyPath = Utf8PathBuf, CaPath = U
     pub key_path: OptionalConfig<CertKeyPath>,
     pub ca_path: OptionalConfig<CaPath>,
     pub bind_addr: SocketAddr,
+    pub mqtt_schema: MqttSchema,
```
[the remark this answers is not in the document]
albinsuresh: Yeah, passing this `MqttSchema` around has been a pain not just here but many other parts of the code where MQTT messages are generated. But, at least in this case, these usages would be removed from here and moved to the entity store actor as suggested [here](https://github.com/thin-edge/thin-edge.io/pull/3230#discussion_r1900976673)

crates/core/tedge_agent/src/entity_manager/server.rs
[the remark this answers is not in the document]
albinsuresh: > So the plan is to change that code to have only one update() call

Yes, the first `update` call is already doing what is done in the subsequent lines registering all the pending entities that are eligible to be registered at that point.

crates/core/tedge_api/src/entity.rs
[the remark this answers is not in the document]
albinsuresh: Resolved by 8751a769 . Even though a single struct is easier to maintain, I'm still tempted to introduce a dedicated `EntityMetadataDto` struct to capture the message payloads, without overloading this internal `EntityMetadata` representation with serde aspects. By using the same struct for everything, I'm a bit worried that it'll make future changes to this internal metadata representation more difficult as it'd affect the external representation as well.

crates/core/tedge_api/src/entity.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    pub fn display_name(&self) -> Option<&str> {
+        self.other.get("name").and_then(|v| v.as_str())
+    }
+
+    pub fn display_type(&self) -> Option<&str> {
+        self.other.get("type").and_then(|v| v.as_str())
+    }
+
+    pub fn external_id_unchecked(&self) -> &EntityExternalId {
```
albinsuresh, line 112: This function was one side effect of reusing the same `EntityMetadata` struct in both entity store and entity cache. When used in the entity store, this value may or may not be present (based on the presence of `@id` in the registration message). But, when used in the entity cache, the external id is guaranteed to be there (auto-generated when not provided via `@id` during registration). So, the mapper code always uses this function with that implicit guarantee, to directly retrieve the external id.

crates/core/tedge_agent/src/agent.rs
albinsuresh: Resolved by 1a85933bd5b546dc611ac9c93230493ad2b6a284

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
albinsuresh: Resolved by a85933b

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
albinsuresh: Resolved by 1a85933b

crates/core/tedge_agent/src/file_transfer_server/http_rest.rs
albinsuresh: Resolved by 1a85933b

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+#[async_trait]
+impl Server for EntityStoreServer {
+    type Request = EntityStoreRequest;
+    type Response = EntityStoreResponse;
+
+    fn name(&self) -> &str {
+        "ConcurrentWorker"
```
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe37

crates/core/tedge_agent/src/file_transfer_server/actor.rs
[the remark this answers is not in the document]
albinsuresh: Resolved by d750f

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe3

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe

crates/core/tedge_api/Cargo.toml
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
albinsuresh: Resolved by d750fe

tests/RobotFramework/tests/cumulocity/mapper_recovery/recover_and_publish_software_update_message.robot
albinsuresh: Resolved by d750fe

crates/extensions/c8y_mapper_ext/src/operations/handlers/firmware_update.rs
```diff
@@ -294,7 +296,7 @@ mod tests {
         .await
         .expect("Send failed");
 
-        mqtt.skip(3).await; // Skip entity registration, mapping and installed firmware messages
+        mqtt.skip(1).await; // Skip entity registration, mapping and installed firmware messages
```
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
albinsuresh: Resolved by d750fe

crates/core/tedge_api/src/store/ring_buffer.rs
```diff
@@ -3,7 +3,7 @@ use std::collections::VecDeque;
 
 /// A bounded buffer that replaces older values with newer ones when full
 #[derive(Debug)]
-pub(crate) struct RingBuffer<T> {
+pub struct RingBuffer<T> {
```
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe

crates/extensions/c8y_mapper_ext/src/operations/handlers/config_snapshot.rs
```diff
@@ -313,7 +313,9 @@ mod tests {
         .await
         .expect("fail to register the child-device");
 
-        mqtt.skip(1).await; // Skip child device registration messages
+        let msg = mqtt.recv().await.unwrap();
+        dbg!(&msg);
+        // mqtt.skip(1).await; // Skip child device registration messages
```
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe3

crates/core/tedge_agent/src/entity_manager/server.rs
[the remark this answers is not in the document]
albinsuresh: Resolved by d750fe3

crates/core/tedge_api/src/entity.rs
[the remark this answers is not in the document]
didier-wenzek: I don't fully get your concerns.

- The extra lines for serde are minimal and only related to naming.
- Is this really using the same struct for everything? These metadata are exchanged over MQTT and stored in-memory. What's wrong?

crates/core/tedge_agent/src/agent.rs
didier-wenzek: Thanks. This simplifies a lot the file transfer server code.

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -0,0 +1,353 @@
...
+/// until those parents are registered as well.
+/// Once the parent is registered, the pending child entities are also registered along with it.
+pub(crate) struct EntityCache {
+    main_device_tid: EntityTopicId,
+    main_device_xid: EntityExternalId,
+    external_id_mapper_fn: ExternalIdMapperFn,
+    external_id_validator_fn: ExternalIdValidatorFn,
+
+    entities: HashMap<EntityTopicId, EntityMetadata>,
+    external_id_map: HashMap<EntityExternalId, EntityTopicId>,
```
didier-wenzek, line 64: An option to address the concerns raised by https://github.com/thin-edge/thin-edge.io/pull/3230/files#r1907088899, is to:

- remove the `external_id` from the `EntityMetadata`
- and let the mapper manipulates `(EntityExternalId, EntityMetadata)` pairs
- these external ids being a mapper specific  concern.

```suggestion
    entities: HashMap<EntityTopicId, (EntityExternalId,EntityMetadata)>,
    external_id_map: HashMap<EntityExternalId, EntityTopicId>,
```

Not a simple change, though.

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -0,0 +1,353 @@
...
+            topic_id: entity.topic_id.clone(),
+            external_id: Some(external_id.clone()),
+            r#type: entity.r#type,
+            parent,
+            other: entity.other,
+            twin_data: Map::new(),
+        };
+
+        let outcome = self.insert(entity.topic_id.clone(), entity_metadata);
+        self.external_id_map.insert(external_id, entity.topic_id);
```
didier-wenzek, line 162: Updating the `external_id_map` must be done *inside* `self.insert()`.

Indeed, the `insert` method is `pub` and can possibly be invoked from elsewhere, breaking the consistency between the `entities` map and its `external_id_map` index.

didier-wenzek (commented):
The major points have been addressed and I will be happy to see this PR merged.

To fix before merging:
 
- coordinate with https://github.com/thin-edge/thin-edge.io/issues/3323 and 1.4.2 release.
- module renaming: `file_transfer_server` -> `tedge_server` or `http_server`
- unwrapping `EntityStore::with_main_device_and_default_service_type` result.
- fragile entity cache updates

To be addressed in follow-up tasks:

- //TODO: Migrate the existing `clean_start` setting which is C8Y specific without breaking backward compatibility. (agent.rs:376)
- //TODO Remove const EARLY_MESSAGE_BUFFER_SIZE: usize = 100;

crates/core/tedge_api/src/entity.rs
didier-wenzek: See https://github.com/thin-edge/thin-edge.io/pull/3230#discussion_r1910506459 for a proposal to fix that.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -352,9 +372,42 @@ impl Agent {
...
+            let clean_start = self.config.entity_store_clean_start;
+            let main_device = EntityRegistrationMessage::main_device(None);
+            let entity_store = EntityStore::with_main_device_and_default_service_type(
+                mqtt_schema.clone(),
+                main_device,
+                self.config.service.ty.clone(),
+                state_dir,
+                clean_start,
+            )
+            .unwrap();
```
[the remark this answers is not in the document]
albinsuresh: The decision to fail fast on startup for such critical errors was intentional, and was done like this when the entity store was earlier used in the mapper as well. Since this issue can be detected during the startup itself, isn't it better to fail fast rather than silently ignoring it and proceeding?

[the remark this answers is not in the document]
didier-wenzek: To fail fast when a key component cannot even be created, yes, definitely. But with an error message telling what is missing. This call being in a method returning `Result<(), anyhow::Error>`, the `unwrap()` looks like an oversight.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -197,12 +125,11 @@ impl EntityStore {
             })?
         };
 
+        let telemetry_cache_size = 0; // Agent need not cache any data messages, the mapper would
```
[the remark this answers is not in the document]
didier-wenzek: > That being said, the existence of this entity store in `tedge_api` is questionable now as I feel it better belongs in the agent code now (Just like `entity_cache` is part of the mapper now).

Moving the telemetry cache out of the entity store is a good idea. To explore in a follow up task.

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
albinsuresh: Resolved by 5c917d1

crates/core/tedge_api/src/entity.rs
```diff
@@ -0,0 +1,168 @@
...
+        self.external_id
+            .as_ref()
+            .expect("External ID must have been set")
+    }
+}
+
+// TODO: Move to entity module
```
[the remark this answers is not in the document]
albinsuresh: Resolved by 5c917d1

crates/core/tedge_agent/src/agent.rs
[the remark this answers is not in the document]
albinsuresh: Resolved by 5c917d1

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -317,47 +312,26 @@ impl CumulocityConverter {
                 }
                 _ => {
                     // if the target entity is unregistered, try to register it first using auto-registration
```
rina23q, line 314: I guess this comment is no longer applicable.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -317,47 +312,26 @@ impl CumulocityConverter {
...
-                    {
-                        let auto_registered_entities = self.try_auto_register_entity(&source)?;
-                        Ok(auto_registered_entities
-                            .into_iter()
-                            .map(|reg_msg| reg_msg.into())
-                            .collect())
-                    } else {
+                    if self.entity_cache.get(&source).is_none() {
                         // On receipt of an unregistered entity data message with custom topic scheme OR
                         // one with default topic scheme itself when auto registration disabled,
```
rina23q, line 317: Also I guess we can remove the `OR ...` sentence as this condition is removed from the mapper.
```suggestion
                        // On receipt of an unregistered entity data message with custom topic scheme
```

rina23q (commented):
Some minor comments that should be easy to address.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+}
+
+impl EntityStoreServer {
+    async fn process_mqtt_message(&mut self, message: MqttMessage) {
+        let (topic_id, channel) = self
+            .mqtt_schema
+            .entity_channel_of(&message.topic)
+            .expect("Topic schema must have been pre-validated");
```
didier-wenzek, line 97: Actually not safe. Got a panic starting the updated agent in my dev environment (which is not clean, I admit - but this should not crash the agent).

```
thread 'tokio-runtime-worker' panicked at crates/core/tedge_agent/src/entity_manager/server.rs:97:14:
Topic schema must have been pre-validated: Channel(InvalidCategory("cmd/builtin/log_upload/sub:log_upload:c8y-mapper-52201"))
```

Updated; any weird topic makes the agent to crash. => This must be fixed.

didier-wenzek (changes requested):

didier-wenzek (changes requested):
I confirm that there are many points to be fixed. None are catastrophic  but as whole the feature is not ready.

## Differences with the plan

Compared to the design [docs](https://github.com/thin-edge/thin-edge.io/blob/main/design/decisions/0005-entity-registration-api.md):

- Neither `PATCH` nor `PUT` are supported
- Querying entities with curl '/v1/entities is not implemented

That's okay for this PR, but the query interface would be really handy for testing.

## Weird log entry

Auto-registration of a child device leads to a weird log entry:

```shell
2025-01-14T13:37:38.769460295Z  INFO tedge_agent::entity_manager::server: Entity: EntityRegistrationMessage { topic_id: EntityTopicId("device/child3//"), external_id: None, type: ChildDevice, parent: Some(EntityTopicId("device/main//")), other: {"name": String("child3")} } not registered but cached for later as its parent is not registered yet
```

This is weird because the parent device is the main device
and because the registration happens as expected but not as warned:

```
[c8y/s/us] 101,demo-device-888:device:child3,child3,thin-edge.io-child
[c8y/measurement/measurements/create] {"externalSource":{"externalId":"demo-device-888:device:child3","type":"c8y_Serial"},"time":"2020-10-15T05:30:47Z","temperature":{"temperature":{"value":25.0}},"type":"environment"}
[c8y/s/us/demo-device-888:device:child3] 117,60
```

The behavior is the same with registration over MQTT or HTTP

```shell
$ tedge mqtt pub 'te/device/child6//' '{"@type": "child-device"}'
```

and over HTTP:

```shell
$ curl http://localhost:8000/tedge/entity-store/v1/entities/device/child07// \
   -H 'Content-Type: application/json' -d '{"@topic-id": "device/child07//", "@type": "child-device"}'
```

- An explicit registration of the main device changes nothing.
- Same INFO is observed when create a grand-child device of an already registered device


## Unhelpful error messages on delete

Deleting a child device works:

```shell
$ curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01 -X DELETE
```

But both the agent and the mapper emit unhelpful error messages:

On the agent:

```
2025-01-14T16:02:52.68266318Z ERROR tedge_agent::entity_manager::server: Failed to update entity store with [te/device/child01// qos=1]
```

and on the mapper:

```
2025-01-14T15:41:08.937846187Z ERROR c8y_mapper_ext::converter: Mapping error: Unexpected error: Invalid entity registration message received on topic: te/device/child01//
```

## Mandatory properties

It's surprising that the topic identifier of entity has to be provided *twice*.

It's required on the topic:

```shell
$ curl http://localhost:8000/tedge/entity-store/v1/entities -v \
  -H "Content-Type: application/json" \
  -d '{
    "@topic-id": "device/child01//", "@type": "child-device", "@parent": "device/main//"
  }'

< HTTP/1.1 404 Not Found
```

and as a property:

```shell
curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01 -v \     
  -H "Content-Type: application/json" \
  -d '{
    "@type": "child-device", "@parent": "device/main//" 
  }'

< HTTP/1.1 422 Unprocessable Entity

Failed to deserialize the JSON body into the target type: missing field `@topic-id` at line 3 column 3
```

leading to surprises!

```shell
$ curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01 -v \
  -H "Content-Type: application/json" \
  -d '{
    "@topic-id": "not/child01//", "@type": "child-device", "@parent": "device/main//"
  }'
  
< HTTP/1.1 200 OK

$ curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01     
Entity not found with topic id: device/child01//%                    
                                                                                                       
$ curl http://localhost:8000/tedge/entity-store/v1/entities/not/child01   
{"@topic-id":"not/child01//","@parent":"device/main//","@type":"child-device"}     
```

crates/core/tedge_agent/src/file_transfer_server/entity_store.rs
```diff
@@ -0,0 +1,373 @@
...
+
+        (status_code, error_message).into_response()
+    }
+}
+
+pub(crate) fn entity_store_router(state: AgentState) -> Router {
+    Router::new()
+        .route(
+            "/v1/entities/*path",
+            post(register_entity)
```
Bravo555, line 68: suggestion: Have `POST` for `v1/entities`, not `v1/entities/*path`

In the proposal, `POST` is declared for `v1/entities` collection, not the resource, so is the change here desired or a leftover from a simpler implementation?
Putting this route on a collection level would also fix the "having to put topic id twice" problem.
If there is a good reason for requiring a path here, it should be documented.

albinsuresh: I've already fixed this already, following Didier's comments. It came from my desire to keep the topics of all the CRUD APIs consistent during the early days. But, following a discussion with the team, we had agreed on dropping the topic id from the path for this POST request.  It just wasn't cleaned up from the impl.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -0,0 +1,60 @@
...
+    ...    c8y/s/us
+    ...    message_contains=101,${DEVICE_SN}:device:child02
+
+    ${get}=    Execute Command    curl http://localhost:8000/tedge/entity-store/v1/entities/device/child02//
+    Should Be Equal
+    ...    ${get}
+    ...    {"@topic-id":"device/child02//","@parent":"device/main//","@type":"child-device"}
+
+Entity auto-registration over MQTT
+    Execute Command    tedge mqtt pub te/device/child01/service/collectd/m/ram '{"current": 6 }'
```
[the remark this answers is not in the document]
albinsuresh: Resolved by 84bf5c3

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -0,0 +1,60 @@
...
+    Should Have MQTT Messages
+    ...    te/device/child01//
+    ...    message_contains="@type":"child-device"
+
+    ${status}=    Execute Command
+    ...    curl -o /dev/null -s -w "%\{http_code\}" -X DELETE http://localhost:8000/tedge/entity-store/v1/entities/device/child01//
+    Should Be Equal    ${status}    200
+
+    ${get}=    Execute Command
+    ...    curl -o /dev/null -s -w "%\{http_code\}" http://localhost:8000/tedge/entity-store/v1/entities/device/child01//
```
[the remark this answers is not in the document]
albinsuresh: Resolved by  84bf5c3

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -1010,6 +1010,18 @@ define_tedge_config! {
...
+
+        entity_store: {
+            /// Enable auto registration feature
+            #[tedge_config(example = "true", default(value = true))]
+            auto_register: bool,
+
+            /// On a clean start, the whole state of the device, services and child-devices is resent to the cloud
+            #[tedge_config(example = "true", default(value = true))]
+            clean_start: bool,
+        },
```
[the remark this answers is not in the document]
albinsuresh: Resolved by d88c054

crates/core/c8y_api/src/json_c8y.rs
```diff
@@ -241,36 +240,34 @@ pub struct C8yClearAlarm {
 impl C8yAlarm {
     pub fn try_from(
```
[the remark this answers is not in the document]
albinsuresh: Resolved by d88c054

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Resolved by d88c054

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Resolved by d88c054

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -0,0 +1,72 @@
...
+    ${get}=    Execute Command    curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01//
+    Should Be Equal
+    ...    ${get}
+    ...    {"@topic-id":"device/child01//","@parent":"device/main//","@type":"child-device"}
+    Should Have MQTT Messages
+    ...    te/device/child01//
+    ...    message_contains="@type":"child-device"
+
+    Execute Command
+    ...    curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities -H 'Content-Type: application/json' -d '{"@topic-id": "device/child01//", "@type": "child-device", "@id": "child01"}'
```
reubenmiller, line 29: Shouldn't updating an entity item use the path to the entity?

```
curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/child01// -H 'Content-Type: application/json' -d '{"@topic-id": "device/child01//", "@type": "child-device", "@id": "child01"}'
```

As the `HTTP GET` is using the entity path, e.g. `curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01//`

albinsuresh: I was a bit confused here as to what the right convention is. Since this API is more of an `upsert` rather than an `update`, I thought it's better to keep the API similar to that of `create`(`POST`). If it was a pure `update` (`PATCH`), the topic id definitely would have been accepted via the path as per the conventions.

didier-wenzek: The convention is that the URL is the identifier of the resource to create, read, update or delete.

```
curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/child01// \
       -H 'Content-Type: application/json' \
      -d '{"@type": "child-device", "@id": "child01"}'
```

albinsuresh: > The convention is that the URL is the identifier of the resource to create, read, update or delete.

Yeah, I also would have liked that consistency in the paths. But when we discussed this last time, esp for the `POST` request, I was informed that the convention for `POST` requests to create a new resource is typically accepted to the path of the collection (`/entity-store/v1/entities` in this case) rather than the resource identifier itself.

The difference in our case, compared to a conventional case is that the unique identifier for the resource is provided by the user rather than the system. So, may be it's okay to go against the above mentioned "generic convention" in our case and accept the topic id in the URL itself for all CRUD operations.

reubenmiller: Yes the `POST` call is fine to use the `/entities` endpoint, but the `PUT`, `DELETE`, `GET` should use the resource name, e.g. `entities/device/child01//`

reubenmiller: The main advantage of having the `POST` not requiring the `/entities/...` path is that on some systems you may not actually know the resource path to an already created item, so creating the resource via the `/entities` endpoint allows the API to then return a reference to the resource url related to the created entity.

reubenmiller: > See Define API operations in terms of HTTP methods

The example shows that the `POST` method does not use a URL pointing to the resource, it only points to the type of resource being created (in their example it is `/customers` instead of our case of `/entities`)

albinsuresh: Reverted the PUT implementation as discussed offline. A proper implementation as suggested here would be introduced in a follow-up PR.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: Resolved by f752408

docs/src/operate/registration/register.md
```diff
@@ -0,0 +1,111 @@
...
+      "@topic-id": "device/child01//",
+      "@id": "child01",
+      "name": "child01"
+  }
+  ```
+
+**Example**
+
+```shell
+curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01
```
reubenmiller, line 86: Does the API also accept leaving out the trailing `//` in the topic-id?

didier-wenzek: Yes, this works.

didier-wenzek (approved):
This PR can be merged once renamed the `file_transfer_server` module.

Required follow-up tasks:

- PUT and PATCH implementation
- Query over registered entities
- Publishing entity properties as twin metadata

rina23q (approved):

Bravo555 (approved):

PR #3231, opened 2024-11-06
fix: match operation without topic only when it was delivered by device control topic

Ruadhri17:
## Proposed changes
This PR fixes issue introduced with #3198. Previously we always matched operation that had no topic provided in the operation file as we assumed the default one is device control topic. Now operation can be run with any topic that user provides in the operation file so this rule no longer applies. Operation without provided topic is added to operation handlers no matter which topic delivered the message, wheras it should be added to handlers only if the message was delivered by device control topic.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/smartrest/operations.rs
+++ b/crates/core/c8y_api/src/smartrest/operations.rs
@@ -1,3 +1,4 @@
+use crate::json_c8y_deserializer::C8yDeviceControlTopic;
 use crate::smartrest::error::OperationsError;
 use crate::smartrest::smartrest_serializer::declare_supported_operations;
 use mqtt_channel::TopicFilter;
@@ -10,6 +11,7 @@ use std::path::Path;
 use std::path::PathBuf;
 use std::time::Duration;
 use tedge_api::substitution::Record;
+use tedge_config::TopicPrefix;
 use tracing::warn;
 
 use super::payload::SmartrestPayload;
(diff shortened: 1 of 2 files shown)


crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -57,11 +59,17 @@ impl Operations {
...
         let mut vec: Vec<(String, Operation)> = Vec::new();
         for op in self.operations.iter() {
             match (op.topic(), op.on_fragment()) {
-                (None, Some(on_fragment)) => vec.push((on_fragment, op.clone())),
+                (None, Some(on_fragment)) if C8yDeviceControlTopic::name(prefix) == topic_name => {
+                    vec.push((on_fragment, op.clone()))
+                }
                 (Some(topic), Some(on_fragment)) if topic == topic_name => {
                     vec.push((on_fragment, op.clone()))
                 }
```
didier-wenzek, line 75: An alternative way to fix the issue is to use `C8yDeviceControlTopic::name(prefix)` as a default value for `OnMessageExec::topic`. Doing so would be less error prone.

didier-wenzek (commented):
As this fix an issue introduced by another PR, it would be good to add a test reproducing the case.

rina23q (commented):
The fix looks good, however, unit test is missing.

crates/core/c8y_api/src/smartrest/operations.rs
Ruadhri17: In therory I agree but I cannot use it as serde default and changing `.topic()` function that unwraps to `C8yDeviceControlTopic::name(prefix)` when topic is `None` requires to provide that prefix almost in every single function in `operations.rs` (also in places that does not need that information) so I find it inconvenient.

didier-wenzek (approved):
Approved

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -393,7 +401,7 @@ pub enum InvalidCustomOperationHandler {
 
 #[cfg(test)]
 mod tests {
-    use std::io::Write;
+    use std::{io::Write, str::FromStr};
```
rina23q, line 404: Format it :)

rina23q (approved):
Looks good now. Just fix the format check, then it will be good to merge.

Issue #2868, opened 2024-05-09
Publish registration and capabilities on te topics when connecting to the MQTT broker


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

If the local MQTT broker (e.g. mosquitto) is not configured with persistence, all retain messages are lost when the service is restarted. As a result all devices registration message, supported commands will not be sent to the platform

1. Start the tedge-agent service (before connection to the cloud)
2. Configure thin-edge.io to connect to Cumulocity IoT (this will al)

    ```sh
    tedge cert create --device-id example001
    tedge config set c8y.url example.c8y.com
    tedge connect c8y
    ```

    The `tedge connect c8y` will also restart the mosquitto broker which will remove all of the previously published retained message by tedge-agent before the tedge-mapper-c8y has seen the messages. This results in the device in the cloud having only partial information about the device.

**Describe the solution you'd like**

Since registration and device capabilities are fairly critical for other components to understand what the device is and what it can do (e.g. does it accept get/set configuration, software_update, firmware_update command etc.), it might make sense to publish these message every time the component connects to the MQTT broker (and not assume these message are persisted when the MQTT broker is restarted).

**Describe alternatives you've considered**

**Additional context**

Below shows an example of the reduced number MQTT message after the MQTT broker is restarted (when it is not configured with persistence).

Before the MQTT broker is restarted, the tedge-agent has previously registered itself and its capabilities:

```sh
% tedge mqtt sub '#'
INFO: Connected
[te/device/main/service/mosquitto-c8y-bridge] {"@id":"rmi-macos:device:main:service:mosquitto-c8y-bridge","@parent":"device/main//","@type":"service","name":"mosquitto-c8y-bridge","type":"service"}
[te/device/main/service/mosquitto-c8y-bridge/status/health] 1
[te/device/main/service/tedge-agent] {"@parent":"device/main//","@type":"service","type":"service"}
[te/device/main/service/tedge-agent/status/health] {"pid":66677,"status":"up","time":1715267898.9250402}
[te/device/main/service/tedge-mapper-c8y] {"@id":"rmi-macos:device:main:service:tedge-mapper-c8y","@parent":"device/main//","@type":"service","name":"tedge-mapper-c8y","type":"service"}
[te/device/main/service/tedge-mapper-c8y/status/health] {"pid":61406,"status":"up","time":1715267770.182519}
[...]

PR #3232, opened 2024-11-06
feat: enable mosquitto persistence by default

reubenmiller:
## Proposed changes

**Open Questions**
* [ ] Should the logic be moved to the install script so that it can check if the user has configured the `persistence true` or not and configured the path (with the correct permissions set).

Enable persistence as thin-edge.io heavily relies on retained messages and if mosquitto persistence is not configured, then it can lead to unexpected behavior.

For example (assuming mosquitto persistence is not enabled), if the *tedge-agent* has already started, and then the user runs `tedge connect c8y`, then any previously retained message published by *tedge-agent* will not be republished, and the tedge-mapper-c8y will not know of the previously published retained messages...this leads to supported operations not being sent to the cloud.

Given the impact of not having mosquitto persistence configured, it should not be left up to the user to configure it.

## Types of changes

## Paste Link to the issue

* Similar issue: https://github.com/thin-edge/thin-edge.io/issues/2868

## Checklist

### DIFF
--- a/crates/core/tedge/src/bridge/common_mosquitto_config.rs
+++ b/crates/core/tedge/src/bridge/common_mosquitto_config.rs
@@ -110,6 +110,10 @@ impl Default for CommonMosquittoConfig {
 
 impl CommonMosquittoConfig {
     pub fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
+        // Enable persistence as thin-edge.io heavily relies on retained messages and
+        // if mosquitto persistence is not configured, then it can lead to unexpected behavior
+        writeln!(writer, "persistence true")?;
+
         writeln!(writer, "per_listener_settings true")?;
 
         writeln!(writer, "connection_messages true")?;
@@ -237,6 +241,7 @@ mod tests {
             .collect();
         let mut expected = std::collections::HashSet::new();
 
+        expected.insert("persistence true");
         expected.insert("listener 1883 127.0.0.1");
         expected.insert("allow_anonymous true");
         expected.insert("connection_messages true");
@@ -280,6 +285,7 @@ mod tests {
 
         let contents = String::from_utf8(buffer).unwrap();
         let expected = concat!(
+            "persistence true\n",
             "per_listener_settings true\n",
             "connection_messages true\n",
             "log_type error\n",


albinsuresh (approved):

reubenmiller:
Though I'm just reviewing one aspect regarding the `persistence_location` (https://mosquitto.org/man/mosquitto-conf-5.html).

If non is defined it will choose the current location, but this may be problematic as I don't see the mosquitto service defining the default working directory, which means the current working directory maybe somewhere that the `mosquitto` user does not have permission to write to.

>persistence_location path
  The path where the persistence database should be stored. If not given, then the current directory is used.

@albinsuresh @didier-wenzek Any thoughts on this? Should we try to set the default path ourselves? Or even consider creating our own mosquitto configuration package?

crates/core/tedge/src/bridge/common_mosquitto_config.rs
```diff
@@ -110,6 +110,10 @@ impl Default for CommonMosquittoConfig {
 
 impl CommonMosquittoConfig {
     pub fn serialize<W: std::io::Write>(&self, writer: &mut W) -> std::io::Result<()> {
+        // Enable persistence as thin-edge.io heavily relies on retained messages and
+        // if mosquitto persistence is not configured, then it can lead to unexpected behavior
+        writeln!(writer, "persistence true")?;
```
didier-wenzek, line 115: Beware that the file `/etc/tedge/mosquitto-conf/tedge-mosquitto.conf` is populated only if there is at least one cloud connected.

I would consider to improve the `tedge refresh-bridges` command to populate this file even if there is no bridge configured. I.e. moving [these 2 lines](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/cli/refresh_bridges.rs#L38) before [that one](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/cli/refresh_bridges.rs#L33)

albinsuresh:
> @albinsuresh @didier-wenzek Any thoughts on this? Should we try to set the default path ourselves? Or even consider creating our own mosquitto configuration package?

Can we not write a new config file with these persistence settings into the `/etc/mosquitto/conf.d` directory once `tedge` is **installed** and not wait for `tedge connect`? Yes, there is the risk of conflicting persistence settings, if the user already has persistence turned on in some other configuration file. But, we can at least check the main `/etc/mosquitto/mosquitto.conf` file to see if it is already turned ON, before adding ours.

reubenmiller:
> > @albinsuresh @didier-wenzek Any thoughts on this? Should we try to set the default path ourselves? Or even consider creating our own mosquitto configuration package?
> 
> Can we not write a new config file with these persistence settings into the `/etc/mosquitto/conf.d` directory once `tedge` is **installed** or as part of `tedge init` and not wait for `tedge connect`? Yes, there is the risk of conflicting persistence settings, if the user already has persistence turned on in some other configuration file. But, we can at least check the main `/etc/mosquitto/mosquitto.conf` file to see if it is already turned ON, before adding ours.

Ok, I'm going to move the state of this PR to draft until we work out a way forward.

reubenmiller:
Closing as more recent versions of mosquitto on OS's such as Debian have persistence enabled by default. Any custom built images using either Yocto or Rugix (formerly Rugpi) also included the required persistence settings.

Issue #3160, opened 2024-10-04
Files under directory `/etc/tedge/operations/c8y` are loaded many times on c8y-mapper startup


rina23q:
**Describe the bug**
The supported operation files under `/etc/tedge/operations/c8y` should be ideally loaded only once on c8y-mapper startup. However, it is actually loaded many times. This can be explained by tracing how [`get_operations()`](https://github.com/thin-edge/thin-edge.io/blob/0866b9b9d7571f1479571d0600a42923a9b2980f/crates/core/c8y_api/src/smartrest/operations.rs#L183) functions are called.

1. init messages
https://github.com/thin-edge/thin-edge.io/blob/0866b9b9d7571f1479571d0600a42923a9b2980f/crates/extensions/c8y_mapper_ext/src/converter.rs#L1278
2. constructing `CumulocityConverter`
https://github.com/thin-edge/thin-edge.io/blob/0866b9b9d7571f1479571d0600a42923a9b2980f/crates/extensions/c8y_mapper_ext/src/converter.rs#L216
3. registering a new command registration message (as many times as command registration messages received)
https://github.com/thin-edge/thin-edge.io/blob/0866b9b9d7571f1479571d0600a42923a9b2980f/crates/extensions/c8y_mapper_ext/src/converter.rs#L1400

---
My comments:

(2) is actually redundant. Since already `operations` get created during `init`, they should parse to `CumulocityConverter`.

(3) is most problematic. `CumulocityConverter::register_operation()` is called by receiving a command registration message. It means, if user has 5 command registration messages, the supported operation directory will be loaded 5 times. I have a doubt if mapper has to load the directory every time when a command registration message arrives.

**To Reproduce**
Easy check is to use validation for custom operation handler. If the contents of a custom handler is invalid, it logs a warning message. This validation is called inside the `get_operations()` function.

Prepare the version of https://github.com/thin-edge/thin-edge.io/pull/3144.

1. Create a invalid custom operation handler file, e.g. /etc/tedge/operations/c8y/my_CustomOperation 
```toml
[exec]
topic = "c8y/devicecontrol/notifications"
on_fragment = "com_cumulocity_model_WebCamDevice"
command_name = "/usr/bin/do_something.sh ${.payload.com_cumulocity_model_WebCamDevice.parameters}"
```
2. Restart c8y-mapper.
3. See the output of the c8y-mapper. You'll find 7 times of the warning message.
```
[...]

PR #3236, opened 2024-11-08
Don't fail if only some operations could not be loaded

Bravo555:
## TODO

- [ ] stop loading all the operations when registering a single operation

## Proposed changes

When loading operations from $CONFIG_DIR/operations/c8y, log an error for operations we were unable to load but return other successfully loaded operations.

Currently this spams the log output a bit, because we statically register some operations by name, but it causes full directory scan and traversal of all operation files anyway. Will be fixed in next commit.
## Types of changes

## Paste Link to the issue
<br/>

## Checklist


Bravo555:
This PR changes logic of loading c8y operations in the `CumulocityConverter` so that it no longer panics, but invalid operation files are logged with `error` level. However, due to #3160, the output gets spammed by multiple errors regarding the same file.

To ensure that only a single log is emitted for a given operation file, this PR originally aimed to address #3160 as well, by refactoring registering operation flow so that operation directory would be read only a single time. However this optimal solution is made impossible by the inability to add an MQTT subscription after the actor is already spawned, so we're forced to scan the directory when creating `C8yConverterConfig` to read all MQTT topics that are possibly defined by custom c8y operations. Also there are some tests that assert some implementation details, which makes it impossible to completely remove duplicate c8y operation directory traversals.

As such, this PR will remove duplicates traversals where it's feasible, but will not completely address #3160, and as such, the log output will still be a bit spammed by duplicate errors, which will be annoying for users. Solving this may require bigger changes to our MQTT actor, so it will be left for another PR.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -28,8 +29,22 @@ pub struct Operations {
...
+        // by adding operations one by one or by directory scan, we can end up with a different order
+        // TODO: refactor to provide more sane API, potentially using a backing `BTreeMap`
+        self.operations
+            .sort_unstable_by(|o1, o2| o1.name.cmp(&o2.name));
+    }
+
+    pub fn remove_operation(&mut self, name: &str) -> Option<Operation> {
+        self.operations.dedup();
+        let pos = self.operations.iter().position(|op| op.name == name);
+        pos.map(|pos| self.operations.remove(pos))
```
albinsuresh, line 42: If the switch to a `BTreeSet` is very minimal, I'm even okay with you making that switch in this PR itself, to avoid complications like this. Or if you prefer a standalone PR, that's also fine.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -3229,6 +3300,85 @@ pub(crate) mod tests {
         );
     }
 
+    /// Check that register_operation correctly registers operations from MQTT and handles error conditions
+    #[tokio::test]
+    async fn test_register_entity() {
```
albinsuresh, line 3305:
```suggestion
    async fn test_register_operation() {
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1551,23 +1558,52 @@ impl CumulocityConverter {
...
+        )?;
+        let operations = self
+            .operations_for_device_mut(target)
+            .expect("entity should've been checked before that it's not a service");
 
-        let need_cloud_update = self.update_operations(ops_dir.as_std_path())?;
+        let prev_operation = operations.remove_operation(&operation.name);
+        // even if the body of the operation is different, as long as it has the same name, supported operations message
+        // will be the same, so we don't need to resend
+        let need_cloud_update = prev_operation.is_none();
```
albinsuresh, line 1588: Why not move this logic into the existing `update_operations` function, as that function already has the contract to return the boolean flag whether the cloud update is required or not?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1551,23 +1558,52 @@ impl CumulocityConverter {
...
+
+                C8yTopic::ChildSmartRestResponse(child_external_id.into()).to_topic(&c8y_prefix)?
+            } else {
+                C8yTopic::upstream_topic(&c8y_prefix)
+            };
+
+            let cloud_update_operations_message = MqttMessage::new(
+                &topic,
+                operations.create_smartrest_ops_message().into_inner(),
+            );
```
albinsuresh, line 1604: Again, why not reuse some of this logic that is already there in the `create_supported_operations` function?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -3229,6 +3300,85 @@ pub(crate) mod tests {
...
+
+        let operation_topic = converter.mqtt_schema.topic_for(
+            &main_device,
+            &Channel::CommandMetadata {
+                operation: OperationType::Custom("my_operation".to_string()),
+            },
+        );
+        let operation_msg = MqttMessage::new(&operation_topic, "{}");
+
+        converter.convert(&operation_msg).await;
```
albinsuresh, line 3323: With no assertion on the outcome of that call, the intention of the test isn't very clear. I was expecting the `114` message as the result of that new custom operation registration.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -3229,6 +3300,85 @@ pub(crate) mod tests {
...
+        let service = main_device.default_service_for_device("service0").unwrap();
+        let operation_topic = converter.mqtt_schema.topic_for(
+            &service,
+            &Channel::CommandMetadata {
+                operation: OperationType::Custom("my_operation".to_string()),
+            },
+        );
+        let operation_msg = MqttMessage::new(&operation_topic, "{}");
+
+        converter.convert(&operation_msg).await;
```
albinsuresh, line 3379: Even though we're not expecting any 114 messages for this service case, it'd be better to do an explicit assertion that the returned vector is empty.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1425,6 +1425,8 @@ async fn mapper_dynamically_updates_supported_operations_for_tedge_device() {
 }
 
 #[tokio::test]
+// TODO: fix or remove test
+#[ignore = "asserts that publishing a single operation capability message causes full rescan of c8y operations directory, which is undesirable behaviour"]
```
albinsuresh, line 1429: I understand that the rescan is not optimal. But, we shouldn't be skipping the whole test for that reason. The dynamic ops update feature is still relevant and must be tested.

crates/core/c8y_api/src/smartrest/operations.rs
didier-wenzek: Yes, better to switch to a `BTreeSet` right now.

Bravo555: I would really prefer to avoid big changes to `Operations` struct in this PR because of parallel PR #3225, after one of these merges we expect to see conflicts which I'd like to minimise.
I'd also like to reorganize the module a bit in the refactor, so I'd rather do it as a separate PR.

crates/extensions/c8y_mapper_ext/src/tests.rs
Bravo555: Yeah, this will need to be addressed before the merge.
Just that from my perspective, it looks like the primary point of this test is to assert the exact behaviour I'm removing (registering any operation via MQTT causes full rescan of c8y operations directory), so I'm not really sure what behaviour we could be checking instead.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: You're right, and to be honest I wrote additional tests because I saw the test coverage decrease, so I meant to cover new branches I added, but in focusing on that I neglected to actually add assertions that verify the output is correct :facepalm:  I'll fix the test.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: Yep, in this case `create_supported_operations` can probably be used and i just missed it.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: fixed in 01ec205849

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: fixed in 01ec205849

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: fixed in 01ec205849

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: Turned out the vector is not empty, there's an MQTT message with an error about operations for services being unsupported

fixed in 01ec205849

crates/extensions/c8y_mapper_ext/src/tests.rs
Bravo555: Okay, I discovered the reason why the test checked that behaviour (#2614), and found that my implementation was incorrect, and in fact, we need to still maintain the old behaviour for child devices.

Fixed in a313601

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: The function changed a bit in a313601, and I ended up with two branches, in the case of a child device I just call `update_operations` outright. Let me know if this addresses it, or if you had something more particular in mind.

rina23q (approved):
As I confirmed that it no longer makes a panic. I would approve it. There are still many small things to be improved regarding `Operations` struct, however, I believe time to move on.

> As such, this PR will remove duplicates traversals where it's feasible, but will not completely address https://github.com/thin-edge/thin-edge.io/issues/3160, and as such, the log output will still be a bit spammed by duplicate errors, which will be annoying for users. Solving this may require bigger changes to our MQTT actor, so it will be left for another PR.

I saw the error messages are now much less (now 2 vs before around 7, the numbers of operation files at tedge-mapper-c8y's startup). Thanks for improving.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1568,51 +1574,59 @@ impl CumulocityConverter {
...
-                let child_external_id = Self::validate_external_id(&child_id)?;
+                let current_operations = self
+                    .operations_for_device_mut(target)
+                    .expect("entity should've been checked before that it's not a service");
 
-                C8yTopic::ChildSmartRestResponse(child_external_id.into()).to_topic(&c8y_prefix)?
-            } else {
-                C8yTopic::upstream_topic(&c8y_prefix)
-            };
+                let prev_operation = current_operations.remove_operation(&operation.name);
```
albinsuresh, line 1602: In your refactoring PR, consider updating the signature of `Operations::add_operation` function to return the previous operation so that we can avoid this removal first and addition one step later.

albinsuresh (approved):
Despite the shortcomings of the `Operations` structure, this is sufficient for the bug fix.

Issue #3240, opened 2024-11-12
allow setting the device.id when using c8y basic auth


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

With the introduction of Basic Auth in https://github.com/thin-edge/thin-edge.io/issues/3036 (to support existing/legacy Cumulocity solutions), the solution still requires a dummy certificate to be set so that the correct `device.id` field can be read. The `device.id` tedge configuration property is a special property which is read from the existing public device certificate, however when using Basic Auth, a certificate is not required, so it is unexpected to demand that the user create a dummy certificate.

**Describe the solution you'd like**

When either of the following conditions is true:
* c8y.auth_mode is set to `basic`
* c8y.auth_mode is set to `auto` and the `c8y.credentials_path` file exists

Then, the `device.id` tedge configuration property should be read from.

```toml
[device]
id = "<device_id>"
```

This would require the user to set the value via the `tedge config set` command.

```sh
tedge config set device.id mydeviceid
```

However, if the above conditions are not met, then the user should be presented with the current error message (when trying to set the `device.id` when using device certificates):

```sh
tedge config set device.id mydeviceid
error: invalid value 'device.id' for '<KEY>': The device id is read from the device certificate and cannot be set directly.
To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.

For more information, try '--help'.
```

**Describe alternatives you've considered**

When using Cumulocity Basic Auth mode, the `device.id` could be derived from the `c8y.username` property in the `c8y.credentials_path` file, for example a value of `t1234/device_myexample` would be interpreted as `myexample`.

This solution is not ideal as it relies on very specific Cumulocity parsing of the username, and also assumes that the username and the device.id are the same (where technically they do not need to match!)

**Additional context**

rina23q:
Resolved by #3326

PR #3242, opened 2024-11-12
feat: read device.id from credentials.toml when basic auth mode

rina23q:
## Proposed changes
Draft for further discussion.

1. Technically, it would be very difficult to switch "read-only" mode or "read-write" mode depending on `c8y.auth_mode`. This PR proposes to keep `device_id` in the `credentials.toml` file alternatively.
2. Reading all configs under `c8y` requires a profile name. However, it cannot be determined when `tedge config get device.id` is called. I parsed `None` as the device profile. Correct me if I am wrong.
```rust
    let c8y_profile: Option<&ProfileName> = None;
    let c8y_config = reader.c8y.try_get(c8y_profile)?;
```

## Types of changes

## Paste Link to the issue
#3240 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/models/auth_method.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/models/auth_method.rs
@@ -57,3 +57,22 @@ impl AuthMethod {
         }
     }
 }
+
+pub fn try_get_device_id_from_credentials_file(credentials_path: &Utf8Path) -> Option<String> {
+    if let Ok(contents) = std::fs::read_to_string(credentials_path) {
+        if let Ok(credentials) = toml::from_str::<CredentialsFile>(&contents) {
+            return Some(credentials.c8y.device_id);
+        }
+    }
+    None
+}
+
+#[derive(Debug, serde::Deserialize)]
+struct CredentialsFile {
+    c8y: C8y,
+}
+
+#[derive(Debug, serde::Deserialize)]
+struct C8y {
+    device_id: String,
+}
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -1,4 +1,5 @@
 use super::models::timestamp::TimeFormat;
+use crate::auth_method::try_get_device_id_from_credentials_file;
 use crate::auth_method::AuthMethod;
 use crate::AptConfig;
 use crate::AutoFlag;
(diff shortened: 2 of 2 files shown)


albinsuresh:
> Technically, it would be very difficult to switch "read-only" mode or "read-write" mode depending on c8y.auth_mode. This PR proposes to keep device_id in the credentials.toml file alternatively.

Why don't we permanently switch this setting to `read-write` mode and apply the following rule:

1. If `auth_mode == basic` then read the written value
2. If `auth_mode == certificate` then read from the certificate

If someone consciously sets `device.id` in the `tedge.toml` and uses the `certificate` mode as well, then we can just validate if both are the same, else fail during `tedge connect`.

rina23q:
Since both `c8y.auth_mode` and `c8y.credentials_path` can differ based on a cloud profile, we should not use any `c8y` config for `device.id`. In essence, `device.id` must be identical regardless of cloud profiles, so it makes sense by design as well.

If tedge config macro has `try_from()` attribute, the implementation would be nice. We already have `from()` attribute. This usecase needs something that returns ` Result<String, ReadError>`. 

Probably the `device.id` definition can be like below:

```rust
#[tedge_config(try_from(
    error = "\
        The device id is read from the device certificate if it exists.\n\
        To create a device certificate, you can use `tedge cert create --device-id <id>`.",
    function = "device_id",
))]
#[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
#[tedge_config(note = "This setting is derived from the device certificate if it is set.")]
#[doku(as = "String")]
id: Result<String, ReadError>,
```

albinsuresh:
> Since both `c8y.auth_mode` and `c8y.credentials_path` can differ based on a cloud profile, we should not use any `c8y` config for `device.id`. In essence, `device.id` must be identical regardless of cloud profiles, so now I'm convinced to use `tedge.toml` to store the value of device ID for basic auth.
> 
> **Proposal 1:**
> 
> `device.id` returns a value from the certificate when the certificate file exists. Else, returns a value from `tedge.toml`.
> 
> **Proposal 2:**
> 
> `device.id` returns a value from the certificate if it is not set (=default). If it is set, the value is read from `tedge.toml`.
> 
> ```

I'd go with proposal 2, as that is sounds straightforward. If someone has explicitly set a value using `tedge config set`, he'd want to use it. And do that extra validation if they are equal in `tedge connect`.

didier-wenzek:
> Since both `c8y.auth_mode` and `c8y.credentials_path` can differ based on a cloud profile, we should not use any `c8y` config for `device.id`. In essence, `device.id` must be identical regardless of cloud profiles, so now I'm convinced to use `tedge.toml` to store the value of device ID for basic auth.
> **Proposal 1:** 
> `device.id` returns a value from the certificate when the certificate file exists. Else, returns a value from `tedge.toml`.
> **Proposal 2:**
> `device.id` returns a value from the certificate if it is not set (=default). If it is set, the value is read from `tedge.toml`.

Things are already too complicated and both proposals are only making things more complicated.

I do think we must stop reading the device id from the certificate:
- today, we have this issue with password authentication where there is no more certificate to read from
- tomorrow, we will have issues with cloud endpoints using each a different certificate.

> If someone consciously sets `device.id` in the `tedge.toml` and then uses the `certificate` mode as well, then we can just validate if both values are the same, else fail during `tedge connect`.

This is the key. I  propose to:
- store `device.id` in `tedge.toml` as a regular setting.
- update the `tedge cert create` command, to make the `--device-id` optional and to update the config if provided.
- add a check on `tedge connect` to reject a certificate which CN is not the device id or a password that is not for the device id.

jarhodes314:
Adding a bit of further detail to Didier's proposed solution as we discussed, we'll need to gracefully handle cases where the device ID isn't explicitly set. This can be done by setting `#[tedge_config(reader(private))]` on all the `device.id` fields in the config macro, preventing any code trying to read the device ID from accessing it directly. Then you can add a method `fn id(&self) -> Result<_, _>` to `TEdgeConfgReaderC8yDevice` etc. that can read the device certificate or use the ID field set in the toml. This should work as a drop in replacement for the existing read-only device ID field, which is already accessed by calling `tedge_config.c8y.try_get(profile)?.device.id()?`.

rina23q:
The PR https://github.com/thin-edge/thin-edge.io/pull/3326 took over the contents, therefore, I would close this PR.

Issue #3235, opened 2024-11-08
tegde connect fails


ck-c8y:
**Describe the bug**
Trying to connect the tegde version 1.3.1 fails:
```
Sending packets to check connection. This may take up to 2 seconds.

ERROR: Io(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
Warning: Bridge has been configured, but Cumulocity connection check failed.


ERROR: Io(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
```
We run a version of the tedge-container-bundle 

We test on 4 different systems.
The error occurs only on two systems. On the other two it works.

**To Reproduce**
The error happen independant of the used bridge: builtin or mosquitto.

**Expected behavior**

The command tedge connect should complete successfully.

buildin log info:
[builtin_tedge_connect_c8y.log](https://github.com/user-attachments/files/17675882/builtin_tedge_connect_c8y.log)
[builtin_tedge_mqtt_sub_hash.log](https://github.com/user-attachments/files/17675885/builtin_tedge_mqtt_sub_hash.log)
[builtin_tedge_config_list.log](https://github.com/user-attachments/files/17675878/builtin_tedge_config_list.log)

mosquitto log info:

[mos_tedge_config_list.log](https://github.com/user-attachments/files/17675895/mos_tedge_config_list.log)
[mos_tedge_connect_c8y.log](https://github.com/user-attachments/files/17675896/mos_tedge_connect_c8y.log)
[mos_tedge_mqtt_sub_hash.log](https://github.com/user-attachments/files/17675897/mos_tedge_mqtt_sub_hash.log)

reubenmiller:
This smells like a tenant configuration problem. Any clues on the relationship between tenants where it works, and where it doesn't?

reubenmiller:
Could you also try connecting to the same tenant which as problems using the `c8y tedge demo start mytest001`?

ck-c8y:
> This smells like a tenant configuration problem. Any clues on the relationship between tenants where it works, and where it doesn't?
Sorry, when I say 2 system are ok, 2 system fails. 
I meant to say it is always the same Cumulocity tenant. But different thin-edges.

ck-c8y:
We debugged the problem further. And it turns out that tedge connect command shows an error:
``` 
ERROR: Io(Os { code: 111, kind: ConnectionRefused, message: "Connection refused" })
[...]

PR #3246, opened 2024-11-13
bug: wait for mosquitto to start rather than blindly sleeping in `tedge connect`

jarhodes314:
## Proposed changes
Change the behaviour of `tedge connect` to wait until mosquitto is listening on the configured port, rather than blindly sleeping for 5 seconds and hoping that's long enough. This drastically speeds up `tedge connect c8y` in most cases, and allows us to increase the timeout (which I've set here to 20 seconds) to allow for cases where mosquitto starts more slowly.

Additionally, I've now added some logic to republish the check connection message when the built-in bridge publishes a health up message. This fixes connection check errors caused when we send the message too soon and the bridge misses the message entirely. We can't rely on message persistence here since the bridge may have never previously connected.

Still TODO:
- [ ] Finish migrating log messages, error messages and warnings to new format
- [ ] Update `tedge connect c8y --test` to do URL check too

## Types of changes

## Paste Link to the issue
- #3235

## Checklist


didier-wenzek (approved):
Approved

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -542,10 +603,20 @@ fn check_device_status_aws(
...
+                    break;
+                } else if is_bridge_health_up_message(&response, &built_in_bridge_health) {
+                    // Built in bridge is now up, republish the message in case it was never received by the bridge
+                    client.publish(
+                        &aws_topic_pub_check_connection,
+                        AtLeastOnce,
+                        false,
+                        REGISTRATION_PAYLOAD,
+                    )?;
+                }
```
albinsuresh, line 619: Isn't this logic valid for the mosquitto bridge as well? Also wondering why don't we move the `is_bridge_health_up_message` check before entering this event loop or even before the test client connection is established.

jarhodes314: For the mosquitto bridge, I don't think this is true. We don't need the cloud connection to exist; we just need the bridge to have a subscription to the message. I haven't read into this to verify it, but my understanding is that the mosquitto bridge is subscribed to the relevant topics before mosquitto is up for others to listen to.

In the same vein, we arguably don't need to check that the built-in bridge health is up, a down message, even a retained one, proves that the subscription exists (even if it's a persisted retained message, as it's not a clean session). That said, using that as a check would be both confusing and not gain us much since we need the cloud connection anyway to make progress.

didier-wenzek (commented):
The added connection checks are correct. And I'm okay to merge this PR.

However, can this PR be an opportunity to improve the output of `tedge connect`? The current output is verbose and adds little value. Something along the following lines:

```
- Connecting
   - device = demo-device-008
   - mqtt = didier.latest.stage.c8y.io
   - certificate = /etc/tedge/device-certs/demo-device-008.pem 
   - bridge = builtin 
- config: ok
- mosquitto: restarted
- bridge: up
- c8y: connected 
```

jarhodes314:
> The added connection checks are correct. And I'm okay to merge this PR.
> 
> However, can this PR be an opportunity to improve the output of `tedge connect`? The current output is verbose and adds little value. Something along the following lines:
> 
> ```
> - Connecting
>    - device = demo-device-008
>    - mqtt = didier.latest.stage.c8y.io
>    - certificate = /etc/tedge/device-certs/demo-device-008.pem 
>    - bridge = builtin 
> - config: ok
> - mosquitto: restarted
> - bridge: up
> - c8y: connected 
> ```

I've had a bit of a hack at this locally. I've gone for something slightly more verbose than this suggestion in order to make it easy to see what is currently being waited on:
[asciicast](https://asciinema.org/a/sVZsr4Nah35MHga2fdXv3gjYJ)

image

The changes for this aren't finished yet, hence they aren't pushed, but they are working for `tedge connect c8y`

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -460,6 +508,13 @@ fn check_device_status_azure(
                     println!("Received expected response message.");
                     exists = true;
                     break;
+                } else if is_bridge_health_up_message(&response, &built_in_bridge_health) {
```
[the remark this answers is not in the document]
jarhodes314: Yeah, particularly with refactoring the output, the overlap has become obvious. I haven't yet attempted to tackle this, but there is a lot of commonality that we should be able to extract.

jarhodes314:
With some additional context added to the config section, a successful reconnection now looks like:
image

Some error examples:
image
image

The output from disconnecting when the device is already disconnected:
image

And the output of `tedge connect c8y --test` when connected to the wrong tenant:
image

didier-wenzek (commented):
Really nice improvement. I will be happy to approve once the tests green.

didier-wenzek (approved):
I confirm my approval. Connection progress and errors are now nicely reported.

There are still 3 failing system tests - but all related to remote access, a feature currently broken on the CI c8y instance.

Issue #3248, opened 2024-11-14
feat: Integration with Cumulocity CA


didier-wenzek:
A planned PKI feature of Cumulocity IoT will allow
a tenant to act as a Certificate Authority (CA) for its fleet of connected devices.
Once enabled, this feature:
- let the devices request their own certificate directly from their tenant
- the legitimacy of these certificate requests being controlled with security tokens
- which have been pre-registered on the tenant by an administrator during a registration step.

To leverage this feature, thin-edge must be extended along the following lines.

### Device Registration

Security tokens are used to prove the legitimacy of certificate requests issued by the devices.
- The tenant and the device must agree on a device common name (CN) and an associated security token.
- A tenant administrator has to register this (CN, token) pair on the tenant and to give the same pair to the device.
- Then the device can request for a certificate for that CN using the token for authentication.

Technically, thin-edge could provide a command combining these steps. However,
- this might be confusing as this combines two different levels of authentication, for the device and for the tenant user
- this is not flexible as this imposes a workflow where all the registration data are coming from the device.

Hence:

- Thin-edge will only provide a command to request a certificate, being given a device id and a security token.
- Users will have to implement their own workflow to generate device identifiers and tokens,
  to pre-register the devices on their tenant, and finally to trigger on the thin-edge device the certificate creation. 
- To support users implementing their own workflow, [c8y-tedge](https://github.com/thin-edge/c8y-tedge) will be extended
  with a sub-command to register a device (CN, token) pair
  and trigger the certificate request on behalf of the device.

### Certificate Creation

A new `tedge cert download c8y` cli sub-command is introduced, designed along similar lines as for `tedge cert upload c8y`.

```shell
$ tedge cert download c8y --help
Request and download the device certificate from Cumulocity

- generate a private key and Signing Certificate Request (CSR) for the device
- upload this CSR on Cumulocity, using the provided device identifier and security token
- loop till the device is registered by an administrator and the CSR accepted 
- store the certificate created by Cumulocity

Usage: tedge cert download c8y [OPTIONS] --device-id <ID> --token <TOKEN>

Options:
      --device-id <ID>
      
          The device identifier to be used as the common name for the certificate

      --token <TOKEN>
      
          The security token assigned to this device when registered to Cumulocity

      --config-dir <CONFIG_DIR>
      
          Use the following settings from the config:
          
          - c8y.http  HTTP Endpoint for the Cumulocity tenant, with optional port.
[...]

PR #3259, opened 2024-11-21
feat: Integration with Cumulocity CA

didier-wenzek:
## Proposed changes

- [x] Impl tedge cert download c8y
- [x] Impl tedge cert renew c8y
- [x] Enhance tedge config with setting for certificate validity period
- [x] Enhance tedge cert show to display number of days before expiration
- [x] Enhance tedge cert show to display certificate serial number
- [x] Enhance tedge cert show to display any cert 
- [x] Get and renew certificates using CSRs that have been generated elsewhere (typically an HSM) 
- [x] Fix tedge cert renew arguments: cloud vs ca
- [ ] ~Make sure the current certificate is not erased on renewal~ This will be done by an external tool (wrapping the renewal with a reconnect and a rollback on error) 

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3248

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -143,7 +143,7 @@ notify-debouncer-full = { version = "0.3.1", default-features = false }
 once_cell = "1.8"
 pad = "0.1"
 path-clean = "0.1"
-pem = "1.0"
+pem = "3.0"
 pin-project = { version = "1.1.3", features = [] }
 postcard = { version = "1.1.0", features = ["alloc"] }
 predicates = "2.1"
@@ -153,6 +153,8 @@ proc-macro2 = "1"
 proptest = "1.0"
 quote = "1"
 rand = "0.8"
+rasn = "0.18" # Not using the latest version which requires rust 1.85
+rasn-cms = "0.18" # Not using the latest version which requires rust 1.85
 rcgen = { version = "0.12", features = ["pem", "zeroize"] }
 regex = "1.4"
 reqwest = { version = "0.12", default-features = false }
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -166,6 +166,28 @@ define_tedge_config! {
         ty: String,
     },
 
+    certificate: {
+        validity: {
+            /// Requested validity duration for a new certificate
+            #[tedge_config(note = "The CA might return certificates valid for period shorter than requested")]
+            #[tedge_config(example = "365d", default(from_str = "365d"))]
+            requested_duration: SecondsOrHumanTime,
+
+            /// Minimum validity duration below which a new certificate should be requested
+            #[tedge_config(note = "This is an advisory setting and the renewal has to be scheduled")]
+            #[tedge_config(example = "30d", default(from_str = "30d"))]
+            minimum_duration: SecondsOrHumanTime,
+        },
+
+        /// Organization name used for certificate signing requests
+        #[tedge_config(example = "ACME", default(value = "Thin Edge"))]
+        organization: Arc<str>,
+
+        /// Organization unit used for certificate signing requests
+        #[tedge_config(example = "IoT", default(value = "Device"))]
+        organization_unit: Arc<str>,
+    },
+
     #[tedge_config(multi)]
     c8y: {
         /// Endpoint URL of Cumulocity tenant
--- a/crates/core/c8y_api/src/http_proxy.rs
+++ b/crates/core/c8y_api/src/http_proxy.rs
@@ -95,13 +95,17 @@ impl C8yEndPoint {
         }
     }
 
-    fn get_base_url(&self) -> String {
+    pub fn get_base_url(&self) -> String {
         let c8y_host = &self.c8y_host;
         if c8y_host.starts_with("http") {
             c8y_host.to_string()
         } else {
             format!("https://{c8y_host}")
         }
+    }
+
+    pub fn get_proxy_url(&self) -> String {
+        self.proxy.base_url()
     }
 
     pub fn proxy_url_for_internal_id(&self, device_id: &str) -> String {
--- a/crates/core/tedge/src/cli/certificate/c8y/download.rs
+++ b/crates/core/tedge/src/cli/certificate/c8y/download.rs
@@ -11,6 +11,7 @@ use anyhow::Error;
 use c8y_api::json_c8y_deserializer::C8yAPIError;
 use camino::Utf8PathBuf;
 use certificate::CloudRootCerts;
+use certificate::CsrTemplate;
 use hyper::StatusCode;
 use reqwest::header::CONTENT_TYPE;
 use reqwest::Response;
@@ -51,6 +52,9 @@ pub struct DownloadCertCmd {
 
     /// Maximum time waiting for the device to be registered
     pub max_timeout: Duration,
+
+    /// CSR template
+    pub csr_template: CsrTemplate,
 }
 
 #[async_trait::async_trait]
@@ -75,6 +79,7 @@ impl DownloadCertCmd {
                 common_name.clone(),
                 self.key_path.clone(),
                 self.csr_path.clone(),
+                self.csr_template.clone(),
             )
             .await
             .with_context(|| format!("Fail to create the device CSR {}", self.csr_path))?;
--- a/crates/core/tedge/src/cli/certificate/c8y/download.rs
+++ b/crates/core/tedge/src/cli/certificate/c8y/download.rs
@@ -149,7 +149,7 @@ impl DownloadCertCmd {
     ) -> Result<Response, reqwest::Error> {
         http.post(url.clone())
             .basic_auth(username, Some(password))
-            .header(CONTENT_TYPE, "text/plain")
+            .header(CONTENT_TYPE, "application/pkcs10")
             .body(csr.to_string())
             .send()
             .await
--- a/crates/core/tedge/src/cli/certificate/c8y/download.rs
+++ b/crates/core/tedge/src/cli/certificate/c8y/download.rs
@@ -1,4 +1,5 @@
 use crate::cli::certificate::c8y::create_device_csr;
+use crate::cli::certificate::c8y::read_csr_from_file;
 use crate::cli::certificate::c8y::store_device_cert;
 use crate::command::Command;
 use crate::error;
@@ -42,6 +43,9 @@ pub struct DownloadCertCmd {
     /// The path where the device CSR file will be stored
     pub csr_path: Utf8PathBuf,
 
+    /// Tell if the CSR has to be generated or is ready to be used
+    pub generate_csr: bool,
+
     /// Delay between two attempts, polling till the device is registered
     pub retry_every: Duration,
 
@@ -66,13 +70,16 @@ impl Command for DownloadCertCmd {
 impl DownloadCertCmd {
     async fn download_device_certificate(&self) -> Result<(), Error> {
         let (common_name, security_token) = self.get_registration_data()?;
-        let csr = create_device_csr(
-            common_name.clone(),
-            self.key_path.clone(),
-            self.csr_path.clone(),
-        )
-        .await
-        .with_context(|| format!("Fail to create the device CSR {}", self.csr_path))?;
+        if self.generate_csr {
+            create_device_csr(
+                common_name.clone(),
+                self.key_path.clone(),
+                self.csr_path.clone(),
+            )
+            .await
+            .with_context(|| format!("Fail to create the device CSR {}", self.csr_path))?;
+        }
+        let csr = read_csr_from_file(&self.csr_path).await?;
 
         let http = self.root_certs.client();
         let c8y_url = &self.c8y_url;
--- a/crates/core/tedge/src/cli/certificate/c8y/download.rs
+++ b/crates/core/tedge/src/cli/certificate/c8y/download.rs
@@ -1,27 +1,21 @@
-use crate::cli::certificate::create_csr::CreateCsrCmd;
+use crate::cli::certificate::c8y::create_device_csr;
+use crate::cli::certificate::c8y::store_device_cert;
 use crate::command::Command;
 use crate::error;
 use crate::get_webpki_error_from_reqwest;
 use crate::log::MaybeFancy;
-use crate::read_cert_to_string;
 use crate::warning;
-use crate::CertError;
 use anyhow::Context;
 use anyhow::Error;
 use camino::Utf8PathBuf;
 use certificate::CloudRootCerts;
-use certificate::NewCertificateConfig;
 use hyper::StatusCode;
 use reqwest::header::CONTENT_TYPE;
 use reqwest::Response;
-use std::fs::Permissions;
 use std::io::Write;
-use std::os::unix::fs::PermissionsExt;
 use std::time::Duration;
 use tedge_config::models::HostPort;
 use tedge_config::models::HTTPS_PORT;
-use tokio::fs::OpenOptions;
-use tokio::io::AsyncWriteExt;
 use url::Url;
 
 /// Command to request and download a device certificate from Cumulocity
@@ -65,10 +59,13 @@ impl Command for DownloadCertCmd {
 impl DownloadCertCmd {
     async fn download_device_certificate(&self) -> Result<(), Error> {
         let (common_name, security_token) = self.get_registration_data()?;
-        let csr = self
-            .create_device_csr(common_name.clone())
-            .await
-            .with_context(|| format!("Fail to create the device CSR {}", self.csr_path))?;
+        let csr = create_device_csr(
+            common_name.clone(),
+            self.key_path.clone(),
+            self.csr_path.clone(),
+        )
+        .await
+        .with_context(|| format!("Fail to create the device CSR {}", self.csr_path))?;
 
         let http = self.root_certs.client();
         let url = format!("https://{}/.well-known/est/simpleenroll", self.c8y_url);
@@ -81,7 +78,7 @@ impl DownloadCertCmd {
             match result {
                 Ok(response) if response.status() == StatusCode::OK => {
                     if let Ok(cert) = response.text().await {
-                        self.store_device_cert(cert).await?;
+                        store_device_cert(&self.cert_path, cert).await?;
                         return Ok(());
                     }
                     error!(
@@ -107,7 +104,7 @@ impl DownloadCertCmd {
                 }
             }
             warning!("Will retry in 5 seconds");
-            std::thread::sleep(Duration::from_secs(5));
+            tokio::time::sleep(Duration::from_secs(5)).await;
         }
     }
 
@@ -135,22 +132,6 @@ impl DownloadCertCmd {
         Ok((device_id, security_token))
     }
 
-    /// Create the device private key and CSR
-    async fn create_device_csr(&self, common_name: String) -> Result<String, CertError> {
-        let config = NewCertificateConfig::default();
-        let create_cmd = CreateCsrCmd {
-            id: common_name,
-            csr_path: self.csr_path.clone(),
-            key_path: self.key_path.clone(),
-            user: "tedge".to_string(),
-            group: "tedge".to_string(),
-        };
-        create_cmd
-            .create_certificate_signing_request(&config)
-            .await?;
-        read_cert_to_string(&self.csr_path)
-    }
-
     /// Post the device CSR
     async fn post_device_csr(
         &self,
@@ -166,18 +147,5 @@ impl DownloadCertCmd {
             .body(csr.to_string())
             .send()
             .await
-    }
-
-    async fn store_device_cert(&self, cert: String) -> Result<(), CertError> {
-        let mut file = OpenOptions::new()
-            .write(true)
-            .create_new(true)
-            .open(&self.cert_path)
-            .await?;
-
-        file.write_all(cert.as_bytes()).await?;
-        file.sync_all().await?;
-        file.set_permissions(Permissions::from_mode(0o444)).await?;
-        Ok(())
     }
 }
--- a/crates/core/tedge/src/cli/certificate/c8y/download.rs
+++ b/crates/core/tedge/src/cli/certificate/c8y/download.rs
@@ -7,6 +7,7 @@ use crate::log::MaybeFancy;
 use crate::warning;
 use anyhow::Context;
 use anyhow::Error;
+use c8y_api::json_c8y_deserializer::C8yAPIError;
 use camino::Utf8PathBuf;
 use certificate::CloudRootCerts;
 use hyper::StatusCode;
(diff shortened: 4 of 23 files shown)


crates/core/tedge/src/cli/certificate/c8y/renew.rs
```diff
@@ -0,0 +1,116 @@
...
+        let csr = create_device_csr(
+            self.device_id.clone(),
+            self.key_path.clone(),
+            self.csr_path.clone(),
+        )?;
+
+        let jwt_token = self.get_jwt_token();
+
+        let http = self.root_certs.blocking_client();
+        let url = format!("https://{}/.well-known/est/simpleenroll", self.c8y_url);
```
didier-wenzek, line 64: It would be simpler to connect c8y through the local c8y http proxy.

crates/core/tedge/src/cli/certificate/download.rs
```diff
@@ -0,0 +1,170 @@
...
+        let http = self.root_certs.blocking_client();
+        let url = format!("https://{}/.well-known/est/simpleenroll", self.c8y_url);
+        let url = Url::parse(&url)?;
+
+        loop {
+            let result = self.post_device_csr(&http, &url, &common_name, &security_token, &csr);
+            match result {
+                Ok(response) if response.status() == StatusCode::OK => {
+                    if let Ok(cert) = response.text() {
+                        self.store_device_cert(cert)?;
```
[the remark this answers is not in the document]
didier-wenzek: Addressed by https://github.com/thin-edge/thin-edge.io/pull/3259/commits/9ef16591aeac636b312b4fa832b72aea39976e3c

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -44,6 +44,10 @@ pub enum TEdgeCertCli {
 
     /// Renew the device certificate
     Renew {
+        /// CA from which the certificate will be renew
+        #[arg(value_enum, default_value = "self-signed")]
+        ca: CertRenewalCA,
+
         #[clap(subcommand)]
         cloud: Option<CloudArg>,
     },
```
didier-wenzek, line 53: Something is wrong here.

In `tedge cert renew c8y`, `c8y` can be understood either as the CA and as the cloud profile. Even `clap` is confused:

- `tedge cert renew self-signed c8y --profile test` is correctly interpreted as a request to renew the self-signed certificate used to connect the test c8y tenant
- while `tedge cert renew c8y c8y --profile test` leads to an error, `clap` incorrectly assuming the default CA (i.e. self-signed) and being then confused by the repeated cloud parameter.

crates/core/tedge/src/cli/certificate/c8y/mod.rs
```diff
@@ -0,0 +1,175 @@
...
+        // With an exception on the thumbprint
+        // assert_eq!(cert.thumbprint().unwrap(), openssl_cert.thumbprint().unwrap());
+        assert_eq!(
+            cert.thumbprint().unwrap(),
+            "A392A53D3D4263A32C779D06CDDE13A847F272B6".to_string()
+        );
+        assert_eq!(
+            openssl_cert.thumbprint().unwrap(),
+            "9C68C7EC9A860366FB8D2697C53B2543D9EA525C".to_string()
+        );
```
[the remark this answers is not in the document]
reubenmiller: After doing a deep dive into the differences in the certificate I've found exactly where the difference is coming from:


I have a working example with refactored code (using [rasn](https://crates.io/crates/rasn)/[rasm-csm](https://crates.io/crates/rasn-cms) over the higher level [cryptographic-message-syntax](https://crates.io/crates/cryptographic-message-syntax) crate) where the thumbprint is the same as the one computed by openssl. The code can be found [on my temp branch](https://github.com/didier-wenzek/thin-edge.io/compare/feat/integration-with-cumulocity-ca...reubenmiller:thin-edge.io:investigate-cert-encoding).

The problem originates from this part in the [x509-certificate](https://github.com/indygreg/cryptography-rs/blob/c078199a1b86191e19a53993691850769e7748d2/x509-certificate/src/rfc5280.rs#L73) where it does the PEM encoding of the Der formatted value when handing the AlgorithmIdentifier and its handling of the optional parameters. The [cryptographic-message-syntax](https://crates.io/crates/cryptographic-message-syntax) crate suffers from the same problem as it uses the same [x509-certificate](https://github.com/indygreg/cryptography-rs) crate.

The x509-certificate crate purposefully uses the ASN.1 NULL value when an AlgorithmIdentifier object does not have any parameters associated with it.

There is a [comment in the code of x509-certificate](https://github.com/indygreg/cryptography-rs/blob/c078199a1b86191e19a53993691850769e7748d2/x509-certificate/src/rfc5280.rs#L73) explaining the motivation of using a ASN.1 NULL tag for (despite the ASN.1 spec clearly stating that the AlgorithmIdentifier parameter object is OPTIONAL).

```rust
fn encoded_values(&self, mode: Mode) -> impl Values + '_ {
    // parameters is strictly OPTIONAL, which means we can omit it completely.
    // However, it is common to see this field encoded as NULL and some
    // parsers seem to insist the NULL be there or else they refuse to
    // parse the ASN.1. So we ensure this field is always set.
    let captured = if let Some(params) = self.parameters.as_ref() {
        params.clone()
    } else {
        AlgorithmParameter(Captured::from_values(mode, ().encode_as(Tag::NULL)))
    };

    encode::sequence((self.algorithm.clone().encode(), captured))
}
```

Other libraries like openssl, and golang x509 libraries do not use ASN.1 NULL values if the AlgorithmIdentifier Parameters are empty, so this results in the PEM encoded x509 certificates having a different thumbprint when compared to the any Rust code which uses the 509-certificate crate.

Whilst this may not be a bug as the certificate should still be functionally equal, the difference in
thumbprints can give the impression that the cert is functionality different, and maybe this is causing the connection problem to Cumulocity (like you said), but I haven't verified this just yet.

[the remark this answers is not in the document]
reubenmiller: And I've found an easier solution for encoding the x509 cert from DER into the PEM format by using the existing dependency to the **pem** crate. Now no custom forks are needed and it adds very little crate dependencies (in comparison with the **x509-certificate** and **cryptographic-message-syntax** crates.

reubenmiller:
Just some UX notes on the `tedge cert show` command with the additional fields.

* Use `humantime::Duration` for a user friend duration (we're already using it elsewhere in thin-edge.io, plus it is more human friendly, shows more detail (but not too much like milliseconds etc.)
* Use `\t` to separate the field and values (makes it easier for users to distinguish)
* Serial number can show both the integer and hex on the same line, e.g. `1741099918675 (0x19561a46553)`. This saves console space and is still very readable
* Add new explicit "status" field which shows if it is valid or not, e.g. `VALID (expires in: 11months 28days 6h 49m 24s)` (using the humantime this is trivial)
* Colour code the "status" value (e.g. > 3 months validity, then green, ). We're already using `yansi` for this which will automatically switch off when the not writing to the console (standard cli behaviour these days)


Below shows an example of the output with the above changes:

```sh
$ cargo run -- cert show
Certificate:    /opt/homebrew/etc/tedge/device-certs/tedge-certificate.pem
Subject:        CN=rmi_example01, O=Thin Edge, OU=Test Device
Issuer:         C=United States, O=Cumulocity, CN=management
Status:         VALID (expires in: 11months 28days 6h 49m 24s)
Valid from:     Tue, 04 Mar 2025 14:51:58 +0000
Valid until:    Wed, 04 Mar 2026 10:19:07 +0000
Serial number:  1741099918675 (0x19561a46553)
Thumbprint:     88FED0755464C2C30F5D61E3BD49B8EA6288D582
```

And here is a screenshot showing with colours as it is much easier for users (if they're looking at the `tedge cert show` on the console) to be visually alerted of the cert's status.

**Cert VALID (and valid for > 30 days)**

image

**Cert Expiring Soon: < 30 days, > 0 days**
*Note*: the example does not show the 30 day threshold as I just hacked the code to show yellow

image

**Status: EXPIRED  *< 0 days**

image

tests/RobotFramework/tasks/E2E_Tutorials/Connect_to_c8y.robot
```diff
@@ -33,11 +33,23 @@ Create the certificate
 
     # You can then check the content of that certificate.
     ${output}=    Execute Command    sudo tedge cert show    # You can then check the content of that certificate.
```
didier-wenzek, line 35: This check is a duplicate of `tests/RobotFramework/tests/cumulocity/registration/cumulocity_ca.robot` and is not executed by the CI.

- @reubenmiller Are the tests under `tests/RobotFramework/tasks` useful?
- If not it would be better to remove all of them.

crates/core/c8y_api/src/http_proxy.rs
```diff
@@ -68,8 +68,8 @@ impl C8yEndPoint {
         c8y_profile: Option<&str>,
     ) -> Result<Self, C8yEndPointConfigError> {
         let c8y_config = tedge_config.c8y.try_get(c8y_profile)?;
-        let c8y_host = c8y_config.proxy.client.host.to_string();
-        let c8y_mqtt_host = c8y_host.clone();
+        let c8y_host = c8y_config.http.or_config_not_set()?.to_string();
+        let c8y_mqtt_host = c8y_config.mqtt.or_config_not_set()?.to_string();
```
didier-wenzek, line 72: I will revert that, as this is unrelated to this PR and makes a system working.

These values are only used for error messages, but setting the proper value here (i.e. the tenant host and not the proxy host) forces the child devices to properly configure c8y.url which is useless (this is why test of `tedge upload c8y` from a child device fails).

crates/core/tedge/src/cli/certificate/cli.rs
didier-wenzek: This has been addressed by https://github.com/thin-edge/thin-edge.io/pull/3259/commits/3e102ff48cf33f2dc85cdca8ed75edd10e9f355d

tests/RobotFramework/tasks/E2E_Tutorials/Connect_to_c8y.robot
reubenmiller: Sorry I missed this question. I've approached your PR (https://github.com/thin-edge/thin-edge.io/pull/3495) to remove these unused tasks. Good suggestion!

reubenmiller:
### Cert renewal

To start out, the certificate renewal will be facilitated by an external process (default using system timer service) to actively renew the certificate.

In order to make it easy for external processes to trigger a certificate renewal, there needs to be a way a process can check if the certificate is about to expire (with the minimum value as defined in `certificate.validity.minimum_duration` in the tedge.toml).

The command used to check the validity of the certificate should produce the following exit code, so that the calling process can respond accordingly. There are several options where the above certificate validity can be placed, below are some options.

Inspiration for this type of renewal is taken from [smallstep](https://smallstep.com/docs/step-ca/renewal/#automated-renewal)

#### Option 1: New subcommand to check if the cert needs renewal or not

```sh
tedge cert needs-renewal c8y
```

Exit codes:

* `0` - certificate needs renewal as it is no longer valid, or it will expire within the duration, `certificate.validity.minimum_duration`
* `1` - certificate is still valid and does not need renewal
* `2` - unexpected error (e.g. certificate does not exist, or can't be read)


#### Option 2: Flag in renew command

```sh
tedge cert renew c8y --needs-renewal
```

Exit codes:

* `0` - certificate needs renewal as it is no longer valid, or it will expire within the duration, `certificate.validity.minimum_duration`
* `1` - certificate is still valid and does not need renewal
* `2` - unexpected error (e.g. certificate does not exist, or can't be read)

didier-wenzek:
> Exit codes:
> * `0` - certificate needs renewal as it is no longer valid, or it will expire within the duration, `certificate.validity.minimum_duration`
> * `1` - certificate is still valid and does not need renewal
> * `2` - unexpected error (e.g. certificate does not exist, or can't be read)
 
I would return a non `0` value when the certificate is *already* expired - because it can no more be renew then. But maybe, this is a design choice to proceed in such in a case (maybe device time is not correctly set and the certificate is actually still valid). Is it?

reubenmiller:
> > Exit codes:
> > 
> > * `0` - certificate needs renewal as it is no longer valid, or it will expire within the duration, `certificate.validity.minimum_duration`
> > * `1` - certificate is still valid and does not need renewal
> > * `2` - unexpected error (e.g. certificate does not exist, or can't be read)
> 
> I would return a non `0` value when the certificate is _already_ expired - because it can no more be renew then. But maybe, this is a design choice to proceed in such in a case (maybe device time is not correctly set and the certificate is actually still valid). Is it?

Yes basically. The renewal might not work, but at least it will be attempted.

didier-wenzek:
> ### Cert renewal

> #### Option 1: New subcommand to check if the cert needs renewal or not


Implemented with https://github.com/thin-edge/thin-edge.io/pull/3259/commits/ea03cc518cb2caded2e55076016dbeb0939f9472

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -149,14 +214,95 @@ impl BuildCommand for TEdgeCertCli {
...
+                // - if the certificate is self signed => create a new self-signed certificate
+                // - if not => assume that the device tenant and its CA tenant are the same.
+                let is_self_signed = match certificate_is_self_signed(&cert_path) {
+                    Ok(is_self_signed) => is_self_signed,
+                    Err(err) => {
+                        return Err(anyhow!("Cannot renew certificate {cert_path}: {err}").into())
+                    }
+                };
+
+                if is_self_signed {
```
reubenmiller, line 269: It would be useful if we allow users to migrate from a self-signed certificate to a certificate issued by the Cumulocity CA, as this is probably the position that a lot of thin-edge.io c8y users are currently in.

So it would be useful to be able to control whether we're re-creating a new self-signed certificate, or doing the default action to request a new certificate via the new Cumulocity CA service.

didier-wenzek: Good point. Indeed, directly asking c8y CA for a renewal should work in such a case despite the current certificate having not been issued by v8y.

didier-wenzek: I see two options, my preference being for the second one.

__implicit__:

When the certificate is used to connect c8y, try first to post the renewal request to c8y CA and fallback to a self-signed certificate when the response is a 404 __and__ the current certificate is self-signed.
- Warn the user when the new certificate is self-signed: this certificate has to be uploaded on the tenant to be trusted.
- Pros: No new command line options
- Cons: the command seems simpler to use, but it is not, as a second critical step might be required.

__explicit__:

Add a `--self-signed` option (which default is false).
- When `--self-signed` is false and the endpoint is c8y, then request the renewal to c8y CA. Let the current certificate unchanged if this fails.
- Return an error if `--self-signed` is true but the certificate is not self-signed.
- Return an error if `--self-signed` is false and the endpoint is not c8y.
- Cons: This changes how `tedge cert renew` behaves
- Pros: Make self-signed certificates second-class citizens

didier-wenzek: > It would be useful if we allow users to migrate from a self-signed certificate to a certificate issued by the Cumulocity CA, as this is probably the position that a lot of thin-edge.io c8y users are currently in.

Similarly, we can support users moving away from device passwords.

> So it would be useful to be able to control whether we're re-creating a new self-signed certificate, or doing the default action to request a new certificate via the new Cumulocity CA service.

And, along the same lines, it would be useful to improve `tedge cert download` or `tedge cert renew`  so a very first certificate can be retrieved using the renewal API and authenticating via JWT.

rina23q (commented):
The changes are very clear and easy to read. I was not able to test with c8y fully since my tenant is missing something for now. I'll test again once my tenant issue is solved.

crates/core/tedge/src/cli/certificate/show.rs
```diff
@@ -16,39 +19,135 @@ macro_rules! print_async {
...
+            "check device validity".into()
+        } else {
+            "show the device certificate".into()
+        }
     }
 
     async fn execute(&self) -> Result<(), MaybeFancy<anyhow::Error>> {
-        self.show_certificate().await?;
-        Ok(())
+        if self.validity_check_only {
```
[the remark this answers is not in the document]
didier-wenzek: I opted for this code organization because both commands share  intent and code a lot, the difference being a more complete output for `show` and more scriptable exit status for `needs-renew`.

crates/core/tedge/src/cli/certificate/c8y/download.rs
```diff
@@ -0,0 +1,178 @@
...
+                Ok(response) if response.status() == StatusCode::OK => {
+                    if let Ok(cert) = response.text().await {
+                        store_device_cert(&self.cert_path, cert).await?;
+                        return Ok(());
+                    }
+                    error!("Fail to extract a certificate from the response returned by {c8y_url}");
+                }
+                Ok(response) => {
+                    let error = Self::c8y_error_message(response).await;
+                    error!("The device {common_name} is not registered yet on {c8y_url}: {error}");
```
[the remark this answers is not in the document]
didier-wenzek: Exactly. The idea is to be able to design on-boarding boot sequences along the following lines: 

- Derive from hardware a unique device name
- Generate a random secret
- Display these registration data over some user-friendly channel (say a QR code)
- Wait for the registration to be validated by the user on the tenant.

rina23q (approved):
It works like a charm, really nice feature! Now it's super easy to register a device in c8y!

reubenmiller (approved):
Approved (from a functional level). It has all of the fundamental elements to build a renewal service on top.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -47,12 +49,55 @@ pub enum TEdgeCertCli {
...
+        /// configured for the given cloud profile.
+        #[clap(long = "csr-path", global = true, value_hint = ValueHint::FilePath)]
+        csr_path: Option<Utf8PathBuf>,
+
+        /// Force the renewal of self-signed certificates as self-signed
+        ///
+        /// This can be used to bypass the default behavior
+        /// which is to forward the renewal request to the cloud CA
+        /// even if the current certificate has not been signed by this CA.
+        /// In most case, the default behavior is what you want:
```
Bravo555, line 67: typo:

```suggestion
        /// In most cases, the default behavior is what you want:
```

Bravo555 (commented):
As for the code itself, LGTM, only managed to find a single typo. But are documentation additions, like a reference or a user guide planned in this PR or in a follow-up?

Cargo.toml
```diff
@@ -153,6 +153,8 @@ proc-macro2 = "1"
 proptest = "1.0"
 quote = "1"
 rand = "0.8"
+rasn = "0.18.0"
```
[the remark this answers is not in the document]
rina23q: Regarding crate versions, long time ago, we decided to specify until minor version (like `0.18`), but I see now our Cargo.toml is mixed of `major.minor.patch` and `major.minor`. I'm also curious why here is 3 digits.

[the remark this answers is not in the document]
didier-wenzek: Because not compatible with our current MSRV.

Added a comment and removed the patch suffix: https://github.com/thin-edge/thin-edge.io/pull/3259/commits/e777d3e9cd06401e4aa0ecf8e4a72f45804c6ded

crates/core/tedge/src/cli/certificate/cli.rs
didier-wenzek: fixed https://github.com/thin-edge/thin-edge.io/pull/3259/commits/bf2f7481150e286b2f0830e648b9c2ea430b1948

Bravo555 (approved):

albinsuresh (approved):
LGTM.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -166,6 +166,28 @@ define_tedge_config! {
...
     },
 
+    certificate: {
+        validity: {
+            /// Requested validity duration for a new certificate
+            #[tedge_config(note = "The CA might return certificates valid for period shorter than requested")]
+            #[tedge_config(example = "365d", default(from_str = "365d"))]
+            requested_duration: SecondsOrHumanTime,
+
+            /// Minimum validity duration bellow which a new certificate should be requested
```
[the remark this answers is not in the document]
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3259/commits/ff6c58f62220cb4d0c7edc03d17570e82a526bc0

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -209,6 +377,72 @@ fn get_device_id(
...
 
+#[derive(clap::Subcommand, Debug)]
+pub enum DownloadCertCli {
+    #[clap(verbatim_doc_comment)]
+    /// Request and download the device certificate from Cumulocity
+    ///
+    /// - Generate a private key and Signing Certificate Request (CSR) for the device
+    /// - Upload this CSR on Cumulocity, using the provided device identifier and security token
+    /// - Loop till the device is registered by an administrator and the CSR accepted
+    /// - Store the certificate created by Cumulocity
```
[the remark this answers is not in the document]
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3259/commits/578a03301e6f1dcd04c80d30922959a2a6770c7a

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -209,6 +377,72 @@ fn get_device_id(
     }
 }
 
+#[derive(clap::Subcommand, Debug)]
+pub enum DownloadCertCli {
+    #[clap(verbatim_doc_comment)]
+    /// Request and download the device certificate from Cumulocity
+    ///
+    /// - Generate a private key and Signing Certificate Request (CSR) for the device
```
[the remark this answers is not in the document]
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3259/commits/578a03301e6f1dcd04c80d30922959a2a6770c7a

crates/core/tedge/src/cli/certificate/c8y/download.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    async fn execute(&self) -> Result<(), MaybeFancy<Error>> {
+        Ok(self.download_device_certificate().await?)
+    }
+}
+
+impl DownloadCertCmd {
+    async fn download_device_certificate(&self) -> Result<(), Error> {
+        let (common_name, security_token) = self.get_registration_data()?;
```
jarhodes314, line 76: This should probably use `spawn_blocking`

didier-wenzek: This has been addressed by https://github.com/thin-edge/thin-edge.io/pull/3525/commits/ec2bf687f3ded1a07585508a7bd7b838cc21a5ab

Issue #3256, opened 2024-11-21
refactor: `Operations` struct refactor


Bravo555:
## Is your refactoring request related to a problem? Please describe.

As mentioned in https://github.com/thin-edge/thin-edge.io/pull/3236#discussion_r1843488529, current design of `Operations` struct and `c8y_api::operations` module presents an API that is a bit too verbose, exposes too much information to `CumulocityConverter` and its implementation doesn't fully account for how we use operations.

As such, this refactor aims to extract some things from `CumulocityConverter` into new `Operations` struct while providing it with a more simplified interface.  

## Describe the solution you'd like

The main aims of the refactor are:

### Phase 1

- [ ] extract all fs operations and path building from `CumulocityConverter` into new operations module
- [ ] Keep track of operations for all devices inside of a single `Operations` struct, to eliminate complicated "find correct `Operations` struct for each device" logic
- [ ] Use correct data structures and expose a sane and simple API to `CumulocityConverter`
- [ ] Simplify the aggregation and loading of multiple `Operation` objects into a single `Operations` object

### Phase 2

- [ ] Decouple "loading operation from fs" from "do substitution on topics and payloads" and "match operation and start correct command when MQTT message comes in"
- [ ] Hide the operation/template distinction from the `CumulocityConverter`
- [ ] Move "loading operations from fs" from `c8y_api` into `c8y_mapper_ext`

## Describe alternatives you've considered

None, will update if any.

PR #3260, opened 2024-11-22
refactor: Extract operation implementation details from `CumulocityConverter`

Bravo555:
## TODO

- [x] cleanup

## Proposed changes

This PR introduces an `SupportedOperations` struct that is used by `CumulocityConverter` to load and save operations.

It achieves following things:

- **Extract all fs operations and path building from `CumulocityConverter` into new operations module**: The main `CumulocityConverter` doesn't create any files or symlinks, or doesn't need to create paths, the logic of mapping supported operations to the filesystem is contained in its own module
- **No main device/child device branching in `CumulocityConverter`**: instead of having separate `operations: Operations` and `children: HashMap<Child, Operations>`, `SupportedOperation` handles this distinction, the converter just passes an external id
- Use `BTreeMap` for to have only one operation under a given name and to ensure consistent ordering

## Types of changes

## Paste Link to the issue
- #3256

## Checklist

### DIFF
--- a/crates/core/c8y_api/src/smartrest/operations.rs
+++ b/crates/core/c8y_api/src/smartrest/operations.rs
@@ -1,6 +1,26 @@
+//! [Supported Operations API]
+//! (https://thin-edge.github.io/thin-edge.io/operate/c8y/supported-operations/#supported-operations-api).
+//!
+//! This module should encompass loading and saving supported operations from and to the filesystem.
+//!
+//! This module works with a c8y operations directory, which by default is `/etc/tedge/operations/c8y`. This directory
+//! contains operation files for the main device, directories that contain operation files for child devices and
+//! operation templates for all devices.
+//!
+//! The names of files in these directories should be the same as names of supported c8y operations, for example
+//! `c8y_Restart`.
+//!
+//! For known c8y operations, the file can be empty. For custom c8y operations, the file should contain a
+//! [`OnMessageExec`] section describing how c8y mapper should convert this c8y operation to a local thin-edge command.
+
 use crate::json_c8y_deserializer::C8yDeviceControlTopic;
 use crate::smartrest::error::OperationsError;
 use crate::smartrest::smartrest_serializer::declare_supported_operations;
+use anyhow::ensure;
+use anyhow::Context;
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+use mqtt_channel::MqttMessage;
 use mqtt_channel::TopicFilter;
 use serde::Deserialize;
 use serde::Deserializer;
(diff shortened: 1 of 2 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1641,15 +1641,11 @@ impl CumulocityConverter {
                     .operations_for_device_mut(target)
                     .expect("entity should've been checked before that it's not a service");
 
-                let prev_operation = current_operations.remove_operation(&operation.name);
+                let prev_operation = current_operations.insert_operation(operation);
```
didier-wenzek: This change seems weird at first ... but is correct!

Previously any operation with the same name was removed before adding a new version of the same operation.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -10,18 +29,326 @@ use std::collections::HashSet;
...
+
 type OperationName = String;
+type OperationNameRef = str;
+
+/// Used to hold and query supported operations for all devices.
+pub struct SupportedOperations {
+    /// External ID of the main device.
+    ///
+    /// Required because when given an external id of the main device when creating an operation, we want to create it
+    /// in a main directory, instead of a subdirectory.
```
didier-wenzek, line 57: Okay. It makes sense for this PR focus on code refactoring without any user-visible changes, but this asymmetry between the main device and the child devices makes things more complicated than they should.

crates/core/c8y_api/src/smartrest/operations.rs
```diff
@@ -10,18 +29,326 @@ use std::collections::HashSet;
...
+        c8y_operation_name: &OperationNameRef,
+    ) -> Result<(), anyhow::Error> {
+        let ops_file = if device_xid == self.device_id {
+            self.base_ops_dir.join(c8y_operation_name)
+        } else {
+            self.base_ops_dir.join(device_xid).join(c8y_operation_name)
+        };
+
+        // Create directory for a device if it doesn't exist yet
+        file::create_directory_with_defaults(ops_file.parent().expect("should never fail"))?;
```
didier-wenzek, line 86: Assuming that this will never fail seems a bit too bold here.

crates/core/c8y_api/src/smartrest/operations.rs
Bravo555: I don't fully get what you mean. Do you mean this PR specifically, or the codebase more generally?

One maybe unnecessary change that I did was instead of having "operations for the main device" in one field and "operations for child devices" in a hashmap, i put it all in a single hashmap, so that `CumulocityConverter` only passes along the external id and doesn't have to care if a device is a child device or a main device. In retrospect the functions can still take an external id but the operations for the main device can be made a normal field instead of a hashmap entry, since "operations for the main device" is something that always exists and putting it in a hashmap where it can be removed introduces an extra failure mode.

crates/core/c8y_api/src/smartrest/operations.rs
Bravo555: Above, we add at least 1 component to the `base_ops_dir` path to create `ops_file`, so I'm pretty sure taking the `.parent()` here never fails.

crates/core/c8y_api/src/smartrest/operations.rs
didier-wenzek: My point is in no case to disagree with this PR which is a real improvement reducing the accidental complexity of the c8y mapper.

My comment is coming from the fact that you (correctly) feel the need to add a comment on why we need the main device external id here (while we don't for the child devices).

crates/core/c8y_api/src/smartrest/operations.rs
didier-wenzek: You are correct.

didier-wenzek (approved):
Approved. This is nice clarification of the code base.

albinsuresh (approved):
LGTM.

PR #3265, opened 2024-11-26
dev: publish devcontainer image

reubenmiller:
## Proposed changes

Publish the dev container to Github container registry to reduce the time it takes to install all of the dependencies to start developing. In addition it makes the dev environment more reliable as apt-get install performance can be unreliable at times as the automatic mirror selection does not always choose a close-by debian mirror.

A workflow has been added to also periodically publish the devcontainer image (or on demand).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -7,7 +7,7 @@ FROM mcr.microsoft.com/devcontainers/rust:1-${VARIANT}
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get -y install --no-install-recommends \
         # apt dependencies
-        apt-transport-https ca-certificates gnupg-agent software-properties-common lsb-release \
+        apt-transport-https ca-certificates gnupg-agent software-properties-common \
         # Debian packaging utilities
         equivs \
         dpkg-dev \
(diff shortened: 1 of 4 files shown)


.github/workflows/devcontainer.yml
```diff
@@ -0,0 +1,34 @@
+name: devcontainer
+
+on:
+  # trigger manually
+  workflow_dispatch:
+  # run automatically every Monday at 01:00 UTC
+  schedule:
+  - cron: "0 1 * * 1"
```
albinsuresh, line 8:
```suggestion
on:
  push:
    paths:
      - '.devcontainer/**'
```
How about publishing updated images only when any one of the devcontainer files change?

.devcontainer/Dockerfile
```diff
@@ -35,61 +35,61 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
...
+    #
     # Remove policy-rc.d file which prevents services from starting
-    && rm -f /usr/sbin/policy-rc.d
-
-# Install gh utility
-RUN curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
+    && rm -f /usr/sbin/policy-rc.d \
+    #
+    # Install gh utility
+    && curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg | sudo dd of=/usr/share/keyrings/githubcli-archive-keyring.gpg \
```
albinsuresh, line 62: What is the benefit of chaining of all these commands one after the other, as opposed to individual `RUN` statements? To avoid caching all the intermediary layers?

Wouldn't it make the debugging harder when any one of those steps fail, as we'd struggle to narrow-down which command in the chain failed?

.github/workflows/devcontainer.yml
reubenmiller: Good idea. We can try it, but we still need the cronjob like thing to update when the base rust image changes

.devcontainer/Dockerfile
reubenmiller: Yes, using a single run reduces the number of layers (`RUN` creates a new layer). Debugging still works as you still get an error when something is wrong, so it is not that difficult (all the other public docker images do it anyway ;)).

.github/workflows/devcontainer.yml
reubenmiller: I've added the suggestion to also build on changes to the devcontainer folder. Commit b0dd53ccbb368eac19a33ba8c1365144c3c5b7c1

albinsuresh (approved):
LGTM

PR #3271, opened 2024-11-27
Add `--log-level` to `tedge-write` and `tedge-apt-plugin`

Bravo555:
## Proposed changes

Add `--log-level` flag to `tedge-write` and `tedge-apt-plugin`.

## Types of changes

## Paste Link to the issue
- #2089

## Checklist

### DIFF
--- a/crates/core/tedge_write/src/bin.rs
+++ b/crates/core/tedge_write/src/bin.rs
@@ -4,6 +4,8 @@ use anyhow::bail;
 use anyhow::Context;
 use camino::Utf8PathBuf;
 use clap::Parser;
+use tedge_config::system_services::log_init;
+use tedge_config::system_services::LogConfigArgs;
 use tedge_utils::atomic::MaybePermissions;
 
 /// tee-like helper for writing to files which `tedge` user does not have write permissions to.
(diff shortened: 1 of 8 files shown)


crates/core/tedge_write/src/bin.rs
```diff
@@ -30,9 +32,23 @@ pub struct Args {
...
+    log_args: LogConfigArgs,
+
+    /// [env: TEDGE_CONFIG_DIR, default: /etc/tedge]
+    #[clap(
+        long = "config-dir",
+        default_value = tedge_config::get_config_dir().into_os_string(),
+        hide_env_values = true,
+        hide_default_value = true,
+    )]
+    config_dir: Utf8PathBuf,
```
albinsuresh, line 46: As @reubenmiller proposed in on of the earlier PRs, wouldn't it make sense to capture these two arguments into a common struct and use a flattened reference to it in all these concrete structs like `MapperOpt`, AgentOpt`?

Bravo555: We could put `config_dir` into the common struct as well, but because it's a multicall binary we'd still have to include it in `Opt` struct for each component and can't really put it in a single place and make all the other `Opts` have references to it. They have to own it directly.

albinsuresh: I think I may have wrongly used the word "reference" here, which caused that confusion. What I meant was to have a flattened field, something like:

```
    #[command(flatten)]
    common_args: CommonArgs,
```

and not really a "reference" to a single instance of that struct, which probably would unnecessarily complicated the code.

albinsuresh (approved):
As discussed, extracting the common args into a common struct and using it at one place to enable logging for all components would be attempted in a follow-up PR.

Issue #1315, opened 2022-08-02
Allow upload of files to the c8y binary storage via tedge cli


mbay-ODW:
**Is your feature improvement request related to a problem? Please describe.**
The configuration/log plugin has implemented a logic to upload files to the binary storage of Cumulocity,
However if I develop an own plugin and need to upload files to the binary storage of c8y, I currently can not do so without implementing own Rest API logic into my plugin, using the jet token etc. .

**Describe the solution you'd like**
Since file upload logic is already available somehow in the code I would basically wish something on the cli such as:

```bash
tedge file upload c8y --file /test/filename.txt
```

For the cli I used the structure  of the cert upload. I would basically expect to get back the binaryID of the uploaded file.

didier-wenzek:
A more generic option would be to have a [local HTTP proxy](https://github.com/thin-edge/thin-edge.io/discussions/915) to the cloud endpoint, so the full REST API of Cumulocity would be available locally.

mbay-ODW:
Any feedback on this here? Got the same request here now:

https://tech.forums.softwareag.com/t/upload-a-base64-picture-string-via-event-message-thinedge-to-cumulocity/265434/7

didier-wenzek:
With the new local c8y auth proxy, we are now in good position to ease file upload to c8y.

@reubenmiller can you take some time to define a solution?

didier-wenzek:
The proposal is to extend the `tedge` cli with a `tedge c8y upload` command,
the plan being to add new Cumulocity-specific commands overtime.

```sh
$ tedge c8y upload --help

Upload a file to Cumulocity

The command create a new event for the device and attach the given file content to this new event.

Usage: tedge c8y upload

Options:
      --text <STRING>
          Text description of the event. Defaults to the event type.

      --time <DATE>
          Time of the event. Defaults to current timestamp.

      --type <STRING>
          Type of the event.
                  
      --json <JSON>
          JSON content attached to the event. Optional.
            
      --file <PATH>
          Path to the uploaded file.

      --mime-type <STRING>
          MIME type of the file content. Defaults to `application/octet-stream`
```


Note: The initial proposal was to add support for a `tedge file upload c8y` command.
However, the need is more to simplify the interactions with the [Cumulocity event API](https://cumulocity.com/api/core/#tag/Attachments)
rather than to upload files to various targets. Hence, the decision to shift the focus toward a `tedge c8y upload` command.

reubenmiller:
[...]

PR #3262, opened 2024-11-25
feat: add profile support to the `tedge` cli and fix mosquitto configuration

jarhodes314:
## Proposed changes
This changes `tedge connect` etc. to use the `<cloud>@<profile>` syntax. For example,

```
tedge connect c8y --profile edge
```

becomes

```
tedge connect c8y@edge
```

It also fixes some issues that were causing mosquitto to fail to start when multiple cloud profiles are configured.

Additionally it adds the missing systemd service definitions, so you can now start `tedge-mapper-c8y@profile` to spawn a mapper instance with a profile selected.

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
<br/>

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #3274, opened 2024-11-29
feat: tedge upload c8y

didier-wenzek:
## Proposed changes

Extend `tedge` cli with a `tedge upload c8y` command to upload a file to Cumulocity.

```sh
Upload a file to Cumulocity

The command create a new event for the device,
attach the given file content to this new event,
and return the event ID.

Usage: tedge upload c8y [OPTIONS] --file <FILE>

Options:
      --file <FILE>
          Path to the uploaded file

      --mime-type <MIME_TYPE>
          MIME type of the file content
          
          [default: application/octet-stream]

      --type <EVENT_TYPE>
          Type of the event
          
          [default: tedge_UploadedFile]

      --text <TEXT>
          Text description of the event. . Defaults to "Uploaded file: <FILE>"

      --json <JSON>
          JSON fragment attached to the event
          
          [default: {}]
```

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/1315

## Checklist


crates/core/tedge/src/cli/c8y/upload.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        println!("{event_id}");
+        Ok(())
+    }
+}
+
+impl C8yUpload {
+    pub fn get_internal_id(&self) -> Result<String, Error> {
+        let url_get_id: String = self.c8y.proxy_url_for_internal_id(&self.device_id);
+        let http_result = blocking::Client::new().get(url_get_id).send()?;
```
didier-wenzek, line 57: As stressed by clippy, I have to use `certificate::CloudRootCerts` type instead to take root_cert_path configurations into account, i.e. to make tedge c8y upload works when the c8y proxy is configured with mTLS.

crates/core/tedge/src/cli/mod.rs
```diff
@@ -98,6 +99,10 @@ pub enum TEdgeOpt {
         relative_links: bool,
     },
 
+    /// Interact with Cumulocity
+    #[clap(subcommand)]
+    C8y(c8y::C8yCmd),
```
albinsuresh, line 104: Wondering if having this concrete enum variant named `C8y` would make it difficult to support profiles, since #3262 is merged and the `c8y@profile` style is preferred over `--profile` arguments.

albinsuresh (commented):
The code looks as straight forward as it can be.

crates/core/tedge/src/cli/mod.rs
didier-wenzek: Good point. I will rebase now to assess what's the actual issues.

didier-wenzek:
This commit https://github.com/thin-edge/thin-edge.io/pull/3274/commits/dd20201e08d53ddf04a2a5509aa39f22e787e687 raises several questions:

- To upload a file to c8y a child device needs to know its external id. How can this be done. The current implementation is a hack.
- The C8yEndPoint struct is working correctly only on the main  device (asking for instance for a c8y.url). Again the current implementation is a workaround.

didier-wenzek:
> > * To upload a file to c8y a child device needs to know its external id. How can this be done. The current implementation is a hack.
> 
> we could limit this feature to the main device until the entity api is available…or allow the use to manually provide the external id if needed

An optional `--device-id` parameter has been added. See https://github.com/thin-edge/thin-edge.io/pull/3274/commits/3a9a526638c356eba69ecf8b4287ecd64c064b65.

- If not provided: the file is attached to the main device.
- To upload a file from a child device, the user must provide the external id of that child device.

didier-wenzek:
>     * The C8yEndPoint struct is working correctly only on the main  device (asking for instance for a c8y.url). Again the current implementation is a workaround.

Fixed by https://github.com/thin-edge/thin-edge.io/pull/3274/commits/3fada8db26e35e1633ab7e86d9ed3dc8371d1d7b

reubenmiller:
@didier-wenzek I just tested the feature out and overall it works nicely, just one slight error scenario.

### Case: Using an invalid mime-type results in an event being created but with no file

If the user is not so versed in mime-types (e.g. they use an invalid mime-type value), then the event is created in Cumulocity, but the binary is not uploaded as the mime-type validation fails (see below)

```sh
tedge upload c8y --text "hello" --file /etc/tedge/tedge.toml --json '{"other":true}' --mime-type invalid
```

*Output*

```sh
Error: failed to upload a file to Cumulocity

Caused by:
    0: builder error: mime parse error: a slash (/) was missing between the type and subtype
    1: mime parse error: a slash (/) was missing between the type and subtype
```

Ideally it would be good to validate the mime-type before creating the event as it would prevent creating an event without an attachment.

crates/core/tedge/src/cli/upload/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+        /// JSON fragment attached to the event
+        #[clap(long, default_value = "{}")]
+        #[arg(value_parser = parse_json)]
+        json: HashMap<String, serde_json::Value>,
+
+        /// Optional c8y cloud profile
+        #[clap(long)]
+        profile: Option<ProfileName>,
+
+        /// Cumulocity external id of the device child on which the file has to be attached.
```
reubenmiller, line 45: The upload could be attached to any device/service which is owned by the main device.

```suggestion
        /// Cumulocity external id of the device/service on which the file has to be attached.
```

didier-wenzek:
> Ideally it would be good to validate the mime-type before creating the event as it would prevent creating an event without an attachment.

Done with https://github.com/thin-edge/thin-edge.io/pull/3274/commits/9b73e34e5c5b07efd87d352f07778d70df63bfff

crates/core/tedge/src/cli/upload/mod.rs
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/3274/commits/cb54862459c13762dfb42f1b5924585a1080f019

crates/core/tedge/src/cli/upload/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+
+mod c8y;
+
+#[derive(clap::Subcommand, Debug)]
+pub enum UploadCmd {
+    /// Upload a file to Cumulocity
+    ///
+    /// The command create a new event for the device,
+    /// attach the given file content to this new event,
+    /// and return the event ID.
```
albinsuresh, line 18:
```suggestion
    /// The command creates a new event for the device,
    /// attaches the given file content to this new event,
    /// and returns the event ID.
```

crates/core/tedge/src/cli/upload/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+    C8y {
+        /// Path to the uploaded file
+        #[clap(long)]
+        file: PathBuf,
```
albinsuresh, line 22: Even though that file path is mandatory argument, I'm assuming that it is accepted as an option rather than a direct argument of the `c8y` subcommand to expand the upload capability to enable uploading to other C8y repositories like software repo, firmware repo etc with future types like `--software`, `--firmware` etc.

albinsuresh (approved):
LGTM.

albinsuresh:
While testing this further, I noticed one minor niggle: If you try to upload a non-existent file, the event is still created without the attachment. So, pre-validating if the file exits and readable might be good.

crates/core/tedge/src/cli/upload/mod.rs
didier-wenzek: fixed by https://github.com/thin-edge/thin-edge.io/pull/3274/commits/af27c794b21cee5d55e904fa4234d366116b5df1

didier-wenzek:
> While testing this further, I noticed one minor niggle: If you try to upload a non-existent file, the event is still created without the attachment. So, pre-validating if the file exits and readable might be good.

Fixed by https://github.com/thin-edge/thin-edge.io/pull/3274/commits/5e456b0e314de5a651e51965a692a1f25f0e2259

crates/core/tedge/src/cli/upload/mod.rs
reubenmiller: > Since the file path is mandatory argument, I thought it would be taken as a direct argument of the c8y subcommand. But, I'm assuming that it is accepted as a --file option rather than a direct argument to expand the upload capability to enable uploading to other C8y repositories like software repo, firmware repo etc with future options like --software, --firmware etc.

I'm pretty sure this is even possible to implement as the device user has restricted permissions. Generally device users can only change data that they own, so uploading to the software/firmware repo won't work successfully as it needs to edit an existing managed object by adding a relationship to the uploaded binary (which is also stored as a binary managed object).

I would avoid supporting uploading directly to the software/firmware repo anyway, as these binaries are stored as [managed object binaries](https://cumulocity.com/api/core/#operation/postBinariesCollectionResource) which aren't cleaned up by the Cumulocity data retention feature. Plus device's shouldn't be responsible for publishing this data directly, as such a binary needs to be validated by a user (or process) before promoting it from one device to other device (imagine if a device uploads a software binary which is in armhf CPU architecture, and another device which has arm64 tries to install it).

reubenmiller (approved):
Approved. Nice addition and we can close one of the more longer standing tickets

PR #3277, opened 2024-12-02
refactor: Extract `supported_operations` module from `c8y_api`

Bravo555:
## Proposed changes

`c8y_api::smartrest::operations` was moved to `c8y_mapper_ext::supported_operations` because:

- it deals with loading operation files from the local filesystem, not `c8y_api` concern
- it was inside `smartrest` module, while not having much to do with smartrest (ie. it only generated "114 Set supported operations" message)

Additionally the `supported_operations` module was divided into 2 parts:

- `supported_operations`: loading, updating, managing the supported operation set, matching on this set, generating messages to set, etc. basically managing the collection and interacting with `CumulocityConverter`
- new `supported_operations::operation`: contains the functions and types related to reading and parsing single operation files from the filesystem and field substitution

## Types of changes

## Paste Link to the issue
- #3256

## Checklist


didier-wenzek (commented):
I'm okay with this refactoring, except that I would have preferred a `git mv crates/core/c8y_api/src/smartrest/operations.rs ...` rather than a delete and two new files.

Bravo555:
> I'm okay with this refactoring, except that I would have preferred a `git mv crates/core/c8y_api/src/smartrest/operations.rs ...` rather than a delete and two new files.

It was necessary to make some changes in `supported_operations.rs` in the first commit because of changes in the imports section (e.g. `c8y_mapper_ext` doesn't depend on `mqtt_channel`, but on `tedge_mqtt_ext`, etc.)

Despite that, when the first commit is viewed separately, it shows as a rename, so old and new is side by side and you can compare that way.

But I can remove the 2nd commit and make it into a separate PR if you'd like.

didier-wenzek:
> But I can remove the 2nd commit and make it into a separate PR if you'd like.

No, that's okay. The move in the first commit is what I was looking for.

didier-wenzek (approved):
Approved

crates/extensions/c8y_mapper_ext/src/supported_operations.rs
```diff
@@ -0,0 +1,501 @@
+//! [Supported Operations API]
```
albinsuresh, line 1: I just find the placement of this module a bit weird, as it is placed directly under `operations` whereas `operations.rs` is directly placed under `supported_operations`. May be you want to move it into a `mod.rs` under `supported_operations` or place both directly under `operations`?

albinsuresh (commented):
LGTM, except for the odd module placement.

crates/extensions/c8y_mapper_ext/src/supported_operations.rs
Bravo555: This style of module placement was supported for some time, and I think it's better to use than the older style for 2 reasons:

- if you split a submodule off a bigger module, you don't have to move the other file and rename it to `mod.rs`, which makes the diffs a bit smaller (bigger pro IMO)
- the file name is still a module name, whereas with `mod.rs` for the module name you have to look at the name of the parent directory, but as the editor tab contains only a file name, you can end up with many `mod.rs` files, so it's a bit harder to keep track of (actually in vscode when two files have the same name then it will display a parent in the tab for context)

But as we're already using the older style (we have 38 `mod.rs` files) I will use this style for consistency.

PR #3283, opened 2024-12-05
Custom operation handler should parse extra arguments for 503/506 messages

rina23q:
### Issues
This PR solves two issues.
1. The new custom operation handlers with workflow execution didn't send the third or more argument(s) of SmartREST `503`/`506` messages (successful operation status update). This issue disabled supports for `c8y_Command` and `c8y_RelayArray`.
2. `501`-`503` messages didn't use the Cumulocity's operation name for custom operations. Instead, the command name (`<cmd_name>`)  from the topic `te/.../cmd/<cmd_name>/tedge-mapper-1234` was used.

### Solution

If user wants extra arguments for `503`/`506`, they have to provide `exec.workflow.output`.
```toml
[exec]
topic = "c8y/devicecontrol/notifications"
on_fragment = "c8y_Command"

[exec.workflow]
operation = "shell_execute"
input.command = "${.payload.c8y_Command.text}"
output = "${.payload.result}"
```

Once c8y-mapper receives a c8y_Command operation, it adds extra internal object `c8y-mapper` (note: not hard-coded value. It's the same as the prefix of the command ID) to include data from the corresponding custom operation handler.

topic: `te/device/main///cmd/set_relay/c8y-mapper-54228`
```json
{
   "c8y-mapper":{
      "on_fragment":"c8y_Command",
      "output":"${.payload.result}"
   },
   "command":"echo helloworld",
   "status":"init"
}
```

The workflow is expected to add `result` field, so that c8y-mapper can convert the output to extra argument of the SmartREST message.

```json
{
  "@version": "3a52d32db97c0cd907624956ed28ed6869452257634e8b87317b5296145a0f92",
  "c8y-mapper": {
    "on_fragment": "c8y_Command",
    "output": "${.payload.result}"
  },
  "logPath": "/var/log/tedge/agent/workflow-set_relay-c8y-mapper-54228.log",
  "command":"echo helloworld",
  "result":"helloworld\n",
  "status": "successful"
}
```

then, the final SmartREST message is
```
503,c8y_Command,"helloworld\n"
```
or
```
506,54228,"helloworld\n"
```

Or `result` also accepts a json array which will be used to translate to multiple parameters in the 503/506 SmartREST message. For example:

```json
{
  "@version": "3a52d32db97c0cd907624956ed28ed6869452257634e8b87317b5296145a0f92",
  "c8y-mapper": {
    "on_fragment": "c8y_RelayArray",
    "output": "${.payload.result}"
  },
  "logPath": "/var/log/tedge/agent/workflow-set_relay-c8y-mapper-54228.log",
  "relay": ["OPEN", "CLOSED"],
  "result": ["OPEN","CLOSED"],
  "status": "successful"
}
```

then, the output SmartREST message is
```
503,c8y_RelayArray,OPEN,CLOSED
```
or
```
506,54228,OPEN,CLOSED
```

(Note: If the `result` property is an array, then normal csv encoding rules should be used to add double quotes around each individual item of the array if necessary)

## Types of changes

## Paste Link to the issue
#3095

## Checklist

## Further comments

**References**
* See the [Cumulocity Documentation Operations page](https://cumulocity.com/docs/smartrest/mqtt-static-templates/#updating-operations) which details which Cumulocity operations accept parameters along with the `503/506` SmartREST 2.0 message

### DIFF
--- a/crates/core/tedge_api/src/workflow/state.rs
+++ b/crates/core/tedge_api/src/workflow/state.rs
@@ -274,9 +274,9 @@ impl GenericCommandState {
         GenericCommandState::extract_text_property(&self.payload, REASON)
     }
 
-    /// Return value of the result field if any
-    pub fn result(&self) -> Option<&str> {
-        GenericCommandState::extract_text_property(&self.payload, RESULT)
+    /// Return value of the result field
+    pub fn result(&self) -> &Value {
+        self.payload.get(RESULT).unwrap_or(&Value::Null)
     }
 
     /// Extract a text property from a Json object
--- a/crates/core/tedge_api/src/workflow/state.rs
+++ b/crates/core/tedge_api/src/workflow/state.rs
@@ -72,6 +72,7 @@ const EXECUTING: &str = "executing";
 const SUCCESSFUL: &str = "successful";
 const FAILED: &str = "failed";
 const REASON: &str = "reason";
+const RESULT: &str = "result";
 
 impl GenericCommandState {
     pub fn new(topic: Topic, status: String, payload: Value) -> Self {
@@ -271,6 +272,11 @@ impl GenericCommandState {
     /// Return the error reason if any
     pub fn failure_reason(&self) -> Option<&str> {
         GenericCommandState::extract_text_property(&self.payload, REASON)
+    }
+
+    /// Return value of the result field if any
+    pub fn result(&self) -> Option<&str> {
+        GenericCommandState::extract_text_property(&self.payload, RESULT)
     }
 
     /// Extract a text property from a Json object
--- a/crates/core/tedge_api/src/workflow/state.rs
+++ b/crates/core/tedge_api/src/workflow/state.rs
@@ -75,8 +75,9 @@ const REASON: &str = "reason";
 const RESULT: &str = "result";
 
 impl GenericCommandState {
-    pub fn new(topic: Topic, status: String, payload: Value) -> Self {
+    pub fn new(topic: Topic, status: String, mut payload: Value) -> Self {
         let invoking_command_topic = Self::infer_invoking_command_topic(topic.as_ref());
+        Self::inject_text_property(&mut payload, STATUS, &status);
         GenericCommandState {
             topic,
             status,
--- a/crates/core/tedge_api/src/workflow/state.rs
+++ b/crates/core/tedge_api/src/workflow/state.rs
@@ -72,7 +72,6 @@ const EXECUTING: &str = "executing";
 const SUCCESSFUL: &str = "successful";
 const FAILED: &str = "failed";
 const REASON: &str = "reason";
-const RESULT: &str = "result";
 
 impl GenericCommandState {
     pub fn new(topic: Topic, status: String, mut payload: Value) -> Self {
@@ -273,11 +272,6 @@ impl GenericCommandState {
     /// Return the error reason if any
     pub fn failure_reason(&self) -> Option<&str> {
         GenericCommandState::extract_text_property(&self.payload, REASON)
-    }
-
-    /// Return value of the result field
-    pub fn result(&self) -> &Value {
-        self.payload.get(RESULT).unwrap_or(&Value::Null)
     }
 
     /// Extract a text property from a Json object
--- a/crates/extensions/c8y_mapper_ext/src/operations/convert.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/convert.rs
@@ -7,6 +7,7 @@ use c8y_api::json_c8y_deserializer::C8yLogfileRequest;
 use c8y_api::json_c8y_deserializer::C8yUploadConfigFile;
 use c8y_api::smartrest::message_ids::SET_SUPPORTED_CONFIGURATIONS;
 use c8y_api::smartrest::message_ids::SET_SUPPORTED_LOGS;
+use serde_json::json;
 use serde_json::Value;
 use std::sync::Arc;
 use tedge_api::commands::CommandStatus;
@@ -433,7 +434,16 @@ impl CumulocityConverter {
             serde_json::Value::Object(serde_json::Map::new())
         };
 
-        let request = GenericCommandState::new(topic, CommandStatus::Init.to_string(), payload);
+        let mapper_id = self.command_id.prefix();
+        let inject_object = json!({
+            mapper_id: {
+                "on_fragment": custom_handler.on_fragment(),
+                "output": custom_handler.workflow_output(),
+            }
+        });
+
+        let request = GenericCommandState::new(topic, CommandStatus::Init.to_string(), payload)
+            .update_with_json(inject_object);
 
         Ok(vec![request.into_message()])
     }
--- a/crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
@@ -1,4 +1,6 @@
 use anyhow::Context;
+use c8y_api::smartrest::smartrest_serializer::succeed_static_operation_with_id;
+use c8y_api::smartrest::smartrest_serializer::succeed_static_operation_with_name;
 use c8y_api::smartrest::smartrest_serializer::CumulocitySupportedOperations;
 use tedge_api::workflow::GenericCommandState;
 use tedge_api::CommandStatus;
@@ -31,10 +33,19 @@ impl OperationContext {
                 extra_messages: vec![],
             }),
             CommandStatus::Successful => {
-                let smartrest_set_operation = self.get_smartrest_successful_status_payload(
-                    CumulocitySupportedOperations::C8yCustom(operation_name.to_string()),
-                    cmd_id,
-                );
+                let payload = command.result();
+
+                let smartrest_set_operation = match self.get_operation_id(cmd_id) {
+                    Some(op_id) if self.smart_rest_use_operation_id => {
+                        succeed_static_operation_with_id(&op_id, payload)
+                    }
+                    _ => {
+                        let operation =
+                            CumulocitySupportedOperations::C8yCustom(operation_name.to_string());
+                        succeed_static_operation_with_name(operation, payload)
+                    }
+                };
+
                 let c8y_notification = MqttMessage::new(sm_topic, smartrest_set_operation);
 
                 Ok(OperationOutcome::Finished {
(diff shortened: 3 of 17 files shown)


crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
```diff
@@ -211,6 +240,73 @@ mod tests {
...
+        ))
+        .await
+        .expect("Send failed");
+
+        // Expect `502` smartrest message on `c8y/s/us` as 'result' contains unsupported type
+        assert_received_contains_str(
+            &mut mqtt,
+            [(
+                "c8y/s/us",
+                "502,command,'result' field must have String or Array",
```
rina23q, line 304: `command` is the name from the thin-edge MQTT message. It should be the Cumulocity operation name, which should be available from a custom operation handler definition...

albinsuresh:
> When the JSON payload contains result, its value will be sent as the third argument of the SmartREST messages.

Instead of relying on a predefined `result` value in the payload, how about giving the freedom to the user to define which all payload fields to as the SmartREST operation parameter values with an `output` definition as follows:

```
[exec.workflow]
operation = "set_relay"
input.relay = "${.payload.c8y_RelayArray}"
output = ["${.payload.relay1Status}","${.payload.relay1Status}"]
```

That way, the user can aggregate multiple different fields to produce the final parameters list as well. The `output` can be a single value or an array of values. That way, we don't leak any C8y contracts into the tedge API. The payloads can be as open as the user wants.

didier-wenzek:
> > When the JSON payload contains result, its value will be sent as the third argument of the SmartREST messages.
> 
> Instead of relying on a predefined `result` value in the payload, how about giving the freedom to the user to define which all payload fields to as the SmartREST operation parameter values with an `output` definition as follows:
> 
> ```
> [exec.workflow]
> operation = "set_relay"
> input.relay = "${.payload.c8y_RelayArray}"
> output = ["${.payload.relay1Status}","${.payload.relay1Status}"]
> ```
> 
> That way, the user can aggregate multiple different fields to produce the final parameters list as well. The `output` can be a single value or an array of values. That way, we don't leak any C8y contracts into the tedge API. The payloads can be as open as the user wants.

Good suggestion. This will be consistent with what is done for operation workflow the [output is extracted from the sub-operation payload](https://thin-edge.github.io/thin-edge.io/references/agent/operation-workflow/#awaiting-sub-command-completion).

`tedge_api` provides all the required building blocks.

- The custom operation definition has to include an `output` filed of type [`StateExcerpt`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/workflow/state.rs#L530)
- and to call [`output.extract_value_from()`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_api/src/workflow/state.rs#L557) on the final state message of the operation.
- This is similar to [what is done for the input](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/operations/convert.rs#L419), but in the reverse direction.

crates/core/tedge_api/src/workflow/state.rs
```diff
@@ -273,6 +274,11 @@ impl GenericCommandState {
         GenericCommandState::extract_text_property(&self.payload, REASON)
     }
 
+    /// Return value of the result field
+    pub fn result(&self) -> &Value {
+        self.payload.get(RESULT).unwrap_or(&Value::Null)
+    }
```
didier-wenzek, line 280: This sounds a bit too specific, this result being only used by C8y custom operations.

The payload being `pub`, this one-liner can be done directly in `c8y_mapper_ext`.

That said, what is proposed by Albin is by far a better approach. See https://github.com/thin-edge/thin-edge.io/pull/3283#issuecomment-2535137925

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
```diff
@@ -31,10 +34,36 @@ impl OperationContext {
                 extra_messages: vec![],
             }),
             CommandStatus::Successful => {
-                let smartrest_set_operation = self.get_smartrest_successful_status_payload(
-                    CumulocitySupportedOperations::C8yCustom(operation_name.to_string()),
-                    cmd_id,
-                );
+                let operation =
+                    CumulocitySupportedOperations::C8yCustom(operation_name.to_string());
```
didier-wenzek, line 38: I understand now the issue highlighted by this comment: https://github.com/thin-edge/thin-edge.io/pull/3283/files#r1876437428

The `OperationContext` gives no context about the c8y operation but only about the tedge operation.

As suggested offline by @rina23q:
- the C8y name of the operation can be injected into the tedge operation on init.
- and extracted back here when the tedge operation is finalized.

However, to implement what is suggested here https://github.com/thin-edge/thin-edge.io/pull/3283#issuecomment-2535137925, more context is required. One would need the whole c8y custom operation definition or at least the except to be used for the response.

-

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
didier-wenzek: Better to fix this bug now as it defeats the purpose of this PR.

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
rina23q: To proceed Albin's proposal, `OperationHandler` (aka `OperationContext`) must hold custom handlers. Once it has done, indeed the issue of c8y's operation name can be solved.

rina23q: Probably the quick fix is, to move `supported_operation` from `CumulocityConverter` to `OperationHandler`.
Then `supported_operation` can be updated by `CumulocityConverter` in runtime, and also `OperationHandler` can access it.

```rust
pub struct CumulocityConverter {
   ...
    // supported_operations: SupportedOperations, // remove from here
}

```

```rust
pub struct OperationHandler {
    context: Arc<OperationContext>,
    pub supported_operations: SupportedOperations,  // added
    running_operations: HashMap<Arc<str>, RunningOperation>,
}
```

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
rina23q: It's addressed in this PR.

crates/extensions/c8y_mapper_ext/src/operations/convert.rs
```diff
@@ -433,7 +434,17 @@ impl CumulocityConverter {
             serde_json::Value::Object(serde_json::Map::new())
         };
 
-        let request = GenericCommandState::new(topic, CommandStatus::Init.to_string(), payload);
+        let mapper_id = self.command_id.prefix();
+        let inject_object = json!({
+            "status": "init",
```
didier-wenzek, line 439: As noticed by @rina23q, this `status` has to be injected because of an unexpected behavior of `GenericCommandState::update_with_json` which expect a status in the injected value and inject an error status if none is provided.

=> I will push a fix to avoid this surprising behavior.

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
```diff
@@ -15,45 +20,110 @@ impl OperationContext {
...
+        if let Some(value) = value.get(mapper_id) {
+            if let Some(c8y_operation) = value.get("on_fragment") {
+                if let Some(c8y_operation) = c8y_operation.as_str() {
+                    return Some(CumulocitySupportedOperations::C8yCustom(
+                        c8y_operation.to_string(),
+                    ));
+                }
+            }
+        }
+        None
```
didier-wenzek, line 86: It's simpler to use [`Value::pointer()`](https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html#method.pointer)

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
```diff
@@ -15,45 +20,110 @@ impl OperationContext {
...
                 })
             }
             _ => Ok(OperationOutcome::Ignored),
+        };
+
+        (result, Some(operation))
+    }
+
+    fn get_c8y_operation(
+        &self,
```
didier-wenzek, line 73: Unused

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
```diff
@@ -15,45 +20,110 @@ impl OperationContext {
...
+    }
+
+    fn convert_output(
+        &self,
+        mapper_id: &str,
+        state: &GenericCommandState,
+        operation: CumulocitySupportedOperations,
+        cmd_id: &str,
+    ) -> SmartrestPayload {
+        match state.payload.get(mapper_id).unwrap().get("output") {
```
didier-wenzek, line 96: The state being received from MQTT, it's not safe to assume the payload has a `c8y` fragment.

This has to be replaced by a function similar to `get_c8y_operation`.

didier-wenzek (commented):
The use of `StateExcerpt` to extract the result from the operation final payload is correct. However, some points can be simpler.

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
didier-wenzek: It's *currently* safe but only thanks to the *current* sequence of function calls, something that might change in the future.

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
rina23q: Addressed in 43172ff512acc90477998570d800e38e4b060795

crates/extensions/c8y_mapper_ext/src/operations/convert.rs
didier-wenzek: Here is the fixup commit: https://github.com/thin-edge/thin-edge.io/pull/3283/commits/e9eb073e9546cc64c64cd2be590a2d47844eb9d5.

In fact the issue was in `GenericCommandState::new()` which was not properly injecting the state in the paylaod.

crates/extensions/c8y_mapper_ext/src/operations/handlers/custom_operation.rs
```diff
@@ -15,47 +20,108 @@ impl OperationContext {
         target: &EntityTarget,
         cmd_id: &str,
         message: &MqttMessage,
-        operation_name: &str,
-    ) -> Result<OperationOutcome, OperationError> {
+    ) -> (
+        Result<OperationOutcome, OperationError>,
+        Option<CumulocitySupportedOperations>,
+    ) {
```
didier-wenzek, line 26: This return type is unusual.

- What's the rational?
- What does it means to return `Err(error), Some(operation))`?

rina23q: 1. `Err` never returned actually, thus, I changed the return type in [064fd77](https://github.com/thin-edge/thin-edge.io/pull/3283/commits/064fd77cc8896951badf4260d290f1eda03f2272).
1. The c8y supported operation name has to be returned during processing a custom operation status transition. It can be `None`, if `.payload.c8y-mapper.on_fragment` is not provided.

I have to admit it is weird and the tuple return is a kind of a workaround for `OperationContext`...

didier-wenzek (approved):
Approved

Issue #3279, opened 2024-12-04
c8y mapper not starting up reliably


reey:
**Describe the bug**
When ever I restart my machine or restart the c8y mapper service (`sudo systemctl restart tedge-mapper-c8y.service`) it is not really reliably starting up again. The mqtt messages on the `te/#` topics are not processed so that no messages are actually published to the `c8y/#` topic.

I've kept the mapper in this not functioning state for ~8 hours to verify if it comes back online at some point, but it does not.
About 1 out of 5 startup attempts of the service is succeeding for me. I had the feeling that with debug logging enabled for the mapper it was starting up a bit more reliable.. Maybe some race condition? I've also got 82 child devices connected right now. I'm not sure what the mapper might be doing on startup, but maybe the number of devices is triggering a race condition?

**To Reproduce**
- Restart the c8y mapper using `sudo systemctl restart tedge-mapper-c8y.service`
- verify whether mqtt messages are still sent to `c8y/#`

**Expected behavior**
Restarting the `tedge-mapper-c8y` service should not cause any issues and the mapping functionality should proceed afterwards.

**Screenshots**
-

**Environment (please complete the following information):**
|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`Raspberry Pi 4 Model B Rev 1.2`|
|System-Architecture|`Linux raspberrypi 6.6.62+rpt-rpi-v8 #1 SMP PREEMPT Debian 1:6.6.62-1+rpt1 (2024-11-25) aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.3.1`|
|mosquitto version|`2.0.11`|

**Additional context**
I've got about 82 device child devices connected to my thin-edge instance.

I've attached an encrypted debug log file: [thin-edge-c8y-mapper.age.log](https://github.com/user-attachments/files/18004636/thin-edge-c8y-mapper.age.log)

Bravo555:
From looking at the log, it looks like mosquitto can't establish a bridge between the local broker and the cloud:
    - there are `MQTT pub` and `MQTT recv` logs for `te/...` topics, but for `c8y/...` topics, there are only `MQTT pub` lines, so we're successfully publishing on local `c8y/...` topics, but the bridge can't be established with c8y, that's why messages we publish probably aren't being sent to the cloud and we're not getting any messages from the cloud as well.

[...]

PR #3289, opened 2024-12-09
fix(#3279): plug `AvailabilityActor` directly to `MqttActor`

Bravo555:
## TODO
- [ ] tests

## Proposed changes

Because the C8yMapperActor synchronously first receives a message from the input channel, and then produces output by sending to the output channel, if one of these channels block, the other one will also not be processed.

In this case, there was a loop: C8yMapperActor sent `MqttMessage`s to AvailabilityActor, and AvailabilityActor send `PublishMessage`s to C8yMapperActor for it to relay it to MqttActor.

When a sender for `PublishMessage`s was full when AvailabilityActor tried sending another message, it would block, so until this message was processed, AvailabilityActor wasn't processing now input. However, if before this output was sent by C8yMapperActor there was another message to relay to AvailabilityActor, and the inbound channel was also full, this would result in a deadlock where neither AvailabilityActor input or output could be sent through.

## Types of changes

## Paste Link to the issue

- #3279

## Checklist

### DIFF
--- a/crates/core/tedge_mapper/src/c8y/mapper.rs
+++ b/crates/core/tedge_mapper/src/c8y/mapper.rs
@@ -266,7 +266,6 @@ impl TEdgeComponent for CumulocityMapper {
             Some(AvailabilityBuilder::new(
                 AvailabilityConfig::try_new(&tedge_config, c8y_profile)?,
                 &mut c8y_mapper_actor,
-                &mut mqtt_actor,
                 &mut timer_actor,
             ))
         } else {
--- a/crates/core/tedge_mapper/src/c8y/mapper.rs
+++ b/crates/core/tedge_mapper/src/c8y/mapper.rs
@@ -266,6 +266,7 @@ impl TEdgeComponent for CumulocityMapper {
             Some(AvailabilityBuilder::new(
                 AvailabilityConfig::try_new(&tedge_config, c8y_profile)?,
                 &mut c8y_mapper_actor,
+                &mut mqtt_actor,
                 &mut timer_actor,
             ))
         } else {
--- a/crates/extensions/c8y_mapper_ext/src/availability/builder.rs
+++ b/crates/extensions/c8y_mapper_ext/src/availability/builder.rs
@@ -1,3 +1,4 @@
+use crate::actor::PublishMessage;
 use crate::availability::actor::AvailabilityActor;
 use crate::availability::AvailabilityConfig;
 use crate::availability::AvailabilityInput;
@@ -31,8 +32,7 @@ pub struct AvailabilityBuilder {
 impl AvailabilityBuilder {
     pub fn new(
         config: AvailabilityConfig,
-        mqtt_in: &mut impl MessageSource<MqttMessage, Vec<ChannelFilter>>,
-        mqtt_out: &mut impl MessageSink<MqttMessage>,
+        mqtt: &mut (impl MessageSource<MqttMessage, Vec<ChannelFilter>> + MessageSink<PublishMessage>),
         timer: &mut impl Service<TimerStart, TimerComplete>,
     ) -> Self {
         let mut box_builder: SimpleMessageBoxBuilder<AvailabilityInput, AvailabilityOutput> =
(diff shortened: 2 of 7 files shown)


crates/extensions/c8y_mapper_ext/src/availability/builder.rs
```diff
@@ -32,19 +31,20 @@ pub struct AvailabilityBuilder {
 impl AvailabilityBuilder {
     pub fn new(
         config: AvailabilityConfig,
-        mqtt: &mut (impl MessageSource<MqttMessage, Vec<ChannelFilter>> + MessageSink<PublishMessage>),
+        mqtt_in: &mut impl MessageSource<MqttMessage, Vec<ChannelFilter>>,
+        mqtt_out: &mut impl MessageSink<MqttMessage>,
```
didier-wenzek, line 35: Nice. You put the finger on the root of the issue.

The badly name `mqtt` connector is not a true MQTT connector (`impl MessageSource<MqttMessage, TopicFilter> + MessageSink<MqttMessage>`),  *but* the c8y mapper forwarding some of the messages to this actor.

You broke the loop (between this actor and the c8y actor), by sending the response messages directly to an MQTT actor. I wonder if we can go a step further. Simply disconnect this actor from the c8y actor and let it consume its input directly from an MQTT actor.

```suggestion
        mqtt: &mut (impl MessageSource<MqttMessage, TopicFilter> + MessageSink<MqttMessage>)
```

You will have not only to adapt `AvailabilityBuilder::mqtt_message_builder()` (exactly as you currently propose) but also the `AvailabilityBuilder::channels()` function but not the tests. Construction of the mapper will have to be changed but lightly.

crates/extensions/c8y_mapper_ext/src/availability/tests.rs
```diff
@@ -414,24 +413,50 @@ async fn timer_send(timer: &mut FakeServerBox<TimerStart, TimerComplete>, event:
...
+    // create two separate Mqtt message boxes to circumvent "can't borrow the same object as mutable more than one time"
+    // use the message boxes because they implement `MessageSource` and `MessageSink` (receivers and senders alone do
+    // not) and then take out their halves not given to the actor and recombine them
+    //
+    // we need separate mqtt in and mqtt out because on the in-side CumulocityMapperActor attaches a `ChannelFilter`
+    // which we use to only get messages we care about, and on the out-side we want to output directly to the MQTT actor
+    // because on CumulocityMapperActor side the halves that send to AvailabilityActor and receive from
+    // AvailabilityActor aren't separate, so one can block the other
+    //
+    // TODO: find a cleaner and more permanent solution
```
didier-wenzek, line 430: One solution, is to fully disconnect the  availability actor from the c8y actor and let it consume its input directly from an MQTT actor.

albinsuresh: > One solution, is to fully disconnect the availability actor from the c8y actor and let it consume its input directly from an MQTT actor.

Routing the MQTT messages from the availability actor to the MQTT actor via the intermediary C8y actor was an intentional design choice to avoid a race condition that we encountered when a direct connection was implemented in the first drop of that availability feature.

With the old design where the availability actor and c8y actor were connected directly to the MQTT actor independently, when a new device was getting registered, the c8y actor would map and publish the device creation message and the availability actor would publish the availability message as well independently. This was causing a race condition where sometimes the availability message was reaching the cloud first, before the device creation message, which messes up the device metadata.

So, we needed a mechanism to make sure that the availability messages for an entity are pushed to the cloud only after its respective registration message is already processed and pushed. By routing the MQTT messages from the availability actor to the C8y actor, we could ensure this sequentiality. So, if we're breaking that design, we must find another solution to the race condition as well.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -266,6 +266,7 @@ impl TEdgeComponent for CumulocityMapper {
             Some(AvailabilityBuilder::new(
                 AvailabilityConfig::try_new(&tedge_config, c8y_profile)?,
                 &mut c8y_mapper_actor,
+                &mut mqtt_actor,
```
albinsuresh, line 269: Breaking the MQTT source and sink was definitely a good idea. But, I feel we need to flip the source and sink. It might be okay for the availability actor to receive the messages directly from the MQTT actor, but the output must be routed through the c8y actor for the reason described in https://github.com/thin-edge/thin-edge.io/pull/3289/files#r1877418069

Bravo555:
Superseded by #3295

PR #3291, opened 2024-12-11
fix: ensure concurrent server always responds to shutdown signal

jarhodes314:
## Proposed changes
This fixes an issue that was causing `tedge-mapper` to fail to shutdown properly in certain cases, namely when the HTTP actor was busy when the shutdown request was received.

The root cause is in these lines of code
https://github.com/thin-edge/thin-edge.io/blob/82cde4094fbfdb073fa26b353393f9f59e3cbef4/crates/core/tedge_actors/src/servers/message_boxes.rs#L51-L65

When we receive a `RuntimeRequest::Shutdown`, `self.requests.recv()` returns `None`. Since this doesn't match `Some(request)`, this branch then becomes inactive and we wait for `self.running_requests_handlers.next()`. If any request handlers are running, when this returns, the code will loop, running `self.requests.recv()` again, meaning we've completely ignored the shutdown request and will never receive another one.

A similar thing happens if the running requests do not finish in a timely manner after the shutdown request is received. We have a unit test for this, however this only tests the case where the actor is at capacity (i.e. we have a task waiting that cannot be scheduled due to the concurrency limit). Given in this case the actor stops running immediately upon receiving the shutdown request, I have replicated this behaviour for the case where the actor is not at capacity.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_actors/src/servers/message_boxes.rs
+++ b/crates/core/tedge_actors/src/servers/message_boxes.rs
@@ -50,8 +50,8 @@ impl<Request: Message, Response: Message> ConcurrentServerMessageBox<Request, Re
 
         loop {
             tokio::select! {
-                Some(request) = self.requests.recv() => {
-                    return Some(request);
+                received = self.requests.recv() => {
+                    return received
                 }
                 Some(result) = self.running_request_handlers.next() => {
                     if let Err(err) = result {
@@ -156,7 +156,7 @@ impl<Request: Message, Response: Message> Sender<Request> for RequestSender<Requ
 
 #[cfg(test)]
 #[cfg(feature = "test-helpers")]
-mod tests {
+mod tests_using_helpers {
     use super::*;
 
     use crate::Builder;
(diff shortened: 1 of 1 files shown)


crates/core/tedge_actors/src/servers/message_boxes.rs
```diff
@@ -339,4 +339,55 @@ mod tests {
             Ok(())
         );
     }
+
```
jarhodes314, line 342: I was slightly horrified to see how complicated the test case above is. I've tested the code more directly (i.e. without the actor actually running, just polling `next_request` manually) as a result. A better solution might be to try and abstract away some of the setup of the above test, and then run essentially the same test just with fewer in-flight requests to test the correct code path.

crates/core/tedge_actors/src/servers/message_boxes.rs
```diff
@@ -339,4 +339,55 @@ mod tests {
...
+    #[tokio::test]
```
didier-wenzek, line 343: Good to have these two new tests actually detecting the issue. However, it would could to put them in a test module that doesn't require `cfg(feature = "test-helpers")`.

didier-wenzek: I pushed a fixup commit to fix that https://github.com/thin-edge/thin-edge.io/pull/3291/commits/f0926246fb4e0f282c7fd7cb328967fa1c0c1258

didier-wenzek (approved):
Approved

PR #3295, opened 2024-12-12
fix #3279: Create a fast path for sending `PublishMessage`s

### LINKED ISSUE #3279: c8y mapper not starting up reliably

Bravo555:
## Proposed changes

Solve the deadlock that can arise when C8yMapperActor can't complete sending MqttMessage to some other actor because that other actor can't complete the send of a `PublishMessage` to C8yMapperActor by creating a fast path to drain PublishMessage buffer ASAP.

Previously when `PublishMessage`s were a part of the main message box we could not prioritize them because we had to process messages in order;

separating out the receiver for `PublishMessage`s makes it so no other messages can block `PublishMessage`s and allows us to use select! to concurrently process them when processing other events blocks

## Types of changes

## Paste Link to the issue

- #3279

## Checklist

## Further comments

About the test: added one in Robot Framework, but it has some problems: for some reasons it takes very long and it publishes a lot of messages to the cloud. Given that this test could be 100% local and would complete almost instantly as a Rust unit test, it would probably be an improvement to move it there but it's easier to add in Robot Framework for now.

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -13,6 +13,7 @@ use tedge_actors::Builder;
 use tedge_actors::ClientMessageBox;
 use tedge_actors::CloneSender;
 use tedge_actors::DynSender;
+use tedge_actors::LoggingReceiver;
 use tedge_actors::LoggingSender;
 use tedge_actors::MessageReceiver;
 use tedge_actors::MessageSink;
@@ -41,6 +42,7 @@ use tedge_uploader_ext::UploadRequest;
 use tedge_uploader_ext::UploadResult;
 use tedge_utils::file::create_directory_with_defaults;
 use tedge_utils::file::FileError;
+use tokio::select;
 
 const SYNC_WINDOW: Duration = Duration::from_secs(3);
 
@@ -68,7 +70,7 @@ impl From<PublishMessage> for MqttMessage {
     }
 }
 
-fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete, PublishMessage] : Debug);
+fan_in_message_type!(C8yMapperInput[MqttMessage, FsWatchEvent, SyncComplete] : Debug);
 
 type C8yMapperOutput = MqttMessage;
 
@@ -79,6 +81,16 @@ pub struct C8yMapperActor {
     timer_sender: LoggingSender<SyncStart>,
     bridge_status_messages: SimpleMessageBox<MqttMessage, MqttMessage>,
     message_handlers: HashMap<ChannelFilter, Vec<LoggingSender<MqttMessage>>>,
+    /// Receiver for MqttMessages sent by other actors meant to be sent to MQTT actor to be published.
+    ///
+    /// Other actors may want to send MqttMessages but we want to ensure that the entity referred to in the message
+    /// topic exists; that's why we route them to the mapper rather than to MqttActor directly.
+    ///
+    /// Needs to be a separate receiver so that we can select! and drain the message queue ASAP. Otherwise
+    /// `process_mqtt_message` may involve sending messages to other actors, and if those other actors are blocked on
+    /// sending a `PublishMessage` because we're trying to send a message to them before processing their responses,
+    /// we'll hit a deadlock.
+    publish_messages: LoggingReceiver<PublishMessage>,
 }
 
 #[async_trait]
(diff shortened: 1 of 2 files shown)


crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -111,20 +123,27 @@ impl Actor for C8yMapperActor {
...
+                        }
+                        C8yMapperInput::FsWatchEvent(event) => {
+                            self.process_file_watch_event(event).await?;
+                        }
+                        C8yMapperInput::SyncComplete(_) => {
+                            self.process_sync_timeout().await?;
+                        }
+                    }
+                }
+                else => break
```
Bravo555, line 146: Good thing that we went over the details of `select!` usage today 😅

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -111,20 +123,28 @@ impl Actor for C8yMapperActor {
...
-                C8yMapperInput::PublishMessage(message) => {
+        loop {
+            select! {
+                // in some tests may not be connected to the sender, so we need to ignore `None`
+                // here, but should get picked up by the main message box anyway
+                Some(message) = self.publish_messages.recv() => {
                     self.mqtt_publisher.send(message.0).await?;
                 }
+                event = self.messages.recv() => {
+                    let Some(event) = event else { break; };
```
didier-wenzek, line 134: The following seems more appropriate: i.e. let consume the published messages till the end even if there is no more input messages.

```suggestion
                Some(event) = self.messages.recv() => {
```

didier-wenzek (commented):
Introducing a second receiver is a nice way to manage priority among messages, here to process first any message that is a reaction of the latest processed message.

Thank you for your perseverance, trying several fixes before finding the more appropriate one.

I will be please to see this PR merged. Just one request: can you add a system test reproducing the issue i.e. when the mapper starts with a long pending list of child device registration.

didier-wenzek:
I confirm that:

- the [test](https://github.com/thin-edge/thin-edge.io/pull/3295/commits/55426e2068a3753530a378a34ec30f031f82023d) correctly reproduce the [issue](https://github.com/thin-edge/thin-edge.io/issues/3279), the mapper blocking on start when overwhelmed with too many child device registration messages
- this PR correctly fix the issue removing the deadlock from the c8y mapper.

crates/extensions/c8y_mapper_ext/src/actor.rs
didier-wenzek: No, I'm wrong this would block the mapper on `^C`, as `self.messages.recv()` returns `None` in such a case.

didier-wenzek:
> About the test: added one in Robot Framework, but it has some problems: for some reasons it takes very long and it publishes a lot of messages to the cloud. Given that this test could be 100% local and would complete almost instantly as a Rust unit test, it would probably be an improvement to move it there but it's easier to add in Robot Framework for now.

I'm okay with this for now.

didier-wenzek (approved):
Approved.

Thank you for your perseverance and the nice solution you arrived at by exploring several alternatives.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -111,20 +123,28 @@ impl Actor for C8yMapperActor {
...
-                C8yMapperInput::FsWatchEvent(event) => {
-                    self.process_file_watch_event(event).await?;
-                }
-                C8yMapperInput::SyncComplete(_) => {
-                    self.process_sync_timeout().await?;
-                }
-                C8yMapperInput::PublishMessage(message) => {
+        loop {
+            select! {
+                // in some tests may not be connected to the sender, so we need to ignore `None`
```
albinsuresh, line 128:
```suggestion
                // in some tests may not be connected to the sender, so we need to ignore `None`
```
What may not be connected to the sender?

albinsuresh (approved):
LGTM

rina23q (commented):
Minor thing. 
I think the system test is more appropriate to be located under the `registration` directory as it's not a part of availability feature, rather (a lot of) child device registration. The test checks `101`, which is SmartREST child device registration message. Though it was connected to the AvailabilityActor.

So my suggestion is moving the test from `/availability` to `/registration`.

crates/extensions/c8y_mapper_ext/src/actor.rs
Bravo555: Corrected the comment to mention the AvailabilityActor

rina23q (approved):

PR #3299, opened 2024-12-16
doc: cloud profiles in Cumulocity

jarhodes314:
## Proposed changes
Adds documentation for using cloud profiles to connect to multiple Cumulocity instances. Hides the `c8y.device.id` and related configurations since those don't yet work due to external ID issues.

Hiding the device ID is done in a single separate commit so it should be easy to revert once the feature works properly.

There is also a small change to ensure profile names are always lowercased, in case someone accesses a key using e.g. `tedge config get c8y.profiles.EDGE.url`, where we weren't previously lowercasing the profile name.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/docs/src/operate/c8y/cloud-profiles.md
+++ b/docs/src/operate/c8y/cloud-profiles.md
@@ -1,0 +1,176 @@
+---
+title: Cloud Profiles
+tags: [Operate, Cumulocity, Cloud Profile]
+description: Connecting %%te%% to multiple Cumulocity tenants
+---
+
+import UserContext from '@site/src/components/UserContext';
+import UserContextForm from '@site/src/components/UserContextForm';
+
+Starting with version 1.4.0, **%%te%%** supports multiple Cumulocity connections
+using cloud profiles. This can be useful for migrating from one Cumulocity
+tenant to another, where Cloud profiles allow you to configure and run multiple
+`tedge-mapper c8y` instances from a single `tedge.toml` configuration file.
+
+:::tip
+#### User Context {#user-context}
+
+You can customize the documentation and commands shown on this page by providing
+relevant settings which will be reflected in the instructions. It makes it even
+easier to explore and use %%te%%.
+
+<UserContextForm settings="DEVICE_ID,C8Y_URL,C8Y_USER" />
+
+The user context will be persisted in your web browser's local storage.
+:::
+
+## Configuration
+There are a few values that need to be configured before we are able to connect
+to a second Cumulocity tenant.
+
+### URL
+To connect to a second tenant, start by configuring the URL of the new tenant:
+
+```sh
+sudo tedge config set c8y.url other.cumulocity.com --profile second
+```
+
+The profile name can be any combination of letters and numbers, and is used only
+to identify the cloud profile within thin-edge. The names are case insensitive,
+so `--profile second` and `--profile SECOND` are equivalent.
+
+You can now see the configuration has been applied to `tedge.toml`:
+
+```sh
+tedge config list url
+```
+
+<UserContext language="sh" title="Output">
+
+```sh
+c8y.url=$C8Y_URL
+c8y.profiles.second.url=other.cumulocity.com
+```
+
+</UserContext>
+
+In addition to the URL there are a couple of other configurations that need to
+be set for the second mapper:
+- the MQTT bridge topic prefix
+- the Cumulocity proxy bind port
+
+### MQTT bridge topic prefix
+```sh
+sudo tedge config set c8y.bridge.topic_prefix c8y-second --profile second
+```
+
+Setting `c8y.bridge.topic_prefix` will change the MQTT topics that the
+Cumulocity bridge publishes to/listens to in mosquitto. The default value is
+`c8y`, so the mappper publishes measurements to `c8y/s/us`, and this is
+forwarded to Cumulocity on the `s/us` topic. In the example above, we set the
+topic prefix to `c8y-second`, so the equivalent local topic would
+`c8y-second/s/us`. It is recommended, but not required, to include `c8y` in the
+topic prefix, to make it clear that the relevant topics are bridge topics that
+will forward to and from Cumulocity.
+
+### Cumulocity proxy bind port
+```sh
+sudo tedge config set c8y.proxy.bind.port 8002 --profile second
+```
+
+Since the Cumulocity mapper hosts a [proxy server for
+Cumulocity](../../references/cumulocity-proxy.md) and there will be a second
+mapper instance running, this configuration also needs to be unique per profile.
+
+### Optional per-profile configurations
+All the Cumulocity-specific configurations (those listed by `tedge config list
+--doc c8y`) can be specified per-profile to match tenant-specific constraints.
+
+## Connecting
+Once the second cloud profile has been configured, you can finally connect the
+second mapper using:
+```sh
+sudo tedge connect --profile second
+```
+<UserContext language="" title="Output">
+```
+Connecting to Cumulocity with config:
+        device id: $DEVICE_ID
+        cloud profile: second
+        cloud host: other.cumulocity.com:8883
+        certificate file: /etc/tedge/device-certs/tedge-certificate.pem
+        bridge: mosquitto
+        service manager: systemd
+        mosquitto version: 2.0.11
+Creating device in Cumulocity cloud... ✓
+Restarting mosquitto... ✓
+Waiting for mosquitto to be listening for connections... ✓
+Verifying device is connected to cloud... ✓
+Enabling tedge-mapper-c8y@second... ✓
+Checking Cumulocity is connected to intended tenant... ✓
+Enabling tedge-agent... ✓
+```
+</UserContext>
+
+Once the mapper is running, you can restart it by running:
+
+```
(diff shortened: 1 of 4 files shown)


docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,156 @@
...
+```sh
+sudo tedge config set c8y.bridge.topic_prefix c8y-second --profile second
+```
+
+Setting `c8y.bridge.topic_prefix` will change the MQTT topics that the
+Cumulocity bridge publishes to/listens to in mosquitto. The default value is
+`c8y`, so the mappper publishes measurements to `c8y/s/us`, and this is
+forwarded to Cumulocity on the `s/us` topic. In the example above, we set the
+topic prefix to `c8y-second`, so the equivalent local topic would
+`c8y-second/s-us`. It is recommended, but not required, to include `c8y` in the
```
reubenmiller, line 58:
```suggestion
`c8y-second/s/us`. It is recommended, but not required, to include `c8y` in the
```

docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,156 @@
...
+```sh
+sudo tedge config set c8y.proxy.bind.port 8002 --profile second
+```
+
+Since the Cumulocity mapper hosts a [proxy server for
+Cumulocity](../../references/cumulocity-proxy.md) and there will be a second
+mapper instance running, this configuration also needs to be unique per profile.
+
+## Connecting
+Connecting the second mapper
```
reubenmiller, line 72:
```suggestion
Once the seconds cloud profile has been configured, you can finally connect the second mapper using:
```

docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,156 @@
...
+</UserContext>
+
+Once the mapper is running, you can restart it by running:
+
+```
+sudo systemctl restart tedge-mapper-c8y@second
+```
+
+This uses a systemd service template to create the `tedge-mapper-c8y@second`
+servicce. If you are not using systemd, you will need to create a service
```
reubenmiller, line 103:
```suggestion
service. If you are not using systemd, you will need to create a service
```

docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,156 @@
+---
+title: Cloud Profiles
+tags: [Operate, Cumulocity, Cloud Profile]
+description: Connecting %%te%% to multiple Cumulocity tenants
+---
+
+import UserContext from '@site/src/components/UserContext'; import
+UserContextForm from '@site/src/components/UserContextForm';
+
+**%%te%%** now supports multiple Cumulocity connections using cloud profiles.
```
didier-wenzek, line 20: *Now* is vague.

Either:

```suggestion
**%%te%%** supports multiple Cumulocity connections using cloud profiles.
```

or :

```suggestion
Starting with version 1.4.0, **%%te%%** supports multiple Cumulocity connections using cloud profiles.
```

I would go with the second suggestion, as the docs are already versioned.

reubenmiller:
@jarhodes314  I pushed one commit to add the `<UserContextForm` at the top of the page, and I've also created a PR on the tedge-docs project to support the passing the `title` property from the `<UserContext>` element to the codeblock (which your PR uses a lot). https://github.com/thin-edge/tedge-docs/pull/96

docs/src/operate/c8y/cloud-profiles.md
reubenmiller: I would also go with the second option.

docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,166 @@
...
+
+```sh
+c8y.url=$C8Y_URL
+c8y.profiles.second.url=other.cumulocity.com
+```
+
+</UserContext>
+
+In addition to the URL there are a couple of other configurations that need to
+be set for the second mapper.
```
didier-wenzek, line 56: I would be explicit here, listing the two mandatory settings:

```suggestion
In addition to the URL there are a couple of other configurations that need to
be set for the second mapper:
- the MQTT bridge topic prefix
- and the Cumulocity proxy bind port.
```

docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,166 @@
...
+
+### Cumulocity proxy bind port
+```sh
+sudo tedge config set c8y.proxy.bind.port 8002 --profile second
+```
+
+Since the Cumulocity mapper hosts a [proxy server for
+Cumulocity](../../references/cumulocity-proxy.md) and there will be a second
+mapper instance running, this configuration also needs to be unique per profile.
+
```
didier-wenzek, line 80: I would  add 3rd sub-section telling that any c8y setting can be adapted.

```suggestion

### Optional per-profile setting

All the Cumulocitu specific settings (those listed by `tedge config list --doc c8y`) can be adapted to match tenant specific constraints.
```

jarhodes314:
> @jarhodes314 I pushed one commit to add the `<UserContextForm>` at the top of the page, and I've also created a PR on the tedge-docs project to support the passing the `title` property from the `<UserContext>` element to the codeblock (which your PR uses a lot). [thin-edge/tedge-docs#96](https://github.com/thin-edge/tedge-docs/pull/96)

Thanks, I made the change locally and then completely forgot that it was separate thing I needed to deal with PR wise.

docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,156 @@
...
+title: Cloud Profiles
+tags: [Operate, Cumulocity, Cloud Profile]
+description: Connecting %%te%% to multiple Cumulocity tenants
+---
+
+import UserContext from '@site/src/components/UserContext'; import
+UserContextForm from '@site/src/components/UserContextForm';
+
+**%%te%%** now supports multiple Cumulocity connections using cloud profiles.
+This can be useful for migrating from one Cumulocity tenant to another, where
```
albinsuresh, line 21: While this is also a genuine use-case, I thought the primary use case is to keep the device connected to multiple tenants parallelly for different kinds of data. E.g: Connected to an edge instance for faster edge telemetry processing and a cloud instance for remote device management. Do we want to promote the feature like that or focus just on migration cases for now(until it is battle tested)?

albinsuresh (commented):
> There is also a small change to ensure profile names are always lowercased, in case someone accesses a key using e.g. tedge config get c8y.profiles.EDGE.url, where we weren't previously lowercasing the profile name.

I have no objections to limiting the profile names to lower case only. But, just curious to know if there was any specific reason why you decided to make it case-insensitive?

reubenmiller:
> I have no objections to limiting the profile names to lower case only. But, just curious to know if there was any specific reason why you decided to make it case-insensitive?

Setting the values via env variables comes to mind where it would benefit from this, e.g. `TEDGE_C8Y_PROFILES_MYNAME_URL` where using uppercase in convention for env variables...`TEDGE_C8Y_PROFILES_MyName_URL` just looks weird.

But normalizing to lowercase just avoids problems from non linux users whom just expect things to be case-insensitive these days, e.g. Windows and MacOS are generally case insensitive when it comes to file/folder names, and linux users would probably already use lower case (because they know better)).

docs/src/operate/c8y/cloud-profiles.md
reubenmiller: yeah we could, but we can always add additional use cases post release

jarhodes314:
I've created a PR on tedge-docs to support `$C8Y_PROFILE_NAME` and `$C8Y_PROFILE_URL`. I don't think it's worth parameterising the topic prefix separately, I've just inferred what I believe to be the most sensible value there from the profile name.

docs/src/operate/c8y/cloud-profiles.md
```diff
@@ -0,0 +1,189 @@
...
+import UserContextForm from '@site/src/components/UserContextForm';
+
+:::tip
+#### User Context {#user-context}
+
+You can customize the documentation and commands shown on this page by providing
+relevant settings which will be reflected in the instructions. It makes it even
+easier to explore and use %%te%%.
+
+<UserContextForm settings="C8Y_PROFILE_NAME,C8Y_PROFILE_URL,C8Y_URL,DEVICE_ID" />
```
jarhodes314, line 17: @reubenmiller I'm wondering whether I actually want `C8Y_URL` and `DEVICE_ID` here. The device id certainly affects nothing people will copy and paste (just the `tedge connect` output, and I think the additional URL is just going to make things more confusing, since people have presumably followed a separate guide.

If we do want `C8Y_URL` here, I'm tempted to put `C8Y_PROFILE_URL` before it in the form since that's the most pertinent configuration value here.

reubenmiller: Though `C8Y_URL` is still used in various places (when showing the user the interaction between setting the default/primary c8y.url and the second cloud profile).

So we can leave it as-is for now, and look at adding some more context to the form fields (e.g. maybe a description of what the field actually means etc.)

didier-wenzek (approved):
Approved

Issue #3297, opened 2024-12-14
parsing error on c8y/devicecontrol/notifications topic if more than 1 operation is present


reubenmiller:
**Describe the bug**

thin-edge.io does not parse the message body of `c8y/devicecontrol/notifications` correctly when there are multiple operations are included in a single message.

The following error is shown on the `te/errors` topic showing the parsing error.

```sh
[te/errors] Failed to convert a message on topic 'c8y/devicecontrol/notifications': trailing characters at line 2 column 1
```

See the additional context section for a full example showing the payload with multiple operations in a single MQTT message payload.

**To Reproduce**

1. Disconnect the device from the internet
2. Wait until the bridge registers the disconnection
3. Create at least 2 operations
4. Subscribe to the local MQTT broker (to monitor the output)

    ```sh
    tedge mqtt sub '#'
    ```

4. Connect the device to the internet and wait for the bridge to connect again
5. Inspect the `tedge mqtt sub` output

**Expected behavior**

thin-edge.io should support parsing the message payload which includes multiple cloud operations in the same message received on the `c8y/devicecontrol/notifications` topic.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`unknown`|
|System-Architecture|`Linux tedge 6.8.0-39-generic #39-Ubuntu SMP PREEMPT_DYNAMIC Sat Jul  6 02:50:39 UTC 2024 aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.3.2~388+g6388202`|

**Additional context**

**MQTT Output**

The snippet below shows an example of a message received from Cumulocity on the `c8y/devicecontrol/notifications` topic which includes multiple operations (one per line)

```sh
[...]

PR #3301, opened 2024-12-16
fix: Collect multiple smartrest messages from c8y devicecontrol topic

Bravo555:
## Proposed changes

c8y/devicecontrol/notifications topic, as many others, can contain in its smartrest payload multiple records. For now ensure that multiple records are parsed from this topic in particular, as C8y sends multiple operations in a single message if e.g. device is offline, multiple operations are started, and then the device connects, it then receives all those operations in a single MQTT message.

But a longer term solution would be to refactor the code to ensure that we're supporting collecting multiple smartrest messages from all the possible MQTT topics.

## Types of changes

## Paste Link to the issue

#3297

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -28,7 +28,7 @@ use c8y_api::json_c8y_deserializer::C8ySoftwareUpdate;
 use c8y_api::smartrest::error::SmartRestDeserializerError;
 use c8y_api::smartrest::inventory::child_device_creation_message;
 use c8y_api::smartrest::inventory::service_creation_message;
-use c8y_api::smartrest::message::collect_smartrest_messages;
+use c8y_api::smartrest::message::collect_c8y_messages;
 use c8y_api::smartrest::message::get_failure_reason_for_smartrest;
 use c8y_api::smartrest::message::get_smartrest_device_id;
 use c8y_api::smartrest::message::get_smartrest_template_id;
@@ -634,7 +634,7 @@ impl CumulocityConverter {
     ) -> Result<Vec<MqttMessage>, ConversionError> {
         // JSON over MQTT messages on c8y/devicecontrol/notifications can contain multiple operations in a single MQTT
         // message, so split them
-        let operation_payloads = collect_smartrest_messages(message.payload_str()?);
+        let operation_payloads = collect_c8y_messages(message.payload_str()?);
 
         let mut output = vec![];
         for operation_payload in operation_payloads {
@@ -884,7 +884,7 @@ impl CumulocityConverter {
         message: &MqttMessage,
     ) -> Result<Vec<MqttMessage>, ConversionError> {
         let mut output: Vec<MqttMessage> = Vec::new();
-        for smartrest_message in collect_smartrest_messages(message.payload_str()?) {
+        for smartrest_message in collect_c8y_messages(message.payload_str()?) {
             let result = self.process_smartrest(smartrest_message.as_str()).await;
             let mut msgs = self.handle_c8y_operation_result(&result, None);
             output.append(&mut msgs)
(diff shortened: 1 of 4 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -632,24 +632,36 @@ impl CumulocityConverter {
         &mut self,
         message: &MqttMessage,
     ) -> Result<Vec<MqttMessage>, ConversionError> {
-        let operation = C8yOperation::from_json(message.payload.as_str()?)?;
-        let device_xid = operation.external_source.external_id;
-        let cmd_id = self.command_id.new_id_with_str(&operation.op_id);
+        // Smartrest messages on c8y/devicecontrol/notifications can contain multiple operations in a single MQTT
+        // message, so split them
```
rina23q, line 636: Terminoligy correction. The messages arrving on `c8y/devicecontrol/notifications` is **JSON over MQTT** messages. **SmartREST** is the comma separated csv-like messages.

```suggestion
        // JSON over MQTT messages on c8y/devicecontrol/notifications can contain multiple operations in a single MQTT
        // message, so split them
```

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -632,24 +632,36 @@ impl CumulocityConverter {
...
+        let operation_payloads = collect_smartrest_messages(message.payload_str()?);
```
rina23q, line 637: I was surprised that `collect_smartrest_messages()` works here. However, after seeing the contents of the function, it makes sense. That means, the function should be renamed now, otherwise it's very misleading. Maybe `collect_c8y_messages()` or so on.

rina23q (commented):
Some renaming required.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: Right, it's not Smartrest, so probably it would also be fine to just split on `\n`, right? As JSON over MQTT doesn't have these escaping problems the function was originally made for?

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -632,24 +632,36 @@ impl CumulocityConverter {
...
+                    operation.op_id.clone(),
+                    &operation.extras,
+                    &operation_message,
+                )
+                .await;
+            let result = self.handle_c8y_operation_result(&result, Some(operation.op_id));
+            output.extend(result);
+        }
 
         Ok(output)
```
rina23q, line 666: My opinion for the future development.

What I don't like from the current way (due to the converter design) is, this return value Ok(output) is a vector of MQTT messages that will be published later at the same time. If 5 operations are included in one message from c8y, 5 of 504 messages will be sent together. But it's definitely out of scope to fix now.

rina23q (approved):
LGTM

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: > What I don't like from the current way (due to the converter design) is, this return value Ok(output) is a vector of MQTT messages that will be published later at the same time.  If 5 operations are included in one message from c8y, 5 of 504 messages will be sent together.

What don't you like in the current design? That the value returned by this method is a vector of MQTT messages or that the five 504 messages are not packed into a single MQTT message? The later can be implemented using the former method signature.

didier-wenzek (approved):
Approved

crates/extensions/c8y_mapper_ext/src/converter.rs
rina23q: `504` changes an operation state to executing in c8y.

Let's say there are 3 operation (Op_A, Op_B, Op_C) are delivered at the same.
Ideally,
1. `Op_A` start proceeding -> Publish `504` for Op_A
2. `Op_B` start proceeding -> Publish `504` for Op_B
3. `Op_C` start proceeding -> Publish `504` for Op_C

However, with the current design, when all of `Op_A`, `Op_B`, and `Op_C` start proceeding, mapper publishes three `504` messages for these 3 operations. 

In short, what I don't like is the timing the `504` message is published by mapper. Not critical, but minor point.

reubenmiller: Though this doesn't occur for operations being processed by a workflow right (as the workflow execution is controlled by the tedge-agent, and currently only processes one type of operation at a time of the same type).

Issue #3305, opened 2024-12-17
tedge-agent ignores in progress workflows without a version after being updated


reubenmiller:
**Describe the bug**

The tedge-agent 1.4.0 ignores in-progress commands if the command does not contain any version information, i.e. the command instance was created by a tedge-agent version < 1.4.0.

This was discovered after updating the [tedge-container-bundle](https://github.com/thin-edge/tedge-container-bundle) project which uses a customer `software_update` workflow to perform a self-update of the container (which includes tedge-agent). The last release of the `tedge-container-bundle` uses thin-edge.io 1.3.1, and during the self-update process, when the new version of the container launch, which includes tedge-agent 1.4.0, the in-progress operation is being ignored. The operation will then remain in the state it was left in before the new container was spawned.

**To Reproduce**

A system test has been added to the thin-edge.io project which includes a workflow to reproduce this issue, so please refer to the [PR](https://github.com/thin-edge/thin-edge.io/pull/3306) for all the details. The process is roughly as follows:

1. Install thin-edge.io 1.3.1

2. Override the default `software_update` workflow which does a self-update of the tedge-agent

3. Create a software update operation from the cloud

The above procedure will result in the software update operation being stuck in the EXECUTING state, and it will never complete.

**Expected behavior**

Command instances which don't have a version should still be respected by the tedge-agent, and the workflow matching the same name (as denoted by the `operation = <name>` field) should be used to continue processing the command instance.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Any`|
|Hardware [incl. revision]|`Any`|
|System-Architecture|`Any`|
|thin-edge.io version|`tedge 1.4.0`|
|mosquitto version|`N/A`|

**Additional context**

reubenmiller:
The CI tests in the [tedge-container-bundle](https://github.com/thin-edge/tedge-container-bundle/actions/runs/12412076916) project now pass with the latest changes from the main branch (after the two PRs have been merged).

A new thin-edge.io release 1.4.1 will be done shortly including this fix.

PR #3307, opened 2024-12-19
fix: Persist workflow definition of pending commands on start

didier-wenzek:
## Proposed changes

When the agent restarts and loads pending operations, these operations might not have been assign a workflow version. This is notably the case after a self update from a previous version of the agent that was oblivious of workflow versions.

In such a case, the agent assumes the current user-defined version of the workflow is the version to be used for the pending command. The workflow definition is persisted in `workflows-in-use` directory and assigned to the command.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3305

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/operation_workflows/persist.rs
+++ b/crates/core/tedge_agent/src/operation_workflows/persist.rs
@@ -346,7 +346,25 @@ impl WorkflowRepository {
         None
     }
 
-    pub fn load_pending_commands(&mut self, commands: CommandBoard) -> Vec<GenericCommandState> {
+    pub async fn load_pending_commands(
+        &mut self,
+        mut commands: CommandBoard,
+    ) -> Vec<GenericCommandState> {
+        // If the resumed commands have been triggered by an agent without workflow version management
+        // then these commands are assigned the current version of the operation workflow.
+        // These currents versions have also to be marked as in use and persisted.
+        for (_, ref mut command) in commands.iter_mut() {
+            if command.workflow_version().is_none() {
+                if let Some(operation) = command.operation() {
+                    if let Some(current_version) = self.workflows.use_current_version(&operation) {
+                        self.persist_workflow_definition(&operation, &current_version)
+                            .await;
+                        *command = command.clone().set_workflow_version(&current_version);
+                    }
+                }
+            }
+        }
+
         self.workflows.load_pending_commands(commands)
     }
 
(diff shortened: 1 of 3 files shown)


crates/core/tedge_agent/src/operation_workflows/persist.rs
```diff
@@ -346,7 +346,25 @@ impl WorkflowRepository {
...
+        // If the resumed commands have been triggered by an agent without workflow version management
+        // then these commands are assigned the current version of the operation workflow.
+        // These currents versions have also to be marked as in use and persisted.
+        for (_, ref mut command) in commands.iter_mut() {
+            if command.workflow_version().is_none() {
+                if let Some(operation) = command.operation() {
+                    if let Some(current_version) = self.workflows.use_current_version(&operation) {
+                        self.persist_workflow_definition(&operation, &current_version)
+                            .await;
+                        *command = command.clone().set_workflow_version(&current_version);
```
jarhodes314, line 362: I'm puzzled by this API where we seemingly are cloning `command` just to mutate it and return it from `set_workflow_version`?

crates/core/tedge_api/src/workflow/supervisor.rs
```diff
@@ -150,6 +132,16 @@ impl WorkflowSupervisor {
...
     }
 
+    /// Mark the current version of an operation workflow as being in use.
+    ///
+    /// Return the current version if any.
+    pub fn use_current_version(&mut self, operation: &OperationName) -> Option<WorkflowVersion> {
+        match self.workflows.get_mut(&operation.as_str().into()) {
+            Some(versions) => versions.use_current_version().cloned(),
+            None => None,
+        }
```
jarhodes314, line 142:
```suggestion
       self.workflows.get_mut(&operation.as_str().into())?.use_current_version().cloned()
```

jarhodes314 (approved):
I'm a bit puzzled by some of what's going on with the APIs, but believe this functions correctly

crates/core/tedge_agent/src/operation_workflows/persist.rs
didier-wenzek: You can be!

Not only `set_workflow_version` but all the updating methods of `GenericCommandState` are consuming `self`. 
Fixing that would not be complicated, but I prefer to fix this bug asap.

crates/core/tedge_api/src/workflow/supervisor.rs
didier-wenzek: Good point. I will fixed that in a follow-up PR.

Issue #2696, opened 2024-02-09
Make tedge-mapper and tedge-agent only available as symlink to tedge 


didier-wenzek:
Thin-edge packaging only includes the `tedge` executable file giving access to the agent, mappers and plugins using symlinks.

However, this is not the case in development where executable files are built for the agent, the mapper and the plugin.
This is misleading and consume resources for nothing.

I propose to
- remove `main.rs` for the agent and the mapper, and all the plugins that can be launched using a symlink on `tedge`.
- create the appropriate symlinks using `just`

reubenmiller:
I'm wondering whether there would be any value in providing an easier way to launch the different components directly via `tedge plugins <sub_component>`, e.g. `tedge plugins tedge-agent` or `tedge plugins tedge-mapper c8y`.

For instance you can enable a similar behaviour by just renaming the `tedge` binary to something other than `tedge`, e.g. `mv tedge tedge2`, then you can see the sub components using `tedge2 --help` (although this is definitely unexpected, so having a more formal way for launching sub components without requiring the symlink might be more useful).

```
# tedge2 --help
tedge is the cli tool for thin-edge.io

Usage: tedge2 <COMMAND>

Commands:
  tedge                     tedge is the cli tool for thin-edge.io
  tedge-mapper              tedge-mapper is the mapper that translates thin-edge.io data model to c8y/az data model.
  tedge-agent               tedge-agent interacts with a Cloud Mapper and one or more Software Plugins
  c8y-firmware-plugin       Thin-edge device firmware management for Cumulocity
  tedge-watchdog            tedge-watchdog checks the health of all the thin-edge.io components/services.
  c8y-remote-access-plugin  Thin-edge.io plugin for the Cumulocity IoT's Cloud Remote Access feature
  tedge-write               A binary used for writing to files which `tedge` user does not have write permissions for, using sudo
  help                      Print this message or the help of the given subcommand(s)

Options:
  -h, --help     Print help
  -V, --version  Print version
```

didier-wenzek:
released with 1.4.2

PR #3313, opened 2025-01-03
feat: Trigger tedge multicall with symlinks as well as sub-commands

didier-wenzek:
## Proposed changes

Simplify the way `tedge-mapper`, `tedge-agent` and `tedge-write` are launched: it's no more mandatory to create symlinks to `tedge` with appropriate names.

The following are now equivalent (for the mapper, but also the agent and  all tedge services and plugins ):

- `tedge run tedge-mapper c8y`
- `tedge-mapper c8y` where `tedge-mapper` is a copy or a symlink of `tedge`

The former `tedge` cli can be invoked as before:

```
$ tedge help
tedge is the cli tool for thin-edge.io

Usage: tedge [OPTIONS] <COMMAND>

Commands:
  init             Initialize Thin Edge
  cert             Create and manage device certificate
  ...
  run              Run thin-edge services and plugins
  ...
```

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2696

## Checklist

## Further comments

This PR supersedes https://github.com/thin-edge/thin-edge.io/pull/3311

### DIFF
--- a/crates/core/tedge/src/cli/mod.rs
+++ b/crates/core/tedge/src/cli/mod.rs
@@ -113,6 +113,15 @@ pub enum TEdgeOpt {
     /// Publish a message on a topic and subscribe a topic.
     #[clap(subcommand)]
     Mqtt(mqtt::TEdgeMqttCli),
+
+    /// Run thin-edge services and plugins
+    Run(ComponentOpt),
+}
+
+#[derive(Debug, clap::Parser)]
+pub struct ComponentOpt {
+    #[clap(subcommand)]
+    pub component: Component,
 }
 
 fn styles() -> clap::builder::Styles {
@@ -174,6 +183,10 @@ impl BuildCommand for TEdgeOpt {
             TEdgeOpt::RefreshBridges => RefreshBridgesCmd::new(&context).map(Command::into_boxed),
             TEdgeOpt::Mqtt(opt) => opt.build_command(context),
             TEdgeOpt::Reconnect(opt) => opt.build_command(context),
+            TEdgeOpt::Run(_) => {
+                // This method has to be kept in sync with tedge::redirect_if_multicall()
+                panic!("tedge mapper|agent|write commands are launched as multicall")
+            }
         }
     }
 }
--- a/crates/core/tedge/src/cli/mod.rs
+++ b/crates/core/tedge/src/cli/mod.rs
@@ -7,6 +7,7 @@ use c8y_firmware_plugin::FirmwarePluginOpt;
 use c8y_remote_access_plugin::C8yRemoteAccessPluginOpt;
 pub use connect::*;
 use tedge_agent::AgentOpt;
+use tedge_apt_plugin::AptCli;
 use tedge_config::cli::CommonArgs;
 use tedge_mapper::MapperOpt;
 use tedge_watchdog::WatchdogOpt;
@@ -51,15 +52,18 @@ pub enum TEdgeOptMulticall {
 
 #[derive(clap::Parser, Debug)]
 pub enum Component {
-    TedgeMapper(MapperOpt),
+    C8yFirmwarePlugin(FirmwarePluginOpt),
+
+    C8yRemoteAccessPlugin(C8yRemoteAccessPluginOpt),
 
     TedgeAgent(AgentOpt),
 
-    C8yFirmwarePlugin(FirmwarePluginOpt),
+    #[clap(alias = "apt")]
+    TedgeAptPlugin(AptCli),
+
+    TedgeMapper(MapperOpt),
 
     TedgeWatchdog(WatchdogOpt),
-
-    C8yRemoteAccessPlugin(C8yRemoteAccessPluginOpt),
 
     TedgeWrite(TedgeWriteOpt),
 }
--- a/crates/core/tedge/src/cli/mod.rs
+++ b/crates/core/tedge/src/cli/mod.rs
@@ -36,6 +36,7 @@ mod upload;
     multicall(true),
 )]
 pub enum TEdgeOptMulticall {
+    /// Command line interface to interact with thin-edge.io
     Tedge {
         #[clap(subcommand)]
         cmd: TEdgeOpt,
--- a/crates/core/tedge/src/main.rs
+++ b/crates/core/tedge/src/main.rs
@@ -32,11 +32,6 @@ static ALLOCATOR: Cap<alloc::System> = Cap::new(alloc::System, usize::MAX);
 fn main() -> anyhow::Result<()> {
     let executable_name = executable_name();
 
-    if matches!(executable_name.as_deref(), Some("apt" | "tedge-apt-plugin")) {
-        let try_opt = AptCli::try_parse();
-        tedge_apt_plugin::run_and_exit(try_opt);
-    }
-
     let opt = parse_multicall(&executable_name, std::env::args_os());
     match opt {
         TEdgeOptMulticall::Component(Component::TedgeMapper(opt)) => {
@@ -64,6 +59,9 @@ fn main() -> anyhow::Result<()> {
             block_on(tedge_watchdog::run(opt))
         }
         TEdgeOptMulticall::Component(Component::TedgeWrite(opt)) => tedge_write::bin::run(opt),
+        TEdgeOptMulticall::Component(Component::TedgeAptPlugin(opt)) => {
+            tedge_apt_plugin::run_and_exit(opt)
+        }
         TEdgeOptMulticall::Tedge { cmd, common } => {
             let tedge_config_location =
                 tedge_config::TEdgeConfigLocation::from_custom_root(&common.config_dir);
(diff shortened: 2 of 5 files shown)


crates/core/tedge/src/main.rs
```diff
@@ -120,20 +124,39 @@ fn executable_name() -> Option<String> {
...
     }
 }
+
+// Transform `tedge mapper|agent|write` commands into multicalls
+//
+// This method has to be kept in sync with TEdgeOpt::build_command
+fn redirect_if_multicall(cmd: TEdgeOpt, common: CommonArgs) -> TEdgeOptMulticall {
+    match cmd {
+        TEdgeOpt::Mapper(opt) => {
+            TEdgeOptMulticall::Component(Component::TedgeMapper(opt.with_common_args(common)))
```
didier-wenzek, line 152: We have to be more cautious here: there is a risk of confusion as `--tedge-config` is provided twice to `clap`.

crates/core/tedge/src/cli/mod.rs
```diff
@@ -36,6 +36,7 @@ mod upload;
     multicall(true),
 )]
 pub enum TEdgeOptMulticall {
+    /// Command line interface to interact with thin-edge.io
```
didier-wenzek, line 39: The docs have to be updated too, to be in sync with the updated `tedge help`

crates/core/tedge/src/main.rs
reubenmiller: I assume you mean `--config-dir`

crates/core/tedge/src/cli/mod.rs
didier-wenzek: See https://github.com/thin-edge/thin-edge.io/pull/3313/commits/6a61c483664967daa102bc216fc4a889c9fffbbd

crates/core/tedge/src/main.rs
didier-wenzek: When a given global argument (as `--config-dir`) is used twice by the command struct (as here where `common` can be accessed directly (`common`) or via the mapper options (`opt.common`), `clap` correctly duplicate the user provided info to all the inner components.

- Hence, there is no need to explicitly copy to common arg.
- As this clap behavior is not documented, tests have been added, see https://github.com/thin-edge/thin-edge.io/pull/3313/commits/b7353194746dbe0a86ab0a17cd48cd7aaafa5975
- There is one caveat: `clap` fails to notice that an argument is provided twice.

reubenmiller:
@didier-wenzek I think it would be useful if we make all of the binaries which can be access via symlinks also available via subcommands, but since there are a few tricker cases (e.g. c8y-firmware-plugin, c8y-remote-access-plugin), so here is the suggested structure (and the associated symlink equivalent in brackets).

* tedge
    * agent (tedge-agent)
    * mapper (tedge-mapper)
    * watchdog (tedge-watchdog)
    * sm-plugins
        * apt (tedge-apt-plugin)
    * plugins
        * c8y-remote-access-plugin (c8y-remote-access-plugin)
        * c8y-firmware-plugin (c8y-firmware-plugin)
    * utils
        * write (tedge-write)

didier-wenzek:
> The second approach with `tedge run xxx` is simpler. I will go with that one.

Done with https://github.com/thin-edge/thin-edge.io/pull/3313/commits/2bada64d93fb84428e4dd415ca1de8c169e7d87b
the docs being updated by https://github.com/thin-edge/thin-edge.io/pull/3313/commits/2c73956c467c272189edd986f9d78f767251e2f5

reubenmiller (changes requested):
Just two minor points:

1. `tedge-apt-plugin` is missing under `tedge run`
2. Can we sort the subcommands under `tedge run`? I don't mind having `help` at the end, however the current order seems to be manually defined.

didier-wenzek:
>     1. `tedge-apt-plugin` is missing under `tedge run`

For some reasons I have to understand better, `tedge-apt-plugin`  is a special case: https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/main.rs#L29

>     2. Can we sort the subcommands under `tedge run`? I don't mind having `help` at the end, however the current order seems to be manually defined.

Indeed, these sub-commands are listed here: https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/cli/mod.rs#L52

Any order in mind?

didier-wenzek:
> > ```
> > 1. `tedge-apt-plugin` is missing under `tedge run`
> > ```
> 
> For some reasons I have to understand better, `tedge-apt-plugin` is a special case: https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/main.rs#L29

This is to exit with a status code of 1 and not 2 when the command line cannot be parsed:
https://github.com/thin-edge/thin-edge.io/blob/main/plugins/tedge_apt_plugin/src/lib.rs#L354

Bravo555 (approved):
Nice improvement, I won't have to manually create dummy binary files for different components to do `cargo run` anymore!

That said, looking at `tedge run --help` there are some inconsistencies that could be fixed:
```
Run thin-edge services and plugins

Usage: tedge run [OPTIONS] <COMMAND>

Commands:
  tedge-mapper              tedge-mapper is the mapper that translates thin-edge.io data model to c8y/az data model.
  tedge-agent               tedge-agent interacts with a Cloud Mapper and one or more Software Plugins
  c8y-firmware-plugin       Thin-edge device firmware management for Cumulocity
  tedge-watchdog            tedge-watchdog checks the health of all the thin-edge.io components/services.
  c8y-remote-access-plugin  Thin-edge.io plugin for the Cumulocity IoT's Cloud Remote Access feature
  tedge-write               tee-like helper for writing to files which `tedge` user does not have write permissions to
  help                      Print this message or the help of the given subcommand(s)

snip...
```

- `tedge-mapper` mentions `c8y/az`, but not `aws`
- the style is inconsistent (capitalization, if the sentence is terminated by a `.`)

These descriptions are taken from `package.description` field of `Cargo.toml` of these crates, so they should either be fixed there, or we could provide the descriptions directly in code by adding doc comments to the `Component` enum variants; however I think updating `Cargo.toml`s would be preferable.

Doing this however would be best in another PR, so approving this one.

didier-wenzek:
> Just two minor points:
> 
>     1. `tedge-apt-plugin` is missing under `tedge run`
> 
>     2. Can we sort the subcommands under `tedge run`? I don't mind having `help` at the end, however the current order seems to be manually defined.

Addressed by https://github.com/thin-edge/thin-edge.io/pull/3313/commits/1bdac4386d487eae3ee3b41e0797f426c9473b7f

crates/core/tedge/src/main.rs
```diff
@@ -120,20 +124,94 @@ fn executable_name() -> Option<String> {
...
+        // the apt plugin must be treated apart
+        // as we want to exit 1 and not 2 when the command line cannot be parsed
+        match AptCli::try_parse() {
+            Ok(apt) => return TEdgeOptMulticall::Component(Component::TedgeAptPlugin(apt)),
+            Err(e) => {
+                eprintln!("{}", RichFormatter::format_error(&e));
+                std::process::exit(1);
+            }
+        }
+    }
```
reubenmiller, line 142: @didier-wenzek How can I verify this? Which command is meant to return an exit code 1 instead of 2?

```sh
cargo build
./target/debug/tedge run tedge-apt-plugin 
```

didier-wenzek: As of now, this only works when using multicall:

```
$ cp target/debug/tedge /tmp/tedge-apt-plugin
$ /tmp/tedge-apt-plugin list --invalid    
error: unexpected argument '--invalid' found

Usage: tedge-apt-plugin list [OPTIONS]

For more information, try '--help'.

$ echo $?
1
```

I tried different ways to overwrite `clap` 's behavior to have something generic or at least working for `tedge run tedge-apt-plugin` but failed.

- [`try_get_matches`](https://docs.rs/clap/latest/clap/struct.Command.html#method.try_get_matches) can be used to avoid the default behavior of [`get_matches`](https://docs.rs/clap/latest/clap/struct.Command.html#method.get_matches) that is to [exit with `2` on parsing errors](https://docs.rs/clap/latest/clap/error/struct.Error.html#method.exit)
- but then we have to manually handle `--help` and `--version`
- and, worse, this messes up the output of `tedge --help`, `clap` being confused by multicall and [returning the help for an unknown executable name](https://github.com/thin-edge/thin-edge.io/issues/2696#issuecomment-1943703691)
- I also tried [`TEdgeOptMulticall::try_parse()`](https://docs.rs/clap/latest/clap/trait.Parser.html#method.try_parse),
- but, again,   `clap` is confused and messes up the help of `tedge`.

I don't think it's worth it spending more time on that.

reubenmiller: yeah I'm ok with that. Thanks for the explanation.

reubenmiller (approved):
Approved

Issue #3315, opened 2025-01-06
`tedge cert create-csr c8y --profile <profile>` doesn't use the CN of the certificate of the cloud profile


rina23q:
**Describe the bug**
`tedge cert create-csr c8y --profile <profile>` tries to use the CN of the main device certificate and private key instead of the ones from the cloud profile.

From the code, the behavour is obvious (wrong).
https://github.com/thin-edge/thin-edge.io/blob/91c4cb508e5e9f80f99cfdd4523d98a11641b0dd/crates/core/tedge/src/cli/certificate/cli.rs#L97-L100

**To Reproduce**
Create a main device certificate.
```
tedge cert create --device-id aaa111
```

Create another device certificate for cloud profile, then create csr.

```
tedge config set c8y.device.cert_path --profile second /etc/tedge/device-certs/tedge@second-certificate.pem
tedge config set c8y.device.key_path --profile second /etc/tedge/device-certs/tedge@second-key.pem
tedge cert create --device-id  bbb222 c8y --profile second
tedge cert create-csr c8y --profile second
```

Run
```
openssl req -noout -subject -in /etc/tedge/device-certs/tedge.csr
```

Then the output has the CN of the mail device certificate, not the one from the cloud profile. It should be `bbb222`.
```
subject=CN = aaa111, O = Thin Edge, OU = Test Device
```

Also, if the main device certificate and private key don't exist, it returns an error.
```
root@0042c94c79fa:/setup# tedge cert remove
Certificate was successfully removed
root@0042c94c79fa:/setup# tedge cert create-csr c8y --profile second
Error: missing configuration parameter

Caused by:
    Config value device.id, cannot be read: The device id is read from the device certificate.
    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`. 
```

**Expected behavior**
`tedge cert create-csr c8y --profile <profile>` should consume the certificate and private key from the profiled config.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [1.4.1]

**Additional context**

PR #3316, opened 2025-01-06
fix: tedge cert create-csr should use cloud profile's CN

rina23q:
## Proposed changes
`tedge cert create-csr <cloud> --profile <profile>` should use the CN of the corresponding device certificate/private key if `--device-id` is not provided.

## Types of changes

## Paste Link to the issue
#3315

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/certificate/cli.rs
+++ b/crates/core/tedge/src/cli/certificate/cli.rs
@@ -3,6 +3,8 @@ use crate::cli::common::CloudArg;
 use camino::Utf8PathBuf;
 use tedge_config::OptionalConfigError;
 use tedge_config::ProfileName;
+use tedge_config::ReadError;
+use tedge_config::TEdgeConfig;
 
 use super::create::CreateCertCmd;
 use super::create_csr::CreateCsrCmd;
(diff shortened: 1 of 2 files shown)


crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -97,7 +97,12 @@ impl BuildCommand for TEdgeCertCli {
                 let cloud: Option<Cloud> = cloud.map(<_>::try_into).transpose()?;
 
                 // Use the current device id if no id is provided
-                let id = id.unwrap_or(get_device_id_from_cloud(&config, &cloud)?);
+                let id = if let Some(id) = id {
```
rina23q, line 100: At first, I wrote this line as below.

```rust
let id = id.unwrap_or(get_device_id_from_cloud(&config, &cloud)?);
```

However, it causes a problem when certificate doesn't exist but id has `Some()` since`unwrap_or()` evaluates the default always regardless of id is `Some` or `None`.

Bravo555: Indeed, and `unwrap_or_else`, makes it so you can't use `?` from within the closure, which is a bit of a pain.

Bravo555 (approved):

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -186,3 +191,17 @@ pub enum UploadCertCli {
         profile: Option<ProfileName>,
     },
 }
+
+fn get_device_id_from_cloud(
```
albinsuresh, line 195:
```suggestion
fn get_device_id_from_config(
```
Since it's not always the id from a cloud profile but could be the main `device.id` as well.

albinsuresh (approved):
LGTM

PR #3318, opened 2025-01-08
refactor: support writable device ids

jarhodes314:
## Proposed changes
This adds support for a writable device ID to aid @rina23q's work. I'm marking this as a refactoring as it doesn't really perform the task in itself, although I have modified `tedge_config.rs` to use the new feature for verifying it actually works as intended.

Still TODO:
- [ ] Support a default value, so `c8y.device.id` defaults to `device.id`
- [ ] Make the device ID logic actually do something useful
- [ ] Check the test names are appropriate

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -428,15 +428,9 @@ define_tedge_config! {
     device: {
         /// Identifier of the device within the fleet. It must be globally
         /// unique and is derived from the device certificate.
-        #[tedge_config(readonly(
-            write_error = "\
-                The device id is read from the device certificate and cannot be set directly.\n\
-                To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-            function = "device_id",
-        ))]
+        #[tedge_config(reader(function = "device_id"))]
         #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
         #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
-        #[tedge_config(reader(private))]
         #[doku(as = "String")]
         id: Result<String, ReadError>,
 
@@ -489,14 +483,10 @@ define_tedge_config! {
         device: {
             /// Identifier of the device within the fleet. It must be globally
             /// unique and is derived from the device certificate.
-            #[tedge_config(readonly(
-                write_error = "\
-                    The device id is read from the device certificate and cannot be set directly.\n\
-                    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-                function = "c8y_device_id",
-            ))]
+            #[tedge_config(reader(function = "c8y_device_id"))]
+            // TODO make this work
+            // #[tedge_config(default(from_optional_key = "device.id"))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
-            #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
             #[doku(as = "String")]
             id: Result<String, ReadError>,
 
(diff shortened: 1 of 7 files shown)


crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -1355,26 +1345,32 @@ fn default_http_bind_address(dto: &TEdgeConfigDto) -> IpAddr {
...
+        .map_err(|err| cert_error_into_config_error(ReadableKey::DeviceId.to_cow_str(), err))?;
     Ok(device_id)
 }
 
-fn device_id(device: &TEdgeConfigReaderDevice) -> Result<String, ReadError> {
+fn device_id(
+    device: &TEdgeConfigReaderDevice,
+    dto_value: &OptionalConfig<String>,
+) -> Result<String, ReadError> {
     device_id_from_cert(&device.cert_path)
```
rina23q, line 1359: If `device.id` is set (= the value exists in dto), the value should be returned. It should be something like below. This applies to all functions for each cloud (c8y, az, aws).

```suggestion
match device.id.or_none() {
    Some(id) => Ok(id.to_string()),
    None => device_id_from_cert(&device.cert_path)
}
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -489,14 +483,10 @@ define_tedge_config! {
...
             /// unique and is derived from the device certificate.
-            #[tedge_config(readonly(
-                write_error = "\
-                    The device id is read from the device certificate and cannot be set directly.\n\
-                    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-                function = "c8y_device_id",
-            ))]
+            #[tedge_config(reader(function = "c8y_device_id"))]
+            // TODO make this work
+            // #[tedge_config(default(from_optional_key = "device.id"))]
```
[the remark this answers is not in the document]
rina23q: I was also thinking that the `default` from `device.id` will work partially. This is the scenario which doesn't work.

`c8y.device.cert_path` and `device.id` exists, but `c8y.device.id` is missing. Then, I expect `c8y.device.id` will return the CN of the `c8y.device.cert_path`. However, my guess the config will return the value of `device.id` due to the default.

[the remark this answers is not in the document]
rina23q: > `c8y.device.cert_path` and `device.id` exists, but `c8y.device.id` is missing. Then, I expect c8y.device.id will return the CN of the c8y.device.cert_path. However, my guess the config will return the value of device.id due to the default.

As we discussed today, this is a corner case. It's fine that `c8y.device.id` returns the value of `device.id` instead of its CN. We just have to document that user should set `c8y.device.id` explicitly in that case.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -428,15 +428,9 @@ define_tedge_config! {
     device: {
         /// Identifier of the device within the fleet. It must be globally
         /// unique and is derived from the device certificate.
-        #[tedge_config(readonly(
-            write_error = "\
-                The device id is read from the device certificate and cannot be set directly.\n\
-                To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-            function = "device_id",
-        ))]
+        #[tedge_config(reader(function = "device_id", private))]
```
rina23q, line 431: There is one usage of direct access to `device.id`, which is `tedge cert create-csr`. This can assume CN is from `device.id` if `--device-id` is not given. Actually conflicting the one I've changed very recently in #3316.

Either removing `private` from here, or force users to input `--device-id` for `tedge cert create-csr`? Any idea?

```rust
fn get_device_id_from_config(
    config: &TEdgeConfig,
    cloud: &Option<Cloud>,
) -> Result<String, ReadError> {
    let id = match cloud {
        None => config.device.id(),
        Some(Cloud::C8y(profile)) => config.c8y.try_get(profile.as_deref())?.device.id(),
        Some(Cloud::Azure(profile)) => config.az.try_get(profile.as_deref())?.device.id(),
        Some(Cloud::Aws(profile)) => config.aws.try_get(profile.as_deref())?.device.id(),
    }?
    .to_owned();
    Ok(id)
}
```

crates/common/tedge_config_macros/impl/src/input/parse.rs
```diff
@@ -5,7 +5,7 @@
 
 // FIXME: if let can be simplified with `.unwrap_or_default()`
 //        for all `#[darling(default)]`
-#![allow(clippy::manual_unwrap_or_default)]
+// #![allow(clippy::manual_unwrap_or_default)]
```
rina23q, line 8: This must be fixed or enable the `allow`. There are many clippy warnings.

crates/common/tedge_config_macros/Cargo.toml
```diff
@@ -21,6 +21,7 @@ tracing = { workspace = true }
 url = { workspace = true }
 
 [dev-dependencies]
+certificate = { workspace = true, features = ["reqwest"] }
```
rina23q, line 24: `Cargo.lock` should be added to this PR as well because of this change.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
jarhodes314: We could move this function into `tedge_config.rs` (it could be a method on `TEdgeConfigReader`), that module is able to access the private field.

rina23q: > We could move this function into `tedge_config.rs` (it could be a method on `TEdgeConfigReader`), that module is able to access the private field.

It sounds the right direction 👍

jarhodes314:
Superseded by #3326

Issue #3323, opened 2025-01-09
tedge config does not support using a symlink to tedge.toml as the symlink is replaced with a file on config updates


reubenmiller:
**Describe the bug**

When using a symlink for `/etc/tedge/tedge.toml`, the symlink is overwritten by thin-edge.io when the configuration file is modified via `tedge config`.

This may be more of a feature improvement, but since using symlinks is fairly standard in Linux, thin-edge.io should respect them.

The use-case where this came across is in the [tedge-container-bundle](https://github.com/thin-edge/tedge-container-bundle) project where a new PR was being developed to persist the `tedge.toml` across container update where the default tedge.toml location would be a symlink pointing to a persistent volume. A volume is not used for the entire `/etc/tedge` folder so that workflows, operation handlers etc. can be updated by container updates (which would not be the case if a named volume would be used for the entire `/etc/tedge` folder).

**To Reproduce**

1. Create a symlink for the tedge.toml file

    ```sh
    ln -sf /data/tedge/tedge.toml /etc/tedge/tedge.toml
    touch /data/tedge/tedge.toml
    ```

    Then verify the symlink exists (used for comparison later on)

    ```sh
    ls -l /etc/tedge/tedge.toml
    ```

    ```sh
    $ ls -l /etc/tedge/tedge.toml
    lrwxrwxrwx    1 tedge    tedge           22 Jan  9 19:23 /etc/tedge/tedge.toml -> /data/tedge/tedge.toml
    ```

2. Run `tedge config list`, this will trigger a writing of the tedge.toml to add the missing version info

3. Check if the tedge.toml is still a symlink (it is expected that it still should be)

    ```sh
    ls -l /etc/tedge/tedge.toml
    ```

    ```sh
    $ ls -l /etc/tedge/tedge.toml
    -rw-r--r--    1 tedge    tedge           31 Jan  9 19:21 /etc/tedge/tedge.toml
    ```

**Expected behavior**

If the `/etc/tedge/tedge.toml` file is a symlink, then it should be preserved, and the target file (where the symlink is pointing to) should be updated.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `any`
 - Hardware [incl. revision]: `any`
 - System-Architecture [e.g. result of "uname -a"]: `any`
 - thin-edge.io version [e.g. 0.1.0]: `1.4.1`

**Additional context**

reubenmiller:
I suspect that the "atomic file write" method is not symlink aware.

PR #3324, opened 2025-01-09
fix: support symlinks in atomic file write methods

reubenmiller:
## Proposed changes

Support resolving symlink targets files when calling `atomically_write_file_sync` and `atomically_write_file_async`.

Note: A symlinks target file will only be resolved if the target file exists, otherwise the path will be the symlink itself.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3323

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/fs.rs
+++ b/crates/common/tedge_utils/src/fs.rs
@@ -48,7 +48,11 @@ pub fn atomically_write_file_sync(
     dest: impl AsRef<Path>,
     mut reader: impl Read,
 ) -> Result<(), AtomFileError> {
-    let dest_dir = parent_dir(dest.as_ref());
+    let dest = dest.as_ref();
+    // resolve path (including symlinks)
+    // if the symlink doesn't exist, (attempt to) create the file it points to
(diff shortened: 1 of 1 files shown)


jarhodes314 (commented):
I believe this would still clobber the symlink in the case where:
- we have write access to the directory containing the symlink
- we don't have read access to the symlink itself

Is there a way we can use file metadata to establish if the file is a symlink and fail if we happen to be in this case? Or do we actually care about solving this particular edge case?

reubenmiller:
> I believe this would still clobber the symlink in the case where:
> 
> * we have write access to the directory containing the symlink
> * we don't have read access to the symlink itself
> 
> Is there a way we can use file metadata to establish if the file is a symlink and fail if we happen to be in this case? Or do we actually care about solving this particular edge case?

Yes you're suggestions are definitely valid (the part about symlinks which don't point to an existing file also surprised me). Would you be able to take over this PR and do the implementation as per your suggestion?

jarhodes314:
> Approved. Better to squash the commits.

Now squashed

PR #3325, opened 2025-01-10
feat: audit journal of commands and trace of actor messages

didier-wenzek:
## Proposed changes

Experiment with the idea of

- an audit journal logging user actions and commands
- and a log tracing all the actor messages. 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -181,6 +181,7 @@ tokio-util = { version = "0.7", features = ["codec"] }
 toml = "0.8"
 tower = "0.4"
 tracing = { version = "0.1", features = ["attributes", "log"] }
+tracing-appender = "0.2"
 tracing-subscriber = { version = "0.3", features = ["time", "env-filter"] }
 try-traits = "0.1"
 tungstenite = "0.20"
--- a/crates/common/tedge_config/src/system_services/log_config.rs
+++ b/crates/common/tedge_config/src/system_services/log_config.rs
@@ -59,8 +59,17 @@ pub fn log_init(
         .with_filter(LevelFilter::INFO)
         .with_filter(filter_fn(|metadata| metadata.target() == "Audit"));
 
+    // Actor traces
+    let trace_appender = tracing_appender::rolling::hourly("/tmp", "tedge.actors.log");
+    let trace_layer = tracing_subscriber::fmt::layer()
+        .with_writer(trace_appender)
+        .with_timer(tracing_subscriber::fmt::time::UtcTime::rfc_3339())
+        .with_filter(LevelFilter::DEBUG)
+        .with_filter(filter_fn(|metadata| metadata.target() == "Actors"));
+
     tracing_subscriber::registry()
         .with(audit_layer)
+        .with(trace_layer)
         .with(log_layer)
         .init();
 
--- a/crates/common/tedge_config/src/system_services/log_config.rs
+++ b/crates/common/tedge_config/src/system_services/log_config.rs
@@ -5,6 +5,11 @@ use crate::system_services::SystemConfig;
 use crate::system_services::SystemServiceError;
 use std::io::IsTerminal;
 use std::str::FromStr;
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -181,6 +181,7 @@ tokio-util = { version = "0.7", features = ["codec"] }
 toml = "0.8"
 tower = "0.4"
 tracing = { version = "0.1", features = ["attributes", "log"] }
+tracing-appender = "0.2"
```
didier-wenzek, line 184: Use for the rolling-file appender

crates/common/tedge_config/src/system_services/log_config.rs
```diff
@@ -28,22 +34,44 @@ pub fn log_init(
...
+            .boxed()
+    } else {
+        let log_level = get_log_level(sname, config_dir)?;
+        log_layer
+            .with_filter(LevelFilter::from_level(log_level))
+            .boxed()
+    };
+
+    // Audit journal
+    let audit_appender = tracing_appender::rolling::hourly("/tmp", "tedge.audit.log");
```
didier-wenzek, line 55: This will have to be configurable ;-)

crates/common/tedge_config/src/system_services/log_config.rs
```diff
@@ -28,22 +34,44 @@ pub fn log_init(
...
+    let audit_layer = tracing_subscriber::fmt::layer()
+        .with_writer(audit_appender)
+        .with_timer(tracing_subscriber::fmt::time::UtcTime::rfc_3339())
+        .with_filter(LevelFilter::INFO)
+        .with_filter(filter_fn(|metadata| metadata.target() == "Audit"));
+
+    // Actor traces
+    let trace_appender = tracing_appender::rolling::hourly("/tmp", "tedge.actors.log");
```
didier-wenzek, line 63: This will have to be configurable ;-)

didier-wenzek: The idea is observe & debug actors and ideally to test actors.

- [ ] Distinguish source messages (MQTT / inotify / service start / ^C) from reactions
- [ ] Add trace identifiers to follow the consequence of an input messages
- [ ] Do we have to interleave actions on the system here (HTTP requests? File Updates? System commands?)

didier-wenzek:
All the system tests are failing for the same reason: `initializing rolling file appender failed: InitError { context: "failed to create initial log file", source: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }`.

crates/common/tedge_config/src/system_services/log_config.rs
Bravo555: > Distinguish source messages (MQTT / inotify / service start / ^C) from reactions
I think ideally we should distinguish between what are strictly inputs/outputs from internal actor messages. To that effect, we do already have `MQTT_sub`, `MQTT_pub` and `MQTT_recv` targets

crates/common/tedge_config/src/system_services/log_config.rs
Bravo555: How about changes to c8y supported operations? Would show when supported commands are added/deleted and when the set is reloaded. Recently when live-debugging with a user we were confused why `c8y_Command` was not present, so it would be ideal to be able to go into the log audit to see why it was removed!

Bravo555 (commented):
We definitely need to improve and formalise our logging a little more, so I like the idea of separate journals for audited operations and actor messages. Another possible category would be also IO (network/disk or both).
For tracking the causes of events, we could definitely make better use of spans. I have one branch where I had to instrument a bit of code to track a bug, so I will also create a PR with that once I cleaned it up a bit.
I reckon there will be more changes, so will review again when the PR is ready.

didier-wenzek:
Closing as being completely outdated

PR #3326, opened 2025-01-10
feat: support read-write device.id in tedge cert/connect

rina23q:
## Proposed changes
This PR aims to open up `device.id` as read-write key. This is the last piece of supporting basic authentication, as it needs `device.id` but not need for device cert.

TODOs:

**tedge cert/connect part**
- [x] ~~`tedge cert create` writes `device.id` in `tedge.toml` explicitly.~~ Reverted.
- [x]  `tedge connect` returns an error if `device.id` mismatches the certificate's CN. The only exception is when `c8y.auth_mode` is `basic`.
- [x] Updated `smart_rest_one` system test to confirm that `tedge connect c8y` uses `device.id` directly when using basic auth.
- [x] ~~`tedge cert create-csr` also sets `device.id` if not configured.~~ Reverted.
- [x] `tedge cert create` should work without `--device-id` option when `device.id` is already set.
- [x] ~~Don't overwrite a certificate and `device.id` when the values of option `--device-id` and config `device.id` are conflicting. This case should return an error.~~ If a certificate exists already in the given path, anyway it will not be overwritten (known behavour)

**tedge_config part** (originally started in the PR #3318 by @jarhodes314 )
- [x] Introduce new read-write macro
- [x] Clippy warnings & doc test error fix
- [x] Move `get_device_id_from_config()` function to `tedge_config` to keep `config.device.id()` as private. https://github.com/thin-edge/thin-edge.io/pull/3318#discussion_r1910487052

**doc part**
- [x] Create Basic auth and SmartREST1.0 support doc

## Types of changes

## Paste Link to the issue
#3242 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -1181,6 +1181,15 @@ impl TEdgeConfigReader {
             Some(Cloud::Aws(profile)) => &self.aws.try_get(profile)?.device.cert_path,
         })
     }
+
+    pub fn device_id<'a>(&self, cloud: Option<impl Into<Cloud<'a>>>) -> Result<&str, ReadError> {
+        Ok(match cloud.map(<_>::into) {
+            None => self.device.id()?,
+            Some(Cloud::C8y(profile)) => self.c8y.try_get(profile)?.device.id()?,
+            Some(Cloud::Az(profile)) => self.az.try_get(profile)?.device.id()?,
+            Some(Cloud::Aws(profile)) => self.aws.try_get(profile)?.device.id()?,
+        })
+    }
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
@@ -1389,6 +1398,32 @@ fn aws_device_id(
     match dto_value.or_none() {
         Some(id) => Ok(id.clone()),
         None => device_id_from_cert(&aws_device.cert_path),
+    }
+}
+
+pub fn explicit_device_id(
+    config_location: &TEdgeConfigLocation,
+    cloud: &Option<Cloud>,
+) -> Option<String> {
+    let dto = config_location.load_dto_from_toml_and_env().ok()?;
+
+    match cloud {
+        None => dto.device.id.clone(),
+        Some(Cloud::C8y(profile)) => {
+            let key = profile.map(|name| name.to_string());
+            let c8y_dto = dto.c8y.try_get(key.as_deref(), "c8y").ok()?;
+            c8y_dto.device.id.clone()
+        }
+        Some(Cloud::Az(profile)) => {
+            let key = profile.map(|name| name.to_string());
+            let az_dto = dto.az.try_get(key.as_deref(), "az").ok()?;
+            az_dto.device.id.clone()
+        }
+        Some(Cloud::Aws(profile)) => {
+            let key = profile.map(|name| name.to_string());
+            let aws_dto = dto.aws.try_get(key.as_deref(), "aws").ok()?;
+            aws_dto.device.id.clone()
+        }
     }
 }
 
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -483,8 +483,7 @@ define_tedge_config! {
             /// Identifier of the device within the fleet. It must be globally
             /// unique and is derived from the device certificate.
             #[tedge_config(reader(function = "c8y_device_id"))]
-            // TODO make this work
-            // #[tedge_config(default(from_optional_key = "device.id"))]
+            #[tedge_config(default(from_optional_key = "device.id"))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
             #[doku(as = "String")]
             id: Result<String, ReadError>,
@@ -1365,14 +1364,20 @@ fn device_id(
     device: &TEdgeConfigReaderDevice,
     dto_value: &OptionalConfig<String>,
 ) -> Result<String, ReadError> {
-    device_id_from_cert(&device.cert_path)
+    match dto_value.or_none() {
+        Some(id) => Ok(id.clone()),
+        None => device_id_from_cert(&device.cert_path)
+    }
 }
 
 fn c8y_device_id(
-    device: &TEdgeConfigReaderC8yDevice,
+    c8y_device: &TEdgeConfigReaderC8yDevice,
     dto_value: &OptionalConfig<String>,
 ) -> Result<String, ReadError> {
-    device_id_from_cert(&device.cert_path)
+    match dto_value.or_none() {
+        Some(id) => Ok(id.clone()),
+        None => device_id_from_cert(&c8y_device.cert_path)
+    }
 }
 
 fn az_device_id(device: &TEdgeConfigReaderAzDevice, _: &()) -> Result<String, ReadError> {
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -428,15 +428,8 @@ define_tedge_config! {
     device: {
         /// Identifier of the device within the fleet. It must be globally
         /// unique and is derived from the device certificate.
-        #[tedge_config(readonly(
-            write_error = "\
-                The device id is read from the device certificate and cannot be set directly.\n\
-                To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-            function = "device_id",
-        ))]
+        #[tedge_config(reader(function = "device_id"))]
         #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
-        #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
-        #[tedge_config(reader(private))]
         #[doku(as = "String")]
         id: Result<String, ReadError>,
 
@@ -489,14 +482,10 @@ define_tedge_config! {
         device: {
             /// Identifier of the device within the fleet. It must be globally
             /// unique and is derived from the device certificate.
-            #[tedge_config(readonly(
-                write_error = "\
-                    The device id is read from the device certificate and cannot be set directly.\n\
-                    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-                function = "c8y_device_id",
-            ))]
+            #[tedge_config(reader(function = "c8y_device_id"))]
+            // TODO make this work
+            // #[tedge_config(default(from_optional_key = "device.id"))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
-            #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
             #[doku(as = "String")]
             id: Result<String, ReadError>,
 
@@ -692,7 +681,6 @@ define_tedge_config! {
                 function = "az_device_id",
             ))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
-            #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
             #[doku(as = "String")]
             id: Result<String, ReadError>,
 
@@ -762,7 +750,6 @@ define_tedge_config! {
                 function = "aws_device_id",
             ))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
-            #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
             #[doku(as = "String")]
             id: Result<String, ReadError>,
 
@@ -1367,26 +1354,32 @@ fn default_http_bind_address(dto: &TEdgeConfigDto) -> IpAddr {
 
 fn device_id_from_cert(cert_path: &Utf8Path) -> Result<String, ReadError> {
     let pem = PemCertificate::from_pem_file(cert_path)
-        .map_err(|err| cert_error_into_config_error(ReadOnlyKey::DeviceId.to_cow_str(), err))?;
+        .map_err(|err| cert_error_into_config_error(ReadableKey::DeviceId.to_cow_str(), err))?;
     let device_id = pem
         .subject_common_name()
-        .map_err(|err| cert_error_into_config_error(ReadOnlyKey::DeviceId.to_cow_str(), err))?;
+        .map_err(|err| cert_error_into_config_error(ReadableKey::DeviceId.to_cow_str(), err))?;
     Ok(device_id)
 }
 
-fn device_id(device: &TEdgeConfigReaderDevice) -> Result<String, ReadError> {
+fn device_id(
+    device: &TEdgeConfigReaderDevice,
+    dto_value: &OptionalConfig<String>,
+) -> Result<String, ReadError> {
     device_id_from_cert(&device.cert_path)
 }
 
-fn c8y_device_id(device: &TEdgeConfigReaderC8yDevice) -> Result<String, ReadError> {
+fn c8y_device_id(
+    device: &TEdgeConfigReaderC8yDevice,
+    dto_value: &OptionalConfig<String>,
+) -> Result<String, ReadError> {
     device_id_from_cert(&device.cert_path)
 }
 
-fn az_device_id(device: &TEdgeConfigReaderAzDevice) -> Result<String, ReadError> {
+fn az_device_id(device: &TEdgeConfigReaderAzDevice, _: &()) -> Result<String, ReadError> {
     device_id_from_cert(&device.cert_path)
 }
 
-fn aws_device_id(device: &TEdgeConfigReaderAwsDevice) -> Result<String, ReadError> {
+fn aws_device_id(device: &TEdgeConfigReaderAwsDevice, _: &()) -> Result<String, ReadError> {
     device_id_from_cert(&device.cert_path)
 }
 
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
@@ -428,7 +428,7 @@ define_tedge_config! {
     device: {
         /// Identifier of the device within the fleet. It must be globally
         /// unique and is derived from the device certificate.
-        #[tedge_config(reader(function = "device_id"))]
+        #[tedge_config(reader(function = "device_id", private))]
         #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
         #[doku(as = "String")]
         id: Result<String, ReadError>,
@@ -673,12 +673,8 @@ define_tedge_config! {
         device: {
             /// Identifier of the device within the fleet. It must be globally
             /// unique and is derived from the device certificate.
-            #[tedge_config(readonly(
-                write_error = "\
-                    The device id is read from the device certificate and cannot be set directly.\n\
-                    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-                function = "az_device_id",
-            ))]
+            #[tedge_config(reader(function = "az_device_id"))]
+            #[tedge_config(default(from_optional_key = "device.id"))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
             #[doku(as = "String")]
             id: Result<String, ReadError>,
@@ -742,12 +738,8 @@ define_tedge_config! {
         device: {
             /// Identifier of the device within the fleet. It must be globally
             /// unique and is derived from the device certificate.
-            #[tedge_config(readonly(
-                write_error = "\
-                    The device id is read from the device certificate and cannot be set directly.\n\
-                    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-                function = "aws_device_id",
-            ))]
+            #[tedge_config(reader(function = "aws_device_id"))]
+            #[tedge_config(default(from_optional_key = "device.id"))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
             #[doku(as = "String")]
             id: Result<String, ReadError>,
@@ -1366,7 +1358,7 @@ fn device_id(
 ) -> Result<String, ReadError> {
     match dto_value.or_none() {
         Some(id) => Ok(id.clone()),
-        None => device_id_from_cert(&device.cert_path)
+        None => device_id_from_cert(&device.cert_path),
     }
 }
 
@@ -1376,16 +1368,28 @@ fn c8y_device_id(
 ) -> Result<String, ReadError> {
     match dto_value.or_none() {
         Some(id) => Ok(id.clone()),
-        None => device_id_from_cert(&c8y_device.cert_path)
+        None => device_id_from_cert(&c8y_device.cert_path),
     }
 }
 
-fn az_device_id(device: &TEdgeConfigReaderAzDevice, _: &()) -> Result<String, ReadError> {
-    device_id_from_cert(&device.cert_path)
+fn az_device_id(
+    az_device: &TEdgeConfigReaderAzDevice,
+    dto_value: &OptionalConfig<String>,
+) -> Result<String, ReadError> {
+    match dto_value.or_none() {
+        Some(id) => Ok(id.clone()),
+        None => device_id_from_cert(&az_device.cert_path),
+    }
 }
 
-fn aws_device_id(device: &TEdgeConfigReaderAwsDevice, _: &()) -> Result<String, ReadError> {
-    device_id_from_cert(&device.cert_path)
+fn aws_device_id(
+    aws_device: &TEdgeConfigReaderAwsDevice,
+    dto_value: &OptionalConfig<String>,
+) -> Result<String, ReadError> {
+    match dto_value.or_none() {
+        Some(id) => Ok(id.clone()),
+        None => device_id_from_cert(&aws_device.cert_path),
+    }
 }
 
 fn cert_error_into_config_error(key: Cow<'static, str>, err: CertificateError) -> ReadError {
--- a/crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
+++ b/crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
@@ -102,6 +102,10 @@ impl TEdgeConfigLocation {
         Ok(TEdgeConfig::from_dto(&dto, self))
     }
 
+    pub fn load_dto_from_toml_and_env(&self) -> Result<TEdgeConfigDto, TEdgeConfigError> {
+        self.load_dto::<FileAndEnvironment>(self.toml_path())
+    }
+
     fn load_dto<Sources: ConfigSources>(
         &self,
         path: &Utf8Path,
--- a/crates/core/tedge/src/cli/certificate/cli.rs
+++ b/crates/core/tedge/src/cli/certificate/cli.rs
@@ -1,12 +1,3 @@
-use crate::cli::common::Cloud;
-use crate::cli::common::CloudArg;
-use camino::Utf8PathBuf;
-use clap::ValueHint;
-use tedge_config::OptionalConfigError;
-use tedge_config::ProfileName;
-use tedge_config::ReadError;
-use tedge_config::TEdgeConfig;
-
 use super::create::CreateCertCmd;
 use super::create_csr::CreateCsrCmd;
 use super::remove::RemoveCertCmd;
@@ -14,6 +5,17 @@ use super::renew::RenewCertCmd;
 use super::show::ShowCertCmd;
 use super::upload::*;
 
+use anyhow::anyhow;
+use camino::Utf8PathBuf;
+use clap::ValueHint;
+use tedge_config::explicit_device_id;
+use tedge_config::OptionalConfigError;
+use tedge_config::ProfileName;
+use tedge_config::TEdgeConfig;
+use tedge_config::TEdgeConfigLocation;
+
+use crate::cli::common::Cloud;
+use crate::cli::common::CloudArg;
 use crate::command::BuildCommand;
 use crate::command::BuildContext;
 use crate::command::Command;
@@ -24,8 +26,8 @@ pub enum TEdgeCertCli {
     /// Create a self-signed device certificate
     Create {
         /// The device identifier to be used as the common name for the certificate
-        #[clap(long = "device-id")]
-        id: String,
+        #[clap(long = "device-id", global = true)]
+        id: Option<String>,
 
         #[clap(subcommand)]
         cloud: Option<CloudArg>,
@@ -80,12 +82,14 @@ impl BuildCommand for TEdgeCertCli {
         let cmd = match self {
             TEdgeCertCli::Create { id, cloud } => {
                 let cloud: Option<Cloud> = cloud.map(<_>::try_into).transpose()?;
+
                 let cmd = CreateCertCmd {
-                    id,
+                    id: get_device_id(id, &config, &context.config_location, &cloud)?,
                     cert_path: config.device_cert_path(cloud.as_ref())?.to_owned(),
                     key_path: config.device_key_path(cloud.as_ref())?.to_owned(),
                     user: user.to_owned(),
                     group: group.to_owned(),
+                    config_location: context.config_location,
                 };
                 cmd.into_boxed()
             }
@@ -97,15 +101,8 @@ impl BuildCommand for TEdgeCertCli {
             } => {
                 let cloud: Option<Cloud> = cloud.map(<_>::try_into).transpose()?;
 
-                // Use the current device id if no id is provided
-                let id = if let Some(id) = id {
-                    id
-                } else {
-                    get_device_id_from_config(&config, &cloud)?
-                };
-
                 let cmd = CreateCsrCmd {
-                    id,
+                    id: get_device_id(id, &config, &context.config_location, &cloud)?,
                     key_path: config.device_key_path(cloud.as_ref())?.to_owned(),
                     // Use output file instead of csr_path from tedge config if provided
                     csr_path: output_path.unwrap_or_else(|| config.device.csr_path.clone()),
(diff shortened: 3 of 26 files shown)


didier-wenzek (commented):
If the device id has been set using `tedge config set device.id`, then the device id should no more be mandatory for `tedge cert create`.

We also need to take care of the paths for the cloud specific certificates.
For instance, the following sequence of command is error prone:

```
# Trying to create a certificate for an unknown cloud profile (okay: the error is detected)
$ tedge cert create --device-id demo-c8y c8y --profile foo
Error: missing configuration parameter

Caused by:
Unknown profile `foo` for the multi-profile property c8y

# Creating the cloud profile
$ tedge config set c8y.device.id device-c8y-foo  --profile foo

# Now that the cloud profile exists, the cert creat command runs
# BUT
#    - --device-id is required while set in the config
#    - the value set on the command line erases the config
#    - the certificate is created globally (because no specific paths were configured for this profile)
$ tedge cert create --device-id some-other-device-id c8y --profile foo
Error: failed to create a test certificate for the device some-other-device-id.

Caused by:
A certificate already exists and would be overwritten.
Existing file: "/etc/tedge/device-certs/demo-device-888.pem"
Run `tedge cert remove` first to generate a new certificate.
```

rina23q:
@didier-wenzek 
Actually this error occurs even now if user doesn't set any keys for a profile beforehand.

> ```
> # Trying to create a certificate for an unknown cloud profile (okay: the error is detected)
> $ tedge cert create --device-id demo-c8y c8y --profile foo
> Error: missing configuration parameter
> 
> Caused by:
> Unknown profile `foo` for the multi-profile property c8y
>```

For the rest, I would like to implement to have the consistent behavour as `create-csr`, which uses the value of `device.id` if `--device-id` is not given.

However, on the condition, if `device.id` is already configured, but `--device-id` option is given with the different name as you tried, what is the expected behaviour? Returns an error as they are conflictiing?

crates/core/tedge/src/error.rs
```diff
@@ -44,4 +45,14 @@ pub enum TEdgeError {
 
     #[error(transparent)]
     FromC8yEndPointConfigError(#[from] c8y_api::http_proxy::C8yEndPointConfigError),
+
+    #[error(
+        r#"The given device ID '{input_id}' doesn't match the one in the config '{config_id}'.
+    Run `tedge config unset {writable_key}` first to unset the device ID."#
+    )]
+    UnmatchedDeviceId {
```
jarhodes314, line 53:
```suggestion
    MismatchedDeviceId {
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
```diff
@@ -102,6 +102,10 @@ impl TEdgeConfigLocation {
         Ok(TEdgeConfig::from_dto(&dto, self))
     }
 
+    pub fn load_dto_with_file_and_env(&self) -> Result<TEdgeConfigDto, TEdgeConfigError> {
```
jarhodes314, line 105:
```suggestion
    pub fn load_dto_from_toml_and_env(&self) -> Result<TEdgeConfigDto, TEdgeConfigError> {
```

jarhodes314 (commented):
From a quick test this seems to be working well. It will be much easier to use the CLI with the persisted device id.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -192,16 +194,187 @@ pub enum UploadCertCli {
...
+    #[test_case("c8y-test", Some("c8y-test"), Some(CloudArg::C8y{ profile: None }), toml::toml!{
+    [device]
+        id = "test"
+    })]
+    #[test_case("c8y-foo-test", Some("c8y-foo-test"), Some(CloudArg::C8y{ profile: Some("foo".parse().unwrap()) }), toml::toml!{
+    [device]
+        id = "test"
+    [c8y.device]
+        id = "c8y-test"
+    })]
```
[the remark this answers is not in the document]
rina23q: In practice, `tedge cert create` will return an error because no profile for "foo" exists. So the following `toml` fie is more realistic. Though the result `get_device_id()` is the same.

```toml
[device]
id = "test"
[c8y.device]
id = "c8y-test"
[c8y.profiles.foo.device]
cert_path = "/path/to/cert.pem"
key_path = "/path/to/key.pem"
```

crates/core/tedge/src/error.rs
rina23q: fixed in https://github.com/thin-edge/thin-edge.io/pull/3326/commits/ed4043332d41a517b08922c90f7a3e55ce4e8c7c

crates/common/tedge_config/src/tedge_config_cli/tedge_config_location.rs
rina23q: fixed in https://github.com/thin-edge/thin-edge.io/pull/3326/commits/bfc6e14b0683c0cb3279a25d0148fd6c20cad744

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +196,187 @@ pub enum UploadCertCli {
...
+        .context("failed to read tedge.toml file")?;
+
+    match id {
+        None => match config_id {
+            None => Err(anyhow!("No device ID is provided. Use `--device-id <name>` option to specify the device ID.").into()),
+            Some(config_id) => Ok(config_id.into()),
+        }
+        Some(input_id) if config_id.is_none() => Ok(input_id),
+        Some(input_id) if input_id == config_id.unwrap() => Ok(input_id),
+        Some(input_id) => match cloud.as_ref().map(<_>::into) {
```
didier-wenzek, line 220: Matching on the `(id, config_id)` pair would make clearer the intent that is to compare the two ids checking for conflicts.

```rust
    match (id, config_id) {
        (None, None) => Err(anyhow!("No device ID is provided. Use `--device-id <name>` option to specify the device ID.").into()),
        (None, Some(config_id)) => Ok(config_id.into()),
        (Some(input_id), None) => Ok(input_id),
        (Some(input_id), Some(config_id)) if input_id == config_id => Ok(input_id),
        (Some(input_id), Some(config_id)) => {
            // ...
        }
    }
```

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +196,187 @@ pub enum UploadCertCli {
...
+            None => Err(MismatchedDeviceId {
+                input_id,
+                config_id: config_id.unwrap().into(),
+                writable_key,
+            }),
+            Some(tedge_config::Cloud::C8y(profile)) => {
+                let key = profile.map(|name| name.to_string());
+                let c8y_dto = dto.c8y.try_get(key.as_deref(), "c8y")?;
+                if c8y_dto.device.id.is_some() {
+                    Err(MismatchedDeviceId {
```
didier-wenzek, line 230: Is this error case really useful?

- It's used only in this function
- and the error case for no device id at all is raised using  `anyhow!()`.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +196,187 @@ pub enum UploadCertCli {
...
+                let az_dto = dto.az.try_get(key.as_deref(), "az")?;
+                if az_dto.device.id.is_some() {
+                    Err(MismatchedDeviceId {
+                        input_id,
+                        config_id: config_id.unwrap().into(),
+                        writable_key,
+                    })
+                } else {
+                    Ok(input_id)
+                }
```
didier-wenzek, line 250: I don't understand what is checked here (for `az` as well as for `c8y` and `aws`).

I seems that we have 3 device ids:  `input_id`, `config_id`, and `az_dto.device.id`, the value of the later being ignored.

rina23q: `dto` is used only to check if the value exists really in `az.device.id`. This cannot be checked by using `reader`.

Some examples:
---

```toml
[device]
id = "generic"

[az.device]
id = "az_id"
```

- `config_id` (= the value from `reader`) has `az_id`.
- `az.dto.device.id` has `Some("az_id")`. So, here the check `az_dto.device.id.is_some()` is `true`.

---

```toml
[device]
id = "generic"
```

- `config_id` (= the value from `reader`) has `generic`.
- - `az.dto.device.id` has `None`. So, here the check `az_dto.device.id.is_some()` is `false`.

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: Yeah, I wanted to find a better solution here. Using `anyhow!()` is definitely fine, even I used it at very first draft. But it will cause repeating the same long error message multiple times in the code. Anyway, I don't like that I had to add `MismatchedDeviceId` under `TEdgeError` as well, so using `anyhow!()` may be better.

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
```diff
@@ -0,0 +1,102 @@
...
+    # device.id is set by tedge cert create
+    ${config_output}=    Execute Command
+    ...    tedge config get c8y.device.id --profile second
+    ...    strip=${True}
+    Should Be Equal    ${config_output}    ${SECOND_DEVICE_SN}
+    ${toml_output}=    Execute Command    cmd=grep -B 1 ${SECOND_DEVICE_SN} /etc/tedge/tedge.toml    strip=${True}
+    Should Be Equal    ${toml_output}    second=[c8y.profiles.second.device]\nid = "${SECOND_DEVICE_SN}"
+
+    # Using another cloud profile with predefined device.id
+    Execute Command    tedge config set c8y.device.id --profile third ${THIRD_DEVICE_SN}
```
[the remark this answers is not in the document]
rina23q: I think my comment caused the confusion. What I wanted to test here are
- Both cases are using cloud profiles
- `c8y.profiles.second.device.id` is not set by `tedge config set`. It is set by `tedge cert create --device-id ${SECOND_DEVICE_SN}`
- `c8y.profiles.third.device.id` is set by `tedge config set c8y.device.id --profile third`. Then, `tedge cert create c8y --profile third` works and it uses the id from config.

So just comment needs to be fixed.

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: Looks good. Since the condition is quite complex, more readability is definitely helpful.

crates/core/tedge/src/cli/certificate/cli.rs
didier-wenzek: Okay, but then why don't we have a check of the cloud specific value. It might be set to the same as for the global device id. For example:

```
[device]
id = "az_id"

[az.device]
id = "az_id"
```

rina23q: @didier-wenzek 
I didn't get which error occurs in your case. In the case, using the value from Reader and Dto is eventually the same.

| key   | Reader |  Dto|
| -------- | ------- | -----|
|device.id  | az_id  | Some("az_id")
| az.device.id | az_id  | Some("az_id")

didier-wenzek: > I didn't get which error occurs in your case. 

This is my point, there should be no error. While, if I understand correctly, the current code is returning an error because `az_dto.device.id.is_some()`.

rina23q: There is one match condition before entering this arm,
```
Some(input_id) if input_id == config_id.unwrap() => Ok(input_id),
```
So, if `input_id` is "az_id", the command will return `Ok(input_id)`.

And, yes, it returns an error if `input_id` != `config_id`and `az_dto.device_id.is_some()`. This should be an error because user tries to create a cert with mismatched ids.

didier-wenzek: So, we are back to my first comment: I don't understand what is checked here.

I will have another look when https://github.com/thin-edge/thin-edge.io/pull/3326#discussion_r1922383288 will be addressed.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -80,12 +85,15 @@ impl BuildCommand for TEdgeCertCli {
         let cmd = match self {
             TEdgeCertCli::Create { id, cloud } => {
                 let cloud: Option<Cloud> = cloud.map(<_>::try_into).transpose()?;
+
                 let cmd = CreateCertCmd {
-                    id,
+                    id: get_device_id(id, &config, &context.config_location, &cloud)?,
```
albinsuresh, line 90: Although I don't have any objections to this, I'm a bit surprised to see this `device.id` match validation here. I was under the impression that a value provided in the cli would simply override the existing value in the config and any validation would happen only during `tedge connect`.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +196,187 @@ pub enum UploadCertCli {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tedge_test_utils::fs::TempTedgeDir;
+    use test_case::test_case;
+
+    #[test_case("test", None, None, toml::toml!{
+    [device]
+        id = "test"
+    })]
```
albinsuresh, line 294: Formatting suggestion:
```suggestion
    #[test_case(
        "test", 
        None, 
        None, 
        toml::toml!{
            [device]
            id = "test"
        }
    )]
```
Same for the rest of the test cases as well.

crates/core/tedge/src/cli/certificate/renew.rs
```diff
@@ -59,6 +59,8 @@ mod tests {
     use std::path::Path;
```
albinsuresh, line 59: Wondering if it'd be safer to call `set_device_id` from this `renew` command `execute()` as well so that any configuration from an older version of `tedge` is updated when the cert is renewed with the newer version.

albinsuresh (commented):
The feature is working as expected. But, simplifying the device id validation done in the `tedge cert` commands as suggested by @didier-wenzek  would go a long way in making that function comprehendible, and easy to review.

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: >  I was under the impression that a value provided in the cli would simply override the existing value in the config

I'm actually now reconsidering this.  If user run like below:
```
tedge config set device.id foo
tedge cert create --device-id bar
```

- If cert/key doesn't exist, the command creates cert/key and set `device.id` to `foo`.
- If cert/key already exists, the command returns an error (known behaviour). No overwriting either id or cert. So, it cannot mess up already existing cert.

Does it really hurt users to overwrite the device id by `tedge cert` although they configured before? I guess not. Also, the command tells `device.id` is set to `foo`.

In that case, error scenario is only there are no input id AND no config id.

> any validation would happen only during tedge connect.

The validation required in `tedge connect` is a different level. It checks if CN matches the configured `device.id` (or profile version `c8y.device.id` / `c8y.profiles.<profile>.device.id`).

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: Which formatter are you using and forces this? My IDE doesn't, it's a pain to do all manually :sweat_smile:

crates/core/tedge/src/cli/certificate/renew.rs
rina23q: Why not :+1:

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: It can be simplified like this.

```rust
fn get_device_id(
    id: Option<String>,
    config: &TEdgeConfig,
    cloud: &Option<Cloud>,
) -> Result<String, anyhow::Error> {
    let config_id = config.device_id(cloud.as_ref()).ok();
    match (id, config_id) {
        (None, None) => Err(anyhow!("No device ID is provided. Use `--device-id <name>` option to specify the device ID.")),
        (None, Some(config_id)) => Ok(config_id.into()),
        (Some(input_id), None) => Ok(input_id),
        (Some(input_id), Some(_config_id)) => Ok(input_id),
}
```

crates/core/tedge/src/cli/certificate/cli.rs
albinsuresh: That was just manually formatted as `cargo fmt` doesn't touch stuff defined inside macros (`test_case` in this case). I found that laset few test cases really hard to read, hence suggested this.

crates/core/tedge/src/cli/certificate/cli.rs
albinsuresh: IMO, overwriting is the simpler contract to enforce. Keeps things simple at the code level as well.

crates/core/tedge/src/error.rs
```diff
@@ -44,4 +45,14 @@ pub enum TEdgeError {
 
     #[error(transparent)]
     FromC8yEndPointConfigError(#[from] c8y_api::http_proxy::C8yEndPointConfigError),
+
+    #[error(
+        r#"The given device ID '{input_id}' doesn't match the one in the config '{config_id}'.
+    Run `tedge config unset {writable_key}` first to unset the device ID."#
```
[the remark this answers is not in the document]
rina23q: Simplified in 99620aa

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: Simplified in 99620aa

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: Simplified in 99620aa

Also see the thread in https://github.com/thin-edge/thin-edge.io/pull/3326#discussion_r1923282326

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: Addressed in https://github.com/thin-edge/thin-edge.io/commit/99620aa1a5cc761ec794b7097e36008398a9360d

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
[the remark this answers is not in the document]
rina23q: Test updated in bb25494d98fde99bbf5535caaeac36eebc5a4ff3

crates/core/tedge/src/cli/certificate/renew.rs
rina23q: Done in 56986f249540b7872f2b8807663fd77b6e6266ad

crates/core/tedge/src/cli/certificate/cli.rs
didier-wenzek: ~~I'm now surprised that there no more conflict detection at all, when the device id is provided on the command line and in the config, but with different values. Wouldn't be better to raise an error stressing the conflict?~~

Okay after having read  https://github.com/thin-edge/thin-edge.io/pull/3326#discussion_r1923282326.

didier-wenzek (commented):
The code looks good and simpler thanks to the latest commits. However, I still need some time to test the updated cli.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -428,15 +428,9 @@ define_tedge_config! {
...
         /// unique and is derived from the device certificate.
-        #[tedge_config(readonly(
-            write_error = "\
-                The device id is read from the device certificate and cannot be set directly.\n\
-                To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-            function = "device_id",
-        ))]
+        #[tedge_config(reader(function = "device_id", private))]
         #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
         #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
```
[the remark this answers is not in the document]
rina23q: done in 9d6dbdbab9d5625cec4255e0382f78364064e69e

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -685,12 +674,8 @@ define_tedge_config! {
...
-            #[tedge_config(readonly(
-                write_error = "\
-                    The device id is read from the device certificate and cannot be set directly.\n\
-                    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-                function = "az_device_id",
-            ))]
+            #[tedge_config(reader(function = "az_device_id"))]
+            #[tedge_config(default(from_optional_key = "device.id"))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
             #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
```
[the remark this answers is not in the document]
rina23q: done in 9d6dbdbab9d5625cec4255e0382f78364064e69e

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -755,12 +740,8 @@ define_tedge_config! {
...
-            #[tedge_config(readonly(
-                write_error = "\
-                    The device id is read from the device certificate and cannot be set directly.\n\
-                    To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`.",
-                function = "aws_device_id",
-            ))]
+            #[tedge_config(reader(function = "aws_device_id"))]
+            #[tedge_config(default(from_optional_key = "device.id"))]
             #[tedge_config(example = "Raspberrypi-4d18303a-6d3a-11eb-b1a6-175f6bb72665")]
             #[tedge_config(note = "This setting is derived from the device certificate and is therefore read only.")]
```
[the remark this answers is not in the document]
rina23q: done in 9d6dbdbab9d5625cec4255e0382f78364064e69e

docs/src/operate/c8y/smartrest-one.md
```diff
@@ -0,0 +1,65 @@
...
+sudo tedge config set device.id <id>
+```
+
+## Adding SmartREST 1.0 template to %%te%% config
+
+Specify the external IDs of the SmartREST 1.0 templates as shown below:
+
+```sh
+sudo tedge config set c8y.smartrest1.templates template-1,template-2
+```
```
didier-wenzek, line 52: *suggestion*:

You have to register on the device all the SmartREST 1.0 templates you plan to use.

This can be done by providing a comma-separated list of the external IDs of the SmartREST 1.0 templates:

```sh
sudo tedge config set c8y.smartrest1.templates template-1,template-2
```

This can also be done by adding these template IDs one by one;

```sh
sudo tedge config add c8y.smartrest1.templates template-1
sudo tedge config add c8y.smartrest1.templates template-2
```

didier-wenzek (commented):
The main feature of this PR (i.e. cloud specific device ids and certificates) is working as described. However, the `tedge` cli starts to be non intuitive and even confusing. Furthermore, `tedge config list XXX` is now buggy missing some entries.

## Error prone CLI

One can make senses of the improved `tedge` command to create device certificates and CSR:

```sh
# Create a device certificate used for all connections unless a more specific one is provided
$ tedge cert create

# Create a device certificate for all __C8Y__ connections unless a more specific one is provided
$ tedge cert create c8y

# Create a device certificate for a specific C8Y tenant
$ tedge cert create c8y --profile foo
```

But this is not so intuitive:

- The cloud is given `tedge cert` as a sub-command, the profile as an option.
- To get/set the device id and cert path with `tedge config`,
  the cloud is given as a prefix (e.g. `c8y.device.id`) and the profile as an option.
- `tedge config list` uses yet another convention (e.g. `c8y.profiles.foo.root_cert_path=/etc/ssl/certs`)

This is even error-prone.
If you want to create a specific certificate for a c8y tenant,
- starting with `tedge config set c8y.device.id c8y.foo.device --profile foo`
- but forgetting to set `c8y.device.key_path` and `c8y.device.cert_path` for the same profile
- then the command `tedge cert create c8y --profile foo`
- creates a certificate used by __all__ the connections.

## Missing tedge config list entries

```sh
# The device being configured with 3 cloud specific key pairs
$ tedge config get device.key_path
/etc/tedge/device-certs/device.key
$ tedge config get c8y.device.key_path
/tmp/c8y.key
$ tedge config get c8y.device.key_path --profile foo
/tmp/foo.c8y.key

# tedge config lists only the main key pair
$ tedge config list key_path                        
device.key_path=/etc/tedge/device-certs/device.key
$ tedge config list | grep key_path                 
device.key_path=/etc/tedge/device-certs/device.key

# while the config contains the missing entries
$ grep key_path /etc/tedge/tedge.toml
key_path = "/etc/tedge/device-certs/device.key"
key_path = "/tmp/foo.c8y.key"
key_path = "/tmp/c8y.key"
```

jarhodes314:
> Furthermore, tedge config list XXX is now buggy missing some entries.
> ## Missing tedge config list entries
> ```shell
> # The device being configured with 3 cloud specific key pairs
> $ tedge config get device.key_path
> /etc/tedge/device-certs/device.key
> $ tedge config get c8y.device.key_path
> /tmp/c8y.key
> $ tedge config get c8y.device.key_path --profile foo
> /tmp/foo.c8y.key
> 
> # tedge config lists only the main key pair
> $ tedge config list key_path                        
> device.key_path=/etc/tedge/device-certs/device.key
> $ tedge config list | grep key_path                 
> device.key_path=/etc/tedge/device-certs/device.key
> 
> # while the config contains the missing entries
> $ grep key_path /etc/tedge/tedge.toml
> key_path = "/etc/tedge/device-certs/device.key"
> key_path = "/tmp/foo.c8y.key"
> key_path = "/tmp/c8y.key"
> ```

This was done in https://github.com/thin-edge/thin-edge.io/commit/ad2209f7d382b40c84ce635fa2892c4d67989b0d. We should be able to revert this commit if the multiple device IDs are handled correctly. Now #3230 is merged, I believe this should be the case @albinsuresh?

docs/src/operate/c8y/smartrest-one.md
rina23q: Done in 54e69907fd978bb7961cf479fe28efeae92916f9

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +187,99 @@ pub enum UploadCertCli {
...
-    .to_owned();
-    Ok(id)
+) -> Result<String, anyhow::Error> {
+    match (id, config.device_id(cloud.as_ref()).ok()) {
+        (None, None) => Err(anyhow!(
+            "No device ID is provided. Use `--device-id <name>` option to specify the device ID."
+        )),
+        (None, Some(config_id)) => Ok(config_id.into()),
+        (Some(input_id), _) => Ok(input_id),
+    }
```
didier-wenzek, line 202: If `config.device_id(cloud.as_ref())` is not ok , meaning that no device id has been *specifically* set for the cloud profile a certificate is created, then the config has to be search for a  more global device id (at the cloud type level then at the device level).

Also if the device id is provided both on the command line *and* the config, but with different values, then are error must be raised. This was okay to ignore the difference when the config was updated, but this is no more the case, leading to weird behavior:

```sh
# force a device id for the c8y foo profile
$ tedge config set c8y.device.id --profile foo device-foo-c8y

# create certificate for the c8y foo profile with a different id
$ tedge cert create c8y --profile foo --device-id not-device-foo-c8y
Certificate was successfully created

Device certificate: /tmp/foo.c8y.cert
Subject: CN=not-device-foo-c8y, O=Thin Edge, OU=Test Device
Issuer: CN=not-device-foo-c8y, O=Thin Edge, OU=Test Device
Valid from: Wed, 22 Jan 2025 09:35:50 +0000
Valid up to: Thu, 22 Jan 2026 09:35:50 +0000
Thumbprint: 1D15451E0DCD20D2C8044A1179EBAC79D257CBA7

# The certificate CN is coming from the command line
# But the config is now misleading
# => an error should have been raised by tedge cert create
$ tedge config get c8y.device.id --profile foo                       
device-foo-c8y
```

Hence the code should be something closer too:

```suggestion
    match (id, config.device_id(cloud.as_ref()).ok()) {
        (None, None) => {
           match get_more_specific_id(...) {
               Some(config_id) => Ok(config_id),
               None => Err(anyhow!(
                 "No device ID is provided. Use `--device-id <name>` option to specify the device ID."
               )),
           }
        }
        None, Some(config_id)) => Ok(config_id.into()),
        (Some(input_id), Some(config_id)) if input_id != config_id => Err(anyhow!(
          "`--device-id` option conflicts with tedge config settings"
        )),
        (Some(input_id), _) => Ok(input_id),
    }
```

didier-wenzek (commented):
I confirm that the following has been correctly implemented as discussed offline:

- `tedge cert create` doesn't update the config file with the device id.
- on success,  `tedge cert create` displays the certificate path and content.

However, I think we are going backward on some points:

- `--device-id` is now a mandatory argument for `tedge cert create` unless set for the *exact same* cloud profile in the config.  If the device id has been set at global level in the config, it should be used as done by the first steps of this PR.
- A conflict, between the config and a  `--device-id`  provided on the cli, leads to a confusing state.

Also missing:

- `device.csr_path` doesn't have its `c8y.device.csr_path` specialization (nor `aws` and `az` specializations), meaning one cannot request a CSR for a specific cloud profile.

albinsuresh:
> > Furthermore, tedge config list XXX is now buggy missing some entries.
> 
> This was done in [ad2209f](https://github.com/thin-edge/thin-edge.io/commit/ad2209f7d382b40c84ce635fa2892c4d67989b0d). We should be able to revert this commit if the multiple device IDs are handled correctly. Now #3230 is merged, I believe this should be the case @albinsuresh?

Yes, the external id aspects were completely moved to the mappers from the `entity-store` and hence the `agent` is not dependent on a `device.id` anymore. So, hiding the configs can be reverted now.

rina23q:
> Also missing:
> 
> `device.csr_path` doesn't have its `c8y.device.csr_path` specialization (nor `aws` and `az` specializations), meaning one cannot > request a CSR for a specific cloud profile.

This can be addressed in a separated PR. It is more like a missed part during the work of cloud profile. Should have been addressed before this PR.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +191,247 @@ pub enum UploadCertCli {
...
+        (None, None) => Err(anyhow!(
+            "No device ID is provided. Use `--device-id <name>` option to specify the device ID."
+        )),
+        (None, Some(config_id)) => Ok(config_id.into()),
+        (Some(input_id), None) => Ok(input_id),
+        (Some(input_id), Some(config_id)) if input_id == config_id => Ok(input_id),
+        (Some(input_id), Some(_config_id)) => {
+            // if ids from input and TEdgeConfigReader mismatch, need to dive into TEdgeConfigDto level
+            validate_input_id_with_dto(&input_id, config_location, cloud)
+        }
```
didier-wenzek, line 211: I don't really get the changes introduced by this commit https://github.com/thin-edge/thin-edge.io/pull/3326/commits/09bf5f92e1dfef2aaee9e44ed90b497b584f960b.

- First, this doesn't fix the issue with `tedge cert create c8y --profile foo` requesting a device id while `tedge config get device.id` returns one (this issue arises when the device id is read from the main device certificate and not the config)
- Second, it seems that there is no difference in behavior between the two ways to read the config (`config.device_id(cloud)` vs what is done inside `validate_input_id_with_dto`). (I have to double check that)

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +191,247 @@ pub enum UploadCertCli {
...
+        Some(tedge_config::Cloud::C8y(profile)) => {
+            let key = profile.map(|name| name.to_string());
+            let c8y_dto = dto.c8y.try_get(key.as_deref(), "c8y")?;
+            let writable_key = WritableKey::C8yDeviceId(key);
+            validate_dto_device_id(&c8y_dto.device.id, input_id, writable_key)
+        }
+        Some(tedge_config::Cloud::Az(profile)) => {
+            let key = profile.map(|name| name.to_string());
+            let az_dto = dto.az.try_get(key.as_deref(), "az")?;
+            let writable_key = WritableKey::AwsDeviceId(key);
```
rina23q, line 238:
```suggestion
            let writable_key = WritableKey::AzDeviceId(key);
```

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -193,16 +191,247 @@ pub enum UploadCertCli {
...
+            validate_input_id_with_dto(&input_id, config_location, cloud)
+        }
+    }
+}
+
+fn validate_input_id_with_dto(
+    input_id: &str,
+    config_location: &TEdgeConfigLocation,
+    cloud: &Option<Cloud>,
+) -> Result<String, anyhow::Error> {
```
didier-wenzek, line 219: I would replace this function by a function that *only* read the config. Check for conflicting parameters can easily be done by the caller.

```
fn explicit_device_id(
    config_location: &TEdgeConfigLocation,
    cloud: &Option<Cloud>,
) -> Option<String> {
    let dto = config_location
        .load_dto_from_toml_and_env().ok()?;

    match cloud.as_ref().map(<_>::into) {
        None => dto.device.id.clone(),
        Some(tedge_config::Cloud::C8y(profile)) => {
            let key = profile.map(|name| name.to_string());
            let c8y_dto = dto.c8y.try_get(key.as_deref(), "c8y").ok()?;
            c8y_dto.device.id.clone()
        }
        Some(tedge_config::Cloud::Az(profile)) => {
            let key = profile.map(|name| name.to_string());
            let az_dto = dto.az.try_get(key.as_deref(), "az").ok()?;
            az_dto.device.id.clone()
        }
        Some(tedge_config::Cloud::Aws(profile)) => {
            let key = profile.map(|name| name.to_string());
            let aws_dto = dto.aws.try_get(key.as_deref(), "aws").ok()?;
            aws_dto.device.id.clone()
        }
    }
}
```

didier-wenzek:
> > > Furthermore, tedge config list XXX is now buggy missing some entries.
> > 
> > 
> > This was done in [ad2209f](https://github.com/thin-edge/thin-edge.io/commit/ad2209f7d382b40c84ce635fa2892c4d67989b0d). We should be able to revert this commit if the multiple device IDs are handled correctly. Now #3230 is merged, I believe this should be the case @albinsuresh?
> 
> @jarhodes314 Yes, the external id aspects were completely moved to the mappers from the `entity-store` and hence the `agent` is not dependent on a `device.id` anymore. So, hiding the configs can be reverted now.

I confirm that the issue has been fixed  by @jarhodes314 's commit https://github.com/thin-edge/thin-edge.io/pull/3326/commits/d6efece89f59c43648c68c1ac05a8a12476310ea

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: Applied in 7cc012da1a080799742af636bff5bab792202c11

didier-wenzek:
Error messages and hints could be improved.

If you set a specific value for `c8y.device.cert_path`, but forget to create a specific certificate for c8y,
then `tedge connect` complains for good reasons but with a misleading hint:

```
$ tedge connect c8y --test  
error: Config value device.id, cannot be read: The device id is read from the device certificate.
To set 'device.id' to some <id>, you can use `tedge cert create --device-id <id>`. 
```

This is misleading, because there is a device certificate and `device.id` can be read.

```
$ tedge config get device.id                         
demo-device-888
$ tedge cert show           
Device certificate: /etc/tedge/device-certs/demo-device-888.pem
Subject: CN=demo-device-888, O=Thin Edge, OU=Test Device
...
```

Could the hint be more explicit about the missing certificate? e.g.:

```
$ tedge connect c8y --test
error: The device id cannot be read from the c8y.device.cert_path certificate:
<PATH to the configured certificate for c8y by c8y.device.cert_path>
```

rina23q:
> Could the hint be more explicit about the missing certificate? e.g.:
> 
> ```
> $ tedge connect c8y --test
> error: The device id cannot be read from the c8y.device.cert_path certificate:
> <PATH to the configured certificate for c8y by c8y.device.cert_path>
> ```

I'm changing the error message that is coming from this part:
https://github.com/thin-edge/thin-edge.io/blob/7cc012da1a080799742af636bff5bab792202c11/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs#L1430-L1438

The difficulty is to give a specific key for cloud profile in the error message. As you see in this line, currently this function always parses `ReadableKey::DeviceId`. This is the reason why you got the message `Config value device.id, cannot be read` although `ReadableKey::DeviceId` exists.
https://github.com/thin-edge/thin-edge.io/blob/7cc012da1a080799742af636bff5bab792202c11/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs#L1357

Then fixing it by giving a precise key, like `ReadableKey::C8yDeviceId("profile")` solves a problem? No. It cannot give the profiled key in the current implementation. Look at this default function for `c8y.device.id`. We cannot get a cloud profile here.

https://github.com/thin-edge/thin-edge.io/blob/7cc012da1a080799742af636bff5bab792202c11/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs#L1374-L1382

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -1353,11 +1353,12 @@ fn default_http_bind_address(dto: &TEdgeConfigDto) -> IpAddr {
 }
 
 fn device_id_from_cert(cert_path: &Utf8Path) -> Result<String, ReadError> {
-    let pem = PemCertificate::from_pem_file(cert_path)
-        .map_err(|err| cert_error_into_config_error(ReadableKey::DeviceId.to_cow_str(), err))?;
-    let device_id = pem
-        .subject_common_name()
-        .map_err(|err| cert_error_into_config_error(ReadableKey::DeviceId.to_cow_str(), err))?;
+    let pem = PemCertificate::from_pem_file(cert_path).map_err(|err| {
+        cert_error_into_config_error(ReadableKey::DeviceId.to_cow_str(), err, cert_path)
```
rina23q, line 1357: This `device_id_from_cert` is called by all device id default functions (`device_id()`, `c8y_device_id()`, `az_device_id()`, and `aws_device_id()`. However, here gives `ReadableKey::DeviceId` as key. This is not 100% correct. It can be `ReadableKey::C8yDeviceId(None)`, `ReadableKey::AzDeviceId(Some("abc"))` and so on.

The difficulty is, the default functions cannot know which profile is used. That's why it is impossible to parse the precise ReadableKey to this fuction from the default device id functions.

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
```diff
@@ -1497,6 +1498,9 @@ pub enum ReadError {
         key: Cow<'static, str>,
         cause: String,
     },
+
+    #[error("The device certificate cannot be read from the path: {path} ")]
+    CertNotFound { path: Utf8PathBuf },
```
rina23q, line 1503: I'm not happy to list this `CertNotFound` as tedge config's `ReadError`. I feel it's too specific.

rina23q (commented):
I was trying to fix the error message that @didier-wenzek highlighted. There are some points discovered why it's not easy.

crates/core/tedge/tests/main.rs
```diff
@@ -85,7 +85,7 @@ mod tests {
         // The remove command can be run when there is no certificate
         remove_cmd.assert().success();
 
-        // We start we no certificate, hence no device id
+        // We start with no certificate, hence no device id
```
didier-wenzek, line 88: This test has to be updated with the new error message.

```rust
        // We start with no certificate, hence no device id
        get_device_id_cmd
            .assert()
            .failure()
            .stderr(predicate::str::contains("The device certificate cannot be read from"));
```

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
didier-wenzek: So the issue to improve the error message is that `device_id_from_cert` has no clue on the actual key. But can this key be passed as a parameter with the appropriate value by the caller (i.e device_id()`, `c8y_device_id()`, `az_device_id()`, and `aws_device_id()`) ?

didier-wenzek (commented):
Since improving error messages and hints when the device id cannot be derived from the config, I propose to give up on this point - at least for now, and to focus on the main point of this PR: the ability to configure a device id with or with a certificate.

Configuring a device to use several cloud profiles each with its specific certificate is working, although this is for now a complicated and error prone task. We can improve the user experience in a second step with the broader scope of configuring multi-profile connection . @reubenmiller do you agree on that?

crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs
rina23q: > So the issue to improve the error message is that `device_id_from_cert` has no clue on the actual key. 

Right.

> But can this key be passed as a parameter with the appropriate value by the caller (i.e `device_id()`, `c8y_device_id()`, `az_device_id()`, and `aws_device_id()`) ?

Also no unfortunately. Look at this function. This function is called by the macro. No chance to get the profile.
https://github.com/thin-edge/thin-edge.io/blob/91eb9bd53406f7b3e9d596ccef80e4719bc4953b/crates/common/tedge_config/src/tedge_config_cli/tedge_config.rs#L1375-L1383

didier-wenzek (approved):
Approved, although the user experience could be simpler (see here for the complicate list of commands required when the use case is not the regular one: https://github.com/thin-edge/thin-edge.io/pull/3326#issuecomment-2607239479).

albinsuresh (approved):
The feature is working fine for the most common cases. The error messages could be improved in some cases though, but don't need to block this PR anymore for that. One simple example of such an error message is when trying to create a certificate with a different `--device-id` from what is set as `device.id`:
```
Error: missing configuration parameter

Caused by:
    `--device-id` option conflicts with tedge config settings.
    Configured value: 'albin_test', but input: 'albin_latest'
    
    Please either update the configuration using `tedge config set <key> <new_id>`
    or provide the correct value with the `--device-id` option.
```

The highlighted error is `Error: missing configuration parameter` whereas this is a pure case of "conflicting configuration". Even though that generic error is misleading, th rest of the message that suggests ways to fix is clear enough for the user. So, this is fine. If I spot more inconsistencies like this for the common use-cases, I'll compile and share it separately, even after this PR is merged.

rina23q:
> The highlighted error is `Error: missing configuration parameter` whereas this is a pure case of "conflicting configuration". 

I'm actually aware of that. However, since this error message comes from very high level, I decided not to touch. Here the message is coming from.

https://github.com/thin-edge/thin-edge.io/blob/e44a0ff2c4ff96f679fbc03283bf1da24d664c27/crates/core/tedge/src/main.rs#L80

PR #3328, opened 2025-01-14
test: missing test for custom operation execution using smartrest templates

Ruadhri17:
## Proposed changes
This PR adds missing test for custom operation execution using smartrest templates. Additionally, it fixes the script content in documentation that currently is not printing any information. 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_smartrest_template/smartrest_template.robot
+++ b/tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_smartrest_template/smartrest_template.robot
@@ -1,0 +1,42 @@
+*** Settings ***
+Resource            ../../../../resources/common.resource
+Library             Cumulocity
+Library             ThinEdgeIO
+
+Test Setup          Custom Setup
+Test Teardown       Get Logs
+
+Test Tags           theme:c8y    theme:operation    theme:custom    theme:smartrest    theme:template
+
+
+*** Test Cases ***
+smartrest template custom operation successful
+    Create SmartREST2 Template    ${CURDIR}/set_wifi.json    ${DEVICE_SN}
+    Should Have SmartREST2 Template    ${DEVICE_SN}
+
+    ${operation}=    Cumulocity.Create Operation
+    ...    description=do something
+    ...    fragments={"set_wifi":{"name":"Factory Wifi","ssid":"factory-onboarding-wifi","type":"WPA3-Personal"}}
+    Operation Should Be SUCCESSFUL    ${operation}
+
+    ${c8y_messages}=    Should Have MQTT Messages
+    ...    c8y/s/dc/${DEVICE_SN}
+    ...    minimum=1
+    ...    maximum=1
+    ...    message_contains=dm101,${DEVICE_SN},Factory Wifi,factory-onboarding-wifi,WPA3-Personal
+
+
+*** Keywords ***
+Custom Setup
+    ${DEVICE_SN}=    Setup
+    Set Suite Variable    $DEVICE_SN
+    Device Should Exist    ${DEVICE_SN}
+
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/set_wifi    /etc/tedge/operations/c8y/set_wifi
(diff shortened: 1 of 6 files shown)


tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_smartrest_template/smartrest_template.robot
```diff
@@ -0,0 +1,35 @@
...
+Library             ThinEdgeIO
+
+Test Setup          Custom Setup
+Test Teardown       Get Logs
+
+Test Tags           theme:c8y    theme:operation    theme:custom    theme:smartrest    theme:template
+
+
+*** Test Cases ***
+smartrest template custom operation successful
```
rina23q, line 13: This test is missing to upload the custom SmartREST template to c8y. I assume that this is the reason why system test failed.

Bravo555: Hmm, it works on my machine when I run the test locally

Bravo555 (changes requested):
I don't fully understand why even though I'm using the same tenant I can complete the test successfully and get the correct `dm101` message when running the test locally, but on the test runner the message is `dm 101` which probably causes the test to fail...

In any case, if the test requires a certain setup on the tenant that causes if the test to fail if it's not done, in this case registering the `custom_devmgmt` SmartREST template, then the test suite documentation should mention that this step should be done and how to do it, or else when running on another tenant one will receive an error and have no quick way of fixing it.

I think the best way would be for the test itself to send the template registration message as mentioned in https://github.com/thin-edge/thin-edge.io/pull/3328/files#r1914766428.

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_smartrest_template/smartrest_template.robot
reubenmiller: As Rina said the system test should create a new SmartREST template collection. It should use a randomized name so that it does not interfere with other existing SmartREST template collections.

reubenmiller: @Ruadhri17 Some pointers on how to implement this (it seems more complicated than it really is...it is not doing anything really new here, so you can look at other parts of the testing framework for clues and examples).

**Part 1:** Add underlying function to interact with SmartREST 2.0 template collections in the new https://github.com/thin-edge/c8y-test-core project

1. Create a new file called "assert_smartrest2.py" using a similar structure as used by [assert_inventory.py](https://github.com/thin-edge/c8y-test-core/blob/main/c8y_test_core/assert_inventory.py)
2. Add an new function which can upload a new SmartREST 2.0 template collection (from a file path, or given dictionary). The name (external identity) should be either given by the caller, or use a randomized alpanumeric name (e.g. `TST_raising_minor_boots`). The upload is just creating the managed object, though be aware you probably have to add a specific `.type` field to it, and also add an external identity it (as a follow up REST API call after the managed object has been created)

**Part 2:** Add new Keyword to the https://github.com/thin-edge/robotframework-c8y project

1. Add a new keyword, something like `Create SmartREST2 Template Collection`. The keyword should be simple as it should just call the underlying `c8y-test-core` method you added in part 1, and also add the SmartREST2 template collection managed object to the auto cleanup list so that the managed object will be automatically deleted after the test suite has finished (see this [example](https://github.com/thin-edge/robotframework-c8y/blob/da35eea0dfa3e4066aa2b8d51113d8cc4c902cdd/Cumulocity/Cumulocity.py#L1531))

**Part 3:** Use the new keyword in your test

1. Last part is to just use the new keyword you created in the previous part.

Bravo555 (approved):
Fixed by creating a smartrest template directly in the test, LGTM :+1:

rina23q (approved):
LGTM. Finally we have a test with a custom SmartREST template :partying_face:

PR #3338, opened 2025-01-17
refactor: fix flaky unit tests

jarhodes314:
## Proposed changes
This fixes some flaky unit tests I observed locally. One was the `mqtt_channel` tests, which were failing due to a genuine bug where the connection was closed before all messages were published. The other was a mapper test, which implicitly assumed messages would be delivered in a fixed order.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -16,6 +16,7 @@ use rumqttc::EventLoop;
 use rumqttc::Incoming;
 use rumqttc::Outgoing;
 use rumqttc::Packet;
+use std::task::Poll;
 use std::time::Duration;
 use tokio::time::sleep;
 
@@ -88,19 +89,22 @@ impl Connection {
 
         let (mqtt_client, event_loop) =
             Connection::open(config, received_sender.clone(), error_sender.clone()).await?;
+        let (awaiting_acks_sender, awaiting_acks_receiver) = mpsc::unbounded();
         tokio::spawn(Connection::receiver_loop(
             mqtt_client.clone(),
             config.clone(),
             event_loop,
             received_sender,
             error_sender.clone(),
+            awaiting_acks_receiver,
+            pub_done_sender,
         ));
         tokio::spawn(Connection::sender_loop(
             mqtt_client,
             published_receiver,
             error_sender,
             config.last_will_message.clone(),
-            pub_done_sender,
+            awaiting_acks_sender,
         ));
 
         Ok(Connection {
@@ -200,8 +204,17 @@ impl Connection {
         mut event_loop: EventLoop,
         mut message_sender: mpsc::UnboundedSender<MqttMessage>,
         mut error_sender: mpsc::UnboundedSender<MqttError>,
+        awaiting_acks: mpsc::UnboundedReceiver<()>,
+        done: oneshot::Sender<()>,
     ) -> Result<(), MqttError> {
+        let awaiting_acks = awaiting_acks.peekable();
+        futures::pin_mut!(awaiting_acks);
         loop {
+            if let Poll::Ready(None) = futures::poll!(awaiting_acks.as_mut().peek()) {
+                // If the channel is dropped, the sender loop has stopped
+                // and we've received an ack for every message published
+                mqtt_client.disconnect().await.unwrap();
+            }
             match event_loop.poll().await {
                 Ok(Event::Incoming(Packet::Publish(msg))) => {
                     if msg.payload.len() > config.max_packet_size {
@@ -252,6 +265,11 @@ impl Connection {
                     break;
                 }
 
+                Ok(Event::Incoming(Incoming::PubAck(_))) => {
+                    // Mark one message as consumed
+                    awaiting_acks.next().await;
+                }
+
                 Err(err) => {
                     error!("MQTT connection error: {err}");
 
@@ -266,6 +284,7 @@ impl Connection {
         // No more messages will be forwarded to the client
         let _ = message_sender.close().await;
         let _ = error_sender.close().await;
+        let _ = done.send(());
         Ok(())
     }
 
@@ -274,7 +293,7 @@ impl Connection {
         mut messages_receiver: mpsc::UnboundedReceiver<MqttMessage>,
         mut error_sender: mpsc::UnboundedSender<MqttError>,
         last_will: Option<MqttMessage>,
-        done: oneshot::Sender<()>,
+        mut awaiting_acks: mpsc::UnboundedSender<()>,
     ) {
         loop {
             match messages_receiver.next().await {
@@ -290,6 +309,8 @@ impl Connection {
                         .await
                     {
                         let _ = error_sender.send(err.into()).await;
+                    } else {
+                        awaiting_acks.send(()).await.unwrap();
                     }
                 }
             }
@@ -303,8 +324,6 @@ impl Connection {
                 .publish(last_will.topic, last_will.qos, last_will.retain, payload)
                 .await;
         }
-        let _ = mqtt_client.disconnect().await;
-        let _ = done.send(());
     }
 
     pub(crate) async fn do_pause() {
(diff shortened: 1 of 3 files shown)


crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -2086,27 +2086,37 @@ async fn json_custom_operation_status_multiple_operations_in_one_mqtt_message()
...
+    for _ in 0..6 {
+        messages.push(mqtt.recv().await.unwrap());
+    }
+    let (mut requests, mut responses): (Vec<_>, Vec<_>) = messages
+        .iter()
+        .map(|msg| (msg.topic.name.as_str(), msg.payload.as_str().unwrap()))
+        .partition(|(_topic, payload)| payload.starts_with("504,"));
 
+    // The messages might get processed out of order, we don't care about the ordering of the messages
+    requests.sort();
```
jarhodes314, line 2099: This probably isn't technically necessary as I imagine the requests are always sent in the order from the original c8y message, but I think the point still stands that the ordering is irrelevant.

crates/tests/mqtt_tests/src/test_mqtt_server.rs
```diff
@@ -115,7 +114,11 @@ fn spawn_broker() -> u16 {
         }
 
         match broker_thread.join() {
-            Ok(Ok(())) => unreachable!("`broker.start()` does not terminate"),
+            Ok(Ok(())) => {
+                // I don't know why it happened, but I have observed this once while testing
+                // So just log the error and retry starting the broker on a new port
+                eprintln!("MQTT-TEST ERROR: `broker.start()` should not terminate until after `spawn_broker` returns")
```
albinsuresh, line 120: But when this happens and when we retry the broker start on the next iteration, the client thread that's started (line. 137) in the last iteration also must be aborted somehow, right? Because once that client thread enters the loop, I don't see how it can break out even on a connection error because of the very specific `if let` check.

crates/tests/mqtt_tests/src/test_mqtt_server.rs
```diff
@@ -184,17 +187,14 @@ fn get_rumqttd_config(port: u16) -> Config {
...
-
     let mut servers = HashMap::new();
     servers.insert("1".to_string(), server_config);
 
     rumqttd::Config {
         id: 0,
         router: router_config,
         cluster: None,
-        console: Some(console_settings),
+        console: None,
```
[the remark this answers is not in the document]
albinsuresh: It appears to be defensive diagnostic mechanism kept in place to debug the server when some rare flaky failure occurs, as running the test again by turning it ON may not reproduce the failure.

albinsuresh (commented):
The main fix that ensures that all the messages to be published are exhausted looks fine. Just some queries/comments on the other bits.

crates/tests/mqtt_tests/src/test_mqtt_server.rs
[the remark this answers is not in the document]
didier-wenzek: We can safely ignore these console settings.

crates/common/mqtt_channel/src/connection.rs
```diff
@@ -252,6 +265,11 @@ impl Connection {
                     break;
                 }
 
+                Ok(Event::Incoming(Incoming::PubAck(_))) => {
+                    // Mark one message as consumed
+                    awaiting_acks.next().await;
+                }
```
didier-wenzek, line 271: We also have to wait for QoS 2 completion.

```suggestion
                Ok(Event::Incoming(Incoming::PubAck(_))) | Ok(Event::Incoming(Incoming::PubComp(_))) => {
                    // Mark one message as consumed
                    awaiting_acks.next().await;
                }
```

crates/common/mqtt_channel/src/connection.rs
```diff
@@ -200,8 +204,17 @@ impl Connection {
...
+        awaiting_acks: mpsc::UnboundedReceiver<()>,
+        done: oneshot::Sender<()>,
     ) -> Result<(), MqttError> {
+        let awaiting_acks = awaiting_acks.peekable();
+        futures::pin_mut!(awaiting_acks);
         loop {
+            if let Poll::Ready(None) = futures::poll!(awaiting_acks.as_mut().peek()) {
+                // If the channel is dropped, the sender loop has stopped
+                // and we've received an ack for every message published
+                mqtt_client.disconnect().await.unwrap();
```
didier-wenzek, line 216: I would rather ignore errors here.

```suggestion
                let _ = mqtt_client.disconnect().await;
```

didier-wenzek (commented):
Thank you for finding and fixing this bug on closing the connection a bit too soon. However, one has to do the same for QoS 2.

crates/common/mqtt_channel/src/connection.rs
```diff
@@ -200,8 +207,25 @@ impl Connection {
         mut event_loop: EventLoop,
         mut message_sender: mpsc::UnboundedSender<MqttMessage>,
         mut error_sender: mpsc::UnboundedSender<MqttError>,
+        awaiting_acks: mpsc::UnboundedReceiver<()>,
```
jarhodes314, line 210: I should replace this with something more appropriate now I'm no longer sending messages to it, I'm just using it to determine when the sender loop has finished executing.

crates/tests/mqtt_tests/src/test_mqtt_server.rs
jarhodes314: The client thread is not part of the loop, it's only started once we have a healthy broker.

didier-wenzek (commented):
Using a semaphore with a unique permit is a nice idea to notify the receiver loop from the sender loop that the connection has to be closed. However, I don't fully get how the receiver loop close when there is stream of input messages.

crates/common/mqtt_channel/src/tests.rs
```diff
@@ -1,30 +1,38 @@
...
-    let mut con = Connection::new(&mqtt_config).await?;
+        .with_session_name(uniquify!("test_client"))
+        .with_subscriptions(topic.try_into().unwrap());
+    let mut con = Connection::new(&mqtt_config).await.unwrap();
 
     // Any messages published on that topic ...
-    broker.publish(topic, "msg 1").await?;
-    broker.publish(topic, "msg 2").await?;
-    broker.publish(topic, "msg 3").await?;
+    broker.publish(topic, "msg 1").await.unwrap();
```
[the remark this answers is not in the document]
jarhodes314: Because there's no panic, `?` doesn't give any indication of which line failed, and that makes cases like this impossible to debug (if I don't know which message didn't publish, I've got almost no clue as to the nature of the bug).

crates/common/mqtt_channel/src/connection.rs
```diff
@@ -200,9 +208,37 @@ impl Connection {
...
+        let mut triggered_disconnect = false;
+        let mut disconnect_permit = None;
+
         loop {
-            match event_loop.poll().await {
+            let remaining_events_empty = event_loop.state.inflight() == 0;
+            if disconnect_permit.is_some() && !triggered_disconnect && remaining_events_empty {
+                let client = mqtt_client.clone();
+                tokio::spawn(async move { client.disconnect().await });
+                // tokio::fs::write("/tmp/thing.txt",format!("{:#?}", &events.event_loop.state)).await.unwrap();
```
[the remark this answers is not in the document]
jarhodes314: Oops. The debug print was enormous, so I had to write it to a file, but forgot to remove this line.

crates/common/mqtt_channel/Cargo.toml
```diff
@@ -17,14 +17,13 @@ log = { workspace = true }
 rumqttc = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
-tokio = { workspace = true, features = ["rt", "time"] }
+tokio = { workspace = true, features = ["rt", "time", "rt-multi-thread"] }
```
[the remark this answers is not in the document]
jarhodes314: It is used when running doctests as we use `#[tokio::main]` there. Trying to test in isolation without it breaks, and that made checking for flakiness hard.

didier-wenzek (approved):
Approved. Thank you for your perseverance in finding the more appropriate solution to properly detect end of connection.

didier-wenzek:
I confirm my approval after this commit using `tokio::select!` https://github.com/thin-edge/thin-edge.io/pull/3338/commits/24838098956ffcb44321cf40b9b35dfad3e6193d

PR #3344, opened 2025-01-21
Testing device registration using random walk

didier-wenzek:
## Proposed changes

- [x] using [proptest](https://proptest-rs.github.io/proptest/) to generate random - but realistic - registration scenario.
- [x] checking actual behavior against a simpler model 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/core/tedge_agent/proptest-regressions/entity_manager/tests.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
+# everyone who runs the test benefits from these saved cases.
+cc eb284ae3160cf4a071a3ca80175b2c72680b7c8ba19b8d577805ef1317eca2db # shrinks to registrations = [(MQTT, AddService { topic: "ac", props: [] }), (MQTT, AddService { topic: "ac", props: [] }), (MQTT, AddDevice { topic: "ac", props: [] }), (HTTP, AddService { topic: "ac", props: [("x", "4")] }), (HTTP, AddDevice { topic: "a", props: [] })]
```
didier-wenzek, line 7: This failing test is a genuine bug!

This translates to the following sequence of commands:

```
tedge mqtt pub te/device/a/service/c '{"@parent":"device/a//","@type":"service"}'
tedge mqtt pub te/device/a/service/c '{"@parent":"device/a//","@type":"service"}'
tedge mqtt pub te/device/ac// '{"@parent":"device/a//","@type":"child-device"}'
curl http://localhost:8000/tedge/entity-store/v1/entities -v -H "Content-Type: application/json" -d '{"@topic-id": "device/a/service/c", "@type": "service", "@parent": "device/a//"}'
curl http://localhost:8000/tedge/entity-store/v1/entities -v -H "Content-Type: application/json" -d '{"@topic-id": "device/a//", "@type": "child-device", "@parent": "device/main//"}'
```

and makes the agent panic:

```
thread 'tokio-runtime-worker' panicked at crates/core/tedge_api/src/store/pending_entity_store.rs:96:73:
called `Option::unwrap()` on a `None` value
```
 
Beyond fixing the bug, two points can be improved:

- further shrinking of the command sequence
- steps displayed as `tedge mqtt` and `curl` commands and not internal representations.

didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/3344/commits/bebfcdda9048a4bbd7d68ce68fda2319a07ab9af

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,542 @@
...
+use tedge_api::entity::EntityMetadata;
+
+#[tokio::test]
+async fn new_entity_store() {
+    let (mut entity_store, _mqtt_output) = entity::server("device-under-test");
+
+    assert_eq!(
+        entity::get(&mut entity_store, "device/main//").await,
+        Some(EntityMetadata::main_device("device-under-test".to_string()))
+    )
```
didier-wenzek, line 14: This test is failing and I will fixed it. However, I would like to understand why the method `EntityRegistrationMessage::main_device()` takes an optional external id for the device but ignores this value.

crates/core/tedge_api/src/store/pending_entity_store.rs
```diff
@@ -93,10 +93,11 @@ impl PendingEntityStore {
         let mut children = vec![];
         if let Some(direct_children) = self.orphans.remove(entity_tid) {
             for child in direct_children {
-                let pending_entity_cache = self.entities.remove(&child).unwrap();
```
didier-wenzek, line 96: The purpose of this method is not clear. It's not obvious why this method updates the pending entity store

- neither from the name (`take_cached_child_entities_data`) 
- nor the doc comment (`/// Returns all the children and their children recursively for a given parent`)

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: My bad. Should have removed that argument after moving the external id aspects from the entity store to the entity cache.

crates/core/tedge_api/src/store/pending_entity_store.rs
albinsuresh: For cached children, once the missing parent is registered, all those children can be registered with the entity store. That's when we extract/remove them from the pending store using this API and register them with the entity store. For some reason I thought the word `take` would imply that we're **taking** the owned values from it rather than references. May be, `remove` was more appropriate.

crates/core/tedge_agent/src/entity_manager/tests.rs
didier-wenzek: Okay, I will do that in this PR.

crates/core/tedge_api/src/store/pending_entity_store.rs
didier-wenzek: I will add appropriate doc comments.

crates/core/tedge_agent/src/entity_manager/tests.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3344/commits/2ae55c313cde87896695f866baeb3fd6c126b95a

crates/core/tedge_api/src/store/pending_entity_store.rs
didier-wenzek: See: https://github.com/thin-edge/thin-edge.io/pull/3344/commits/74db7e1c3dc37822ecf592f784dc84cc8ce8b48d

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,628 @@
...
+    //#![proptest_config(proptest::prelude::ProptestConfig::with_cases(1000))]
+    #[test]
+    fn it_works_for_any_registration_order(registrations in model::walk(10)) {
+        tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(async {
+            let (mut entity_store, _mqtt_output) = entity::server("device-under-test");
+            let mut state = model::State::new();
```
albinsuresh, line 27: Just FYI: Came across this [proptest-state-machine](https://proptest-rs.github.io/proptest/proptest/state-machine.html) crate which looked to be made for this kind of stuff.

TBH, after looking at some of their examples, I found this verbose version easier to comprehend than the abstractions and macro magic provided by that crate.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,628 @@
...
+    prop_compose! {
+        pub fn random_action_on(topic: String)(
+            protocol in random_protocol(),
+            action in 1..5,
+            props in random_props(2)
+        ) -> Command {
+            let topic = topic.to_owned();
+            let action = match action {
+                1|2 => Action::AddDevice{ topic, props },
+                3|4 => Action::AddService{ topic, props },
```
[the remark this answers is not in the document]
didier-wenzek: > Was it done to increase the probability of these 2 actions getting generated more than the remove counterparts?

Yes, this was the purpose. But I'm not convinced this a good idea. I will remove it.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,628 @@
...
+                5 => Action::RemService{ topic },
+                _ => Action::RemDevice{ topic },
```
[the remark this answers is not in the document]
didier-wenzek: Oops. A confirmation that https://github.com/thin-edge/thin-edge.io/pull/3344#discussion_r1936742433 was not a good idea!

crates/core/tedge_agent/src/entity_manager/tests.rs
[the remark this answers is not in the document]
didier-wenzek: Removed by https://github.com/thin-edge/thin-edge.io/pull/3344/commits/90d242ceb801cde2d056dc0b50530032fda4049b

crates/core/tedge_agent/src/entity_manager/tests.rs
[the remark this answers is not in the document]
didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/3344/commits/90d242ceb801cde2d056dc0b50530032fda4049b

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,628 @@
...
+                actions.0.push(action);
+                Just(actions)
+            })
+    }
+
+    pub fn walk(max_length: u32) -> impl Strategy<Value = Commands> {
+        if max_length == 0 {
+            Just(Commands(vec![])).boxed()
+        } else if max_length == 1 {
+            prop::collection::vec(random_action(), 0..1)
```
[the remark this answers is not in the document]
didier-wenzek: [`prop::collection::vec`](https://docs.rs/proptest/latest/proptest/collection/fn.vec.html) expects a [`SizeRange`](https://docs.rs/proptest/latest/proptest/collection/struct.SizeRange.html) which [constructor](https://docs.rs/proptest/latest/proptest/collection/struct.SizeRange.html#method.new) takes a [__`RangeInclusive`__](https://doc.rust-lang.org/nightly/core/ops/struct.RangeInclusive.html).

I changed the code to `prop::collection::vec(random_action(), 0..=1)`. See https://github.com/thin-edge/thin-edge.io/commit/90d242ceb801cde2d056dc0b50530032fda4049b

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,628 @@
...
+                1|2 => Action::AddDevice{ topic, props },
+                3|4 => Action::AddService{ topic, props },
+                5 => Action::RemService{ topic },
+                _ => Action::RemDevice{ topic },
+            };
+            Command { protocol, action }
+        }
+    }
+
+    pub fn random_action() -> impl Strategy<Value = Command> {
```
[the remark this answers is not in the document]
didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3344/commits/1d7234a357cf2b9bf9ae91aa48dbf4b537ac3d9b

crates/core/tedge_agent/src/entity_manager/tests.rs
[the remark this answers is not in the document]
albinsuresh: How the whole thing was working with the earlier value `(0..1)` which was just an exclusive `Range` and not `RangeInclusive` is still a mystery to me.

crates/core/tedge_agent/proptest-regressions/entity_manager/tests.txt
```diff
@@ -0,0 +1,9 @@
...
+cc aa8e0b76fd3142bf3cd1df1f453a4468defd437111b406887d760f68d84ad5c3 # shrinks to registrations = tedge http delete /tedge/entity-store/v1/entities/device/a// && tedge http delete /tedge/entity-store/v1/entities/device/main/service/a && tedge mqtt pub -r device/main/service/a '{"@parent":"device/main//","@type":"service","z":"1"}' && tedge http delete /tedge/entity-store/v1/entities/device/main/service/a && tedge mqtt pub -r device/ab// '' && tedge mqtt pub -r device/a// '' && tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/a//","@type":"child-device"}' && tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/a//","@type":"child-device"}' && tedge mqtt pub -r device/a// '{"@parent":"device/main//","@type":"child-device","z":"4"}'
```
albinsuresh, line 9: Shouldn't the earlier values also be normalised to this format?

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,628 @@
...
+        fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+            // mimicking a sequence of cli commands, with no extra quotes
+            // e.g:
+            //     tedge mqtt pub -r device/main/service/a '{"@parent":"device/main//","@type":"service","x":"9"}' \
+            //  && tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/c//","@type":"child-device","z":"5"}'
+            let mut sep = "";
+            for command in &self.0 {
+                f.write_str(sep)?;
+                f.write_str(format!("{command}").as_str())?;
+                sep = " && ";
```
albinsuresh, line 152: Optional:
```suggestion
                if f.alternate() {
                    sep = "  &&\n";
                } else {
                    sep = " && ";
                }
```
Just for an easy to read prettier version while pretty printing like:

```
tedge mqtt pub -r device/a// '{"@parent":"device/main//","@type":"child-device","z":"3"}' &&
 tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/a//","@type":"child-device"}' &&
 tedge mqtt pub -r device/a// '{"@parent":"device/main//","@type":"child-device"}' &&
 tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/a//","@type":"child-device"}' &&
 tedge mqtt pub -r device/cc/service/a '{"@parent":"device/cc//","@type":"service","y":"7"}' &&
 tedge mqtt pub -r device/cca// '{"@parent":"device/cc//","@type":"child-device","z":"5"}' &&
 tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/main/service/a","@type":"service","z":"8"}' &&
 tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/main/service/a","@type":"service","y":"5"}' &&
 tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/main/service/a","@type":"service","y":"8"}'
 ```
 
 **vs**

```
tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/b//","@topic-id":"device/b/service/b","@type":"service"}' && tedge mqtt pub -r device/bb// '{"@parent":"device/b//","@type":"child-device"}' && tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/b//","@topic-id":"device/b/service/b","@type":"service"}' && tedge mqtt pub -r device/bb// '{"@parent":"device/b//","@type":"child-device","z":"8"}' && tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/b//","@topic-id":"device/b/service/b","@type":"service","z":"8"}' && tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/b//","@topic-id":"device/b/service/b","@type":"service","z":"9"}' && tedge mqtt pub -r device/bb// '{"@parent":"device/b//","@type":"child-device","x":"0"}' && tedge mqtt pub -r device/bb// '{"@parent":"device/b//","@type":"child-device","z":"4"}' && tedge mqtt pub -r device/b/service/b '{"@parent":"device/b//","@type":"service"}'
```

File: crates/core/tedge_agent/proptest-regressions/entity_manager/tests.txt (hunk 9 above)
Comment (didier-wenzek) [reply crates/core/tedge_agent/proptest-regressions/entity_manager/tests.txt:9]: Yeah, but I have to do it manually :-(

File: crates/core/tedge_agent/src/entity_manager/tests.rs (hunk 10 above)
Comment (didier-wenzek) [reply crates/core/tedge_agent/src/entity_manager/tests.rs:152]: An issue is that `proptest` removes the newlines in `proptest-regressions/entity_manager/tests.txt`. I will try to find a way that displays well in proptest file as well as test unit output.

Comment (didier-wenzek) [reply crates/core/tedge_agent/src/entity_manager/tests.rs:152]: `Formatter.alternate()` works nicely. I will push a fix.

Comment (didier-wenzek) [reply crates/core/tedge_agent/src/entity_manager/tests.rs]: See https://github.com/thin-edge/thin-edge.io/pull/3344/commits/beaa728d9f93695c70710aaf927f1e5f04d6c39e

File: crates/core/tedge_agent/proptest-regressions/entity_manager/tests.txt (hunk 9 above)
Comment (didier-wenzek) [reply crates/core/tedge_agent/proptest-regressions/entity_manager/tests.txt:9]: Done: https://github.com/thin-edge/thin-edge.io/pull/3344/commits/9283fb7d0976ecaecd3e5c3516f276edad1aba5f

File: crates/core/tedge_agent/src/entity_manager/tests.rs [hunk 11]
```diff
@@ -0,0 +1,655 @@
...
+    assert_eq!(
+        entity::get(&mut entity_store, "device/main//").await,
+        Some(EntityMetadata::main_device())
+    )
+}
+
+#[tokio::test]
+async fn removing_an_unknown_child_using_mqtt() {
+    let registrations = vec![
+        // tedge mqtt pub -r te/device/a// ''
```
didier-wenzek, line 24: What's the expected behavior of sending an empty MQTT message to an entity topic?

I was expecting this to delete the entity (which in this case has not been registered yet),
but this half register the entity:

- it is inserted in the `EntityTree` maintained by the entity store (using default metadata: i.e as a child device of main)
- but it is not listed as an entity (despite not being pending).

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -0,0 +1,655 @@
...
+
+proptest! {
+    //#![proptest_config(proptest::prelude::ProptestConfig::with_cases(1000))]
+    #[test]
+    fn it_works_for_any_registration_order(registrations in model::walk(10)) {
+        tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap()
+        .block_on(check_registrations(registrations))
```
didier-wenzek, line 43: Extracting the check from the proptest helps to replay issues found with proptest.

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: > What's the expected behavior of sending an empty MQTT message to an entity topic?

The agent(entity store) will do nothing as deletions are supported only over HTTP. But the mapper would still react to that MQTT message by removing the that entry from its entity cache.

> but this half registers the entity: (using default metadata: i.e as a child device of main).

That's a bug in the auto-registration logic. A side effect of the entity manager actor of subscribing to `te/+/+/+/+/#` topic that includes `te/+/+/+/+` topic as well. It should have been `te/+/+/+/+/+/#` or to be completely on the safer side subscribe to the data topics specifically like `[te/+/+/+/+/m/+, te/+/+/+/+/e/+, te/+/+/+/+/a/+, te/+/+/+/+/cmd/+/+, ...]`.

didier-wenzek: > That's a bug in the auto-registration logic.

Okay, I will fix that.

didier-wenzek: Doing nothing on `tedge mqtt pub -r te/device/a// ''` is somehow inconsistent with the fact that an `HTTP DELETE /tedge/entity-store/v1/entities/device/a//` publish a retain empty message on `te/device/a//`. How can a client like the mapper tell the difference between the two cases?

It would be better for the agent to react on clearing an entity topic by deleting the entity.

albinsuresh: > It would be better for the agent to react on clearing an entity topic by deleting the entity.

I agree. Now that `tedge-agent` is the sole owner of the entity store, we can treat empty MQTT messages also as a request to delete the entire tree.

Bravo555 (commented):
Very nice change, I didn't find any other blockers, apart from the one found by Albin regarding needing to clearly define and change behaviour for empty registration messages to mean device deletion. Because that makes the tests not pass, I'm not approving the PR yet, but after it's fixed, it's good to merge as far as I'm concerned.

Regarding hypothetical improvements to how we utilise proptest:
- Rewriting sequences of actions from proptest output to unit test is a bit tedious, but when looking at proptest book it seemed to me that creating new unit test out of sequences found with proptest was a relatively common operation, so I'm wondering if rather than for ease-of-reading, the `Commands` `Debug impl` shouldn't be optimised for ease-of-copying to quickly create new Rust struct values.
- Currently the command sequences lack shrinking support, but I believe that's being worked on.
- As for checking the entity store implementation correctness: I believe this would be a good opportunity to try to fit in the invariant testing approach. An insertion operation has invariants that it doesn't modify other entities apart from target and its parent, and has effects that `Get` operation should return the entity and it should be a part of its parent's children. Similarly, a delete operation doesn't affect other entities and has the opposite effect that after executing it the target entity shouldn't be returned by a `Get` operation. I believe by defining and asserting these effects and invariants for each operation type to be a more scalable approach, having the additional advantage that in the test we formalize the semantics of an operation from the POV of the caller, instead of just restating the implementation.

didier-wenzek:
> clearly define and change behaviour for empty registration messages to mean device deletion. Because that makes the tests not pass, I'm not approving the PR yet, but after it's fixed, it's good to merge as far as I'm concerned.

Yeah, I will fix that. Discovering such issues is the main point of this PR. Hence I see this failing test as a success ;-)

> Rewriting sequences of actions from proptest output to unit test is a bit tedious, but when looking at proptest book it seemed to me that creating new unit test out of sequences found with proptest was a relatively common operation, so I'm wondering if rather than for ease-of-reading, the `Commands` `Debug impl` shouldn't be optimised for ease-of-copying to quickly create new Rust struct values.

Indeed, the test `removing_an_unknown_child_using_mqtt` has been created after a failing sequence found by proptest.
I will update the failing test output to include both the `tedge` commands and the Rust struct commands, to ease two levels of testing.

> Currently the command sequences lack shrinking support, but I believe that's being worked on.

Yeah, I would be nice to have this.

> As for checking the entity store implementation correctness: I believe this would be a good opportunity to try to fit in the invariant testing approach. An insertion operation has invariants that it doesn't modify other entities apart from target and its parent, and has effects that `Get` operation should return the entity and it should be a part of its parent's children. Similarly, a delete operation doesn't affect other entitieremoving_an_unknown_child_using_mqtts and has the opposite effect that after executing it the target entity shouldn't be returned by a `Get` operation. I believe by defining and asserting these effects and invariants for each operation type to be a more scalable approach, having the additional advantage that in the test we formalize the semantics of an operation from the POV of the caller, instead of just restating the implementation.

Great idea! Indeed implementing a model against which the real implementation is tested is a pain. Adding external contract checks and internal contract `debug_assert()` will strengthen the checks.

didier-wenzek:
> > Rewriting sequences of actions from proptest output to unit test is a bit tedious, but when looking at proptest book it seemed to me that creating new unit test out of sequences found with proptest was a relatively common operation, so I'm wondering if rather than for ease-of-reading, the `Commands` `Debug impl` shouldn't be optimised for ease-of-copying to quickly create new Rust struct values.

> Indeed, the test `removing_an_unknown_child_using_mqtt` has been created after a failing sequence found by proptest.
I will update the failing test output to include both the `tedge` commands and the Rust struct commands, to ease two levels of testing.

Implemented by https://github.com/thin-edge/thin-edge.io/pull/3344/commits/57823db1debb6733d2c3f22ccaf13d146bbdd82f

Unit test failures are now displayed along these lines:

```
minimal failing input: registrations = 
    // tedge http post /tedge/entity-store/v1/entities '{"@parent":"device/main//","@topic-id":"device/main/service/a","@type":"service","y":"5"}'
    (protocol:HTTP,action:AddService(topic:"a",props:[("y","5")]))
    // tedge mqtt pub -r te/device/main/service/b '{"@parent":"device/main//","@type":"service"}'
    (protocol:MQTT,action:AddService(topic:"b",props:[]))
```

crates/core/tedge_agent/src/entity_manager/tests.rs
didier-wenzek: >> It would be better for the agent to react on clearing an entity topic by deleting the entity.

> I agree. Now that `tedge-agent` is the sole owner of the entity store, we can treat empty MQTT messages also as a request to delete the entire tree.

Implemented by https://github.com/thin-edge/thin-edge.io/pull/3344/commits/5009140e9b8980ef68f5b38ebd7e9bebdb417698

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -196,19 +196,23 @@ mod model {
 
             for command in &self.0 {
                 f.write_str(sep)?;
-                f.write_str(format!("{command}").as_str())?;
                 if f.alternate() {
-                    sep = "\n    "; // On test unit output, print each command on a new line
+                    // On test unit output, print each command on a new line (using Rust notation)
+                    sep = "\n    ";
+                    f.write_str(format!("// {command}{sep}").as_str())?;
```
albinsuresh, line 260: I was wondering why this ron representation wasn't added at the `Command` level itself and not limited to `Commands`, so that we get these visual improvements even when we have any other collection of `Command` instances . I'm assuming it was done intentionally to keep the `tests.txt` content minimal with just one representation.

albinsuresh (approved):
LGTM.

crates/core/tedge_agent/src/entity_manager/tests.rs
didier-wenzek: > I'm assuming it was done intentionally to keep the tests.txt content minimal with just one representation.

Yes. I try to keep `test.txt` as simple as possible

Issue #3345, opened 2025-01-21
tedge-agent is not restarted during tedge connect resulting in potential mismatch between mapper and tedge-agent


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

If thin-edge.io configuration is changed after already running `tedge connect c8y` once, then the **tedge-agent** service will not be restarted and pick up the updated settings. This is mostly applicable if settings such as the `mqtt.client.port` are changed.

In standard setups, the user is unaware that `tedge connect c8y` starts the **tedge-agent** service in the background. This means that when tedge configuration is changed, then they are unaware that **tedge-agent** service might need to be restarted.

The following is an example how users can get into thin-edge.io into a "half-working" state where the mapper is successfully connected to Cumulocity, but the tedge-agent is still listening to the MQTT broker on the incorrect port (and hence won't respond to any operations):

1. Configure thin-edge.io and try 

    ```sh
    tedge cert create --device-id example
    tedge config set c8y.url example.com
    tedge cert upload c8y
    tedge connect c8y
    ```

2. Discover there is a problem with the mosquitto port being used

3. Change the MQTT port used by mosquitto and thin-edge.io

    ```sh
    tedge config set mqtt.bind.port 1884
    tedge config set mqtt.client.port 1884
    ```

4. Try connecting thin-edge.io to Cumulocity (again)

    ```sh
    tedge connect c8y
    ```

Currently the way to "fix" the problem would be to get the user to restart the already running **tedge-agent** using the following command:

```sh
systemctl restart tedge-agent
```

**Describe the solution you'd like**

Restart the **tedge-agent** service instead of just starting it during `tedge connect c8y` and `tedge reconnect c8y`.

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
An alternative solution would be to add some configuration detection in the **tedge-agent** where it detects and configuration changes, and reloads it if necessary. For example, if the user changes the mqtt.client.port, then it would be smart enough to stop the existing MQTT client and reconnect to the given port. Though this might be tricky when there are dependent configuration changes which are done over time (e.g. changing the mqtt client host AND port), the user had to issue multiple commands, and it is only valid once the changes the MQTT bind settings have been changed, and mosquitto has been restarted (so the complexity increases).

reubenmiller:
Given the complexity to determine whether the tedge-agent needs to be restarted or not, we'll stick with the default not to restart it, and the user must do this.

PR #3347, opened 2025-01-21
feat: restart tedge-agent on tedge connect

reubenmiller:
## Proposed changes

Restart the tedge-agent service (instead of just starting it) when calling `tedge connect <cloud>` to ensure it matches any settings that may have changed (e.g. mqtt.client.port) since it was last started.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3345

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/system_services/manager_ext.rs
+++ b/crates/common/tedge_config/src/system_services/manager_ext.rs
@@ -4,12 +4,20 @@ use anyhow::Context as _;
 /// Extension trait for `SystemServiceManager`.
 pub trait SystemServiceManagerExt {
     fn start_and_enable_service(&self, service: SystemService<'_>) -> anyhow::Result<()>;
+    fn restart_and_enable_service(&self, service: SystemService<'_>) -> anyhow::Result<()>;
     fn stop_and_disable_service(&self, service: SystemService<'_>) -> anyhow::Result<()>;
 }
 
 impl SystemServiceManagerExt for &dyn SystemServiceManager {
     fn start_and_enable_service(&self, service: SystemService<'_>) -> anyhow::Result<()> {
         self.start_service(service)
+            .with_context(|| format!("Failed to start {service}"))?;
+        self.enable_service(service)
+            .with_context(|| format!("Failed to enable {service}"))
+    }
+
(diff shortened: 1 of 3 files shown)


reubenmiller:
Looks like we have a system test which covers the opposite case (introduced in 415ffaaee7f10188ef472d4786c78584e9d6db57). Investigating the PR where it was introduced and compare the two conflicting requirements.

didier-wenzek:
> Do you think dynamically reloading the configuration used by the tedge-agent would be feasible here?

Why not but this will be tedious. The issue is that the impact of a config change is deeply hidden in numerous independent components.

albinsuresh:
From what I understood, the primary intent of the PR #2505 was just to prevent the agent from being stopped on a disconnect (which is even more relevant now since profiled connections are possible). Since this PR doesn't change that aspect, it is fine from that POV.

My only concern about restart on `connect` is the possible disruption to any operations that were in-progress (esp the built-in ones like software update). For example, if two profiled c8y connections: `cloud` and `edge` were running and a config download operation triggered from `cloud` was in-progress, `tedge connect c8y --profile edge` would unnecessarily disrupt the operation of the `cloud`. If `tedge-agent` supported seamless resumption of operations, this isn't a big deal. But that's not the case AFAIK (any incomplete operations are marked failed on restart), at least for the built-in operations.

reubenmiller:
> From what I understood, the primary intent of the PR #2505 was just to prevent the agent from being stopped on a disconnect (which is even more relevant now since profiled connections are possible). Since this PR doesn't change that aspect, it is fine from that POV.
>
> My only concern about restart on `connect` is the possible disruption to any operations that were in-progress (esp the built-in ones like software update). For example, if two profiled c8y connections: `cloud` and `edge` were running and a config download operation triggered from `cloud` was in-progress, `tedge connect c8y --profile edge` would unnecessarily disrupt the operation of the `cloud`. If `tedge-agent` supported seamless resumption of operations, this isn't a big deal. But that's not the case AFAIK (any incomplete operations are marked failed on restart), at least for the built-in operations.

I don't think it will be an issue, as the `tedge connect c8y` and `tedge reconnect c8y` are generally only run when the user is fixing something, so the likelihood that the tedge-agent is already doing something like a software-update is unlikely. And even it it is, then the persistence of operations shouldn't be an issue for it.

Issue #3339, opened 2025-01-20
REST API for querying entity store


albinsuresh:
**Is your feature request related to a problem? Please describe.**

Currently the visibility into the entity store managed by the agent is limited to just debug logs of the agent. An external API for users to query the same would be useful for the customers and would be helpful in debugging issues as well.

**Describe the solution you'd like**

Add a `GET /v1/entities` REST API to the entity store service of the tedge http server to query the entity store as proposed in this [spec](https://github.com/thin-edge/thin-edge.io/blob/main/design/decisions/0005-entity-registration-api.md#query-entities)

**Additional context**

Filtering of entities to be supported using query parameters as proposed in the spec.

PR #3355, opened 2025-01-23
feat: REST API for querying entity store

albinsuresh:
## Proposed changes

Basic entity store query REST API implementation: `GET /tedge/entity-store/v1/entities` to list all entities.
Additional filtering capabilities to be added in follow-up PRs.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3339

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -33,6 +33,7 @@ use serde_json::Value as JsonValue;
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::collections::HashSet;
+use std::collections::VecDeque;
 use std::path::Path;
 
 // In the future, root will be read from config
@@ -514,6 +515,14 @@ impl EntityStore {
     pub fn cache_early_data_message(&mut self, message: MqttMessage) {
         self.pending_entity_store.cache_early_data_message(message)
     }
+
+    pub fn list_entity_tree<'a>(
+        &'a self,
+        root: Option<&'a EntityTopicId>,
+    ) -> Result<Vec<&'a EntityMetadata>, entity_store::Error> {
+        self.entities
+            .list_entity_tree(root.unwrap_or_else(|| self.main_device()))
+    }
 }
 
 /// In-memory representation of the entity tree
@@ -648,6 +657,26 @@ impl EntityTree {
 
             self.entities.remove(topic_id);
             removed_entities.push(topic_id.clone())
+        }
+    }
+
+    pub fn list_entity_tree<'a>(
+        &'a self,
+        root: &'a EntityTopicId,
+    ) -> Result<Vec<&'a EntityMetadata>, Error> {
+        if let Some(metadata) = self.entities.get(root).map(|node| node.metadata()) {
+            let mut topic_ids = VecDeque::from(vec![root]);
+            let mut entities = vec![metadata];
+
+            while let Some(topic_id) = topic_ids.pop_front() {
+                let (child_topics, children): (Vec<_>, Vec<_>) =
+                    self.children(topic_id).into_iter().unzip();
+                topic_ids.extend(child_topics);
+                entities.extend(children);
+            }
+            Ok(entities)
+        } else {
+            Err(Error::UnknownEntity(root.to_string()))
         }
     }
 }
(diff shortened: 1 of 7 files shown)


tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
```diff
@@ -962,6 +962,67 @@ def delete_smartrest_one_template(self, template_id: str):
...
+        if not device:
+            log.info("Device has not been setup, so no logs to collect")
+            return
+
+        output = device.execute_command("curl -f http://localhost:8000/tedge/entity-store/v1/entities")
+        entities = json.loads(output.stdout)
+        return entities
+
+
+    @keyword("Assert Contains Entity")
```
reubenmiller, line 996: It's convention to use `Should Contain ...`in the keyword. It does not have to match the python method name.

```suggestion
    @keyword("Should Container Entity")
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -33,6 +35,65 @@ CRUD apis
...
+    Register entity    device/child1//    child-device    device/main//
+    Register entity    device/child2//    child-device    device/main//
+    Register entity    device/child20//    child-device    device/child2//
+    Register entity    device/child21//    child-device    device/child2//
+    Register entity    device/child210//    child-device    device/child21//
+    Register entity    device/child2100//    child-device    device/child210//
+    Register entity    device/child22//    child-device    device/child2//
+
+    ${output}=    Execute Command    curl http://localhost:8000/tedge/entity-store/v1/entities
+    ${entities}=    Convert String To Json    ${output}
```
reubenmiller, line 63: Better the use the Keyword that you created in the ThinEdgeIO.py library.

```suggestion
    ${entities}=    List Entities
```

reubenmiller: After that, you can remove the `Collections` and `JSONLibrary` imports at the top of the test file.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -33,6 +35,65 @@ CRUD apis
...
+
+    ${entities}=    Assert Contains Entity    {"@topic-id": "device/main//", "@type": "device"}    ${entities}
+    ${entities}=    Assert Contains Entity
```
reubenmiller, line 66: You're overwriting the `${entities}` values. So you can just drop the variable assignment in all of the assertions, e.g.

```
Should Contain Entity    {"@topic-id": "device/main//", "@type": "device"}    entities=${entities}
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -58,3 +119,9 @@ Entity auto-registration over MQTT
 Custom Setup
     ${DEVICE_SN}=    Setup
     Set Suite Variable    $DEVICE_SN
+
+Register entity
+    [Arguments]    ${topic_id}    ${type}    ${parent}
+
+    Execute Command
+    ...    curl -X POST http://localhost:8000/tedge/entity-store/v1/entities -H 'Content-Type: application/json' -d '{"@topic-id": "${topic_id}", "@type": "${type}", "@parent": "${parent}"}'
```
reubenmiller, line 127: It might be worth adding this as a keyword to ThinEdgeIO.py

E.g. `Register Entity`

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
```diff
@@ -962,6 +962,67 @@ def delete_smartrest_one_template(self, template_id: str):
...
+        """
+        if not entities:
+            entities = self.list_entities()
+
+        if isinstance(entity, str):
+            entity = json.loads(entity)
+
+        matches = entities
+        if entity:
+            matches = [entity for entity in entities if entity == entity]
```
reubenmiller, line 1021: Since renaming the function's `item` argument, it makes the list comprehension very difficult to read due to the variable shadowing happening (or I wouldn't be surprised if it does not work at all!)

Why not just keep the `item` name?

```suggestion
            matches = [entity for entity in entities if entity == item]
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Oops... I guess the broken list comprehension made the tests pass.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -650,6 +656,36 @@ impl EntityTree {
             removed_entities.push(topic_id.clone())
         }
     }
+
+    pub fn list<'a>(&'a self, topic_id: &'a EntityTopicId) -> Vec<&'a EntityMetadata> {
+        if let Some(metadata) = self.entities.get(topic_id).map(|node| node.metadata()) {
+            let mut topic_ids = VecDeque::from(vec![topic_id]);
+            let mut entities = vec![metadata];
+
+            self.list_internal(&mut topic_ids, &mut entities);
```
didier-wenzek, line 665: `list_internal` seems over-complicated and doesn't need to be recursive. I would even use a simple `while` loop.

```suggestion
            while let Some(topic_id) = topic_ids.pop_front() {
                let (child_topics, children): (Vec<_>, Vec<_>) =
                    self.children(topic_id).into_iter().unzip();
                topic_ids.extend(child_topics);
                entities.extend(children);
            }
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
reubenmiller: In that case I would recommend just creating an small test case which just calls the endpoint and checks that the response status code is ok...though thinking about it, I don't think explicitly testing the api endpoint is really required..the ThinEdgeIO is already part of the code so it is closely linked to the test anyway.

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
```diff
@@ -962,6 +962,67 @@ def delete_smartrest_one_template(self, template_id: str):
...
 
+    @keyword("List Entities")
+    def list_entities(
+            self, name: str = None 
+    ) -> List[Dict[str, Any]]:
+        """
+        Get entity list from the device using the entity store query REST API
+
+        Args:
+            name (str, optional): Device name to fetch the entity list from
```
albinsuresh, line 973: I'm having second thoughts about this parameter. Actually, this entity store service or the HTTP server itself is only available on the main device. Even if `tedge-agent` is installed on a child device, these endpoints won't be available. So, accepting this parameter is pointless. In fact, it should be restricted to the main device only. How can I do that?

reubenmiller: Though it is needed as the api can be reachable from other devices, so we still need it.

It would be very easy to detect if a child device is being used and switch to the correct hostname/ip address instead of using localhost. When the tedge-agent is running on a child device it needs to know the main device hostname/ip address, so it should already have the info needed.

reubenmiller: But we can live it simple for now and keep the `name` (device name) parameter. The test will just fail if the user tries to run the keyword on a child device (which would still be technically correct ;))

crates/core/tedge_api/src/entity_store.rs
didier-wenzek: fixed by https://github.com/thin-edge/thin-edge.io/pull/3355/commits/55975e07897183d946916560587f00d700882735

didier-wenzek (approved):
Approved

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
albinsuresh: Resolved by 55975e0

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by 55975e0

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
albinsuresh: Leaving it as-is.

didier-wenzek (approved):
I confirm my approval.

reubenmiller (approved):
Tests (and test library changes) look good.

Issue #3369, opened 2025-01-29
device.id should be derived from device certificates when in use


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

If the cloud connection uses a certificate, then the device.id used for any cloud communication should be derived from the certificate's Common Name (CN).

**Describe the solution you'd like**

* When using certificates, the `device.id` should only be used as a default value when creating a certificate via the `tedge cert create` command if the `--device-id` flag is not provided by the user.
* Cloud connections should only use the device.id from the certificate being used for the connection

The change in behaviour is best highlighted in the follow cases:

### Case 1: Use explicit device id during cert creation

```sh
tedge config set device.id foo
tedge cert create --device-id bar

# => CN=bar
```

### Case 2: Use default device.id

```sh
tedge config set device.id foo
tedge cert create

# => CN=foo
```

### Case 3: Use default device.id

```sh
tedge cert create --device-id foo
tedge config set device.id bar
tedge connect c8y

# => connect: device.id=foo
```

### Case 4: Use default device.id to create the cert

```sh
tedge config set device.id foo
tedge cert create
# CN=foo
tedge config set device.id bar

tedge connect c8y
# => connect: device.id=foo
```

**Describe alternatives you've considered**

**Additional context**

didier-wenzek:
> I think we also have to change cloud mappers to use the value from CN, not from the internal API device.id().

I would really avoid to do that. This sounds like saying `device.id()` is not the device id! Somehow this is puzzling everything only to support a legacy case we do not recommend (i.e. smartrest 1 and password authentication)

I do think that:

- If there is a certificate, `device.id()` must return the CN of that certificate, ignoring any entry in `tedge.toml`.
- `device.id()` should only use the id set in `tedge.toml` when there is no certificate.

And I would even go a step further:

- `tedge config get device.id` should returns the CN of the certificate when there is one - ignoring the id set in the toml file if any.
- this will be less error prone, the value being returned by `tedge config get device.id` being the value actually used
- this might be a bit confusing but only on the effect of `tedge config set device.id`
- this works well for users using only certificates
- this works well for users using only passwords
- this might confuse users mixing passwords and certificates.

So revisiting the use cases:

### Case 1: Use explicit device id during cert creation

```sh
$ tedge config set device.id foo
$ tedge cert create --device-id bar
$ tedge config get device.id
bar

# => CN=bar && used by tedge connect and the mapper.
```

### Case 2: Use default device.id

[...]

PR #3372, opened 2025-01-31
fix: device.id is derived from device certificate if certificate exists

rina23q:
## Proposed changes

* `device.id` always uses the value from a device certificate. If no certificate exists, it returns a value from the `tedge.toml` file. Both don't exist, finally it returns an error.

* `tedge connect` doesn't validate if device certficate's CN matches the value from the `tedge.toml` file. It just consume the value from `device.id`.

* `tedge cert create` creates a cert, which CN is determined by this order. 1) the value given by the `--device-id` option, 2) the value from `tedge config get <corresponding_device_id_key>` 

## Types of changes

## Paste Link to the issue
#3369

## Checklist


reubenmiller:
> tedge cert create always creates a cert with the value given by the --device-id option. It doesn't matter if user configure device.id in the tedge config settings explicitly.

The phrasing sounds a little strange here. The precedence of source of the device.id value just follows the cli convention of: argument, env variable, configuration, where the first value present is used.

rina23q:
> > tedge cert create always creates a cert with the value given by the --device-id option. It doesn't matter if user configure device.id in the tedge config settings explicitly.
> 
> The phrasing sounds a little strange here. The precedence of source of the device.id value just follows the cli convention of: argument => env variable => configuration, where the first value present is used.

Yeah I fixed the phrasing. Prioritizing the argument is what this PR fixed. Note that the order of env variable => configuration is the responsibility of the tedge config settings and should be this order for all configs.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -8,11 +8,9 @@ use super::upload::*;
 use anyhow::anyhow;
 use camino::Utf8PathBuf;
 use clap::ValueHint;
-use tedge_config::explicit_device_id;
```
didier-wenzek, line 11: `tedge_config::explicit_device_id` can be fully removed as it was only used here.

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
```diff
@@ -78,54 +78,114 @@ Run tedge cert create with cloud profile
...
+    ...    tedge config set c8y.device.key_path --profile foo /etc/tedge/device-certs/tedge-private-key@foo.pem
+
+    ### No device ids are explicitly set
+    # create certificate without explicit device.id in tedge config setting
     ${output}=    Execute Command    tedge cert create --device-id input
     Should Contain    ${output}    CN=input
+    ${output}=    Execute Command    tedge config get device.id    strip=${True}
+    Should Be Equal    ${output}    input
+    ${output}=    Execute Command    tedge config get c8y.device.id    exp_exit_code=1
+    ${output}=    Execute Command    tedge config get c8y.device.id --profile foo    exp_exit_code=1
```
didier-wenzek, line 96: I was expecting a cascading behavior here: i.e. if there is  no specific device id for a cloud profile, then the general device id is used.

~~Is this because explicit paths have set for the misc cloud profiles?~~

I got it: this is because no device id has been set in the config file.

I would be good to have a more consistent cascading behavior.

didier-wenzek (commented):
Okay we have now a more consistent behavior regarding device id /  tedge config / tedge connect.

However, it would be good to add 4 independent tests matching exactly the 4 cases described by https://github.com/thin-edge/thin-edge.io/issues/3369. The current tests are useful but a bit complicated to follow because very dependent of the whole sequence of check vs tear-down steps.

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
rina23q: > I would be good to have a more consistent cascading behavior.

If no specific `device.cert_path` is set for a cloud profile, the result here will be different.

```sh
# Remove specific paths 
$ tedge config unset c8y.device.cert_path
$ tedge config unset c8y.device.cert_path --device foo

# create a new cert for the generic one
$ tedge config cert create --device-id input

# get device ids
$ tedge config get device.id
input
$ tedge config get c8y.device.id
input
$ tedge config get c8y.device.id --profile foo
input
```

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
```diff
@@ -78,54 +78,114 @@ Run tedge cert create with cloud profile
...
+    Execute Command    tedge config get device.id    exp_exit_code=1
+    Execute Command    tedge config set c8y.device.cert_path /etc/tedge/device-certs/tedge-certificate@default.pem
+    Execute Command    tedge config set c8y.device.key_path /etc/tedge/device-certs/tedge-private-key@default.pem
+    Execute Command
+    ...    tedge config set c8y.device.cert_path --profile foo /etc/tedge/device-certs/tedge-certificate@foo.pem
+    Execute Command
+    ...    tedge config set c8y.device.key_path --profile foo /etc/tedge/device-certs/tedge-private-key@foo.pem
+
+    ### No device ids are explicitly set
+    # create certificate without explicit device.id in tedge config setting
```
reubenmiller, line 90: I wouldn't call this "explicit" as inferring a value from configuration is really be explicit.

```suggestion
    # create certificate without device.id in tedge config setting
```

crates/core/tedge/src/cli/certificate/cli.rs
rina23q: removed in ef49eeb23438484e78e27beffeba18a9e9e950b5

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
rina23q: rephrased some comment lines in ef49eeb23438484e78e27beffeba18a9e9e950b5

rina23q:
> However, it would be good to add 4 independent tests matching exactly the 4 cases described by #3369. The current tests are useful but a bit complicated to follow because very dependent of the whole sequence of check vs tear-down steps.

Added a new suite to cover all the four cases in ef49eeb23438484e78e27beffeba18a9e9e950b5

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
```diff
@@ -78,54 +78,122 @@ Run tedge cert create with cloud profile
...
 
 Device ID derivation
+    Execute Command    tedge config get device.id    exp_exit_code=1
+    Execute Command    tedge config set c8y.url example.com --profile foo
+
+    ### All cert/key paths are the default
+    ${output}=    Execute Command    tedge cert create --device-id input
+    Should Contain    ${output}    CN=input
+    Validate device IDs    input    input    input
+    Execute Command    tedge cert remove
```
didier-wenzek, line 88: This is the cascading behavior I was expecting.

didier-wenzek (approved):
Approved. Thanks for taking the time to make more consistent how the device id is used across the misc tedge cert / config and connect API.

tests/RobotFramework/tests/tedge/device_id.robot
```diff
@@ -0,0 +1,69 @@
...
+    ${output}=    Execute Command    tedge cert create
+    Should Contain    ${output}    CN=${FOO}
+
+    ${output}=    Execute Command    tedge config get device.id    strip=${True}
+    Should Be Equal    ${output}    ${FOO}
+
+    ${output}=    Execute Command    tedge connect c8y    ignore_exit_code=${True}
+    Should Contain    ${output}    device id: ${FOO}
+
+Use default device.id 2
```
reubenmiller, line 38: Can you add a more descriptive test name here? It is too similar to `Use default device.id` which makes it unclear what the test case is about.

tests/RobotFramework/tests/tedge/tedge_cert_create.robot
reubenmiller: This behaviour also caught me by surprise as setting a cloud profile certificate path changes the `device.id` value returned, even if the cloud profile certificate does not exist...though after reviewing it, it seems to make sense as the act of creating a different cert path for a cloud profile, will decouple it from the generic/default certificate...so it makes it easier for system users to control the device.id inheritance without forcing them to create the certificate.

reubenmiller (approved):
Approved (only 1 minor test case renaming required)

Overall it would benefit from splitting the "Device ID derivation" into individual test cases (and possibly moving it to a new Test Suite so it can use a single "Suite Setup" and reuse the same device for all test cases with suitable cleanup logic between tests)...however this is not a blocker for merging the PR.

tests/RobotFramework/tests/tedge/device_id.robot
rina23q: Renamed to `Use device id from cert`

tests/RobotFramework/tests/tedge/device_id/device_id_usecases.robot
```diff
@@ -0,0 +1,69 @@
+*** Settings ***
+Documentation       This suite covers all usecases in the issue #3369
+
+Resource            ../../../resources/common.resource
+Library             Cumulocity
+Library             ThinEdgeIO
+Library             ../../.venv/lib/python3.11/site-packages/robot/libraries/String.py
```
reubenmiller, line 7: This looks to be an accidental import. Check if the `String` RobotFramework is required, if so, then change to:

```suggestion
Library             String
```

rina23q: Thanks for catching this. IDE added it automatically, it was not intended and safe to remove.

Issue #3375, opened 2025-02-03
Fix `acceptor_rejects_untrusted_client_certificates` following hyper upgrade


jarhodes314:
In #3371, I ignored `acceptor_rejects_untrusted_client_certificates` since I wasn't getting the desired error message, but I believe the underlying functionality is still behaving as expected. The point of checking the error is correct is intended to ensure the process failed in the correct way. If we didn't do this, the client could, for example, fail to trust the server certificates and the connection would be immediately rejected rather than failing on a client certificate error. The type of error we get is also important as when working, it makes it easy for clients to diagnose the issue on their side rather than just getting some generic TLS failure which is near impossible to debug. The issue may well be caused by some refactoring on `reqwest`'s side, as I had to change some error handling there.

PR #3382, opened 2025-02-05
test: re-enable client certificate test

jarhodes314:
## Proposed changes
Remove the `#[ignore]` attribute on one of the client certificate tests. Due to a change in the underlying certificate verifier, the error  rustls produces is slightly different to what it was previously, and this was causing the test to fail. We aren't interested in the specifics of the error message produced, we simply care that the error arose from the client certificate verification and not some bug in the test code. Having updated the test to check for the updated error, I've verified that it still fails for otherwise failing connections (e.g. the client not trusting the server certificate). Therefore I think a simple update to the test to reflect the new behaviour is sufficient here.

## Types of changes

## Paste Link to the issue
- #3375

## Checklist

### DIFF
--- a/crates/common/axum_tls/src/acceptor.rs
+++ b/crates/common/axum_tls/src/acceptor.rs
@@ -171,7 +171,6 @@ mod tests {
     }
 
     #[tokio::test]
-    #[ignore]
     async fn acceptor_rejects_untrusted_client_certificates() {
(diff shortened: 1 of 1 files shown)


albinsuresh (approved):

didier-wenzek (approved):
Approved

PR #3383, opened 2025-02-05
feat: Query parameter support for entity store query API

albinsuresh:
## Proposed changes

- [x] Add query parameter support for the query API
- [x] Supported parameters:
  - [x] `root`
  - [x] `parent`
  - [x] `type`
- [x] Documentation

Decided not to add the `depth` parameter as agreed [here](https://github.com/thin-edge/thin-edge.io/pull/3383#issuecomment-2639211046).
Multiple values for the same parameter is also not supported.

## Types of changes

## Paste Link to the issue

#3339 

## Checklist

### DIFF
--- a/docs/src/operate/registration/register.md
+++ b/docs/src/operate/registration/register.md
@@ -85,17 +85,27 @@ GET /v1/entities/{topic-id}
 curl http://localhost:8000/tedge/entity-store/v1/entities/device/child01
 ```
 
-## Query Entities
-
-### Query All Entities
-
-List all entities registered with thin-edge starting from the `main` device.
+## Query entities
 
 **Endpoint**
 
 ```
 GET /v1/entities
 ```
+
+**Query parameters**
+
+| Parameter | Description                                                            | Examples                            |
+|-----------|------------------------------------------------------------------------|-------------------------------------|
+| `root`    | Entity tree starting from the given `root` node (including it) | `device/child2//`                   |
+| `parent`  | Direct child entities of the given `parent` entity (excluding it)             | `device/main//`                     |
+| `type`    | Entities of the given entity `type`                          | `main`, `child-device` or `service` |
+
+The following restrictions apply:
+* Multiple values can not be specified for the same parameter.
+* The same parameter can not be repeated multiple times.
+* The `root` and `parent` parameters can not be used together.
+
 
 **Responses**
 
@@ -107,7 +117,7 @@ GET /v1/entities
           "@type": "device"
       },
       {
-          "@topic-id": "device/main/service/tedge-agent",
+          "@topic-id": "device/main/service/service0",
           "@type": "service",
           "@parent": "device/main//"
       },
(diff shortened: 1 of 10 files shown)


didier-wenzek:
I don't see a `depth` parameter as useful, and I would not spent time implementing support for that.

In practice the entity tree will be mostly never deep.

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
```diff
@@ -1015,20 +1015,22 @@ def register_entity(
 
     @keyword("List Entities")
     def list_entities(
-            self, device_name: str = None 
+            self, params: Dict[str, str] = None, device_name: str = None
```
reubenmiller, line 1018: I don't think the putting all of the query parameter mappings in the params is a good idea here (it just makes it a lot harder to use).

```suggestion
            self, entity_type: Optional[str, str] = None, parent: Optional[str, str] = None, root: Optional[str, str] = None, device_name: str = None
```

And if people still want to use a dictionary, then we could just add kwargs support to the function signature, and then people can pass this stuff as a dictionary (as that is more python like)

didier-wenzek (approved):
Approved

docs/src/operate/registration/register.md
```diff
@@ -108,39 +112,320 @@ GET /v1/entities
...
-          "@type": "child-device",
-          "@parent": "device/child0//"
       }
+      ...
   ]
   ```
+* 404: Not Found
+  ```
+  Entity not found with topic id: device/unknown/topic/id
+  ```
```
reubenmiller, line 130: I'm not to sure about returning 404 when the query is empty.

Generally having no results in a query is not an error. Trying to access a resource that does not exist, would warrant a 404 response.

If you look at the REST API from Cumulocity, the query endpoints (e.g. `GET /inventory/managedObjects' don't return a 404 NOT FOUND HTTP status code when there are no matching results, it just returns an empty array. Technically speaking, I believe the "resource" in `GET /v1/entities` is '/v1/entities' itself...which does exist.

albinsuresh: Yeah, this was a confusing case for me as well and my original implementation was retuning empty array for non-existent entity. I changed it to 404 because empty array is returned for a leaf entity that doesn't have any children as well. So, when we get an empty array, we won't be able to differentiate if it's because there are no entities matching the query or if it's a non-existent entity. That's why I changed it to 404 for the latter case so that there is a clear distinction.

But, if the convention is empty array, then I'll change it back.

reubenmiller:
### Item 1: Response body Content-Type on errors.

Currently errors responses are returned to the user with the `Content-Type: text/plain`. This makes it difficult for simple clients to parse the responses (especially when writing scripts).

Given that the responses of successful queries are json, having the error messages in the same format makes it easier to parse by clients, for example:

Using an invalid value for the `type` query parameter yields an error, however that error is not visible by the user if they're piping it to `jq`:

```sh
curl -s 'http://localhost:8000/tedge/entity-store/v1/entities?type=asdfasdf' | jq
parse error: Invalid numeric literal at line 1, column 8
```

So for example if the error message is in json format, it could look something like this:

```sh
curl -s 'http://localhost:8000/tedge/entity-store/v1/entities?type=asdfasdf' | jq

{
    "error": "Invalid entity type: asdfasdf"
}
```

The user is still free to control whether they want to pipe results at all from a failing curl command using `curl -f`, but still in this case the error message is still lost.

reubenmiller:
### Item 2: Empty query parameter values should be treated as missing query parameter values

Treating empty values as missing values makes it very easy for users to write a simple wrapper around the calls without having to complicated string matching.

For example, the following currently returns an error.

```sh
curl -s 'http://localhost:8000/tedge/entity-store/v1/entities?type=' -v
```

But if we were to treat an empty query parameter value the same as an absent value, then it makes it much easier to build the query parameters.

For example, below shows two implementations that do exactly the same thing (wrap the type filter so the user can provide a given value), and it shows how simple it could be versus if we treat empty values as an error (due to being an invalid type).

```sh
# Option 1: treat empty query param values as non-existing values
FILTER_BY_TYPE="${1:-}"
curl -s 'http://localhost:8000/tedge/entity-store/v1/entities?type=${FILTER_BY_TYPE}' -v

# Option 2:
# vs. having to add a lot of empty string checking when building the query parameters
FILTER_BY_TYPE="${1:-}"
FILTER_ARGS=
if [ -n "$FILTER_BY_TYPE" ]; then
    if [ -n "$FILTER_ARGS" ]; then
        FILTER_ARGS="$FILTER_ARGS&type=${FILTER_BY_TYPE}"
    else
        FILTER_ARGS="type=${FILTER_BY_TYPE}"
    fi
fi
curl -s 'http://localhost:8000/tedge/entity-store/v1/entities?${FILTER_ARGS}' -v
```

docs/src/operate/registration/register.md
reubenmiller: Yes I would change it back to returning an empty array. This is much easier to deal with for people consuming the API as generally error handling generally wants to split "there are no results" from "the api request failed".

reubenmiller: And if you want to check the existence of a device, then you should be using `GET http://localhost:8000/tedge/entity-store/v1/entities/{entity}`

docs/src/operate/registration/register.md
```diff
@@ -86,18 +86,22 @@ GET /v1/entities/{topic-id}
...
 
 ```
 GET /v1/entities
 ```
 
+**Query parameters**
+
+| Parameter | Description                                                  | Examples                            |
+|-----------|--------------------------------------------------------------|-------------------------------------|
+| `root`    | The topic id of the root node where to start the search from | `device/child2//`                   |
```
reubenmiller, line 101: Might be worthwhile mentioned that using the `root` query parameter includes the root element.

"The topic id of the root node where to start the search from (including the root entity)"

reubenmiller:
### Item 3: Deleting a non-existent entity should return 404 status code

Deleting a resource that does not exist should return a HTTP 404 status code, currently it returns 200, and an empty json array.

```sh
curl -XDELETE http://localhost:8000/tedge/entity-store/v1/entities/device/child2// | jq
```

reubenmiller:
### Item 4: Return the full entity representation 

It is more consistent if the list of deleted entites objects is returned in its entirety instead of just the `@topic-id` field.

It is also easier to interpret the returned data, because the fields are clearly labeld.

**Current**

```
curl -XDELETE http://localhost:8000/tedge/entity-store/v1/entities/device/child2// | jq
[
  "device/child21/service/service210",
  "device/child210/service/service2100",
  "device/child2100//",
  "device/child210//",
  "device/child21//",
  "device/child22//",
  "device/child2/service/service20",
  "device/child2/service/service21",
  "device/child20//",
  "device/child2//"
]
```

**Expected**

```
curl -XDELETE http://localhost:8000/tedge/entity-store/v1/entities/device/child2// | jq
[
    {
        "@topic-id": "device/child20//",
        "@parent": "device/child2//",
        "@type": "child-device"
    },
  // ..
]
```

reubenmiller (changes requested):
Ok, overall looks nice, just a few minor things to change.

But I'm sorry I just realised that I've been reviewing the entire entity api not just the query api (probably because I was looking at the "register.md" file without a diff view in my editor). So all of the non-query related changes can be ignored and done in a follow up PR (we shouldn't corrupt the scope).

docs/src/operate/registration/register.md
albinsuresh: Reverted that change in 1de9671a

docs/src/operate/registration/register.md
albinsuresh: Resolved by https://github.com/thin-edge/thin-edge.io/commit/1de9671aaddddc34a5e2b00b8d3076a72018064b

albinsuresh:
> Ok, overall looks nice, just a few minor things to change.

Addressed Items 1 and 2 in 416d499 and 1de9671a respectively. Items 3 and 4 related to the DELETE API would be addressed in a separate PR.

reubenmiller (approved):
Approved

Issue #3341, opened 2025-01-20
Accept initial twin data in the entity registration message


albinsuresh:
**Is your feature improvement request related to a problem? Please describe.**

When an entity is registered using an entity registration message, in addition to the key fields like `@type`, `@parent` and `@id`, two additional twin properties: `name` and `type` are also accepted, which are pushed to the cloud along with the registration message. Any additional key-value pairs published in the registration message are not mapped/ignored.

It'd be better if these additional key-value pairs are treated like initial twin data for that entity and published to the cloud as additional inventory fragments. Further updates to these fragments can be done via their corresponding `twin/<data-type>` topics.

**Describe the solution you'd like**

Accept all additional non-key fields in the entity registration message, including `name` and `type` as initial twin data for that entity and publish those fragments to their corresponding `twin/<data-type>` topics so that the mapper picks those twin updates and publishes those to the cloud. Any further updates to these twin topics would also be handled automatically by the mapper.

**Additional context**

To avoid having two copies of the same data on two different topics, the agent may re-publish a registration message back to the broker with all the non-key fragments from the user-provided registration message stripped out, as copies of those fragments are available via their `twin` topics.

The agent can refuse further updates to the entity registration topic by republishing the original registration message with only the key fields back to the broker.

PR #3392, opened 2025-02-11
refactor: No default service type in entity store

albinsuresh:
## Proposed changes

The entity store need not add a default service type to the entity if the service type is not provided in the registration message. The service type is mandatory only for the Cumulocity mapper and not for any other tedge components.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -377,10 +377,9 @@ impl Agent {
             let telemetry_cache_size = 0; // Agent need not cache any data messages, the mapper would
 
             let main_device = EntityRegistrationMessage::main_device(None);
-            let entity_store = EntityStore::with_main_device_and_default_service_type(
+            let entity_store = EntityStore::with_main_device(
                 mqtt_schema.clone(),
                 main_device,
-                self.config.service.ty.clone(),
                 telemetry_cache_size,
                 state_dir,
                 clean_start,
--- a/crates/core/tedge_agent/src/entity_manager/tests.rs
+++ b/crates/core/tedge_agent/src/entity_manager/tests.rs
@@ -197,15 +197,13 @@ mod entity {
     ) -> (EntityStoreServer, SimpleMessageBox<MqttMessage, NoMessage>) {
         let mqtt_schema = MqttSchema::default();
         let main_device = EntityRegistrationMessage::main_device(Some(device_id.to_string()));
-        let default_service_type = "default_service_type".to_string();
         let telemetry_cache_size = 0;
         let log_dir = TempDir::new().unwrap();
         let clean_start = true;
         let entity_auto_register = true;
-        let entity_store = EntityStore::with_main_device_and_default_service_type(
+        let entity_store = EntityStore::with_main_device(
             mqtt_schema.clone(),
             main_device,
-            default_service_type,
             telemetry_cache_size,
             log_dir,
             clean_start,
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -66,10 +66,9 @@ const MQTT_ROOT: &str = "te";
 /// );
 /// let registration_message = EntityRegistrationMessage::try_from(&mqtt_message).unwrap();
 ///
-/// let mut entity_store = EntityStore::with_main_device_and_default_service_type(
+/// let mut entity_store = EntityStore::with_main_device(
 ///     MqttSchema::default(),
 ///     registration_message,
-///     "service".into(),
 ///     0,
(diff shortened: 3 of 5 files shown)


reubenmiller:
@albinsuresh Though the tedge setting used to control the default service type is `service.type` which has no affiliations to clouds, so from that point of view, it is not such as bad thing if the type is added to the entity store object representation.

albinsuresh:
> I believe the mapper was adding it before just because the entity store was previously being handled by the mapper...so it might be better to re-evaluate this.

No, the mapper was (and still is) using this `service.type` value while converting service registration messages (without an explicit `type` in the payload) and health status messages, as the `type` is mandatory in the `102` message (This config setting was indeed added because of this requirement). That's why it's still working even after I dropped it from the registration message in this PR.

> Another advantage to adding the value from the service.type at the entity store level, is it makes the registered entities more consistent when users want to read from the entity store or subscribe to different services

In that case, wouldn't it make sense to add the `type` field for the auto-registered `child-device` entities as well so that the behaviour is consistent for both entity types? I was just trying to bring this consistency. I opted for removing it from the service registration message because it was never a mandatory field for any component other than the c8y mapper. So, it just felt like the agent adding some mapper specific metadata into an otherwise agnostic payload, when the mapper could have handled it on its own.

But, if you feel that this `type` information would be consumed by other components as well, then I can keep it the way it is (drop this PR) and bring the consistency by adding the `type` for auto-registered `child-device`s as well (in another PR).

didier-wenzek (commented):
I'm in favor of removing the *default* service type from the entity store, as proposed by this PR.

- Why should the agent be responsible for a property it doesn't used?
- The default service type, being Cumulocity related, should be configured under `c8y.*` and, why not, be cloud profile specific.

reubenmiller:
> I'm in favor of removing the _default_ service type from the entity store, as proposed by this PR.
> 
> * Why should the agent be responsible for a property it doesn't used?
> * The default service type, being Cumulocity related, should be configured under `c8y.*` and, why not, be cloud profile specific.

ok. Seems like there is more pushback than expected, and it is not that important of a feature.
which is not using the data so even though it does not 

Though long term I think we need to revisit how we approach making thin-edge.io data available locally (e.g. accessible by local machine UI's, other components etc.). So having the entity store api applying a normalization to entity registration does have an advantage as local clients/UIs can subscribe to the data and get a unified view (e.g. the service type field will always be set...either from the component publishing the registration message, or the main tedge-agent enriches the registration messages with the missing information).

albinsuresh:
> Ok. Seems like there is more pushback than expected, and it is not that important of a feature.

Don't get me wrong. I was just trying to fix the consistency issue of only services having the `type` and child-devices not. Which could be achieved by either dropping one or keeping both. I'm fine with either.

That being said, considering #3341 in the pipeline, we're going to drop both the `name` and `type` from the registration message and move them as `twin` messages. So, dropping the `type` field as done in this PR should be fine anyway.

didier-wenzek:
> Though long term I think we need to revisit how we approach making thin-edge.io data available locally (e.g. accessible by local machine UI's, other components etc.). So having the entity store api applying a normalization to entity registration does have an advantage as local clients/UIs can subscribe to the data and get a unified view (e.g. the service type field will always be set...either from the component publishing the registration message, or the main tedge-agent enriches the registration messages with the missing information).

I agree with that, although I'm also in favor of merging this PR as a first cleaning step.

- Setting the default service type is currently hard-coded deeply in the entity store implementation, while this is both specific (only related to Cumulocity) and incomplete (no default type for child devices).
- To get an unified view, why not a template approach? Giving the users the ability to provide a bunch of properties with default values to be injected by the agent on registration if not set by the component publishing the registration message.

albinsuresh:
> To get an unified view, why not a template approach? Giving the users the ability to provide a bunch of properties with default values to be injected by the agent on registration if not set by the component publishing the registration message.

Something along the lines of the `inventory.json` file currently managed by the mapper, but at the agent level?

didier-wenzek:
> > To get an unified view, why not a template approach? Giving the users the ability to provide a bunch of properties with default values to be injected by the agent on registration if not set by the component publishing the registration message.
> 
> Something along the lines of the `inventory.json` file currently managed by the mapper, but at the agent level?

Yes, with specific templates for child-devices and services.   ... but not in this PR.

didier-wenzek (approved):
Approved.

Even if a better solution that has still to be designed, I see this PR as a first cleaning step.

reubenmiller (approved):
Approved.

Though this small PR did raise a few questions about the exact role of the tedge-agent when running as the main device and the new tedge http api which now offers entity store api...but all of these points can be discussed in the future.

PR #3400, opened 2025-02-14
feat: Initial twin data from entity registration message

albinsuresh:
## Proposed changes

- [x] `tedge-agent` treats all non-tedge fragments received in the entity registration message as initial twin data and publishes them to the corresponding twin topics.
- [x] The c8y-mapper, while converting these registration messages, publishes these twin values as inventory updates.
- [x] Treat `name` and `type` fragments just like any other twin data and allow those to be updated via the `twin` topics.
- [x] Prevent the mapper from clearing `name` and `type` fragments of entities as C8y doesn't support the same. 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/entity_manager/server.rs
+++ b/crates/core/tedge_agent/src/entity_manager/server.rs
@@ -1,4 +1,5 @@
 use async_trait::async_trait;
+use serde_json::Value;
 use tedge_actors::LoggingSender;
 use tedge_actors::MessageSink;
 use tedge_actors::Sender;
@@ -129,13 +130,12 @@ impl EntityStoreServer {
                 Ok(registered) => {
                     for entity in registered {
                         for (fragment_key, fragment_value) in entity.reg_message.other {
-                            let twin_channel = Channel::EntityTwinData { fragment_key };
-                            let topic = self
-                                .mqtt_schema
-                                .topic_for(&entity.reg_message.topic_id, &twin_channel);
-                            let message =
-                                MqttMessage::new(&topic, fragment_value.to_string()).with_retain();
-                            self.publish_message(message).await;
+                            self.publish_twin_data(
+                                &entity.reg_message.topic_id,
+                                fragment_key,
+                                fragment_value,
+                            )
+                            .await;
                         }
                     }
                 }
@@ -159,11 +159,7 @@ impl EntityStoreServer {
                 Ok(entities) => {
                     for entity in entities {
                         let message = entity.to_mqtt_message(&self.mqtt_schema).with_retain();
-                        if let Err(err) = self.mqtt_publisher.send(message).await {
-                            error!(
-                                "Failed to publish auto-registration messages for the topic: {topic_id} due to {err}",
-                            )
-                        }
+                        self.publish_message(message).await;
                     }
                 }
                 Err(err) => {
@@ -174,6 +170,18 @@ impl EntityStoreServer {
                 }
             }
         }
+    }
+
+    async fn publish_twin_data(
+        &mut self,
+        topic_id: &EntityTopicId,
+        fragment_key: String,
+        fragment_value: Value,
+    ) {
+        let twin_channel = Channel::EntityTwinData { fragment_key };
+        let topic = self.mqtt_schema.topic_for(topic_id, &twin_channel);
+        let message = MqttMessage::new(&topic, fragment_value.to_string()).with_retain();
+        self.publish_message(message).await;
     }
 
     async fn publish_message(&mut self, message: MqttMessage) {
@@ -205,11 +213,7 @@ impl EntityStoreServer {
 
         if !registered.is_empty() {
             let message = entity.to_mqtt_message(&self.mqtt_schema);
-            if let Err(err) = self.mqtt_publisher.send(message.clone()).await {
-                error!(
-                    "Failed to publish the entity registration message: {message:?} due to {err}",
-                )
-            }
+            self.publish_message(message).await;
         }
         Ok(registered)
     }
@@ -224,9 +228,7 @@ impl EntityStoreServer {
                 .with_retain()
                 .with_qos(QoS::AtLeastOnce);
 
-            if let Err(err) = self.mqtt_publisher.send(clear_entity_msg).await {
-                error!("Failed to publish clear message for the topic: {topic} due to {err}",)
-            }
+            self.publish_message(clear_entity_msg).await;
         }
 
         deleted
--- a/crates/core/tedge_agent/src/entity_manager/server.rs
+++ b/crates/core/tedge_agent/src/entity_manager/server.rs
@@ -129,7 +129,7 @@ impl EntityStoreServer {
             Ok(entity) => match self.entity_store.update(entity.clone()) {
                 Ok(registered) => {
                     for entity in registered {
-                        for (fragment_key, fragment_value) in entity.reg_message.other {
+                        for (fragment_key, fragment_value) in entity.reg_message.twin_data {
                             self.publish_twin_data(
                                 &entity.reg_message.topic_id,
                                 fragment_key,
--- a/crates/core/tedge_agent/src/entity_manager/server.rs
+++ b/crates/core/tedge_agent/src/entity_manager/server.rs
@@ -125,13 +125,26 @@ impl EntityStoreServer {
         }
 
         match EntityRegistrationMessage::try_from(&message) {
-            Ok(entity) => {
-                if let Err(err) = self.entity_store.update(entity.clone()) {
+            Ok(entity) => match self.entity_store.update(entity.clone()) {
+                Ok(registered) => {
+                    for entity in registered {
+                        for (fragment_key, fragment_value) in entity.reg_message.other {
+                            let twin_channel = Channel::EntityTwinData { fragment_key };
+                            let topic = self
+                                .mqtt_schema
+                                .topic_for(&entity.reg_message.topic_id, &twin_channel);
+                            let message =
+                                MqttMessage::new(&topic, fragment_value.to_string()).with_retain();
+                            self.publish_message(message).await;
+                        }
+                    }
+                }
+                Err(err) => {
                     error!(
                         "Failed to register entity registration message: {entity:?} due to {err}"
                     );
                 }
-            }
+            },
             Err(()) => error!("Failed to parse {message} as an entity registration message"),
         }
     }
@@ -160,6 +173,13 @@ impl EntityStoreServer {
                     );
                 }
             }
+        }
+    }
+
+    async fn publish_message(&mut self, message: MqttMessage) {
+        let topic = message.topic.clone();
+        if let Err(err) = self.mqtt_publisher.send(message).await {
+            error!("Failed to publish the message on topic: {topic:?} due to {err}");
         }
     }
 
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -347,8 +347,8 @@ impl CumulocityConverter {
         let mut messages = vec![];
 
         // Parse the optional fields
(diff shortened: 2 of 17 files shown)


tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
```diff
@@ -254,8 +254,8 @@ Nested child devices support sending inventory data via tedge topic
     Execute Command    tedge mqtt pub --retain "te/device/${nested_child}///twin/type" ''
     Execute Command    tedge mqtt pub --retain "te/device/${nested_child}///twin/name" ''
     ${mo}=    Device Should Have Fragments    type
-    Should Be Equal    ${mo["type"]}    thin-edge.io-child
-    Should Be Equal    ${mo["name"]}    ${nested_child}
+    Should Be Equal    ${mo["type"]}    NewType
+    Should Be Equal    ${mo["name"]}    NewName
```
albinsuresh, line 258: I honestly don't know why we're repeating this same test for main device, child device, nested child device and with builtin bridge. The logic is the same for all cases. So, I'm even considering keeping these checks only for one case and removing the rest.

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -170,8 +169,8 @@ impl EntityCache {
             external_id: Some(external_id.clone()),
             r#type: entity.r#type,
             parent,
-            other: entity.other,
-            twin_data: Map::new(),
+            other: entity.other.clone(),
```
albinsuresh, line 172: Now that this `other` field represents initial twin data, this could be merged with the `twin_data` field. I can't think of any benefits in keeping the **initial** twin data cached separately from the **current** twin data forever.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
didier-wenzek: I would not remove these tests as they are testing things which are different from the user point of view: twin data can be attached to the main device or any child devices. Running the same tests with and without the builtin doesn't harm.

The problem is that the test *code* is duplicated. Could the test be rewritten so the feature (setting twin data in the cloud) can be checked with a single test that is launched is several context (main vs child device, mosquitto vs builtin bridge)?

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -125,13 +125,26 @@ impl EntityStoreServer {
...
+                Ok(registered) => {
+                    for entity in registered {
+                        for (fragment_key, fragment_value) in entity.reg_message.other {
+                            let twin_channel = Channel::EntityTwinData { fragment_key };
+                            let topic = self
+                                .mqtt_schema
+                                .topic_for(&entity.reg_message.topic_id, &twin_channel);
+                            let message =
+                                MqttMessage::new(&topic, fragment_value.to_string()).with_retain();
+                            self.publish_message(message).await;
```
didier-wenzek, line 138: Rather than a `publish_message()` method, I would add a `publish_twin_data()` method, to focus on the purpose rather than the mechanism.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -344,49 +344,75 @@ impl CumulocityConverter {
...
+        if let Some(reg_message) = reg_message {
+            messages.push(reg_message);
         }
+
+        for (fragment_key, fragment_value) in input.other.iter() {
+            if fragment_key == "name" || fragment_key == "type" {
+                // Skip converting the name and type fields as they are already included in the registration message
+                continue;
+            }
+            let twin_messages = self.convert_twin_fragment(
```
didier-wenzek, line 406: Isn't the mapper reacting *twice* on twin data? On the registration messages and on the twin messages derived by the agent from the registration messages?

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
didier-wenzek: I have always been confused by these two fields, `other` vs `twin_data`, and I would be happy to have one removed if there are no impacts from a user perspective.

By the way, why is `other` tagged with as `serde(flatten)` and `twin_data` as `serde(skip)`?

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: When the same twin data is delivered again, they are filtered out as "duplicate values" in the `inventory::try_convert_twin_fragment` function. Twin data is converted to inventory updates only if they really update the existing values. Else they are ignored.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
albinsuresh: Yeah, we could do something like that with test templates. Will create a follow-up task for that.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: I'll add the `publish_twin_data` wrapper on top of `publish_message`. I introduced `publish_message` to be reused by all other functions in this module.

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
reubenmiller: Though sometimes abstracting the tests too much makes it more difficult to run selected tests (e.g. RobotFramework Templates don't actually allow you to run an individual test case using the VSCode plugin)...so sometimes copying/pasting is a more pragmatic approach (given the test cases don't change that much)

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
albinsuresh: Resolved by 36b07fb

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Resolved by a49d4cb

rina23q (commented):
Overall it looks really good, especially I really like the data struct is simplified (`other` and `twin_data` are merged to `twin_data`). Mainly I have questions.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -344,49 +344,75 @@ impl CumulocityConverter {
...
                     self.entity_cache.parent_external_id(entity_topic_id)?;
+                let display_name = display_name.unwrap_or(external_id.as_ref());
+                let display_type = display_type.unwrap_or("thin-edge.io-child");
 
                 let child_creation_message = child_device_creation_message(
                     external_id.as_ref(),
-                    display_name,
-                    display_type,
+                    Some(display_name),
+                    Some(display_type),
```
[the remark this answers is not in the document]
albinsuresh: From the SmartREST point-of-view of the `child_device_creation_message` API, accepting `name` and `type` as optional is still valid, even though our usages are always providing concrete values.

rina23q (approved):
LGTM

tests/RobotFramework/tests/cumulocity/telemetry/child_device_telemetry.robot
albinsuresh: Considering both the feedback, I went ahead with a far simpler approach in 45e1149 instead of using templates.

didier-wenzek (approved):
Approved

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -56,16 +56,38 @@ Register child device with defaults via MQTT
...
+    ...    message_contains=${CHILD_SN}
+    Should Have MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/maintenance_mode
+    ...    message_contains=true
+    Should Have MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/maintenance_window
+    ...    message_contains=5
+
+    Device Should Exist    ${CHILD_SN}
+    Device Should Have Fragments    name
```
reubenmiller, line 87: We should check the values as well (as we're assuming the mapping translated the values correctly here) by using `Device Should Have Fragment Values`.

The `Device Should Have Fragments` is useful just to check the presence of a "value", but it does check what it actually is.

Below shows an example of the amended assertion.

```
Device Should Have Fragment Values    name\="${CHILD_SN}"
```

albinsuresh: Resolved by d22d521

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
```diff
@@ -56,16 +56,36 @@ Register child device with defaults via MQTT
...
+    ...    te/device/${CHILD_SN}///twin/name
+    ...    message_contains=${CHILD_SN}
+    Should Have MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/maintenance_mode
+    ...    message_contains=true
+    Should Have MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/maintenance_window
+    ...    message_contains=5
+
+    ${foo}=    Device Should Exist    ${CHILD_SN}
```
reubenmiller, line 86: Is this variable used for anything? If not, then it should be deleted (as I'm sure the Robocop linter will complain about it).

tests/RobotFramework/tests/cumulocity/inventory/inventory_update.robot
```diff
@@ -33,3 +63,49 @@ Custom Setup
...
     ThinEdgeIO.Disconnect Then Connect Mapper    c8y
+
+Validate inventory fragment updates via twin topics
+    [Arguments]    ${device_xid}
+
+    IF    '${device_xid}' == '${DEVICE_SN}'
+        ${device_id}=    Set Variable    main
+    ELSE
+        ${device_id}=    Set Variable    ${device_xid}
+    END
```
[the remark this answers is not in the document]
reubenmiller: Long term we might need to evaluate how we write tests and how much should keywords need to know about their tests, as the topic scheme is defined in two places (once in the calling function, and once in the test)...so we need to make sure the topic scheme is aligned in both.

But in the future we'll create a more generic Keyword to help users get the correct topic e.g. `Get tedge Topic` once we open up the RobotFramework to users.

tests/RobotFramework/tests/cumulocity/registration/registration_lifecycle.robot
albinsuresh: Oops. Resolved in 8eaef40.

tests/RobotFramework/tests/cumulocity/inventory/inventory_update.robot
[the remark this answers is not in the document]
reubenmiller: For clarity this is the part I'm referring to:

```
Update child device inventory fragments via twin topics
    ${child_sn}=    Get Random Name
    Execute Command    tedge mqtt pub --retain 'te/device/${child_sn}//' '{"@type":"child-device","@id":"${child_sn}"}'

    Validate inventory fragment updates via twin topics    ${child_sn}
```

The test case already knows the topic root and topic prefix, `te/device/${child_sn}//'` as it is publishing to it, but then it only passes the `${child_sn}` to the `Validate inventory fragment updates via twin topics` keyword, where the keyword also does the same topic reconstruction, so the keyword is assume that we have the same topic scheme...passing the topic directly to the function would just eliminate the chance of misalignment...but this is more a minor point so I won't block because of it.

PR #3406, opened 2025-02-17
test: avoid restarting the tedge-mapper during the test

reubenmiller:
## Proposed changes

Remove the (now) unnecessary restarting of the tedge-mapper during the test as the mapper should be able to detect such changes dynamically.

A fixed sleep is used instead to give the mapper a grace period to detect the change (e.g. via hanlding of the the inotify event).

The change was motivated by observing a local test failure during a flake finder run, where it seemed that restarting the mapper would negatively affect the test by introducing unnecessary noise (though exact reason does not matter too much as the restart is now unnecessary).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_timeout/custom_operation.robot
+++ b/tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_timeout/custom_operation.robot
@@ -12,7 +12,7 @@ Test Tags           theme:c8y    theme:operation    theme:custom
 *** Test Cases ***
 Custom operation successful
     ThinEdgeIO.Transfer To Device    ${CURDIR}/c8y_Command_Success    /etc/tedge/operations/c8y/c8y_Command
-    ThinEdgeIO.Restart Service    tedge-mapper-c8y
+    Should Contain Supported Operations    c8y_Command
     ${operation}=    Cumulocity.Create Operation
     ...    description=do something
     ...    fragments={"c8y_Command":{"text":"sleep 5"}}
@@ -20,7 +20,7 @@ Custom operation successful
 
 Custom operation fails
     ThinEdgeIO.Transfer To Device    ${CURDIR}/c8y_Command_Fails    /etc/tedge/operations/c8y/c8y_Command
-    ThinEdgeIO.Restart Service    tedge-mapper-c8y
+    Should Contain Supported Operations    c8y_Command
     ${operation}=    Cumulocity.Create Operation
     ...    description=do something
     ...    fragments={"c8y_Command":{"text":"sleep 20"}}


tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_timeout/custom_operation.robot
```diff
@@ -12,15 +12,15 @@ Test Tags           theme:c8y    theme:operation    theme:custom
 *** Test Cases ***
 Custom operation successful
     ThinEdgeIO.Transfer To Device    ${CURDIR}/c8y_Command_Success    /etc/tedge/operations/c8y/c8y_Command
-    ThinEdgeIO.Restart Service    tedge-mapper-c8y
+    Sleep    2s    reason=Give the mapper a chance to detect the change
```
albinsuresh, line 15: Wouldn't that (updated supported operations list) be better?
```suggestion
    Should Have MQTT Messages
    ...    c8y/s/us
    ...    message_pattern="114,.*c8y_Command,.*"
```

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_timeout/custom_operation.robot
```diff
@@ -12,15 +12,15 @@ Test Tags           theme:c8y    theme:operation    theme:custom
...
     ${operation}=    Cumulocity.Create Operation
     ...    description=do something
     ...    fragments={"c8y_Command":{"text":"sleep 5"}}
     Operation Should Be SUCCESSFUL    ${operation}
 
 Custom operation fails
     ThinEdgeIO.Transfer To Device    ${CURDIR}/c8y_Command_Fails    /etc/tedge/operations/c8y/c8y_Command
-    ThinEdgeIO.Restart Service    tedge-mapper-c8y
+    Sleep    2s    reason=Give the mapper a chance to detect the change
```
albinsuresh, line 23: Same as above.

tests/RobotFramework/tests/cumulocity/custom_operation/custom_operation_timeout/custom_operation.robot
reubenmiller: Yeah good suggestion, for the first test this should be fine. I'll have to check if the same check will work for the second test given that we're only updating the contents (and not changing the file name), so I'm not sure if the known operations list is cached or not...but it should be easy enough to check.

reubenmiller: Forget what I just said.. @albinsuresh you're 100% right...for some reason I thought the test was reusing the same container (it is not), and testing that the contents of the custom operation handler was being tested (it isn't)...so you're proposal is spot on, and I'll implement that now.

reubenmiller: I used a slight variant of your suggestion. I opted for using the c8y interface `Should Contain Supported Operations` rather than relying knowing the cloud specific c8y topic.

```
Should Contain Supported Operations    c8y_Command
```

albinsuresh (approved):
LGTM

Issue #3405, opened 2025-02-17
Include shell completions scripts in linux packages for debian and rpm


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Using the tab completions scripts which were included as part of https://github.com/thin-edge/thin-edge.io/issues/3331, include the scripts into the linux packages for debian and rpm.

**Describe the solution you'd like**

Add the following shell completion scripts for the following shells:
* bash
* zsh
* fish

Use the recommended paths as per the guidelines for each corresponding shell.

**Describe alternatives you've considered**

**Additional context**

Issue #3331, opened 2025-01-15
Generate shell completions for `tedge` CLI


jarhodes314:
Ideally this should support both the fixed values (e.g. `config` and `get`) and configuration keys too (e.g. `device.id`). It should support zsh and bash.

jarhodes314:
The notes I made from my initial investigations into this. There are a couple of different ways the shell completion can work, and they have different trade-offs.

## Static
We generate a static completion script. This tells the shell everything it needs to know to complete all possible `tedge ...` commands, resulting in a very long shell script that will run completely autonomously from the `tedge` command. This should always do something sensible, however:
- Fish doesn't support the config key completions (although may be able to following a PR to clap)
- The completions won't change if `tedge` changes, so if they're manually installed, they'll need to be manually updated

## Dynamic
We generate a dynamic completion script. This basically just tells the shell to invoke `tedge` with some particular environment variables, and if we see those, `tedge` will output some information for the completion script to use and will exit. This has some slightly different characteristics:
- It will always be up-to-date with the current `tedge` binary, so you only need to install once
- If you have the completion script and try and use it with an outdated tedge binary (i.e. one that doesn't have the completions feature), you will get some error messages in your terminal when trying to use completions
- It works out of the box on fish
- The clap-complete feature is marked as `unstable`, although from a code perspective it's a very small API that we're using to enable it, and it appears to work perfectly at the moment, so I don't forsee much of a maintenance burden here

## Preview (on fish)
![image](https://github.com/user-attachments/assets/648af974-70af-4a4b-ae15-7cab1a37f6db)

reubenmiller:
From my experience with go-c8y-cli, I'd recommend the dynamic completion options. This is much more flexible and we can write custom completion logic if needed in Rust.

jarhodes314:
Nice, that's much simpler than trying to install it. I've also now realised that the dynamic values are the way to get partial completions (e.g. we suggest `c8y.url` but not `c8y.profiles.new.url`) to work properly. In my original testing, I used [`ValueEnum`](https://docs.rs/clap/latest/clap/trait.ValueEnum.html) to achieve this, but this also restricts the values we can insert.

We could also try and add (best-effort) handling of profile names (for the separate argument). I think for this I would approximate to "any valid profile name for any cloud is a sufficient completion" (the chance of someone actually using cloud profiles and multiple clouds simultaneously must be miniscule)

reubenmiller:
[...]

PR #3407, opened 2025-02-17
feat: install tedge tab completions during installation

reubenmiller:
## Proposed changes

Install the shell completions during the `tedge` post installation phase if the detected shell scripts are found.

Adding the script as part of the installation allows for a more seamless completion experience as sometimes no setup needs to be done if the user has already enabled shell completion for their preferred shell.

The system test image now includes the shell completion out of the box to help developers during testing and demos.

Note: The shell completion script is added as part of the post installation and not a static file as the clap generated script will use the full path to the binary that called the `tedge completions` command which makes running at build time problematic as it would use the build time path of the `tedge` binary instead of the destinations path.

## Types of changes

## Paste Link to the issue

* https://github.com/thin-edge/thin-edge.io/issues/3405
* https://github.com/thin-edge/thin-edge.io/issues/3331

## Checklist

### DIFF
--- a/docs/src/install/index.md
+++ b/docs/src/install/index.md
@@ -223,43 +223,120 @@ The packages can be viewed directly from the [Cloudsmith.io](https://cloudsmith.
 </table>
 
 ## Shell completions
+
 :::note
 Shell-completions are supported with `tedge` >= 1.5.0
 :::
 
-To make using the CLI easier, `tedge` provides tab-completion support. This can
-be enabled by adding a small configuration to your shell environment:
+To make using the Command Line Interface (CLI) easier, `tedge` provides shell completion support (also known as tab-completions). Shell completions allow users to discover commands and configuration directly on the command line without having to check the online documentation.
+For all `tedge` commands, you can activate the tab completion by pressing `<TAB>` twice in quick succession. You will be presented with the available options based on the given the command line's current input.
 
-```sh title="~/.bashrc" tab={"label":"bash"}
-source <(tedge completions bash)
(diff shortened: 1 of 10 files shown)


didier-wenzek (approved):
Approved.

I only checked things related to `zsh`.

reubenmiller:
> Approved.
> 
> I only checked things related to `zsh`.

Though the system tests include all three shells (bash, zsh and fish) and the profile changes are part of the container definition, so any system test can be used to try it out.

docs/src/install/index.md
```diff
@@ -223,43 +223,120 @@ The packages can be viewed directly from the [Cloudsmith.io](https://cloudsmith.
...
 
 ## Shell completions
+
 :::note
 Shell-completions are supported with `tedge` >= 1.5.0
 :::
 
-To make using the CLI easier, `tedge` provides tab-completion support. This can
-be enabled by adding a small configuration to your shell environment:
+To make using the Command Line Interface (CLI) easier, `tedge` provides shell completion support (also known as tab-completions). Shell completions allow users to discover commands and configuration directly on the command line without having to lookup up online documentation.
```
rina23q, line 231: One "up" should be removed I suppose.

```suggestion
To make using the Command Line Interface (CLI) easier, `tedge` provides shell completion support (also known as tab-completions). Shell completions allow users to discover commands and configuration directly on the command line without having to look up online documentation.
```

rina23q (commented):
Small correction in docs.

docs/src/install/index.md
reubenmiller: I also wasn't sure whether it should be "lookup" or "look up"...so I changes it to "check"

PR #3417, opened 2025-02-24
fix: make device profile `skip` field optional and skip operation fragment when it is set to `true` 

Ruadhri17:
## Proposed changes
This PR fixes an issue where the `skip` field in the device profile was treated as mandatory by serde. Now, if the field is omitted, it defaults to `false`. Additionally, the field has been renamed to `@skip` so that it is properly recognized by the workflow and aligns with our documentation. Finally, this PR addresses a problem where the mapper did not verify whether an operation fragment sent to the cloud was marked as skipped.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/device_profile.rs
+++ b/crates/core/tedge_api/src/device_profile.rs
@@ -41,6 +41,7 @@ impl CommandPayload for DeviceProfileCmdPayload {
 pub struct DeviceProfileOperation {
     #[serde(flatten)]
     pub operation: OperationPayload,
+    #[serde(default, rename = "@skip")]
     pub skip: bool,
 }
 
--- a/crates/extensions/c8y_mapper_ext/src/operations/handlers/device_profile.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/handlers/device_profile.rs
@@ -57,32 +57,34 @@ impl OperationContext {
                 let mut messages = Vec::new();
 
                 for device_profile_operation in command.payload.operations {
-                    let message = match device_profile_operation.operation {
-                        OperationPayload::Firmware(firmware) => {
-                            let twin_metadata_topic = self.mqtt_schema.topic_for(
-                                &target.topic_id,
-                                &Channel::EntityTwinData {
-                                    fragment_key: "firmware".to_string(),
-                                },
-                            );
+                    if !device_profile_operation.skip {
+                        let message = match device_profile_operation.operation {
+                            OperationPayload::Firmware(firmware) => {
+                                let twin_metadata_topic = self.mqtt_schema.topic_for(
+                                    &target.topic_id,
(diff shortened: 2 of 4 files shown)


crates/extensions/c8y_mapper_ext/src/operations/handlers/device_profile.rs
```diff
@@ -1811,4 +1813,96 @@ mod tests {
...
+                                            "action": "install"
+                                        }
+                                    ]
+                                }
+                            ]
+                        }
+                    },
+                    {
+                        "operation": "config_update",
+                        "@skip": true,
```
didier-wenzek, line 1867: I don't get how this test checks that the config update is skipped, but software & firmware update applied.

crates/extensions/c8y_mapper_ext/src/operations/handlers/device_profile.rs
```diff
@@ -1811,4 +1813,96 @@ mod tests {
...
+        // Simulate config_update command with "successful" state
+        mqtt.send(MqttMessage::new(
+            &Topic::new_unchecked("te/device/main///cmd/device_profile/c8y-mapper-123456"),
+            json!({
+                "status": "successful",
+                "name": "test-profile",
+                "operations": [
+                    {
+                        "operation": "firmware_update",
+                        "@skip": false,
```
didier-wenzek, line 1835: To check the default value.

```suggestion
```

didier-wenzek (commented):
minor: The docs still use the keyword `skip` instead of `@skip` in one place (*The `"skip"` field is optional and the value is `false`, by default.*)

crates/core/tedge_api/src/device_profile.rs
```diff
@@ -41,6 +41,7 @@ impl CommandPayload for DeviceProfileCmdPayload {
 pub struct DeviceProfileOperation {
     #[serde(flatten)]
     pub operation: OperationPayload,
+    #[serde(default, rename = "@skip")]
```
albinsuresh, line 44: I don't think we can do that now as this has been part of a public API, esp a mandatory field. So, if we change it now, it might affect existing users.

albinsuresh: Scratch that. It is indeed `@skip` that the agent supports: https://github.com/thin-edge/thin-edge.io/blob/e509ae6bf27526f03a45c9c3e74cfa11f40c03df/crates/core/tedge_api/src/workflow/mod.rs#L414

didier-wenzek (approved):
Approved

reubenmiller:
@Ruadhri17 I'd recommend rebasing as there has been a lot of system test improvements to fix flaky tests (e.g. the two failing tests on the last run have been fixed).

PR #3422, opened 2025-02-26
test: fix `tedge_c8y_upload.robot` not setting device context

Bravo555:
## Proposed changes

Because the first test didn't set device context like the other ones, library defaulted to returning events for other devices, so if there are other devices that run this test that were not deleted, the test will fail.

The library will be updated to warn/err if device context is not set, because in tests 99% of the time we want to get events only for the device we are testing, but for now update the test to fix failing CI.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


reubenmiller (approved):
Good find.

I'm also creating a new PR for the underlying test framework to fail when a c8y source is not provided (explicit or via the current context). see https://github.com/thin-edge/c8y-test-core/pull/25

Issue #2427, opened 2023-11-07
flaky system test: Test if all c8y services are up in service_monitoring.robot


albinsuresh:
**Describe the bug**

Test failure in CI run:
* https://github.com/thin-edge/thin-edge.io/actions/runs/6783385820/job/18437730578?pr=2424

```
2023-11-07 11:13:28.840047 [PID:58115] [3] [ID:26] Execution failed in Tests.Cumulocity.Service Monitoring.Service Monitoring with 1 failing test(s)
==============================================================================
Tests                                                                         
==============================================================================
Tests.Cumulocity                                                              
==============================================================================
Tests.Cumulocity.Service Monitoring                                           
==============================================================================
Tests.Cumulocity.Service Monitoring.Service Monitoring                        
==============================================================================
Test if all c8y services are up                                       | FAIL |
AssertionError
------------------------------------------------------------------------------
Test if all c8y services are down                                     | PASS |
------------------------------------------------------------------------------
Test if all c8y services are using configured service type            | PASS |
------------------------------------------------------------------------------
Test if all c8y services using default service type when service t... | PASS |
------------------------------------------------------------------------------
Check health status of tedge-mapper-c8y service on broker stop start  | PASS |
------------------------------------------------------------------------------
Check health status of tedge-mapper-c8y service on broker restart ... | PASS |
------------------------------------------------------------------------------
Check health status of child device service :: Test service status... | PASS |
------------------------------------------------------------------------------
Tests.Cumulocity.Service Monitoring.Service Monitoring                | FAIL |
7 tests, 6 passed, 1 failed
==============================================================================
Tests.Cumulocity.Service Monitoring                                   | FAIL |
7 tests, 6 passed, 1 failed
==============================================================================
```

**Screenshots**

### Robot Results
| :white_check_mark: Passed | :x: Failed | :next_track_button: Skipped | Total | Pass % | :stopwatch: Duration |
| --- | --- | --- | :---: | :---: | :---: |
| 355 | 1 | 3 | 356 | 99.72 | 57m29.898999999s |

### Failed Tests
| Name | Message | :stopwatch: Duration | Suite |
| --- | --- | :---: | :---: |
| Test if all c8y services are up | AssertionError | 174.204 s | `Service Monitoring` |

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [e.g. 0.1.0]

[...]

PR #3433, opened 2025-02-27
tests: ensure tedge-mapper-c8y is running before running assertions

reubenmiller:
## Proposed changes

Improve a flaky system test by ensuring the `tedge-mapper-c8y` service is running.

Whilst it might include some duplicate in the `tedge-mapper-c8y` test case, where the tedge-mapper-c8y service is started twice, this does not cause any problems as the second service start will be a no-operation as systemd does not restart an already started service.

**Flakey System Test Root cause**

The Robotframework test cases are not isolated so they are run in the same container. When the **tedge-mapper-c8y** test case ends, it will stop the service which can cause some of the following services to fail if the services (which are started by default in the `Test Setup`) were published to Cumulocity before the `tedge-mapper-c8y` service was stopped.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2427

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
+++ b/tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
@@ -144,6 +144,9 @@ Custom Setup
 
 Check if a service is up
     [Arguments]    ${service_name}
+    # Ensure tedge-mapper-c8y is running because it publishes messages to the cloud
+    ThinEdgeIO.Start Service    tedge-mapper-c8y
+    ThinEdgeIO.Service Should Be Running    tedge-mapper-c8y
     ThinEdgeIO.Service Should Be Running    ${service_name}
 
     Device Should Exist    ${DEVICE_SN}:device:main:service:${service_name}    show_info=False
@@ -157,6 +160,9 @@ Check if a service is up
 
 Check if a service is down
     [Arguments]    ${service_name}
+    # Ensure tedge-mapper-c8y is running because it publishes messages to the cloud
+    ThinEdgeIO.Start Service    tedge-mapper-c8y
+    ThinEdgeIO.Service Should Be Running    tedge-mapper-c8y
     ThinEdgeIO.Start Service    ${service_name}
     Device Should Exist    ${DEVICE_SN}:device:main:service:${service_name}    show_info=False
     ThinEdgeIO.Stop Service    ${service_name}


Bravo555 (approved):

tests/RobotFramework/tests/cumulocity/service_monitoring/service_monitoring.robot
```diff
@@ -144,6 +144,8 @@ Custom Setup
 
 Check if a service is up
     [Arguments]    ${service_name}
+    # Ensure tedge-mapper-c8y is running because it publishes messages to the cloud
+    ThinEdgeIO.Start Service    tedge-mapper-c8y
```
albinsuresh, line 148: Just to be as safe as possible.
```suggestion
    ThinEdgeIO.Start Service    tedge-mapper-c8y
    ThinEdgeIO.Service Should Be Running    tedge-mapper-c8y
```

reubenmiller: Though technically that does not offer any improvements as checking that status won't make the test pass more or less (as the following assertions are "smart" assertions and don't rely on the condition being immediately true).

reubenmiller: But I've added the suggestion as it like at least speed up the diagnosis phase if it does fail rather than digging through the logs to confirm that the mapper is actually running.

albinsuresh (approved):
This fix is fine.

Actually the real weird thing about that test is that we're doing a `ThinEdgeIO.Stop Service    ${service_name}` at the end of the function `Check if a service is up` that was just supposed to validate the given condition and not really change the test env. I don't really understand why it was originally written like that.

reubenmiller:
> This fix is fine.
> 
> Actually the real weird thing about that test is that we're doing a `ThinEdgeIO.Stop Service ${service_name}` at the end of the function `Check if a service is up` that was just supposed to validate the given condition and not really change the test env. I don't really understand why it was originally written like that.

Yeah the stopping of the service is also a bit strange...I guess the idea was to pick up cases where 1 services status could be potentially written to another one?

albinsuresh:
> I guess the idea was to pick up cases where 1 services status could be potentially written to another one?

That would have been a really convoluted way to test that. Anyway, if these tests become flaky again, I'd be tempted to remove that first.

Issue #3368, opened 2025-01-29
flaky system test: Tests.Cumulocity.Registration.Processes Many Devices


albinsuresh:
**Describe the bug**

Flaky system test: `Tests.Cumulocity.Registration.Processes Many Devices`

```
Processes many devices                                                | FAIL |
Matching messages on topic 'c8y/s/us' is less than minimum.
wanted: 1
got: 0
```

The assertion failed after multiple successful iterations validating earlier child devices (45th child registration was missing in this case)

**Failed instances**

- https://github.com/thin-edge/thin-edge.io/actions/runs/13025075002/attempts/1
- https://github.com/thin-edge/thin-edge.io/actions/runs/13028578945
- https://github.com/thin-edge/thin-edge.io/actions/runs/13323710315/job/37213216092?pr=3400

reubenmiller:
Checking the logs it seems that the tedge-mapper c8y either stops receiving the local tedge registration messages, or stops processing the messages (even though all of the tedge registration messages are received by the mqtt-logger).

So it looks like this is something worth investing on the Rust code level.

rina23q:
I did some experiments. To debug, I added more debug messages on [this function ](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge_actors/src/channels.rs#L76-L80) like below and ran the test. 

```rust
#[async_trait]
impl<M: Message, N: Message + Into<M>> Sender<N> for mpsc::Sender<M> {
    async fn send(&mut self, message: N) -> Result<(), ChannelError> {
        dbg!(&message);
        let result = SinkExt::send(&mut self, message.into()).await; // doesn't complete sometimes
        dbg!(&result);
        result?;
        Ok(())
    }
}
```

`SinkExt::send()` doesn't return sometimes and stuck there forever when the test failed.

I replaced the `SinkExt::send()` by  [`SinkExt::feed()`](https://docs.rs/futures-util/latest/futures_util/sink/trait.SinkExt.html#method.feed) and [`SinkExt::flush()`](https://docs.rs/futures-util/latest/futures_util/sink/trait.SinkExt.html#method.flush) as below. 

```rust
#[async_trait]
impl<M: Message, N: Message + Into<M>> Sender<N> for mpsc::Sender<M> {
    async fn send(&mut self, message: N) -> Result<(), ChannelError> {
        dbg!(&message);
        let result = SinkExt::feed(&mut self, message.into()).await;
        dbg!(&result);
        result?;
[...]

PR #3440, opened 2025-02-27
refactor: Remove from C8yMapperActor the publish_messages receiver

didier-wenzek:
## Proposed changes

Remove from C8yMapperActor the publish_messages receiver

The motivation, for this receiver of wrapped MqttMessages, was that: "other actors may want to send MqttMessages but we want to ensure that the entity referred to in the message topic exists; that's why we route them to the mapper rather than to MqttActor directly."

However, the C8yMapperActor was sending these messages unchanged. And futhermore, this was requiring a specific mechanism to avoid deadlocks. Better to remove all that, and let the other actors (actually only the AvailabilityActor) send their messages directly over MQTT.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3368

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/availability/builder.rs
+++ b/crates/extensions/c8y_mapper_ext/src/availability/builder.rs
@@ -1,4 +1,3 @@
-use crate::actor::PublishMessage;
 use crate::availability::actor::AvailabilityActor;
 use crate::availability::AvailabilityConfig;
 use crate::availability::AvailabilityInput;
@@ -32,7 +31,7 @@ pub struct AvailabilityBuilder {
 impl AvailabilityBuilder {
     pub fn new(
         config: AvailabilityConfig,
-        mqtt: &mut (impl MessageSource<MqttMessage, Vec<ChannelFilter>> + MessageSink<PublishMessage>),
+        mqtt: &mut (impl MessageSource<MqttMessage, Vec<ChannelFilter>> + MessageSink<MqttMessage>),
         timer: &mut impl Service<TimerStart, TimerComplete>,
     ) -> Self {
         let mut box_builder: SimpleMessageBoxBuilder<AvailabilityInput, AvailabilityOutput> =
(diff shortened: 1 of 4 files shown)


rina23q (approved):
As @didier-wenzek stated above, this PR is the logical solution of the deadlock https://github.com/thin-edge/thin-edge.io/issues/3368.

Exactly the line 244 `publish_message_to_subscribed_handles()` was detected 
 that it causes the deadlock from the ticket investigation.
https://github.com/thin-edge/thin-edge.io/blob/6222270c25836edf5043c5307e9e389b108ad8c5/crates/extensions/c8y_mapper_ext/src/actor.rs#L241-L244

Since this PR changed the routing of the messages across actors, it makes sense why tests didn't fail even after 70 iterations. Compared to the failure rate 10%, no failure in 70 interactions can be said that the issue is fixed.

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
```diff
@@ -32,7 +31,7 @@ pub struct AvailabilityBuilder {
 impl AvailabilityBuilder {
     pub fn new(
         config: AvailabilityConfig,
-        mqtt: &mut (impl MessageSource<MqttMessage, Vec<ChannelFilter>> + MessageSink<PublishMessage>),
+        mqtt: &mut (impl MessageSource<MqttMessage, Vec<ChannelFilter>> + MessageSink<MqttMessage>),
```
albinsuresh, line 34: That was subtle(accessing the MQTT actor indirectly via the mapper). Wondering if accepting `MessageSink` as a separate parameter and connecting directly to the MQTT actor would have made the connection flow easier follow.

albinsuresh (approved):
LGTM.

I'm also wondering if the sender code update that we used to debug the issue(using `try_send` and logging when it fails) could also be merged separately, to easily catch such issues in the future.

crates/extensions/c8y_mapper_ext/src/availability/builder.rs
didier-wenzek: Yeah. My motivation was mostly to avoid changing the builders.

didier-wenzek:
> I'm also wondering if the sender code update that we used to debug the issue(using `try_send` and logging when it fails) could also be merged separately, to easily catch such issues in the future. I guess we can update the `LoggingSender`(which is used everywhere) to achieve the same.

Indeed, this is something to consider. However, this would have to be improved as the current logging was quite poor missing the actor name.

Issue #3412, opened 2025-02-21
New CLI command to collect diagnostic logs in one step


rina23q:
**Is your feature request related to a problem? Please describe.**
Introduce a new CLI command to collect all the diagnostic logs in one step so that we can support users to report bugs or questions easily.

**Describe the solution you'd like**

The first idea is to create a new CLI subcommand `tedge logs`.
```sh
tedge logs [--output-dir <DIR>]
```

It collects
- ~~output of mqtt-logger (need to mask JWT token values)~~ refer to [comment](https://github.com/thin-edge/thin-edge.io/issues/3412#issuecomment-2677751024)
- operation logs under `/var/log/tedge/agent/*`
- ~~message cache `/etc/tedge/.tedge-mapper-c8y/entity_store.json`~~ Contents of the entity store, GET by HTTP API call. (`tedge http get /tedge/entity-store/v1/entities`) 
- tedge-mapper and tedge-agent logs (journalctl. how to do if the device is not using systemd)
- mosquitto log (if using mosquitto)
- and more?

and zips them, then outputs to the target directory.

**Describe alternatives you've considered**

Alternatively we can create a community plugin to collect logs.

**Additional context**

A good reference of the type of logs is what we collect in our system tests.
https://github.com/thin-edge/thin-edge.io/blob/main/tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py#L162

reubenmiller:
> message cache /etc/tedge/.tedge-mapper-c8y/entity_store.json

Though we should definitely make it clear that this file is an internal file only (useful for us, but we don't want users interacting with it directly). In addition getting the entity store via the new HTTP API would be in addition useful (if the API endpoint is available)

reubenmiller:
> output of mqtt-logger (need to mask JWT token values)

You can exclude this as the `mqtt-logger` is not a community plugin and only exists in the context of our system tests. This feature is more for customers who are using thin-edge.io in development or production.

reubenmiller:
> > message cache /etc/tedge/.tedge-mapper-c8y/entity_store.json
> 
> Though we should definitely make it clear that this file is an internal file only (useful for us, but we don't want users interacting with it directly). In addition getting the entity store via the new HTTP API would be in addition useful (if the API endpoint is available)

[...]

PR #3445, opened 2025-03-04
feat: add optional break conditions to `tedge mqtt sub`

rina23q:
## Proposed changes

This PR introduces two new options to `tedge mqtt sub` to give a break to the command.

```shell
  -W, --duration <DURATION>
          Set a timeout duration (e.g., 60s, 1h)

  -C, --count <COUNT>
          Set the number of packets before stopping
```

* `--duration`: `tedge mqtt sub` breaks after given time. e.g.,
```
tedge mqtt sub "#" --duration 10s
```

* `--count`: `tedge mqtt sub` breaks after the given number of packets received. e.g.,
```
$ tedge mqtt sub "#" --count 3
INFO: Connected
[te/device/main/service/c8y-firmware-plugin] {"@parent":"device/main//","@type":"service","type":"service"}
[te/device/main/service/c8y-firmware-plugin/status/health] {"pid":1400,"status":"down"}
[te/device/main/service/mosquitto-c8y-bridge] {"@id":"rina0010:device:main:service:mosquitto-c8y-bridge","@parent":"device/main//","@type":"service","name":"mosquitto-c8y-bridge","type":"service"}
INFO: Break
```

## Types of changes

## Paste Link to the issue
#3412

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/mqtt/cli.rs
+++ b/crates/core/tedge/src/cli/mqtt/cli.rs
@@ -9,6 +9,7 @@ use clap_complete::ArgValueCandidates;
 use clap_complete::CompletionCandidate;
 use mqtt_channel::Topic;
 use rumqttc::QoS;
+use tedge_config::SecondsOrHumanTime;
 
 const PUB_CLIENT_PREFIX: &str = "tedge-pub";
 const SUB_CLIENT_PREFIX: &str = "tedge-sub";
@@ -45,6 +46,12 @@ pub enum TEdgeMqttCli {
         /// Avoid printing the message topics on the console
         #[clap(long = "no-topic")]
         hide_topic: bool,
+        /// Set a timeout duration (e.g., 60s, 1h)
+        #[clap(long, short = 'W')]
+        duration: Option<SecondsOrHumanTime>,
+        /// Set the number of packets before stopping
+        #[clap(long, short = 'C')]
+        count: Option<u32>,
     },
 }
 
@@ -77,6 +84,8 @@ impl BuildCommand for TEdgeMqttCli {
                     topic,
                     qos,
                     hide_topic,
+                    duration,
+                    count,
                 } => MqttSubscribeCommand {
                     host: config.mqtt.client.host.clone(),
                     port: config.mqtt.client.port.into(),
@@ -87,6 +96,8 @@ impl BuildCommand for TEdgeMqttCli {
                     ca_file: auth_config.ca_file,
                     ca_dir: auth_config.ca_dir,
                     client_auth_config: auth_config.client,
+                    duration: duration.map(|v| v.duration()),
+                    count,
                 }
                 .into_boxed(),
             }
--- a/crates/core/tedge/src/cli/mqtt/subscribe.rs
+++ b/crates/core/tedge/src/cli/mqtt/subscribe.rs
@@ -13,6 +13,7 @@ use rumqttc::Outgoing;
 use rumqttc::Packet;
 use rumqttc::QoS;
 use std::sync::atomic::Ordering;
+use std::time::Duration;
 use tedge_config::MqttAuthClientConfig;
 
 const DEFAULT_QUEUE_CAPACITY: usize = 10;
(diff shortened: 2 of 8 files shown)


reubenmiller:
@rina23q Just naming things. It might be more useful to align with naming used by [mosquitto_sub](https://mosquitto.org/man/mosquitto_sub-1.html) (to make it easier to switch between the tools)

* `--window` => `--duration` and with a short option match mosquittos_sub `-W`. I'd originally thought about `--timeout`...but that is to unclear e.g. timeout of what, connecting to the broker?
* `--packets` => `--count` and with a short option match mosquittos_sub `-C` (uppercase)

Plus, it the previous names (window and packets) are super obvious what they mean.

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -87,8 +90,15 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
         options.set_transport(rumqttc::Transport::tls_with_config(tls_config.into()));
     }
 
+    // the default keepalive in rumqttc is 60 sec
+    if !cmd.window.is_zero() && cmd.window.as_secs() < 60 {
+        options.set_keep_alive(cmd.window);
```
albinsuresh, line 95: Smart

albinsuresh (commented):
Agree to Reuben's naming suggestions. Also would be nice to add some basic system tests.

crates/core/tedge/src/cli/mqtt/cli.rs
```diff
@@ -45,6 +46,12 @@ pub enum TEdgeMqttCli {
         /// Avoid printing the message topics on the console
         #[clap(long = "no-topic")]
         hide_topic: bool,
+        /// Set a timeout duration (e.g., 60s, 1h). Use 0 for no timeout
+        #[clap(long, default_value = "0")]
+        window: SecondsOrHumanTime,
+        /// Set the number of packets before stopping. Use 0 for unlimited
+        #[clap(long, default_value = "0")]
+        packets: u32,
```
didier-wenzek, line 54: A cleaner approach is to use optional inputs instead of magic values.

```suggestion
        /// Set a timeout duration (e.g., 60s, 1h).
        #[clap(long)]
        window: Option<SecondsOrHumanTime>,
        /// Set the number of packets before stopping.
        #[clap(long)]
        packets: Option<u32>,
```

crates/core/tedge/src/cli/mqtt/subscribe.rs
didier-wenzek: I'm okay with that. But be aware that the precision is 60 sec if the timeout is greater than 60s.

An option, is to set the keep alive to 5 seconds to have a more precise timeout.

rina23q: If user gives 1m10s, then indeed with 60s keepalive, the command will break after 2m.
On the other hand, if user provides 1h, every 5s ping sounds a bit too much.

Does it make simplify if we set the same keepalive value as provided unless it's smaller than the maximum value?
> the maximum Keep Alive interval is 18 hours, 12 minutes, and 15 seconds. (https://www.hivemq.com/blog/mqtt-essentials-part-10-alive-client-take-over/)

didier-wenzek: > if user provides 1h, every 5s ping sounds a bit too much.

I would not bother too much with that. This command being used locally and for test purposes.

crates/core/tedge/src/cli/mqtt/cli.rs
rina23q: Done in 20884905d1e22b2e6db220a89fc9b773abd8c334

crates/core/tedge/src/cli/mqtt/subscribe.rs
rina23q: Changed in 20884905d1e22b2e6db220a89fc9b773abd8c334

rina23q:
Added system tests. The good highlight is, now we can use `tedge mqtt sub --duration 1` instead of `timeout 1` in system test :)

Screenshot 2025-03-05 at 19 05 39

albinsuresh (approved):

Issue #3439, opened 2025-02-27
registration message order


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Registering child devices and services is often done asynchronously with all of the other messages. This can cause problems when sending the registration messages to Cumulocity as some of the telemetry data messages can be received before the registration messages which causes the cloud entity to be registered with default info (e.g. name and type) or some of the messages being silently ignored by the cloud as the device is not registered prior to receiving the message.

**Describe the solution you'd like**

A review needs to be done on how exactly to solve this problem reliably, but below is the order of messages that needs to be sent to Cumulocity, and there should be a brief delay (~N seconds) before any follow up messages are sent.

1. Main device (`101`)
1. device registration (e.g. `101`)
2. Service registration (`102`)
3. Telemetry data or service status

**Describe alternatives you've considered**

**Additional context**

It should be noted the registration problem is mostly related to child entities (child devices and services) as the main device is registered via `tedge connect c8y`. This means that the main device already exists in the platform prior to the mapper being started for the first time.

**System Test failures**

The following flaky system test tickets are open where the suspected root cause is due to the ordering of registration messages:

* https://github.com/thin-edge/thin-edge.io/issues/3393
* https://github.com/thin-edge/thin-edge.io/issues/3429

PR #3446, opened 2025-03-05
refactor: Delay after publishing entity creation message

albinsuresh:
## Proposed changes

Introduce a delay after the entity registration is converted and published to C8Y, so that the telemetry data published afterwards doesn't race with it, leading to the auto-creation of a managed object with same ID.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3439

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/actor.rs
+++ b/crates/extensions/c8y_mapper_ext/src/actor.rs
@@ -181,6 +181,9 @@ impl C8yMapperActor {
             self.process_registration_message(pending_entity.reg_message)
                 .await?;
 
+            // Wait for a while to allow the registration message to be processed by the cloud before sending the data messages
+            tokio::time::sleep(Duration::from_secs(1)).await;
+
             // Convert and publish cached data messages
             for pending_data_message in pending_entity.data_messages {
                 self.process_message(pending_data_message).await?;


crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -181,6 +181,9 @@ impl C8yMapperActor {
             self.process_registration_message(pending_entity.reg_message)
                 .await?;
 
+            // Wait for a while to allow the registration message to be processed by the cloud before sending the data messages
+            tokio::time::sleep(Duration::from_secs(1)).await;
```
jarhodes314, line 185: This feels like it's going to break if the bridge loses connection. Is there a particular reason why we can't do what we do for the main device, i.e. publish the registration message multiple times and wait until we receive a suitable `41` response to confirm the device definitely exists?

albinsuresh: Yeah, this is a known limitation. We're doing this low cost fix to just reduce the probability of this flakiness in 80-90% of the cases when everything is working fine. The plan was to try this  low cost fix and see if it reduces the flakiness considerably.

The other two approaches considered were:

1. Waiting for the duplicate device error over MQTT as done for the main device
2. Performing the entity creation over HTTP

Since both approaches fundamentally change the control flow within the mapper, with huge impact on all existing tests(esp unit tests) as well, we just decided to do this minimal fix for now. But, we can revisit these approaches later once we achieve some stability.

Issue #3447, opened 2025-03-06
tedge mqtt sub panics when using a duration less than 1 second


reubenmiller:
**Describe the bug**

The new `tedge mqtt sub --duration <duration>` feature introduced in https://github.com/thin-edge/thin-edge.io/pull/3445 does not cleanly handle sub second durations due to mqttc rejecting keep alives less than a given value.

```sh
$ tedge mqtt sub --count 100 --duration 100ms '#' --log-level warn

thread 'main' panicked at /home/runner/.cargo/git/checkouts/rumqtt-e490340c454254e0/8c489fa/rumqttc/src/lib.rs:578:9:
Keep alives should be specified in seconds. Durations less than a second are not allowed, except for Duration::ZERO.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**To Reproduce**

The bug is easy to reproduce by running:

```
tedge mqtt sub --count 100 --duration 100ms '#' --log-level warn
```

**Expected behavior**

A minimum duration period should be applied, and a clean error message should be returned to the user explaining why the value was rejected.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`unknown`|
|System-Architecture|`Linux tedge 6.8.0-39-generic #39-Ubuntu SMP PREEMPT_DYNAMIC Sat Jul  6 02:50:39 UTC 2024 aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.4.3~224+g1e7027f`|

**Additional context**

rina23q:
Fixed by #3448

PR #3448, opened 2025-03-06
fix: don't panic when --duration has less than 1s value

rina23q:
## Proposed changes
The PR #3445 uses the MQTT keepalive mechanism to achieve a timeout. When an outgoing event `PINGREQ` occurs, the code checks the elapsed time, and if the elapsed time is larger than the specified duration, `tedge mqtt sub` exits.

MQTT keepalive must be at least 1s, that's why a panic occurred for the `100ms` input as the ticket #3447 describes.

After trying several approaches, I decided to take this approach https://github.com/thin-edge/thin-edge.io/pull/3448#discussion_r1984854857. It requires the least code changes compared to other approaches and `--duration` can even accept `100ms` input.

Also, this PR fixes regarding a `DISCONNECT` packet. Now it is sent before exiting the command when timeout occurred or message count limit is reached.

## Types of changes

## Paste Link to the issue
#3447

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/mqtt/cli.rs
+++ b/crates/core/tedge/src/cli/mqtt/cli.rs
@@ -46,10 +46,10 @@ pub enum TEdgeMqttCli {
         /// Avoid printing the message topics on the console
         #[clap(long = "no-topic")]
         hide_topic: bool,
-        /// Set a timeout duration (e.g., 60s, 1h)
+        /// Disconnect and exit after the specified timeout (e.g., 60s, 1h)
         #[clap(long, short = 'W')]
         duration: Option<SecondsOrHumanTime>,
-        /// Set the number of packets before stopping
+        /// Disconnect and exit after receiving the specified number of messages
         #[clap(long, short = 'C')]
         count: Option<u32>,
     },
--- a/crates/core/tedge/src/cli/mqtt/subscribe.rs
+++ b/crates/core/tedge/src/cli/mqtt/subscribe.rs
@@ -90,19 +90,8 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
         options.set_transport(rumqttc::Transport::tls_with_config(tls_config.into()));
     }
 
-    match cmd.duration {
-        Some(Duration::ZERO) | None => {}
-        Some(duration) if duration.as_secs() < 5 => {
-            options.set_keep_alive(duration);
-        }
-        Some(_) => {
-            options.set_keep_alive(Duration::from_secs(5));
-        }
-    }
-
     let (client, mut connection) = Client::new(options, DEFAULT_QUEUE_CAPACITY);
-    let interrupted = super::disconnect_if_interrupted(client.clone());
-    let started = std::time::Instant::now();
+    let interrupted = super::disconnect_if_interrupted(client.clone(), cmd.duration);
     let mut n_packets = 0;
 
     for event in connection.iter() {
@@ -122,18 +111,12 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
                         }
                         n_packets += 1;
                         if matches!(cmd.count, Some(count) if count > 0 && n_packets >= count) {
-                            eprintln!("INFO: Break");
+                            eprintln!("INFO: Received {n_packets} messages");
+                            let _ = client.disconnect();
                             break;
                         }
                     }
                     Err(err) => error!("{err}"),
-                }
-            }
-            Ok(Event::Outgoing(Outgoing::PingReq)) => {
-                if matches!(cmd.duration, Some(duration) if !duration.is_zero() && started.elapsed() > duration)
-                {
-                    eprintln!("INFO: Timeout");
-                    break;
                 }
             }
             Ok(Event::Incoming(Incoming::Disconnect)) => {
--- a/crates/core/tedge/src/cli/mqtt/subscribe.rs
+++ b/crates/core/tedge/src/cli/mqtt/subscribe.rs
@@ -92,7 +92,7 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
 
     let (client, mut connection) = Client::new(options, DEFAULT_QUEUE_CAPACITY);
     let interrupted = super::disconnect_if_interrupted(client.clone(), cmd.duration);
-    let mut n_packets = 0;
+    let mut n_messages = 0;
 
     for event in connection.iter() {
         match event {
@@ -109,10 +109,9 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
                         } else {
                             println!("[{}] {}", &message.topic, payload);
                         }
-                        n_packets += 1;
-                        if matches!(cmd.count, Some(count) if count > 0 && n_packets >= count) {
-                            eprintln!("INFO: Received {n_packets} messages");
-                            let _ = client.disconnect();
+                        n_messages += 1;
+                        if matches!(cmd.count, Some(count) if count > 0 && n_messages >= count) {
+                            eprintln!("INFO: Received {n_messages} message/s");
                             break;
                         }
                     }
@@ -136,6 +135,18 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
                 }
                 error!("{err}");
                 std::thread::sleep(std::time::Duration::from_secs(1));
+            }
+            _ => {}
+        }
+    }
+
+    let _ = client.disconnect();
+    for event in connection.iter() {
+        match event {
+            Err(_)
+            | Ok(Event::Incoming(Incoming::Disconnect))
+            | Ok(Event::Outgoing(Outgoing::Disconnect)) => {
+                break;
             }
             _ => {}
         }
--- a/crates/core/tedge/src/cli/mqtt/mod.rs
+++ b/crates/core/tedge/src/cli/mqtt/mod.rs
@@ -4,7 +4,7 @@ use rumqttc::Client;
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
-use std::sync::Mutex;
+use std::time::Duration;
 
 mod cli;
 mod error;
@@ -13,19 +13,28 @@ mod subscribe;
 
 const MAX_PACKET_SIZE: usize = 268435455; // 256 MB
 
-fn disconnect_if_interrupted(client: Client) -> Arc<AtomicBool> {
-    let interrupter = Arc::new(Mutex::new(client));
+fn disconnect_if_interrupted(client: Client, timeout: Option<Duration>) -> Arc<AtomicBool> {
     let interrupted = Arc::new(AtomicBool::new(false));
     for signal in signal_hook::consts::TERM_SIGNALS {
-        let interrupter = interrupter.clone();
+        let client = client.clone();
         let interrupted = interrupted.clone();
         unsafe {
             let _ = signal_hook::low_level::register(*signal, move || {
                 interrupted.store(true, Ordering::Relaxed);
-                let client = interrupter.lock().unwrap();
                 let _ = client.disconnect();
             });
         }
     }
+
+    if let Some(timeout) = timeout {
+        let timeout_elapsed = interrupted.clone();
+        std::thread::spawn(move || {
+            std::thread::sleep(timeout);
+            eprintln!("INFO: Timeout");
+            timeout_elapsed.store(true, Ordering::Relaxed);
+            let _ = client.disconnect();
+        });
+    }
+
     interrupted
 }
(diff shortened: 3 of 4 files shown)


crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -92,6 +93,9 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
 
     match cmd.duration {
         Some(Duration::ZERO) | None => {}
+        Some(duration) if duration.lt(&Duration::from_secs(1)) => {
+            return Err(anyhow!("--duration <DURATION> must be at least 1 second"));
```
didier-wenzek, line 97: Rather returning an error, I would simply round the timeout value.

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -90,75 +93,100 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
...
+                ) {
+                    break;
                 }
             }
-            Ok(Event::Outgoing(Outgoing::PingReq)) => {
-                if matches!(cmd.duration, Some(duration) if !duration.is_zero() && started.elapsed() > duration)
-                {
+        }
+        Some(duration) => loop {
+            match connection.recv_timeout(duration) {
```
didier-wenzek, line 115: This is does not behave as you expect.

Instead of waiting at most the given `duration` between *two* events, you want the *loop* to run at most the given `duration`.

```suggestion
        Some(duration) => while started.elapsed() < duration {
            match connection.recv_timeout(duration - started.elapsed() ) {
```

crates/core/tedge/src/cli/mqtt/subscribe.rs
reubenmiller: Yeah we talked about this, I'm ok with both (error or rounding)...rounding will be better once the piggy-backing the MQTT PINGREQ logic is avoided (as it unexpectedily compromises the user expectation of that the given duration will be accurate to ~1 second etc.)

Though if we do rounding and it affects the user agreement (e.g. changes the duration from 500ms to 1s), then we should be loud about it by printing sometime to the console

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -90,75 +93,100 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
...
-                        n_packets += 1;
-                        if matches!(cmd.count, Some(count) if count > 0 && n_packets >= count) {
-                            eprintln!("INFO: Break");
-                            break;
-                        }
-                    }
-                    Err(err) => error!("{err}"),
+    match cmd.duration {
+        None => {
+            for event in connection.iter() {
```
didier-wenzek, line 102: I'm not convinced that having two different loops helps understanding.

Why not use either `connection.recv()` or `connection.recv_timeout()`, both returning the same type of values.

crates/core/tedge/src/cli/mqtt/subscribe.rs
didier-wenzek: > the piggy-backing the MQTT PINGREQ logic [...] unexpectedly compromises the user expectation of that the given duration will be accurate to ~1 second.

This could be fixed by sending a PING each and every second. I agree that this is a hack, but I think this is acceptable in the context of `tedge mqtt sub` which is more of an helper than a production tool.

Anyway, looking for a better solution.

didier-wenzek: > Anyway, looking for a better solution.

The approach proposed by @rina23q (i.e. using `rumqttc::connection.recv_timeout()`) works but is a bit invasive.

A simple alternative is to use a background thread that exits when the timeout elapses.

```
    fn execute(&self) -> Result<(), MaybeFancy<anyhow::Error>> {
        std::thread::spawn(move || {
            std::thread::sleep(self.timeout);
            eprintln!("Timeout");
            std::process::exit(0);
        });
        Ok(subscribe(self)?)
    }
```

Yet another alternative, less brutal, is to spawn this thread from [`disconnect_if_interrupted`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/core/tedge/src/cli/mqtt/mod.rs#L16) which is used by `tedge mqtt` to be interrupted on `^C`. 

```
fn disconnect_if_interrupted(client: Client, timeout: Duration) -> Arc<AtomicBool> {
    let interrupter = Arc::new(Mutex::new(client));
    let interrupted = Arc::new(AtomicBool::new(false));
    for signal in signal_hook::consts::TERM_SIGNALS {
         // ..
    }

    let timeout_elapsed = interrupted.clone();
    std::thread::spawn(move || {
        std::thread::sleep(timeout);
        eprintln!("Timeout");
        timeout_elapsed.store(true, Ordering::Relaxed);
        let client = interrupter.lock().unwrap();
        let _ = client.disconnect();
    });

    interrupted
}
```

jarhodes314: Is there a particular reason why `disconnect_if_interrupted` is using a `Mutex` around `client` when `rumqttc::Client: Clone` (since it's just wrapping an `mpsc::Sender<rumqttc::Request>` if I recall correctly)?

didier-wenzek: > Is there a particular reason why `disconnect_if_interrupted` is using a `Mutex` around `client` when `rumqttc::Client: Clone` (since it's just wrapping an `mpsc::Sender<rumqttc::Request>` if I recall correctly)?

Indeed, this `Mutex` is useless and can be safely removed.

```
fn disconnect_if_interrupted(client: Client, timeout: Duration) -> Arc<AtomicBool> {
    let interrupted = Arc::new(AtomicBool::new(false));
    for signal in signal_hook::consts::TERM_SIGNALS {
         // ..
    }

    let timeout_elapsed = interrupted.clone();
    std::thread::spawn(move || {
        std::thread::sleep(timeout);
        eprintln!("Timeout");
        timeout_elapsed.store(true, Ordering::Relaxed);
        let _ = client.disconnect();
    });

    interrupted
}
```

crates/core/tedge/src/cli/mqtt/cli.rs
```diff
@@ -49,7 +49,7 @@ pub enum TEdgeMqttCli {
         /// Set a timeout duration (e.g., 60s, 1h)
         #[clap(long, short = 'W')]
         duration: Option<SecondsOrHumanTime>,
-        /// Set the number of packets before stopping
+        /// Disconnect and exit after receiving the specified number of messages
```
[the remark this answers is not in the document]
rina23q: Rephrased in d0f2861fa1786c45a688e4c78ee666c004e4d0d4

crates/core/tedge/src/cli/mqtt/subscribe.rs
rina23q: Thanks @didier-wenzek. I took your approach on `disconnect_if_interrupted()`. The commit is here. 4441db63cde5a7d8d120f87876a514bed66c90c8

didier-wenzek (approved):
Approved. It works now with user-surprises nor internal hacks. Nice!

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -122,20 +111,14 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
                         }
                         n_packets += 1;
                         if matches!(cmd.count, Some(count) if count > 0 && n_packets >= count) {
-                            eprintln!("INFO: Break");
+                            eprintln!("INFO: Received {n_packets} messages");
+                            let _ = client.disconnect();
```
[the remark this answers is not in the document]
didier-wenzek: In order to be sure the disconnect is effective, the simplest is to add another active loop waiting for that.

- `break` from the main loop without a call to `client.disconnect()` 
- call  `client.disconnect()` the after the main loop
- the starts a new loop waiting the disconnect

```
    let _ = client.disconnect();
    for event in connection.iter() {
        match event {
            Err(_)
            | Ok(Event::Incoming(Incoming::Disconnect))
            | Ok(Event::Outgoing(Outgoing::Disconnect)) => {
                break;
            }
            _ => {}
        }
    }

```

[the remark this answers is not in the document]
didier-wenzek: I pushed a commit to properly disconnect: https://github.com/thin-edge/thin-edge.io/pull/3448/commits/cfaf55dbdc7efa6718bdc823829684aa71e0968f

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -122,25 +111,19 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
                         }
                         n_packets += 1;
                         if matches!(cmd.count, Some(count) if count > 0 && n_packets >= count) {
-                            eprintln!("INFO: Break");
+                            eprintln!("INFO: Received {n_packets} messages");
```
reubenmiller, line 114: Allow for both singular and plural form (without doing the extra logic)

```suggestion
                            eprintln!("INFO: Received {n_packets} message/s");
```

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -90,19 +90,8 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
...
-        Some(_) => {
-            options.set_keep_alive(Duration::from_secs(5));
-        }
-    }
-
     let (client, mut connection) = Client::new(options, DEFAULT_QUEUE_CAPACITY);
-    let interrupted = super::disconnect_if_interrupted(client.clone());
-    let started = std::time::Instant::now();
+    let interrupted = super::disconnect_if_interrupted(client.clone(), cmd.duration);
     let mut n_packets = 0;
```
reubenmiller, line 95: `messages` would be more descriptive of what the variable is actually counting ;)

```suggestion
    let mut n_messages = 0;
```

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -122,25 +111,19 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
...
-                    break;
-                }
-            }
             Ok(Event::Incoming(Incoming::Disconnect)) => {
                 eprintln!("INFO: Disconnected");
                 break;
             }
             Ok(Event::Outgoing(Outgoing::Disconnect)) => {
+                eprintln!("INFO: Interrupted");
                 break;
```
reubenmiller, line 127: Is the `INFO: Interrupted` message really needed here? When using `--duration` it seems to print out both "Timeout" and "Interrupted" (which is a bit noisy from a UX point of view).

```sh
$ tedge mqtt sub '#' --duration 100ms >/dev/null
INFO: Connected
INFO: Timeout
INFO: Interrupted
```

In contrast to use the `--count`, it only prints out the expected message after fulfilling the exit condition.

```sh
$ tedge mqtt sub '#' --duration 100ms --count 1 >/dev/null
INFO: Connected
INFO: Received 1 messages
```

So the differences in log messages is a bit unexpected.

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -158,6 +140,18 @@ fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
         }
     }
 
+    let _ = client.disconnect();
+    for event in connection.iter() {
+        match event {
+            Err(_)
+            | Ok(Event::Incoming(Incoming::Disconnect))
+            | Ok(Event::Outgoing(Outgoing::Disconnect)) => {
```
jarhodes314, line 148: Is this `Outgoing::Disconnect` correct? If we break before the other server acknowledges we're disconnected, that still feels incorrect. Maybe we need to `return` instead of `break` where we receive `Incoming::Disconnect` above (in the main loop, not this loop), since I don't think we want to disconnect twice?

didier-wenzek: > Is this `Outgoing::Disconnect` correct? If we break before the other server acknowledges we're disconnected, that still feels incorrect.

It is correct. I checked the disconnection on the broker.

An `Incoming::Disconnect` is only received when the disconnect is initiated by the broker.  The only event that could be used as a sentinel for an actual disconnect is `Err(MqttState(ConnectionAborted))`

> Maybe we need to `return` instead of `break` where we receive `Incoming::Disconnect` above (in the main loop, not this loop), since I don't think we want to disconnect twice?

Disconnecting twice is harmless.

crates/core/tedge/src/cli/mqtt/subscribe.rs
rina23q: done in 30906508844a93cf741398048174360e1f1da43d

crates/core/tedge/src/cli/mqtt/subscribe.rs
rina23q: done in 30906508844a93cf741398048174360e1f1da43d

crates/core/tedge/src/cli/mqtt/subscribe.rs
reubenmiller: Output looks great now, thanks.

reubenmiller (approved):
Apprved (functionality wise). Really nice improvements.

PR #3451, opened 2025-03-07
ci: add cargo-deny check

reubenmiller:
## Proposed changes

Adding [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) checks and config to enforce the following checks:

* Avoid using yanked versions of a crate (this did require some dependencies to be updated in order for the cargo-deny check to pass)
* Only allow crates following a pre-approved list of licenses (this will avoid accidental usage of a GPLv3 license which some customers forbid)
* Highlight unmaintained crates (some crates were added to the exceptions for now until we can properly review them )

Usage of duplicated creates are just warnings, but I've left the check as a warning as it highlights the potential need to consolidate some of our crates to reduce the dependency tree of the project.

cargo-deny has also been added to `just check` to ensure changes are picked up early.

```
just check
```

However the dependency linting/checks can be run using:

```
just check-dependencies
```

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


Bravo555 (approved):

deny.toml
```diff
@@ -0,0 +1,244 @@
...
+# https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
+[licenses]
+# List of explicitly allowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+allow = [
+    "MIT",
+    "Apache-2.0",
+    "Unicode-3.0",
+    "Unlicense",
```
rina23q, line 96: `Unlicense` is acceptable?

rina23q (approved):

deny.toml
reubenmiller: This list is purely generated by the list of software licenses currently used by thin-edge.io and its dependencies. We can always remove the dependencies which are using the `Unlicense` license, but that should be done in a follow up PR.

reubenmiller:
@rina23q @Bravo555 I had to make some mi
* Created a new just task `check-dependencies` so it can be called on its own, though it is still run via `just check` as well. It also updated the advisory db to ensure it is checking against up to date information
* the duplicate crate version warning is now ignored as this was just spamming the output.

rina23q:
> the duplicate crate version warning is now ignored as this was just spamming the output.

`--allow duplicate` is really helpful for the clean input indeed!

PR #3454, opened 2025-03-07
feat: de-duplicate operations from c8y/devicecontrol/notifications

jarhodes314:
## Proposed changes
De-duplicate operations using an in-memory store of operation IDs. This is an attempt at a fairly simple solution Cumulocity sending operations multiple times. There is a chance we need some persistence layer to make sure we don't execute an operation multiple times, but I believe the flakiness we are currently seeing in tests are due to the mapper publishing a 500 message on startup, and the issues occur while the mapper runs continuously. The window of time where this should be a problem is likely very limited as we only re-receive `PENDING` operations, and as soon as the operation is received by `tedge-agent`, it should get marked as `EXECUTING`.

One issue that exists with this solution is a (admittedly likely very slow) memory leak caused by the mapper storing the operation IDs indefinitely. Ideally, we should delete these at some point where we know the operation is no longer marked as pending, though I'm not sure how to do this.

Still TODO:
- [ ] Test the case where we don't initially handle the operation, e.g. we receive an unrecognised custom operation that's later registered by the time we next re-request pending operations
- [ ] Handle deleting the unneeded IDs

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -190,6 +190,8 @@ pub struct CumulocityConverter {
 
     supported_operations: SupportedOperations,
     pub operation_handler: OperationHandler,
+
+    processed_ids: HashSet<String>,
 }
 
 impl CumulocityConverter {
@@ -279,6 +281,7 @@ impl CumulocityConverter {
             command_id,
             active_commands: HashSet::new(),
             operation_handler,
+            processed_ids: HashSet::new(),
         })
     }
 
@@ -642,6 +645,9 @@ impl CumulocityConverter {
         let mut output = vec![];
         for operation_payload in operation_payloads {
             let operation = C8yOperation::from_json(operation_payload)?;
+            if self.processed_ids.contains(&operation.op_id) {
+                continue;
+            }
             let device_xid = operation.external_source.external_id;
             let cmd_id = self.command_id.new_id_with_str(&operation.op_id);
 
@@ -662,7 +668,10 @@ impl CumulocityConverter {
                     &operation_message,
                 )
                 .await;
-            let result = self.handle_c8y_operation_result(&result, Some(operation.op_id));
+            let result = self.handle_c8y_operation_result(&result, Some(operation.op_id.clone()));
+            if !result.is_empty() {
+                self.processed_ids.insert(operation.op_id);
+            }
             output.extend(result);
         }
 
@@ -1680,6 +1689,7 @@ pub(crate) mod tests {
     use tedge_http_ext::HttpResult;
     use tedge_mqtt_ext::test_helpers::assert_messages_matching;
     use tedge_mqtt_ext::MqttMessage;
+    use tedge_mqtt_ext::QoS;
     use tedge_mqtt_ext::Topic;
     use tedge_test_utils::fs::TempTedgeDir;
     use test_case::test_case;
@@ -2515,6 +2525,40 @@ pub(crate) mod tests {
         // a result between now and elapsed that is not 0 probably means that the operations are
         // blocking and that you probably removed a tokio::spawn handle (;
         assert_eq!(now.elapsed().as_secs(), 0);
+    }
+
+    #[tokio::test]
+    async fn operations_are_deduplicated() {
+        let tmp_dir = TempTedgeDir::new();
+        let (mut converter, _http_proxy) = create_c8y_converter(&tmp_dir).await;
+
+        let original = MqttMessage::new(&Topic::new_unchecked("c8y/devicecontrol/notifications"), json!(
+            {"delivery":{"log":[],"time":"2025-03-05T08:49:24.986Z","status":"PENDING"},"agentId":"1916574062","creationTime":"2025-03-05T08:49:24.967Z","deviceId":"1916574062","id":"16574089","status":"PENDING","c8y_Restart":{},"description":"do something","externalSource":{"externalId":"test-device","type":"c8y_Serial"}}
+        ).to_string());
+        let redelivery = MqttMessage::new(&Topic::new_unchecked("c8y/devicecontrol/notifications"), json!(
+            {"delivery":{"log":[{"time":"2025-03-05T08:49:24.986Z","status":"PENDING"},{"time":"2025-03-05T08:49:25.000Z","status":"SEND"},{"time":"2025-03-05T08:49:25.162Z","status":"DELIVERED"}],"time":"2025-03-05T08:49:25.707Z","status":"PENDING"},"agentId":"1916574062","creationTime":"2025-03-05T08:49:24.967Z","deviceId":"1916574062","id":"16574089","status":"PENDING","c8y_Restart":{},"description":"do something","externalSource":{"externalId":"test-device","type":"c8y_Serial"}}
+        ).to_string());
+        assert_eq!(
+            converter
+                .parse_c8y_devicecontrol_topic(&original)
+                .await
+                .unwrap(),
+            vec![MqttMessage {
+                topic: Topic {
+                    name: "te/device/main///cmd/restart/c8y-mapper-16574089".into(),
+                },
+                payload: json!({"status":"init"}).to_string().into(),
+                qos: QoS::AtLeastOnce,
+                retain: true,
+            },]
+        );
+        assert_eq!(
+            converter
+                .parse_c8y_devicecontrol_topic(&redelivery)
+                .await
+                .unwrap(),
+            vec![]
+        );
     }
 
     #[tokio::test]


albinsuresh (commented):
> One issue that exists with this solution is a (admittedly likely very slow) memory leak caused by the mapper storing the operation IDs indefinitely. Ideally, we should delete these at some point where we know the operation is no longer marked as pending, though I'm not sure how to do this.

Although clearing this entry when the operation transitions to `successful/failed` states is good enough for majority of the cases, I understand that it is not a fool-proof solution as there is still the possibility of getting a duplicate message while these state transition messages are still in transit to the cloud (or buffered for processing either locally or on the cloud). But since the terminal state transitions usually happen after the `executing` transition has already happened(which changes the `PENDING` status of the op in the cloud), the risk is reduced even further, although not fully eliminated.

I'm in favour of risking a duplicate operation execution in such rare cases(where the duplicate is delivered even after the terminal state transition is published), compared to the risk of that slow memory leak.

albinsuresh:
> we receive an unrecognised custom operation that's later registered

I don't see how this can happen as a device is not expected to receive an operation that it hasn't declared as a supported operation (this is my expectation from C8Y) via the operation capability registration.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -190,6 +190,8 @@ pub struct CumulocityConverter {
 
     supported_operations: SupportedOperations,
     pub operation_handler: OperationHandler,
+
+    processed_ids: HashSet<String>,
```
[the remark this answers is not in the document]
didier-wenzek: Looking at the doc comment for [`active_commands`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/converter.rs#L189) this sounds indeed as the correct place for that fix.

However, this raises an other point: why do we have this issue with duplicated commands while there is already a mechanism supposed to handle that?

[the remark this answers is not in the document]
jarhodes314: I think `active_commands` is a broken solution to the problem. I think there we are de-duplicating `c8y/devicecontrol/notifications`, but we are only tracking an active command once we receive the relevant tedge-topic command message, which will happen a short while later. I think this is what leaves open the window for an operation to be duplicated.

Assuming I've understood correctly, that would indicate that the problem could simply be solved by moving the `active_commands` insertion to where I'm currently inserting to `processed_ids`, and deleting the `processed_ids` stuff?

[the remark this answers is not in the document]
didier-wenzek: > Assuming I've understood correctly, that would indicate that the problem could simply be solved by moving the `active_commands` insertion to where I'm currently inserting to `processed_ids`, and deleting the `processed_ids` stuff?

One point is sure: one should keep only a single de-duplication mechanism. What you propose makes sense: it's better to remove duplicates before any processing.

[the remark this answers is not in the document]
albinsuresh: > why do we have this issue with duplicated commands while there is already a mechanism supposed to handle that?

For the cases where the duplicate messages are delivered after a restart, it is the lack of persistence of this set. But for duplicate messages delivered while the mapper is still live, this should have been sufficient.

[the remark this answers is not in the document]
jarhodes314: I've now modified `active_commands` to insert immediately post-conversion, rather than when we receive our outgoing message. As a result, I've deleted `processed_ids`.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2517,6 +2523,118 @@ pub(crate) mod tests {
...
+        let (mut converter, _http_proxy) = create_c8y_converter(&tmp_dir).await;
+
+        let original = MqttMessage::new(&Topic::new_unchecked("c8y/devicecontrol/notifications"), json!(
+            {"delivery":{"log":[],"time":"2025-03-05T08:49:24.986Z","status":"PENDING"},"agentId":"1916574062","creationTime":"2025-03-05T08:49:24.967Z","deviceId":"1916574062","id":"16574089","status":"PENDING","c8y_Restart":{},"description":"do something","externalSource":{"externalId":"test-device","type":"c8y_Serial"}}
+        ).to_string());
+        let redelivery = MqttMessage::new(&Topic::new_unchecked("c8y/devicecontrol/notifications"), json!(
+            {"delivery":{"log":[{"time":"2025-03-05T08:49:24.986Z","status":"PENDING"},{"time":"2025-03-05T08:49:25.000Z","status":"SEND"},{"time":"2025-03-05T08:49:25.162Z","status":"DELIVERED"}],"time":"2025-03-05T08:49:25.707Z","status":"PENDING"},"agentId":"1916574062","creationTime":"2025-03-05T08:49:24.967Z","deviceId":"1916574062","id":"16574089","status":"PENDING","c8y_Restart":{},"description":"do something","externalSource":{"externalId":"test-device","type":"c8y_Serial"}}
+        ).to_string());
```
didier-wenzek, line 2536: It would be good to make obvious that `original` and `redelivery` only differ on the `delivery` field`.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2517,6 +2523,118 @@ pub(crate) mod tests {
...
+                .parse_json_custom_operation_topic(&redelivery)
+                .await
+                .unwrap(),
+            vec![],
+            "Operation redelivery is ignored by converter"
+        );
+    }
+
+    #[tokio::test]
+    async fn custom_operations_are_not_deduplicated_before_registration() {
```
didier-wenzek, line 2595: I struggle to understand what is checked by this test and how.

`CumulocityConverter.supported_operations` field is patched before the first request then restored before the second. Okay, but why?

jarhodes314: This tests is checking what happens if we receive a custom operation that is unrecognised and later get redelivered the custom operation after it is registered with the mapper. If the converter naïvely assumes that the operation is active after we first receive it, the de-duplication mechanism will ignore the redelivery. But since we haven't yet processed this message, we should process such a redelivery. This is obviously dependent on something sending a 500 message once the operation is registered, but that could be the case for some custom operation handling service.

The patching of the `supported_operations` was intended as an easy way of ensuring the registered operations are made clear, since I'm not trying to test how we update `supported_operations` in this case.

didier-wenzek: This is clearer. Might be good to add this response as a comment to the test.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2517,6 +2523,118 @@ pub(crate) mod tests {
...
+            {"delivery":{"log":[{"time":"2025-03-05T08:49:24.986Z","status":"PENDING"},{"time":"2025-03-05T08:49:25.000Z","status":"SEND"},{"time":"2025-03-05T08:49:25.162Z","status":"DELIVERED"}],"time":"2025-03-05T08:49:25.707Z","status":"PENDING"},"agentId":"1916574062","creationTime":"2025-03-05T08:49:24.967Z","deviceId":"1916574062","id":"16574089","status":"PENDING","my_op":{},"description":"do something","externalSource":{"externalId":"test-device","type":"c8y_Serial"}}
+        ).to_string());
+        assert_ne!(
+            converter
+                .parse_json_custom_operation_topic(&original)
+                .await
+                .unwrap(),
+            vec![],
+            "Initial operation delivery produces outgoing message"
+        );
```
albinsuresh, line 2583: A slightly stricter check that validates at least the `cmd` topic would have been better than this non-empty check. To avoid false positives like a converted error message sent to `te/errors` instead of the expected `cmd` to beat this assertion.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -2517,6 +2523,118 @@ pub(crate) mod tests {
...
+                .parse_json_custom_operation_topic(&original)
+                .await
+                .unwrap(),
+            vec![],
+            "Operation is ignored before the operation is registered"
+        );
+
+        converter.supported_operations = after_registration;
+
+        assert_ne!(
```
albinsuresh, line 2628: Same comment as above, regarding a stricter check.

crates/extensions/c8y_mapper_ext/src/converter.rs
albinsuresh: Technically this is fine. But seeing the detailed discussion above, it felt like this case could have been better represented as an integration test in `tests.rs`, where we can better simulate the dynamic custom operation registration during the execution of the test.

albinsuresh (commented):
Changes look fine. Some minor suggestions on the tests.

crates/extensions/c8y_mapper_ext/src/converter.rs
jarhodes314: I've now removed some of the extraneous fields and changed things so we send the original message twice, since the converter doesn't care about the `delivery` field.

crates/extensions/c8y_mapper_ext/src/converter.rs
jarhodes314: I think I want the opposite of that, @albinsuresh. The point of this being a unit test is I don't also want to test the operation registration logic at the same time. I've added a comment to explain what it is I'm trying to test.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1399,6 +1410,21 @@ impl CumulocityConverter {
         &mut self,
         message: &MqttMessage,
     ) -> Result<Vec<MqttMessage>, ConversionError> {
+        if dbg!(self.active_commands_last_cleared.elapsed(&*self.clock)) > Duration::from_secs(3600)
+        {
+            let mut to_remove = vec![];
+            for (id, time) in &self.active_commands {
+                // Expire tasks after 12 hours
+                if dbg!(time.elapsed(&*self.clock)) > Duration::from_secs(3600 * 12) {
```
didier-wenzek, line 1418: As above:

```suggestion
                if self.clock.elapsed_since(time) > Duration::from_secs(3600 * 12) {
```

crates/common/clock/src/lib.rs
```diff
@@ -16,3 +20,58 @@ impl Clock for WallClock {
...
+/// A wrapper type to prevent accidental use of [Instant::elapsed]
+///
+/// Once a time is produced by [MonotonicClock], we don't want to use
+/// [Instant::elapsed] as that will compare to [Instant::now] and not
+/// [MontonicClock::now].
+pub struct TedgeInstant(Instant);
+
+pub struct SystemClock;
+
+impl MonotonicClock for SystemClock {
```
didier-wenzek, line 39: I'm okay with that. However, have you considered [`tokio::time::pause`](https://docs.rs/tokio/latest/tokio/time/fn.pause.html) and  [`tokio::time::advance`](https://docs.rs/tokio/latest/tokio/time/fn.advance.html)?

didier-wenzek (commented):
Two forgotten `dbg!` to be removed and some questions.

crates/common/clock/src/lib.rs
jarhodes314: I'd forgotten about those, I'll change to using them since this introduces unnecessary complexity.

didier-wenzek (approved):
Approved. Thank you.

crates/common/clock/src/lib.rs
didier-wenzek: Indeed, far simpler!

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1403,6 +1406,20 @@ impl CumulocityConverter {
         &mut self,
         message: &MqttMessage,
     ) -> Result<Vec<MqttMessage>, ConversionError> {
+        if self.active_commands_last_cleared.elapsed() > Duration::from_secs(3600) {
```
albinsuresh, line 1409: I was actually thinking that you'd use the timer actor from the mapper to get timeout notification message at desired intervals. But, this was smart.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1399,6 +1406,20 @@ impl CumulocityConverter {
...
+            let mut to_remove = vec![];
+            for (id, time) in &self.active_commands {
+                // Expire tasks after 12 hours
+                if time.elapsed() > Duration::from_secs(3600 * 12) {
+                    to_remove.push(id.to_owned());
```
albinsuresh, line 1414: Although it's highly unlikely for a normal operation to keep running for over 12 hours, there could be cases where an operation like firmware update may get delayed by the workflow because it is scheduled to be executed during non-busy hours(late night/early morning). For e.g, the device receives the firmware update at 9 AM in the morning, but postpones the execution because the device is "busy" until 11 PM at night, or until some point in future where it can decide that the device is "not busy". In such cases, this 12 hours assumption might not be safe enough.

An easier solution would be to use an even higher value, but even that would be arbit. The safest option would be not to remove until the the operation has fully completed (in its terminal state or cleared)

albinsuresh (commented):
The changes look much simpler now. I've gone one concern though, regarding the premature eviction from the cache, before the operation really completes.

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: This cache only prevents creation of multiple commands for an operation and stores no state. So clearing the cache will no break a pending operation.

jarhodes314: We do also remove operations when they are complete, but to do this for c8y custom operations too we would need to encode the logic for how c8y processes incoming smartrest commands (including updating the operation by either type or id). Because we only use this for de-duplication, it really only need represent pending operations, non-completed operations is overkill.

albinsuresh (approved):
LGTM.

didier-wenzek (approved):
Approved

Issue #3456, opened 2025-03-10
tedge-agent should not block processing commands if one command is unacknowledged by the creator


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Workflows processed by `tedge-agent` currently await the workflow result to be cleared by the user/component which created it.

New command requests for the same workflow are then queued until any previously in-use workflow (of the same type) has been cleared. This makes the `tedge-agent` fragile to clients who either forget to clear an existing command, which then will block other users from processing their requests.

**Problem 1: Badly behaving client which fails to acknowledge the result**

1. Client 1: Create workflow for firmware_update (cmd_id=1)
1. tedge-agent: processes firmware_update (cmd_id=1), and sets the status to "successful" or "failed"
1. Client 2: Create workflow for firmware_update (cmd_id=2)
1. tedge-agent: waits until cmd_id=1 has been acknowledged by Client 1 (and cleared) until processing cmd_id=2, but if Client 1 never acknowledges the result, then the tedge-agent will be blocked indefinitely

**Problem 2: tedge-agent fails to see acknowledgement**

If some (which is currently the case with mosquitto > 2.0.11, <= 2.0.21), then it is possible that the clearing of the command is not seen by the tedge-agent, and therefore it will block any future commands until the tedge-agent is restarted (as on startup it will check for any existing retain messages, and reconcile the any in-progress commands). Whilst this is mainly due to existing [mosquitto bug #2618](https://github.com/eclipse-mosquitto/mosquitto/issues/2618), the problem would also exist if the MQTT server isn't configured with persistence.

Below shows the potentially problematic sequ

1. Client 1: Create workflow for firmware_update (cmd_id=1)
1. tedge-agent: processes firmware_update (cmd_id=1), and sets the status to "successful" or "failed"
1. tedge-agent: Gets disconnected from the MQTT broker (possibly due to the device being restarted)
1. Client 1: Clears the cmd_id=1 message (tedge-agent is still disconnected)
1. tedge-agent: Reconnects to the MQTT broker, but does not receive the cleared retain message, so it will not process any future command of the same type until the tedge-agent is restarted or another client re-sends the clearing message (but this is unlikely as it would be difficult to find the correct message id)

**Describe the solution you'd like**

The solution is open up for discussion, but below are some questions to think about:

[...]

PR #3462, opened 2025-03-11
fix: subscribe to topics on reconnection when session persistence is not enabled or broken

reubenmiller:
## Proposed changes

Resubscribe to MQTT topics on reconnection when the no session is present. No session is present when mosquitto persistence has not been configured, or inaccessible (e.g. corrupt or incorrect ownership/permissions).

Previously, the reconnection logic was assuming that the only time the client needs to resubscribe to the MQTT topics is when an anonymous session is used. However if the MQTT broker persistence layer is not working, then the client should be checking if a session is present or not, and then resubscribing to the topics if no session was restored.

The most notable symptom without this change was that the tedge-agent would stop reacting to operations sent via MQTT if it would be disconnected from the local MQTT broker and if the MQTT broker persistence layer was not working or enabled. Restarting the tedge-agent service would "fix" the situation but only until the next reconnection event.

## Types of changes

## Paste Link to the issue

* Discovered during the investigation of this ticket: https://github.com/thin-edge/thin-edge.io/issues/3456

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -271,9 +271,10 @@ impl Connection {
                                 .await?;
                         }
 
-                        if config.session_name.is_none() {
+                        if config.session_name.is_none() || !ack.session_present {
                             // Workaround for  https://github.com/bytebeamio/rumqtt/issues/250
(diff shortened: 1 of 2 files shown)


didier-wenzek (approved):
Approved

PR #3474, opened 2025-03-14
feat: Entity twin data CRUD REST APIs

albinsuresh:
## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/http_server/entity_store.rs
+++ b/crates/core/tedge_agent/src/http_server/entity_store.rs
@@ -32,6 +32,7 @@ use tedge_api::entity::EntityMetadata;
 use tedge_api::entity::InvalidEntityType;
 use tedge_api::entity_store;
 use tedge_api::entity_store::EntityRegistrationMessage;
+use tedge_api::entity_store::EntityTwinMessage;
 use tedge_api::entity_store::ListFilters;
 use tedge_api::mqtt_topics::EntityTopicId;
 use tedge_api::mqtt_topics::TopicIdError;
@@ -115,6 +116,9 @@ enum Error {
 
     #[error(transparent)]
     InvalidTwinData(#[from] InvalidTwinData),
+
+    #[error("Path: {0} not found")]
+    PathNotFound(String),
 }
 
 impl IntoResponse for Error {
@@ -131,6 +135,7 @@ impl IntoResponse for Error {
             Error::InvalidEntityStoreResponse => StatusCode::INTERNAL_SERVER_ERROR,
             Error::InvalidInput(_) => StatusCode::BAD_REQUEST,
             Error::InvalidTwinData(_) => StatusCode::BAD_REQUEST,
+            Error::PathNotFound(_) => StatusCode::NOT_FOUND,
         };
         let error_message = self.to_string();
 
@@ -143,9 +148,10 @@ pub(crate) fn entity_store_router(state: AgentState) -> Router {
         .route("/v1/entities", post(register_entity).get(list_entities))
         .route(
             "/v1/entities/{*path}",
-            get(get_entity)
-                .patch(patch_entity)
-                .delete(deregister_entity),
+            get(get_resource)
+                .put(put_resource)
+                .patch(patch_resource)
+                .delete(delete_resource),
         )
         .with_state(state)
 }
@@ -170,13 +176,91 @@ async fn register_entity(
     ))
 }
 
-async fn patch_entity(
+async fn get_resource(
+    State(state): State<AgentState>,
+    Path(path): Path<String>,
+) -> Result<impl IntoResponse, Error> {
+    let segments = path.split('/').collect::<Vec<&str>>();
+    match segments.as_slice() {
+        [seg1, seg2, seg3, seg4] => {
+            let topic_id = topic_id_from_path_segments(seg1, seg2, seg3, seg4)?;
+            get_entity(state, topic_id)
+                .await
+                .map(|res| res.into_response())
+        }
+        [seg1, seg2, seg3, seg4, "twin", fragment_key] => {
+            let topic_id = topic_id_from_path_segments(seg1, seg2, seg3, seg4)?;
+            get_entity_twin_fragment(state, topic_id, fragment_key.to_string())
+                .await
+                .map(|res| res.into_response())
+        }
+        _ => Err(Error::PathNotFound(path)),
+    }
+}
+
+async fn put_resource(
+    State(state): State<AgentState>,
+    Path(path): Path<String>,
+    Json(value): Json<Value>,
+) -> impl IntoResponse {
+    let segments = path.split('/').collect::<Vec<&str>>();
+    match segments.as_slice() {
+        [seg1, seg2, seg3, seg4, "twin", fragment_key] => {
+            let topic_id = topic_id_from_path_segments(seg1, seg2, seg3, seg4)?;
+            Ok(upsert_entity_twin_fragment(state, topic_id, fragment_key.to_string(), value).await)
+        }
+        _ => Err(Error::PathNotFound(path)),
+    }
+}
+
+async fn patch_resource(
     State(state): State<AgentState>,
     Path(path): Path<String>,
     Json(twin_fragments): Json<Map<String, Value>>,
 ) -> impl IntoResponse {
-    let topic_id = EntityTopicId::from_str(&path)?;
-    let twin_data = EntityTwinData::try_new(topic_id, twin_fragments)?;
+    let segments = path.split('/').collect::<Vec<&str>>();
+    match segments.as_slice() {
+        [seg1, seg2, seg3, seg4] => {
+            let topic_id = topic_id_from_path_segments(seg1, seg2, seg3, seg4)?;
+            Ok(patch_entity(state, topic_id, twin_fragments).await)
+        }
+        _ => Err(Error::PathNotFound(path)),
+    }
+}
+
+async fn delete_resource(
+    State(state): State<AgentState>,
+    Path(path): Path<String>,
+) -> Result<Response, Error> {
+    let segments = path.split('/').collect::<Vec<&str>>();
+    match segments.as_slice() {
+        [seg1, seg2, seg3, seg4] => {
+            let topic_id = topic_id_from_path_segments(seg1, seg2, seg3, seg4)?;
+            deregister_entity(state, topic_id).await
+        }
+        [seg1, seg2, seg3, seg4, "twin", fragment_key] => {
+            let topic_id = topic_id_from_path_segments(seg1, seg2, seg3, seg4)?;
+            delete_entity_twin_fragment(state, topic_id, fragment_key.to_string()).await
+        }
+        _ => Err(Error::PathNotFound(path)),
+    }
+}
+
+fn topic_id_from_path_segments(
+    seg1: &str,
+    seg2: &str,
+    seg3: &str,
+    seg4: &str,
+) -> Result<EntityTopicId, TopicIdError> {
+    EntityTopicId::from_str(&format!("{}/{}/{}/{}", seg1, seg2, seg3, seg4))
+}
+
+async fn patch_entity(
+    state: AgentState,
+    topic_id: EntityTopicId,
+    twin_fragments: Map<String, Value>,
+) -> impl IntoResponse {
+    let twin_data = EntityTwinData::try_new(topic_id.clone(), twin_fragments)?;
 
     let response = state
         .entity_store_handle
@@ -188,17 +272,15 @@ async fn patch_entity(
     };
     res?;
 
-    let entity = get_entity(State(state), Path(path)).await?;
+    let entity = get_entity(state, topic_id).await?;
 
     Ok(entity)
 }
 
 async fn get_entity(
-    State(state): State<AgentState>,
-    Path(path): Path<String>,
-) -> Result<Json<EntityMetadata>, Error> {
-    let topic_id = EntityTopicId::from_str(&path)?;
-
+    state: AgentState,
+    topic_id: EntityTopicId,
+) -> Result<impl IntoResponse, Error> {
     let response = state
         .entity_store_handle
         .clone()
@@ -210,18 +292,13 @@ async fn get_entity(
     };
 
     if let Some(entity) = entity_metadata {
-        Ok(Json(entity.clone()))
+        Ok(Json(entity))
     } else {
         Err(Error::EntityNotFound(topic_id))
     }
 }
 
-async fn deregister_entity(
-    State(state): State<AgentState>,
-    Path(path): Path<String>,
-) -> Result<Response, Error> {
-    let topic_id = EntityTopicId::from_str(&path)?;
-
+async fn deregister_entity(state: AgentState, topic_id: EntityTopicId) -> Result<Response, Error> {
     let response = state
         .entity_store_handle
         .clone()
@@ -255,6 +332,68 @@ async fn list_entities(
     };
 
     Ok(Json(entities))
+}
+
+async fn get_entity_twin_fragment(
+    state: AgentState,
+    topic_id: EntityTopicId,
+    fragment_key: String,
+) -> Result<impl IntoResponse, Error> {
+    let response = state
+        .entity_store_handle
+        .clone()
+        .await_response(EntityStoreRequest::GetTwinFragment(topic_id, fragment_key))
+        .await?;
+    let EntityStoreResponse::GetTwinFragment(fragment_value) = response else {
+        return Err(Error::InvalidEntityStoreResponse);
+    };
+
+    Ok(Json(fragment_value))
+}
+
+async fn upsert_entity_twin_fragment(
+    state: AgentState,
+    topic_id: EntityTopicId,
+    fragment_key: String,
+    fragment_value: Value,
+) -> impl IntoResponse {
+    let twin_data = EntityTwinMessage::new(topic_id.clone(), fragment_key, fragment_value);
+
+    let response = state
+        .entity_store_handle
+        .clone()
+        .await_response(EntityStoreRequest::UpdateTwinFragment(twin_data))
+        .await?;
+    let EntityStoreResponse::UpdateTwinFragment(res) = response else {
+        return Err(Error::InvalidEntityStoreResponse);
+    };
+    res?;
+
+    let entity = get_entity(state, topic_id).await?;
+
+    Ok(entity)
+}
+
+async fn delete_entity_twin_fragment(
+    state: AgentState,
+    topic_id: EntityTopicId,
+    fragment_key: String,
+) -> Result<Response, Error> {
+    let twin_data = EntityTwinMessage::new(topic_id.clone(), fragment_key, Value::Null);
+
+    let response = state
+        .entity_store_handle
+        .clone()
+        .await_response(EntityStoreRequest::UpdateTwinFragment(twin_data))
+        .await?;
+    let EntityStoreResponse::UpdateTwinFragment(res) = response else {
+        return Err(Error::InvalidEntityStoreResponse);
+    };
+    res?;
+
+    let entity = get_entity(state, topic_id).await?;
+
+    Ok(entity.into_response())
 }
 
 #[cfg(test)]
--- a/tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
+++ b/tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
@@ -1,6 +1,7 @@
 *** Settings ***
 Resource            ../../../resources/common.resource
 Library             ThinEdgeIO
+Library             Collections
 
 Suite Setup         Custom Setup
 Test Teardown       Get Logs    ${DEVICE_SN}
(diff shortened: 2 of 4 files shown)


crates/core/tedge_agent/src/http_server/entity_store.rs
```diff
@@ -170,13 +181,113 @@ async fn register_entity(
...
+        [seg1, seg2, seg3, seg4, "twin", fragment_key] => {
+            let topic_id = topic_id_from_path_segments(seg1, Some(seg2), Some(seg3), Some(seg4))?;
+            Ok((
+                topic_id,
+                Channel::EntityTwinData {
+                    fragment_key: fragment_key.to_string(),
+                },
+            ))
+        }
+        _ => Err(Error::PathNotFound(path.to_string())),
```
didier-wenzek, line 266: The error cases must give move insight to the user.

Currently, the followings lead to the same error message telling the path is not found.

* Incorrect channel: `device/main///cmd/maintenance_window`
* Missing fragment key: `device/main///twin`
* Incorrect fragment key: `device/main///twin/maintenance/window not`

didier-wenzek (commented):
The code is correct but the error messages have to be improved.

crates/core/tedge_agent/src/http_server/entity_store.rs
```diff
@@ -143,9 +153,10 @@ pub(crate) fn entity_store_router(state: AgentState) -> Router {
         .route("/v1/entities", post(register_entity).get(list_entities))
         .route(
             "/v1/entities/{*path}",
-            get(get_entity)
-                .patch(patch_entity)
-                .delete(deregister_entity),
+            get(get_resource)
+                .put(put_resource)
+                .patch(patch_resource)
```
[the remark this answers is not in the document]
albinsuresh: > It's not really clear how patch is intended to be used.

The `PATCH` method is only implemented for `entities/<entity-topic-id>` endpoint, where the resource is the entity itself, as patch here means you're patching the existing entity with additional twin data. But, `PATCH` is not implemented for the `twin/<fragment_key>` endpoint, where the resource is the twin fragment itself, as inserts/updates of the twin values were conceived as `PUT` instead of a `PATCH`. I didn't overload `PATCH` here to keep that option open for patching an existing twin fragments by adding/updating sub-fragments.

> But the variant above gives an error message that is difficult to understand

Yeah, that's misleading. It should have been a 405 Method Not Allowed.

crates/core/tedge_agent/src/http_server/entity_store.rs
albinsuresh: I was just trying to mimic the simple `Not Found` message emitted by axum when a non-existent resource path is queried. Will fix that by adding more context.

crates/core/tedge_agent/src/http_server/entity_store.rs
```diff
@@ -115,6 +117,12 @@ enum Error {
 
     #[error(transparent)]
     InvalidTwinData(#[from] InvalidTwinData),
+
+    #[error("Path: {0} not found")]
+    PathNotFound(String),
```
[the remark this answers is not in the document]
albinsuresh: I was just trying to mimic the simple Not Found message emitted by axum when a non-existent resource path is queried. Will fix that by adding more error variants.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -126,6 +127,28 @@ Delete entity tree
     ...    {"@topic-id":"device/child2//","@parent":"device/main//","@type":"child-device"}
     ...    ${entities}
 
+Entity twin update apis
+    ${entity}=    Execute Command
+    ...    curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin/maintenance_window -H 'Content-Type: application/json' -d '5'
+    ${entity}=    Evaluate    ${entity}
+    Should Be Equal    ${entity["maintenance_window"]}    ${5}
```
reubenmiller, line 134: When setting a digital twin value, the `PUT` response is different to the `GET` response for the same API endpoint, where it is expected that it should be the same.

For example, below shows the response from adding a new digital twin value:

```
$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/hello --data '"hello"'
{"@topic-id":"device/main//","@type":"device","hello":"hello"}
```

Then retrieving the digital twin value using `GET` returns:

```
$ tedge http get /tedge/entity-store/v1/entities/device/main///twin/hello
"hello"
```

So for this test case I would expect that the user is only interacting with the `twin` data, and not the entity store data.

reubenmiller:
It might be worth change the api endpoint path, as there is some duplication of information which then makes it harder to extend functionality in the future.

**From**
```
/tedge/entity-store/v1/entities/device/main///twin/hello
```

**To**

```sh
/tedge/v1/entities/device/main///twin/hello
```

I also wouldn't be opposed to changing out the word "entities" for a more use-friendly name...but until we have another suggestion, then we will stick with it.

crates/core/tedge_agent/src/http_server/entity_store.rs
reubenmiller: We should also prevent empty fragment keys as well as the following leads to an empty key value.

```sh
$ tedge http put '/tedge/entity-store/v1/entities/device/main///twin/' '{"hello":"foo"}'
{"@topic-id":"device/main//","@type":"device","":{"hello":"foo"},"hello":"foo2","hello2":"foo2"}
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Both `PUT` and `DELETE` are returning the updated entity definition, which I thought the users would benefit from. Of course, they could do another `GET` to get the same. If that isn't desired, then what should the response be? Just `200 OK` is fine?

crates/core/tedge_agent/src/http_server/entity_store.rs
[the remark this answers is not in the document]
reubenmiller: I'm also having problems reconciling the replacing of all of the existing digital twin keys using the following command:

```sh
$ tedge http put /tedge/entity-store/v1/entities/device/main// '{"foo":"bar"}'
Error: failed to PUT http://127.0.0.1:8000/tedge/entity-store/v1/entities/device/main//

Caused by:
    HTTP client error: 404 Not Found
    {"error":"Path: device/main// not found"}
```

Using `patch` would work to add/edit values, but not to replace all other values.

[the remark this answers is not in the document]
reubenmiller: Or should we keep the entity registration information separate to the digital twin information, and instead it would be a put to:

```sh
$ tedge http put /tedge/entity-store/v1/entities/device/main///twin '{"foo":"bar"}'

# or 
$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/ '{"foo":"bar"}'
```

albinsuresh:
> It might be worth change the api endpoint path, as there is some duplication of information which then makes it harder to extend functionality in the future.
> 
> **From**
> 
> ```
> /tedge/entity-store/v1/entities/device/main///twin/hello
> ```
> 
> **To**
> 
> ```shell
> /tedge/v1/entities/device/main///twin/hello
> ```
> 
> I also wouldn't be opposed to changing out the word "entities" for a more use-friendly name...but until we have another suggestion, then we will stick with it.

Agree about the duplication. The `/tedge/entity-store` prefix was added just to keep it consistent with the `/tedge/file-transfer` endpoints which enabled a clear separation of those service endpoints as well. But, I feel we can achieve the same by using `/v1/entities` prefix as well.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -1,5 +1,6 @@
 *** Settings ***
 Resource            ../../../resources/common.resource
+Library             Collections
```
reubenmiller, line 3: Is this line actually needed? I ran the test without it and it passed.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
reubenmiller, line 134: Robotframework has a keyword to do this rather than using the expression evaluation, `${5}`.

```suggestion
    Should Be Equal As Integers    ${entity["maintenance_window"]}    5
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -126,6 +127,28 @@ Delete entity tree
...
+    Should Be Equal    ${get}    5
+
+    ${timestamp}=    Get Unix Timestamp
+    ${entity}=    Execute Command
+    ...    curl -X DELETE http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin/maintenance_window
+    ${entity}=    Evaluate    ${entity}
+    Should Not Contain    ${entity}    maintenance_window
+    Should Have MQTT Messages
+    ...    te/device/main///twin/maintenance_window
+    ...    date_from=${timestamp}
```
reubenmiller, line 150: It would be better to check if the retained message has been cleared rather than just checking if any message has been received (as it could also be a value update).

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
reubenmiller: I think the main point is that the `twin` api endpoint is solely about the twin data, and not the entity data. So the `PUT` response should be exactly the same as the `GET` response and both return a HTTP 200 status code.

For the `DELETE` twin command, I would just return a `204` (No Content). This is different to the `DELETE` on the entity store, as the entity store was deleting a tree of devices, so it was unclear which devices are being deleted (as it behaved more like a query), where as for deleting a digital twin property, the context is already clear in the URL, and the user does not care about the value (that's why they are deleting it).

crates/core/tedge_agent/src/http_server/entity_store.rs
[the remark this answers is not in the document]
albinsuresh: Resolved by f837935c

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by f837935c

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by f837935c

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by f837935c

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by f837935c

crates/core/tedge_agent/src/http_server/entity_store.rs
didier-wenzek: > Resolved by f837935c

Now, working as expected

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
didier-wenzek: > Resolved by f837935

Perfect. Now, `PUT` returns no content and `GET` returns what has been `PUT`. `DELETE` returns `204 No Content` as expected.

crates/core/tedge_agent/src/http_server/entity_store.rs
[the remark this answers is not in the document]
didier-wenzek: >> But the variant above gives an error message that is difficult to understand

> Yeah, that's misleading. It should have been a 405 Method Not Allowed.

Okay now, the following returns a `405 Method Not Allowed`.

```
$ tedge http patch /tedge/entity-store/v1/entities/device/main///twin/maintenance_window '{"days":43}'
Error: failed to PATCH http://127.0.0.1:8000/tedge/entity-store/v1/entities/device/main///twin/maintenance_window

Caused by:
    HTTP client error: 405 Method Not Allowed
    {"error":"Method Not Allowed"}
```

However it would be a bit better to tell what is actually not allowed: i.e. to patch a fragment.

Also, as patching a fragment is not allowed it would be better to return a `405 Method Not Allowed` in the following case:

```
$ tedge http patch /tedge/entity-store/v1/entities/device/main///twin/maintenance_window '"42"'
Error: failed to PATCH http://127.0.0.1:8000/tedge/entity-store/v1/entities/device/main///twin/maintenance_window

Caused by:
    HTTP client error: 422 Unprocessable Entity
    Failed to deserialize the JSON body into the target type: invalid type: string "42", expected a map at line 1 column 4
```

didier-wenzek (commented):
I will be happy to approve - once the unit tests fixed. Also, I have some suggestions to improve the error messages.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
reubenmiller: Something is still a bit strange. Doing a PUT does return something (which is expected imo), however it only returns something when the value has changed (which is unexpected).

For example, the following shows the unexpected behaviour:

```sh
# first time returns the value which was 'put', 5
$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/maintenance_window '5'

# second time, no response is given
$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/maintenance_window '5'
```

Let's keep the response stateless, and just always return the current value (yes this may be the same as what was sent), but it is there to confirm that the server's interpretation of the payload is in fact the same.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -126,6 +113,134 @@ Delete entity tree
...
+
+    ${timestamp}=    Get Unix Timestamp
+    ${http_code}=    Execute Command
+    ...    curl -o /dev/null --silent --write-out "%\{http_code\}" -X DELETE http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin/maintenance_window
+    Should Be Equal    ${http_code}    204
+    Should Have MQTT Messages
+    ...    te/device/main///twin/maintenance_window
+    ...    date_from=${timestamp}
+    ${retained_message}=    Execute Command
+    ...    tedge mqtt sub --no-topic te/device/main///twin/maintenance_window --duration 1
```
reubenmiller, line 136: Yes the duration will default to seconds, but it would be better to be explicit, as the default unit is not clear to the reader.

```suggestion
    ...    tedge mqtt sub --no-topic te/device/main///twin/maintenance_window --duration 1s
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -126,6 +113,134 @@ Delete entity tree
...
+    ${new_payload}=    Set Variable    {"maintainer":"John Doe","maintenance_mode":true,"maintenance_window":5}
+    ${put}=    Execute Command
+    ...    curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin -H 'Content-Type: application/json' -d '${new_payload}'
+    Should Be Equal    ${put}    ${new_payload}
+    Should Have MQTT Messages
+    ...    te/device/main///twin/maintenance_mode
+    ...    message_contains=true
+    Should Have MQTT Messages
+    ...    te/device/main///twin/maintenance_window
+    ...    message_contains=5
```
reubenmiller, line 151: It would be worth while also adding a test for replacing all of the existing values with an empty object, as currently there is a bit of an inconsistency when putting `{}` to the `/twin/` endpoint.

For example:

```sh
tedge http put /tedge/entity-store/v1/entities/device/main///twin/ '{}'

# Current: <no response>
# Expected: {}
```

reubenmiller:
It's shaping up nicely. I just found some slightly unexpected results whilst testing.

### Unexpected json payloads which are accepted

This is probably due to the crate we're using to parse the JSON, but I found some unexpected results when sending invalid json payloads (I wouldn't say that I'm against it, but wanted to point it out as it might lead to some unexpected results for users). I'm assuming the crate is doing some json token parsing and only parsing the first "token" (e.g. integer, string, object, array etc.)

All of the following digital twin updates work and don't return an error (despite the payloads technically not being full valid json (from start to finish))

```sh
$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/test '12-2'
=> 12

$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/test '12.12.1'
=> 12.12

$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/test 'true true'
=> true

$ tedge http put /tedge/entity-store/v1/entities/device/main///twin/test '{"a":1}{"b":2}'
=> {"a":1}
```


### What size limit to we impose on the payloads?

This would be useful to users to at least guide them to not publishing huge payloads, and we should make sure it aligns with the max payload size used by the local MQTT broker (if we have some limit set).

Below shows the current response when trying to upload a 5MB json content where the request is rejected (which is good), however I suspect it wasn't an active check, and just the default buffer size that we're using.

```sh
$ ls -lh body.json
-rw-r--r-- 1 root root 5.7M Mar 20 18:39 body.json

$ curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin/maintenance_window -d @body.json -H "Content-Type: application/json"
Failed to buffer the request body: length limit exceeded
```

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
```diff
@@ -126,6 +108,138 @@ Delete entity tree
...
+    ${get}=    Execute Command
+    ...    curl http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin
+    Should Be Equal
+    ...    ${get}
+    ...    ${new_payload}
+
+    # Replace existing twins
+    ${timestamp}=    Get Unix Timestamp
+    ${new_payload}=    Set Variable    {"maintenance_mode":false}
+    ${put}=    Execute Command
```
reubenmiller, line 157: This looks to be the offending line which is causing the RF lint error (e.g. `${put}` assigned to be never used.

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by 32305a5

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by 7d725d7

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by 6156bcb

tests/RobotFramework/tests/tedge_agent/http_server/entity_store_api.robot
albinsuresh: Resolved by 32305a5. I was trying to give a more accurate response to the user whether the update really updated something or not. But, since keeping these methods as idempotent is the convention/desired, I've changed it.

didier-wenzek (approved):
Approved. Thank you for your perseverance polishing this new API.

I would just consider to simplify the api endpoint prefix as suggested https://github.com/thin-edge/thin-edge.io/pull/3474#issuecomment-2733462317

albinsuresh:
> I would just to consider to simplify the api endpoint prefix as suggested [#3474 (comment)](https://github.com/thin-edge/thin-edge.io/pull/3474#issuecomment-2733462317)

Yup, that's the plan. Had discussed it with @reubenmiller offline and felt that it would be better to do it as an independent PR.

reubenmiller:
> > ### What size limit to we impose on the payloads?
> > This would be useful to users to at least guide them to not publishing huge payloads, and we should make sure it aligns with the max payload size used by the local MQTT broker (if we have some limit set).
> > Below shows the current response when trying to upload a 5MB json content where the request is rejected (which is good), however I suspect it wasn't an active check, and just the default buffer size that we're using.
> > ```shell
> > $ ls -lh body.json
> > -rw-r--r-- 1 root root 5.7M Mar 20 18:39 body.json
> > 
> > $ curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin/maintenance_window -d @body.json -H "Content-Type: application/json"
> > Failed to buffer the request body: length limit exceeded
> > ```
> 
> Resolved by [989c944](https://github.com/thin-edge/thin-edge.io/commit/989c944b665bef7cf9066b67a101a7fb421593c0).

This doesn't seem to be working as expected...I get a buffer size error for a file size around 6.6MB before it hits the upper limit of the defined threshold)

```sh
curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin -d @body2.json -H "Content-Type: application/json"
Failed to buffer the request body: length limit exceeded
```

It seems that we need to set the max body size on the [axum server](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html).

crates/core/tedge_agent/src/http_server/entity_store.rs
```diff
@@ -170,35 +188,141 @@ async fn register_entity(
...
+async fn put_resource(
+    State(state): State<AgentState>,
+    Path(path): Path<String>,
+    payload: String,
+) -> impl IntoResponse {
+    if payload.len() > HTTP_MAX_PAYLOAD_SIZE {
+        return Err(Error::EntityStoreError(
+            entity_store::Error::PayloadSizeExceeded(payload.len()),
+        ));
+    }
```
albinsuresh, line 224: Wasted a lot of time trying to use the [RequestBodyLimitLayer](https://docs.rs/tower-http/0.6.2/tower_http/limit/struct.RequestBodyLimitLayer.html) with axum `Router`, but failed. So, falling back to this instead.

albinsuresh: Well... It was my bad. I was wrongly trying to integrate the `tower-http` abstraction [RequestBodyLimitLayer](https://docs.rs/tower-http/0.6.2/tower_http/limit/struct.RequestBodyLimitLayer.html) instead [DefaultBodyLimit](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html) which is the right `axum` router level abstraction. Resolved by 69a6f6f.

albinsuresh:
> This doesn't seem to be working as expected...I get a buffer size error for a file size around 6.6MB before it hits the upper limit of the defined threshold)
> 
> ```shell
> curl -X PUT http://localhost:8000/tedge/entity-store/v1/entities/device/main///twin -d @body2.json -H "Content-Type: application/json"
> Failed to buffer the request body: length limit exceeded
> ```
> 
> It seems that we need to set the max body size on the [axum server](https://docs.rs/axum/latest/axum/extract/struct.DefaultBodyLimit.html).

Resolved in 69a6f6f by applying the 1 MB limit as agreed.

reubenmiller (approved):
Approved. Thanks for the persistence!

Issue #3476, opened 2025-03-15
tedge connect c8y shows warnings about certificates when connect via basic auth credentials


reubenmiller:
**Describe the bug**

When connecting to Cumulocity using Basic Auth (e.g. `tedge config set c8y.auth_method basic`), the `tedge connect c8y` prints out unnecessary warnings about failing to change the ownership of the public and private certificate, which is unexpected as a device certificate is not needed when connecting via Basic Auth credentials.

Below shows the `tedge connect c8y` lines with the two `WARN` lines which should not be present when using basic auth.

```sh
root@tedge:~# tedge connect c8y
Connecting to Cumulocity with config:
        device id: rmi_test1001
        cloud profile: <none>
        cloud host: thin-edge-io.eu-latest.cumulocity.com:8883
        auth method: basic
        credentials path: /etc/tedge/credentials.toml
        bridge: mosquitto
        service manager: systemd
        mosquitto version: 2.0.11
Creating device in Cumulocity cloud... ✓
2025-03-15T09:04:20.14318654Z  WARN tedge::cli::connect::command: Failed to change ownership of /etc/tedge/device-certs/tedge-certificate.pem to mosquitto:mosquitto: The path is not accessible. "/etc/tedge/device-certs/tedge-certificate.pem"
2025-03-15T09:04:20.143328557Z  WARN tedge::cli::connect::command: Failed to change ownership of /etc/tedge/device-certs/tedge-private-key.pem to mosquitto:mosquitto: The path is not accessible. "/etc/tedge/device-certs/tedge-private-key.pem"
Restarting mosquitto... ✓
Waiting for mosquitto to be listening for connections... ✓
Verifying device is connected to cloud... ✓
Enabling tedge-mapper-c8y... ✓
Enabling tedge-agent... ✓
```

**To Reproduce**

1. Configure thin-edge.io to use basic auth when connecting to Cumulocity

    ```sh
    sudo tedge config set c8y.auth_method basic
    cat <<EOT > "$(tedge config get c8y.credentials_path)"
    [c8y]
    username = "t12345/device_rmi_test1001"
    password = "<some_password>"
    EOT
    sudo tedge config set device.id rmi_test1001
    ```

2. Connect to c8y

    ```sh
    sudo tedge connect c8y
    ```

**Expected behavior**

The following log messages should not be present in the output as it is expected that the certificate files do not exist when using basic auth.

```log
[...]

PR #3479, opened 2025-03-17
fix: skip chown for cert and key when using Basic Auth

Ruadhri17:
## Proposed changes
Skip changing ownership for device certificate as it is not needed when connecting via Basic Auth credentials.

## Types of changes

## Paste Link to the issue
#3476
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/connect/command.rs
+++ b/crates/core/tedge/src/cli/connect/command.rs
@@ -1022,6 +1022,11 @@ async fn new_bridge(
 }
 
 pub async fn chown_certificate_and_key(bridge_config: &BridgeConfig) {
+    // Skip chown when using Basic Auth
+    if bridge_config.auth_type == AuthType::Basic {
+        return;
+    }
+
     let (user, group) = match bridge_config.bridge_location {
         BridgeLocation::BuiltIn => ("tedge", "tedge"),
         BridgeLocation::Mosquitto => (crate::BROKER_USER, crate::BROKER_GROUP),
--- a/crates/common/tedge_config/src/tedge_toml/models/auth_method.rs
+++ b/crates/common/tedge_config/src/tedge_toml/models/auth_method.rs
@@ -34,7 +34,7 @@ impl FromStr for AuthMethod {
     }
 }
 
-#[derive(PartialEq, Eq)]
+#[derive(PartialEq, Eq, Debug, Clone, Copy, Display)]
 pub enum AuthType {
     Certificate,
     Basic,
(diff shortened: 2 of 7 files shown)


crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -991,6 +992,14 @@ fn new_bridge(
 }
 
 pub fn chown_certificate_and_key(bridge_config: &BridgeConfig) {
+    // Skip chown when using Basic Auth
+    if bridge_config
+        .auth_method
+        .is_some_and(|auth_method| auth_method == AuthMethod::Basic)
```
rina23q, line 998: Actually, this condition doesn't cover when `auth_method == AuthMethod::Auto` and auto chooses basic. You could use `auth_method.is_basic(<path_to_credentials_toml>)` instead, however, you need to parse the credentials file path, which is not available in this function. 

Alternatively, if both `remote_username` and `remote_password` are set, you can consider that auth method is basic. 

```suggestion
    if bridge_config.remote_username.is_some() && bridge_config.remote_password.is_some() 
```

Ruadhri17: When [Bridge Config](https://github.com/thin-edge/thin-edge.io/blob/8c2f74f9f9759982c260b4870ce0a734a0b34233/crates/core/tedge/src/bridge/c8y.rs#L151) is created, based on `remote_username` it is either `AuthMethod::Certificate` or `AuthMethod::Basic`. For `aws` and `azure` it is set to `None`.

rina23q: > When [Bridge Config](https://github.com/thin-edge/thin-edge.io/blob/8c2f74f9f9759982c260b4870ce0a734a0b34233/crates/core/tedge/src/bridge/c8y.rs#L151) is created, based on remote_username it is either AuthMethod::Certificate or AuthMethod::Basic. For aws and azure it is set to None.

This is correct. So, just to fix the issue, your change is enough.

However, there is a mess in the code. `BridgeConfig::auth_method` should be `Option<AuthType>`, `AuthType` has only `Basic` or `Certificate`. Furthermore, what is `None` mean? AWS/Azure, we use certificate based authentication. So, for me, even `BridgeConfig::auth_method` should be no longer Option, just `AuthType`.

didier-wenzek: As the condition seems no so obvious, it would be better to extract it into a `BridgeConfig` method.

rina23q (approved):
Thank you!

Issue #3475, opened 2025-03-15
c8y-remote-access-plugin broken on main branch


reubenmiller:
**Describe the bug**

The c8y-remote-access-plugin on the main branch is broken and any attempt to connect via WebSSH to the device fails and the operation failure reason is not very helpful, as it only shows ```note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace```.

But when tedge downgraded to 1.4.2, then the c8y-remote-access-plugin functions as expected and the WebSSH works again.

The workflow log is however more helpful (from the failed attempt with the latest version from the main branch, `1.4.3~266+gf4ee98b`):

```sh
root@tedge:/var/log/tedge/agent# cat c8y_RemoteAccessConnect-2025-03-15T09:21:36.906024766Z.log
----- $ c8y-remote-access-plugin "530,rmi_test1001,127.0.0.1,22,0ec273b9-00b0-40aa-953d-640b680bfaea"
Exit status: 101 (ERROR)

stderr <<EOF
sock: Connected to Unix socket at /run/c8y-remote-access-plugin.sock
sock: Writing message (530,rmi_test1001,127.0.0.1,22,0ec273b9-00b0-40aa-953d-640b680bfaea) to socket
sock: Message sent
sock: Shutting down writing on the stream, waiting for response...
sock: Shut down successful
sock: Reading response...
2025-03-15T09:21:36.922104493Z  INFO mqtt_channel::connection: MQTT connecting to broker: host=127.0.0.1:8883, session_name=None    
2025-03-15T09:21:36.967091954Z  INFO mqtt_channel::connection: MQTT connection established    
2025-03-15T09:21:36.987835558Z  INFO c8y_api::http_proxy: JWT token requested
2025-03-15T09:21:37.026258082Z  INFO c8y_api::http_proxy: JWT token received
2025-03-15T09:21:37.028113892Z  INFO mqtt_channel::connection: MQTT connection closed    
thread 'main' panicked at crates/core/tedge/src/main.rs:57:58:
called `Result::unwrap()` on an `Err` value:   × Connecting to Websocket
  ├─▶ host wss://thin-edge-io.eu-latest.cumulocity.com/service/remoteaccess/
  │   device/0ec273b9-00b0-40aa-953d-640b680bfaea
  ├─▶ WebSocket protocol error: SubProtocol error: Server sent a subprotocol
  │   but none was requested
  ╰─▶ SubProtocol error: Server sent a subprotocol but none was requested

[...]

PR #3480, opened 2025-03-18
test: add rust tests for c8y-remote-access-plugin

jarhodes314:
## Proposed changes
Tests the c8y-remote-access-plugin in rust, so we have a basic sanity check that it is behaving as expected against a simulated Cumulocity. There are a couple of tests, one checks that we cope with the mandatory sub-protocol `binary` enforced by Cumulocity (the root cause of #3475), and one that tests the basic functionality of the proxy using an axum-based websocket server. We still primarily rely on robot tests to ensure the remote access works as intended in practice, but this will ensure we have _some_ test coverage if we have to temporarily disable tests due to Cumulocity platform issues.

Also updates direct usage of base64 to the latest version, and replace the dependency on `sha-1` with a dependency on `sha1`, since they are identical. The base64 upgrade has led to the JWT decoding becoming more strict since the code as of this PR disallows padding (aka trailing '=' characters), but the JWT spec specifies that the encoding should be ["with all trailing '=' characters omitted"](https://datatracker.ietf.org/doc/html/rfc7515#section-2). It also switches the code to use the URL-safe base64 alphabet, as the spec also requires use of that.

## Types of changes

## Paste Link to the issue
- #3475

## Checklist


Cargo.toml
```diff
@@ -168,7 +168,8 @@ rustls-pemfile = "2.2"
 serde = "1.0"
 serde_ignored = "0.1"
 serde_json = "1.0"
-sha-1 = "0.10"
+serial_test = "3.2"
```
rina23q, line 171: It was removed by #3482 and revoked here accidentally?

rina23q (approved):

Cargo.toml
jarhodes314: Good catch, fixed now

PR #3500, opened 2025-03-25
chore: Add missing `format!` and resolve trivial clippy 1.87 lints

Bravo555:
## Proposed changes

Adds a missing `format!` in one command error to print the command line and resolve trivial clippy 1.78 lints.

The suggested fixes use version 1.74, our MSRV is 1.78 so it should be okay.

There are two categories of lints left to address:

- ~~`large_enum_variant`: this is in `tedge_config_macros`, and probably only has very very small impact on macro compilation speed, so we can most likely ignore it and apply `#[allow]`, but defering to @jarhodes314 as the author.~~
  Added `#[allow]` because it's low impact macro code, see https://github.com/thin-edge/thin-edge.io/pull/3500#issuecomment-2751082148
- `result_large_err`: there's loads and loads of instances of this lint being triggered by the amount of big error enums that we have (by both byte-size and number of variants). Lower level errors being included directly in high-level errors instead of via an indirection contribute to this, so while we could probably `Box` the selected number of big low-level errors to fix this, reducing the number of variants of some of these enums would probably also have some value in making the code simpler and easier to maintain.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config_macros/impl/src/input/parse.rs
+++ b/crates/common/tedge_config_macros/impl/src/input/parse.rs
@@ -87,6 +87,7 @@ fn not_tedge_config(attr: &syn::Attribute) -> bool {
     !is_tedge_config
 }
 
+#[allow(clippy::large_enum_variant)] // macro code, low impact
 #[derive(Debug)]
 pub enum FieldOrGroup {
     Field(ConfigurableField),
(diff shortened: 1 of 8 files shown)


jarhodes314:
> large_enum_variant: this is in tedge_config_macros, and probably only has very very small impact on macro compilation speed, so we can most likely ignore it and apply #[allow], but defering to @jarhodes314 as the author.

@Bravo555 I think `#[allow(...)]` is a perfectly sensible option here, I'm not concerned about the macro using a tiny bit more memory than it plausibly could

rina23q (approved):

PR #3503, opened 2025-03-26
refactor: Shorten entity store service REST API paths

albinsuresh:
## Proposed changes

Drop the `/entity-store` prefix:
`/tedge/entity-store/v1/entities -> /tedge/v1/entities`

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


didier-wenzek (commented):
API version sounds inconsistent now, with `/tedge/v1/entities` which includes a version vs `/tedge/file-transfer` which doesn't.

albinsuresh:
> API version sounds inconsistent now, with `/tedge/v1/entities` which includes a version vs `/tedge/file-transfer` which doesn't.

To keep them consistent, we can have a new versioned route like `/tedge/v1/files` pointing to the same handlers that serve the legacy `/tedge/file-transfer` route as well. Chose `/files` here instead of `/file-transfer` to be consistent with the entity-store APIs having the resource name in the path rather than the service name. To make it more explicit that the legacy path is deprecated, we can add a `Deprecated` header to the legacy responses as well.

@reubenmiller Since you suggested redirect as another option, yes we could do that as well, as it conveys the path migration to the users very explicitly. But I'm wondering if getting redirection responses instead of the actual responses would still be annoying to the users expecting the actual response, as not all tools automatically follow the redirected paths by default. Having both legacy and new routes pointing to the same handler seems less invasive from the user's perspective. 

So which one should we go with? Or rather, which one is the standard(community accepted) way to handle deprecation?

reubenmiller:
> > API version sounds inconsistent now, with `/tedge/v1/entities` which includes a version vs `/tedge/file-transfer` which doesn't.
> 
> To keep them consistent, we can have a new versioned route like `/tedge/v1/files` pointing to the same handlers that serve the legacy `/tedge/file-transfer` route as well. Chose `/files` here instead of `/file-transfer` to be consistent with the entity-store APIs having the resource name in the path rather than the service name. To make it more explicit that the legacy path is deprecated, we can add a `Deprecated` header to the legacy responses as well.
> 
> @reubenmiller Since you suggested redirect as another option, yes we could do that as well, as it conveys the path migration to the users very explicitly. But I'm wondering if getting redirection responses instead of the actual responses would still be annoying to the users expecting the actual response, as not all tools automatically follow the redirected paths by default. Having both legacy and new routes pointing to the same handler seems less invasive from the user's perspective.
> 
> So which one should we go with? Or rather, which one is the standard(community accepted) way to handle deprecation?

If the implementation effort is the same, then avoid the redirect (as you said if someone is using curl already, then it does not follow redirects by default, so it could break things).

albinsuresh:
> Convert main device alarm topic

Resolved by [e60ecd7](https://github.com/thin-edge/thin-edge.io/pull/3503/commits/e60ecd7f398a14f0a8806e7a5483d07b3fdcfaed). The `Convert main device alarm topic` test failure is unrelated (flaky).

didier-wenzek (commented):
In order to fully deprecate `/tedge-filetransfert`, many points are missing:

- the examples given by`tedge http --help`
- the documentation
- the internal uses (notably in the c8y mapper)

Better to address at least the first two points in this PR.

didier-wenzek:
> In order to fully deprecate `/tedge-filetransfert`, many points are missing:
> 
>     * the examples given by`tedge http --help`
> 
>     * the documentation
> 
>     * the internal uses (notably in the c8y mapper)
> 
> 
> Better to address at least the first two points in this PR.

Apologize, it seems that I reviewed an old version as these points have been correctly addressed.

didier-wenzek (approved):
Appoved

PR #3510, opened 2025-03-26
ci: parameterize the tarball contents

reubenmiller:
## Proposed changes

Fix inconsistency in the building of the tarball which assumes that `tedge` is the only binary that should be pacakged. In addition the existing code was very misleading as a list of packages were passed to it, but these were being ignored by the function.

Instead, just specify which binary should be included as part of the tarball (though it is still limited to 1 binary)

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/ci/build_scripts/package.sh
+++ b/ci/build_scripts/package.sh
@@ -234,6 +234,7 @@ get_package_arch() {
 build_tarball() {
     local name="$1"
     local target="$2"
+    local binary="$3"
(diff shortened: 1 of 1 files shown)


rina23q (approved):
> In addition the existing code was very misleading as a list of packages were passed to it, but these were being ignored by the function.

Good catch, the third argument (a list of package) was not used at all...

Issue #3506, opened 2025-03-26
tedge-p11-server usability improvements


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Further development of the tedge-p11-server component which adds PKCS11 support usage of keys stored in HSM.

**Describe the solution you'd like**

The following items could be improved:

#### Item 1: Log an error if the module does not exist or is not configured

If the module path is configured but does not exist, or the module path is unset, then `tedge-p11-server` should log an error, then exit with a non-zero exit code.

Currently there is no warning presented to the user, so they don't know that it is misconfigured.

```log
Mar 25 15:26:13 rpi5-d83addab8e9f systemd[1]: Started tedge-p11-server.service - tedge-p11-server.
Mar 25 15:26:13 rpi5-d83addab8e9f tedge-p11-server[1946135]: 2025-03-25T15:26:13.806406Z  INFO tedge_p11_server: crates/extensions/tedge-p11-server/src/main.rs:67: Using cryptoki configuration cryptoki_config=CryptokiConfigDirect { module_path: "", pin: "", serial: None }
Mar 25 15:26:13 rpi5-d83addab8e9f tedge-p11-server[1946135]: 2025-03-25T15:26:13.806447Z  INFO tedge_p11_server: crates/extensions/tedge-p11-server/src/main.rs:84: Server listening listener=Some("/run/tedge-p11-server/tedge-p11-server.sock")
```

#### Item 2: add log-level flag to tedge-p11-server

In line with other tedge cli commands, the `tedge-p11-server` should supported the `--log-level <level>` flag.

```sh
tedge-p11-server --log-level info
```

#### Item 3: tedge connect - add information that the key is being used via the `device.cryptoki.mode` setting

At the moment, it is not obvious to the user that they are using the cryptoki/pkcs11 feature.

```sh
$ tedge reconnect c8y
Disconnecting from Cumulocity
Removing bridge config file... ✓
Bridge doesn't exist. Device is already disconnected from Cumulocity.
Reconnecting with config:
	device id: rpi5-d83addab8e9f_hsm
	cloud profile: <none>
	cloud host: thin-edge-io.eu-latest.cumulocity.com:8883
	auth type: Certificate
	certificate file: /etc/tedge/device-certs/tedge-certificate.pem
	bridge: built-in
	service manager: tedgectl
	mosquitto version: 2.0.18
Creating device in Cumulocity cloud... ✓
Restarting mosquitto... ✓
[...]

PR #3511, opened 2025-03-26
`tedge-p11-server` improvements:  Try to load cryptoki lib immediately, `--log-level`, cryptoki connect config summary

Bravo555:
## Proposed changes

1. If the cryptoki dynamic module doesn't exist or loading it causes some other error, then we exit with an error immediately since we can't handle any requests.
    
    Notably, when module path is wrong, the process exits with the following message:
    
    ```
    Error: Failed to start the signing service
    
    Caused by:
        0: Failed to get handle of pkcs11 signing key
        1: Failed to load PKCS#11 dynamic object
        2: /lib/x86_64-linux-gnu/libykcs.so: cannot open shared object file: No such file or directory
    ```

2. Add `--log-level` flag to `tedge-p11-server`. 
3. When connecting/reconnecting, we now display whether or not cryptoki is used in config summary (example):
    
    ```
    Reconnecting with config:
            device id: marcel-hsm-device
            cloud profile: <none>
            cloud host: thin-edge-io.eu-latest.cumulocity.com:8883
            auth type: Certificate
            bridge: built-in
            service manager: systemd
            mosquitto version: 2.0.20
            cryptoki: true
    ```
    
    Currently this is only a boolean value, to show more information here would require some more changes in the bridge configuration structs.

## Types of changes

## Paste Link to the issue
#3506 

## Checklist


crates/core/tedge/src/cli/log.rs
```diff
@@ -319,6 +321,7 @@ impl fmt::Display for ConfigLogger<'_> {
         if let Some(mosquitto_version) = self.mosquitto_version {
             self.log_single_entry(f, "mosquitto version", &mosquitto_version)?;
         }
+        self.log_single_entry(f, "cryptoki", &self.cryptoki)?;
```
didier-wenzek, line 324: It would be better to display this close to the auth type. And, even better, only when the `auth type` is  `Certificate`.

Bravo555: fixed in 8f8a82a8b5

didier-wenzek (approved):
Approved.

However an error message could still be improved. When trying to reconnect after a first failed attempt (because the module path was incorrect), one got an error related to a socket still in use with not clue on what this socket is and not hint on what can be the next step.  

```
$ tedge-p11-server --module-path /second/attempt --log-level DEBUG

2025-03-28T09:27:44.621915Z DEBUG tedge_p11_server: crates/extensions/tedge-p11-server/src/main.rs:88: No sockets from systemd, creating a standalone socket
Error: Failed to bind to socket

Caused by:
    Address already in use (os error 98)
```

Bravo555:
> However an error message could still be improved. When trying to reconnect after a first failed attempt (because the module path was incorrect), one got an error related to a socket still in use with not clue on what this socket is and not hint on what can be the next step.

This will be a part of a next PR which adds proper removal of the socket after server shutdown.

Bravo555:
> However an error message could still be improved. When trying to reconnect after a first failed attempt (because the module path was incorrect), one got an error related to a socket still in use with not clue on what this socket is and not hint on what can be the next step.

Fixed in https://github.com/thin-edge/thin-edge.io/pull/3519.

PR #3514, opened 2025-03-26
feat: tedge-p11-server packaging

reubenmiller:
## Proposed changes

Package tedge-p11-server as linux packages and a tarball. The tedge-p11-server binary must not be compiled with the MUSL target as it relies on dynamically loading libraries (e.g. .so or .dylib files) which MUSL targets do not support, however the required build changes which ensures that the tedge-p11-server is compiled correctly was done in https://github.com/thin-edge/thin-edge.io/pull/3529.

A system test was added to test the tedge-p11-server interaction with thin-edge.io by using softhsm2 (which is a PKCS11 compliant HSM which allows testing of the PKCS11 interface without any additional hardware).

Minor changes:
* Changed the default socket path used by tedge and tedge-p11-server to match the systemd socket path

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/configuration/package_scripts/tedge-p11-server/preinst
+++ b/configuration/package_scripts/tedge-p11-server/preinst
@@ -1,0 +1,55 @@
+#!/bin/sh
+set -e
+
+command_exists() {
+    command -V "$1" >/dev/null 2>&1
+}
+
+group_exists() {
+    name="$1"
+    if command_exists id; then
+        id -g "$name" >/dev/null 2>&1
+    elif command_exists getent; then
+        getent group "$name" >/dev/null 2>&1
+    else
+        # Fallback to plain grep, as busybox does not have getent
+        grep -q "^${name}:" /etc/group
+    fi
+}
+
+user_exists() {
+    name="$1"
+    if command_exists id; then
+        id -u "$name" >/dev/null 2>&1
+    elif command_exists getent; then
+        getent passwd "$name" >/dev/null 2>&1
+    else
+        # Fallback to plain grep, as busybox does not have getent
+        grep -q "^${name}:" /etc/passwd
+    fi
+}
+
+### Create groups
+if ! group_exists tedge; then
+    if command_exists groupadd; then
+        groupadd --system tedge
+    elif command_exists addgroup; then
+        addgroup -S tedge
+    else
+        echo "WARNING: Could not create group: tedge" >&2
+    fi
+fi
+
+### Create users
+# Create user tedge with no home(--no-create-home), no login(--shell) and in group tedge(--gid)
+if ! user_exists tedge; then
+    if command_exists useradd; then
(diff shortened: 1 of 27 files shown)


albinsuresh (commented):
Though everything looked fine to me, let someone else more familiar with the subject approve the PR.

tests/images/debian-systemd/debian-systemd.dockerfile
```diff
@@ -75,6 +78,9 @@ COPY files/http-server/nginx.conf /etc/nginx/nginx.conf
 COPY files/http-server/*.sh /usr/bin/
 RUN systemctl disable nginx
 
+# HSM
+# RUN usermod -a -G softhsm tedge
```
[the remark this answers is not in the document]
reubenmiller: Yes, good catch, and fixed in 2cfee4c77871ada9836283f24167f9103c2fcdab.

I'd hoped to add it to the test setup, but that didn't work because the tedge user does not exist during the container image build phase...and I didn't want to add it there as we have plenty of "fresh installation" tests which I didn't want to influence.

configuration/package_scripts/tedge-p11-server/preinst
```diff
@@ -0,0 +1,55 @@
...
+    else
+        echo "WARNING: Could not create group: tedge" >&2
+    fi
+fi
+
+### Create users
+# Create user tedge with no home(--no-create-home), no login(--shell) and in group tedge(--gid)
+if ! user_exists tedge; then
+    if command_exists useradd; then
+        useradd -u "999" --system --no-create-home --shell /sbin/nologin --gid tedge tedge
```
didier-wenzek, line 47: I guess that using a well-know user and group id (aka  `992` and `999`) is to ease shared-access to the socket across containers. However, here this sounds more as magic numbers, as these are documented no where.

reubenmiller: Actually, after thinking about it a bit more, I'll remove the fixed assignment, and just make it a pre-requisite for users to setup before the package is installed (which can be documented in the future PR doc).

Bravo555: Also, tedge-p11-server is supposed to be installed on a host, so we have to create user and group here as we do in the preinst script for tedge, but don't we also have to make sure the ids match for access control?

reubenmiller: Changed in 084b00ff0f6eb400fdb7a33cb4ae94a85c950c51

The usage of explicit GID and UID has been removed to avoid conflicts where the GID/UID are already assigned to different users. The downside is that the user MUST now configure a static GID/UID before using it in any container environments (but this is generally done anyway on the image building layer, so shouldn't be a big barrier).

Bravo555 (approved):
LGTM

Issue #3394, opened 2025-02-11
Config option values containing only digits passed via env variables are incorrectly parsed as a number when they are a string


Bravo555:
**Describe the bug**

Executing `TEDGE_HTTP_CLIENT_HOST="123456" tedge config get http.client.host` results in a following error:

```
Error: missing configuration parameter

Caused by:
    invalid type: found unsigned int `123456`, expected a string for key "HTTP.CLIENT.HOST" in TEDGE_HTTP_CLIENT_HOST environment variable
```

Value coming from an env variable is attempted to be parsed as an int, even though the type of `http.client.host` is actually `Arc<str>`.

By contrast, setting the value via `tedge config set http.client.host 123456` works:

```
$ tedge config set http.client.host 123456
$ tedge config get http.client.host
123456
```

**To Reproduce**

See above.

**Expected behavior**

The env variable value should not be eagerly parsed as an int and should be passed as a string.

**Screenshots**

**Additional context**

The `http.client.host` is not a very good example, because generally, hostnames shouldn't consists of only digits.
However, in #3366 a new `device.cryptoki.pin` option is added that contains a PIN value for the HSM. This option MUST be a string, because if parsed as an int, leading zeroes would be stripped, so a correct pin "012345" would be parsed as "12345", not to say it would also probably be parsed as an octal value.

I'm not marking it as a bug, since it's very unlikely anybody has been affected by this and one would have to do some objectionable things to trigger this (like setting hostnames consisting of only digits and leading zeroes).

Bravo555:
Note: there is a workaround - it works when the env variable contains escaped doublequotes:
```
TEDGE_HTTP_CLIENT_HOST=\"012345\" tedge config get http.client.host
012345
```

PR #3516, opened 2025-03-28
feat: read environment variables with leading zeroes as strings

jarhodes314:
**Superseded by #3518, unless I discover some critical issue with that better approach**

## Proposed changes
Captures environment variables containing leading zeroes losslessly. When specifying pins for cryptoki, the customer may wish to configure a value with a leading zero, and may wish to do this via an environment variable. Currently, figment will attempt to parse the value to an integer, and if that succeeds, will parse the value as an integer. This doesn't stop us deserialising the value to a string (though we will need a custom type for that), but if the number has one or more leading zeroes, those will be lost in the conversion. This PR changes the deserialisation of the environment variables to parse as strings if they are an integer with a leading zero.

Technically, this could be seen as a breaking change, but in a practical sense it shouldn't be.

*TODO before merging*: add proptest to this, this feature is a great candidate for property-based testing, and I suspect I've probably overlooked something.

## Types of changes

## Paste Link to the issue
- #3394

## Checklist

## Further comments

In an ideal world, serde would be able to look at the target type we are deserialising to, notice that it's a string, and copy the value verbatim without any numeric parsing. Unfortunately, serde's design requires the value we're deserialising from to dictate what "type" it is (string, u64, etc.). This great for formats like TOML where the different types have unambiguous syntax, but for environment variables (and formats like CSV), this is a pain. We also [can't fine-tune our deserialisation logic depending on where a value came from](https://stackoverflow.com/a/69691696). However, we can create a type that can deserialise from multiple possible source types, like `TemplatesSet` (which parses `TEDGE_C8Y_SMARTREST_TEMPLATES=test,value` to `["test", "value"]` just fine). This should enable the cryptoki pin to be parsed via a newtype wrapper that can be deserialised from strings or integers.

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/figment.rs
+++ b/crates/common/tedge_config/src/tedge_toml/figment.rs
@@ -7,9 +7,15 @@ use std::sync::Mutex;
 
 use figment::providers::Format;
 use figment::providers::Toml;
+use figment::util::nest;
+use figment::value::Dict;
+use figment::value::Map;
 use figment::value::Uncased;
+use figment::value::Value;
 use figment::Figment;
 use figment::Metadata;
+use figment::Profile;
+use figment::Provider;
 use once_cell::sync::Lazy;
 use serde::de::DeserializeOwned;
 
@@ -207,10 +213,10 @@ impl TEdgeEnv {
         })
     }
 
-    fn provider(&self) -> figment::providers::Env {
+    fn provider(&self) -> TEdgeEnvProvider {
         let prefix = self.prefix;
         static WARNINGS: Lazy<Mutex<HashSet<String>>> = Lazy::new(<_>::default);
-        figment::providers::Env::prefixed(self.prefix).ignore(&["CONFIG_DIR", "CLOUD_PROFILE"])
+        let provider = figment::providers::Env::prefixed(self.prefix).ignore(&["CONFIG_DIR", "CLOUD_PROFILE"])
         .filter(move |key| {
             std::env::vars()
             .find(|(k, _)| k.strip_prefix(prefix).is_some_and(|k| k == key))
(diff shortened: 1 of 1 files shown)


reubenmiller:
@jarhodes314 Maybe I'm missing something...you're basically saying that an env variable `0123456` would be then serialized into `123456` (only because it is starting with a zero)? If this is correct, then I don't think it is a viable option, as in the pin's case, we're talking about a PIN being provided in the exact form that it needed (because it is like a password). It would be a strange UX to expect the PIN's value to altered just to suite some questionable design decision in figment

jarhodes314:
> @jarhodes314 Maybe I'm missing something...you're basically saying that an env variable `0123456` would be then serialized into `123456` (only because it is starting with a zero)? If this is correct, then I don't think it is a viable option, as in the pin's case, we're talking about a PIN being provided in the exact form that it needed (because it is like a password). It would be a strange UX to expect the PIN's value to altered just to suite some questionable design decision in figment

I think I explained things badly. What you've discussed there is the default behaviour we get from figment, so what happens with the current tedge version. With this change, `0123456` would be deserialised as `"0123456"`. `123456` will still be parsed as an integer, but we can convert that back to a string quite happily without any issues.

Bravo555:
> you're basically saying that an env variable `0123456` would be then serialized into `123456` (only because it is starting with a zero)? If this is correct, then I don't think it is a viable option

My impression is that would be the default behaviour if we use figment as-is, but the PR is changing it to make it so that now, when parsing env variables that are not delimited as strings (with `"`) numbers with leading zeros are parsed as strings.

previously:
- `TEDGE_DEVICE_CRYPTOKI_PIN=\"012345\"` -> String("0123456")
- `TEDGE_DEVICE_CRYPTOKI_PIN=012345` -> number(12345)
- `TEDGE_DEVICE_CRYPTOKI_PIN=123450` -> number(123450)

now:
- `TEDGE_DEVICE_CRYPTOKI_PIN=\"012345\"` -> String("0123456")
- `TEDGE_DEVICE_CRYPTOKI_PIN=012345` -> String("012345")
- `TEDGE_DEVICE_CRYPTOKI_PIN=123450` -> number(123450)

Ideally, we'd want the 3rd case to parse as string as well, but only for `TEDGE_DEVICE_CRYPTOKI_PIN=123450`, but how parser works is global for all values. In the case here, we're fine with both string and number, since we can convert number into a string with no errors, but for other values in the config that are intended to be numbers, this can break something. AFAIK converting 2nd case is fine because they don't really get leading zero number values in practice (that's I assume what _Technically, this could be seen as a breaking change, but in a practical sense it shouldn't be_ comment is referring to).

reubenmiller:
> > @jarhodes314 Maybe I'm missing something...you're basically saying that an env variable `0123456` would be then serialized into `123456` (only because it is starting with a zero)? If this is correct, then I don't think it is a viable option, as in the pin's case, we're talking about a PIN being provided in the exact form that it needed (because it is like a password). It would be a strange UX to expect the PIN's value to altered just to suite some questionable design decision in figment
> 
> @reubenmiller I think I explained things badly. What you've discussed there is the default behaviour we get from figment, so what happens with the current tedge version. With this change, `0123456` would be deserialised as `"0123456"`. `123456` will still be parsed as an integer, but we can convert that back to a string quite happily without any issues.
> 
> Where this would cause a breakage is if someone specified their MQTT port as `01883` rather than `1883`. I don't expect anyone is doing that currently, and it's an easy fix if they are. It would possibly also attempt to parse the value as octal, which would be unlikely to be what any.usrr actually wants.

Awesome :) Right so you were just talking about the internal env variable to Rust type conversions.

crates/common/tedge_config/src/tedge_toml/figment.rs
```diff
@@ -269,6 +340,90 @@ mod tests {
...
+                extract_data::<Config, FileAndEnvironment>(&PathBuf::from("tedge.toml"))
+                    .unwrap()
+                    .0
+                    .value,
+                0.123,
+            );
+            Ok(())
+        })
+    }
+
```
Bravo555, line 426: suggestion: Would it be possible to have a test when we deserialize to both a number and to a string from all the 3 cases from my comment: https://github.com/thin-edge/thin-edge.io/pull/3516#issuecomment-2761494136 ?

The changes look roughly correct, but I still have some trouble understanding the implication of this change for consumers that deserialize sequences of digits into strings (`device.cryptoki.pin`) or into proper numbers.
> Technically, this could be seen as a breaking change, but in a practical sense it shouldn't be
This is one question I had, does this change result in a thing that was previously possible now being an error? It would be nice to have a confirmation, if any, in a form of a test.

PR #3518, opened 2025-03-28
feat: parse `TEDGE_` environment variables using `FromStr` implementation

jarhodes314:
## Proposed changes
Supersedes #3516 

This is an attempt at a better solution to the lossy parsing we currently encounter when processing `TEDGE_` environment variables. Since we don't need to process environment variables via serde now (I believe we did when figment was introduced pre-macro), serde/figment are proving a hinderance for processing the variables as we discovered with #3394.

The changes (will) remove figment as a dependency, and move to a model where we parse `tedge.toml` to a DTO, then insert the values from the environment variables into the DTO (parsing them in the same way `tedge config set` does). Unlike `tedge config set`, this will also accept TOML-formatted values for backwards compatibility with the current figment behaviour.

The code changes are still quite a way from being finished at the time of opening this PR, but ignoring error cases/warnings, this I believe has all the intended behaviour, so can be manually tested. If you do specify an unknown environment variable, this will currently result in a hard error, whereas it will ultimately result in a warning like we get from figment currently.

### Examples
Here are some examples of how the new solution works, showing how it handles leading zeroes intelligently depending on the target key, and how it parses TOML syntax like figment did.

```bash
$ TEDGE_MQTT_CLIENT_PORT=1883 tedge config get mqtt.client.port
1883 # simple case, works as before
$ TEDGE_MQTT_CLIENT_PORT=01883 tedge config get mqtt.client.port
1883 # leading zero, works as before
$ TEDGE_HTTP_CLIENT_HOST=012345 tedge config get http.client.host
012345 # parsed as a string, previously this was rejected
$ TEDGE_C8Y_SMARTREST_TEMPLATES=test,values tedge config get c8y.smartrest.templates
["test","values"] # custom FromStr implementation, works as before
$ TEDGE_C8Y_SMARTREST_TEMPLATES=[\"test\",\"values\"] tedge config get c8y.smartrest.templates
["test","values"] # TOML syntax in environment variable, works as before
$ TEDGE_C8Y_SMARTREST_TEMPLATES=[\"single,value,with,comma\"] tedge config get c8y.smartrest.templates
["single,value,with,comma"] # TOML syntax, works as before
```

### Minor breaking change
This is again technically a breaking change, but not really in practice. Previously, the environment variables overrode the `tedge.toml` values before deserialisation occurred, which could cause an invalid `tedge.toml` to be masked by environment variables. Thus the following would previously have worked:

```bash
$ cat /etc/tedge/tedge.toml
mqtt.client.port = "not a number"
$ TEDGE_MQTT_CLIENT_PORT=1883 tedge config get mqtt.client.port
1883
```

With this change, this will (correctly) fail when parsing `tedge.toml`.

### TODOs
- [ ] Handle empty environment variables
- [ ] Tests
- [ ] Preserve the existing warning behaviour
- [ ] Remove figment now it's not needed
- [ ] Proptest? - probably should do this, I suspect there are things we can do that aren't that difficult

## Types of changes

## Paste Link to the issue
- #3394

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
@@ -9,6 +9,7 @@ use crate::TEdgeConfig;
 use crate::TEdgeConfigDto;
 use crate::TEdgeConfigError;
 use crate::TEdgeConfigReader;
+use anyhow::Context as _;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use serde::Serialize;
@@ -21,6 +22,8 @@ use tedge_utils::fs::atomically_write_file_async;
 use tedge_utils::fs::atomically_write_file_sync;
 use tracing::debug;
 use tracing::warn;
+
+use super::tedge_config;
 
 const DEFAULT_TEDGE_CONFIG_PATH: &str = "/etc/tedge";
 const ENV_TEDGE_CONFIG_DIR: &str = "TEDGE_CONFIG_DIR";
@@ -181,6 +184,31 @@ impl TEdgeConfigLocation {
                 // Reload DTO to get the settings in the right place
                 (dto, warnings) = super::figment::extract_data::<_, Sources>(self.toml_path())?;
             }
+        }
+
+        for (key, value) in std::env::vars() {
+            let writable_key = match key.strip_prefix("TEDGE_") {
+                Some("CONFIG_DIR") => continue,
+                Some("CLOUD_PROFILE") => continue,
+                // TODO failure to parse should be a warning
+                Some(key) => key
+                    .to_ascii_lowercase()
+                    .parse::<tedge_config::WritableKey>()
+                    .with_context(|| format!("Failed to parse `TEDGE_{key}`"))
+                    .unwrap(),
+                None => continue,
+            };
+            // TODO error context
+            if value.starts_with('"') || value.starts_with('[') {
+                if let Ok(mut tmp_dto) =
+                    toml::from_str::<TEdgeConfigDto>(&format!("{writable_key} = {value}"))
+                {
+                    // TODO error handling and testing
+                    dto.take_value_from(&mut tmp_dto, &writable_key).unwrap();
+                    continue;
+                }
+            }
+            dto.try_update_str(&writable_key, &value).unwrap()
         }
 
         Ok((dto, warnings))
--- a/crates/common/tedge_config/src/tedge_toml/figment.rs
+++ b/crates/common/tedge_config/src/tedge_toml/figment.rs
@@ -51,11 +51,11 @@ pub fn extract_data<T: DeserializeOwned, Sources: ConfigSources>(
     let env = TEdgeEnv::default();
     let figment = Figment::new().merge(Toml::file(path));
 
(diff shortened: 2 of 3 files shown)


crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
```diff
@@ -268,9 +277,132 @@ impl TEdgeConfigLocation {
...
+                            tedge_key.to_ascii_lowercase()
+                        ));
+                        continue;
+                    }
+                }
+            }
+            None => continue,
+        };
+
+        // TODO test these warnings are vaguely sensibly formatted
```
jarhodes314, line 336: I should have some unit tests that checks these warnings appear as intended. I believe the warnings could only ever appear in cases where we interact with profiles (which is unlikely to be seen in combination with environment variables, since env vars generally negate the need for profiles).

didier-wenzek:
I'm surprised by the following. I was expecting an error but got a warning and the environment variable has been ignored.

```
$ TEDGE_MQTT_CLIENT_PORT="port" tedge config get mqtt.client.port
2025-04-03T08:59:38.774718549Z  WARN tedge_config::tedge_toml::tedge_config_location: Failed to process TEDGE_MQTT_CLIENT_PORT: Failed to parse input
1883
```

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
```diff
@@ -1,16 +1,14 @@
 use std::path::Path;
 
-use crate::tedge_toml::figment::ConfigSources;
-use crate::tedge_toml::figment::FileAndEnvironment;
-use crate::tedge_toml::figment::FileOnly;
-use crate::tedge_toml::figment::UnusedValueWarnings;
+use crate::tedge_toml::DtoKey;
```
didier-wenzek, line 3: I don't get where this `DtoKey` is defined.

jarhodes314:
> I'm surprised by the following. I was expecting an error but got a warning and the environment variable has been ignored.
> 
> ```
> $ TEDGE_MQTT_CLIENT_PORT="port" tedge config get mqtt.client.port
> 2025-04-03T08:59:38.774718549Z  WARN tedge_config::tedge_toml::tedge_config_location: Failed to process TEDGE_MQTT_CLIENT_PORT: Failed to parse input
> 1883
> ```

@didier-wenzek I've now changed this to error:
```
$ TEDGE_MQTT_CLIENT_PORT="port" tedge config get mqtt.client.port
Error: Failed to set value for mqtt.client.port to "port" from environment variable TEDGE_MQTT_CLIENT_PORT

Caused by:
    0: Failed to parse input
    1: invalid digit found in string
```

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
jarhodes314: It's defined along with `WritableKey` etc in [query.rs](https://github.com/thin-edge/thin-edge.io/pull/3518/files#diff-cd95e9aea2dc9e842e4cd738eaf06c1e284ccab0754a684defe0851b188a7faeR53)

didier-wenzek (approved):
Approved. Nice to see reducing the number of dependencies.

PR #3529, opened 2025-04-03
ci: refactor build script and workflow

reubenmiller:
## Proposed changes

Motivation: Building for the new tedge-p11-server component adds some additional build challenges due having to build this component with gnu libc (not musl!).

Normalize the building of thin-edge.io so that both local developers and the build workflow use the same entrypoint to build via `just release`, though each environment can still control which tooling is used to actually build the required binaries (e.g. clang (current default for the workflow), and cargo-zigbuild (previous default when running locally).

There are provisions in the build.sh script which:

* Select gnu vs musl based on binary target (tedge will use the given target, where as tedge-p11-server will be forced to use the musl equivalent)
* Use cargo-zigbuild to control which gnu libc version is used when building for gnu libc targets (default glibc version is 2.17 for all targets except for riscv which is 2.27)
* All artifacts are combined into a single folder (currently based on the target given by the caller)

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/ci/package_list.sh
+++ b/ci/package_list.sh
@@ -12,8 +12,8 @@ RELEASE_PACKAGES=(
 )
 export RELEASE_PACKAGES
 
-# Deprecated packages are still built but not explicitly tested
-# This allows users to still access the packages if needed, however
-# it is only reserved for packages with a more public facing API
-DEPRECATED_PACKAGES=()
-export DEPRECATED_PACKAGES
+# List of binaries which should be built
+BINARIES=(
+    tedge
+)
+export BINARIES
--- a/.github/workflows/build-workflow.yml
+++ b/.github/workflows/build-workflow.yml
@@ -18,11 +18,6 @@ on:
         type: string
         required: false
         default: "10"
-      run_rust_tests:
-        description: Run Rust tests (generally this is not required as the PR check will run full Rust tests)
-        type: boolean
-        required: false
-        default: false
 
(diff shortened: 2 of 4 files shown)


ci/package_list.sh
```diff
@@ -12,8 +12,8 @@ RELEASE_PACKAGES=(
...
 export RELEASE_PACKAGES
 
-# Deprecated packages are still built but not explicitly tested
-# This allows users to still access the packages if needed, however
-# it is only reserved for packages with a more public facing API
-DEPRECATED_PACKAGES=()
-export DEPRECATED_PACKAGES
+# List of binaries which should be built
+BINARIES=(
+    tedge
```
rina23q, line 17: Why is `tedge-p11-server` not listed here? Is `tedge-p11-server` supposed to be released as one package?

rina23q (commented):
Looks good, however, to be honest, I am lost in understanding the changes in `build.sh`...

rina23q (approved):
I had an offline talk with @reubenmiller and was convinced of this change. Thanks for all the efforts.

ci/package_list.sh
reubenmiller: no, the tedge-p11-server will be added in the next PR which adds the additional packaging files (as it is useless just to build the tedge-p11-server binary without packaging it with system service/socket etc.)

Issue #3538, opened 2025-04-07
allow users to specify which pkcs11 private key to use by setting a uri


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When using a HSM module with the PKCS11 interface, the HSM module can contain multiple keys (either in the same PKCS11 module, or from different ones), so users should be able to direct thin-edge.io to which one should be used for a given certificate.

**Describe the solution you'd like**

thin-edge.io pkcs11 support show allow users to device which private key to use by specifying the PKCS11 URI via the tedge (client side) configuration file, tedge.toml.

* support specifying a PKCS11 URI for each private key (similar to the `*.key_path` properties, but called `key_uri)

Below shows an example of the private key used for all mappers.

**File: tedge.toml**

```toml
[device]
key_uri = "pkcs11:model=SoftHSM%20v2;manufacturer=SoftHSM%20project;token=tedge"
```

**Describe alternatives you've considered**

**Additional context**

PR #3544, opened 2025-04-08
feat: tedge-p11-server Select token using serial number

Bravo555:
## Proposed changes

Allow the user to select a token via a serial number. If a serial number is provided, we attempt to select a token with that number and fail if such token doesn't exist.

As of now this is implemented for tedge-p11-server, so works with `socket` cryptoki mode, but doesn't work yet with `module` mode.

## Types of changes

## Paste Link to the issue
- #3538

## Checklist

## Further comments

I'm aware that the issue asks for the URL and not serial numbers, but this was easy to do and should allow users to select which token to use if multiple are available, while supporting the [PKCS #11 URL scheme](https://www.rfc-editor.org/rfc/rfc7512.html) will involve more work, such as possibly writing a parser for these URLs.

### DIFF
--- a/configuration/init/systemd/tedge-p11-server.service
+++ b/configuration/init/systemd/tedge-p11-server.service
@@ -6,7 +6,7 @@ Requires=tedge-p11-server.socket
 Type=simple
 StandardError=journal
 EnvironmentFile=-/etc/tedge/plugins/tedge-p11-server.conf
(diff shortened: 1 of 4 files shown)


reubenmiller:
@Bravo555 It would be better just to accept a PKCS11 URI (https://github.com/thin-edge/thin-edge.io/issues/3538) as this is easier to manage as you can be as specific as you need to be (e.g. use the same setting in the entire fleet).

Bravo555:
Superseded by #3555.

PR #3555, opened 2025-04-10
feat: Key selection using PKCS #11 URI

Bravo555:
## TODO

- [x] PKCS11 URI scheme parsing
- [x] selection of token by serial number and label and private key by label
- [x] select private key by id (need to handle percent encoding correctly)
- [x] tests
- [x] ~~require~~ warn a user ~~to provide a URI that uniquely identifies key to use~~ if there are multiple tokens/keys and URI is either not provided or not sufficiently precise (https://github.com/thin-edge/thin-edge.io/pull/3555#discussion_r2047408293)

## Follow-up

- [ ] warn user if used URI points to an object that is not a private key but the private key exists in the token
- [ ] code cleanup 

## Proposed changes

Allows the user to denote private key that will be used by `tedge-p11-server` for signing by a new optional `--uri` parameter.

The PKCS #&NoBreak;11 URI is explained in [RFC 7512](https://www.rfc-editor.org/rfc/rfc7512.html). It can be used to identify any cryptoki resource by its properties. In our case it's used to select a key, but the URI given by the user doesn't have to identify a key, but can identify e.g a token which we can select and ~~find its first~~ if the token contains only one private key, use that key.

## Types of changes

## Paste Link to the issue

- #3538 

## Checklist

### DIFF
--- a/crates/extensions/tedge-p11-server/Cargo.toml
+++ b/crates/extensions/tedge-p11-server/Cargo.toml
@@ -15,6 +15,7 @@ asn1-rs.workspace = true
 camino.workspace = true
 clap.workspace = true
 cryptoki.workspace = true
+percent-encoding.workspace = true
 postcard.workspace = true
 rustls.workspace = true
 sd-listen-fds.workspace = true
--- a/crates/extensions/tedge-p11-server/src/pkcs11.rs
+++ b/crates/extensions/tedge-p11-server/src/pkcs11.rs
@@ -46,26 +46,17 @@ impl Debug for CryptokiConfigDirect {
     }
 }
 
-#[derive(Debug)]
-pub enum Pkcs11SigningKey {
-    Rsa(RSASigningKey),
-    Ecdsa(ECSigningKey),
+#[derive(Debug, Clone)]
+pub struct Cryptoki {
+    context: Pkcs11,
+    config: CryptokiConfigDirect,
 }
 
-impl Pkcs11SigningKey {
-    pub fn from_cryptoki_config(
-        cryptoki_config: &CryptokiConfigDirect,
-    ) -> anyhow::Result<Pkcs11SigningKey> {
-        let CryptokiConfigDirect {
-            module_path,
-            pin,
-            // TODO(marcel): select modules by serial if multiple are connected
-            serial: _,
-        } = cryptoki_config;
-
-        debug!(%module_path, "Loading PKCS#11 module");
+impl Cryptoki {
+    pub fn new(config: CryptokiConfigDirect) -> anyhow::Result<Self> {
+        debug!(module_path = %config.module_path, "Loading PKCS#11 module");
         // can fail with Pkcs11(GeneralError, GetFunctionList) if P11_KIT_SERVER_ADDRESS is wrong
-        let pkcs11client = match Pkcs11::new(module_path) {
+        let pkcs11client = match Pkcs11::new(&config.module_path) {
             Ok(p) => p,
             // i want to get inner error but i don't know if there is a better way to do this
             Err(Error::LibraryLoading(e)) => {
@@ -78,18 +69,33 @@ impl Pkcs11SigningKey {
 
         pkcs11client.initialize(CInitializeArgs::OsThreads)?;
 
+        Ok(Self {
+            context: pkcs11client,
+            config,
+        })
+    }
+
+    pub fn signing_key(&self) -> anyhow::Result<Pkcs11SigningKey> {
+        let CryptokiConfigDirect {
+            pin,
+            // TODO(marcel): select modules by serial if multiple are connected
+            serial: _,
+            ..
+        } = &self.config;
+
         // Select first available slot. If it's not the one we want, the token
         // used in p11-kit-server should be adjusted.
-        let slot = pkcs11client
+        let slot = self
+            .context
             .get_slots_with_token()?
             .into_iter()
             .next()
             .context("Didn't find a slot to use. The device may be disconnected.")?;
-        let slot_info = pkcs11client.get_slot_info(slot)?;
-        let token_info = pkcs11client.get_token_info(slot)?;
+        let slot_info = self.context.get_slot_info(slot)?;
+        let token_info = self.context.get_token_info(slot)?;
         debug!(?slot_info, ?token_info, "Selected slot");
 
-        let session = pkcs11client.open_ro_session(slot)?;
+        let session = self.context.open_ro_session(slot)?;
         session.login(UserType::User, Some(pin))?;
         let session_info = session.get_session_info()?;
         debug!(?session_info, "Opened a readonly session");
@@ -110,6 +116,12 @@ impl Pkcs11SigningKey {
 
         Ok(key)
     }
+}
+
+#[derive(Debug)]
+pub enum Pkcs11SigningKey {
+    Rsa(RSASigningKey),
+    Ecdsa(ECSigningKey),
 }
 
 impl SigningKey for Pkcs11SigningKey {
--- a/crates/extensions/tedge-p11-server/src/pkcs11.rs
+++ b/crates/extensions/tedge-p11-server/src/pkcs11.rs
@@ -4,6 +4,8 @@
 //! - thin-edge: docs/src/references/hsm-support.md
 //! - PKCS#11: https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/os/pkcs11-base-v2.40-os.html
 
+pub use cryptoki::types::AuthPin;
+
 use anyhow::Context;
 use asn1_rs::ToDer;
 use camino::Utf8PathBuf;
@@ -16,24 +18,24 @@ use cryptoki::object::AttributeType;
 use cryptoki::object::KeyType;
 use cryptoki::session::Session;
 use cryptoki::session::UserType;
-pub use cryptoki::types::AuthPin;
 use rustls::sign::Signer;
 use rustls::sign::SigningKey;
 use rustls::SignatureAlgorithm;
 use rustls::SignatureScheme;
+use tracing::debug;
 use tracing::trace;
+use tracing::warn;
 
 use std::fmt::Debug;
 use std::sync::Arc;
 use std::sync::Mutex;
-use tracing::debug;
-use tracing::warn;
 
 #[derive(Clone)]
 pub struct CryptokiConfigDirect {
     pub module_path: Utf8PathBuf,
     pub pin: AuthPin,
     pub serial: Option<Arc<str>>,
+    pub uri: Option<Arc<str>>,
 }
 
 impl Debug for CryptokiConfigDirect {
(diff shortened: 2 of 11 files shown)


crates/extensions/tedge-p11-server/Cargo.toml
```diff
@@ -15,6 +15,7 @@ asn1-rs.workspace = true
 camino.workspace = true
 clap.workspace = true
 cryptoki.workspace = true
+percent-encoding = "2.3.1"
```
reubenmiller, line 18: This probably needs to be a workspace dependency (to be in-line with other deps)

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -31,6 +31,69 @@ Use Private Key in SoftHSM2 using tedge-p11-server
...
+    ...    certificates, we'll only be using one key and we'll only import it to the chosen token, keeping other tokens
+    ...    empty.
+
+    # configure tedge
+    Execute Command    tedge config set c8y.url "$(echo ${C8Y_CONFIG.host} | sed 's|https?://||g')"
+    Execute Command    tedge config set mqtt.bridge.built_in true
+    Execute Command    tedge config set device.cryptoki.mode socket
+
+    Execute Command
+    ...    cmd=printf 'TEDGE_DEVICE_CRYPTOKI_MODULE_PATH=%s\nTEDGE_DEVICE_CRYPTOKI_PIN=%s\nTEDGE_DEVICE_CRYPTOKI_URI=%s\n' "/usr/lib/softhsm/libsofthsm2.so" "123456" "" | sudo tee /etc/tedge/plugins/tedge-p11-server.conf
```
reubenmiller, line 48: It would be better to create a keyword in the robot file to set the configuration.

e.g. 

```
Set tedge-p11-server Configuration
    [Arguments]    ${pkcs11_url}
    Execute Command
    ...    cmd=printf 'TEDGE_DEVICE_CRYPTOKI_MODULE_PATH=%s\nTEDGE_DEVICE_CRYPTOKI_PIN=%s\nTEDGE_DEVICE_CRYPTOKI_URI=%s\n' "/usr/lib/softhsm/libsofthsm2.so" "123456" "${pkcs11_url}" | sudo tee /etc/tedge/plugins/tedge-p11-server.conf
```

Then use it in the tests like so (note the usage of named arguments here, e.g. `pkcs11_url=`):

```
Set tedge-p11-server Configuration     pkcs11_url=pkcs11:token=token1
```

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -31,6 +31,69 @@ Use Private Key in SoftHSM2 using tedge-p11-server
...
+    Execute Command
+    ...    cmd=sudo env C8Y_USER='${C8Y_CONFIG.username}' C8Y_PASSWORD='${C8Y_CONFIG.password}' tedge cert upload c8y
+
+    Execute Command    tedge reconnect c8y
+
+    # expect failure if we try to use a token that doesnt exist
+    Execute Command
+    ...    cmd=printf 'TEDGE_DEVICE_CRYPTOKI_MODULE_PATH=%s\nTEDGE_DEVICE_CRYPTOKI_PIN=%s\nTEDGE_DEVICE_CRYPTOKI_URI=%s\n' "/usr/lib/softhsm/libsofthsm2.so" "123456" "pkcs11:token=asdf" | sudo tee /etc/tedge/plugins/tedge-p11-server.conf
+    Restart Service    tedge-p11-server
+    Run Keyword And Expect Error    *    Execute Command    tedge reconnect c8y
```
reubenmiller, line 61: Use `exp_exit_code=!0` instead of the Robotframework command, otherwise the tedge reconnect c8y will be run multiple times before it fails.

```suggestion
    Execute Command    tedge reconnect c8y    exp_exit_code=!0
```

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -31,6 +31,69 @@ Use Private Key in SoftHSM2 using tedge-p11-server
...
+
+    Execute Command
+    ...    cmd=printf 'TEDGE_DEVICE_CRYPTOKI_MODULE_PATH=%s\nTEDGE_DEVICE_CRYPTOKI_PIN=%s\nTEDGE_DEVICE_CRYPTOKI_URI=%s\n' "/usr/lib/softhsm/libsofthsm2.so" "123456" "pkcs11:token=tedge" | sudo tee /etc/tedge/plugins/tedge-p11-server.conf
+    Restart Service    tedge-p11-server
+    Execute Command    tedge reconnect c8y
+
+    # generate some other keys on the token that has valid tedge key
+    Execute Command
```
reubenmiller, line 81: Using named arguments is much nicer than having to escape the `=` sign everywhere.

```
Execute Command
    ...    cmd=p11tool --set-pin=123456 --login --generate-privkey ECDSA --curve=secp256r1 --label "key2" "pkcs11:token=tedge"
```

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
reubenmiller: Resolved by 29c496474acb2fb0daf2325fb456e88ffd926985

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
reubenmiller: Resolved by 29c496474acb2fb0daf2325fb456e88ffd926985

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
reubenmiller: Resolved by 29c496474acb2fb0daf2325fb456e88ffd926985

crates/extensions/tedge-p11-server/src/pkcs11.rs
```diff
@@ -78,38 +70,128 @@ impl Pkcs11SigningKey {
...
+                let Attribute::KeyType(keytype) = key_type else {
+                    anyhow::bail!("can't get key type");
+                };
+                let pkcs11 = PKCS11 {
+                    session: Arc::new(Mutex::new(session)),
+                };
+                match keytype {
+                    KeyType::EC => Pkcs11SigningKey::Ecdsa(ECSigningKey { pkcs11 }),
+                    KeyType::RSA => Pkcs11SigningKey::Rsa(RSASigningKey { pkcs11 }),
+                    _ => panic!("unsupported key type"),
```
didier-wenzek, line 141: Is a panic really appropriate here?

crates/extensions/tedge-p11-server/src/pkcs11.rs
```diff
@@ -78,38 +70,128 @@ impl Pkcs11SigningKey {
...
+                let pkcs11 = PKCS11 {
+                    session: Arc::new(Mutex::new(session)),
+                };
+                let key_type = get_key_type(pkcs11.clone()).context("Failed to read key")?;
+                trace!("Key Type: {:?}", key_type.to_string());
+                match key_type {
+                    KeyType::EC => Pkcs11SigningKey::Ecdsa(ECSigningKey { pkcs11 }),
+                    KeyType::RSA => Pkcs11SigningKey::Rsa(RSASigningKey { pkcs11 }),
+                    _ => {
+                        panic!("Unsupported key type. Only EC and RSA keys are supported");
```
didier-wenzek, line 154: Is a panic really appropriate here? I would return an error as this can be due to a misconfiguration.

crates/extensions/tedge-p11-server/src/pkcs11.rs
```diff
@@ -383,3 +465,102 @@ fn write_asn1_integer(writer: &mut dyn std::io::Write, b: &[u8]) {
...
+            let input = "pkcs11:token=The%20Software%20PKCS%2311%20Softtoken;\
+            manufacturer=Snake%20Oil,%20Inc.;\
+            model=1.0;\
+            object=my-certificate;\
+            type=cert;\
+            id=%69%95%3E%5C%F4%BD%EC%91;\
+            serial=\
+            ?pin-source=file:/etc/token_pin";
+
+            // let input = "pkcs11:model=SoftHSM%20v2;manufacturer=SoftHSM%20project;serial=5c1becd5222a84f6;token=token123;id=%01;object=object123;type=private";
```
didier-wenzek, line 549: Is there a plan to test also this input?

didier-wenzek (commented):
I will be happy to approve. There are small points still to address though.

crates/extensions/tedge-p11-server/src/pkcs11.rs
```diff
@@ -76,21 +77,40 @@ impl Cryptoki {
...
-        let slot = self
-            .context
-            .get_slots_with_token()?
+        let CryptokiConfigDirect { pin, uri, .. } = &self.config;
+
+        let uri_attributes = uri
+            .as_deref()
+            .map(|uri| uri::Pkcs11Uri::parse(uri).context("Failed to parse PKCS #11 URI"))
+            .transpose()?;
+        dbg!(&uri_attributes);
```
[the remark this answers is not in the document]
Bravo555: removed in 0208319f

crates/extensions/tedge-p11-server/src/pkcs11.rs
Bravo555: changed to `anyhow::bail!` in 0208319f

crates/extensions/tedge-p11-server/src/pkcs11.rs
Bravo555: changed to `anyhow::bail!` in 0208319f

crates/extensions/tedge-p11-server/src/pkcs11.rs
Bravo555: All of these attributes are already tested in the example above, so removed this comment in 0208319f.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -31,6 +31,63 @@ Use Private Key in SoftHSM2 using tedge-p11-server
...
+    # Upload the self-signed certificate
+    Execute Command
+    ...    cmd=sudo env C8Y_USER='${C8Y_CONFIG.username}' C8Y_PASSWORD='${C8Y_CONFIG.password}' tedge cert upload c8y
+
+    Execute Command    tedge reconnect c8y
+
+    # expect failure if we try to use a token that doesnt exist
+    Set tedge-p11-server Configuration    pkcs11_url=pkcs11:token=asdf
+    Restart Service    tedge-p11-server
+    Execute Command    tedge reconnect c8y    exp_exit_code=!0
```
didier-wenzek, line 59: There are so many reason for `tedge reconnect` to fail that I would assess that we really get the error we expect.

```suggestion
    ${stderr}=    Execute Command    tedge reconnect c8y    exp_exit_code=!0    stdout=false    stderr=true
    Should Contain    ${stderr}    tedge_p11_server
    Should Contain    ${stderr}    Failed to find a signing key: Didn't find a slot to use
```

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -31,6 +31,63 @@ Use Private Key in SoftHSM2 using tedge-p11-server
...
+
+    # create tokens with no keys on them, so key selection fails if wrong token is selected
+    Execute Command    softhsm2-util --init-token --free --label token2 --pin "123456" --so-pin "123456"
+    Execute Command    softhsm2-util --init-token --free --label token3 --pin "123456" --so-pin "123456"
+    Execute Command    softhsm2-util --init-token --free --label token4 --pin "123456" --so-pin "123456"
+    Execute Command    softhsm2-util --init-token --free --label token5 --pin "123456" --so-pin "123456"
+    Execute Command    softhsm2-util --init-token --free --label token6 --pin "123456" --so-pin "123456"
```
didier-wenzek, line 66: I don't get the point of these fake tokens. The tests pass without those.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
Bravo555: fixed in a8a732e. The suite was also refactored a bit in the next commit so it should be a bit easier to read.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
Bravo555: removed in a8a732e.

This is a leftover I forgot to remove from the previous approach which I abandoned because it was bad. Currently when there are many keys, we just use the first one returned to us by the PKCS#11 implementation. The idea was to create many new tokens, such that one of them would be picked up as "first" and used and later return an error because it does not contain a key.

But relying on an arbitrary implementation-specific order is bad. In situations where there are many tokens/keys to be used and we can't automatically select a good one, we should fail right there and require a user to provide a URI that uniquely identifies a key to be used.

So the test was changed to explicitly point to a token without the key and then to a token with the correct key. As for actually returning an error where a key cannot be uniquely identified, it will be implemented in the next commit.

Bravo555: warning implemented in 7aa3921

didier-wenzek (approved):
Approved

Issue #3487, opened 2025-03-19
option to mark child devices with c8y_IsDevice fragment via c8y smartrest 2.0 template


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

An upcoming Cumulocity change to the SmartREST 2.0 `101` message will add support for creating child devices with the `c8y_IsDevice` fragment, which means the device can show up in the Cumulocity All Devices list.

New SmartREST 2.0 `101` message example:

```csv
101,child_IsDevice_2,"child2 name",childType,<true|false>
```

Where the 5th column in the CSV controls whether the `c8y_IsDevice` fragment is to be added to the child device or not (for newly created child devices only). The options are as follows:

* `true` => Add the `c8y_IsDevice` fragment
* `false` => Don't add `c8y_IsDevice` (the current default behaviour)

The additional fragment can already be added to the outgoing `101` messages, because if Cumulocity does not support this new fields it will just ignore it (so no Cumulocity version/feature checking is required).

**Describe the solution you'd like**

The following changes are required

* Add new tedge.toml configuration to control whether child devices should have the 5th columns set to `true` or `false`. Default to `true` (this is the new default being adopted by other Cumulocity products). 
* Add the 5th column, where the value is based on the new tedge.toml configuration, to all outgoing `101` Cumulocity SmartREST 2.0 messages (regardless of the Cumulocity version)

**Describe alternatives you've considered**

The following decisions were
* Controlling the c8y_IsDevice setting per child device is not required as it would lead to inconsistencies (why is child-device1 in the All Devices list, but child-device2 isn't). We can change this stance in the future if there is enough customer feedback, but for the moment we will favour fleet consistency over flexibility 

**Additional context**

rina23q:
> Controlling the c8y_IsDevice setting per child device is not required as it would lead to inconsistencies (why is child-device1 in the All Devices list, but child-device2 isn't). We can change this stance in the future if there is enough customer feedback, but for the moment we will favour fleet consistency over flexibility

As of now, if user wants to add `c8y_IsDevice` for a specific device after the child device creation, they can use `twin` topic.

```
tedge mqtt pub -r te/device/child1///twin/c8y_IsDevice '{}'
```

That updates the MO of the child device and makes it shown in the all devices list.

PR #3560, opened 2025-04-11
feat: append the option whether to add c8y_IsDevice to child devices

rina23q:
## Proposed changes
The static SmartREST template 101 is extended to accept boolean if a newly created child device should hold the `c8y_IsDevice` fragment.

This change adds a new tedge configuration to control it.

todos:

- [x] Get the consensus for the name of the new tedge config. => `c8y.smartrest.child_device.create_with_device_marker`
- [ ] ~~Add (or modify if needed) a system test~~ -> eu-latest is not supporting this new feature yet anyway

## Types of changes

## Paste Link to the issue
#3487 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -246,6 +246,12 @@ define_tedge_config! {
             /// Switch using 501-503 (without operation ID) or 504-506 (with operation ID) SmartREST messages for operation status update
             #[tedge_config(example = "true", default(value = true))]
             use_operation_id: bool,
+
+            child_device: {
+                /// Attach the c8y_IsDevice fragment to child devices on creation
+                #[tedge_config(example = "false", default(value = false))]
+                create_with_device_marker: bool,
+            }
         },
 
         smartrest1: {
--- a/crates/extensions/c8y_mapper_ext/src/config.rs
+++ b/crates/extensions/c8y_mapper_ext/src/config.rs
@@ -62,6 +62,7 @@ pub struct C8yMapperConfig {
     pub bridge_service_name: String,
     pub bridge_health_topic: Topic,
     pub smartrest_use_operation_id: bool,
+    pub smartrest_child_device_create_with_device_marker: bool,
 
     pub data_dir: DataDir,
     pub config_dir: Arc<Utf8Path>,
@@ -102,6 +103,7 @@ impl C8yMapperConfig {
         software_management_with_types: bool,
         auto_log_upload: AutoLogUpload,
         smartrest_use_operation_id: bool,
+        smartrest_child_device_create_with_device_marker: bool,
         max_mqtt_payload_size: u32,
     ) -> Self {
         let ops_dir = config_dir
@@ -143,6 +145,7 @@ impl C8yMapperConfig {
             bridge_service_name,
             bridge_health_topic,
             smartrest_use_operation_id,
+            smartrest_child_device_create_with_device_marker,
 
             config_dir,
             logs_path,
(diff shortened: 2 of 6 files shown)


crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -246,6 +246,12 @@ define_tedge_config! {
             /// Switch using 501-503 (without operation ID) or 504-506 (with operation ID) SmartREST messages for operation status update
             #[tedge_config(example = "true", default(value = true))]
             use_operation_id: bool,
+
+            child: {
+                /// Attach the c8y_IsDevice fragment to child devices
+                #[tedge_config(example = "false", default(value = false))]
+                append_device_fragment: bool,
+            }
```
albinsuresh, line 254:
```suggestion
            child_device: {
                /// Attach the c8y_IsDevice fragment to child devices
                #[tedge_config(example = "false", default(value = false))]
                is_device: bool,
            }
```
Combined property name: `tedge.smartrest.child_device.is_device`. I'm not 100% convinced about that name either. But the intention was to convey the effect of the setting rather than what it technically does under-the-hood.

albinsuresh (approved):
Feel free to merge once the name is finalized.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
reubenmiller: This is a bit verbose, but it does following [101 SmartREST message description](https://cumulocity.com/docs/smartrest/mqtt-static-templates/#101)

```
smartrest.child_device.create_with_device_marker
```

albinsuresh: That name is as direct as it can be. So, totally fine with that. But, if we really want to make it shorter, I might consider dropping the `create_with` prefix and just keep just `smartrest.child_device.device_marker` or even `smartrest.child_device.is_device_marker`.

reubenmiller: Though I think the `smartrest.child_device.is_device_marker` is lacking the intent of the config, and do we know if the `c8y_IsDevice` fragment is added/removed if the values changes after the device actually exists (I would suspect that it won't)...if so, then "create_with_device_marker" is more descriptive that it will only be added when creating the device.

rina23q: `101` message doesn't override the existing settings. You will just get an error `41,101,Device with serial 'test-54321' already existing`. It means user can control `c8y_IsDevice` only for the device creation time by `101`.

So, I will update this PR with:
```
smartrest.child_device.create_with_device_marker
```

rina23q: updated: 70a4a5dc106aac91d2c66a29d458ddfa428bde0d

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -62,7 +62,7 @@ pub struct C8yMapperConfig {
     pub bridge_service_name: String,
     pub bridge_health_topic: Topic,
     pub smartrest_use_operation_id: bool,
-    pub smartrest_child_append_device_fragment: bool,
+    pub smartrest_child_device_create_with_device_marker: bool,
```
albinsuresh, line 65: Not a request for change: I'm just surprised clippy isn't freaking out with that name 😉.

albinsuresh (approved):
LGTM

reubenmiller (approved):
Approved. It worked as expected (when testing it on a tenant that supports the new SmartREST 2.0 template)

Issue #3547, opened 2025-04-08
tedge cert create-csr output-path flag fails when given a relative address


reubenmiller:
**Describe the bug**

When creating a local CSR via `tedge cert create-csr c8y` and giving an explicit output path, the command fails
when a relative path is given.

```sh
$ $ tedge cert create-csr c8y --device-id example --output-path ./local.csr

Error: failed to Generate a Certificate Signing Request.

Caused by:
    Invalid device.cert_path path: Relative path: "./local.csr" is not permitted. Provide an absolute path instead.
```

However the command creates the CSR if the full path is given, however enforcing the user to provide the full path when calling a CLI command is unexpected.

```sh
$ tedge cert create-csr c8y --output-path "$(pwd)/local.csr"
```

**To Reproduce**

It can be reproduced by running:

```sh
$ tedge cert create-csr c8y --device-id example --output-path ./local.csr
```

**Expected behavior**

Relative paths should be supported when provided via the `--output-path` flag. The flag output should be expanded to the full path after parsing the command.

So it is expected that the following produce a `local.csr` file in the current working directory.

```sh
$ tedge cert create-csr c8y --device-id example --output-path ./local.csr
```

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `macOS`
 - Hardware [incl. revision]: `aarch64`
 - System-Architecture [e.g. result of "uname -a"]: `Darwin example.local 23.6.0 Darwin Kernel Version 23.6.0: Thu Sep 12 23:35:29 PDT 2024; root:xnu-10063.141.1.701.1~1/RELEASE_ARM64_T6000 arm64`
 - thin-edge.io version [e.g. 0.1.0]: `tedge 1.4.3~428+g42bf3a1`

**Additional context**

PR #3562, opened 2025-04-12
fix: support relative output paths when creating a csr

reubenmiller:
## Proposed changes

Allow users to provide a relative output path when using `tedge cert create-csr`.

**Example**

```
tedge cert create-csr c8y --device-id test-user --output-path ./local.csr
```

Note: In the future the code can be simplified once the MSRV can be increased beyond 1.78, however the current MSRV is required to support Yocto environments

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3547

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/certificate/create_csr.rs
+++ b/crates/core/tedge/src/cli/certificate/create_csr.rs
@@ -44,7 +44,7 @@ impl Command for CreateCsrCmd {
 impl CreateCsrCmd {
     pub async fn create_certificate_signing_request(&self) -> Result<(), CertError> {
         let id = &self.id;
-        let csr_path = &self.csr_path;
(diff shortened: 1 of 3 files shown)


didier-wenzek (changes requested):
I don't think this is the correct approach to solve the issue.

- The create CSR command is not the only command that can be given a relative path.
- There is no issue to read or write a file given a relative path.
- The root cause of the issue is that in many places there is a call to: [validate_parent_dir_exists](https://github.com/thin-edge/thin-edge.io/blob/main/crates/common/tedge_utils/src/paths.rs#L145)
- Does this check make sense?
- If so this check has to be fixed, not the path.

reubenmiller:
Closing in favour of https://github.com/thin-edge/thin-edge.io/pull/3564

PR #3564, opened 2025-04-14
fix: support relative path args when using tedge cert

didier-wenzek:
## Proposed changes

Allow users to provide a relative  path when using `tedge cert` sub-commands a `create-csr` or `show`.

- [x] Improve `tedge config set` to make sure no relative paths is stored in the TOML config  (because used later from any working directory)
- [x] Improve `tedge config set` to store cannonalized  paths when given relative paths
- [x] Remove from `tedge cert` the check that was preventing a path argument to be relative. 

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3547

## Checklist


albinsuresh (commented):
Changes looked fine. But just wanted to confirm if we wanted to restrict relative paths not just from the TOML file, but also in the CLI when they are set. Will approve after this confirmation.

rina23q (approved):
The changes really make sense. CLI accepts both absolute and relative paths while `tedge config` accepts only absolute.

albinsuresh (approved):

tests/RobotFramework/tests/mqtt/basic_pub_sub.robot
```diff
@@ -53,5 +53,5 @@ tedge configure MQTT server authentication
     Execute Command    tedge config set mqtt.client.auth.ca_file /etc/mosquitto/ca_certificates/ca.crt
 
 tedge configure MQTT client authentication
-    Execute Command    tedge config set mqtt.client.auth.cert_file client.crt
-    Execute Command    tedge config set mqtt.client.auth.key_file client.key
+    Execute Command    tedge config set mqtt.client.auth.cert_file "$(pwd)/client.crt"
```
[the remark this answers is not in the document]
didier-wenzek: Here a proposal to implement this idea to normalize a path configured using `tedge config set`

https://github.com/thin-edge/thin-edge.io/pull/3564/commits/368ec48c7569bcb5b38b90f7ee518f84136d5b31

crates/common/tedge_config/src/tedge_toml/models/path.rs
```diff
@@ -20,11 +20,18 @@ impl AbsolutePath {
...
     pub fn from_path(path: Utf8PathBuf) -> Result<Self, InvalidAbsolutePath> {
-        if path.is_relative() {
-            Err(InvalidAbsolutePath(path))
+        let absolute_path = if path.is_absolute() {
+            path
         } else {
-            Ok(AbsolutePath(path))
-        }
+            let Ok(Ok(pwd)) = std::env::current_dir().map(Utf8PathBuf::try_from) else {
+                return Err(InvalidAbsolutePath(path));
```
albinsuresh, line 27: Minor:
```suggestion
                return Err(InvalidRelativePath(path));
```
Else the error message will be slightly misleading, right? As the issue is not at all with the input `path` but from the derived `pwd`.

crates/core/tedge_agent/src/http_server/request_files.rs
```diff
@@ -88,5 +88,6 @@ fn local_path_for_file(
 }
 
 fn clean_utf8_path(path: &Utf8Path) -> Utf8PathBuf {
-    Utf8PathBuf::from(path_clean::clean(path.as_str()))
+    // unwrap is safe because clean returns an utf8 path when given an utf8 path
+    Utf8PathBuf::try_from(path_clean::clean(path.as_std_path())).unwrap()
```
albinsuresh, line 92: I'm assuming that we're using `path_clean::clean()` instead of `std::path::absolute()` because our MSRV is still 1.78 and the latter is only available from `1.79`.

albinsuresh (approved):
LGTM. Just one minor remark on an error message.

crates/common/tedge_config/src/tedge_toml/models/path.rs
didier-wenzek: I changed the error message to "Cannot be converted to an absolute path" but let the error type name unchanged because it makes sense for `AbsolutePath::try_new` to fail with `InvalidAbsolutePath`.


See https://github.com/thin-edge/thin-edge.io/pull/3564/commits/4134c9e92038199220a156b659fe208ad546fb4a

reubenmiller (approved):

Issue #3543, opened 2025-04-08
support pkcs11 in aws and azure mappers


reubenmiller:
**Is your feature request related to a problem? Please describe.**

The tedge-mapper should also support accessing private keys stored in HSMs via the PKCS11 interface (which c8y now supports).

**Describe the solution you'd like**

Add support for AWS and Azure mappers to use private keys stored in a HSM.

**Describe alternatives you've considered**

**Additional context**

PR #3566, opened 2025-04-16
feat: PKCS#11 support for aws and azure

Bravo555:
## Proposed changes

Add PKCS#11 support for AWS and Azure built-in bridges.

I have manually tested `tedge connect aws` and `tedge connect az` using a self-signed certificate created by `tedge cert create` which now creates a cert that uses the ECDSA P-256 with SHA256 sigscheme.

## Types of changes

## Paste Link to the issue

- #3543 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
@@ -111,6 +111,19 @@ impl MqttAuthConfig {
 }
 
 impl TEdgeConfig {
+    /// Returns a [`rustls::ClientConfig`] for an MQTT client that will connect to the MQTT broker
+    /// of a cloud given in the parameter.
+    pub fn mqtt_client_config_rustls(
+        &self,
+        cloud: &dyn CloudConfig,
+    ) -> anyhow::Result<rustls::ClientConfig> {
+        let client_config = self
+            .mqtt_auth_config_cloud_broker(cloud)?
+            .to_rustls_client_config()?;
+
+        Ok(client_config)
+    }
+
     pub fn mqtt_config(&self) -> Result<mqtt_channel::Config, CertificateError> {
         let host = self.mqtt.client.host.as_str();
         let port = u16::from(self.mqtt.client.port);
@@ -146,8 +159,6 @@ impl TEdgeConfig {
         &self,
         cloud: &dyn CloudConfig,
     ) -> anyhow::Result<MqttAuthConfigCloudBroker> {
-        let ca_path = cloud.root_cert_path().to_path_buf();
-
         // if client cert is set, then either cryptoki or key file must be set
         let client_auth = match self.device.cryptoki.config()? {
             Some(cryptoki_config) => MqttAuthClientConfigCloudBroker {
@@ -161,7 +172,7 @@ impl TEdgeConfig {
         };
 
         Ok(MqttAuthConfigCloudBroker {
-            ca_path,
+            ca_path: cloud.root_cert_path().to_path_buf(),
             client: Some(client_auth),
         })
     }
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
@@ -13,18 +13,16 @@ use certificate::parse_root_certificate::AuthPin;
 use certificate::CertificateError;
 use tedge_config_macros::all_or_nothing;
 
+use super::CloudConfig;
 use super::TEdgeConfigReaderDeviceCryptoki;
 
 use certificate::parse_root_certificate::CryptokiConfig;
 use certificate::parse_root_certificate::CryptokiConfigDirect;
 
 /// An MQTT authentication configuration for connecting to the remote cloud broker.
-///
-/// Either ca_dir or ca_file must be set.
 #[derive(Debug, Clone, Default)]
 pub struct MqttAuthConfigCloudBroker {
-    pub ca_dir: Option<Utf8PathBuf>,
-    pub ca_file: Option<Utf8PathBuf>,
+    pub ca_path: Utf8PathBuf,
     pub client: Option<MqttAuthClientConfigCloudBroker>,
 }
 
@@ -43,10 +41,6 @@ pub enum PrivateKeyType {
 
 impl MqttAuthConfigCloudBroker {
     pub fn to_rustls_client_config(self) -> anyhow::Result<rustls::ClientConfig> {
-        let Some(ca) = self.ca_dir.or(self.ca_file) else {
-            anyhow::bail!("No CA file or directory provided");
-        };
-
         let Some(MqttAuthClientConfigCloudBroker {
             cert_file,
             private_key,
@@ -57,11 +51,15 @@ impl MqttAuthConfigCloudBroker {
 
         let client_config = match private_key {
             PrivateKeyType::File(key_file) => {
-                certificate::parse_root_certificate::create_tls_config(ca, key_file, cert_file)
+                certificate::parse_root_certificate::create_tls_config(
+                    self.ca_path,
+                    key_file,
+                    cert_file,
+                )
             }
             PrivateKeyType::Cryptoki(cryptoki_config) => {
                 certificate::parse_root_certificate::create_tls_config_cryptoki(
-                    ca,
+                    self.ca_path,
                     cert_file,
                     cryptoki_config,
                 )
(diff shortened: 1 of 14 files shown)


crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -52,7 +52,15 @@ impl TEdgeComponent for AwsMapper {
...
-            use_key_and_cert(&mut cloud_config, aws_config)?;
+
+            // set tls config
+            let cloud_broker_auth_config = tedge_config
+                .mqtt_auth_config_cloud_broker(aws_config)
+                .expect("error getting cloud broker auth config");
+            let client_config = cloud_broker_auth_config
+                .to_rustls_client_config()
+                .expect("error getting cloud broker auth config");
+            cloud_config.set_transport(Transport::tls_with_config(client_config.into()));
```
didier-wenzek, line 63: This piece of code is repeated for c8y, aws and azure mappers. This will also be useful to let `tedge connect` validate a new certificate using a direct connection to the cloud.

Hence, I would add a new method in tedge config that returns a `rustls::ClientConfig`, to write:

```suggestion
            let tls_config =  tedge_config.rustls_client_config(aws_config)?;
            cloud_config.set_transport(Transport::tls_with_config(tls_config.into()));
```


Furthermore; the function `use_key_and_cert` is now only used by a test and can be safely removed.

Bravo555: fixed in 884a5caa43

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -153,13 +153,10 @@ impl TEdgeComponent for CumulocityMapper {
...
-                let client_config = cloud_broker_auth_config
-                    .to_rustls_client_config()
-                    .expect("error getting cloud broker auth config");
-                cloud_config.set_transport(Transport::tls_with_config(client_config.into()));
+                let tls_config = tedge_config
+                    .mqtt_client_config_rustls(c8y_config)
+                    .context("Failed to create MQTT TLS config")?;
+                cloud_config.set_transport(Transport::tls_with_config(tls_config.into()));
             } else {
                 // TODO(marcel): integrate credentials auth into MqttAuthConfig?
```
reubenmiller, line 161: Not related to this PR but we need to be careful to not introduce named TODO's into the code. Generally we like to avoid TODO's if possible all together, but sometimes they do make sense, but should not be tied to a specific person.

crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
```diff
@@ -142,34 +153,30 @@ impl TEdgeConfig {
...
+        cloud: &dyn CloudConfig,
     ) -> anyhow::Result<MqttAuthConfigCloudBroker> {
-        let c8y = self.c8y.try_get(c8y_profile)?;
-
-        let mut mqtt_auth = MqttAuthConfigCloudBroker {
-            ca_dir: Some(c8y.root_cert_path.clone().into()),
-            ca_file: None,
-            client: None,
-        };
+        let ca_path = cloud.root_cert_path().to_path_buf();
```
reubenmiller, line 162: Does this still work when the user is using a directory path for the `<cloud>.root_cert_path`? As it seems that later on the `ca_path` variable is being assigned to the `cert_file` property of the `MqttAuthClientConfigCloudBroker` which seems to suggest that it would break something.

crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
```diff
@@ -142,34 +153,30 @@ impl TEdgeConfig {
...
 
         // if client cert is set, then either cryptoki or key file must be set
-        if let Some(cryptoki_config) = self.device.cryptoki.config()? {
-            mqtt_auth.client = Some(MqttAuthClientConfigCloudBroker {
-                cert_file: c8y.device.cert_path.clone().into(),
+        let client_auth = match self.device.cryptoki.config()? {
+            Some(cryptoki_config) => MqttAuthClientConfigCloudBroker {
+                cert_file: cloud.device_cert_path().to_path_buf(),
```
reubenmiller, line 167: `cloud.device_cert_path` can be either a file or path to a directory. Does this still work as expected?

crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
Bravo555: In this particular method the only use of `ca_path` I see is in the final return, but I've inlined the binding for clarity in 0848bf1a

crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
Bravo555: AFAIK `device_cert_path` the path is to a device certificate, so it can only be a file. It's derived from `device/c8y/aws/az.cert_path`, which point to files.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -13,7 +13,7 @@ use std::path::Path;
 use std::sync::Arc;
 use zeroize::Zeroizing;
 
-pub const MAX_PACKET_SIZE: usize = 268435455;
+pub const MAX_PACKET_SIZE: usize = 268435456;
```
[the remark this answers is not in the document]
Bravo555: My bad, was just experimenting and that was not meant to be included or changed. Removed in 0848bf1a

crates/core/tedge_mapper/src/c8y/mapper.rs
Bravo555: Right, my intention in this case is to note some possible improvement to the code that for the time being I've avoided doing and noting down the author (me) to contact about this possible improvement, because the information about who's the author of the comment might change if the code is e.g. moved. So I'm not necessarily tying myself to that particular TODO item.

> Generally we like to avoid TODO's if possible all together
I'll keep it in mind and avoid adding any more TODOs.

crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
reubenmiller: Sorry my mistake. I was mixing this up with the ca_path...you're right this is just about the device certificate which is a file.

reubenmiller (changes requested):
**Update: ✅  resolved**

When testing it with AWS it seems that the `tedge connect aws` does not display the correct value for the `cryptoki: false` field and there is an erroneous error message when the change of ownership of the non-existent key file is attempted. Below shows the current output which needs to be fixed:

```
stdout:
Disconnecting from Aws
Removing bridge config file... ✓
Bridge doesn't exist. Device is already disconnected from Aws.
reconnect to Aws cloud.:
	device id: TST_bag_broad_option
	cloud profile: <none>
	cloud host: example-ats.iot.us-east-1.amazonaws.com:8883
	auth type: Certificate
	cryptoki: false
	bridge: built-in
	service manager: systemd
Restarting mosquitto... ✓
Waiting for mosquitto to be listening for connections... ✓
Enabling tedge-mapper-aws... ✓
Verifying device is connected to cloud... ✓

stderr:
2025-04-23T15:07:24.471004841Z  WARN tedge::cli::connect::command: Failed to change ownership of /etc/tedge/device-certs/tedge-private-key.pem to tedge:tedge: The path is not accessible. "/etc/tedge/device-certs/tedge-private-key.pem"
```

I'd assuming the Azure suffers from the same problem but I haven't checked that explicitly yet.

crates/core/tedge/src/bridge/c8y.rs
```diff
@@ -195,7 +193,8 @@ impl From<BridgeConfigC8yParams> for BridgeConfig {
             auth_type,
             mosquitto_version,
             keepalive_interval,
-            use_cryptoki,
+            // this is orthogonal to
+            // use_cryptoki: false,
```
reubenmiller, line 197: This can be removed

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -383,13 +383,22 @@ impl ConnectCommand {
...
-            std::thread::sleep(std::time::Duration::from_secs(2));
+            tokio::time::sleep(std::time::Duration::from_secs(2)).await;
         }
         self.check_connection().await
     }
 
     async fn check_connection(&self) -> Result<DeviceStatus, Fancy<ConnectError>> {
         let config = &self.config;
+        // IDEA: Report if any errors were returned by the mapper
+        //
```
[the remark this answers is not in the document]
reubenmiller: I'd just remove the entire IDEA block, as proposed solutions should not be tied to any specific technology (e.g. systemd/journald), as relying on this to very that "the connection is ok" will be very fragile. If you want to explore the idea then creating a ticket is a better approach.

crates/core/tedge/src/bridge/c8y.rs
Bravo555: removed in 38cb22ef6c

crates/core/tedge/src/cli/connect/command.rs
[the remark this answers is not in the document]
Bravo555: removed in 4367536b11

reubenmiller (approved):
Approved (from a functional viewpoint). But please wait for another Rust reviewer before merging.

rina23q (approved):
The code change looks good. Also, I confirmed it works with Azure IoT Hub.

Issue #3559, opened 2025-04-11
flaky system test: Updating entities from a child device


reubenmiller:
**Describe the bug**

Flaky test: `Updating entities from a child device`

```
Updating entities from a child device                                 | FAIL |
'' does not contain '"name":"Child 01"'
```

**Failed builds**

* https://github.com/thin-edge/thin-edge.io/actions/runs/14400287940/job/40384869398?pr=3558
* https://github.com/thin-edge/thin-edge.io/actions/runs/14400216580/job/40384541649?pr=3551
* https://github.com/thin-edge/thin-edge.io/actions/runs/14400287940?pr=3558
* https://github.com/thin-edge/thin-edge.io/actions/runs/14465440292/job/40569077998?pr=3563
* https://github.com/thin-edge/thin-edge.io/actions/runs/14465440292/job/40569077998?pr=3563

reubenmiller:
I was able to reproduce the issue using the flake-finder and reducing the CPU quote allocated to the system test container:

```sh
export DOCKER_OPTIONS_CPU_QUOTA=5000
export DOCKER_OPTIONS_CPU_PERIOD=20000
invoke flake-finder --test-name "Updating entities from a child device" --iterations 3 --clean
```
This resulted in at least 1 run failing from 3.

```
Overall: FAILED
Results: 3 iterations, 2 passed, 1 failed
Elapsed time: 0:04:01.207033
Failed iterations: [2]
```


Then adding a sleep 1s between the PUT and the GET then results in the tests passing:

```
Overall: PASSED
Results: 3 iterations, 3 passed, 0 failed
Elapsed time: 0:04:19.898668
```

reubenmiller:
It seems that it is a legitimate test failure due to a race condition in the entity store logic.

reubenmiller:
Rerunning the above test now passes.

```sh
invoke flake-finder --test-name "Updating entities from a child device" --iterations 5 --clean
```

```
Overall: PASSED
Results: 5 iterations, 5 passed, 0 failed
Elapsed time: 0:05:31.424670
```

reubenmiller:
Running the test with the fix in https://github.com/thin-edge/thin-edge.io/pull/3581 it now passes 100% of the time.

```
$ invoke flake-finder --test-name "Updating entities from a child device" --iterations 100 --clean
Overall: PASSED
Results: 100 iterations, 100 passed, 0 failed
Elapsed time: 0:16:13.157208
```

PR #3570, opened 2025-04-18
fix: Avoid clearing updated keys on twin update

albinsuresh:
## Bug RCA

The test flakiness where the `twin/name` fragment was becoming `null` after the same was updated with a new value because of a faulty logic in how twin updates are handled. When a given twin fragment is updated with a new value, instead of overwriting the existing value with the new one, the current code was trying to clear the old value with a null message first, followed by publishing the updated value. This is the cause of the race, as sometimes, the `twin` query in the test was happening immediately after the `null` value is processed which results in empty response.

## Proposed changes

When existing twin data like {"a": 1, "b": 2, "c": 3} is updated with {"b": 2, "c": 4, "d": 5}, the update is handled as follows:
1. The removed key: "a" is cleared
2. The unchanged key: "b" is not republished
3. The updated key: "c" is republished with the new value 4
4. The new key: "d" is published as new
## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3559

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/entity_manager/server.rs
+++ b/crates/core/tedge_agent/src/entity_manager/server.rs
@@ -123,7 +123,7 @@ impl Server for EntityStoreServer {
                 EntityStoreResponse::GetTwinFragments(res.cloned())
             }
             EntityStoreRequest::SetTwinFragments(topic_id, fragments) => {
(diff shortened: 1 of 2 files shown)


didier-wenzek (approved):
Approved.

Issue #3582, opened 2025-04-27
Preserve workflow failure reasons if set by a state


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When using workflows, if the operation is set to failed, then the default failure reason generally isn't very helpful as it it just includes the exit code from the last state, and it overrides any more meaningful reason already set in previous states. Below shows an example of the generic error reason which overrides any existing reason.

```
/usr/bin/rugpi_workflow.sh returned exit code 1
```

Below shows a more detailed failure reason was given (set via the `begin/end-tedge` blocks), but it is overridden with the generic reason, below shows the snippet from the workflow log where the reason is set.

```sh
:::begin-tedge:::
{"reason":"Refusing to install update as the current (hot) partition is not the default partition. This indicates that you may be in the middle of an update. Please reboot to switch to the default partition"}
:::end-tedge:::
```

Having a more descriptive failure reason helps users understand more quickly what the problem is without having to download the full workflow log (as the failure reason is generally visible in the Cumulocity Device Management Application).

**Describe the solution you'd like**

The tedge-agent should only use the generic "x returned exit code y" when an existing `.reason` property is not already set. This would allow the workflow to define it's own more detailed failure reason.

**Describe alternatives you've considered**

**Additional context**

The full workflow log taken from the Rugix firmware update workflow which shows a non-trivial process where a failure reason was set by the failing workflow, but not propagated back to the UI as it was overridden.

```log
==================================================================
Triggered firmware_update workflow
==================================================================

topic:     te/device/main///cmd/firmware_update/c8y-mapper-45569495
operation: firmware_update
cmd_id:    c8y-mapper-45569495
time:      2025-04-27T18:43:50.748997844Z

==================================================================

----------------------[ firmware_update @ init | time=2025-04-27T18:43:50.763384777Z ]----------------------

State:    {"logPath":"/var/log/tedge/agent/workflow-firmware_update-c8y-mapper-45569495.log","name":"tedge-raspios-arm64","remoteUrl":"https://repo-thinedgeio.s3.us-east-1.amazonaws.com/tedge-raspios-arm64/tedge-raspios-arm64_20250427.1105.img.xz","status":"init","tedgeUrl":"https://repo-thinedgeio.s3.us-east-1.amazonaws.com/tedge-raspios-arm64/tedge-raspios-arm64_20250427.1105.img.xz","version":"20250427.1105-img-aws"}

Action:   move to scheduled state

[...]

PR #3583, opened 2025-04-28
fix: Overwritten workflow failure reason

didier-wenzek:
## Proposed changes

- Preserve workflow failure reasons if set by a state

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3582

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/workflow/handlers.rs
+++ b/crates/core/tedge_api/src/workflow/handlers.rs
@@ -88,34 +88,38 @@ impl ExitHandlers {
         outcome: std::io::Result<std::process::Output>,
     ) -> Value {
         match outcome {
-            Ok(output) => match output.status.code() {
-                None => self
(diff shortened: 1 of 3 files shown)


rina23q (commented):
Now if workflow has already provided the failure reason, user can check the exit code only in the workflow log file. Would it helpful if we add an exit code field, like `{"code": 1}` to JSON payload?

reubenmiller:
> Now if workflow has already provided the failure reason, user can check the exit code only in the workflow log file. Would it helpful if we add an exit code field, like `{"code": 1}` to JSON payload?

I'm not necessarily against that proposal, however generally the exit code is already part of the branching logic between states anyway, so the workflow has most of the information it already needs to make the state transition (which is already based on the exit code).

rina23q (approved):
Approved

reubenmiller (approved):
Approved. I experimented with the new behaviour and it works well.

PR #3587, opened 2025-04-29
refactor: parse entity registration message using serde

albinsuresh:
## Proposed changes

Parse entity registration message using serde for better error messages.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/entity_store.rs
+++ b/crates/core/tedge_api/src/entity_store.rs
@@ -37,9 +37,6 @@ use std::collections::VecDeque;
 use std::mem;
 use std::path::Path;
 
-// In the future, root will be read from config
-const MQTT_ROOT: &str = "te";
-
 /// A store for topic-based entity metadata lookup.
 ///
 /// This object is a hashmap from MQTT identifiers to entities (devices or
@@ -61,11 +58,11 @@ const MQTT_ROOT: &str = "te";
 /// # use mqtt_channel::{MqttMessage, Topic};
 /// # use tedge_api::mqtt_topics::MqttSchema;
 /// # use tedge_api::entity_store::{EntityStore, EntityRegistrationMessage};
-/// let mqtt_message = MqttMessage::new(
-///     &Topic::new("te/device/main//").unwrap(),
-///     r#"{"@type": "device"}"#.to_string(),
-/// );
-/// let registration_message = EntityRegistrationMessage::try_from(&mqtt_message).unwrap();
+///
+/// let registration_message = EntityRegistrationMessage::try_from(
+///     "device/main//".parse().unwrap(),
+///     r#"{"@type": "device"}"#.as_bytes()
+/// ).unwrap();
 ///
 /// let mut entity_store = EntityStore::with_main_device(
 ///     MqttSchema::default(),
@@ -167,7 +164,10 @@ impl EntityStore {
                                 match channel {
                                     Channel::EntityMetadata => {
                                         if let Ok(register_message) =
-                                            EntityRegistrationMessage::try_from(&message)
+                                            EntityRegistrationMessage::try_from(
+                                                source.clone(),
+                                                message.payload_bytes(),
+                                            )
                                         {
                                             if let Err(err) = self.register_entity(register_message)
                                             {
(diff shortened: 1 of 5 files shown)


crates/core/tedge_api/src/entity_store.rs
```diff
@@ -939,114 +940,64 @@ pub enum InitError {
     Custom(String),
 }
 
-/// An object representing a valid entity registration message.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct EntityRegistrationMessage {
-    // fields used by thin-edge locally
+pub struct EntityRegistrationPayload {
```
didier-wenzek, line 944: A doc comment would help to understand why we have two similar structs, aka `EntityRegistrationMessage` and `EntityRegistrationPayload` and what are their relationship.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -939,114 +940,64 @@ pub enum InitError {
...
-            health_endpoint: health_service,
-            twin_data: other,
-        })
+            .and_then(|s| s.strip_prefix('/'))
+            .ok_or_else(|| {
+                serde::de::Error::custom("Failed to parse MQTT topic as a valid entity topic")
+            })?;
+        let topic_id = EntityTopicId::from_str(topic_id).map_err(|_| {
+            serde::de::Error::custom("Failed to parse MQTT topic as a valid entity topic")
+        })?;
```
didier-wenzek, line 990: I would use `MqttSchema::parse` to extract the topic-id and check this is a metadata channel.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -939,114 +940,64 @@ pub enum InitError {
...
+
+        let mut payload: EntityRegistrationPayload =
+            serde_json::from_slice(message.payload_bytes())?;
+        if payload.topic_id.is_some() {
+            return Err(serde::de::Error::custom(
+                "@topic-id should not be present in the MQTT payload as it must be specified in the topic",
+            ));
+        }
+        payload.topic_id = Some(topic_id.clone());
```
didier-wenzek, line 999: Wouldn't it be simpler to remove the topic-id field from `EntityRegistrationPayload` as it must indeed never be used?

Sure, we can no more simply return `payload.try_into()` to get the expected `EntityRegistrationMessage`, but this would really better convey the intent.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -1123,9 +1074,25 @@ impl TryFrom<&MqttMessage> for EntityRegistrationMessage {
...
-            "Failed to parse entity registration message: {value}",
-        ))
+        EntityRegistrationMessage::new(value)
+    }
+}
+
+impl TryFrom<EntityRegistrationPayload> for EntityRegistrationMessage {
+    type Error = serde_json::Error;
+
+    fn try_from(value: EntityRegistrationPayload) -> Result<Self, Self::Error> {
```
didier-wenzek, line 1084: Implementing this trait makes things more complicated than necessary, while used in a single place.

I would:
- remove the `topic_id` field from `EntityRegistrationPayload`
- implement an ad-hoc method `EntityRegistrationMessage::try_from(TopicId, EntityRegistrationPayload)`

crates/core/tedge_api/src/entity_store.rs
albinsuresh: > Wouldn't it be simpler to remove the topic-id field from EntityRegistrationPayload as it must indeed never be used?

The `@topic-id` was added to this struct since the registration payload received via HTTP would have the topic-id in the payload and not the URL path. The other alternative would have been to drop the `@topic-id` from this struct and then introduce another struct specifically for HTTP which includes the `topic-id` and this `EntityRegistrationPayload` as a "flattened" field.

crates/core/tedge_api/src/entity_store.rs
rina23q: Actually I failed to understand why we need `EntityRegistrationPayload`. If it needs for serde, we can implement custom Serialize/Deseralize for `EntityRegistrationMessage`...

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Resolved by [8654090](https://github.com/thin-edge/thin-edge.io/pull/3587/commits/8654090f76cf664e56dff2f493927e9a1b324c18)

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Resolved by [8654090](https://github.com/thin-edge/thin-edge.io/pull/3587/commits/8654090f76cf664e56dff2f493927e9a1b324c18)

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Resolved by [8654090](https://github.com/thin-edge/thin-edge.io/pull/3587/commits/8654090f76cf664e56dff2f493927e9a1b324c18). Instead of parsing the `topic-id` here, I've updated the signature in such a way that the already parsed `topic-id` is taken as a parameter, along with the payload.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Resolved by [8654090](https://github.com/thin-edge/thin-edge.io/pull/3587/commits/8654090f76cf664e56dff2f493927e9a1b324c18)

crates/core/tedge_api/src/entity_store.rs
didier-wenzek: > I've updated the signature in such a way that the already parsed topic-id is taken as a parameter,

Even better.

didier-wenzek (approved):
Approved. The code is much simpler now.

rina23q (approved):
Nice refactoring 👍

Issue #3601, opened 2025-05-08
tedge-p11-server client signing failed due to selected different keys to the given uri


reubenmiller:
**Describe the bug**

A user which is exploring stores stored in a Nitrokey HSM 2 module which has multiple private keys (at least 3 different ones) failed to connect to Cumulocity using the private key which is accessed via the tedge-p11-server service.

When doing a `tedge connect c8y`, the following error appears on the tedge-p11-server logs and the client fails to connect to the cloud.

```
May 08 23:27:09 100666-2-nc-hmi tedge-p11-server[1907642]:     1: Function::Sign: PKCS11 error: The plaintext input data to a cryptographic operation has a bad length.  Depending on the operation’s mechanism, this could mean that the plaintext data is too short, too long, or is not a multiple of some particular block size.  This return value has higher priority than CKR_DATA_INVALID.
```

Upon closer inspection it seems that the tedge-p11-server key selection (which occurs multiple times in the handshake) does not always find the correct key using the given key's url (either specified on the tedge-p11-server --uri flag, or using the client side setting `device.key_uri`). The algorithm selection seems to be ok, but when it tries to sign the data, it selects a different key which results in the error.

For instance, in this case, the private key is an EC based key, however for signing, the tedge-p11-server picked another key which was RSA based.

It seems that the tedge-p11-server key selection does not always honour the given key's uri. When hard coding the key's label, tedge was then able to connect successfully, below shows the hard-coded key template criteria.

```rust
let mut key_template = vec![
    Attribute::Token(true),
    Attribute::Private(true),
    Attribute::Sign(true),
];
// TEST: Hard code a unique label just to prove there was nothing wrong with the key itself
key_template.push(Attribute::Label("device-cert".as_bytes().to_vec()));
```

**To Reproduce**

**Setup**

* tedge-p11-server running on the host
* tedge running inside a container and using the tedge-p11-server socket mounted inside the container (with correct permissions)
* Nitrokey HSM 2 with at least 3 different keys on the same slot

**Expected behavior**

The TLS scheme selection and signing should use the same key as specified by the PKCS11 URI which is either specified via the `tedge-p11-server --uri <uri>` flag, or via the client.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`unknown`|
[...]

PR #3602, opened 2025-05-09
fix: pkcs11 signing with a wrong key and a ready check when `TedgeP11Client` is created

Bravo555:
## Proposed changes

This PR has 2 parts:

1. Fix a bug where we didn't keep the handle to the key we found using the URI if it was declared, but used another search to find a new key capable of signing. Sometimes it worked, because the library could technically return the least recently used key as "first", which worked sometimes but not always. Now, within the session we keep track of the object handle in the key struct so the key is guaranteed to be correct.
2. Add a ready ready check when creating the `TedgeP11Client`. During a shared debugging session with @reubenmiller, we discovered that on certain setups loading the PKCS11 library can be unreasonably slow, so we decided to make a request that starts the server before the actual handshake to ensure the server is warm. More rationale in the method doc comment.   

## Types of changes

## Paste Link to the issue
- #3601 

## Checklist

### DIFF
--- a/crates/extensions/tedge-p11-server/src/client.rs
+++ b/crates/extensions/tedge-p11-server/src/client.rs
@@ -13,10 +13,41 @@ use super::service::SignRequest;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TedgeP11Client {
-    pub socket_path: Arc<Path>,
+    socket_path: Arc<Path>,
 }
 
 impl TedgeP11Client {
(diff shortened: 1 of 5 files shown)


reubenmiller (approved):
Approved. We tested this on the device where the initial problem was discovered and it worked as expected.

didier-wenzek (approved):
Approved

rina23q (approved):

Issue #3512, opened 2025-03-26
flaky system test: Convert child device alarm topic


reubenmiller:
**Describe the bug**

Flaky test: `Convert child device alarm topic`

```
Convert child device alarm topic                                      | FAIL |
Matching messages on topic 'te/device/child1///a/test_alarm' is less than minimum.
wanted: 1
got: 0

messages:
[]
```

**Failed builds**

* https://github.com/thin-edge/thin-edge.io/actions/runs/14085576561/job/39449192224

reubenmiller:
There is some evidence that the sometimes the retained messages are not being received by the tedge-mapper-c8y or the mqtt-logger (which is used in tests for running assertions).

* https://github.com/thin-edge/thin-edge.io/actions/runs/14390043365?pr=3557

albinsuresh:
**RCA**

The issue occurs when the broker, at times, does not deliver the alarm message to other subscribers like `tedge-agent` and even the `mqtt-logger` because sometimes the `tedge mqtt pub` client prematurely closes the connection with the broker before the `PUBREL` is even published for QoS 2 messages as seen here:

```
Mar 26 14:37:11 90db3c0d2616 mosquitto[623]: 1742999831: Received PUBLISH from tedge-pub-1176 (d0, q2, r1, m1, 'tedge/alarms/major/test_alarm/child1', ... (23 bytes))
Mar 26 14:37:11 90db3c0d2616 mosquitto[623]: 1742999831: Sending PUBREC to tedge-pub-1176 (m1, rc0)
Mar 26 14:37:11 90db3c0d2616 mosquitto[623]: 1742999831: Received DISCONNECT from tedge-pub-1176
Mar 26 14:37:11 90db3c0d2616 mosquitto[623]: 1742999831: Client tedge-pub-1176 disconnected.
```

Without receiving the `PUBREL` packet, the broker can not  publish that message to the subscribing `tedge-agent` which results in that message never getting converted as well, leading to the test failure.

For the others tests, where the alarms were properly converted, the publishing client stayed long enough, sending the `PUBREL` to the broker and receiving the `PUBCOMP` from the broker as seen here:

```
Mar 26 14:37:43 90db3c0d2616 mosquitto[623]: 1742999863: Received PUBLISH from tedge-pub-1302 (d0, q2, r1, m1, 'tedge/alarms/major/test_alarm/child2', ... (0 bytes))
[...]

Issue #1932, opened 2023-04-27
c8y-firmware-plugin timer actor does not react to a shutdown message


reubenmiller:
**Describe the bug**

The `c8y-firmware-plugin` does not restart cleanly (e.g. less than 60 seconds) when a child firmware operation is in progress.

The logs seem to indicate that the `Timer` actor does not react to the `shutdown` message. All other actors within the `c8y-firmware-plugin` actors respond (e.g. `HealthMonitorActor`, `C8YJwtRetriever`, `MQTT`, `Downloader`, `Signal-Handler` and `FirmwareManager`)

**To Reproduce**

Given the following setup

* Main device running tedge
* Child device with support for the c8y_Firmware operation

Then

1. Create a Cumulocity firmware operation for the child device

2. Whilst the operation is processing, restart the firmware plugin

    ```sh
    systemctl restart c8y-firmware-plugin
    ```

    **Note:** The restart command will take 60 seconds, and the following log message is shown in the journald logs (e.g. `journalctl -fu c8y-firmware-plugin -n 100`)

    ```
    ERROR Runtime: Timeout waiting for all actors to shutdown
    ```

**Expected behavior**

The `Timer` actor should listen to the `shutdown` message and stop the timer action so that it the actor `Runtime` does not need to forcefully shutdown the process.

**Screenshots**

Below shows the log output of the `c8y-firmware-plugin` after the `systemctl restart c8y-firmware-plugin` command was executed.
```log
Apr 27 11:17:04 1090b1469950 systemd[1]: Stopping Thin-edge device firmware management for Cumulocity...
Apr 27 11:17:04 1090b1469950 c8y-firmware-plugin[60]: 2023-04-27T11:17:04.485696536Z  INFO Signal-Handler: send Shutdown
Apr 27 11:17:04 1090b1469950 c8y-firmware-plugin[60]: 2023-04-27T11:17:04.485792703Z  INFO Runtime: Shutting down
Apr 27 11:17:04 1090b1469950 c8y-firmware-plugin[60]: 2023-04-27T11:17:04.485810953Z  INFO Runtime: Successfully sent shutdown request to HealthMonitorActor-6
Apr 27 11:17:04 1090b1469950 c8y-firmware-plugin[60]: 2023-04-27T11:17:04.485822536Z  INFO Runtime: Successfully sent shutdown request to C8YJwtRetriever-2
[...]

PR #3603, opened 2025-05-09
fix: ensure that `tedge mqtt pub` fully publishes messages sent with QoS 2

jarhodes314:
## Proposed changes
Ensure that all outstanding publishes are sent when running `tedge mqtt pub -q 2 ...`.

The root cause was the queue in rumqttc's event loop. When we attempt to publish a message, this is sent to the event loop via a channel. In the case of this bug, the event loop didn't receive this publish notification before we detected the sender loop had shut down, and thus we started the disconnection process before publishing the message. The publish would still be processed before the disconnect request, but the disconnection caused us not to respond to the broker's PubRec acknowledgement, thus the QoS 2 message wasn't fully processed.

## .Types of changes

## Paste Link to the issue
- #3512 

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -14,9 +14,11 @@ use log::warn;
 use rumqttc::AsyncClient;
 use rumqttc::Event;
 use rumqttc::EventLoop;
-use rumqttc::Incoming;
 use rumqttc::Outgoing;
 use rumqttc::Packet;
+use std::collections::HashSet;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::OwnedSemaphorePermit;
@@ -94,6 +96,7 @@ impl Connection {
             Connection::open(config, received_sender.clone(), error_sender.clone()).await?;
         let permits = Arc::new(Semaphore::new(1));
         let permit = permits.clone().acquire_owned().await.unwrap();
+        let pub_count = Arc::new(AtomicUsize::new(0));
         tokio::spawn(Connection::receiver_loop(
             mqtt_client.clone(),
             config.clone(),
@@ -102,6 +105,7 @@ impl Connection {
             error_sender.clone(),
             pub_done_sender,
             permits,
+            pub_count.clone(),
         ));
         tokio::spawn(Connection::sender_loop(
             mqtt_client,
@@ -109,6 +113,7 @@ impl Connection {
             error_sender,
             config.last_will_message.clone(),
             permit,
+            pub_count,
         ));
 
         Ok(Connection {
@@ -202,6 +207,7 @@ impl Connection {
         Ok((mqtt_client, event_loop))
     }
 
+    #[allow(clippy::too_many_arguments)]
     async fn receiver_loop(
         mqtt_client: AsyncClient,
         config: Config,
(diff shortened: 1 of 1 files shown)


crates/common/mqtt_channel/src/connection.rs
```diff
@@ -210,15 +216,18 @@ impl Connection {
...
         let mut disconnect_permit = None;
+        let mut awaiting_ack = HashSet::new();
 
         loop {
             // Check if we are ready to disconnect. Due to ownership of the
             // event loop, this needs to be done before we call
             // `event_loop.poll()`
-            let remaining_events_empty = event_loop.state.inflight() == 0;
+            let remaining_events_empty =
+                event_loop.state.inflight() == 0 && pub_count.load(Ordering::SeqCst) == 0;
```
albinsuresh, line 230: Couldn't that be just the latter?
```suggestion
            let remaining_events_empty = pub_count.load(Ordering::SeqCst) == 0;
```

crates/common/mqtt_channel/src/connection.rs
```diff
@@ -234,7 +243,12 @@ impl Connection {
                 // but will immediately be returned by `event_loop.poll()`
                 biased;
 
-                event = event_loop.poll() => event,
+                event = event_loop.poll() => {
+                    // If we receive more events, we want to reset the disconnect permit
+                    // so the event loop doesn't block disconnection
+                    disconnect_permit.take();
```
albinsuresh, line 249: Does that really help? After the permit is acquired in line no 254 during an iteration, we start the next iteration and would have triggered the disconnection in line no. 235. Since this `event_loop.poll()` happens after that, setting it back to None doesn't really help, does it? As the damage (triggering the disconnect task) has already been done.

crates/common/mqtt_channel/src/connection.rs
```diff
@@ -287,11 +301,26 @@ impl Connection {
...
+                        pub_count.fetch_sub(1, Ordering::SeqCst);
+                    }
+                    awaiting_ack.insert(p);
+                }
+
+                Ok(Event::Incoming(Packet::PubAck(p))) => {
+                    awaiting_ack.remove(&p.pkid);
+                }
+
+                Ok(Event::Incoming(Packet::PubRec(p))) => {
```
albinsuresh, line 320: Isn't it safer to remove the packet once `PUBCOMP` is received rather than PUBREC? Since that guarantees that the broker has released the message to other subscribers?
```suggestion
                Ok(Event::Incoming(Packet::PubComp(p))) => {
```

albinsuresh (commented):
Unrelated to the issue that we're fixing, but, seeing the additional handling required for `Outgoing` QoS 2 packets published by this client, I'm wondering whether we should do something similar for `Incoming` QoS 2 packets as well, to make sure that this client properly acknowledges those packets received from the broker.

crates/common/mqtt_channel/src/connection.rs
jarhodes314: No, because pub_count == 0 tells us that everything we intended to publish has been published, and inflight == 0 tells us that all the messages are acknowledged.

crates/common/mqtt_channel/src/connection.rs
jarhodes314: I thought there was a chance it might not since we definitely have cases where we acquire the disconnect permit, but then can't send the disconnect due to pending packets, but if we have pending packets that means we have to wait for event_loop.poll() to complete before we can achieve the required precondition for sending the disconnect. So thinking about it more thoroughly, I think you're correct, this is redundant.

crates/common/mqtt_channel/src/connection.rs
jarhodes314: Yes this is probably true

didier-wenzek: Actually true, as `PUBCOMP` is the QoS 2 equivalent of `PUBACK` of QoS 1. By closing on `PUBREC`, the `PUBREL` might be missing and the broker let waiting for it.

didier-wenzek (approved):
Approved.

didier-wenzek:
Rerunning the failed test. Flaky, but no so:

```
$ invoke flake-finder --test-name "Firmware plugin supports restart via service manager #1932" --iterations 100 --outputdir output_ff --clean
...
```

reubenmiller:
> Rerunning the failed test. Flaky, but no so:
> 
> ```
> $ invoke flake-finder --test-name "Firmware plugin supports restart via service manager #1932" --iterations 100 --outputdir output_ff --clean
> ...
> Results: 100 iterations, 100 passed, 0 failed
> ```

yeah i looked like it was caused by a duplicate operation being received…we could relax the assertion to not check a max message count, but not sure if that would cause downstream failures anyway…but yes all unrelated

albinsuresh (approved):
Approved.

PR #3614, opened 2025-05-13
refactor: flush after calling write_all

jarhodes314:
## Proposed changes
Improve the file handling by:
- Making sure we flush any buffers once we have written to the file
- Making sure we use a `BufWriter` to improve file performance
- For cases where we write all the contents at once (often the case in tests, use `fs::write(...)` instead of manually opening a file

There are probably many cases here where it's not strictly necessary to do a flush (anywhere where we close a file before we rely on the data being persisted), but there are definitely quite a few cases where we this assumption doesn't hold for us.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/download/src/download.rs
+++ b/crates/common/download/src/download.rs
@@ -415,6 +415,7 @@ async fn save_chunks_to_file_at(
     while let Some(bytes) = response.chunk().await? {
         writer.write_all(&bytes)?;
     }
+    writer.flush()?;
     Ok(())
 }
 
@@ -800,6 +801,7 @@ mod tests {
                         let msg = format!("{header}\r\n{size}\r\n{body}\r\n");
                         debug!("sending message = {msg}");
                         writer.write_all(msg.as_bytes()).await.unwrap();
+                        writer.flush().await.unwrap();
                     } else {
                         let header = "\
                             HTTP/1.1 200 OK\r\n\
@@ -811,6 +813,7 @@ mod tests {
                         let body = "AAAA";
                         let msg = format!("{header}\r\n4\r\n{body}\r\n");
                         writer.write_all(msg.as_bytes()).await.unwrap();
+                        writer.flush().await.unwrap();
                     }
                 };
                 tokio::spawn(response_task);
@@ -1005,6 +1008,7 @@ mod tests {
         }
 
         file.write_all(buffer.as_bytes()).unwrap();
+        file.flush().unwrap();
 
         Ok(file)
     }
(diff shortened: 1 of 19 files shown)


crates/common/tedge_utils/src/file.rs
```diff
@@ -325,6 +325,7 @@ impl PermissionEntry {
                             from: e,
                         })
                         .await?;
+                    f.flush().await?;
```
didier-wenzek, line 328: The current sequence of calls (`sync_all`, `write_all`, `flush`) has to be rewritten as  (`write_all`, `flush`, `sync_all`)

crates/common/tedge_utils/src/file.rs
```diff
@@ -359,6 +360,7 @@ pub async fn overwrite_file(file: &Path, content: &str) -> Result<(), FileError>
                     from: e,
                 })
                 .await?;
+            f.flush().await?;
```
didier-wenzek, line 363: The current sequence of calls (`sync_all`, `write_all`, `flush`) has to be rewritten as  (`write_all`, `flush`, `sync_all`)

crates/common/download/src/download.rs
```diff
@@ -415,6 +415,7 @@ async fn save_chunks_to_file_at(
     while let Some(bytes) = response.chunk().await? {
         writer.write_all(&bytes)?;
     }
+    writer.flush()?;
```
Bravo555, line 418: note: this is an example where there's no functional change, because [`flush` on a `File` is a noop](https://doc.rust-lang.org/std/fs/struct.File.html#platform-specific-behavior-14), but still it's probably good to have for consistency or if somebody added a buffer and forgot to add a `flush` call.

If we actually want to ensure that all writes reach the file, I think `sync_data` or `sync_all` might be better here.

crates/core/tedge/src/bridge/common_mosquitto_config.rs
```diff
@@ -137,6 +137,7 @@ impl CommonMosquittoConfig {
 
         self.internal_listener.write(writer).await?;
         self.external_listener.write(writer).await?;
+        writer.flush().await?;
```
Bravo555, line 140: note: if the caller calls `serialize` multiple times, then adding `flush` here means some of the flushes will be redundant. The point is to flush the writer before it's dropped, so the caller should probably do it either way, but keeping it here is also fine.

crates/common/tedge_utils/src/file.rs
Bravo555: note that [`flush` on a `File` is a noop](https://doc.rust-lang.org/std/fs/struct.File.html#platform-specific-behavior-14) so `sync_all` probably suffices, but we can keep the `flush` to be consistent.

Bravo555 (approved):
Looks alright, some `flush`es could maybe be `sync`s, but I'm not sure if it's necessary either way, so approving.

didier-wenzek (approved):
Approved

PR #3623, opened 2025-05-14
feat: Connect summary cryptoki show mode and key URI

Bravo555:
## Proposed changes

Change `cryptoki:` line in connection summary from simple `true/false` to display mode and key URI:

- If mode is off: `cryptoki: off`
- If mode is not off and `key_uri` is not set: `cryptoki: MODE`
- If mode is not off and `key_uri` is set: `cryptoki: MODE (key: KEY)`

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

I'm not sure if this is the best way to format the information in the summary, I'm open to suggestions for changes.

An alternative would be to instead of separate `cryptoki:` line, to add a `private key:` line, and put there `key_uri` and `key_path` config settings, depending whether cryptoki is enabled or not.

### DIFF
--- a/crates/core/tedge/src/cli/connect/command.rs
+++ b/crates/core/tedge/src/cli/connect/command.rs
@@ -32,6 +32,7 @@ use anyhow::bail;
 use c8y_api::http_proxy::read_c8y_credentials;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
+use certificate::parse_root_certificate::CryptokiConfig;
 use mqtt_channel::Topic;
 use std::borrow::Cow;
 use std::collections::HashMap;
@@ -100,7 +101,24 @@ impl Command for ConnectCommand {
             bridge_config(&tedge_config, &self.cloud).map_err(anyhow::Error::new)?;
         let credentials_path =
             credentials_path_for(&tedge_config, &self.cloud).map_err(anyhow::Error::new)?;
-        let use_cryptoki = tedge_config.device.cryptoki.mode.enabled();
+
+        let cloud = tedge_config
+            .as_cloud_config((&self.cloud).into())
+            .map_err(anyhow::Error::new)?;
+
+        let cryptoki_key_uri = tedge_config
+            .device
+            .cryptoki_config(cloud)?
+            .map(|c| match c {
+                CryptokiConfig::Direct(d) => d.uri,
+                CryptokiConfig::SocketService { uri, .. } => uri,
+            });
+        let cryptoki_mode = tedge_config.device.cryptoki.mode.clone();
+        let cryptoki_status = match cryptoki_key_uri {
+            None => "off".to_string(),
+            Some(None) => format!("{cryptoki_mode}"),
+            Some(Some(key)) => format!("{cryptoki_mode} (key: {key})"),
+        };
 
         ConfigLogger::log(
             self.description(),
@@ -108,7 +126,7 @@ impl Command for ConnectCommand {
             &*self.service_manager,
             &self.cloud,
             credentials_path,
-            use_cryptoki,
+            &cryptoki_status,
             tedge_config.proxy.address.or_none(),
             tedge_config.proxy.username.or_none().map(|u| u.as_str()),
         );
--- a/crates/core/tedge/src/cli/log.rs
+++ b/crates/core/tedge/src/cli/log.rs
@@ -256,7 +256,7 @@ pub struct ConfigLogger<'a> {
     mosquitto_version: Option<&'a str>,
     cloud: &'a MaybeBorrowedCloud<'a>,
     credentials_path: Option<&'a Utf8Path>,
-    cryptoki: bool,
+    cryptoki: &'a str,
     proxy_url: Option<&'a ProxyUrl>,
     proxy_username: Option<&'a str>,
 }
@@ -270,7 +270,7 @@ impl<'a> ConfigLogger<'a> {
         service_manager: &'a dyn SystemServiceManager,
         cloud: &'a MaybeBorrowedCloud<'a>,
         credentials_path: Option<&'a Utf8Path>,
-        use_cryptoki: bool,
+        cryptoki: &'a str,
         proxy_url: Option<&'a ProxyUrl>,
         proxy_username: Option<&'a str>,
     ) {
@@ -287,7 +287,7 @@ impl<'a> ConfigLogger<'a> {
                 service_manager,
                 mosquitto_version: config.mosquitto_version.as_deref(),
                 cloud,
-                cryptoki: use_cryptoki,
+                cryptoki,
                 proxy_url,
                 proxy_username,
             }
--- a/tests/RobotFramework/tests/pkcs11/private_key_storage.robot
+++ b/tests/RobotFramework/tests/pkcs11/private_key_storage.robot
@@ -5,6 +5,7 @@ Documentation       Test thin-edge.io MQTT client authentication using a Hardwar
 ...                 cryptographic tokens that will be read by thin-edge. In real production environments a dedicated
 ...                 hardware device would be used.
 
+# it would be good to explain here why we use the tedge-p11-server exclusively and not the module mode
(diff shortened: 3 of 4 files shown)


reubenmiller:
> If mode is off: cryptoki: disabled

It might make more sense to show the same value as used in the tedge.toml, e.g. `off`, `socket` or `module`...this would be more consistent, and also easier to debug (e.g. are you using a module or socket etc.)

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -103,15 +104,29 @@ impl Command for ConnectCommand {
...
+        let cryptoki_key_uri = config.device.cryptoki_config(cloud)?.map(|c| match c {
+            CryptokiConfig::Direct(d) => d.uri,
+            CryptokiConfig::SocketService { uri, .. } => uri,
+        });
+        let cryptoki_mode = config.device.cryptoki.mode.clone();
+        let cryptoki_status = match cryptoki_key_uri {
+            None => "off".to_string(),
+            Some(None) => format!("{cryptoki_mode} (key: unspecified)"),
+            Some(Some(key)) => format!("{cryptoki_mode} (key: {key})"),
+        };
```
didier-wenzek, line 121: This is working but I'm surprised by the `mode` extracted twice via two different paths (indirectly with `config.device.cryptoki_config` and then explicitly with `config.device.cryptoki.mode`). Isn't there a risk of inconsistency?

didier-wenzek (approved):
Approved, with minor comments.

reubenmiller:
@Bravo555 I noticed that we currently don't have a system test (which is run by default) for setting the `device.key_uri` (since the only place it is being used is the AWS test which is currently marked as on_demand only).

It would make sense adding a new system tests which just uses Cumulocity and sets the PKCS11 URI on the client side (and not on the tedge-p11-server)...having this test would also make it easier to see a real example of how the `tedge connect xx` info looks like in a real-world scenario.

crates/core/tedge/src/cli/log.rs
```diff
@@ -256,7 +257,7 @@ pub struct ConfigLogger<'a> {
     mosquitto_version: Option<&'a str>,
     cloud: &'a MaybeBorrowedCloud<'a>,
     credentials_path: Option<&'a Utf8Path>,
-    cryptoki: bool,
+    cryptoki: &'a dyn Display,
```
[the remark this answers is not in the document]
Bravo555: fixed in 8f4fdf0f

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -103,15 +104,29 @@ impl Command for ConnectCommand {
...
+            .map_err(anyhow::Error::new)?;
+
+        let cryptoki_key_uri = config.device.cryptoki_config(cloud)?.map(|c| match c {
+            CryptokiConfig::Direct(d) => d.uri,
+            CryptokiConfig::SocketService { uri, .. } => uri,
+        });
+        let cryptoki_mode = config.device.cryptoki.mode.clone();
+        let cryptoki_status = match cryptoki_key_uri {
+            None => "off".to_string(),
+            Some(None) => format!("{cryptoki_mode} (key: unspecified)"),
```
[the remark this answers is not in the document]
Bravo555: fixed in c057753

Bravo555:
> It would make sense adding a new system tests which just uses Cumulocity and sets the PKCS11 URI on the client side

New test added in f3ac02b8. Is it sufficient?

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -53,6 +58,26 @@ Select Private key using PKCS#11 URI
...
+Select Private key using a request URI
+    [Documentation]    Like above, we select the key using a URI, but this time we include it in a request, which means
+    ...    we can select different keys without restarting tedge-p11-server.
+
+    Set tedge-p11-server Uri    value=
+    Tedge Reconnect Should Succeed
+
+    Execute Command    softhsm2-util --init-token --free --label token10 --pin "123456" --so-pin "123456"
+    Execute Command    cmd=tedge config set device.key_uri pkcs11:token=token10
+    Tedge Reconnect Should Fail With    Failed to find a signing key
```
reubenmiller, line 72: This surprises me a little that without the `object=tedge` component of the pkcs11 uri, that they key is not found.

reubenmiller: It might be more obvious to the reader by using an incorrect token name rather than just leaving the `object=tedge`.

```
tedge config set device.key_uri pkcs11:token=token123
```

Bravo555: The logic is that if `token` attribute is present, we select that token. Only then we try to find any keys on it (in this case there are none).

The docs section mentions this although I realize now that it doesn't sound very clear:

> It is also possible to use a URI that identifies a token in settings like device.key_uri. The URI will then be used to select a token, but the key will be selected automatically, though the selected key may be wrong if there are multiple to choose from. [1]

[1]: https://thin-edge.github.io/thin-edge.io/references/hsm-support/#key-selection

i.e. after selecting the token, we automatically try to find a matching key **on that token**.

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -103,15 +103,25 @@ impl Command for ConnectCommand {
...
+        let cloud = config
+            .as_cloud_config((&self.cloud).into())
+            .map_err(anyhow::Error::new)?;
+
+        let cryptoki_mode = config.device.cryptoki.mode.clone();
+        let cryptoki_key_uri = cloud.key_uri();
+        let cryptoki_status = match cryptoki_key_uri {
+            Some(key_uri) => format!("{cryptoki_mode} (key: {key_uri})"),
+            None => format!("{cryptoki_mode}"),
+        };
```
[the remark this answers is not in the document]
Bravo555: Okay, so there is a reason why i did [this](https://github.com/thin-edge/thin-edge.io/pull/3623#discussion_r2095198037) on friday...

I forgot that `cloud.key_uri` will be `None`, when we actually use `device.key_uri`. I'll revert to the previous version which will be a tad more complicated, but it'll work.

crates/core/tedge/src/cli/connect/command.rs
Bravo555: I had to revert this because I forgot why I did that originally: `cloud.device.key_uri` is empty if we use `device.key_uri`, and `.cryptoki_config()` actually read both and applied `device.key_uri` if the one from `cloud` was `None`.

crates/core/tedge/src/cli/connect/command.rs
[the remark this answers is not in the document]
Bravo555: fixed in 6479650 and added checks if key is displayed correctly in 0dc27e1d7

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
reubenmiller: Though if there is only one key for the given token, wouldn't that be enough to find it?

Bravo555: In this case, we create a new empty token, with no keys in it, which, looking back, is indeed unnecessary, because we can just use an incorrect token name.
Switched to just using a non-existent token in a21f01bb.

reubenmiller: Yeah that makes the test easier to read

crates/core/tedge/src/cli/connect/command.rs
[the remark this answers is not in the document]
reubenmiller: Though one system test seems to be failing now.

reubenmiller (approved):
Approved. Nice improvement to make it more transparent about the settings being used.

PR #3517, opened 2025-03-28
POC: tedge wasm mapper

didier-wenzek:
## Proposed changes

A thin-edge mapper that can be extended by users with mapping rules implemented in WebAssembly.

Features:
- [x] Chain stateless filters along pipelines: `MQTT sub| filter-1 | filter-2 | ... | filter-n | MQTT pub`
- [ ] Configure filters
- [x] Implement state-full filters (deferring messages up to the end of a time window)

Criteria:
- [x] Use Rust to implement filters
- [x] Use javascript, python or go to implement filters
- [ ] Dev tooling ?
- [ ] Performance? Memory footprint?
- [ ] Debuggability? 

Examples:

- [x] Add missing timestamps
- [ ]  Discard measurements with timestamp out of the expected range (too old or in the future)
- [x] Translate thin-edge json measurment into cumulocity json
- [ ] Add measurement units - read from topic metadata
- [x] Translate Collectd data into thin-edge JSON
- [x] Group messages received during the same time window
- [ ] Compute 5-minute rolling average

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io-roadmap/issues/86

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s. You can activate automatic signing by running `just prepare-dev` once)
- [x] I ran `just format` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I used `just check` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #3626, opened 2025-05-15
poc: Generic Mapper using JavaScript

didier-wenzek:
## Proposed changes

This POC explores an alternative to https://github.com/thin-edge/thin-edge.io/pull/3517

A thin-edge mapper that can be extended by users with mapping rules implemented in __JavaScript__.

Features:
- [x] Chain stateless filters along pipelines: `MQTT sub| filter-1 | filter-2 | ... | filter-n | MQTT pub`
- [x] Configure filters
- [x] Update filter config using MQTT 
- [x] Implement state-full filters (deferring messages up to the end of a time window)
- [ ] Combine user-provided JS filters with built-in Rust filters

Criteria:
- [ ] Expressive power?
- [ ] Dev tooling ?
- [ ] Performance? Memory footprint?
- [ ] Debuggability? 

Examples:

- [x] Add missing timestamps
- [x]  Discard measurements with timestamp out of the expected range (too old or in the future)
- [x] Translate thin-edge json measurment into cumulocity json
- [x] Add measurement units - read from topic metadata
- [x] Translate Collectd data into thin-edge JSON
- [ ] Group messages received during the same time window
- [ ] Compute 5-minute rolling average

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge_gen_mapper/src/actor.rs
```diff
@@ -34,11 +41,26 @@ impl Actor for GenMapper {
...
+                                continue;
+                            };
+                            if matches!(path.extension(), Some("js" | "ts")) {
+                                self.reload_filter(path).await;
+                            } else if path.extension() == Some("toml") {
+                                self.reload_toml(path).await;
+                            }
+                        },
+                        Some(InputMessage::FsWatchEvent(e)) => {
+                            tracing::warn!("TODO do something with {e:?}")
```
didier-wenzek, line 64: We can definitely ignore anything that is is related to a file with an unknown extension (such as `.collectd.toml.swp` and `collectd.toml~`).

- Deletion of a pipeline should be okay, unless if we try to unsubscribe to the input topics.
- Deletion of a script is more problematic, if still used by one of the pipeline.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -120,22 +209,60 @@ impl FromPeers {
     async fn relay_messages_to(
         &mut self,
         outgoing_mqtt: &mut mpsc::UnboundedSender<MqttMessage>,
+        client: mqtt_channel::AsyncClient,
+        subscriptions: Arc<Mutex<TopicFilter>>,
```
didier-wenzek, line 213: I would combine these two fields of an `mqtt_chanel::Connection`  in a single one (say a  `subscriber`)  that handles the whole logic below to compute the new subscription set and update `mqtt_channel::AsyncClient`.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -120,22 +209,60 @@ impl FromPeers {
...
+                    for o in overlap {
+                        if request.0.remove(&o).is_none() {
+                            to_unsub.add_unchecked(&o);
+                        }
+                    }
+                    let to_sub = request.0;
+                    let to_sub = to_sub.filters();
+                    if !to_sub.is_empty() {
+                        info!("Updating MQTT subscription to include {to_sub:?}");
+                        client.subscribe_many(to_sub).await.unwrap();
```
didier-wenzek, line 239: > NB This doesn't currently work since dynamic subscriptions aren't hooked up fully. tedge_mqtt_ext needs some further modifications to support sending the received messages on newly subscribed topics to the actor that requested the subscription

Okay,  I understand this comment now. The MQTT subscription is correctly updated *but* the messages are not forwarded by this actor which `ToPeers::peer_senders` subscriptions list has not been updated.

Indeed, not easy to update  from the `FromPeers` event loop this `peer_senders` list which is owned by the `ToPeers` event loop. A new channel, would be not enough as these peer `Sender`s are anonymous. Which one would have to be updated?

didier-wenzek: Here is a proposal to fix that: https://github.com/thin-edge/thin-edge.io/pull/3626/commits/43932cf2250b37807ce77e3408ad8c35029e240c

PR #3631, opened 2025-05-19
feat: PKCS11 support EC SECP384R1 and SECP521R1 keys

Bravo555:
## TODO
- [x] tests
- [x] impl

## Proposed changes

Currently we support only EC keys using curve SECP256R1, as during initial development we confirmed that C8y, Azure and AWS all offered `ecdsa_secp256r1_sha256` sigscheme. However, if that ever changes, or if a user simply wishes to use a bigger key, it is impossible to connect. Therefore we should support the other sizes of EC keys. 

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
<br/>

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s. You can activate automatic signing by running `just prepare-dev` once)
- [x] I ran `just format` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `just check` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #3635, opened 2025-05-21
refactor: Use one SigningKey struct

Bravo555:
## Proposed changes

This is a small refactor extracted from #3631 and made into a separate PR to get a cleaner diff and for easier comparison of code required to support EC keys to code required to support RSA keys, once RSA PR is made.

The main improvements of this refactor are:
- Remove separate structs for EC keys and RSA keys, use a single struct that contains a sigscheme to be used
- create a struct for sigschemes we support, so it's easier to see at a glance what's supported and extraneous unwraps/checks can be removed because we rely on the type system more
- remove duplicate attribute query when signing, use a sigscheme we saved when creating the signing key struct

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments
Previously the behaviour was a little bit wrong because if we got an EC key we put 3 different signature schemes as "supported" for the purposes of choosing a scheme from those offered by the server, but that was wrong because for EC keys, each type of EC key can only do one scheme - P256 can do SHA256, P384 can do SHA384, P521 can do SHA512, so previously we could've theoretically had P256 key and select a SHA512 sigscheme, which was wrong and signing would just fail later, but now we only set a scheme we can actually use and fail earlier

### DIFF
--- a/crates/extensions/tedge-p11-server/src/pkcs11.rs
+++ b/crates/extensions/tedge-p11-server/src/pkcs11.rs
@@ -147,15 +147,32 @@ impl Cryptoki {
         let pkcs11 = PKCS11 {
             session: Arc::new(Mutex::new(session)),
         };
+
+        // we need to select a signature scheme to use with a key - each type of key can only have one signature scheme
+        // ideally we'd simply get a cryptoki mechanism that corresponds to this sigscheme but it's not possible;
+        // instead we have to manually parse additional attributes to select a proper sigscheme; currently don't do it
+        // and just select the most common sigscheme for both types of keys
+
+        // NOTE: cryptoki has AttributeType::AllowedMechanisms, but when i use it in get_attributes() with opensc-pkcs11
+        // module it gets ignored (not present or supported) and with softhsm2 module it panics(seems to be an issue
+        // with cryptoki, but regardless):
+
+        // thread 'main' panicked at library/core/src/panicking.rs:218:5:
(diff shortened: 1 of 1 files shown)


crates/extensions/tedge-p11-server/src/pkcs11.rs
```diff
@@ -194,25 +211,32 @@ impl Cryptoki {
...
+    sigscheme: SigScheme,
 }
 
 impl SigningKey for Pkcs11SigningKey {
     fn choose_scheme(&self, offered: &[SignatureScheme]) -> Option<Box<dyn Signer>> {
-        match self {
-            Self::Ecdsa(e) => e.choose_scheme(offered),
-            Self::Rsa(r) => r.choose_scheme(offered),
+        debug!("Offered signature schemes. offered={:?}", offered);
+        // safe: we only create RSA and ECDSA SigningKeys
```
jarhodes314, line 224: I don't understand what this is referring to. None of the code indicates to me it's potentially going to do anything unsafe/panic.

Bravo555: Oh, my bad, missed that one. Previously there was an unwrap somewhere here because I converted between `rustls::SignatureScheme` and `cryptoki::MechanismType` and these functions were partial, but later on I added `SigScheme` which removed those unwraps. Going to remove that comment.

didier-wenzek (approved):
Approved

Issue #3642, opened 2025-05-26
regression: setting a value via an environment variable results in a deprecation warning


reubenmiller:
**Describe the bug**

tedge prints unexpected warnings when using environment variables to control the tedge configuration. It seems to be a regression which occurred some time after 1.5.1 (but 1.5.1 is not affected). It seems to occurs for most configuration variable names.

Below shows a comparison with the latest main branch version and the last known ok version (though I didn't check for the exact commit which broke it).

**tedge 1.5.2~56+g2609c72 (ERROR)**

```sh
$ tedge --version
tedge 1.5.2~56+g2609c72

$ TEDGE_MQTT_CLIENT_HOST=1.2.3.4 tedge config get mqtt.client.host
2025-05-26T17:22:22.238019Z  WARN tedge_config::tedge_toml::tedge_config: The key 'mqtt_client_host' is deprecated. Use 'mqtt.client.host' instead.
1.2.3.4
```

**tedge 1.5.1 (OK)**

```sh
$ tedge --version
tedge 1.5.1

$ TEDGE_MQTT_CLIENT_HOST=1.2.3.4 tedge config get mqtt.client.host
1.2.3.4
```

**To Reproduce**

1. Install the latest version from the main branch

2. Set any tedge toml value via an environment variable then get it using `tedge config get`

    ```
    TEDGE_MQTT_CLIENT_HOST=1.2.3.4 tedge config get mqtt.client.host
    ```

**Expected behavior**

Setting a valid tedge configuration key name via an environment variable should not result in a warning message being printed.

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`MacOS`|
|Hardware [incl. revision]|`unknown`|
|System-Architecture|`arm64`|
|thin-edge.io version|`tedge 1.5.2~56+g2609c72`|

**Additional context**

reubenmiller:
The fix works as expected.

PR #3643, opened 2025-05-27
fix: setting a value via an environment variable results in a deprecation warning

reubenmiller:
## Proposed changes

TODO:

* [x] Update system test to provide coverage for the bug
* [x] Implementation

Fix deprecation false positives when setting tedge config values using environment variables.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3642

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/tedge/tedge_config_get.robot
+++ b/tests/RobotFramework/tests/tedge/tedge_config_get.robot
@@ -53,6 +53,16 @@ Set unknown configuration via environment variables
     ...    ${stderr}
     ...    Unknown configuration field "c8y_unknown_configuration" from environment variable TEDGE_C8Y_UNKNOWN_CONFIGURATION
 
+Read deprecated key
+    ${stdout}    ${stderr}=    Execute Command
+    ...    cmd=tedge config get mqtt.external.capath
+    ...    stdout=${True}
+    ...    stderr=${True}
+    ...    exp_exit_code=!0
+    Should Be Empty    ${stdout}
+    Should Contain    ${stderr}    The key 'mqtt.external.capath' is deprecated. Use 'mqtt.external.ca_path' instead.
+    Should Contain    ${stderr}    The provided config key: 'mqtt.external.ca_path' is not set
+
 Normalize paths configured with tedge config
     Execute Command    tedge config set c8y.device.csr_path c8y-device.csr
     ${path}=    Execute Command    tedge config get c8y.device.csr_path
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4445,6 +4445,7 @@ dependencies = [
  "tokio",
  "toml 0.8.8",
  "tracing",
+ "tracing-subscriber",
  "url",
  "which",
  "x509-parser 0.16.0",
--- a/crates/common/tedge_config/src/system_toml/log_level.rs
+++ b/crates/common/tedge_config/src/system_toml/log_level.rs
@@ -1,10 +1,21 @@
-use camino::Utf8Path;
-
 use super::SystemConfig;
 use super::SystemTomlError;
 use crate::cli::LogConfigArgs;
+use camino::Utf8Path;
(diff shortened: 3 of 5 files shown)


reubenmiller (commented):
Approved. Thanks for fixing the system test that I added

tests/RobotFramework/tests/tedge/tedge_config_get.robot
```diff
@@ -66,10 +66,20 @@ Custom Setup
...
 Check known tedge environment settings
     [Arguments]    ${ENV_NAME}    ${KEY_NAME}    ${VALUE}
-    ${output}=    Execute Command    cmd=env ${ENV_NAME}=${VALUE} tedge config get ${KEY_NAME}
-    Should Be Equal    ${output}    ${VALUE}\n
+    ${stdout}    ${stderr}=    Execute Command
+    ...    cmd=env ${ENV_NAME}=${VALUE} tedge config get ${KEY_NAME}
+    ...    stdout=${True}
+    ...    stderr=${True}
+    ...    retries=1
+    Should Be Equal    ${stdout}    ${VALUE}\n
```
Bravo555, line 74: nit: I think `Execute Command    strip=True` could be used here to remove that extra `\n`

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
```diff
@@ -346,6 +344,14 @@ fn update_with_environment_variables(
     Ok(())
 }
 
+fn parse_key_without_warnings(tedge_key: &str) -> Result<WritableKey, ParseKeyError> {
+    tracing::subscriber::with_default(NoSubscriber::new(), || {
+        tedge_key
+            .to_ascii_lowercase()
+            .parse::<tedge_config::WritableKey>()
+    })
```
Bravo555, line 352: question: is the warning ultimately emitted by figment and that's why we can't just not emit it in the first place?

`WritableKey` parse impl is emitted by a macro, so I can't seem to find it and confirm it myself.

Bravo555 (approved):
Small notes, but the fix is working, so approving.

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
```diff
@@ -346,6 +344,14 @@ fn update_with_environment_variables(
...
+}
```
albinsuresh, line 353: I didn't fully understand this fix, as it seems to ignore all warnings while reading the provided key: not just unknown ones, but even the ones that we have actually deprecated in the past. What am I missing here? Aren't there some old deprecated keys for which we want the deprecation warning to be emitted?

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
jarhodes314: https://github.com/thin-edge/thin-edge.io/blob/1ff8cf459af0ba4a463c460774b9e968abd1f5ef/crates/common/tedge_config_macros/impl/src/query.rs#L1109-L1126 shows what this expands to in a simple case

tests/RobotFramework/tests/tedge/tedge_config_get.robot
reubenmiller: Though we're checking the exact output here, so we want to be sure that we're only outputting multiple newline chars (which would be masked if using the strip option)

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
jarhodes314: The deprecation warning is shown when running e.g. `tedge config get mqtt.external.certfile` (or it is as of 64dc06e, turns out there was also a bug in logging the deprecated keys when running `tedge config ...`), so it is useful.

albinsuresh (approved):
LGTM.

PR #3645, opened 2025-05-27
tools: Remove obsolete clippy disallowed rules

Bravo555:
## Proposed changes

Remove rules from clippy.toml for types and methods that were unreachable and were being marked as warnings by clippy from nightly-2025-05-25.

On April 1, a PR was merged into clippy to check if paths in `disallowed_*` configurations are actually reachable. In our case, some rules relating to hyper and reqwest::blocking were outdated probably due to upgrades of versions of these crates.

This commit removes the outdated rules, but alternatively we could set `allow-invalid` field to silence the warnings but keep the rules, e.g. if we expect to add `reqwest::blocking` back in the future.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


jarhodes314 (approved):
This looks good to me. `reqwest::blocking` is no longer enabled due to the `tedge` CLI refactoring, and `hyper::client` doesn't exist as of `hyper` 1.0. `hyper-util` has a replacement for this, but doesn't have the `Client::new` method, so we don't need to replace this rule.

PR #3650, opened 2025-05-28
feat: tedge flows, a configurable mapper using JavaScript

didier-wenzek:
## Proposed changes

A first step toward configurable mappers that users can extend and adapt to their use-cases
with their own filtering and message transformation rules.

The following steps are described here: https://github.com/thin-edge/thin-edge.io/issues/3756

Features:

- [x] Define the behavior of a mapper as a combination of *flows*, *steps* and transformation *scripts*
- [x] Consume, transform and produce MQTT messages.
- [x] Use JavaScript to define step functions 
- [x] Define flows as chain of transformation *steps*:  `MQTT sub| step-1 |step-2 | ... | step-n | MQTT pub`
- [x] Configure flows and steps
- [x] Dynamically update step config using MQTT 
- [x] Implement state-full filters (deferring messages up to the end of a time window)
- [ ] ~Combine user-provided JS filters with built-in Rust filters~

Criteria:
- [ ] Expressive power?
- [ ] Dev tooling ?
- [ ] Performance? Memory footprint?
- [ ] Debuggability? 

Examples:

- [x] Add missing timestamps
- [x]  Discard measurements with timestamp out of the expected range (too old or in the future)
- [x] Translate thin-edge json measurment into cumulocity json
- [x] Add measurement units - read from topic metadata
- [x] Translate Collectd data into thin-edge JSON
- [x] Group messages received during the same time window
- [x] Compute 5-minute rolling average, min, max
- [x] A circuit breaker filter detecting that too many messages are being sent within a given period
- [x] See also [real use-case examples](https://github.com/thin-edge/tedge-mapper-generic-examples)

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3756

## Checklist

## Further comments

This PR uses [quickJS](https://github.com/DelSkayn/rquickjs)  as an alternative to:

- WASM gen mapper https://github.com/thin-edge/thin-edge.io/pull/3517
- RustyScript gen mapper https://github.com/thin-edge/thin-edge.io/pull/3626

### DIFF
--- a/design/decisions/0006-tedge-flows.md
+++ b/design/decisions/0006-tedge-flows.md
@@ -1,13 +1,13 @@
 # Extensible mapper and user-provided flows
 
-* Date: __2025-07-1-__
-* Status: __New__
+* Date: __2025-08-08__
+* Status: __Approved__
 
 ## Motivation
 
 In theory, %%te%% users can implement customized mappers to transform data published on the MQTT bus
 or to interact with the cloud. In practice, they don't.
-Implementing a mapper is costly while what is provided out-the-box by %%te%% already meets most requirements.
+Implementing a mapper is costly, while what is provided out-the-box by %%te%% already meets most requirements.
 The need is not to write new mappers but to adapt existing ones.
 
 The aim of the extensible mapper it to let users extend and adapt the mappers with their own filtering and mapping rules,
@@ -18,10 +18,14 @@ leveraging the core mapping rules and mapper mechanisms (bridge connections, HTT
 The %%te%% mappers for Cumulocity, Azure, AWS and Collectd are implemented on top of an extensible mapper
 which is used to drive all MQTT message transformations.
 - Transformations are implemented by flows which consume MQTT messages, apply a sequence of transformation steps and produce MQTT messages.
+  - a flow is composed of multiple steps
   - `MQTT sub| step-1 | step-2 | ... | step-n | MQTT pub`
 - A flow can combine builtin and user-provided steps.
-- The user can configure all the transformations used by a mapper,
-  editing MQTT sources, flows, steps and MQTT sinks.
+- Flows can be deployed by users
+  - The users can configure all the transformations used by a mapper, editing MQTT sources, flows, steps and MQTT sinks.
+  - Flows can be deployed multiple times.
+  - For instance, given the flow "log-surge", it can be deployed as multiple instances of the same flow,
+    where each instance can be customized.
 - By contrast with the current implementation, where the translation of measurements from %%te%% JSON to Cumulocity JSON
   is fully hard-coded, with the generic mapper a user can re-use the core of this transformation while adding customized steps:
   - consuming measurement from a non-standard topic
@@ -51,28 +55,27 @@ steps = [
 ]
 ```
 
-- A flow script has to export at least one `process` function.
-  - `process(t: Timestamp, msg: Message, config: Json) -> Vec<Message>` 
+- A flow script has to export at least one `onMessage` function.
+  - `onMessage(message: Message, config: object): null | Message | Message[]` 
   - This function is called for each message to be transformed
   - The arguments passed to the function are:
-    - The current time as `{ seconds: u64, nanoseconds: u32 }` 
-    - The message `{ topic: string, payload: string }`
+    - The message `{ topic: string, payload: string, timestamp: { seconds: u64, nanoseconds: u32 } }`
     - The config as read from the flow config or updated by the script
   - The function is expected to return zero, one or many transformed messages `[{ topic: string, payload: string }]`
   - An exception can be thrown if the input message cannot be transformed.
-- A flow script can also export an `update_config` function
+- A flow script can also export an `onConfigUpdate` function
   - This function is called on each message received on the `meta_topics` as defined in the config.
   - The arguments are:
     - The message to be interpreted as a config update `{ topic: string, payload: string }`
     - The current config
    - The returned value (an arbitrary JSON value) is then used as the new config for the flow script.
-- A flow script can also export a `tick` function
+- A flow script can also export an `onInterval` function
   - This function is called at a regular pace with the current time and config.
   - The flow script can then return zero, one or many transformed messages
-  - By sharing an internal state between the `process` and `tick` functions,
+  - By sharing an internal state between the `onMessage` and `onInterval` functions,
     the flow script can implement aggregations over a time window.
-    When messages are received they are pushed by the `process` function into that state
-    and the final outcome is extracted by the `tick` function at the end of the time window.
+    When messages are received they are pushed by the `onMessage` function into that state
+    and the final outcome is extracted by the `onInterval` function at the end of the time window.
 
 ## First release
 
@@ -97,6 +100,71 @@ To keep things simple for the first release, the following questions are deferre
 - To handle operations, would the plugins be expanded to do more complex things like HTTP calls, file-system interactions, etc.? 
 - What are the pros and cons to persist flow step states?
 - Split a flow, forwarding transformed messages to different flows for further processing
+
+## Discussion
+
+### Naming
+
+Why flow?
+
+- Simple language for beginner users
+- Similar terminology to NodeRED.
+  Hopefully this will lead to a quick understanding rather than confusion with NodeRED,
+  but essentially they are doing the same thing.
+- Aligning (mostly) to terminology used by IDP project, though technically they might be using "flow logic"
+
+### Function names and configuration
+
+**Option 1 (selected): Add timestamp to message object**
+
+```
+function process(message, config) {
+  console.log("Processing message": {
+    payload: message.payload,
+    timestamp: message.timestamp,
+    config,
+  });
+}
+```
+
+* disadvantages - How to communicate that it isn't the timestamp of the runner
+  and not necessarily the timestamp of the message (as that would be part of the message's payload),
+  though `mosquitto_sub -t '#' -F "%J"` does this.
+
+
+**Option 2: Add timestamp at the end**
+
+```
+function process(message, {
+  output_topic = "foo/bar",
+  clamp_min = 10,
+  clamp_max = 20,
+}, timestamp) {
+console.log("Processing message using config", {
+    output_topic,
+    clamp_min,
+    clamp_max,
+  });
+}
+```
+
+* disadvantages - harder to deconstruct configuration in the function signature, which makes it slightly less readable
+
+**Option 3: timestamp at second position**
+
+```
+function process(message, _timestamp, {
+  output_topic = "foo/bar",
+  clamp_min = 10,
+  clamp_max = 20,
+}) {
+  console.log("Processing message using config", {
+    output_topic,
+    clamp_min,
+    clamp_max,
+  });
+}
+```
 
 ### API
 
@@ -134,6 +202,30 @@ Other ideas to explore to make the API more flexible:
 - Allow a flow to subscribe to topics related to the device/entity it is running on
 - Feed flow scripts with message excerpts as done for the workflows
 
+### Context
+
+The current API tries to free the flow scripts from state managements and callbacks.
+This is done by passing around state and context data.
+For instance the `update_config` function is given a config value and is expected to return an updated config.
+Along the same lines, a current `timestamp` is passed to all functions, instead of `time.now()` callback.
+
+This works. But is this the more natural way of working in Javascript?
+Also, this quickly leads to complicated function signatures when we want to provide a context or persist a flow state.
+And not easy to extend in the future without breaking function signatures.
+Last but not least, the current API forces the flow scripts to have a static state,
+which is not easy to persist (even if not done for the first version, this will be likely requested).
+
+What about the approach proposed by NodeRed to [work with context](https://nodered.org/docs/user-guide/context)?
+Thin-edge could provide 3 context levels:
+
+- `context.step`: only visible to the flow step that set the value
+- `context.flow`: visible to all steps on the same flow
+- `context.device` or `context.entities`:  visible to all flows, notably populated with device and child device internal ids.
+
+However, by contrast with NodeRed, these contexts could be regular JSON values.
+We still have to check that a script can update sub-fields in these values.
+If not we will have to provide a `set` method as done by NodeRed.
+
 ### Devops tools
 
 The flexibility to customize MQTT message processing with user-provided functions comes with risks:
@@ -142,16 +234,16 @@ The flexibility to customize MQTT message processing with user-provided function
 - builtin flows might be accidentally disconnected or broken
 - a step might introduce a performance bottleneck.
 
-To help mitigating these risks, the `tedge mapping` sub-commands provide the tools to test, monitor and debug steps and flows.
+To help mitigating these risks, the `tedge flows` sub-commands provide the tools to test, monitor and debug steps and flows.
 
-- `tedge mapping list [topic]` displays flows and steps messages received on this topic will flow through
+- `tedge flows list [topic]` displays flows and steps messages received on this topic will flow through
   - can be used with a set of flows not configured yet for a mapper
-- `tedge mapping test [flow]` feeds a step or flow with input messages and produces the transformed output messages
+- `tedge flows test [flow]` feeds a step or flow with input messages and produces the transformed output messages
   - allow users to run an assertion based on the input/output of a flow
-  - ability to pipe `tedge mqtt sub` and `tedge mapping test`
+  - ability to pipe `tedge mqtt sub` and `tedge flows test`
   - control of the timestamps
   - test aggregation over ticks
   - can be used with a set of flows not configured yet for a mapper
-- `tedge mapping stats [flow]` returns statistics on the messages processed by a flow
+- `tedge flows stats [flow]` returns statistics on the messages processed by a flow
   - count message in, message out
   - processing time min, median, max for each flow and step
--- a/docs/src/references/mappers/flows.md
+++ b/docs/src/references/mappers/flows.md
@@ -28,13 +28,13 @@ which rule how to consume, transform and produce MQTT messages.
 - A *step* function transforms one input message into zero, one or more output messages.
   - Steps are effect-free functions, with no access to MQTT, HTTP or the file-system.
   - The focus is on message transformation, format conversion, content extraction and completion as well as filtering and redacting.
-- A *connector* is used by the mapper to consume messages from and produce messages to.
+- A *connector* is used by the mapper to consume messages from a source and produce messages to a sink.
   - MQTT is the primary message source and target, but overtime others can be added.
   - Connectors can be seen as streams of messages all with the same shape so they can be processed by any step.
 - A *flow* applies a chain of transformation *steps* to input messages producing fully processed output messages.
   - The *flows* put things in motion, actually interacting with the system, consuming and producing messages.
   - Messages received on a flow are passed to the first step; and the transformed messages, if any,
-    are pushed to the subsequent steps upto the output connector.
+    are pushed to the subsequent steps up to the output connector.
 - A flow can combine builtin and user-provided steps.
   - Builtin steps provide generic building blocks such as %%te%% JSON translation into Cumulocity JSON.
   - Users can implement specific steps using JavaScript or TypeScript to refine transformations to their use cases. 
@@ -53,11 +53,11 @@ which rule how to consume, transform and produce MQTT messages.
 
 ## Step API
 
-A transformation *scripts* is a JavaScript or TypeScript module that exports:
+A transformation *script* is a JavaScript or TypeScript module that exports:
 
-- at least a function `onMessage()`, aimed to transform one input message into zero, one or more output messages,
-- possibly a function `onInterval()`, called at regular intervals to produce aggregated messages,
-- possibly a function `onConfigUpdate()`, used to update the step config.
+- at least, a function `onMessage()`, aimed to transform one input message into zero, one or more output messages,
+- possibly, a function `onInterval()`, called at regular intervals to produce aggregated messages,
+- possibly, a function `onConfigUpdate()`, used to update the step config.
 
 ```ts
 interface FlowStep {
@@ -90,6 +90,25 @@ type Timestamp = {
 A `config` is an object freely defined by the step module, to provide default values such as thresholds, durations or units.
 These values are configured by the flow and can be dynamically updated on reception of config update messages.
 
+The `onMessage` function is called for each message to be transformed
+  - The arguments passed to the function are:
+    - The message `{ topic: string, payload: string, timestamp: { seconds: u64, nanoseconds: u32 } }`
+    - The config as read from the flow config or updated by the script
+  - The function is expected to return zero, one or many transformed messages `[{ topic: string, payload: string }]`
+  - An exception can be thrown if the input message cannot be transformed.
+- If defined and associated in the step config with `meta_topics`, the `onConfigUpdate` function is called on each message received on these `meta_topics`.
+  - The arguments are:
+    - The message to be interpreted as a config update `{ topic: string, payload: string }`
+    - The current config
+  - The returned value (an arbitrary JSON value) is then used as the new config for the flow script.
+- A flow script can also export a `onInterval` function
+  - This function is called at a regular pace with the current time and config.
+  - The flow script can then return zero, one or many transformed messages
+  - By sharing an internal state between the `onMessage` and `onInterval` functions,
+    the flow script can implement aggregations over a time window.
+    When messages are received they are pushed by the `onMessage` function into that state
+    and the final outcome is extracted by the `onInterval` function at the end of the time window.
+
 ## Flow configuration
 
 - The generic mapper loads flows and steps stored in `/etc/tedge/flows/`.
@@ -97,10 +116,12 @@ These values are configured by the flow and can be dynamically updated on recept
 - A step is defined by a JavaScript file with `.js` extension.
   - This can also be a TypeScript module with a `.ts` extension.
 - The definition of flows must provide a list of MQTT topics to subscribe to.
-  - The flow will be feed with all the messages received on these topics.
+  - The flow will be fed with all the messages received on these topics.
 - A flow definition provides a list of steps.
   - Each step is built from a javascript and is possibly given a config (arbitrary json that will be passed to the script)
-  - Each step can also subscribe to a list of MQTT topics (which messages will be passed to the script to update its config)
+  - Each step can also subscribe to a list of MQTT meta topics where the metadata about the actual data message is stored
+    (e.g, meta topic of a measurement type where its units threshold values are defined).
+    The messages received on these topics will be passed to the `onConfigUpdate` letting the script update its config.
 
 ```toml
 input.mqtt.topics = ["te/+/+/+/+/m/+"]
@@ -112,26 +133,57 @@ steps = [
 ]
 ```
 
-## POC API
+## %%te%% flow mapper
 
-- A flow script has to export at least one `onMessage` function.
-  - `onMessage(msg: Message, config: Json) -> Vec<Message>` 
-  - This function is called for each message to be transformed
-  - The arguments passed to the function are:
-    - The message `{ topic: string, payload: string, timestamp: { seconds: u64, nanoseconds: u32 } }`
-    - The config as read from the flow config or updated by the script
-  - The function is expected to return zero, one or many transformed messages `[{ topic: string, payload: string }]`
-  - An exception can be thrown if the input message cannot be transformed.
-- A flow script can also export an `onConfigUpdate` function
-  - This function is called on each message received on the `meta_topics` as defined in the config.
-  - The arguments are:
-    - The message to be interpreted as a config update `{ topic: string, payload: string }`
-    - The current config
-   - The returned value (an arbitrary JSON value) is then used as the new config for the flow script.
-- A flow script can also export a `onInterval` function
-  - This function is called at a regular pace with the current time and config.
-  - The flow script can then return zero, one or many transformed messages
-  - By sharing an internal state between the `onMessage` and `onInterval` functions,
-    the flow script can implement aggregations over a time window.
-    When messages are received they are pushed by the `onMessage` function into that state
-    and the final outcome is extracted by the `onInterval` function at the end of the time window.
+The extensible mapper is launched as a regular mapper:
+
+```shell
+tedge-mapper flows
+```
+
+This mapper:
+
+- loads all the flows defined in `/etc/tedge/flows`
+- reloads any flow or script that is created, updated or deleted while the mapper is running
+- subscribes to each flow `input.mqtt.topics`, dispatching the messages to the `onMessage` functions
+- subscribes to each step `meta_topics`, dispatching the messages to the `onConfigUpdate` functions
+- triggers at the configured pace the `onInterval` functions
+- publishes memory usage statistics
+- publishes flows and steps usage statistics
+
+## %%te%% flow cli
+
+Flows and steps can be tested using the `tedge flows test` command.
+- These tests are done without any interaction with MQTT and `tedge-mapper flows`,
+  meaning that tests can safely be run on a device in production
+- By default, tests run against the flows and scripts used by `tedge-mapper flows`.
+  However, a directory of flows under development can be provided using the `--flows-dir <FLOWS_DIR>` option.
+- A test can be specific to a flow or script using the `--flow <OPTION>` option.
+ 
+A test can be given a test message on the command line.
+
+```shell
+$ tedge flows test te/device/main///m/environment '{ "temperature": 29 }'
+
+[c8y/measurement/measurements/create] {"type":"environment","temperature":{"temperature":29},"time":"2025-08-07T12:47:26.152Z"}
+```
+
+Alternatively a test can be given a sequence of messages via its stdin.
+
+```shell
+$ echo '[te/device/main///m/environment]' '{ "temperature": 29 }' | tedge flows test
+
+[c8y/measurement/measurements/create] {"type":"environment","temperature":{"temperature":29},"time":"2025-08-07T12:47:26.152Z"}
+```
+
+__Note__ that when the input of a test is received from its stdin,
+the topic is given using a bracket syntax `[<TOPIC>] <PAYLOAD>`
+similar to the output of `tedge mqtt sub` and `tedge flows test` itself.
+
+This can be used to chain tests:
+
+```shell
+$ tedge flows test collectd/mandarine/cpu/percent-active '1754571280.572:2.07156308851224' | tedge flows test
+
+[c8y/measurement/measurements/create] {"type":"collectd","time":"2025-08-07T12:54:40.572Z","cpu":{"percent-active":2.07156308851224}}
+```


docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,121 @@
...
+  - Each stage is built from a javascript and is possibly given a config (arbitrary json that will be passed to the script)
+  - Each stage can also subscribe to a list of MQTT topics (which messages will be passed to the script to update its config)
+
+```toml
+input_topics = ["te/+/+/+/+/m/+"]
+
+stages = [
+    { filter = "add_timestamp.js" },
+    { filter = "drop_stragglers.js", config = { max_delay = 60 } },
+    { filter = "te_to_c8y.js", meta_topics = ["te/+/+/+/+/m/+/meta"] }
```
albinsuresh: Wondering if stage-wise `meta_topics` is really necessary or a pipe-line level `meta_topics` would be sufficient (and simpler to maintain).

albinsuresh: Expressing the correlation between the `input_topic` and `meta_topic` with wildcards appears slightly inaccurate/misleading. For example, when the above definition is applied on a `te/+/+/+/+/m/temperature` measurement, the corresponding `meta_topic` should have only been `te/+/+/+/+/m/temperature/meta`. But with the wildcard `te/+/+/+/+/m/+/meta` subscription, it covers the meta topics of other measurements like `m/pressure/meta`, `m/humidity/meta` etc as well. Wondering if we should introduce some sort of variable substitution syntax like `m/{m_type}` and `m/{m_type}/meta` so that this correlation is better expressed.

docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,121 @@
...
+- `process(t: Timestamp, msg: Message, config: Json) -> Vec<Message>`
+- `tick(t: Timestamp) -> Vec<Message>`
+- `update_config(msg: Message, config: Json) -> Json`
+
+An alternative is to let the user implement more specific functions with simpler type signatures:
+
+- `filter(msg: Message, config: Json) -> bool`
+- `map(msg: Message, config: Json) -> Message`
+- `filter_map(msg: Message, config: Json) -> Option<Message>`
+- `flat_map(msg: Message, config: Json) -> Vec<Message>`
```
albinsuresh: I prefer these APIs with these well-known names/signatures as that makes the pipeline definition easier to comprehend. I believe it also makes it easier to expand this API set in future beyond the signature of the `process` function.

docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,121 @@
...
+- A pipeline is defined by a TOML file with `.toml` extension.
+- A filter is defined by a Javascript file with `.js` extension.
+- The definition of pipeline must provide a list of MQTT topics to subscribe to.
+  - The pipeline will be feed with all the messages received on these topics.
+- A pipeline definition also provides a list of stages.
+  - Each stage is built from a javascript and is possibly given a config (arbitrary json that will be passed to the script)
+  - Each stage can also subscribe to a list of MQTT topics (which messages will be passed to the script to update its config)
+
+```toml
+input_topics = ["te/+/+/+/+/m/+"]
```
albinsuresh: I was initially wondering why this is an array of distinct topics and not a single (possibly wildcard) topic. I'm assuming it is to support generic transformations like adding the timestamp various types of messages with a single pipeline definition.

docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,121 @@
...
+
+stages = [
+    { filter = "add_timestamp.js" },
+    { filter = "drop_stragglers.js", config = { max_delay = 60 } },
+    { filter = "te_to_c8y.js", meta_topics = ["te/+/+/+/+/m/+/meta"] }
+]
+```
+
+- A filter has to export at least one `process` function.
```
albinsuresh: Having the function signature also defined here or prior to this section would be good.

docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,121 @@
...
+- A filter can also export an `update_config` function
+  - This function is called on each message received on the `meta_topics` as defined in the config.
+  - The arguments are:
+    - The message to be interpreted as a config update `{ topic: string, payload: string }`
+    - The current config
+   - The returned value (an arbitrary JSON value) is then used as the new config for the filter.
+- A filter can also export a `tick` function
+  - This function is called at a regular pace with the current time and config.
+  - The filter can then return zero, one or many transformed messages
+  - By sharing an internal state between the `process` and `tick` functions,
```
albinsuresh: I think I get the purpose of `tick` in isolation, as a way of aggregating data points. But, I'm failing to understand the use-case for combining both `process` and `tick` together. I hope it'll be clearer as more examples are added.

albinsuresh (commented):
Here is my first round of review comments focussing purely on the pipeline APIs and not much of the rust infra code. Unfortunately, I'm having some trouble testing the feature due to some certificate misconfiguration preventing the mapper from even starting with the following error:

```
Jun 10 06:36:23 2a9636a9acc5 tedge-mapper[3107]: 2025-06-10T06:36:23.630655575Z ERROR Runtime: Actor MQTT-2 has finished unsuccessfully: ActorError(InvalidPrivateKey(InvalidCertificate(Other(OtherError(UnsupportedCertVersion)))))
```

Will provide additional usability feedback once that issue is resolved.

docs/src/references/mappers/gen-mapper.md
didier-wenzek: What we want here is for the messages received on the meta topics to be directly delivered to the stage. The other stages have no way to do something meaningful with these messages (In this case, units for various measurement types).

A better name might be `config_topics`.

didier-wenzek: What you propose seems too specific to this specific example (adding units to measurements leveraging the `meta` topics as defined by thin-edge). Sure the script (here `te_to_c8y.js`) has to appropriately match data and meta data, but why should the generic mapper enforce that from the outside?

docs/src/references/mappers/gen-mapper.md
didier-wenzek: I agree these APIs convey better specific intent. However, having a `process` API that encompasses all the specific cases is a key point to be able to chain stages into pipelines and to expend the framework in the future. We need to find a way for both with an usual API built on top of a sound and expressive definition of what is a transformation stage.

docs/src/references/mappers/gen-mapper.md
didier-wenzek: An MQTT subscription is actually a list of topics possibly with wildcards. Having a single topic is quickly annoying. This is something I miss with `tedge mqtt sub` for instance.

docs/src/references/mappers/gen-mapper.md
didier-wenzek: For a stage to compute an aggregation over a time window, the `process` and `tick` functions have to share a state where messages are pushed when received by the `process` functions and the outcome extracted by the `tick` function at the end of the time window.

I will make this clearer in the doc.

didier-wenzek: See https://github.com/thin-edge/thin-edge.io/pull/3650/commits/b5a9c83c8495c266eb51a34b3832bfc41a5c4bd9

docs/src/references/mappers/gen-mapper.md
didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3650/commits/b5a9c83c8495c266eb51a34b3832bfc41a5c4bd9

docs/src/references/mappers/gen-mapper.md
didier-wenzek: I have yet another proposal, which is to be very liberal with javascript function outputs.
e.g. to accept a `null` return as no messages and to accept a single message to be returned instead of forcing the script to return a singleton array. 

 See https://github.com/thin-edge/thin-edge.io/pull/3650/commits/b5a9c83c8495c266eb51a34b3832bfc41a5c4bd9

crates/extensions/tedge_gen_mapper/pipelines/loop.toml
```diff
@@ -0,0 +1,7 @@
+# This pipeline is on purpose looping: the messages are published to the same topic
+input_topics = ["loopback/#"]
+
+stages = [
+    { filter = "add_timestamp.js" },
+    { filter = "circuit-breaker.js", tick_every_seconds = 1, config = { stats_topic = "te/error", too_many = 10000, message_on_too_many = { topic = "te/device/main///a/too-many-messages", payload = "too many messages" }, message_on_back_to_normal = { topic = "te/device/main///a/too-many-messages", payload = "back to normal" } } }
```
didier-wenzek: This line highlights pain points with the current configuration format.

- There are two levels of config which is confusing. `tick_every_second` is used by the engine, while `config.too_many` is passed to the script. Flattening the `config` might be simpler.
- TOML inline tables are intended to appear on a single line. Leading to painfully long lines as here.
- Passing a JSON payload to the filter config is not straightforward. Currently the engine expects a string and the JS filter has to stringyfy  the payload.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -161,6 +169,54 @@ impl MessageSource<MqttMessage, TopicFilter> for MqttActorBuilder {
...
+impl MessageSource<MqttMessage, DynSubscriptions> for MqttActorBuilder {
+    fn connect_sink(
+        &mut self,
+        subscriptions: DynSubscriptions,
+        peer: &impl MessageSink<MqttMessage>,
+    ) {
+        let client_id = self.connect_id_sink(subscriptions.init_topics(), peer);
+        subscriptions.set_client_id(client_id);
+    }
+}
```
didier-wenzek: @jarhodes314 Here is my proposal to connect the MQTT actor using the usual traits rather than a concrete impl.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -161,6 +169,54 @@ impl MessageSource<MqttMessage, TopicFilter> for MqttActorBuilder {
     }
 }
 
+impl MessageSource<MqttMessage, DynSubscriptions> for MqttActorBuilder {
+    fn connect_sink(
+        &mut self,
+        subscriptions: DynSubscriptions,
```
jarhodes314: Is there a particular reason why we can't just pass `&mut DynSubscriptionsInner` here and avoid the interior mutability?

didier-wenzek: I tried but failed to work around the lifetime constraints on the call site.

How to tell that the subscriptions is not borrowed longer than the `connect_sink` invocation?

```
56 |           mqtt.connect_mapped_sink(&mut inner, &self.message_box, |msg| {
                                       ^^^^^^^^^^ borrowed value does not live long enough
59 |           let client_id = inner.client_id.unwrap();
   |                           ^^^^^^^^^^^^^^^ use of borrowed `inner`

```

crates/extensions/tedge_gen_mapper/src/pipeline.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    pub async fn update_config(
+        &mut self,
+        js_runtime: &JsRuntime,
+        message: &Message,
+    ) -> Result<(), FilterError> {
+        for stage in self.stages.iter_mut() {
+            if stage.config_topics.accept_topic_name(&message.topic) {
+                stage.filter.update_config(js_runtime, message).await?
```
albinsuresh: I'm not sure how this meta data message handling would work with different devices ahving diffrent metadata. For example, assume the following metadata messages for `child01` and `child02`:

* `te/device/child01///m/temp/meta`: `{"unit": "C"}`
* `te/device/child02///m/temp/meta`: `{"unit": "F"}`

depending on the order in which these messages are delivered, the final unit in config would be either `C` or `F`. When the actual measurement from the other device is received later, it would be published with the wrong unit.

To prevent this, the java script logic would have to maintain a mapping of metadata for each entity topic id. Else the pipeline engine should somehow maintain this association or redefine the API in such a way that `process()` function gets the metadata message as well from the same target entity.

crates/extensions/tedge_gen_mapper/src/actor.rs
```diff
@@ -0,0 +1,163 @@
...
+                        },
+                        Some(InputMessage::FsWatchEvent(FsWatchEvent::Modified(path))) => {
+                            let Ok(path) = Utf8PathBuf::try_from(path) else {
+                                continue;
+                            };
+                            if matches!(path.extension(), Some("js" | "ts")) {
+                                self.processor.reload_filter(path).await;
+                            } else if path.extension() == Some("toml") {
+                                self.processor.reload_pipeline(path).await;
+                                self.send_updated_subscriptions().await?;
```
albinsuresh: If a newly added pipeline also subscribes to the exact same retained config topics that were previously subscribed by some other pipeline (e.g: `te/+/+/+/+/m/+/meta`), then the new pipeline wouldn't get the messages on those topics when activated, as they were already received and delivered to the mapper actor earlier.

crates/extensions/tedge_gen_mapper/src/pipeline.rs
```diff
@@ -0,0 +1,190 @@
...
+        let mut messages = vec![message.clone()];
+        for stage in self.stages.iter() {
+            let mut transformed_messages = vec![];
+            for message in messages.iter() {
+                let filter_output = stage.filter.process(js_runtime, timestamp, message).await;
+                transformed_messages.extend(filter_output?);
+            }
+            messages = transformed_messages;
+        }
+        Ok(messages)
```
albinsuresh: We may have to add a check whether the transformed messages have a topic different from the incoming message, to avoid infinte loops (the transformed message triggering the pipeline again). Unfortunately we can't statically detect this and reject such pipelines without even loading them, unless we enforce the terminal step in a pipeline to define a static topic prefix.

crates/extensions/tedge_gen_mapper/pipelines/average.js
```diff
@@ -0,0 +1,99 @@
+// Compute the average value of a series of measurements received during a time windows
+// - Take care of the topic: messages received over different topics are not mixed
+// - Ignore messages which are not formated as thin-edge JSON
+// - Ignore values which are not numbers
+// - Use the first timestamp as the timestamp for the aggregate
+class State {
+    static agg_for_topic = {}
```
albinsuresh: Since the state is maintained in a static field, would it cause some sort of interference if mutiple pipelines share the same script? Or each pipeline somehow gets its own static instance?

crates/extensions/tedge_gen_mapper/src/pipeline.rs
```diff
@@ -0,0 +1,190 @@
...
+        self.update_config(js_runtime, message).await?;
+        if !self.input_topics.accept_topic_name(&message.topic) {
+            return Ok(vec![]);
+        }
+
+        let mut messages = vec![message.clone()];
+        for stage in self.stages.iter() {
+            let mut transformed_messages = vec![];
+            for message in messages.iter() {
+                let filter_output = stage.filter.process(js_runtime, timestamp, message).await;
```
albinsuresh: It would have been better if there was a mechanism to short-circuit the execution of a pipeline on error, without relying on that stage returning a message to `te/error` topic. With `te/errors` message, it would also get propagated to the rest of the pipeline, which they must reject. It would have been better if there was an implicit mechanism for the script to return an error that this pipeline engine would use to break out of the pipeline execution and would convert to the `te/errors` messages.

crates/extensions/tedge_gen_mapper/pipelines/te_to_c8y.js
```diff
@@ -0,0 +1,132 @@
...
+///         "altitude": {
+///             "value": 98.6
+///         }
+///     },
+///     "pressure": {
+///         "pressure": {
+///             "value": 98
+///         }
+///     }
+/// }
```
albinsuresh: As discussed offline, we still need to figure out a way to provide the external id of the entities to this `process` function.

crates/extensions/tedge_gen_mapper/pipelines/average.js
didier-wenzek: Each pipeline must be given its own static instance. This has not been done yet, though.

didier-wenzek: Fixed by https://github.com/thin-edge/thin-edge.io/pull/3650/commits/094119f261d95cd80fbc760db6245b89fd1b664b

crates/extensions/tedge_gen_mapper/src/js_filter.rs
```diff
@@ -0,0 +1,360 @@
...
+    pub config: JsonValue,
+    pub tick_every_seconds: u64,
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct JsonValue(serde_json::Value);
+
+impl JsFilter {
+    pub fn new(pipeline: PathBuf, index: usize, path: PathBuf) -> Self {
+        let module_name = format!("{}|{}|{}", pipeline.display(), index, path.display());
```
didier-wenzek: The `index` is used to distinguish instances of javascript module used by the same pipeline on different stages.

- I still have to add a test for that case.
- The current full name is a bit large. What about using a hash?

crates/extensions/tedge_gen_mapper/src/pipeline.rs
didier-wenzek: See this test which addresses your point: https://github.com/thin-edge/thin-edge.io/pull/3650/files#diff-909c558385e14854043758513bb6b54e395642e2265df934c0fa92c96f3e2980

> To prevent this, the java script logic would have to maintain a mapping of metadata for each entity topic id.

Yes this is exactly what is done.

crates/extensions/tedge_gen_mapper/src/actor.rs
didier-wenzek: Yeah this is an issue. This should be addressed by https://github.com/thin-edge/thin-edge.io/pull/3710.

As that is not so easy to fix, an option is to let the `tedge_gen_mapper` maintains a connection to the MQTT actor *per* pipeline.

crates/extensions/tedge_gen_mapper/src/pipeline.rs
didier-wenzek: One could add a check at runtime (i.e. checking that the transformed messages are sent on a different topic than the source message). However, this will not protect again loops over several pipelines.

What I propose, is to add an extra pipeline config to tell where the messages are expected to be sent.

```
input_topics = ["te/+/+/+/+/m/+"]

stages = [
    { filter = "add_timestamp.js" },
    { filter = "drop_stragglers.js", config = { max_delay = 60 } },
    { filter = "te_to_c8y.js", meta_topics = ["te/+/+/+/+/m/+/meta"] }
]

output_topics = ["c8y/#"]
```

crates/extensions/tedge_gen_mapper/src/pipeline.rs
didier-wenzek: > It would have been better if there was a mechanism to short-circuit the execution of a pipeline on error, without relying on that stage returning a message to `te/error` topic. With `te/errors` message, it would also get propagated to the rest of the pipeline, which they must reject.

I agree that forwarding error messages to the next stage is really confusing as this leads to cascading errors.

> It would have been better if there was an implicit mechanism for the script to return an error that this pipeline engine would use to break out of the pipeline execution and would convert to the `te/errors` messages.

A script can throw an error which is catched by the engine. The current behavior is to log this error and to abort the pipeline. What you propose, i.e. to publish the error on a pipeline error topic is a good idea.

docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,168 @@
...
+- feature the ability to customize MEA processing by combining builtin filters with user-provided functions written in JavaScript
+- provide tools to create, test, monitor and debug filters and pipelines
+- be stable enough that user-defined filters will still work without changes with future releases.
+
+To keep things simple for the first release, the following questions are deferred:
+
+- Could a generic mapper let users define bridge rules as well as message transformation pipelines?
+- Does it make sense to run such a mapper on child-devices?
+- Could a pipeline send HTTP messages? Or could a filter tell the runtime to send messages over HTTP?
+- How to handle binary payloads on the MQTT bus?
```
reubenmiller: Whilst it might not be a focus, we should not prevent users from using binary payloads

crates/core/tedge/src/cli/mapping/cli.rs
```diff
@@ -0,0 +1,126 @@
...
+        filter: Option<PathBuf>,
+
+        /// Send a tick after all the message samples
+        #[clap(long = "final-tick")]
+        final_tick: bool,
+
+        /// Topic of the message sample
+        ///
+        /// If none is provided, messages are read from stdin expecting a line per message:
+        /// [topic] payload
```
reubenmiller: This nomenclature can be slightly confusing as on the command line, `[xxx]` generally means "optional", where as the topic is mandatory.

Though I assume `[]` was chosen because that is how we format the messages on the console from `tedge mqtt sub`. So given that, maybe we can just provide an example in addition to the syntax,

e.g.

```
/// [topic] payload
/// For example,
/// [te/example/one] {"text":"test message"}
```

didier-wenzek: Yes, the bracketed topic syntax has been chosen to be able to pipe `tedge mqtt sub` into `tedge mapping test`.

However, if the topic is provided on the command line then no brackets are expected:

```
$ tedge mapping test te/example/one '{"text":"test message"}'
````

**Comment by didier-wenzek** [comment]
Breaking changes: all the flow TOML files have to be updated, due to the following naming fixes:

- Pipeline -> Flow : https://github.com/thin-edge/thin-edge.io/pull/3650/commits/c31c06846b7e5f00b8a4005e4dc6b6c57242b2bb
- Stage -> Step : https://github.com/thin-edge/thin-edge.io/pull/3650/commits/8aa71c753e6d71b3ea271483b7de1028b29b090c
- Filter -> Flow Step (or Flow Script depending on the context) https://github.com/thin-edge/thin-edge.io/pull/3650/commits/b71ad0987086e6432ca4567295325ae64ba56a7a

File: crates/extensions/tedge_gen_mapper/src/js_runtime.rs [hunk 18]
```diff
@@ -97,6 +107,28 @@ impl JsRuntime {
...
+            .send(request)
+            .await
+            .map_err(|err| anyhow!(err))?;
+
+        // FIXME: The following timeout is not working
+        //  - see unit test: js_script::while_loop
+        //  - the issue is that the quickjs runtime fails to yield when executing `while(true)`
+        //  - Using task::spawn_blocking to launch the quickjs runtime doesn't help
+        //    - A timeout is the properly raised
+        //    - but the JS runtime keeps executing `while(true)` and is no more responsive.
```
didier-wenzek: This is a concern. We might have to:

- abort the whole JS runtime
- reload all the flows
- disable the faulty script and flow

didier-wenzek:
Breaking changes: the commands have been renamed:

- `tedge mapping test` -> `tedge flows test` (see https://github.com/thin-edge/thin-edge.io/pull/3650/commits/0163a0e406b6220899286cbbba68fa4549e3526a)
- `tedge-mapper gen` -> `tedge-mapper flows` (see https://github.com/thin-edge/thin-edge.io/pull/3650/commits/8a0798db87d4606c3e2432912f189836991befaf)
- ` /etc/tedge/gen-mapper` -> `/etc/tedge/flows` (see https://github.com/thin-edge/thin-edge.io/pull/3650/commits/68513d83c76716564e982d9d6c06f08e11a3461e)

docs/src/references/mappers/gen-mapper.md
```diff
@@ -9,160 +9,100 @@ import ProposalBanner from '@site/src/components/ProposalBanner'
...
+  - Connectors can be seen as streams of messages all with the same shape so they can be processed by any step.
+- A *flow* applies a chain of transformation *steps* to input messages producing fully processed output messages.
+  - The *flows* put things in motion, actually interacting with the system, consuming and producing messages.
+  - Messages received on a flow are passed to the first step; and the transformed messages, if any,
+    are pushed to the subsequent steps upto the output connector.
+- A flow can combine builtin and user-provided steps.
+  - Builtin steps provide generic building blocks such as %%te%% JSON translation into Cumulocity JSON.
+  - Users can implement specific steps using JavaScript or TypeScript to refine transformations to their use cases. 
+- If some message transformations can be fully defined only from the input message, most require a *context*.
+  - What is the Cumulocity internal id of the device? What are the units used by a sensor? Does the location of the device matter?
```
albinsuresh: This cloud mapping aspect is something that I'm still a bit confused about. If this single dynamic mapper is expected to replace all the existing cloud mappers with the cloud-specific mappings just being some inbuilt transformation steps, then the flows can't be just sequential as they currently are. We'll need conditional step routing(to take some steps only on some input messages), branching out to multiple steps (to map the same source message to multiple cloud formats parallely, as done by the mappers running parallely now) etc. This would have a significant impact on the syntax.

docs/src/references/mappers/gen-mapper.md
```diff
@@ -9,160 +9,100 @@ import ProposalBanner from '@site/src/components/ProposalBanner'
...
+    to produce transformed messages on time-window boundaries.
+
+## Step API
+
+A transformation *scripts* is a JavaScript or TypeScript module that exports:
+
+- at least a function `onMessage`, aimed to transform one input message into zero, one or more output messages
+- possibly a function `onInterval`, called at regular intervals to produce aggregated messages.
+
+
```
albinsuresh: I'd list the message signatures here.

crates/extensions/tedge_gen_mapper/src/js_script.rs
```diff
@@ -0,0 +1,486 @@
...
+use std::path::Path;
+use std::path::PathBuf;
+use tracing::debug;
+
+#[derive(Clone)]
+pub struct JsScript {
+    pub module_name: String,
+    pub path: PathBuf,
+    pub config: JsonValue,
+    pub tick_every_seconds: u64,
```
albinsuresh: Syntactic sugar: how about accepting the interval in human readable form?
```suggestion
    pub tick_interval: SecondsOrHumanTime,
```

docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,108 @@
...
+  - This can also be a TypeScript module with a `.ts` extension.
+- The definition of flows must provide a list of MQTT topics to subscribe to.
+  - The flow will be feed with all the messages received on these topics.
+- A flow definition provides a list of steps.
+  - Each step is built from a javascript and is possibly given a config (arbitrary json that will be passed to the script)
+  - Each step can also subscribe to a list of MQTT topics (which messages will be passed to the script to update its config)
+
+```toml
+input_topics = ["te/+/+/+/+/m/+"]
+
```
albinsuresh: We should probably take an `error_topic` also as an input, which can be used to publish the error message when a step execution fails. Just a single flow level topic would be enough IMO. We can provide step-level error topics when the need arises.

docs/src/references/mappers/gen-mapper.md
```diff
@@ -0,0 +1,108 @@
...
+steps = [
+    { script = "add_timestamp.js" },
```
albinsuresh: This step adding timestamp would be a good candidate to be provided as an inbuilt transformation. Doing that would give us some more clarity on the syntax changes required to accommodate inbuilt transformations.

albinsuresh (commented):
> What about the approach proposed by NodeRed to [work with context](https://nodered.org/docs/user-guide/context)?
Thin-edge could provide 3 context levels:
context.step: only visible to the flow step that set the value
context.flow: visible to all steps on the same flow
context.device or context.entities: visible to all flows, notably populated with device and child device internal ids.

The `context` parameter would be a great addition as that'll make the state management  between steps a lot cleaner and make future enhancements like giving the pipeline access to tedge config or entity store a lot easier.

docs/src/references/mappers/gen-mapper.md
didier-wenzek: We decided to have a single function, named `onMessage`, which can return zero, one or more messages.

docs/src/references/mappers/gen-mapper.md
didier-wenzek: We decided to rename this function `onInterval` to better convey the intent.

crates/extensions/tedge_mqtt_ext/src/lib.rs
didier-wenzek: Here is how to avoid interior mutability See https://github.com/thin-edge/thin-edge.io/pull/3710#discussion_r2215876319

crates/extensions/tedge_gen_mapper/pipelines/te_to_c8y.js
didier-wenzek: The proposal is to use 3 context levels  à la NodeRed https://github.com/thin-edge/thin-edge.io/pull/3650#issuecomment-3075283984

docs/src/references/mappers/gen-mapper.md
didier-wenzek: Indeed, this needs to be clarified.

What I have in mind and has to be discussed, is to keep the c8y, aws, az and collectd mappers as today except that those will use the extensible mapper with a set of flows provided out of the box. Users will have then the ability to adapt these flows to their purpose.

design/decisions/0006-tedge-flows.md
```diff
@@ -0,0 +1,157 @@
...
+- To handle operations, would the plugins be expanded to do more complex things like HTTP calls, file-system interactions, etc.? 
+- What are the pros and cons to persist flow step states?
+- Split a flow, forwarding transformed messages to different flows for further processing
+
+### API
+
+The POC expects the flow scripts to implement a bunch of functions. This gives a quite expressive interface
+(filtering, mapping, splitting, dynamic configuration, aggregation over time windows), but at the cost of some complexity.
+
+- `process(t: Timestamp, msg: Message, config: Json) -> Vec<Message>`
```
albinsuresh, line 174:
```suggestion
- `process(t: Timestamp, msg: Message, context: Context) -> Vec<Message>`
```
Since we have plans to add context in the future anyway, how about adding it to the method signature in the initial version itself and expose the config JSON via this context object as `context.config`? We can keep adding more fields to the context like `context.device`, `context.tedgeConfig` etc in future as well.

crates/extensions/tedge_gen_mapper/src/js_runtime.rs
```diff
@@ -19,11 +20,19 @@ pub struct JsRuntime {
...
 impl JsRuntime {
     pub async fn try_new() -> Result<Self, LoadError> {
         let runtime = rquickjs::AsyncRuntime::new()?;
         runtime.set_memory_limit(16 * 1024 * 1024).await;
         runtime.set_max_stack_size(256 * 1024).await;
+        runtime
+            .set_interrupt_handler(Some(Box::new(|| {
+                let credits = TIME_CREDITS.fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
+                credits == 0
+            })))
```
didier-wenzek: This is not really a timeout as the frequency at which the interrupt handler is called is unknown.

- The number of credits has to be tuned. I just guessed some numbers for now.
- We could have something using a clock, though.

crates/extensions/tedge_gen_mapper/src/js_runtime.rs
didier-wenzek: Fix by https://github.com/thin-edge/thin-edge.io/pull/3650/commits/8212eac9e3c23f52f496538b04e126bf4ba5a5c2

docs/src/references/mappers/gen-mapper.md
didier-wenzek: Done with https://github.com/thin-edge/thin-edge.io/pull/3650/commits/edf1c5d11975b68df984607a484d96efb162de7f and https://github.com/thin-edge/thin-edge.io/pull/3650/commits/c46629ea745e82fc5717efe5f88137564779ab55

crates/extensions/tedge_gen_mapper/src/js_script.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/56a2b9103eede3621b80400a7874378f611b46d0

tests/RobotFramework/tests/tedge_flows/flows/average.js
```diff
@@ -0,0 +1,108 @@
+// Compute the average value of a series of measurements received during a time windows
+// - Take care of the topic: messages received over different topics are not mixed
+// - Ignore messages which are not formated as thin-edge JSON
```
rina23q:
```suggestion
// - Ignore messages which are not formatted as thin-edge JSON
```

design/decisions/0006-tedge-flows.md
```diff
@@ -0,0 +1,157 @@
+# Extensible mapper and user-provided flows
+
+* Date: __2025-07-1-__
```
rina23q: It looks like a mistake. The suggested date is just my guess...

```suggestion
* Date: __2025-07-01__
```

design/decisions/0006-tedge-flows.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Devops tools
+
+The flexibility to customize MQTT message processing with user-provided functions comes with risks:
+- a step might not behave as expected,
+- flows might be overlapping or conflicting, possibly sending duplicate messages or creating infinite loops
+- builtin flows might be accidentally disconnected or broken
+- a step might introduce a performance bottleneck.
+
+To help mitigating these risks, the `tedge mapping` sub-commands provide the tools to test, monitor and debug steps and flows.
```
rina23q: Since it's decision doc, I'm not sure how much it should be the same as actually implemented. But I think `tedge mapping` here is actually `tedge flows`. 

```suggestion
To help mitigating these risks, the `tedge flows` sub-commands provide the tools to test, monitor and debug steps and flows.
```

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+- A *step* function transforms one input message into zero, one or more output messages.
+  - Steps are effect-free functions, with no access to MQTT, HTTP or the file-system.
+  - The focus is on message transformation, format conversion, content extraction and completion as well as filtering and redacting.
+- A *connector* is used by the mapper to consume messages from and produce messages to.
+  - MQTT is the primary message source and target, but overtime others can be added.
+  - Connectors can be seen as streams of messages all with the same shape so they can be processed by any step.
+- A *flow* applies a chain of transformation *steps* to input messages producing fully processed output messages.
+  - The *flows* put things in motion, actually interacting with the system, consuming and producing messages.
+  - Messages received on a flow are passed to the first step; and the transformed messages, if any,
+    are pushed to the subsequent steps upto the output connector.
```
rina23q:
```suggestion
    are pushed to the subsequent steps up to the output connector.
```

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
+---
+title: Extensible mapper and user-provided Flows
+tags: [Reference, Mappers, Cloud]
+sidebar_position: 2
+draft: true
```
rina23q, line 5: Is it intended to keep this doc draft?

didier-wenzek: As of now, yes. This will be marked as visible when packaging will be available.

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+
+<ProposalBanner/>
+
+## Concepts
+
+Users can extend and adapt the built-in mappers for Cumulocity, Azure and AWS
+with their own filtering and message transformation rules,
+leveraging the core mapping rules and mapper mechanisms (bridge connections, HTTP proxies, operations).
+
+As an example, users can now adapt to their use cases the translation of measurements from %%te%% JSON to Cumulocity JSON:
```
albinsuresh, line 18: This sentence talks specifically about `te` to `c8y` translation whereas the bullet points below describe generic translations from "non-standard" input topics and output topics. Either change this sentence or update the bullet points to reflect the specific `te` inputs and `c8y` outputs.
```suggestion
As an example, users can now adapt to their use cases the translation of measurements from any input message format on arbitrary topics to any cloud message format:
```
At the end, may be we can talk about the possibility of overriding existing mapper with the user's own `te to c8y` mapper as an example.

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+  - adding units read from device config
+  - producing transformed measurements on a non-standard topic.
+
+The behavior of a mapper is defined by a set of *connectors*, *flows*, *steps* and transformation *scripts*
+which rule how to consume, transform and produce MQTT messages.
+
+- A *step* function transforms one input message into zero, one or more output messages.
+  - Steps are effect-free functions, with no access to MQTT, HTTP or the file-system.
+  - The focus is on message transformation, format conversion, content extraction and completion as well as filtering and redacting.
+- A *connector* is used by the mapper to consume messages from and produce messages to.
```
```suggestion
- A *connector* is used by the mapper to consume messages from a source and produce messages to a target.
```

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+    - enriched by some flows and steps with context info extracted from metadata and capability messages,
+    - used by all flows and steps to adapt their behavior
+- %%te%% provides some support to steps aggregating messages over time windows.
+  - For each aggregating step, the mapper persists a state (a JSON object)
+    which can be updated by the step function on each message and at regular intervals
+    to produce transformed messages on time-window boundaries.
+
+## Step API
+
+A transformation *scripts* is a JavaScript or TypeScript module that exports:
```
```suggestion
A transformation *script* is a JavaScript or TypeScript module that exports:
```

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+These values are configured by the flow and can be dynamically updated on reception of config update messages.
+
+## Flow configuration
+
+- The generic mapper loads flows and steps stored in `/etc/tedge/flows/`.
+- A flow is defined by a TOML file with `.toml` extension.
+- A step is defined by a JavaScript file with `.js` extension.
+  - This can also be a TypeScript module with a `.ts` extension.
+- The definition of flows must provide a list of MQTT topics to subscribe to.
+  - The flow will be feed with all the messages received on these topics.
```
```suggestion
  - The flow will be fed with all the messages received on these topics.
```

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+- A flow definition provides a list of steps.
+  - Each step is built from a javascript and is possibly given a config (arbitrary json that will be passed to the script)
+  - Each step can also subscribe to a list of MQTT topics (which messages will be passed to the script to update its config)
```
```suggestion
  - Each step can also subscribe to a list of MQTT meta topics where the metadata about the actual data message is stored (e.g, meta topic of a measurement type where its units threshold values are defined). The messages received on these topics will be passed to the `onConfigUpdate` function in the script to update its config.
```

design/decisions/0006-tedge-flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/f28cf8587f04b833eafc42bbd1e4e97c84b5465f

docs/src/references/mappers/flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/f28cf8587f04b833eafc42bbd1e4e97c84b5465f

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+
+```toml
+input.mqtt.topics = ["te/+/+/+/+/m/+"]
+
+steps = [
+    { script = "add_timestamp.js" },
+    { script = "drop_stragglers.js", config = { max_delay = 60 } },
```
albinsuresh, line 131: Wasn't there a discussion on whether to rename this `meta_topics` to `config_topics` to better align with the script function? The agreement was to stick with `meta_topics`?

design/decisions/0006-tedge-flows.md
didier-wenzek: Yeah, this decision doc doesn't have to be updated after subsequent decisions.

However, for this first PR it's better to have things aligned.
=> addressed https://github.com/thin-edge/thin-edge.io/pull/3650/commits/f28cf8587f04b833eafc42bbd1e4e97c84b5465f

tests/RobotFramework/tests/tedge_flows/flows/average.js
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/fb54f77dd9955219cf8e341970328ac81842918b

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -0,0 +1,526 @@
...
+    /// - the current timestamp
+    /// - the message to be transformed
+    /// - the flow step config (as configured for the flow step, possibly updated by onConfigUpdate messages)
+    ///
+    /// The returned value is expected to be an array of messages.
+    pub async fn on_message(
+        &self,
+        js: &JsRuntime,
+        timestamp: &DateTime,
+        message: &Message,
```
albinsuresh, line 91: Why take a reference of `Message` here when we always have to pass a clone to the js engine? The reference is passed down the entir chain starting from the actor -> flow -> step -> script finally to clone it here, when the original owned message is simply discarded after passing it down this chain.

I usually don't bother much about clones, but just raising it here as this is telemetry data(possibly high in volume to make an impact).

That being said, I see that we're doing the same in the c8y mapper as well where `&MqttMessage` is passed to the converter and not the original message itself.

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -0,0 +1,526 @@
...
+    ///
+    /// The "onConfigUpdate" function of the JS module is passed 2 arguments
+    /// - the message
+    /// - the current flow step config
+    ///
+    /// The value returned by this function is used as the updated flow step config
+    pub async fn on_config_update(
+        &mut self,
+        js: &JsRuntime,
+        message: &Message,
```
albinsuresh, line 119: Same reasoning as the comment above: owned vs reference

docs/src/references/mappers/flows.md
```diff
@@ -0,0 +1,137 @@
...
+    { script = "te_to_c8y.js", meta_topics = ["te/+/+/+/+/m/+/meta"] }
+]
+```
+
```
albinsuresh, line 135: It'd be nice to document the `tedge flows` cli here as well.

crates/core/tedge/src/cli/flows/cli.rs
```diff
@@ -0,0 +1,122 @@
...
+        ///
+        /// Default to /etc/tedge/flows
+        #[clap(long)]
+        flows_dir: Option<PathBuf>,
+
+        /// Path to the flow step script or TOML flow definition
+        ///
+        /// If none is provided, applies all the matching flows
+        #[clap(long)]
+        flow: Option<PathBuf>,
```
albinsuresh, line 43: Would have been nice if just the file name of the `toml` file or the `js` file could be accepted here as `--flow` value as their parent directory could be derived form `flows_dir` to save some typing.

```
tedge flows test --flow /etc/tedge/flows/count-measurements.toml test/device/main///m/temp '50'
```
vs
```
tedge flows test --flow count-measurements.toml test/device/main///m/temp '50'
```

crates/core/tedge/src/cli/flows/test.rs
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+}
+
+fn parse(line: String) -> Result<Option<Message>, Error> {
+    let line = line.trim();
+    if line.is_empty() {
+        return Ok(None);
+    }
+    if !line.starts_with("[") {
```
albinsuresh, line 95: Not a request for change but we can discuss this post-merge: it would have been better if this tool accepted topic without this opening and closing parenthesis as well, to make it compatible with tools like `mosquitto_sub` as well, in addition to `tedge mqtt sub`. This restriction ties it too close to `tedge mqtt sub` utility.

If for some reason we can't simultaneously handle inputs with and without parenthesis for topics, I'd rather provide a `--raw` option to the `tedge mqtt sub` utility that prints the output without the parenthesis.

crates/core/tedge/src/cli/flows/cli.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// Trigger onInterval after all the message samples
+        #[clap(long = "final-on-interval")]
```
albinsuresh, line 46: Optional suggestion as I found the `final-on-interval` a bit unintuitive: I'm assuming that's how we're meant to use this option, where the EOF signal triggers the `on_interval` function.
```suggestion
        #[clap(long = "eof-on-interval")]
```

To be honest, I'm not 100% convinced about the proposed name either.

docs/src/references/mappers/flows.md
didier-wenzek: Yeah. I prefer the original wordings because more concrete. The non-standard input and output make sense, although not really precise. A thin-edge json message can be received from some `custom/topic` and a c8y json measurement can be published to another cloud.

docs/src/references/mappers/flows.md
didier-wenzek: No agreement has been done. We still have to discuss this idea of *meta* /  *config* topics when combined with *contexts* à la NodeRed. There are some overlaps that have to be clarified.

crates/core/tedge/src/cli/flows/test.rs
didier-wenzek: > This restriction ties it too close to `tedge mqtt sub` utility.

This is precisely the goal.

> If for some reason we can't simultaneously handle inputs with and without parenthesis for topics, I'd rather provide a `--raw` option to the `tedge mqtt sub` utility that prints the output without the parenthesis.

Not really keen to do that.

crates/core/tedge/src/cli/flows/cli.rs
didier-wenzek: Yeah, I'm not fully convinced by my proposal too.

But can a flag convey everything? For instance, here the purpose of this flag is to "trigger the `onInterval` method when EOF is reached".

docs/src/references/mappers/flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/78f51985c55660302dc1347714cb78c6c56b0bf6

docs/src/references/mappers/flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/78f51985c55660302dc1347714cb78c6c56b0bf6

docs/src/references/mappers/flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/78f51985c55660302dc1347714cb78c6c56b0bf6

docs/src/references/mappers/flows.md
didier-wenzek: applied: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/78f51985c55660302dc1347714cb78c6c56b0bf6

docs/src/references/mappers/flows.md
didier-wenzek: Added as well as basic docs for `tedge-mapper flows`. See https://github.com/thin-edge/thin-edge.io/pull/3650/commits/78f51985c55660302dc1347714cb78c6c56b0bf6

crates/extensions/tedge_flows/src/js_script.rs
didier-wenzek: Passing a reference along the chain  (actor -> flow -> step -> script) makes sense as the same message can be processed by several flows and steps.

The JS engine is running in a different thread and even on a different memory heap. So some form of cloning is required here. That said, we currently have two cloning related steps. A first one when sending the processing request to the JS worker via a channel, then a transformation form rust to js memory. Could this be avoid? Maybe, but not easily.

Later, when introducing builtin steps written in Rest, the processing will be doable without any cloning. Hence the use of references.

rina23q:
@didier-wenzek I'm sure that you should just rebase as I see `1.6.0` in your `Cargo.lock` file. We recently bumped up to `1.6.1`. It will solve the `Cargo.lock` issue. I confirmed it in my forked repo. https://github.com/rina23q/thin-edge.io/pull/46

```toml
[[package]]
name = "aws_mapper_ext"
version = "1.6.0"
```

crates/core/tedge/src/cli/flows/test.rs
albinsuresh: > > This restriction ties it too close to `tedge mqtt sub` utility.
> 
> This is precisely the goal.
>

To be clear, I was proposing to accept both kinds of input: topic with or without the parenthesis around it (effectively making the parenthesis optional) so that it's consistent with the input format when accepted as cli args as well.

But, are you saying that we're intentionally restricting this to the `tedge mqtt sub` output format **only** for some reason? Just curious why.

Anyway, as mentioned in the original comment, this wasn't a request for change, but just a discussion point.

> > If for some reason we can't simultaneously handle inputs with and without parenthesis for topics, I'd rather provide a `--raw` option to the `tedge mqtt sub` utility that prints the output without the parenthesis.
> 
> Not really keen to do that.

That's fine, since that choice was intentional.

crates/extensions/tedge_flows/src/js_script.rs
albinsuresh: > Passing a reference along the chain (actor -> flow -> step -> script) makes sense as the same message can be processed by several flows and steps.

Good point. Clones definitely can't be avoided while passing the same source message across flows. But may be, within the flows.

> A first one when sending the processing request to the JS worker via a channel, then a transformation form rust to js memory. Could this be avoid? Maybe, but not easily.

Yeah, I didn't even think that the in-memory cloning required to transform the rust JSON representation to js JSON representation can be avoided. I was just talking about a few clones done at the rust level itself, while the message is passed through the steps in a flow. For example we have the following code in `Flow::on_message`:

```
        let mut messages = vec![message.clone()];
        for step in self.steps.iter() {
            // redacted
            for message in messages.iter() {
                // redacted
                let step_output = step.script.on_message(js_runtime, timestamp, message).await;
```

where there first clone is done while creating the `messages` vector, which is consumed during the iteration as a reference and passed down to `JsScript::on_message()` only to be cloned there again:

```
let mut message = message.clone();
```

If the `JsScript:on_message` would accept an owned message, then the entries from the `messages` vector could be passed directly with an iteration using `into_iter()`.

But, these are marginal perf improvements that we can evaluate later as well. Just capturing my thoughts.

> Later, when introducing builtin steps written in Rest, the processing will be doable without any cloning. Hence the use of references.

Good point.

crates/core/tedge/src/cli/flows/cli.rs
albinsuresh: > But can a flag convey everything?

It was just an attempt to convey as much as possible with minimal changes. My first thought was `on-interval-on-eof` which was too much. So yeah, let's stick with the current one until we can come up with something better.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -0,0 +1,269 @@
...
+        if script.no_js_on_interval_fun && script.interval_secs != 0 {
+            warn!(target: "flows", "Flow script with no 'onInterval' function: {}; but configured with an 'interval' in {flow}", script.path.display());
+        }
+    }
+
+    pub(crate) fn fix(&mut self) {
+        let script = &mut self.script;
+        if !script.no_js_on_interval_fun && script.interval_secs == 0 {
+            // 0 as a default is not appropriate for a script with an onInterval handler
+            script.interval_secs = 1;
```
albinsuresh, line 178: Why **silently** fix a user misconfiguration? Wouldn't it be better to just log this as misconfiguration in the `check() itself, and avoid loading/reloading the contained flow?

albinsuresh (approved):
Just added a few remarks on some points, which can be discussed/improved post-merge. Approving, since none of those are blockers.

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: Yeah, this is a hack. But this not related to user misconfiguration.

The problem is due to how flow steps are built. A first version is created reading the flow `.toml` file, and this version is updated later when the `.js` script is loaded. Whether or not an `onInterval` function is given the appropriate default is not the same.

Adding this to the list of point to fix for the public release: https://github.com/thin-edge/thin-edge.io/issues/3756

Cargo.lock
Bravo555, line 1: todo: revert `Cargo.lock` changes that are not related to `tedge-flows`, because currently it looks like we upgraded all of the dependencies, unless we're okay with that :)

crates/core/tedge/src/cli/flows/cli.rs
```diff
@@ -0,0 +1,122 @@
...
+
+        /// Topic of the message sample
+        ///
+        /// If none is provided, messages are read from stdin expecting a line per message:
+        /// [topic] payload
+        topic: Option<String>,
+
+        /// Payload of the message sample
+        ///
+        /// If none is provided, payloads are read from stdin
```
jarhodes314, line 57: This should probably clarify the same "line per message" format as we do for topic, assuming that is the case?

Further thought: when I first read this, I thought what this meant was that you need to supply topic if you supply payload, but you could supply a topic and have a series of payloads read from stdin. Looking at the validation though, that isn't the case (you need both or neither set). This behaviour is sensible, but I think these doc comments should be more explicit in saying that you either need both topic and payload set or neither.

crates/core/tedge/src/cli/flows/cli.rs
```diff
@@ -0,0 +1,122 @@
...
+            } => {
+                let flows_dir = flows_dir.unwrap_or_else(|| Self::default_flows_dir(config));
+                let message = match (topic, payload) {
+                    (Some(topic), Some(payload)) => Some(Message {
+                        topic,
+                        payload,
+                        timestamp: None,
+                    }),
+                    (Some(_), None) => Err(anyhow!("Missing sample payload"))?,
+                    (None, Some(_)) => Err(anyhow!("Missing sample topic"))?,
```
jarhodes314, line 85:
```suggestion
                    (Some(_), None) => bail!("Missing sample payload"),
                    (None, Some(_)) => bail!("Missing sample topic"),
```

crates/core/tedge/src/cli/flows/cli.rs
```diff
@@ -0,0 +1,122 @@
+use crate::cli::flows::list::ListCommand;
+use crate::cli::flows::test::TestCommand;
+use crate::command::BuildCommand;
+use crate::command::Command;
+use crate::ConfigError;
+use anyhow::anyhow;
```
jarhodes314, line 6: To go with below suggestion

```suggestion
use anyhow::bail;
```

crates/core/tedge/src/cli/flows/list.rs
```diff
@@ -0,0 +1,44 @@
...
+
+pub struct ListCommand {
+    pub flows_dir: PathBuf,
+    pub topic: Option<String>,
+}
+
+#[async_trait::async_trait]
+impl Command for ListCommand {
+    fn description(&self) -> String {
+        format!("list flows and flow steps in {:}", self.flows_dir.display())
```
jarhodes314: I'm not sure what the `{:}` means here? I'm familiar with `{:?}` for debug formatting and `{:#}` for alt formatting, but don't know how `{:}` changes the output?

Cargo.lock
didier-wenzek: Reverted

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -0,0 +1,83 @@
+*** Settings ***
+Library             ThinEdgeIO
+
+Test Setup          Custom Setup
+Test Teardown       Get Logs
```
Bravo555: question: why `Test Setup/Teardown` and not `Suite Setup/Teardown`?

This creates a container per test suite, so possibly it was unintended?

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
```diff
@@ -0,0 +1,129 @@
...
+///
+/// These metadata are expected to have the same shape of the actual values.
+///
+/// ```
+/// [te/device/main///m/example/meta] { "temperature": { "unit": "°C" }}
+/// ```
+///
+/// and:
+/// ```
+/// [te/device/main///m/example] { "temperature": { "unit": 23 }}
```
Bravo555: suggestion: typo?

```suggestion
/// [te/device/main///m/example] { "temperature": { "value": 23 }}
```

crates/core/tedge/src/cli/flows/cli.rs
didier-wenzek: Unfortunately, this leads to mismatched types: expected `TEdgeError`, found `Error`

crates/core/tedge/src/cli/flows/test.rs
```diff
@@ -0,0 +1,139 @@
...
+                    return Some(line.to_string());
+                }
+            }
+            Err(err) => {
+                eprintln!("Fail to read input stream {}", err);
+                return None;
+            }
+        }
+    }
+}
```
Bravo555, line 129: note: just for future convenience, could also be written as:

```suggestion
async fn next_line(input: &mut BufReader<Stdin>) -> Option<String> {
    let mut input = input.lines();
    input
        .next_line()
        .await
        .inspect_err(|err| {
            eprintln!("Fail to read input stream {}", err);
        })
        .unwrap_or(None)
}
```

crates/core/tedge/src/cli/flows/list.rs
didier-wenzek: > I'm not sure what the {:} means here? 

Me neither ;-) Fixed: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/2b5dd76a8980f43f1390230cb142f3ab803f68ad

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3650/commits/3564eedb4c1183df05c8912bf10dfe6e0bcfe036

The test suite is now faster.

crates/core/tedge/src/cli/flows/cli.rs
didier-wenzek: Added as a follow up task https://github.com/thin-edge/thin-edge.io/issues/3756

Issue #3640, opened 2025-05-26
tedge connect aws with untrusted certificate passes the direct connection step successfully


reubenmiller:
**Describe the bug**

When connect to AWS IoT Core using a device certificate that **IS NOT** trusted by AWS, then the manual MQTT client connection performed in the `tedge connect aws` shows incorrectly that the verification was successful (via a green tick) when the following connection test fails, see below for the output in question:

```sh
Verifying device is connected to cloud... ✓
```

It is possible that the AWS is terminating the MQTT connection abruptly and not via the MQTT Error Code 5 (Connection Refused. The client is not authorized to connect), or whether the client needs to check if the connection is still active after a minimum amount of time (e.g. 2 seconds).

The full connection test does fail (which is expected), however it is difficult to verify the reason for the failure.

Below is the full command output when the device certificate is not trusted by AWS IoT Core

**Command: tedge connect aws**

```text
connect to Aws cloud.:
        device id: tedge0001
        cloud profile: <none>
        cloud host: a1vto0hq4ubn5n-ats.iot.us-east-1.amazonaws.com:8883
        auth type: Certificate
        certificate file: /etc/tedge/device-certs/tedge-certificate.pem
        cryptoki: false
        bridge: mosquitto
        service manager: systemd
        proxy: Not configured
Restarting mosquitto... ✓
Waiting for mosquitto to be listening for connections... ✓
Verifying device is connected to cloud... ✓
Connection test failed, attempt 1 of 5

Verifying device is connected to cloud... ✓
Connection test failed, attempt 2 of 5

Verifying device is connected to cloud... ✓
Connection test failed, attempt 3 of 5

Verifying device is connected to cloud... ✓
Connection test failed, attempt 4 of 5

Verifying device is connected to cloud... ✓
warning: Bridge has been configured, but Aws connection check failed.
Enabling tedge-mapper-aws... ✓
```

**To Reproduce**

You can do this manually, however it is easier to get setup using the tedge-demo-container and the c8y-tedge go-c8y-cli extension

1. Start a tedge demo container without running the bootstrapping, and open a shell into it

    ```sh
    c8y tedge demo start tedge0001 --skip-bootstrap
    c8y tedge demo shell tedge0001
    ```

2. Create a self-signed certificate

    ```sh
    tedge cert create --device-id tedge0001
    ```
3. Set a valid AWS IoT Core URL

    ```sh
[...]

PR #3654, opened 2025-05-30
Reenable connect checks for aws and azure

Bravo555:
## Proposed changes

Reenable connect checks for aws and azure.

It was disabled in f40a93c90, possibly by mistake because we set a variable used only when connecting to c8y. However, aside from setting that variable we emit a warning if the check is not successful, which is desired. As such, connection checks for these clouds were reenabled.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/connect/command.rs
+++ b/crates/core/tedge/src/cli/connect/command.rs
@@ -227,9 +227,7 @@ impl ConnectCommand {
             self.start_mapper().await;
         }
 
-        #[cfg(feature = "c8y")]
         let mut connection_check_success = true;
-        #[cfg(feature = "c8y")]
         if !self.offline_mode {
             match self
                 .check_connection_with_retries(
@@ -445,7 +443,6 @@ impl ConnectCommand {
         }
     }
 
-    #[cfg(feature = "c8y")]
     async fn check_connection_with_retries(
         &self,
         tedge_config: &TEdgeConfig,


reubenmiller:
@Bravo555 Would this explain https://github.com/thin-edge/thin-edge.io/issues/3640?

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -227,9 +227,7 @@ impl ConnectCommand {
             self.start_mapper().await;
         }
 
-        #[cfg(feature = "c8y")]
         let mut connection_check_success = true;
```
rina23q, line 230: I guess that this unused variable warning was misleading and caused the mistake.
 
```
warning: variable `connection_check_success` is assigned to, but never used
   --> crates/core/tedge/src/cli/connect/command.rs:231:17
    |
231 |         let mut connection_check_success = true;
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: consider using `_connection_check_success` instead
    = note: `#[warn(unused_variables)]` on by default
```

However, I couldn't come up with a good refactoring idea to solve this... connection check has to be done before the mapper starts, and `tenant_matches_configured_url()` needs a connection to the cloud. So, we can't remove the connection check condition here.

```rust
if !use_basic_auth && !self.offline_mode && connection_check_success {
    let _ = self.tenant_matches_configured_url(tedge_config).await;
}
```

Though, honestly, this one can be simplified to this. `tenant_matches_configured_url()` checks `basic_auth` and `offline_mode` inside the function. But this is unrelated to this PR's purpose.
```rust
if connection_check_success {
    let _ = self.tenant_matches_configured_url(tedge_config).await;
}
```

rina23q (approved):
Definitely a good catch and truly connection check was skipped when I built a binary by only `azure` feature (`cargo build --no-default-features --features "azure"
`)

```
$ sudo ./target/debug/tedge connect az 
connect to Azure cloud.:
	device id: rina0010
	cloud profile: <none>
	cloud host: thin-edge-test-hub.azure-devices.net:8883
	auth type: Certificate
	cryptoki: off
	bridge: mosquitto
	service manager: systemd
	proxy: Not configured
Restarting mosquitto... ✓
Waiting for mosquitto to be listening for connections... ✓
Enabling tedge-mapper-az... ✓
```

`Verifying device is connected to cloud... ✓` was missing from the output, so it's skipped.

And after removing feature flags as in this PR, the message is back.

PR #3656, opened 2025-05-30
fix: aws and az fail when no response on connection check

Bravo555:
## TODO

- [x] tests

## Proposed changes

When connecting to aws and azure, during "Verifying device is connected to cloud..." stage, if we don't get response from a remote broker, mark the spinner as failed instead of OK.

It was being marked as OK, because the code was a bit too c8y-centric: for c8y we connect directly to the remote broker to create a device and after that switch to the bridge. When using the bridge, if remote broker doesn't respond, we know that the bridge is connected with remote broker because it's using the same configuration as the direct connection, which was successful, so we consider device status unknown instead of failing the connection.

## Types of changes

## Paste Link to the issue

#3640

## Checklist


crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -274,7 +270,11 @@ impl ConnectCommand {
             Cloud::Azure(_) => (),
         }
 
-        Ok(())
+        if connection_check_success {
+            Ok(())
+        } else {
+            Err(anyhow::anyhow!("Connection check failed").into())
+        }
```
Bravo555, line 277: note: To make the `tedge connect aws` return a non-zero exit code when the connection check failed had to add this here, which can influence `c8y` where failing the connection check is apparently not a failure? I'm not sure why, and if it's intended or accidental.

didier-wenzek: `tedge connect c8y` ignores connection check errors since https://github.com/thin-edge/thin-edge.io/pull/218.
@rina23q do you remember the motivation for this behavior?

That said this is not my main concern regarding `connect/command.rs`. There is so many alternative paths for c8y, aws and az that its really difficult to tell what is done for each. For instance, a few lines above this one, the agent is restarted for c8y but not for aws and az. Why? Also, you have to know that `enable_software_management` means restart the agent.

rina23q: Yeah I remember the decision.
At the time, the purpose of `tedge connect` was to establish the bridge to the cloud, and the connection check was considered an optional check because it may fail, for example, when the server is not responded, that's not thin-edge side of failure. That's why we decided to have exit code `0`.

However, now we have `--offline` option. So, I don't mind if we return non-zero exist code when connection check fails.

didier-wenzek (approved):
Approved

reubenmiller (approved):
Approved. Tested with AWS and it was good (both negative and positive test cases).

I pushed a minor commit to improve the readability of some of the error messages for all clouds, b347575e9c9887d0d049950de1f3b3d26440aeb0

PR #3661, opened 2025-06-02
refactor: support dynamic MQTT subscriptions in `mqtt-channel`/`tedge-mqtt-ext`

jarhodes314:
## Proposed changes
For the custom mapper, the user specifies which topics to subscribe the mapping pipelines to. In order to support dynamically reloading these pipelines, we need to be able to update the subscriptions. This was originally done in #3626, but this PR extracts out the changes for separate review.

To properly support dynamic subscriptions, the list of subscribers and their topic filters has been replaced with a trie.  The general idea is that this structure maps topics to ids of subscribed clients, allowing us to efficiently retrieve the subscribed clients without checking every possible subscription. It manages which topics we subscribe to, ensuring we subscribe to a minimal set of topics to cover the currently subscribed clients, and (crucially) allows us to unsubscribe when topics are no longer consumed. Since this datastructure is separate, the goal is to have 100% unit test coverage here before merging (at the time of writing it is close to this, but not quite there).

How the trie is structured is detailed in https://github.com/thin-edge/thin-edge.io/blob/d8f33027b280b116d919f3b20ac05c040651a7a5/crates/extensions/tedge_mqtt_ext/src/trie.rs#L7-L77

One thing this PR doesn't do currently is handle retain messages. The effect of this is that if client is subscribed to a topic with a retain message and then another client subscribes, this new client does not receive the retain message since no MQTT subscription is created in this case.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/config.rs
+++ b/crates/common/mqtt_channel/src/config.rs
@@ -11,7 +11,6 @@ use std::fmt::Formatter;
 use std::ops::Deref;
 use std::path::Path;
 use std::sync::Arc;
-use std::sync::Mutex;
 use zeroize::Zeroizing;
 
 pub const MAX_PACKET_SIZE: usize = 268435455;
@@ -33,7 +32,7 @@ pub struct Config {
     /// The list of topics to subscribe to on connect
     ///
     /// Default: An empty topic list
-    pub subscriptions: Arc<Mutex<TopicFilter>>,
+    pub subscriptions: TopicFilter,
 
     /// Clean the MQTT session upon connect if set to `true`.
     ///
@@ -164,7 +163,7 @@ impl Default for Config {
                 authentication: None,
             },
             session_name: None,
-            subscriptions: Arc::new(Mutex::new(TopicFilter::empty())),
+            subscriptions: TopicFilter::empty(),
             clean_session: false,
             queue_capacity: 1024,
             max_packet_size: 16 * 1024 * 1024,
@@ -214,8 +213,8 @@ impl Config {
     /// Add a list of topics to subscribe to on connect
     ///
     /// Can be called several times to subscribe to many topics.
-    pub fn with_subscriptions(self, topics: TopicFilter) -> Self {
-        self.subscriptions.lock().unwrap().add_all(topics);
+    pub fn with_subscriptions(mut self, topics: TopicFilter) -> Self {
+        self.subscriptions.add_all(topics);
         self
     }
 
--- a/crates/common/mqtt_channel/src/config.rs
+++ b/crates/common/mqtt_channel/src/config.rs
@@ -6,12 +6,12 @@ use log::debug;
 use rumqttc::tokio_rustls::rustls;
 use rumqttc::tokio_rustls::rustls::pki_types::CertificateDer;
 use rumqttc::LastWill;
-use std::sync::Mutex;
 use std::fmt::Debug;
 use std::fmt::Formatter;
 use std::ops::Deref;
 use std::path::Path;
 use std::sync::Arc;
+use std::sync::Mutex;
 use zeroize::Zeroizing;
 
 pub const MAX_PACKET_SIZE: usize = 268435455;
--- a/crates/common/mqtt_channel/src/config.rs
+++ b/crates/common/mqtt_channel/src/config.rs
@@ -6,6 +6,7 @@ use log::debug;
 use rumqttc::tokio_rustls::rustls;
 use rumqttc::tokio_rustls::rustls::pki_types::CertificateDer;
 use rumqttc::LastWill;
+use std::sync::Mutex;
 use std::fmt::Debug;
 use std::fmt::Formatter;
 use std::ops::Deref;
@@ -32,7 +33,7 @@ pub struct Config {
     /// The list of topics to subscribe to on connect
     ///
     /// Default: An empty topic list
-    pub subscriptions: TopicFilter,
+    pub subscriptions: Arc<Mutex<TopicFilter>>,
 
     /// Clean the MQTT session upon connect if set to `true`.
     ///
@@ -163,7 +164,7 @@ impl Default for Config {
                 authentication: None,
             },
             session_name: None,
-            subscriptions: TopicFilter::empty(),
+            subscriptions: Arc::new(Mutex::new(TopicFilter::empty())),
             clean_session: false,
             queue_capacity: 1024,
             max_packet_size: 16 * 1024 * 1024,
@@ -213,8 +214,8 @@ impl Config {
     /// Add a list of topics to subscribe to on connect
     ///
     /// Can be called several times to subscribe to many topics.
-    pub fn with_subscriptions(mut self, topics: TopicFilter) -> Self {
-        self.subscriptions.add_all(topics);
+    pub fn with_subscriptions(self, topics: TopicFilter) -> Self {
+        self.subscriptions.lock().unwrap().add_all(topics);
         self
     }
 
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -164,9 +164,15 @@ impl Connection {
         let (published_sender, published_receiver) = mpsc::unbounded();
         let (error_sender, error_receiver) = mpsc::unbounded();
         let (pub_done_sender, pub_done_receiver) = oneshot::channel();
+        let subscriptions = Arc::new(Mutex::new(config.subscriptions.clone()));
 
-        let (mqtt_client, event_loop) =
-            Connection::open(config, received_sender.clone(), error_sender.clone()).await?;
+        let (mqtt_client, event_loop) = Connection::open(
+            config,
+            received_sender.clone(),
+            error_sender.clone(),
+            subscriptions.clone(),
+        )
+        .await?;
         let permits = Arc::new(Semaphore::new(1));
         let permit = permits.clone().acquire_owned().await.unwrap();
         let pub_count = Arc::new(AtomicUsize::new(0));
@@ -179,6 +185,7 @@ impl Connection {
             pub_done_sender,
             permits,
             pub_count.clone(),
+            subscriptions.clone(),
         ));
         tokio::spawn(Connection::sender_loop(
             mqtt_client.clone(),
@@ -194,7 +201,7 @@ impl Connection {
             published: published_sender,
             errors: error_receiver,
             pub_done: pub_done_receiver,
-            subscriptions: SubscriberHandle::new(mqtt_client, config.subscriptions.clone()),
+            subscriptions: SubscriberHandle::new(mqtt_client, subscriptions),
         })
     }
 
@@ -207,6 +214,7 @@ impl Connection {
         config: &Config,
         mut message_sender: mpsc::UnboundedSender<MqttMessage>,
         mut error_sender: mpsc::UnboundedSender<MqttError>,
+        subscriptions: Arc<Mutex<TopicFilter>>,
     ) -> Result<(AsyncClient, EventLoop), MqttError> {
         const INSECURE_MQTT_PORT: u16 = 1883;
         const SECURE_MQTT_PORT: u16 = 8883;
@@ -234,7 +242,7 @@ impl Connection {
                     };
                     info!(target: "MQTT", "Connection established");
 
-                    let subscriptions = config.subscriptions.lock().unwrap().filters();
+                    let subscriptions = subscriptions.lock().unwrap().filters();
 
                     // Need check here otherwise it will hang waiting for a SubAck, and none will come when there is no subscription.
                     if subscriptions.is_empty() {
@@ -291,6 +299,7 @@ impl Connection {
         done: oneshot::Sender<()>,
         permits: Arc<Semaphore>,
         pub_count: Arc<AtomicUsize>,
+        subscriptions: Arc<Mutex<TopicFilter>>,
     ) -> Result<(), MqttError> {
         let mut triggered_disconnect = false;
         let mut disconnect_permit = None;
@@ -363,7 +372,7 @@ impl Connection {
                             // If session_name is not provided or if the broker session persistence
                             // is not enabled or working, then re-subscribe
 
-                            let subscriptions = config.subscriptions.lock().unwrap().filters();
+                            let subscriptions = subscriptions.lock().unwrap().filters();
                             // Need check here otherwise it will hang waiting for a SubAck, and none will come when there is no subscription.
                             if subscriptions.is_empty() {
                                 break;
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -4,6 +4,7 @@ use crate::MqttError;
 use crate::MqttMessage;
 use crate::PubChannel;
 use crate::SubChannel;
+use crate::TopicFilter;
 use futures::channel::mpsc;
 use futures::channel::oneshot;
 use futures::SinkExt;
@@ -20,6 +21,7 @@ use std::collections::HashSet;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use std::sync::Arc;
+use std::sync::Mutex;
 use std::time::Duration;
 use tokio::sync::OwnedSemaphorePermit;
 use tokio::sync::Semaphore;
@@ -38,6 +40,10 @@ pub struct Connection {
 
     /// A channel to notify that all the published messages have been actually published.
     pub pub_done: oneshot::Receiver<()>,
+
+    pub client: AsyncClient,
+
+    pub subscriptions: Arc<Mutex<TopicFilter>>,
 }
 
 impl Connection {
@@ -108,7 +114,7 @@ impl Connection {
             pub_count.clone(),
         ));
         tokio::spawn(Connection::sender_loop(
-            mqtt_client,
+            mqtt_client.clone(),
             published_receiver,
             error_sender,
             config.last_will_message.clone(),
@@ -121,6 +127,8 @@ impl Connection {
             published: published_sender,
             errors: error_receiver,
             pub_done: pub_done_receiver,
+            client: mqtt_client,
+            subscriptions: config.subscriptions.clone(),
         })
     }
 
@@ -160,7 +168,7 @@ impl Connection {
                     };
                     info!(target: "MQTT", "Connection established");
 
-                    let subscriptions = config.subscriptions.filters();
+                    let subscriptions = config.subscriptions.lock().unwrap().filters();
 
                     // Need check here otherwise it will hang waiting for a SubAck, and none will come when there is no subscription.
                     if subscriptions.is_empty() {
@@ -289,7 +297,7 @@ impl Connection {
                             // If session_name is not provided or if the broker session persistence
                             // is not enabled or working, then re-subscribe
 
-                            let subscriptions = config.subscriptions.filters();
+                            let subscriptions = config.subscriptions.lock().unwrap().filters();
                             // Need check here otherwise it will hang waiting for a SubAck, and none will come when there is no subscription.
                             if subscriptions.is_empty() {
                                 break;
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -17,6 +17,7 @@ use rumqttc::Event;
 use rumqttc::EventLoop;
 use rumqttc::Outgoing;
 use rumqttc::Packet;
+use rumqttc::SubscribeFilter;
 use std::collections::HashSet;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
@@ -41,9 +42,75 @@ pub struct Connection {
     /// A channel to notify that all the published messages have been actually published.
     pub pub_done: oneshot::Receiver<()>,
 
-    pub client: AsyncClient,
+    pub subscriptions: SubscriberHandle,
+}
 
-    pub subscriptions: Arc<Mutex<TopicFilter>>,
+#[derive(Clone)]
+/// A client for changing the subscribed topics
+pub struct SubscriberHandle {
+    client: AsyncClient,
+    subscriptions: Arc<Mutex<TopicFilter>>,
+}
+
+impl SubscriberHandle {
+    pub fn new(client: AsyncClient, subscriptions: Arc<Mutex<TopicFilter>>) -> Self {
+        Self {
+            client,
+            subscriptions,
+        }
+    }
+}
+
+#[async_trait::async_trait]
+pub trait SubscriberOps {
+    async fn subscribe_many(
+        &self,
+        topics: impl IntoIterator<Item = String> + Send,
+    ) -> Result<(), MqttError>;
+    async fn unsubscribe_many(
+        &self,
+        topics: impl IntoIterator<Item = String> + Send,
+    ) -> Result<(), MqttError>;
+}
+
+#[async_trait::async_trait]
+impl SubscriberOps for SubscriberHandle {
+    async fn subscribe_many(
+        &self,
+        topics: impl IntoIterator<Item = String> + Send,
+    ) -> Result<(), MqttError> {
+        let topics = topics.into_iter().collect::<Vec<_>>();
+        {
+            let mut subs = self.subscriptions.lock().unwrap();
+            for topic in &topics {
+                subs.add(topic)?;
+            }
+        }
+        self.client
+            .subscribe_many(topics.into_iter().map(|path| SubscribeFilter {
+                path,
+                qos: rumqttc::QoS::AtLeastOnce,
+            }))
+            .await?;
+        Ok(())
+    }
+
+    async fn unsubscribe_many(
+        &self,
+        topics: impl IntoIterator<Item = String> + Send,
+    ) -> Result<(), MqttError> {
+        let topics = topics.into_iter().collect::<Vec<_>>();
+        {
+            let mut subs = self.subscriptions.lock().unwrap();
+            for topic in &topics {
+                subs.add(topic)?;
+            }
+        }
+        for topic in topics {
+            self.client.unsubscribe(topic).await?;
+        }
+        Ok(())
+    }
 }
 
 impl Connection {
@@ -127,8 +194,7 @@ impl Connection {
             published: published_sender,
             errors: error_receiver,
             pub_done: pub_done_receiver,
-            client: mqtt_client,
-            subscriptions: config.subscriptions.clone(),
+            subscriptions: SubscriberHandle::new(mqtt_client, config.subscriptions.clone()),
         })
     }
 
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -49,7 +49,7 @@ pub struct Connection {
 /// A client for changing the subscribed topics
 pub struct SubscriberHandle {
     client: AsyncClient,
-    subscriptions: Arc<Mutex<TopicFilter>>,
+    pub(crate) subscriptions: Arc<Mutex<TopicFilter>>,
 }
 
 impl SubscriberHandle {
@@ -103,7 +103,7 @@ impl SubscriberOps for SubscriberHandle {
         {
             let mut subs = self.subscriptions.lock().unwrap();
             for topic in &topics {
-                subs.add(topic)?;
+                subs.remove(topic);
             }
         }
         for topic in topics {
--- a/crates/common/mqtt_channel/src/lib.rs
+++ b/crates/common/mqtt_channel/src/lib.rs
@@ -54,3 +54,4 @@ pub use futures::Stream;
 pub use futures::StreamExt;
 
 pub use rumqttc::QoS;
+pub use rumqttc::AsyncClient;
--- a/crates/common/mqtt_channel/src/lib.rs
+++ b/crates/common/mqtt_channel/src/lib.rs
@@ -53,5 +53,4 @@ pub use futures::SinkExt;
 pub use futures::Stream;
 pub use futures::StreamExt;
 
-pub use rumqttc::AsyncClient;
 pub use rumqttc::QoS;
--- a/crates/common/mqtt_channel/src/lib.rs
+++ b/crates/common/mqtt_channel/src/lib.rs
@@ -53,5 +53,5 @@ pub use futures::SinkExt;
 pub use futures::Stream;
 pub use futures::StreamExt;
 
-pub use rumqttc::QoS;
 pub use rumqttc::AsyncClient;
+pub use rumqttc::QoS;
--- a/crates/extensions/tedge_mqtt_ext/Cargo.toml
+++ b/crates/extensions/tedge_mqtt_ext/Cargo.toml
@@ -18,6 +18,7 @@ test-helpers = ["dep:assert-json-diff"]
 assert-json-diff = { workspace = true, optional = true }
 async-trait = { workspace = true }
 mqtt_channel = { workspace = true }
+rumqttc = { workspace = true }
 serde_json = { workspace = true }
 tedge_actors = { workspace = true }
 tedge_utils = { workspace = true }
--- a/crates/extensions/tedge_mqtt_ext/Cargo.toml
+++ b/crates/extensions/tedge_mqtt_ext/Cargo.toml
@@ -27,6 +27,7 @@ tracing = { workspace = true }
 [dev-dependencies]
 futures = { workspace = true }
 mqtt_tests = { path = "../../tests/mqtt_tests" }
+proptest = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/extensions/tedge_mqtt_ext/Cargo.toml
+++ b/crates/extensions/tedge_mqtt_ext/Cargo.toml
@@ -18,7 +18,6 @@ test-helpers = ["dep:assert-json-diff"]
 assert-json-diff = { workspace = true, optional = true }
 async-trait = { workspace = true }
 mqtt_channel = { workspace = true }
-rumqttc = { workspace = true }
 serde_json = { workspace = true }
 tedge_actors = { workspace = true }
 tedge_utils = { workspace = true }
--- a/crates/extensions/tedge_mqtt_ext/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_ext/src/lib.rs
@@ -5,8 +5,14 @@ mod tests;
 
 use async_trait::async_trait;
 use mqtt_channel::Connection;
+pub use mqtt_channel::DebugPayload;
+pub use mqtt_channel::MqttError;
+pub use mqtt_channel::MqttMessage;
+pub use mqtt_channel::QoS;
 use mqtt_channel::SinkExt;
 use mqtt_channel::StreamExt;
+pub use mqtt_channel::Topic;
+pub use mqtt_channel::TopicFilter;
 use std::convert::Infallible;
 use std::sync::Arc;
 use std::sync::Mutex;
@@ -25,19 +31,13 @@ use tedge_actors::Sender;
 use tracing::info;
 
 pub type MqttConfig = mqtt_channel::Config;
-pub use mqtt_channel::DebugPayload;
-pub use mqtt_channel::MqttError;
-pub use mqtt_channel::MqttMessage;
-pub use mqtt_channel::QoS;
-pub use mqtt_channel::Topic;
-pub use mqtt_channel::TopicFilter;
 
 pub struct MqttActorBuilder {
     mqtt_config: mqtt_channel::Config,
     input_receiver: InputCombiner,
     pub_or_sub_sender: PubOrSubSender,
     publish_sender: mpsc::Sender<MqttMessage>,
-    pub subscriber_addresses: Vec<(TopicFilter, DynSender<MqttMessage>)>,
+    subscriber_addresses: Vec<DynMqttSender>,
     signal_sender: mpsc::Sender<RuntimeRequest>,
 }
 
@@ -48,7 +48,7 @@ struct InputCombiner {
 }
 
 #[derive(Debug)]
-pub enum PublishOrSubscribe{
+pub enum PublishOrSubscribe {
     Publish(MqttMessage),
     Subscribe(SubscriptionRequest),
 }
@@ -119,8 +119,8 @@ impl MqttActorBuilder {
 
     pub(crate) fn build_actor(self) -> MqttActor {
         let mut combined_topic_filter = TopicFilter::empty();
-        for (topic_filter, _) in self.subscriber_addresses.iter() {
-            combined_topic_filter.add_all(topic_filter.to_owned());
+        for subscriber in self.subscriber_addresses.iter() {
+            combined_topic_filter.add_all(subscriber.topic_filter());
         }
 
         let removed = combined_topic_filter.remove_overlapping_patterns();
@@ -143,15 +143,71 @@ impl AsMut<MqttConfig> for MqttActorBuilder {
 }
 
 impl MessageSource<MqttMessage, TopicFilter> for MqttActorBuilder {
-    fn connect_sink(&mut self, subscriptions: TopicFilter, peer: &impl MessageSink<MqttMessage>) {
-        let sender = peer.get_sender();
-        self.subscriber_addresses.push((subscriptions, sender));
+    fn connect_sink(&mut self, topics: TopicFilter, peer: &impl MessageSink<MqttMessage>) {
+        let sender = DynMqttSender {
+            topics: Subscription::Static(topics),
+            sender: peer.get_sender(),
+        };
+        self.subscriber_addresses.push(sender);
     }
 }
 
 impl MessageSink<MqttMessage> for MqttActorBuilder {
     fn get_sender(&self) -> DynSender<MqttMessage> {
         self.publish_sender.clone().into()
+    }
+}
+
+impl MessageSource<MqttMessage, Subscription> for MqttActorBuilder {
+    fn connect_sink(&mut self, topics: Subscription, peer: &impl MessageSink<MqttMessage>) {
+        let sender = DynMqttSender {
+            topics,
+            sender: peer.get_sender(),
+        };
+        self.subscriber_addresses.push(sender);
+    }
+}
+
+pub enum Subscription {
+    Static(TopicFilter),
+    Dynamic(Arc<Mutex<TopicFilter>>),
+}
+
+impl Subscription {
+    pub fn topic_filter(&self) -> TopicFilter {
+        match self {
+            Subscription::Static(topic_filter) => topic_filter.clone(),
+            Subscription::Dynamic(topic_filter) => topic_filter.lock().unwrap().clone(),
+        }
+    }
+
+    pub fn accept(&self, message: &MqttMessage) -> bool {
+        match self {
+            Subscription::Static(topic_filter) => topic_filter.accept(message),
+            Subscription::Dynamic(topic_filter) => topic_filter.lock().unwrap().accept(message),
+        }
+    }
+}
+
+struct DynMqttSender {
+    topics: Subscription,
+    sender: DynSender<MqttMessage>,
+}
+
+impl DynMqttSender {
+    pub fn topic_filter(&self) -> TopicFilter {
+        self.topics.topic_filter()
+    }
+}
+
+#[async_trait]
+impl Sender<MqttMessage> for DynMqttSender {
+    async fn send(&mut self, message: MqttMessage) -> Result<(), ChannelError> {
+        if self.topics.accept(&message) {
+            self.sender.send(message).await
+        } else {
+            Ok(())
+        }
     }
 }
 
@@ -166,10 +222,14 @@ struct PubOrSubSender {
 
 #[async_trait]
 impl Sender<PublishOrSubscribe> for PubOrSubSender {
-    async fn send(&mut self, message:PublishOrSubscribe) ->  Result<(), ChannelError> {
+    async fn send(&mut self, message: PublishOrSubscribe) -> Result<(), ChannelError> {
         match message {
-            PublishOrSubscribe::Publish(msg) => Sender::<_>::send(&mut self.publish_sender, msg).await,
-            PublishOrSubscribe::Subscribe(sub) => Sender::<_>::send(&mut self.subscription_request_sender, sub).await,
+            PublishOrSubscribe::Publish(msg) => {
+                Sender::<_>::send(&mut self.publish_sender, msg).await
+            }
+            PublishOrSubscribe::Subscribe(sub) => {
+                Sender::<_>::send(&mut self.subscription_request_sender, sub).await
+            }
         }
     }
 }
@@ -203,7 +263,7 @@ pub struct FromPeers {
 }
 
 pub struct ToPeers {
-    peer_senders: Vec<(TopicFilter, DynSender<MqttMessage>)>,
+    peer_senders: Vec<DynMqttSender>,
 }
 
 impl FromPeers {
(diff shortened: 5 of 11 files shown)


crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -109,33 +258,71 @@ impl Builder<MqttActor> for MqttActorBuilder {
...
-    peer_senders: Vec<(TopicFilter, DynSender<MqttMessage>)>,
+    peer_senders: Vec<DynMqttSender>,
 }
 
 impl FromPeers {
     async fn relay_messages_to(
         &mut self,
         outgoing_mqtt: &mut mpsc::UnboundedSender<MqttMessage>,
+        client: mqtt_channel::AsyncClient,
+        subscriptions: Arc<Mutex<TopicFilter>>,
```
didier-wenzek: I would combine these two fields into a single one  that handles the whole logic below to compute the new subscription set and update `mqtt_channel::AsyncClient`. I would even move all these details in the `mqtt_channel` module.

Indeed, these two fields are both extracted from the same `mqtt_client` when calling `relay_messages_to`. Furthermore, they are used no where else.

crates/common/mqtt_channel/src/lib.rs
```diff
@@ -53,4 +53,5 @@ pub use futures::SinkExt;
 pub use futures::Stream;
 pub use futures::StreamExt;
 
+pub use rumqttc::AsyncClient;
```
didier-wenzek: I would avoid to leak the internal client type, exposing only `mqtt_channel::Connection`.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -32,7 +33,7 @@ pub struct Config {
     /// The list of topics to subscribe to on connect
     ///
     /// Default: An empty topic list
-    pub subscriptions: TopicFilter,
+    pub subscriptions: Arc<Mutex<TopicFilter>>,
```
didier-wenzek: As the subscriptions are never updated through the config but using the connection, I would consider to remove the use of interior mutability here, and to rather use these subscriptions as an initial value on connect.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -109,33 +258,71 @@ impl Builder<MqttActor> for MqttActorBuilder {
...
+                        let mut subscriptions = subscriptions.lock().unwrap();
+                        subscriptions.add_all(request.0.clone());
+                        subscriptions.remove_overlapping_patterns()
+                    };
+                    let mut to_unsub = TopicFilter::empty();
+                    for o in overlap {
+                        if request.0.remove(&o).is_none() {
+                            to_unsub.add_unchecked(&o);
+                        }
+                    }
```
Bravo555: this logic is a bit difficult to follow
if current and new subscriptions were sets, i'd expect something like this:

```rust
let (to_sub, to_unsub) = {
    let mut current_subscriptions = subscriptions.lock().unwrap();
    let new_subscriptions = request.0;

    let to_sub = new_subscriptions.difference(current_subscriptions);
    let to_unsub = current_subscriptions.difference(new_subscriptions);

    (to_sub, to_unsub)
};
```

But i understand that what we also need to do here is collapsing overlapping topics, e.g. `["topic/#", "topic/a"] => ["topic/#"]`
however, I fail to understand how this is done using `remove_overlapping_patterns`. E.g. in this example of current and requested topic subscriptions:

```
current:
a
b
c

request:
a
b
d

after adding all:
a (+)
b (+)
c
d

overlap:
a
b

to_unsub (overlap - request):


to_sub (request - overlap):
d
```

if I've traced the steps correctly, we should unsub from c, but we don't. However we do if `c` is part of the overlap.

In any case I think this part could use a bit more explanation, as well as an example of what an overlap is and is not.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -17,7 +17,7 @@ Test Tags           adapter:docker    theme:cryptoki
 
 
 *** Variables ***
-${CERT_TEMPLATE}     /etc/tedge/hsm/cert.template    # created by init_softhsm script
+${CERT_TEMPLATE}    /etc/tedge/hsm/cert.template    # created by init_softhsm script
```
Bravo555: I added commit resolving this to my PR here
https://github.com/thin-edge/thin-edge.io/pull/3656/commits/2f09de69832865f761152dee7a676e388419ac80

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -90,6 +157,88 @@ impl MessageSink<MqttMessage> for MqttActorBuilder {
...
+            topics,
+            sender: peer.get_sender(),
+        };
+        self.subscriber_addresses.push(sender);
+    }
+}
+
+pub enum Subscription {
+    Static(TopicFilter),
+    Dynamic(Arc<Mutex<TopicFilter>>),
```
albinsuresh: Do we really need 2 different variants? Why not just the mutable variant which can be initialised with the static value as well when needed?

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -20,42 +27,99 @@ use tedge_actors::RuntimeError;
...
+struct InputCombiner {
+    publish_receiver: mpsc::Receiver<MqttMessage>,
+    subscription_request_receiver: mpsc::Receiver<SubscriptionRequest>,
+    signal_receiver: mpsc::Receiver<RuntimeRequest>,
+}
+
+#[derive(Debug)]
+pub enum PublishOrSubscribe {
+    Publish(MqttMessage),
+    Subscribe(SubscriptionRequest),
```
albinsuresh, line 64: Are we expecting an `Unsubscribe` request variant as well? Or the idea is to derive the topics to unsubscribed, by comparing the new subscriptions with the old subscriptions and extracting the difference?

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -109,33 +258,71 @@ impl Builder<MqttActor> for MqttActorBuilder {
...
+                    let to_sub = request.0;
+                    let to_sub = to_sub.filters();
+                    if !to_sub.is_empty() {
+                        info!("Updating MQTT subscription to include {to_sub:?}");
+                        client.subscribe_many(to_sub).await.unwrap();
+                    }
+                    if !to_unsub.patterns().is_empty() {
+                        info!("Updating MQTT subscription to not include {to_unsub:?}");
+                        for topic in to_unsub.patterns() {
+                            client.unsubscribe(topic).await.unwrap();
```
albinsuresh: This unsubscribe logic appears problematic in an environment with multiple peer actors having different subscription needs. If one peer actor is already subscribed to `topic1` and `topic2`, and a second peer subscribes to `topic3` only, wouldn't that make this mqtt actor unsubscribe from `topic1` and `topic2` as they are not included in the new subscription list, leaving peer1 broken without its subscriptions?

albinsuresh (commented):
Adding some integration tests in `tests.rs` to showcase the usage of the dynamic subscribe/unsubscribe feature would be good, as a system test probably isn't feasible in this state.

crates/extensions/tedge_mqtt_ext/src/lib.rs
didier-wenzek: This is to avoid the cost of locking a mutex for each and every MQTT messages.

Actually, this mutex is my main concern with this PR. This is my proposal but I'm not happy with it. Not only there is a full scan over all the subscribers each time a message is received (to find those interested by this message), but there is a mutex lock for each dynamic subscriber (even when this is to conclude that the subscriber is not interested by the message).

What is the extra costs in practice? A benchmark could help here.  

An alternative design is to organize the senders in a prefix tree (aka a trie), where a list of senders is attach to each node of the subscription topic filters. The sender of a subscriber subscribing to many topics will have to be cloned but no lock nor full scan will be required. The implementation is a bit more involved though.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: It could definitely use more explanation, and I will add that. The logic is designed to only unsubscribe in a specific case which is when we upgrade from a subscription to a single topic to a wildcard (e.g. if we were previously subscribed to `a/b` and now wish to subscribe to `a/#`). This avoids us having overlapping subscriptions. It will also avoid subscribing to topics that are already subscribed to, as you identified.

Unsubscribing from unused topics would require knowing what all consumers of this actor are currently subscribed to and we don't have that information here. All we have here is a list of topics that one consumer wants to be subscribed to, and the list of topics we are subscribed to.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: No, this is precisely why the logic discussed in https://github.com/thin-edge/thin-edge.io/pull/3661#discussion_r2123481024 is weird. We only unsubscribe from topics in the case where we were subscribed to a single topic and this is "upgraded" to a wildcard subscription.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: The idea with the current implementation is to never unsubscribe, apart from where duplicate subscriptions would be created.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,894 @@
...
+                .map(|t| format!("{prefix}/{t}"))
+                .collect(),
+        }
+    }
+}
+
+impl PartialEq for Diff {
+    fn eq(&self, other: &Self) -> bool {
+        let mut s = HashSet::new();
+        s.extend(&self.subscribe);
```
didier-wenzek: I would consider to directly use `HashSet` instead of `Vec` for `Diff`.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1098 @@
...
+                        diff
+                    }
+                } else {
+                    Diff::empty()
+                }
+            }
+        }
+    }
+
+    // TODO this should only accept valid message topic, not containing wildcards
```
didier-wenzek: In practice, this is check has already be done (because the message has been received from MQTT or wrapped into a `Topic`).

So I would simply enforced that at the top level: `MqtTrie::matches<'a>(&'a self, topic: &Topic) -> Vec<&'a T>` and adds a `debug_assert!` here on the head.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1098 @@
...
+                (Some(_), Some(_), _) => break None,
+                (None, Some("#"), Some(Winner::This)) => break None,
+                (None, Some("#"), _) => break Some(Ordering::Less),
+                (None, Some(_), _) => break None,
+                (Some("#"), None, Some(Winner::Other)) => break None,
+                (Some("#"), None, _) => break Some(Ordering::Greater),
+                (Some(_), None, _) => break None,
+                (None, None, Some(Winner::This)) => break Some(Ordering::Greater),
+                (None, None, Some(Winner::Other)) => break Some(Ordering::Less),
+                (None, None, None) => break Some(Ordering::Equal),
```
didier-wenzek, line 245: It's not easy to grasp the logic behind this long list of cases.

The tests help a lot, though.

But is my understand correct? Is the intent to define `topic_filter_1 <= topic_filter_2` as "any topic accepted by topic_filter_1 is also accepted by topic_filter_2"?

I would add a doc comment to clarify this.

didier-wenzek (commented):
The `struct MqtTrie` introduced here is the right abstraction to address dynamic MQTT subscriptions, although a bit involved. Nicely implemented with a good code coverage.

crates/extensions/tedge_mqtt_ext/src/trie.rs
didier-wenzek: I added some missing tests https://github.com/thin-edge/thin-edge.io/pull/3661/commits/236a2ac16696c3cf3fd322f766bf22f4ad9b7c12

jarhodes314: Still need to add a doc comment, although I have now grouped and commented the cases to broadly explain why we need them. Your understanding is correct though

crates/common/mqtt_channel/src/config.rs
jarhodes314: I don't quite understand. The reason this is mutable is because we need to maintain the list of subscribed topics so we can resubscribe if the connection drops at any point.

crates/extensions/tedge_mqtt_ext/Cargo.toml
```diff
@@ -18,6 +18,7 @@ test-helpers = ["dep:assert-json-diff"]
 assert-json-diff = { workspace = true, optional = true }
 async-trait = { workspace = true }
 mqtt_channel = { workspace = true }
+rumqttc = { workspace = true }
```
didier-wenzek: I would avoid adding a direct dependency on `rumqttc`. See also https://github.com/thin-edge/thin-edge.io/pull/3661/files#r2121521308

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -20,54 +29,124 @@ use tedge_actors::RuntimeError;
...
-        for pattern in combined_topic_filter.patterns() {
+        let mut topic_filter = TopicFilter::empty();
+        for pattern in &self.subscription_diff.subscribe {
+            topic_filter.add(pattern).unwrap();
             tracing::info!(target: "MQTT sub", "{pattern}");
         }
-        for pattern in removed {
+        for pattern in self.subscription_diff.unsubscribe {
             tracing::warn!(target: "MQTT sub", "ignoring overlapping subscription to {pattern}");
         }
```
didier-wenzek: Is this message still accurate?

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -78,9 +157,28 @@ impl AsMut<MqttConfig> for MqttActorBuilder {
...
+            client: ClientId(self.current_id),
+            sender: peer.get_sender(),
+        };
+        let client_id = sender.client;
+        self.current_id += 1;
+        for topic in topics.patterns() {
+            self.subscription_diff += self.trie.trie.insert(topic, client_id);
+        }
+        self.subscriber_addresses.push(sender);
+        client_id
```
didier-wenzek: I would re-arrange a bit this code to better encapsulate  the strong relationship between a client id and the index of its sender in the vector of `subscriber_addresses`.

```suggestion
        let client_id =  self.add_new_subscriber(peer.get_sender());
        for topic in topics.patterns() {
            self.subscription_diff += self.trie.trie.insert(topic, client_id);
        }
        client_id
```

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -90,6 +188,103 @@ impl MessageSink<MqttMessage> for MqttActorBuilder {
...
+fan_in_message_type!(TrieRequest[SubscriptionRequest, MatchRequest]: Clone, Debug);
+
+#[derive(Debug)]
+pub enum TrieResponse {
+    Diff(SubscriptionDiff),
+    Matched(Vec<ClientId>),
+}
+
+#[async_trait]
+impl Server for TrieService {
```
didier-wenzek, line 225: What is the motivation to encapsulate the `MqttTrie`  behind a `Service`? Couldn't the `MqttActor` directly include this trie into its state?

crates/extensions/tedge_mqtt_ext/src/tests.rs
```diff
@@ -180,6 +191,65 @@ async fn communicate_over_mqtt() {
...
+        .unwrap();
+    assert_eq!(timeout(msgs_0.recv()).await.unwrap(), msg);
+
+    // Verify that messages
+    let msg = MqttMessage::new(&Topic::new_unchecked("a/b"), "hello");
+    msgs_0
+        .send(PublishOrSubscribe::Publish(msg.clone()))
+        .await
+        .unwrap();
+    assert_eq!(timeout(msgs_1.recv()).await.unwrap(), msg);
```
didier-wenzek: I fail to get the point of this latest check. This seems unrelated to new subscription.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: `FromPeers` and `ToPeers` operate independently (so they can execute concurrently), and both depend on the state of the trie.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -20,54 +29,124 @@ use tedge_actors::RuntimeError;
...
-    input_receiver: CombinedReceiver<MqttMessage>,
+    input_receiver: InputCombiner,
+    pub_or_sub_sender: PubOrSubSender,
     publish_sender: mpsc::Sender<MqttMessage>,
-    pub subscriber_addresses: Vec<(TopicFilter, DynSender<MqttMessage>)>,
+    subscriber_addresses: Vec<DynMqttSender>,
     signal_sender: mpsc::Sender<RuntimeRequest>,
+    trie: TrieService,
+    current_id: usize,
+    subscription_diff: SubscriptionDiff,
```
albinsuresh, line 52: Seeing the `subscription_diff` here felt a bit weird at first, till I realised that it is essentially capturing all the topics requested by peer actors, along with the overlapping ones so that only the "broader topics" are used for subscriptions and the overlapping ones are ignored/logged. That just made me realise that our earlier implementation was allowing overlapping subscriptions and we were probably saved by Mosquitto broker de-duplicating overlapping subscriptions.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1278 @@
...
+    valid_topics
+}
+
+impl<T: Debug + Eq> TrieNode<T> {
+    fn subscribers_matching(&self, topic: &str) -> Vec<String> {
+        match topic.split_once("/") {
+            Some(("+", rest)) => {
+                if self.sub_nodes.contains_key("#") {
+                    vec![]
+                } else {
```
didier-wenzek: I fail to understand this function.

- Why do we have a match for `Some(("+", rest))` but none for Some(("#", rest))` on the input topic?
- Why do we have a check for `self.sub_nodes.contains_key("#")` but none on `self.sub_nodes.contains_key("+")` ?

crates/extensions/tedge_mqtt_ext/src/tests.rs
```diff
@@ -180,6 +191,65 @@ async fn communicate_over_mqtt() {
...
 }
 
+#[tokio::test]
+async fn dynamic_subscriptions() {
+    let broker = mqtt_tests::test_mqtt_broker();
+    let mqtt_config = MqttConfig::default().with_port(broker.port);
+    let mut mqtt = MqttActorBuilder::new(mqtt_config);
+
+    let mut msgs_0 = SimpleMessageBoxBuilder::<_, PublishOrSubscribe>::new("dyn-subscriber", 16);
+    let mut msgs_1 = SimpleMessageBoxBuilder::<_, PublishOrSubscribe>::new("dyn-subscriber1", 16);
```
albinsuresh: Minor readability improvement:
```suggestion
    let mut client_0 = SimpleMessageBoxBuilder::<_, PublishOrSubscribe>::new("dyn-subscriber", 16);
    let mut client_1 = SimpleMessageBoxBuilder::<_, PublishOrSubscribe>::new("dyn-subscriber1", 16);
```

crates/extensions/tedge_mqtt_ext/src/tests.rs
```diff
@@ -180,6 +191,65 @@ async fn communicate_over_mqtt() {
...
+    assert_eq!(timeout(msgs_0.recv()).await.unwrap(), msg);
+
+    // Verify that messages
+    let msg = MqttMessage::new(&Topic::new_unchecked("a/b"), "hello");
+    msgs_0
+        .send(PublishOrSubscribe::Publish(msg.clone()))
+        .await
+        .unwrap();
+    assert_eq!(timeout(msgs_1.recv()).await.unwrap(), msg);
+}
```
albinsuresh, line 250: One additional case worth adding would be to validate that a message published to `a/c` is only received by `client_1` and not `client_0`.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1278 @@
...
+                (Some("#"), Some(_), _) => break Some(Ordering::Greater),
+                (Some(_), Some("#"), Some(Winner::This)) => break None,
+                (Some(_), Some("#"), _) => break Some(Ordering::Less),
+
+                // + > static segment, but keep going as there may be more segments
+                (Some("+"), Some(_), Some(Winner::Other)) => break None,
+                (Some("+"), Some(_), _) => current_winner = Some(Winner::This),
+                (Some(_), Some("+"), Some(Winner::This)) => break None,
+                (Some(_), Some("+"), _) => current_winner = Some(Winner::Other),
+                (Some(_), Some(_), _) => break None,
```
albinsuresh, line 232: Just additional comments:
```suggestion

                // a/.. and b/.. are not comparable
                (Some(_), Some(_), _) => break None,
```

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1278 @@
...
+                None => (),
+            }
+        }
+        valid_topics.insert(topics[i].as_ref().to_owned());
+    }
+    valid_topics
+}
+
+impl<T: Debug + Eq> TrieNode<T> {
+    fn subscribers_matching(&self, topic: &str) -> Vec<String> {
```
albinsuresh:
```suggestion
    fn subscribed_topics_matching(&self, topic: &str) -> Vec<String> {
```
As the term `subscribers` could be wrongly interpreted as the "subscribed clients" and what's returned is the topics and not the client ids.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1278 @@
...
+                .sub_nodes
+                .get(topic)
+                .filter(|node| node.has_subscribers())
+                .map(|_| topic.to_owned())
+                .into_iter()
+                .collect(),
+        }
+    }
+
+    fn subscribers(&self) -> Vec<String> {
```
albinsuresh:
```suggestion
    fn subscribed_topics(&self) -> Vec<String> {
```
For the same reason mentioned above.

albinsuresh (commented):
That's some impressive trie code, and reviewing it was a bit challenging TBH, but the extensive test coverage definitely helped a lot. Though things look fine from the MQTT actor integration perspective, I still need to go deeper into the trie insertion logic, as a single pass isn't enough to fully absorb the complexity of it.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1278 @@
...
+
+    fn is_active(&self) -> bool {
+        !self.subscribers.is_empty() || !self.sub_nodes.is_empty()
+    }
+
+    fn has_subscribers(&self) -> bool {
+        !self.subscribers.is_empty()
+    }
+
+    fn insert(&mut self, topic: &str, subscriber: T) -> SubscriptionDiff {
```
Bravo555: suggestion: The first argument is not really a (full) topic, but a _topic suffix_

```suggestion
    fn insert(&mut self, topic_suffix: &str, subscriber: T) -> SubscriptionDiff {
```

crates/common/mqtt_channel/src/lib.rs
Bravo555: We could also add that type to `clippy.toml`.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: In theory, yes, since unsubscribe represents any subscriptions that were superseded when inserting future subscriptions into the trie (e.g. if we had inserted `a/b` then inserted `a/+`, the latest diff would unsubscribe `a/b`). However in practice we start with an empty trie, so any topics we unsubscribe from must be in another diff's `subscribe` set. This means the subscribe and unsubscribe entries would cancel each other out and be removed when combining the diffs. As a result, unsubscribe should always be empty at this point in the code. Therefore I'll remove these lines.

Bravo555: +1 on that. If I understand correctly, as we connect other actors and they set their subscriptions, which can overlap, as we build the trie we will end up with some items in the `unsubscribe` list because the Trie implicitly assumes that we're immediately changing the MQTT subscriptions, but in this case, as we haven't connected to MQTT broker yet, we can just ignore them.

So really, as trie deduplicating the topics is working as intended and having some duplicate topics is nothing out of the ordinary, perhaps we shouldn't emit this warning and worry the user unnecessarily. What matters is that the user is aware which topics we're subscribing to, and which exact topics were deduplicated is probably a small detail, so maybe `debug` level would be appropriate here.

As for whether it should be removed/changed, possibly, but _technically_ this is a refactor and this is _technically_ a functional change, though it's so small it's probably fine, so if I'm not getting something wrong, I'm for removing it.

crates/extensions/tedge_mqtt_ext/src/trie.rs
Bravo555: Also the `topic` here is really a topic suffix

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1278 @@
...
+                    if node.has_subscribers() {
+                        subs.push(key.clone());
+                    }
+                    subs
+                })
+                .collect()
+        }
+    }
+
+    fn remove(&mut self, topic: &str, id: &T) -> SubscriptionDiff {
```
Bravo555: `topic` is a topic suffix

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1278 @@
...
+                            ..<_>::default()
+                        },
+                    );
+                    self.insert_diff_for(topic)
+                }
+            }
+        }
+    }
+
+    fn insert_diff_for(&self, topic: &str) -> SubscriptionDiff {
```
Bravo555: `topic` is a topic suffix

crates/extensions/tedge_mqtt_ext/src/tests.rs
jarhodes314: Yeah I think I got part way through writing the check before realising it was difficult to actually assert what I wanted to. I've now replaced this with Albin's suggestion.

crates/extensions/tedge_mqtt_ext/src/tests.rs
jarhodes314: Now done

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: Now done

crates/extensions/tedge_mqtt_ext/Cargo.toml
jarhodes314: Now removed in favour of mqtt channel exporting a `SubscriberHandle`

crates/extensions/tedge_mqtt_ext/src/lib.rs
Bravo555: Weren't the overlapping subscriptions being discarded in `build_actor`? See the comment about the warning there.

Bravo555 (commented):
Good job on the Trie, looks solid.

crates/common/mqtt_channel/src/lib.rs
didier-wenzek: This can now be safely removed

crates/extensions/tedge_mqtt_ext/src/trie.rs
jarhodes314: > Why do we have a match for Some(("+", rest)) but none for Some(("#", rest))` on the input topic?

`Some(("#", rest))` is an invalid topic filter since it would be of the form `#/...`.

> Why do we have a check for self.sub_nodes.contains_key("#") but none on self.sub_nodes.contains_key("+") ?

For `#`, we can short circuit since any subscriber to `#` supersedes any subsequent subscriptions. For `+` however, there may be sub nodes we need to care about. That said, having looked at this further, I'm not entirely convinced this current logic is correct, and if it is, the method name is not communicating what it actually does.

jarhodes314: I've now tested this a bit further and have demonstrated with the following test that this code is definitely not working as I intended:

```rust
#[test]
fn test() {
    let mut t = MqtTrie::default();
    t.insert("a/+/c/d", 0);
    t.insert("a/b/+/e", 1);
    t.insert("a/b/c/d", 2);

    assert_eq!(
        t.remove("a/+/c/d", &0),
        SubscriptionDiff { subscribe: ["a/b/c/d".into()].into(), unsubscribe: ["a/+/c/d".into()].into()}
    );
}
```

This fails because it doesn't subscribe. The reasoning for this is that when searching for a suffix matching `c/d`, the `node.get("+")` line succeeds, so it investigates the `a/b/+/e` section of the trie and completely ignores  `a/b/c/d`. This is obviously not correct - it should return both and rely on the de-duplication to handle things correctly.

jarhodes314: I've now pushed a fix for this issue, also including `#` in the results on the short-circuiting branches. This doesn't have any impact in use of `subscribed_topics_matching` by the trie, but does make the implementation make more sense and true to its name.

crates/extensions/tedge_mqtt_ext/src/lib.rs
albinsuresh: > Weren't the overlapping subscriptions being discarded in build_actor?

Yes. The earlier log statement was just informative to the user, highlighting that some of the overlapping topics were ignored as they are already covered by broader topics. But, I agree with you that the same being emitted as a warning might unnecessarily confuse the user. So, we're better off without that.

crates/common/mqtt_channel/src/connection.rs
```diff
@@ -38,6 +41,76 @@ pub struct Connection {
...
+
+    async fn unsubscribe_many(
+        &self,
+        topics: impl IntoIterator<Item = String> + Send,
+    ) -> Result<(), MqttError> {
+        let topics = topics.into_iter().collect::<Vec<_>>();
+        {
+            let mut subs = self.subscriptions.lock().unwrap();
+            for topic in &topics {
+                subs.add(topic)?;
```
didier-wenzek: One should remove and not add unsubscribed topic to the list of subscriptions.

```suggestion
                subs.remove(topic);
```

didier-wenzek (commented):
This is a really nice PR really closed to be merged.

The trie implementation is complex - mainly because MQTT topic matching rules are complicated, but unit-tests code coverage is really great. I'm confident that this code can be maintained.

P.S:
`cargo mutant` reports a single miss in `trie.rs` (trie.rs:269:21: replace + with * in remove_unneeded_topics) which is not really a miss: the mutant is still correct but is doing more work.

crates/common/mqtt_channel/src/connection.rs
jarhodes314: Good spot, now fixed (and tested)

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: It's technically a functional change, although my thoughts on the topic here are:
a) this is not necessarily a warning the user can do anything about, it likely means the root cause of the warning is something in the thin-edge codebase
b) even if we do want to preserve the message, these lines as they were written when Didier first commented won't achieve anything. As I mentioned previously, the unsubscribe set is always empty here
c) if we're connecting to mosquitto (which we almost certainly are), this detail makes no difference. I imagine this information was potentially useful in debugging nanomq issues, but the deduplication should mean that we won't run into such issues (apart from in the edge case of simultaneous subscriptions to e.g. `a/+/c` and `a/b/+`)

didier-wenzek: >  I imagine this information was potentially useful in debugging nanomq issues

Indeed, the now removed warnings have been introduced in that context (nanomq sending message duplicates when there are overlapping subscriptions). These  warnings were mostly only useful in a debug context where the subscription of an actor was removed because of a more general subscription done by an independent actor.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -20,54 +30,121 @@ use tedge_actors::RuntimeError;
...
+}
+
+struct InputCombiner {
+    publish_receiver: mpsc::Receiver<MqttMessage>,
+    subscription_request_receiver: mpsc::Receiver<SubscriptionRequest>,
+    signal_receiver: mpsc::Receiver<RuntimeRequest>,
+}
+
+#[derive(Debug)]
+pub enum PublishOrSubscribe {
```
albinsuresh, line 62:
```suggestion
pub enum MqttRequest {
```
Just to ease future extensions where we might want to add "dynamic connection" requests (as highlighted in https://github.com/thin-edge/thin-edge.io/pull/3681#discussion_r2142332893) or "complete unsubscribe" requests etc.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -20,54 +30,121 @@ use tedge_actors::RuntimeError;
...
-    input_receiver: CombinedReceiver<MqttMessage>,
+    input_receiver: InputCombiner,
+    pub_or_sub_sender: PubOrSubSender,
     publish_sender: mpsc::Sender<MqttMessage>,
-    pub subscriber_addresses: Vec<(TopicFilter, DynSender<MqttMessage>)>,
+    subscriber_addresses: Vec<DynSender<MqttMessage>>,
     signal_sender: mpsc::Sender<RuntimeRequest>,
+    trie: TrieService,
+    current_id: usize,
+    subscription_diff: SubscriptionDiff,
```
albinsuresh, line 52: Optional: An `MqtTrie::subscriptions()` API that lets you query the trie to return the minimal `TopicFilter` that covers all the "broader topics" would have eliminated the need to maintain this `SubscriptionDiff` here. The `MqtTrie` struct could maintain this `subscription_diff` as well, keeping it updated as new topics are added/removed. And then, in the `build_actor()` function, we could just do `self.trie.trie.subscriptions()`.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -0,0 +1,1346 @@
...
+        let mut res = Vec::new();
+        for (topic, node) in &self.sub_nodes {
+            if node.has_subscribers() {
+                res.push(topic.to_owned());
+            }
+        }
+        res
+    }
+
+    fn all_sub_topics(&self) -> Vec<String> {
```
albinsuresh: It'd be good to briefly document the subtle difference between this method and `subscribed_topics`.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1343,4 +1383,63 @@ mod tests {
...
+        #[test]
+        fn tedge_agent() {
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]: [crates/extensions/tedge_mqtt_ext/src/lib.rs:172:66] topic = "te/+/+/+/+/cmd/config_update/+"
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]: [crates/extensions/tedge_mqtt_ext/src/lib.rs:172:39] self.trie.trie.insert(dbg!(topic), client_id) = SubscriptionDiff {
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]:     subscribe: {},
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]:     unsubscribe: {
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]:         "te/+/+/+/+/#",
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]:         "te/device/main///cmd/+/+",
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]:     },
+            // Jun 16 16:14:13 james-Precision-3591 tedge-agent[357845]: }
```
didier-wenzek, line 1788: Oops. What a bug!

It's really nice to see that bug fix. However, this raises a concern. Test coverage were apparently excellent, but still missing large area! How can we be more confident that there is not yet another unexplored area and latent bugs?

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: It might be nice, but this is a good way to test the logic against existing subscriptions, and also made it really easy to detect the issue affecting tedge-agent in the system tests once I added some debug logging. Since we only need to store this to manage the initial subscription list before connecting to the broker, the builder feels like the correct place for this.

crates/extensions/tedge_mqtt_ext/src/trie.rs
jarhodes314: It turns out also that the fix for this was still buggy. This was illuminated by a test where we subscribe multiple times to the same topic

```
        #[test]
        fn resubscribing_to_the_mid_level_segment_wildcard_produces_an_empty_diff() {
            let mut t = MqtTrie::default();

            assert_eq!(t.insert("a/+/c", 1), SubscribeTo("a/+/c"));
            assert_eq!(t.insert("a/+/c", 2), SubscriptionDiff::empty()); // Fails as the trie returned SubscribeTo("a/+/c")
        }
```

I believe this class of cases is now prevented by a guard where we don't add any unsubscribe cases if the list of topics to subscribe to is empty. We definitely shouldn't ever have an insertion that causes unsubscriptions and not also subscriptions because the `insert` method only unsubscribes when overriding an subscription with a larger wildcard subscription.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1473,5 +1672,19 @@ mod tests {
                 SubscriptionDiff::empty()
             );
         }
+
+        #[test]
+        fn issue_1() {
+            let mut t = MqtTrie::default();
+            t.insert("/#", 0);
+            assert!(t.insert("+/+", 1).unsubscribe.is_empty(),);
+        }
```
didier-wenzek, line 1803: This test reproduces the case `cc c2af50542642d11276b3414120dba5d319b7e79bc5c4b593b0f0e3e551bb216e # shrinks to subscriptions = ["/#", "+/+"]` of `crates/extensions/tedge_mqtt_ext/proptest-regressions/trie.txt`

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1473,5 +1672,19 @@ mod tests {
...
+
+        #[test]
+        fn issue_2() {
+            let mut t = MqtTrie::default();
+            t.insert("a/+/a", 0);
+            assert!(t.remove("a", &0).subscribe.is_empty());
+        }
```
didier-wenzek: This test reproduces the case `cc c305db36e62ac105b2011d441fca4e88224159738d637aae88679fdeb967de3d # shrinks to subscriptions = ["a/+/a"], unsubscriptions = ["a"]` of `crates/extensions/tedge_mqtt_ext/proptest-regressions/trie.txt`

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1067,6 +1087,18 @@ mod tests {
...
+            fn cumulated_diff_over_added_subscriptions_should_contains_no_unsubscriptions(
+                subscriptions in random_subscriptions(10, 5)
+            ) {
+                let mut diff = SubscriptionDiff::empty();
+                let mut t = MqtTrie::default();
+                for s in &subscriptions {
+                    diff += t.insert(s,1);
+                }
+                assert!(diff.unsubscribe.is_empty());
+            }
```
didier-wenzek: This is a check following that discussion https://github.com/thin-edge/thin-edge.io/pull/3661#discussion_r2140359677

crates/extensions/tedge_mqtt_ext/src/trie.rs
didier-wenzek: As noted by @jarhodes314 "that feels like an "undefined behaviour" for the trie, we could plausibly solve it, but I don't see there being a lot of point in doing so".

Indeed, the first version of this test was removing a subscription that has never been inserted.

=> The test has been updated and matches a genuine bug. See https://github.com/thin-edge/thin-edge.io/pull/3661/commits/5983f9d0cd72f16f929ffe3a0d5bbf3cacdf6c00

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1058,12 +1102,12 @@ mod tests {
...
                     let diff = t.insert(&s,1);
-                    for removed in diff.unsubscribe {
+                    assert!(diff.subscribe.len() <= 1, "Insert subscribed more than one topic at once: {diff:?}");
+                    for removed in &diff.unsubscribe {
                         // If one unsubscribes to some topic, then one must subscribe to a larger one
-                        assert!(diff.subscribe.iter().any(|added| {
-                            let cmp = RankTopicFilter(added).partial_cmp(&RankTopicFilter(&removed));
-                            cmp == Some(Ordering::Greater)
-                        }));
+                        assert!(diff.subscribe.iter().all(|added| {
```
didier-wenzek: I think this should be `any` not `all`.

As a the count of new subscribers cannot be greater than 1, this makes a difference only when no subscription has been added but at least one remove. Such a case should be rejected.

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1058,12 +1102,12 @@ mod tests {
                 let mut t = MqtTrie::default();
                 for s in subscriptions {
                     let diff = t.insert(&s,1);
-                    for removed in diff.unsubscribe {
+                    assert!(diff.subscribe.len() <= 1, "Insert subscribed more than one topic at once: {diff:?}");
```
didier-wenzek: I would add one more check, to make the contract crystal clear

```suggestion
                    assert!(diff.subscribe.len() <= 1, "Insert subscribed more than one topic at once: {diff:?}");
                    if let Some(added) = diff.subscribe.iter().next() {
                        assert_eq!(added, &s, "Insert({s}) subscribed to a different topic: {added}");
                    }

```

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1078,12 +1122,29 @@ mod tests {
...
                         // If one subscribes to some topic, then one must unsubscribe from a larger one
-                        assert!(diff.unsubscribe.iter().any(|removed| {
-                            let cmp = RankTopicFilter(removed).partial_cmp(&RankTopicFilter(&added));
-                            cmp == Some(Ordering::Greater)
-                        }));
+                        assert!(
+                            RankTopicFilter(&s) > RankTopicFilter(added),
+                            "Unsubscribing from {s} resubscribed to topic that isn't larger ({added}): {diff:?}");
+                    }
+                }
```
didier-wenzek: I think it would be better to keep symmetric the two properties:

- `mqttrie_extend_is_growing_when_adding_subscriptions`
- `mqttrie_extend_is_decreasing_when_removing_subscriptions`

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1095,9 +1156,10 @@ mod tests {
                 let mut diff = SubscriptionDiff::empty();
                 let mut t = MqtTrie::default();
                 for s in &subscriptions {
+                    let prev_diff = diff.clone();
                     diff += t.insert(s,1);
+                    assert!(diff.unsubscribe.is_empty(), "Inserting {s} (prev_diff={prev_diff:?}, diff={diff:?})");
                 }
-                assert!(diff.unsubscribe.is_empty());
```
didier-wenzek: So even stronger. Okay.

didier-wenzek (approved):
Now, we can be confident that the mqttrie implementation is correct.

crates/extensions/tedge_mqtt_ext/src/trie.rs
jarhodes314: Such a case should be rejected, but all that change achieves in practice is causing this test to overlap with the cumulative case, which will catch exactly the same problem (unless `AddAssign for SubscriptionDiff` is totally broken, but that should be caught by separate tests)

didier-wenzek: >  that should be caught by separate tests

Agree. My point is more to document the expectation we have. For instance just looking at the data structure of a `SubscriptionDiff`, it's not obvious that one of the two sets has to be a singleton or the empty set.

didier-wenzek: I removed the use of `any`/`all` to make the property easier to understand (leveraging the fact that the `added` topic can only be the inserted topic `s`).

Adding a subscription to some topic `T` means:
- either do nothing (because already covered)
- or subscribe to `T` and possibly unsubscribe to some topic covered by `T`.

See https://github.com/thin-edge/thin-edge.io/pull/3661/commits/1c0ca94201076d862610ca32719c3c66506f1d49

crates/extensions/tedge_mqtt_ext/src/trie.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3661/commits/1c0ca94201076d862610ca32719c3c66506f1d49

crates/extensions/tedge_mqtt_ext/src/trie.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3661/commits/1c0ca94201076d862610ca32719c3c66506f1d49

crates/extensions/tedge_mqtt_ext/src/trie.rs
```diff
@@ -1473,5 +1794,146 @@ mod tests {
                 SubscriptionDiff::empty()
             );
         }
+
+        #[test]
+        fn issue_1() {
+            let mut t = MqtTrie::default();
+            t.insert("/#", 0);
```
albinsuresh, line 1801: Never realised that the root topic also could be empty, until seeing this test. What's weird is that both `Topic::new("")` and `Topic::new("/")` are valid topics, but neither `Topic::new("/+")` nor `Topic::new("/#")` are valid. And yet the trie accepts them. Hope that doesn't break any of the assumptions.

But that makes me wonder how the MQTT actor should react to invalid topics in the `SubscriptionRequest`, since `SubscriptionDiff` also stores the subscriptions as `Hashset<String>` and not `Hashset<Topic>`.

jarhodes314: They aren't valid topics since they contain wildcards, but they are valid topic filters to subscribe to. So you can't publish a message to `/#` but you can subscribe to it and receive a message on e.g. `/a` or the empty topic name ``

albinsuresh: > but neither Topic::new("/+") nor Topic::new("/#") are valid

Aah yes, my bad. Wrongly used `Topic` instead of `TopicFilter`.

> But that makes me wonder how the MQTT actor should react to invalid topics in the `SubscriptionRequest`

I see that the trie is still silently accepting bad topic filters like `a/#/a` but internally keeping it as `a/#`. We probably need to use stronger types like `TopicFilter` at least in the `SubscriptionRequest` to prevent that.

didier-wenzek: > I see that the trie is silently accepting bad topic filters like `a/#/a` but internally keeping it as `a/#`. We probably need to use stronger types like `TopicFilter` to prevent that.

Yeah, but only at the entry point of the API i.e the 3 methods of `MqtTrie`.

See also for a similar comment: https://github.com/thin-edge/thin-edge.io/pull/3661#discussion_r2131849740

PR #3668, opened 2025-06-04
fix: wrong message being shown on `system.toml` parse error

Bravo555:
## Proposed changes

A wrong variant of ServiceCommand was used which caused an error message to wrongly report that `ServiceCommand::Enable` instead of `ServiceCommand::Start` failed when `start` command in `system.toml` did not contain a placeholder for an argument.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

Some of the service manager code could be further improved:

1. Many arguments are passed down unnecessarily just for sake of error reporting, we could instead not pass them down and instead attach the context up the call stack.
2. Could use docs comments describing why we use `system.toml` file and why handling it takes quite a bit of code

### DIFF
--- a/crates/core/tedge/src/system_services/managers/general_manager.rs
+++ b/crates/core/tedge/src/system_services/managers/general_manager.rs
@@ -193,48 +193,26 @@ impl ServiceCommand<'_> {
         service_manager: &GeneralServiceManager,
     ) -> Result<ExecCommand, SystemServiceError> {
         let config_path = service_manager.config_path.clone();
+        let config = match self {
+            Self::CheckManager => &service_manager.init_config.is_available,
+            Self::Stop(_) => &service_manager.init_config.stop,
+            Self::Restart(_) => &service_manager.init_config.restart,
+            Self::Start(_) => &service_manager.init_config.start,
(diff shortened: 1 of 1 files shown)


crates/core/tedge/src/system_services/managers/general_manager.rs
```diff
@@ -213,7 +213,7 @@ impl ServiceCommand<'_> {
             ),
             Self::Start(service) => ExecCommand::try_new_with_placeholder(
                 service_manager.init_config.start.clone(),
-                ServiceCommand::Enable(service),
+                ServiceCommand::Start(service),
```
jarhodes314: I believe the following should work since `ServiceCommand: Copy`. Assuming it does, that will do a better job of preventing these errors.

```suggestion
                self,
```

didier-wenzek: It works. Indeed less error prone.

Bravo555: Went a step further and shortened the whole thing in 23bc247.

PR #3670, opened 2025-06-04
fix: clear all retained messages on entity deregistration

albinsuresh:
## Proposed changes

Track all retained message topics for an entity and clear them on entity deregistration.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/entity_manager/server.rs
+++ b/crates/core/tedge_agent/src/entity_manager/server.rs
@@ -211,6 +211,23 @@ impl EntityStoreServer {
             }
         }
 
+        if channel.is_alarm()
+            || channel.is_command()
+            || channel.is_command_metadata()
+            || channel.is_health()
+        {
+            if !message.payload_bytes().is_empty() {
+                // Messages published to any persistent/retained channels which must be tracked
+                // so that they can be cleared when the entity is deregistered.
+                self.entity_store
+                    .track_persistent_channel(&topic_id, channel.clone());
+            } else {
+                // Untrack previously tracked channels when they are cleared from the broker.
+                self.entity_store
+                    .untrack_persistent_channel(&topic_id, &channel);
+            }
+        }
+
         if let Channel::EntityTwinData { fragment_key } = channel {
             let fragment_value = serde_json::from_slice(message.payload_bytes())?;
             let twin_message = EntityTwinMessage::new(topic_id, fragment_key, fragment_value);
@@ -307,16 +324,11 @@ impl EntityStoreServer {
     async fn deregister_entity(&mut self, topic_id: &EntityTopicId) -> Vec<EntityMetadata> {
         let deleted = self.entity_store.deregister_entity(topic_id);
         for entity in deleted.iter().rev() {
-            for twin_key in entity.twin_data.keys() {
-                let topic = self.mqtt_schema.topic_for(
-                    &entity.topic_id,
-                    &Channel::EntityTwinData {
-                        fragment_key: twin_key.to_string(),
-                    },
-                );
-                let clear_twin_msg = MqttMessage::new(&topic, "").with_retain();
+            for channel in &entity.persistent_channels {
+                let topic = self.mqtt_schema.topic_for(&entity.topic_id, channel);
+                let clear_data_msg = MqttMessage::new(&topic, "").with_retain();
 
-                self.publish_message(clear_twin_msg).await;
+                self.publish_message(clear_data_msg).await;
             }
             let topic = self
                 .mqtt_schema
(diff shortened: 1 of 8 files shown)


crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -211,6 +211,23 @@ impl EntityStoreServer {
             }
         }
 
+        if channel.is_alarm()
+            || channel.is_command()
+            || channel.is_command_metadata()
+            || channel.is_health()
```
albinsuresh, line 217: We're only clearing the retained messages on this specified list of channels where we expect retained messages, excluding measurement, event and any other random channels under the entity topic id. I'm open to clear any measurements and events that the user may have accidentally published as retained. But, for other random channels, I feel we shouldn't be touching them, as tedge doesn't recognise/own those topics.

didier-wenzek: That makes sense.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -211,6 +211,23 @@ impl EntityStoreServer {
...
+        {
+            if !message.payload_bytes().is_empty() {
+                // Messages published to any persistent/retained channels which must be tracked
+                // so that they can be cleared when the entity is deregistered.
```
didier-wenzek, line 221: Do we really need to track all these channels? Can't  we simply blindly clear all related retained topics when an entity is de-registered? There is no harm in clearing a retained message that doesn't exist.

albinsuresh: By establishing an dynamic subscription to `te/<entity-topic-id>/#` topic to identify all the retained topics and then publish the clear messages? This was my first thought to fix this issue, but did fall back to this tracking approach since MQTT actor doesn't support independent dynamic subscriptions. Not just a new subscription, but we need a new independent connection with that subscription.

But, we can bypass the MQTT actor by spawning an independent MQTT connection with this wildcard subscription specifically for this case, wait for a second or something to get all retained messages (this is the non-deterministic part that I don't like, based on the alarms experience), and then clear them. Is that what you had in mind as well?

didier-wenzek: I have in mind something different. By *blindly* clear all related retained messages, I mean publishing clear retained messages on the 4 channels an entity can have retained messages attached to - but without any attempt to subscribe to those messages.

More precisely, in the `deregister_entity` method I would replace the loop over the persisted channels for an entity (`for channel in &entity.persistent_channels { ...  }` by a generic list of channels independently of the entity under de-registration:

```rust
for channel in [ChannelFilter::Alarm, AnyCommand, CommandMetadata, Health] {
        let topic = self.mqtt_schema.topic_for(&entity.topic_id, channel);
        let clear_data_msg = MqttMessage::new(&topic, "").with_retain();

        self.publish_message(clear_data_msg).await;
}
```

albinsuresh: > I mean publishing clear retained messages on the 4 channels an entity can have retained messages attached to - but without any attempt to subscribe to those messages.

That'd work for static channels like `status/health` but won't work for dynamic channels where the data types and ids like alarm type, command type, command id etc (e.g: `a/high_temp`, `cmd/restart/123`) are present, right? That's why we need to either track those dynamic channels or subscribe on demand.

didier-wenzek: I see the problem now. You are correct, we need something dynamic, to collect all these free-form channels on the fly.

albinsuresh: So, which one would be the better approach: tracking messages as done in this PR which is deterministic, but with run-time memory consumption OR collecting all retained messages on-demand with a fresh dynamic connection/subscription without any runtime cost, but with that element of non-determinism due to the timed waiting reuiqred to receive all retained messages?

didier-wenzek: Run-time memory consumption should not be a problem (in practice, there are less than 1000 entries). For me the main criteria here is code clarity. To this regard, having a function grouping all the cleaning logic is appealing even if there is a timeout issue.

albinsuresh: Will re-implement this fix with a dynamic connection/subscription. For simplicity, I'd bypass the MQTT actor and just establish a direct connection instead.

albinsuresh:
Replaced by #3681

Issue #3664, opened 2025-06-04
Extend tedge-p11-server to support creating a Certificate Signing Request


reubenmiller:
**Is your feature request related to a problem? Please describe.**

When thin-edge.io is using a private key stored in a HSM, it uses the tedge-p11-server to interact with the private key to connect to the Cloud.

As a result, it means the `tedge cert renew c8y` and `tedge cert create-csr` commands no longer work as the commands expect to have direct access to the private key which does not exist when the tedge-p11-server (PKCS11 interface) is being used.

Today, the only way to support the cert renewal is to use tools such as [gnutls-bin certtool](https://www.gnutls.org/manual/html_node/certtool-Invocation.html) to generate the CSR, and then use the `tedge cert renew --csr-path <local.csr>`.

**Describe the solution you'd like**

Extend the tedge-p11-server to also support generating a Certificate Signing Request for keys stored in a PKCS11 compatible HSM, and integrate this support in the following commands.

```sh
tedge cert create-csr
tedge cert renew c8y
```

This would eliminate the dependency on additional tooling, and improve the integration (e.g. the tedge commands already know where the tedge-p11-server is, and what key it is using), so reduce chances of misconfiguration.

**Describe alternatives you've considered**

**Additional context**

PR #3677, opened 2025-06-11
feat: `tedge cert renew c8y` using HSM

Bravo555:
## TODO
- [x] cleanup
- [x] consider using a pubkey derived from privkey (more complexity than reusing the public key in the certificate, different per cryptosystem, need to add some dependencies - reusing cert pubkey is simpler) 

## Proposed changes

This PR introduces support for `tedge cert renew c8y` when the private key is located on the HSM.

To generate a CSR, one needs to put the public key in the CSR (it will become a part of the certificate (SPKI)) and to sign the certificate using the private key to prove the ownership of it. Because the public key can be derived from the private key, and because we need to sign anyway, until this PR we just read the private key to do both these things.

Now, with an HSM, we can't read the private key, so instead for CSR creation with HSM being enabled, we read the public key from the currently used certificate and sign using the `TedgeP11Client`.

It's also perhaps possible to derive the public key from the public attributes of the private key in the HSM (that's what [gnutls seems to be doing](https://github.com/gnutls/gnutls/blob/af6a39894e0dc8e1dd3f9690f7fc011d0ffe86b5/lib/pkcs11_privkey.c#L1434)), so that would be an alternative approach to reusing the pubkey from the certificate that needs to be considered as well.

## Types of changes

## Paste Link to the issue
#3664

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/certificate/create_csr.rs
+++ b/crates/core/tedge/src/cli/certificate/create_csr.rs
@@ -4,11 +4,15 @@ use crate::log::MaybeFancy;
 use crate::override_public_key;
 use crate::persist_new_private_key;
 use crate::reuse_private_key;
+use anyhow::Context;
 use camino::Utf8PathBuf;
+use certificate::parse_root_certificate::CryptokiConfig;
 use certificate::CsrTemplate;
 use certificate::KeyCertPair;
 use certificate::KeyKind;
 use tedge_config::TEdgeConfig;
+use tracing::debug;
+use tracing::instrument;
 
 /// Create a certificate signing request (CSR)
 pub struct CreateCsrCmd {
@@ -16,7 +20,10 @@ pub struct CreateCsrCmd {
     pub id: String,
 
     /// The path where the device private key will be stored
-    pub key_path: Utf8PathBuf,
+    pub key: Key,
+
+    /// Path to current certificate
+    pub current_cert: Option<Utf8PathBuf>,
 
     /// The path where the device CSR will be stored
     pub csr_path: Utf8PathBuf,
@@ -27,6 +34,12 @@ pub struct CreateCsrCmd {
 
     /// CSR template
     pub csr_template: CsrTemplate,
+}
+
+#[derive(Debug, Clone)]
+pub enum Key {
+    Local(Utf8PathBuf),
+    Cryptoki(CryptokiConfig),
 }
 
 #[async_trait::async_trait]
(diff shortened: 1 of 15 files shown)


reubenmiller:
@Bravo555  The feature works nicely. The only open question would be if the errors messages could be a bit more informative when the `tedge-p11-server.socket` is not reachable. I checked to see what would happen if I just disabled both the tedge-p11-server and socket and only got the following output:

```sh
# tedge cert renew c8y
Error: failed to renew the device certificate via Cumulocity HTTP proxy http://127.0.0.1:8001

Caused by:
    0: Cryptography related error
    1: Remote key error
```

The above error doesn't tell anything about the tedge-p11-server not being reachable, which socket is being used, etc, so it would definitely help people debug a bit of the error message included more context.

Bravo555:
> The above error doesn't tell anything about the tedge-p11-server not being reachable, which socket is being used, etc, so it would definitely help people debug a bit of the error message included more context

To sign the CSR using key in the HSM I had to implement a function `rcgen::RemoteKeyPair::sign` that returns `rcgen::Error`. This error type has a variant `rcgen::Error::RemoteKeyError` that's intended to be used here, but it can't contain any other data or context. So I used another variant, `rcgen::Error::PemError` that contains a string, so we can put some information into the error message, but a part of the error message is misleading (error is not PEM-related):

```
Error: failed to renew the device certificate via Cumulocity HTTP proxy http://127.0.0.1:8001

Caused by:
    0: Cryptography related error
    1: PEM error: Failed to connect to tedge-p11-server UNIX socket at '/opt/run/tedge-p11-server/pkcs11.sock'
```

I will see if I adding a string parameter, or a `dyn Error` object to this error variant can be merged upstream to `rcgen`. Alternatively for now we could use a patched version of `rcgen` and add this parameter ourselves.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -7,6 +7,7 @@ Documentation       Test thin-edge.io MQTT client authentication using a Hardwar
 
 # it would be good to explain here why we use the tedge-p11-server exclusively and not the module mode
```
didier-wenzek, line 8: It might be good to resolve this todo.

crates/core/tedge/src/cli/certificate/create_csr.rs
```diff
@@ -43,27 +55,41 @@ impl Command for CreateCsrCmd {
...
+        let previous_key = match &self.key {
+            Key::Local(key_path) => reuse_private_key(key_path)
+                .await
+                .map_err(|e| CertError::IoError(e).key_context(key_path.clone()))?,
+
+            Key::Cryptoki(cryptoki) => {
+                let current_cert = self
+                    .current_cert
+                    .clone()
+                    .expect("Need an existing cert when using an HSM");
```
didier-wenzek: A bit surprising, but okay as a temporary solution.

didier-wenzek (approved):
Approved

PR #3681, opened 2025-06-12
fix: clear all retained entity messages on entity deregistration

albinsuresh:
## Proposed changes

On entity deregistration, fetch all retained messages associated with that entity and its children and then clear them all.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/entity_manager/server.rs
+++ b/crates/core/tedge_agent/src/entity_manager/server.rs
@@ -12,13 +12,17 @@ use tedge_api::entity_store::EntityTwinMessage;
 use tedge_api::entity_store::EntityUpdateMessage;
 use tedge_api::entity_store::ListFilters;
 use tedge_api::mqtt_topics::Channel;
+use tedge_api::mqtt_topics::ChannelFilter;
+use tedge_api::mqtt_topics::EntityFilter;
 use tedge_api::mqtt_topics::EntityTopicId;
 use tedge_api::mqtt_topics::MqttSchema;
 use tedge_api::pending_entity_store::RegisteredEntityData;
 use tedge_api::EntityStore;
+use tedge_mqtt_ext::MqttConnector;
 use tedge_mqtt_ext::MqttMessage;
-use tedge_mqtt_ext::QoS;
 use tedge_mqtt_ext::TopicFilter;
+use tokio::time::timeout;
+use tokio::time::Duration;
 use tracing::error;
 
 #[derive(Debug)]
@@ -52,6 +56,7 @@ pub enum EntityStoreResponse {
 pub struct EntityStoreServer {
     entity_store: EntityStore,
     mqtt_schema: MqttSchema,
+    mqtt_connector: Box<dyn MqttConnector>,
     mqtt_publisher: LoggingSender<MqttMessage>,
     entity_auto_register: bool,
 }
@@ -60,6 +65,7 @@ impl EntityStoreServer {
     pub fn new(
         entity_store: EntityStore,
         mqtt_schema: MqttSchema,
+        mqtt_connector: Box<dyn MqttConnector>,
         mqtt_actor: &mut impl MessageSink<MqttMessage>,
         entity_auto_register: bool,
     ) -> Self {
@@ -68,6 +74,7 @@ impl EntityStoreServer {
         Self {
             entity_store,
             mqtt_schema,
+            mqtt_connector,
             mqtt_publisher,
             entity_auto_register,
         }
@@ -306,24 +313,56 @@ impl EntityStoreServer {
 
     async fn deregister_entity(&mut self, topic_id: &EntityTopicId) -> Vec<EntityMetadata> {
         let deleted = self.entity_store.deregister_entity(topic_id);
-        for entity in deleted.iter().rev() {
-            for twin_key in entity.twin_data.keys() {
-                let topic = self.mqtt_schema.topic_for(
-                    &entity.topic_id,
-                    &Channel::EntityTwinData {
-                        fragment_key: twin_key.to_string(),
-                    },
-                );
-                let clear_twin_msg = MqttMessage::new(&topic, "").with_retain();
 
-                self.publish_message(clear_twin_msg).await;
+        let mut topics = TopicFilter::empty();
+        for entity in deleted.iter() {
+            for channel_filter in [
+                ChannelFilter::MeasurementMetadata,
+                ChannelFilter::EventMetadata,
+                ChannelFilter::AlarmMetadata,
+                ChannelFilter::Alarm,
+                ChannelFilter::EntityTwinData,
+                ChannelFilter::AnyCommand,
+                ChannelFilter::AnyCommandMetadata,
+            ] {
+                let topic = self
+                    .mqtt_schema
+                    .topics(EntityFilter::Entity(&entity.topic_id), channel_filter);
+                topics.add_all(topic);
             }
+        }
+
+        // A single connection to retrieve all retained metadata messages for all deleted entities
+        match self.mqtt_connector.connect(topics).await {
+            Ok(mut connection) => {
+                while let Ok(Some(message)) =
+                    timeout(Duration::from_secs(1), connection.next_message()).await
+                {
+                    if message.retain && !message.payload_bytes().is_empty() {
+                        let clear_msg = MqttMessage::new(&message.topic, "").with_retain();
+                        if let Err(err) = self.mqtt_publisher.send(clear_msg).await {
+                            error!(
+                                    "Failed to clear retained message on topic {} while de-registering {} due to: {err}",
+                                    topic_id,
+                                    message.topic
+                                );
+                        }
+                    }
+                }
+
+                connection.disconnect().await;
+            }
+            Err(err) => {
+                error!("Failed to create MQTT connection for clearing entity data: {err}");
+            }
+        }
+
+        // Clear the entity metadata of all deleted entities bottom up
+        for entity in deleted.iter().rev() {
             let topic = self
                 .mqtt_schema
                 .topic_for(&entity.topic_id, &Channel::EntityMetadata);
-            let clear_entity_msg = MqttMessage::new(&topic, "")
-                .with_retain()
-                .with_qos(QoS::AtLeastOnce);
+            let clear_entity_msg = MqttMessage::new(&topic, "").with_retain();
 
             self.publish_message(clear_entity_msg).await;
         }
--- a/crates/core/tedge_agent/src/entity_manager/tests.rs
+++ b/crates/core/tedge_agent/src/entity_manager/tests.rs
@@ -6,19 +6,28 @@ use crate::entity_manager::tests::model::Command;
 use crate::entity_manager::tests::model::Commands;
 use crate::entity_manager::tests::model::Protocol::HTTP;
 use crate::entity_manager::tests::model::Protocol::MQTT;
+use async_trait::async_trait;
+use futures::channel::mpsc::UnboundedReceiver;
+use futures::channel::mpsc::UnboundedSender;
+use futures::StreamExt;
 use proptest::proptest;
 use serde_json::json;
 use std::collections::HashSet;
 use tedge_actors::test_helpers::MessageReceiverExt;
+use tedge_actors::Sender;
 use tedge_actors::Server;
 use tedge_api::entity::EntityMetadata;
 use tedge_api::entity::EntityType;
 use tedge_mqtt_ext::test_helpers::assert_received_contains_str;
+use tedge_mqtt_ext::MqttConnection;
+use tedge_mqtt_ext::MqttConnector;
+use tedge_mqtt_ext::MqttError;
 use tedge_mqtt_ext::MqttMessage;
+use tedge_mqtt_ext::TopicFilter;
 
 #[tokio::test]
 async fn new_entity_store() {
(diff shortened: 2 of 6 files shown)


crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -307,23 +315,39 @@ impl EntityStoreServer {
...
+                .mqtt_config
+                .clone()
+                .with_no_session()
+                .with_subscriptions(self.mqtt_schema.topics(
+                    EntityFilter::Entity(&entity.topic_id),
+                    ChannelFilter::AnyData,
+                ));
+
+            // Clear all retained metadata/data associated with the entity
+            match mqtt_channel::Connection::new(&mqtt_config).await {
```
albinsuresh: Ideally the connection shouldn't be established directly from here, but the mqtt actor should be enhanced to support new dynamic **connection** requests as well. But taking this simpler approach for now to avoid any conflict with #3661 by adding a new MQTT input variant (`MqttInput::Connect`) as that PR is already adding support for dynamic **subscription** requests (`MqttInput::SubscribeRequest`) in the mqtt actor.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -396,6 +397,7 @@ mod entity {
         let server = EntityStoreServer::new(
             entity_store,
             mqtt_schema,
+            MqttConfig::default(),
```
albinsuresh: To avoid any dependency on test-mqtt-server from this unit test code, a different abstraction is needed here so that dummy messages can be passed to the test code without using a broker..

crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -180,6 +180,7 @@ impl MqttSchema {
             ChannelFilter::AnyCommandMetadata => "/cmd/+".to_string(),
             ChannelFilter::CommandMetadata(operation) => format!("/cmd/{operation}"),
             ChannelFilter::Health => "/status/health".to_string(),
+            ChannelFilter::AnyData => "/+/#".to_string(),
```
albinsuresh: Instead of this broad topic filter that covers "everything", we can selectively subscribe to the known channels like [`/a/+/meta`, `/cmd/+`, `/cmd/+/+`, ...] as well.

crates/core/tedge_agent/Cargo.toml
```diff
@@ -24,6 +24,7 @@ http-body = { workspace = true }
 http-body-util = { workspace = true }
 hyper = { workspace = true, features = ["full"] }
 log = { workspace = true }
+mqtt_channel = { workspace = true }
```
albinsuresh: This won't be needed once MQTT actor is capable of establishing dynamic connections as well.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: Introduced an additional level of abstraction for dynamic connections in [8cd679f](https://github.com/thin-edge/thin-edge.io/pull/3681/commits/8cd679f654f36029c69f42e2990911c3af3416a0) to make this code unit-testable.

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: Done in [8cd679f](https://github.com/thin-edge/thin-edge.io/pull/3681/commits/8cd679f654f36029c69f42e2990911c3af3416a0)

crates/core/tedge_agent/Cargo.toml
albinsuresh: Resolved by bf84271

crates/extensions/tedge_mqtt_ext/Cargo.toml
```diff
@@ -20,6 +20,7 @@ async-trait = { workspace = true }
 mqtt_channel = { workspace = true }
 serde_json = { workspace = true }
 tedge_actors = { workspace = true }
+tedge_api = { workspace = true }
```
didier-wenzek: This dependency raises concerns.

`tedge` is more specific than `mqtt`. Hence, the latter should not depend on the former.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -223,3 +230,59 @@ impl Actor for MqttActor {
         .await
     }
 }
+
+#[async_trait]
+pub trait MqttConnector: Send {
```
didier-wenzek, line 229: These concepts of `MqttConnector` and `MqttConnectionFactory` have to be moved to `tedge_agent`.

This will make the dependency graph clearer.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -223,3 +230,59 @@ impl Actor for MqttActor {
...
+
+    fn get_message_receiver(
+        &mut self,
+        entity: &EntityTopicId,
+    ) -> Option<&mut UnboundedReceiver<MqttMessage>>;
+
+    async fn disconnect(&mut self, entity: &EntityTopicId);
+}
+
+pub struct MqttConnectionFactory {
```
didier-wenzek: Minor - This name sounds a bit java-ish ;-)

didier-wenzek (changes requested):

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: This sounds as a bit too broad. Even capturing MEAs.

crates/extensions/tedge_mqtt_ext/src/lib.rs
albinsuresh: As discussed offline, will retain these abstractions here, but without the dependency on `tedge_api` abstractions like `EntityTopicId`.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -52,6 +55,7 @@ pub enum EntityStoreResponse {
 pub struct EntityStoreServer {
     entity_store: EntityStore,
     mqtt_schema: MqttSchema,
+    mqtt_connector: Box<dyn MqttConnector>,
```
didier-wenzek, line 59: Why do we a dynamic object here. I see only one implementation.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -307,23 +313,40 @@ impl EntityStoreServer {
...
-                    &entity.topic_id,
-                    &Channel::EntityTwinData {
-                        fragment_key: twin_key.to_string(),
-                    },
-                );
-                let clear_twin_msg = MqttMessage::new(&topic, "").with_retain();
-
-                self.publish_message(clear_twin_msg).await;
+            // Clear all retained metadata/data associated with the entity
+            match self.mqtt_connector.connection_for(&entity.topic_id).await {
```
didier-wenzek: I would consider to spawn the whole loop in a background task (including open & close of the connection).

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -307,23 +313,40 @@ impl EntityStoreServer {
...
+                                    "Failed to clear retained message on topic {} while de-registering {} due to: {err}",
+                                    topic_id,
+                                    message.topic
+                                );
+                                }
+                            }
+                        }
+                    }
+
+                    self.mqtt_connector.disconnect(&entity.topic_id).await;
```
didier-wenzek: Why does `disconnect` need to known the topic id (or filter). Is the MQTT connection shared among all the virtual connection for a topic id? Wouldn't be simpler to have a fresh MQTT connection each time an entity is deregistered (something that really only rarely happens - making no point is caching/sharing resources)?

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -307,23 +313,40 @@ impl EntityStoreServer {
...
-                    },
-                );
-                let clear_twin_msg = MqttMessage::new(&topic, "").with_retain();
-
-                self.publish_message(clear_twin_msg).await;
+            // Clear all retained metadata/data associated with the entity
+            match self.mqtt_connector.connection_for(&entity.topic_id).await {
+                Ok(_) => {
+                    if let Some(messages) =
+                        self.mqtt_connector.get_message_receiver(&entity.topic_id)
```
didier-wenzek: The `MqttConnector` API seems a bit complicated for nothing: a regular MQTT connection is created, cached, retrieved by name, used and removed locally. The same can be done without caching nor naming.

didier-wenzek (commented):
I would consider to fully remove the concept of `MqttConnector` API and to directly use regular mqtt_channel connections.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: It was introduced only to make the deregister code easily unit-testable by using a mocked `MqttConnector` in tests. The intention was to avoid spawning even the test broker, which we have been trying to move away from.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: It was done directly using the `mqtt_channel` in the first commit: 188b87a. But then I was forced to spawn a test broker so that this deregister code could be unit tested. Since the test broker requires serial execution of tests, and since the `proptests` in `entity_manager::tests` are also using the deregister API, it would have significantly affected out test runtime. So, I was trying to create an abstraction to mock the connection process without a  broker, by just passing test channels from the test code (something like the `SimpleMessageBox` without the overhead of connecting it like an actor).

The `MqttConnector::get_message_receiver` abstraction was my attempt to do that, so that test channels could be used from the unit tests by using the `TestMqttConnector`. Since this API was only returning the receiver half of the channel and not the connection itself, I was forced to add the `disconnect` API as well so that the underlying connection is also closed once the receiver is exhausted.

didier-wenzek: Testability is indeed a key point. However, the proposed API seems still over complicated. Why a cache per topic id while entities are one after the other? Why `connection_for` returns nothing and forces a subsequent `get_message_receiver` ?

albinsuresh: To simplify the contract, especially with the disjoint `get_message_receiver` and `disconnect` and caching of connections, I'm considering the following trait definitions:

```
#[async_trait]
pub trait MqttConnector: Send {
    async fn connect(&mut self, topics: TopicFilter) -> Result<Box<dyn MqttConnection>, MqttError>;
}

#[async_trait]
pub trait MqttConnection: Send {
    fn get_message_receiver(&mut self) -> &mut UnboundedReceiver<MqttMessage>;

    async fn disconnect(self: Box<Self>);
}
```

How about that?

albinsuresh: Here is the implementation of the new trait definition: 67a05b3 which better represents the real `mqtt_channel::Connection` as well.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: Resolved by 67a05b3

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: Does that really help? Because, even if we spawn it as a different task, we'll still have to wait for that task to finish here, as we can't return from this `deregister` call before everything is cleared.

didier-wenzek: >  as we can't return from this `deregister` call before everything is cleared.

Why not? This was my proposal. I see clearing all these retained messages as a follow-up task of entity de-registration. But, I agree that this can be discussed.

crates/core/tedge_agent/src/entity_manager/server.rs
didier-wenzek: By far simpler.

I would consider one step further, but the current proposal is okay.

```rust
#[async_trait]
pub trait MqttConnection: Send {
    async fn next_message(&mut self) -> Option<MqttMessage>;

    async fn disconnect(self: Box<Self>);
}
```

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: > I see clearing all these retained messages as a follow-up task of entity de-registration. But, I agree that this can be discussed.

Since one of the main highlights of the HTTP API was the feedback after something is complete, returning while the operation is still in-progress in the background feels a bit weird. As a user, I personally prefer the deterministic behaviour, esp since this is a rare operation and hence I'm okay with waiting a bit more until everything is fully complete. But I'll let @reubenmiller also pitch in, if the proposed eventual cleanup is sufficient.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: Resolved by 5cec94fa2

crates/core/tedge_api/src/mqtt_topics.rs
albinsuresh: Resolved by 67a05b3c

Issue #3646, opened 2025-05-27
Follow-up tasks for tedge diagnostic feature


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
Tracker for the follow-up tasks for the `tedge diag collect` feature, which initial PR is #3608.

- [x] Add the summary output to the collection; point 6 in https://github.com/thin-edge/thin-edge.io/pull/3608#issuecomment-2877225225 => PR #3687 
- [x] Security enhancement; https://github.com/thin-edge/thin-edge.io/pull/3608#discussion_r2102563802 => PR #3698
- [x] Document user guide => PR #3663 
- [x] `Spinner` should output to `stderr`; point 7 in https://github.com/thin-edge/thin-edge.io/pull/3608#pullrequestreview-2868755733 => PR #3658

---
From https://github.com/thin-edge/thin-edge.io/issues/3646#issuecomment-2930392519
- [x] Plugin improvements (1.-4.) => PR #3682 
- [x] Rename an option from `--tarball-name` to `--name` => PR #3662 

---
- [x] Output the tarball file path to stdout to enable the following usecase => PR #3696
```shell
OUTPUT_PATH=$(tedge diag collect --name some_file_name)
tedge upload c8y --file "$OUTPUT_PATH"
``` 

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

albinsuresh:
A few other issues/improvements that I noticed, which you may consider in your follow-up tasks:

1. Why not collect the `tedge.toml` file as it is one of the most important things to look at while debugging a customer env? Are we skipping this file because there could be some sensitive information in it? Since passwords are not part of this file, I don't know what else could be problematic.
2. The tedge plugin does not collect the logs of the `collectd` mapper.
3. While collecting the MQTT messages, there's a 5s timeout for collecting all messages but only 1s for the retained messages. Was that lower value intentional? It might be safer to have a higher value.
4. Why is the mosquitto plugin collecting the mosquitto log both from the journal as well as the file system, when they are essentially the same? I understand that not all devices would have `journalctl` available, but since the plugin already checks for the existence of that command before fetching the logs, I feel we only need to fall back to the `mosquitto.log` file when this command is not available.
5. Optional suggestion: The `--tarball-name` option seems a bit too long. How about just `--name` as it is the name used for the diagnostic directory as well and not just the tar file?

didier-wenzek:
[...]

PR #3682, opened 2025-06-12
chore: improve diag-plugins and add system tests

rina23q:
## Proposed changes

According to the review 1.-4. points from https://github.com/thin-edge/thin-edge.io/issues/3646#issuecomment-2930392519

* `01_tedge.sh` collects tedge.toml and tedge config list output (->1.)
* `01_tedge.sh` collects also collectd mapper log (->2.)
* `07_mosquitto.sh` collects either journalctl output and mosquitto.log (->4.)

In addition, add system tests for all predefined diag plugins.

## Types of changes

## Paste Link to the issue
#3646 

## Checklist

### DIFF
--- a/configuration/contrib/diag-plugins/01_tedge.sh
+++ b/configuration/contrib/diag-plugins/01_tedge.sh
@@ -2,6 +2,7 @@
 set -e
 
 OUTPUT_DIR=""
+TEDGE_CONFIG_DIR=${TEDGE_CONFIG_DIR:-/etc/tedge}
 COMMAND=""
 
 # Parse arguments
@@ -9,6 +10,10 @@ while [ $# -gt 0 ]; do
     case "$1" in
         --output-dir)
             OUTPUT_DIR="$2"
+            shift 2
+            ;;
+        --config-dir)
+            TEDGE_CONFIG_DIR="$2"
             shift 2
             ;;
         collect)
@@ -46,6 +51,17 @@ collect() {
             collect_logs "tedge-mapper-${cloud}"
         fi
     done
+
+    # Collectd mapper log
+    if systemctl list-unit-files tedge-mapper-collectd.service  >/dev/null 2>&1; then
+        collect_logs tedge-mapper-collectd
+    fi
+
+    # Copy tedge.toml
+    cp "$TEDGE_CONFIG_DIR"/tedge.toml "$OUTPUT_DIR"/tedge.toml
+
+    # tedge config list
+    tedge config list --all > "$OUTPUT_DIR"/tedge-config-list.log
 }
 
 # Execute the specified command
--- a/configuration/contrib/diag-plugins/07_mosquitto.sh
+++ b/configuration/contrib/diag-plugins/07_mosquitto.sh
@@ -28,9 +28,7 @@ if [ -n "$OUTPUT_DIR" ] && [ ! -d "$OUTPUT_DIR" ]; then
 fi
 
 mosquitto_journal() {
-    if command -V journalctl >/dev/null 2>&1; then
-        journalctl -u "mosquitto" -n 1000 --no-pager > "$OUTPUT_DIR/mosquitto-journal.log" 2>&1 ||:
-    fi
+    journalctl -u "mosquitto" -n 1000 --no-pager > "$OUTPUT_DIR/mosquitto-journal.log" 2>&1 ||:
 }
 
 mosquitto_log() {
@@ -42,17 +40,16 @@ mosquitto_log() {
 }
 
 collect() {
-    if [ "$(tedge config get mqtt.bridge.built_in)" = "false" ]; then
-        if command -V mosquitto > /dev/null 2>&1; then
+    if command -V mosquitto > /dev/null 2>&1; then
+        if command -V journalctl >/dev/null 2>&1; then
             mosquitto_journal
-            mosquitto_log
-        else
-            echo "mosquitto not found" >&2
         fi
+        mosquitto_log
     else
-      # built-in bridge is used, hence this plugin should be skipped
-      exit 2
-    fi 
+        echo "mosquitto not found" >&2
+        # this plugin is not applicable when mosquitto doesn't exist
+        exit 2
+    fi
 }
 
 case "$COMMAND" in
--- a/tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
+++ b/tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
@@ -1,0 +1,88 @@
+*** Settings ***
+Library             String
+Library             ThinEdgeIO
+
+Suite Setup         Custom Suite Setup
+Suite Teardown      Get Suite Logs
+
+Test Tags           theme:troubleshooting    theme:cli    theme:plugins
+
+
+*** Test Cases ***
+01_tedge
+    ${log_names}=    Create List
+    ...    output.log
+    ...    tedge-agent.log
+    ...    tedge-mapper-c8y.log
+    ...    tedge-mapper-collectd.log
+    ...    tedge-config-list.log
+    ...    tedge.toml
+    FOR    ${log_name}    IN    @{log_names}
+        File Size Is Not Zero    ${log_name}
+    END
+
+    Log Should Contain    tedge-agent.log    Starting tedge-agent.service
+    Log Should Contain    tedge-mapper-c8y.log    Starting tedge-mapper-c8y.service
+    Log Should Contain    tedge-mapper-collectd.log    Starting tedge-mapper-collectd.service
+    Log Should Contain    tedge-config-list.log    c8y.url
+    Execute Command    diff /etc/tedge/tedge.toml /results/test/01_tedge/tedge.toml
+
+02_os
+    File Size Is Not Zero    output.log
+    Log Should Contain    output.log    Linux
+
+03_mqtt
+    ${log_names}=    Create List
+    ...    output.log
+    ...    tedge-mqtt-sub.log
+    ...    tedge-mqtt-sub-retained-only.log
+    FOR    ${log_name}    IN    @{log_names}
+        File Size Is Not Zero    ${log_name}
+    END
+    Log Should Contain    tedge-mqtt-sub.log    [te/device/main///cmd/restart] {}
+    Log Should Contain    tedge-mqtt-sub-retained-only.log    [te/device/main///twin/type] "thin-edge.io"
+
+04_workflow
+    File Size Is Not Zero    output.log
+    Execute Command    diff -qr /var/log/tedge/agent/workflow-* /results/test/04_workflow/workflow-*
+
+05_entities
+    File Size Is Not Zero    output.log
+    Log Should Contain    output.log    "@topic-id": "device/main//"
+
+06_internal
+    ${log_names}=    Create List
+    ...    output.log
+    ...    entity_store.jsonl
+    FOR    ${log_name}    IN    @{log_names}
+        File Size Is Not Zero    ${log_name}
+    END
+    Execute Command    diff /etc/tedge/.agent/entity_store.jsonl /results/test/06_internal/entity_store.jsonl
+
+07_mosquitto
+    ${log_names}=    Create List
+    ...    output.log
+    ...    mosquitto.log
+    ...    mosquitto-journal.log
+    FOR    ${log_name}    IN    @{log_names}
+        File Size Is Not Zero    ${log_name}
+    END
+    Execute Command    diff /var/log/mosquitto/mosquitto.log /results/test/07_mosquitto/mosquitto.log
+    Log Should Contain    mosquitto-journal.log    Starting mosquitto.service
+
+
+*** Keywords ***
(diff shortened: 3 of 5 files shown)


configuration/contrib/diag-plugins/07_mosquitto.sh
```diff
@@ -28,9 +28,7 @@ if [ -n "$OUTPUT_DIR" ] && [ ! -d "$OUTPUT_DIR" ]; then
 fi
 
 mosquitto_journal() {
-    if command -V journalctl >/dev/null 2>&1; then
-        journalctl -u "mosquitto" -n 1000 --no-pager > "$OUTPUT_DIR/mosquitto-journal.log" 2>&1 ||:
-    fi
+    journalctl -u "mosquitto" -n 1000 --no-pager > "$OUTPUT_DIR/mosquitto-journal.log" 2>&1 ||:
```
albinsuresh, line 31: Why limit the lines to 1000? Why not the whole log? Since it is only collected locally and not uploaded to the cloud or anything, the size doesn't really matter, right? If we're really keen to  limit the size, I'd at least use the `-b` option so that everything from the last restart is available.

rina23q: No specific reason behind. I can remove the size limitation

rina23q: Removed `-n 1000` in 09d5ee5394c2638dc2cc048a1d70d454261e2133

albinsuresh (commented):
The changes look fine.

But, we'll have to extend the `tedge_diag_collect` test to do deeper validations beyond the presence of the `output.log` files for each plugin, bbut also validate what's collected by the factory plugins as well, so that changes like these can be easily validated. We probably don't need to validate each and every line of the collected logs, but some basic validation like the presence of a non-empty file with the expected starting line or something like that would suffice. For static files like `tedge.toml`, they can be compared for equality as well.

configuration/contrib/diag-plugins/07_mosquitto.sh
reubenmiller: Whilst the file is produced locally, the file is only really usable if it is transferred away from the device...so having a file size limit is generally not the worst idea...however I guess since a logrotate definition might be included within the mosquitto package, then you can rely on the logrotate to limit the file size in this instance.

reubenmiller: Sorry my statement was incorrect. You're using a journalctl command so the output MUST be limited, otherwise this could provide logs from a very long time back, and using the `-b` is not a good fit here because you may want to have log messages from a previous boot (if there is infact logs)...so limiting to collecting recent log messages is a good thing here.

configuration/contrib/diag-plugins/01_tedge.sh
```diff
@@ -46,6 +51,17 @@ collect() {
             collect_logs "tedge-mapper-${cloud}"
         fi
     done
+
+    # Collectd mapper log
+    if command -V collectd >/dev/null 2>&1; then
```
rina23q: Not sure if this condition is the best. Alternatively, the condition can be as below. (checking if `tedge-collectd-mapper` is active)

```sh
if command -v systemctl >/dev/null 2>&1; then
    if systemctl is-active --quiet tedge-collectd-mapper.service; then
        collect_logs tedge-mapper-collectd
    fi
fi
```

configuration/contrib/diag-plugins/01_tedge.sh
```diff
@@ -2,6 +2,7 @@
 set -e
 
 OUTPUT_DIR=""
+CONFIG_DIR=${TEDGE_CONFIG_DIR:-/etc/tedge}
```
reubenmiller: This might be easier, and it avoids having to remember what "CONFIG_DIR" actually means (e.g. plugin config dir, thin-edge.io config dir)?

```suggestion
TEDGE_CONFIG_DIR=${TEDGE_CONFIG_DIR:-/etc/tedge}
```

configuration/contrib/diag-plugins/01_tedge.sh
reubenmiller: You can simplify the logic and just use the `systemctl list-unit-files` to check both if the systemctl command exists, and the service exists

```sh
if systemctl list-unit-files tedge-collectd-mapper.service  >/dev/null 2>&1; then
    collect_logs tedge-mapper-collectd
fi
```

configuration/contrib/diag-plugins/07_mosquitto.sh
```diff
@@ -44,8 +42,11 @@ mosquitto_log() {
 collect() {
     if [ "$(tedge config get mqtt.bridge.built_in)" = "false" ]; then
         if command -V mosquitto > /dev/null 2>&1; then
-            mosquitto_journal
-            mosquitto_log
+            if command -V journalctl >/dev/null 2>&1; then
+                mosquitto_journal
+            else
+                mosquitto_log
+            fi
```
reubenmiller: You should still collect mosquitto logs even if using the built-in bridge, as mosquitto is still used as the local message bus.

rina23q: Ah ok. It means I explained wrongly from the day 1, "this plugin is not applicable when build-in used" :D

configuration/contrib/diag-plugins/01_tedge.sh
rina23q: Applied in 2cd9a2df854007380a5a9d93cc0f75e7ece1b9fc (apologize for the wrong service name in my comment, `tedge-mapper-collectd` is the right one)

configuration/contrib/diag-plugins/01_tedge.sh
rina23q: Done in [32454e0](https://github.com/thin-edge/thin-edge.io/pull/3682/commits/32454e009d094a2bc9f81a7e9e8ed507089a8b89)

configuration/contrib/diag-plugins/07_mosquitto.sh
rina23q: Fixed in [32db02b](https://github.com/thin-edge/thin-edge.io/pull/3682/commits/32db02b36d56502c6a9fc0c60e81d817b7e3e466)

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
```diff
@@ -0,0 +1,71 @@
...
+01_tedge
+    ${log_names}=    Create List
+    ...    output.log
+    ...    tedge-agent.log
+    ...    tedge-mapper-c8y.log
+    ...    tedge-mapper-collectd.log
+    ...    tedge-config-list.log
+    ...    tedge.toml
+    FOR    ${log_name}    IN    @{log_names}
+        File Size Is Not Zero    ${log_name}
```
albinsuresh, line 21: Since these are the dedicated tests for each plugin, I'd go one step further and assert at least some content: like the first line or some known/guaranteed content, just to be sure that what's captured is not some junk.

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
```diff
@@ -0,0 +1,71 @@
...
+    END
+
+04_workflow
+    File Size Is Not Zero    output.log
+    Execute Command    diff -qr /var/log/tedge/agent/workflow-* /results/test/04_workflow/workflow-*
+
+05_entities
+    File Size Is Not Zero    output.log
```
albinsuresh, line 50: We can assert if the content is valid json and contains the entry for the main device, which is guaranteed.

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
```diff
@@ -0,0 +1,71 @@
...
+02_os
+    File Size Is Not Zero    output.log
+
+03_mqtt
+    ${log_names}=    Create List
+    ...    output.log
+    ...    tedge-mqtt-sub.log
+    ...    tedge-mqtt-sub-retained-only.log
+    FOR    ${log_name}    IN    @{log_names}
+        File Size Is Not Zero    ${log_name}
```
albinsuresh, line 40: We could assert the presence of some known message like `tedge-agent` registration message, or something similar.

albinsuresh (commented):
Happy to approve with some test coverage improvements.

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
```diff
@@ -0,0 +1,71 @@
...
+
+*** Keywords ***
+File Size Is Not Zero
+    [Arguments]    ${filename}
+    Execute Command    test -s /results/test/${TEST NAME}/${filename}
+
+Custom Suite Setup
+    Setup
+    Execute Command    mkdir -p /results
+    Start Service    tedge-mapper-collectd
```
jarhodes314, line 87: Is there possibly a race here? Are the tests assuming the diagnostic collection happens after `tedge-mapper-collectd` publishes something?

rina23q: Nope, `tedge-mapper-collectd` doesn't have to publish something. As long as some content is logged in `journalctl`, it's fine. So even this line, indicating the service is starting, is enough.
```
Jun 16 16:01:26 c2e84027fd54 systemd[1]: Starting tedge-mapper-collectd.service - tedge-mapper-collectd converts Thin Edge JSON measurements to Cumulocity JSON format....
```

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
rina23q: All your suggestions are applied in e2af4a12ef684b014577d417eb64c3281be914c6

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
```diff
@@ -20,10 +20,16 @@ Test Tags           theme:troubleshooting    theme:cli    theme:plugins
     FOR    ${log_name}    IN    @{log_names}
         File Size Is Not Zero    ${log_name}
     END
+
+    Log Should Contain    tedge-agent.log    Starting tedge-agent.service
```
albinsuresh, line 24: The earlier `File Size Is Not Zero` check is redundant now, as we're checking the content itself here.

configuration/contrib/diag-plugins/07_mosquitto.sh
```diff
@@ -42,17 +40,17 @@ mosquitto_log() {
...
-    if [ "$(tedge config get mqtt.bridge.built_in)" = "false" ]; then
-        if command -V mosquitto > /dev/null 2>&1; then
+    if command -V mosquitto > /dev/null 2>&1; then
+        if command -V journalctl >/dev/null 2>&1; then
             mosquitto_journal
-            mosquitto_log
         else
-            echo "mosquitto not found" >&2
+            mosquitto_log
         fi
```
reubenmiller: We'll need to change this logic to check both locations, journald logs and the default location of the mosquitto file log, as you can configure mosquitto to log to file instead of syslog (journald) even if you have journalctl installed.

reubenmiller: So just move the "mosquitto_log" out of the journalctl if/else condition.

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
reubenmiller: Though one advantage to checking the file size is zero is that it would produce a more meaningful error if the file is empty (e.g. "File is empty" rather than "File does not contain "xxxx")

configuration/contrib/diag-plugins/07_mosquitto.sh
reubenmiller: The log limit has been added in again.

configuration/contrib/diag-plugins/07_mosquitto.sh
rina23q: Done in 5a88ac55b943e2b5aa94aa4c3c08e6cd268cd40a

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
rina23q: > Though one advantage to checking the file size is zero is that it would produce a more meaningful error if the file is empty (e.g. "File is empty" rather than "File does not contain "xxxx")

Exactly this is the reason why I kept it. Also, it just calls `test -s xxx`, so it should be a light job and finish quickly.

reubenmiller (approved):
Approved

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
albinsuresh: In that case I'd have done the `File Size Is Not Zero` from the `Log Should Contain` function itself. But anyway, minor thing.

Issue #3684, opened 2025-06-13
Change order in which C8y_DeviceProfile is mapped to operations


evanorsouw:
**Is your feature improvement request related to a problem? Please describe.**
The current implementation of applying a C8y_DeviceProfile to a set of operations uses the fixed order; 

1. firmware
2. software
3. configuration

The concrete use-cases I have are;
1. installing a container-group (software) pulling from a registry that needs a certificate (configuration).
==> **fails because certificate not yet installed at time of pulling**.
2. install a generic software package that needs a valid configuration to work.
==> **installing package succeeds, running it fails due to missing configuration, restart required (not always obvious to user)**. 

In general it seems that software always depends on configuration and never the other way around.

**Describe the solution you'd like**
Change the mapping order to;
1. firmware
2. configuration
3. software

**Describe alternatives you've considered**

The alternatives considered;
- adding workflow scripts to reshuffle the operations order
- wrap otherwise off-the-shelf packages as new packages that include application configuration
- manually installing configuration upfront outside the device-profile
- change applications to deal with temporarily missing configurations

At best these all work, but they typically complicate the system as a whole.

**Additional context**
The change (swap software and configuration in this function [convert.rs](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/c8y_mapper_ext/src/operations/convert.rs#L335)) is trivial.
I realize that this breaks behavior for some users and cannot oversee the consequences.
So an option to trigger this new behavior may be desirable.

reubenmiller:
After discussing the proposal with the team, we've come to the conclusion that changing default device profile processing order to `firmware -> configuration -> software` will offer more out-of-the-box benefits to users as a whole.

We've just pushed a draft PR which also explains the advantages etc, but we should be able to merge sometime next week as it is not a big change, and whilst it is technically a breaking change, I believe most users are more suffering from this, so we don't believe it to negatively affect existing users to warrant making it configurable (outside of overriding the device_profile workflow).

https://github.com/thin-edge/thin-edge.io/pull/3685

reubenmiller:
@evanorsouw The change has now been merged into main, so you can try it out already using the main channel which is installed via the following command:

```sh
wget -O - thin-edge.io/install.sh | sh -s -- --channel main
```

The install script is just configuring a different linux repository called [tedge-main](https://cloudsmith.io/~thinedge/repos/tedge-main/packages/) instead of "tedge-release" on cloudsmith.io.

[...]

PR #3685, opened 2025-06-13
feat: process configuration before software in device profile operations

reubenmiller:
## Proposed changes

Change the default processing order of artifacts included in a device profile operation, where configuration is processed before software. The new order supports more use-cases out-of-the-box.

Technically this is a breaking change as it changed the behaviour, however we believe it will only positively impact users and remove common annoyances with the existing processing order.

**Old Processing Order**

* Firmware
* Software
* Configuration

**New Processing Order**

* Firmware
* Configuration
* Software

Below lists the advantages and disadvantages of the change:

**Advantages**
* Repositories can be configured (with credentials or proxy settings) before applications/containers are installed via software
* Configuration can be applied before the application/container starts the first time (also removes need for restarting the software)

**Disadvantages**
* Short period of time where potentially incompatible configuration is applied before the new application version has been installed
* Potential that application could overwrite config (generally only occurs if app builders haven't defined the linux package definition correctly)

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3684

## Checklist

### DIFF
--- a/docs/src/references/agent/device-profiles.md
+++ b/docs/src/references/agent/device-profiles.md
@@ -386,7 +386,7 @@ Each artifact type is mapped to the corresponding operation type in thin-edge
 
 Since the thin-edge payload is an ordered list of operations offering flexibility in defining them in any order,
 the C8y payload is mapped to the equivalent tedge JSON format by applying an implicit order between the artifact types,
-starting with the `firmware_update` operation followed by `software_update` and then `config_update`.
+starting with the `firmware_update` operation followed by `config_update` and then `software_update`.
 Since both `software` and `configuration` values are arrays with a defined order, it is maintained during the mapping as well.
 
 The above payload, meant for the main device, is mapped to thin-edge JSON format as follows:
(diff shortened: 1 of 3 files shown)


docs/src/references/agent/device-profiles.md
```diff
@@ -386,7 +386,7 @@ Each artifact type is mapped to the corresponding operation type in thin-edge
 
 Since the thin-edge payload is an ordered list of operations offering flexibility in defining them in any order,
 the C8y payload is mapped to the equivalent tedge JSON format by applying an implicit order between the artifact types,
-starting with the `firmware_update` operation followed by `software_update` and then `config_update`.
+starting with the `firmware_update` operation followed by `config_update` and then `software_update`.
```
albinsuresh, line 389: Shouldn't we add a note or warn here which highlights the exact same `Disadvantages` highlighted in the code comments, esp since it is a significant behavioural change from the previous versions? I'm even wondering if we should highlight that the behaviour was different in earlier versions.

reubenmiller: I don't think the change in behaviour is worth noting in the docs as it will just add legacy knowledge into the docs, and since the previous behaviour was generally problematic is most situations, I don't think it will break any users existing setups.

However, I've added a note about how users can managed breaking configuration changes to applications which are already installed. 315bf8ef3b505daf4def799c8b9b222c8e3a063c

reubenmiller: After an offline discussion with @albinsuresh, it was decided that the note invokes more questions rather than answers.

Instead, we'll rely on user feedback to extend the docs in the future.

albinsuresh (approved):
LGTM.

PR #3687, opened 2025-06-13
feat: tedge diag adds summary output to the archive

rina23q:
## Proposed changes

The copy of the console output by `tedge diag collect` is now added as `summary.log` to a tarball archive.

## Types of changes

## Paste Link to the issue
#3646

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/diag/collect.rs
+++ b/crates/core/tedge/src/cli/diag/collect.rs
@@ -1,9 +1,7 @@
+use crate::cli::diag::logger::DualLogger;
 use crate::command::Command;
-use crate::error;
-use crate::info;
 use crate::log::MaybeFancy;
 use crate::log::Spinner;
-use crate::warning;
 use anyhow::Context;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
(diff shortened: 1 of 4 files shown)


crates/core/tedge/src/cli/diag/logger.rs
```diff
@@ -0,0 +1,28 @@
...
+        let msg = format!($($arg)+);
+        eprintln!("{} {}", "warning:".yellow().bold(), msg);
+        super::log_to_file($log_path, &format!("warning: {msg}"));
+    }};
+}
+
+pub fn log_to_file<P: AsRef<Path>>(path: P, message: &str) {
+    if let Ok(mut file) = OpenOptions::new().create(true).append(true).open(path) {
+        let _ = writeln!(file, "{message}");
+    }
```
didier-wenzek: As this is always called from `DiagCollectCommand` passing `self.summary_file`, I would  open the file just once in `DiagCollectCommand::new()`.

This would also help to highlight a bug. As the directory aimed to collect the data is created too late, most of the calls to `diag_log` and `diag_warning` logs nothing in `summary.log`

crates/core/tedge/src/cli/diag/collect.rs
```diff
@@ -52,31 +55,48 @@ impl Command for DiagCollectCommand {
 
         let mut skipped_count = 0;
         let mut error_count = 0;
+        let mut log_buffer = Vec::new(); // To avoid output while spinner is running
```
didier-wenzek: Why defer logging in `summary.log`? This is working fine without.

didier-wenzek (changes requested):
There is a bug. The call to `file::create_directory_with_defaults(&self.diag_dir)` in `DiagCollectCommand::execute()` at the very first step before `self.read_diag_plugins().await?`, otherwise all the warnings emitted by `diag_log!` and `diag_warning!` are not pushed to `summary.log`.

crates/core/tedge/src/cli/diag/collect.rs
```diff
@@ -52,31 +55,48 @@ impl Command for DiagCollectCommand {
...
             match spinner.finish(res) {
-                Ok(exit_status) if exit_status.success() => {}
-                Ok(exit_status) if exit_status.code() == Some(2) => {
-                    skipped_count += 1;
-                    info!("{plugin} is marked skipped");
-                }
                 Ok(exit_status) => {
-                    error_count += 1;
-                    error!("{plugin} failed with exit status: {exit_status}");
+                    log_buffer.push(format!("{banner}... ✓"));
```
albinsuresh: While testing, I noticed something weird. The ✔️  signs were not captured/rendered properly. Instead, the output was as follows:

```
Executing /usr/share/tedge/diag-plugins/01_tedge.sh... ?~\~S
Executing /usr/share/tedge/diag-plugins/02_os.sh... ?~\~S
Executing /usr/share/tedge/diag-plugins/03_mqtt.sh... ?~\~S
...
```

reubenmiller: The file contains the correct utf-8 characters, I think it is just the vim which isn't showing property.

You can tell vim to use utf-8 character by entering the command `set encoding=utf-8`

reubenmiller: Catting the file also shows that it is fine (assuming your console supports utf-8 characters as well)

reubenmiller:
@rina23q One of the system tests should also be adjusted to add a check for the creation of the `summary.log` file within the tarball.

crates/core/tedge/src/cli/diag/collect.rs
rina23q: ~~Because it didn't work when I tried without the buffer and just call `diag_warning!()` or so on. The `summary.log` was created correctly. However, duplicated messages are printed in the terminal.~~

Now I'm not 100% sure what I tried and didn't work. All I remember was to get duplicated outputs on terminal. I'll do again and come back.

didier-wenzek: Calling `diag_warning!` duplicates messages (on purpose), but directly calling `log_to_fil` will do what you expect.

crates/core/tedge/src/cli/diag/logger.rs
rina23q: I changed the code as the file is opened only once and the bug is fixed in 499da14ff5f1da771893b51e2bb127494fd935cb (Rust) and system test 83e2485520ed29a50b8ef7026b7f8eceb529a959

crates/core/tedge/src/cli/diag/collect.rs
rina23q: You're right. The buffer is removed.

rina23q:
> @rina23q One of the system tests should also be adjusted to add a check for the creation of the `summary.log` file within the tarball.

Done in cf090a6cf9fac09083f4ba17f2ab2652b2f4f572

crates/core/tedge/src/cli/diag/collect.rs
reubenmiller: @albinsuresh Are there any open questions about this?

albinsuresh: @reubenmiller Yeah, it was vim not detecting the encoding properly. Working fine, once it was set explicitly.

Issue #3686, opened 2025-06-13
allow passing c8y.url by a flag to the tedge cert download c8y


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Whilst creating a device migration workflow which moves a device from one Cumulocity tenant, to another, a problem arose when trying to use the `tedge cert download c8y` which made it difficult to download a new device certificate from an alternative Cumulocity instance without changing any the existing `c8y.url` (to avoid breaking the existing connection before a new device certificate is downloaded).

Normally using environment variables could be used to modify the `c8y.url` setting, however it does not work if `sudo` is being used and the `SETENV` setting is not allowed, which then prevents the `tedge` user from passing environment variables to the process with elevated privileges.

```sh
sudo -u tedge sudo TEDGE_C8Y_URL=example.com tedge cert download c8y --device-id example --one-time-password example
sudo: sorry, you are not allowed to set the following environment variables: TEDGE_C8Y_URL
```

Or trying to use the sudo preserve environment (`-E`) option:

```
sudo -u tedge sudo -E TEDGE_C8Y_URL=example.com tedge cert download c8y --device-id example --one-time-password example
sudo: sorry, you are not allowed to preserve the environment
```

Whilst this could be "corrected" by modifying the sudoers rule to allow the tedge user to preserve environment variables, it would be easier (and more secure) to just allow the user to pass the Cumulocity url as a flag.

**Describe the solution you'd like**

Add a new flag, `--url` to the `tedge cert download c8y` command which controls the url used to try to attempt to download the device certificate from.

**Example (proposal)**

```sh
sudo -u tedge sudo tedge cert download c8y --device-id example --one-time-password example --url example.com
```

**Describe alternatives you've considered**

**Additional context**

PR #3689, opened 2025-06-16
feat: add url flag to tedge cert download c8y

reubenmiller:
## Proposed changes

Add `--url` flag to the `tedge cert download c8y` command which allows users to provde the url where the certificate should be downloaded from.

By having a flag, it avoids problems with setting environment variables via sudo if the user does not have the SETENV sudo option set.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3686

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/certificate/cli.rs
+++ b/crates/core/tedge/src/cli/certificate/cli.rs
@@ -17,6 +17,8 @@ use camino::Utf8PathBuf;
 use certificate::CsrTemplate;
 use clap::ValueHint;
 use std::time::Duration;
+use tedge_config::models::HostPort;
+use tedge_config::models::HTTPS_PORT;
 use tedge_config::tedge_toml::OptionalConfigError;
 use tedge_config::tedge_toml::ProfileName;
 use tedge_config::TEdgeConfig;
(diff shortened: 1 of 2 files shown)


didier-wenzek (commented):
The code is correct, however I wonder why a specific treatment for this setting. There are many place where `tedge` uses settings that can be overwritten by env variable but not with a command argument. A more systematic fix is what you propose here https://github.com/thin-edge/thin-edge.io/pull/3688

reubenmiller:
> The code is correct, however I wonder why a specific treatment for this setting. There are many place where `tedge` uses settings that can be overwritten by env variable but not with a command argument. A more systematic fix is what you propose here #3688

The motivation for the flag is to avoid relying on sudo to be configured correctly, as if someone has installed thin-edge.io without the linux packaging, then they will also face the same issue.

And looking in the future, having the host be configurable, would possible make the `tedge cert download c8y` command also work with another EST server (not just Cumulocity), for example if there was a local EST server to issue certificates to clients in the local network.

didier-wenzek (approved):
Approved

PR #3696, opened 2025-06-18
chore: tedge diag's tarball path prints out to stdout

rina23q:
## Proposed changes
Only the tarball path prints out to `stdout` to make the following usecase work.

```sh
OUTPUT_PATH=$(tedge diag collect --name some_file_name)
tedge upload c8y --file "$OUTPUT_PATH"

# or in a one-liner
tedge upload c8y --file "$(tedge diag collect)"
```

## Types of changes

## Paste Link to the issue
#3646 

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/diag/collect.rs
+++ b/crates/core/tedge/src/cli/diag/collect.rs
@@ -201,6 +201,7 @@ impl DiagCollectCommand {
 
         // Cannot write this message to summary.log since the tarball has already been created
         eprintln!("Diagnostic information saved to {tarball_path}");
+        println!("{tarball_path}");
         Ok(tarball_path)
     }
 }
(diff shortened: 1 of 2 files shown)


crates/core/tedge/src/cli/diag/collect.rs
```diff
@@ -200,7 +200,8 @@ impl DiagCollectCommand {
         tar.finish()?;
 
         // Cannot write this message to summary.log since the tarball has already been created
-        eprintln!("Diagnostic information saved to {tarball_path}");
+        eprint!("Diagnostic information saved to: ");
+        println!("{tarball_path}");
```
didier-wenzek: Splitting a message over two channels is a bit unusual!

I would simply duplicate the message on each channel, serving two specific purposes (user feedback, script integration).

```suggestion
        eprintln!("Diagnostic information saved to {tarball_path}");
        println!("{tarball_path}");
```

reubenmiller: And worst case if people complain about that the "duplicate messages" we can always add a conditional to only print the stdout message if the output is not a terminal

rina23q: Applied in a68a83b48df828de37d37c36a246ebeb4a8032c2

didier-wenzek (approved):
Approved

PR #3698, opened 2025-06-20
feat: tedge diag collect security enhancement

rina23q:
## Proposed changes

To enhance the security, if the process is run by root, the plugin must meet those conditions:
* owned by root
* not writable by other users

Otherwise, these files will be skipped.
```sh
warning: Skipping file: /setup/diag-plugins/owned_by_non_root.sh (not owned by root)
warning: Skipping file: /setup/diag-plugins/writable_by_other_users.sh (writable by non-root users)
```

(P.S. This is the last follow-up task for `tedge diag collect`!)

## Types of changes

## Paste Link to the issue
#3646

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/diag/collect.rs
+++ b/crates/core/tedge/src/cli/diag/collect.rs
@@ -8,6 +8,7 @@ use camino::Utf8PathBuf;
 use flate2::write::GzEncoder;
 use flate2::Compression;
 use std::collections::BTreeSet;
+use std::os::unix::fs::MetadataExt;
 use std::os::unix::fs::PermissionsExt;
 use std::process::ExitStatus;
 use std::time::Duration;
@@ -17,6 +18,7 @@ use tedge_config::models::AbsolutePath;
 use tedge_config::TEdgeConfig;
 use tedge_utils::file;
 use tracing::debug;
+use uzers;
 
 #[derive(Debug)]
 pub struct DiagCollectCommand {
@@ -139,13 +141,9 @@ impl DiagCollectCommand {
 
         while let Some(entry) = entries.next_entry().await? {
             if let Ok(path) = Utf8PathBuf::from_path_buf(entry.path()) {
-                if path.extension() == Some("ignore") {
-                    debug!("Skipping ignored file: {:?}", entry.path());
-                } else if path.is_file() && is_executable(&path).await {
+                if validate_plugin(&path, logger).await {
                     plugins.insert(path);
                     continue;
-                } else {
-                    logger.warning(&format!("Skipping non-executable file: {:?}", entry.path()));
                 }
             } else {
                 logger.warning(&format!("Ignoring invalid path: {:?}", entry.path()));
@@ -160,7 +158,7 @@ impl DiagCollectCommand {
     ) -> Result<ExitStatus, anyhow::Error> {
         let plugin_name = plugin_path
             .file_stem()
-            .with_context(|| format!("No file name for {}", plugin_path))?;
+            .with_context(|| format!("No file name for {plugin_path}"))?;
         let plugin_output_dir = self.diag_dir.join(plugin_name);
         let output_file = plugin_output_dir.join("output.log");
         file::create_directory_with_defaults(&plugin_output_dir)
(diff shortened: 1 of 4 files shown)


crates/core/tedge/src/cli/diag/collect.rs
```diff
@@ -206,9 +204,40 @@ impl DiagCollectCommand {
...
     match tokio::fs::metadata(path).await {
-        Ok(metadata) => metadata.permissions().mode() & 0o111 != 0,
+        Ok(metadata) => {
+            if metadata.permissions().mode() & 0o111 == 0 {
+                logger.warning(&format!("Skipping file: {path} (not executable)"));
+                return false;
+            }
+
+            // extra metadata check when the process is running by the root user for the security
+            if get_current_uid() == 0 {
```
albinsuresh: Do we really need the additional `uzers` crate to fetch the uid? Could have fetched it from `MetadataExt`, right?

crates/core/tedge/src/cli/diag/collect.rs
```diff
@@ -206,9 +204,40 @@ impl DiagCollectCommand {
...
+                if metadata.uid() != 0 {
+                    logger.warning(&format!("Skipping file: {path} (not owned by root)"));
+                    return false;
+                }
+                if metadata.permissions().mode() & 0o022 != 0 {
```
albinsuresh, line 231: I didn't really understand the reasoning behind this restriction. Even the one above. What if the normal user on the device created the script? Why reject that? Do we have similar restrictions for the software management plugins as well?

tests/RobotFramework/tests/tedge/diag/tedge_diag_collect.robot
```diff
@@ -68,8 +70,21 @@ Invalid plugin is skipped
...
+    ...    touch /setup/diag-plugins/owned_by_non_root.sh && chown tedge:tedge /setup/diag-plugins/owned_by_non_root.sh && chmod +x /setup/diag-plugins/owned_by_non_root.sh
+    Execute Command
+    ...    touch /setup/diag-plugins/writable_by_other_users.sh && chmod a+wx /setup/diag-plugins/writable_by_other_users.sh
+    ${stderr}=    Execute Command
+    ...    sudo tedge diag collect --plugin-dir /setup/diag-plugins
+    ...    stdout=${False}
+    ...    stderr=${True}
+    ...    ignore_exit_code=${True}
+    Should Contain    ${stderr}    (not owned by root)
+    Should Contain    ${stderr}    (writable by non-root users)
```
didier-wenzek: To be meaningful, these checks should include filenames.

```suggestion
    Should Contain    ${stderr}   owned_by_non_root.sh  (not owned by root)
    Should Contain    ${stderr}   writable_by_other_users.sh (writable by non-root users)
```

crates/core/tedge/src/cli/diag/collect.rs
didier-wenzek: > Do we really need the additional `uzers` crate to fetch the uid? Could have fetched it from `MetadataExt`, right?

The point is to get the id of the current user, not the id of the file owner.

As two kinds of user id are used here, I would be explicit:

 ```suggestion
            if uzers.get_current_uid() == 0 {
```

crates/core/tedge/src/cli/diag/collect.rs
didier-wenzek: The goal is to prevent a user from misleading the root user, i.e. to write a malware script that would be executed with admin privileges by `sudo tedge diag collect`.

Software management plugins are not executed directly by the agent but delegated to `sudo`.

crates/core/tedge/src/cli/diag/collect.rs
rina23q: > The point is to get the id of the current user, not the id of the file owner.

This is correct.

Changed as suggested in 8b7b7875af34e50bb1ce906740fe9a1bc90b1b45

tests/RobotFramework/tests/tedge/diag/tedge_diag_collect.robot
rina23q: Done in https://github.com/thin-edge/thin-edge.io/commit/8b7b7875af34e50bb1ce906740fe9a1bc90b1b45

didier-wenzek (approved):
Approved

albinsuresh (approved):
LGTM

PR #3706, opened 2025-06-24
test: improve self-signed certificate cleanup

reubenmiller:
## Proposed changes

Improve the tracking of self-signed certificates within the tests. With the changes, the system tests no longer leave self-signed certificate which where generated within the tests.

* New keyword to allow a certificate to be registered for cleanup
* Record the certificates after each test (instead of assuming the cert still exists at the end of the suite)
* Fix thumbprint parsing logic
* Use Cumulocity CA to register devices in the HSM tests, however a the RSA private key with a bit length of 1024 had to be removed as it isn't supported by the Cumulocity CA and is anyway deemed insecure by today's security recommendatons

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/pkcs11/private_key_storage.robot
+++ b/tests/RobotFramework/tests/pkcs11/private_key_storage.robot
@@ -88,7 +88,7 @@ Connects to C8y using an RSA key
     type=rsa    bits=4096
     type=rsa    bits=3072
     type=rsa    bits=2048
-    type=rsa    bits=1024
+    # type=rsa    bits=1024    # RSA 1024 is considered to be insecure is not supported when using the Cumulocity Certificate Authority feature
 
 Connects to C8y supporting all TLS13 ECDSA signature algorithms
     [Documentation]    Check that we support all ECDSA sigschemes used in TLS1.3, i.e: ecdsa_secp256r1_sha256,


Bravo555:
Instead of manually tracking individual certificates, couldn't we perhaps just remove ALL the certificates from the tenant that match the test device name (and children)? These names should be globally unique, so no other device should be affected and additionally:

- if I debug a suite and create more certificates from the shell these would also get cleaned up
- if the test suite mistakenly didn't use the new keyword and uploaded the cert manually, this would also get cleaned up automatically
- the implementation could be much simplified

I think this would be a good enough solution assuming that from within the tests we only create certs with the name of the device (and children) or I'm otherwise not missing anything else.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -181,52 +181,46 @@ Connect to C8y using new keypair
     ELSE
         Fail    Wrong key type provided.
     END
-    VAR    ${cert_path}=    /etc/tedge/device-certs/${object_name}.pem
+    ${cert_path}=    Execute Command    cmd=tedge config get device.cert_path    strip=${True}
```
Bravo555: todo: this is left unused

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -73,11 +73,11 @@ Select Private key using a request URI
     Should Contain    ${stderr}    item=cryptoki: socket (key: pkcs11:token=token123)
 
     Execute Command    cmd=tedge config unset device.key_uri
-    Execute Command    cmd=tedge config set c8y.device.key_uri pkcs11:token=token123
+    Execute Command    cmd=tedge config set device.key_uri pkcs11:token=token123
```
Bravo555, line 76: question: why this change here? Does cert registration keyword not work when key is under `[c8y]`, or is it style-related?

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
```diff
@@ -299,30 +325,60 @@ def log_operations(self, mo_id: str, status: str = None):
...
-        if result.return_code != 0:
-            log.info("Failed to get device certificate fingerprint. %s", result.stdout)
-            return
+        # Parse the certificate details via tedge cert show
+        lines = []
+        try:
+            command = "tedge cert show"
+            if cloud_profile:
+                command += f" c8y --profile {cloud_profile}"
+            lines = self.execute_command(command, ignore_exit_code=True).splitlines()
```
Bravo555: suggestion: Move `c8y` to base command and mention in the main keyword docs that cert cleanup works only for c8y

Ah, so that's why `c8y.` prefix from some config sets in some tests were removed. We only pass `c8y` if the profile is explicitly set.
The feature is used in the `multi_cloud` suite, but still we only use `c8y` in all relevant test suites, so how about just using `c8y` here? That way both `c8y.device.cert_path` and `device.cert_path` will be considered if profiles aren't used.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
reubenmiller: Resolved by 5e6476fbd0d2e9a5f1b8793dabfdb6d5c4ead204

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
reubenmiller: The reason was just to simplify the setup and use the default thin-edge.io wide configuration rather than the c8y specific setup.

You only really need to use the cloud specific settings if you wish to maintain multiple clouds with independent certificates (which I would anyway recommend against as it generally is harder to manage).

reubenmiller: The secondary motivation was that the certificate cleanup logic wasn't written to check all of the possible certificate locations (as there are just too many to list)...so I went for the easier way and just use the sensible defaults.

tests/RobotFramework/libraries/ThinEdgeIO/ThinEdgeIO.py
reubenmiller: I don't think it's worth supporting all configurations, as the usage is very niche anyway. The multi-cloud setup will anyway use c8y-ca in the future which will eliminate the need for cleaning up the certificate...so again, favouring simplicity over catering for all possible use-cases

Bravo555: Still, I think at the very least a test writer would expect the certificate registered by this keyword to be the same as the one `tedge connect c8y` uses, which isn't currently the case when `c8y.device.*` settings are used.

`tedge cert show c8y` shows cert under `c8y.device.cert_path` if it's defined or `device.cert_path` if it doesn't and it's consistent with `tedge connect c8y`. I'd expect here we could also just do `tedge cert show c8y` and append only `--profile {cloud_profile}` if it's necessary?

reubenmiller: I guess I was opting for the more compatible `tedge cert show` command (as `c8y` was only added in newer versions)...but yeah it kind of makes slightly more sense.

Resolved by be02c4930128406c39a75e2d13df2720a117b304

reubenmiller:
@Bravo555 I pushed a fix, ab5152152d6e62f79b38304ea94e6462f9a93893, as I noticed that the Cumulocity device/user cleanup logic wasn't being run when self-signed certificates were not being used...but the latest run leaves a clean environment now.

Bravo555 (approved):
Approved

PR #3707, opened 2025-06-25
feat: tedge connect to c8y mqtt service endpoint

albinsuresh:
## Proposed changes

- [x] Specification (with multiple approaches) for connecting thin-edge to the new Cumulocity MQTT service endpoint.
- [x] Impl finalised approach (connect to mqtt service along with core mqtt)
- [x] mosquitto bridge support
- [x] builtin bridge support

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_mqtt_bridge/src/config.rs
+++ b/crates/extensions/tedge_mqtt_bridge/src/config.rs
@@ -144,7 +144,7 @@ impl BridgeRule {
 
 impl BridgeConfig {
     pub fn new() -> Self {
-        Self::default()
+        Default::default()
     }
 
     pub fn forward_from_local(
@@ -214,6 +214,7 @@ impl BridgeConfig {
             local_to_remote,
             remote_to_local,
             bidirectional_topics,
+            ..
         } = self;
 
         let (bidir_local_topics, bidir_remote_topics) = bidirectional_topics.into_iter().unzip();
--- a/tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
+++ b/tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
@@ -45,7 +45,7 @@ Connect to Cumulocity MQTT Service endpoint builtin bridge
     ${DEVICE_SN}=    Setup    connect=${False}
     Execute Command    tedge config set mqtt.bridge.built_in true
     Execute Command    tedge config set c8y.mqtt_service.enabled true
-    Execute Command    tedge config set c8y.mqtt_service.topics demo/topic
+    Execute Command    tedge config set c8y.mqtt_service.topics 'sub/topic,demo/topic'
     Execute Command    tedge connect c8y
 
     External Identity Should Exist    ${DEVICE_SN}:device:main:service:tedge-mapper-bridge-c8y-mqtt    show_info=False
--- a/crates/extensions/tedge_mqtt_bridge/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_bridge/src/lib.rs
@@ -504,7 +504,11 @@ async fn half_bridge(
                 let topics = topics.clone();
                 // We have to subscribe to this asynchronously (i.e. in a task) since we might at
                 // this point have filled our cloud event loop with outgoing messages
-                tokio::spawn(async move { recv_client.subscribe_many(topics).await.unwrap() });
+                tokio::spawn(async move {
+                    for topic in topics {
+                        recv_client.subscribe(topic).await.unwrap()
+                    }
+                });
 
                 session_present = Some(conn_ack.session_present);
 
@@ -605,7 +609,7 @@ impl MqttEvents for EventLoop {
 }
 #[async_trait::async_trait]
 trait MqttClient: MqttAck + Clone + Send + Sync {
-    async fn subscribe_many(&self, topics: Vec<SubscribeFilter>) -> Result<(), ClientError>;
+    async fn subscribe(&self, topic: SubscribeFilter) -> Result<(), ClientError>;
     async fn publish(
         &self,
         topic: String,
@@ -617,9 +621,10 @@ trait MqttClient: MqttAck + Clone + Send + Sync {
 
 #[async_trait::async_trait]
 impl MqttClient for AsyncClient {
-    async fn subscribe_many(&self, topics: Vec<SubscribeFilter>) -> Result<(), ClientError> {
-        AsyncClient::subscribe_many(self, topics).await
+    async fn subscribe(&self, topic: SubscribeFilter) -> Result<(), ClientError> {
+        AsyncClient::subscribe(self, topic.path, topic.qos).await
     }
+
     async fn publish(
         &self,
         topic: String,
--- a/crates/extensions/tedge_mqtt_bridge/src/test_helpers.rs
+++ b/crates/extensions/tedge_mqtt_bridge/src/test_helpers.rs
@@ -137,7 +137,7 @@ impl MqttAck for BlockingSubscribeClient {
 
 #[async_trait::async_trait]
 impl MqttClient for BlockingSubscribeClient {
-    async fn subscribe_many(&self, _: Vec<SubscribeFilter>) -> Result<(), ClientError> {
+    async fn subscribe(&self, _: SubscribeFilter) -> Result<(), ClientError> {
         pending().await
     }
 
@@ -178,8 +178,8 @@ impl MqttAck for ActionLogger {
 }
 #[async_trait::async_trait]
 impl MqttClient for ActionLogger {
-    async fn subscribe_many(&self, topics: Vec<SubscribeFilter>) -> Result<(), ClientError> {
-        self.log(Action::SubscribeMany(topics));
+    async fn subscribe(&self, topic: SubscribeFilter) -> Result<(), ClientError> {
+        self.log(Action::SubscribeMany(vec![topic]));
         Ok(())
     }
 
@@ -322,7 +322,7 @@ impl MqttAck for ChannelClient {
 
 #[async_trait::async_trait]
 impl MqttClient for ChannelClient {
-    async fn subscribe_many(&self, _: Vec<SubscribeFilter>) -> Result<(), ClientError> {
+    async fn subscribe(&self, _: SubscribeFilter) -> Result<(), ClientError> {
         Ok(())
     }
 


design/decisions/0007-c8y-mqtt-service-connect.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Solutions
+
+To establish a bridge connection to the mqtt service endpoint, the following info must be provided in the connect request:
+
+* host: Same as legacy mqtt endpoint
+* port: 9883 by default
+* auth:
+  * for username/password based authentication
+    * username: `<tenant_id>/<user_name>`
```
didier-wenzek, line 38: What is the relationship between `user_name` and device id?

design/decisions/0007-c8y-mqtt-service-connect.md
```diff
@@ -0,0 +1,171 @@
...
+
+To connect to the mqtt service, it must be enabled before `tedge connect c8y` is executed:
+
+```sh
+tedge config set c8y.mqtt_service.enabled "true"
+```
+
+:::note
+The tenant id which must be passed in the username is retrieved using the `https://<c8y.url>/tenant/currentTenant` REST API.
+An alternative is to take this as an additional `c8y.tenant_id` configuration.
```
didier-wenzek: Requesting the tenant id using C8Y REST API requires to be already connected to C8Y. Using an additional  `c8y.tenant_id` settings imposes less constraints. Also, can the tenant id be derived from C8Y url? At least in the usual cases.

design/decisions/0007-c8y-mqtt-service-connect.md
```diff
@@ -0,0 +1,171 @@
...
+
+- Clear separation from `tedge connect c8y` allows the second connection to be established and disconnected on demand.
+- Existing users are not affected, as the `tedge connect c8y` behavior stays the same.
+- Allows customers to connect exclusively to the new mqtt service, if they don't need the core mqtt connection.
+- There is no real dependency on Cumulocity as multiple `remote` profiles can be used to connect to any generic broker.
+
+**Cons**
+
+- Many of the config parameters defined in the `c8y` profile like the `url`, `cert_path` etc
+  will have to be duplicated in `remote` profile as well.
```
didier-wenzek, line 147: Can we use similar inheritance rules as done for the device certificate which can be globally specified with optional cloud-specific settings?

design/decisions/0007-c8y-mqtt-service-connect.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### Configuration
+
+The mqtt service connection related configs are defined under the `[c8y]` table in `tedge.toml` :
+
+| Config | Description | Default |
+| ------ | ----------- | ------- |
+| `c8y.mqtt_service.enabled` | Whether the mqtt server connection must be established or not | `false` |
+| `c8y.mqtt_service.url` | The config URL for the MQTT service | `<c8y.url>:9883` |
+| `c8y.mqtt_service.topic_prefix` | topic prefix used to bridge mqtt topics | `c8y-mqtt` |
```
didier-wenzek, line 71: What will be the bridge rules?

- Publish messages received on `c8y-mqtt/$topic` to the suffix `$topic`. Okay.
- But what about for the subscriptions?
- As subscribing to wildcard topics (`#` and `+`)  is not allowed, a new tedge config settings is required to let users provide an explicit list of topics to subscribe to.

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: The `device_id` and `username` need not be the same. The `device_id` is only used to retrieve the device credentials from Cumulocity as part of the [initial bootstrap process](https://cumulocity.com/docs/device-integration/mqtt/#step-0-request-device-credentials_).

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: > Requesting the tenant id using C8Y REST API requires to be already connected to C8Y

By "connected to C8Y" if you meant the core mqtt bridge connection, then no, as the proposal was to use the direct REST API here, as we do for device creation in the existing `tedge connect c8y`. But, if you meant "network connection to C8Y", then yes, that's the primary limitation of this choice. We won't be able to do an "offline connect", using this mechanism.

> Also, can the tenant id be derived from C8Y url?

I don't think we can do this reliably, esp with tenant aliases. For example, when I connect to my c8y tenant, I use the aliased url `albin.latest.c8y.io` when the real tenant id is something like `t9836373`. And the mqtt service expects the real tenant id in the `username` and and not the alias.

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: Yes we can inherit the global `device.cert_path` and `device.key_path`, as it is already done for `c8y`, `az` and `aws`. But, if the `c8y.device.cert_path` has overridden the global value, then we diverege, unless the new `remote` profile inherits from `c8y_device` settings as well. Such a coupling would go against the spirit of modelling this new `remote` profile as a generic cloud profile.

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: > But what about for the subscriptions?

Aah yes, missed that in this proposal. For the second approach with `remote` cloud profile, the plan was to replicate the same configs from other cloud profiles like `az`, `aws` that already includes the `topics` config setting that captures the topics to subscribe to.

design/decisions/0007-c8y-mqtt-service-connect.md
```diff
@@ -0,0 +1,171 @@
...
+  Publishes are rejected when the queue is full.
+- Each published message has a time-to-live, after which it is cleared from the queue.
+  The subscribed clients are not notified of expired messages.
+
+## Solutions
+
+To establish a bridge connection to the mqtt service endpoint, the following info must be provided in the connect request:
+
+* host: Same as legacy mqtt endpoint
+* port: 9883 by default
```
albinsuresh, line 34:
```suggestion
* port: 9883 by default
* client id: device id (but need not be the device id itself)
```

design/decisions/0007-c8y-mqtt-service-connect.md
reubenmiller: I believe the requirement of having the tenant id as the username may be lifted soon (though it will still be required if the trusted certificate is present on more than 1 tenant within the same Cumulocity instance).

reubenmiller:
Given that there is a clear vision that the Cumulocity mqtt-service will replace the MQTT core functionality, then approach 1 is probably the best approach for delivering the immediate "mqtt-service support" feature. As it would also add a smooth transition to only using the mqtt-service without changing much on the device side, e.g. `tedge connect c8y` would still be the same UX but it would just do something differently behind the scenes.

Though I still think we should explore the idea of support the connection to any MQTT broker by allowing option 2, but we could do that specific stuff in a different feature implementation.

design/decisions/0007-c8y-mqtt-service-connect.md
didier-wenzek: > The `device_id` and `username` need not be the same.

Then better to be explicit.

```suggestion
    * username: `<tenant_id>/<device_id>`
```

design/decisions/0007-c8y-mqtt-service-connect.md
didier-wenzek: > By "connected to C8Y" if you meant the core mqtt bridge connection, then no, as the proposal was to use the direct REST API here, as we do for device creation in the existing `tedge connect c8y`.

My point is related to this use of a direct REST API call. One needs to be authenticated.

- For password based authentication, things are okay as a password is given along the device id.
- But for cert based authentication, one needs to fetch a JWT token using the regular C8Y MQTT connection.

reubenmiller: The `/tenant/loginOptions` is an endpoint that doesn't require authentication, however you'd have to parse a bit more information from it to check potentially get the tenant id.

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: Resolved by 8675e0f

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: Introduced the tedge config: `c8y.tenant_id` for the same in 9979965, instead of the REST call.

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: No it is not the tedge `device_id`. The tedge `device.id` is used as the remote client id of the mqtt bridge, but the username would be the username of any user on the tenant or the username received with the device bootstrap process.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -433,6 +438,28 @@ define_tedge_config! {
...
             interval: SecondsOrHumanTime,
         },
+
+        mqtt_service: {
+            /// Wheather to connect to the MQTT service endpoint or not
+            #[tedge_config(example = "true", default(value = false))]
+            enabled: bool,
+
+            /// MQTT service endpoint for the Cumulocity tenant, with optional port.
+            #[tedge_config(example = "mqtt.your-tenant.cumulocity.com:1234")]
```
didier-wenzek: 1n example using the default port value would be more helpful.

```suggestion
            #[tedge_config(example = "mqtt.your-tenant.cumulocity.com:9883")]
```

design/decisions/0007-c8y-mqtt-service-connect.md
didier-wenzek: __Could `c8y.tenant_id` be computed when not set?__

Having an explicit setting for `c8y.tenant_id` is simpler. However as a user this leads to something a bit surprising. 

Simply enabling the `mqtt_service` breaks `tedge connect c8y --test` because  `c8y.tenant_id`  is not set.

```shell
$ tedge connect c8y --test          
test connection to Cumulocity cloud.:
        device id: demo-1.5
        cloud profile: <none>
        cloud host: didier.latest.stage.c8y.io:8883
        auth type: Certificate
        cryptoki: off
        bridge: mosquitto
        service manager: systemd
        proxy: Not configured
Verifying device is connected to cloud... ✓
Checking Cumulocity is connected to intended tenant... ✓
Connection check to c8y cloud is successful.

$ tedge config set c8y.mqtt_service.enabled true

$ tedge connect c8y --test                      
Error: failed to test connection to Cumulocity cloud.

Caused by:
    A value for 'c8y.tenant_id' is missing.
        A value can be set with `tedge config set c8y.tenant_id <value>`
```

tests/RobotFramework/tests/cumulocity/bridge_config/builtin_bridge.robot
```diff
@@ -10,7 +10,7 @@ Test Teardown       Get Logs
 *** Test Cases ***
 Connection test
     [Documentation]    Repeatedly test the cloud connection
-    FOR    ${attempt}    IN RANGE    0    10    1
+    FOR    ${attempt}    IN RANGE    0    1    1
```
didier-wenzek: Why this change?

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
```diff
@@ -0,0 +1,55 @@
...
+Test Teardown       Get Logs
+
+
+*** Test Cases ***
+Connect to Cumulocity MQTT Service endpoint
+    ${DEVICE_SN}=    Setup    connect=${False}
+    Execute Command    tedge config set c8y.tenant_id t37070943
+    Execute Command    tedge config set c8y.mqtt_service.enabled true
+    Execute Command    tedge config set c8y.mqtt_service.topics 'sub/topic,demo/topic'
+    Execute Command    tedge connect c8y
```
didier-wenzek, line 15: How can I check that the connection to the MQTT service is actually established?

- Neither `tedge connect c8y` nor `tedge connect c8y --test` tells a word about the MQTT service and status.
- I tried on purpose to break the config, no errors are returned by `tedge connect`.

Good. The bridge status published on `te/device/main/service/mosquitto-c8y-mqtt-bridge/status/health` is working and reports broken as well as correct settings.

=> I would add a test assertion that a status message of `1` is received on this topic.

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
```diff
@@ -0,0 +1,55 @@
...
+    # TODO: Validate message received on test/topic on the other client
+
+    Sleep    1s
+
+Connect to Cumulocity MQTT Service endpoint builtin bridge
+    ${DEVICE_SN}=    Setup    connect=${False}
+    Execute Command    tedge config set mqtt.bridge.built_in true
+    Execute Command    tedge config set c8y.tenant_id t37070943
+    Execute Command    tedge config set c8y.mqtt_service.enabled true
+    Execute Command    tedge config set c8y.mqtt_service.topics 'sub/topic,demo/topic'
```
albinsuresh, line 48: There's something really weird with multiple subscriptions on MQTT service. MQTT service is rejecting subscription requests with multiple topics from the built-in bridge, which is why this test case was failing. Here is the same reproduced with `mosquitto_sub` client:

```sh
$ mosquitto_sub -v -h albin.latest.stage.c8y.io -p 9883 -i albin-123 --key device.key --cert device.crt --cafile c8y-root.crt -u t37070943 -t sub/topic -t demo/topic
All subscription requests were denied.
```

Once you limit the subscription to a single topic, it works. This test also works when there is only one subscription topic.

But to my surprise, the mosquitto bridge with multiple subscriptions works just fine (as done in the first test case).

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -360,3 +216,248 @@ pub fn service_monitor_client_config(
...
+
+    let mut tc = BridgeConfig::new();
+    let local_prefix = format!("{}/", c8y_config.mqtt_service.topic_prefix.as_str());
+
+    for topic in cloud_topics {
+        tc.forward_from_remote(topic, local_prefix.clone(), "")?;
+    }
+
+    // Templates
+    tc.forward_from_local("#", local_prefix.clone(), "")?;
```
albinsuresh, line 415: Unfortunately this bridge rule triggers infinite loops. This `forward_from_local` rule bridges anything received on `c8y-mqtt/#` to the cloud and the above `forward_from_remote` rule bridges anything received on the subscribed topics to local topics with the `c8y/` prefix. And this creates an infinite loop when a message is received from the cloud.

But again, the mosquitto bridge detects and handles these cycles somehow, but the built-in bridge doesn't. So, either we should improve the built-in bridge to handle the same, or have either one of the following rules enforced for the mqtt service bridge:

1. Use different prefixes to bridge outbound and inbound messages
2. Do not bridge all published messages on `c8y-mqtt/#`, but make the user explicitly specify the publish topics to be bridged as well.

tests/RobotFramework/tests/cumulocity/bridge_config/builtin_bridge.robot
albinsuresh: Accidental change. Reverted in 0a066063d

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: Hmm... Not aware of any mechanisms to derive it statically from other config fields, other than the dynamic methods mentioned above like fetching it via REST. We could fall back to those methods when this value is not set.

That being said, in the context of the issue that you highlighted about `c8y --test`, we need to define what that `--test` even means for the mqtt service endpoint and how/what we can test.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -433,6 +438,28 @@ define_tedge_config! {
...
         },
+
+        mqtt_service: {
+            /// Wheather to connect to the MQTT service endpoint or not
+            #[tedge_config(example = "true", default(value = false))]
+            enabled: bool,
+
+            /// MQTT service endpoint for the Cumulocity tenant, with optional port.
+            #[tedge_config(example = "mqtt.your-tenant.cumulocity.com:9883")]
+            #[tedge_config(default(from_optional_key = "c8y.url"))]
```
jarhodes314: Is this not better derived from `c8y.mqtt`, which will likely already have the correct domain for MQTT connections (I'm assuming this is the same issue as for standard MQTT connections)?

crates/core/tedge_mapper/src/c8y/mapper.rs
jarhodes314: There is some deduplication logic we already have for the AWS shadow topics. Unfortunately there isn't a nice way to naturally avoid loops with MQTT v3 since there is no way to opt out of receiving messages you published (assuming you are subscribed to the relevant topic). So to detect and avoid the loops, the bridge needs to know the topic is bidirectional, then cache the relevant publish packet until we receive the message back from the broker, at which point we can immediately acknowledge the message and not forward it to the cloud.

We can probably fix things in this (relatively) simple case by telling the bridge to treat the cloud topics as bidirectional but not (directly) subscribe to them locally. For testing this works against mosquitto, forwarding the cloud topics bidirectionally (using `tc.forward_bidirectionally`) and subscribing separately to `#` should be fine, since mosquitto will essentially ignore the double subscription (but this would be problematic with nanomq).

In short, provided we don't have complex overlapping rules, it should be quite possible to make this _just work_.

albinsuresh: > the bridge needs to know the topic is bidirectional, then cache the relevant publish packet until we receive the message back from the broker, at which point we can immediately acknowledge the message and not forward it to the cloud.

That sounds fine. But, to be on the safer side with that caching (esp with half open connections), it'd be better if the cached entries have a time-to-live as well, post-which they are cleared whether or not the duplicate is received from the cloud or not. Sure, this might lead to a late duplicate message getting resent to the cloud once more, but that should settle after a few loops once the network stabilises.

> We can probably fix things in this (relatively) simple case by telling the bridge to treat the cloud topics as bidirectional but not (directly) subscribe to them locally

I didn't understand that proposal. How does a bidirectional subscription to the same topic (e.g: `demo/topic`) both on the cloud and the local help? Esp since the expectation on the local broker is to bridge messages on `c8y-mqtt/demo/topic` (included in `c8y-mqtt/#`) and **not** `demo/topic` directly. Could you explain where/how the loop is broken when a message is received first from the cloud on `demo/topic`?

jarhodes314: If we do a bidirectional forward, this tells the bridge to break loops on those messages. It can have different prefixes on local and cloud topics, which is exactly how the AWS shadow topic works at the moment.

Because the bridge knows in advance the topic will cause a loop, it can cache the published message when it forwards to local/cloud and then use that cached message to ignore the same message when the bridge receives the message from the target broker.

The only difference from the AWS shadow topics is that we have a wildcard subscription on local and not the cloud so if we add bidirectional rules to perform de-duplication for the cloud topics, we would then subscribe to `c8y-mqtt/#` on the local broker and duplicate all of those subscriptions. This isn't going to cause a problem with mosquitto, so should be a valid way to verify that the existing logic works, but we will need to ensure we don't have duplicate subscriptions in the final version as that will misbehave with NanoMQ.

didier-wenzek: >  1. Use different prefixes to bridge outbound and inbound messages

I see this as the simplest and more robust route.

reubenmiller: But still using a common prefix to group the c8y-mqtt messages together, but just adding two nested topics:
* `c8y-mqtt/in`
* `c8y-mqtt/out`

docs/src/operate/c8y/connect-mqtt-service.md
```diff
@@ -0,0 +1,111 @@
...
+Most of the configurations used to connect to the MQTT service endpoint of Cumulocity are same as
+the ones used to connect to the core MQTT endpoint.
+
+1. Configure Cumulocity tenant URL:
+
+   <UserContext>
+   
+   ```sh
+   sudo tedge config set c8y.url $C8Y_URL
+   ```
```
didier-wenzek: I would use here `c8y.mqtt_service.url` and add a note about `c8y.mqtt` and `c8y.url`.

albinsuresh: It was intentional, as it is closer to the existing connection flow, and the user doesn't have to set the `c8y.mqtt_service.url` explicitly in most cases. Instead, how about keeping the `c8y.url` here and add a note about the possibility of overriding it with `c8y.mqtt_service.url` when needed?

didier-wenzek: > Instead, how about keeping the `c8y.url` here and add a note about the possibility of overriding it with `c8y.mqtt_service.url` when needed?

A good alternative. The point being to tell the relationship between these settings.

docs/src/operate/c8y/connect-mqtt-service.md
```diff
@@ -0,0 +1,111 @@
...
+import UserContext from '@site/src/components/UserContext';
+import UserContextForm from '@site/src/components/UserContextForm';
+import BrowserWindow from '@site/src/components/BrowserWindow';
+
+:::tip
+#### User Context {#user-context}
+
+You can customize the documentation and commands shown on this page by providing relevant settings which will be reflected in the instructions. It makes it even easier to explore and use %%te%%.
+
+<UserContextForm settings="DEVICE_ID,C8Y_URL,C8Y_TENANT_ID,C8Y_USERNAME,C8Y_PASSWORD" />
```
albinsuresh: Accompanying [tedge-docs PR](https://github.com/thin-edge/tedge-docs/pull/119) that introduces the keys `C8Y_TENANT_ID` and `C8Y_PASSWORD`.

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: Resolved by [702e370](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/702e370811520b6c5ac40d0993b222b0a4d03c1d)

docs/src/operate/c8y/connect-mqtt-service.md
albinsuresh: Resolved by  9d4ebb0

crates/core/tedge_mapper/src/c8y/mapper.rs
albinsuresh: Fixed in [b25610a](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/b25610ab1c3556b6be04801683d1185b4245c637) as @jarhodes314 suggested as that was also a trivial one-liner.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
albinsuresh: Resolved by [015ee3d](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/015ee3db5afa211849a314ba61e1b7e47511caf8)

albinsuresh:
Here are some of the rough edges around this feature in its current state:
1. When an existing user wants to try the mqtt service feature, he's forced to do a reconnect, since mqtt service connection is part of the same `tedge connect c8y` command. Though this was done for a seamless migration from core mqtt to mqtt service under-the-hood in the future, during this transition phase, it could be a bit annoying to some.
1. The `tedge connect c8y` command is not very explicit about 2 connections being established to 2 different endpoints, expect for the following line in the output: `Creating mosquitto bridge to MQTT service... ✓`
1. The `tedge connect c8y --test` currently relies on the health status message of the mqtt service bridge to determine if it's healthy or not, unlike the core mqtt bridge test that really sends the JWT request message and waits for the response. With the mqtt service, since there is no such inbuilt request/response topic pairs, we were left with no choice but to rely only on the health status. The issue with this approach is that the health status is not updated immediately on events such as a network disconnection as it can take a while (the ping interval) for the bridge to detect this.
    
    The saving grace here is that we're checking both the core mqtt connection and the mqtt service connection together, and hence for issues like network disconnection, the jwt token check would detect it anyway and fail the `c8y --test`. It is only an issue when only the connection to mqtt service is broken while the core mqtt connection is still intact.
1. Even though the mqtt service rejects subscription requests to certain reserved topics like `s/ds`, if that topic is included in the `c8y.mqtt_service.topics` list along with others, the bridge still gets established with this problematic topic silently ignored. This lack of feedback is a bit problematic. With mosquitto bridge, only the problematic topics are ignored. With built-in bridge, all the topics are ignored. So, effectively you'd end up with a bridge that appears healthy, but some/none of the subscriptions are working.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -433,6 +438,28 @@ define_tedge_config! {
             #[tedge_config(example = "60m", default(from_str = "60m"))]
             interval: SecondsOrHumanTime,
         },
+
+        mqtt_service: {
+            /// Wheather to connect to the MQTT service endpoint or not
```
reubenmiller: Typo.

```suggestion
            /// Whether to connect to the MQTT service endpoint or not
```

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
reubenmiller: This was discussed offline, but the outcome was that it seems to be a current limitation of the MQTT Service, however we provided the appropriate RnD team that supporting multiple subscriptions would mean that the MQTT service would be more compatible with different MQTT tooling (and what users expect when interacting with such as service)

docs/src/operate/c8y/connect-mqtt-service.md
reubenmiller: Please remove the `C8Y_PASSWORD`, as it encourages bad security practices (e.g. getting the user to expose their password by storing it in unencrypted browser storage, and potentially exposing the password via the examples if they are sharing a screen with someone).

docs/src/operate/c8y/connect-mqtt-service.md
```diff
@@ -0,0 +1,120 @@
...
+The test client must use a client id that is different from the device id.
+:::
+
+Once subscribed, publish to the same topic from the %%te%% device:
+
+```sh
+tedge mqtt pub c8y-mqtt/test/topic "hello world"
+```
+
+Now, validate that the same message is received by the test client.
```
reubenmiller: We're going to need to remove this section, as device isolation is starting to be rolled out this week, which will invalidate these instructions...and given that there won't be an easy way to validate this for now, it's better to have it removed.

docs/src/operate/c8y/connect-mqtt-service.md
```diff
@@ -0,0 +1,120 @@
...
+   
+   </UserContext>
+
+1. Enable connection to MQTT service endpoint:
+
+   ```sh
+   sudo tedge config set c8y.mqtt_service.enabled true
+   ```
+
+1. Provide a topic to subscribe to (only a single topic is allowed at the moment):
```
reubenmiller: Is there is a visible error if the bridge tries to subscribe to multiple topics? As I noticed that the system tests are using multiple topics (at least the mosquitto bridge test, but I also changed the in-built bridge tests to use multiple topics and it connects fine now...though given the lack of validation it is difficult to say if it actually will receive messages on that topic).

E.g. the mosquitto bridge uses the following config.

```
tedge config set c8y.mqtt_service.topics 'sub/topic,demo/topic'
```

docs/src/operate/c8y/connect-mqtt-service.md
```diff
@@ -0,0 +1,120 @@
...
+   If the MQTT service url is different from the one that would be derived from `c8y.url` or `c8y.mqtt`,
+   set`c8y.mqtt_service.url` explicitly.
+   :::
+
+1. If you're using device certificates to connect to Cumulocity, set the tenant id (skip this when using username/password):
+
+   <UserContext>
+   
+   ```sh
+   sudo tedge config set c8y.tenant_id $C8Y_TENANT_ID
```
reubenmiller: It seems this restriction has been lifted, so it would be nice if tenant_id wasn't a required property any more...sure we can still use it if the user provides it, but we should fail if it isn't set.

Below shows the mosquitto_sub command that I used within a system test (with the mapper turned off).

```
mosquitto_sub -h thin-edge-io.eu-latest.cumulocity.com -p 9883 --key /etc/tedge/device-certs/tedge-private-key.pem --cert /etc/tedge/device-certs/tedge-certificate.pem -t 'hello/topic' --debug -i $(tedge config get device.id) --capath /etc/ssl/certs 
Client TST_merge_intense_gulf sending CONNECT
Client TST_merge_intense_gulf received CONNACK (0)
Client TST_merge_intense_gulf sending SUBSCRIBE (Mid: 1, Topic: hello/topic, QoS: 0, Options: 0x00)
Client TST_merge_intense_gulf received SUBACK
Subscribed (mid: 1): 0
^CClient TST_merge_intense_gulf sending DISCONNECT
```

albinsuresh: Resolved by [ea87f84](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/ea87f84fa29d7cac117f62d68bb0555e0a0e8ab4)

docs/src/operate/c8y/connect-mqtt-service.md
albinsuresh: Resolved by [73cca52](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/73cca52944ad68153eecbcb0ec3ccd0a5bec728d)

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
albinsuresh: Resolved by [ea87f84](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/ea87f84fa29d7cac117f62d68bb0555e0a0e8ab4)

docs/src/operate/c8y/connect-mqtt-service.md
albinsuresh: MQTT service does not support multiple topics in a subscription request. Mosquitto seems to be handling this fine by sending multiple subscription requests. But, our built-in bridge does not. This MQTT service restriction may not be fixed before GA. So, we need to handle it on our end.

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
```diff
@@ -0,0 +1,53 @@
...
+*** Test Cases ***
+Connect to Cumulocity MQTT Service endpoint
+    ${DEVICE_SN}=    Setup    connect=${False}
+    Execute Command    tedge config set c8y.mqtt_service.enabled true
+    Execute Command    tedge config set c8y.mqtt_service.topics 'sub/topic,demo/topic'
+    Execute Command    tedge connect c8y
+
+    # TODO: Subscribing to test/topic from another client
+    Execute Command    tedge mqtt pub c8y-mqtt/test/topic '"hello"'
+    # TODO: Validate message received on test/topic on the other client
```
didier-wenzek: We really need to find a way to test a connection the MQTT service. Exchanging messages from one client to another will not work.

docs/src/operate/c8y/connect-mqtt-service.md
didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3707/commits/73cca52944ad68153eecbcb0ec3ccd0a5bec728d

docs/src/operate/c8y/connect-mqtt-service.md
didier-wenzek: Shouldn't this step, related to `c8y.tenant_id`, fully removed, including the note below?

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -226,6 +227,10 @@ define_tedge_config! {
         #[tedge_config(reader(private))]
         url: ConnectUrl,
 
+        /// Cumulocity tenant ID
+        #[tedge_config(example = "t12345678")]
+        tenant_id: String,
+
```
didier-wenzek: It the `tenant_id` still useful?

didier-wenzek (commented):
The feature is working as described. However, the `tenant_id` has to be removed from the config - as being no more useful.

docs/src/operate/c8y/connect-mqtt-service.md
albinsuresh: Resolved by 584b438

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
albinsuresh: Unfortunately there isn't any server side API that we can rely on. So, for now I've just added the mqtt service bridge service health check validation in 7a827c8

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
albinsuresh: That was the first line that I had removed, but it accidentally got stashed away. Resolved again in 7a827c8

docs/src/operate/c8y/connect-mqtt-service.md
albinsuresh: Accidentally got stashed away. Resolved in 7a827c8

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
albinsuresh: Addressed in [7a827c8](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/7a827c8972eaa63314bc6c0217e5e4c837252e17) by checking the health status of the bridge service in C8Y.

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
albinsuresh: It is already working with mosquitto bridge as it sends separate subscription requests for each bridge rule. The same behaviour was replicated in built-in bridge as well with [584b438](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/584b43856836d724d072fa0b268568131f0313a1)

crates/extensions/tedge_mqtt_bridge/src/config.rs
```diff
@@ -40,6 +40,7 @@ pub struct BridgeConfig {
     local_to_remote: Vec<BridgeRule>,
     remote_to_local: Vec<BridgeRule>,
     bidirectional_topics: Vec<(Cow<'static, str>, Cow<'static, str>)>,
+    pub(crate) remote_subscribe_once: bool,
```
jarhodes314: Given mosquitto seemingly does this by default, I wonder whether we actually don't want a flag and just want to change the behaviour to always use subscribe per topic.

design/decisions/0007-c8y-mqtt-service-connect.md
```diff
@@ -0,0 +1,166 @@
+# Connect to C8Y MQTT service endpoint
+
+* Date: __2024-06-23__
```
rina23q: (trivial) Isn't it **2025**-06-23?

crates/core/tedge/src/cli/connect/c8y.rs
```diff
@@ -198,30 +205,24 @@ pub(crate) async fn check_device_status_c8y(
...
+            .await?;
+    }
+
+    let mut mqtt_service_bridge_healthy = false;
+    let mut core_mqtt_connected = false;
 
     let mut err = None;
     loop {
         match event_loop.poll().await {
-            Ok(Event::Incoming(Packet::SubAck(_))) => {
```
rina23q, line 214: Why the connection check logic needed to be changed? Surprised to see that the JWT token request/receive check is removed.

albinsuresh: I just moved it below, where it is done only after making sure that the bridge is healthy.

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
```diff
@@ -0,0 +1,57 @@
...
+    ${DEVICE_SN}=    Setup    connect=${False}
+    Execute Command    tedge config set c8y.mqtt_service.enabled true
+    Execute Command    tedge config set c8y.mqtt_service.topics 'sub/topic,demo/topic'
+    Execute Command    tedge connect c8y
+
+    External Identity Should Exist    ${DEVICE_SN}:device:main:service:mosquitto-c8y-mqtt-bridge    show_info=False
+    Device Should Have Fragment Values    status\=up
+
+    Execute Command    tedge mqtt pub c8y-mqtt/test/topic '"hello"'
+    # TODO: Validate message received on test/topic on C8Y
```
albinsuresh, line 21: Since August 7 release of MQTT service, device isolation is [enabled by default](https://cumulocity.com/docs/change-logs/#mqtt-service-0.9.6-device-isolation-preview) preventing another MQTT client from receiving the message published by this device. Since there isn't any another easy option to test the message passing between the device and the cloud, you may turn device isolation off as follows (using your tenant admin credentials):

```
curl --location --request PUT "https://<TENANT_DOMAIN>/features/mqtt-service.tenant.isolation/by-tenant" \
--header "Authorization: Basic <AUTHORIZATION>" \
--header "Accept: application/json" \
--header "Content-Type: application/json" \
--data-raw '{ "active":true }'
```

before connecting the test client to subscribe/publish to the device topics.

rina23q (approved):
Approved

crates/extensions/tedge_mqtt_bridge/src/config.rs
albinsuresh: Resolved by [c06940e](https://github.com/thin-edge/thin-edge.io/pull/3707/commits/c06940e570c1ce564c7fa8bc3b33cd193ef0c7b5)

design/decisions/0007-c8y-mqtt-service-connect.md
reubenmiller, line 1: The actual decision is missing from the decision document (considering that we've done the implementation as per Approach 1 (to support a smooth transition once the MQTT Core SmartREST topics are supported by the MQTT Service)).

design/decisions/0007-c8y-mqtt-service-connect.md
reubenmiller: I assume that this change is still being done?

tests/RobotFramework/tests/tedge_connect/tedge_connet_mqtt_service.robot
reubenmiller: Or using go-c8y-cli

```
c8y features tenants enable --key mqtt-service.tenant.isolation
```

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: Resolved by 8bc7e0b

design/decisions/0007-c8y-mqtt-service-connect.md
albinsuresh: Resolved by 8bc7e0b

PR #3709, opened 2025-06-26
feat: `tedge cert create-key` command

Bravo555:
## TODO
- [ ] handle all supported keys (EC 256/384, RSA 2048/3072/4096)
- [ ] allow selecting token, key label, key type and size using command line arguments
- [ ] show more information after key is created
- [ ] test failure modes
- [ ] cleanup

## Proposed changes

Implements a `tedge cert create-key` command that can be used to create a private key on a PKCS11 token without additional tools.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge-p11-server/Cargo.toml
```diff
@@ -19,6 +19,7 @@ cryptoki.workspace = true
 pem.workspace = true
 percent-encoding.workspace = true
 postcard.workspace = true
+rand = "0.9.1"
```
didier-wenzek: The workspace version can be updated to "0.9.1"  or even "0.10.0-rc.0 "

```suggestion
rand.workspace = true
```

crates/extensions/tedge-p11-server/Cargo.toml
```diff
@@ -41,3 +42,6 @@ tempfile.workspace = true
 
 [lints]
 workspace = true
+
+# [patch.pem-rfc7468]
+# base64ct = "=1.6"
```
didier-wenzek: Is this still useful?

docs/src/references/hsm-support.md
```diff
@@ -250,3 +250,73 @@ will then be used to select a token, but the key will be selected automatically,
...
+
+```sh command="tedge cert create-key -h" title="tedge cert create-key -h"
+Create a new keypair using a PKCS11 token
+
+Usage: tedge cert create-key [OPTIONS] --label <LABEL> --type <TYPE> [COMMAND]
+
+Commands:
+  c8y   
+  az    
+  aws
```
didier-wenzek, line 301: I don't understand the purpose of this sub-command. To create a key, one only needs to interact with the  HSM not the cloud. Am I missing something?

docs/src/references/hsm-support.md
```diff
@@ -250,3 +250,73 @@ will then be used to select a token, but the key will be selected automatically,
...
+    Public key:
+    -----BEGIN PUBLIC KEY-----
+    BJ2vhTSHOh6KBB5NHG3wWJBj7sFAfW/W0GAenEwTttIWk+7EMT2mFgonVw1U1OPK
+    0CjBq4TKf4EtuzV+UYorBE4=
+    -----END PUBLIC KEY-----
+    ```
+
+    The command also creates a CSR at `device.csr_path`, which should be used when requesting a new certificate.
+
+4. Change the `device.key_uri` setting to point to the new private key. This is most easily done by
```
didier-wenzek: Do we really have to set `device.key_uri` twice? First before creating a key and then once created?

This sounds as two independent settings conflated into one.

docs/src/references/hsm-support.md
```diff
@@ -250,3 +250,73 @@ will then be used to select a token, but the key will be selected automatically,
...
+
+Commands:
+  c8y   
+  az    
+  aws   
+  help  Print this message or the help of the given subcommand(s)
+
+Options:
+      --config-dir <CONFIG_DIR>  [env: TEDGE_CONFIG_DIR, default: /etc/tedge]
+      --label <LABEL>
```
didier-wenzek: What is this?

docs/src/references/hsm-support.md
```diff
@@ -250,3 +250,73 @@ will then be used to select a token, but the key will be selected automatically,
...
+
+    Example:
+
+    ```title="Previous URI"
+    pkcs11:model=SoftHSM%20v2;manufacturer=SoftHSM%20project;serial=83f9cf49039c051a;token=c8y%20token;id=%02;object=azure%20keypair;type=private
+    ```
+
+    ```title="New URI"
+    pkcs11:model=SoftHSM%20v2;manufacturer=SoftHSM%20project;serial=83f9cf49039c051a;token=c8y%20token;object=my-key
+    ```
```
didier-wenzek: I don't understand what I have to do here.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -107,19 +111,6 @@ Can use PKCS11 key to renew the public certificate
...
-    ...    tedge cert renew c8y
-    ...    error=Failed to connect to tedge-p11-server UNIX socket at '/run/tedge-p11-server/tedge-p11-server.sock'
-
-    Execute Command    systemctl start tedge-p11-server.socket
-
-    Execute Command    cmd=tedge config set c8y.device.key_uri pkcs11:object=nonexistent_key
-    Command Should Fail With
-    ...    tedge cert renew c8y
-    ...    error=PKCS #11 service failed: Failed to find a key
-    Execute Command    cmd=tedge config unset c8y.device.key_uri
```
didier-wenzek, line 121: Why have these checks been removed? They sound as still meaningful.

crates/core/tedge/src/cli/certificate/create_key.rs
```diff
@@ -0,0 +1,94 @@
...
+        // should probably verify the keys before using them
+        let cryptoki = tedge_p11_server::tedge_p11_service(self.cryptoki_config.clone())?;
+        let key = cryptoki.create_key(None, params)?;
+        let pubkey_pem = key.pem;
+
+        eprintln!("New keypair was successfully created.");
+        eprintln!("Public key:\n{pubkey_pem}\n");
+
+        eprintln!("Insert this as a new value of device.key_uri:");
+        eprintln!("{}", key.uri);
```
didier-wenzek: Is there a reason not to set `device.key_uri` on behalf of the user?

crates/extensions/tedge-p11-server/Cargo.toml
Bravo555: changed to use workspace `rand` in 20946af

crates/extensions/tedge-p11-server/Cargo.toml
Bravo555: removed in 20946af

docs/src/references/hsm-support.md
Bravo555: Analogous to `tedge cert create c8y/az/aws` which changes the path of the certificate (if default settings are changed), these subcommands just read `device.key_uri` from under `[c8y]`/`[az]`/`[aws]` tables. Will add help text to make it more clear.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -139,6 +130,60 @@ Can renew the certificate using different keypair
     ${stdout}=    Tedge Reconnect Should Succeed
     Should Contain    ${stdout}    The new certificate is now the active certificate
 
+Can create a private key on the PKCS11 token and download new cert from c8y
```
reubenmiller: The test description does not match the test steps. The test is only creating the private key and not actually using it.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -51,6 +56,43 @@ pub enum TEdgeCertCli {
...
+    /// After the key is generated, tedge config is updated to use the new key using
+    /// `device.key_uri` property. Depending on the selected cloud, we use `device.key_uri` setting
+    /// for that cloud, e.g. `create-key c8y` will write to `c8y.device.key_uri`.
+    ///
+    /// When multiple tokens are connected, if `device.key_uri` setting is present, the token
+    /// identified by this URI will be used. Otherwise, the first token returned by the system will
+    /// be used.
+    CreateKey {
+        /// Human readable description (CKA_LABEL attribute) for the key.
+        #[arg(long)]
```
reubenmiller: Setting a default would be useful here.


```suggestion
        #[arg(long, default_value = "tedge")]
```

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -51,6 +56,43 @@ pub enum TEdgeCertCli {
...
+        label: String,
+
+        /// The type of the key.
+        #[arg(long)]
```
reubenmiller, line 84: Setting a default which reflects the same default key type used in the `tedge cert create` command.


```suggestion
        #[arg(long, default_value = "ecdsa")]
```

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -51,6 +56,43 @@ pub enum TEdgeCertCli {
...
+        /// The curve (size) of the ECSA key. Should only be used with --type ecdsa.
+        #[arg(long, default_value = "p256", group = "key_params")]
+        curve: EcCurve,
+
+        // can't document subcommands here because one would have to document variants of the enum
+        // but this type is used in other places
+        #[clap(subcommand)]
+        cloud: Option<CloudArg>,
+
+        token: Option<String>,
```
reubenmiller, line 124: A doc string would be useful to provide more info about what the `[TOKEN]` positional argument refers to (feel free to change the wording a bit).

```suggestion
        /// PKCS#11 URI of the token where the keypair should be created
        token: Option<String>,
```

docs/src/references/hsm-support.md
```diff
@@ -250,3 +250,96 @@ will then be used to select a token, but the key will be selected automatically,
...
+```
+
+`tedge cert create-key` command generates a new keypair on the PKCS #11 token.
+
+1. Configure cryptoki in `module` or `socket` mode as described in previous sections.
+2. Run the `tedge cert create-key` command. You'll need to provide key type, size and label of the
+   key object.
+
+    ```sh
+    tedge cert create-key --type ecdsa --curve 256 --label my-key
```
reubenmiller: Typo.

```suggestion
    tedge cert create-key --type ecdsa --curve p256 --label my-key
```

crates/core/tedge/src/cli/certificate/create_key.rs
```diff
@@ -0,0 +1,222 @@
...
+                key = format!("profiles.{profile}.{key}");
+            }
+
+            if let Some(cloud) = cloud {
+                key = format!("{cloud}.{key}");
+            }
+
+            let key = key.parse().expect("should be valid WritableKeys");
+            let r = dto.try_update_str(&key, uri).map_err(|e| e.into());
+            eprintln!("Value of `{key}` was updated to point to the new key");
```
reubenmiller: Slight improvement to the wording

```suggestion
            eprintln!("The `{key}` configuration setting was updated with the newly created key's pkcs11 uri");
```

docs/src/references/hsm-support.md
```diff
@@ -250,3 +250,96 @@ will then be used to select a token, but the key will be selected automatically,
 key may be wrong if there are multiple to choose from. Also if the URI contain attributes that
 identify a key, but doesn't contain attributes that identify a token, still the first token will be
 selected, even if another token contains the intended key.
+
+## Key generation
```
reubenmiller, line 285: A general question, can this command be used without a HSM? It doesn't seem so as the help is more centered around the HSM and the `TOKEN` positional argument is mandatory (which doesn't make sense for a file based key).

```sh
# tedge cert create-key 
error: the following required arguments were not provided:
--label <LABEL>
--type <TYPE>

Usage: tedge cert create-key --label <LABEL> --type <TYPE> [TOKEN]

For more information, try '--help'.
```

docs/src/references/hsm-support.md
reubenmiller: This is way too easy to accidentally create too many keys. This will create two private key objects where both have the token=tedge attribute set

```sh
tedge cert create-key --type ecdsa --label tedge "pkcs11:token=tedge"
tedge cert create-key --type ecdsa --label tedge "pkcs11:token=tedge"
```

But if the user tries to prevent duplications by providing an `id` attribute, the `id` attribute is ignored.

```sh
tedge cert create-key --type ecdsa --label tedge 'pkcs11:token=tedge;id=other'
```

Is there a way we can prevent the user from accidentally doing this? e.g. can we check if the key_uri already exists or something before creating another key (similar to how the `tedge cert create` would fail again if you run it twice)?

docs/src/references/hsm-support.md
```diff
@@ -250,3 +250,96 @@ will then be used to select a token, but the key will be selected automatically,
...
+  az    
+  aws   
+  help  Print this message or the help of the given subcommand(s)
+
+Options:
+      --config-dir <CONFIG_DIR>
+          [env: TEDGE_CONFIG_DIR, default: /etc/tedge]
+
+      --label <LABEL>
+          Human readable description (CKA_LABEL attribute) for the key
```
reubenmiller, line 315: Would it also make sense to also allow the user to control the private key's `id` attribute (in addition to the label)? Would this make it easier for user to identify the unique object?

```
tedge cert create-key --label tedge --id '3'
```

crates/core/tedge/src/cli/certificate/create_key.rs
```diff
@@ -0,0 +1,222 @@
...
+                bits: self.bits.into(),
+            },
+            KeyType::Ecdsa => KeyTypeParams::Ec {
+                curve: self.curve.into(),
+            },
+        };
+
+        let cryptoki = tedge_p11_server::tedge_p11_service(self.cryptoki_config.clone())?;
+        let Some(token) = self.token.clone() else {
+            eprintln!("No token URL was provided for this operation; the available tokens are:");
```
reubenmiller, line 95: It seems that the token list may not be complete if a new token was added after the the tedge-p11-server was started (same problem that we previously had with the client keys not being found by the tedge-p11-server).

Could we somehow force refreshing the token list?

crates/core/tedge/src/cli/certificate/cli.rs
Bravo555: added in 5ea0610422

crates/core/tedge/src/cli/certificate/cli.rs
Bravo555: added in 5ea0610422

crates/core/tedge/src/cli/certificate/cli.rs
Bravo555: added in 5ea0610422

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: added in 5ea0610422

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: Indeed this will happen if softhsm2 is used as a library, and we should be able fix this by always reloading before attempting to create the key. I'll implement it.

docs/src/references/hsm-support.md
Bravo555: Indeed this command only operates on the HSM, so we can change the naming to something like `create-key-pkcs11` to make it clear.

docs/src/references/hsm-support.md
Bravo555: We could allow the user to set it, but it would allow the user to generate objects that are not unique and have the same id. Handling these object after could get tricky, particularly `p11tool` gets confused when operating on objects that have the same URI.

What we do now is generate a random id which we include in the URI, so the object is always unique.

docs/src/references/hsm-support.md
Bravo555: A random id is generated such that when running the command twice, with the same parameters, the returned URI will be different.

The `TOKEN` positional argument is a URI identifying the token, while `id` attribute is related to objects on the token, so it's not relevant and thus ignored.

We could allow user to provide their `id` attribute for the key via `--id` flag, but it's not going to prevent them from making duplicates, but will allow them to make duplicates.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
Bravo555: fixed in fe62dc7871

docs/src/references/hsm-support.md
Bravo555: fixed in 3d69946369

docs/src/references/hsm-support.md
Bravo555: The reliance on `device.key_uri` was removed and now to select the token one needs to provide the token URI as a positional argument to `tedge cert create-key`. The guide was updated accordingly.

docs/src/references/hsm-support.md
Bravo555: The reliance on device.key_uri was removed and now to select the token one needs to provide the token URI as a positional argument to tedge cert create-key. The guide was updated accordingly.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
Bravo555: reverted removal of checks in 52917d7462

docs/src/references/hsm-support.md
reubenmiller: So PKCS11 doesn't reject an object which is created with the same ID? I would of expected that the API would fail somewhere. Plus I would assume that the id would not be globally unique, but only unique within the label (or token).

I still think it is worthwhile allow power-users to set the id value, and then use an autogenerated id by default (though something a bit shorter than the current generation would be nice).

docs/src/references/hsm-support.md
reubenmiller: Though we're talking about creating a key for tedge to use, so it should check first if the device.key_uri in the tedge.toml exists, and only create a new key if nothing is found (and ideally use the value defined in the tedge.toml)

docs/src/references/hsm-support.md
reubenmiller: Though having a command to create a file based private key would be beneficial as it would allow users to integrate more easily into existing PKI services.

Ideally it would be nice to have one single command but given that there are a few pkcs11 only argments, and don't want users to accidentally create a file based key when they were expecting to create a HSM based key...so with that, I'm ok with calling it the following (though I'm not super thrilled about the name, but I don't have a better suggestion).

```sh
tedge cert create-key-pkcs11
```

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: fixed in cff302ce7

crates/core/tedge/src/cli/certificate/create_key.rs
```diff
@@ -0,0 +1,283 @@
...
+
+        // generate a keypair
+        // should probably verify the keys before using them
+        let key = cryptoki.create_key(&token, params)?;
+        let pubkey_pem = key.pem;
+        let uri = key.uri;
+
+        eprintln!("New keypair was successfully created.");
+        eprintln!("Key URI: {uri}");
+        eprintln!("Public key:\n{pubkey_pem}\n");
```
reubenmiller, line 121: I'm wondering if adding an option to save the public key to file would make more sense, as it would make it more compatible with other 3rd party tooling  (e.g. gnutls's certtool).

Something like `--out-public-key` (though happy with a better name)?

And the user could provide `--out-public-key -` to denote writing the output to stdout.

docs/src/references/hsm-support.md
didier-wenzek: About the name: `create-key-pkcs11`:

- Why `pkcs11`, while elsewhere we use the term `cryptoki`?
- Wouldn't `create-pkcs11-key` be better?
- Personally, I would suggest `create-hsm-key`, but this introduces yet another acronym.

docs/src/references/hsm-support.md
```diff
@@ -281,3 +281,96 @@ will then be used to select a token, but the key will be selected automatically,
...
+
+`tedge cert create-key` command generates a new keypair on the PKCS #11 token.
+
+1. Configure cryptoki in `module` or `socket` mode as described in previous sections.
+2. Run the `tedge cert create-key` command. You'll need to provide key type, size and label of the
+   key object.
+
+    ```sh
+    tedge cert create-key-pkcs11 --type ecdsa --curve p256 --label my-key
+    ```
```
didier-wenzek: Some parts of the text is using `create-key`, others `create-key-pkcs11`. This will have to be fixed, once the final term is agreed.

docs/src/references/hsm-support.md
```diff
@@ -281,3 +281,96 @@ will then be used to select a token, but the key will be selected automatically,
...
+
+    ```sh title="Output"
+    New keypair was successfully created.
+    Key URI: pkcs11:model=SoftHSM%20v2;manufacturer=SoftHSM%20project;serial=a30ed1ca6244fc5f;token=test-token;id=%51%05%87%75%6F%B7%28%EC%5E%5D%1F%B8%EB%CF%FD%96%B7%E4%28%B6;object=my-key
+    Public key:
+    -----BEGIN PUBLIC KEY-----
+    BEsjmiXDdko90IDdjlAb/bWyTf6kd6S+/KPlj2Yd3zjHZe54evLyHJ1e8dSDhpy7
+    2Tcml9ZcHWBHA+MM0NFAbaw=
+    -----END PUBLIC KEY-----
```
didier-wenzek, line 387: Is the public key written somewhere?

docs/src/references/hsm-support.md
reubenmiller: I don’t mind HSM as it is more user friendly (even with the additional acronym)

crates/core/tedge/src/cli/certificate/create_key.rs
```diff
@@ -0,0 +1,283 @@
...
+            .context("invalid id")?;
+
+        let cryptoki = tedge_p11_server::tedge_p11_service(self.cryptoki_config.clone())?;
+        let Some(token) = self.token.clone() else {
+            eprintln!("No token URL was provided for this operation; the available tokens are:");
+            let tokens = cryptoki.get_tokens_uris()?;
+            for token_uri in tokens {
+                eprintln!("{token_uri}");
+            }
+            return Ok(());
```
didier-wenzek: Shouldn't we return a non-zero exit code at this point as no key has been generated?

crates/core/tedge/src/cli/certificate/create_key.rs
```diff
@@ -0,0 +1,283 @@
...
+async fn save_key_uri_to_config(
+    config: TEdgeConfig,
+    cloud: Option<&Cloud>,
+    uri: &str,
+) -> anyhow::Result<()> {
+    config
+        .update_toml(&|dto, _reader| {
+            // XXX: can break if the keys ever change and having to use strings sucks
+            let profile = cloud.as_ref().and_then(|c| c.profile_name());
+            // XXX: this should be elsewhere!
```
didier-wenzek: I agree this code is fragile and coupled with unrelated aspects, but I would avoid such comments, that are not really helpful.

Even if not ideal, I would move to `crate::cli::common` the transformation of a key (in that case `"device.key_uri"`) into something related to a cloud (`"{cloud}.{key}"`) or a profile (`"profiles.{profile}.{key}"`) or both (`"{cloud}.profiles.{profile}.{key}"`)

crates/core/tedge/src/cli/certificate/create_key.rs
```diff
@@ -0,0 +1,283 @@
...
+            );
+            r
+        })
+        .await
+        .map_err(anyhow::Error::new)?;
+
+    Ok(())
+}
+
+/// Parses id provided as a sequence of hex digits without `0x` prefix, optionally separated by spaces.
```
didier-wenzek:
```suggestion
/// Parses id provided as a sequence of bytes encoded as pair of hex digits without `0x` prefix, optionally separated by spaces.
```

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -134,6 +193,48 @@ impl TedgeP11Service for Cryptoki {
...
+            .collect();
+
+        Ok(uris)
+    }
+
+    fn create_key(&self, uri: &str, params: CreateKeyParams) -> anyhow::Result<CreateKeyResponse> {
+        let session_params = SessionParams {
+            uri: Some(uri.to_string()),
+            // TODO: do we want to use client-provided (device.key_pin) PIN in create-key-pkcs11 (we can still use
+            // extract PIN from URI)? Options are:
```
didier-wenzek: Is this planned for this PR? If yes, what is the retained approach? The 1st (let the user passes the pin in the URI) and 4th (add a --pin flag that also updates the config) sound reasonable.

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: fixed by returning exit code 1 in 068fbade

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: applied in bcde3b375

docs/src/references/hsm-support.md
Bravo555: My bad, missed it. Renamed to `create-key-hsm` in 458dc110

docs/src/references/hsm-support.md
Bravo555: renamed to `create-key-hsm` in 458dc1100

> Why pkcs11, while elsewhere we use the term cryptoki

I've seen both terms used somewhat interchangably and haven't made an effort to only use one of them for consistency and also used both :/
Looking back, _PKCS #11_ is the name of the standard, and _Cryptoki_ is the name of the interface defined by the standard.

> Wouldn't create-pkcs11-key be better?

I view it as `create-key-hsm` being an HSM-specific variant of `create-key`, and if we add a generic `create-key` command in the future, the suffixed version fits better.
So decided to go with `create-key-hsm`.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: resolved in b1f68c1

The approach we decided to go with Reuben was to enable passing the PIN via a flag, but not writing it automatically to tedge-config, as it is a secret, so the user has to input it manually just that they're aware.

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: Added `--outfile-pubkey` flag in 67f9a6648

gnutls tools only calls it `--outfile`, added `-pubkey` suffix to be a bit more precise.

docs/src/references/hsm-support.md
Bravo555: Added `--outfile-pubkey` flag in 67f9a664

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: Added writing key_uri to config automatically in 62ba8f2

docs/src/references/hsm-support.md
Bravo555: this pointed to an outdated version of the guide, but was replaced and updated.

crates/core/tedge/src/cli/certificate/create_key.rs
Bravo555: modified in ed52d71411

Indeed the second comment lacks context. I've chosen to extract this snippet to a separate function and given current generic mapper configuration work (#3835), which is relevant for `device.key_uri`, so given it will have to probably be updated then, I've chosen to leave it as is for now and provide more context in the comments.

crates/extensions/tedge-p11-server/src/proxy/client.rs
```diff
@@ -206,6 +227,20 @@ impl TedgeP11Client {
...
+    pub fn create_key(
+        &self,
+        request: CreateKeyRequest,
+    ) -> anyhow::Result<crate::service::CreateKeyResponse> {
+        let request = Frame1::CreateKeyRequest(request);
+        let response = self.do_request(request)?;
+
+        let Frame1::CreateKeyResponse(pubkey) = response else {
+            bail!("protocol error: bad response, expected create_key, received: {response:?}");
+        };
```
didier-wenzek, line 239: One needs to properly handle the case where the server has not been updated and is not able to create a key. Nothing more can be done than failing the command, but the error message must be explicit and not user-puzzling as currently.

crates/extensions/tedge-p11-server/src/proxy/server.rs
```diff
@@ -128,6 +129,34 @@ impl TedgeP11Server {
             // ensures the PKCS11 library is loaded and ready to serve signing requests. In
             // practice, this only occurs with a client calls TedgeP11Client::with_ready_check.
             Frame1::Ping => Frame1::Pong,
+
+            Frame1::CreateKeyRequest(request) => {
```
didier-wenzek, line 133: This must explicitly be a new version.


```suggestion
            Frame2::CreateKeyRequest(request) => {
```

crates/extensions/tedge-p11-server/src/proxy/connection.rs
```diff
@@ -95,6 +97,10 @@ pub enum Frame1 {
     GetPublicKeyPemResponse(String),
     Ping,
     Pong,
+    CreateKeyRequest(CreateKeyRequest),
+    CreateKeyResponse(CreateKeyResponse),
+    GetTokensUrisRequest,
+    GetTokensUrisResponse(Vec<String>),
```
didier-wenzek, line 103: This must added into a new `Frame::Version1(Frame1)` case.

Bravo555:
Integration tests consistently get stuck at `Tests.Tedge Flows.Tedge Flows`, so PR can't be merged.
This should be fixed by #3846, so I'll try merging again after that PR is merged.

PR #3710, opened 2025-06-26
refactor: handle retain messages with dynamic subscriptions

jarhodes314:
## Proposed changes
Handle retain messages when dynamic subscribers subscribe to a topic with an existing subscription. With the original implementation, new subscribers to topics that are already subscribed to wouldn't receive retain messages, e.g.

retain message exists on a/b
client a connections with subscription to a/b
client b connects with no subscriptions
mqtt actor connects -> retain message is received and forwarded to client a
client b subscribes to a/b
the message won't be resent to the mqtt actor for forwarding since an existing subscription exists

With these changes, the subscription from client b will cause a connection to be created dynamically to retrieve the relevant retain messages (i.e. the ones not covered by a new subscription).

In the case of an existing subscription to `a/b` and a new subscription to `a/+`, both the existing client and the newly subscribed client will receive the retain message. I don't believe this is a huge issue, as spontaneous redelivery of retain messages can occur regardless in the case of connection flakiness, for example.

Still TODO:
- [x] Integrate these changes with the entity store retain message clearing

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/connection.rs
+++ b/crates/common/mqtt_channel/src/connection.rs
@@ -454,9 +454,13 @@ impl Connection {
         // one has first to explicitly send the last will message.
         if let Some(last_will) = last_will {
             let payload = Vec::from(last_will.payload_bytes());
-            let _ = mqtt_client
+            if mqtt_client
                 .publish(last_will.topic, last_will.qos, last_will.retain, payload)
-                .await;
+                .await
+                .is_ok()
+            {
+                pub_count.fetch_add(1, Ordering::SeqCst);
+            }
         }
 
         // At this point, `_disconnect_permit` is dropped
--- a/crates/common/mqtt_channel/src/tests.rs
+++ b/crates/common/mqtt_channel/src/tests.rs
@@ -615,3 +615,28 @@ async fn connections_from_cloned_configs_are_independent() -> Result<(), anyhow:
 
     Ok(())
 }
+
+#[tokio::test]
+async fn connection_can_be_closed_after_last_will_published() -> Result<(), anyhow::Error> {
+    // This test arose from an issue with dynamic subscriptions where
+    // subscriptions were shared between different MQTT channel instances
+    let broker = mqtt_tests::test_mqtt_broker();
+    let last_will_topic = uniquify!("last/will");
+
+    let mqtt_config = Config::default()
+        .with_port(broker.port)
+        .with_last_will_message(MqttMessage {
+            topic: Topic::new_unchecked(last_will_topic),
+            payload: "test".to_owned().into(),
+            qos: QoS::AtLeastOnce,
+            retain: true,
+        });
+
+    let con = Connection::new(&mqtt_config).await?;
+
+    timeout(Duration::from_secs(1), con.close())
+        .await
+        .expect("Connection did not close");
+
+    Ok(())
+}


crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -345,6 +365,30 @@ impl FromPeers {
                             client.unsubscribe_many(diff.unsubscribe).await.unwrap();
                         }
                     });
+                    if !tf.is_empty() {
```
albinsuresh, line 512: I'm wondering if this would meet the dynamic subscription requirements of the dynamic mapper, where in most cases when new pipelines are dynamically added, they are most likely going to subscribe to already subscribed topics like `te/+/+/+/+/m/temp/meta` and would be expecting the retained messages from those topics. But, since dynamic connections are established only when a new subscription is added or an existing one is upgraded, the newly added pipelines won't get the existing retained messages as far as I understand. This won't meet our requirement, right? I'm just wondering if blindly establishing a new connection per peer/client (pipeline) would be better here.

jarhodes314: > But, since dynamic connections are established only when a new subscription is added or an existing one is upgraded

No, the exact opposite. We establish a dynamic connection when we receive a subscription that _does_ already exist.

> I'm just wondering if blindly establishing a new connection would be better here.

If we did this blindly for all topics, all that would achieve is retain messages sent to newly subscribed topics would be received twice.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -139,9 +141,10 @@ impl MqttActorBuilder {
             tracing::info!(target: "MQTT sub", "{pattern}");
         }
 
-        let mqtt_config = self.mqtt_config.with_subscriptions(topic_filter);
+        let mqtt_config = self.mqtt_config.clone().with_subscriptions(topic_filter);
         MqttActor::new(
             mqtt_config,
+            self.mqtt_config,
```
didier-wenzek: The `base_config` used on the side to read retained messages cannot be a simple clone of the main MQTT connection config. Using the same session name would trigger session clashes. Also better to use a clean session, as only retained messages matter.

```
let base_config = mqtt_config.clone().with_no_session();
```

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -345,6 +365,30 @@ impl FromPeers {
...
+                        let dynamic_connection_config =
+                            self.base_config.clone().with_subscriptions(tf);
+                        let mut sender = tx_to_peers.clone();
+                        tokio::spawn(async move {
+                            let mut conn =
+                                mqtt_channel::Connection::new(&dynamic_connection_config)
+                                    .await
+                                    .unwrap();
+                            while let Ok(msg) =
+                                tokio::time::timeout(Duration::from_secs(10), conn.received.next())
```
didier-wenzek: 10 seconds seems like too much. All the retained messages will be published in a row on connect. That's okay though, as adding a new subscription is a rare event.

crates/extensions/tedge_mqtt_ext/src/lib.rs
albinsuresh: > No, the exact opposite. We establish a dynamic connection when we receive a subscription that does already exist.

Not sure what I was doing wrong then. When I tested the same by tweaking the `dynamic_subscribers_receive_retain_messages` test with the following code at the end:

```
client_0
        .send(PublishOrSubscribe::Subscribe(SubscriptionRequest {
            diff: SubscriptionDiff {
                subscribe: ["a/b".into()].into(),
                unsubscribe: [].into(),
            },
            client_id: client_id_1,
        }))
        .await
        .unwrap();
assert_eq!(timeout(client_0.recv()).await.unwrap(), msg);
```

the assertion validating if `client0` gets the same retained message again, timed out.

jarhodes314: That is the intention. A new client subscribing shouldn't redeliver the retain messages to an existing subscription?

albinsuresh: That's where I see a problem. For all the pipelines, the generic mapper is the only peer of the mqtt actor, right? While it is connected to the mqtt actor while being built, it would have declared a set of subscriptions based on all the pipelines that existed at that time. And when a new pipeline is dynamically registered, but with some topics that the previous pipelines had already subscribed, it should still get the retained messages on those previously registered topics as well, right? Since the new pipeline is not registering as a new peer/client, but issuing the subscription request via the same gen mapper peer which has already received the retained messages for the same topics, it wouldn't get those retained messages even though they are "new" to the new pipeline, right? That's my concern.

didier-wenzek: As that is not so easy to fix, an option is to let the `tedge_gen_mapper` maintains a connection to the MQTT actor *per* pipeline.

crates/extensions/tedge_mqtt_ext/src/lib.rs
didier-wenzek: I pushed a fix https://github.com/thin-edge/thin-edge.io/pull/3710/commits/95f654dd2e7a2bffa60916926c61f30598a0e0bc.

Without that fix, the test `dynamic_subscribers_receive_retain_messages` is failing when updated to match a more realistic case, i.e. where the main connection has a session name.

crates/extensions/tedge_mqtt_ext/src/tests.rs
```diff
@@ -249,6 +250,119 @@ async fn dynamic_subscriptions() {
...
+            diff: SubscriptionDiff {
+                subscribe: ["a/+".into()].into(),
+                unsubscribe: [].into(),
+            },
+            client_id: client_id_1,
+        }))
+        .await
+        .unwrap();
+
+    // client_1 should receive both "a/b" and "b/c" retain messages upon subscribing
```
didier-wenzek: The comment is not accurate.

```suggestion
    // client_1 should receive "a/b" retain message upon subscribing to "a/+"
```

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -36,62 +39,180 @@ use tedge_actors::Server;
...
+}
+
+#[derive(Clone)]
+pub struct DynSubscriptions {
+    inner: Arc<Mutex<DynSubscriptionsInner>>,
+}
+pub struct DynSubscriptionsInner {
+    init_topics: TopicFilter,
+    client_id: Option<ClientId>,
+}
```
didier-wenzek: Now that a `mut ref` to a `DynSubscriptionsInner` can be directly used as a `MessageSource` config, one can deprecate the use of `Arc<Mutex<_>>`.

```suggestion
pub struct DynSubscriptions {
    init_topics: TopicFilter,
    client_id: Option<ClientId>,
}
```

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -56,27 +67,94 @@ pub enum EntityStoreResponse {
...
-        mqtt_connector: Box<dyn MqttConnector>,
-        mqtt_actor: &mut impl MessageSink<MqttMessage>,
+        mqtt_actor: &mut M,
         entity_auto_register: bool,
-    ) -> Self {
-        let mqtt_publisher = LoggingSender::new("MqttPublisher".into(), mqtt_actor.get_sender());
+    ) -> Self
+    where
+        M: MessageSink<MqttRequest>
+            + for<'a> MessageSource<MqttMessage, &'a mut DynSubscriptionsInner>,
```
didier-wenzek: Thanks!

This `for<'a>` is what I was missing to use a `mut ref` to a `DynSubscriptionsInner` as a `MessageSource` config.

didier-wenzek (commented):
Nice improvement. I just have a proposal to simplify actor connections.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -56,27 +67,94 @@ pub enum EntityStoreResponse {
...
+
+struct DeregistrationActor {
+    messages: SimpleMessageBox<MqttMessage, NoMessage>,
+    mqtt_publish: LoggingSender<MqttMessage>,
+}
+
+#[async_trait::async_trait]
+impl Actor for DeregistrationActor {
+    fn name(&self) -> &str {
+        todo!()
```
albinsuresh:
```suggestion
        "EntityDeregistrationActor"
```

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -321,6 +399,10 @@ impl EntityStoreServer {
             return deleted;
         }
 
+        if deleted.is_empty() {
+            return vec![];
+        }
```
albinsuresh: Redundant `if` block? I believe you just wanted to update the previous one by explicitly returning `return vec![]` rather than `return deleted`.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -340,33 +422,7 @@ impl EntityStoreServer {
...
-                    }
-                }
-
-                connection.disconnect().await;
-            }
-            Err(err) => {
-                error!("Failed to create MQTT connection for clearing entity data: {err}");
-            }
-        }
+        self.retain_requests.send(topics).await.unwrap();
```
albinsuresh: Though I'm not too worried about the order in which the entity data is cleared, I'm slightly concerned about this happening asynchronously, which doesn't guarantee that all data messages would be cleared before the deregistration request completes/returns. This is one guarantee that I'd still like to give to the users that everything is cleared before the `DELETE` call returns. I'm not too bothered about that request taking a bit more time, as it it is such rare operation.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -208,11 +338,20 @@ impl TrieService {
...
+#[cfg(feature = "test-helpers")]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ClientId(pub usize);
+#[cfg(not(feature = "test-helpers"))]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct ClientId(usize);
 
 type MatchRequest = String;
 
+struct InsertRequest {
```
albinsuresh:
```suggestion
struct ConnectRequest {
```

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -188,19 +204,62 @@ async fn delete_entity_tree_clears_entities_bottom_up() {
...
+    mqtt_box
+        .assert_received([MqttRequest::RetrieveRetain(ClientId(0), topic_filter)])
+        .await;
     mqtt_box
         .assert_received([
-            MqttMessage::from(("te/device/child000/service/service0", "")).with_retain(),
-            MqttMessage::from(("te/device/child000//", "")).with_retain(),
-            MqttMessage::from(("te/device/child00//", "")).with_retain(),
-            MqttMessage::from(("te/device/child0//", "")).with_retain(),
+            MqttRequest::Publish(
```
albinsuresh: We can avoid this explicit wrapping with a `impl From<MqttMessage> for MqttRequest`, right?

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -36,62 +39,180 @@ use tedge_actors::Server;
...
+    pub fn client_id(&self) -> ClientId {
+        self.inner.lock().unwrap().client_id.unwrap()
+    }
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum MqttRequest {
     Publish(MqttMessage),
     Subscribe(SubscriptionRequest),
+    RetrieveRetain(ClientId, TopicFilter),
```
albinsuresh: I was expecting a slightly different dynamic connection API, where the clients would dynamically create a `Sender<MqttMessage>` and include it in the dynamic connection request, which this actor would use to send the messages from that dynamic connection. This design is also fine and functional. I just found the peer connection logic a but unintuitive. I believe it's done this way to ease unit testability, as I've had to do something similar when I introduced the `MqttConnection` as well.

crates/extensions/tedge_mqtt_ext/src/lib.rs
albinsuresh: With the dynamic mapper more likely to use the newly added dynamic **connection** requests, I'm just wondering who/when the dynamic **subscription** requests would be used.

albinsuresh (commented):
Things look fine from a functionality perspective, except for one remark on the async nature of clearing entity data (which we can discuss). But, it feels like the connection logic with the MQTT actor has gotten a lot more complicated now. Unfortunately, I don't have any concrete suggestions yet on how to improve it.

crates/core/tedge_agent/src/entity_manager/server.rs
jarhodes314: Now deleted this entirely

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: Now done

crates/core/tedge_agent/src/entity_manager/tests.rs
jarhodes314: I'm not sure whether I want such an implementation to exist, but instead I've modified the assertion logic to allow comparisons between different types, and added `impl PartialEq<MqttMessage> for MqttRequest`

albinsuresh: > I've modified the assertion logic to allow comparisons between different types, and added impl PartialEq<MqttMessage> for MqttRequest

That works too. Suggested the explicit `From` impl since the `fan_in_message_type!` macro does the same as well.

> I'm not sure whether I want such an implementation to exist

Are you trying to highlight something wrong or non-idiomatic with having such impls or just trying to say that adapting the equality check is just better?

crates/core/tedge_actors/src/test_helpers.rs
```diff
@@ -182,7 +183,7 @@ pub trait MessageReceiverExt<M: Message>: Sized {
     async fn assert_received_unordered<Samples>(&mut self, expected: Samples)
     where
         Samples: IntoIterator + Send,
-        M: From<Samples::Item>;
+        M: From<Samples::Item> + PartialEq;
```
albinsuresh: Wondering why the generic signature of this function is not consistent with that of the updated `assert_received`(with the `Item` parameter added). To be done in a follow-up commit or it was a conscious omission?

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -407,6 +417,7 @@ impl FromPeers {
         rx_to_peers: &mut mpsc::UnboundedReceiver<MqttRequest>,
     ) -> Result<(), RuntimeError> {
         while let Ok(Some(message)) = self.try_recv(rx_to_peers).await {
+            let message = message;
```
albinsuresh: Relics of some debugging?

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -56,27 +62,38 @@ pub enum EntityStoreResponse {
...
+    ) -> Self
+    where
+        M: MessageSink<MqttRequest>,
+    {
+        let mqtt_publisher = LoggingSender::new(
+            "MqttPublisher".into(),
+            Box::new(MappingSender::new(mqtt_actor.get_sender(), |msg| {
+                [MqttRequest::Publish(msg)]
+            })),
+        );
```
albinsuresh: Just an additional `sender_clone()` would have been sufficient, provided there was an `impl From<MqttMessage> for MqttRequest`:
```suggestion
        let mqtt_publisher = LoggingSender::new(
            "MqttPublisher".into(),
            mqtt_actor.get_sender().sender_clone(),
        );
```

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -340,31 +357,13 @@ impl EntityStoreServer {
...
-                            error!(
-                                    "Failed to clear retained message on topic {} while de-registering {} due to: {err}",
-                                    topic_id,
-                                    message.topic
-                                );
-                        }
-                    }
-                }
+        let (tx, mut rx) = mpsc::unbounded();
+        self.retain_requests.send((tx, topics)).await.unwrap();
```
albinsuresh, line 355: Seeing that unwrap, I'm also a bit confused on what the right strategy on error would be. Inability to send something over a channel sounds like a fatal issue for which stopping (hoping of a restart) seems like the best thing to do. Logging the error and proceeding (as done in the earlier code) doesn't seem right either. But panicking also feels like an extreme. Ideally this actor should have returned a `RuntimeError` so that the actor runtime gets a chance for a clean shutdown of the other actors. Since updating the error propogation strategy of this actor is way out of of the scope of this ticket, I guess this is fine for now.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -130,19 +124,38 @@ async fn delete_entity_clears_retained_data() {
             true,
         ),
     ] {
-        mqtt_sender
-            .send(MqttMessage::from((topic, &payload.to_string())).with_retain_flag(retain))
+        mqtt_input
+            .send(MqttRequest::Publish(
```
albinsuresh: Another explicit wrapping that we could avoid with an `impl From<MqttMessage> for MqttRequest`, by relying on the `send()` doing that conversion implicitly.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -432,71 +488,81 @@ mod entity {
...
-
-    async fn disconnect(self: Box<Self>) {
-        // Do nothing
+    impl MessageSink<MqttRequest> for TestMqttActorBuilder {
+        fn get_sender(&self) -> tedge_actors::DynSender<MqttRequest> {
+            let mut cached_sender = self.sender.lock().unwrap();
+            if let Some(sender) = &*cached_sender {
+                return sender.sender_clone();
+            }
+            let retain_messages: Mutex<Vec<MqttMessage>> = <_>::default();
```
albinsuresh, line 473: That was smart, internally caching the retained messages at the sender level.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -481,14 +494,16 @@ impl FromPeers {
...
-    fn forward_retain_messages_to(
+    fn forward_retain_messages_to<Packet: Send + 'static>(
         &self,
-        mut sender: mpsc::UnboundedSender<(ClientId, MqttMessage)>,
+        mut sender: mpsc::UnboundedSender<Packet>,
         topics: TopicFilter,
-        client_id: ClientId,
+        prepare_msg: impl (Fn(MqttMessage) -> Packet) + Send + 'static,
     ) {
+        println!("TESTOING");
```
albinsuresh: To be removed.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -481,14 +494,16 @@ impl FromPeers {
...
         let dynamic_connection_config = self.base_config.clone().with_subscriptions(topics);
         tokio::spawn(async move {
+            println!("TESTOING");
```
albinsuresh: To be removed.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -525,7 +523,45 @@ mod entity {
...
+            let sender =
+                Box::new(MappingSender::new(
+                    self.messages.get_sender(),
+                    move |req| match req {
+                        MqttRequest::Publish(msg) if msg.retain => {
+                            dbg!(&msg);
```
albinsuresh: To be removed?

albinsuresh (commented):
LGTM. Happy to approve after some of the debugging code is cleaned up.

crates/core/tedge_agent/src/entity_manager/server.rs
```diff
@@ -56,27 +62,38 @@ pub enum EntityStoreResponse {
 pub struct EntityStoreServer {
     entity_store: EntityStore,
     mqtt_schema: MqttSchema,
-    mqtt_connector: Box<dyn MqttConnector>,
     mqtt_publisher: LoggingSender<MqttMessage>,
     entity_auto_register: bool,
+    retain_requests: SimpleMessageBox<NoMessage, (mpsc::UnboundedSender<MqttMessage>, TopicFilter)>,
```
didier-wenzek: Better to make explicit that `retain_requests` is only used to send messages.

```suggestion
    retain_requests: LoggingSender<(mpsc::UnboundedSender<MqttMessage>, TopicFilter)>,
```

crates/core/tedge_agent/src/entity_manager/tests.rs
jarhodes314: Ah, I've realised the original reason why I threw out the `From` implementation is that it doesn't work for the `assert_received` implementation, as that tries to convert before comparing. Since `MqttRequest::RetrieveRetain` now takes a channel, we can't reasonably implement `PartialEq` for `MqttRequest`, so we have to compare `MqttRequest` to `MqttMessage` using `PartialEq`.

crates/core/tedge_agent/src/entity_manager/server.rs
jarhodes314: The only case in which this should be returned is if the channel is closed (which means the MQTT actor has shut down) at the point at which we attempt to request retain messages. If that's happened, we've probably already received a shutdown request.

crates/core/tedge_agent/src/entity_manager/tests.rs
jarhodes314: Now done

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: Now done

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: Now done

crates/core/tedge_agent/src/entity_manager/tests.rs
jarhodes314: Now done

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: Yes, probably

crates/core/tedge_actors/src/test_helpers.rs
jarhodes314: I think I didn't quite manage it because I no longer needed the `unordered` comparison, but I've now updated it to match `assert_received`

crates/core/tedge_agent/src/entity_manager/server.rs
jarhodes314: For some reason, `sender_clone` doesn't work here, but a second call to `get_sender` does, so I've simplified this to that.

crates/core/tedge_agent/src/entity_manager/server.rs
jarhodes314: Now done

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: I've instead renamed it to `TrieInsertRequest`, since that explains more clearly what this actually represents

crates/core/tedge_agent/src/entity_manager/server.rs
jarhodes314: Now addressed

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: I think the mapper should be using both (if we change the input topic for a mapping rule, for instance, that will cause a change in the subscription, but shouldn't need a new MQTT actor connection).

Also, the "connection" part has nothing to do with MQTT connections and is merely a dynamic connection between two actors.

crates/core/tedge_agent/src/entity_manager/server.rs
jarhodes314: Now done

crates/core/tedge_agent/src/entity_manager/tests.rs
jarhodes314: I've also now implemented the From so it can be used for sending.

crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -120,6 +159,9 @@ impl TryFrom<MqttRequest> for MqttMessage {
 pub enum MqttRequest {
     Publish(MqttMessage),
     Subscribe(SubscriptionRequest),
+    /// A one-shot request for all the retain messages for a set of topics
+    ///
+    /// The provided sender is used to
```
albinsuresh:
```suggestion
    /// The provided sender is used to send those retained messages back to the requesting peer
```

albinsuresh (approved):
LGTM.

crates/core/tedge_agent/src/entity_manager/server.rs
albinsuresh: Hmm... Weird. I had tested that change before submitting that suggestion, but when I just tested it again with your latest changes, it doesn't compile. I may have changed something else as well yesterday. Anyway, this is also good enough.

Issue #3713, opened 2025-06-27
tedge-mapper-c8y built-in bridge stopped sending mqtt messages to Cumulocity


reubenmiller:
**Describe the bug**

A incident was detected on the OSADL QA farm device, osadl-rackfslot2, where the tedge-mapper-c8y using the built-in bridge stopped sending data back to Cumulocity.

The device could still receive operations from Cumulocity (so the device could be "rescued" by restarting the tedge-mapper-c8y service), however the operation status and metrics stopped being sent back to Cumulocity.

**Symptoms**

* Operation status is not updated, but operations were being processed (e.g. could get tedge-configuration-plugin and even connect using remote access)
* Cumulocity Availability status displayed as RED/GREEN.
* Alarm in Cumulocity, "No data received from device within required interval" since May 28, 2025.May 28, 2025, 10:03:35 AM CEST, but since the required interval is set to 60mins, it's likely that the connected stopped up to 60mins before the alarm date.
* tedge-mapper-c8y log entries shows that there are waiting messages that don't seem to change (e.g. "waiting=15")
* A remote access session (via SSH) could be established successfully
* Restarting the "tedge-mapper-c8y" service resumed "normal" function

**Log Snippet**

Below shows a log snippet taken before the "tedge-mapper-c8y" service was restarted

```log
Jun 27 06:16:10 rackfslot1s tedge-mapper[2973910]: 2025-06-27T04:16:10.342187594Z DEBUG tedge_mqtt_bridge: crates/extensions/tedge_mqtt_bridge/src/lib.rs:475: Bridge cloud connection: received=16 forwarded=16 published=51880 waiting=15 acknowledged=51857 finalized=51857
Jun 27 06:16:10 rackfslot1s tedge-mapper[2973910]: 2025-06-27T04:16:10.353437447Z DEBUG tedge_mqtt_bridge: crates/extensions/tedge_mqtt_bridge/src/lib.rs:474: Received notification (cloud) Incoming(Publish(Topic = s/dat, Qos = AtLeastOnce, Retain = false, Pkid = 1, Payload Size = 810))
Jun 27 06:16:10 rackfslot1s tedge-mapper[2973910]: 2025-06-27T04:16:10.353505261Z DEBUG tedge_mqtt_bridge: crates/extensions/tedge_mqtt_bridge/src/lib.rs:475: Bridge cloud connection: received=16 forwarded=16 published=51881 waiting=15 acknowledged=51857 finalized=51857
[...]

PR #3717, opened 2025-06-30
fix: manually republish messages on built-in bridge reconnect

jarhodes314:
## Proposed changes
Manually republish messages on bridge reconnect to simulate non-clean sessions for Cumulocity.

Since v0.24, rumqttc has fixed how it handles reconnection to a broker, only republishing the existing packets when a session is present (i.e. we have connected to a broker with `clean_session=false` and the broker has acknowledged this in the connack packet). For Cumulocity, we ideally want a non-clean (i.e. persistent) session, but this is explicitly recommended against.

Still TODO:
- [ ] Test this in a reproducible way, with a broker we can control the responses of, so we can ensure that the messages get published exactly once

## Types of changes

## Paste Link to the issue
- #3713 

## Checklist


crates/extensions/tedge_mqtt_bridge/tests/test_broker/mod.rs
jarhodes314, line 1: This file is a `mod.rs` so it doesn't get compiled into its own test binary. It will be compiled as a submodule of `republish.rs` so the tests in this file will run as part of this.

The first draft of this code was created by gemini, hence the heavy commenting. I intend to remove some of the less useful ones, although the details of what the broker should be expecting are broadly useful.

crates/extensions/tedge_mqtt_bridge/Cargo.toml
```diff
@@ -21,12 +21,14 @@ bytes = { workspace = true }
 certificate = { workspace = true }
 futures = { workspace = true }
 mqtt_channel = { workspace = true }
+mqttbytes = { workspace = true }
```
didier-wenzek: This is a dev dependency.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -468,6 +472,11 @@ async fn half_bridge(
                     info!("Waiting {time:?} until attempting reconnection to {name} broker");
                 }
                 tokio::time::sleep(time).await;
+                if session_present.map_or(true, |session_present| !session_present) {
+                    let msgs = recv_event_loop.take_pending();
+                    debug!("Extending pending with: {msgs:?}");
+                    pending.extend(msgs);
+                }
```
didier-wenzek: I'm reading this as "if the session is *not* managed by the connection, then we have to republish ourself the pending messages". Am I correct?

I think it would be good to add a comment here.

crates/extensions/tedge_mqtt_bridge/Cargo.toml
```diff
@@ -21,12 +21,14 @@ bytes = { workspace = true }
...
 mutants = { workspace = true }
 rumqttc = { workspace = true, features = ["proxy"] }
 tedge_actors = { workspace = true }
 tedge_config = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, default-features = false, features = ["macros"] }
+tokio-util = { workspace = true }
```
didier-wenzek: This is also a dev dependency.

didier-wenzek (commented):
I really like these new tests driven by small precise actions on the local and remote brokers.

The code of the `test_broker` used underneath is a bit weird but that's okay as the method to control its behavior from the test suite are clear enough.

I will be happy to approve this PR once the clippy warnings fixed (or simply marked as ignored, since they are related to the `test_broker`).

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: That understanding is correct. I will add a comment.

Since `pending` shouldn't actually be written to by rumqttc until we have established a connection, in the `session_present == None` case, this shouldn't actually matter. Rewriting the logic as `if session_present == Some(false)` would be clearer as to the condition under which (I believe) this is satisfied, but I'm not sure it's any clearer as to why it's correct.

didier-wenzek: `if session_present == Some(false)` might be a bit simpler, but a comment would help in any case.

jarhodes314: I have now added a comment, and went for `session_present != Some(true)`.

crates/extensions/tedge_mqtt_bridge/Cargo.toml
jarhodes314: Yep, initially I couldn't remember how to have submodules of integration tests so I just had the test broker as part of the crate. Fixed now

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -453,6 +454,9 @@ async fn half_bridge(
     let mut published = 0; // Count of messages published (by the companion)
     let mut acknowledged = 0; // Count of messages acknowledged (by the MQTT end-point of the companion)
 
+    let mut session_present: Option<bool> = None;
```
albinsuresh, line 461: I was wondering why this is not initiated by the `clean_session` value used in the mqtt config used to establish the connection. Got the answer from the code below that it is decided based on the broker response, since some brokers like C8Y can ignore the client's request for a persistent session. A comment would be helpful for future readers.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -479,13 +493,24 @@ async fn half_bridge(
...
         match notification {
-            Event::Incoming(Incoming::ConnAck(_)) => {
+            Event::Incoming(Incoming::ConnAck(conn_ack)) => {
                 info!("Bridge {name} connection subscribing to {topics:?}");
-                let recv_client = recv_client.clone();
-                let topics = topics.clone();
-                // We have to subscribe to this asynchronously (i.e. in a task) since we might at
-                // this point have filled our cloud event loop with outgoing messages
-                tokio::spawn(async move { recv_client.subscribe_many(topics).await.unwrap() });
+                {
```
albinsuresh: Why was this code moved into a sub-block? To drop the `recv_client` handle as quickly as possible? Any subtle impact on the fix with this change?

jarhodes314: Ah, I think that's a relic of the initial implementation where I attempted to resend the messages manually and needed multiple clones of the client and topics, so I put this into a block to allow me to reuse names.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: Now added a comment

Issue #3693, opened 2025-06-17
Add ability to create missing intermediate paths when installing configuration files


evanorsouw:
**Is your feature improvement request related to a problem? Please describe.**

My use-case was the following;
* extend `tedge-configuration-plugin.toml` with certificate of local local self-signed container registry.
```
[[files]]
type = "harbor-certificate"
path = "/etc/containers/certs.d/myregistry/ca.crt"
```
* with `podman` installed, only the following path exists; `/etc/containers/certs.d`
* problem now is that path component `myregistry` is application specific hence not known upfront.  
Therefor the deployment would fail with a file (path) not found exception.

**Describe the solution you'd like**
Proposed solution is to add the option `makedirs` to `tedge-configuration-plugin.toml`;

```
[[files]]
type = "harbor-certificate"
path = "/etc/containers/certs.d/<hostname>/ca.crt"
makedirs = true
```

reubenmiller:
@evanorsouw Though this goes beyond just creating the directory structure, as it would be also doing some variable expansion within the given `path` property, e.g. evaluating the `<hostname>` with the current device's hostname, or did I misunderstand something?

evanorsouw:
@reubenmiller no, that is not the intention. I used `<hostname>` just as a placeholder for something realname, `path` is still an unambiguous absolute path.

reubenmiller:
@evanorsouw  Though I would propose that the default behaviour should be to always create the parent directory hierarchy if required.

For instance, with upcoming change in the configuration (https://github.com/thin-edge/thin-edge.io/pull/3685), the configuration can be applied before the required directory structure has been created by the application. Leaving the users to specify the "makedirs" flag in the configuration file would be easy to forget.

The only disadvantage I can think of would be that it could lead to creating configuration files that aren't read by anything, however this wouldn't impact the users experience that much (imo). Or do you have another viewpoint why automatically creating the folder structure won't be desired?

reubenmiller:
I did a quick test and by using the `std::fs::create_dir_all` before creating the files would create a directory structure which is owned by root (as it is run under elevated priviliedges using `tedge-write`), but the file (leaf) gets the expected permissions.

In some instances this may be problematic if an application requires write access to folders, however I would suspect that the correct folder permissions would be updated by the application itself at install time.

Below shows an example of the created directory structure for "/etc/containers/certs.d/example/ca.crt":

```sh
# namei -l /etc/containers/certs.d/example/ca.crt
[...]

PR #3718, opened 2025-07-02
feat: create parent directories if missing for config_update operation

rina23q:
## Proposed changes
Previously, the `config_update` operation would fail if any parent directories of the target file were missing.

This change ensures that missing parent directories are automatically created when the operation is executed.

The mode, user, and group of the immediate parent directory can be configured via `tedge-configuration-plugin.toml` using the following keys:
- `parent_mode`
- `parent_user`
- `parent_group`

If `parent_user` and `parent_group` are not explicitly specified, but the target file's ownership is provided, the parent directory will inherit the file's user and group.

## Types of changes

## Paste Link to the issue
#3693 

## Checklist

### DIFF
--- a/crates/extensions/tedge_config_manager/src/actor.rs
+++ b/crates/extensions/tedge_config_manager/src/actor.rs
@@ -2,6 +2,7 @@ use anyhow::Context;
 use async_trait::async_trait;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
+use log::debug;
 use log::error;
 use log::info;
 use serde_json::json;
@@ -33,7 +34,9 @@ use tedge_uploader_ext::UploadRequest;
 use tedge_uploader_ext::UploadResult;
 use tedge_utils::atomic::MaybePermissions;
 use tedge_write::CopyOptions;
+use tedge_write::CreateDirsOptions;
 
+use crate::FileEntry;
 use crate::TedgeWriteStatus;
 
 use super::config::PluginConfig;
@@ -355,11 +358,57 @@ impl ConfigManagerWorker {
         let from_path = Utf8Path::from_path(&from)
             .with_context(|| format!("path is not utf-8: '{}'", from.to_string_lossy()))?;
 
+        let file_entry = self
+            .plugin_config
+            .get_file_entry_from_type(&request.config_type)?;
+        let to = Utf8PathBuf::from(&file_entry.path);
+
+        if let Some(parent) = to.parent() {
+            if !parent.exists() {
+                self.create_parent_dirs(parent, file_entry)?;
+            }
+        }
+
         let deployed_to_path = self
-            .deploy_config_file(from_path, &request.config_type)
+            .deploy_config_file(from_path, file_entry)
             .context("failed to deploy configuration file")?;
 
         Ok(deployed_to_path)
+    }
+
+    /// Creates the parent directories of the target file if they are missing,
+    /// and applies the permissions and ownership that are specified.
+    /// First, if `use_tedge_write` is enabled, it tries to use tedge-write to create the missing parent directories.
+    /// If it's disabled or creation with elevated privileges fails, fall back to the current user.
+    fn create_parent_dirs(&self, parent: &Utf8Path, file_entry: &FileEntry) -> anyhow::Result<()> {
+        if let TedgeWriteStatus::Enabled { sudo } = self.config.use_tedge_write.clone() {
+            debug!("Creating the missing parent directories with elevation at '{parent}'");
+            let result = CreateDirsOptions {
+                dir_path: parent,
+                sudo,
+                mode: file_entry.parent_permissions.mode,
+                user: file_entry.parent_permissions.user.as_deref(),
+                group: file_entry.parent_permissions.group.as_deref(),
+            }
+            .create();
+
+            match result {
+                Ok(()) => return Ok(()),
+                Err(err) => {
+                    info!("Failed to create the missing parent directories with elevation at '{parent}' with error: {err}. \
+            Falling back to the current user to create the directories.");
+                }
+            }
+        }
+
+        debug!("Creating the missing parent directories without elevation at '{parent}'");
+        std::fs::create_dir_all(parent)
+            .with_context(|| format!("failed to create parent directories. Path: '{parent}'"))?;
+
+        file_entry.parent_permissions.clone().apply_sync(parent.as_std_path())
+            .with_context(|| format!("failed to change permissions or mode of the parent directory. Path: '{parent}'"))?;
+
+        Ok(())
     }
 
     /// Deploys the new version of the configuration file and returns the path under which it was
@@ -377,35 +426,10 @@ impl ConfigManagerWorker {
     fn deploy_config_file(
         &self,
         from: &Utf8Path,
-        config_type: &str,
+        file_entry: &FileEntry,
     ) -> anyhow::Result<Utf8PathBuf> {
-        let file_entry = self.plugin_config.get_file_entry_from_type(config_type)?;
-
         let to = Utf8PathBuf::from(&file_entry.path);
-
-        let permissions = MaybePermissions {
-            uid: file_entry
-                .file_permissions
-                .user
-                .as_ref()
-                .map(|u| {
-                    uzers::get_user_by_name(&u).with_context(|| format!("no such user: '{u}'"))
-                })
-                .transpose()?
-                .map(|u| u.uid()),
-
-            gid: file_entry
-                .file_permissions
-                .group
-                .as_ref()
-                .map(|g| {
-                    uzers::get_group_by_name(&g).with_context(|| format!("no such group: '{g}'"))
-                })
-                .transpose()?
-                .map(|g| g.gid()),
-
-            mode: file_entry.file_permissions.mode,
-        };
+        let permissions = MaybePermissions::try_from(&file_entry.file_permissions)?;
 
         let src = std::fs::File::open(from)
             .with_context(|| format!("failed to open source temporary file '{from}'"))?;
--- a/crates/extensions/tedge_config_manager/src/config.rs
+++ b/crates/extensions/tedge_config_manager/src/config.rs
@@ -113,6 +113,9 @@ pub struct RawFileEntry {
     user: Option<String>,
     group: Option<String>,
     mode: Option<u32>,
+    parent_user: Option<String>,
+    parent_group: Option<String>,
+    parent_mode: Option<u32>,
 }
 
 #[derive(Debug, Eq, PartialEq, Default, Clone)]
@@ -125,6 +128,7 @@ pub struct FileEntry {
     pub path: String,
     pub config_type: String,
     pub file_permissions: PermissionEntry,
+    pub parent_permissions: PermissionEntry,
 }
 
 impl Hash for FileEntry {
@@ -147,16 +151,27 @@ impl Borrow<str> for FileEntry {
 
 impl FileEntry {
     pub fn new(
-        path: String,
-        config_type: String,
-        user: Option<String>,
-        group: Option<String>,
-        mode: Option<u32>,
+        path: impl Into<String>,
+        config_type: impl Into<String>,
+        file_permissions: PermissionEntry,
+        parent_permissions: PermissionEntry,
     ) -> Self {
+        let parent_user = parent_permissions
+            .user
+            .or_else(|| file_permissions.user.clone());
+        let parent_group = parent_permissions
+            .group
+            .or_else(|| file_permissions.group.clone());
+
         Self {
-            path,
-            config_type,
-            file_permissions: PermissionEntry { user, group, mode },
+            path: path.into(),
+            config_type: config_type.into(),
+            file_permissions,
+            parent_permissions: PermissionEntry {
+                user: parent_user,
+                group: parent_group,
+                mode: parent_permissions.mode,
+            },
         }
     }
 }
@@ -198,10 +213,9 @@ impl PluginConfig {
     fn new_with_config_file_entry(config_file_path: &Path) -> Self {
         let file_entry = FileEntry::new(
             config_file_path.display().to_string(),
-            DEFAULT_PLUGIN_CONFIG_TYPE.into(),
-            None,
-            None,
-            None,
+            DEFAULT_PLUGIN_CONFIG_TYPE,
+            PermissionEntry::default(),
+            PermissionEntry::default(),
         );
         Self {
             files: HashSet::from([file_entry]),
@@ -226,9 +240,12 @@ impl PluginConfig {
             let entry = FileEntry::new(
                 raw_entry.path,
                 config_type.clone(),
-                raw_entry.user,
-                raw_entry.group,
-                raw_entry.mode,
+                PermissionEntry::new(raw_entry.user, raw_entry.group, raw_entry.mode),
+                PermissionEntry::new(
+                    raw_entry.parent_user,
+                    raw_entry.parent_group,
+                    raw_entry.parent_mode,
+                ),
             );
 
             if !self.files.insert(entry) {
--- a/docs/src/operate/c8y/configuration-management.md
+++ b/docs/src/operate/c8y/configuration-management.md
@@ -26,13 +26,25 @@ files = [
   { path = '/etc/tedge/mosquitto-conf/c8y-bridge.conf', type = 'c8y-bridge.conf' },
   { path = '/etc/tedge/mosquitto-conf/tedge-mosquitto.conf', type = 'tedge-mosquitto.conf' },
   { path = '/etc/mosquitto/mosquitto.conf', type = 'mosquitto.conf' },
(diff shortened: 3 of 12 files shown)


tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin.toml
```diff
@@ -4,4 +4,5 @@ files = [
     { path = '/etc/config1.json', type = 'CONFIG1', user = 'tedge', group = 'tedge', mode = 0o640 },
     { path = '/etc/binary-config1.tar.gz', type = 'CONFIG1_BINARY', user = 'tedge', group = 'tedge', mode = 0o640 },
     { path = '/etc/config-root.json', type = 'CONFIG-ROOT', user = 'root', group = 'root', mode = 0o600 },
+    { path = '/etc/containers/certs.d/example/ca.crt', type = 'harbor-certificate', user = 'tedge', group = 'tedge', mode = 0o640 },
```
rina23q: If `parent_user` and `parent_group` are specified to `root`, the failed test *Set Configuration Should Create Parent Directories* will pass.

```suggestion
    { path = '/etc/containers/certs.d/example/ca.crt', type = 'harbor-certificate', user = 'tedge', group = 'tedge', mode = 0o640, parent_user = 'root', parent_group = 'root' },
```

Since `user` and `group` are specified for the file, the current logic assumes the immediate parent directory has to be owned by `tedge` as well. This logic was not a part of the initial requirement. If it's better to get rid of the assumption, just tell me.

docs/src/operate/c8y/configuration-management.md
```diff
@@ -26,13 +26,18 @@ files = [
...
-  { path = '/etc/tedge/c8y/example.txt', type = 'example', user = 'tedge', group = 'tedge', mode = 0o644 }
+  { path = '/etc/tedge/c8y/example.txt', type = 'example', user = 'tedge', group = 'tedge', mode = 0o644 },
+  { path = '/etc/containers/certs.d/example/ca.crt', type = 'harbor-certificate', user = 'tedge', group = 'tedge', mode = 0o640, parent_user = 'root', parent_group = 'root', parent_mode = 0o755 },
 ]
 ```
 
 * `path` is the full path to the configuration file.
 * `type` is a unique alias for each file entry which will be used to represent that file in Cumulocity UI.
-* `user`, `group` and `mode` are UNIX file permission settings to be used to create a configuration file. If not provided, the files will be created with `root` user. If the file exists already, its ownership will be retained.
+* `user`, `group` and `mode` are UNIX file permission settings to be used to create a configuration file. If not provided, the files will be created with the current user and group, or `root`.
```
didier-wenzek: This sentence is not clear: "with the current user and group, or `root`". Which one is actually use?

docs/src/operate/c8y/configuration-management.md
```diff
@@ -26,13 +26,18 @@ files = [
...
+  If the file exists already, its ownership will be retained.
```
didier-wenzek, line 38: *Preserved* as used bellow seems more appropriate.

```suggestion
  If the file exists already, its ownership will be preserved.
```

crates/extensions/tedge_config_manager/src/config.rs
```diff
@@ -146,17 +150,29 @@ impl Borrow<str> for FileEntry {
 }
 
 impl FileEntry {
+    #[allow(clippy::too_many_arguments)]
```
didier-wenzek: I would consider to pass `PermissionEntry` structs instead of creating those from argument lists (indeed too large and error prone).

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -383,57 +383,62 @@ impl ConfigManagerWorker {
...
+                    if let Err(err) = std::fs::create_dir_all(parent) {
+                        bail!(
+                            "failed to create parent directories. path: '{parent}', error: '{err}'"
+                        );
+                    }
+                } else {
+                    needs_elevation = true;
+                }
+            }
+        }
```
didier-wenzek: The algorithm used to detect if an elevation is required or not is incorrect.

- If the user has no write access to the grand-parent directory, they cannot create a parent directory there whatever the  target ownership is.
- The rules to have this correct are complex.
- => I would simply try to create the directory with the current user and call `tedge-writer` as a fallback.
-

docs/src/operate/c8y/configuration-management.md
rina23q: First of all, I would note that this change was made as an update based on the current implementation, regardless of this PR. (= the description was incorrect)

If the first attempt to create a file is successful, the current user and group are used. 
https://github.com/thin-edge/thin-edge.io/blob/c4964d2ed2530e931ddaba774d7a8f8a532b0499/crates/extensions/tedge_config_manager/src/actor.rs#L413

If it's failed, `tedge-write` will be called with `sudo` when it's enabled by the config. Then, the second attempt to create a file by `tedge-write` will use `root`.

crates/extensions/tedge_config_manager/src/actor.rs
rina23q: > If the user has no write access to the grand-parent directory, they cannot create a parent directory there whatever the target ownership is.

True. This was exactly the reason why the test *Set Configuration Should Create Parent Directories* failed before this fixup commit https://github.com/thin-edge/thin-edge.io/pull/3718/commits/6a1bef4ca53c71cf734e6eb4ecf7db47a3b0151c.

> => I would simply try to create the directory with the current user and call tedge-writer as a fallback.

To be honest, this was my first implementation trial not recorded in git. There is a small problem with this.
Let's use this configuration as an example. (given that `a/b/c.txt` doesn't exist yet)
```toml
files = [
    { path = '/etc/tedge/plugins/a/b/c.txt', type = 'abc', user = 'rina', group = 'rina', parent_user = 'rina', parent_group = 'rina' }
]
```
The first attempt to create directories`create_dir_all` for `/etc/tedge/plugins/a/b/c.txt` will be successful and the owner of `a` and `a/b` will be `tedge:tedge` since `tedge-agent` is running by `tedge`. However, applying the specified ownership `rina:rina` to `a/b` (the immediate parent) will fail.

Then, this case needs elevation. `tedge-write` can change the permission. However, for `tedge-write`, the parent directory already exist because it's created by the first attempt by tedge-agent. This means, `tedge-write` must be able to change the ownership and permissions regardless of the condition if `parent.exist()`. This is contradicting the rule, "we don't change the perms and ownership if files/directories already exist." It's not possible for `tedge-write` to distinguish if the immediate parent did exist before, or just freshly created by tedge-agent. 

Maybe this can be controlled by expanding `tedge-write` and call it multiple times.
* `create_dir_all` success && changing perms failed
* -> call `tedge-write` to change only perms of the parent dir
* try to create a file && failed due to permission denied
* -> call `tedge-write` to create a file

docs/src/operate/c8y/configuration-management.md
didier-wenzek: > First of all, I would note that this change was made as an update based on the current implementation, regardless of this PR. (= the description was incorrect)

I fully appreciate that.

> If the first attempt to create a file is successful, the current user and group are used.
> ...
> If the first attempt to create a file is successful, the current user and group are used. 

```suggestion
* `user`, `group` and `mode` are UNIX file permission settings to be used to create a configuration file. If not provided, the files will be created with the current user and group, and, if not feasible, using `root`.
```

didier-wenzek: > If not provided, the files will be created with the current user

Is the current user the current user of the agent? I.e. `tedge`?

crates/extensions/tedge_config_manager/src/actor.rs
didier-wenzek: Things seem to be quite complicated :-( 

> The first attempt to create directories`create_dir_all` for `/etc/tedge/plugins/a/b/c.txt` will be successful and the owner of `a` and `a/b` will be `tedge:tedge` since `tedge-agent` is running by `tedge`.

Created as `tedge:tedge` ? It makes sense, but this is what is said in that comment https://github.com/thin-edge/thin-edge.io/pull/3718#discussion_r2184731769

> However, applying the specified ownership `rina:rina` to `a/b` (the immediate parent) will fail.
> ...
> And, by design, `tedge_write` will not change the ownership of `a/b`

This makes sense and simply implies that we must *never* try to create parent directories without privileges.

> This would contradicting the rule, "we don't change the perms and ownership if files/directories already exist." 

This is an important rule and we must keep respecting it.

> Maybe this can be controlled by expanding `tedge-write` and call it multiple times.

Why not simply call it just once in all and every cases?

- i.e. stop creating missing directories in the agent.

docs/src/operate/c8y/configuration-management.md
rina23q: > Is the current user the current user of the agent? I.e. `tedge`?

Yes, correct.

crates/extensions/tedge_config_manager/src/actor.rs
rina23q: Now the code is much simplified here. https://github.com/thin-edge/thin-edge.io/pull/3718/commits/79b3414a38721a832f830e0569d9c1bbbb8b90d0

The higtlight is:
* If `tedge-write` is enabled, use `tedge-write` directly to create dirs and a file.
* If `tedge-write` is disabled, try to create dirs and a file with the current user.

crates/extensions/tedge_config_manager/src/config.rs
rina23q: Addressed in https://github.com/thin-edge/thin-edge.io/pull/3718/commits/79b3414a38721a832f830e0569d9c1bbbb8b90d0

docs/src/operate/c8y/configuration-management.md
rina23q: Updated to reflect the current decision. b42bb22fb30b099dfd62868031af710c0cca1055
(I might have explained too much details about the ownership as "Getting Started"...)

docs/src/operate/c8y/configuration-management.md
rina23q: The text changed in b42bb22fb30b099dfd62868031af710c0cca1055

crates/extensions/tedge_config_manager/src/actor.rs
didier-wenzek: Indeed, the code is really far simpler. I will be happy to approve once the failing system test fixed.

rina23q: The test failure of *Set Configuration when file exists and tedge run by root* highlighted that I did revert of the decision made in #3073 :( Thanks to the great test coverage and the tag to the ticket.

#3073 says,
> if we have permissions, avoid calling tedge-write (https://github.com/thin-edge/thin-edge.io/pull/3069)

Hmm, what can we do...

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -378,8 +389,25 @@ Set Configuration from URL
     Log    Test Description: ${test_desc}
 
     ThinEdgeIO.Set Device Context    ${device}
-    ${hash_before}=    Execute Command    md5sum ${device_file}
-    ${stat_before}=    Execute Command    stat ${device_file}
+    ${hash_before}=    Execute Command    md5sum ${device_file} ||true
+    ${stat_before}=    Execute Command    stat ${device_file} ||true
```
albinsuresh: That `|| true` was intentional? If the expected file doesn't exist upfront (when it is expected to exist), wouldn't it be better to have the tests fail here with the clear "file not found error" rather than in the file content equiality check failing later on(which could be misleading)?

docs/src/operate/c8y/configuration-management.md
```diff
@@ -18,49 +22,72 @@ The configuration management functionality is provided by the **tedge-agent** se
...
 
 :::note
 You can also configure the `tedge-configuration-plugin.toml` from the cloud later.
 :::
 
 **Step 2**
 Navigate to your Cumulocity Device Management and the desired device. Open its **Configuration** tab.
-You can find `tedge-configuration-plugin` and more are listed as supported configuration types, as declared in the plugin configuration file in step 1.
+You can find `tedge-configuration-plugin` and more are listed as supported configuration types, as declared in the
+plugin configuration file in step 1.
```
didier-wenzek: [Minor] From a personal point of view, I feel harder to read doc source files where sentences are arbitrary cut to fit on 80 columns.

crates/extensions/tedge_config_manager/src/actor.rs
didier-wenzek: The motivation is to avoid a double copy of the file. A solution is to call tedge-write twice. A first time to create missing directories (unless the parent dir already exists); another call to copy the file (unless we can do it without privileges).

docs/src/operate/c8y/configuration-management.md
rina23q: Ah, that's something unexpected. Should be because of IDE settings. I would revert the places that I didn't want to change.

rina23q: Now this formatting rule is removed

tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin.toml
```diff
@@ -4,4 +4,5 @@ files = [
     { path = '/etc/config1.json', type = 'CONFIG1', user = 'tedge', group = 'tedge', mode = 0o640 },
     { path = '/etc/binary-config1.tar.gz', type = 'CONFIG1_BINARY', user = 'tedge', group = 'tedge', mode = 0o640 },
     { path = '/etc/config-root.json', type = 'CONFIG-ROOT', user = 'root', group = 'root', mode = 0o600 },
+    { path = '/etc/containers/certs.d/example/ca.crt', type = 'harbor-certificate', user = 'tedge', group = 'tedge', mode = 0o640, parent_user = 'root', 'parent_group' = 'root' },
```
reubenmiller: NIT, should change any behaviour, but a bit strange in having the single quotes around the property name.


```suggestion
    { path = '/etc/containers/certs.d/example/ca.crt', type = 'harbor-certificate', user = 'tedge', group = 'tedge', mode = 0o640, parent_user = 'root', parent_group = 'root' },
```

rina23q: Fixed here. Also, added the control and check for `parent_mode`. 8c6e9e11839350469a01e91d95da35c3759ae063

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -69,6 +69,19 @@ Set Configuration when tedge-write is in another location
...
+
+    ${config_url}=    Cumulocity.Create Inventory Binary    temp_file    harbor-certificate    contents=DUMMY CONTENTS
+    ${operation}=    Cumulocity.Set Configuration    harbor-certificate    url=${config_url}
+    ${operation}=    Operation Should Be SUCCESSFUL    ${operation}    timeout=120
+
+    ThinEdgeIO.Set Device Context    ${PARENT_SN}
+    ${contents}=    Execute Command    cat /etc/containers/certs.d/example/ca.crt    strip=${True}
+    Should Be Equal    ${contents}    DUMMY CONTENTS
+    ${mode}=    Execute Command    stat -c '%a' /etc/containers/certs.d/example    strip=${True}
+    Should Be Equal    ${mode}    700
```
reubenmiller, line 83: Since the feature also involves setting the directory ownership, we should also check that (both owner and group)

rina23q: Added in 778ea01898d33a928b8462db65d5080e03438865

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
rina23q: @reubenmiller Can I remove `|| true` as Albin suggested?

reubenmiller: Though the downside is that the `Execute Command` will take 30 seconds to fail if the file isn't found (due to the in-built "smart retry" mechanism.

If you want clarify I would suggest using the "File Exists" check prior to getting it, as that will be the most descriptive.

reubenmiller: Though given that the `||true` was only added afterwards, then it might be better to remove it as it seems to be just a by-product of the PR and not directly required by it.

rina23q: Added `ThinEdgeIo.File Should Exist` and remove `||true`. 6a61165e96d3448a3e6271d3900ba534de6dac4c

Issue #3721, opened 2025-07-03
Duplicate Cumulocity devicecontrol messages aren't being ignored


reubenmiller:
**Describe the bug**

https://github.com/thin-edge/thin-edge.io/pull/3454 was created to de-duplicate any incoming operations coming from Cumulocity, to avoid executing the operation twice.

However, this [test run](https://github.com/thin-edge/thin-edge.io/actions/runs/16043523325) shows that the operation handler is triggered twice (about 40 milliseconds later)

Below shows the mqtt log showing that the `c8y_Command` operation (id: `19830895`) is received twice AND processed twice (by noting the `506,19830895` messages which have different contents).

```
---- mqtt messages ----
2025-07-03T06:59:26.418551Z      c8y/s/us                                                               114,c8y_Command,c8y_DeviceProfile,c8y_DownloadConfigFile,c8y_LogfileRequest,c8y_RemoteAccessConnect,c8y_Restart,c8y_SoftwareUpdate,c8y_UploadConfigFile
2025-07-03T06:59:26.542605Z      c8y/s/dat                                                              71,<redacted>
2025-07-03T06:59:26.553273Z      c8y/devicecontrol/notifications                                        {"delivery":{"log":[],"time":"2025-07-03T06:59:26.457Z","status":"PENDING"},"agentId":"7019830853","creationTime":"2025-07-03T06:59:26.415Z","deviceId":"7019830853","id":"19830895","status":"PENDING","c8y_Command":{"text":"echo helloworld1"},"description":"echo helloworld","externalSource":{"externalId":"TST_hurtle_rectilinear_heater","type":"c8y_Serial"}}
2025-07-03T06:59:26.553402Z      c8y/s/ds                                                               511,TST_hurtle_rectilinear_heater,echo helloworld1
2025-07-03T06:59:26.594604Z      c8y/s/us                                                               504,19830895
2025-07-03T06:59:26.594820Z      c8y/s/us                                                               506,19830895,"helloworld1
"
[...]

PR #3728, opened 2025-07-11
fix: continue to de-duplicate incoming operations once operation is complete

jarhodes314:
## Proposed changes
Fix a flaky system test caused by duplicate operations being executed by the mapper.

The reason why the existing de-duplication didn't work is that the operation had been completed locally before the operation notification was received for a second time from Cumulocity. Since the command was no longer "active", the mapper totally ignored the fact the operation has already been executed.

This PR adds a new `recently_completed_commands` store to the mapper, which stores the completed command IDs for ~1 hour to cope with this race condition where the Cumulocity operation status takes some time to update.

## Types of changes

## Paste Link to the issue
- #3721 

## Checklist


didier-wenzek (approved):
Approved. Thank you for fixing that race condition.

Having this very specific and local fix is okay for now.
However, I don't think this is the only place we have to sync local and remote states.

PR #3733, opened 2025-07-15
bump rcgen to 0.14.2

Bravo555:
## Proposed changes

Recent versions of `rcgen` contain nice features, so I decided to bump the crate and fix all breaking changes first before our usage of it is expanded by the pkcs11 create key PR.

Relevant features:
- `0.13.0`
	- P521 support
	- The API used to create/issue key pairs, certificates, certificate signing requests (CSRs), and certificate revocation lists (CRLs) has been restructured to emphasize consistency and avoid common errors with serialization. (better to fix breaking changes ASAP)
	- Support for using `aws-lc-rs` in `fips` mode
- `0.13.1`
	- Several improvements to the capabilities available when working with certificate signing requests
		- Enable signing without private key
- `0.14.0`
	- better remote signing API (`RemoteKeyPair` changed to `SigningKey`)

Also we'd like to move from RSA PKCS#1 v1.5, which is vulnerable to [Marvin Attack](https://people.redhat.com/~hkario/marvin/) to RSA PSS signatures for signing the CSRs, which new versions of `rcgen` may add in the future.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/certificate/src/lib.rs
+++ b/crates/common/certificate/src/lib.rs
@@ -1,7 +1,6 @@
 use anyhow::Context;
 use camino::Utf8Path;
 use device_id::DeviceIdError;
-use rcgen::Certificate;
 use rcgen::CertificateParams;
 use rcgen::KeyPair;
 use sha1::Digest;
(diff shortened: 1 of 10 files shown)


crates/common/certificate/src/lib.rs
```diff
@@ -245,14 +244,43 @@ impl rcgen::RemoteKeyPair for RemoteKeyPair {
...
             .map_err(|e| rcgen::Error::PemError(e.to_string()))
     }
+}
+
+pub struct KeyCertPair {
+    certificate: rcgen::Certificate,
+    // in rcgen 0.14 params are necessary to generate the CSR
+    params: rcgen::CertificateParams,
+    signing_key: SigningKeyWrapper,
+}
```
didier-wenzek, line 254: Just to confirm.

Previously, the `certificate` was protected under `Zeroizing`. Now this protection has been narrowed under the `signing_key` in the case there is actually a private key in memory. This makes sense. However, can you confirm that, with version 0.14, `rcgen::Certificate` no more contains data to zeroize on drop?

Bravo555: Indeed, wrapping `rcgen::Certificate` with `Zeroizing<>` generates the following error:

```
error[E0277]: the trait bound `rcgen::Certificate: zeroize::DefaultIsZeroes` is not satisfied
   --> crates/common/certificate/src/lib.rs:250:18
    |
250 |     certificate: Zeroizing<rcgen::Certificate>,
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `zeroize::DefaultIsZeroes` is not implemented for `rcgen::Certificate`
    |
```

And in 0.14 only [`rcgen::KeyPair` implements `Zeroize`](https://docs.rs/rcgen/latest/src/rcgen/lib.rs.html#882-887), which makes sense because it's the only thing holding sensitive data.

PR #3737, opened 2025-07-21
fix: pkcs11 generate correct CSR signature

Bravo555:
## Proposed changes

Make the signer use a PKCS 1.5 signature when signing CSRs instead of currently used RSA-PSS.

The signature generated so far was invalid because, when generating a CSR with a private key on a PKCS11 token, and if that key was an RSA key, the signer signed the CSR using RSA-PSS signature scheme, but the SignatureAlgorithm identifier in the CSR indicated that the signature was a PKCS 1.5 signature.

This was previously unnoticed because Cumulocity did not verify CSR signature is correct when renewing, but as of time of making this fix, Cumulocity now correctly verifies the signatures.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -331,7 +341,20 @@ impl From<SigScheme> for rustls::SignatureAlgorithm {
...
+
+impl From<rustls::SignatureScheme> for SigScheme {
+    fn from(value: rustls::SignatureScheme) -> Self {
+        match value {
+            rustls::SignatureScheme::ECDSA_NISTP256_SHA256 => SigScheme::EcdsaNistp256Sha256,
+            rustls::SignatureScheme::ECDSA_NISTP384_SHA384 => SigScheme::EcdsaNistp384Sha384,
+            rustls::SignatureScheme::ECDSA_NISTP521_SHA512 => SigScheme::EcdsaNistp521Sha512,
+            rustls::SignatureScheme::RSA_PSS_SHA256 => SigScheme::RsaPssSha256,
+            rustls::SignatureScheme::RSA_PKCS1_SHA256 => SigScheme::RsaPkcs1Sha256,
+            _ => todo!(),
```
jarhodes314: If we can't map all possible source values to a valid target value here, this should probably be a TryFrom

Bravo555: fixed in 08af0dceca by directly using `SigScheme` instead of `rustls::SignatureScheme` and converting to `SigScheme`, so the partial `From` implementation was entirely removed

jarhodes314 (approved):

PR #3739, opened 2025-07-22
build: Make build tools build to different dirs so `just release` doesn't rebuild from scratch every time

Bravo555:
## Proposed changes

Until now, when build.sh script used different build tools, e.g. cargo-zigbuild or clang, all these tools built to the same directory. Because their output is different, they generated different cachedir tags, so this resulted in no artifact caching when running `just release` multiple times on the same code: cargo zigbuild first run and wrote its tag, then clang and overwrote zigbuild's cache tag and on next `just release` invocation zigbuild would have to build from scratch becuase its stuff was overwritten by clang, and vice versa.

Now they output to different directories so they don't overwrite each other's cache. This might increase the size of the build directory.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

Feel free to amend the commit since the change is maybe not the cleanest (mutating a variable defined outside the function, may easily break), but not sure how to do it better.

### DIFF
--- a/ci/build_scripts/build.sh
+++ b/ci/build_scripts/build.sh
@@ -203,16 +203,21 @@ build() {
     shift
     case "$build_tool" in
         zig|ziglang|cargo-zigbuild)
+            # make different tools build to different directories so they don't trash each other's cache
+            BUILD_SUBDIR=target/cargo-zigbuild
+            BUILD_DIR=${BUILD_DIR/target/$BUILD_SUBDIR}
             # shellcheck disable=SC2086
-            cargo-zigbuild $TOOLCHAIN zigbuild "$@"
+            cargo-zigbuild $TOOLCHAIN zigbuild --target-dir "$BUILD_SUBDIR" "$@"
             ;;
         clang)
+            BUILD_SUBDIR=target/clang
+            BUILD_DIR=${BUILD_DIR/target/$BUILD_SUBDIR}
             # shellcheck disable=SC2086
-            mk/cargo.sh $TOOLCHAIN build "$@"
+            mk/cargo.sh $TOOLCHAIN build --target-dir "$BUILD_SUBDIR" "$@"
             ;;
         native|*)
             # shellcheck disable=SC2086
-            cargo $TOOLCHAIN build "$@"
+            cargo $TOOLCHAIN build --target-dir "$BINARY_TARGET" "$@"
             ;;
     esac
 }


Bravo555:
> Though isn't this only a problem if you're switching between the different tools

By "tools" I meant cargo-zigbuild and clang, which we use both in the build script and are referred to as "build tools" in the script. In `get_build_tool_for_binary()` we set zig as build tool for the `tedge-p11-server` and `clang` for other packages. When we run the default `just release` both of these tools are used and they overwrite each other's output, so consecutively running `just release` leads to everything being rebuilt from scratch every time.

I'm under the impression that this is the default behaviour since `tedge-p11-server` binary was added (previously build artifacts were cached, so running `just release` multiple time would build just one time and on next invocations would just repackage). Just to be clear, are you not observing the behaviour where running `just release` multiple times it always builds everything from scratch? This is the case on my machine and it's a bit annoying when running `just integration-test` runs `just release` and when I run the tests multiple times it rebuilds everything even if code wasn't changed.

reubenmiller:
> > Though isn't this only a problem if you're switching between the different tools
> 
> By "tools" I meant cargo-zigbuild and clang, which we use both in the build script and are referred to as "build tools" in the script. In `get_build_tool_for_binary()` we set zig as build tool for the `tedge-p11-server` and `clang` for other packages. When we run the default `just release` both of these tools are used and they overwrite each other's output, so consecutively running `just release` leads to everything being rebuilt from scratch every time.
> 
> I'm under the impression that this is the default behaviour since `tedge-p11-server` binary was added (previously build artifacts were cached, so running `just release` multiple time would build just one time and on next invocations would just repackage). Just to be clear, are you not observing the behaviour where running `just release` multiple times it always builds everything from scratch? This is the case on my machine and it's a bit annoying when running `just integration-test` runs `just release` and when I run the tests multiple times it rebuilds everything even if code wasn't changed.

No, the default for running `just release` locally is both zig...I think you're just seeing the rebuilding because if you change a commit, the build.rs must run again (as it included the SEMVER env variable to embed the git-version number into the binary).

Bravo555:
Indeed looks like my hypothesis was wrong because after bump to 1.85 (not sure if related) without this change `just release` does the correct thing, i.e. doesn't rebuild from scratch. Also not sure why the change seemed to fix the problem, indeed maybe my environment cursed somehow, again. In any case, because it works now and the "fix" wasn't correct, closing.

PR #3745, opened 2025-07-28
ci: update to ubuntu-24.04 runner

reubenmiller:
## Proposed changes

Update the runner to use ubuntu-24.04 for a more modern OS version that seems to have less issues when cross compiling C libraries (required in upcoming tedge flows feature which needs to compile the quickjs C library).

[ubuntu-24.04](https://github.com/actions/runner-images/blob/ubuntu22/20250720.1/images/ubuntu/Ubuntu2404-Readme.md) does not include cargo-audit, so it needs to be explicitly installed (which is better anyway).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/.github/workflows/pull-request-checks.yml
+++ b/.github/workflows/pull-request-checks.yml
@@ -193,6 +193,9 @@ jobs:
       - name: Install stable
         uses: dtolnay/rust-toolchain@stable
 
+      - name: cargo install cargo-audit
+        uses: taiki-e/install-action@cargo-audit
+
       - name: Cargo audit
         run: cargo audit
 
--- a/.github/workflows/pull-request-checks.yml
+++ b/.github/workflows/pull-request-checks.yml
@@ -10,7 +10,7 @@ env:
 jobs:
   block-fixup:
     name: Block fixup commits
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
     if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-request-checks.yml
```diff
@@ -10,7 +10,7 @@ env:
 jobs:
   block-fixup:
     name: Block fixup commits
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
rina23q, line 13: Just trivial thought. I found some jobs run on `ubuntu-latest`. I can understand why some jobs need a specific version, especially the jobs that requires build. But, for example, this `block-fixup`, `Filter changes`, and maybe more, I don't think it has to be on the specific OS version. 
Anyway it's too trivial and I don't think it's worth changing. If we want consistency, changing all `ubuntu-latest` to `ubuntu-24.04` might not be a bad idea.

rina23q (approved):
LGTM

.github/workflows/pull-request-checks.yml
reubenmiller: Yeah I left at least the jobs which are marked as ubuntu-latest as that generally indicates that the runner environment does not matter so match, and things should work if the underlying runner image changes. We'll do a review of the runner types in a different PR to see if we can consolidate, but not a huge priority right now.

PR #3751, opened 2025-07-31
refactor(pkcs11): `TedgeP11Service`

Bravo555:
## Proposed changes

Create a `TedgeP11Service` trait for a standard interface implemented by both `Cryptoki` (direct interaction with P11 modules) and `TedgeP11Client`.

 In #3709, to create keys we can use `TedgeP11Client` or `Cryptoki` directly. However, to not have to branch unnecessarily, we'd like these both to have the same interface for creating keys. With previous design this was not possible, as the only trait we had was `TedgeP11Signer` which represented an already present signing key located on the token, but we wanted to create a new one.
 
The new `TedgeP11Service` (name reused from a previous trait) is the main top-level trait for all P11-related operations (not only signing, but in #3709 also creating new keys, and in the future maybe also listing all the keys on the token), implemented both by `Cryptoki` and `TedgeP11Client`.

One additional advantage is now, `TedgeP11Server` uses `Cryptoki` using its `TedgeP11Service` API the same as we would directly by other callers, so we achieved some additional deduplication.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge-p11-server/src/proxy/client.rs
```diff
@@ -43,24 +70,27 @@ impl TedgeP11Client {
...
 
         // make any request to make sure the service is online and it will respond
-        let _ = client.choose_scheme(&[], None);
+        let _ = client._choose_scheme(&[], None);
 
         client
     }
 
-    pub fn choose_scheme(
+    pub fn _choose_scheme(
```
didier-wenzek: It's a bit surprising to see an underscore for a name that is actually used.

That said, I have no nice alternative. Maybe: `do_choose_scheme`?

crates/extensions/tedge-p11-server/src/proxy/server.rs
```diff
@@ -5,19 +5,21 @@ use tracing::error;
...
 pub struct TedgeP11Server {
-    service: Box<dyn SigningService + Send + Sync>,
+    service: Box<dyn TedgeP11Service>,
 }
 
 impl TedgeP11Server {
     pub fn new<S>(service: S) -> anyhow::Result<Self>
     where
-        S: SigningService + Send + Sync + 'static,
+        S: TedgeP11Service + Send + Sync + 'static,
```
didier-wenzek:
```suggestion
        S: TedgeP11Service + 'static,
```

didier-wenzek (approved):
Approved, with minor comments.

crates/extensions/tedge-p11-server/src/proxy/client.rs
Bravo555: Was able to just call it `choose_scheme`, seems there are no problems with name shadowing as correct function is selected.

crates/extensions/tedge-p11-server/src/proxy/server.rs
Bravo555: done in 0816b98.

Issue #3430, opened 2025-02-27
Add inventory.json file support for child devices via the agent


albinsuresh:
**Is your feature improvement request related to a problem? Please describe.**

Currently the mapper is processing the `inventory.json` file by publishing the content to the corresponding twin topics and to the cloud. But, keeping it as part of the mapper limits this functionality to the main device only whereas even child devices could benefit from the same.

**Describe the solution you'd like**

Move the responsibility of processing the `inventory.json` file from the mapper to the agent so that this functionality can be used by child devices as well. Propagating the same to the cloud will be handled automatically by the mapper, when it receives those twin messages.

PR #3755, opened 2025-08-06
feat: inventory.json support for child devices

albinsuresh:
## Proposed changes

Enable users to specify initial inventory data for child devices running tedge-agent  using an inventory.json file.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3430

## Checklist


tests/RobotFramework/tests/cumulocity/inventory/inventory_update_child_device.robot
```diff
@@ -0,0 +1,68 @@
+*** Settings ***
+Resource            ../../../resources/common.resource
+Library             DateTime
+Library             String
+Library             Cumulocity
+Library             ThinEdgeIO
+
+Test Setup          Custom Setup
+Test Teardown       Custom Teardown
```
Bravo555, line 9:
```suggestion
Suite Setup          Custom Setup
Suite Teardown       Custom Teardown
```

crates/core/tedge_agent/src/twin_manager/actor.rs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        };
+        let inventory_json: Value = serde_json::from_reader(file).map_err(|err| {
+            RuntimeError::ActorError(
+                format!("Failed to parse inventory file contents as JSON: {}", err).into(),
+            )
+        })?;
+        let Value::Object(twin_map) = inventory_json else {
+            return Err(RuntimeError::ActorError(
+                "Invalid inventory.json format: expected a JSON object".into(),
```
Bravo555: todo: attach the filename to the error

crates/core/tedge_agent/src/twin_manager/actor.rs
```diff
@@ -0,0 +1,135 @@
...
+                    Ok(Some(next)) => next,
+                    _ => break,
+                };
+            }
+        }
+
+        let device_id = self.config.device_topic_id.clone();
+        for (key, value) in inventory_map {
+            self.publish_twin_data(&device_id, key.clone(), value.clone())
+                .await;
```
Bravo555: suggestion: don't have to clone here

```suggestion
            self.publish_twin_data(&device_id, key, value).await;
```

crates/core/tedge_agent/src/twin_manager/actor.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        // This is to prevent the MQTT actor from crashing while trying to send a twin message to this actor later
+        // after this actor has finished and closed its message box, but the MQTT actor still holds the sender half of it.
+        while self.messages.recv().await.is_some() {
+            // Do nothing
+        }
+        Ok(())
+    }
```
Bravo555: thought: So as I understand, basically what we do here is:

1. Read the fragments from the inventory file
2. Wait until the end of the "init phase", during which we remove fragments that are already published (our heuristic for "end of the init phase" being messages not being published for 1s)
3. Publish all the remaining fragments
4. Effectively exit, but we're forced to loop because other actors may send us messages despite us not wanting to process them anymore.

So the actor stays alive, despite not doing anything, which sucks a bit, but as the comment says, it's hard to exit without the MQTT actor possibly receiving failures after we exit.

Funnily enough, this usage pattern here could be easily handled by an in-memory broker: we could subscribe for a bit, listen to some messages, and exit without other actors even being aware of us, because sources and sinks would be decoupled, but the actor runtime works more on a peer-to-peer basis, so that's why it's hard.

Even still, perhaps upstream producers being coupled to their downstream consumers such that the producer (actor A) can't output after the consumer (actor B) exits isn't a nice property. Perhaps the runtime should simply remove the consumer message box from producer's sender list and let the producer proceed, only if A doesn't depend on B in any way. But there's also a pattern where A sends message to B and then waits for B's response, in which case it does depend on it and should know ASAP if B exits (it should then probably also exit with an error).

So we have both request/response and publish/subscribe patterns, in this case publish/susbcribe is a bit awkward.

crates/core/tedge_agent/src/twin_manager/actor.rs
Bravo555, line 1: todo: https://thin-edge.github.io/thin-edge.io/references/mappers/c8y-mapper/#base-inventory-model will also have to be updated.

Bravo555 (changes requested):
One error message adjustment request and docs update, apart from that, looks good and will approve after these changes.

crates/core/tedge_agent/src/twin_manager/actor.rs
albinsuresh: > Even still, perhaps upstream producers being coupled to their downstream consumers such that the producer (actor A) can't output after the consumer (actor B) exits isn't a nice property.

Completely agree. I am really tempted to change the MQTT actor logic to not fail on send when a channel is closed, but just log and proceed (potentially removing that sender from its `peer_senders` list as well). I just wasn't fully sure if some other part of the code is expecting on the existing behaviour to handle things like shutdown requests. That's why I went with this not-so-pretty workaround. But, I'm open to do that change in a separate PR so that we can discuss the impact.

Bravo555: I'm happy to accept this workaround. As for changing only the MQTT actor (because indeed actors themselves have a choice to exit or keep going if send fails!), also approve making the modification in a separate PR, I'm curious what the test results would be.

tests/RobotFramework/tests/cumulocity/inventory/inventory_update_child_device.robot
albinsuresh: Resolved by [8754d74](https://github.com/thin-edge/thin-edge.io/pull/3755/commits/8754d74929df9149acb72f6d9f37bcd9a8a531c9)

crates/core/tedge_agent/src/twin_manager/actor.rs
albinsuresh: Resolved by [8754d74](https://github.com/thin-edge/thin-edge.io/pull/3755/commits/8754d74929df9149acb72f6d9f37bcd9a8a531c9)

crates/core/tedge_agent/src/twin_manager/actor.rs
albinsuresh: Resolved by [8754d74](https://github.com/thin-edge/thin-edge.io/pull/3755/commits/8754d74929df9149acb72f6d9f37bcd9a8a531c9)

tests/RobotFramework/tests/cumulocity/inventory/inventory_update_child_device.robot
reubenmiller: Am I missing something? What is the reasoning from using Test vs Suite setups here? Given that there is only one system test in the suite, there will be zero effect, however I don't get the motivation from switching from Test to Suite.

tests/RobotFramework/tests/cumulocity/inventory/inventory_update_child_device.robot
```diff
@@ -0,0 +1,68 @@
...
+
+*** Keywords ***
+Custom Setup
+    # Parent
+    ${PARENT_SN}=    Setup    connect=${False}
+    Set Suite Variable    $PARENT_SN
+
+    ${parent_ip}=    Get IP Address
+    Set Suite Variable    $PARENT_IP    ${parent_ip}
+    Execute Command    sudo tedge config set c8y.enable.log_upload true
```
rina23q: I think this is unnecessary.

tests/RobotFramework/tests/cumulocity/inventory/inventory_update_bootstrap.robot
```diff
@@ -6,7 +6,7 @@ Library             ThinEdgeIO
 Test Setup          Custom Setup
 Test Teardown       Get Logs
 
-Test Tags           theme:c8y    theme:telemetry
+Test Tags           theme:c8y
```
rina23q: You can also remove `theme:telemetry` in `inventory_update.robot`.

reubenmiller: The `theme:telemetry` tag is still ok here as we're talking about digital twin data (which is a kind of telemetry data)

reubenmiller: I would actually keep the `theme:telemetry` as digital twin properties is telemetry data.

tests/RobotFramework/tests/cumulocity/inventory/inventory_update_child_device.robot
reubenmiller: Closing the comment as generally the users can change the Test/Suite setup to their liking, however a Test Setup/Teardown would be safer than a suite-wide setup as it offers test isolation (with the downside of taking longer to execute).

tests/RobotFramework/tests/cumulocity/inventory/inventory_update_bootstrap.robot
albinsuresh: Resolved in [4612f52](https://github.com/thin-edge/thin-edge.io/pull/3755/commits/4612f524e50ea96e99778d3451efe981b66e8303)

albinsuresh: Resolved in [4612f52](https://github.com/thin-edge/thin-edge.io/pull/3755/commits/4612f524e50ea96e99778d3451efe981b66e8303)

tests/RobotFramework/tests/cumulocity/inventory/inventory_update_child_device.robot
albinsuresh: Resolved in [4612f52](https://github.com/thin-edge/thin-edge.io/pull/3755/commits/4612f524e50ea96e99778d3451efe981b66e8303)

crates/core/tedge_agent/src/twin_manager/actor.rs
albinsuresh: Resolved in [4612f52](https://github.com/thin-edge/thin-edge.io/pull/3755/commits/4612f524e50ea96e99778d3451efe981b66e8303)

PR #3757, opened 2025-08-12
feat: generating CSR using any PKCS11 key

Bravo555:
## Proposed changes

This PR removes a limitation for generating CSRs using PKCS11 private keys that they can only be generated for the same certificate that is already present. Now, when generating CSRs with PKCS11 keys, previous certificate is no longer reused, so one can create CSRs using new keys.

This unlocks two usecases that were previously impossible:

- using `tedge cert renew` to install a new certificate when we already have a certificate but a different keypair is used (previously `SubjectPublicKeyInfo` was reused from previous cert so using different key didn't work)
- using `tedge cert download c8y` when we don't yet have a certificate and register the device to C8y CA and download the initial certificate (previously some fields of CSR were reused from older cert so had no way to fill these fields without some certificate already being present)

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


tests/RobotFramework/tests/pkcs11/private_key_storage.robot
```diff
@@ -111,13 +111,34 @@ Can use PKCS11 key to renew the public certificate
...
     Execute Command    systemctl start tedge-p11-server.socket
+
     Execute Command    cmd=tedge config set c8y.device.key_uri pkcs11:object=nonexistent_key
     Command Should Fail With
     ...    tedge cert renew c8y
-    ...    error=PEM error: protocol error: bad response, expected sign, received: Error(ProtocolError("PKCS #11 service failed: Failed to find a signing key: Failed to find a private key"))
+    ...    error=PKCS #11 service failed: Failed to find a key
+    Execute Command    cmd=tedge config unset c8y.device.key_uri
+
+Can use tedge cert download c8y to download a certificate
```
albinsuresh, line 124: Since these tests are using suite level setup and hence this test can be accidentally influenced by other tests in future, wouldn't it be safer to do a `tedge cert delete` before proceeding with the rest of the test to make sure that there weren't any prior certificates present?

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -263,6 +269,89 @@ impl Cryptoki {
...
+        KeyType::RSA => {
+            let attrs = session.get_attributes(
+                key,
+                &[AttributeType::Modulus, AttributeType::PublicExponent],
+            )?;
+            trace!(?attrs);
+            let mut attrs = attrs.into_iter();
+
+            let Attribute::Modulus(modulus) = attrs.next().context("Not modulus")? else {
+                anyhow::bail!("No modulus");
```
albinsuresh, line 317: Surprised to see so many panics in the server code for a client request. Can a malformed configuration/input trigger these panics? Shouldn't those be sent in the response? The same comment for the rest of the `bails` introduced below.

crates/extensions/tedge-p11-server/src/proxy/server.rs
```diff
@@ -103,6 +104,20 @@ impl TedgeP11Server {
...
+            Frame1::GetPublicKeyPemRequest(uri) => {
+                let response = self.service.get_public_key_pem(uri.as_deref());
+                match response {
+                    Ok(pubkey_pem) => Frame1::GetPublicKeyPemResponse(pubkey_pem),
+                    Err(err) => {
+                        let response = Frame1::Error(ProtocolError(format!(
+                            "PKCS #11 service failed: {err:#}"
+                        )));
+                        connection.write_frame(&response)?;
+                        anyhow::bail!(err);
```
albinsuresh, line 117: Another bail that appears problematic, if it can be triggered by a misconfiguration.

tests/RobotFramework/tests/pkcs11/private_key_storage.robot
Bravo555: Yeah, currently generated keys and certs have a unique random suffix added so they don't conflict, but in hindsight just removing them and restoring the device to the state before the test case runs would be better because right now it's a bit of a mess because there's state in so many different places and it's hard to remember all of it.
So in some next refactor PR I'll make sure no stuff gets persisted between test cases, which will be much simpler.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: `anyhow::bail!` does not panic, it's just short for `return Err(anyhow!($args))`, so an error is returned (even though now I see I didn't add any `.context()` calls above, so if returned, the error would say something like: "PKCS #11 service failed: No modulus" which would be hard to understand).

However, this error will be triggered only when a pathologically uncompliant PKCS11 library is used, because the standard guarantees these objects (object class `CKO_PUBLIC_KEY`, key type `CKK_RSA`/`CKK_EC`) should have these attributes. But this isn't very well expressed in the current code, there should be clear distinction between expected errors, and unexpected ones, such as this. Comment on line 303 also relates to this.

So while it won't panic and is very unlikely to be observed, this is probably a good candidate for a follow-up refactor.

crates/extensions/tedge-p11-server/src/proxy/server.rs
Bravo555: As above, `anyhow::bail!` is expected, just returns the error to the caller.

Issue #3768, opened 2025-08-22
allow non-numeric meta properties in tedge measurements


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Publishing a thin-edge.io measurement which includes non-numeric fragments is not supported where-as the Cumulocity measurement format does support additional meta-information fragments which allows users to contextualize the measurement with additional properties (and avoid those properties being presented as measurements).

```sh
tedge mqtt pub 'te/device/main///m/machine_energy_measurement' '{
  "robot_mech_energy": {
     "accumulated_energy": 0.1,
     "ROB_1_axis_1_energy": 0.1,
     "ROB_1_axis_2_energy": 0.2,
     "ROB_1_axis_3_energy": 0.3,
     "ROB_1_axis_4_energy": 0.4,
     "ROB_1_axis_5_energy": 0.5,
     "ROB_1_axis_6_energy": 0.6,
     "ROB_2_axis_1_energy": 0.1,
     "ROB_2_axis_2_energy": 0.2,
     "ROB_2_axis_3_energy": 0.3,
     "ROB_2_axis_4_energy": 0.4,
     "ROB_2_axis_5_energy": 0.5,
     "ROB_2_axis_6_energy": 0.6
  },
  "measurement_context": {
    "status": "nominal",
    "cycle_id": "cycle-98765"
  }
}'
```

The above measurement format will result in a parsing error due to the **measurement_context** fragment containing string values.

```sh
$ sudo journalctl -fu tedge-mapper-c8y
Aug 22 09:33:08 tedge tedge-mapper[5392]: 2025-08-22T09:33:08.257174875Z ERROR c8y_mapper_ext::converter: Mapping error: Failed to convert a message on topic 'te/device/main///m/machine_energy_measurement': Invalid JSON: Unexpected type within a group at line 18 column 23: `",
Aug 22 09:33:08 tedge tedge-mapper[5392]:     "cycle_id": "cycle-98765"
Aug 22 09:33:08 tedge tedge-mapper[5392]:   }
Aug 22 09:33:08 tedge tedge-mapper[5392]: }
Aug 22 09:33:08 tedge tedge-mapper[5392]: `
```

**Describe the solution you'd like**

Allow users to publish measurements which contain additional non-numeric properties outside of the fragments which contain a value.

* If the user mixes value

### Example 1

```sh
tedge mqtt pub 'te/device/main///m/machine_energy_measurement' '{
  "robot_mech_energy": {
[...]

PR #3774, opened 2025-08-28
feat: publish properties along tedge measurements

didier-wenzek:
## Proposed changes

Allow  meta properties to be published along tedge measurements, and forward these properties unchanged to Cumulocity.

```
tedge mqtt pub te/device/main///m/example '{
  "time": "2020-10-15T05:30:47+00:00",
  "temperature": 25,
  "properties": {
    "sensor": "DS18B20",
    "has_alarm": false,
    "samples": [ 25.1, 25.0, 24.9],
    "range": {
       "min": "-55°C",
       "max": "+125°C"
    }
  }
}'
```

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3768

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/measurement/mod.rs
+++ b/crates/core/tedge_api/src/measurement/mod.rs
@@ -81,6 +81,10 @@ pub use serialize::*;
 ///    fn visit_text_property(&mut self, _name: &str, _value: &str) -> Result<(), Self::Error>{
 ///         Ok(())
 ///    }
+///
+///    fn visit_json_property(&mut self, _name: &str, _value: serde_json::Value) -> Result<(), Self::Error>{
+///         Ok(())
+///    }
 /// }
 /// ```
 pub trait MeasurementVisitor {
@@ -95,6 +99,13 @@ pub trait MeasurementVisitor {
 
     /// Add a text property, attached to the current group if any.
     fn visit_text_property(&mut self, _name: &str, _value: &str) -> Result<(), Self::Error>;
+
+    /// Add a free form property that is not a measurement nor a text.
+    fn visit_json_property(
+        &mut self,
+        name: &str,
+        value: serde_json::Value,
+    ) -> Result<(), Self::Error>;
 
     /// Start to gather measurements for a group.
     fn visit_start_group(&mut self, group: &str) -> Result<(), Self::Error>;
--- a/crates/core/tedge_api/src/measurement/serialize.rs
+++ b/crates/core/tedge_api/src/measurement/serialize.rs
@@ -9,6 +9,7 @@ pub struct ThinEdgeJsonSerializer {
     is_within_group: bool,
     default_timestamp: Option<OffsetDateTime>,
     timestamp_present: bool,
+    properties: Option<serde_json::Value>,
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -59,6 +60,7 @@ impl ThinEdgeJsonSerializer {
             is_within_group: false,
             default_timestamp,
             timestamp_present: false,
+            properties: None,
         }
     }
 
@@ -71,6 +73,11 @@ impl ThinEdgeJsonSerializer {
             if let Some(default_timestamp) = self.default_timestamp {
                 self.visit_timestamp(default_timestamp)?;
             }
+        }
+
+        if let Some(properties) = &self.properties {
+            self.json.write_key("properties")?;
+            self.json.write_value(properties)?;
         }
 
         self.json.write_close_obj();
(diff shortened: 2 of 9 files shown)


docs/src/understand/thin-edge-json.md
```diff
@@ -162,6 +162,41 @@ and hence must not be used as measurement keys:
...
+  "time":"2020-10-15T05:30:47Z",
+  "temperature":{"temperature":{"value":25.0}},
+  "sensor":"DS18B20",
+  "has_alarm":false,
+  "samples":[25.1,25.0,24.9],
+  "range":{"min":"-55°C","max":"+125°C"},
+  "type":"example"
+}
+```
+
```
albinsuresh, line 202: Would be good to add an additional note that nesting in object metadata values is limited to just one level.

docs/src/understand/thin-edge-json.md
```diff
@@ -162,6 +162,41 @@ and hence must not be used as measurement keys:
...
+  "has_alarm": false,
+  "samples": [ 25.1, 25.0, 24.9],
+  "range": {
+     "min": "-55°C",
+     "max": "+125°C"
+  }
+}'
+```
+
+These additional non-numeric properties (texts, booleans, arrays as well as records of those)
```
albinsuresh: Though it's mentioned that non-numeric values are expected here, we might wanna stress it a bit more as users trying to capture numeric metadata values might get surprised by the current behaviour (although it can be explained/defended as existing behaviour). For example, I accidentally published the following message:
```
tedge mqtt pub 'te/device/main///m/ac_temp' '{
    "temp": 22,
    "unit": "C"
    "range" : { 
        "min": 18, 
        "max": 32
    }
}'
```
... expecting the `range` object to be captured as metadata, but its values were interpreted as measurements. This is perfectly fine as that has always been the behaviour with multi-valued measurements. It's just that a user might accidentally get surprised. Of course they can get around this by converting those numeric values as strings, as you've done in these examples.

I won't be surprised if users of this feature soon requests for numeric metadata value support as well, but that's gonna be tricky with backward compatibility.

crates/core/tedge_api/src/measurement/mod.rs
```diff
@@ -96,6 +100,13 @@ pub trait MeasurementVisitor {
     /// Add a text property, attached to the current group if any.
     fn visit_text_property(&mut self, _name: &str, _value: &str) -> Result<(), Self::Error>;
 
+    /// Add a free form property that is not a measurement nor a text.
+    fn visit_json_property(
```
albinsuresh, line 104: Could this have replaced the existing `visit_text_property`, as this one covers string values as well? From what I remember, `visit_text_property` was introduced to send the `type` fields as-is to the cloud. And that's exactly what would be done with non-string metadata as well. So, wondering if we really need to differentiate between these.

crates/core/tedge_api/src/measurement/serialize.rs
```diff
@@ -73,6 +75,16 @@ impl ThinEdgeJsonSerializer {
             }
         }
 
+        if !self.metadata.is_empty() {
+            self.json.write_key("properties")?;
```
albinsuresh, line 79: What's this special `"properties"` key? It looked like all non-numeric keys at the root of the JSON would be captured into a special "properties" object (which would have been weird). But, I don't see that happening in practice, which is good. But that makes me wonder when this `properties ` key is used.

crates/core/tedge_api/src/measurement/mod.rs
didier-wenzek: I avoided this for now because the `type` property is kind of specific. I will consider a simplification.

crates/core/tedge_api/src/measurement/serialize.rs
didier-wenzek: This serializer is only used by unit tests.

docs/src/understand/thin-edge-json.md
didier-wenzek: > ... expecting the range object to be captured as metadata, but its values were interpreted as measurements. 

I don't see how this can be avoided without breaking backward compatibility.

`"range" : { "min": 18, "max": 32  }` is interpreted since day one as a group of measurements with two values.

An option could be to explicitly group *all* the metadata in a group with a reserved name (e.g. `meta`):

```
tedge mqtt pub 'te/device/main///m/ac_temp' '{
    "temp": 22,
    "meta": {
        "unit": "C"
        "range" : { 
           "min": 18, 
           "max": 32
         }
    }
}'
```

docs/src/understand/thin-edge-json.md
didier-wenzek: Added: https://github.com/thin-edge/thin-edge.io/pull/3774/commits/b18c973f2a54b0348b7ff14cff02f6115c4b9a09

docs/src/understand/thin-edge-json.md
didier-wenzek: I added a suggestion along these lines: https://github.com/thin-edge/thin-edge.io/issues/3768#issuecomment-3249453648

didier-wenzek: We decided to group all the properties under a `properties` object.

docs/src/understand/thin-edge-json.md
didier-wenzek: This does no more apply, as we decided to group all the properties under a `properties` object.

rina23q (approved):
LGTM

Issue #982, opened 2022-03-07
Allow units via thin-edge json


mbay-SAG:
**Describe the solution you'd like**
Currently devices can not send unit via the thin-edge json approach. However this is needed for some use cases where the device is actively sending the unit.
**Describe alternatives you've considered**
Data point library does work works on platform side, however at some points  the unit should come directly from the device itself (e.g. switch mode metric <-> imperial).
**Additional context**

didier-wenzek:
Is the unit sent along all and every data points? Or is the unit defined as metadata of the measurement source ?

didier-wenzek:
There is broader discussion about thin-edge json here https://github.com/thin-edge/thin-edge.io-specs/pull/19

albinsuresh:
@mbay-SAG @mbie-sag 

Is it really useful to send the unit along with each and every measurement? Or you'd rather want the ability to set/update the unit of a measurement (its metadata) either as a static configuration or as a dynamic property? Expecting the unit with every measurement would definitely be the easiest thing to implement. But wouldn't that put unnecessary burden on the end user in terms of wasted bandwidth on every measurement sent?

As an extended query, when a measurement is sent **without a unit** from the device, what would be the preferred behaviour? 

* That measurement remains unit-less?
* It has the same unit as the last measurement?
* Inherit the unit from some metadata associated with that measurement (assuming we support some action to set/update the unit for a measurement at a device level)?

mbay-SAG:
Currently on C8Y the unit can be send, but is optional. So I would expect that if I pack the unit into my json, that it is used by the mapper. However if not, then not. 

How the ui handles that is nothing that is implemented within thin-edge, is it? I personally would prefer that the ui uses the unit from the last measurement where a unit was provided but this seems to be problematic considering retentions etc.

reubenmiller:
We accidentally published docs describing this feature, though it isn't implemented. Below is the docs that can be added to the `docs/src/references/mappers/c8y-mapper.md` page once the implementation is done:

```markdown
#### Measurement with unit

The unit is a metadata associated with measurements which can be registered as a metadata message for a given measurement type.
If the following metadata message is registered for the `environment` measurement type:

```sh te2mqtt formats=v1
tedge mqtt pub -r te/device/main///m/environment/meta '{
  "units": {
    "temperature": "°C"
  }
}'
```

Then subsequent messages will be mapped with the registered unit value as follows.

<div class="code-indent-left">

**%%te%% (input)**

[...]

PR #3776, opened 2025-09-04
feat: Define measurement units using topic metadata

didier-wenzek:
## Proposed changes

- [x] Combine thin-edge JSON measurements with measurement metadata to produce C8Y JSON measurements with units
- [ ] Use the `te/+/+/+/+/m/+/meta` topics to update measurement metadata
- [ ] Doc & Test

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/982

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -471,7 +471,7 @@ impl CumulocityConverter {
         if let Some(entity) = self.entity_cache.get(source) {
             // Need to check if the input Thin Edge JSON is valid before adding a child ID to list
             let c8y_json_payload =
-                json::from_thin_edge_json(input.payload_str()?, entity, measurement_type)?;
+                json::from_thin_edge_json(input.payload_str()?, entity, measurement_type, None)?;
 
             if c8y_json_payload.len() < self.size_threshold.0 {
                 mqtt_messages.push(MqttMessage::new(
--- a/crates/extensions/c8y_mapper_ext/src/json.rs
+++ b/crates/extensions/c8y_mapper_ext/src/json.rs
@@ -20,6 +20,7 @@ use crate::entity_cache::CloudEntityMetadata;
 use crate::serializer;
 use clock::Clock;
 use clock::WallClock;
+use std::collections::HashMap;
 use tedge_api::measurement::*;
 use time::OffsetDateTime;
 use time::{self};
@@ -38,9 +39,10 @@ pub fn from_thin_edge_json(
     input: &str,
     entity: &CloudEntityMetadata,
     m_type: &str,
+    units: Option<&Units>,
 ) -> Result<String, CumulocityJsonError> {
     let timestamp = WallClock.now();
-    let c8y_vec = from_thin_edge_json_with_timestamp(input, timestamp, entity, m_type)?;
+    let c8y_vec = from_thin_edge_json_with_timestamp(input, timestamp, entity, m_type, units)?;
     Ok(c8y_vec)
 }
 
@@ -49,10 +51,77 @@ fn from_thin_edge_json_with_timestamp(
     timestamp: OffsetDateTime,
     entity: &CloudEntityMetadata,
     m_type: &str,
+    units: Option<&Units>,
 ) -> Result<String, CumulocityJsonError> {
-    let mut serializer = serializer::C8yJsonSerializer::new(timestamp, entity, m_type);
+    let mut serializer = serializer::C8yJsonSerializer::new(timestamp, entity, m_type, units);
     parse_str(input, &mut serializer)?;
     Ok(serializer.into_string()?)
+}
+
+/// Units used for measurements of a given type
+#[derive(Default)]
+pub struct Units {
+    units: HashMap<String, String>,
+    groups: HashMap<String, Units>,
+}
+
+impl Units {
+    /// An empty set of measurement units
+    ///
+    /// This is the default when no measurement metadata is published for a measurement topic
+    pub fn new() -> Units {
+        Units {
+            units: HashMap::new(),
+            groups: HashMap::new(),
+        }
+    }
+
+    /// True if not units are actually defined
+    pub fn is_empty(&self) -> bool {
+        self.units.is_empty() && self.groups.is_empty()
+    }
+
+    /// Measurement units as defined by metadata published on a measurement topic
+    pub fn from_metadata(meta: serde_json::Value) -> Self {
+        let mut units = Units::new();
+        if let serde_json::Value::Object(map) = meta {
+            for (k, v) in map {
+                units.set_unit(k, v);
+            }
+        }
+        units
+    }
+
+    pub fn set_unit(&mut self, measurement: String, meta: serde_json::Value) {
+        if let serde_json::Value::String(unit) = meta {
+            // "Temperature": "°C"
+            self.units.insert(measurement, unit);
+        } else if let Some(unit) = meta.get("unit") {
+            // "Temperature": {"unit": "°C"},
+            self.set_unit(measurement, unit.clone());
+        } else {
+            // "Climate": { "Temperature": {"unit": "°C"}, "Humidity": {"unit": "%RH"} }
+            let group = measurement;
+            self.set_group_units(group, meta);
+        }
+    }
+
+    pub fn set_group_units(&mut self, group: String, meta: serde_json::Value) {
+        let units = Units::from_metadata(meta);
+        if !units.is_empty() {
+            self.groups.insert(group, units);
+        }
+    }
+
+    /// Retrieve the unit to be used for a measurement, if any
+    pub fn get_unit(&self, measurement: &str) -> Option<&str> {
+        self.units.get(measurement).map(|x| x.as_str())
+    }
+
+    /// Retrieve the units to be used for a measurement group, if any
+    pub fn get_group_units(&self, group: &str) -> Option<&Units> {
+        self.groups.get(group)
+    }
 }
 
 #[cfg(test)]
@@ -77,8 +146,13 @@ mod tests {
         let timestamp = datetime!(2021-04-08 0:00:0 +05:00);
 
         let entity = CloudEntityMetadata::new("foo".into(), EntityMetadata::main_device(None));
-        let output =
-            from_thin_edge_json_with_timestamp(single_value_thin_edge_json, timestamp, &entity, "");
+        let output = from_thin_edge_json_with_timestamp(
+            single_value_thin_edge_json,
+            timestamp,
+            &entity,
+            "",
+            None,
+        );
 
         let expected_output = json!({
             "time": timestamp
@@ -114,8 +188,13 @@ mod tests {
         let timestamp = datetime!(2021-04-08 0:00:0 +05:00);
 
         let entity = CloudEntityMetadata::new("foo".into(), EntityMetadata::main_device(None));
-        let output =
-            from_thin_edge_json_with_timestamp(single_value_thin_edge_json, timestamp, &entity, "");
+        let output = from_thin_edge_json_with_timestamp(
+            single_value_thin_edge_json,
+            timestamp,
+            &entity,
+            "",
+            None,
+        );
 
         let expected_output = json!({
             "time": timestamp
@@ -160,7 +239,7 @@ mod tests {
                   }"#;
 
         let entity = CloudEntityMetadata::new("foo".into(), EntityMetadata::main_device(None));
-        let output = from_thin_edge_json(single_value_thin_edge_json, &entity, "");
+        let output = from_thin_edge_json(single_value_thin_edge_json, &entity, "", None);
 
         assert_eq!(
             expected_output.split_whitespace().collect::<String>(),
@@ -183,8 +262,13 @@ mod tests {
         let timestamp = datetime!(2021-04-08 0:00:0 +05:00);
 
         let entity = CloudEntityMetadata::new("foo".into(), EntityMetadata::main_device(None));
-        let output =
-            from_thin_edge_json_with_timestamp(multi_value_thin_edge_json, timestamp, &entity, "");
+        let output = from_thin_edge_json_with_timestamp(
+            multi_value_thin_edge_json,
+            timestamp,
+            &entity,
+            "",
+            None,
+        );
 
         let expected_output = json!({
             "time": timestamp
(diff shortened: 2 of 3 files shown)


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -190,6 +191,8 @@ pub struct CumulocityConverter {
 
     supported_operations: SupportedOperations,
     pub operation_handler: OperationHandler,
+
+    units: Units,
```
albinsuresh, line 195: Wondering if we should keep a `meta` wrapper at this level which internally contains the unit definitions, instead of keeping `units` directly here, since units is just one aspect of the metadata.

crates/extensions/c8y_mapper_ext/src/json.rs
```diff
@@ -49,12 +51,79 @@ fn from_thin_edge_json_with_timestamp(
...
+                units.set_unit(k, v);
+            }
+        }
+        units
+    }
+
+    pub fn set_unit(&mut self, measurement: String, meta: serde_json::Value) {
+        if let serde_json::Value::String(unit) = meta {
+            // "Temperature": "°C"
+            self.units.insert(measurement, unit);
```
albinsuresh: Not quite sure about this case where we assume the value to be units, if we expect `meta` to be expanded beyond units in future. But, if we're pretty confident that `unit` is the most reasonable "default" `meta` for a measurement, then this is fine.

crates/extensions/c8y_mapper_ext/src/json.rs
```diff
@@ -49,12 +51,79 @@ fn from_thin_edge_json_with_timestamp(
...
+    let mut serializer = serializer::C8yJsonSerializer::new(timestamp, entity, m_type, units);
     parse_str(input, &mut serializer)?;
     Ok(serializer.into_string()?)
 }
 
+/// Units used for measurements of a given type
+#[derive(Default)]
+pub struct Units {
+    units: HashMap<String, String>,
+    groups: HashMap<String, Units>,
```
albinsuresh:
```suggestion
    group_units: HashMap<String, Units>,
```

crates/extensions/c8y_mapper_ext/src/converter.rs
didier-wenzek: Possibly. However, this can easily be added later when actually required.

crates/extensions/c8y_mapper_ext/src/json.rs
didier-wenzek: Yeah. Better to remove support for this shortcut. Otherwise we will have to support it sooner or later, even if not documented ;-)

crates/extensions/c8y_mapper_ext/src/json.rs
didier-wenzek: done https://github.com/thin-edge/thin-edge.io/pull/3776/commits/5b0032d283ab95e49602ddc4239333206fa7c842

crates/extensions/c8y_mapper_ext/src/json.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3776/commits/c7c208d50ddbc7d0f00af76588821dc310696869

crates/extensions/c8y_mapper_ext/src/json.rs
```diff
@@ -49,12 +51,84 @@ fn from_thin_edge_json_with_timestamp(
...
+    ///
+    /// This is the default when no measurement metadata is published for a measurement topic
+    pub fn new() -> Units {
+        Units {
+            units: HashMap::new(),
+            group_units: HashMap::new(),
+        }
+    }
+
+    /// True if not units are actually defined
```
Bravo555: todo: typo
```suggestion
    /// True if no units are actually defined
```

crates/extensions/c8y_mapper_ext/src/json.rs
```diff
@@ -49,12 +51,84 @@ fn from_thin_edge_json_with_timestamp(
...
+        if let Some(unit) = meta.get("unit") {
+            // "Temperature": {"unit": "°C"},
+            if let serde_json::Value::String(unit_name) = unit {
+                self.units.insert(measurement, unit_name.to_owned());
+            }
+        } else {
+            // "Climate": { "Temperature": {"unit": "°C"}, "Humidity": {"unit": "%RH"} }
+            let group = measurement;
+            self.set_group_units(group, meta);
+        }
```
Bravo555: question: So, for a payload like `{"unit: "u", "Temperature": {"unit": "°C"}, "Humidity": {"unit": "%RH"}}` we only grab the `"unit"` and ignore the others?

I think we should only allow only one and print an error and don't add any units if both are present, ideally also adding this case to the tests. Or really, for any garbage data that can't be parsed published on these topics we should emit an ERROR event?

Also perhaps TRACE for when we actually set a new unit, but it's not as important.

crates/extensions/c8y_mapper_ext/src/serializer.rs
Bravo555, line 1: note: No tests changes in this file because the serializer is called by tests in `json.rs`

Without many comments it's a bit hard to understand but it seems correct and the tests in `json.rs` pass, so evidently nothing's broken.

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -378,6 +378,81 @@ c8y/measurement/measurements/create
...
+The Cumulocity mapper uses the measurement meta topics to attach units to measurement values.
+
+The metadata for a measurement topic:
+
+```mermaid
+graph LR
+  te --/--- entity_id["&lt;entity id&gt;"] --/--- m --/--- measurement_type["&lt;measurement type&gt;"]
+```
+
+are to by published to the associated meta topic:
```
rina23q: I think it's typo.

```suggestion
are to be published to the associated meta topic:
```

docs/src/references/mappers/c8y-mapper.md
```diff
@@ -378,6 +378,81 @@ c8y/measurement/measurements/create
...
+  "time":"2025-09-03T10:05:47.226Z"
+}
+```
+
+- A message received on `te/device/main///m/<type>` uses the units defined on `te/device/main///m/<type>/meta`, if any.
+- If the unit for a measurement is unknown, the measurement value is simply sent with no unit.
+- Other metadata such as the precision or the min and max values can be attached to a measurement.
+  However, these are ignored by the Cumulocity mapper.
+- `"Temperature"`, `"Climate": { "Temperature" }` and `"Engine": { "Temperature" }` can be given different units.
+
```
rina23q, line 458: I think it would be nice to mention how to clear the metadata with one sentence. Something like `To clear the metadata,  publish an empty retained message to the topic.`

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -213,6 +214,9 @@ impl C8yMapperConfig {
         topics.add_all(mqtt_schema.topics(AnyEntity, AnyCommand));
         topics.add_all(mqtt_schema.topics(AnyEntity, AnyCommandMetadata));
 
+        // Add measurement metadata topic
+        topics.add_all(mqtt_schema.topics(AnyEntity, MeasurementMetadata));
```
rina23q: I think this one should be added to the tedge config `c8y.topics` since it has `te/+/+/+/+/a+` settings.

didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3776/commits/bd8892c66b6a6cface51f3a462ce6c82a2a45a2d

docs/src/references/mappers/c8y-mapper.md
didier-wenzek: Added https://github.com/thin-edge/thin-edge.io/pull/3776/commits/20f874ac440b087954aed8b2114aaacda8c2de4c

docs/src/references/mappers/c8y-mapper.md
didier-wenzek: Fixed https://github.com/thin-edge/thin-edge.io/pull/3776/commits/20f874ac440b087954aed8b2114aaacda8c2de4c

crates/extensions/c8y_mapper_ext/src/json.rs
didier-wenzek: There is a tension here between flexibility and adherence to thin-edge JSON conventions.

- We want the freedom to add not only units but also other info about measurements such as precision as with `"temperature": { "unit": "°C", "precision": "±0.1°C" }`
- We want to  be able to use different units for  `"Temperature"`, `"Climate": { "Temperature" }` and `"Engine": { "Temperature" }`, e.g. with `"Climate": { "Temperature": {"unit": "°C"} }`

However, this makes things a bit complicated. As raised by @Bravo555, `"Foo": {"unit: "u", "Temperature": {"unit": "°C"}}` can then be understood as metadata assigning units to `"Foo":20` values as well as `"Foo" : { "Temperature": 20 }` (the current code ignoring the latter, though).

An alternative is to flatten the structure of (measurement, info) pairs where group measurements are identified by a path. 

```
 "Temperature": { "unit": "°C", "precision": "±0.1°C" },
 "Climate.Temperature": { "unit": "°C", "precision": "±0.1°C" },
 "Engine.Temperature": { "unit": "°F", "precision": "±1°F" },
```

=> I will give a try to this idea.

crates/extensions/c8y_mapper_ext/src/serializer.rs
didier-wenzek: The main reason I didn't add unit tests here is that the serializer API is really low level with a method call per json field. Working on the user input, i.e. json, is less subject to changes.

crates/extensions/c8y_mapper_ext/src/json.rs
didier-wenzek: > => I will give a try to this idea.

https://github.com/thin-edge/thin-edge.io/pull/3776/commits/7ddb8ed36ad8b6021c41514a39ab72db8b10eaa7

crates/extensions/c8y_mapper_ext/src/json.rs
```diff
@@ -93,15 +93,20 @@ impl Units {
...
+                None => {
+                    // "Temperature": {"unit": "°C"},
+                    self.units.insert(measurement, unit.to_owned());
+                }
+                Some((group, measurement)) => {
+                    // "Climate.Temperature": {"unit": "°C"},
+                    self.group_units
+                        .entry(group.to_owned())
+                        .or_default()
+                        .set_unit(measurement.to_owned(), meta);
```
Bravo555, line 107: question: If the messages have flat key-value structures, with no nesting, do we still need separate hashmaps for group units? Can't we put all the keys in a single hashmap?

didier-wenzek: Indeed, not required. This would simplify the mapping and its memory footprint. However this would also implies building paths (concatenating group and measurement names) each time a grouped measurement value is received.

Bravo555 (approved):

tests/RobotFramework/tests/cumulocity/telemetry/thin-edge_device_telemetry.robot
```diff
@@ -20,6 +20,99 @@ Thin-edge devices support sending simple measurements
...
+    Log    ${measurements}
+    Should Be Equal    ${measurements[0]["temperature"]["temperature"]["unit"]}    °C
+    Should Be Equal    ${measurements[1]["temperature"]["temperature"]["unit"]}    °F
+
+Measurement units can be cleared
+    # Using a test-specific measurement is required to not interfer with other tests
+    Execute Command    tedge mqtt pub -r te/device/main///m/t6/meta '{ "temperature": { "unit": "°C" } }'
+    Execute Command    tedge mqtt pub te/device/main///m/t6 '{ "temperature": 25 }'
+    Execute Command    tedge mqtt pub -r te/device/main///m/t6/meta ''
+    Execute Command    tedge mqtt pub te/device/main///m/t6 '{ "temperature": 298.15 }'
```
reubenmiller: @didier-wenzek Do you think this might introduce a flakey test if the meta topic is not processed before receiving the second temperature measurement?

reubenmiller: If so, it might be worth doing two measurement assertions (one to check that the units are present, and the second to check that the units are not present)

didier-wenzek: > @didier-wenzek Do you think this might introduce a flakey test if the meta topic is not processed before receiving the second temperature measurement?

Good point. The mapper sequentially processes all the measurement related messages. But the broker might change the message order as they are not published on the same topic.

didier-wenzek: Actually, not flaky.

```
$ invoke flake-finder --test-name "Measurement units can be cleared" --iterations 100 --outputdir output_ff --clean
Overall: PASSED
Elapsed time: 0:20:33.596564

```

didier-wenzek: I will remove that useless commit when re-basing https://github.com/thin-edge/thin-edge.io/pull/3776/commits/92c40f6f67c940c8c18420bd00f9573b594bbe40

reubenmiller: Yeah I was also surprised that it didn't seem to be flakey, though I pushed a slightly refactored test that also passed 50 flake finder runs (and the test removes the potential race condition)

commit: 40bea55959ca1cd20bd95876c55d3bfa67f404a1

albinsuresh:
To ease the discoverability of this feature, wouldn't it be better to mention the metadata support in https://github.com/thin-edge/thin-edge.io/blob/main/docs/src/start/send-measurements.md as well, rather than "hidden" in a few reference docs?

didier-wenzek:
> To ease the discoverability of this feature, wouldn't it be better to mention the metadata support in https://github.com/thin-edge/thin-edge.io/blob/main/docs/src/start/send-measurements.md as well, rather than "hidden" in a few reference docs?

Good point. I will push a PR to improve that.

Issue #510, opened 2021-10-15
Persitence Service Addon (DB based buffering)


marco2901:
Some Devices are running in very network weak regions or even have only a cyclic transmission rate (e.g. once per day). For those devices it would be good to have an persistence service, which is storing data (not just in-memory) over a longer period and flush them at connection time.

It would be cool to have addons for connecting databases like influx, mongo or round-robin mechanisms based ones, which then could take care about data persistence during disconnected operation time.

phcrb:
Hi @marco2901, @didier-wenzek,
I'm planning to deploy some gateways with thin-edge in places with potentially no wired connection available, so transmission will be performed by a 3G/4G cellular link. The devices attached to the gateway typically send measurements every 5 to 15 minutes. A mecanism to store data locally when the cellular link is not available would be strongly apreciated.

My gateway does not run a debian-like os, but rather a custom 32 bit linux with a simple busybox and no apt facility, built by the hardware manufacturer, so it's nearly impossible to install a 
complex database engine such as mongo or influxdb as they may need many dependencies. A simple file-based storage buffer would be more adapted to my use case.

didier-wenzek:
A first step is to [configure mosquitto](https://mosquitto.org/man/mosquitto-conf-5.html) to enable persistence.

For instance, to store on disk all the pending messages every 60 seconds, `/etc/mosquitto/mosquitto.conf` can be updated with:

```
persistence true
persistence_location /var/lib/mosquitto/
autosave_interval 60
autosave_on_changes false
```

If you have few but important messages, you can persist the pending messages more frequently or when x new messages are pending. Say every 5 messages:

```
persistence true
persistence_location /var/lib/mosquitto/
autosave_interval 5
autosave_on_changes true
```

For a more robust solution, using a local small foot print database, one also needs some kind of application acknowledgement from the cloud peer telling the device when these messages have been successfully transmitted and can be safely removed from the local database.

PR #3779, opened 2025-09-12
poc: rebase tedge mea db onto main

jarhodes314:
## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/510

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/actor.rs
+++ b/crates/extensions/tedge_flows/src/actor.rs
@@ -44,10 +44,7 @@ impl Actor for FlowsMapper {
         while let Some(message) = self.next_message().await {
             match message {
                 InputMessage::Tick(_) => {
-                    let drained_messages = self.drain_db().await?;
-                    self.on_messages(MessageSource::MeaDB, drained_messages)
-                        .await?;
-
+                    self.poll_ready_sources().await?;
                     self.on_interval().await?;
                 }
                 InputMessage::MqttMessage(message) => match Message::try_from(message) {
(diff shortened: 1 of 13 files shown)


crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -36,12 +39,15 @@ impl Actor for FlowsMapper {
         loop {
             tokio::select! {
                 _ = interval.tick() => {
+                    let drained_messages = self.drain_db().await?;
+                    self.on_messages(MessageSource::MeaDB, drained_messages).await?;
```
didier-wenzek: This design is a legacy of the POC and is actually not so neat.

What is wrong, is that *all* the flows are requested for messages drained from the db (in `self.drain_db()`), and then these messages are produced to *all* the flows (in `self.on_messages()`) while at most one flow is interested by each: the former source flow!  This is:
- inefficient: messages drained from the DB are even produced to flows unrelated to the MEA DB
- complicated: each message must be attached a source such as `MessageSource::MeaDB`
- not extensible: a new method is required for each kind of message sources (similar to `self.drain_db()`).

A better approach would be to add a `on_interval()` method on the sources themselves. Only the messages produced by the source of a flow will then be processed by the flow. This can even be improved with `MessageSource` trait. But the key point is that the flow actor has no more to be updated for each new source kind.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -58,8 +85,17 @@ pub enum FlowError {
...
 
 impl Flow {
+    pub fn accept(&self, source: MessageSource, message_topic: &str) -> bool {
+        match &self.input {
+            FlowInput::Mqtt {
+                topics: input_topics,
+            } => source == MessageSource::Mqtt && input_topics.accept_topic_name(message_topic),
+            FlowInput::MeaDB { .. } => source == MessageSource::MeaDB,
+        }
+    }
```
didier-wenzek, line 102: With the proposed design, a flow is no more requested to `accept` or not a message that has been produced by a random source. A flow will only have to process messages produced by its *own* source.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -161,29 +198,33 @@ impl FlowStep {
...
-            // 0 as a default is not appropriate for a script with an onInterval handler
-            script.interval_secs = 1;
+        if !script.no_js_on_interval_fun && script.interval.is_zero() {
+            // Zero as a default is not appropriate for a script with an onInterval handler
+            script.interval = std::time::Duration::from_secs(1);
         }
     }
 }
 
 impl FlowInput {
```
didier-wenzek, line 230: The main idea of the proposed design is to let the source generates messages on some configured interval.

```
impl FlowInput {
    pub async fn on_interval(
        &mut self,
        timestamp: &DateTime,
      match self {
            FlowInput::Mqtt { topics } => Ok(vec![]),
            FlowInput::MeaDB { .. } => {
                drain_db().await
            }
        }
    }
}
```

didier-wenzek (commented):
The current design must be revisited before adding new kind of message sources.

PR #3781, opened 2025-09-15
feat: support for flows operating on binary payloads 

didier-wenzek:
## Proposed changes

Let flow step functions operate on binary payload.

- [x] [TextDecoder](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder)
    - [x] utf-8
    - [ ] ~utf-16be, utf16le~
    - [x] ignore BOM
    - [x] substitute malformed data with a replacement character
    - [x] decode(buffer)
- [x] [TextEncoder](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder)
    - [x] utf-8
    - [x] encode()
    - [x] encodeInto()
- [x] Step functions processing and producing messages with binary payloads
- [x] Improve `tedge mqtt` with an `--base64` option to ease pub/sub of binary payloads
- [x] Improve `tedge flows test` with an `--base64` option to test flows of binary payloads

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/docs/src/references/mappers/flows.md
+++ b/docs/src/references/mappers/flows.md
@@ -71,12 +71,14 @@ interface FlowStep {
 }
 ```
 
-A message contains the message topic and payload as well as an ingestion timestamp: 
+A message contains the message topic and payload as well as an ingestion timestamp.
+The bytes of the raw message payload are also accessible as an array of unsigned bytes: 
 
 ```ts
 type Message = {
   topic: string,
   payload: string,
+  raw_payload: Uint8Array,
   timestamp: Timestamp
 }
 
@@ -91,7 +93,7 @@ These values are configured by the flow and can be dynamically updated on recept
 
 The `onMessage` function is called for each message to be transformed
   - The arguments passed to the function are:
-    - The message `{ topic: string, payload: string, timestamp: { seconds: u64, nanoseconds: u32 } }`
+    - The message `{ topic: string, payload: string, raw_payload: Uint8Array, timestamp: { seconds: u64, nanoseconds: u32 } }`
     - The config as read from the flow config or updated by the script
   - The function is expected to return zero, one or many transformed messages `[{ topic: string, payload: string }]`
   - An exception can be thrown if the input message cannot be transformed.
@@ -186,3 +188,17 @@ $ tedge flows test collectd/mandarine/cpu/percent-active '1754571280.572:2.07156
 
 [c8y/measurement/measurements/create] {"type":"collectd","time":"2025-08-07T12:54:40.572Z","cpu":{"percent-active":2.07156308851224}}
 ```
+
+## Builtin Objects
+
+%%te%% flows uses the [QuickJS](https://bellard.org/quickjs/) engine and supports [ECMAScript® 2023](https://tc39.es/ecma262/2023/).
+
+Flows and step functions are executed in a sandbox with no access to the system, the disk or the network.
+
+The following builtin objects are exported:
+
+- [`console.log()`](https://developer.mozilla.org/en-US/docs/Web/API/console/log_static)
+  - Output log messages to the mapper log
+- [`TextDecoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder)
+  - Only `utf-8` is supported 
+- [`TextEncoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder)


crates/extensions/tedge_flows/src/js_value.rs
```diff
@@ -0,0 +1,315 @@
...
+    fn from(value: Message) -> Self {
+        let raw_payload = JsonValue::Bytes(value.payload.clone());
+        let payload = match String::from_utf8(value.payload) {
+            Ok(utf8) => JsonValue::string(utf8),
+            Err(_) => JsonValue::Null,
+        };
+        JsonValue::object([
+            ("topic", JsonValue::string(value.topic)),
+            ("payload", payload),
+            ("raw_payload", raw_payload),
```
didier-wenzek, line 64: I'm not convinced by this approach which is to pass to jS both a binary payload and an utf8 string with the same content.

Starting with this hack to have something working end to end. Alternatives have to be explored.

tests/RobotFramework/tests/tedge/tedge_mqtt.robot
```diff
@@ -45,6 +45,18 @@ tedge mqtt sub breaks after receiving first non-retailed message
...
+    Execute Command
+    ...    tedge mqtt pub -r --base64 te/device/main///m/ eyAidGVtcGVyYXR1cmUiOiAzMC4yLCAiaHVtaWRpdHkiOiAyMyB9
+    ${messages}=    Execute Command    tedge mqtt sub te/device/main///m/ --count 1
+    Should Contain    ${messages}    { "temperature": 30.2, "humidity": 23 }
+
+receiving binary payload with tedge mqtt
+    Execute Command
+    ...    tedge mqtt pub -r te/device/main///m/ '{ "temperature": 30.2, "humidity": 23 }'
+    ${messages}=    Execute Command    tedge mqtt sub --base64 te/device/main///m/ --count 1
+    Should Contain    ${messages}    eyAidGVtcGVyYXR1cmUiOiAzMC4yLCAiaHVtaWRpdHkiOiAyMyB9
```
Bravo555, line 58: note: Because the payload is the same between two tests I thought there was a dependency from the first test to the second, but only when I was successful I realized that `sub --base64` converts all payloads to base64 representation.

I see in the second test suite we do something like:
```
    VAR     ${payload}  { "temperature": 30.2, "humidity": 23 }
    Execute Command
    ...    tedge mqtt pub -r te/device/main///m/ '${payload}'
    ${messages}=    Execute Command    tedge mqtt sub --base64 te/device/main///m/ --count 1 --no-topic | base64 -d -    strip=True
    Should Contain    ${messages}    ${payload}
```

Why the choice of raw base64 here?

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -83,27 +86,38 @@ async fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
             }
         };
 
+        if cmd.retained_only && !message.retain {
+            info!(target: "MQTT", "Received first non-retained message. topic={}", message.topic.name);
```
Bravo555: suggestion: can also record fields like this
```suggestion
            info!(target: "MQTT", topic = message.topic.name, "Received first non-retained message");
```

crates/core/tedge/src/cli/flows/test.rs
```diff
@@ -87,7 +91,7 @@ fn print(messages: Result<Vec<Message>, FlowError>) {
     }
 }
 
-fn parse(line: String) -> Result<Option<Message>, Error> {
+fn parse(base64_input: bool, line: String) -> Result<Option<Message>, Error> {
```
Bravo555: todo: Fine for prototyping, but otherwise a bit of a bad smell for how many times we add a boolean arg for this, will probably require some cleaup here before merge.

crates/extensions/tedge_flows/src/js_lib/text_decoder.rs
```diff
@@ -0,0 +1,82 @@
...
+
+        let text = if self.fatal {
+            std::str::from_utf8(bytes)
+                .map_err(|err| Exception::throw_message(&ctx, &err.to_string()))?
+                .to_string()
+        } else {
+            match String::from_utf8_lossy(bytes) {
+                Cow::Owned(text) => text,
+                Cow::Borrowed(text) => text.to_owned(),
+            }
```
Bravo555: suggestion: if we always want owned, just use `.to_string()

```suggestion
            String::from_utf8_lossy(bytes).to_string()
```

crates/extensions/tedge_flows/src/js_lib/text_encoder.rs
```diff
@@ -0,0 +1,44 @@
...
+                    if let Some(bytes) = object.as_typed_array::<u8>() {
+                        return Ok(bytes.clone());
+                    }
+                }
+                "".to_string()
+            }
+            Some(js_string) => js_string.to_string()?,
+        };
+        TypedArray::new(ctx.clone(), string.as_bytes())
+    }
```
Bravo555: suggestion: found it a bit hard to follow the logic, so tried to simplify it a bit. I'm assuming that if `.as_object()` returns `Some()`, then `.as_string()` can't return `Some()` as well (i.e. the value can't be a string and an object simultaneously)

```suggestion
    pub fn encode(&self, ctx: Ctx<'js>, value: Value<'js>) -> Result<TypedArray<'js, u8>> {
        // if we already have a bytearray, return it
        if let Some(bytes) = value.as_object().and_then(|o| o.as_typed_array::<u8>()) {
            return Ok(bytes.clone());
        }

        // otherwise encode if it's a js string
        let string = value
            .as_string()
            .map(|s| s.to_string())
            .transpose()?
            .unwrap_or_default();
        TypedArray::new(ctx.clone(), string.as_bytes())
    }
```

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -234,17 +248,35 @@ impl Message {
...
+impl std::fmt::Display for Message {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        '['.fmt(f)?;
+        self.topic.fmt(f)?;
+        "] ".fmt(f)?;
+        match &self.payload_str() {
+            Some(str) => str.fmt(f),
+            None => format!("{:?}", self.payload).fmt(f),
+        }
+    }
```
Bravo555, line 255: suggestion: avoids allocating and is a little simpler

```suggestion
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "[{}]", self.topic)?;
        match &self.payload_str() {
            Some(str) => write!(f, "{str}"),
            None => write!(f, "{:?}", self.payload),
        }
    }
```

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -218,12 +218,26 @@ impl TryFrom<OffsetDateTime> for DateTime {
...
 impl Message {
+    pub fn new(topic: impl ToString, payload: impl Into<Vec<u8>>) -> Self {
+        Message {
+            topic: topic.to_string(),
+            payload: payload.into(),
+            timestamp: None,
+        }
+    }
+
+    #[cfg(test)]
```
Bravo555, line 229: thought: realized that multiple impl blocks are nice for these test methods, e.g.:

```
impl Message {
    // normal methods
}

#[cfg(test)]
impl Message {
    pub fn sent_now(mut self) -> Self {
        self.timestamp = Some(DateTime::now());
        self
    }

    pub(crate) fn new_binary(topic: &str, payload: impl Into<Vec<u8>>) -> Self {
        Message {
            topic: topic.to_string(),
            payload: payload.into(),
            timestamp: None,
        }
    } 
}
```

tests/RobotFramework/tests/tedge/tedge_mqtt.robot
didier-wenzek: > Why the choice of raw base64 here?

From a user perspective, this is indeed useless: the payloads used in these two tests are actually regular utf8 strings and no encoding is required.

The purpose is only to test the `base64` of `tedge mqtt pub` and `tedge mqtt sub`. The former is expected to decode the encoded payload before publishing it. The latter is expected to encode the received message payloads before printing those.

crates/core/tedge/src/cli/flows/test.rs
didier-wenzek: Yeah. I would not bother so much as this is only used only private methods. I will see what can be improved, though.

crates/core/tedge/src/cli/mqtt/subscribe.rs
didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3781/commits/e926d2ffd5c15af305b46df1228e99f0e9616558

crates/extensions/tedge_flows/src/js_lib/text_decoder.rs
didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3781/commits/2ee0de776ddf2faa9028b5b45ee09f84514a7ed7

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3781/commits/7ff76048a9e1b15d8095846240383e2afb43f201 and oops https://github.com/thin-edge/thin-edge.io/pull/3781/commits/93a9794d6ff540760d15b36f547f695a580e4aa4

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: Yeah, not sure this is better

crates/extensions/tedge_flows/src/js_lib/text_encoder.rs
didier-wenzek: That one is indeed over-complicated. Looking closer to the [TextDecode specs](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder), both `encode` and `encodeInto` take a javascript string for input. Hence, all this coercion code is simply useless. See https://github.com/thin-edge/thin-edge.io/pull/3781/commits/93014c560fd3c2dffb630fbbf828e9a3ec7a95a9

crates/core/tedge/src/cli/flows/test.rs
didier-wenzek: Indeed, this simplifies the code at lot. https://github.com/thin-edge/thin-edge.io/pull/3781/commits/33fd1a9ff23f8b6d4587aa930a94d7beb6636f2f

crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -174,7 +175,9 @@ impl JsWorker {
 
     async fn run(mut self) {
         rquickjs::async_with!(self.context => |ctx| {
-            console::init(&ctx);
+            js_lib::console::init(&ctx);
+            js_lib::text_decoder::init(&ctx);
+            js_lib::text_encoder::init(&ctx);
```
jarhodes314, line 180: Is there a reason why we're doing this twice, or was this just a typo?

didier-wenzek: Two different classes are registered: decoder vs encoder.

That said, it might be good to have here a single call to a global library registration (say `js_lib::init(&ctx)` so only the `js_lib` module would have to be updated when adding new global values and classes.

jarhodes314: Ah, yes, this was just me being blind and not spotting the encoder/decoder difference

Bravo555:
> So I propose the following:
> 
>     With the --base64 option, tedge mqtt sub encodes all the received payloads
>     Without the --base64 option, tedge mqtt sub encodes any payload that is not utf8, but makes this explicit.
> 
> I think this improves the UX, as there is no more useless error messages (such as "invalid utf-8 sequence of 1 bytes from index 3"):
> 
> ```
> $ tedge mqtt sub 'te/#'
> [te/device/main///m/] {"temperature": 20}
> [te/device/main///m/protobuf] <ERR=NON-UTF8> ChIJAAAAAAAAPUARAAAAAACASUA=
> [te/device/main///m/protobuf]  
> 	=@I@
> ```
> 
> However, note that the 3rd message is by chance an UTF-8 valid sequence, while this a protobuf valid message for { "temperature": 29, "humidity": 51 }. The 2nd message is also a protobuf message but for { "temperature": 29, "humidity": 50 }. One single digit difference and the behavior is opposite.
> 
> For a reliable output, the --base64 option can be used:
> 
> ```
> $ tedge mqtt sub --base64 'te/#'
> [te/device/main///m/] eyAiVGVtcGVyYXR1cmUiOiAyNywgIkh1bWlkaXR5IjogNjcgfQ==
> [te/device/main///m/protobuf] ChIJAAAAAAAAPUARAAAAAACASUA=
> [te/device/main///m/protobuf] ChIJAAAAAAAAPUARAAAAAAAASUA=
> ```

An alternative would be to use [`slice::escape_ascii`](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.escape_ascii) function to print ASCII printable characters, but escape other characters.

Using this snippet:

```rust
        let payload = if cmd.base64 {
            BASE64_STANDARD.encode(message.payload_bytes()).into_bytes()
        } else {
            message.payload.as_bytes().to_vec()
        };
        let payload = payload.escape_ascii();

        let line = if cmd.hide_topic {
            format!("{payload}\n")
        } else {
            format!("[{}] {payload}\n", &message.topic)
        };
```

Base64 sequence `ChIJAAAAAAAAPUARAAAAAACASUA=` is printed as the following:

```
[te/device/main///m/protobuf] \n\x12\t\x00\x00\x00\x00\x00\x00=@\x11\x00\x00\x00\x00\x00\x80I@
```

This has the advantage of printing regular printable ASCII characters if there are any, so if e.g. there's only one character that can't be displayed, the rest of the message can still be displayed. For humans who run `tedge mqtt sub` this is a better UX. We also don't have to treat differently messages that are valid UTF-8 but contain non-printable characters, vs invalid UTF-8; all get printed escaped.

Bravo555 (approved):
No blockers left. Changing the formatting of binary messages is optional - I think it looks nicer and gives better UX, but we can proceed with the current formatting using base64.

reubenmiller:
> This has the advantage of printing regular printable ASCII characters if there are any, so if e.g. there's only one character that can't be displayed, the rest of the message can still be displayed. For humans who run `tedge mqtt sub` this is a better UX. We also don't have to treat differently messages that are valid UTF-8 but contain non-printable characters, vs invalid UTF-8; all get printed escaped.

Though the problem with this proposal is that is not very usable...as rarely printing a portion of a binary message is useful to the user, and it just makes it harder to pass to other applications, e.g. copy/pasting the message and then using it with `base64 -d` and saving it to file.

crates/extensions/tedge_flows/src/js_value.rs
```diff
@@ -0,0 +1,315 @@
...
+use std::collections::BTreeMap;
+
+/// Akin to serde_json::Value with an extra case for binary data
+#[derive(Clone, Debug)]
+pub enum JsonValue {
+    Null,
+    Bool(bool),
+    Number(serde_json::Number),
+    String(String),
+    Bytes(Vec<u8>), // <= This case motivates the use of JsonValue vs serde_json::Value
```
albinsuresh, line 17: **Not a request for change**
I'm assuming that this was done to ease some impl aspects, but purely from the type perspective this feels slightly problematic as this is a "recursive enum" structure where the `Array` and `Object` variants having self-references to the same `JsonValue` type seems to suggest that the following payload is also allowed:
```
{
   "some_key": <binary-payload>
}
```

In my mind it should have been something like this:
```
enum Payload {
   Bytes(Vec<u8>),
   Json(serde_json::Value)
}
```
But I'll admit that I don't understand the js engine deep enough to say if this is even feasible.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -218,12 +218,26 @@ impl TryFrom<OffsetDateTime> for DateTime {
...
+    pub fn sent_now(mut self) -> Self {
+        self.timestamp = Some(DateTime::now());
+        self
+    }
+
     #[cfg(test)]
-    pub(crate) fn new(topic: &str, payload: &str) -> Self {
+    pub(crate) fn new_binary(topic: &str, payload: impl Into<Vec<u8>>) -> Self {
```
albinsuresh: This `new_binary` seems redundant with the `new` function above, as they both essentially accept `impl Into<Vec<u8>>` payload. I could just swap the usages of this one with the other one.

didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3781/commits/29f774d68916ab8e7d9fce01f6059f87300cecf6

crates/extensions/tedge_flows/src/js_value.rs
didier-wenzek: > This seems to suggest that the following payload is also allowed:
> ```
> {
>   "some_key": <binary-payload>
> }
> ```

Yes this is the goal. For instance a message is translated in an object with a string topic and a binary payload.

crates/extensions/tedge_flows/src/flow.rs
albinsuresh: I guess you forgot to remove that constructor itself, which is now unused ;)

didier-wenzek: Oops https://github.com/thin-edge/thin-edge.io/pull/3781/commits/985c37fa70016aa149930884b235b911f812d42c

PR #3782, opened 2025-09-15
docs: Mention measurement units in the starting guide

didier-wenzek:
## Proposed changes

Mention measurement units in the starting guide

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/pull/3776#issuecomment-3290681704

## Checklist


docs/src/start/send-measurements.md
```diff
@@ -87,6 +87,29 @@ The `time` field is not a regular measurement like `temperature` or `pressure` b
...
+    "pressure": { "unit": "bar" },
+    "current.L1": { "unit": "A" },
+    "current.L2": { "unit": "A" },
+    "current.L3": { "unit": "A" }
+}'
+```
+
+Note the `--retain` flag. This marks the message as retained making it available to all applications
+not only for the subsequent measurements but also across application restarts.
+
```
albinsuresh, line 112: I don't know what's the best way to convey that tedge itself does not limit this additional metadata payload to `unit` values alone, but accepts other values as well, but only `unit` fields are processed by the **c8y mapper**. Is that distinction worth highlighting?

docs/src/start/send-measurements.md
```diff
@@ -87,6 +87,29 @@ The `time` field is not a regular measurement like `temperature` or `pressure` b
...
+For details, refer to [%%te%% JSON Specification](../../references/mappers/c8y-mapper/#measurement-units)
+
```
albinsuresh, line 114: I'm a bit confused about what' the right place to document `measurement` API extensions like these. I thought that this would be the right place, but then saw that the `properties` key for inline metadata support added in #3774 is documented in `docs/src/understand/thin-edge-json.md` and not here. There's quite a bit of overlap between that file and the contents of this file along with its siblings `send-events.md`, `raise-alarm.md` etc. So, not sure what's the right target. If these `getting-started` files are meant to be short with only the minimal interface intro and the `understand` one is supposed document the details/extensions, then perhaps this new doc content also belongs there.

docs/src/start/send-measurements.md
didier-wenzek: As this is a starting guide, I think it's better to not overload the reader. This is why I have been silent about other properties, especially since those are currently used nowhere.

docs/src/start/send-measurements.md
didier-wenzek: > So, not sure what's the right target

Not sure either. We have two types of content. On one side, we describe the format and  content of MEA messages and their metadata. On the other side, we tell how this content is interpreted and used by the mappers and more specifically the Cumulocity mapper.

docs/src/start/send-measurements.md
```diff
@@ -87,6 +87,29 @@ The `time` field is not a regular measurement like `temperature` or `pressure` b
...
+    "temperature": { "unit": "°C" },
+    "pressure": { "unit": "bar" },
+    "current.L1": { "unit": "A" },
+    "current.L2": { "unit": "A" },
+    "current.L3": { "unit": "A" }
+}'
+```
+
+Note the `--retain` flag. This marks the message as retained making it available to all applications
+not only for the subsequent measurements but also across application restarts.
```
rina23q: Cosmetic suggestion

```suggestion
:::note 
The `--retain` flag marks the message as retained making it available to all applications
not only for the subsequent measurements but also across application restarts.
:::
```

docs/src/start/send-measurements.md
reubenmiller: Ok, I would recommend linking to [references/mqtt-api/#telemetry-type-metadata](https://thin-edge.github.io/thin-edge.io/next/references/mqtt-api/#telemetry-type-metadata) as in the context of the paragraph it is describing the registration of the measurement type info, and not the c8y mapping.

We can always add an additional link to the c8y mapping from the mqtt api details, but we can do that in a different PR.

didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3782/commits/1fcdb3e8b370a0d123701e1165eb4f4b40def1cd

docs/src/start/send-measurements.md
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3782/commits/49f62dd4461a51340decae3243a403ebc7388b11

reubenmiller (approved):
Approved, thanks for the addition which should make more users aware of the feature

Issue #3766, opened 2025-08-22
tedge-p11-server reports "Failed to find a private key"


ck-c8y:
**Describe the bug**

We run **tedge-p11-server**  (1.5.1) on the host and **tedge** (1.5.1) as podman container).

**tedge-p11-server** is started and configured with a pin:
```
[device.cryptoki]
pin = "1111111"
module_path = "/usr/lib/x86_64-linux-gnu/pkcs11/opensc-pkcs11.so"
```

**tedge** is running in a podman container with the following start script:
```
podman run -d \
--userns keep-id \
--name tedge \
--network tedge \
-p "1883:1883" \
-p "8000:8000" \
-p "8001:8001" \
-v "tedge-data:/data/tedge" \
-v "/run/tedge-p11-server:/run/tedge-p11-server" \
-v "/mnt/data/tedge/config:/local-conf" \
-v "/run/user/$UID/podman/podman.sock:/var/run/docker.sock:rw" \
-v "/mnt/data/home/tedge/mosquitto.conf:/etc/mosquitto/mosquitto.conf:rw" \
-e TEDGE_C8Y_OPERATIONS_AUTO_LOG_UPLOAD=always \
-e TEDGE_MQTT_BRIDGE_BUILT_IN=true \
-e TEDGE_DEVICE_CRYPTOKI_MODE=socket \
-e TEDGE_DEVICE_CERT_PATH=/local-conf/tedge-certificate.pem \
-e S6_CMD_WAIT_FOR_SERVICES_MAXTIME=$S6_CMD_WAIT_FOR_SERVICES_MAXTIME \
--env-file /mnt/data/tedge/config/wec-iot.cfg \
<OUR_REGISTRY>/thinedge:stable
```
Before phase 1 no cert is issued in `/mnt/data/tedge/config`, so **tedge** tries to reconnect.

The sequence of the following actions is as follows:

- phase 1: 
    - tedge-p11-server running, configured with user PIN
    - tedge-container running , new certificate arrived, after 30 seconds restarted manually
- phase 2: 
    - tedge-p11-server unchanged
    - tedge-container unchanged after previous restart
- phase 3: 
    - tedge-p11-server restarted
    - tedge-container restarted

In the logs of the **tedge-p11-server** the following error is reported:
```
tedge_p11_server::server: Incoming request failed: Failed to find a signing key
Aug 21 11:25:08 tedge-device tedge-p11-server[463]: Caused by:
[...]

PR #3785, opened 2025-09-18
improve: PKCS11 slots/tokens/objects logging

Bravo555:
## Proposed changes

When opening a session to a token as part of a request, all slots and tokens visible by the system will be printed.
We will also print URIs of all objects that we can see on the token.

NOTE: this PR also reverts "https://github.com/thin-edge/thin-edge.io/pull/3772", because this fix introduces TLS timeouts. For details see the commit message https://github.com/thin-edge/thin-edge.io/commit/7bcf8a91d7da11a4c5e7491063ff1ae48bda0190

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

This is a follow-up to #3766 that we thought was solved but is not. The issue can't be reliably reproduced, so we need to improve diagnostics for when it eventually reappears.

### DIFF
--- a/crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
+++ b/crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
@@ -136,7 +136,7 @@ impl Cryptoki {
     /// again ([C_GetSlotList]).
     ///
     /// [C_GetSlotList]: https://docs.oasis-open.org/pkcs11/pkcs11-base/v2.40/errata01/os/pkcs11-base-v2.40-errata01-os-complete.html#_Toc441755804
-    fn reinit(&self) -> anyhow::Result<()> {
+    fn _reinit(&self) -> anyhow::Result<()> {
         // load a new client before locking so if error we don't poison the mutex
         let new_client = Self::load(&self.config.module_path)?;
 
@@ -195,7 +195,9 @@ impl Cryptoki {
         let wanted_label = uri_attributes.token.as_ref();
         let wanted_serial = uri_attributes.serial.as_ref();
 
-        self.reinit()?;
+        // avoid reinitializing for now because of Nitrokey slowness
+        // self.reinit()?;
+
         let context = match self.context.lock() {
             Ok(c) => c,
             Err(e) => e.into_inner(),
--- a/crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
+++ b/crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
@@ -284,8 +284,7 @@ impl CryptokiSession<'_> {
         let session = &self.session;
 
         // get the signing key
-        let key =
-            Self::find_key_by_attributes(&self.uri_attributes, session, ObjectClass::PRIVATE_KEY)?;
+        let key = self.find_key_by_attributes(&self.uri_attributes, ObjectClass::PRIVATE_KEY)?;
         let key_type = session
             .get_attributes(key, &[AttributeType::KeyType])?
             .into_iter()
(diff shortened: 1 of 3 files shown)


crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -201,21 +235,128 @@ impl Cryptoki {
...
+    uri_attributes: uri::Pkcs11Uri<'a>,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum CryptokiSessionType {
+    ReadOnly,
+    _ReadWrite,
+}
+
+impl<'a> CryptokiSession<'a> {
```
didier-wenzek: Clippy warning:

```
warning: the following explicit lifetimes could be elided: 'a
   --> crates/extensions/tedge-p11-server/src/pkcs11/mod.rs:354:6
    |
354 | impl<'a> CryptokiSession<'a> {
    |      ^^                  ^^
```

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -201,21 +235,128 @@ impl Cryptoki {
...
+
+fn get_all_slots_info(cryptoki: &Pkcs11) -> Vec<SlotInfo> {
+    let slots = match cryptoki.get_all_slots() {
+        Ok(slots) => slots,
+        Err(err) => {
+            error!(?err, "failed to get slots");
+            return vec![];
+        }
+    };
+    let slotinfos = slots
```
didier-wenzek: Clippy warning

```
warning: returning the result of a `let` binding from a block
   --> crates/extensions/tedge-p11-server/src/pkcs11/mod.rs:316:5
```

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -201,21 +235,128 @@ impl Cryptoki {
...
+struct CryptokiSession<'a> {
+    session: Session,
+    token_info: TokenInfo,
+    uri_attributes: uri::Pkcs11Uri<'a>,
+}
+
+#[derive(Debug, Clone, Copy)]
+enum CryptokiSessionType {
+    ReadOnly,
+    _ReadWrite,
```
didier-wenzek, line 347: Why introduce a case that is never used? Somehow this even makes `CryptokiSessionType` useless.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -164,16 +174,40 @@ impl Cryptoki {
...
+        // avoid reinitializing for now because of Nitrokey slowness
+        // self.reinit()?;
+
         let context = match self.context.lock() {
             Ok(c) => c,
             Err(e) => e.into_inner(),
         };
 
+        debug!(slots = ?get_all_slots_info(&context));
+        debug!(tokens = ?get_all_token_info(&context));
```
didier-wenzek, line 209: Is debug appropriate? Wouldn't info be better as these lists of slots and tokens are critical to check that a setting is correct.

What I'm missing is how often `open_session()` is called.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: It's a part of #3709 where we do need it because we need to open a RW session, this variant will be used when that's eventually merged.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: >Is debug appropriate?

We already do print the `SlotInfo` and `TokenInfo` of the token we select using debug, so kept debug for printing all to be consistent.

 >Wouldn't info be better as these lists of slots and tokens are critical to check that a setting is correct.
>
> What I'm missing is how often open_session() is called.

`open_session()` is called for every request, so having this be info by default would spam the output quite a bit. And it has to be for every request, because slots and objects can change in between.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: fixed in 014adb5

they must've disabled this lint in later versions because I'm developing on the current stable and I didn't got it.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: fixed in bcd3df27

PR #3789, opened 2025-09-18
fix: MQTT messages are forwarded twice to subscribers with overlapping subscriptions

didier-wenzek:
When an actor subscribes to overlapping topics, e.g. `a/+` and `+/b`, then all messages received on the overlapping topics (in that case `a/b`) are forwarded twice to the actor, and even four times with some MQTT brokers (rumqttc and nanomq).

- Little can be done when the broker itself duplicates the messages. But at least, the `tedge_mqtt_ext` layer should not add its own set of duplicates. 
- Its difficult to avoid such overlapping subscriptions with `tedge flows` where the subscriptions are independently configured for each flow.

## Proposed changes

- [x] Let `tedge_mqtt_ext` removes duplicates in the list of clients for a message

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_mqtt_ext/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_ext/src/lib.rs
@@ -14,6 +14,7 @@ use mqtt_channel::StreamExt;
 use mqtt_channel::SubscriberOps;
 pub use mqtt_channel::Topic;
 pub use mqtt_channel::TopicFilter;
+use std::collections::HashSet;
 use std::convert::Infallible;
 use std::sync::Arc;
 use std::time::Duration;
@@ -356,10 +357,10 @@ impl TrieService {
 }
 
 #[cfg(feature = "test-helpers")]
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct ClientId(pub usize);
 #[cfg(not(feature = "test-helpers"))]
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
 pub struct ClientId(usize);
 
 type MatchRequest = String;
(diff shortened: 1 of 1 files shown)


crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -627,7 +628,7 @@ impl ToPeers {
         let TrieResponse::Matched(matches) = subscribed else {
             unreachable!("MatchRequest always returns Matched")
         };
-        for client in matches {
+        for client in HashSet::<ClientId>::from_iter(matches) {
```
didier-wenzek, line 631: This fixes the issue, but might not be the best way. Can this de-duplication be done internally when collecting the subscribers from the trie?

crates/extensions/tedge_mqtt_ext/src/tests.rs
```diff
@@ -250,6 +250,57 @@ async fn dynamic_subscriptions() {
...
+    let mut client = client.build();
+
+    // The client then dynamically subscribes to an overlapping topic
+    client
+        .send(MqttRequest::Subscribe(SubscriptionRequest {
+            diff: SubscriptionDiff {
+                // a more convincing example would be to subscribe to "+/b"
+                // but in that case rumqttd sends the message twice
+                // something we can do little about
+                subscribe: ["a/b".into()].into(),
```
didier-wenzek, line 278: Adding a subscription to `+/b` leads to 4 messages being delivered to the client.

albinsuresh (commented):
The fix looks fine and AFAIK, this topic overlap issue was highlighted as a known limitation when dynamic subscription was implemented. So, I doubt if anything else can be done. So, I'll let @jarhodes314 approve the PR.

crates/extensions/tedge_mqtt_ext/src/lib.rs
jarhodes314: I believe we _could_ deduplicate in the trie logic, but I don't think we would gain anything over this solution by doing that.

PR #3795, opened 2025-09-22
fix: Only reload PKCS11 library if key is not found

Bravo555:
## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge-p11-server/src/proxy/server.rs
```diff
@@ -118,6 +119,14 @@ impl TedgeP11Server {
                     }
                 }
             }
+
+            // This branch runs only when a client calls TedgeP11Client::with_ready_check.
+            // It essentially does nothing, but when the server is not yet running and is triggered
+            // by a systemd socket, at a point client makes the request systemd starts the server,
+            // Cryptoki object is initialized before TedgeP11Server is constructed (in main) and at
+            // the point we return the response here, the PKCS11 library is loaded and ready to
+            // serve the subsequent request.
```
didier-wenzek: I feel this comment a bit difficult to understand. The content is here (reading twice, I got it), but having several independent sentences might help.

Here is my proposal:

The Ping/Pong request does nothing per-se, but can be used by a client to start the service. Indeed, systemd being configured to start the service when a request is received on the associated socket, a Ping/Pong request triggers a service start and ensures the PKCS11 library is loaded and ready to serve signing requests. In practice, this only occurs with a client calls TedgeP11Client::with_ready_check.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -87,30 +88,72 @@ pub struct Cryptoki {
...
+                // dropping the signing key should drop the session
+                drop(signing_key);
+                self.reinit()?;
+                self.open_session_ro(request.uri.as_deref())
+                    .and_then(|s| s.signing_key())
+                    .context("Failed to find a signing key")?
+            }
+
+            Err(e) => return Err(e),
+        };
```
didier-wenzek: This piece of code is critical and tricky, and must therefore not be duplicated to `choose_scheme` and `sign`.

=> Introduce a `get_signing_key(self, uri)` method.

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
```diff
@@ -87,30 +88,72 @@ pub struct Cryptoki {
...
+            // refresh the slots only if the key couldn't be found, i.e.:
+            // - we didn't find a slot with the token that matches the URI
+            // - we didn't find an object on the token that matches the URI
+            Err(ref e)
+                if format!("{e:#}").contains("Didn't find a slot to use")
+                    || format!("{e:#}").contains("Failed to find a key") =>
+            {
+                warn!("Failed to find a signing key, reloading the library to retry");
+                // dropping the signing key should drop the session
+                drop(signing_key);
```
didier-wenzek: The term *should* makes me wonder. Is the session actually dropped?

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: fixed in bc09733709

crates/extensions/tedge-p11-server/src/pkcs11/mod.rs
Bravo555: Indeed, the comment sounds unsure. However, I'm sure the session must be dropped here. The `signing_key` is a `Pkcs11Session` which has `session: Arc<Mutex<Session>`, but after construction `signing_key` is immediately dropped without us cloning the session anywhere, so it must be dropped in entirety.

I have some in-progress refactoring that removes this Mutex, will be more clearer then. For now, giving this comment a little self-confidence in 931ae6bb06

crates/extensions/tedge-p11-server/src/proxy/server.rs
Bravo555: Indeed, the comment describes what happens, but not _why_ it happens, assuming the context of a problem that a reader will have no notion of. Used your comment and changed the beginning a bit to add a bit more context.
Fixed in c4bf1ad.

PR #3801, opened 2025-10-01
refactor: improve handling of intervals in tedge-flows

jarhodes314:
## Proposed changes
Refactor tedge-flows to improve interval handling.

Previously, tedge-flows relied on an exact number of seconds passing to execute an interval, and this had two negative effects:
- If tedge-flows was experiencing resource-pressure/needed to execute a moderately long-running javascript process, it could miss a tick entirely just by being a second late
- If we wanted to test the intervals for real, the intervals used need to be multiple seconds long to definitively assert the interval is behaving, which is non-ideal for unit/integration tests

This PR fixes both of these issues, allowing arbitrary precision intervals that are guaranteed to execute once the interval has elapsed, regardless of how long it takes tedge-flows to inspect the pending tasks. Additionally, the new approach supports the `tokio::time` mocking API, so I have added some deterministic integration tests that don't require sleeping whatsoever.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/Cargo.toml
+++ b/crates/extensions/tedge_flows/Cargo.toml
@@ -34,7 +34,9 @@ toml = { workspace = true, features = ["parse"] }
 tracing = { workspace = true }
 
 [dev-dependencies]
+tedge_mqtt_ext = { workspace = true, features = ["test-helpers"] }
 tempfile = { workspace = true }
+tokio = { workspace = true, features = ["test-util"] }
 
 [lints]
 workspace = true
--- a/crates/extensions/tedge_flows/tests/interval.rs
+++ b/crates/extensions/tedge_flows/tests/interval.rs
@@ -1,0 +1,497 @@
+use std::sync::Arc;
+use std::sync::Mutex;
+use std::time::Duration;
+use tedge_actors::Actor;
+use tedge_actors::Builder;
+use tedge_actors::CloneSender;
+use tedge_actors::MappingSender;
+use tedge_actors::MessageSink;
+use tedge_actors::MessageSource;
+use tedge_actors::NoConfig;
+use tedge_actors::SimpleMessageBoxBuilder;
+use tedge_flows::FlowsMapperBuilder;
+use tedge_mqtt_ext::DynSubscriptions;
+use tedge_mqtt_ext::MqttMessage;
+use tedge_mqtt_ext::MqttRequest;
+use tempfile::TempDir;
+
+#[tokio::test(start_paused = true)]
+async fn interval_executes_at_configured_frequency() {
+    let config_dir = create_test_flow_dir();
+
+    write_file(
+        &config_dir,
+        "counter.js",
+        r#"
+        let count = 0;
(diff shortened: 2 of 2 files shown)


crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -116,24 +119,51 @@ impl JsScript {
...
+    }
+
+    /// Check if this script should execute its interval function now
+    /// Returns true and updates next_execution if it's time to execute
+    pub fn should_execute_interval(&mut self) -> bool {
+        if self.no_js_on_interval_fun || self.interval.is_zero() {
+            return false;
+        }
+
+        let now = Instant::now();
```
didier-wenzek: I would pass `now` as an argument of the method. Having the time externalized makes things easier to test. Using `tokio::time` mocking helps for unit tests, but is useless for `tedge flows test`

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -116,24 +119,51 @@ impl JsScript {
...
     /// Trigger the onInterval function of the JS module
     ///
     /// The "onInterval" function is passed 2 arguments
     /// - the current timestamp
     /// - the current flow step config
     ///
     /// Return zero, one or more messages
+    ///
+    /// Note: Caller should check should_execute_interval() before calling this
     pub async fn on_interval(
```
didier-wenzek, line 161: I see pros and cons to externalize or not the check if this call is timely. What are your motivation to move this check out?

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -116,24 +119,51 @@ impl JsScript {
...
+    /// Should be called after the script is loaded and interval is set
+    pub fn init_next_execution(&mut self) {
+        if !self.no_js_on_interval_fun && !self.interval.is_zero() {
+            self.next_execution = Some(Instant::now() + self.interval);
+        }
+    }
+
+    /// Check if this script should execute its interval function now
+    /// Returns true and updates next_execution if it's time to execute
+    pub fn should_execute_interval(&mut self) -> bool {
```
didier-wenzek: It's okay for this PR to have a `should_execute_interval` function inside the `js_script` module. However, this abstraction will also be useful for rust-based flow steps as well as for [sources producing data at a regular pace](https://github.com/thin-edge/thin-edge.io/pull/3779#discussion_r2391809781).

crates/extensions/tedge_flows/src/js_script.rs
jarhodes314: For the functions the javascript module exports for tedge-flows to call, my expectation is that the methods on `JsScript` serve as a lightweight wrapper, doing minimal work beyond calling the function with the provided input. It also feels like a bit of a hack to pretend the function produced no output when it wasn't called at all (or if it entirely doesn't exist).

didier-wenzek (approved):
Approved. I appreciate the really nice unit test suite.

PR #3802, opened 2025-10-02
refactor: Improve error reporting on sudo execution error

didier-wenzek:
## Proposed changes

Improve error reporting when:

- [x] a command has not been configured to be executable by tedge with sudo
- [x] a sm plugin has not been properly configured to be used with sudo
- [x] a log plugin  has not been properly configured to be used with sudo

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/sudo.rs
+++ b/crates/common/tedge_config/src/sudo.rs
@@ -1,5 +1,6 @@
 use std::ffi::OsStr;
 use std::process::Command;
+use std::process::Stdio;
 use std::sync::Arc;
 use tracing::warn;
 
@@ -65,4 +66,79 @@ impl SudoCommandBuilder {
             }
         }
     }
+
+    /// Ensure the command can be executed using sudo
+    ///
+    /// Be warned, that the command is actually executed.
+    pub fn ensure_command_succeeds<S: AsRef<OsStr>>(
+        &self,
+        program: &impl AsRef<OsStr>,
+        args: &Vec<S>,
+    ) -> Result<(), SudoError> {
+        let output = self
+            .command(program)
+            .args(args)
+            .stdout(Stdio::null())
+            .stderr(Stdio::piped())
+            .output();
+        match output {
+            Ok(output) if output.status.success() => Ok(()),
+            Ok(output) => {
+                tracing::error!(target: "sudo", "{} failed with stderr: <<EOF\n{}\nEOF",
+                    program.as_ref().to_string_lossy(),
+                    String::from_utf8_lossy(output.stderr.as_ref()));
+                match output.status.code() {
+                    Some(exit_code) => {
+                        if self.command_is_sudo_enabled(program, args) {
+                            Err(SudoError::ExecutionFailed(exit_code))
+                        } else {
+                            Err(SudoError::CannotSudo)
+                        }
+                    }
+                    None => Err(SudoError::ExecutionInterrupted),
+                }
+            }
+            Err(err) => Err(SudoError::CannotExecute(err)),
+        }
+    }
+
+    /// Check that sudo is enabled and the user authorized to run the command with sudo
+    ///
+    /// This is done by running `sudo --list <command> <args>`.
+    fn command_is_sudo_enabled<S: AsRef<OsStr>>(
+        &self,
+        program: &impl AsRef<OsStr>,
+        args: &Vec<S>,
+    ) -> bool {
+        if !self.enabled {
+            return false;
+        }
+        let Ok(sudo) = which::which(self.sudo_program.as_ref()) else {
+            return false;
+        };
+        let status = Command::new(sudo)
+            .arg("-n")
+            .arg("--list")
+            .arg(program)
+            .args(args)
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .status();
+        matches!(status, Ok(status) if status.success())
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum SudoError {
+    #[error("The user has not been authorized to run the command with sudo")]
+    CannotSudo,
+
+    #[error(transparent)]
+    CannotExecute(#[from] std::io::Error),
+
+    #[error("The command returned a non-zero exit code: {0}")]
+    ExecutionFailed(i32),
+
+    #[error("The command has been interrupted by a signal")]
+    ExecutionInterrupted,
 }
--- a/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
+++ b/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
@@ -134,7 +134,8 @@ Overriding a log plugin
     # Add an extra location for local log plugins
     Execute Command    mkdir -p /usr/local/tedge/log-plugins
     Execute Command    tedge config set log.plugin_paths '/usr/local/tedge/log-plugins,/usr/share/tedge/log-plugins'
-    Execute Command    cmd=echo 'tedge ALL = (ALL) NOPASSWD:SETENV: /usr/local/tedge/log-plugins/[a-zA-Z0-9]*' | sudo tee -a /etc/sudoers.d/tedge
+    Execute Command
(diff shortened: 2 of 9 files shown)


crates/common/tedge_config/src/sudo.rs
```diff
@@ -65,4 +66,69 @@ impl SudoCommandBuilder {
             }
         }
     }
+
+    /// Check that the command can be executed using sudo
+    pub fn command_is_executable<S: AsRef<OsStr>>(
```
jarhodes314: From the name, I'd expect this to just tell me if I _can_ execute the command, but it looks like this actually _executes_ the command too, which I found suprising. Is there a better name we can give this to more clearly communicate the actual behaviour?

didier-wenzek: Yes, this is actually running the command, the point being to filter out executables that are not plugins .i.e. not following the Log or SM plugin API.

So, I agree that this method is badly named. However, I don't have any better suggestion.

- `check_command_accepts_args` ?

jarhodes314: Maybe something more like `ensure_command_succeeds`?

didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3802/commits/94c822d5b666d1435b0eb0b6691205cd6c4c3340

jarhodes314 (approved):

reubenmiller:
@didier-wenzek I've been trying to run the "Overriding a log plugin" system test locally and removing the sudoers rule to see the behaviour, but I'm not seeing any log output indicating a sudoer rule issue. I only see the following error (which is similar to what we had before):

```log
Oct 06 12:04:17 0529f55f43d7 tedge-agent[969]: 2025-10-06T12:04:17.538831466Z ERROR log plugins: Skipping /usr/local/tedge/log-plugins/fake_plugin: does not support list operation and may not be a valid plugin
```

crates/common/tedge_config/src/sudo.rs
```diff
@@ -65,4 +66,74 @@ impl SudoCommandBuilder {
...
+    pub fn ensure_command_succeeds<S: AsRef<OsStr>>(
+        &self,
+        program: &impl AsRef<OsStr>,
+        args: &Vec<S>,
+    ) -> Result<(), SudoError> {
+        let status = self
+            .command(program)
+            .args(args)
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
```
reubenmiller: It might be a misunderstanding on my side, but I find redirecting stdout/stderr to dev null surprising given the the stdout and stderr maybe part of the plugin contract (e.g. the log plugin uses stdout to get the list of supported log types from a plugin).

jarhodes314: I believe the intention of the method is that we're just verifying that the plugin obeys this bit of the contract (i.e. it's executable and in that specific case, supports a list command) rather than anything where we actually particularly care about the output.

reubenmiller: I'm reopening this as redirecting stdout and stderr means that it is very difficult to debug faulty plugins as any stdout/stderr generated from the plugins is discarded.

I just had an example where a new plugin accidentally used "print" instead of "printf", and the only log output I received was the following (which wasn't very helpful to locate the issue):

```sh
Oct 06 14:16:40 5e8ba00eef29 sudo[859]:    tedge : PWD=/ ; USER=root ; COMMAND=/usr/share/tedge/log-plugins/journald list
Oct 06 14:16:40 5e8ba00eef29 sudo[859]: pam_unix(sudo:session): session opened for user root(uid=0) by (uid=997)
Oct 06 14:16:40 5e8ba00eef29 sudo[859]: pam_unix(sudo:session): session closed for user root
Oct 06 14:16:40 5e8ba00eef29 tedge-agent[741]: 2025-10-06T14:16:40.348138712Z ERROR log plugins: File /usr/share/tedge/log-plugins/journald in log plugin directory does not support list operation and may not be a valid plugin, skipping.
```

In the end, I had to run the plugin locally to see the printf typo.

```sh
$ /usr/share/tedge/log-plugins/journald list
c8y-firmware-plugin
mosquitto
mqtt-logger
ssh
tedge-agent
tedge-flows
tedge-mapper-c8y
/usr/share/tedge/log-plugins/journald: 44: print: not found
```

reubenmiller: We could potentially limit the output to stderr only...as stdout is part of the `list` contract, which would just flood the text which is not relevant for the check being executed.

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -158,6 +158,23 @@ Overriding a log plugin
...
+    ThinEdgeIO.Transfer To Device
+    ...    ${CURDIR}/plugins/dummy_plugin
+    ...    /etc/share/tedge/log-plugins/dummy_plugin
+    Execute Command    chmod a+x /etc/share/tedge/log-plugins/dummy_plugin
+
+    ${started_at}=    Execute Command    date "+%Y-%m-%d %H:%M:%S"    strip=${True}
+    Start Service    tedge-agent
+    Execute Command    sleep 3
+    ${agent-log}=    Execute Command    cmd=journalctl --no-pager --unit=tedge-agent --since="${started_at}"
+    Should Contain   ${agent-log}    ERROR log plugins: Skipping /etc/share/tedge/log-plugins/dummy_plugin: not properly configured to run with sudo
```
reubenmiller: I would suggest changing the assertion a little to be more reliable:

* Use grep within the `journalctl` command to avoid potential flakiness around the 3 second sleep (e.g. if the test runner is under load and things are slow to start). The `Execute Command` is actually a "smart assertion" which will retry the command automatically if a non-zero exit code, and eventually give up after 30 seconds (by default). Yes it makes the assertion a bit harder to read, however stability is preferred to readability here.

* Use `_SYSTEMD_INVOCATION_ID=` (see https://man7.org/linux/man-pages/man1/journalctl.1.html) to only return the logs related to the currently running instance of tedge-agent (to remove the reliance on a time based log retrieval). Unfortunately we can't use the `--invocation` flag as that is only available in >= systemd 257 (and we're on systemd 252 in Debian bookworm)

```suggestion
    Start Service    tedge-agent
    Execute Command    cmd=journalctl --no-pager _SYSTEMD_INVOCATION_ID="$(systemctl show -p InvocationID --value tedge-agent.service)" | grep -F "ERROR log plugins: Skipping /etc/share/tedge/log-plugins/dummy_plugin: not properly configured to run with sudo"
```

reubenmiller: And the readability can be improved if we add a dedicated keyword to do the same thing in the DeviceLibrary (but I can do that later on).

reubenmiller (dismissed):
Approved. Nice improvement

crates/common/tedge_config/src/sudo.rs
didier-wenzek: stderr is now captured: https://github.com/thin-edge/thin-edge.io/pull/3802/commits/6946ea1081b58af89b65186ec1880681c3800982

This has also the benefits to highlight that a password was required by sudo. Something `sudo --list <cmd> <args>` fails to tell.

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
reubenmiller: I've added a new keyword in the underlying testing framework to make this check easier. The usage is in this fixup commit, ff494f437295e9b3bad3c247b70cbe8b3d992bda

```
Service Logs Should Contain
    ...    tedge-agent
    ...    current_only=${True}
    ...    text=ERROR log plugins: Skipping /etc/share/tedge/log-plugins/dummy_plugin: not properly configured to run with sudo
```

reubenmiller (approved):
Approved

PR #3805, opened 2025-10-06
feat: tedge flows consuming and producing file content

didier-wenzek:
## Proposed changes

- [x] Add a watcher actor consuming commands stdout
- [x] Tedge flows consuming bg process stdout
- [x] Add ways to control when a failing bg process is spawned again.
      - Currently no configurable.
      - Retry after 5s delay when the command fails to run
      - The process is immediately re-spawned when reaching end-of-stream
- [x] Feed a flow with data produced by a command run at a regular pace
- [x] Tedge flows appending to files
- [x] Tedge flows redirecting errors to mqtt or a file
- [x] Resolve flaky system tests

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -91,12 +91,69 @@ Each instance of a script must have its own static state
...
     ...    ${expected_msg}
 
+Running tedge-flows
+    # Assuming the flow count-events.toml has been properly installed
+    Execute Command    tedge mqtt sub test/count/e --duration 2s | grep '{}'
+
+Consuming messages from a process stdout
+    # Assuming the flow journalctl-follow.toml has been properly installed
+    Execute Command    (sleep 1;tedge mqtt pub --retain te/device/main///cmd/software_list/test-follow '{"status":"init"}')&
+    Execute Command    tedge mqtt sub log/journalctl-follow --duration 2s | grep software_list
```
didier-wenzek: This test is unfortunately flaky

```
$  invoke flake-finder --test-name "Consuming messages from a process stdout" --iterations 100 --outputdir output_ff --clean
Overall: FAILED
Results: 100 iterations, 80 passed, 20 failed
```

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -156,28 +198,125 @@ impl FlowsMapper {
...
+                    .open(path)
+                    .await
+                    .map_err(|err| {
+                        error!(target: "flows", "{flow}: cannot open {path}: {err}");
+                    })
+                else {
+                    return Ok(());
+                };
+                for message in messages {
+                    if let Err(err) = file.write_all(format!("{message}\n").as_bytes()).await {
```
jarhodes314, line 336: As we're writing potentially many times to the file, I think we might want to consider using `tokio::io::BufWriter`

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -156,28 +198,125 @@ impl FlowsMapper {
...
+            } => self.publish(flow, messages, &output).await,
+            FlowResult::Err {
+                flow,
+                error,
+                output,
+            } => self.publish_error(flow, error, &output).await,
+        }
+    }
+
+    async fn publish(
```
jarhodes314, line 302: This flow is now much clearer with the separation of message processing from output format handling.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -184,11 +329,51 @@ impl FlowStep {
     }
 }
 
-impl FlowInput {
-    pub fn topics(&self) -> &TopicFilter {
+impl FlowOutput {
+    pub async fn publish_messages(
```
jarhodes314: From a very brief look, it seems like this is duplicating some logic in actor.rs. Can you explain what's going on/what I've overlooked?

jarhodes314 (commented):
The refactorings here are really nice along with file functionality being very useful. It's now a lot clearer how the different possible sources can coexist happily, and future sources and sinks can be added in a far less invasive manner.

I'm broadly very happy with the changes, I've just left a couple of questions.

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: Oops: code that should have been removed after the refactoring.

Fixed: https://github.com/thin-edge/thin-edge.io/pull/3805/commits/322b31f447b4aef06bf1e15e2a850d39710869b1

crates/extensions/tedge_flows/src/actor.rs
didier-wenzek: Added https://github.com/thin-edge/thin-edge.io/pull/3805/commits/9e483f61ffac232f0f6303d9ef6e57624b192816

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
reubenmiller: I'll take over this task to see if the system tests can be refactored to remove the flakiness (as this is a non-trivial task given all of the background processing going on)

reubenmiller: Should be resolved in bab77b2056cade2bd783a4430a782c9a439a9880

```sh
$ invoke flake-finder --test-name "Consuming messages from a process stdout" --iterations 100 --outputdir output_ff --clean

Overall: PASSED
Elapsed time: 0:20:18.769239
```

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
reubenmiller, line 1: There seems to be some additional flakiness in some system tests which I'm trying to resolve.

Running the following command results in some test failures:

```
invoke flake-finder --suite "tedge_flows" --iterations 5 --clean
```

reubenmiller: af42fa245404b8c9445b49fcd74f928b52b0ac7b should resolve most of the flakiness. The root cause seems to be that the journalctl flows were flooding the MQTT broker with messages causing other tests to fail due to the mqtt buffer being overloaded.

Moving the journalctl flows to a separate folder and only installing them for a given test seems to improve the tests reliability. A run of 20 iterations was successful, though I'll run another flake-finder with 100 iterations to confirm.

reubenmiller: The test suite looks good after 100 iterations.

```sh
$ invoke flake-finder --suite "tedge_flows" --iterations 100 --clean
Overall: PASSED
Elapsed time: 2:25:55.762151
```

docs/src/references/mappers/flows.md
```diff
@@ -134,6 +132,82 @@ steps = [
...
+input.file.path = "/var/log/some-app.log"
+```
+
+By default, this file is followed as done by `tail -F`, waiting for new lines to be appended and consumed as messages.
+This behavior can be changed, the file being then read at regular intervals.
+
+```toml
+[input.file]
+path = "/var/log/some-app.log"
+topic = "some-app-log"
```
albinsuresh, line 160: It'd be worth calling out the default input topic (the path) is also overridden by this explicit `topic`. And also the fact that this effectively becomes the `output` topic as well, when no further transformations happen in the flow.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -91,12 +92,106 @@ Each instance of a script must have its own static state
...
 
+Running tedge-flows
+    # Assuming the flow count-events.toml has been properly installed
+    Execute Command    tedge mqtt sub test/count/e --duration 2s | grep '{}'
+
+Consuming messages from a process stdout
+    # Assuming the flow journalctl-follow.toml has been properly installed
+    Install Journalctl Flow    journalctl-follow.toml
+    ${test_start}    Get Unix Timestamp
+    Restart Service    tedge-agent
```
albinsuresh, line 103: Is this restart needed? Since flows are dynamically reloaded by the file watcher mechanism

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -91,12 +92,106 @@ Each instance of a script must have its own static state
...
+    Should Not Be Empty    ${messages[0]}    msg=Output should not be empty lines
+    [Teardown]    Uninstall Journalctl Flow    journalctl-cursor.toml
+
+Consuming messages from the tail of file
+    # Assuming the flow tail-named-pipe.toml has been properly installed
+    ${start}    Get Unix Timestamp
+    Execute Command    echo hello>/tmp/events
+    Should Have MQTT Messages    topic=log/events    message_contains=hello    minimum=1    date_from=${start}
+
+Consuming messages from of file, periodically
```
albinsuresh:
```suggestion
Consuming messages from the tail of file, periodically
```

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -91,12 +92,106 @@ Each instance of a script must have its own static state
...
+    # Assuming the flow read-file-periodically.toml has been properly installed
```
albinsuresh, line 124: Copying these relevant flow definitions as part of the test and then removing them with a test teardown as done for the journalctl flow tests would have been better, as some of the flows like `count-events.toml`, `count-measurement.toml` etc add quite a lot of noise on the MQTT broker, making debugging difficult.

docs/src/references/mappers/flows.md
```diff
@@ -134,6 +132,82 @@ steps = [
...
+### Output connectors
+
+Transformed messages and errors can be published over MQTT or appended to files.
+
+The default is to publish the transformed messages over MQTT on the topics specified by each message.
+And to direct all the errors to a specific topic, the `te/error` topic.
+
+```toml
+[output.mqtt]
+
```
albinsuresh, line 193:
```suggestion
```
Unless that empty table had some relevance.

docs/src/references/mappers/flows.md
```diff
@@ -134,6 +132,82 @@ steps = [
...
+topic = "tedge-agent-journalctl"
+command = "journalctl --no-pager --cursor-file=/tmp/tedge-agent-cursor --unit tedge-agent"
+interval = "1h"
+```
+
+### Output connectors
+
+Transformed messages and errors can be published over MQTT or appended to files.
+
+The default is to publish the transformed messages over MQTT on the topics specified by each message.
```
albinsuresh, line 188:
```suggestion
The default is to publish the transformed messages over MQTT on the topics specified in the `input`.
```

docs/src/references/mappers/flows.md
didier-wenzek: Indeed not required. I only added it for symmetry.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
reubenmiller: The reason is to ensure that there are log entries that will be fed into the flow. Without restarting the tedge-agent, we can't guarantee that there will be log entries.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
reubenmiller: Alternatively we could just switch to a per-test setup, though I guess there will also still be noise if all flows are installed by default.

docs/src/references/mappers/flows.md
didier-wenzek: > And also the fact that this effectively becomes the `output` topic as well, when no further transformations happen in the flow.

Stressing this without being confusing might be tricky, even if you are correct (that the input topic might finally be used for the output if no steps assigned a different one).

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
didier-wenzek: To rephrase what has been said by Reuben, restarting the agent is not required by the flows (which are are dynamically reloaded  and by the way, *not* managed by the agent), but supporting the tests: we want to check that journald entries are actually consumed and processed by the flow defined by `journalctl-follow.toml`).

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
didier-wenzek: When reading a file *periodically* then the whole file is read, not juts its tail.

docs/src/references/mappers/flows.md
didier-wenzek: No, your understanding is incorrect. If no output topic is configured for a flow, then the topic assigned to each messages dictates where the message is published. Your suggested change applies only one corner case where no steps changed the topic assigned by the source.

albinsuresh: Aah yes. I didn't consider the flow steps modifying the topics.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
albinsuresh: Yeah, installing just the relevant flow for the test was the key, to avoid the noise from others.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
albinsuresh: Aah right. Then just drop the `of` from that name.

didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3805/commits/9ca104fa4f7739a2cec79c20914e43e223bb70fb

crates/extensions/tedge_flows/src/input_source.rs
```diff
@@ -0,0 +1,296 @@
...
+use crate::flow::SourceTag;
+use async_trait::async_trait;
+use camino::Utf8PathBuf;
+use std::fmt::Display;
+use std::time::Duration;
+use tedge_mqtt_ext::TopicFilter;
+use tedge_watch_ext::WatchRequest;
+use tokio::time::Instant;
+
+pub trait FlowInput: Display + StreamingSource + PollingSource {}
```
albinsuresh, line 12: Something feels wrong with this type definition. In my head, a `FlowInput` need not always be both a `StreamingSource` and a `PollingSource` but could have been either one as well. Like for example, the Mqtt source which can only be a `StreamingSource` and not a `PollingSource`, right?

crates/extensions/tedge_flows/src/input_source.rs
```diff
@@ -0,0 +1,296 @@
...
+impl<T: Display + StreamingSource + PollingSource> FlowInput for T {}
+
+/// Trait for input sources that stream messages out of continuously running processes
+pub trait StreamingSource {
+    /// MQTT topics subscribed by this source
+    fn topics(&self) -> TopicFilter {
```
albinsuresh, line 18: That feels like an MQTT source specific aspect leaking into what should have been a generic abstraction.

crates/extensions/tedge_flows/src/input_source.rs
```diff
@@ -0,0 +1,296 @@
...
+}
+
+impl Display for MqttFlowInput {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "MQTT topics: {:?}", self.topics)
+    }
+}
+
+#[async_trait]
+impl PollingSource for MqttFlowInput {
```
albinsuresh, line 69: This dymmy impl highlights the issue with the `FlowInput` definition highlighted above. Ideally, an `MqttFlowInput` shouldn't have had to implement `PollingSource` at all, as it really isn't a "real polling source".

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -118,17 +171,49 @@ impl StepConfig {
...
+            InputConfig::Mqtt { topics } => Box::new(MqttFlowInput {
                 topics: topic_filters(topics)?,
             }),
-        }
+            InputConfig::File {
+                topic,
+                path,
+                interval,
+            } => {
+                let topic = topic.unwrap_or_else(|| path.clone().to_string());
```
albinsuresh, line 185: For that default fallback, shouldn't we reject any paths with special characters that can't be in a topic? Esp since these same `input` topics can end up being the `output` topics as well.

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -118,17 +171,49 @@ impl StepConfig {
...
+                Box::new(FileFlowInput::new(flow_name, topic, path, interval))
+            }
+            InputConfig::Process {
+                topic,
+                command,
+                interval,
+            } => {
+                let topic = topic.unwrap_or_else(|| command.clone());
```
albinsuresh, line 193: Same comment as above on possible special characters.

crates/extensions/tedge_flows/src/input_source.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    fn next_deadline(&self) -> Option<Instant> {
+        None
+    }
+
+    fn is_ready(&self, _now: Instant) -> bool {
+        false
+    }
+
+    fn update_after_poll(&mut self, _now: Instant) {}
```
albinsuresh, line 82: Could this have been done as part of the `poll` itself rather than invoking it explicitly?

crates/extensions/tedge_flows/src/input_source.rs
```diff
@@ -0,0 +1,296 @@
...
+            SourceTag::Process { flow } => flow == &self.flow,
+            SourceTag::Poll { flow } if self.poll.is_polling() => flow == &self.flow,
+            SourceTag::Poll { .. } => false,
+        }
+    }
+}
+
+struct PollInterval {
+    polling_interval: Option<Duration>,
+    next_deadline: Option<Instant>,
```
albinsuresh, line 262: Rather than keeping all the values of this struct itself as optional, wouldn't it be better to make the `poll` field of the `XYZFlowInput` structs as `Option<>`?

docs/src/references/mappers/flows.md
reubenmiller: The flow determines what it does with the input topic...so I don't think we need to put all potential possibilities.

crates/extensions/tedge_flows/src/input_source.rs
didier-wenzek: You are right that in practice a *given* source is either a `StreamingSource` or a `PollingSource`. However, having a trait that handles both case, let the flow actor uniformly manage *all* the sources with a *single* source per flow. In a previous version, each flow was dealing with two sources - one for each type, making the actor juggling with the appropriate flow depending of the case. Somehow this replaces pattern cases like `if let Some(streaming_source) ...` from the actor to the source implementation which simply does nothing when not of the appropriate kind. More elaborated source implementation, yes, but in counter part, a simpler actor.

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: Possibly

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: Possibly

crates/extensions/tedge_flows/src/input_source.rs
didier-wenzek: As said here https://github.com/thin-edge/thin-edge.io/pull/3805#discussion_r2452647715, the idea is to simplify the actor logic at the cost of some dummy implementation as this one.

crates/extensions/tedge_flows/src/input_source.rs
didier-wenzek: You are correct, the `trait StreamingSource` is more a bag of streaming related sources rather than a generic abstraction. I struggled to find a better design. The main problem I have is that to be generic a source would have to not only specify types for the messages and registration requests but also to interact with actor builders.

didier-wenzek: What bothers me most about this trait is not really that there is a mix of concerns, but that this is done at the wrong level, i.e. at the level of a flow definition while only the actor can actually populate these sources with messages. This flaw is obvious in `tedge flows test` where concrete implementations of `StreamingSource` are injected into each flow but are not used by the test tool because it doesn't have the mechanism to do so - no actors, no MQTT connections, no sub-processes, but only test samples. This can even be seen as a bug: `tedge flows test` cannot be used to test a flow which source is not MQTT.

=> So, I will look for a more appropriate abstraction, along the following lines.
- The goal is to have a better separation of message processing from input gathering
- Keep flow definitions and the runtime free of any mechanisms to pull messages in, the focus of the runtime being only on processing messages.
- Introduce a concept of `ConnectedFlow` used by the flow actor to feed the runtime with messages.

albinsuresh: > The goal is to have a better separation of message processing from input gathering

Yeah, that coupling feels like the root of the problem. Would be nice if that is eliminated.

PR #3806, opened 2025-10-06
feat: add journald log plugin

reubenmiller:
## Proposed changes

Add journald log plugin which is included in the deb and rpm packages only.

The plugin includes a special log type called "all-units" which when used will return the logs for all systemd units/services. It is the equivalent to doing `journalctl -u "*"`. Having an option to return the logs for all units is useful when solving complex scenarios where behaviour across multiple services needs to be observed.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


configuration/contrib/log-plugins/journald
```diff
@@ -0,0 +1,128 @@
...
+    if ! command -V systemctl >/dev/null 2>&1; then
+        echo "Error: 'systemctl' command was not found" >&2
+        exit 1
+    fi
+    if ! command -V journalctl >/dev/null 2>&1; then
+        echo "Error: 'journalctl' command was not found" >&2
+        exit 1
+    fi
+    status=0
+    { systemctl list-units --type=service --no-legend --no-pager --plain || status=$?; } | awk '!/^systemd/ { sub(/\.service$/, "", $1); print $1 }' || status=$?
```
reubenmiller, line 40: Unfortunately the pipefail option is not in the posix spec (or at least not in any recent versions), so we have to do use subshells and store the status if the `systemctl` command fails.

albinsuresh (commented):
The updated plugin looks fine. Happy to approve once the log plugin system test is fixed, which wasn't previously expecting the types reported by the `journald` plugin.

albinsuresh:
> Should we have an "all" option which will return the logs from all units (not just the selected unit?)

Sounds good to me.

albinsuresh (approved):

Issue #3716, opened 2025-06-30
allow clients to provide the PKCS11 pin to the tedge-p11-server when interacting in the socket mode


reubenmiller:
**Is your feature request related to a problem? Please describe.**

It can be difficult for users to configure the PKCS11 pin to be used by the tedge-p11-server when the customer's applications (including tedge) are running within a container, and don't have access to the tedge-p11-server process or its configuration file.

**Describe the solution you'd like**

Allow clients to specify the PKCS11 pin to be used for interactions with the tedge-p11-server. Since the tedge-mapper already supports providing the key_uri to tedge-p11-server, allowing the client to also specify the required PIN would all the variable tedge-p11-server configuration to be controlled by the client (all by the PKCS11 `module_path`, however this is a static value which can be hardcoded when installing tedge-p11-server).

**Describe alternatives you've considered**

**Additional context**

PR #3809, opened 2025-10-06
feat: PKCS11 PIN per request

Bravo555:
## TODO

- [x] ~~documentation~~ expanded a bit on what the options are doing and what happens when both `cryptoki.pin` and `device.key_pin` are set

## Proposed changes

Allows the user to provide a PKCS#11 User PIN value to be used when logging into PKCS11 token using a new `device.key_pin` config setting.

```
tedge config set device.key_pin my-pin
```

Additionally, pin also may be provided in `device.key_uri` using a standard PKCS#11 URI `pin-value` query attribute:

```
pkcs11:token=mytoken;object=mykey?pin-value=my-pin
```

tedge-p11-server will then attempt to login to the token using `my-pin` instead of the default PIN tedge-p11-server was started with.

When multiple pin values are present, we use: URI -> pin field -> provider pin.

Because it is sensitive, there are security considerations, of interest to us being this part of the RFC7512:

> Section 7.5 of [RFC3986] applies since a PKCS #11 URI may be used in world-readable command-line arguments to run applications, stored in public configuration files, or otherwise used in clear text.  For that reason, the "pin-value" attribute should only be used if the URI string itself is protected with the same level of security as the token PIN itself otherwise is.

When using the `pin-value` attribute, because URIs are just strings, parts of the code that just pass these URIs may log them, potentially exposing secrets in the logs. On tedge-p11-server side as soon as we're parsing URI we can put in in a secrecy wrapper, to not be able to print it from then on, but there's nothing preventing printing the value before it's parsed on the server, or on the client making the request. For this reason, using `pin-value` is not recommended, although it's supported when setting `device.key_pin` is for some reason not possible, as `pin-value` is a standardised attribute so there's no reason for socket or module provider to not support it.

When using `device.key_uri` config setting, the value of the secret appears in `tedge.toml`, but after reading it will not be printed in the logs.

## Types of changes

## Paste Link to the issue

#3716

## Checklist


crates/extensions/tedge-p11-server/src/service.rs
```diff
@@ -37,13 +41,57 @@ pub struct ChooseSchemeResponse {
...
+    pub pin: Option<SecretString>,
+}
+
+/// A secret string that should not be printed by accident.
+///
+/// Rolling our own type because `secrecy::SecretString` doesn't impl Serialize,
+/// and we don't need eager zeroizing, we only need to make sure not to print
+/// the value.
+#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct SecretString(String);
```
didier-wenzek, line 62: I don't fully get the motivation here.

- Something that can be serialized is no more something that cannot be printed. 
- A `SecretString` can be exposed as a string with `expose()`
-  A `SecretString`  can be cast into a `secrecy::SecretString` which can then be exposed with `expose_secret()`.
- The secret is stored in `tedge.toml` as clear-text. And is displayed by `tedge config get`

This sounds overly complicated. Why no simply use `String`? Or `secrecy::SecretString`  with appropriate can to `expose_secret()` when sending a request over the unix socket?

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -152,6 +152,16 @@ define_tedge_config! {
...
 
+        /// User PIN value for logging into the PKCS#11 token provided by the consumer.
+        ///
+        /// This differs from cryptoki.pin in that cryptoki.pin is used by PKCS#11 provider, e.g. tedge-p11-server as a
+        /// default PIN for all tokens, but device.key_pin is the PIN provided by the consumer (tedge) with a given
+        /// `key_uri`.
+        ///
+        /// In practice, this can be used to define separate keys and separate PINs for different connection profiles.
+        #[tedge_config(example = "123456", example = "my-pin")]
+        key_pin: Arc<str>,
```
didier-wenzek, line 163: How can this be used to define separate keys and PINs while this `device.key_pin` is not a profiled configuration?

Why is this setting not under `cryptoki` ?

Bravo555: Made the `device.key_pin` profiled in 56876a5.

Settings under `cryptoki` are like a global configuration. When using tedge-p11-server, `cryptoki.pin` is a default value used when a client does not provide a pin. `cryptoki.uri` is used when client does not provide a URI, or used as a base to limit what keys are available to the client. This setting is not under `cryptoki` because it's used alongside `device.key_uri` when a client wants to use a specific key and use a pin that `tedge-p11-server` is not configured with.

crates/extensions/tedge-p11-server/src/service.rs
Bravo555: > - Something that can be serialized is no more something that cannot be printed.

We just don't want to log it accidentally, i.e. when logging the request with "{:?}" we'd like to see other fields but have PIN be redacted. When serializing, it's only done in `connection.rs` and when calling `.serialize()`, one has to provide a serializer, so it's not going to be accidental.

> - The secret is stored in `tedge.toml` as clear-text. And is displayed by `tedge config get`

Indeed config may contain secrets (we also have `cryptoki.pin` in `tedge.toml`), but we still don't want to have any secrets in the logs if we can help it (similarly to how `cryptoki.pin` is redacted when starting `tedge-p11-server`). Having pin as clear text in `tedge.toml` is probably not ideal and we could handle it like private keys, i.e. secret in a separate file and path to that file in `tedge.toml`, which we can always do in the future.

> Why no simply use String

Requests and Responses implement `Debug` and can be logged, so when using `String` we'd log the pin value in the logs

> Or secrecy::SecretString with appropriate call to expose_secret() when sending a request over the unix socket?

We're serializing the entire request to write it to the socket, to call `expose_secret()` one would have to write `impl Serialize` manually, which we're currently deriving.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
didier-wenzek: > Made the `device.key_pin` profiled in [56876a5](https://github.com/thin-edge/thin-edge.io/commit/56876a5b964623201246cb29fd3040162028d6b7).

A bit unfortunate that profile support implies so much duplicated code. Unrelated to this PR, though.

crates/extensions/tedge-p11-server/src/service.rs
didier-wenzek: > we could handle it like private keys, i.e. secret in a separate file and path to that file in `tedge.toml`, which we can always do in the future.

Why not in this PR? This would makes more sense.

By the way, this is also what is done for `c8y.credentials_path`. 

> We just don't want to log it accidentally, i.e. when logging the request with "{:?}" we'd like to see other fields but have PIN be redacted. 

I appreciate that. However, I feel the current implementation is over complicated. Do we need the `secrecy` crate as we have to work around it? Why not simply keep this `SecretString` wrapper with redacting `Display` and `Debug` implementations?

Bravo555: Removed `secrecy` and replaced with our own `SuperString` everywhere, in dca30ef.

 > Why not in this PR? This would makes more sense.
 
I think then we'd have to add an option for `cryptoki.pin_path` as well, so both pin options support it, point users to use this new option and keep the old `cryptoki.pin_path` for backwards compatibility but say it's not recommended, and then in this PR I could change `key_pin` to `key_pin_path`. I could do the `cryptoki.pin_path` changes in a separate PR for readability.

reubenmiller: As discussed offline, it has been decided to keep with storing the pin within the tedge.toml. The user can still protect this file by changing the permissions so that it is only readable by the `tedge` user (e.g. `chmod 600 /etc/tedge/tedge.toml`).

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -168,13 +178,28 @@ define_tedge_config! {
...
+            /// the scope of keys or tokens that can be used by the clients.
+            ///
+            /// For example, if `cryptoki.uri=pkcs11:token=token1` and `device.key_uri=pkcs11:token2;object=key1`,
+            /// `tedge-p11-server` will use URI `pkcs11:token1;object=key1`.
+            ///
+            /// For more information about PKCS11 URIs, see RFC7512.
+
+            // NOTE: combining URI behaviour seems unintuitive and surprising. If a client asks for a key on `token2`,
+            // but the server only allows accessing `token1` then it probably should receive a KeyNotFound instead of
+            // getting a signature from key on `token1` and then NotAuthorized because the signature is wrong.
```
didier-wenzek, line 201: Maybe this doc comment is trying to say too many things, notably when displayed by `tedge config list --doc`. Yes PKCS#11 is subtel, but I seems to me that a proper doc would do better that a command helper.

Bravo555: This is a regular comment, not a doc comment, so it's not displayed in `tedge config list --doc`. It's related to the behaviour explained in the doc comment, so decided to include it below.

Bravo555: Unless you don't really mean the `NOTE`, but also the entire doc comment above talking about how the URIs are combined?
In this case I'd agree that if every option had descriptions of this size, then yeah `tedge config list --doc` output would be a bit large, but still it's nice to have longer descriptions for the options, explaining their behaviour and also potentially interactions with other options, directly in tedge-config instead of only doing that in Markdown documentation.

My preferred solution would be to either tweak a bit the format of `tedge config list --doc` such that these longer descriptions look better.

didier-wenzek: > My preferred solution would be to either tweak a bit the format of `tedge config list --doc` such that these longer descriptions look better.

We can see that later.

PR #3814, opened 2025-10-08
fix: update file-log-plugin alias to avoid future conflicts

albinsuresh:
## Proposed changes

The `file` alias used for `tedge-file-log-plugin` installed at `/usr/share/tedge/log-plugins/file` would conflict with the upcoming `tedge-file-config-plugin` installed at `/usr/share/tedge/config-plugins/file`.  Though they exist in two different directories on the file-system, the executable name resolution of the `tedge` multi-call binary wouldn't support two different subcommands having the same alias. So, this fix just uses a non-conflicting name `file-log` for the log plugin. The upcoming config plugin would be called `file-config`.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -158,7 +158,7 @@ impl LogManagerActor {
         let (log_type, plugin_name) = request_payload
             .log_type
             .split_once("::")
-            .unwrap_or((&request_payload.log_type, "file"));
+            .unwrap_or((&request_payload.log_type, "file-log"));
```
albinsuresh: This change highlights the coupling that we have between the plugin name on the file-system and the internal handling of that "default" plugin which is assumed to be the `file-log` plugin. Wondering if this warrants the introduction of a new  configurable default log type (e.g: `log.plugin.default`) with the default value being `file-log`.

crates/core/tedge/src/cli/init.rs
```diff
@@ -100,7 +100,13 @@ impl TEdgeInitCmd {
         }
         // The last directory of the log plugin path list is the location for thin-edge provided plugins
         if let Some(main_log_plugins_dir) = config.log.plugin_paths.0.last() {
-            create_symlinks_for("file", target, Path::new(main_log_plugins_dir), &RealEnv).await?;
+            create_symlinks_for(
+                "file-log",
```
albinsuresh: @reubenmiller and I  had an offline discussion on this fragile "symlink contract", which requires the `tedge` mult-call binary to be aware of the symlink names as well and was exploring alternative approaches to avoid this. One proposal was to create wrapper shell scripts that calls the respective tedge component using the `tedge run` command, instead of these symlinks. For example,  `/usr/share/tedge/log-plugins/file` would be a simple script as follows instead of a symlink:

```
#!/bin/sh

set -eu

tedge run tedge-file-log-plugin $@
```

didier-wenzek: I don't get the point. The `tedge` multi-call binary doesn't need to know the actual names  of the commands (as long as they are properly listed as sub-commands). The issues with the `apt` and the log `file` plugins are coming from elsewhere: we want to replace the exit code used by clap and also we want the `init` command to create the symlinks.

I see elsewhere the root cause of the issue motivating this PR: the filename of the plugin being used as the log type exposed to the users.

reubenmiller: The problem is that when configuration plugins are introduced, then there will be a second "file" plugin which will be related to the configuration file management and the log file management...so it is really a namespacing issue.

One option to resolve the future conflict would be to use log/config within the file, but this then presents an odd UX for the names of the types when it is published to the cloud...e.g. `mosquitto::file-log`...which looks inconsistent as other plugin names wouldn't have the redundant "-log" suffix, e.g. `tedge-agent::journald`.

The other option, is to use a small script instead of a symlink (in the log/plugin cases at least), as this could be even more flexible. The only downside is that it would launch a small shell to launch but then replace the process with the process that should be called (without forking it):

**file: /usr/share/tedge/log-plugins/file**
```sh
#!/bin/sh
set -e
exec tedge run tedge-file-log-plugin "$@"
```

reubenmiller: Of course we can look at other solutions. Symlinks aren't necessarily the main requirement behind having a multi-call binary...the point for the multi-call binary was more to decrease the overall binary size of the thin-edge.io components.

albinsuresh: > The tedge multi-call binary doesn't need to know the actual names of the commands (as long as they are properly listed as sub-commands). The issues with the apt and the log file plugins are coming from elsewhere: we want to replace the exit code used by clap and also we want the init command to create the symlinks.

The exit codes were not the only issue. The multi-call binary detects which subcommand to run based on the executable name that it was invoked with, right? The following lookup code:
```
    let is_known_subcommand = executable_name
        .as_deref()
        .is_some_and(|name| cmd.find_subcommand(name).is_some());
```
looks for subcommand names or their aliases for this matching.

Because of this design, the `tedge-log-file-plugin` subcommand will be executed only if the symlink name is `tedge-log-file-plugin` itself or its current alias `file`. Even if you create another symlink named `xyz` that points to `/usr/bin/tedge-log-file-plugin`, when it is executed, it gets resolved to the `tedge` command itself and not the `tedge run tedge-log-file-plugin` subcommand.

This hard linking of the alias `file` to mean the `tedge-log-file-plugin` becomes a problem when we soon introduce the `tedge-config-file-plugin` which can't be use the alias `file` anymore. It can take a new alias `file-config`, sure, but then it becomes a bit weird that `file` mean `file-log`. So, to avoid that confusion and keep things consistent, I just switched to the explicit alias: `file-log`.

> I see elsewhere the root cause of the issue motivating this PR: the filename of the plugin being used as the log type exposed to the users.

Yes, that's a different issue highlighted in the [previous comment](https://github.com/thin-edge/thin-edge.io/pull/3814#discussion_r2414078250)

albinsuresh: > One option to resolve the future conflict would be to use log/config within the file, but this then presents an odd UX for the names of the types when it is published to the cloud...e.g. mosquitto::file-log...which looks inconsistent as other plugin names wouldn't have the redundant "-log" suffix, e.g. tedge-agent::journald.

Just to be clear, even though this PR renamed the plugin from `file` to `file-log`, there was no externally visible change, like the log types reported by this plugin getting a new `::file-log` suffix. That's because we don't include this suffix for the types retrieved from the "implicitly default" file plugin. So, a `mosquitto` log entry in the `tedge-log-plugin.toml` would still be reported as `mosquitto` only and not `mosquitto::file-log`. That way, the name of the "default" plugin is hidden from the end-user.

didier-wenzek: > The multi-call binary detects which subcommand to run based on the executable name that it was invoked with, right? 

Of course, this is the principle of multi-call binary. So I should say that `tedge` doesn't need to repeat itself about sub-command names once properly listed as sub-commands.

My point is that by pointing multi-call as the culprit is barking up the wrong tree. The true issue is a namespace problem.

> Just to be clear, even though this PR renamed the plugin from `file` to `file-log`, there was no externally visible change ... So, a `mosquitto` log entry in the `tedge-log-plugin.toml` would still be reported as `mosquitto` only and not `mosquitto::file-log`. That way, the name of the "default" plugin is hidden from the end-user.

So why bother with aliases ? `tedge-log-file-plugin` will work. Isn't it?

reubenmiller: The problem I have with that approach is that the moment the default plugin is changed, then we'd have "::file-log" suffix on its entries.

The current mix of some types having a suffix and some not (i.e. the default), is already an unexpected UX as I'm sure we'll get questions about this.

albinsuresh: > So why bother with aliases ? tedge-log-file-plugin will work. Isn't it?

Yes, as long as the default plugin is not changed. Since the name itself isn't externally visible and not reported to the cloud (as long as it is the default plugin), then having shorter name (`file` or `file-log`) doesn't really matter. The shorter name is only desirable when it is visible from the cloud (when the default is changed).

> The current mix of some types having a suffix and some not (i.e. the default), is already an unexpected UX as I'm sure we'll get questions about this.

Yes, skipping the suffixes when there are no conflicts sound good, but all the conflict resolution strategies that I can think of have some limitations. One approach is, when the plugins report non-conflicting types, report them as-is, and when there is a conflict, add a suffix to both types. What I don't like about this approach is the non-determinism whether a suffix would be present or not. For example, if `plugin_y` reported `type_a` without any conflicts initially, when a new `plugin_x` is installed and if it also supports the same `type_a`, the previously supported `type_a` suddenly becomes `type_a::plugin_y`, which would be unacceptable.

configuration/package_manifests/nfpm.tedge.yaml
```diff
@@ -94,6 +94,13 @@ contents:
     packager: rpm
 
   # preset log plugins
+
+  # file
+  - src: ./configuration/contrib/log-plugins/file
```
albinsuresh, line 99: When I tested locally, the `file` symlink from an old installation is getting replaced with this new script with the same name. So, I haven't updated the `preinst` script to remove the symlink first.

crates/core/tedge/src/cli/init.rs
albinsuresh: As discussed offline, decided to replace the `file` symlink with the wrapper script mentioned above.

configuration/package_manifests/nfpm.tedge.yaml
albinsuresh: Apparently, `dpkg` avoids overwriting a file only if it is "owned by" another package. But it'll happily overwrite any other "local" files (not owned by another package). Since the `file` symlink was earlier crated by the `tedge init` command, and not included in the debian manifest, it is treated as a local file.

Issue #3812, opened 2025-10-07
Interface to retrigger the tedge-agent log-plugins list commands


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When the output of a log plugin has a new log entry, there isn't any easy mechanism to communicate to the tedge-agent that it should re-execute the log plugin list commands.

For instance below shows the problematic workflow:

1. Install [tedge-container-plugin](https://github.com/thin-edge/tedge-container-plugin) which will include a "container" log plugin which reports the list of container on the device.
2. Install a new container called "app1" (via the container software management plugin)
3. The user wishes to get the logs from the command, however the "app1::container" log type isn't included in the log type list, so it is not selectable in the Cumulocity Device Management UI

**Describe the solution you'd like**

Add support for a mechanism that a component can get the tedge-agent to re-evaluate the log types list without having to restart the tedge-agent service.

**Describe alternatives you've considered**

**Additional context**

PR #3817, opened 2025-10-13
feat: publish supported log types when software_update or config_update finishes

albinsuresh:
## Proposed changes

Refresh and publish supported log types when `software_update` or `config_update` operations finish.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3812

## Checklist

### DIFF
--- a/crates/extensions/tedge_log_manager/src/config.rs
+++ b/crates/extensions/tedge_log_manager/src/config.rs
@@ -63,10 +63,14 @@ impl LogManagerConfig {
             ChannelFilter::Command(OperationType::LogUpload),
         );
 
-        let log_metadata_sync_topics = mqtt_schema.topics(
+        let mut log_metadata_sync_topics = mqtt_schema.topics(
             EntityFilter::Entity(&mqtt_device_topic_id),
             ChannelFilter::Command(OperationType::SoftwareUpdate),
         );
+        log_metadata_sync_topics.add_all(mqtt_schema.topics(
+            EntityFilter::Entity(&mqtt_device_topic_id),
+            ChannelFilter::Command(OperationType::ConfigUpdate),
+        ));
 
         Ok(Self {
             mqtt_schema,
--- a/crates/extensions/tedge_log_manager/src/lib.rs
+++ b/crates/extensions/tedge_log_manager/src/lib.rs
@@ -262,6 +262,6 @@ impl MessageSink<CmdMetaSyncSignal> for LogManagerBuilder {
 impl SyncOnCommand for LogManagerBuilder {
     /// Return the list of operations for which this actor wants to receive sync signals
     fn sync_on_commands(&self) -> Vec<OperationType> {
-        vec![OperationType::SoftwareUpdate]
+        vec![OperationType::SoftwareUpdate, OperationType::ConfigUpdate]
     }
 }
--- a/crates/extensions/tedge_log_manager/src/tests.rs
+++ b/crates/extensions/tedge_log_manager/src/tests.rs
@@ -99,8 +99,9 @@ async fn new_log_manager_builder(
     SimpleMessageBox<NoMessage, FsWatchEvent>,
     UploaderMessageBox,
 ) {
-    let log_metadata_sync_topics =
+    let mut log_metadata_sync_topics =
         TopicFilter::new_unchecked("te/device/main///cmd/software_update/+");
+    log_metadata_sync_topics.add_unchecked("te/device/main///cmd/config_update/+");
 
     let config = LogManagerConfig {
         mqtt_schema: MqttSchema::default(),
@@ -661,6 +662,82 @@ async fn log_types_published_on_software_update_message() -> Result<(), anyhow::
     .await?;
 
     // The log manager should publish the log types again
+    assert_eq!(
+        mqtt.recv().await,
+        Some(
+            MqttMessage::new(&log_reload_topic, r#"{"types":["type_one","type_two"]}"#)
+                .with_retain()
+        )
+    );
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn log_types_published_on_config_update_message() -> Result<(), anyhow::Error> {
+    let tempdir = prepare()?;
+    let (mut mqtt, _fs, _uploader) = spawn_log_manager_actor(tempdir.path()).await;
+
+    let log_reload_topic = Topic::new_unchecked("te/device/main///cmd/log_upload");
+    let config_update_topic = Topic::new_unchecked("te/device/main///cmd/config_update/1234");
+
+    // Skip the initial log types message on startup
+    mqtt.skip(1).await;
+
+    mqtt.send(MqttMessage::new(
+        &config_update_topic,
+        r#"
+        {
+            "status": "init",
+            "updateList": []
+        }"#,
+    ))
+    .await?;
+
+    mqtt.send(MqttMessage::new(
+        &config_update_topic,
+        r#"
+        {
+            "status": "executing",
+            "updateList": []
+        }"#,
+    ))
+    .await?;
+
+    // The log manager does not react to `config_update` in state other than "successful" or "failed"
+    assert!(mqtt.recv().await.is_none());
+
+    // Send a config_update message in terminal state that trigger log types reload
+    let config_update_message = r#"
+        {
+            "status": "successful",
+            "updateList": []
+        }"#;
+    mqtt.send(MqttMessage::new(
+        &config_update_topic,
+        config_update_message,
+    ))
+    .await?;
+
+    // The log manager should publish the log types again
+    assert_eq!(
+        mqtt.recv().await,
+        Some(
+            MqttMessage::new(&log_reload_topic, r#"{"types":["type_one","type_two"]}"#)
+                .with_retain()
+        )
+    );
+
+    // Failed config update should also trigger log types reload
+    mqtt.send(MqttMessage::new(
+        &config_update_topic,
+        r#"
+        {
+            "status": "failed",
+            "updateList": []
+        }"#,
+    ))
+    .await?;
     assert_eq!(
         mqtt.recv().await,
         Some(
--- a/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
+++ b/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
@@ -37,6 +37,21 @@ Supported log types updated on software update
     ...    message_contains=cron::journald
     Should Support Log File Types    cron::journald    includes=${True}
 
+Supported log types updated on config update
+    [Documentation]    Updating any configuration should trigger supported log types update
+    ${config_url}=    Cumulocity.Create Inventory Binary
+    ...    tedge-configuration-plugin
+    ...    tedge-configuration-plugin
+    ...    contents=files=[]
+    ${start_time}=    Get Unix Timestamp
+    ${operation}=    Cumulocity.Set Configuration    tedge-configuration-plugin    url=${config_url}
+    Operation Should Be SUCCESSFUL    ${operation}    timeout=120
+
+    Should Have MQTT Messages
+    ...    topic=te/device/main///cmd/log_upload
+    ...    date_from=${start_time}
+    ...    message_contains=software-management
+
 Log operation journald plugin can return logs for all units
     Should Support Log File Types    all-units::journald    includes=${True}
     ${start_timestamp}=    Get Current Date    UTC    -1 hours    result_format=%Y-%m-%dT%H:%M:%S+0000


crates/extensions/tedge_log_manager/src/lib.rs
```diff
@@ -126,31 +126,41 @@ impl LogManagerBuilder {
...
+                    .unwrap_or(None)
+                    .map(|cmd| cmd.into())
+            } else if log_metadata_sync_topics.accept(&message) {
+                if let Ok(cmd) = GenericCommandState::from_command_message(&message) {
+                    if cmd.is_finished() {
+                        return Some(LogInput::LogMetadataSync(()));
+                    }
+                }
+                return None;
+            } else {
```
albinsuresh, line 177: These changes and the changes to the `subscriptions` functions are only used in the test scope. The actual "subscription" happens through the workflow actor which is captured below in the `IntoIterator` implementation. This was a major source of confusion for me, where the unit tests were working but the actual code wasn't.

Planning to move all these test scoped functions starting from `connect_mqtt` into a separate `test_utils` module in a follow-up commit, to avoid any confusions in future.

crates/extensions/tedge_log_manager/src/lib.rs
```diff
@@ -230,6 +240,21 @@ impl IntoIterator for &LogManagerBuilder {
...
+                if cmd.is_finished() {
+                    Some(LogInput::LogMetadataSync(()))
+                } else {
+                    None
+                }
+            });
+        vec![
+            (OperationType::LogUpload.to_string(), sender.into()),
+            (
+                OperationType::SoftwareUpdate.to_string(),
```
albinsuresh: This is the key change, making the log actor receive `software_update` state transition messages as well.

crates/extensions/tedge_log_manager/src/actor.rs
```diff
@@ -80,6 +81,10 @@ impl Actor for LogManagerActor {
                 LogInput::LogUploadResult((topic, result)) => {
                     self.process_uploaded_log(&topic, result).await?;
                 }
+                LogInput::LogMetadataSync(_) => {
+                    warn!("Received reload types trigger message");
```
albinsuresh: Ignore that `warn` log. To be removed.

crates/extensions/tedge_log_manager/src/config.rs
```diff
@@ -62,6 +63,15 @@ impl LogManagerConfig {
             ChannelFilter::Command(OperationType::LogUpload),
         );
 
+        let mut log_type_reload_topics = mqtt_schema.topics(
```
albinsuresh:
```suggestion
        let mut log_metadata_sync_topics = mqtt_schema.topics(
```

albinsuresh: Resolved by [d78057d](https://github.com/thin-edge/thin-edge.io/pull/3817/commits/d78057d95c98628cfbe58e4abd9564929476bb55)

crates/extensions/tedge_log_manager/src/actor.rs
albinsuresh: Resolved by [d78057d](https://github.com/thin-edge/thin-edge.io/pull/3817/commits/d78057d95c98628cfbe58e4abd9564929476bb55)

crates/extensions/tedge_log_manager/src/tests.rs
```diff
@@ -609,3 +612,61 @@ async fn read_log_from_file_that_does_not_exist() -> Result<(), anyhow::Error> {
...
+        &software_update_topic,
+        r#"
+        {
+            "status": "executing",
+            "updateList": []
+        }"#,
+    ))
+    .await?;
+
+    // The log manager does not react to `software_update` in state than "successful" or "failed"
```
didier-wenzek:
```suggestion
    // The log manager does not react to `software_update` in state other than "successful" or "failed"
```

But shouldn't this even be:


```suggestion
    // The log manager does not react to `software_update` in state other than "successful"
```

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -47,10 +47,19 @@ Software list should be populated during startup with advanced software manageme
...
     Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
     Device Should Have Installed Software    c8y-remote-access-plugin
 
+    # Verify
+    Should Have MQTT Messages
+    ...    topic=te/device/main///cmd/log_upload
+    ...    minimum=1
+    ...    maximum=1
+    ...    date_from=${start_time}
+    ...    message_contains=software-management
```
didier-wenzek: What would be ideal is to test the updated log capabilities in a context where there is actually a new log type. For example, installing a new systemd service which should then be detected by the journald log plugin.

crates/extensions/tedge_log_manager/src/lib.rs
didier-wenzek: > These changes and the changes to the `subscriptions` functions are only used in the test scope. 

This seems to be a legacy of the former log manager which was directly connected to MQTT while with the new setting all the MQTT messages are received by the agent.

> Planning to move all these test scoped functions starting from `connect_mqtt` into a separate `test_utils` module in a follow-up commit, to avoid any confusions in future.

The bare minimum is to mark these methods as `#[cfg(test)]`. Another option is to supply remove all that code and to test the actor using the same connection type as the agent uses. And doing so, actually testing the real thing.

crates/extensions/tedge_log_manager/src/lib.rs
```diff
@@ -230,6 +240,21 @@ impl IntoIterator for &LogManagerBuilder {
             MappingSender::new(self.box_builder.get_sender(), |cmd: GenericCommandState| {
                 LogUploadCmd::try_from(cmd).map(LogInput::LogUploadCmd).ok()
             });
-        vec![(OperationType::LogUpload.to_string(), sender.into())].into_iter()
+        let metadata_sync_sender =
+            MappingSender::new(self.box_builder.get_sender(), |cmd: GenericCommandState| {
+                if cmd.is_finished() {
```
didier-wenzek: Does it make sense to emit updated log capabilities after a failure? I would consider the following:


```suggestion
                if cmd.is_successful() {
```

crates/extensions/tedge_log_manager/src/lib.rs
didier-wenzek: I was about to say that we also have to react on `configuration_update`, but realized then this design is limited. Reacting to `configuration_update` can be done, but what about reacting to MQTT messages published on `signal/sync`? Don't we have to also make the agent broadcasting synchronization messages to all operation handlers?

tests/RobotFramework/tests/cumulocity/software_management/software.robot
albinsuresh: I am adding such a test at the moment.

crates/extensions/tedge_log_manager/src/lib.rs
albinsuresh: I also had started with `is_successful`, but then thought that it makes sense to include `is_failed` as well to cover for partial failures when a single software update includes multiple items and te failure happens after a few are installed. Yes, I agree that 90% of failure cases wouldn't be such and we'd be unnecessarily refreshing the types for all those cases, but it's just a trade-off. Operations in general being rare in a production env, I hope this is fine.

crates/extensions/tedge_log_manager/src/tests.rs
albinsuresh: Response in https://github.com/thin-edge/thin-edge.io/pull/3817#discussion_r2428267367

crates/extensions/tedge_log_manager/src/lib.rs
albinsuresh: Yeah, I did the same thing for `config_update` as well, but you're right, this wouldn't work for the `sync` signal, as this `IntoIterator` impl is designed for operations/commands only, with the `GenericCommandState` input.

I'll have to change the input from `GenericCommandState` to a wrapper enum that can take either a `GenericCommandState` or an `MqttMessage` as well. Since this would be a fairly significant change, I'd prefer to do that in a dedicated PR where the `signal/sync` support is added.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
rina23q: I think we miss the keyword that checks supported log types in c8y. I mean something similar to `Cumulocity.Should Contain Supported Operations`.

crates/extensions/tedge_log_manager/src/tests.rs
albinsuresh: Resolved by [3445aeb](https://github.com/thin-edge/thin-edge.io/pull/3817/commits/3445aeb4ed69b81182e921dcb0915716200f162f)

tests/RobotFramework/tests/cumulocity/software_management/software.robot
albinsuresh: Resolved by [4750d33](https://github.com/thin-edge/thin-edge.io/pull/3817/commits/4750d331684d46a660936ef612fa3cc726b5fb2d)

crates/extensions/tedge_log_manager/src/lib.rs
albinsuresh: Added support for `config_update` in 3f87e0e

crates/extensions/tedge_log_manager/src/lib.rs
albinsuresh: Moved those functions to a test-scoped `impl` block for now. I'll try to rewrite the tests as suggested as well in a follow-up commit.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -46,11 +46,6 @@ Software list should be populated during startup with advanced software manageme
     Should Have MQTT Messages    c8y/s/us    message_contains=140,    minimum=1
     Device Should Have Installed Software    tedge    timeout=120
 
-Install software via Cumulocity
-    ${OPERATION}=    Install Software    c8y-remote-access-plugin    # TODO: Use different package
-    Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
-    Device Should Have Installed Software    c8y-remote-access-plugin
-
```
reubenmiller, line 53: This change seems unrelated to the PR

reubenmiller:
@albinsuresh Just an general question as I couldn't read it from the Rust code...

Does the refreshing of the log plugin lists have any influence on the software update or configuration update operation? For instance if a log plugin is misbehaving, it can't affect the result of the operation right?

tests/RobotFramework/tests/cumulocity/software_management/software.robot
reubenmiller: Will this change be reverted then?

albinsuresh: Yes, it will be reverted. I meant to delete that additional assertion that I had added in a previous commit (which was moved to another test), but accidentally ended up deleting the whole test instead.

crates/extensions/tedge_log_manager/src/lib.rs
```diff
@@ -230,6 +247,25 @@ impl IntoIterator for &LogManagerBuilder {
...
+                }
+            });
+        vec![
+            (OperationType::LogUpload.to_string(), sender.into()),
+            (
+                OperationType::SoftwareUpdate.to_string(),
+                metadata_sync_sender.sender_clone(),
+            ),
+            (
+                OperationType::ConfigUpdate.to_string(),
```
albinsuresh: This line is the causing the `Config update request not processed when operation is disabled for tedge-agent` system test to fail, as this makes the `WorkflowActor` "think" that this `LogManagerActor` is a built-in operation handler for `config_update` operations rather than treating it just as an "observer" of `config_update` state transitions. So, when the `WorkflowActor` receives the `init` message, it changes it to `scheduled` and `executing` states and sends it to the `LogManagerActor` which just ignores those messages as it really can't "handle" `config_update` operations.

To fix this, we need to bring in this distinction between an operation handler and a casual observer so that `LogManagerActor` can register just as the observer and not the handler.

didier-wenzek: > To fix this, we need to bring in this distinction between an operation handler and a casual observer so that LogManagerActor can register just as the observer and not the handler.

This has to be done with this other point in mind:

> Yeah, I did the same thing for `config_update` as well, but you're right, this wouldn't work for the `sync` signal, as this `IntoIterator` impl is designed for operations/commands only, with the `GenericCommandState` input.

I would try to address both requirements with a single abstraction. The need is not really for the `LogManagerActor` to observe other operations, but to be warned that a synchronization is required. Can this be handled by a function  (provided by the `LogManagerActor` to the `WorkflowActor`) to tell which MQTT message triggers a  `signal/sync`.

crates/core/tedge_actors/src/builders.rs
```diff
@@ -223,6 +223,14 @@ pub trait RuntimeRequestSink {
     fn get_signal_sender(&self) -> DynSender<RuntimeRequest>;
 }
 
+pub type SyncSignal = ();
+
+/// The [Builder] of an [Actor](crate::Actor) must implement this trait
+/// to receive sync signals that can be used to sync any actor metadata with the external world.
+pub trait SyncSignalSink {
```
albinsuresh: I'm not very happy with the placement of this trait here. But, couldn't find a better spot due to the dependency on the `DynSender` type. Suggestions welcome.

crates/extensions/tedge_log_manager/src/lib.rs
albinsuresh: Resolved by [f9ecb4c](https://github.com/thin-edge/thin-edge.io/pull/3817/commits/f9ecb4c36ecd10d8c4cb65bcf50555a3940bcbd4)

The `SyncSignalSink` abstraction should make it easy for us to integrate the `signal/sync` handling in future.

crates/core/tedge_actors/src/builders.rs
didier-wenzek: What is the difference between this trait and `MessageSink<SyncSignal>` ?

albinsuresh: Oh yeah, that's true. I'll replace this with `MessageSink<SyncSignal>`.

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -428,10 +430,16 @@ impl WorkflowActor {
...
+            self.finalize_builtin_command_update(new_state).await?;
+
+            Ok(())
+        } else {
+            // As not finalized, the builtin state is sent back
+            // to the builtin operation actor for further processing.
+            let builtin_state = new_state.clone();
+            Ok(self.builtin_command_dispatcher.send(builtin_state).await?)
         }
-        Ok(self.builtin_command_dispatcher.send(new_state).await?)
```
didier-wenzek: Why  duplicate `self.builtin_command_dispatcher.send()`  in both branches?

crates/core/tedge_actors/src/builders.rs
albinsuresh: Resolved by c5317dd

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -253,9 +253,14 @@ impl WorkflowActor {
 
                 // Move to the next state to await the builtin operation outcome
                 let new_state = state.update(handlers.on_exec.clone());
+                info!("Publishing new state: {} to MQTT", &new_state.status);
```
albinsuresh: This log and the next one are remnants of some debugging. To be removed before merge.

justfile
```diff
@@ -84,7 +84,7 @@ format: check-tools
     taplo fmt
 
     if [ ! -d tests/RobotFramework/.venv ]; then
-        just -f {{justfile()}} setup-integration-test
+        just -f {{justfile()}} `setup-integration-test`
```
didier-wenzek: Better to fix that in a specific PR.

didier-wenzek (approved):
Approved. The updated design is by far cleaner.

- Don't forget: https://github.com/thin-edge/thin-edge.io/pull/3817/files#r2439285063

justfile
albinsuresh: That was another accidental change. Have reverted it.

PR #3819, opened 2025-10-14
refactor: handle renamed fields correctly in profiled configurations

jarhodes314:
## Proposed changes
While trying to work on a new feature, I discovered renamed fields don't work as expected when stored inside renamed fields. They are read correctly, but we get unknown configuration key warnings when reading tedge-config, and (for the same underlying reason) the `group.profiles.name.field` syntax doesn't respect the renaming.

More concretely:
```rust
define_tedge_config! {
    #[tedge_config(multi)]
    mapper: {
        #[tedge_config(rename = "type")]
        ty: String,
    }
}
```

will successfully parse `mapper.profiles.c8y.ty` instead of `mapper.profiles.c8y.type`. Using the `--profile` flag works as expected though: `tedge config get --profile c8y mapper.type` succeeds (but emits the warning `Unknown configuration field "mapper.profiles.c8y.type" from toml file /etc/tedge/tedge.toml`).

Labelling this as an improvement rather than bugfix since the issue only affects possible future changes to configurations, not any existing configurations.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config_macros/impl/src/query.rs
+++ b/crates/common/tedge_config_macros/impl/src/query.rs
@@ -926,8 +926,8 @@ fn enum_variant(segments: &VecDeque<&FieldOrGroup>) -> ConfigurationKey {
                 FieldOrGroup::Multi(m) => {
                     format!("{}(?:[\\._]profiles[\\._]([^\\.]+))?", m.ident)
                 }
-                FieldOrGroup::Field(f) => f.ident().to_string(),
-                FieldOrGroup::Group(g) => g.ident.to_string(),
+                FieldOrGroup::Field(f) => f.name().to_string(),
+                FieldOrGroup::Group(g) => g.name().to_string(),
             })
             .collect::<Vec<_>>()
             .join("[\\._]");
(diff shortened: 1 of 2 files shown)


crates/common/tedge_config_macros/impl/src/query.rs
```diff
@@ -1639,6 +1639,71 @@ mod tests {
...
+                        _ => {
+                            if let Ok(key) = <ReadOnlyKey as ::std::str::FromStr>::from_str(value) {
+                                Err(ParseKeyError::ReadOnly(key))
+                            } else {
+                                Err(ParseKeyError::Unrecognised(value.to_owned()))
+                            }
+                        }
+                    };
+                    if let Some(captures) = ::regex::Regex::new(
+                            "^mapper(?:[\\._]profiles[\\._]([^\\.]+))?[\\._]type$",
```
jarhodes314, line 1688: **Not a suggestion, just illustrating what the code did previously**

Previously, the macro outputted the same code, except with this change:
```suggestion
                            "^mapper(?:[\\._]profiles[\\._]([^\\.]+))?[\\._]ty$",
```

PR #3821, opened 2025-10-15
fix: Avoid spurious unknown pkid warning on bridge healthcheck

jarhodes314:
## Proposed changes
Prior to these changes, whenever it runs, the built-in bridge produces two warnings like this:

```
2025-10-15T13:40:55.752977Z  WARN MQTT bridge: [local] Received ack for unknown pkid=2
```

This is caused by the healthcheck message, which we ignore in the ack processing as we don't need to forward the acknowledgement to the other broker like we do for forwarded messages. Since this case isn't actually the bridge or broker misbehaving, this PR prevents the bridge warning when it receives the ack for the healthcheck message.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


didier-wenzek (approved):
Approved. Thank you for taking care of what could be seen as a detail, but actually improve the confidence we can have on thin-edge.

Issue #3807, opened 2025-10-06
Support connecting tedge components to a local mqtt broker with username/password authentication


reubenmiller:
**Is your feature request related to a problem? Please describe.**

thin-edge.io. components only support connecting to a local MQTT broker that either has no authentication, or cert based client authentication.

In scenarios where users are using an already configured local MQTT broker and it is fairly common that username/password authentication is used (as username/password requires less setup/Maintenance then cert based authentication).

Allow users to connect thin-edge.io components to a local MQTT broker which is using a username/password.

**Describe the solution you'd like**

Allow users to connect thin-edge.io components to a local MQTT broker which is using a username/password.

For example, the following are the suggested configuration value to allow users to configure using username/password based authentication when connecting to the local MQTT broker.

```sh
tedge config set mqtt.client.auth.username "<username>"
tedge config set mqtt.client.auth.password "<password>"
```

The above values should also be able to read from the env variables

* TEDGE_MQTT_CLIENT_AUTH_USERNAME
* TEDGE_MQTT_CLIENT_AUTH_PASSWORD

**Open questions**

* Should the password be stored in the tedge.toml, or should the tedge.toml only contain a file path to a file which contains the password so the file can be made to only be readable by the tedge user?

**Describe alternatives you've considered**

**Additional context**

PR #3823, opened 2025-10-17
feat: Support username/password for local broker authentication

rina23q:
## Proposed changes
Support username/password authentication to the MQTT local broker.

---

Prepare a `/etc/tedge/.password` file. The first line is read as the password.
```
testpassword
```

Then, configure these parameters.
```
tedge config set mqtt.client.auth.username testuser
tedge config set mqtt.client.auth.password_file /etc/tedge/.password
```

If encrypted connection is activated in the broker, configure  any secure port (e.g. 8883) and give the CA certificate file or the directory (below is example)
```
tedge config set mqtt.client.port 8883
tedge config set mqtt.client.auth.ca_file /etc/mosquitto/certs/ca.crt
```

Todo:
- [x] Confirm `tedge mqtt pub/sub` works
- [x] Confirm `tedge-agent` & `tedge-mapper-c8y` work with mosquitto bridge
- [x] Confirm `tedge-agent` & `tedge-mapper-c8y` works with build-in bridge
- [x] Confirm it works with non-c8y cloud
- [x] Write system tests
- [x] Extend the user guide

## Types of changes

## Paste Link to the issue
#3807 

## Checklist

### DIFF
--- a/docs/src/operate/security/mqtt-tls-configuration.md
+++ b/docs/src/operate/security/mqtt-tls-configuration.md
@@ -5,11 +5,17 @@ description: Setting up TLS for secure local MQTT communication
 ---
 
 %%te%% supports certificate-based authentication when communicating with
-an MQTT broker. Three levels of security are supported:
+an MQTT broker. Four levels of security are supported:
 
 1. No authentication (default)
 2. Server authentication
 3. Server + client authentication
+4. Username and password authentication
+
+:::note
+Username and password authentication can be used as the sole authentication method, or it can be layered on top of server or server + client authentication for enhanced security.
+See [here](#username-password-authentication) for details.
+:::
 
 ## MQTT Configuration
 
@@ -225,6 +231,42 @@ openssl x509 -req \
     -out client.crt \
     -days 365
 ```
+
+## Username/Password authentication {#username-password-authentication}
+
+When a local MQTT broker requires username/password authentication,
+the %%te%% components need to provide this username and password.
+
+Configure the username and the path to the password file using `tedge config`:
+
+```sh
+sudo tedge config set mqtt.client.auth.username "USERNAME"
+sudo tedge config set mqtt.client.auth.password_file "PATH_TO_PASSWORD_FILE"
+```
+
+In the password file, write the password on the first line. The second line and any subsequent lines will be ignored.
+
+```sh title="file: PATH_TO_PASSWORD_FILE"
+YOUR_PASSWORD
+```
+
+For security reasons, we recommend changing the password file's ownership to `tedge:tedge`
+and setting the permissions to `600`.
+This ensures that only %%te%% services can read the password.
+
+```sh
+sudo chown tedge:tedge "PATH_TO_PASSWORD_FILE"
+sudo chmod 600 "PATH_TO_PASSWORD_FILE"
+```
+
+In addition, if TLS is enabled on the broker, configure the secure port (8883 is the standard port) and provide the path to the trusted CA certificate file using `tedge config`:
+```sh
+sudo tedge config set mqtt.client.port 8883
+sudo tedge config set mqtt.client.auth.ca_file "PATH_TO_CA_CERTIFICATE"
+```
+
+If any %%te%% services (e.g. tedge-agent) are already running, they must be restarted for the changes to take effect.
+
 
 ## Next steps
 


crates/common/mqtt_channel/src/config.rs
```diff
@@ -376,3 +426,25 @@ impl Config {
...
+    let first_line = match reader.lines().next() {
+        Some(Ok(password)) => Ok(password),
+        Some(Err(error)) => Err(CertificateError::IoError {
+            error,
+            path: path.as_ref().to_owned(),
+        }),
+        None => Err(CertificateError::IoError {
+            path: path.as_ref().to_owned(),
+            error: std::io::Error::other("File content is empty"),
+        }),
```
rina23q: This error has to be removed as the MQTT specification states that you can send a username without password. However, both username and password are required for `MqttOptions::set_credentials()` in rumqttc, so I'm not sure how empty password is addressed in rumqttc.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -107,13 +110,19 @@ impl Default for AuthenticationConfig {
     }
 }
 
+#[derive(Debug, Clone)]
+enum ClientAuthConfig {
+    Cert(ClientAuthCertConfig),
+    User(ClientAuthUserConfig),
```
didier-wenzek: Password is clearer.


```suggestion
    Pass(ClientAuthPassConfig),
```

crates/common/mqtt_channel/src/config.rs
```diff
@@ -130,6 +139,12 @@ impl zeroize::Zeroize for PrivateKey {
     }
 }
 
+#[derive(Debug, Clone)]
+struct ClientAuthUserConfig {
+    username: String,
+    password: String,
```
didier-wenzek: I would consider to use [`zeroize`](https://docs.rs/zeroize/latest/zeroize/)


```suggestion
    password: Zeroizing<String>,
```

crates/common/tedge_config/src/tedge_toml/tedge_config/mqtt_config.rs
```diff
@@ -108,6 +121,21 @@ impl MqttAuthConfig {
 
         Ok(Some(client_config))
     }
+
+    // Return username and password if they are set
+    pub fn username_and_password(&self) -> Option<(String, String)> {
```
didier-wenzek: Better to return a `ClientAuthUserConfig` as defined in the `mqtt_channel::config` module.


```suggestion
    pub fn username_and_password(&self) -> Option<mqtt_channel::config::ClientAuthUserConfig> {
```

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -94,6 +94,10 @@ impl MqttBridgeActorBuilder {
         );
         // TODO cope with certs but not ca_dir, or handle that case with an explicit error message?
         let auth_config = tedge_config.mqtt_client_auth_config();
+
+        if let Some((username, password)) = auth_config.username_and_password() {
+            local_config.set_credentials(username, password);
+        }
```
didier-wenzek: I would design the config API in such a way that no change is required for the clients connecting the local MQTT bus. So that, all the required auth settings are captured in a single method call to `mqtt_client_auth_config` whatever their kind (certificates or password based).

crates/core/tedge/src/cli/mqtt/subscribe.rs
```diff
@@ -66,8 +66,15 @@ async fn subscribe(cmd: &MqttSubscribeCommand) -> Result<(), anyhow::Error> {
...
+    if let Some(MqttAuthClientConfig::Cert(client_auth)) = cmd.client_auth_config.as_ref() {
+        config.with_client_auth_cert_key(&client_auth.cert_file, &client_auth.key_file)?;
+    }
+
+    if let Some(MqttAuthClientConfig::User(client_auth_user)) = cmd.client_auth_config.as_ref() {
+        config.with_client_auth_user_pass(
+            &client_auth_user.username,
+            client_auth_user.password_file.clone(),
+        )?;
     }
```
didier-wenzek: An MQTT client should be agnostic to how the user is authenticated. I would move this logic into the tedge config so we can write:

```rust
if let Some(client_auth) = cmd.client_auth_config.as_ref() {
        config.with_client_auth(client_auth)?;
```

This will avoid code duplication across `tedge mqtt sub`, `pub`, mappers and the agent.

crates/common/mqtt_channel/src/config.rs
rina23q: Agreed.
 
By the way, I created this enum because at first I thought user would use **either** certificate or username/password for authentication. However, after I saw the MQTT specifications, client certificate is in the Transport Layer and username/password is in the Application Layer. Then, I started feeling, we should support the case using **both**. Can you confirm this?

didier-wenzek: While a broker can accept both, a thin-edge MQTT client only needs to be configured to authenticate itself either using a certificate or a password, not both. Using a certificate is recommended.

reubenmiller: But the username can still be required when using certificates right? At least that is my experience when using the Azure EventGrid MQTT endpoint (not that this is a cloud related topic, but more to give an example of an MQTT broker being configured a specific way).

rina23q: > But the username can still be required when using certificates right?

I think so. For example, mosquitto has `use_identity_as_username` settings, if it's set to `true`, the certificate's CN is used as a username. Password won't be set.

reubenmiller: Ok good to know. I don't think we need to go to the level of having a "use_identity_as_username" equivalent in thin-edge.io...I'd much rather let the users configure the values themselves, and then possibly allow users to set a value using variables `${.common_name}` (but we can do template evaluation that later).

docs/src/operate/security/mqtt-tls-configuration.md
```diff
@@ -11,6 +11,8 @@ an MQTT broker. Three levels of security are supported:
 2. Server authentication
 3. Server + client authentication
 
+In addition, authentication with username and password is supported. See [here](#username-password-authentication) for details.
```
albinsuresh: I'd have added that to the above list, especially since `No authentication` is a part of that list.

docs/src/operate/security/mqtt-tls-configuration.md
```diff
@@ -226,6 +228,33 @@ openssl x509 -req \
     -days 365
 ```
 
+## Username/Password authentication {#username-password-authentication}
+
+When a local MQTT broker requires username/password authentication,
+the %%te%% components also need to know this username and password.
```
albinsuresh:
```suggestion
the %%te%% components need to provide this username and password.
```

docs/src/operate/security/mqtt-tls-configuration.md
```diff
@@ -226,6 +228,33 @@ openssl x509 -req \
...
+sudo tedge config set mqtt.client.auth.username "USERNAME"
+sudo tedge config set mqtt.client.auth.password_file "PATH_TO_PASSWORD_FILE"
+```
+
+In the password file, write the password on the first line. The second line and any subsequent lines will be ignored.
+
+```sh title="file: <file_path>"
+YOUR_PASSWORD
+```
+
```
albinsuresh, line 252: Add one line about restricting the access levels(`600`) to that file as soon as it is created?

tests/RobotFramework/tests/mqtt/local_broker_authentication/local_broker_authentication.robot
```diff
@@ -0,0 +1,131 @@
...
+    Connect Mapper    c8y
+
+    # Child
+    ${CONTAINER_2}=    Setup    bootstrap_args=--no-secure    register=${False}    connect=${False}
+    Set Test Variable    $CONTAINER_2
+
+    Configure MQTT client    host=${CONTAINER_1_HOSTNAME}    port=1884
+    Start Service    tedge-agent
+
+TLS Setup
```
albinsuresh, line 99: Optional suggestion: Instead of two variants `TLS Setup` and `No TLS Setup`, why not just keep a single `Setup` keyword that takes `tld_enabled` as an argument? Especially since these keywords are already taking the `use_builtin_bridge` argument?

tests/RobotFramework/tests/mqtt/local_broker_authentication/local_broker_authentication.robot
```diff
@@ -0,0 +1,131 @@
...
+
+    Cumulocity.Should Have Services    name=tedge-mapper-c8y    status=up
+    Cumulocity.Should Have Services    name=tedge-agent    status=up
+    Cumulocity.Should Have Services    name=c8y-firmware-plugin    status=up
+
+No TLS Setup
+    [Arguments]    ${use_builtin_bridge}
+
+    # Parent
+    ${CONTAINER_1}=    Setup    bootstrap_args=--no-secure    register=${True}    connect=${False}
```
albinsuresh:
```suggestion
    ${PARENT_SN}=    Setup    bootstrap_args=--no-secure    register=${True}    connect=${False}
```
We've always used these naming conventions to distinguish the containers easily.

tests/RobotFramework/tests/mqtt/local_broker_authentication/local_broker_authentication.robot
```diff
@@ -0,0 +1,131 @@
...
+    Execute Command    sudo tedge config set mqtt.bridge.built_in ${use_builtin_bridge}
+    Execute Command    mosquitto_passwd -c -b /etc/mosquitto/pwfile testuser testpassword
+
+    Transfer To Device    ${CURDIR}/unencrypted-listener.conf    /etc/mosquitto/conf.d/
+    Configure MQTT client    host=127.0.0.1    port=1884
+
+    Connect Mapper    c8y
+
+    # Child
+    ${CONTAINER_2}=    Setup    bootstrap_args=--no-secure    register=${False}    connect=${False}
```
albinsuresh:
```suggestion
    ${CHILD_SN}=    Setup    bootstrap_args=--no-secure    register=${False}    connect=${False}
```

crates/common/mqtt_channel/src/config.rs
```diff
@@ -314,13 +329,35 @@ impl Config {
...
+    }
+
+    /// Provide client username/password for authentication.
+    pub fn with_client_auth_user_pass<P: AsRef<Path>>(
+        &mut self,
+        username: &str,
+        pass_file: P,
+    ) -> Result<&mut Self, CertificateError> {
+        debug!(target: "MQTT", "Using client username: {username}");
+        debug!(target: "MQTT", "Using client password: {}", pass_file.as_ref().display());
```
albinsuresh:
```suggestion
        debug!(target: "MQTT", "Using client password from: {}", pass_file.as_ref().display());
```

crates/common/mqtt_channel/src/config.rs
```diff
@@ -107,13 +110,19 @@ impl Default for AuthenticationConfig {
     }
 }
 
+#[derive(Debug, Clone)]
+enum ClientAuthConfig {
+    Cert(ClientAuthCertConfig),
+    User(ClientAuthPassConfig),
```
albinsuresh:
```suggestion
    Cred(ClientAuthCredConfig),
```
Short for "credentials".

crates/core/tedge/src/cli/mqtt/publish.rs
```diff
@@ -67,8 +67,16 @@ async fn publish(cmd: &MqttPublishCommand) -> Result<(), anyhow::Error> {
...
+    if let Some(MqttAuthClientConfig::Cert(client_auth_cert)) = cmd.client_auth_config.as_ref() {
+        config
+            .with_client_auth_cert_key(&client_auth_cert.cert_file, &client_auth_cert.key_file)?;
+    }
+
+    if let Some(MqttAuthClientConfig::User(client_auth_user)) = cmd.client_auth_config.as_ref() {
+        config.with_client_auth_user_pass(
+            &client_auth_user.username,
+            client_auth_user.password_file.clone(),
+        )?;
```
albinsuresh:
```suggestion
if let Some(client_auth_config) = cmd.client_auth_config.as_ref() {
    match client_auth_config {
        MqttAuthClientConfig::Cert(client_auth_cert) => {
            config
                .with_client_auth_cert_key(&client_auth_cert.cert_file, &client_auth_cert.key_file)?;
        }
        MqttAuthClientConfig::User(client_auth_user) => {
            config.with_client_auth_user_pass(
                &client_auth_user.username,
                client_auth_user.password_file.clone(),
            )?;
        }
    }
}```

File: crates/common/mqtt_channel/src/config.rs (hunk 14 above)
Comment (albinsuresh) [reply crates/common/mqtt_channel/src/config.rs]: Okay, I just saw the older [discussion](https://github.com/thin-edge/thin-edge.io/pull/3823#discussion_r2452010235) on this. For some reason, it didn't appear in the files view during the review.

File: crates/common/mqtt_channel/src/config.rs (hunk 2 above)
Comment (rina23q) [reply crates/common/mqtt_channel/src/config.rs]: I'll refactor the code so that user has an opportunity to set username (and password if necessary) together with client certificates since it's possible in the MQTT specifications. I believe it's better thin-edge supports whatever the MQTT specifications allow.

File: tests/RobotFramework/tests/mqtt/local_broker_authentication/local_broker_authentication.robot (hunk 10 above)
Comment (rina23q) [reply tests/RobotFramework/tests/mqtt/local_broker_authentication/local_broker_authentication.robot:99]: `use_builtin_bridge` makes really one-liner difference (`sudo tedge config set mqtt.bridge.built_in ${use_builtin_bridge}`), while TLS/no TLS need more different setup (bootstrap_args, port, set/unset some tedge config keys, different listener file). Then eventually I thought it would be easier to read, if there are two different Setup keys.

File: crates/common/mqtt_channel/src/config.rs (hunk 1 above)
Comment (rina23q) [reply crates/common/mqtt_channel/src/config.rs]: I checked how `MqttOptions::set_cedentials()` behaves in the packet level. 

In the MQTT specs level, there are username flag and password flag. So, "empty password" and "password not set" are different. The former,  the password flag is set but password is empty string. The latter, the password flag itself isn't set.

"Empty password" exmaple:
```sh
mosquitto_pub -h 127.0.0.1 -u testuser -P "" -t test -p 1883 -m message
```

"Password not set" example:
```sh
mosquitto_pub -h 127.0.0.1 -u testuser -t test -p 1883 -m message
```

Back to rumqttc. Their API offers only `set_credentials(username, password)`. I provided the empty string to the second argument as below.
```rust
    mqttoptions.set_credentials("username", "");
```
Then, I ran the rumqttc client and confirmed the password flag was not set in the CONNECT packet. This means, rumqttc treats "empty password" equals to "password not set".

crates/common/mqtt_channel/src/config.rs
rina23q: This enum no longer exists, so I would mark this comment resolved.

crates/core/tedge/src/cli/mqtt/publish.rs
rina23q: This conditional arm no longer exists. So I would mark this comment resolved.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -93,10 +93,14 @@ impl MqttBridgeActorBuilder {
             tedge_config.mqtt.client.port.into(),
         );
         // TODO cope with certs but not ca_dir, or handle that case with an explicit error message?
-        let auth_config = tedge_config.mqtt_client_auth_config();
-        let local_tls_config = auth_config.to_rustls_client_config().unwrap();
-        if let Some(tls_config) = local_tls_config {
-            local_config.set_transport(Transport::tls_with_config(tls_config.into()));
+        let auth_config: mqtt_channel::AuthenticationConfig =
+            tedge_config.mqtt_client_auth_config().try_into().unwrap();
```
rina23q: This `unwrap()` is not safe. However, the original code also used `unwrap()` for loading certificates as the comment on line 82 says. To address this error correctly, more code changes will be required. I didn't, because it's out of scope of this PR.

docs/src/operate/security/mqtt-tls-configuration.md
rina23q: `600` is questionable because this file also has be read by `tedge mqtt pub/sub`. Given that the file is owned by `tedge:tedge`, then `tedge-agent`, `tedge-mapper` etc. are okay. However, normal users cannot use `tedge mqtt pub/sub`.

docs/src/operate/security/mqtt-tls-configuration.md
rina23q: Changed in e2c9962572885f88325151bb2661eb24c40f784c

docs/src/operate/security/mqtt-tls-configuration.md
rina23q: Applied in e2c9962572885f88325151bb2661eb24c40f784c

tests/RobotFramework/tests/mqtt/local_broker_authentication/local_broker_authentication.robot
rina23q: Applied in 8fd08377b5666bbfa168999facd0890871bbe2bf

tests/RobotFramework/tests/mqtt/local_broker_authentication/local_broker_authentication.robot
rina23q: Applied in 8fd08377b5666bbfa168999facd0890871bbe2bf

crates/core/tedge/src/cli/mqtt/subscribe.rs
rina23q: Did refactoring to remove code duplications regarding setting credentials and tls config in fa0b8a901caa8dc0c8d65d10f5368d26b619d410

crates/extensions/tedge_mqtt_bridge/src/lib.rs
rina23q: Changed the API in https://github.com/thin-edge/thin-edge.io/commit/fa0b8a901caa8dc0c8d65d10f5368d26b619d410

crates/common/mqtt_channel/src/config.rs
rina23q: Now username/password can be set together with certificate and key. https://github.com/thin-edge/thin-edge.io/commit/fa0b8a901caa8dc0c8d65d10f5368d26b619d410

crates/common/mqtt_channel/src/config.rs
rina23q: Rephrased it together with other debug messages in feec2ba3771116cdc5e77891a0164d5405d54fbe.

crates/common/mqtt_channel/src/config.rs
```diff
@@ -91,29 +90,88 @@ pub struct BrokerConfig {
...
         }
     }
 }
 
-#[derive(Debug, Clone)]
-enum ClientAuthConfig {
-    Cert(ClientAuthCertConfig),
-    User(ClientAuthPassConfig),
+impl AuthenticationConfig {
+    pub fn get_mut_cert_store(&mut self) -> &mut rustls::RootCertStore {
```
albinsuresh:
```suggestion
    pub fn get_cert_store_mut(&mut self) -> &mut rustls::RootCertStore {
```
That's he naming convention that I've seen in Rust like `iter_mut`.

docs/src/operate/security/mqtt-tls-configuration.md
albinsuresh: Good point. Wondering if we can make this file owned by the current user and the `tedge` group and no access to others.

crates/common/mqtt_channel/src/lib.rs
```diff
@@ -30,7 +30,7 @@
 #![forbid(unsafe_code)]
 
 mod channel;
-mod config;
+pub mod config;
```
didier-wenzek: I would remove this `pub`, to be consistent with the structure of this crate.

There is a `pub use config::*;` at the crate level.

crates/core/tedge/src/cli/mqtt/publish.rs
```diff
@@ -61,15 +58,7 @@ async fn publish(cmd: &MqttPublishCommand) -> Result<(), anyhow::Error> {
...
-    if let Some(ca_file) = &cmd.ca_file {
-        config.with_cafile(ca_file)?;
-    }
-    if let Some(ca_dir) = &cmd.ca_dir {
-        config.with_cadir(ca_dir)?;
-    }
-    if let Some(client_auth) = cmd.client_auth_config.as_ref() {
-        config.with_client_auth(&client_auth.cert_file, &client_auth.key_file)?;
-    }
+    config.with_client_auth(cmd.auth_config.clone().try_into()?)?;
```
didier-wenzek, line 61: This is a nice improvement.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
didier-wenzek: At least replace this `unwrap()` by an `expect("<a hint of what is wrong>")`

crates/common/mqtt_channel/src/config.rs
```diff
@@ -376,3 +404,22 @@ impl Config {
...
+    let first_line = match reader.lines().next() {
+        Some(Ok(password)) => Ok(password),
+        Some(Err(error)) => Err(CertificateError::IoError {
+            error,
+            path: path.as_ref().to_owned(),
+        }),
+        None => Ok("".to_string()),
+    }?;
+
+    Ok(first_line)
```
didier-wenzek:
```suggestion
     match reader.lines().next() {
        Some(Ok(password)) => Ok(password),
        Some(Err(error)) => Err(CertificateError::IoError {
            error,
            path: path.as_ref().to_owned(),
        }),
        None => Ok("".to_string()),
    }
```

crates/common/mqtt_channel/src/config.rs
```diff
@@ -88,35 +90,114 @@ pub struct BrokerConfig {
...
+                cert_config.key.deref().0.clone_key(),
+            )?,
+            None => tls_config.with_no_client_auth(),
+        };
+        Ok(Some(tls_config))
+    }
+
+    /// When the password is empty, this returns an empty string.
+    /// This is because `rumqttc::MqttOptions::set_credentials()` always requires a value for password.
+    pub fn get_credentials(&self) -> Option<(String, String)> {
```
didier-wenzek: To have a consistent use of zeroize.


```suggestion
    pub fn get_credentials(&self) -> Option<(String, Zeroize<String>)> {
```

crates/common/mqtt_channel/src/lib.rs
rina23q: Indeed this `pub` was useless :/

crates/common/mqtt_channel/src/config.rs
rina23q: Renamed in a2967abe786384db544c6dcf7a63ed4146be2da5

crates/extensions/tedge_mqtt_bridge/src/lib.rs
rina23q: Done in a2967abe786384db544c6dcf7a63ed4146be2da5

crates/common/mqtt_channel/src/config.rs
```diff
@@ -376,3 +404,22 @@ impl Config {
         Ok(mqtt_options)
     }
 }
+
+/// Read the first line of the given file and return it.
+pub fn read_password(path: impl AsRef<Path>) -> Result<String, CertificateError> {
```
didier-wenzek: To have a consistent use of zeroize.


```suggestion
pub fn read_password(path: impl AsRef<Path>) -> Result<Zeroize<String>, CertificateError> {
```

didier-wenzek (approved):
Approved with 3 suggestions

docs/src/operate/security/mqtt-tls-configuration.md
didier-wenzek: > Add one line about restricting the access levels(600) to that file as soon as it is created?

This is a good suggestion, even if this prevent regular users to `tedge mqtt pub/sub` because somehow this is the point to have passwords and private keys.

crates/common/mqtt_channel/src/config.rs
rina23q: Changed in 3fa0c0d50f4c39daa91e0bacc16eff645cc99cdf

crates/common/mqtt_channel/src/config.rs
rina23q: Changed in 3fa0c0d50f4c39daa91e0bacc16eff645cc99cdf

crates/common/mqtt_channel/src/config.rs
rina23q: Changed in 3fa0c0d50f4c39daa91e0bacc16eff645cc99cdf

jarhodes314 (approved):

docs/src/operate/security/mqtt-tls-configuration.md
rina23q: Updated the doc in 27d1ab56e001a7481d1036c6f6d4eccd3c422ef3

docs/src/operate/security/mqtt-tls-configuration.md
```diff
@@ -250,6 +250,15 @@ In the password file, write the password on the first line. The second line and
 YOUR_PASSWORD
 ```
 
+For the security reasons, we recommend changing the password file's ownership to `tedge:tedge`
```
didier-wenzek:
```suggestion
For security reasons, we recommend changing the password file's ownership to `tedge:tedge`
```

rina23q: Applied it and rebased it. (since the required change is tiny)

PR #3830, opened 2025-10-22
feat: add dmesg log plugin

reubenmiller:
## Proposed changes

Add a dmesg log plugin to the out-of-the-box features. The dmesg plugin only provide one type called "all" due to the vast differences in the dmesg implementations (e.g. busybox/bsd don't support filtering by log facility or datetime).

The plugin will attempt to call dmesg with a datetime filter, however if it fails for any reason, then dmesg will be called without any arguments (and in this case all the messages will be returned). This is due to the various implementations of dmesg.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/configuration/contrib/log-plugins/dmesg
+++ b/configuration/contrib/log-plugins/dmesg
@@ -1,0 +1,138 @@
+#!/bin/sh
+set -eu
+
+NOT_SUPPORTED=2
+
+usage() {
+    cat <<EOT
+dmesg log plugin for thin-edge.io
+
+IMPORTANT: Some dmesg implementations don't support date/time filters, so it will fallback to
+calling dmesg without any filtering should the datetime filters fail for any reason.
+
+The log plugin allows users to access kernel messages from dmesg.
+
+Usage:
+  $0 list
+      List all log types supported by this plugin
+
+  $0 get <log-type> [options]
+      Fetch logs for the given facility (<log-type>)
+
+Options for "get":
+  --since <timestamp>   Show logs since the given timestamp (seconds since epoch)
+  --until <timestamp>   Show logs up to the given timestamp  (seconds since epoch)
+  --help                Show this help message and exit
+
+Examples:
+  $0 list
+  $0 get all --since 1696250000 --until 1696260000
+EOT
+}
+
+is_busy_box() {
+    if dmesg --help 2>&1 | grep -q "facility"; then
+        return 1
+    fi
+    return 0
+}
+
+list_log_types() {
+    if ! command -V dmesg >/dev/null 2>&1; then
+        exit "$NOT_SUPPORTED"
+    fi
+
+    # Note: The dmesg implementations vary a lot on different
+    # operating systems, so facility filtering is difficult to provide
+    printf 'all\n'      # Default type
+}
+
+format_timestamp() {
+    value="$1"
+    case "$value" in
+        ''|*[!0-9]*)
+            echo "$value"
+            ;;
+        *)
+            # convert unix timestamp to iso date, as dmesg does not support unix timestamps
+            date -d @"$value" +%Y-%m-%d\ %H:%M:%S
+            ;;
+    esac
+}
+
+get_logs() {
+    # Note: the log type is unused, but still enforce the cli contract
+    if [ $# -lt 1 ]; then
+        echo "Error: 'get' command expects a mandatory <log-type> arguments" >&2
+        usage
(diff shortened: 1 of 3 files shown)


configuration/contrib/log-plugins/dmesg
```diff
@@ -0,0 +1,164 @@
...
+    # as busybox implementations don't seem to support it
+    if ! is_busy_box; then
+        printf 'kern\n'     # kernel messages
+        printf 'user\n'     # random user-level messages
+        printf 'mail\n'     # mail system
+        printf 'daemon\n'   # system daemons
+        printf 'auth\n'     # security/authorization messages
+        printf 'syslog\n'   # messages generated internally by syslogd
+        printf 'lpr\n'      # line printer subsystem
+        printf 'news\n'     # network news subsystem
```
albinsuresh: The standard set seems to be [15](https://en.wikipedia.org/wiki/Syslog) although I came across a few other blogs where that number was 11. But most of the online docs had 15 listed.

rina23q (approved):

configuration/contrib/log-plugins/dmesg
reubenmiller: @rina23q  @albinsuresh Ok, since the list seems to vary significantly (depends on the dmesg implementation), I'll refactor to remove all the types (as it will be more predictable for users), and just have a single "all" type.

Do you both agree?

reubenmiller: Though I'll still have to add a check to find out when the datetime filtering can be used (as that is also dependent on the implementation).

rina23q: I'm fine to keep only `all`.
The 15 facilities are based on RFC5424 (syslog protocol). I assume and it seems `dmesg` doesn't have to align with the facilities defined on RFC5424?

reubenmiller: Though the problem is that I don't think that all dmesg implementations support it anyway...so it would just be adding an extra 15 items to the types, where we don't know which ones would work or not.

So if we keep things simple for now, and just have one type, as it is anyway for more advanced usecases (as this is more for powerusers who want to monitor kernel messages to debug driver problems etc.)

reubenmiller: I've updated the PR to used the single "all" type. The PR's description also reflects the updated scope.

rina23q (approved):
LGTM

configuration/contrib/log-plugins/dmesg
```diff
@@ -0,0 +1,137 @@
...
+    if [ $# -lt 1 ]; then
+        echo "Error: 'get' command expects a mandatory <log-type> arguments" >&2
+        usage
+        exit 1
+    fi
+    _log_type="$1"
+    shift
+
+    since="1 day ago"
+    until="0 second ago"
```
didier-wenzek: Something is wrong with these defaults. I got no entries when no values is given for `--since` and `--until`.

reubenmiller: Though that might be because there are no values within the time range (which isn't uncommon)

reubenmiller: But you can confirm this by just increasing the look back.

```sh
/var/share/tedge/log-plugins/dmesg get all --since "180 days ago" --until "0 second ago"
```

reubenmiller: We could change the default to something higher, like 90 days?

configuration/contrib/log-plugins/dmesg
albinsuresh: Yeah, `all` is fine. If someone wants fine-grained filtering, they can easily add a thin wrapper plugin overriding this one with the exact types available on their device.

configuration/contrib/log-plugins/dmesg
didier-wenzek: So, it was my bad. I have entries. I have been confused by the GUI with unexpected values for the time window.

> We could change the default to something higher, like 90 days?

Why not.

didier-wenzek (approved):
Approved. The default time range could be increased as suggested.

configuration/contrib/log-plugins/dmesg
reubenmiller: The default has been changed to the last 90 days in 4ae70f87448a760841045da0eab2fd5d8268bfe6

PR #3831, opened 2025-10-24
fix: missing reload events

didier-wenzek:
## Proposed changes

Applying this suggestion https://github.com/thin-edge/thin-edge.io/pull/3805#pullrequestreview-3375663764
highlighted two bugs, previously masked because all the tests were pre-loaded and sharing timers.

- [x] Isolate flow system tests, installating a flow per test
- [x] Fix timer of periodic sources
- [x] Fix flow reloading on inotify events

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/runtime.rs
+++ b/crates/extensions/tedge_flows/src/runtime.rs
@@ -249,22 +249,15 @@ impl MessageProcessor {
 
     pub async fn add_flow(&mut self, path: Utf8PathBuf) {
         let flow_id = Self::flow_id(&path);
-        if !self.flows.contains_key(&flow_id) && self.load_flow(flow_id, path.clone()).await {
-            info!(target: "flows", "Loaded new flow {path}");
-        }
-    }
-
-    pub async fn reload_flow(&mut self, path: Utf8PathBuf) {
-        let flow_id = Self::flow_id(&path);
-        if self.flows.contains_key(&flow_id) && self.load_flow(flow_id, path.clone()).await {
-            info!(target: "flows", "Reloaded updated flow {path}");
+        if self.load_flow(flow_id, path.clone()).await {
+            info!(target: "flows", "Loading flow {path}");
         }
     }
 
     pub async fn remove_flow(&mut self, path: Utf8PathBuf) {
         let flow_id = Self::flow_id(&path);
         self.flows.remove(&flow_id);
-        info!(target: "flows", "Removed deleted flow {path}");
+        info!(target: "flows", "Removing flow {path}");
     }
 }
 
--- a/crates/extensions/tedge_flows/src/actor.rs
+++ b/crates/extensions/tedge_flows/src/actor.rs
@@ -73,7 +73,7 @@ impl Actor for FlowsMapper {
                     if matches!(path.extension(), Some("js" | "ts" | "mjs")) {
                         self.processor.reload_script(path).await;
                     } else if path.extension() == Some("toml") {
-                        self.processor.reload_flow(path).await;
+                        self.processor.add_flow(path).await;
                         self.send_updated_subscriptions().await?;
                     }
                 }


tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -116,31 +116,38 @@ Consuming messages from a process stdout, periodically
...
-    ...    tedge mqtt sub test/file/input --duration 1s | grep 'Error in /etc/tedge/flows/read-file-periodically.toml'
+    ...    tedge mqtt sub test/file/input --duration 1s | grep 'Fail to poll /tmp/file.input'
     Execute Command    echo 'hello world' >/tmp/file.input
     Execute Command    tedge mqtt sub test/file/input --duration 1s | grep 'hello world'
+    [Teardown]    Uninstall Flow    read-file-periodically.toml
 
 Appending messages to a file
     # Assuming the flow append-to-file.toml has been properly installed
+    Install Flow    append-to-file.toml
     Execute Command    for i in $(seq 3); do tedge mqtt pub seq/events "$i"; done
```
didier-wenzek, line 137: This might still be flacky, even if okay after 10 runs.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -116,31 +116,38 @@ Consuming messages from a process stdout, periodically
 
 Consuming messages from the tail of file
     # Assuming the flow tail-named-pipe.toml has been properly installed
+    Install Flow    tail-named-pipe.toml
```
albinsuresh, line 116: The comment in the previous line isn't valid anymore. Same applicable to the rest of the tests as well.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
albinsuresh: To avoid that flakiness, we need to wait till the file is modified before doing the `grep`, right?

Came across this `inotifywait` tool that makes it easier to do that with a simple command:
```
inotifywait --timeout 1 --event modify
```

But it's not available out-of-the-box. The `inotify-tools` package must be installed first.

crates/extensions/tedge_flows/src/runtime.rs
```diff
@@ -95,10 +95,18 @@ impl MessageProcessor {
...
+            .flows
             .values()
             .flat_map(|flow| &flow.steps)
-            .filter_map(|step| step.script.next_execution)
+            .filter_map(|step| step.script.next_execution);
+
+        let source_deadlines = self
+            .flows
+            .values()
+            .filter_map(|flow| flow.input.next_deadline());
```
albinsuresh, line 107: I'm a bit surprised that none of the periodic tests failed with that bug. I don't even fully get how it was getting masked by interference from other tests, as none of the tests with the periodic polling should have worked with the old code.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
didier-wenzek: We need to wait for the flow to be fully loaded before sending MQTT messages to the flow input topic.

For that we need to improve tedge flow so status or stats messages are published over MQTT when a flow is loaded.

crates/extensions/tedge_flows/src/runtime.rs
didier-wenzek: The bug was masked because for tick events to be sent it's enough to have one source or step requesting a tick. Once sent the tick event is passed to all sources and steps, triggering a call to `on_interval` if time has elapsed. In other words, the sources were ready to process tick events but was failing to register a proper pace for the ticks.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
didier-wenzek: Still valid, only redundant. Anyway, removed: https://github.com/thin-edge/thin-edge.io/pull/3831/commits/1ccbc2f57e8187c65caceab1c8527b9e2ee33ed7

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
albinsuresh: > We need to wait for the flow to be fully loaded before sending MQTT messages to the flow input topic.

Aah okay. I thought the flakiness was because we're doing the grep on the output file immediately after publishing the message, without giving enough time for the mapper to process that input and write to the file.

> For that we need to improve tedge flow so status or stats messages are published over MQTT when a flow is loaded.

Yeah, that'd fix both.

PR #3834, opened 2025-10-27
feat: sync signal support for agent

albinsuresh:
## Proposed changes

Agent to react to a `signal/sync` message by notifying all built-in actors to sync their states. Currently only the log manager actor reacts to this signal by resending the supported log types. Operation specific sync messages like `sync_log_upload` can be used to trigger a sync of that specific actor.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/operation_workflows/actor.rs
+++ b/crates/core/tedge_agent/src/operation_workflows/actor.rs
@@ -154,7 +154,7 @@ impl WorkflowActor {
             }
             _ => {
                 error!("Unsupported channel: {}", channel);
-                return Ok(());
+                Ok(())
             }
         }
     }
@@ -168,13 +168,19 @@ impl WorkflowActor {
             SignalType::Sync => {
                 info!("Received sync signal, requesting all builtin actors to sync");
                 self.sync_signal_dispatcher.sync_all().await?;
-                Ok(())
+            }
+            SignalType::SyncOperation(operation) => {
+                info!(
+                    "Received sync signal for {}, requesting the corresponding actor to sync",
+                    operation
+                );
+                self.sync_signal_dispatcher.sync(operation).await?;
             }
             SignalType::Custom(_) => {
                 // Custom signal types are not handled yet
-                Ok(())
             }
         }
+        Ok(())
     }
 
     async fn process_command_message(
@@ -469,7 +475,7 @@ impl WorkflowActor {
         new_state: GenericCommandState,
     ) -> Result<(), RuntimeError> {
         if new_state.is_finished() {
-            self.sync_dependent_actors(&new_state).await?;
+            self.sync_listener_actors(&new_state).await?;
             self.finalize_builtin_command_update(new_state).await?;
 
             Ok(())
@@ -502,13 +508,13 @@ impl WorkflowActor {
         self.process_command_update(adapted_state).await
     }
 
-    async fn sync_dependent_actors(
+    async fn sync_listener_actors(
         &mut self,
         command: &GenericCommandState,
     ) -> Result<(), RuntimeError> {
         if let Some(command) = command.operation() {
             self.sync_signal_dispatcher
-                .send(command.as_str().into())
+                .sync_listener(command.as_str().into())
                 .await?;
         }
         Ok(())
--- a/crates/core/tedge_agent/src/operation_workflows/builder.rs
+++ b/crates/core/tedge_agent/src/operation_workflows/builder.rs
@@ -28,7 +28,7 @@ use tedge_api::mqtt_topics::ChannelFilter;
 use tedge_api::mqtt_topics::EntityFilter;
 use tedge_api::mqtt_topics::EntityTopicId;
 use tedge_api::mqtt_topics::MqttSchema;
-use tedge_api::mqtt_topics::SignalType;
+use tedge_api::mqtt_topics::OperationType;
 use tedge_api::workflow::GenericCommandData;
 use tedge_api::workflow::GenericCommandState;
 use tedge_api::workflow::OperationName;
@@ -74,7 +74,11 @@ impl WorkflowActorBuilder {
 
         let mqtt_publisher = mqtt_actor.get_sender();
         mqtt_actor.connect_sink(
-            Self::subscriptions(&config.mqtt_schema, &config.device_topic_id),
+            Self::subscriptions(
+                &config.mqtt_schema,
+                &config.device_topic_id,
+                &config.service_topic_id,
+            ),
             &input_sender,
         );
         let mqtt_publisher = LoggingSender::new("MqttPublisher".into(), mqtt_publisher);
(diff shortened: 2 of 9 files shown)


crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -798,6 +798,7 @@ impl OperationType {
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum SignalType {
     Sync,
+    SyncLogUpload,
```
albinsuresh: I wasn't fully convinced with that name (mapped to the topic: `signal/sync_log_upload`). A few other alternatives were:
1. `SyncLogTypes` (`signal/sync_log_types`)
2. `SyncLogMetadata` (`signal/sync_log_metadata`)
3. `LogUpload` (`signal/log_upload`), keeping it the same as the operation type itself.

crates/core/tedge_agent/src/operation_workflows/builder.rs
```diff
@@ -110,19 +112,33 @@ impl WorkflowActorBuilder {
...
+        mqtt_schema.topics(
+            EntityFilter::Entity(device_topic_id),
+            ChannelFilter::AnyCommand,
+        ) + mqtt_schema.topics(
+            EntityFilter::Entity(device_topic_id),
+            ChannelFilter::Signal(SignalType::Sync),
+        ) + mqtt_schema.topics(
+            EntityFilter::Entity(device_topic_id),
+            ChannelFilter::Signal(SignalType::SyncLogUpload),
+        )
```
albinsuresh: The second commit with the `Add` trait impl for `TopicFilter` just came out of my desire to easily combine topic filters like this.

crates/common/mqtt_channel/src/topics.rs
```diff
@@ -93,7 +94,7 @@ impl TopicFilter {
     }
 
     /// Check if the pattern is valid and add it to this topic filter.
-    pub fn add(&mut self, pattern: &str) -> Result<(), MqttError> {
+    pub fn try_add(&mut self, pattern: &str) -> Result<(), MqttError> {
```
albinsuresh: An unfortunate side effect of the `Add` impl for `TopicFilter` causing a name collision. But, now I feel this is the right name for this function.

didier-wenzek: Okay. However, I see using `+` to combine topic filters as disputable.

crates/core/tedge_agent/src/operation_workflows/message_box.rs
```diff
@@ -49,10 +49,20 @@ impl CommandDispatcher {
 
 #[derive(Default)]
 pub(crate) struct SyncSignalDispatcher {
+    self_senders: HashMap<OperationType, DynSender<CmdMetaSyncSignal>>,
     senders: HashMap<OperationType, Vec<DynSender<CmdMetaSyncSignal>>>,
 }
 
 impl SyncSignalDispatcher {
+    /// Register sender for the operation type to receive sync signals to itself
```
didier-wenzek: I don't understand this comment nor the purpose of the `self_senders` map.

Why do we need `senders: HashMap<OperationType, Vec<DynSender<CmdMetaSyncSignal>>>` *and*
`self_senders: HashMap<OperationType, DynSender<CmdMetaSyncSignal>>`. The latter seems to be implied by the former.

crates/core/tedge_api/src/mqtt_topics.rs
didier-wenzek: Do we need a new `SyncLogUpload` case? I don't think so.

Introducing enum types with a case per operation as we did for `OperationType` adds little value (these are just glorified strings, as no specific data is attached to each case) and leads to repeated pointless code.

I would:
- either use a constant `SignalType::Custom("log_upload")` for the same purpose (simple).
- or leverage the `OperationType` enum (effective type support).


```suggestion
    SyncOperation(OperationType),
```

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -139,11 +140,56 @@ impl WorkflowActor {
...
+                self.sync_signal_dispatcher.sync_all().await?;
+            }
+            SignalType::SyncLogUpload => {
+                info!(
+                    "Received log_upload sync signal, requesting the corresponding actor to sync"
+                );
+                self.sync_signal_dispatcher
+                    .sync(OperationType::LogUpload)
+                    .await?;
+            }
```
didier-wenzek: This highlights why I think enum types with a case per operation add little value.

- Duplicated pointless code: this will have to be duplicated for each operation.
- Error prone, with no support from the type system: there is no relationship between `SignalType::SyncLogUpload` and `OperationType::LogUpload`.

The second point suggests the following:


```suggestion
            SignalType::SyncOperation(operation_type)  => {
                info!(
                    "Received {operation_type} sync signal, requesting the corresponding actor to sync"
                );
                self.sync_signal_dispatcher
                    .sync(operation_type)
                    .await?;
            }
```

crates/core/tedge_api/src/mqtt_topics.rs
albinsuresh: I like that suggestion a lot.

crates/core/tedge_agent/src/operation_workflows/message_box.rs
albinsuresh: I feel it's a naming issue.

The separation is needed because the purpose of both maps are different. The `senders` map keep track of the operation types and their corresponding "listeners". When the log manager subscribes to software update operation updates, it gets tracked in that `senders` map. So, based on our existing dependency requirements, the `senders` map would be something like this:

* `software_update` -> [`log_manager`, `config_manager`]
* `config_update` -> [`log_manager`]
* `log_upload` -> []

With this setup, whenever a `software_update` command finishes, the `log_manager` and `config_manager` are synced. Now to send `sync_log_upload` signal to the `log_manager`, I can't add the `log_upload` -> [`log_manager`] mapping to the exact same map, because that'll lead to the `log_manager` getting synced after the completion of every `log_upload` operation.

That's why I kept a different `self_senders` map of the operation types and their own senders so that when a sync signal is received for that operation, it is sent to their corresponding senders.

To avoid this confusion, I feel I should rename that `senders` map to something else like, `listeners` or `dependents` or something like that. And may be add this detailed commentary as well.

crates/core/tedge_api/src/mqtt_topics.rs
albinsuresh: Resolved in [10e49db](https://github.com/thin-edge/thin-edge.io/pull/3834/commits/10e49dbc2b5d9a4c31a7b5b8362687d48362504b) as suggested

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: Resolved in [10e49db](https://github.com/thin-edge/thin-edge.io/pull/3834/commits/10e49dbc2b5d9a4c31a7b5b8362687d48362504b) as suggested

crates/core/tedge_agent/src/operation_workflows/message_box.rs
albinsuresh: Renamed the fields and updated the docs in 6dad35f

crates/core/tedge_agent/src/operation_workflows/builder.rs
albinsuresh: Moved that unrelated commit into a separate PR: https://github.com/thin-edge/thin-edge.io/pull/3836

crates/common/mqtt_channel/src/topics.rs
albinsuresh: Moved that unrelated commit into a separate PR: https://github.com/thin-edge/thin-edge.io/pull/3836

crates/core/tedge_agent/src/operation_workflows/message_box.rs
didier-wenzek: The comment is definitely clearer and the naming less confusing.

jarhodes314 (approved):

PR #3836, opened 2025-10-28
refactor: add topic filters to combine them

albinsuresh:
## Proposed changes

Add topic filters with the `+` operator to combine them.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/topics.rs
+++ b/crates/common/mqtt_channel/src/topics.rs
@@ -8,6 +8,8 @@ use std::collections::HashSet;
 use std::convert::TryInto;
 use std::fmt::Display;
 use std::fmt::Formatter;
+use std::ops::Add;
+use std::ops::AddAssign;
 
 /// An MQTT topic
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
@@ -93,14 +95,10 @@ impl TopicFilter {
     }
 
     /// Check if the pattern is valid and add it to this topic filter.
-    pub fn add(&mut self, pattern: &str) -> Result<(), MqttError> {
-        let pattern = String::from(pattern);
-        if rumqttc::valid_filter(&pattern) {
-            self.patterns.push(pattern);
-            Ok(())
-        } else {
-            Err(MqttError::InvalidFilter { pattern })
-        }
+    pub fn try_add(&mut self, pattern: &str) -> Result<(), MqttError> {
+        let other = TopicFilter::new(pattern)?;
+        *self += other;
+        Ok(())
     }
 
     /// Assuming the pattern is valid and add it to this topic filter.
@@ -240,7 +238,7 @@ impl TryInto<TopicFilter> for Vec<&str> {
     fn try_into(self) -> Result<TopicFilter, Self::Error> {
         let mut filter = TopicFilter::empty();
         for pattern in self.into_iter() {
-            filter.add(pattern)?
+            filter.try_add(pattern)?
         }
         Ok(filter)
     }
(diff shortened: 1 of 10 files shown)


crates/common/mqtt_channel/src/topics.rs
```diff
@@ -93,7 +94,7 @@ impl TopicFilter {
     }
 
     /// Check if the pattern is valid and add it to this topic filter.
-    pub fn add(&mut self, pattern: &str) -> Result<(), MqttError> {
+    pub fn try_add(&mut self, pattern: &str) -> Result<(), MqttError> {
```
rina23q, line 98: I see mostly this `try_add()` is used in our code and the new`std::ops::Add` is used only in one place.
I did a quick experiment to improve `try_add()`.

try_add():
```rust
pub fn try_add(&mut self, pattern: &str) -> Result<(), MqttError> {
    let other = TopicFilter::new(pattern)?;
    *self += other;
    Ok(())
}
```

and implement also `std::ops::AddAssign`
```rust
impl Add for TopicFilter {
    type Output = TopicFilter;

    fn add(mut self, other: TopicFilter) -> TopicFilter {
        self += other;
        self
    }
}

impl AddAssign for TopicFilter {
    fn add_assign(&mut self, rhs: Self) {
         self.add_all(rhs);
    }
}
```

How do you think about this? Too much?

jarhodes314: This sounds like a sensible suggestion to me, the proposed `try_add` implementation is much more elegant. And since we are effectively performing `AddAssign` within `Add`, it seems very logical to also support `AddAssign`.

As a slight side note, I found it slightly surprising that the `+` operator mutates the existing `TopicFilter` (made particularly apparent here where you implement `+` in terms of `+=`), rather than creating a new one (despite the fact it is clearly more reasonable here to mutate). Of course this is a great point of ownership in Rust; this `Add` implementation is free to mutate or not and the consumer of this method is unable to make a mistake caused by surprise mutation.

albinsuresh: Yeah, that `AddAssign` totally makes sense. In fact, I had initially implemented `AddAssign` as well, but then took it out as I felt that calling `topics.add_all(other_topics)` reads slightly better than `topics += other_topics` because of its explicitness. But that's purely my opinion and I guess there's no harm in having that as well for those who prefers to use it.

didier-wenzek: > I felt that calling `topics.add_all(other_topics)` reads slightly better than `topics += other_topics` 

I'm really surprised that you feel that while proposing with this PR to write  `topics + other_topics`. What's the difference?

albinsuresh: > I'm really surprised that you feel that while proposing with this PR to write `topics + other_topics`. What's the difference?

I believe my point-of-view slightly changed in this case after seeing a lot of the usages of `add_all` which were done in loops. So the pattern was somewhat like this:

```
let mut filters = TopicFilter::empty()
for topic_str in some_topics {
    filters.add(TopicFilter::new(topic_str)?)
}
```

That minor annoying thing that prompted me to impl the `Add` in the first place was the need to create that mutable initial filter, when you just wanted to combine a couple of static topic filters. But, in the looped cases, it was a necessity anyway. So, I felt that changing that (**in a lot of places**) would raise more eyebrows for no significant value add, when my priority was only improving the static cases as follows 😄 

**From**
```
        let mut topics = config.logfile_request_topic.clone();
        topics.add_all(config.log_metadata_sync_topics.clone());
        topics
```

**To**
```
config.logfile_request_topic.clone() + config.log_metadata_sync_topics.clone()
```

albinsuresh: Resolved by 510b5665bd as suggested by @rina23q. I didn't go overboard and replace all existing usages of `add_all` with `+=`, but only those that could be made concise.

PR #3838, opened 2025-10-29
feat: filter log types using include/exclude patterns

albinsuresh:
Allow users to filter log types reported by log plugins using include and exclude patterns in the plugin configuration.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


docs/src/extend/log-management.md
```diff
@@ -69,6 +69,52 @@ then sudoers entries must be created for those directories as well.
...
+## Filtering Plugin Log Types
+
+When a plugin is listing too many log types that that user is not interested in,
+the irrelevant entries can be filtered out by specifying `include` and `exclude` filter patterns
+in the `tedge-log-plugin.toml` configuration file.
+
+### Configuration Format
+
+Use the `[[plugins.<plugin-name>]]` sections to define filters.
+Both `include` and `exclude` support [glob patterns](https://en.wikipedia.org/wiki/Glob_(programming)).
```
reubenmiller: It might be worth using regex instead of glob patterns to be consistent with other parts of tedge (e.g. the sm plugin filtering)

docs/src/extend/log-management.md
```diff
@@ -69,6 +69,80 @@ then sudoers entries must be created for those directories as well.
...
+Multiple `[[plugins.<plugin-name>]]` sections can be defined for the same plugin
+so that it is easier to extend by just appending new entries.
+
+```toml title="file: /etc/tedge/plugins/tedge-log-plugin.toml"
+[[plugins.journald]]
+include = "tedge-*"
+
+[[plugins.journald]]
+exclude = "systemd*"
```
didier-wenzek: According to the filtering rules, this `include` clause is useless. I would be good to come with an example that better demonstrates the usefulness of `include` / `exclude` clauses.

The point here is only to show something sensible, the details being explained in the following sections.

It can be  "exclude all `systemd` services except `systemd-logind`".

Or better, something simpler with two plugins:


```suggestion
[[plugins.docker]]
include = "tedge-*"

[[plugins.journald]]
exclude = "systemd*"
```

crates/extensions/tedge_log_manager/src/config.rs
```diff
@@ -87,3 +89,317 @@ impl LogManagerConfig {
...
+        };
+
+        let mut exclude_patterns = Vec::new();
+        let mut include_patterns = Vec::new();
+
+        for filter in filters {
+            if let Some(pattern) = &filter.include {
+                if let Ok(compiled) = glob::Pattern::new(pattern) {
+                    include_patterns.push(compiled);
+                } else {
```
didier-wenzek: I would extract this compilation stage out of the `filter_log_types` method. Can this be done when the file is read?

crates/extensions/tedge_log_manager/src/config.rs
```diff
@@ -87,3 +89,317 @@ impl LogManagerConfig {
...
+}
+
+/// Individual filter entry for a plugin
+#[derive(Clone, Deserialize, Debug, PartialEq, Eq)]
+pub struct PluginFilterEntry {
+    #[serde(default)]
+    pub include: Option<String>,
+    #[serde(default)]
+    pub exclude: Option<String>,
+}
```
didier-wenzek: Looking how this is used, an `enum` seems more appropriate.

```suggestion
pub enum PluginFilterEntry {
     Include(String),
     Exclude(String),
}
```

I would even go a step further with:


```suggestion
pub enum PluginFilterEntry {
     Include(glob::Pattern),
     Exclude(glob::Pattern),
}
```

crates/extensions/tedge_log_manager/src/config.rs
```diff
@@ -87,3 +89,317 @@ impl LogManagerConfig {
...
+                    tracing::warn!(
+                        "Failed to parse plugin config from {}: {}",
+                        path.display(),
+                        err
+                    );
+                    Self::default()
+                }
+            },
+            Err(_) => {
+                // File doesn't exist or not readable - this is OK, just use defaults
```
didier-wenzek: Okay, but it would be good to warn the user.

docs/src/extend/log-management.md
albinsuresh: > Or better, something simpler with two plugins:

Yup, I'll do that since this is just the introduction section

> It can be "exclude all systemd services except systemd-logind".

Have included that example in the examples section below.

crates/extensions/tedge_log_manager/src/config.rs
albinsuresh: > Can this be done when the file is read?

Yes. We can keep these patterns compiled when the config file is read/reloaded.

docs/src/extend/log-management.md
```diff
@@ -69,6 +69,80 @@ then sudoers entries must be created for those directories as well.
...
+```
+
+#### Exclude all systemd services except systemd-logind
+
+```toml
+[[plugins.journald]]
+include = "systemd-logind"
+
+[[plugins.journald]]
+exclude = "systemd-*"
```
reubenmiller: I'm not sure about the chosen data format here, as it involves arrays of objects, which makes it a bit difficult to manipulate. Could we just have a flat object structure and change the "include" and "exclude" keys to be an array?

e.g.

```toml
[plugin.journald]
include = ["systemd-logind"]
exclude = ["systemd-*"]
```

crates/extensions/tedge_log_manager/src/config.rs
```diff
@@ -87,3 +89,317 @@ impl LogManagerConfig {
...
+    #[serde(default)]
+    pub include: Option<String>,
+    #[serde(default)]
+    pub exclude: Option<String>,
+}
+
+impl PluginConfig {
+    /// Load plugin configuration from the given path
+    pub fn from_file(path: &std::path::Path) -> Self {
+        match std::fs::read_to_string(path) {
```
jarhodes314:
```suggestion
    pub async fn from_file(path: &std::path::Path) -> Self {
        match tokio::fs::read_to_string(path).await {
```

crates/extensions/tedge_log_manager/src/config.rs
jarhodes314: This is a good idea. I would be careful with the wording if this is potentially not actually an issue, just highlighting that we didn't read the configuration from the specified file. If the default behaviour of thin-edge is that this file doesn't exist, this warning will appear often, and we should make it clear that this is not an issue. I may also be misassuming how things work in this case, and it may be the case that we can cope but this is a weird edge case that the user should fix, in which case a normal warning would be entirely justified.

crates/extensions/tedge_log_manager/src/config.rs
```diff
@@ -87,3 +89,317 @@ impl LogManagerConfig {
...
+        }
+
+        // include a type if it matches any include pattern OR doesn't match any exclude pattern
+        log_types
+            .into_iter()
+            .filter(|log_type| {
+                let matches_include = include_patterns.iter().any(|p| p.matches(log_type));
+                let matches_exclude = exclude_patterns.iter().any(|p| p.matches(log_type));
+
+                // When both include and exclude patterns exist, use OR logic
```
jarhodes314, line 251: I'm slightly confused by this logic. Essentially this means if I have an `include` filter and I add an `exclude` filter, that will potentially cause _more_ items to match than before we added the second filter rule. I think my natural expectation is for the two to be `AND`ed together.

docs/src/extend/log-management.md
```diff
@@ -69,6 +69,80 @@ then sudoers entries must be created for those directories as well.
...
 
+## Factory Plugins
+
+* The default `file` plugin is included in the `tedge` installation package itself on all distributions.
+* A `journald` plugin that can gather systemd service logs using the `journalctl` command is also included
+  in the `tedge` packages for systemd based distributions like Debian, Ubuntu, RHEL etc.
+
+## Filtering Plugin Log Types
+
+When a plugin is listing too many log types that that user is not interested in,
```
rina23q:
```suggestion
When a plugin is listing too many log types that user is not interested in,
```

docs/src/extend/log-management.md
rina23q: Note that we are using glob patterns for file paths in config/log management. However, I also agree with using regex patterns for filtering.

docs/src/extend/log-management.md
```diff
@@ -69,6 +69,80 @@ then sudoers entries must be created for those directories as well.
...
+```
+
+### Reloading After Configuration Changes
+
+Whenever the plugin filters in `tedge-log-plugin.toml` are modified,
+the agent automatically reloads the plugins and applies the new filters.
+You can verify the filtered log types by checking the published metadata on the MQTT topic:
+
+```sh
+tedge mqtt sub 'te/device/main///cmd/log_upload'
```
rina23q, line 144: I think you can add "For the main device" or some note, since this settings can be applied to child devices as well. For child devices, the topic ID will be different.

crates/extensions/tedge_log_manager/src/config.rs
```diff
@@ -87,3 +89,317 @@ impl LogManagerConfig {
...
+    }
+
+    /// Apply filtering to log types from a plugin
+    ///
+    /// Filtering logic:
+    /// - If no filters defined, return all types as-is
+    /// - When only include is specified: only types matching any include filter are returned
+    /// - When only exclude is specified: all types not matching any exclude filter are returned
+    /// - When both are specified: include the types that matches any include filter OR doesn't match any exclude filter
+    pub fn filter_log_types(&self, plugin_name: &str, log_types: Vec<String>) -> Vec<String> {
```
rina23q: Though it's unrelated to this change, I just realized all functions are taking/using `Vec` for log_types - that means we accept the duplicated log types from the list command.
Isn't it better to change it to `HashSet` to avoid the duplications? 

Actually the same thing can be applied to the filters. If two or more completely same filters are added, they will be used here multiple times.

docs/src/extend/log-management.md
albinsuresh: Yeah, I've had much deliberation on this format. And I agree that the very first format that came to my mind was exactly what you've proposed, which is fairly intuitive as well. But I went with the array of objects for easier manipulation and extensibility.

I agree that all `include` patterns in a single array would be easier for manual manipulation, but if a user wants to programatically add a new `include` pattern, it'd be a lot more difficult to append to an existing array, right?

With the proposed array of objects patterns, it can be a simple append of the following block, without having to know the existing values in the pattern:
```
[[plugin.journald]]
include = "new-pattern-*"
```

Programmatically removing an existing pattern looks equally difficult with both formats, but I'm expecting more dynamic appends than removals.

This format would also come in handy in future, if/when we add support for drop-in extension files for `tedge-log-plugin.toml`. In that case also, a new include/exclude pattern for a plugin can be defined in an extension file the exact same way, even if include/exclude patterns for the same plugin are defined in the main tedge-log-plugin.toml` or any other extension file. But if we go with the simple format that you've proposed, a new pattern cannot be added to an extension file if the include patterns for that same plugin is defined elsewhere.

These were the reasons why I went with the array of objects format, although I agree that it is not the most intuitive format for someone new to TOML.

crates/extensions/tedge_log_manager/src/config.rs
albinsuresh: TBH, I agree with you that the natural expectation would be an `AND`. But this OR combination was implemented because the sm-plugin filtering also does the same and we wanted to keep it consistent. Actually there are use-cases for both.

If it was `AND` I could define rules like:

```
[[plugin.journald]]
include = "tedge-*"
exclude = "tedge-watcher"
```

where we want to include everything matching `tedge-*` but selectively exclude `tedge-watcher` from it.

Now, the `OR`can also be used to define the reverse rule:

```
[[plugin.journald]]
exclude = "systemd-*"
include = "systemd-logind"
```

where we exclude everything `systemd-*` but would like to selectively override that blanket exclusion by including `systemd-logind`.

Both are useful in my opinion and for some reason we chose the latter for sm-plugins and we're still sticking to it.

crates/extensions/tedge_log_manager/src/config.rs
albinsuresh: The `Vec` is used for deterministic ordering as well, which helps in tests. Yes, there is the possibility of a badly written plugin lisiting the same type multiple times. So, an alternative to address both concerns would be a `BTreeSet`.

docs/src/extend/log-management.md
reubenmiller: I'm just wondering if the data model is essentially an array of objects, rather than an array of strings, that this makes it harder to manipulate the data in the future.

Though is there anything stopping people also providing both the include and exclude properties within one array item, e.g.

```toml
[[plugins.journald]]
include = "systemd-logind"
exclude = "systemd-*"
```

albinsuresh: > Though is there anything stopping people also providing both the include and exclude properties within one array item, e.g.
> 
> ```toml
> [[plugins.journald]]
> include = "systemd-logind"
> exclude = "systemd-*"
> ```

That's perfectly valid as the keys are different (`include` and `exclude`).

Or did you mean to ask if multiple `include` filters are allowed under the same object array as follows:

```
[[plugins.journald]]
include = "abc"
include = "xyz"
```

OR

```
[[plugins.journald]]
include = ["abc", "def"]
include = ["xyz"]
```


and expect all the `include` patterns to be combined into a single `include` array?

**That's an invalid toml** due to key duplication under the same table. The second `include` pattern must be placed in a separate `[[plugins.journald]]` block.

crates/extensions/tedge_log_manager/src/config.rs
albinsuresh: After [this](https://github.com/thin-edge/thin-edge.io/pull/3838#discussion_r2492813337) comment from  @reubenmiller I'm having second thoughts on converting that struct definition into an enum as it prevents users from defining both `include` and `exclude` filters under the same plugin entry as follows:

 ```
 [[plugins.journald]]
exclude = "systemd-*"
include = "systemd-logind"
```

The struct might be more future-proof as well if we want to support more fields beyond `include` and `exclude` in future.

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 8 above)
Comment (didier-wenzek) [reply crates/extensions/tedge_log_manager/src/config.rs:251]: What confuses me is the current implementation is the default values.

- When no `exclude`  rules are given it means exclude everything, i.e. take only what is explicitly included.
- When no `include` rules are given it means include everything, *unless* there is an exclude rule in which case it means include nothing.

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 4 above)
Comment (didier-wenzek) [reply crates/extensions/tedge_log_manager/src/config.rs]: My suggestion is not to prevent a combination of include and exclude rules for a given plugin, but to simplify the representation and the implementation logic. Note that currently a `Vec<PluginFilterEntry>>` is attached to each log type, making the enum enough and clearer. Two different structs might be required, though. A first one for the TOML file, aka `PluginConfig`, a second one to control the extent of log types to manage.

File: docs/src/extend/log-management.md (hunk 6 above)
Comment (reubenmiller) [reply docs/src/extend/log-management.md]: No I just meant the example that I [posted](https://github.com/thin-edge/thin-edge.io/pull/3838#discussion_r2492813337). I just wanted to make sure that it also worked, as adding two array items seemed strange when you want to just define an include and an exclude.

After the demo, I'm ok with the overall data structure (as long as the user can also provided include and exclude on the same array item).

File: docs/src/extend/log-management.md (hunk 2 above)
Comment (albinsuresh) [reply docs/src/extend/log-management.md]: Resolved by [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618)

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 3 above)
Comment (albinsuresh) [reply crates/extensions/tedge_log_manager/src/config.rs]: Resolved by [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618)

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 4 above)
Comment (albinsuresh) [reply crates/extensions/tedge_log_manager/src/config.rs]: Have done the compilation to regex pattern in [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618). The enum couldn't be done due to the possibility of both `include` and `exclude` filters under the `[[plugins.journald]]` entry, unless we use an enum like:
```
enum Filter {
   Include(Regex)
   Exclude(Regex)
   Both(Regex)
}
```
which would make little difference in the filtering logic.

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 5 above)
Comment (albinsuresh) [reply crates/extensions/tedge_log_manager/src/config.rs]: Resolved by [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618).

File: docs/src/extend/log-management.md (hunk 1 above)
Comment (albinsuresh) [reply docs/src/extend/log-management.md]: Switched to regex patterns in [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618).

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 7 above)
Comment (albinsuresh) [reply crates/extensions/tedge_log_manager/src/config.rs]: Resolved by [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618).

File: docs/src/extend/log-management.md (hunk 9 above)
Comment (albinsuresh) [reply docs/src/extend/log-management.md]: Resolved by [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618).

File: docs/src/extend/log-management.md (hunk 10 above)
Comment (albinsuresh) [reply docs/src/extend/log-management.md:144]: Resolved by [2d9f927](https://github.com/thin-edge/thin-edge.io/pull/3838/commits/2d9f9274e2ce50737d30f15de11a32fa0a05e618).

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 8 above)
Comment (albinsuresh) [reply crates/extensions/tedge_log_manager/src/config.rs:251]: > What confuses me is the current implementation is the default values.
> 
> * When no `exclude`  rules are given it means exclude everything, i.e. take only what is explicitly included.
> * When no `include` rules are given it means include everything, _unless_ there is an exclude rule in which case it means include nothing.

I totally agree that it is difficult to easily put the meaning of that combined case into words as can be done for the other cases as follows:

*  Neither `include` or `exclude`: Include everything
* `include` only: Include just what is specified
* `exclude` only: Include everything that is not excluded

But the combined case is really difficult to put in a sentence:
* `include` and `exclude`: Include everything that is not excluded + what  an be selectively included from the excluded list

If we had done the `AND` case, it was a lot easier: Include everything matching `include`, but further filter that with those specified in `exclude`.

For my brain, the AND is easier to comprehend than the OR.

But, I believe the reasoning for that choice was that users are more likely to provide `exclusion` filters for what they don't want, rather than providing `inclusion` filters for everything that they want. Once an `exclude` filter is provided, people would mostly want to selectively include a few out of those.

Comment (didier-wenzek) [reply crates/extensions/tedge_log_manager/src/config.rs:251]: > But, I believe the reasoning for that choice was that users are more likely to provide `exclusion` filters for what they don't want, rather than providing `inclusion` filters for everything that they want. Once an `exclude` filter is provided, people would mostly want to selectively include a few out of those.

If so, the doc must highlight it and sounds natural.

One way to do so might be to a layman AND ALSO as a substitute for the formal OR.
As done with a `+` in the following sentence:

> Include everything that is not excluded + what  an be selectively included from the excluded list

Are retained all the log types that are not excluded and also those explicitly included.

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 4 above)
Comment (didier-wenzek) [reply crates/extensions/tedge_log_manager/src/config.rs]: Okay let's keep the struct.

However, I disagree with the rationale. The in-memory representation of the filters doesn't have to match the external schema of the configuration file. We could have a vector of enum values with two cases or two vectors (one for the included regex the other for the excluded ones)

File: crates/extensions/tedge_log_manager/src/config.rs (hunk 11 above)
Comment (albinsuresh) [reply crates/extensions/tedge_log_manager/src/config.rs]: Resolved by 34480cf

File: tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot [hunk 12]
```diff
@@ -6,7 +6,7 @@ Library             String
 Library             Cumulocity
 Library             ThinEdgeIO
 
-Suite Setup         Custom Setup
+Test Setup          Custom Setup
```
rina23q, line 9: Since you changed Suite Setup to Test Setup, I think `Set Suite Variable` should also be changed to `Set Test Variable` in the `Custom Setup`.

```
Custom Setup
    ${DEVICE_SN}=    Setup
    Set Suite Variable    $DEVICE_SN # this line
```

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -220,6 +220,85 @@ Agent resilient to plugin dirs removal
     ThinEdgeIO.Service Health Status Should Be Up    tedge-agent
     Should Have MQTT Messages    c8y/s/us    date_from=${date_from}    message_pattern=118,
 
+Plugin include filter
+    # Clean up any existing plugins for test isolation
+    Execute Command    rm /usr/share/tedge/log-plugins/file
+    Execute Command    rm /usr/share/tedge/log-plugins/journald
+
```
rina23q, line 248: Why `dmsg` is not removed in this test?
```
Execute Command    rm /usr/share/tedge/log-plugins/dmesg
```

and if it's forgotten, probably it's better to create a key, like `Cleanup existing plugins` since these removals repeat in several tests.

crates/extensions/tedge_log_manager/src/config.rs
albinsuresh: > We could have a vector of enum values with two cases or two vectors (one for the included regex the other for the excluded ones)

Aah yes. For the internal representation, we can absolutely do that. I was only thinking of the external struct used by serde.

albinsuresh: Resolved that as well in 9d4bbb63d1 (squashed commit) with a `TomlPluginConfig` struct used for parsing and a compiled `PluginConfig` struct that maintains `include_patterns` and `exclude_patterns` separately.

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
albinsuresh: Resolved in the squashed commit bd7c2a95c0

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
albinsuresh: Resolved in the squashed commit bd7c2a95c0

PR #3842, opened 2025-10-30
refactor: clearly separate message processing from input gathering

didier-wenzek:
## Proposed changes

This PR addresses refactoring requests on the design of [tedge flows input](https://github.com/thin-edge/thin-edge.io/pull/3805):

- [x] Separate flow input definitions from flow input connectors which stream/poll the sources 
- [x] Add `trait FlowRegistry` to handle different kinds of flows,
- [x] Add `struct ConnectedFlow` to wrap a `Flow` with the mechanisms to stream/poll its source
- [x] [Remove `trait FlowInput: StreamingSource + PollingSource`](https://github.com/thin-edge/thin-edge.io/pull/3805#discussion_r2452349719) which combines unrelated abstraction
- [x] [Remove MQTT concerns from StreamingSource](https://github.com/thin-edge/thin-edge.io/pull/3805#discussion_r2452355409)
- [x] [Remove meaningless Streaming and Polling Source implementation for MQTT](https://github.com/thin-edge/thin-edge.io/pull/3805#discussion_r2452384949)
- [x] [Remove complicated use of cascaded Option](https://github.com/thin-edge/thin-edge.io/pull/3805#discussion_r2452445452)

## Types of changes

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/connected_flow.rs
+++ b/crates/extensions/tedge_flows/src/connected_flow.rs
@@ -5,10 +5,12 @@ use crate::flow::FlowError;
 use crate::flow::FlowInput;
 use crate::flow::FlowResult;
 use crate::flow::Message;
-use crate::input_source::CommandFlowInput;
-use crate::input_source::FileFlowInput;
-use crate::input_source::FlowSource;
-use crate::input_source::MqttFlowInput;
+use crate::input_source::CommandPollingSource;
+use crate::input_source::CommandStreamingSource;
+use crate::input_source::FilePollingSource;
+use crate::input_source::FileStreamingSource;
+use crate::input_source::PollingSource;
+use crate::input_source::StreamingSource;
 use crate::registry::FlowRegistry;
 use crate::registry::FlowStore;
 use camino::Utf8Path;
@@ -18,7 +20,8 @@ use tokio::time::Instant;
 /// A flow connected to a source of messages
 pub struct ConnectedFlow {
     flow: Flow,
-    pub(crate) input: Box<dyn FlowSource>,
+    streaming_source: Option<Box<dyn StreamingSource>>,
+    polling_source: Option<Box<dyn PollingSource>>,
 }
 
 impl AsRef<Flow> for ConnectedFlow {
@@ -35,9 +38,13 @@ impl AsMut<Flow> for ConnectedFlow {
 
 impl ConnectedFlow {
     pub fn new(flow: Flow) -> Self {
-        let name = flow.name().to_string();
-        let input = connect(name, flow.input.clone());
-        ConnectedFlow { flow, input }
+        let streaming_source = streaming_source(flow.name().to_owned(), flow.input.clone());
+        let polling_source = polling_source(flow.input.clone());
+        ConnectedFlow {
+            flow,
+            streaming_source,
+            polling_source,
+        }
     }
 
     pub fn name(&self) -> &str {
@@ -45,11 +52,17 @@ impl ConnectedFlow {
     }
 
     pub fn input_topic(&self) -> &str {
-        self.input.enforced_topic().unwrap_or_default()
+        self.flow.input.enforced_topic().unwrap_or_default()
     }
 
     pub fn watch_request(&self) -> Option<WatchRequest> {
-        self.input.watch_request()
+        self.streaming_source
+            .as_ref()
+            .and_then(|source| source.watch_request())
+    }
+
+    pub fn next_deadline(&self) -> Option<Instant> {
+        self.polling_source.as_ref().and_then(|p| p.next_deadline())
     }
 
     pub async fn on_source_poll(&mut self, timestamp: DateTime, now: Instant) -> FlowResult {
@@ -62,7 +75,9 @@ impl ConnectedFlow {
         timestamp: DateTime,
         now: Instant,
     ) -> Result<Vec<Message>, FlowError> {
-        let source = &mut self.input;
+        let Some(source) = &mut self.polling_source.as_mut() else {
+            return Ok(vec![]);
+        };
         if !source.is_ready(now) {
             return Ok(vec![]);
         };
@@ -77,30 +92,37 @@ impl ConnectedFlow {
     }
 }
 
-fn connect(flow_name: String, input: FlowInput) -> Box<dyn FlowSource> {
+fn streaming_source(flow_name: String, input: FlowInput) -> Option<Box<dyn StreamingSource>> {
     match input {
-        FlowInput::Mqtt { topics } => Box::new(MqttFlowInput { topics }),
+        FlowInput::StreamFile { topic: _, path } => {
+            Some(Box::new(FileStreamingSource::new(flow_name, path)))
+        }
+
+        FlowInput::StreamCommand { topic: _, command } => {
+            Some(Box::new(CommandStreamingSource::new(flow_name, command)))
+        }
+
+        _ => None,
+    }
+}
+
+fn polling_source(input: FlowInput) -> Option<Box<dyn PollingSource>> {
+    match input {
         FlowInput::PollFile {
             topic,
             path,
             interval,
-        } => Box::new(FileFlowInput::new(flow_name, topic, path, Some(interval))),
+        } => Some(Box::new(FilePollingSource::new(topic, path, interval))),
+
         FlowInput::PollCommand {
             topic,
             command,
             interval,
-        } => Box::new(CommandFlowInput::new(
-            flow_name,
-            topic,
-            command,
-            Some(interval),
-        )),
-        FlowInput::StreamFile { topic, path } => {
-            Box::new(FileFlowInput::new(flow_name, topic, path, None))
-        }
-        FlowInput::StreamCommand { topic, command } => {
-            Box::new(CommandFlowInput::new(flow_name, topic, command, None))
-        }
+        } => Some(Box::new(CommandPollingSource::new(
+            topic, command, interval,
+        ))),
+
+        _ => None,
     }
 }
 
@@ -139,10 +161,7 @@ impl FlowRegistry for ConnectedFlowRegistry {
             .flat_map(|flow| &flow.as_ref().steps)
             .filter_map(|step| step.script.next_execution);
 
-        let source_deadlines = self
-            .flows
-            .flows()
-            .filter_map(|flow| flow.input.next_deadline());
+        let source_deadlines = self.flows.flows().filter_map(|flow| flow.next_deadline());
 
         script_deadlines.chain(source_deadlines)
     }
--- a/crates/extensions/tedge_flows/src/flow.rs
+++ b/crates/extensions/tedge_flows/src/flow.rs
@@ -7,6 +7,8 @@ use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use serde_json::json;
 use serde_json::Value;
+use std::fmt::Display;
+use std::fmt::Formatter;
 use std::time::Duration;
 use tedge_mqtt_ext::MqttMessage;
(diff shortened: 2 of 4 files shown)


crates/extensions/tedge_flows/src/config.rs
```diff
@@ -101,6 +103,53 @@ pub enum ConfigError {
...
+        match FlowConfig::load_flow(flow).await {
+            Ok(flow) => Some(flow),
+            Err(err) => {
+                error!(target: "flows", "Failed to load flow {flow}: {err}");
+                None
+            }
+        }
+    }
+
+    pub async fn load_single_script(script: &Utf8Path) -> FlowConfig {
```
jarhodes314: Since this doesn't touch the filesystem, it doesn't need to be `async`:

```suggestion
    pub fn load_single_script(script: &Utf8Path) -> FlowConfig {
```

With that in mind, I find the nomenclature here slightly confusing. The other methods `load` flows and this means reading something from disk, and this method also `load`s a script, but it doesn't touch the filesystem, nor can it fail at this stage. I don't have a suggestion spring to mind for what I would rename this to, but hopefully it's understandable why this feels slightly confusing to me.

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -171,27 +219,42 @@ impl StepConfig {
...
                 let topic = topic.unwrap_or_else(|| command.clone());
-                Box::new(CommandFlowInput::new(flow_name, topic, command, interval))
+                match interval {
+                    Some(interval) if !interval.is_zero() => FlowInput::PollCommand {
+                        topic,
+                        command,
+                        interval,
+                    },
+                    _ => FlowInput::StreamCommand { topic, command },
+                }
```
jarhodes314, line 257: I really like this new approach, particularly with how elegantly it conveys the possible polling vs streaming approaches, and how these are configured by the user.

crates/extensions/tedge_flows/src/connected_flow.rs
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+fn streaming_source(flow_name: String, input: FlowInput) -> Option<Box<dyn StreamingSource>> {
+    match input {
+        FlowInput::PollFile {
+            topic: _,
+            path,
+            interval,
+        } if interval.is_zero() => Some(Box::new(FileStreamingSource::new(flow_name, path))),
```
jarhodes314: I'm slightly suprised to see this logic here rather than when we generate the `FlowInput`, particularly as it's reliant on `streaming_source` and `polling_source` being in agreement about what we need to do to handle this case.

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: > but hopefully it's understandable why this feels slightly confusing to me.

It is understandable! I got confused myself more than once. I will come with a better name.

crates/extensions/tedge_flows/src/connected_flow.rs
didier-wenzek: > I'm slightly suprised to see this logic here rather than when we generate the `FlowInput`

Good point. I will move the logic there.

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: With [this suggestion](https://github.com/thin-edge/thin-edge.io/pull/3842#discussion_r2478370681) this will even be clearer.

crates/extensions/tedge_flows/src/connected_flow.rs
didier-wenzek: Actually, already done there but duplicated here. Fixed https://github.com/thin-edge/thin-edge.io/pull/3842/commits/8b5b0e23e17a6ac13c368b2976ef81d95f1dc3ea

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: My proposal is `wrap_script_into_flow`: https://github.com/thin-edge/thin-edge.io/pull/3842/commits/06c64961675d6c5b7a344b7804012639a366bdd4

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -171,27 +219,42 @@ impl StepConfig {
...
+            },
             InputConfig::File {
                 topic,
                 path,
                 interval,
             } => {
                 let topic = topic.unwrap_or_else(|| path.clone().to_string());
-                Box::new(FileFlowInput::new(flow_name, topic, path, interval))
+                match interval {
+                    Some(interval) if !interval.is_zero() => FlowInput::PollFile {
```
albinsuresh, line 236: This  fallback to `FlowInput::StreamFile`  when `interval` is misconfigured with `0` value is intentional? Wouldn't it be better to flag it as a misconfiguration? Same comment for the `FlowInput::StreamCommand` as well.

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -0,0 +1,237 @@
...
+
+#[async_trait]
+pub trait FlowRegistryExt: FlowRegistry {
+    fn config_dir(&self) -> Utf8PathBuf;
+
+    fn flow(&self, name: &str) -> Option<&Self::Flow>;
+    fn flow_mut(&mut self, name: &str) -> Option<&mut Self::Flow>;
+
+    fn flows(&self) -> impl Iterator<Item = &Self::Flow>;
+    fn flows_mut(&mut self) -> impl Iterator<Item = &mut Self::Flow>;
```
albinsuresh, line 69: Optional suggestion: These functions could be moved into `FlowRegistry` itself as default implementations, right? As they only rely on the `store()` and `store_mut()` functions? That'll make the separation between `FlowRegistry` and `FlowRegistryExt` clearer as well, where the former deals with all the core access primitives and the latter deals with the dependency on `JsRuntime`.

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    async fn remove_script(&mut self, path: Utf8PathBuf) {
+        for flow in self.store().flows() {
+            let flow_id = flow.as_ref().name();
+            for step in flow.as_ref().steps.iter() {
+                if step.script.path() == path {
+                    warn!(target: "flows", "Removing a script used by a flow {flow_id}: {path}");
+                    return;
```
albinsuresh, line 173: Nothing is being removed here.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -117,6 +143,18 @@ pub enum FlowError {
...
+        self
+    }
+}
+
+impl AsMut<Flow> for Flow {
+    fn as_mut(&mut self) -> &mut Flow {
+        self
+    }
+}
+
```
albinsuresh, line 157: These impls are implicitly added by the compiler for all types, right? Or am I missing something special with the `Flow` type?

crates/extensions/tedge_flows/src/connected_flow.rs
```diff
@@ -0,0 +1,168 @@
...
+    ) -> Result<Vec<Message>, FlowError> {
+        let Some(source) = &mut self.polling_source.as_mut() else {
+            return Ok(vec![]);
+        };
+        if !source.is_ready(now) {
+            return Ok(vec![]);
+        };
+
+        let messages = source.poll(timestamp).await?;
+        source.update_after_poll(now);
```
albinsuresh, line 86: Seeing `update_after_poll` called immediately after `poll` makes me wonder why separate them? Why not do the latter as part of `poll` itself? Initially I thought this separation would help with some unit testing. But I don't really see any such usages. Am I missing some hidden benefits with this seperation?

crates/extensions/tedge_flows/src/flow.rs
albinsuresh: Well, to my surprise, I just realised that such blanket implementations don't exist for `AsRef` and `AsMut`. I really thought that an impl like this would be there for all types:

```
impl<T> AsRef<T> for T {
    fn as_ref(&self) -> &T { self }
}
```

but just learned the Rust coherence/orphan rules preventing such an impl. TBH, I'm still not fully convinced by the arguments made against it. Still trying to understand it better. I expected the compiler to be smarter than that to differentiate between `impl AsRef<AnotherType> for T` from `impl AsRef<T> for T`, without conflicting `AnotherType` with `T`. But I'm sure they have their reasons.

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    async fn reload_script(&mut self, js_runtime: &mut JsRuntime, path: Utf8PathBuf) {
+        for flow in self.store_mut().flows_mut() {
+            for step in &mut flow.as_mut().steps {
+                if step.script.path() == path {
+                    match js_runtime.load_script(&mut step.script).await {
+                        Ok(()) => {
+                            step.script.init_next_execution();
+                            info!(target: "flows", "Reloaded flow script {path}");
```
rina23q: All other info messages are using present progressive (e.g. Removing, Adding, loading, ...)

```suggestion
                            info!(target: "flows", "Reloading flow script {path}");
```

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: Yes, this is intentional. Raising an error can be considered.

crates/extensions/tedge_flows/src/registry.rs
didier-wenzek: Indeed, nothing is removed. To remove a script from the JS engine (or to load a new one), you have to edit a flow configuration, removing (or adding) a step. The point of this method is only to warn the user when removing a script actually used by a flow, because that flow can then no more be reloaded.

crates/extensions/tedge_flows/src/registry.rs
didier-wenzek: > These functions could be moved into `FlowRegistry` itself as default implementations, right? As they only rely on the `store()` and `store_mut()` functions?

Technically, yes. However, it makes no sense to have another implementation. Hence grouped into an `Ext` trait.

> That'll make the separation between `FlowRegistry` and `FlowRegistryExt` clearer as well, where the former deals with all the core access primitives and the latter deals with the dependency on `JsRuntime`.

This is not the intent of that separation.  `FlowRegistryExt` groups the methods for which no other implementation makes sense.

crates/extensions/tedge_flows/src/registry.rs
didier-wenzek: fixed

crates/extensions/tedge_flows/src/registry.rs
didier-wenzek: That said there is a memory leak. Dropping or updating a flow doesn't drop the step functions from the JS runtime.

```
INFO flows: Memory usage:
INFO flows:   - malloc size: 1227696
INFO flows:   - used memory size: 989842
INFO flows:   - function count: 337
INFO flows:   - object count: 2881
INFO flows:   - array count: 707
INFO flows:   - string count: 523
INFO flows:   - atom count: 1642
INFO flows: Loading flow /etc/tedge/flows/measurements.toml           <= removing a step
INFO flows: Memory usage:
INFO flows:   - malloc size: 1241856
INFO flows:   - used memory size: 1001620
INFO flows:   - function count: 347                                <= should not increase
INFO flows:   - object count: 2907
INFO flows:   - array count: 707
INFO flows:   - string count: 523
INFO flows:   - atom count: 1642
```

This will be addressed in an independent PR.

PR #3843, opened 2025-10-30
feat: tedge flows statistics are published over MQTT

didier-wenzek:
## Proposed changes

- [x] Publish basic processing stats over MQTT
- [x]  Publish memory usage
- [x] Configure periodicity
- [ ] ~Configure what is published, what is not~

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -1372,6 +1372,14 @@ define_tedge_config! {
             #[tedge_config(example = "1h", default(from_str = "1h"))]
             #[tedge_config(note = "Any value less than one second is rounded up to one second.")]
             interval: SecondsOrHumanTime,
+
+            /// Enable or disable statistics for step onMessage
+            #[tedge_config(default(value = false))]
+            on_message: bool,
+
+            /// Enable or disable statistics for step onInterval
+            #[tedge_config(default(value = false))]
+            on_interval: bool,
         },
     },
 }
--- a/crates/core/tedge_mapper/src/az/mapper.rs
+++ b/crates/core/tedge_mapper/src/az/mapper.rs
@@ -103,9 +103,13 @@ impl TEdgeComponent for AzureMapper {
             .persist_builtin_flow("mea", az_converter.builtin_flow().as_str())
             .await?;
         let te = &tedge_config.mqtt.topic_root;
-        let dump_interval = tedge_config.flows.stats.interval.duration();
-        let service_config =
-            FlowsMapperConfig::new(&format!("{te}/{service_topic_id}"), dump_interval);
+        let stats_config = &tedge_config.flows.stats;
+        let service_config = FlowsMapperConfig::new(
+            &format!("{te}/{service_topic_id}"),
+            stats_config.interval.duration(),
+            stats_config.on_message,
+            stats_config.on_interval,
+        );
         let mut fs_actor = FsWatchActorBuilder::new();
         let mut cmd_watcher_actor = WatchActorBuilder::new();
 
--- a/crates/core/tedge_mapper/src/flows/mod.rs
+++ b/crates/core/tedge_mapper/src/flows/mod.rs
@@ -22,9 +22,13 @@ impl TEdgeComponent for GenMapper {
 
         let te = tedge_config.mqtt.topic_root.as_str();
         let service_topic_id = EntityTopicId::default_main_service(service_name)?;
-        let dump_interval = tedge_config.flows.stats.interval.duration();
-        let service_config =
-            FlowsMapperConfig::new(&format!("{te}/{service_topic_id}"), dump_interval);
+        let stats_config = &tedge_config.flows.stats;
+        let service_config = FlowsMapperConfig::new(
+            &format!("{te}/{service_topic_id}"),
+            stats_config.interval.duration(),
+            stats_config.on_message,
+            stats_config.on_interval,
+        );
 
         let mut fs_actor = FsWatchActorBuilder::new();
         let mut cmd_watcher_actor = WatchActorBuilder::new();
--- a/crates/extensions/tedge_flows/src/stats.rs
+++ b/crates/extensions/tedge_flows/src/stats.rs
@@ -27,6 +27,12 @@ pub enum Sample {
 }
 
 #[derive(Default)]
+pub struct StatsFilter {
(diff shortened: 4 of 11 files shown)


crates/core/tedge_mapper/src/flows/mod.rs
```diff
@@ -14,11 +17,18 @@ impl TEdgeComponent for GenMapper {
...
     ) -> Result<(), anyhow::Error> {
-        let (mut runtime, mut mqtt_actor) =
-            start_basic_actors("tedge-flows", &tedge_config).await?;
+        let service_name = "tedge-flows";
+        let te = &tedge_config.mqtt.topic_root;
+        let service_id = EntityTopicId::default_main_service(service_name).unwrap();
+        let service_config = FlowsMapperConfig {
+            statistics_topic: Topic::new(&format!("{te}/{service_id}/statistics"))?,
+            status_topic: Topic::new(&format!("{te}/{service_id}/status"))?,
+        };
```
didier-wenzek: It's now time to discuss the appropriate topics for flows status and statistics https://github.com/thin-edge/thin-edge.io/pull/3846#discussion_r2485287239

I think that it would make sense to publish the statistics as measurements under `{te}/{service_id}/m/statistics`. But then these stats will be pushed by default to the cloud.

reubenmiller: Yeah publishing under the standard thin-edge.io mqtt topic structure would be more consistent, but we might need to use a different telemetry endpoint to publish this information to avoid sending too much information to the cloud by default.

We could create a convention where some local/internal information which is meant for local/on-device consumption (rather than intended for the cloud):

```sh
{te}/{service_id}/local/statistics
```

Alternatively, we could use `flows` as the type since the flows statistics maybe published on serveral services in the future, e.g. tedge-mapper-c8y, tedge-mapper-az, tedge-flows etc.

```sh
{te}/{service_id}/flows/statistics
```

reubenmiller: A new proposal for the topic would be:

```
{te}/{service_id}/status/metrics
```

didier-wenzek: > A new proposal for the topic would be:
> 
> ```
> {te}/{service_id}/status/metrics
> ```

Done: https://github.com/thin-edge/thin-edge.io/pull/3843/commits/f0f75b7b648331f80211c8fda1b172d7ddcb2c1f

crates/extensions/tedge_flows/src/stats.rs
```diff
@@ -188,4 +203,28 @@ impl Dimension {
...
+
+    fn filename(path: &str) -> &str {
+        path.split('/').next_back().unwrap_or(path)
+    }
+}
+
+pub trait StatsPublisher {
+    type Record;
+
+    fn publish(&self, dim: &impl Display, stats: Value) -> Option<Self::Record>;
```
albinsuresh:
```suggestion
    fn publish_record(&self, dim: &impl Display, stats: Value) -> Option<Self::Record>;
```
`publish` felt like a misnomer as this function doesn't do the publishing itself.

didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3843/changes/eab8b0846a4c67f32e9c72843bf836944541eec1

didier-wenzek:
> > Making the 5 minutes stats collection interval configurable would be good, esp for manual testing and to have some system tests as well. The good thing is that we've got good unit test coverage for the same.

> Yes, I will add that.

Added: https://github.com/thin-edge/thin-edge.io/pull/3843/changes/161844e7fbe4db18ddf8fa85ee0982a4b4f6ccde

crates/core/tedge_mapper/src/az/mapper.rs
```diff
@@ -103,7 +103,9 @@ impl TEdgeComponent for AzureMapper {
             .persist_builtin_flow("mea", az_converter.builtin_flow().as_str())
             .await?;
         let te = &tedge_config.mqtt.topic_root;
-        let service_config = FlowsMapperConfig::new(&format!("{te}/{service_topic_id}"));
+        let dump_interval = tedge_config.flows.stats.dump_interval.duration();
```
albinsuresh: I guess this is fine for now and we can introduce mapper-specific overrides if/when there is demand.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -1364,8 +1364,15 @@ define_tedge_config! {
...
         #[tedge_config(example = "/usr/share/tedge/config-plugins,/usr/local/share/tedge/config-plugins", default(value = "/usr/share/tedge/config-plugins"))]
         plugin_paths: TemplatesSet,
-    }
+    },
 
+    flows: {
+        stats: {
+            /// The interval in seconds between flow statistics dumps
+            #[tedge_config(example = "1h", default(from_str = "1h"))]
+            dump_interval: SecondsOrHumanTime,
```
albinsuresh: **Optional**:
```suggestion
            publish_interval: SecondsOrHumanTime,
```

didier-wenzek: What about simply `interval`? i.e. `flows.stats.interval`?

reubenmiller:
@didier-wenzek Is there an example of some of the expected stats? For testing I changed the stats interval to `1s` but I seem to be only getting the memory stats and not the processing (though I have one flow installed). I was expecting to see some processing stats like number of messages processed etc. as the tedge-flows log entries seems to indicate that some "processing statistics" would be present.


Below is the only message that I'm seeing (at least at the interval as set in the config)

```text
[te/device/main/service/tedge-flows/status/metrics/memory] {"function_count":0,"malloc_bytes":91095,"memory_used_bytes":82848,"object_count":192}
```

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
didier-wenzek: > What about simply `interval`? i.e. `flows.stats.interval`?

Done: https://github.com/thin-edge/thin-edge.io/pull/3843/changes/e441e159a733f3a34ce22cc0eadd89975009f858

reubenmiller (approved):
Approved, great persistence  :)

PR #3846, opened 2025-10-31
fix: flaky tedge flows tests

didier-wenzek:
## Proposed changes

- [x] Tedge flows publish enabled/updated/removed status
- [x] Fix flaky tedge flows tests needing to wait a flow be activated before sending input messages

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -133,16 +139,27 @@ Consuming messages from a file, periodically
     [Teardown]    Uninstall Flow    read-file-periodically.toml
 
 Appending messages to a file
+    ${start}    Get Unix Timestamp
     Install Flow    append-to-file.toml
+    Should Have MQTT Messages
+    ...    topic=te/device/main/service/tedge-flows/status/flows
+    ...    minimum=1
+    ...    message_contains=append-to-file
+    ...    date_from=${start}
```
didier-wenzek: This is not working, but why? The messages are correctly published by tedge flows:

```
[te/device/main/service/tedge-flows/status/flows] {"flow":"/etc/tedge/flows/append-to-file.toml","status":"updated","time":1761916482}
```

didier-wenzek: I also failed to improve `Install Flow` with the following:

```
Install Flow
    [Arguments]    ${definition_file}
    ThinEdgeIO.Transfer To Device    ${CURDIR}/input-ext/${definition_file}    /etc/tedge/flows/
    Should Have MQTT Messages
    ...    topic=te/device/main/service/tedge-flows/status/flows
    ...    minimum=1
    ...    message_contains=${definition_file}
    ...    date_from=${start}
```

Any clue about what I'm missing?

didier-wenzek: So the main error is here:

```
ThinEdgeIO.Transfer To Device    ${CURDIR}/input-ext/${definition_file}    /etc/tedge/flows/
```

- `${CURDIR}/input-ext` doesn't exist. The correct directory is `${CURDIR}/flow/input-ext`
- In such a case `ThinEdgeIO.Transfer To Device` doesn't fail, making the error less visible
- Also `ThinEdgeIO.Transfer To Device` copies nothing when the second arg doesn't have a trailing `/`.

But why things were sometimes working?

- All the flows under  `${CURDIR}/flow/input-ext` were actually installed by `Copy Configuration Files` and loaded by tedge flows (because being under a sub-directory)
- In other words, the changes introduced [here](https://github.com/thin-edge/thin-edge.io/pull/3831) were changing *nothing* from a test perspective: all the flows were loading on start :-(.

didier-wenzek: => A fix has been pushed (and squashed as the previous commit was fully incorrect):

- The input flows examples have been out the main directory of samples (so no more silently installed)
- `Copy Configuration Files` has been fixed.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -163,6 +171,38 @@ impl FlowsMapper {
...
+        } else {
+            "removed"
+        };
+        let status = Self::flow_status(flow, status, &now);
+        self.mqtt_sender.send(status).await?;
+        Ok(())
+    }
+
+    fn flow_status(flow: &str, status: &str, time: &OffsetDateTime) -> MqttMessage {
+        let topic = Topic::new_unchecked("te/device/main/service/tedge-flows/status/flows");
```
didier-wenzek, line 197: Hard-coded for now. This will be done properly in https://github.com/thin-edge/thin-edge.io/pull/3843

albinsuresh, line 197:
```suggestion
        let topic = Topic::new_unchecked("te/device/main/service/tedge-flows/status/flow");
```
Since the payload is the status of a single flow and not an aggregate stat across all flows.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -163,6 +171,38 @@ impl FlowsMapper {
...
+        let payload = json!({
+            "flow": flow,
```
albinsuresh, line 199: I'm wondering if it'd be better to include the `flow` name in the **topic** instead of the payload so that these status messages can be published as **retained** messages per flow. That'll help the user get an overview of the status of all the flows by subscribing to `status/flow/#` topic.

@reubenmiller Thoughts?

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -180,19 +180,31 @@ Custom Setup
     Start Service    tedge-flows
 
 Copy Configuration Files
-    ThinEdgeIO.Transfer To Device    ${CURDIR}/flows/*    /etc/tedge/flows/
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/flows/*.js    /etc/tedge/flows/
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/flows/*.toml    /etc/tedge/flows/
```
albinsuresh, line 184: Side comment unrelated to this PR: We could consider installing the relevant flow at the test level for older tests as well in an independent PR.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -52,6 +56,7 @@ impl Actor for FlowsMapper {
 
     async fn run(mut self) -> Result<(), RuntimeError> {
         self.send_updated_subscriptions().await?;
+        self.notify_flows_status().await?;
```
albinsuresh, line 59: Not a deal breaker or anything, but I see duplicate `status/flow` messages whenever flows are dynamically installed. I believe it's because of the fs-notify issue where a newly created file emits both `Modified` and `Updated` events. Wondering if we can react to `Modified` events only.

reubenmiller: Don't we have some generic logic which debounces inotify notifications?

crates/extensions/tedge_flows/src/actor.rs
reubenmiller: I think we're going to address the exact topic/s which the stats are published to in a follow up PR...so I assume a single topic was just chosen for this PR, and the follow-up PR will publish them on individual topics.

crates/extensions/tedge_flows/src/actor.rs
didier-wenzek: Sure, but the feature is `tedge flows`

crates/extensions/tedge_flows/src/actor.rs
didier-wenzek: > include the `flow` name in the **topic** instead of the payload so that these status messages can be published as **retained** messages per flow.

I considered that but decided against it for now. Indeed this raises other questions, such as "what about a flow which definition is removed while the mapper was down?". The goal of this PR was more to quickly come with a fix.

crates/extensions/tedge_flows/src/actor.rs
didier-wenzek: > Don't we have some generic logic which debounces inotify notifications?

Yes we have. But there are still some surprising event sequences.

> I believe it's because of the fs-notify issue where a newly created file emits both Modified and Updated events. Wondering if we can react to Modified events only.

For now, we are on the safe side - with duplicated events. I will need some time to investigate if only reacting to `Modified` is correct.

crates/extensions/tedge_flows/src/actor.rs
albinsuresh: Okay. We can take a final call on this based on whether flow names would be included in the topic or not. In the current state where all flow stats are published to the same topic, this seems appropriate.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
reubenmiller: Note: A [PR](https://github.com/thin-edge/thin-edge.io/pull/3856) has been created to update the robotframework-devicelibrary which adds a check to the `Transfer to Device` keyword where at least 1 file must be found, otherwise an error is thrown. This should help identifier simple mistakes like using the wrong source path more quickly.

PR #3847, opened 2025-11-04
fix: hanging workflow awaiting a sub-operation resumed on agent restart

reubenmiller:
## Proposed changes

`await-agent-restart` never completes when its `on_success` target is a final state, and the operation is invoked as a sub-operation. The invoking workflow stays stuck in `await-operation-completion` until it is timed out.

On restart, `WorkflowSupervisor::load_pending_commands` resumes every pending command — moving those in an `await-agent-restart` step to their `on_success` state — but it only returned the resumed states to the caller: the command board itself kept the states persisted before the agent was interrupted.

This goes unnoticed as long as the resumed state triggers an action that persists a new state on its way (e.g. an intermediate `action = "proceed"` step). It is fatal when `on_success` points directly at a final state, because the resulting `cleanup` action updates no state at all. The invoking command is then resumed, reads its sub-command from the board, still sees the stale `await-agent-restart` status, concludes the sub-operation is running, and waits
forever.

The fix stores the resumed states on the command board, so a command awaiting a sub-command observes the actual state of that sub-command and not the state persisted before the interruption. The agent also skips the resumed commands that have been cleared in the meantime — the pending commands are resumed in a non deterministic order, and a sub-command may be finalized and cleared while resuming its invoking command; without the check its retained state would be published again and left stuck on the broker.

This also fixes the same class of hang for the built-in `config_update` workflow when it is wrapped by a custom or device-profile operation.

**Verification**

The system test was added first and confirmed to fail in the [CI Run](https://github.com/thin-edge/thin-edge.io/actions/runs/30989686873) before the fix was applied.

## Types of changes

## Paste Link to the issue

* https://github.com/thin-edge/thin-edge.io/issues/3848

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
+++ b/tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
@@ -253,6 +253,22 @@ The max log file count can be configured per operation
     Should Contain    ${log_files}    workflow-lite_software_update-n205.log
     Should Contain    ${log_files}    workflow-lite_software_update-n206.log
 
+Trigger wrapped workflow without an intermediate state
+    [Documentation]    A sub-operation awaiting the agent restart and moving on success directly to a
+    ...    final state, must be seen as successful by the operation awaiting its completion.
+    ${pid_before}    Get Service PID    tedge-agent
+    Execute Command
+    ...    tedge mqtt pub --retain te/device/main///cmd/restart-tedge-agent-wrapper/robot-4 '{"status":"init"}'
+    Should Have MQTT Messages
+    ...    te/device/main///cmd/restart-tedge-agent-wrapper/robot-4
+    ...    message_pattern=.*successful.*
+    ...    timeout=120
+    ${pid_after}    Get Service PID    tedge-agent
+    Should Not Be Equal    ${pid_before}    ${pid_after}    msg=tedge-agent should have been restarted
+    [Teardown]    Run Keywords
+    ...    Execute Command    tedge mqtt pub --retain te/device/main///cmd/restart-tedge-agent-wrapper/robot-4 ''
+    ...    AND    Get Logs
+
 
 *** Keywords ***
 Custom Test Setup
@@ -277,6 +293,8 @@ Copy Configuration Files
     ThinEdgeIO.Transfer To Device    ${CURDIR}/restart-tedge-agent.toml    /etc/tedge/operations/
     ThinEdgeIO.Transfer To Device    ${CURDIR}/tedge-agent-pid.sh    /etc/tedge/operations/
     ThinEdgeIO.Transfer To Device    ${CURDIR}/native-reboot.toml    /etc/tedge/operations/
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/restart-tedge-agent-wrapper.toml    /etc/tedge/operations/
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/restart-tedge-agent-internal.toml    /etc/tedge/operations/
     ThinEdgeIO.Transfer To Device    ${CURDIR}/gp_command.toml    /etc/tedge/operations/
     ThinEdgeIO.Transfer To Device    ${CURDIR}/super_command.toml    /etc/tedge/operations/
     ThinEdgeIO.Transfer To Device    ${CURDIR}/sub_command.toml    /etc/tedge/operations/
--- a/crates/core/tedge_agent/src/operation_workflows/actor.rs
+++ b/crates/core/tedge_agent/src/operation_workflows/actor.rs
@@ -742,6 +742,16 @@ impl WorkflowActor {
                     .load_pending_commands(pending_commands)
                     .await
                 {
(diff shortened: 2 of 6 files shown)


tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -253,6 +253,21 @@ The max log file count can be configured per operation
     Should Contain    ${log_files}    workflow-lite_software_update-n205.log
     Should Contain    ${log_files}    workflow-lite_software_update-n206.log
 
+Trigger wrapped workflow without an intermediate state
+    [Documentation]    A sub-operation awaiting the agent restart and moving on success directly to a
+    ...    final state, must be seen as successful by the operation awaiting its completion.
+    Execute Command
+    ...    cmd=echo 'tedge ALL = (ALL) NOPASSWD:SETENV: /usr/bin/tedge, /usr/bin/systemctl, /etc/tedge/sm-plugins/[a-zA-Z0-9]*, /bin/sync, /sbin/init, /sbin/shutdown, /usr/sbin/reboot, /usr/bin/on_shutdown.sh' > /etc/sudoers.d/tedge
+    ${pid_before}    Execute Command    sudo systemctl show --property MainPID tedge-agent
```
rina23q: Can be replaced with `Get Service PID` keyword as other test uses.

```suggestion
    ${pid_before}    Get Service PID    tedge-agent
```

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -253,6 +253,21 @@ The max log file count can be configured per operation
     Should Contain    ${log_files}    workflow-lite_software_update-n205.log
     Should Contain    ${log_files}    workflow-lite_software_update-n206.log
 
+Trigger wrapped workflow without an intermediate state
+    [Documentation]    A sub-operation awaiting the agent restart and moving on success directly to a
+    ...    final state, must be seen as successful by the operation awaiting its completion.
+    Execute Command
+    ...    cmd=echo 'tedge ALL = (ALL) NOPASSWD:SETENV: /usr/bin/tedge, /usr/bin/systemctl, /etc/tedge/sm-plugins/[a-zA-Z0-9]*, /bin/sync, /sbin/init, /sbin/shutdown, /usr/sbin/reboot, /usr/bin/on_shutdown.sh' > /etc/sudoers.d/tedge
```
rina23q: The difference from the sudores file created in `Custom Test Setup` is `/usr/bin/reboot`. The new workflow files added by this PR doesn't use it. So this can be removed.

```suggestion
```

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -253,6 +253,21 @@ The max log file count can be configured per operation
...
+    ${pid_before}    Execute Command    sudo systemctl show --property MainPID tedge-agent
+    Execute Command
+    ...    tedge mqtt pub --retain te/device/main///cmd/restart-tedge-agent-wrapper/robot-1 '{"status":"init"}'
+    Should Have MQTT Messages
+    ...    te/device/main///cmd/restart-tedge-agent-wrapper/robot-1
+    ...    message_pattern=.*successful.*
+    ...    timeout=120
+    ${pid_after}    Execute Command    sudo systemctl show --property MainPID tedge-agent
+    Should Not Be Equal    ${pid_before}    ${pid_after}    msg=tedge-agent should have been restarted
+
```
rina23q, line 271: No one clears the `restart-tedge-agent-wrapper/robot-1`. Add it here, the end of the test.

```suggestion
    Execute Command
    ...    tedge mqtt pub --retain te/device/main///cmd/restart-tedge-agent-wrapper/robot-1 ''
```

rina23q (commented):
It looks good overall. I found several improvements on the robot test.

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
reubenmiller: Good call, I'll add it to the teardown

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
reubenmiller: Resolved in 39dae278b729cffa656c22b1d462880aedbc695e

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
reubenmiller: Resolved in 39dae278b729cffa656c22b1d462880aedbc695e

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
reubenmiller: Resolved in 39dae278b729cffa656c22b1d462880aedbc695e

rina23q (approved):

PR #3851, opened 2025-11-06
refactor: Enable mapper configuration to be separated from tedge.toml

jarhodes314:
## Proposed changes
An alternative solution to generalised mapper configuration (original proposal is in #3835).

This separates the mapper configurations entirely from tedge.toml.

TODO:
- [x] Adopt compatibility layer for non-c8y clouds
- [x] Remove `CloudConfig` implementations for `TEdgeConfigReaderC8y` etc.
- [x] Attempt to read the configuration only once
- [x] Better handle `TryFrom<...> for BridgeConfigC8yParams`
- [x] Fix the bridge_include nonsense - this is done by making `bridge.include.local_cleansession` a configuration for all clouds
- [ ] ~Work out how to handle `bridge.include.local_cleansession` for non c8y clouds~ This will be fixed in the following PR which uses existing `TEdgeConfig` deserialisation layer
- [x] Improve URL missing error message for the compatibility layer
- [x] Replace hardcodings of `mappers` config directory with a single source of truth
- [x] `mapper.url` should be private
- [x] Unify check for matching url/device id in `c8y` and generalised configs
- [x] Remove unnecessary `derive(Deserialize)` in mapper config
- [ ] ~Add code interface for cli support and use this in `tedge connect` validations for unique proxy bind ports~ This will be fixed in next PR
- [ ] Environment variable support?

### Example configuration
To use the feature, you need to create a mapper config file. In `/etc/tedge/mappers/c8y.toml`, you add the `c8y` fields from `tedge.toml`

```toml
url = "example.cumulocity.com"
root_cert_path = "/etc/my/c8y/certs"

proxy.bind.port = 8123
```

If you want to try it out, `tedge-mapper c8y` will read from the new configuration if it exists. For profiles, the config goes in `/etc/tedge/mappers/c8y.d/profile-name.toml`.

### Code changes
The bulk of the code this PR touches at the moment is in `mapper_config.rs`. This is essentially extracting the mapper related configurations from tedge.toml, allowing us to deserialise the configuration. It doesn't add any support for editing the mapper configuration file, so for the moment this will need to be done by hand. I anticipate adding some CLI support for the mapper configuration file later, but this shouldn't have a major impact on the more critical changes made by this PR.

Most of `mapper_config.rs` was written by Claude. It could do with some restructuring/splitting out, my focus was to get the infrastructure in place to allow the existing mapper code to read the new configuration. The other changes, at least as of the first commit are to support `tedge-mapper c8y`. As of the first commit, `tedge-mapper c8y` sources its configuration from `mapper.config_path` (with no profile set). I will need to add a compatibility layer later.

### Differences between the two proposals
The "mapper config in tedge.toml" proposal (#3835) stored the "specialised" (e.g. c8y-specific) configurations inside a `c8y` sub-field, so you would set e.g. `mapper.profiles.c8y.c8y.proxy.bind_port`. This proposal keeps the existing hierarchy, so the cloud-specific fields are at the same level of the configuration as everything else in the toml file (e.g. `proxy.bind.port` above is at the top-level like everything else).

When deserialising the old configuration, the specialised mapper configuration was read as an enum. This is a pain when handling the configuration since you have to pass the specialised configuration around in addition to the mapper config that contains it. By comparison, the new approach uses generics to handle the specialised mapper configurations. This means the c8y mapper can check for the given profile that `mapper.type = "c8y"`, then deserialise with only the c8y specialisations supported.

Previous solution:
```rust
let mapper_config = tedge_config.mapper.try_get(profile)?;

match mapper_config.ty {
    CloudType::C8y { c8y } => {
        // Access c8y specific fields through `c8y`
        println!("{}", c8y.proxy.bind_port);

        // We need to pass `c8y` around now to save us doing this enum destructuring again
        C8yMapperConfig::from(&tedge_config, &mapper_config, &c8y),
    },
    CloudType::Az { .. } => panic!("c8y mapper requires mapper.type to be set to 'c8y'")
}
```

This solution:
```rust
let mapper = tedge_config.mapper.try_get(profile)?;
assert_eq!(mapper.ty, "c8y");

let mapper_config = tedge_config::mapper_config::load_mapper_config(&mapper.config_path.or_config_not_set()?, &tedge_config).await?;

// Access c8y specific fields through the cloud type
println!("{}", mapper_config.cloud_specific.proxy.bind_port);

// We only need to pass mapper_config around
let c8y_mapper_config = C8yMapperConfig::from(&tedge_config, &mapper_config),
```

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -525,7 +525,7 @@ impl TEdgeConfig {
 }
 
 /// Return the trust store as a stream, for easy chaining
-/// 
+///
 /// Note: This does not stream the certificates as they are read from disk. It
 /// simply reads all the certificates, then returns them as a stream.
 fn stream_trust_store<T: SpecialisedCloudConfig>(
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
@@ -599,11 +599,6 @@ pub enum MapperConfigError {
     #[error("Failed to parse mapper configuration: {0}")]
     TomlParse(#[from] toml::de::Error),
 
-    // TODO remove me
-    /// Required field is missing after applying defaults
-    #[error("Required field '{field}' is missing from mapper configuration")]
-    MissingField { field: &'static str },
-
     /// Failed to read from tedge config
     #[error("Failed to read from tedge config: {0}")]
     ConfigRead(String),
--- a/crates/common/certificate/src/cloud_root_certificate.rs
+++ b/crates/common/certificate/src/cloud_root_certificate.rs
@@ -2,12 +2,12 @@ use anyhow::Context;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use reqwest::Certificate;
-use tokio_stream::Stream;
-use tokio_stream::StreamExt as _;
-use tokio_stream::wrappers::ReadDirStream;
 use std::io::Cursor;
 use std::pin::Pin;
 use std::sync::Arc;
+use tokio_stream::wrappers::ReadDirStream;
+use tokio_stream::Stream;
+use tokio_stream::StreamExt as _;
 
 #[derive(Debug, Clone)]
 pub struct CloudHttpConfig {
@@ -94,7 +94,10 @@ async fn stream_file_or_directory(
         Box::pin(dir_stream.map(move |file| match file {
             Ok(file) => {
                 let mut path = base_path.clone();
-                path.push(file.file_name().to_str().ok_or(anyhow::anyhow!("encountered non-utf8 path: {}", file.file_name().display()))?);
+                path.push(file.file_name().to_str().ok_or(anyhow::anyhow!(
+                    "encountered non-utf8 path: {}",
+                    file.file_name().to_string_lossy()
+                ))?);
                 Ok(path)
             }
             Err(e) => Err(e).with_context(|| format!("reading metadata for file in {base_path}")),


jarhodes314:
> The separation of mapper configs from the main `tedge.toml` has opened up a lot of possibilities, esp for multi-profile management. I've got a few suggestions on how things are laid out on the file-system, which I feel might simplify the management of these configs. How about having a config file structure as follows:
> 
> ```
> /etc/tedge
> |    |-- tedge.toml
> |    |-- c8y.toml
> |    |-- c8y.d/
> |    |    |-- profile1.toml
> |    |    |-- profile2.toml
> |    |-- az.toml
> |    |-- az.d/
> |    |    |-- profile1.toml
> |    |    |-- profile2.toml
> |    |-- generic1.toml
> |    |-- generic2.toml
> ```
> 
> Where the `c8y.toml` and `az.toml` represents the default profile configs as it is today. All the additional profiles are kept under the associated config extension directories like `c8y.d`, `az.d` etc. When a default profile needs to be changed, we can just make the `c8y.toml` a symlink that points to a named profile. When `c8y.toml` is converted from a toml file to a symlink, we can provide additional tooling to backup the existing contents of it into another named profile file. Or if we don't want the `c8y.toml` file to change from a regular file to symlink from time to time, we can keep it as a symlink all the time, that points to a profile of the same name as follows: `/etc/tedge/c8y.toml --> /etc/tedge/c8y.d/c8y.toml` by default.
> 
> I hope that the TOML schema enforcement becomes easier for different clouds as there is a clear correlation between the names of the config files/directories and their types. WDYT?

The reason why the current design defines the paths and associated cloud types in `tedge.toml`, and a large part of this reason is so we can deserialise the specialised configuration (e.g. the c8y-only fields) appropriately. Unfortunately `#[serde(flatten)]`, which we use for the specialised configuration, precludes us from using `#[serde(deny_unknown_fields)]`, so it's hard to give the user an error or warning for anything that shouldn't be in the configuration. I think we can add schema enforcement ourselves though, it would just be less "zero-cost" than the serde approach, which detects the unwanted fields as they're read from the toml file.

I'm not entirely sure how to handle different profiles at the moment. With the current design, `tedge-mapper c8y` will use `mapper.config_path` (i.e. the default, non-profiled configuration) to determine where to read its configuration from, and if `mapper.type` isn't set to `c8y`, then `tedge-mapper c8y` will fail to start. This essentially means with the current design, customers with devices connected to two different clouds would need to use profiles for at least one of the cloud configurations to use the new generalised configuration.

I think a possible solution to this is that we make profile names required, so `tedge-mapper c8y` (with no explicit profile specified) would actually use the `c8y` profile. This does raise the question of how to handle the `tedge.toml` aspects, and just getting rid of them in favour of an approach like what you suggested may well be a good solution to this, and I don't see any reason why this wouldn't work. My only hesitation is that at first glance to me it looked like the proposed configuration is hierarchical, where `c8y.d/profile1.toml` inherits from `c8y.toml`. I believe this isn't actually the case, `c8y.toml` is just the default profile, and `c8y.d/profile1.toml` is separate, but I just thought it's worth mentioning since it was the first thing that came to mind looking as the example file structure.

jarhodes314:
> The separation of mapper configs from the main `tedge.toml` has opened up a lot of possibilities, esp for multi-profile management. I've got a few suggestions on how things are laid out on the file-system, which I feel might simplify the management of these configs. How about having a config file structure as follows:
> 
> ```
> /etc/tedge
> |    |-- tedge.toml
> |    |-- c8y.toml
> |    |-- c8y.d/
> |    |    |-- profile1.toml
> |    |    |-- profile2.toml
> |    |-- az.toml
> |    |-- az.d/
> |    |    |-- profile1.toml
> |    |    |-- profile2.toml
> |    |-- generic1.toml
> |    |-- generic2.toml
> ```
> 
> Where the `c8y.toml` and `az.toml` represents the default profile configs as it is today. All the additional profiles are kept under the associated config extension directories like `c8y.d`, `az.d` etc. When a default profile needs to be changed, we can just make the `c8y.toml` a symlink that points to a named profile. When `c8y.toml` is converted from a toml file to a symlink, we can provide additional tooling to backup the existing contents of it into another named profile file. Or if we don't want the `c8y.toml` file to change from a regular file to symlink from time to time, we can keep it as a symlink all the time, that points to a profile of the same name as follows: `/etc/tedge/c8y.toml --> /etc/tedge/c8y.d/c8y.toml` by default.
> 
> I hope that the TOML schema enforcement becomes easier for different clouds as there is a clear correlation between the names of the config files/directories and their types. WDYT?

I have now implemented this

didier-wenzek:
> I have now implemented this

An intermediate `mappers` directory has been introduced and that's a good idea that removes any confusion with the existing `c8y` directory.

```
/etc/tedge
|    |-- tedge.toml
|    |-- c8y
|    |   -- operation1
|    |-- mappers
|    |    |-- c8y.toml
|    |    |-- c8y.d/
|    |    |    |-- profile1.toml
|    |    |    |-- profile2.toml
|    |    |-- az.toml
|    |    |-- az.d/
|    |    |    |-- profile1.toml
|    |    |    |-- profile2.toml
|    |    |-- generic1.toml
|    |    |-- generic2.toml
```

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -313,7 +319,8 @@ fn core_mqtt_bridge_config(
         tc.forward_from_local("s/uat", local_prefix.clone(), "")?;
     }
 
-    let c8y = c8y_config.mqtt.or_config_not_set()?;
+    // TODO mapper.mqtt should probably be a global setting
+    let c8y = c8y_config.mqtt().or_config_not_set()?;
```
didier-wenzek: I don't get this TODO comment
 
- What do you mean by a global setting?
- Is this related to `c8y_config.cloud_specific`?
- Is the difference between global/specific, hidden behind `c8y_config.mqtt()`function call ?
- Would a `c8y_config.mqtt` property be global?

crates/core/tedge/src/command.rs
```diff
@@ -147,10 +148,14 @@ pub trait Command {
 ///     }
 /// }
 /// ```
+#[async_trait::async_trait]
 pub trait BuildCommand {
```
didier-wenzek, line 152: Can this change be done along another PR?

Asking because this sounds unrelated and broadens the scope of things to be tested. E.g. adding "command completion" to "tedge config", "mapper config", "mapper profiles" ...

crates/common/tedge_config/src/tedge_toml/models/c8y_software_management.rs
```diff
@@ -6,6 +6,7 @@ use strum::Display;
 #[derive(
     Debug, Clone, Copy, serde::Serialize, serde::Deserialize, Eq, PartialEq, doku::Document, Display,
 )]
+#[serde(rename_all = "kebab-case")]
 #[strum(serialize_all = "camelCase")]
 pub enum SoftwareManagementApiFlag {
```
didier-wenzek: What's the motivation/impact for using kebab-case here?

crates/core/tedge/src/command.rs
jarhodes314: The motivation for this was that the `mapper_config` method (on `TEdgeConfig`) is quite complicated, as a result I didn't want to have sync and async versions of the same method. Possibly there could be a sync version that detects the current runtime and just runs the async version, but this felt like a nicer way to tackle that problem. 

I guess the more fundamental change this represents is we have gone from loading all the configuration in one go (with the exception of `device.id` since that is derived from the certificate as and when it's needed) to loading the mapper configuration lazily, we only open the mapper configuration file when we (first) need it.

jarhodes314: The motivation for this was that the `mapper_config` method (on `TEdgeConfig`) is quite complicated, as a result I didn't want to have sync and async versions of the same method. Possibly there could be a sync version that detects the current runtime and just runs the async version, but this felt like a nicer way to tackle that problem. 

I guess the more fundamental change this represents is we have gone from loading all the configuration in one go (with the exception of `device.id` since that is derived from the certificate as and when it's needed) to loading the mapper configuration lazily, we only open the mapper configuration file when we (first) need it.

didier-wenzek: > I didn't want to have sync and async versions of the same method.

I'm fully aligned with that. 

> loading the mapper configuration lazily, we only open the mapper configuration file when we (first) need it.

Indeed a change that makes sense now that we have independent config files.

> Can this change be done along another PR?

Okay, let's keep this change in that PR.

didier-wenzek:
I like a lot the new way to configure mappers and profiles using independent file.

However, it would be good to raise at least a warning when a non-profiled c8y configuration is given both in `/etc/tedge/tedge.toml` and `/etc/tedge/mappers/c8y.toml`. It nice to give the priority to the latter, but it's confusing when you edit the former without any visible effect.

crates/common/tedge_config/src/tedge_toml/models/c8y_software_management.rs
jarhodes314: No particular motivation, I should have used `camelCase`. The impact point is more important though; indeed this change breaks configuration parsing for existing users that have manually set `c8y.software_management.api`. I will change the `rename_all` to be `camelCase` and add `serde(alias = "...")` attributes to the individual values for backwards-compatibility (and add a regression test to make sure the existing toml files can be deserialised).

This is working under the assumption that we want the toml file to as closely as possible match the values users use and observe via the `tedge config` cli, which I believe is what we want

At the moment there is an inconsistency if users wish to use environment variables with toml syntax (there isn't a particularly good reason why people would use toml syntax here, but given we can make it work sensibly, I think we ought to):

```shell
# Unquoted, the value matches the `tedge config` capitalisation
$ env TEDGE_C8Y_SOFTWARE_MANAGEMENT_API=legacy tedge config get c8y.software_management.api
legacy

# Note that when quoted, the value needs to be capitalised, since serde is used, not strum
$ env TEDGE_C8Y_SOFTWARE_MANAGEMENT_API=\"Legacy\" tedge config get c8y.software_management.api
legacy

# The error message is particularly confusing for quoted values with the wrong casing
$ env TEDGE_C8Y_SOFTWARE_MANAGEMENT_API=\"legacy\" tedge config get c8y.software_management.api
Error: Failed to set value for c8y.software_management.api to "\"legacy\"" from environment variable TEDGE_C8Y_SOFTWARE_MANAGEMENT_API

Caused by:
    0: Failed to parse input
    1: Failed to parse flag: "legacy". Supported values are: legacy, advanced

# The other way round, the error case makes much more sense
$ env TEDGE_C8Y_SOFTWARE_MANAGEMENT_API=Legacy tedge config get c8y.software_management.api
Error: Failed to set value for c8y.software_management.api to "Legacy" from environment variable TEDGE_C8Y_SOFTWARE_MANAGEMENT_API

Caused by:
    0: Failed to parse input
    1: Failed to parse flag: Legacy. Supported values are: legacy, advanced
```

jarhodes314: Now implemented, along with tests.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1746 @@
...
+pub trait SpecialisedCloudConfig:
+    Sized + DeserializeOwned + ApplyRuntimeDefaults + ExpectedCloudType + FromCloudConfig + Send + Sync + 'static
+{
+    /// The mapper-specific configuration type for this cloud
+    type SpecialisedMapperConfig: DeserializeOwned + std::fmt::Debug + Document + Default + Send + Sync;
+}
+
+/// Base mapper configuration with common fields and cloud-specific fields via generics
+#[derive(Debug, Document)]
+pub struct MapperConfig<T: SpecialisedCloudConfig> {
```
albinsuresh, line 117: I'm wondering if this generic representation of a `MapperConfig` with the cloud specific type wrapped as a generic type is worth the complexity of all the indirections, esp with the current design that keeps each cloud mapper config in their own "named" files and directories. Why not let the mappers parse their config files into cloud specific mapper config types like `C8yMapperConfig`, `AzMapperConfig` and even a `GenericMapperConfig` directly rather than going thorugh `tedge_config::mapper_config()` that returns this generic type? These common fields like `url`, `root_cert_path` etc can be captured into a `BaseMapperConfig` (or even the `GenericMapperConfig` itself) and wrapped into the cloud specific ones as a `flattened` field so that each `toml` file can be directly parsed into one of these config types.

In my simplistic view, the `TEdgeConfig` struct doesn't really have to read all mapper configs and keep them cached, as every mapper instance is only interested in the common `tedge` configurations and the cloud config for its own type and profile. So, let the mappers parse the cloud config files that are relevant for its type and profile directly, rather than letting `tedge_config` parse everything into generic `MapperConfig` instances and have the mappers looking them up later on.

With such a scheme, the `tedge_config` layer just needs to act like the router between the cloud specific config types and their corresponding file locations (or TOML tables in the old config scheme), along with the magic for `device.id`.

albinsuresh: While looking at `tedge_config` only from the mapper process's PoV, may be I've oversimplified the requirements here, without fully appreciating the complexities of having a shared definition of the config structs between the `tedge config` cli utility as well as the consumers of the configs like the mapper.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -100,6 +135,445 @@ impl TEdgeConfig {
...
+            .await
+        {
+            ConfigDecision::LoadNew { .. }
+            | ConfigDecision::NotFound { .. }
+            | ConfigDecision::PermissionError { .. } => Box::new(
+                futures::stream::once(ready(None)).chain(
+                    tokio_stream::wrappers::ReadDirStream::new(
+                        tokio::fs::read_dir(self.mappers_config_dir().join(format!("{ty}.d")))
+                            .await
+                            .unwrap(),
```
didier-wenzek: Having a complete mapper config hierarchy should not be mandatory. 

```
$ sudo tedge-mapper c8y               
2025-11-28T09:26:46.883035301Z  INFO Runtime: Started
2025-11-28T09:26:46.883403123Z  INFO Runtime: Running Signal-Handler-0
2025-11-28T09:26:46.883467269Z  INFO Runtime: Running HealthMonitorActor-1

thread 'main' panicked at /home/didier/production/thin-edge.io/crates/common/tedge_config/src/tedge_toml/tedge_config.rs:316:30:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


$ fd . /etc/tedge/mappers                                                                       
/etc/tedge/mappers/c8y.d/
/etc/tedge/mappers/c8y.d/foo.toml
/etc/tedge/mappers/c8y.toml
```

The fix is simple but should not be mandatory:

```
$ mkdir /etc/tedge/mappers/az.d
$ mkdir /etc/tedge/mappers/aws.d
$ sudo systemctl start tedge-mapper-c8y
$ sudo systemctl status tedge-mapper-c8y

     Active: active (running) since Fri 2025-11-28 10:43:50 CET; 3min 29s ago

```

jarhodes314: Ah, this turned out to be a bug where we always consulted the c8y entry for where the configuration lies, so something was determining the az or aws configuration, discovering that c8y is migrated to the new format, and attempting to read the az/aws configuration from the new format. That's easy to fix.

However, this does raise an additional point about whether `c8y.d` must exist to read `c8y.toml`. I think the answer is "no, we should cope with the profiles directory not existing" (even if we automatically create it regardless of whether profiles are in use when performing a hypothetical automated migration from tedge.toml to separate mapper config in the future).

didier-wenzek: > However, this does raise an additional point about whether `c8y.d` must exist to read `c8y.toml`. I think the answer is "no, we should cope with the profiles directory not existing"

I agree and also think thin-edge should not mandate a complete file hierarchy when a single configuration file is enough.

jarhodes314: I've now fixed the original bug, and improved the error message in the case of the default profile existing but not the profiled-configs directory. I'm still working on actually preventing the error entirely, though.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -100,6 +135,445 @@ impl TEdgeConfig {
...
+        Ok(CloudHttpConfig::new(roots.clone(), proxy))
+    }
+}
+
+fn stream_trust_store<T: SpecialisedCloudConfig>(
+    mapper_config: Arc<MapperConfig<T>>,
+) -> impl Stream<Item = Certificate> {
+    futures::stream::once(async move {
+        read_trust_store(&mapper_config.root_cert_path)
+            .await
```
didier-wenzek, line 536: [minor] I don't get the motivation to return an asynchronous `Stream` of `Certificate` here. Indeed, the method `read_trust_store` is not `async` (or more precisely only made artificially `async` by this PR).

crates/core/tedge/src/cli/config/cli.rs
```diff
@@ -183,3 +187,61 @@ impl BuildCommand for ConfigCmd {
         }
     }
 }
+
+pub async fn restrict_cloud_config_update(
```
didier-wenzek, line 191: As of now this function is a *check* used by `tedge config set` to display an error message when the setting to update is not in the usual location. What amount of work would be required to transform this function into one that *directs* `tedge config set` to the appropriate file?

The appropriate file to update is correctly retrieved by `tedge_config.decide_config_source()` and displayed to the user. What prevents the command to *use* and *update* that file?
- Is this only because the cloud prefix is not given there?
- Would that help to keep the prefix in the cloud configuration file? (i.e. having `c8y.url = xxx` in `/etc/tedge/mappers/c8y.toml` exactly as in `/etc/tedge/tedge.toml`)
- Is this due to something deeper?

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1738 @@
...
+#[derive(Debug, Deserialize, Document)]
+pub struct AwsMapperSpecific {
+    /// Whether to add timestamps to messages
+    #[serde(default = "default_timestamp")]
+    pub timestamp: bool,
+
+    /// The timestamp format to use
+    #[serde(default = "default_timestamp_format")]
+    pub timestamp_format: TimeFormat,
+}
```
didier-wenzek, line 153: With the vision I have for `tedge flows`, the configuration of the AWS mapper should be reduced to a combination of a bridge and a generic flow adding timestamps. The configuration of the latter will be externalized in a flow configuration file, i.e. no more under tedge config.

- Would this help mapper configuration?
- Would this remove the need for specialized config in ` MapperConfig<T: SpecialisedCloudConfig>`?

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1738 @@
...
+pub struct AzMapperSpecificConfig {}
+
+/// AWS IoT-specific mapper configuration fields
+#[derive(Debug, Deserialize, Document)]
+pub struct AwsMapperSpecificConfig {}
+
+/// CloudConfig implementation for C8y
+impl SpecialisedCloudConfig for C8yMapperSpecificConfig {
+    type SpecialisedMapperConfig = EmptyMapperSpecific;
+}
```
didier-wenzek, line 570: I don't get the difference between `C8yMapperSpecificConfig` and `SpecialisedMapperConfig`?

Looking the following lines related to `Az` and `Aws`, I'm expecting here something like:

 
```suggestion
/// CloudConfig implementation for C8y
impl SpecialisedCloudConfig for C8yMapperSpecificConfig {
    type SpecialisedMapperConfig = C8yMapperSpecific;
}
```

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1738 @@
...
+    FileRead(#[from] std::io::Error),
+
+    /// Failed to parse TOML configuration
+    #[error("Failed to parse mapper configuration: {0}")]
+    TomlParse(#[from] toml::de::Error),
+
+    // TODO remove me
+    /// Required field is missing after applying defaults
+    #[error("Required field '{field}' is missing from mapper configuration")]
+    MissingField { field: &'static str },
```
didier-wenzek: This case is indeed not used.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/compat.rs
```diff
@@ -0,0 +1,700 @@
...
+            MapperConfigError::ConfigRead(format!("C8y profile '{}' not found", profile.unwrap()))
+        })?;
+
+        build_mapper_config(c8y_config, tedge_config, profile)
+    }
+
+    fn from_cloud_config(c8y: &Self::CloudConfigReader, profile: Option<&str>) -> Self {
+        C8yMapperSpecificConfig {
+            auth_method: c8y.auth_method,
+            credentials_path: c8y.credentials_path.clone(),
```
didier-wenzek, line 59: Looking this code, I feel that it would be simpler the keep the `c8y` prefix in `/etc/tedge/mappers/c8y.toml` even if this will lead to a strange TOML file with a single `c8y` root.

crates/core/tedge/src/cli/config/cli.rs
jarhodes314: The way `tedge config set` actually updates the file is by using the `try_update_str` method generated by the macro, and this updates the DTO which can then be serialised back to `tedge.toml`. We don't currently have this method for the mapper config, nor the DTO structure. In theory we could mitigate this by writing values directly to a `toml::Table`, however for this we would need to know the type to parse the user's input to before writing it to a file.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
jarhodes314: I've now added a comment to explain this, and properly updated `read_trust_store` to use async methods for reading the file. I've also added a comment explaining that this is simply intended to produce a `Stream` so we can manipulate it like one (in this case, to read the different mapper configs, then read the trust stores for each, then chain the results together).

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: Yeah, I need to improve the naming here. The `MapperSpecific` stuff relates to the specialisation of the `mapper` field within the mapper toml config (so `mapper.timestamp_format` can exist for `az`/`aws` but not `c8y`, but `mapper.mqtt.max_payload_size` exists for all three). At the moment it's confusing distinguishing between "mapper config" (i.e. the toml file) and the `mapper` field within that toml file

didier-wenzek: > so `mapper.timestamp_format` can exist for `az`/`aws` but not `c8y`, but `mapper.mqtt.max_payload_size` exists for all three

If this help to simplify an already too complicated piece of code, we don't have to be fully compatible with the monolithic configuration file. We can decide to make things more uniform by adding a `timestamp_format` to the c8y mapper config even if not actually used.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/compat.rs
jarhodes314: I'm not sure that would necessarily help, however I think we might be able to reuse the existing DTO logic rather than separate deserialisation logic for the non tedge.toml use-case.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/compat.rs
```diff
@@ -0,0 +1,700 @@
...
+        },
+        cloud_specific,
+    })
+}
+
+/// Trait to abstract over different cloud config readers
+///
+/// This trait provides a uniform interface for accessing common fields
+/// from different cloud configuration readers (C8y, Az, Aws).
+trait CloudConfigAccessor {
```
didier-wenzek, line 220: The fact, we need to introduce this trait to make the new config compatible with the former,  is a sign we missed something in the legacy config definition. We have too many duplicated definitions there. I see this PR as an opportunity to be in a better position to fix that.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1733 @@
...
+    /// Set of MQTT topics the mapper should subscribe to
+    pub topics: TemplatesSet,
+
+    /// Bridge configuration
+    pub bridge: BridgeConfig,
+
+    pub mapper: MapperMapperConfig<T::SpecialisedMapperConfig>,
+
+    /// Cloud-specific configuration fields (flattened into the same level)
+    pub cloud_specific: T,
```
albinsuresh, line 136: This separation of `SpecialisedCloudConfig` in the `cloud_specific` field and its associated `SpecialisedMapperConfig` into a seperate `mapper` fields feels like a source of a lot of indirection to access cloud specific fields. I understand that the intent was to separate the common fields from cloud-specific fields. But I found it really difficult to follow the hierarchy.

I'm just wondering if introducing a `MapperConfigAccessor` trait (similar to the `CloudConfigAccessor` trait) to expose these common configs like `url`, `device`, `mapper` etc and use these trait functions to access these configs rather that direct struct field access And then have structs like `C8yMapperSpecificConfig`, `AzMapperSpecificConfig` etc to implement this trait. For serde, we can have a `CommonMapperConfig` struct that captures all these common fields and included in the cloud specific structs as a flattened field.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1733 @@
...
+pub struct AwsMapperSpecificConfig {}
+
+/// CloudConfig implementation for C8y
+impl SpecialisedCloudConfig for C8yMapperSpecificConfig {
+    type SpecialisedMapperConfig = EmptyMapperSpecific;
+}
+
+/// CloudConfig implementation for Azure
+impl SpecialisedCloudConfig for AzMapperSpecificConfig {
+    type SpecialisedMapperConfig = AzMapperSpecific;
```
albinsuresh, line 574: I really didn't get the benefit of this indirection, exposing the `mapper` fields via this `SpecialisedCloudConfig`. Why not define those `mapper` fields directly under the `AzMapperSpecificConfig` struct, just like other cloud specific fields?

TBH, I even feel that this conceptual separation between `CloudSpecificConfig` and `CloudSpecificMapperConfig` ended up creating too many abstractions complicating these structs. From my POV as a user, the latter is also just yet another cloud-specific config.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1733 @@
...
+            timestamp: default_timestamp(),
+            timestamp_format: default_timestamp_format(),
+        }
+    }
+}
+
+/// Cumulocity-specific mapper configuration fields
+#[derive(Debug, Deserialize, Document)]
+#[serde(default)]
+pub struct C8yMapperSpecificConfig {
```
albinsuresh, line 497: It'd be better if these mapper specific config structs and their trait impls could be moved to dedicated modules for better isolation, keeping just the common abstractions here.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1733 @@
...
+impl<T: Default> Default for PartialMapperMapperConfig<T> {
+    fn default() -> Self {
+        Self {
+            mqtt: PartialMqttConfig::default(),
+            cloud_specific: T::default(),
+        }
+    }
+}
+
+fn set_key_if_blank<T>(field: &mut OptionalConfig<T>, value: Cow<'static, str>) {
```
albinsuresh, line 1155:
```suggestion
fn set_optional_key_if_blank<T>(field: &mut OptionalConfig<T>, value: Cow<'static, str>) {
```

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -0,0 +1,1733 @@
...
+    use OptionalConfig as OC;
+    match field {
+        OC::Present { ref mut key, .. } | OC::Empty(ref mut key) if key.is_empty() => *key = value,
+        _ => (),
+    }
+}
+
+fn set_key_if_blank2<T>(field: &mut Keyed<T>, value: Cow<'static, str>) {
```
albinsuresh, line 1163:
```suggestion
fn set_key_if_blank<T>(field: &mut Keyed<T>, value: Cow<'static, str>) {
```
To be consistent with the previous naming suggestion.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: The `cloud_specific` field will be reduced to the single field at the top-level of mapper config with the upcoming changes to use `TEdgeConfigDto`s to deserialize, we will no longer need the `MapperMapperConfig` specialisation, nor the `PartialMapperConfig` structs that we currently have. It will also move `bridge.include.local_cleansession` back under the c8y-specific fields.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: As noted above, this will be going away in the follow-up PR (and already has on my local copy of the code). The fundamental issue being solved here is that `#[serde(flatten)]` is only aware of what's going on at the current level of a struct, so if we consider:

```rust
#[derive(Deserialize)]
struct Example {
    mapper: ExampleMapper,
    #[serde(flatten)]
    other: OtherFields
}

#[derive(Deserialize)]
struct OtherFields {
    mapper: OtherFieldsSpecificMapper,
    something_else_cloud_specific: String,
}

#[derive(Deserialize)]
struct ExampleMapper {
    all_clouds_have_this: String,
}

#[derive(Deserialize)]
struct OtherFieldsSpecificMapper {
    something_cloud_specific: String,
}
```

In this case, `mapper.all_clouds_have_this` will be deserialised properly, but `mapper.something_cloud_specific` (which is via `OtherFields`) will never be deserialised as serde will attempt to deserialise `something_cloud_specific` as part of the `ExampleMapper` struct, and will just discard the field when it fails. `something_else_cloud_specific` will get deserialised just fine as that isn't a field in `Example` though, so `serde` will deserialise it as part of `OtherFields`.

What this means for us is where we have configuration groups like `mapper` and `bridge` that contain some configurations shared between clouds, as well as some cloud-specific configurations, we have to deserialise these cloud-specific configurations separately to the other, entirely-cloud-specific groups (like `smartrest`).

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: I will look into this in the follow-up PR.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: This is a better name, however this function will be deleted as part of the follow-up, so I'll leave this as-is for this PR.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: This is a better name, however this function will be deleted as part of the follow-up, so I'll leave this as-is for this PR.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
albinsuresh: What if the flattening was done the other-way around, where the shared fields are flattened, like this:

```
    #[derive(Deserialize, Debug)]
    struct ExampleCloud {
        #[serde(flatten)]
        common_fields: ExampleMapper,
        cloud_specific: String,
        other_cloud_specific: OtherFields,
    }

    #[derive(Deserialize, Debug)]
    struct ExampleMapper {
        all_clouds_have_this: String,
    }

    #[derive(Deserialize, Debug)]
    struct OtherFields {
        mapper: OtherFieldsSpecificMapper,
        something_else_cloud_specific: String,
    }

    #[derive(Deserialize, Debug)]
    struct OtherFieldsSpecificMapper {
        something_cloud_specific: String,
    }
```

I could parse a TOML config as follows with the above structure:

```toml
all_clouds_have_this = "common value"
cloud_specific = "cloud specific value"

[other_cloud_specific]
something_else_cloud_specific = "other cloud specific value"

[other_cloud_specific.mapper]
something_cloud_specific = "mapper specific value"
```

Am I missing something here?

jarhodes314: You haven't got any overlap in the field names in your example. They key thing is that whatever struct you flatten needs to contain a key of the same name as a key in the struct you flatten into.

If we changed the flattening to handle the common fields, in my example, you would end up with the opposite problem, where the global `mapper` config values wouldn't be deserialised by any clouds with a specialised `mapper` key.

jarhodes314: You haven't got any overlap in the field names in your example. They key thing is that whatever struct you flatten needs to contain a key of the same name as a key in the struct you flatten into.

If we changed the flattening to handle the common fields, in my example, you would end up with the opposite problem, where the global `mapper` config values wouldn't be deserialised by any clouds with a specialised `mapper` key.

jarhodes314: You haven't got any overlap in the field names in your example. They key thing is that whatever struct you flatten needs to contain a key of the same name as a key in the struct you flatten into.

If we changed the flattening to handle the common fields, in my example, you would end up with the opposite problem, where the global `mapper` config values wouldn't be deserialised by any clouds with a specialised `mapper` key.

albinsuresh: > You haven't got any overlap in the field names in your example. They key thing is that whatever struct you flatten needs to contain a key of the same name as a key in the struct you flatten into.

That's exactly why I felt that reversing the flattening would solve that problem as the name of the field that's flattened (`mapper: ExampleMapper`) doesn't matter, right? Just to confirm, I tried that as well:

```rust
struct ExampleCloud {
    #[serde(flatten)]
    mapper: ExampleMapper,
    other: OtherFields
}

#[derive(Deserialize, Debug)]
struct OtherFields {
    mapper: OtherFieldsSpecificMapper,
    
}
```

... and that also worked. I could parse the following TOML:

```toml
            all_clouds_have_this = "common value"
            cloud_specific = "cloud specific value"

            [other_cloud_specific]
            something_else_cloud_specific = "other cloud specific value"

            [other_cloud_specific.mapper]
            something_cloud_specific = "mapper specific value"

```

into:

```
ExampleCloud {
    mapper: ExampleMapper {
        all_clouds_have_this: "common value",
    },
    cloud_specific: "cloud specific value",
    other_cloud_specific: OtherFields {
        mapper: OtherFieldsSpecificMapper {
            something_cloud_specific: "mapper specific value",
        },
        something_else_cloud_specific: "other cloud specific value",
    },
}
```

There would have been a problem if the `ExampleMapper` struct had a field that is named `other` or any other top-level fields in `ExampleCloud`. But we don't have that problem, do we?

jarhodes314: The issue is that we need the `mapper` field to be exposed in the resulting configuration, so that `az.mapper.timestamp` becomes `mapper.timestamp` in the new configuration file. But we also have `mapper.max_payload_size` in the same configuration file and that is a generic configuration.

```rust
#[derive(Deserialize)]
struct AzConfiguration {
    #[serde(flatten)]
    generic: GenericFields,
    mapper: AzMapper,
}

#[derive(Deserialize)]
struct GenericFields {
    mapper: MapperFields,
}

#[derive(Deserialize)]
struct MapperFields {
    max_payload_size: u32,
}

#[derive(Deserialize)]
struct AzMapper {
    timestamp: bool,
}
```

With this, we end up in the opposite scenario, we can't deserialise `GenericFields::mapper`.

Runnable example which panics for this reason: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d85f970d3c08bf4f9313c3067a736b92

albinsuresh: > The issue is that we need the mapper field to be exposed in the resulting configuration, so that az.mapper.timestamp becomes mapper.timestamp in the new configuration file. But we also have mapper.max_payload_size in the same configuration file and that is a generic configuration.

Okay. Now the issue is clear. And to simplify this problem we can definitely consider adding the `timestamp` field also into the generic config set as @didier-wenzek suggested [here](https://github.com/thin-edge/thin-edge.io/pull/3851#discussion_r2585230049). It does make sense for the c8y mapper as well, now that it can map free form MQTT messages with the MQTT service support.

PR #3854, opened 2025-11-07
tests: allow using a custom Cumulocity mqtt endpoint in tests

reubenmiller:
## Proposed changes

**Goals**
* Allow to run all tests against custom Cumulocity endpoints (decouple the explicit configuration of the Cumulocity endpoints)

Extend the system test framework to allow to run all of the tests using a custom Cumulocity mqtt endpoint to make it easier to check the overall compatibility with the tests.

The test framework will now set the appropriate tedge c8y settings based upon the presence of a custom mqtt endpoint or not. Below shows the relationship of the environment variables and the resulting tedge settings:

```sh
C8Y_BASEURL=example.cumulocity.com

# the above will result in setting the following
tedge config set c8y.url example.cumulocity.com
```

Where as,

```sh
C8Y_BASEURL=example.cumulocity.com
C8Y_MQTT=example.cumulocity.com:9883

# the above will result in setting the following
tedge config set c8y.http example.cumulocity.com
tedge config set c8y.mqtt example.cumulocity.com:9883
```

This means that users can configure all system tests to run against the MQTT service by setting the following environment variable (typically set via the .env file):

```sh
C8Y_MQTT=example.com:9883
```

This is to enable testing against the newer Cumulocity mqtt service to ensure compatibility with the Cumulocity Core MQTT endpoints. It should have no impact on existing tests which use the default setup.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/devdata/env.template
+++ b/tests/RobotFramework/devdata/env.template
@@ -19,6 +19,8 @@ PYTHON_KEYRING_BACKEND=keyring.backends.null.Keyring
 C8Y_BASEURL=
 C8Y_USER=
 C8Y_PASSWORD=
+# Custom MQTT endpoint, e.g. example.cumulocity:9883
+#C8Y_MQTT=
 
 #
 # AWS IoT


tests/RobotFramework/tests/tedge_connect/tedge_connect_profile_validation.robot
```diff
@@ -21,7 +21,10 @@ Bridge topic prefix must be unique to test connection
 
 Device ID must be unique if cloud URLs are the same
     ThinEdgeIO.Execute Command
-    ...    sudo tedge config set c8y.url --profile second "$(tedge config get c8y.url)"
+    ...    sudo tedge config set c8y.mqtt --profile second "$(tedge config get c8y.mqtt)"
+    ...    timeout=0
+    ThinEdgeIO.Execute Command
+    ...    sudo tedge config set c8y.http --profile second "$(tedge config get c8y.http)"
```
didier-wenzek, line 27: As the test is only checking that connecting the new profile fails, it's not easy to be confident that this change is actually required - although it makes sense.

reubenmiller: Yeah I guess if it is checking that the connection doesn't work then you're right...so in the end I guess it is more of a consistency thing...though I'm running into more and more places where checking if c8y.url or c8y.http and c8y.mqtt is proving to be a real pain (e.g. when writing scripts to cater for both scenarios).

Issue #3850, opened 2025-11-05
refactor tedge flows api


reubenmiller:
**Is your refactoring request related to a problem? Please describe.**

Refactor the tedge flows interface and data types.

**Describe the solution you'd like**

The following items should be refactored to improve the overall tedge flows concept:

### Item 1: Remove onConfigUpdate function

Remove the `onConfigUpdate()` function from the flow concept. Whilst the underlying functionality is useful, it is not a common case yet, and we might be able to solve it in a different way in the future. Removing this interface reducing the size of the public API for the initial launch of the feature.

###  Item 2: Replace Timestamp type with the JavaScript Date object - Done

Replace the `Timestamp` type with the JavaScript native `Date` object.

The type should be replaced in the following areas:

* In the onInterval function, the first argument
* In the `message.timestamp`, and rename the property to `message.time`

Below shows an example of the function signatures with the change in the timestamp data type.

```js
export function onMessage(message: Message, config: any) {
    console.log(`Message received at: ${message.time}`);
    return [];
}

export function onInterval(time: Date, config: any) {
    return [];
}
```

### Item 3: Replace config with context in functions

Change the function signatures of `onMessage` and `onInterval` to rename the `config` argument to `context`, however `context` is an object which should include a `config` property (which contains the step's configuration). Below shows an example of the context object and its relationship to the "config" property.

**Example Context object**

```js
{
    config: {
        foo: "bar"
    }
}
```

Below shows an example of the renamed function signatures which use the new "context" type.

```sh
export function onMessage(message: Message, context: any) {
    console.log(`Message has context: ${context.config}`);
    return [];
}

export function onInterval(time: Date, context: any) {
    return [];
}
```

### Item 4: Rename payload_raw to payload in the message data object

In the `message` type, rename `payload_raw` to `payload` (overwriting the previous string type used in `payload`). This will prevent duplicating the data being passed to the flows. JavaScript helper functions can be provided in the meantime to make it easier to encode and decode payloads to and from the JavaScript `Uint8Array` type.

```js
export interface Message {
  time?: Date; // Previously a custom Timestamp object, but now a JavaScript Date
  topic: string;
  payload?: Uint8Array; // Previously was a string, and was duplicated in payload_raw
  retain: boolean;
}
```

**Describe alternatives you've considered**

**Additional context**

PR #3863, opened 2025-11-14
feat: Use JS Date for tedge flows timestamps

didier-wenzek:
## Proposed changes

- [x] Use [JS Date objects](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) for tedge flow timestamps
- [x] Add a `--processing-time` option to `tedge flows test`
- [x] Deprecate adhoc `DateTime` in favor of std `SystemTime`
- [x] Rename `Message.timestamp` to `Message.time`

## Types of changes

## Paste Link to the issue

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/js_script.rs
+++ b/crates/extensions/tedge_flows/src/js_script.rs
@@ -319,7 +319,7 @@ export async function onMessage(message, config) {
     async fn using_date() {
         let js = r#"
 export function onMessage(message, config) {
-    let time = message.timestamp;
+    let time = message.time;
     return {
         "topic": message.topic,
         "payload": JSON.stringify({
--- a/crates/extensions/tedge_flows/src/js_value.rs
+++ b/crates/extensions/tedge_flows/src/js_value.rs
@@ -61,7 +61,7 @@ impl From<Message> for JsonValue {
             ("topic", JsonValue::string(value.topic)),
             ("payload", payload),
             ("raw_payload", raw_payload),
-            ("timestamp", JsonValue::option(value.timestamp)),
+            ("time", JsonValue::option(value.timestamp)),
         ])
     }
 }
--- a/docs/src/references/mappers/flows.md
+++ b/docs/src/references/mappers/flows.md
@@ -64,14 +64,14 @@ interface FlowStep {
     onMessage(message: Message, config: object): null | Message | Message[],
   
     // called at regular intervals to produce aggregated messages
-    onInterval(timestamp: Timestamp, config: object): null | Message | Message[],
+    onInterval(time: Date, config: object): null | Message | Message[],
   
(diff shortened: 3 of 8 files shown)


crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -314,6 +314,38 @@ export async function onMessage(message, config) {
...
+        let (runtime, script) = runtime_with(js).await;
+
+        let datetime = DateTime {
+            seconds: 1763050414,
+            nanoseconds: 0,
+        };
+        let input = Message::new("clock", "");
+        let output = Message::new(
+            "clock",
+            r#"{"milliseconds":1763050414000,"date":"Thu Nov 13 2025 17:13:34 GMT+0100"}"#
```
jarhodes314: It looks like this is failing in CI due to the build server being in UTC and not CET.

didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3863/commits/3a687f04f6f3e37cc645c39f4df0250226419395

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -113,17 +112,11 @@ impl FlowResult {
...
-    pub seconds: u64,
-    pub nanoseconds: u32,
-}
-
 #[derive(Clone, serde::Deserialize, serde::Serialize, Eq, PartialEq)]
 pub struct Message {
     pub topic: String,
     pub payload: Vec<u8>,
-    pub timestamp: Option<DateTime>,
+    pub timestamp: Option<SystemTime>,
```
albinsuresh, line 119: Why did we choose `SystemTime` over `OffsetDateTime` when we widely use the latter in the rest of the tedge code?

didier-wenzek: Mostly because we need nothing related to date formatting here.

albinsuresh: Okay. I believe the lack of timezone information in `SystemTime` is okay in this case as both the rust runtime and the js engine are on the same device, and hence can be aligned with the same device timezone.

reubenmiller (approved):
Approved. Works nicely and should be at least more convenient for users to convert the timestamp to different formats using standard javascript Date functions.

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: Here is test aimed to remove any timezone concern: https://github.com/thin-edge/thin-edge.io/pull/3863/commits/c4aa07f5cd37376b8ef51eaa570d8e18d72572de

PR #3869, opened 2025-11-20
feat: tedge flows context

didier-wenzek:
## Proposed changes

- [x] Deprecate flow script `onConfigUpdate` method
- [x] Add a context argument to `onMessage` and `onInterval` to give the scripts access to the shared data
- [x] `context.mapper` gives the scripts a read access to a store shared by all the flows
- [x] `context.mapper` is populated by regular flows which output is configured to be the context.
- [x] `context.flow` is shared by all the scripts of a flow
- [x] `context.script` is private to a script instance
- [x] `context.config` is the config provided by the flow to a script instance
- [x] the keys of a context can be listed with the `keys` method.
- [x] `context` can be updated using `get`, `set`, `remove` methods.
 
## Types of changes

## Paste Link to the issue

## Checklist

### DIFF
--- a/docs/src/references/mappers/flows.md
+++ b/docs/src/references/mappers/flows.md
@@ -38,13 +38,20 @@ which rule how to consume, transform and produce MQTT messages.
   - Builtin steps provide generic building blocks such as %%te%% JSON translation into Cumulocity JSON.
   - Users can implement specific steps using JavaScript or TypeScript to refine transformations to their use cases. 
 - If some message transformations can be fully defined only from the input message, most require a *context*.
-  - What is the Cumulocity internal id of the device? What are the units used by a sensor? Does the location of the device matter?
-  - Such a context can only be specific and has to be built from various sources, configuration, metadata and capability messages.  
-  - For that purpose, %%te%% maintain a context object which is
-    - created, cached and populated by %%te%% using configuration data,
+  - What is the Cumulocity internal id of the device? What are the units used by a sensor? What is the location of the device?
+  - For that purpose, %%te%% maintain a `context` object which is
     - passed to all invocations of transformation steps,
-    - enriched by some flows and steps with context info extracted from metadata and capability messages,
+    - structured along 3 namespaces
+      - `context.mapper` is shared by all the flows of a mapper
+      - `context.flow` is private to the flow, shared by all the steps of that flow
+      - `context.script` is private to a script instance and persisted across script reloads
+    - created, cached and populated by %%te%% using configuration data,
+    - possibly enriched by the flows with data extracted from metadata and capability messages,
     - used by all flows and steps to adapt their behavior
+  - The `context` object also provides a `context.config` object
+    - which content is specific to each script instance and provided by flows configuration
+    - to set values such as thresholds, ranges, durations, units or endpoints
+    - giving the users a way to control the behavior of a step without rewriting scripts.
 - %%te%% provides some support to steps aggregating messages over time windows.
   - For each aggregating step, the mapper persists a state (a JSON object)
     which can be updated by the step function on each message and at regular intervals
@@ -60,10 +67,10 @@ A transformation *script* is a JavaScript or TypeScript module that exports:
 ```ts
 interface FlowStep {
     // transform one input message into zero, one or more output messages
-    onMessage(message: Message, config: object): null | Message | Message[],
+    onMessage(message: Message, context: Context): null | Message | Message[],
   
     // called at regular intervals to produce aggregated messages
-    onInterval(time: Date, config: object): null | Message | Message[]
+    onInterval(time: Date, context: Context): null | Message | Message[]
 }
 ```
 
@@ -79,17 +86,49 @@ type Message = {
 }
 ```
 
-A `config` is an object freely defined by the step module, to provide default values such as thresholds, durations or units.
-These values are configured by the flow and can be dynamically updated on reception of config update messages.
+The `context` object passed to `onMessage()` and `onInterval()` gives scripts and flows a way to share data.
+
+```ts
+type Context = {
+  // A set of (key, value) pairs shared by all the scripts of a mapper
+  mapper: KVStore,
+  
+  // A set of (key, value) pairs shared by all the scripts of a flow
+  flow: KVStore,
+  
+  // A set of (key, value) pairs private to a script, persisted across module reloads
+  script: KVStore,
+  
+  // A value provided by the flow configuration of that step
+  config: any,
+}
+
+type KVStore = {
+  // List the keys for which this store holds a value  
+  keys(): string[]
+  
+  // Get the value attached to a key (returning null, if none)
+  get(key: string): any,
+  
+  // Set the value attached to a key (removing the key if the provided value is null)
+  set(key: string, value: any),
+  
+  // Remove any value attache to a key
+  remove(key: string),
+}
+```
+
+The `context.config` is an object freely defined by the step module, to provide default values such as thresholds, durations or units.
 
 The `onMessage` function is called for each message to be transformed
   - The arguments passed to the function are:
     - The message `{ topic: string, payload: string, raw_payload: Uint8Array, time: Date }`
-    - The config as read from the flow config or updated by the script
+    - A context object with the config and state
   - The function is expected to return zero, one or many transformed messages `[{ topic: string, payload: string }]`
   - An exception can be thrown if the input message cannot be transformed.
-- A flow script can also export a `onInterval` function
-  - This function is called at a regular pace with the current time and config.
+
+A flow script can also export a `onInterval` function
+  - This function is called at a regular pace with the current time and context.
   - The flow script can then return zero, one or many transformed messages
   - By sharing an internal state between the `onMessage` and `onInterval` functions,
     the flow script can implement aggregations over a time window.
@@ -189,6 +228,15 @@ accept_topics = "c8y/#"
 
 [errors.file]
 path = "/var/run/tedge/flows.log"
+```
+
+The output of flow can also be directed to the global context aka `context.mapper`.
+The main usage is for a flow to store and share metadata received as retained MQTT messages,
+so this metadata can be used by other transformation flows;
+the canonical example being a context flow populating the context with measurement units to be used by a measurement publisher flow.
+
+```toml
+[output.context]
 ```
 
 ## %%te%% flow mapper


tests/RobotFramework/tests/tedge_flows/flows/load_metadata.toml
```diff
@@ -0,0 +1,7 @@
+#
```
albinsuresh, line 1: This way of feeding the `context` with data using a standalone flow is just a temporary mechanism, right? I'm assuming that in the final API, the relevant "meta inputs" for a flow would be specified directly in the flow definition itself?

crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -35,10 +37,12 @@ impl JsRuntime {
             })))
             .await;
         let context = rquickjs::AsyncContext::full(&runtime).await?;
-        let worker = JsWorker::spawn(context).await;
+        let store = KVStore::default();
```
albinsuresh, line 40: In addition to this global store (shared across all the flows), is there any plans to provide a single flow-scoped store as well? To share any runtime data between scripts in a flow?

crates/extensions/tedge_flows/src/js_lib/kv_store.rs
```diff
@@ -0,0 +1,75 @@
...
+
+#[derive(Clone, Default, JsLifetime)]
+pub struct KVStore {
+    data: Arc<Mutex<BTreeMap<String, JsonValue>>>,
+}
+
+impl KVStore {
+    pub fn init(&self, ctx: &Ctx<'_>) {
+        let globals = ctx.globals();
+        let _ = Class::<FlowStore>::define(&globals);
```
albinsuresh: Minor: While reading the code, this would look like an odd coupling from `KVStore` to `FlowStore` when we expect it to be only the other way around. Though, logically it doesn't make any difference, as this is an `init` function called only once, I feel this code would have been better placed in `JsWorker::run()` itself.

tests/RobotFramework/tests/tedge_flows/flows/load_metadata.toml
didier-wenzek: > This way of feeding the `context` with data using a standalone flow is just a temporary mechanism, right?

No, on the contrary, this is the core of the proposal.

> I'm assuming that in the final API, the relevant "meta inputs" for a flow would be specified directly in the flow definition itself?

The idea is that the context provides a store which content is shared by *all* the flows. Any flow can use entity metadata for instance. One can add a flow configuration to tell what is relevant for that flow, but this would be only informational: the data being feed in a more general manner.

It seems that what you have in mind is similar to the mechanism of `onConfigUpdate()` we decided to deprecate because begin too specific to a flow step. Each step having to provide its own logic to extract, store and use metadata. The proposal is to separate these 3 aspects:

- Context specific flows *extract* contextual data from MQTT retain messages or other configuration sources
- The engine *stores* contextual data and makes this context available to all flows
- Any flow has a *read access* to  contextual data.

didier-wenzek: One point that is confusing with the current state of the PR is the following excerpt from the `te_to_c8y.js` example. How a fresh new store can ever contain some data? 

```
  let store = new FlowStore()
  let meta = store.get(`${message.topic}/meta`) || {}
```

That one is  just a temporary mechanism.

The plan is to pass a store to the script via a `context` parameter:

```
  let store = context.store.
  let meta = store.get(`${message.topic}/meta`) || {}
}
```

That said the content of that store will have to be documented.

albinsuresh: > The idea is that the context provides a store which content is shared by all the flows. Any flow can use entity metadata for instance.

Since the flow context currently is global, I'd agree that having this notion of global flows building the context for all flows make sense.

But, from the user perspective, it felt slightly convoluted when I wanted to add a new flow. For example, when I'm going to add a new flow for `temperature` measurement processing, which has dependency on `temperature/meta` topics as well, I'm forced to write two different flows: one for the data processing itself and the new one for metadata input gathering. As a user I was wondering why not specify the `meta_inputs` at the flow definition level itself so that it feels contextual to the processing logic itself defined in the same flow.

> It seems that what you have in mind is similar to the mechanism of onConfigUpdate() we decided to deprecate because begin too specific to a flow step. Each step having to provide its own logic to extract, store and use metadata.

Taking that  `extract` and `store` responsibility from the user was a good thing. That really wasn't my concern. My only concern was about splitting the "runtime inputs" and "meta inputs" into two different flow definitions, which I feel is less intuitive. I'd have preferred `meta_input.mqtt.topics` config defined in the message processing flow definition itself as follows:

```
input.mqtt.topics = ["te/+/+/+/+/m/temp"]
meta_input.mqtt.topics = ["te/+/+/+/+/m/temp/meta"]

steps = [
    { script = "add_timestamp.js" },
    { script = "te_to_c8y.js", meta_topics = ["te/+/+/+/+/m/+/meta"] }
]
```

The mapper, while loading the flows, can then use these `meta_input.mqtt.topics` to build that context for the js_runtime.

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
```diff
@@ -51,7 +51,8 @@ export function onMessage(message, config) {
     type: type
   }
 
-  let meta = config[`${message.topic}/meta`] || {}
+  let store = new FlowStore()
+  let meta = store.get(`${message.topic}/meta`) || {}
```
albinsuresh: That `|| {}` would mask a possible race condition between this flow and the `load_metadata.toml` flow, right? Since both the flows could be loaded independently, this required context may not have been populated by the other flow, before this step is executed by this flow as it could theoretically receive the measurement message first, before the other flow receives and processes the metadata message for the same.

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
```diff
@@ -90,40 +91,3 @@ export function onMessage(message, config) {
...
-///   "type": "example",
-///   "temperature": {
-///     "temperature": {
-///       "value": 23,
-///       "unit": "°C"
-///     }
-///   }
-/// }
-/// ```
-export function onConfigUpdate(message, config) {
```
albinsuresh, line 123: Though I'm fully aligned with `tedge` taking over the `extract` and `store` responsibilities of the `meta` messages, one additional "feature" that we've lost in the process is the flows awareness of any dynamic meta changes. For example, if there was a tick based flow doing some processing with an `interval` of 1h, if the meta definition changed after 30 mins, the flow should have been aware of this change. The flow can choose whether to react to that change or not for the current batch of messages. But, at least getting notified would have been useful.

This is probably not a big deal in most fleets where the telemetry metadata is pretty much static.

crates/extensions/tedge_flows/src/js_runtime.rs
didier-wenzek: Having 3 context levels (device/flow/step) was my initial plan. However, do we really have a *concrete* needs for the intermediate level? It would be good to have examples before adding this extra level.

crates/extensions/tedge_flows/src/js_lib/kv_store.rs
didier-wenzek: Yeah, this line is likely subjected to changes and I would not bother too much.

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
didier-wenzek: > That `|| {}` would mask a possible race condition between this flow and the `load_metadata.toml` flow, right? 

As long as the flow mapper received the metadata first and the measurement second, then the mapper guarantees this line to be processed where the context metadata is available. But yes, a race can happen at the MQTT level - exactly as for any other mapper or agent logic.

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
didier-wenzek: Correct

tests/RobotFramework/tests/tedge_flows/flows/load_metadata.toml
didier-wenzek: > But, from the user perspective, it felt slightly convoluted when I wanted to add a new flow. For example, when I'm going to add a new flow for `temperature` measurement processing, which has dependency on `temperature/meta` topics as well, I'm forced to write two different flows: one for the data processing itself and the new one for metadata input gathering. As a user I was wondering why not specify the `meta_inputs` at the flow definition level itself so that it feels contextual to the processing logic itself defined in the same flow.

This was precisely the reasoning of the initial design with `meta_topics` per script and an `onConfigUpdate()` method. Note that if a user wants that this is still feasible with a script subscribing to both data and metadata as in  `input.mqtt.topics = ["te/+/+/+/+/m/temp", "te/+/+/+/+/m/temp/meta"]` and using a function akin to  `onConfigUpdate()` to process the metadata.

>  I'm forced to write two different flows: one for the data processing itself and the new one for metadata input gathering.

In practice, I think only one or two flows will handle metadata and populate the context for all the other flows with metadata about entities, MEAs and operations.

> The mapper, while loading the flows, can then use these `meta_input.mqtt.topics` to build that context for the js_runtime.

This could make sense for a context which scope is the flow.

reubenmiller: There are two use-cases for the store (which is available from the future context object). Only case 1 is currently addressed by this PR (which is fine).

### Case 1: Global state which is accessible to all flows

Each flow can access data from the entity stores. I don't think these flows will be changing very often as they are basically extending the flow engine to a degree by making it easy to inject global state into all flows that wish to use it...but these can be independently maintained and extended (via non-breaking data model changes).

In practice, we could provide the following context stores out of the box:

* telemetry meta information (i.e. measurement units)
* entity store data

However, users can extend this very easily to add their own stores which are populated from additional sources like databases, and then it is accessible to all flows without each flow having to declare its dependency. I would view the context dependency more as a "weak reference" to additional data rather than a hard dependency...it means the flow can be run without the dependency, however it might just influence whether it actually produces any data or not (but that is up to the flow's own logic).

### Case 2: A flow would like to persist its own state and access it later (non-shared state)

Different flows may need to store some additional state for instances where the flow engine will be restarted (or the flow itself will be reloaded). Currently each flow is individually instantiated in the runtime, so technically it can create a local variable within the module and store state there, however I believe this behaviour will be changing due to some technical limitations of the rquickjs integration which may result in having to completing re-instantiate each flow module when the flow code or configuration changes.

Example:

A flow is calculating an uptime of a service over a long period of time (e.g. 7 days)...it wants to calculate the uptime across device restarts. Since flows are running in a sandboxed environment, they aren't able to write to a filesystem directly, so they could use the provided context backend to access.

crates/extensions/tedge_flows/src/js_runtime.rs
reubenmiller: I've added a very high level example of a flow level state here, https://github.com/thin-edge/thin-edge.io/pull/3869#discussion_r2558530292

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
reubenmiller: Though wouldn't the race conditional only occur for the first message? Any message delivered in the future will then be processed after the store has been populated.

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
reubenmiller: Do we allow for multiple outputs? If so, then technically users could also publish the context to an MQTT topic as well, and the flow could subscribe to both the original flow messages, and the context updates (which are also published on some local MQTT topic)

tests/RobotFramework/tests/tedge_flows/flows/load_metadata.toml
albinsuresh: > In practice, I think only one or two flows will handle metadata and populate the context for all the other flows with metadata about entities, MEAs and operations.

If the user doesn't have to explicitly create the "meta input" flow separately, but just consume the built-in metadata maintained by tedge, then this is fine.

But, that raises another question. Since the metadata includes entity metadata, operations metadata and even MEA metadata, we'd end up maintaining another fatter "entity store" at the mapper level as well, in addition to the one already maintained by the agent. That won't be good.

reubenmiller: Ok, maybe we might not end up doing an entity store via this method, but still the idea to be able to expose additional context to multiple flows is useful

crates/extensions/tedge_flows/src/js_runtime.rs
albinsuresh: > However, do we really have a concrete needs for the intermediate level?

If the `device` context derived from MEA metadata for **all** telemetry types is maintained by the mapper itself, then no I don't really see a use-case for it. But, if the user needs to define the meta topics specifically required by his flow, I feel it'd be natural for him to define it in the `flow` context itself rather than externally.

Regarding `step` level context, I'm not sure as I've always thought that different steps in a flow would be using different aspects of the same metadata message. For example, if a measurement metadata message has `min`, `max` and `unit` one step might be filtering the measurements that are out of that `min-max` range and the later one adding the unit.

So, a `flow` context felt like a decent balance between a `device` context and `step` context.

tests/RobotFramework/tests/tedge_flows/flows/load_metadata.toml
didier-wenzek: > There are two use-cases for the store (which is available from the future context object). Only case 1 is currently addressed by this PR (which is fine).

Yeah the current state of this PR is focused on Case 1.

For case 2, the idea is to "Extend the context to hold an arbitrary state value with read / write access by the owner script."

> But, that raises another question. Since the metadata includes entity metadata, operations metadata and even MEA metadata, we'd end up maintaining another fatter "entity store" at the mapper level as well, in addition to the one already maintained by the agent. That won't be good.

The intent is not to implement another "entity store" but to give the flow scripts an access to this entity store. That said, I agree that this PR is lacking something. The Rust object backing the store passed to the script should not be created by the flow engine but passed to the flow engine by the mapper. The C8Y mapper will have then an opportunity to add its own rules.

tests/RobotFramework/tests/tedge_flows/flows/te_to_c8y.js
albinsuresh: > Though wouldn't the race conditional only occur for the first message? Any message delivered in the future will then be processed after the store has been populated.

As discussed offline, the race condition would happen not only when the flow is loaded for the very first time, but everytime the flow is reloaded as well, on mapper restarts. Even though the subscriptions made by these flows are persistent after the flow is loaded for the first time, the race between messages from two different topics can still happen whenever the MQTT connection is re-established. This is a fundamental issue with the MQTT protocol itself that we've had to workaround for the entity store, by caching telemetry messages for an entity before it is registered. The flow scripts will also have to take a similar approach by caching the data in the `flow` context.

tests/RobotFramework/tests/tedge_flows/flows/events.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+    let event = JSON.parse(message.payload)
+    let c8y_event = {
+        type: topic_parts[6] || "ThinEdgeEvent"
+    }
+    Object.assign(c8y_event, event)
+
+    let entity_id = `${topic_parts[1]}/${topic_parts[2]}/${topic_parts[3]}/${topic_parts[4]}`
+    if (entity_id != "device/main//") {
+        let entity = context.get(entity_id)
```
albinsuresh: Since the `context` object is likely to evolve in the future exposing more and more things beyond the "state", how do we plan on accommodating those additions in the future?

For example, to expose things like the entity store handle, the config or the additional flow-level state, how would they be exposed? Via additional functions like `context.get_entity_meadata()`, `context.config()` etc? Or they'll be added as sub-fields to this context object like `context.entity_store`, `context.config`, `context.timestamp` etc, where each sub-field can take different structural forms? If it's the former, the current API with just the `get()` function for now, looks good enough (though it might need a different name to convey that it is the global state that it is referring to). If it's the latter, for technical reasons around the handling of objects in different shapes, we'll have to adjust this API as well, like `context.device.get()` or something like that.

crates/extensions/tedge_flows/src/js_lib/kv_store.rs
didier-wenzek: > Yeah, this line is likely subjected to changes and I would not bother too much.

This line is now removed. A script can no more create a `new FlowStore()` and has to use the context provided to `onMessage()`. See https://github.com/thin-edge/thin-edge.io/pull/3869/commits/ea57579f8e2e8f02934d78338f6cff95e95a485e

tests/RobotFramework/tests/tedge_flows/flows/events.js
didier-wenzek: > Since the `context` object is likely to evolve in the future exposing more and more things beyond the "state", how do we plan on accommodating those additions in the future?

This is indeed a critical point. Technically, the current proposal with a single namespace is enough. After all thin-edge is already using such a namespace with the MQTT topics. In practice, we need more clearly separated namespaces: with context data provided by thin-edge, the mapper or the users. Stability and documentation also matter. For instance what is provided in the context of the c8y mapper must be clear.

> For example, to expose things like the entity store handle, the config or the additional flow-level state, how would they be exposed? Via additional functions like `context.get_entity_meadata()`, `context.config()` etc? Or they'll be added as sub-fields to this context object like `context.entity_store`, `context.config`, `context.timestamp` etc, where each sub-field can take different structural forms?

We can introduce context specific namespaces such as  `context.entity_metadata.get()` or `context.c8y.get()`.  This would help stability and documentation. The question is then to we want something extensible or not. Could the az mapper adds his own `context.az.get()` context?

> If it's the former, the current API with just the `get()` function for now, looks good enough (though it might need a different name to convey that it is the global state that it is referring to). If it's the latter, for technical reasons around the handling of objects in different shapes, we'll have to adjust this API as well, like `context.device.get()` or something like that.

The technical difference between `context.foo.get()` (a method call) and `context.foo` (a property) is that the latter has to be serialized in its entirety to be passed to `onMessage()`, while for the former only a handle is passed to the `onMessage()` and a context value is serialized only on demand.

It makes sense to pass the script `config` as a value: this config is only intended to that script and the script will in all cases use the config in its entirety. However, a script invocation needs access to very few items of the entity cache (e.g. the c8y external id of the measurement being processed). Hence an on-demand approach is more appropriate here.

albinsuresh: > The question is then to we want something extensible or not. Could the az mapper adds is one context.az.get() context?

You mean whether we want to expose any "cloud context" in generic manner via something like `context.cloud` or allow different mappers to choose that namespace like `context.c8y` or `context.az` etc? I don't really see a strong case for the latter, especially if the static namespaces are easier to implement from the technical point-of-view as well. Just having the ability to define more than one namespace like `context.device` or `context.xyz` in future is more than sufficient in my opinion.

> It makes sense to pass the script config as a value: this config is only intended to that script and the script will in all cases use the config in its entirety.

Exposing the `config` object as-is is completely fine. But, if we're introducing the notion of multiple namespaces under the context, exposing the config via `context.config.get("config_key")` would just enable reusing the same underlying model for the future state namespaces as well. But TBH, I don't really have any strong preferences here. Fine with `config` object being exposed directly as well.

didier-wenzek: > You mean whether we want to expose any "cloud context" in generic manner via something like `context.cloud` or allow different mappers to choose that namespace like `context.c8y` or `context.az` etc?

My question was more about the content than the name. A generic name like  `context.cloud`  can make sense. But a method such as `context.cloud.getExternalId()` is more cloud specific. This is why I prefer a context with a generic method `get()` over free form keys, possibly organized along different namespaces.

crates/extensions/tedge_flows/src/js_lib/kv_store.rs
```diff
@@ -0,0 +1,115 @@
...
+        ctx: &Ctx<'js>,
+        flow_name: &str,
+        script_name: &str,
+        config: &JsonValue,
+    ) -> Result<Value<'js>> {
+        let context = Object::new(ctx.clone())?;
+
+        context.set("mapper", FlowStore::new(MAPPER_NAMESPACE))?;
+        context.set("flow", FlowStore::new(flow_name))?;
+        context.set("script", FlowStore::new(script_name))?;
```
albinsuresh, line 37: What's the use-case you had in mind for `script` context? From a user POV, how is it different from a `static` state variable defined in a script?

I'm asking this based on the following interpretation of the context hierarchy:
* `mapper`: Shared across all flows
* `flow`: Shared between steps
* `script`: Shared with what?

tests/RobotFramework/tests/tedge_flows/flows/events.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+    let event = JSON.parse(message.payload)
+    let c8y_event = {
+        type: topic_parts[6] || "ThinEdgeEvent"
+    }
+    Object.assign(c8y_event, event)
+
+    let entity_id = `${topic_parts[1]}/${topic_parts[2]}/${topic_parts[3]}/${topic_parts[4]}`
+    if (entity_id != "device/main//") {
+        let entity = context.mapper.get(entity_id)
```
albinsuresh, line 18: While trying to access a non-exported context namespace like `context.abc.get()` from the script, it was a silent failure with no trace of that error in the log. But, testing the same broken script with the `tedge flows test` utility was giving the proper error message:

```
2025-11-28T06:29:00.114082517Z ERROR tedge::cli::flows::test: Error in /etc/tedge/flows/measurements.toml: No messages can be processed due to an incorrect setting: JS raised exception: Error: cannot read property 'get' of undefined
    at onMessage (/etc/tedge/flows/measurements.toml|0|/etc/tedge/flows/add_timestamp.js:2:11)
```

Was it not feasible to get the same error message during live message processing?

albinsuresh (commented):
Happy to approve once the doc is updated.

A side note, unrelated to this PR: When `console.log()` outputs are logged in the mapper log, it appears directly as follows:
```
Nov 28 06:37:49 f6641df9de15 tedge-mapper[530]: JavaScript.Console: "test log"
```
It would have been better if the context(`module_name`) of that log is also included in that log statement.

crates/extensions/tedge_flows/src/js_lib/kv_store.rs
didier-wenzek: > What's the use-case you had in mind for `script` context? From a user POV, how is it different from a `static` state variable defined in a script?

The plan is to persist the script context across restarts. Not done yet. Currently the state is only kept in memory and persists module reload.

> I'm asking this based on the following interpretation of the context hierarchy:
> * `mapper`: Shared across all flows
> * `flow`: Shared between steps
> * `script`: Shared with what?

* `script`: Persisted across reloads and restarts

tests/RobotFramework/tests/tedge_flows/flows/events.js
didier-wenzek: > no trace of that error in the log.

This is surprising. Both `tedge-mapper flows` and `tedge flows test` should log the same. I will check.

docs/src/references/mappers/flows.md
```diff
@@ -38,13 +38,20 @@ which rule how to consume, transform and produce MQTT messages.
...
-  - Such a context can only be specific and has to be built from various sources, configuration, metadata and capability messages.  
-  - For that purpose, %%te%% maintain a context object which is
-    - created, cached and populated by %%te%% using configuration data,
+  - What is the Cumulocity internal id of the device? What are the units used by a sensor? What is the location of the device?
+  - For that purpose, %%te%% maintain a `context` object which is
     - passed to all invocations of transformation steps,
-    - enriched by some flows and steps with context info extracted from metadata and capability messages,
+    - structured along 3 namespaces
+      - `context.mapper` is shared by all the flows of a mapper
+      - `context.flow` is shared by all the steps of a flow
```
```suggestion
      - `context.flow` is private to the flow, shared by all the steps of that flow
```

docs/src/references/mappers/flows.md
```diff
@@ -83,21 +86,48 @@ type Message = {
...
+  remove(key: string),
+}
+```
+
+The `context.config` is an object freely defined by the step module, to provide default values such as thresholds, durations or units.
 
 The `onMessage` function is called for each message to be transformed
   - The arguments passed to the function are:
     - The message `{ topic: string, payload: string, raw_payload: Uint8Array, time: Date }`
     - The config as read from the flow config or updated by the script
```
```suggestion
    - A context object with the config and state
```

docs/src/references/mappers/flows.md
```diff
@@ -83,21 +86,48 @@ type Message = {
...
   - An exception can be thrown if the input message cannot be transformed.
-- If defined and associated in the step config with `meta_topics`, the `onConfigUpdate` function is called on each message received on these `meta_topics`.
-  - The arguments are:
-    - The message to be interpreted as a config update `{ topic: string, payload: string }`
-    - The current config
-  - The returned value (an arbitrary JSON value) is then used as the new config for the flow script.
-- A flow script can also export a `onInterval` function
+
+A flow script can also export a `onInterval` function
   - This function is called at a regular pace with the current time and config.
```
```suggestion
  - This function is called at a regular pace with the current time and context.
```

docs/src/references/mappers/flows.md
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3869/commits/b4012f17378f1ef8e3240c7fc06ff575731a4f3c

docs/src/references/mappers/flows.md
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3869/commits/b4012f17378f1ef8e3240c7fc06ff575731a4f3c

docs/src/references/mappers/flows.md
didier-wenzek: https://github.com/thin-edge/thin-edge.io/pull/3869/commits/b4012f17378f1ef8e3240c7fc06ff575731a4f3c

tests/RobotFramework/tests/tedge_flows/flows/events.js
didier-wenzek: So there is indeed a difference between `tedge-mapper flows` and `tedge flows test`.

- `tedge flows test` ignores all the input, output and error configuration: all the messages are consumed from stdin and published to stdout or stderr.
-  `tedge-mapper flows`  uses the configured input, output and error. And the default is to send errors to `te/errors`.

One can improve flow error logging with an option to tee error messages to the mapper log.

reubenmiller (approved):
Approved

PR #3871, opened 2025-11-24
tests: tedge-p11-server add `Frame1` deserialization snapshot tests

Bravo555:
## Proposed changes

This PR extracts `Frame` and `Frame1` into separate modules and adds some tests to them, in particular for `Frame1` deserialisation snapshot tests are added to make sure changes to `Frame1` are not breaking changes.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge-p11-server/src/proxy/frame1.rs
+++ b/crates/extensions/tedge-p11-server/src/proxy/frame1.rs
@@ -34,3 +34,203 @@ pub enum Frame1 {
 /// An error that can be returned to the client by the server.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ProtocolError(pub String);
+
+/// Snapshot tests for `Frame1` Serialize/Deserialize implementations.
+///
+/// This is to make sure that recent versions of TedgeP11Client can work with older versions of TedgeP11Server and vice
+/// versa. The `input` bytes should not be updated when `Frame1` is updated: these are reference snapshots.
+#[cfg(test)]
+mod tests {
+    use crate::pkcs11::CreateKeyParams;
+    use crate::pkcs11::KeyTypeParams;
+    use crate::pkcs11::SigScheme;
+    use crate::service::SignatureAlgorithm;
+    use crate::service::SignatureScheme;
+    use crate::SecretString;
+
+    use super::*;
+
+    #[test]
+    fn test_deserialize_error() {
+        let input = vec![0, 1, 97];
+        let request: Frame1 = postcard::from_bytes(&input).unwrap();
+        assert_eq!(request, Frame1::Error(ProtocolError("a".to_string())));
+    }
+
+    #[test]
+    fn test_deserialize_choose_scheme_request() {
+        let input = vec![1, 0, 0, 0];
+        let request: Frame1 = postcard::from_bytes(&input).unwrap();
+        assert_eq!(
+            request,
+            Frame1::ChooseSchemeRequest(ChooseSchemeRequest {
+                offered: vec![],
+                uri: None,
+                pin: None,
+            })
+        );
+    }
+
+    #[test]
+    fn test_deserialize_choose_scheme_request_2() {
+        let input = vec![1, 1, 131, 8, 1, 3, 117, 114, 105, 1, 3, 112, 105, 110];
+        let request: Frame1 = postcard::from_bytes(&input).unwrap();
+        assert_eq!(
+            request,
+            Frame1::ChooseSchemeRequest(ChooseSchemeRequest {
+                offered: vec![SignatureScheme(
+                    rustls::SignatureScheme::ECDSA_NISTP256_SHA256,
+                )],
+                uri: Some("uri".to_string()),
+                pin: Some(SecretString::new("pin".to_string())),
+            })
+        );
+    }
+
+    #[test]
+    fn test_deserialize_sign_request() {
+        let input = vec![
+            2, 4, 100, 97, 116, 97, 1, 3, 107, 101, 121, 1, 4, 100, 97, 116, 97,
+        ];
+        let request: Frame1 = postcard::from_bytes(&input).unwrap();
+        assert_eq!(
+            request,
+            Frame1::SignRequest(SignRequest {
+                to_sign: b"data".to_vec(),
+                uri: Some("key".to_string()),
+                pin: Some(SecretString::new("data".to_string())),
+            })
+        );
+    }
+
+    #[test]
+    fn test_deserialize_choose_scheme_response() {
+        let input = vec![3, 1, 132, 16, 1];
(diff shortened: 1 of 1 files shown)


crates/extensions/tedge-p11-server/src/proxy/frame.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[test]
+    fn cant_deserialize_older_as_newer_when_new_fields() {
+        let f1 = F1::V1 { a: 42 };
+        let f1 = postcard::to_allocvec(&f1).unwrap();
+        let f2 = postcard::from_bytes::<F2>(&f1);
+
+        f2.expect_err("can't deserialize older type into newer type");
+    }
+
+    /// Shows that when a new field is added to a request, it can't be deserialized if sent as older type.
```
didier-wenzek: This comment doesn't match the test and is therefore more puzzling than helpful

crates/extensions/tedge-p11-server/src/proxy/frame.rs
```diff
@@ -0,0 +1,94 @@
...
+#[non_exhaustive]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum Frame {
+    Version1(Frame1),
+}
+
+/// Documents the properties of the serialization/deserialization format for purposes of adding new requests while
+/// maintaining compatibility.
+#[cfg(test)]
+mod tests {
```
didier-wenzek, line 25: The knowledge provided by these tests is definitely valuable. However, does it make sense to include them here?

Indeed, their purpose is more to assess the behavior of `postcard` when reader and writer are not using the same version of the struct to be de-serialized, than to check that tedge-p11 server and client can communicate despite running different versions.

What we want is that a new version of the tedge p11 client can still communicate with an older version of tedge p11 server. Sure the new features will not work (say creating a private key) but the oldest requests must still work (mainly signing a TLS request).

I appreciate that this is your goal. However I'm not sure that these tests clearly communicate the intent to the maintainer. What we need is clear documented rules for the maintainer:

- Can a request type be changed?
- What changes are okay?
- When a new frame type must be introduced. e.g. adding a case `Frame::VersionN(FrameU)` ?
- What are the rules adding this new case. e.g. adding new request type, overriding former request types?

crates/extensions/tedge-p11-server/src/proxy/frame.rs
Bravo555: My bad, comment above got accidentally duplicated.

fixed in 92ca02646c

crates/extensions/tedge-p11-server/src/proxy/frame.rs
Bravo555: > What we want is that a new version of the tedge p11 client can still communicate with an older version of tedge p11 server. Sure the new features will not work (say creating a private key) but the oldest requests must still work (mainly signing a TLS request).

To test full builds of TedgeP11Server and TedgeP11Client is a bit tricky in unit tests because we sort of want to use client and server from different commits.
But to effectively test different versions of server and client, I will be adding 2 things next:

- snapshot tests of serialize/deserialize implementations for the messages, to make sure requests/responses will be serialized to and deserialized from the same byte sequence. In this way we can verify the server and the client know and don't "forget" any messages in case some incompatible change gets introduced into the message types. This however doesn't guarantee the features will keep working, only that server and client can understand each other
- Robot Framework feature tests will be modified to test each feature with 2 tedge-p11-server versions: the initial version where the feature was introduced (i.e. for `tedge cert renew` that would be 1.6.1) and the current build version. This will ensure that recent version of tedge are compatible with the initial tedge-p11-server version that supports the feature, and every successive one (because they will be run for every new version). Though these tests won't verify if older tedge versions are compatible with recent tedge-p11-server versions (but it seems this is an unlikely scenario and forward compatibility is less of an issue than backwards compatibility in this case).

Still, I'd opt for keeping these tests below as IMO they provide some value.

> - Can a request type be changed?
> - What changes are okay?

Cases where a request can't be changed should be caught by the snapshot tests that I still have to push, instead of creating a new PR i can add them to this one if you feel this current PR doesn't provide much value without them.

> - When a new frame type must be introduced. e.g. adding a case Frame::VersionN(FrameU)

I'd say new frame version is required when:

- the serialization format is changed to something else
- if the semantics of some request type have to be changed such that it needs to be handled differently than before (i think it's unlikely given that we can just add a new request)
- if fields have to be removed or changed in a way that isn't just appending new fields (for example, if, instead of having a few requests that have `pin` parameter we wanted to change the structure such that `pin` is always sent, without having to add it to each and every request type)

So we can do a lot just adding new types of requests.

> - What are the rules adding this new case. e.g. adding new request type, overriding former request types?

Generally new request types can be added freely. When overriding former request types, I'm not entirely sure, but I think the rule is that new optional fields can be added to requests, but not to responses (deserializer expects explicit `1` for `Some` and `0` for `None`, so if older tedge-p11-server doesn't send the field, that's a parse failure).

(EDIT: When new optional fields are added to requests, that's breaking the forward compatibility - recent tedge-p11-server will not be able to decode it when sent by older client that skips this new optional field.)

One more thing that can be improved, when adding lots of new frame variants with slight differences for compatibility reasons is to have a layer for mapping frames to requests, because many frames can map to the same request (e.g. `SignRequest` and `SignRequestWithSigScheme` with `sigscheme: None`).

didier-wenzek: > This however doesn't guarantee the features will keep working, only that server and client can understand each other

Perfect

> Still, I'd opt for keeping these tests below as IMO they provide some value.

Their value is to give some insight on `postcard`, not on thin-edge correctness. But that's okay to keep them.

> Cases where a request can't be changed should be caught by the snapshot tests

Indeed a good use case for snapshot testing

> I'd say new frame version is required when ...

The key point is to document these rules ... otherwise the snapshots might be updated by an unaware maintainer.

crates/extensions/tedge-p11-server/src/proxy/frame1.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+/// An error that can be returned to the client by the server.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ProtocolError(pub String);
+
+/// Snapshot tests for `Frame1` Serialize/Deserialize implementations.
+///
+/// This is to make sure that recent versions of TedgeP11Client can work with older versions of TedgeP11Server and vice
+/// versa.
```
didier-wenzek:
```suggestion
/// This is to make sure that recent versions of TedgeP11Client can work with older versions of TedgeP11Server and vice
/// versa. The `input` bytes should not be updated when `Frame1` is updated: these are reference snapshots.
```

Issue #3868, opened 2025-11-19
tedge init fails when running as non-root user due permissions problems creating /usr/share/tedge


reubenmiller:
**Describe the bug**

Running `tedge init` on an operation system that doesn't have permission to create `/usr/share` will result in the command failing.

Below shows an example of running thin-edge.io on MacOS homebrew formula which installs thin-edge.io under a custom location and also runs the tedge commands as a non-root user:

```sh
$ /opt/homebrew/bin/tedge init \
    --config-dir /opt/homebrew/etc/tedge \
    --user=myuser \
    --group=staff

Error: failed to Initialize tedge

Caused by:
    0: Failed to initialize tedge. You have to run tedge with sudo.
    1: Creating the directory failed: "/usr/share/tedge".
/usr/share/tedge
```

**To Reproduce**

Try running `tedge init` as a non-root user which does not have permission to create `/usr/share/tedge`.

**Expected behavior**

The `/usr/share/tedge` folder should be configurable to allow users to customize the location when using thin-edge.io in non-standard ways

* read-only file systems
* running as non-root (e.g. in containers)

If the `/usr/share/tedge` folder would be configurable, then the user could configure it to create the share folder in a location that is read-writable and where the current user would have permissions to create the directory.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `macOS`
 - Hardware [incl. revision]: `Macbook`
 - System-Architecture [e.g. result of "uname -a"]: `aarch64`
 - thin-edge.io version [e.g. 0.1.0]: `tedge 1.6.2~330+gf5297bb`

**Additional context**

reubenmiller:
> Actually, `tedge init` doesn't create `/usr/share/tedge` directory directly. I think you have `/usr/share/tedge/log-plugins` in the tedge config `log.plugin_paths`. From the code, `tedge init` tries to create parent directories of `log.plugin_paths` if they are missing.
> 
> [thin-edge.io/crates/core/tedge/src/cli/init.rs](https://github.com/thin-edge/thin-edge.io/blob/c2e46794c0ea7a8cb2cc97d293017ada00d825b5/crates/core/tedge/src/cli/init.rs#L98-L100)
> 
> Lines 98 to 100 in [c2e4679](/thin-edge/thin-edge.io/commit/c2e46794c0ea7a8cb2cc97d293017ada00d825b5)
> 
>  for log_plugins_dir in &config.log.plugin_paths.0 { 
[...]

PR #3872, opened 2025-11-24
feat: add share directory path to tedge config

rina23q:
## Proposed changes
- Add `share.path` to control the shared directory path that is used for config, log, diag plugins. (default: `/usr/share`)
- Also, the default values of `diag.plugin_paths` and `log.plugin_paths` respect the `share.path` value.
- `tedge init` tries to create `{share.path}/tedge`.

```sh
# default
tedge config list path
share.path=/usr/share
diag.plugin_paths=["/usr/share/tedge/diag-plugins"]
log.plugin_paths=["/usr/share/tedge/log-plugins"]

# change share.path
sudo tedge config set share.path /foo/bar

# diag/log plugin paths also changed
tedge config list paths
diag.plugin_paths=["/foo/bar/tedge/diag-plugins"]
log.plugin_paths=["/foo/bar/tedge/log-plugins"]
```

## Types of changes

## Paste Link to the issue
#3868 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -917,6 +917,12 @@ define_tedge_config! {
         path: AbsolutePath,
     },
 
+    share: {
+        /// The directory used to store plugins (e.g., log, config, diag)
+        #[tedge_config(example = "/usr/share", default(from_str = "/usr/share"))]
+        path: AbsolutePath,
+    },
+
     firmware: {
         child: {
             update: {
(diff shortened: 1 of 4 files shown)


reubenmiller (changes requested):
**TLDR;**

Since the paths are already configurable, I don't think we need the extra `share.path` variable, as using it adds some unexpected behaviour when adding custom paths to `diag.plugin_paths` and `log.plugin_paths`.

I would suggest removing the `share.paths` and just removing the creation of the folders within the `tedge init` command.

**Detailed example showing the problem**

The introduction of the `share.path` brings in some unexpected behaviour that doesn't work well (and is ignored) when users add their own custom paths.

Below shows the steps which leads to the share.path being ignored and having no effect.

1. Change the default share.path
    ```sh
    tedge config set share.path /foo/bar
    ```

1. Check that the values are changed

    ```sh
    $ tedge config list paths
    diag.plugin_paths=["/foo/bar/tedge/diag-plugins"]
    log.plugin_paths=["/foo/bar/tedge/log-plugins"]
    ```

1. Add an additional path to the diag plugins and check how the paths are affected

    ```sh
    tedge config add diag.plugin_paths /lib/solution/diag
    ```

    Below shows the unexpected behaviour, where the custom `share.path` is no longer used in the default value as the value as been evaluated during the `tedge config add` and added as a fixed value.

    ```sh
    $ tedge config list paths
    diag.plugin_paths=["/lib/solution/diag", "/usr/share/tedge/diag-plugins"]
    log.plugin_paths=["/usr/share/tedge/log-plugins"]
    ```


    Looking at the `tedge.toml` file, it shows why it is showing the default `/usr/share` path.

    ```toml
    [diag]
    plugin_paths = [
        "/lib/solution/diag",
        "/usr/share/tedge/diag-plugins",
    ]
    ```

rina23q:
This PR is replaced with #3876

Issue #3879, opened 2025-11-26
Upgrade nix dependency to 0.30.1 (or latest available)


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The current version of nix (0.26.4) is from August 2023. Whilst there doesn't seem to be any significant features that are impacting thin-edge.io, upgrading the version regularly will help minimize the effort when new versions come out (as generally some breaking changes are introduced).

**Describe the solution you'd like**

Upgrade the package to the following version (though pick a newer version if available):

|dependency|current|upgrade|
|----------|-------|-------|
|nix|0.26.4|0.30.1|

See the [nix Changelog](https://github.com/nix-rust/nix/blob/master/CHANGELOG.md) for details on the breaking changes

**Describe alternatives you've considered**

**Additional context**

PR #3881, opened 2025-11-26
chore: upgrade nix to 0.30.1

didier-wenzek:
## Proposed changes

- [x] upgrade nix to 0.30.1
- [x] remove usage of deprecated flock method

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3879

## Checklist

### DIFF
--- a/crates/common/flockfile/src/unix.rs
+++ b/crates/common/flockfile/src/unix.rs
@@ -1,6 +1,6 @@
 use nix::errno::Errno::EACCES;
 use nix::errno::Errno::EAGAIN;
-use nix::fcntl::flock;
+use nix::fcntl::Flock;
 use nix::fcntl::FlockArg;
 use nix::unistd::write;
 use std::fs::File;
@@ -8,7 +8,6 @@ use std::fs::OpenOptions;
 use std::fs::{self};
 use std::io;
 use std::os::unix::fs::PermissionsExt;
-use std::os::unix::io::AsRawFd;
 use std::path::Path;
 use std::path::PathBuf;
 use tracing::debug;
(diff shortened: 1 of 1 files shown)


crates/common/flockfile/src/unix.rs
```diff
@@ -181,21 +171,6 @@ mod tests {
...
     use tempfile::NamedTempFile;
 
-    #[test]
-    fn lock_access_remove() {
-        let path = NamedTempFile::new().unwrap().into_temp_path().to_owned();
-        let lockfile = Flockfile::new_lock(&path).unwrap();
-
-        assert_eq!(lockfile.path, path);
-
-        lockfile.unlock().unwrap();
```
didier-wenzek, line 191: The `unlock` method was only used by this test and was also adding some useless complications to `Flockfile`. => Removed

rina23q: Now the new struct `Flock` has `unlock()` implementation, so even if needed it, we wouldn't have to implement on our own :)

crates/common/flockfile/src/unix.rs
```diff
@@ -122,21 +116,17 @@ impl Flockfile {
 }
 
 impl Drop for Flockfile {
```
rina23q, line 118: I saw `nix::fcntl::Flock` has `drop()` implementation now, and reading our comment "if the program exits unexpectedly and drop is not called the lock will be removed by the system", I am doubting if we really need `drop()` for `Flockfile`.

rina23q (approved):
Approved since I believe this can be merged as it is now. It's also nice to see how new `nix::fcntl::Flock` can be used. I have one question.

crates/common/flockfile/src/unix.rs
didier-wenzek: `impl Drop for Flockfile` indirectly uses `impl Drop for Flock` and adds a step: removing the file, while `Flock` only remove the lock on drop.

PR #3882, opened 2025-11-26
feat: extensible config management with plugins

albinsuresh:
## Proposed changes

Get and set configuration using external plugins installed at /usr/share/tedge/config-plugins by default.

Extracting existing file-based configuration management logic into a `file` plugin to be done in a follow-up PR.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/docs/src/extend/config-management.md
+++ b/docs/src/extend/config-management.md
@@ -1,0 +1,269 @@
+---
+title: Configuration Management
+tags: [Extend, Configuration Management]
+sidebar_position: 3
+description: Extend configuration management capabilities with plugins
+---
+
+This document describes how to extend the configuration management capabilities of %%te%% using plugins.
+
+By default, %%te%% supports managing file-based configurations.
+But, this out-of-the-box support is limited to reading or updating a file on the file system.
+
+Custom configuration plugins enable more flexible configuration workflows:
+1. Configuration files can be read from any source, such as the file system, a database, or a registry.
+1. While fetching the configuration for a type, even if it is split across multiple files (e.g: config extensions),
+   they can be combined into a single "effective configuration" before it is returned.
+1. Pre-processing or post-processing steps can be performed while updating a configuration.
+   Common pre-processing steps, include:
+   - validating the new config
+   - taking a backup of the existing config before the new config is applied
+   Similarly, some commonly seen post-processing step are:
+   - reloading/restarting the service using that config, for the updated config to take effect
+   - restore the previously backed up config, if applying the new one fails.
+
+## Overview
+
+The %%te%% agent supports extensible configuration management through a plugin system:
+
+* Plugins can be installed to support any kind of configuration sources, including non-file formats
+* Each plugin can handle multiple configuration types
+* Plugins are discovered and executed automatically by `tedge-agent`
+
+## How Config Plugins Work
+
+A config plugin is an executable that implements a simple command-line interface with the following requirements:
+
+* Implement three sub-commands:
+  1. **`list`** - Returns all config types supported by the plugin
+  2. **`get <type>`** - Retrieves the current configuration for the specified type
+  3. **`set <type> <new-config-path>`** - Updates the configuration for the specified type from the provided config file path
+* Must exit with code 0 for when successful.
+* Should output all config types supported by the plugin, one per line, for the `list` command.
+* Should output the configuration content to stdout for the `get` command.
+* Should handle validation, backup, restoration as well as any required service restart in the `set` command.
+* Should handle errors gracefully, emitting those to `stderr`, and exit with non-zero codes on failure.
+
+The agent automatically:
+* Discovers plugins at startup by running their `list` command to gather their supported config types.
+  skipping the ones returning a non zero exit code.
+* Publishes the supported config types to the device's `config_snapshot` and `config_update` command metadata topics
+  (e.g: `te/device/main///cmd/config_snapshot` and `te/device/main///cmd/config_update`)
+  with a plugin suffix in the format `<config-type>::<plugin-name>` (e.g., `lighttpd.conf::lighttpd`)
+* Routes `config_snapshot` requests to the `get` command of the appropriate plugin based on the type suffix.
+* Routes `config_update` requests to the `set` command of the appropriate plugin based on the type suffix.
+* Detects any new plugin installations dynamically.
+* Refreshes the supported config types by reloading the plugins when any new software is installed or configuration is updated.
+
+Config plugins are installed at `/usr/share/tedge/config-plugins`, by default.
+
+This plugin root directory can be changed using:
+```sh
+sudo tedge config set configuration.plugin_paths /usr/local/share/tedge/config-plugins,/usr/share/tedge/config-plugins
+```
+
+Multiple plugin directories can be specified if a layered directory structure is desired,
+where plugins from directories earlier in the list gets precedence over the latter ones.
+For example, with the above configuration, a `file` plugin in `/usr/local/share/tedge/config-plugins`
+would override the one with the same name in `/usr/share/tedge/config-plugins`.
+This mechanism can be used to override the factory plugins when needed.
+
+## Permissions
+
+Plugins are executed by the agent with `sudo` privileges.
+When %%te%% is installed using the official packages available for the supported package managers,
+the `tedge` package automatically creates the following sudoers entry,
+giving sudo rights to all plugins installed at `/usr/share/tedge/config-plugins`:
+
+```
+tedge    ALL = (ALL) NOPASSWD:SETENV: /usr/share/tedge/config-plugins/[a-zA-Z0-9]*
+```
+
+If %%te%% is installed using some alternative installation methods,
+then the above sudoers entry must also be added explicitly. 
+
+If the `config.plugin_paths` config is updated with additional directories as shown in the previous section,
+then sudoers entries must be created for those directories as well.
+
+Additionally, ensure your plugin has appropriate permissions to access the configuration sources it needs.
+
+## Creating a Custom Config Plugin
+
+### Example: lighttpd plugin
+
+The following example demonstrates a complete plugin implementation for managing the lighttpd web server configuration, showcasing how to handle multi-file configurations, validation, backup/restore, and service reloading.
+
+Key features of this plugin are:
+
+* The `list` command outputs the main configuration file: `lighttpd.conf`
+* The `get` command retrieves the "effective configuration":
+  - Combined the main configuration with all config extensions
+  - Applies the default values for those not provided explicitly
+* The `set` command implements a safe update process:
+  - Validates the new configuration before applying it
+  - Creates a backup of the existing configuration
+  - Applies the new configuration
+  - Restarts the `lighttpd` service
+  - Restores from backup if the reload fails
+  - Cleans up the backup on success
+
+Here is the complete plugin:
+
+```sh
+#!/bin/sh
+set -eu
+
+usage() {
+    echo "Usage: $0 <command> [args...]"
+    echo "Commands:"
+    echo "  list                                    List all config types supported by this plugin"
+    echo "  get <type>                              Print the config for the specified type to stdout"
+    echo "  set <type> <new-config-path>            Update the config for the specified type from the new config path"
+    exit 1
+}
+
+list_config_types() {
+    echo lighttpd.conf
+}
+
+get_config() {
+    config_type="$1"
+
+    if [ "$config_type" != "lighttpd.conf" ]; then
+        echo "Error: Unsupported config type '$config_type'" >&2
+        exit 1
+    fi
+
+    # Print the effective configuration, aggregating all config extensions and applying defaults
+    lighttpd -p -f /etc/lighttpd/lighttpd.conf
+}
+
+set_config() {
+    config_type="$1"
+    target_config="$2"
+    config_path="/etc/lighttpd/lighttpd.conf"
+    backup_path="${config_path}.backup"
+
+    if [ "$config_type" != "lighttpd.conf" ]; then
+        echo "Error: Unsupported config type '$config_type'" >&2
+        exit 1
+    fi
+
+    # Verify the new configuration before applying it
+    if ! lighttpd -t -f "$target_config" 2>&1; then
+        echo "Error: Configuration validation failed" >&2
+        exit 1
+    fi
+
+    # Backup existing configuration
+    if [ -f "$config_path" ]; then
+        cp "$config_path" "$backup_path"
+    fi
+
+    # Apply new configuration
+    mv "$target_config" "$config_path"
+
+    # Restart the service and check if it succeeded
+    if ! systemctl restart lighttpd; then
+        echo "Error: Failed to reload lighttpd service, restoring backup" >&2
+        if [ -f "$backup_path" ]; then
+            mv "$backup_path" "$config_path"
+            systemctl reload lighttpd
+        fi
+        exit 1
+    fi
+
+    rm -f "$backup_path"
+}
+
+main() {
+    if [ $# -lt 1 ]; then
+        usage
+    fi
+
+    command="$1"
+    shift
+
+    case "$command" in
+        list)
+            list_config_types
+            ;;
+        get)
+            if [ $# -lt 1 ]; then
+                echo "Error: 'get' command requires a <type> argument" >&2
+                exit 1
+            fi
+            get_config "$@"
+            ;;
+        set)
+            if [ $# -lt 2 ]; then
+                echo "Error: 'set' command requires <type> and <new-config-path> arguments" >&2
+                exit 1
+            fi
+            set_config "$@"
+            ;;
+        *)
+            echo "Error: Unknown command '$command'" >&2
+            usage
+            ;;
+    esac
+}
+
+main "$@"
+```
+
+### Installation
+
+Copy the plugin to the plugins directory and make it executable:
+
+```sh
+sudo cp /path/to/lighttpd /usr/share/tedge/config-plugins/lighttpd
+sudo chmod +x /usr/share/tedge/config-plugins/lighttpd
+```
+
+### Testing the Plugin
+
+List all config types the plugin supports:
+
+```sh
+sudo /usr/share/tedge/config-plugins/lighttpd list
+```
+
+Retrieve the current configuration:
+
+```sh
+sudo /usr/share/tedge/config-plugins/lighttpd get lighttpd.conf
+```
+
+Update the configuration (note: this will actually modify your system):
+
+```sh
+sudo /usr/share/tedge/config-plugins/lighttpd set lighttpd.conf /path/to/new-lighttpd.conf
+```
+
+## Best Practices for Config Plugins
+
+When creating custom config plugins, follow these best practices:
+
+1. Validate the new configuration before applying it.
+2. Always create a backup before updating a configuration, and restore it if the update fails.
+3. After updating a configuration, reload or restart the service as appropriate.
+
+## Refresh Supported Config Types
+
+To ensure that any newly installed services or configuration sources are immediately available for configuration management,
+`tedge-agent` automatically reloads the plugins and refreshes the supported config types on the following events:
+
+* A new config plugin is installed in the plugin directory (`/usr/share/tedge/config-plugins`)
+* The `tedge-config-plugin.toml` file is updated
+* A new software is installed with the agent (via the `software_update` command)
+
+A refresh can also be triggered manually by sending sync signals to the agent as follows:
+
+```sh
+tedge mqtt pub te/device/main/service/tedge-agent/signal/sync_config '{}'
+```
+
+The agent reacts to all these events by gathering the latest supported config types from all the installed plugins
+by invoking the `list` command on them,
+and publishes the aggregated types to the `te/device/main///cmd/config_snapshot` and `te/device/main///cmd/config_update` meta topics.
--- a/docs/src/extend/log-management.md
+++ b/docs/src/extend/log-management.md
@@ -1,5 +1,5 @@
 ---
-title: Extensible Log Management
+title: Log Management
 tags: [Extend, Log Management]
 sidebar_position: 2
 description: Extend log management capabilities with plugins
@@ -57,11 +57,12 @@ This mechanism can be used to override the factory plugins when needed.
 ## Permissions
 
 Plugins are executed by the agent with `sudo` privileges.
-The agent automatically creates the following sudoers entry,
-giving sudo rights to all plugins installed at `/usr/local/lib/tedge/log-plugins`:
+When %%te%% is installed using the official packages available for the supported package managers,
+the `tedge` package automatically creates the following sudoers entry,
+giving sudo rights to all plugins installed at `/usr/share/tedge/log-plugins`:
 
 ```
-tedge    ALL = (ALL) NOPASSWD:SETENV: /usr/local/lib/tedge/log-plugins/[a-zA-Z0-9]*
+tedge    ALL = (ALL) NOPASSWD:SETENV: /usr/share/tedge/log-plugins/[a-zA-Z0-9]*
 ```
 
 If the `log.plugin_paths` config is updated with additional directories as shown in the previous section,


tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
reubenmiller, line 1: We might need to rethink this example as it doesn't highlight the usefulness of the feature, as it is a lot of effort to go through just to restart the service after the config is changed.

Of course you can do anything within the script, however having an example with a real-world use-case would be good to highlight why we're doing this feature in the first place.

So some ideas:
* Configuration validator (checks if the config is valid before applying it)
* use config to perform a backup and restore (e.g. the get will collect multiple files and return a tar.gz file, and then the backup file can be "restored" by using the config set mechanism)

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -995,6 +995,13 @@ define_tedge_config! {
         #[tedge_config(example = "/usr/share/log-plugins,/etc/tedge/log-plugins", default(value = "/usr/share/tedge/log-plugins"))]
         plugin_paths: TemplatesSet,
     },
+
+    configuration: {
+        /// The directories where log plugins are stored
+        #[tedge_config(example = "/usr/share/config-plugins,/usr/share/local/config-plugins", default(value = "/usr/share/tedge/config-plugins"))]
```
didier-wenzek:
```suggestion
        /// The directories where configuration plugins are stored
        #[tedge_config(example = "/usr/share/config-plugins,/usr/share/local/config-plugins", default(value = "/usr/share/tedge/config-plugins"))]
```

crates/extensions/tedge_config_manager/src/config.rs
```diff
@@ -24,6 +25,7 @@ use tedge_utils::file::PermissionEntry;
 
 use super::error::InvalidConfigTypeError;
 
+pub const DEFAULT_PLUGIN_DIR: &str = "config-plugins";
```
didier-wenzek: Why do we need this constant while there is already a default for `configuration.plugin_paths`?

tests/RobotFramework/tests/cumulocity/configuration/plugins/hello
```diff
@@ -0,0 +1,74 @@
+#!/bin/bash
+
+set -euo pipefail
+
+usage() {
+    echo "Usage: $0 <command> [args...]"
+    echo "Commands:"
+    echo "  list                                    List all config types supported by this plugin"
+    echo "  get <type> <target-file-path>           Get config for the specified type and write to the target path"
+    echo "  set <type> <new-config-path>            Set config for the specified type with the new config path"
```
didier-wenzek: The contract is not clear as this comment is not aligned with the code.

I understand the comment as:

```suggestion
    echo "  get <type> <target-file-path>          Copy the config for the specified type to the target path"
    echo "  set <type> <new-config-path>        Update the config for the specified type from new config path"
```

But the script behaves as:


```suggestion
    echo "  get <type> <target-file-path>           Copy the config for the specified type to stdout"
    echo "  set <type> <new-config-path>         Update the config for the specified type from new config path"
```

tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
albinsuresh: > as it is a lot of effort to go through just to restart the service after the config is changed

I was under the impression that this missing restart, although minimal, was the most common use-case and the key step we're missing for a lot of software like `collectd`, `mosquitto` or even our own services `tedge-agent` and `tedge-mapper` for certain kind of config changes. Initially I was thinking of writing a `collectd` plugin, but then chose an even more minimal example to avoid additional dependencies and to keep the plugin as simple as possible. The intention was to keep the focus on the plugin contract itself instead of its sophisticated logic.

That being said, I agree that we need to write a sophisticated example as well, more for the documentation than the system test as the user might appreciate that more than the simplistic example.

But, I completely agree on the effort part. That's why the proposal is to add the `exec` command support to the `file` plugin itself, so that the users can easily specify that missing step with the existing stuff, without the overhead of writing an entire plugin.

reubenmiller: Yes the restarting of a service after a config is definitely the most common use-case, though as you said that we don't expect people to use this side of plugins to perform this task. So it is mostly about seeing if we can validate the design by creating a realistic (but simple) example which showcases functionality that was not possible before.

crates/core/tedge_api/src/commands.rs
```diff
@@ -905,9 +905,9 @@ impl CommandPayload for ConfigUpdateCmdPayload {
 }
 
 impl ConfigUpdateCmdPayload {
-    pub fn successful(&mut self, path: impl Into<String>) {
+    pub fn successful(&mut self, path: Option<impl Into<String>>) {
         self.status = CommandStatus::Successful;
-        self.path = Some(path.into())
+        self.path = path.map(|p| p.into());
```
jarhodes314: Not actually terribly related to your change, but it feels like `ToString` better captures the intent than `Into<String>`

```suggestion
    pub fn successful(&mut self, path: Option<impl ToString>) {
self.status = CommandStatus::Successful;
        self.path = path.map(|p| p.to_string());
```

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -228,9 +235,37 @@ impl ConfigManagerWorker {
...
+
+                    target_file.to_path_buf()
+                } else {
+                    return Err(ConfigManagementError::PluginError {
+                        plugin_name: plugin_name.to_string(),
+                        reason: "Plugin not found".to_string(),
+                    });
+                }
+            } else {
+                // File-based config retrieval
```
jarhodes314: I think this comment is possibly useful, but doesn't really explain what's going on at the moment (either what "file-based" really means, or why we are doing that). I presume what this is intended to convey is something along the lines of "no plugin is specified, fall back to the legacy file-based behaviour".

crates/extensions/tedge_config_manager/src/plugin.rs
```diff
@@ -0,0 +1,162 @@
...
+            .filter(|line| !line.is_empty())
+            .collect();
+
+        Ok(config_types)
+    }
+
+    pub(crate) async fn get(
+        &self,
+        config_type: &str,
+        target_file_path: &Path,
```
jarhodes314: I don't understand why we're taking `&Path` instead of `&Utf8Path` here, particularly given when we call this function, it looks like we call `Utf8Path::as_std_path` to convert the value to a `Path` in the first place.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -228,9 +235,37 @@ impl ConfigManagerWorker {
         topic: &Topic,
         request: &mut ConfigSnapshotCmdPayload,
     ) -> Result<Utf8PathBuf, ConfigManagementError> {
-        let file_entry = self
-            .plugin_config
-            .get_file_entry_from_type(&request.config_type)?;
+        let config_path =
+            if let Some((config_type, plugin_name)) = request.config_type.split_once("::") {
```
didier-wenzek: It would be really convenient for a maintainer to have two functions (sitting side by side), one to make a c8y config type from a plugin name and a plugin config type, another to split a c8y config type. This would also include the logic related to the legacy file-based plugin (which is unnamed)

Grepping for `::` over Rust code in not really helpful ;-)

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -505,15 +557,45 @@ impl ConfigManagerWorker {
...
+            if let Some(plugin) = self.external_plugins.by_plugin_type(&plugin_type) {
+                match plugin.list(None).await {
+                    Ok(conf_types) => {
+                        info!(
+                            target: "config plugins",
+                            "Plugin {} supports config types: {:?}", plugin_type, conf_types
+                        );
+
+                        for conf_type in conf_types {
+                            config_types.push(format!("{}::{}", conf_type, plugin_type));
```
didier-wenzek: See the comment on the reverse logic `config_type.split_once("::")`. The code has to be reorganized to have these two points side by side.

design/decisions/0008-extensible-config-management.md
```diff
@@ -118,9 +117,8 @@ The agent uses the plugins as follows:
 - The `file` plugin's is implemented as follows:
   - `list`: list all the config types listed in `tedge-configuration-plugin.toml`.
   - `get`: Copy the contents of the existing configuration to the temp target file argument passed to it.
-  - `set`: Replace the existing config file with new config file in the argument.
-  - `finalize`: Perform the command specified in the `exec` field.
-    No-op when no `exec` command is provided.
+  - `set`: Replace the existing config file with new config file in the argument
+     and execute the command specified in the `exec` field, if one is provided.
```
didier-wenzek, line 121: Is this implemented? If so, I failed to find where.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -358,22 +389,43 @@ impl ConfigManagerWorker {
...
-        let to = Utf8PathBuf::from(&file_entry.path);
-
-        if let Some(parent) = to.parent() {
-            if !parent.exists() {
-                self.create_parent_dirs(parent, file_entry)?;
-            }
-        }
+        let config_path =
+            if let Some((config_type, plugin_name)) = request.config_type.split_once("::") {
+                if let Some(plugin) = self.external_plugins.by_plugin_type(plugin_name) {
```
didier-wenzek: This logic being repeated, I would consider to add a method `Actor.get_plugin(c8y_plugin_type) -> Result<Option<Plugin>>`

design/decisions/0008-extensible-config-management.md
albinsuresh: No. This is not done yet. The plan is to extract the file-based config handling into a standalone `file` plugin in a follow-up PR and then add this functionality to it. Thought of keeping the scope of this PR limited to just the plugin support.

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: Resolved by a16fff1cbf

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: Resolved by a16fff1cbf

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: Resolved by a16fff1cbf

crates/extensions/tedge_config_manager/src/plugin.rs
albinsuresh: Resolved by a16fff1cbf

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: Resolved by a16fff1cbf

crates/core/tedge_api/src/commands.rs
albinsuresh: Resolved by a16fff1cbf

tests/RobotFramework/tests/cumulocity/configuration/plugins/hello
albinsuresh: Resolved by a16fff1cbf

crates/extensions/tedge_config_manager/src/config.rs
albinsuresh: Resolved by a16fff1cbf

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
albinsuresh: Resolved by a16fff1cbf

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -228,9 +236,31 @@ impl ConfigManagerWorker {
         topic: &Topic,
         request: &mut ConfigSnapshotCmdPayload,
     ) -> Result<Utf8PathBuf, ConfigManagementError> {
-        let file_entry = self
-            .plugin_config
-            .get_file_entry_from_type(&request.config_type)?;
+        let config_path = if let Some(plugin) = self.get_plugin(&request.config_type)? {
+            let (config_type, plugin_name) = split_cloud_config_type(&request.config_type)
+                .expect("get_plugin returned Some, so config_type must have a plugin");
```
didier-wenzek, line 243: Minor. A way to remove the `expect` is to let the `get_plugin()` method to return the config type along the plugin.


```suggestion
        let config_path = if let Some((plugin, config,_type)) = self.get_plugin(&request.config_type)? {
```

However, this makes the `get_plugin()` signature a bit too adhoc. So the `expect` is fine.

tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
albinsuresh: Resolved by de2b3030fd and 81df499, using `lighttpd` as the guinea pig service, as it supports config validation without really applying the config.

tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
```diff
@@ -0,0 +1,102 @@
...
+    if [[ -f "$config_path" ]]; then
+        cp "$config_path" "$backup_path"
+    fi
+
+    # Apply new configuration
+    mv "$target_config" "$config_path"
+
+    # Reload service and check if it succeeded
+    if ! systemctl restart lighttpd; then
+        echo "Error: Failed to reload lighttpd service, restoring backup" >&2
```
albinsuresh, line 56: We could move this backup/restore logic to the agent so that all configs get this functionality out-of-the-box.

I'm not sure if doing this implicit restore of the old config is "safe" for all config types and hence wondering if the plugins should have an option that lets them express if they want to opt into or out of this automated backup/restore functionality.

One option would be to overload the `set` command exit code, with a predefined value like `100`, to convey if they want automated restore or not.

An alternate option is to introduce an additional sub-command like:
```
./plugin auto-restore-on-failed-update <config-type>
```

design/decisions/0008-extensible-config-management.md
rina23q, line 1: Probably you can update the date and status in the page header.

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,255 @@
...
+where plugins from directories earlier in the list gets precedence over the latter ones.
+For example, with the above configuration, a `file` plugin in `/usr/local/share/tedge/config-plugins`
+would override the one with the same name in `/usr/share/tedge/config-plugins`.
+This mechanism can be used to override the factory plugins when needed.
+
+## Permissions
+
+Plugins are executed by the agent with `sudo` privileges.
+The agent automatically creates the following sudoers entry,
+giving sudo rights to all plugins installed at `/usr/local/lib/tedge/config-plugins`:
```
rina23q: To keep the directory consistency with the example above.

```suggestion
giving sudo rights to all plugins installed at `/usr/local/share/tedge/config-plugins`:
```

Also, 
```
tedge    ALL = (ALL) NOPASSWD:SETENV: /usr/local/share/tedge/config-plugins/[a-zA-Z0-9]*
```

tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
```diff
@@ -0,0 +1,113 @@
...
+Library             DateTime
+Library             OperatingSystem
+Library             String
+Library             Cumulocity
+Library             ThinEdgeIO
+
+Suite Setup         Custom Setup
+Test Teardown       Get Logs
+
+Test Tags           theme:c8y    theme:log
```
rina23q:
```suggestion
Test Tags           theme:c8y    theme:configuration
```

crates/extensions/tedge_config_manager/src/plugin.rs
```diff
@@ -0,0 +1,153 @@
...
+        let output = command
+            .execute(command_log)
+            .await
+            .map_err(|err| self.plugin_error(err))?;
+
+        if !output.status.success() {
+            let stderr = String::from_utf8_lossy(&output.stderr);
+            return Err(ConfigManagementError::PluginError {
+                plugin_name: self.name.clone(),
+                reason: format!("Set command failed: {}", stderr),
```
rina23q: I think this `execute()` is called by not only `set`, but also `list` and `get`, so the error message is misleading?

tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
```diff
@@ -0,0 +1,102 @@
...
+    local config_path="/etc/lighttpd/lighttpd.conf"
+    local backup_path="${config_path}.backup"
+
+    if [[ "$config_type" != "lighttpd.conf" ]]; then
+        echo "Error: Unsupported config type '$config_type'" >&2
+        exit 1
+    fi
+
+    # Verify the new configuration before applying it
+    if ! lighttpd -t -f "$target_config" 2>&1; then
```
didier-wenzek, line 41: A nice idea! Both for a system to provide this configuration validation option and to use it in a configuration update script.

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,255 @@
...
+1. Non-file based configuration sources are not supported
+2. While updating a configuration, no pre-processing or post-processing of the config can be performed.
+
+Custom configuration plugins address these limitations and enable more flexible configuration workflows.
+
+## Overview
+
+The %%te%% agent supports extensible configuration management through a plugin system:
+
+* Plugins can be installed to support any kind of configuration sources, including non-file formats
```
didier-wenzek, line 29: I don't get what you mean by *non-file formats*. In all cases what is provided by the user is a file. I guess that you mean that the configuration might actually be stored in some opaque registry; but it might be confusing.

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,255 @@
...
+
+## How Config Plugins Work
+
+A config plugin is an executable that implements a simple command-line interface with the following requirements:
+
+* Implement three sub-commands:
+  1. **`list`** - Returns all config types supported by the plugin (one per line)
+  2. **`get <type>`** - Retrieves the current configuration for the specified type and outputs it to stdout
+  3. **`set <type> <new-config-path>`** - Updates the configuration for the specified type from the provided config file path
+* Must exit with code 0 for successful command (used to validate the plugin using the `list` command).
```
didier-wenzek: The use of the exit code is beyond the list command.

```suggestion
* Must exit with code 0 for successful command.
```

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,255 @@
...
+* Should output the configuration content to stdout for the `get` command.
+* Should handle validation, backup, and restoration in the `set` command.
+* Should handle errors gracefully and exit with non-zero codes on failure.
+
+The agent automatically:
+* Discovers plugins at startup by running their `list` command to gather their supported config types.
```
didier-wenzek, line 48: I would move here the note about the list command exit code.

```suggestion
* Discovers plugins at startup by running their `list` command to gather their supported config types (skipping plugins returning a non zero exit code on `list`).
```

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,255 @@
...
+
+A config plugin is an executable that implements a simple command-line interface with the following requirements:
+
+* Implement three sub-commands:
+  1. **`list`** - Returns all config types supported by the plugin (one per line)
+  2. **`get <type>`** - Retrieves the current configuration for the specified type and outputs it to stdout
+  3. **`set <type> <new-config-path>`** - Updates the configuration for the specified type from the provided config file path
+* Must exit with code 0 for successful command (used to validate the plugin using the `list` command).
+* Should output the configuration content to stdout for the `get` command.
+* Should handle validation, backup, and restoration in the `set` command.
```
didier-wenzek:
```suggestion
* Should handle validation, backup, and restoration in the `set` command, as well as any required service restart.
```

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,255 @@
...
+Multiple plugin directories can be specified if a layered directory structure is desired,
+where plugins from directories earlier in the list gets precedence over the latter ones.
+For example, with the above configuration, a `file` plugin in `/usr/local/share/tedge/config-plugins`
+would override the one with the same name in `/usr/share/tedge/config-plugins`.
+This mechanism can be used to override the factory plugins when needed.
+
+## Permissions
+
+Plugins are executed by the agent with `sudo` privileges.
+The agent automatically creates the following sudoers entry,
```
didier-wenzek: This is not correct. The agent changes nothing in the sudoers file.

This is done by the installers provided by thin-edge and has to be done manually for custom install.

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,255 @@
...
+            echo "Error: Unknown command '$command'" >&2
+            usage
+            ;;
+    esac
+}
+
+main "$@"
+```
+
+Key features of this plugin:
```
didier-wenzek: I would list these features before the script itself, as an appetizer.

didier-wenzek (commented):
The `lighttpd` example nicely highlights the motivation for configuration plugins: configuration checks / service restart / backup / rollback

Note that:
- Shellcheck is failing: https://github.com/thin-edge/thin-edge.io/actions/runs/19862804596/job/56917189687?pr=3882

docs/src/extend/config-management.md
reubenmiller: The correct location is actually `/usr/share` not `/usr/local/share`...the full rule would be:

```
/usr/share/tedge/config-plugins/[a-zA-Z0-9]*
```

tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
didier-wenzek: I would prefer to stay on the safer and simpler side of doing nothing here. A restoration logic is context specific and depends on all the previous steps. The current proposal is flexible enough to address this is a few lines of code as demonstrated by the `lighttpd` example.

docs/src/extend/config-management.md
albinsuresh: Resolved by 7e4c33a

docs/src/extend/config-management.md
albinsuresh: Resolved by 7e4c33a

docs/src/extend/config-management.md
albinsuresh: Resolved by 7e4c33a

docs/src/extend/config-management.md
albinsuresh: Resolved by 7e4c33a

docs/src/extend/config-management.md
albinsuresh: Resolved by 7e4c33a

docs/src/extend/config-management.md
albinsuresh: Resolved by 7e4c33a

crates/extensions/tedge_config_manager/src/plugin.rs
albinsuresh: Resolved by [d8817e6](https://github.com/thin-edge/thin-edge.io/pull/3882/commits/d8817e6b0910c39afad6248b347ccc24938ba097)

tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
albinsuresh: Resolved by [d8817e6]

design/decisions/0008-extensible-config-management.md
albinsuresh: Resolved by [d8817e6]

docs/src/extend/config-management.md
albinsuresh: Resolved by https://github.com/thin-edge/thin-edge.io/commit/7e4c33a856ffbb6f1bf3604e78f28003d5e1c308

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,272 @@
...
+   Each file needs to be fetched individually, using independent types.
+1. While updating a configuration, no pre-processing or post-processing of the config can be performed.
+   Common pre-processing steps, includes:
+   - validating the new config
+   - taking a backup of the existing config before the new config is applied
+   Similarly, commonly seen post-processing step are:
+   - reloading/restarting the service using that config, for the updated config to take effect
+   - restore the previously backed up config, if applying the new one fails.
+
+Custom configuration plugins address these limitations and enable more flexible configuration workflows.
```
didier-wenzek: I wonder if this can be turned in a positive style. I initially read this as limitations of the configuration plugins and I have been a bit confused!

Something along the lines:

```
By default, %%te%% supports managing file-based configurations.
But, this out-of-the-box support is limited to reading or updating a file on the file system.

Custom configuration plugins enable more flexible configuration workflows:
1. Configuration files can be read from any source, from the file system as well as a database or a registry.
2. Pre-processing or post-processing steps can be performed while updating a configuration.
```

didier-wenzek (approved):
Approving, with a suggestion to improve the doc.

This is a great feature which benefits are nicely demonstrated by the `lighttpd` example.

P.S. Shellcheck is still unhappy with the *generated* `preinst` scripts which are no more aligned with the reference script.

configuration/package_manifests/nfpm.tedge.yaml
```diff
@@ -93,6 +93,10 @@ contents:
       mode: 0644
     packager: rpm
 
+  # config plugins directory
+  - dst: /usr/share/tedge/config-plugins
+    type: dir
+
```
reubenmiller: We can skip this folder as we don't need to create the folder by default (as we don't do it from the log plugins).

Once the `file` config plugin exists, then it will automatically create the folder.

tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
```diff
@@ -0,0 +1,109 @@
...
+
+    Command Metadata Should Have Refreshed    ${start_time}
+
+Supported config types updated on sync signal
+    ${start_time}=    Get Unix Timestamp
+    Execute Command    tedge mqtt pub te/device/main/service/tedge-agent/signal/sync '{}'
+    Command Metadata Should Have Refreshed    ${start_time}
+
+    ${start_time}=    Get Unix Timestamp
+    Execute Command    tedge mqtt pub te/device/main/service/tedge-agent/signal/sync_config_snapshot '{}'
```
reubenmiller: I'm a bit surprised seeing that we have two different signals to sync a `sync_config_snapshot` and a `sync_config_update`...do we really need both since they are both operating on a single config type list (e.g. the supported configuration types)?

configuration/package_manifests/nfpm.tedge.yaml
albinsuresh: Yeah, this directory creation will be replaced with the `file` plugin copy (as done for the log plugin) once that is available. Creating it for now is still harmless, right? As someone wanting to try out this feature has one less step to do? And avoids any potential directory permission issues as well, if the user creates it himself.

tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
albinsuresh: Ideally, no. I just kept it consistent with the `cmd` meta topics which has the same config list duplicated on both `cmd/config_snapshot` and `cmd/config_update` topics as well. IIRC, the rationale for this design was that the types that can be fetched vs updated could differ in some environments, as not all configs that can be read are safe to be updated.  This disjoint design lets the users express that as well, by having a smaller subset of types as updatable compared to the readable ones. Unfortunately c8y doesn't support this difference, but that's just the c8y limitation while we can remain flexible.

But, I agree that in most cases, both the lists would be the same. So, do you propose to keep these separate signals to retain that consistency and add an additional 2-in-1 `signal/config_operation` that triggers both? Or don't want to to have the separate signals at all?

configuration/package_manifests/nfpm.tedge.yaml
reubenmiller: It has the potential to cause unnecessary headaches with folder ownership (in the sense of which folders are owned by the debian packages that created them). Since we'll be transitioning to just putting files inside the folder (and not explicitly creating the folder), then that will just eliminate any potential headaches from changes from folder to file ownership within the linux package.

tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
reubenmiller: I think we only need one, as trigger both seems unusual as it just gives the potential for the two lists to get out of sync...and since the values within those fragments are both driven by the same config plugin list, then it doesn't really need both (but I don't mind if we still maintain two lists of identical values)...though it would be interesting to see which value the tedge-mapper-c8y uses to publish it to Cumulocity ;)

I'd just go with `signal/sync_config` that is the common prefix for both commands, and remove the confusion and annoyance of having to send two signals to update both lists when a new config plugin is added or removed.

docs/src/extend/config-management.md
albinsuresh: Resolved by bd6ea9a

tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
albinsuresh: Resolved by [695ba2f](https://github.com/thin-edge/thin-edge.io/pull/3882/commits/695ba2f6a6d12b2c9f4246a46265ca4dd5294b56)

configuration/package_manifests/nfpm.tedge.yaml
albinsuresh: Resolved by [c9e0eb6](https://github.com/thin-edge/thin-edge.io/pull/3882/commits/c9e0eb65309a3849adbaac3c2e4e270f54469424)

docs/src/extend/config-management.md
```diff
@@ -0,0 +1,269 @@
...
+`tedge-agent` automatically reloads the plugins and refreshes the supported config types on the following events:
+
+* A new config plugin is installed in the plugin directory (`/usr/share/tedge/config-plugins`)
+* The `tedge-config-plugin.toml` file is updated
+* A new software is installed with the agent (via the `software_update` command)
+
+A refresh can also be triggered manually by sending sync signals to the agent as follows:
+
+```sh
+tedge mqtt pub te/device/main/service/tedge-agent/signal/sync_config_snapshot '{}'
```
albinsuresh:
```suggestion
tedge mqtt pub te/device/main/service/tedge-agent/signal/sync_config '{}'
```

albinsuresh: Will address this before the final rebase, to avoid yet another build for this one-liner.

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -169,15 +170,30 @@ impl WorkflowActor {
...
+                    self.sync_signal_dispatcher.sync(operation).await?;
+                }
+            },
+            SignalType::Custom(signal) => {
+                match signal.as_str() {
+                    "sync_config" => {
+                        info!("Received sync_config signal, requesting the builtin config actor to sync");
+                        self.sync_signal_dispatcher
+                            .sync(OperationType::ConfigSnapshot)
+                            .await?;
```
didier-wenzek, line 193: Don't we have to synchronize both `ConfigSnapshot` and `ConfigUpdate` ?

albinsuresh: The actor syncs both `cmd/config_snapshot` and `cmd/config_update` metadata on receipt of either of these signals. So, it doesn't really matter which one we send. I'll just add a code comment to clarify this point.

Issue #3884, opened 2025-12-01
ThinEdge C8Y Proxy sent the Identity API GET request until ";" character of the actual request


Oezge-Akin:
**Describe the bug**

The issue is produced with the request to the Identity API but this might be applicable to all C8Y API requests.

Using the thinEdge as the C8Y proxy, when the URL contains ";" in the URL path, the thinEdge seems to send the request URL until ";" character to the platform.

**To Reproduce**

- Have the thinEdge connected to your tenant.
- Have a device in the C8Y platform and with identity containing ";" character in the value.
This is what I have:

<img width="1788" height="401" alt="Image" src="https://github.com/user-attachments/assets/3a2eac50-940d-4f3c-99cf-fde434197243" />
 
- Retrieve the device object from the external Id that contains ";" character in the value
For this example:

`GET http://localhost:8001/c8y/identity/externalIds/c8y_OpcuaDevice/tId%3A50175621%3BsId%3D37316412%3Bnsu%3Durn%3Acumulocity%3Aopcua%3Atest%3Aserver%3Bs%3DDynamic%20Playground`

**Expected behavior**
It returns the result with the device managed object that is bound to, like the following:

```
{
    "self": "...",
    "externalId": "tId:50175621;sId=37316412;nsu=urn:cumulocity:opcua:test:server;s=Dynamic Playground",
    "type": "c8y_OpcuaDevice",
    "managedObject": {
        "self": "...",
        "id": "25314167"
    }
}
```

**Actual behavior**

It returns 404 Not Found with message, like the following:

```
{
    "message": "External id not found; external id = ID [type=c8y_OpcuaDevice, value=tId:50175621]",
    "error": "identity/Not Found",
    "info": "https://cumulocity.com/api/core/"
}
```

The external ID value from the error message is the substring of the actual value until the first  ";" character.

**Environment (please complete the following information):**
 - thin-edge is targeting an eu-latest tenant.
 - tedge version is 1.6.0

**Additional context**

When doing the request to my tenant domain directly, it's able to return response correctly:

```
GET https://<myTenant>.eu-latest.cumulocity.com/identity/externalIds/c8y_OpcuaDevice/tId%3A50175621%3BsId%3D37316412%3Bnsu%3Durn%3Acumulocity%3Aopcua%3Atest%3Aserver%3Bs%3DDynamic%20Playground

{
    "self": "...",
[...]

PR #3886, opened 2025-12-02
fix: prevent URI percent-decoding in c8y_auth_proxy path handling

rina23q:
## Proposed changes
[`axum::extract::Path`](https://docs.rs/axum/latest/axum/extract/struct.Path.html) forces any percent encoded parameters decoded, that causes a url containing encoded semicolon `%3B` changed into decoded `;`, and c8y responses with an error.

 > Any percent encoded parameters will be automatically decoded. The decoded parameters must be valid UTF-8, otherwise Path will fail and return a 400 Bad Request response.

This PR uses a workaround to avoid using `axum::extract::Path`.

## Types of changes

## Paste Link to the issue
#3884

## Checklist

### DIFF
--- a/crates/extensions/c8y_auth_proxy/src/server.rs
+++ b/crates/extensions/c8y_auth_proxy/src/server.rs
@@ -3,7 +3,6 @@ use anyhow::Context;
 use axum::extract::ws::rejection::WebSocketUpgradeRejection;
 use axum::extract::ws::WebSocket;
 use axum::extract::FromRef;
-use axum::extract::Path;
 use axum::extract::State;
 use axum::extract::WebSocketUpgrade;
 use axum::http::HeaderValue;
@@ -395,17 +394,20 @@ async fn respond_to(
     State(host): State<TargetHost>,
     State(client): State<reqwest::Client>,
     retrieve_token: State<SharedTokenManager>,
-    path: Option<Path<String>>,
     uri: hyper::Uri,
     method: Method,
     mut headers: HeaderMap<HeaderValue>,
     ws: Result<WebSocketUpgrade, WebSocketUpgradeRejection>,
     small_body: crate::body::PossiblySmallBody,
 ) -> Result<Response, ProxyError> {
-    let path = match &path {
-        Some(Path(p)) => p.as_str(),
-        None => "",
-    };
+    // Workaround to avoid using `axum::extract::Path`
+    // as it always decodes percent encoded characters
+    let base = uri
+        .path()
+        .strip_prefix("/c8y")
+        .expect("The path must start with /c8y");
+    let path = base.strip_prefix("/").unwrap_or(base);
+
     let auth: fn(reqwest::RequestBuilder, &str) -> reqwest::RequestBuilder =
         if headers.contains_key(AUTHORIZATION) {
             |req, _auth_value| req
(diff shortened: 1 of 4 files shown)


crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -395,17 +394,16 @@ async fn respond_to(
...
     ws: Result<WebSocketUpgrade, WebSocketUpgradeRejection>,
     small_body: crate::body::PossiblySmallBody,
 ) -> Result<Response, ProxyError> {
-    let path = match &path {
-        Some(Path(p)) => p.as_str(),
-        None => "",
-    };
+    // Workaround to avoid using `axum::extract::Path`
+    // as it always decodes percent encoded characters
+    let path = uri.path().strip_prefix("/c8y/").unwrap_or_default();
```
jarhodes314: I don't love the use of `.unwrap_or_default` here. While this does work, if the prefix we strip here was changed so it slightly mismatched the value in the router, the proxy would subtly break, forwarding every request to `/`.

I think the solution to this problem is to strip `/c8y`, then optionally trim the remaining `/` (or potentially just not add the remaining `/` back in)

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -988,6 +986,28 @@ mod tests {
         assert_eq!(res.status(), 200);
     }
 
+    #[tokio::test]
+    async fn sends_percent_encoded_string_request_3884() {
```
jarhodes314: While this test represents an instance of the bug we are fixing here, I think we could do with another test that verifies the case where the client sends `;` in the request path, and we check the proxy uses `;`. This will check the true goal of this change, which is that the proxy is transparent (and ensure we're not simply blindly percent-encoding the path segments).

Side note: this would be a really nice candidate for a property-based test, where we check the target path always corresponds to the input path.

jarhodes314 (commented):
Nice fix. A couple of suggested improvements, but the changes are broadly good.

While we're fixing this, I suspect we've got a similar potential bug with the handling of query parameters, it would be good to check if that's the case and handle it similarly if so.

crates/extensions/c8y_auth_proxy/src/server.rs
rina23q: Fixed in 46b34663bf2326b877d5fc6c029503eae04d3e30

crates/extensions/c8y_auth_proxy/src/server.rs
rina23q: I added more symbols in the unit test in 46b34663bf2326b877d5fc6c029503eae04d3e30. Do we need proptest additionally?

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -395,17 +394,20 @@ async fn respond_to(
...
-    let path = match &path {
-        Some(Path(p)) => p.as_str(),
-        None => "",
-    };
+    // Workaround to avoid using `axum::extract::Path`
+    // as it always decodes percent encoded characters
+    let base = uri
+        .path()
+        .strip_prefix("/c8y")
+        .unwrap_or_else(|| uri.path());
```
jarhodes314: I'd still prefer a "normal" `unwrap` or an `expect` here. My reasoning for this is that our expectation is this route will only be triggered if the path starts with `/c8y`, so if that assumption is violated, there's probably a bug in our code and we aren't going to be able to derive the correct path. As a result, I think panicking (which I believe will be converted into a  500 Internal Server Error response by axum) is a perfectly reasonable thing to do here. If we just fall back to `uri.path()`, I expect if this has a bug, the symptoms will now be a 404 response from Cumulocity as the proxy path is never going to match the path we request from Cumulocity

crates/extensions/c8y_auth_proxy/src/server.rs
jarhodes314: The updated code looks fine. I don't think there's a particular _need_ to use proptest here, it just struck me that this is quite a good example of something that looks like it could be relatively easily proptested.

crates/extensions/c8y_auth_proxy/src/server.rs
rina23q: Changed to `expect` in 65eadf76d08dccef43a33705508398229ca2469b

PR #3889, opened 2025-12-08
refactor: Deserialise separate mapper config file using existing logic for tedge.toml

jarhodes314:
## Proposed changes
Use the existing `TEdgeConfigC8yDto` (etc.) structs to deserialise separate mapper config files, with a view to supporting the `tedge config` CLI fully.

### Still TODO
- [x] Environment variable support
- [x] Add test for environment variable support
- [x] Full `tedge config` support
- [x] Ensure new profiles are created using migrated mapper config if relevant
- [ ] A tool to perform the initial migration from tedge.toml to separate mapper config
- [ ] ensure we handle default first cloud connections sensibly and can set things to later default to separate config files 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/compat.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/compat.rs
@@ -1,3 +1,15 @@
+//! Compatibility layer for converting tedge_config data into a [MapperConfig]
+//!
+//! The full flow for reading all mapper configs (regardless of whether sourced
+//! from tedge.toml or a separate file) is:
+//!
+//! - Read tedge.toml into a `TEdgeConfigDto`
+//! - Update `TEdgeConfigDto` with any migrated mapper configs
+//! - When filling the migrated configs in `TEdgeConfigDto`, store the config
+//!   path in the `mapper_config_dir` hidden field. This will be used if we
+//!   write back to the toml to ensure things get written to the correct file
+//! - When a mapper accesses the mapper config, they do this through the generic
+//!   [MapperConfig] type, which is done via this `compat` layer
 use super::*;
 use crate::tedge_toml::tedge_config::TEdgeConfigReaderAws;
 use crate::tedge_toml::tedge_config::TEdgeConfigReaderAz;
@@ -6,9 +18,6 @@ use crate::tedge_toml::ReadableKey;
 use crate::TEdgeConfig;
 
 /// Trait for creating cloud-specific mapper configuration from tedge.toml cloud sections
-///
-/// This trait enables backward compatibility by loading the new `MapperConfig<T>` format
-/// from the legacy c8y/az/aws configuration sections in tedge.toml.
 pub trait FromCloudConfig: Sized {
     /// The corresponding TEdgeConfigReader type for this cloud
     type CloudConfigReader: CloudConfigAccessor + Send + Sync + 'static;
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
@@ -468,6 +468,7 @@ fn keys_in_inner(prefix: &str, table: &toml::map::Map<String, toml::Value>) -> V
 mod tests {
     use crate::models::AbsolutePath;
     use crate::tedge_toml::mapper_config::AzMapperSpecificConfig;
+    use crate::tedge_toml::mapper_config::C8yMapperSpecificConfig;
     use crate::tedge_toml::Cloud;
     use once_cell::sync::Lazy;
     use tedge_config_macros::ProfileName;
@@ -633,6 +634,26 @@ type = "a-service-type""#;
         env.set_var("TEDGE_DIAG_PLUGIN_PATHS", "[\"test\",\"values\"]");
         let config = t.load().await.unwrap();
         assert_eq!(config.diag.plugin_paths, ["test", "values"]);
+    }
+
+    #[tokio::test]
+    async fn environment_variables_are_read_with_migrated_mapper_config() {
+        let (ttd, t) = create_temp_tedge_config("").unwrap();
+        ttd.dir("mappers")
+            .file("c8y.toml")
+            .with_raw_content("url = \"example.com\"");
+        let mut env = EnvSandbox::new().await;
+        env.set_var("TEDGE_C8Y_ROOT_CERT_PATH", "/root/cert/path");
+        let config = t.load().await.unwrap();
+        let c8y_config = config
+            .mapper_config::<C8yMapperSpecificConfig>(&None::<ProfileName>)
+            .unwrap();
+        assert_eq!(
+            c8y_config.http().or_none().unwrap().host().to_string(),
+            "example.com",
+            "Verify that c8y.toml file has actually been read"
+        );
+        assert_eq!(c8y_config.root_cert_path.to_string(), "/root/cert/path");
     }
 
     #[tokio::test]


crates/core/tedge/src/cli/config/commands/add.rs
```diff
@@ -20,7 +20,7 @@ impl Command for AddConfigCommand {
     }
 
     async fn execute(&self, tedge_config: TEdgeConfig) -> Result<(), MaybeFancy<anyhow::Error>> {
-        restrict_cloud_config_update("add", &self.key, &tedge_config).await?;
+        restrict_cloud_config_access("add", &self.key, &tedge_config).await?;
```
jarhodes314: At the moment, all the `tedge config` commands are still as before, however I intend to fix that (it's just slightly tricky figuring out exactly how to achieve the fine details, but the broad "generate a TEdgeConfigDto" for the mapper config aspect is working and in use for deserialisation).

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/compat.rs
```diff
@@ -1,17 +1,48 @@
...
+    }
+}
+
+fn cloud_type_for(key: Cow<'static, str>) -> Option<(CloudType, Option<ProfileName>)> {
+    match key.split_once(".") {
+        Some(("c8y", rest)) => Some((CloudType::C8y, extract_profile_name(rest))),
+        Some(("az", rest)) => Some((CloudType::Az, extract_profile_name(rest))),
+        Some(("aws", rest)) => Some((CloudType::Aws, extract_profile_name(rest))),
+        _ => None,
+    }
```
didier-wenzek: It's a nice idea to move this logic out of `restrict_cloud_config_acces`. We have now two independent steps clearly separated.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -2113,7 +2117,8 @@ mod tests {
         let error = tedge_config
             .mapper_config::<C8yMapperSpecificConfig>(&profile_name(None))
             .await
-            .unwrap_err();
+            .err()
```
albinsuresh: I see that `MapperConfig` not being `Debug` prompted this change. Any reason why we can't add the `Debug` bound to `SpecialisedCloudConfig` so that `MapperConfig` can also be `Debug`?

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -81,40 +82,35 @@ impl DeviceConfig {
...
 pub trait SpecialisedCloudConfig:
-    Sized
-    + DeserializeOwned
-    + ApplyRuntimeDefaults
-    + ExpectedCloudType
-    + FromCloudConfig
-    + Send
-    + Sync
-    + 'static
+    Sized + ExpectedCloudType + FromCloudConfig + Send + Sync + 'static
```
albinsuresh, line 100:
```suggestion
    Sized + ExpectedCloudType + FromCloudConfig + Send + Sync + std::fmt::Debug + 'static
```
So that the `MapperConfig` can also be `Debug`.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -130,169 +126,97 @@ pub struct MapperConfig<T: SpecialisedCloudConfig> {
...
-
-#[derive(Debug, Deserialize, Document, Default)]
-pub struct PartialMqttConfig {
-    #[serde(default)]
-    pub max_payload_size: Option<MqttPayloadLimit>,
-}
-
-#[derive(Debug, Document)]
-pub struct MapperMapperConfig<M> {
+pub struct MapperMapperConfig {
```
albinsuresh:
```suggestion
pub struct CommonMapperConfig {
```

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -81,40 +82,35 @@ impl DeviceConfig {
...
     /// The topic prefix for the bridge MQTT topic
     pub topic_prefix: Keyed<TopicPrefix>,
 
     /// The amount of time after which the bridge should send a ping
     pub keepalive_interval: SecondsOrHumanTime,
+}
 
-    pub include: BridgeIncludeConfig,
+pub struct C8yBridgeConfig {
+    pub include: C8yBridgeIncludeConfig,
```
albinsuresh, line 95: This could have been left as a generic mapper config, no? As it is about the local bridge connection which is valid for any mapper, whether they use it or not?

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -307,276 +231,253 @@ impl<T: PartialEq> PartialEq<T> for Keyed<T> {
...
-            timestamp: default_timestamp(),
-            timestamp_format: default_timestamp_format(),
-        }
-    }
-}
-
 /// Cumulocity-specific mapper configuration fields
-#[derive(Debug, Deserialize, Document)]
-#[serde(default)]
 pub struct C8yMapperSpecificConfig {
```
albinsuresh, line 387: The vision is to remove the duplication between this struct and the `TEdgeConfigDtoC8y` generated by the `define_tedge_config!` macro in a subsequent step, right? Or this is the end?

jarhodes314: Yes, this should be the case. I think my ideal vision at this stage is for the shared configurations to be defined as a single `mapper` group in the `define_tedge_config!` macro to avoid duplication, and that will define the separate cloud configurations as well as the generalised mapper config that can be used internally to remove duplication in the various mappers.

At the moment, I think I am tending towards a solution where `TEdgeConfig` is proactively populated with the mapper configuration tomls where appropriate, as opposed to reading them on demand, as that makes implementing `tedge config` commands far simpler, and reduces the amount of code split we need between the handling the two ways of configuring mappers.

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -81,40 +84,70 @@ impl DeviceConfig {
...
+        dto: Self::CloudDto,
+        base_config: &TEdgeConfig,
+        profile: Option<&str>,
+    ) -> Self::CloudConfigReader;
+}
+
+pub trait HasPath {
+    fn set_path(&mut self, path: Utf8PathBuf);
+    fn get_path(&self) -> Option<&Utf8Path>;
+}
```
didier-wenzek, line 131: Using this trait to assign an origin to a mapper setting is a nice idea.

I was initially a bit confused by the `tedge_config` fields (such as `c8y.read_from`) which are not populated by serde but by `TEdgeConfigDto::populate_single_mapper` *after* the source file has been serialized.

I still have some confusion on the global `tedge_config` which is manipulated as a single giant struct but populated by independent readers that use sub-type such as `C8yMapperSpecificConfig::CloudDto`. I fail to understand how the knot is tied.

Also, what's the plan for the `compat` layer? Is that layer still useful?

jarhodes314: At the moment, the `compat` layer is used for everything that reads the `MapperConfig<T>` struct. The full flow goes:

- Read tedge.toml into a `TEdgeConfigDto`
- Update `TEdgeConfigDto` with any migrated mapper configs (/possibly fail if there is a conflict, although I'm not entirely sure if that's actually the case)
- When filling the migrated configs in `TEdgeConfigDto`, store the config path in the `read_from` hidden field. This will be used if we write back to the toml to ensure things get written to the correct file
- If a mapper accesses the mapper config, they do this through the generic `MapperConfig` type, which is done via the `compat` layer.
 
There is definitely at least some code in the `mapper_config` module that can be tidied up. If we want to keep the concept of generic mapper config for properties shared by the different mappers, possibly this is best achieved through a simple accessor trait with methods like `root_cert_path()`, `bridge_built_in()` etc. implemented for `TEdgeConfigDtoC8y` and the others.

Alternatively, now there is internal compatibility reading the mapper configs into `TEdgeConfig`, we could just undo the changes to the mappers and how they read configuration since they could plausibly read configuration direct from `tedge_config.c8y` etc. and that should just work now.

didier-wenzek: > The full flow goes:
> - Read tedge.toml into a `TEdgeConfigDto`
> - Update `TEdgeConfigDto` with any migrated mapper configs (/possibly fail if there is a conflict, although I'm not entirely sure if that's actually the case)
> - When filling the migrated configs in `TEdgeConfigDto`, store the config path in the `read_from` hidden field. This will be used if we write back to the toml to ensure things get written to the correct file
> - If a mapper accesses the mapper config, they do this through the generic `MapperConfig` type, which is done via the `compat` layer.

This is a clear description of tedge-config read flow. It deserves to be promoted as a doc comment. That said, I don't have a precise idea where would be the best place for this comment 

> If we want to keep the concept of generic mapper config for properties shared by the different mappers,  [...]
> Alternatively, now there is internal compatibility reading the mapper configs into `TEdgeConfig`, [...]

As both paths seem a bit involved, it would be good to discuss this from a DX point of view first. What will have a thin-edge dev to write to add a new mapper config option?

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: It could have been, however the fact it wasn't c8y-specific was simply an artefact of the deserialisation being made significantly more complicated previously.

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -514,17 +504,15 @@ impl BuildCommand for TEdgeCertCli {
...
 
                     let cloud_config = match cloud.as_ref() {
-                        Some(c) => Some(config.as_cloud_config(c.into()).await?),
+                        Some(c) => Some(config.as_cloud_config(c.into())?),
                         None => None,
                     };
                     let cryptoki = config
                         .device
-                        .cryptoki_config(cloud_config.as_ref().map(|c| c as &dyn CloudConfig))?;
+                        .cryptoki_config(cloud_config.as_ref().map(|c| &**c as &dyn CloudConfig))?;
```
didier-wenzek: This is a detail and a matter of taste, I agree, but I would prefer:

```suggestion
                        .cryptoki_config(cloud_config.as_ref().map(|c| c.as_ref() as &dyn CloudConfig))?;
```

However, I'm surprised that the following is not compiling:

```
512 |                         .cryptoki_config(cloud_config.as_ref().map(|c| c.as_ref()))?;
    |                          --------------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait `tedge_config::tedge_toml::CloudConfig`, found trait `tedge_config::tedge_toml::CloudConfig + std::marker::Send + std::marker::Sync`                                                                                                                          
    |                          |
    |                          arguments to this method are incorrect
    |
    = note: expected enum `std::option::Option<&dyn tedge_config::tedge_toml::CloudConfig>`
               found enum `std::option::Option<&dyn tedge_config::tedge_toml::CloudConfig + std::marker::Send + std::marker::Sync>`
```

Why extra traits are breaking the expectations? Is this because of virtual table constraints? `&dyn A+B` having to combine the virtual tables of `A` and `B`? Even for *marker* traits such as `Send` and `Sync`?

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: This won't help. If you `#[derive(Debug)]` for a generic struct, it will automatically apply a `Debug` constraint for all generic parameters (which is why we sometimes need to implement `Clone` and `Debug` manually, e.g. implementing `Clone` for a type containing `Arc<T>` (since we don't need `T: Clone` to clone `Arc<T>`)). For example:

```rust
#[derive(Debug)]
struct Wrapper<T>(T);

// will expand to
impl<T: std::fmt::Debug> std::fmt::Debug for Wrapper<T> {
    // ...
}
```

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
jarhodes314: Now done

crates/core/tedge/src/cli/certificate/cli.rs
jarhodes314: Now fixed by changing the receiving method to accept `&(impl CloudConfig + ?Sized)` rather than `&dyn CloudConfig`, effectively making it generic over `CloudConfig`. That meant the conversion here merely became `cloud_config.as_deref()`

Even with marker traits, the vtables for `dyn` objects in Rust are separate for each and every `dyn` trait in use, so `A + B` counts as a different trait to `A`, and thus `dyn A + B` cannot be dereferenced to `dyn A` (but you can cast from `dyn A + B` to `dyn A` as you do have all the information to construct the new vtable, but this has to be done explicitly).

Issue #3890, opened 2025-12-09
`tedge cert create` should use HSM configuration


Bravo555:
**Is your feature request related to a problem? Please describe.**
Currently, HSM configuration is used when generating CSRs, but not when generating self-signed certificates, i.e. `tedge cert download` will use HSM, but `tedge cert create` will not.

**Describe the solution you'd like**
`tedge cert create` should use HSM configuration to sign the self-signed certificate.

**Describe alternatives you've considered**

**Additional context**

PR #3891, opened 2025-12-09
feat: `tedge cert create` use HSM configuration

Bravo555:
## TODO

- [x] handle scenario when there's no key on the HSM
- [x] handle other errors (tedge-p11-server errors, etc.)

## Proposed changes

Make `tedge cert create` command use the current HSM configuration, i.e. if we're currently using a PKCS11-based private key, sign the self-signed certificate using that key. Otherwise the certificate is broken as it's signed with a different private key that doesn't correspond to the public key info in the certificate.

Normally `tedge cert create` persists to disk newly generated certificate and the private key, but now, if already existing PKCS11-based private key is used, the private key is not persisted (because it can't be).   

## Types of changes

## Paste Link to the issue

- #3890 

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/certificate/cli.rs
+++ b/crates/core/tedge/src/cli/certificate/cli.rs
@@ -38,6 +38,11 @@ pub enum TEdgeCertCli {
     ///
     /// This command creates the device certificate and private key, and persists them, if they
     /// don't already exist.
+    ///
+    /// If thin-edge is configured to use private key from an HSM (see
+    /// https://thin-edge.github.io/thin-edge.io/references/hsm-support), the key needs to be
+    /// present on the token (see `tedge cert create-key-hsm` command if you need to create it) and
+    /// selected. After invoking the command, only the new certificate will be persisted.
     Create {
         /// The device identifier to be used as the common name for the certificate
         #[clap(long = "device-id", global = true)]


tests/RobotFramework/tests/pkcs11/tedge_cert_create/initial.robot
```diff
@@ -0,0 +1,13 @@
+*** Settings ***
+Resource        tedge_cert_create.resource
+
+Suite Setup     tedge-p11-server Setup    ${TEDGE_P11_SERVER_VERSION}
+
+
+*** Variables ***
+${TEDGE_P11_SERVER_VERSION}     ${EMPTY}
```
didier-wenzek: We will have to update that line with the correct version once the PR actually merged.

crates/core/tedge/src/cli/certificate/create.rs
```diff
@@ -53,8 +57,23 @@ impl Command for CreateCertCmd {
 }
 
 impl CreateCertCmd {
-    pub async fn create_test_certificate(&self, config: &CsrTemplate) -> Result<(), CertError> {
-        let cert = KeyCertPair::new_selfsigned_certificate(config, &self.id, &KeyKind::New)?;
+    // test certificate here means it's not yet permanent - establishing the connection using this
+    // new certificate will be tested during `tedge connect` and if an error happens, we will revert
+    // to the previous certificate. If not, the certificate will be set as a permanent active
+    // certificate and will not be tested on further connections.
+    pub async fn create_test_certificate(
```
didier-wenzek: Actually, this method name refers to a test certificate for a different reason. The point is to make clear that a self-signed certificate should not be used in prod (unless a tenant has a really low count of devices attached to).

crates/core/tedge/src/cli/certificate/create.rs
```diff
@@ -66,15 +85,18 @@ impl CreateCertCmd {
...
-        persist_new_private_key(
-            key_path,
-            cert.private_key_pem_string()?,
-            &self.user,
-            &self.group,
-        )
-        .await
-        .map_err(|err| err.key_context(key_path.clone()))?;
+        // only persist to filesystem if we just created a new key (not using HSM)
+        if let KeyKind::New = key_kind {
```
didier-wenzek: Instead of one method with an optional `CryptokiConfig` and internal HSM/not-HSM cases, I would consider to have two separate methods, one unchanged and another one for the HSM case.

tests/RobotFramework/tests/pkcs11/tedge_cert_create/initial.robot
Bravo555: Since there were no changes to `tedge-p11-server`, I actually replaced the version with `1.7.0`, which works.

crates/core/tedge/src/cli/certificate/create.rs
Bravo555: Got it. I'm still not entirely sure if calling it "test certificate" in code conveys the meaning exactly (seems like a recommendation to the user, but the user probably won't be reading the source code), but since I've missed the mark on that one, I'll remove that comment.

crates/core/tedge/src/cli/certificate/create.rs
Bravo555: fixed in 92fb7db

crates/core/tedge/src/cli/certificate/create.rs
Bravo555: fixed in 71abdb0

PR #3892, opened 2025-12-10
feat: az mapper defined using regular flows

didier-wenzek:
## Proposed changes

__breaking change__: The flows definitions for `tedge-mapper flows` are now under `/etc/tedge/mappers/flows/flows` (and no more under  `/etc/tedge/flows`). This has been done to have a consistent location across the mappers.

- [x] Tedge flows can invoke transformation functions implemented in Rust
- [x] Tedge flows provide builtin transformations (such as add-timestamp or set-topic)
- [x] A mapper can provide its own builtin transformations
- [x] A mapper can define builtin flows, which definition is persisted on disk and can be updated and disabled by users
- [x] The az mapper is defined using a builtin flow that can be updated and disabled by users
- [x] The flow-based az mapper is fully compatible with the legacy version
- [x] (test) az mapper is backward compatible
- [x] (test) Update / disable / restore az builtin workflow
- [x] (doc) How to update / disable / restore a builtin workflow
- [x] (refactoring) Remove dead code
- [x] Tedge flows definitions are stored under `/etc/tedge/mappers/$NAME.$PROFILE/flows`
- [x] `tedge flows config-dir` displays the config directory for a mapper profile

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/az_mapper_ext/src/lib.rs
+++ b/crates/extensions/az_mapper_ext/src/lib.rs
@@ -51,7 +51,7 @@ input.mqtt.topics = {input_topics}
 steps = [
     {{ builtin = "skip-mosquitto-health-status" }},
     {timestamp_step}
-    {{ builtin = "cap-payload-size", config = {{ max_size = {max_size} }} }},
+    {{ builtin = "limit-payload-size", config = {{ max_size = {max_size} }} }},
 ]
 
 output.mqtt.topic = "{output_topic}"
--- a/crates/extensions/tedge_flows/src/transformers/mod.rs
+++ b/crates/extensions/tedge_flows/src/transformers/mod.rs
@@ -7,8 +7,8 @@ use std::collections::HashMap;
 use std::time::SystemTime;
 
 mod add_timestamp;
-mod cap_payload_size;
 mod ignore_topics;
+mod limit_payload_size;
 mod set_topic;
 
 pub trait Transformer: Send + Sync + 'static {
@@ -51,7 +51,7 @@ impl Default for BuiltinTransformers {
             transformers: HashMap::default(),
         };
         transformers.register(add_timestamp::AddTimestamp::default());
-        transformers.register(cap_payload_size::CapPayloadSize::default());
+        transformers.register(limit_payload_size::LimitPayloadSize::default());
         transformers.register(ignore_topics::IgnoreTopics::default());
         transformers.register(set_topic::SetTopic::default());
         transformers
--- a/docs/src/references/mappers/builtin-flows.md
+++ b/docs/src/references/mappers/builtin-flows.md
@@ -45,7 +45,7 @@ input.mqtt.topics = ["te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+
 steps = [
     { builtin = "skip-mosquitto-health-status" },
     { builtin = "add-timestamp", config = { property = "time", format = "unix", reformat = true } },
-    { builtin = "cap-payload-size", config = { max_size = 262144 } },
+    { builtin = "limit-payload-size", config = { max_size = 262144 } },
 ]
 
 output.mqtt.topic = "az/messages/events/"
--- a/docs/src/references/mappers/flows.md
+++ b/docs/src/references/mappers/flows.md
@@ -328,18 +328,18 @@ Add a timestamp to JSON messages
   This can be changed with the `reformat` config so any timestamp is reformated to the requested format. 
 - `{ builtin = "add-timestamp", config = { format = "rfc3339", reformat = true }}`
 
-### `cap-payload-size`
-
-Filter out messages which payload is too large
-- Must be configured with the `max_size` for the messages (maximum number of bytes)
-- Can be configured to `discard` the messages instead of raising an error (the latter being the default)
-- `{ builtin = "cap-payload-size", config = { max_size = 64000, discard = true }}`
-
 ### `ignore-topics`
 
 Filter out messages with specific topics
 - Must be configured with a list of `topics` and topic filters to be ignored
 - `{ builtin = "ignore-topics", config.topics = ["te/device/main/service/mosquitto-c8y-bridge/#"] }`
+
+### `limit-payload-size`
+
+Filter out messages which payload is too large
+- Must be configured with the `max_size` for the messages (maximum number of bytes)
+- Can be configured to `discard` the messages instead of raising an error (the latter being the default)
+- `{ builtin = "limit-payload-size", config = { max_size = 64000, discard = true }}`
 
 ### `set-topic`
 
--- a/crates/extensions/tedge_flows/src/transformers/limit_payload_size.rs
+++ b/crates/extensions/tedge_flows/src/transformers/limit_payload_size.rs
@@ -6,14 +6,14 @@ use crate::Message;
 use std::time::SystemTime;
 
 #[derive(Clone, Default)]
-pub struct CapPayloadSize {
+pub struct LimitPayloadSize {
     max_size: Option<usize>,
     discard: bool,
 }
 
-impl Transformer for CapPayloadSize {
+impl Transformer for LimitPayloadSize {
     fn name(&self) -> &str {
-        "cap-payload-size"
+        "limit-payload-size"
     }
 
     fn set_config(&mut self, config: JsonValue) -> Result<(), ConfigError> {
--- a/tests/RobotFramework/tests/azure/updated_mea.toml
+++ b/tests/RobotFramework/tests/azure/updated_mea.toml
@@ -4,7 +4,7 @@ input.mqtt.topics = ["te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+
 steps = [
     { builtin = "skip-mosquitto-health-status" },
     { builtin = "add-timestamp", config = { property = "time", format = "unix", reformat = false } },
-    { builtin = "cap-payload-size", config = { max_size = 262144 } },
+    { builtin = "limit-payload-size", config = { max_size = 262144 } },
 ]
 
 output.mqtt.topic = "az/messages/events/"


tests/RobotFramework/tests/tedge_flows/context-flows/device_events.toml
```diff
@@ -1,6 +1,6 @@
 input.mqtt.topics = ["te/+/+/+/+/e/+"]
 
 steps = [
-    { script = "add_timestamp.js" },
+    { builtin = "add-timestamp" },
```
didier-wenzek, line 4: This is an example of a flow combining builtin transformation steps with java script transformations.

crates/extensions/az_mapper_ext/src/converter.rs
```diff
@@ -142,6 +142,27 @@ impl AzureConverter {
...
+input.mqtt.topics = ["{te}/+/+/+/+/m/+","{te}/+/+/+/+/e/+","{te}/+/+/+/+/a/+", "{te}/+/+/+/+/status/health"]
+
+steps = [
+    {{ builtin = "add-timestamp", config = {{ property = "time", format = "{time_format}" }} }},
+    {{ builtin = "cap-payload-size", config = {{ max_size = {max_size} }} }},
+]
+
+output.mqtt.topic = "{output_topic}"
+errors.mqtt.topic = "{errors_topic}"
+"#,
```
didier-wenzek: This is the builtin flow of the az mapper. This flow definition includes user settings and is persisted in `/etc/tedge/az/flows/mea.toml` and can be edited by users, adding / tuning / removing steps. Here is an example:

```
input.mqtt.topics = ["te/+/+/+/+/m/+","te/+/+/+/+/e/+","te/+/+/+/+/a/+", "te/+/+/+/+/status/health"]

steps = [
    { builtin = "add-timestamp", config = { property = "time", format = "unix" } },
    { builtin = "cap-payload-size", config = { max_size = 262144 } },
]

output.mqtt.topic = "az/messages/events/"
errors.mqtt.topic = "te/errors"
```

I still have to write the doc.

crates/extensions/tedge_flows/src/transformers/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+        timestamp: SystemTime,
+        message: &Message,
+        config: &JsonValue,
+    ) -> Result<Vec<Message>, FlowError>;
+
+    fn is_periodic(&self) -> bool {
+        false
+    }
+
+    fn on_interval(
```
albinsuresh: What's the built-in transformer that you have in mind that could be periodic?

crates/core/tedge_mapper/src/az/mod.rs
```diff
@@ -1 +1,38 @@
+use std::time::SystemTime;
+use tedge_flows::FlowError;
+use tedge_flows::JsonValue;
+use tedge_flows::Message;
+use tedge_flows::Transformer;
+
 pub mod mapper;
+
+#[derive(Clone, Default)]
+struct SkipMosquittoHealthStatus;
```
albinsuresh, line 11: This could be modelled as a built-in transformer as well, where messages can be filtered based on the target entity in the topic.

crates/extensions/tedge_flows/src/transformers/mod.rs
```diff
@@ -0,0 +1,159 @@
...
+use crate::LoadError;
+use crate::Message;
+use std::collections::HashMap;
+use std::time::SystemTime;
+
+mod add_timestamp;
+mod cap_payload_size;
+mod set_topic;
+
+pub trait Transformer: Send + Sync + 'static {
```
albinsuresh, line 14:
```suggestion
pub trait BuiltinStep: Send + Sync + 'static {
```
Just to avoid introducing yet another name/concept for the user, since the keyword in the flow definition is also `builtin`. Especially, if the answer to a user asking "what a transformer is ", would be: "it's a built-in step".

Otherwise, establish the notion of `steps` being `transformers`, either `builtin` or external(`script`), in the doc.

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -5,73 +5,41 @@ use crate::js_runtime::JsRuntime;
...
 pub struct JsScript {
     pub module_name: String,
+    pub flow: Utf8PathBuf,
     pub path: Utf8PathBuf,
-    pub config: JsonValue,
-    pub interval: Duration,
-    pub next_execution: Option<Instant>,
-    pub no_js_on_message_fun: bool,
-    pub no_js_on_interval_fun: bool,
+    pub is_defined: bool,
```
albinsuresh, line 16:
```suggestion
    pub is_transformer: bool,
```

crates/extensions/az_mapper_ext/src/lib.rs
```diff
@@ -1,3 +1,66 @@
...
+            add_timestamp,
+            time_format,
+            size_threshold,
+        }
+    }
+
+    pub fn builtin_flow(&self) -> String {
+        let timestamp_step = if self.add_timestamp {
+            format!(
+                r#"{{ builtin = "add-timestamp", config = {{ property = "time", format = "{time_format}" }} }},"#,
```
albinsuresh: Optional:
```suggestion
                r#"{{ builtin = "add-timestamp", config = {{ property_name = "time", format = "{time_format}" }} }},"#,
```
Or fragment_name or fragment_key.

crates/extensions/az_mapper_ext/src/lib.rs
```diff
@@ -1,3 +1,66 @@
...
+        };
+
+        format!(
+            r#"
+input.mqtt.topics = {input_topics}
+
+steps = [
+    {{ builtin = "skip-mosquitto-health-status" }},
+    {timestamp_step}
+    {{ builtin = "cap-payload-size", config = {{ max_size = {max_size} }} }},
```
albinsuresh: Optional:
```suggestion
    {{ builtin = "limit-payload-size", config = {{ max_size = {max_size} }} }},
```

albinsuresh (commented):
A really good showcase of the flows mapper. I've just got a few minor suggestions, most being optional.

crates/extensions/az_mapper_ext/src/lib.rs
albinsuresh: I'm assuming that this `config` is not just a temporary arrangement while `tedge config` is not exposed to tha flows, but can be used to override the values from `tedge config` even when we expose the same to these builtin flows/transformers.

crates/extensions/tedge_flows/src/transformers/mod.rs
didier-wenzek: I see no reasons to give less power to built-in transformers. Monitoring & limiting message rate is an example of a  generic transformer that would be useful.

crates/extensions/tedge_flows/src/transformers/mod.rs
didier-wenzek: > `s/Transformer/BuiltinStep/g`

Strictly speaking, a step is a transformer *instance*: the same code may be used with different config by several flows. This is why, I'm not inclined to rename a transformer to a `BuiltinStep`.

That said, I agree that we must avoid introducing yet another concept for the user.

Also, I'm not happy with these names of `Transformer` vs `JsScript` currently used for very similar concepts which difference is only the programming language: Rust vs JavaScript.

 I have considered:
- `RsScript` vs  `JsScript`
- `RsTransformer` vs  `JsTransformer`
- `RsTransformation` vs  `JsTransformation`

There are also requests to introduce the concept of *smart functions*:
 
- `BuiltinSmartFunction` vs  `SmartFunction`

> Otherwise, establish the notion of `steps` being `transformers`, either `builtin` or external(`script`), in the doc.

No, this would expose the word `transformer` which is currently uniquely used internally. Also, the concept of *step* in a *multi-steps transformation* is  self-explanatory and I really prefer to keep it.

crates/extensions/tedge_flows/src/js_script.rs
didier-wenzek: No, this flag is actually only used for the case a script doesn't export an `onMessage()` function. We can even consider here to reject the script and raise an error.

crates/extensions/az_mapper_ext/src/lib.rs
didier-wenzek: More precise and correct, I agree. But a bit too lengthy while I don't think there is any ambiguities.

didier-wenzek: Yeah. The point is to be backward compatible with the current az mapper which behavior is configured by the `tedge config`.

crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -49,18 +49,37 @@ impl JsRuntime {
...
+        &mut self,
+        script: &mut JsScript,
+        source: impl Into<Vec<u8>>,
+    ) -> Result<(), LoadError> {
+        let exports = self.load_js(script.module_name.to_owned(), source).await?;
+        Self::set_exports(script, exports);
+        Ok(())
+    }
+
+    fn set_exports(script: &mut JsScript, exports: Vec<&str>) {
```
jarhodes314: Minor: We definitely don't require taking ownership of a `Vec<_>` here

```suggestion
    fn set_exports(script: &mut JsScript, exports: &[&str]) {
```

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// This name is built after : the flow name, the script name (or builtin transformer)
+    /// and the index that step among all the steps of the flow (so two instances of the same script
+    /// in the same flow are given different instance name).
+    pub fn instance_name(flow: impl ToString, script: impl ToString, index: usize) -> String {
+        format!(
+            "{flow}|{index}|{script}",
+            flow = flow.to_string(),
+            script = script.to_string()
+        )
+    }
```
jarhodes314: If you depend on `std::fmt::Display` (which is pretty much guaranteed to be implemented for anything that implements `ToString`, the format! invocation can be greatly simplified

```suggestion
    pub fn instance_name(flow: impl Display, script: impl Display, index: usize) -> String {
        format!("{flow}|{index}|{script}")
    }
```

crates/extensions/tedge_flows/src/transformers/add_timestamp.rs
```diff
@@ -0,0 +1,58 @@
...
+        message: &Message,
+        config: &JsonValue,
+    ) -> Result<Vec<Message>, FlowError> {
+        let Ok(serde_json::Value::Object(json_message)) =
+            serde_json::from_slice(message.payload.as_slice())
+        else {
+            return Ok(vec![message.clone()]);
+        };
+
+        let time_property = config.string_property("property").unwrap_or("time");
```
jarhodes314: I find it somewhat surprising that the `Transformer` implementations parse the config on every incoming message. Is there a reason that these values might change between `on_message` invocations?

crates/extensions/tedge_flows/src/js_value.rs
```diff
@@ -53,6 +54,34 @@ impl JsonValue {
...
+    pub fn string_property(&self, property: &str) -> Option<&str> {
+        self.property(property).and_then(|v| match v {
+            JsonValue::String(string) => Some(string.as_str()),
+            _ => None,
+        })
+    }
+
+    pub fn number_property(&self, property: &str) -> Option<Number> {
+        self.property(property).and_then(|v| match v {
+            JsonValue::Number(n) => Some(n.clone()),
```
jarhodes314: What is the rationale behind cloning here? It's not an expensive clone (in fact, the private enum within the public `Number` struct in serde_json implements Copy: https://docs.rs/serde_json/latest/src/serde_json/number.rs.html#21-34), but it's a little suprising, particularly when we avoid cloning in `string_property`?

crates/extensions/tedge_flows/src/js_runtime.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/907420024ee6165033d8ef6029b0dbf0f96aabb5

crates/extensions/tedge_flows/src/steps.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/2e38683081a6f71880094dd9decf1fbb1f77b7d4

crates/extensions/tedge_flows/src/transformers/add_timestamp.rs
didier-wenzek: > Is there a reason that these values might change between on_message invocations?

No, these values can only be changed via the config file. I will fix that.

tests/RobotFramework/tests/azure/azure_telemetry.robot
```diff
@@ -116,14 +116,88 @@ Publish health status message for main device service
...
+    ...    date_from=${timestamp}
+    ...    minimum=1
+    ...    maximum=1
+
+Discard mosquitto health status
+    Execute Command    (sleep 1; tedge mqtt pub te/device/main/service/mosquitto-c8y-bridge/status/health 1)&
+    ${message}=    Execute Command
+    ...    tedge mqtt sub 'az/#' --duration 2s | grep -v '"status":"up"'
+    ...    ignore_exit_code=${True}
+    Should Be Empty    ${message}
```
albinsuresh:
```suggestion
    ${timestamp}=    Get Unix Timestamp
    Execute Command    tedge mqtt pub te/device/main/service/mosquitto-c8y-bridge/status/health 1
    Should Have MQTT Messages
        ...    c8y/s/us
        ...    message_contains="status":"up"
        ...    date_from=${timestamp}
        ...    minimum=0
        ...    maximum=0
```

It would have been even better to introduce a `Should Not Have MQTT Messages` keyword to better communicate the intent of that test.

tests/RobotFramework/tests/azure/azure_telemetry.robot
```diff
@@ -116,14 +116,88 @@ Publish health status message for main device service
...
     Execute Command    sudo systemctl restart tedge-mapper-az.service
     ThinEdgeIO.Service Health Status Should Be Up    tedge-mapper-az
 
 Custom Teardown
     Get Suite Logs
     Execute Command    sudo tedge config unset az.topics
+
+Update Builtin Flow
+    [Arguments]    ${flow_definition}
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/${flow_definition}    /etc/tedge/az/flows/mea.toml
```
albinsuresh: Seeing these different flow directories like `/etc/tedge/flows` and `/etc/tedge/az/flows`
and presumably the future ones like `/etc/tedge/c8y/flows` and `/etc/tedge/aws/flows` as well, it feels a bit fragmented directly under `/etc/tedge`. Would it be better to gather all the flows into a hierarchical structure as follows:
```
/etc/tedge/flows
|-- c8y
|-- az
|-- aws
```

Another factor that could influence this decision is the introduction of the `/etc/tedge/mapper` directory for the mapper configs. Since both the configs and the flows are mapper specific, may be we should put them both under the same roof like:

```
/etc/tedge/mapper
|-- c8y
|   |-- flows
|   |-- config
|-- az
| ...
|-- aws
| ...
```

Another question is whether we'd need to support flows that are defined once (at one location) but used by all cloud mappers. If yes, we need to accommodate that `common` directory as well.

docs/src/references/mappers/flows.md
```diff
@@ -306,3 +313,9 @@ The following builtin objects are exported:
 - [`TextDecoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder)
   - Only `utf-8` is supported 
 - [`TextEncoder`](https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder)
+
+## Builtin transformations
+
+- `{ builtin = "set-topic" }`: assign a topic to messages
+- `{ builtin = "add-timestamp" }`: add a timestamp to JSON messages
+- `{ builtin = "cap-payload-size" }`: filter out messages which payload is too large
```
albinsuresh: Including the `config` examples for each builtin transformer also would have been nice.

docs/src/references/mappers/builtin-flows.md
```diff
@@ -0,0 +1,84 @@
...
+- `az.mapper.timestamp`: whether the mapper should add a timestamp or not
+- `az.mapper.timestamp_format`: the date format to be used for the timestamp (`unix` or `rfc-3339`)
+- `az.mapper.mqtt.max_payload_size`: the maximum payload for a message (`262144` by default)
+- `az.bridge.topic_prefix`: the prefix used for the bridge local MQTT topic  (`az` by default)
+
+:::note
+For a change of one of these settings to be effective, the Azure mapper has to be restarted:
+```
+$ sudo tedge config set az.mapper.timestamp_format rfc-3339
+$ sudo systemctl restart tedge-mapper-c8y
```
albinsuresh:
```suggestion
$ sudo systemctl restart tedge-mapper-az
```

docs/src/references/mappers/builtin-flows.md
```diff
@@ -0,0 +1,84 @@
...
+:::note
+If the builtin flow is updated, then the Azure mapper will not override its content,
+even if some `tedge config` settings have been updated. In such a case, the Azure mapper only
+update its flow template, i.e. the file `/etc/tedge/az/flows/mea.toml.template`.
+:::
+
+If other flow definitions are provided along the builtin flow in the `/etc/tedge/az/flows/` directory,
+then these flows are loaded by the Azure mapper.
+
+Finally, the builtin flow is disabled when replaced by a file with the same name and a `.toml.disabled` extension
```
albinsuresh, line 85:
```suggestion
Finally, the builtin flow is disabled when replaced by a file with the same name and a `.toml.disabled` extension, or by deleting it.
```

That being said, I'm just wondering if the mapper should recreate the built-in flow `toml` file when it is missing.

crates/extensions/tedge_flows/src/js_value.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/cb6c4cab9455450967d84c9a42b4e60fa2108a2c

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -195,6 +238,49 @@ impl<T: FlowRegistry + Send> FlowRegistryExt for T {
...
+    ) -> Result<(), UpdateFlowRegistryError> {
+        let dir = self.store().config_dir();
+        let flow_path = dir.join(name).with_extension("toml");
+        let disabled_flow_path = flow_path.with_extension("toml.disabled");
+        let template_path = flow_path.with_extension("toml.template");
+
+        // Don't update the flow definition if overridden or disabled
+        let prior_flow = tokio::fs::read(&flow_path).await.ok();
+        let prior_template = tokio::fs::read(&template_path).await.ok();
+        let overridden = prior_flow != prior_template;
```
albinsuresh, line 255: If the `toml` file was accidentally deleted while the `toml.template` is still present, then I expected the mapper to recreate the built-in `toml` file with the same template content. But this check prevents that, as removal in interpreted as overridden. Was this intentional?

didier-wenzek: Yes, this is intentional. If the `.toml.template` file exists and differs from the associated `.toml` file (if any), then the `.toml` file is not updated (nor created if missing).

The idea is to let users override or remove a builtin flow.

If a user wants a builtin flow to be recreated or reset, she has to copy the   `.toml.template` file  into the associated `.toml` file. As an alternative, she can remove both the associated  `.toml.template` and `.toml` files and then restart the mapper.

docs/src/references/mappers/builtin-flows.md
didier-wenzek: > I'm just wondering if the mapper should recreate the built-in flow `.toml` file when it is missing.

I think it should not, unless the `.toml.template` file is also missing. Because this would introduce a difference between builtin and user-defined flows; the latter being disabled by removing a file, and the former by renaming a file.

tests/RobotFramework/tests/azure/azure_telemetry.robot
didier-wenzek: We definitely have to improve the hierarchy currently used.

docs/src/references/mappers/builtin-flows.md
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/8ce830f0edd407eb91f1e9c5fc9b3e3ddce6588b

docs/src/references/mappers/builtin-flows.md
didier-wenzek: Updated: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/8ce830f0edd407eb91f1e9c5fc9b3e3ddce6588b

crates/extensions/tedge_flows/src/transformers/add_timestamp.rs
didier-wenzek: Actually, this is a legacy of the very first API proposal where the configuration was updated by the `onConfigUpdate` method which has been deprecated meantime. With the current API, the config can be parsed only once when the transformer is built. I will update the `Transformer` trait to enable this.

didier-wenzek: Addressed by https://github.com/thin-edge/thin-edge.io/pull/3892/changes/7fbe6a5c95cd616920f336c26a35f5e2a832a22e

crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -49,18 +49,37 @@ impl JsRuntime {
     }
 
     pub async fn load_script(&mut self, script: &mut JsScript) -> Result<(), LoadError> {
-        let exports = self.load_file(script.module_name(), script.path()).await?;
+        let exports = self
+            .load_file(script.module_name.to_owned(), script.path())
+            .await?;
+        Self::set_exports(script, exports.as_slice());
```
jarhodes314: I'm fairly sure we can rely on `Deref` instead of explicitly calling `.as_slice()` here

```suggestion
        Self::set_exports(script, &exports);
```

crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -49,18 +49,37 @@ impl JsRuntime {
...
+        Ok(())
+    }
+
+    pub async fn load_script_literal(
+        &mut self,
+        script: &mut JsScript,
+        source: impl Into<Vec<u8>>,
+    ) -> Result<(), LoadError> {
+        let exports = self.load_js(script.module_name.to_owned(), source).await?;
+        Self::set_exports(script, exports.as_slice());
```
jarhodes314:
```suggestion
        Self::set_exports(script, &exports);
```

docs/src/references/mappers/builtin-flows.md
```diff
@@ -0,0 +1,87 @@
...
+---
+
+import ProposalBanner from '@site/src/components/ProposalBanner'
+
+<ProposalBanner/>
+
+## Concepts
+
+The behavior of the mappers provided out-of-the-box by %%te%% for Cumulocity, Azure and AWS, 
+is defined using [__tedge flows__](./flows.md) which definition can be adapted exactly as user defined flows:
```
jarhodes314:
```suggestion
The out-of-the-box mappers for Cumulocity, Azure, and AWS are defined as [tedge flows](./flows.md).
Their definitions can be customized in the same way as any user-defined flow:
```

docs/src/references/mappers/builtin-flows.md
```diff
@@ -0,0 +1,87 @@
...
+- Each builtin flow is materialized by two files:
+  - a flow definition with a `.toml` extension, which can be updated by users and defines the behavior of the mapper
+  - a copy of the builtin definition with a `.toml.template` extension,
+    which is not supposed to be updated and can be used as a reference when the active definition is updated.
+- These flow definitions - user-defined or builtin, updated or not, define the message transformations applied by the mapper.
+- When updating a builtin flow, the user can:
+  - Change step configurations
+  - Add, remove steps
+  - Combine builtin transformations with custom ones implemented in JavaScript.
+- A builtin flow can also be disabled and fully replaced by user-defined flows.
```
jarhodes314: I may have slightly misunderstood the "active definition" part here, but I think we can make this somewhat more concise:

```suggestion
- These definitions are persisted on disk alongside the user provided flows for this mapper

- Each mapper generates two files:
  - the "active" flow definition with a `.toml` extension
  - a copy of the generated definition with a `.toml.template` extension
  
  The active definition can be customised by the user.
  The template is intended to be used as a reference when the active definition is updated.

- This flow defines the message transformations applied by the mapper,
  regardless of whether the user customises the behaviour

- When updating a builtin flow, the user can:
  - Change step configurations
  - Add or remove steps
  - Combine builtin transformations with custom ones implemented in JavaScript

- A builtin flow can also be disabled and fully replaced by user-defined flows.
```

docs/src/references/mappers/builtin-flows.md
```diff
@@ -0,0 +1,87 @@
...
+```
+$ sudo tedge config set az.mapper.timestamp_format rfc-3339
+$ sudo systemctl restart tedge-mapper-az
+```
+:::
+
+This file can also be manually edited to:
+- change step configurations
+- add, remove steps
+- substitute a JavaScript implementation for the builtin transformation.
```
jarhodes314:
```suggestion
- Change step configurations
- Add or remove steps
- Substitute a JavaScript implementation for the builtin transformation.
```

crates/extensions/tedge_flows/src/js_runtime.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/22115f842cc65f5015f34cf18c9cbcdc09ba527b

crates/extensions/tedge_flows/src/js_runtime.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/22115f842cc65f5015f34cf18c9cbcdc09ba527b

docs/src/references/mappers/builtin-flows.md
didier-wenzek: Applied: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/fd2bfc19f8906b956efdd1b6cd9b01d7b98da5a1

docs/src/references/mappers/builtin-flows.md
didier-wenzek: Applied: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/fd2bfc19f8906b956efdd1b6cd9b01d7b98da5a1

docs/src/references/mappers/builtin-flows.md
didier-wenzek: Applied: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/fd2bfc19f8906b956efdd1b6cd9b01d7b98da5a1 with a slight modification to avoid the term *active*.

docs/src/references/mappers/flows.md
albinsuresh: I guess you missed this comment.

crates/core/tedge_mapper/src/az/mod.rs
albinsuresh: What I meant by this comment was that this az specific transformer could be modelled as a generic transformer that takes the target topic as the input and drop all messages on that topic, so that it can be used by any mapper/flow. Something like:

```
{ builtin = "ignore-topics", config = { topics = ["te/device/main/service/mosquitto-az-bridge/#", "te/device/main/service/builtin-bridge-az/#"] } }
```

didier-wenzek: Good point, we can have a more generic builtin transformer aimed to ignore a subset of topics. => I will add it.

However to be fully compatible with the legacy `az` mapper we need also to check the message payload.

docs/src/references/mappers/flows.md
didier-wenzek: Improved: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/cfdeb8d22c674f6b283860b5916360cc2a96e0cf

crates/core/tedge_mapper/src/az/mod.rs
didier-wenzek: Added an `ignore-topics` transformer https://github.com/thin-edge/thin-edge.io/pull/3892/changes/95738b16718b6f7fe5fe46149babe7c557e18446

Not used yet for the `az` mapper. For two reasons. First, to avoid any behavior changes. Second, to provide an example of a mapper that provides its own transformers.

tests/RobotFramework/tests/azure/azure_telemetry.robot
didier-wenzek: The directory `/etc/tedge/flows` is not meant to be permanent. It will likely be deprecated as well as the mapper specifically named `flows` in favor of user-configured mappers, each with specific name, topic prefix, flows and config as well as specific bridge mappings and certificates.

=> Moving to a hierarchy with a config directory per mapper is a good move.

From a flow perspective, the proposed hierarchy (`/etc/tedge/mapper/$CLOUD/flows`) would work.
However:

- We also need to take care of the cloud profiles
- It would be good for a cloud profile to inherit flows from the main mapper
- A mapper is not necessarily related to a cloud (e.g. the `collectd` mapper)
- Having config directories for mapper config (`/etc/tedge/mapper/$CLOUD/config`) is foreign to what we have now (`/etc/tedge/mapper/c8y.toml` and `/etc/tedge/mapper/c8y.d/$PROFILE.toml`)
- We need to be able to extend this in the future to store certificates specific to a cloud profile or user-provided bridge rules.

What about the following:

```
/etc/tedge/mapper/
|-- c8y/
|   |-- tedge.toml
|   |-- flows/
|   |-- certs/
|-- c8y.foo/     # a c8y mapper for the cloud profile foo
|   |-- tedge.toml
|   |-- flows/
|   |-- certs/
|-- az/
| ...
|-- aws/
| ...
|-- eventgrid/  # a mapper that is not natively supported
|   |-- tedge.toml
|   |-- bridge.toml
|   |-- flows/
|   |-- certs/
|-- collectd/   # a mapper that is not connected to the cloud
|   |-- flows/
```

- For flows, this would work (with some small complexities to let a cloud profile inherit flows)
- @jarhodes314 what would be the amount of work to adapt the mapper config along these lines? Do you see blockers or better ideas?
- @reubenmiller, @albinsuresh Do you see blockers or better ideas?
- It would be good to have an agreement on this as well as an implementation of  the target flows hierarchy before this PR been merged.

jarhodes314: I like this proposal, particularly the way it puts profiled and non-profiled configurations at the same depth in the directory structure.

I'm slightly unclear about the details of how we handle custom mappers like eventgrid vs collectd (which is not a custom mapper but isn't cloud connected). However, I think this is a point we can refine when we come to migrate the collectd mapper to use flows like we do here for the az mapper.

reubenmiller: The structure makes sense to me. Though just a few open question:

* Where do we put flows are are generic (e.g. they are an extension to the te topics), so they aren't specifically related to a cloud mapper. I'd be happy to keep the `tedge-mapper flows` to run these flows individually (or would this be a job for the tedge-agent?)
* how do we manage the installation and removal of flows, as the command will also need to be aware of the cloud profile name now

docs/src/references/mappers/builtin-flows.md
jarhodes314: That's now much clearer, thanks.

tests/RobotFramework/tests/azure/azure_telemetry.robot
jarhodes314: I've realised I totally failed to answer the question re implementation effort.

It's essentially trivial to implement this change to my work, though there are a few sources of truth for the mapping between `(CloudType, Option<ProfileName>)` and the configuration path and I would like to rectify that (regardless of whether we change the directory structure or not). Even then, this would not be a difficult task, and the consistent directory structure between non-profiled and profiled clouds makes things easier.

jarhodes314: I've just had another thought through a conversation with @reubenmiller, I suspect for supporting `tedge config` commands (and generally solving the issues re combination of builtin vs custom mappers), we'd be better off grouping the custom/generic mappers and relying on profiles as the mechanism by which people name said mappers, for example with eventgrid like this:
```
/etc/tedge/mapper/
|-- c8y/
|   |-- tedge.toml
|   |-- flows/
|   |-- certs/
|-- c8y.foo/     # a c8y mapper for the cloud profile foo
|   |-- tedge.toml
|   |-- flows/
|   |-- certs/
|-- az/
| ...
|-- aws/
| ...
|-- custom.eventgrid/  # a mapper that is not natively supported
|   |-- tedge.toml
|   |-- bridge.toml
|   |-- flows/
|   |-- certs/
|-- collectd/   # a mapper that is not connected to the cloud
|   |-- flows/
```

didier-wenzek: > Where do we put flows are are generic (e.g. they are an extension to the te topics), so they aren't specifically related to a cloud mapper. I'd be happy to keep the `tedge-mapper flows` to run these flows individually  (or would this be a job for the tedge-agent?)

First of all, it's better to keep the agent unaware of any flow logic. Flows are about mapping rules and as such driven by `tedge-mapper`.

If we look only to the smaller scope of flows (ignoring for now the cloud connection aspects), when the mapper is started with `tedge-mapper  $X --profile $Y` it only has to look for its flow definition in the directory `/etc/tedge/mapper/$X.$Y/flows`.

> how do we manage the installation and removal of flows, as the command will also need to be aware of the cloud profile name now

The `tedge flows` cli will have to be aware of the cloud profiles too, in order to list or test flows for a specific cloud, profile or mapper. So an option is to add a `tedge flows config-dir` sub command returning the directory for a specific cloud profile. Installation and removal of flows can then be mostly directory based.

didier-wenzek: We conclude the following:

- All the mapper related configuration will be stored under `/etc/tedge/mappers` using a sub-directory per mapper.
  - This PR will be updated accordingly the hierarchy for flow definitions
  - The per-mapper configuration hierarchy will be updated along the same lines, but independently of this PR. 
  - The others aspects such as c8y operation templates or certificates will be addressed later.
- The sub-directory for a mapper is named after the mapper name and the cloud profile if any.
  - `$NAME.$PROFILE` 
  - `tedge-mapper c8y` config directory is  `/etc/tedge/mappers/c8y` 
  - `tedge-mapper c8y --profile foo` config directory is  `/etc/tedge/mappers/c8y.foo`
- All the flows for a mapper are under a `flows` sub-directory
  -  `tedge-mapper az`  flows are defined under `/etc/tedge/mappers/az/flows`
  -  `tedge-mapper flows`  flows are defined under `/etc/tedge/mappers/flows/flows`
- The `tedge flows` command is extended to handle mapper names and profiles
 - A  `tedge flows config-dir` sub-command is added to display the flows directory for a mapper profile.

reubenmiller (commented):
Generally the mapping worked nice, and I was able to customize the flow to add some additional properties in the payload. Though I found one slightly unexpected behaviour:

Updating the az flow definition resulted in a message being published on the `tedge-flows` service rather than the `tedge-mapper-az` service

```sh
[te/device/main/service/tedge-flows/status/flows] {"flow":"/etc/tedge/az/flows/mea.toml","status":"updated","time":1767941019}
```

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -241,7 +241,7 @@ Install Flow
     [Arguments]    ${directory}    ${definition_file}
     ${start}    Get Unix Timestamp
     ThinEdgeIO.Transfer To Device    ${CURDIR}/${directory}/${definition_file}    /etc/tedge/mappers/flows/flows/
-
+    Restart Service    tedge-flows
```
didier-wenzek: This commit will have to be reverted before this PR merged.

For some reasons I fail to understand dynamic flow definition reloading in no more working within the test framework, while this seems to be still correct outside.

reubenmiller: Resolved by fc5d808fbcc01cdde49dda81493c4a5bfabd8dc3

docs/src/references/mappers/flows.md
```diff
@@ -306,3 +316,34 @@ The following builtin objects are exported:
...
+### `ignore-topics`
+
+Filter out messages with specific topics
+- Must be configured with a list of `topics` and topic filters to be ignored
+- `{ builtin = "ignore-topics", config.topics = ["te/device/main/service/mosquitto-c8y-bridge/#"] }`
+
+### `set-topic`
+
+Assign a topic to messages
+- Must be configured with the `topic` to be assigned.
```
reubenmiller: It might be helpful to give a hint where this would be typically used, e.g. for flows which data comes from other sources like a command's output instead of directly from MQTT (hence why the message might not have a topic already configured).

docs/src/references/mappers/flows.md
```diff
@@ -306,3 +316,34 @@ The following builtin objects are exported:
...
+- The `format` to be used is either `unix` (the default) or `rfc3339`.
+- Unless specified otherwise the name for the added timestamp `property` is `time`.
+- When the input message already has a timestamp property the default is to let it unchanged.
+  This can be changed with the `reformat` config so any timestamp is reformated to the requested format. 
+- `{ builtin = "add-timestamp", config = { format = "rfc3339", reformat = true }}`
+
+### `cap-payload-size`
+
+Filter out messages which payload is too large
+- Must be configured with the `max_size` for the messages
```
reubenmiller: We should add the unit of measurement that the max_size is defined, e.g. bytes

docs/src/references/mappers/flows.md
```diff
@@ -306,3 +316,34 @@ The following builtin objects are exported:
...
+- Can be configured to `discard` the messages instead of raising an error (the latter being the default)
+- Expect a single configuration setting, the `topic` to be assigned
```
reubenmiller: This comment is not clear what the `topic` is referring to. Do you mean that a `topic` can be defined where the error messages are sent to if the payload limit is exceeded?

didier-wenzek: Oops, just a copy-pasted line that has to be removed.

configuration/package_manifests/nfpm.tedge-flows.yaml
```diff
@@ -36,7 +36,7 @@ contents:
       mode: 0644
     packager: rpm
 
-  - dst: /etc/tedge/flows
+  - dst: /etc/tedge/mappers/flows/flows
```
albinsuresh, line 39:
```suggestion
  - dst: /etc/tedge/mappers/custom/flows
```
As it was looking a bit off with nested `flows` directories.

The actual root of that issue is the name of the mapper/service itself being named `tedge-flows`. Since the naming is yet to be finalised AFAIK, this is probably outside the scope of this PR. Just raising it here as we've been trying to lock-in the directory paths so that they don't change between versions.

If we're still accepting naming suggestions, I've got a few: 
* tedge-stream (aligns well with flows)
* tedge-router (rhymes with `mapper` 😉 )
* tedge-transformer (still rhymes)
* tedge-adapter (rhymes but generic)
* tedge-composer (somewhat rhymes)

That being said, naming of the service itself is definitely out of scope for this PR.

docs/src/references/mappers/flows.md
didier-wenzek: > hence why the message might not have a topic already configured

Messages always have a topic, whatever their source. The point here is to assign a target topic.

Here my proposal: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/e966a01cec6f02e4744a7a8e3e5fcf523646e99c

docs/src/references/mappers/flows.md
didier-wenzek: added: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/e966a01cec6f02e4744a7a8e3e5fcf523646e99c

docs/src/references/mappers/flows.md
didier-wenzek: Removed: https://github.com/thin-edge/thin-edge.io/pull/3892/changes/e966a01cec6f02e4744a7a8e3e5fcf523646e99c

crates/extensions/az_mapper_ext/src/lib.rs
didier-wenzek: Done https://github.com/thin-edge/thin-edge.io/pull/3892/changes/99e11dce997d150811fd42f658591d2fac4cc20f

albinsuresh (approved):
Re-confirming my approval.

reubenmiller (approved):
Approved

PR #3902, opened 2025-12-18
feat: Support automated config migration for separate mapper config

jarhodes314:
## Proposed changes
Add a migration tool for converting `tedge.toml` based configs to separate mapper configs.

~To use the tool, run `tedge-mapper migrate-config c8y` (or with the cloud of your choosing).~
To use the tool, run `tedge config upgrade`.

TODO:
- [x] System tests for the separate mapper config functionality
- [x] Add line to `tedge connect` summary showing where the mapper is configured
- [x] Ensure there is a mechanism for changing the default mapper config location for new users
- [x] Include the directory cleanup logic in the system tests
- [ ] Test that config can be read from tedge.toml if mapper has flows directory but not migrated config
- [ ] Ensure we are happy with `tedge connect` config summary showing the feature/temporarily hide it if not

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/configuration/mapper_config_migration.robot
+++ b/tests/RobotFramework/tests/configuration/mapper_config_migration.robot
@@ -130,7 +130,7 @@ Migration With Write Protected Tedge Toml
     Setup Test Config    c8y    test.c8y.io
 
     # Ensure mappers config dir exists (it might not if this test is run on its own)
-    Execute Command    sudo -u tedge mkdir /etc/tedge/mappers
+    Execute Command    sudo -u tedge mkdir -p /etc/tedge/mappers
     Execute Command    sudo chmod 755 /etc/tedge/mappers
 
     # Make tedge directory read-only


crates/core/tedge_mapper/src/lib.rs
```diff
@@ -117,6 +119,10 @@ pub enum MapperName {
     Collectd,
     #[cfg(feature = "tedge-flows")]
     Flows,
+    #[clap(hide = true)]
+    MigrateConfig {
+        cloud_name: CloudType,
+    },
```
didier-wenzek: Why use a hidden command for this migration tool?

didier-wenzek (commented):
`tedge-mapper migrate-config` is nicely working, including for cloud profile.

It could be improved a bit to be more explicit when no config is to be migrated (because not set or already migrated).

crates/core/tedge_mapper/src/lib.rs
jarhodes314: At the moment, I didn't want to expose the details for undocumented features, particularly one where the user has nothing obvious to gain by using the feature

jarhodes314:
> I noticed that now `tedge config set` remove unknown properties, while this was not the case before #3889. I'm okay with that, as I don't think it was a documented behavior, but it would be good to consider restoring the former behavior.

I believe this is the existing behaviour for `tedge.toml` even though it is new for the mapper toml files. Since pretty much no users edit these files directly, I don't think this matters.

crates/core/tedge_mapper/src/lib.rs
jarhodes314: Particularly with @albinsuresh's point about unifying directory structure between this and the mapper flow definitions in https://github.com/thin-edge/thin-edge.io/pull/3892#discussion_r2663835565, there's a chance we may need to slightly revise the structure, which is a good reason for people not to adopt this feature until we are happy it's stable.

jarhodes314:
> 1. Why not restart the corresponding service after the config file migration? If we have intentionally left it to the user, thenit'd be good to give a hit about that to the user in the output. I prefer implicit restart as leaving the config migrated without the restart feels risky, as the user might forget about it and it'll unknowingly take effect with a future restart. Doing the restart inline gives us the opportunity to revert the changes as well, if it fails.

We could potentially do this, however it's also the case that this migration shouldn't change anything. If we did support relative paths in config files in the future then this could be a problematic assumption, but I don't see much 

> 2. Why expose the migration step via the `tedge-mapper` cli and not `tedge config`? That feels like an impl detail leak that only the mapper configs are migrated. In future if we decide to move more configs around, having a singular "migration interface" would have been nice. I'm thinking something like `tedge config upgrade` that will detect the existing config version and migrate it to the target version that the command was invoked with. Even at the code level, the `migrate-config` subcommand felt misplaced.

This is a good point, particularly since we only install tab-completions for `tedge` and not `tedge-mapper` so if it's under `tedge` it will be easier to use.

> 3. The current API forces the user to do the migration one-by-one for each cloud, presumably to make the transition less error prone as the user would be able to validate things before proceeding to the next migration. But for all cloud profiles of a given cloud, the migration is done in one-shot and here we are inconsistent with the other behaviour. I'd be i favour of migrating all at once.

Particularly given the new default behaviour is currently set to only create separate mapper configs if no cloud configs exist in `tedge.toml`, this makes sense. I think it adds to the whole point of encapsulating the details of what the config migration/update is.

> 4. Wouldn't it be safer to keep a backup of the old config at `/etc/tedge/tedge.toml.old` so that the user can revert back to it just in case something fails after the migration?

Maybe. If we did this, I would like there to be a clear process to remove this file because I think leaving things like that around will just cause confusion otherwise. This is particularly true since we generally encapsulate all of the configuration files so the users generally never need touch them.

albinsuresh:
> > 4. Wouldn't it be safer to keep a backup of the old config at `/etc/tedge/tedge.toml.old` so that the user can revert back to it just in case something fails after the migration?
> 
> Maybe. If we did this, I would like there to be a clear process through which we remove this file automatically, since leaving it around will mostly likely cause confusion. This is particularly true since we generally encapsulate all of the configuration files so the users generally never need touch them.

If we do the service restarts as part of the "migration" then we can cleanly remove the backup after the service restarts and connection tests are deemed successful.

jarhodes314:
This now supports the updated directory structure as discussed in https://github.com/thin-edge/thin-edge.io/pull/3892#discussion_r2663835565. There is one known failing rust test due to an error message not quite being as clear as it was, and one TODO that I need to address regarding some minor code cleanup.

I have set things up so the migrated config only maintains mapper config directories that have a config file (or some other contents, such as persisted flow definitions). If we remove the configuration for a given cloud and there is nothing else in its mapper directory, that directory will be cleaned up. I still need to cover this in the system tests.

didier-wenzek (approved):
Approved.

This is working nicely. I have even been happily surprised to see the following working:

```
$ tedge config upgrade
Error: failed to upgrade the configuration format
Caused by:
    0: Writing the content to the file "/etc/tedge/mappers/az/tedge.toml" failed
    1: Permission denied (os error 13)

$ checking
Indeed a previously created az directory is not writable
Good: the c8y config has been correctly migrated
Good: the az config has been left unchanged

$ fix az owner

$ tedge config upgrade

$ checking
Good: the az has now been migrated
Good: the c8y config is left unchanged  (already upgraded)
```

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/mod.rs
```diff
@@ -108,26 +108,30 @@ pub trait SpecialisedCloudConfig:
...
+            Some(profile) => {
+                let profile = profile.as_ref();
+                dir.join(format!("{ty}.{profile}/tedge.toml"))
+            }
         }
     }
 }
 
 pub trait HasPath {
     fn set_mapper_config_dir(&mut self, path: Utf8PathBuf);
```
albinsuresh:
```suggestion
    fn set_mappers_root_dir(&mut self, path: Utf8PathBuf);
```
As it is not just about the configs anymore and the config dir really is `mappers/{cloud}`.

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
```diff
@@ -87,44 +107,86 @@ impl TEdgeConfigLocation {
...
+        MapperConfigPath {
+            base_dir: Cow::Owned(self.mappers_config_dir()),
+            cloud_type: T::expected_cloud_type(),
+        }
+    }
+
     /// Decide which configuration source to use for a given cloud and profile
     ///
     /// This function centralizes the decision logic for mapper configuration precedence:
     /// 1. New format (`mappers/[cloud].toml` or `mappers/[cloud].d/[profile].toml`) takes precedence
```
albinsuresh:
```suggestion
    /// 1. New format (`mappers/[cloud].toml` or `mappers/[cloud].[profile].toml`) takes precedence
```

crates/common/tedge_config/src/lib.rs
```diff
@@ -27,13 +29,36 @@ impl TEdgeConfig {
         config_location.load().await
     }
 
+    /// Load [TEdgeConfig], using a separate mapper config file as the default
+    /// behaviour if no clouds are already configured
+    ///
+    /// As of 2026-01-05, this is only used for testing how the new default will
+    /// work before we fully adopt the new format. When we do this, we should
+    /// probably also include `tedge-mapper config-migrate` commands in the
```
albinsuresh:
```suggestion
    /// probably also include `tedge config upgrade` commands in the
```

tests/RobotFramework/tests/configuration/mapper_config_connect.robot
```diff
@@ -0,0 +1,108 @@
...
+
+
+*** Test Cases ***
+Connection Succeeds With Migrated Config
+    [Documentation]    Verify actual connection works with migrated config
+    # Setup and migrate
+    Migrate Cloud Configs
+
+    # Connect (actual connection, not test mode)
+    Execute Command    sudo tedge reconnect c8y
```
albinsuresh, line 23: Is a `reconnect` really necessary? Just a mapper service restart would have sufficed, right?

tests/RobotFramework/tests/configuration/mapper_config_migration.robot
```diff
@@ -0,0 +1,219 @@
...
+    Setup Test Config    c8y    test.c8y.io
+    Execute Command    sudo tedge config set c8y.topics "te/+/+/+/+/e/+"
+
+    # Execute migration
+    Migrate Cloud Configs
+
+    # Verify: Topics preserved in migrated config
+    Verify Mapper Config File Contains Pattern    c8y    topics = ["te/+/+/+/+/e/+"]
+
+Migration Preserves C8y Smartrest Templates
```
albinsuresh: Just curious what is/was special with SmartREST templates and `c8y.topics` configs from the previous test for just those to have dedicated tests?

tests/RobotFramework/tests/configuration/mapper_config_migration.robot
```diff
@@ -0,0 +1,219 @@
...
+
+Migration With Empty C8y Config
+    [Documentation]    Verify migration only creates directories if config exists
+    # Setup: Just set a minimal config
+    Execute Command    sudo tedge config unset c8y.url
+
+    # Execute migration
+    Migrate Cloud Configs
+
+    # Verify: File created (even if mostly empty)
```
albinsuresh:
```suggestion
    # Verify: File not created (as it would have been empty)
```

crates/common/tedge_config/src/tedge_toml/tedge_config/mapper_config/compat.rs
```diff
@@ -163,8 +189,21 @@ impl FromCloudConfig for AwsMapperSpecificConfig {
...
+            .try_get(profile, "aws")
+            .unwrap()
+            .mapper_config_file
+            .clone()
+            .unwrap_or(
+                tedge_config
+                    .location
+                    .tedge_config_root_path()
+                    .join("tedge.toml"),
+            );
```
albinsuresh:
```suggestion
            .unwrap_or_else(|| {
                tedge_config
                    .location
                    .tedge_config_root_path()
                    .join("tedge.toml")
            });
```

tests/RobotFramework/tests/configuration/mapper_config_migration.robot
```diff
@@ -0,0 +1,219 @@
...
+    ...    sudo -u tedge tedge config upgrade
+    ...    exp_exit_code=!0
+    ...    stderr=${True}
+    ...    stdout=${False}
+    Should Contain    ${result}    Permission denied
+    # The error should talk about how `/etc/tedge/mappers` can't be created
+    Should Contain    ${result}    /etc/tedge/mappers
+
+    # Cleanup: Restore permissions
+    Execute Command    sudo chmod 755 /etc/tedge
```
albinsuresh, line 203: One more test that might be worth adding is multi-cloud, multi-profile:

```
Migrate Multiple Clouds And Profiles
    [Documentation]    Verify migration handles multiple cloud and their profiles simltaneously
    # Setup: Configure default and two named profiles
    Setup Test Config    c8y    default.c8y.io
    Setup Test Config    c8y    prod.c8y.io    profile=prod
    Setup Test Config    c8y    test.c8y.io    profile=test

    Setup Test Config    az    default.azure.com
    Setup Test Config    az    prod.azure.com    profile=prod

    Setup Test Config    aws    default.amazonaws.com
    Setup Test Config    aws    test.amazonaws.com    profile=test

    # Execute migration
    Migrate Cloud Configs

    # Verify: All three config files created
    Verify Mapper Config File Exists    c8y
    Verify Mapper Config File Contains    c8y    url    default.c8y.io

    Verify Mapper Config File Exists    c8y    profile=prod
    Verify Mapper Config File Contains    c8y    url    prod.c8y.io    profile=prod

    Verify Mapper Config File Exists    c8y    profile=test
    Verify Mapper Config File Contains    c8y    url    test.c8y.io    profile=test

    Verify Mapper Config File Exists    az
    Verify Mapper Config File Contains    az    url    default.azure.com
    Verify Mapper Config File Exists    az    profile=prod
    Verify Mapper Config File Contains    az    url    prod.azure.com    profile=prod

    Verify Mapper Config File Exists    aws
    Verify Mapper Config File Contains    aws    url    default.amazonaws.com
    Verify Mapper Config File Exists    aws    profile=test
    Verify Mapper Config File Contains    aws    url    test.amazonaws.com    profile=test
```

tests/RobotFramework/tests/configuration/mapper_config_connect.robot
reubenmiller: The `tedge reconnect c8y` is probably preferred here as the commands takes care of the service abstraction so you don't have to run systemd commands yourself, and you also get a formal verification everything also works correctly (e.g. the connection test etc.)

tests/RobotFramework/tests/configuration/mapper_config_migration.robot
jarhodes314: Yeah, I can't remember. I'll delete one of them as I don't think it tells us anything the other tests don't. It's unlikely to cause an issue, but it probably does make some sense testing not just a string config.

Issue #3907, opened 2026-01-06
Concurrent execution of tedge connect --test causes "connection closed by peer" errors due to ID collision


d-timotijevic:
**Describe the bug**
The command `tedge connect <cloud> --test` is not process-safe. If two separate processes or services invoke this command simultaneously to check connectivity, they fail with a disconnection error.

This happens because the connectivity check implementation uses a **hardcoded, static MQTT Client ID** (e.g., `check_connection_c8y`).

When a second process attempts the check while the first is running, the local MQTT broker (Mosquitto) detects a duplicate Client ID and forcibly disconnects the existing session (`connection closed by peer`). This results in a race condition where valid connectivity checks report failure.

**To Reproduce**
1. Ensure `tedge` is configured with a valid cloud connection (e.g., Cumulocity).
2. Run the connection check concurrently in a single command:
   ```bash
   sudo tedge connect c8y --test & sudo tedge connect c8y --test
   ```
3. Observe that one or both processes fail immediately.

**Actual Logs / Output**
```text
$ sudo tedge connect c8y --test & sudo tedge connect c8y --test
[2] 273151
Testing Cumulocity connection with config:
        ...
        bridge: mosquitto
Testing Cumulocity connection with config:
        ...
        bridge: mosquitto

Verifying device is connected to cloud... ✗
error: Failed to verify device is connected to Cumulocity: Failed to connect to mosquitto for connection check: Mqtt state: Io error: Custom { kind: ConnectionAborted, error: "connection closed by peer" }: Io error: Custom { kind: ConnectionAborted, error: "connection closed by peer" }: connection closed by peer

Verifying device is connected to cloud... ✗
error: Failed to verify device is connected to Cumulocity: Failed to connect to mosquitto for connection check: Mqtt state: Io error: Custom { kind: ConnectionAborted, error: "connection closed by peer" }: Io error: Custom { kind: ConnectionAborted, error: "connection closed by peer" }: connection closed by peer
[2]    273151 exit 1     sudo tedge connect c8y --test
```

**Expected behavior**
Multiple processes (e.g., a background service monitor and a user CLI session) should be able to run connection tests concurrently without interfering with one another. The CLI should generate a unique Client ID for the check (e.g., by appending the process ID) to avoid broker collisions.

**Environment (please complete the following information):**
 - OS: Linux
 - Hardware: x86_64
 - thin-edge.io version: 1.4.2

**Additional context**
[...]

PR #3908, opened 2026-01-06
fix: allow concurrent `tedge connect <cloud> --test` processes

rina23q:
## Proposed changes

To support running multiple tedge connect test processes simultaneously, this change introduces a random suffix to the session name. This prevents naming conflicts when concurrent processes are executed.

Note: Although removing `.with_session_name(CLIENT_ID)` would technically resolve the issue, I chose to introduce `.with_session_prefix()` for a specific reason: keeping the prefix in the session name improves debuggability in MQTT logs, as the session name clearly identifies the source of each MQTT session.

## Types of changes

## Paste Link to the issue
#3907 

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/config.rs
+++ b/crates/common/mqtt_channel/src/config.rs
@@ -294,6 +294,18 @@ impl Config {
         }
     }
 
+    /// Set the random session name with prefix and clear the session
+    pub fn with_session_prefix(self, prefix: impl Into<String>) -> Self {
+        let random: String = std::iter::repeat_with(fastrand::lowercase)
+            .take(10)
+            .collect();
+        Self {
+            session_name: Some(format!("{}-{}", prefix.into(), random)),
+            clean_session: true,
+            ..self
+        }
+    }
+
     /// Unset the session name and clear the session
     pub fn with_no_session(self) -> Self {
         Self {
--- a/tests/RobotFramework/tests/tedge_connect/tedge_connect_test.robot
+++ b/tests/RobotFramework/tests/tedge_connect/tedge_connect_test.robot
@@ -18,6 +18,10 @@ tedge_connect_test_positive
     ${output}=    Execute Command    sudo tedge connect c8y --test    stdout=${False}    stderr=${True}
     Should Contain    ${output}    Connection check to c8y cloud is successful.
 
+Multiple tedge connect test processes can run concurrently
+    [Tags]    \#3907
+    Execute Command    sudo tedge connect c8y --test & sudo tedge connect c8y --test    timeout=5
+
 Non-root users should be able to read the mosquitto configuration files #2154
     [Tags]    \#2154
     Execute Command    sudo tedge connect c8y || true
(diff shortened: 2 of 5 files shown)


crates/common/mqtt_channel/src/config.rs
```diff
@@ -294,6 +294,17 @@ impl Config {
         }
     }
 
+    /// Set the random session name with prefix
+    pub fn with_session_prefix(self, prefix: impl Into<String>) -> Self {
+        let random: String = std::iter::repeat_with(fastrand::lowercase)
+            .take(10)
+            .collect();
+        Self {
+            session_name: Some(format!("{}-{}", prefix.into(), random)),
```
didier-wenzek, line 303: Better to force a clean session here.


```suggestion
            session_name: Some(format!("{}-{}", prefix.into(), random)),
            clean_session: true,
```

This is notably missing for the az and aws connections.

rina23q: Which one is better, forcing a clean session flag `true` in the `with_session_prefix()` API as you suggested now, or adding it with `.with_clean_session(true)` to`MqttOptions` as below?(already `.with_clean_session(true)` for c8y's connection check and jwt retrieval)

```rust
    let mut mqtt_options = tedge_config
        .mqtt_config()?
        .with_session_prefix(CLIENT_ID)
        .with_clean_session(true)
        .rumqttc_options()?;
    mqtt_options.set_keep_alive(RESPONSE_TIMEOUT);
```

rina23q: I guessed the former is preferable as the session that requires a random name always wants a clean session for sure. Fixed in 5db505ebba9e4aa7a4b000a53ba3b1df792853d1

didier-wenzek: Better inside the `with_session_prefix()` method because its easy to forget to clear the session while this adds burden to the broker for nothing (we will never reconnect using the same random name). As the `with_session_prefix()`  assigns *some* session name this defeats the check done here: https://github.com/thin-edge/thin-edge.io/blob/main/crates/common/mqtt_channel/src/config.rs#L390

tests/RobotFramework/tests/tedge_connect/tedge_connect_test.robot
```diff
@@ -18,6 +18,10 @@ tedge_connect_test_positive
     ${output}=    Execute Command    sudo tedge connect c8y --test    stdout=${False}    stderr=${True}
     Should Contain    ${output}    Connection check to c8y cloud is successful.
 
+Multiple tedge connect test processes can run concurrently
```
reubenmiller, line 21: @rina23q Can you check that this test is ok in the flake-finder? I'd suspect there might be a risk of some inconsistency (though I'm 100% confident that your changes are ok, just the testing part is by nature indeterminate (e.g. if the first process takes longer to start up then the second process for unexpected reasons)

rina23q: 100/100 passed 👍 Actually, if two processes don't run concurrently, this test will succeed anyway.
I confirmed this test failed with the same session name (before the change).

Issue #3894, opened 2025-12-11
The builtin-bridge should be configurable


didier-wenzek:
**Is your feature improvement request related to a problem? Please describe.**

The builtin bridge embedded by the c8y, az and aws mappers is currently using hard-coded rules and a user can not add, modify or remove a topic mapping rule. Notably, the current az mapper is using rules to map thin-edge topics with Azure IoT Hub topics and cannot be used to connect Azure Event Grid which topic schema is different.

**Describe the solution you'd like**

The configuration of the builtin-bridge for a given mapper instance should be externalized in a file, and this file should be editable by users. Ideally, this bridge configuration file should be a template with placeholders for the cloud MQTT endpoint as well as the device identifier and credentials.

**Describe alternatives you've considered**

**Additional context**

PR #3918, opened 2026-01-10
feat: configurable built-in bridge rules

jarhodes314:
## Proposed changes
Enable the built-in bridge to be configured with custom rules, ultimately to help enable a generic mapper.

### Templating functionality
Users can define a series of toml templates for various bits of bridge logic:

```toml
local_prefix = "${config.c8y.bridge.topic_prefix}/" # you can also specify these per-rule
remote_prefix = ""

[[rule]] # you can define a series of rules
topic = "s/us/#"
direction = "outbound" # alternatively, can be: inbound or bidirectional

[[rule]]
if = "${config.c8y.mqtt_service.enabled}" # boolean config variables can be used to turn rules on/off
local_prefix = "${config.c8y.bridge.topic_prefix}/mqtt/out/" # local prefix is overridden for this rule
topic = "#"

[[rule]]
if = "${connection.auth_method} == 'password'" # we can also toggle rules based on the authentication method the bridge is using
topic = "s/ul/#"

[[template_rule]] # template rules iterate over something to generate multiple rules
for = "${config.c8y.smartrest.templates}" # e.g. we can iterate over a template set
topic = "s/dc/${item}" # and interpolate the iterated value in the topic
direction = "inbound"

[[template_rule]]
for = {item = "mode", in = ['s', 't', 'q', 'c'] } # we can also iterate over array literals
topic = "${mode}/us/#"
direction = "outbound"
```

### Trying it out
For testing it out, save the following template to a file and run `cargo run -p tedge_mqtt_bridge <filename>`. You can play around with setting `c8y.smartrest.templates` and seeing the generated output change as a result. I've experimented with using [`serde_spanned`](https://docs.rs/serde_spanned/latest/serde_spanned/) and [`ariadne`](https://docs.rs/ariadne/latest/ariadne/) for contextual error reporting with invalid templates.

It is also now used by the c8y mapper. The templates are currently loaded from the rust code directly, they aren't persisted, but you can edit the `smartrest.toml` and `mqtt-service.toml` files as appropriate to experiment with the bridge rules/error handling on invalid templates. With the files as committed, the behaviour should be identical to the existing bridge rules.

```toml
local_prefix = "${config.c8y.bridge.topic_prefix}/"
remote_prefix = ""

[[rule]]
topic = "s/ds"
direction = "inbound"

[[rule]]
local_prefix = "c8y/" # you can override the variables within rules
remote_prefix = ""
topic = "s/us"
direction = "outbound"

[[template_rule]]
for = "${config.c8y.smartrest.templates}"
topic = "s/uc/${item}"
direction = "outbound"

[[template_rule]]
for = "${config.c8y.smartrest.templates}"
topic = "s/dc/${item}"
direction = "inbound"

[[template_rule]]
if = "${connection.auth_method == 'password'"
for = ['s', 't', 'q', 'c']
topic = "${item}/ul/#"
direction = "outbound"
```
TODO testing/error handling:
- [x] Do some sanity checking on disabled rules (e.g. if the `if` condition isn't satisfied, but the template is genuinely invalid
- [x] Collect all errors at once when parsing a template, don't just stop on the first error
- [x] Better parse errors for `if`
- [x] Suppress spillover errors (e.g. local_prefix doesn't parse, don't then claim local_prefix is missing)
- [x] Rule negation
- [x] Relax tedge-mapper-c8y tests to prevent new rules causing failures

TODO:
- [x] Parse a configuration file of bridge rules
- [x] Make the configuration variable expansion profile-aware
- [x] Configure the bridge with said rules
- [x] Integrate this into a mapper so it can run custom rules
- [x] Persist the default rules for the c8y mapper
- [ ] ~Persist the default rules from each cloud mapper like we do for flows in #3892~ See #3965
- [ ] ~Formalise the debug cli~ See #3964
- [x] Document the configuration format

## Types of changes

## Paste Link to the issue
- #3894

## Checklist

### DIFF
--- a/crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/condition.rs
+++ b/crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/condition.rs
@@ -80,7 +80,7 @@ fn lexer<'src>(
     // Operators: any combination of `!` and `=` (e.g., `==`, `!=`, `=`)
     // We don't currently support != but may wish to in the future, and users may try this
     // so recognising it but explicitly rejecting it is sensible
-    let op = one_of("!=")
+    let op = one_of("!=&|")
         .repeated()
         .at_least(1)
         .to_slice()


crates/core/tedge_mapper/src/c8y/bridge/smartrest.toml
```diff
@@ -0,0 +1,90 @@
...
+topic = "s/dl/${@for}"
+direction = "inbound"
+
+[[template_rule]]
+for = "${.c8y.smartrest.templates}"
+topic = "s/dc/${@for}"
+direction = "inbound"
+
+[[rule]]
+if = "auth_method(certificate)"
```
didier-wenzek: Adding this `if` template rule is a nice idea.

jarhodes314: Thanks. I guess the alternative option is like with the flows where we read the config and persist it by way of only including the enabled steps in the generated flow, but given we need to have a way to reference the configuration from the bridge templates, it feels like we should be able to access other logic, and this felt like the sensible way to generalise things.

reubenmiller: Though we're quickly building our own DSL as we have some fairly complicated mechanisms (for, templating etc.)

I'm wondering whether supporting executing a javascript where it can be a simple or complicated as you like...so below shows the equivalent, where I've assuming that the function is called with two objects, connection (which includes connection details like "auth_method", and "config" which is an object of the tedge config (more or less).

```js
export function bridge_config(connection, config) {
  const rules = [
    {
      direction: "outbound",
      topic: "s/us/#",
      enabled: true,
    },
    {
      direction: "outbound",
      topic: "s/ul/#",
      enabled: connection.auth_method === "password",
    },

    // mqtt service topics
    {
      local_prefix: `${config.c8y.bridge.topic_prefix}/mqtt/out/`,
      topic: "#",
      enabled: config.c8y.mqtt_service.enabled,
    },

    // SmartREST message including different processing modes
    ...["s","t","q","c"].map((item) => ({
      direction: "outbound",
      topic: `${item}/us/#`,
      enabled: true,
    })),
    
    // Custom SmartREST templates
    ...(config.c8y.smartrest.templates || []).map((item) => ({
      direction: "inbound",
      topic: `s/dc/${item}`,
      enabled: true,
    })),
  ];

  // return full configuration object
  // as json
  return JSON.stringify({
    local_prefix: config.c8y.bridge.topic_prefix,
    remote_prefix: "",
    rule: rules.filter(item => item.enabled),
  });
}
```

Or just a simple example using static rules (with no fancy templating)

```js
export function bridge_config(connection, config) {
  return JSON.stringify({
    local_prefix: config.c8y.bridge.topic_prefix,
    remote_prefix: "",
    rule: [
      {
        direction: "outbound",
        topic: "s/us/#",
      },
      {
        direction: "outbound",
        topic: "s/ul/#",
      },
      {
        direction: "inbound",
        topic: "s/ds/#",
      },
    ],
  });
}
```

jarhodes314: I think the downside of javascript is it's a lot more difficult for us thin-edge to help the user detect errors if they screw something up (and my primary concern here is the templates we ship as those have some logic to cover the various config cases). With a toml file and special logic, we can perform error handling on the entire thing, checking that the templated variables are valid regardless of whether the rule is actually enabled/exists with the current configuration.

reubenmiller: Potentially, though it runs the risk of not being to meet requirements for custom bridges in the future. But yes you can't evaluate the javascript config without an tedge engine, so that would make it more difficult to validate offline.

The DLS is definitely not straight forward, and some parts like `if = "auth_method(password)"` seem to be inconsistent between defining a literal and defining a template variable. For example are all of the following equivalent/valid expressions?

```toml
if = "auth_method(password)"
if = "${.c8y.mqtt_service.enabled}"
if = true
if = false
if = "true"
if = "false"
```

jarhodes314: > Potentially, though it runs the risk of not being to meet requirements for custom bridges in the future. But yes you can't evaluate the javascript config without an tedge engine, so that would make it more difficult to validate offline.
> 
> The DLS is definitely not straight forward, and some parts like `if = "auth_method(password)"` seem to be inconsistent between defining a literal and defining a template variable. For example are all of the following equivalent/valid expressions?
> 
> ```toml
> if = "auth_method(password)"
> if = "${.c8y.mqtt_service.enabled}"
> if = true
> if = false
> if = "true"
> if = "false"
> ```

At the moment, the first two are valid, and none of the others are valid (so the only options are auth_method expressions and boolean config variables). I'm in agreement about it being somewhat inconsistent.

With the current design, we do have quite a lot of flexibility in the sense that it's not a breaking change to make something valid when it was previously invalid, so if we discover a large gap in functionality, we should be able to resolve that without causing any major issues.

I think the syntax definitely needs some consideration, and we have plenty of flexibility there. I erred on the side of avoiding bikeshedding, but it would be good to have a more coherent/predictable system here.

didier-wenzek: I agree that there is a risk to build our own DSL, in practice  incomplete and possibly inconsistent. However, using Javascript would be going too far. An intermediate solution could be to use a proper template engine such as [mustache](https://mustache.github.io/). However, I'm not even convinced that it's worth it.

1. Currently, users has no way to tune bridge definitions. Even plain configuration files with no templating/scripting supports would be a real improvement.
2. Using tedge config to repeat or opt-out bridge rules  (as proposed here) , will meet 99% of the requirements.
3. For niche requirements we might have in the future (say topic names are generated along some fancy logic), *external tools* can be used to generate the expected bridge rules.
4. We have to focus on *externalized* bridge rules.

jarhodes314: > I agree that there is a risk to build our own DSL, in practice incomplete and possibly inconsistent. However, using Javascript would be going too far. An intermediate solution could be to use a proper template engine such as [mustache](https://mustache.github.io/). However, I'm not even convinced that it's worth it.

I broadly agree with this assessment. In my mind, we're essentially trying to serve as an alternative to mosquitto, and that doesn't provide any sort of templating capabilities. My goal here is generally to provide a templating system that covers our needs for supporting the existing `tedge.toml` configurations while avoiding the complexity of the bridge template needing to change.

In an attempt to try the javascript approach, I've created https://github.com/thin-edge/thin-edge.io/pull/3938. Writing the template wasn't too difficult, and it is fairly ergonomic for writing the more complex rules (I would say more so than in a typical templating engine), and readability is pretty good. However, when I did encounter errors, they were quite tricky to pin down as it either results in a rule not being generated, or results in an error message which tells the user where in the json output something went wrong (and in that case, the line number it gives you to track down the source of the error is just totally meaningless).

jarhodes314: It's probably also worth considering the argument against a DSL. For me, this primarily boils down to "parsing an entire file generally comes with a lot of edge cases that need handling and this is tricky to get right". With the "DSL-in-toml" approach, we mostly avoid that complexity since the bulk of what needs parsing is done for us by the `toml` and `serde` crates.

For making the conditional rules more coherent, I suggest we do something like @reubenmiller's example, where we have some concept of connection and config and reference variables within those, for example:

```toml
local_prefix = "${config.c8y.bridge.topic_prefix}/"
remote_prefix = ""

[[rule]]
topic = "s/us/#"
direction = "outbound"

[[rule]]
if = "${config.c8y.mqtt_service.enabled}" 
local_prefix = "${config.c8y.bridge.topic_prefix}/mqtt/out/"
topic = "#"

[[rule]]
# previously: if = "auth_method(password)"
if = "${connection.auth_method} == 'password'"
topic = "s/ul/#"

[[template_rule]]
for = "${config.c8y.smartrest.templates}"
topic = "s/dc/${@for}"
direction = "inbound"

[[template_rule]]
for = ['s', 't', 'q', 'c']
topic = "${@for}/us/#"
direction = "outbound"
```

I think the current syntax is also a bit succinct. I think there is significant value to making the special things look a bit different. The proposal above makes `if` an expression, which helps it stand out.

For achieving this with the iteration syntax, I suggest the following:
```toml
[[template_rule]]
# previously: for = "${.c8y.smartrest.templates}"
each = { list = "${config.c8y.smartrest.templates}", as = "template_id" }
topic = "s/dc/${template_id}" # previously "s/dc/${@for}"
direction = "inbound"

[[template_rule]]
each = { list = ['s', 't', 'q', 'c'], as = "mode" }
topic = "${mode}/us/#"
direction = "outbound"
```

This gets rid of the special `@for` syntax that looks like it might be able to refer to multiple variables and allows for better documentation of what's going on. To me, `each` is also a clearer name for what's going on (to me, `each` implies iteration whereas `for` could mean a huge range of things).

I also think that `if` could be renamed to `enabled_if`, to make it clearer what happens when the condition is satisfied.

With all of these suggested improvements, this is what the example would look like:
```toml
local_prefix = "${config.c8y.bridge.topic_prefix}/"
remote_prefix = ""

[[rule]]
topic = "s/us/#"
direction = "outbound"

[[rule]]
enabled_if = "${config.c8y.mqtt_service.enabled}" 
local_prefix = "${config.c8y.bridge.topic_prefix}/mqtt/out/"
topic = "#"

[[rule]]
enabled_if = "${connection.auth_method} == 'password'"
topic = "s/ul/#"

[[template_rule]]
each = { list = "${config.c8y.smartrest.templates}", as = "template_id" }
topic = "s/dc/${template_id}"
direction = "inbound"

[[template_rule]]
each = { list = ['s', 't', 'q', 'c'], as = "mode" }
topic = "${mode}/us/#"
direction = "outbound"
```

didier-wenzek: > It's probably also worth considering the argument against a DSL. For me, this primarily boils down to "parsing an entire file generally comes with a lot of edge cases that need handling and this is tricky to get right". With the "DSL-in-toml" approach, we mostly avoid that complexity since the bulk of what needs parsing is done for us by the `toml` and `serde` crates.

Indeed a point in favor of the "DSL-in-toml" approach taken by this PR.

> This gets rid of the special `@for` syntax that looks like it might be able to refer to multiple variables and allows for better documentation of what's going on. To me, `each` is also a clearer name for what's going on (to me, `each` implies iteration whereas `for` could mean a huge range of things).

I like the new proposal and this ability to bind the values, that are iterated over, to a variable which name is chosen by the users.

> I also think that `if` could be renamed to `enabled_if`, to make it clearer what happens when the condition is satisfied.

I have no strong preferences regarding the names.

* `if` sounds clear enough to me.
* `for = { item = `template`,  in = "${config.c8y.smartrest.templates}"  }` would sound a bit more familiar.

jarhodes314: > I have no strong preferences regarding the names.
> 
> * `if` sounds clear enough to me.
> * `for = { item = `template`,  in = "${config.c8y.smartrest.templates}"  }` would sound a bit more familiar.

For now at least, I have implemented the new syntax like this.

crates/extensions/tedge_mqtt_bridge/Cargo.toml
```diff
@@ -13,20 +13,35 @@ repository = { workspace = true }
...
+
 [dependencies]
 anyhow = { workspace = true }
+ariadne = { workspace = true }
 async-trait = { workspace = true }
 backoff = { workspace = true }
 bytes = { workspace = true }
+camino = { workspace = true }
 certificate = { workspace = true }
+chumsky = { workspace = true }
```
didier-wenzek, line 28: A bit surprised to see a dependency on a parser combinator while using toml as the main structure for bridge templates.

crates/extensions/tedge_mqtt_bridge/Cargo.toml
```diff
@@ -13,20 +13,35 @@ repository = { workspace = true }
 # No features on by default
 default = []
 
+[[bin]]
+name = "test-bridge-rules"
```
didier-wenzek, line 17: This test command is really convenient and will have to be installed and documented.

crates/extensions/tedge_mqtt_bridge/src/main.rs
```diff
@@ -0,0 +1,40 @@
...
+    bridge_config.add_rules_from_template(
+        toml_path,
+        &toml_content,
+        &tedge_config,
+        AuthMethod::Certificate,
+        None,
+    )?;
+
+    // Debug print the expanded rules
+    println!("{:#?}", bridge_config);
```
didier-wenzek, line 37: Simple and effective. However, can this be improved with some `--debug` info to tell which tedge config setting has been used to skip or repeat a rule template.

didier-wenzek (commented):
A really nice improvement!

I still have to review the doc.

crates/extensions/tedge_mqtt_bridge/Cargo.toml
jarhodes314: Yes, it's probably best to polish this a bit and make it part of the `tedge` cli as at the moment it's a very quickly LLM'd together debug tool for mostly my purposes. Possibly we could also do with some sort of `test` command, where a user can observe what happens to a given topic from a given broker.

crates/extensions/tedge_mqtt_bridge/Cargo.toml
jarhodes314: Yeah, it's only for parsing the conditions (i.e. `if = "..."`), and I was trying to avoid it.

We can _just about_ parse these without resorting to a full parsing library, but with the initial manual implementation, it was starting to show the strain even with this. I think if we need to extend the syntax for these, the cracks will appear with a manual implementation. By comparison, I extended the current syntax with an optional negation operator for boolean configs, and that was lovely and simple: https://github.com/thin-edge/thin-edge.io/pull/3918/changes/cd6857e9885826fa4850eae76eac6ff504113e6d.

I should go and check what impact this has on binary size though, as it may not be worth the cost.

didier-wenzek: I appreciate that manual parsing is only okay in very limited cases. But don't we have such a simple case, a dot separated list of names?

I acknowledge that error reporting is great.

```
Error: Failed to expand bridge configuration
   ╭─[ /tmp/mqtt-service.toml:2:40 ]
   │
 2 │ if = "${config.c8y.mqtt_service.enabled"
   │                                        │ 
   │                                        ╰─ found end of input expected '.', or '}'
───╯
Error: Invalid bridge config template
```

> I should go and check what impact this has on binary size though, as it may not be worth the cost.

Yes, it would be good to know.

jarhodes314: > I appreciate that manual parsing is only okay in very limited cases. But don't we have such a simple case, a dot separated list of names?

We also have `${connection.auth_method} == 'password'` type expressions too. With `chumsky`, it's pretty easy to handle whatever operator people might try instead of `==` (e.g. `=` or `===` and not have too much trouble), and maybe also accept differently quoted strings (e.g. not valid right now, but `if = '${connection.auth_method} == "certificate"'` is a plausible thing for someone to write).

> > I should go and check what impact this has on binary size though, as it may not be worth the cost.
> 
> Yes, it would be good to know.

Before chumsky (7bdf98fe): 20,055,584 bytes (19.12 MiB)
With chumsky (7cd55a0a): 20,100,640 bytes (19.16 MiB)
Difference: +45,056 bytes (+45 KiB, +0.63%)
Adding chumsky increased the release binary size by about 45 KiB (~0.2%)

So it's a pretty small impact.

crates/extensions/tedge_mqtt_bridge/src/main.rs
jarhodes314: Got it to produce more user-friendly output (only locally atm), although I would like to extend this to explain more about what's going on in the template (e.g. "x is disabled because y" as opposed to just ignoring the rules that are disabled and similar handling for template rules that are effectively disabled due to an empty iterator).

image

crates/extensions/tedge_mqtt_bridge/Cargo.toml
didier-wenzek: So,  let's go with chumsky

crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/condition.rs
```diff
@@ -0,0 +1,534 @@
...
+        .labelled("auth method condition")
+}
+
+/// Main condition parser
+fn condition_parser<'tokens, 'src: 'tokens, I>(
+) -> impl Parser<'tokens, I, Condition, extra::Err<Rich<'tokens, Token<'src>, SimpleSpan>>> + Clone
+where
+    I: ValueInput<'tokens, Token = Token<'src>, Span = SimpleSpan>,
+{
+    choice((config_condition(), auth_condition())).labelled("condition")
```
didier-wenzek, line 191: `chumsky`'s error handling is amazingly fantastic!

In the following, the fact that only 'config' or 'connection' makes sense is independently defined by the `config_condition` and `auth_condition` parsers where this is not even the fist token: 

```
Error: Failed to expand bridge configuration
   ╭─[ /tmp/mqtt-service.toml:2:9 ]
   │
 2 │ if = "${auth.c8y.mqtt_service.enabled} == 'yeah' "
   │         ──┬─  
   │           ╰─── found 'auth' expected 'config', or 'connection'
───╯
Error: Invalid bridge config template
```

crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/condition.rs
```diff
@@ -0,0 +1,534 @@
...
+fn auth_condition<'tokens, 'src: 'tokens, I>(
+) -> impl Parser<'tokens, I, Condition, extra::Err<Rich<'tokens, Token<'src>, SimpleSpan>>> + Clone
+where
+    I: ValueInput<'tokens, Token = Token<'src>, Span = SimpleSpan>,
+{
+    let auth_method_var = just(Token::VarStart)
+        .ignore_then(just(Token::Ident("connection")))
+        .ignore_then(just(Token::Dot))
+        .ignore_then(just(Token::Ident("auth_method")))
+        .then_ignore(just(Token::VarEnd));
```
didier-wenzek, line 165: I'm a bit surprised by this mix of parsing and semantics, i.e to hard code at the parser level that `connection.auth_method == something` makes senses but not `config..auth_method == something`.

That said, this let us leverage `chumsky`'s error reports.

crates/extensions/tedge_mqtt_bridge/src/config_toml/mod.rs
```diff
@@ -0,0 +1,1643 @@
...
+
+impl<Target> FromStr for ConfigReference<Target> {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s = s
+            .strip_prefix("${config.")
+            .ok_or_else(|| anyhow::anyhow!("Config variable must start with ${{config."))?;
+        let s = s
+            .strip_suffix("}")
+            .ok_or_else(|| anyhow::anyhow!("Config variable must end with }}"))?;
```
didier-wenzek: Okay, but wouldn't be better to have to same parsing logic for `if` and `for` templating rules?

At least to have both parsing logic side by side in `parse_condition.rs` and `parse_repetition.rs` to make this more obvious.

docs/src/references/mappers/configurable-bridge.md
```diff
@@ -0,0 +1,173 @@
...
+[[template_rule]]
+if = "${connection.auth_method} == 'password'"
+for = { item = "mode", in = ['s', 't', 'q', 'c'] }
+topic = "${mode}/ul/#"
+direction = "outbound"
+```
+
+## Template file structure
+
+The Cumulocity built-in behavior is configured in two files: `/etc/tedge/mappers/c8y/bridge/smartrest.toml` and `/etc/tedge/mappers/c8y/bridge/mqtt-service.toml`.
```
didier-wenzek: Better to have a sub-section per cloud as the templates will differ.

```suggestion
## Template file structure

### Cumulocity mapper
The Cumulocity built-in behavior is configured in two files: `/etc/tedge/mappers/c8y/bridge/smartrest.toml` and `/etc/tedge/mappers/c8y/bridge/mqtt-service.toml`.
```

crates/extensions/tedge_mqtt_bridge/src/config_toml/mod.rs
jarhodes314: Yes, it would make sense to use this parsing logic for both, particularly since the `for` syntax is merely a subset of the `if` syntax, and the span information will be more targeted. I'll make that change in a bit.

crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/condition.rs
jarhodes314: Yeah, finding the right balance of layers here can be tricky. I decided to separate lexing and parsing to improve the error handling with slightly incorrect operators (e.g. `=` or `===` instead of `==`). The code is inspired by this chumsky example: https://github.com/zesterer/chumsky/blob/main/examples/nano_rust.rs. Beyond that, I went for the simple option of parsers that are pretty specific to the current implementation as it makes all the logic much simpler.

The bulk of the implementation here was claude's doing, prompted along the lines of "please rewrite this parsing logic in chumsky using the following example: https://github.com/zesterer/chumsky/blob/main/examples/nano_rust.rs", but I've had a good play with trying to add rules and it has been remarkably easy to do.

docs/src/references/mappers/configurable-bridge.md
```diff
@@ -0,0 +1,173 @@
...
+```toml
+# We can also make rules conditional
+[[rule]]
+if = "${connection.auth_method} == 'certificate'"
+topic = "s/dat"
+direction = "inbound"
+
+[[template_rule]]
+if = "${connection.auth_method} == 'password'"
+for = { item = "mode", in = ['s', 't', 'q', 'c'] }
```
albinsuresh: Though this `if` condition syntax feels more natural, I feel a slight inconsistency here. The iteration is expressed in a pure toml way using an inline table with `item` and `in` keys, wheras  the `if` directive is more of a templated style. I'm wondering why not use the same format for `if` as well, like:

```suggestion
if = { item = "${connection.auth_method}", eq = 'password'}
for = { item = "mode", in = ['s', 't', 'q', 'c'] }
```

That way we don't need to do any fancy parsing of the values. In addition to `eq`, we can provide more operators like `neq`, `gt`, `lt` etc as well.

docs/src/references/mappers/configurable-bridge.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Template rules
+
+You may wish to define multiple rules using a pattern:
+
+```toml
+[[template_rule]]
+# Iterating over an array in tedge config
+for = { item = "template", in = "${config.c8y.smartrest.templates}" }
+topic = "s/uc/${template}"
```
albinsuresh: There is definitely no harm with this syntaxt, and it definitely reads natural as well. But, I'm wondering if the ability to name the `item` reference is really useful or not. We could also just reserve the key `${item}` and directlly use that in the rule.
```suggestion
for = "${config.c8y.smartrest.templates}"
topic = "s/uc/${item}"
```

docs/src/references/mappers/configurable-bridge.md
jarhodes314: `for` definitely has a well defined structure, anything that `for` can reasonably be used for is something along the lines of `for item in iterable`, whereas `if` is a bit more complex. Already, we can do `if = "${c8y.mqtt_service.enabled}"` as well as checking the auth method, so I think the open-ended syntax is justified there.

docs/src/references/mappers/configurable-bridge.md
jarhodes314: This might be a sensible solution. Originally I implemented something similar, execute using `${@for}` instead of `${item}`. I decided in the end against `${@for}` because (a) it's another new bit of syntax and (b) it makes it look like you can reference more than just `for` (e.g. why can't I do `${@local_prefix}`?).

`item` might be a simpler option, and fits in nicely with there being a short, predefined list of possible values at the start of a `${...}` variable expansion (the others being `config` and `connection`). It also gives a nice resolution to your other point about the syntactic inconsistencies, so I think I'm in favour of this as a solution.

didier-wenzek: > I think I'm in favour of this as a solution.

So let's go with this simplification.

crates/extensions/tedge_mqtt_bridge/src/config_toml/mod.rs
jarhodes314: Now implemented all the parsing logic in chumsky

docs/src/references/mappers/configurable-bridge.md
jarhodes314: I've now implemented this change in syntax, though I still need to document it

crates/core/tedge_mapper/src/c8y/bridge/smartrest.toml
didier-wenzek: > For now at least, I have implemented the new syntax like this.

This has been simplified even further as `for = "${config.c8y.smartrest.templates}"` where the items binding name is fixed to `item`. See https://github.com/thin-edge/thin-edge.io/pull/3918#discussion_r2758990767

docs/src/references/mappers/configurable-bridge.md
```diff
@@ -0,0 +1,173 @@
...
+## Syntax
+
+A basic bridge rule looks something like:
+```toml
+[[rule]]
+local_prefix = "c8y/"
+remote_prefix = ""
+topic = "s/us"
+direction = "outbound"
+```
```
rina23q, line 30: Is QoS supported? Mosquitto bridge has that configuration.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -399,5 +306,242 @@ fn mqtt_bridge_config(
...
+) -> anyhow::Result<BridgeConfig> {
+    let bridge_config_dir = tedge_config
+        .mapper_config_dir::<C8yMapperSpecificConfig>(cloud_profile)
+        .join("bridge");
+
+    // Persist the built-in bridge configuration templates
+    persist_bridge_config_file(
+        &bridge_config_dir,
+        "smartrest",
+        include_str!("bridge/smartrest.toml"),
```
albinsuresh: Minor:
```suggestion
        include_str!("bridge/mqtt-core.toml"),
```
Since it's not just the **SmartREST** rules but covers **JSON-over-MQTT** topics as well. [Core MQTT](https://cumulocity.com/docs/device-integration/mqtt/) is the official term used by C8Y to refer to that endpoint.

crates/core/tedge_mapper/src/c8y/bridge/mqtt-service.toml
```diff
@@ -0,0 +1,13 @@
...
+
+[[rule]]
+local_prefix = "${config.c8y.bridge.topic_prefix}/mqtt/out/"
+topic = "#"
+direction = "outbound"
+
+[[template_rule]]
+for = { item = "topic", in = "${config.c8y.mqtt_service.topics}" }
+local_prefix = "${config.c8y.bridge.topic_prefix}/mqtt/in/"
+topic = "${topic}"
```
albinsuresh:
```suggestion
for = "${config.c8y.mqtt_service.topics}"
local_prefix = "${config.c8y.bridge.topic_prefix}/mqtt/in/"
topic = "${item}"
```

crates/core/tedge_mapper/src/c8y/bridge/smartrest.toml
```diff
@@ -0,0 +1,90 @@
+local_prefix = "${config.c8y.bridge.topic_prefix}/"
+remote_prefix = ""
+
+[[template_rule]]
+for = { item = "template", in = "${config.c8y.smartrest1.templates}" }
```
albinsuresh: New `for` syntax to be used across this file.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -399,5 +306,242 @@ fn mqtt_bridge_config(
...
+    cloud_profile: Option<&ProfileName>,
+) -> anyhow::Result<BridgeConfig> {
+    let bridge_config_dir = tedge_config
+        .mapper_config_dir::<C8yMapperSpecificConfig>(cloud_profile)
+        .join("bridge");
+
+    // Persist the built-in bridge configuration templates
+    persist_bridge_config_file(
+        &bridge_config_dir,
+        "smartrest",
```
albinsuresh: Minor:
```suggestion
        "mqtt-core.toml",
```
Since it's not just the **SmartREST** rules but covers **JSON-over-MQTT** topics as well. [Core MQTT](https://cumulocity.com/docs/device-integration/mqtt/) is the official term used by C8Y to refer to that endpoint.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -399,5 +306,242 @@ fn mqtt_bridge_config(
...
+        include_str!("bridge/smartrest.toml"),
+    )
+    .await?;
+    persist_bridge_config_file(
+        &bridge_config_dir,
+        "mqtt-service",
```
albinsuresh: Is this `smartrest.toml` vs `mqtt-service.toml` separation needed? With **MQTT Service** already in GA, this separation would be obsolete soon, as Cumulocity will be serving both the legacy topics and the custom topics via the same endpoint. So, I'd rather add those two mqtt-service specific rules into the main bridge file itself.

docs/src/references/mappers/configurable-bridge.md
jarhodes314: We could do, but we don't currently expose this in the internal API in the code. I'm also not entirely confident that the underlying QoS 2 support that exists actually works, since we've never used it. It's probably something that would be nice to support in future but I don't think this is worth solving in this PR.

crates/core/tedge_mapper/src/c8y/bridge/mqtt-service.toml
jarhodes314: Ah yes, I got carried away running the tests for `tedge_mqtt_bridge` locally and totally forgot to update the templates which are tested by the mapper.

crates/core/tedge_mapper/src/c8y/bridge/mqtt-core.toml
```diff
@@ -0,0 +1,103 @@
+local_prefix = "${config.c8y.bridge.topic_prefix}/"
```
Bravo555, line 6: suggestion: For out-of-the-box default rulesets like this one, would be nice to include a comment with description and a link to the documentation, so it's easier to learn for new users.

```suggestion
# A collection of MQTT bridge rules to convert between thin-edge MQTT and Cumulocity MQTT.
# Cumulocity MQTT Quick Reference: https://cumulocity.com/docs/smartrest/quick-reference/
# thin-edge.io Configurable built-in bridge reference: https://thin-edge.github.io/thin-edge.io/next/references/mappers/configurable-bridge/

local_prefix = "${config.c8y.bridge.topic_prefix}/"
```

crates/extensions/tedge_mqtt_bridge/Cargo.toml
Bravo555: could also place it in `tedge_mqtt_bridge/src/bin/test-bridge-rules.rs` instead and it works without needing to include this `[[bin]]` section.

docs/src/references/mappers/configurable-bridge.md
Bravo555: Also no longer true after 572a9133. Now there's only one file, `mqtt-core.toml`.

crates/core/tedge_mapper/src/c8y/mapper.rs
jarhodes314: Now done

crates/core/tedge_mapper/src/c8y/mapper.rs
jarhodes314: Now combined into `core-mqtt.toml`

crates/core/tedge_mapper/src/c8y/bridge/smartrest.toml
jarhodes314: Now done

crates/extensions/tedge_mqtt_bridge/src/config_toml/mod.rs
```diff
@@ -0,0 +1,1706 @@
...
+    where
+        Self: 'a;
+
+    fn expand(
+        &self,
+        config: TemplateConfig,
+        cloud_profile: Option<&ProfileName>,
+    ) -> Result<Self::Target, ExpandError> {
+        // TODO warn if there is no reference to the iterated variable
+        // TODO maybe add warnings if there are duplicate rules
```
Bravo555: todo: resolve TODOs

crates/extensions/tedge_mqtt_bridge/Cargo.toml
jarhodes314: This is true, but I think in the follow-up PR about debugging and linting tooling, I'll remove it anyway in favour of something in the `tedge` CLI.

crates/extensions/tedge_mqtt_bridge/src/config_toml/mod.rs
jarhodes314: I have moved these into #3964 (313ffaa)

crates/core/tedge_mapper/src/c8y/bridge/mqtt-core.toml
jarhodes314: Fixed 313ffaa

docs/src/references/mappers/configurable-bridge.md
jarhodes314: Created #3966 to capture this idea

crates/extensions/tedge_mqtt_bridge/src/persist.rs
```diff
@@ -0,0 +1,340 @@
...
+///
+/// - `{name}.toml` - active config (updated only if not overridden or disabled)
+/// - `{name}.toml.template` - canonical template (always updated)
+/// - `{name}.toml.disabled` - marker file indicating the config is disabled
+///
+/// This pattern allows users to customize their bridge configuration while still
+/// receiving template updates. If a user has modified the `.toml` file (making it
+/// differ from the `.toml.template`), or if a `.toml.disabled` file exists, the
+/// active config will not be updated.
+pub async fn persist_bridge_config_file(
```
didier-wenzek, line 29: We have to consider sharing this logic across definitions for flows and bridges. This version is more polished.

crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/template.rs
```diff
@@ -0,0 +1,618 @@
...
+            Token::Text(s) => write!(f, "{s}"),
+        }
+    }
+}
+
+/// Lexer for template strings
+///
+/// Produces a flat token stream including both variable references and literal text.
+/// Uses a recursive approach to handle the inside/outside variable context.
+fn lexer<'src>() -> impl Parser<
```
didier-wenzek, line 86: Why no reusing the same lexer as for condition strings?

Sure the latter introduces more token kinds (2), but this should not be a problem for the parser.

jarhodes314: They are two very distinct languages, the templates are whitespace-sensitive, whereas the conditions very much aren't. The point of introducing the separate lexing stages in the first place is to improve error messages ny breaking things down into meaningful tokens (e.g. some number of repeated `=` to form an operator) and we would totally go against that if we used the lowest common denominator tokenisation for both.

I should probably put these thoughts in doc comments though as it is not immediately obvious, and I have myself asked this question too.

crates/extensions/tedge_mqtt_bridge/src/persist.rs
jarhodes314: I would like to make it a semantic comparison (so parse the file into a `toml::Table` and compare those) so the comparison ignores e.g. bonus whitespace. I think it would be sensible to do this as part of unifying the logic for the bridge and flows so they both behave in the same way.

crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/template.rs
jarhodes314: Added doc comments: 550d47d15b

crates/extensions/tedge_mqtt_bridge/src/config_toml/parsing/condition.rs
```diff
@@ -0,0 +1,542 @@
...
+) -> impl Parser<'src, OffsetInput<'src>, Vec<Span<Token<'src>>>, extra::Err<Rich<'src, char, OffsetSpan>>>
+{
+    let var_start = just("${").to(Token::VarStart);
+    let var_end = just('}').to(Token::VarEnd);
+    let dot = just('.').to(Token::Dot);
+
+    // Operators: any combination of `!` and `=` (e.g., `==`, `!=`, `=`)
+    // We don't currently support != but may wish to in the future, and users may try this
+    // so recognising it but explicitly rejecting it is sensible
+    let op = one_of("!=")
```
albinsuresh: Minor:
```suggestion
    let op = one_of("!=&|")
```
Just for a more user firendly error message as users are likely to try binary `AND` and `OR` operators as well like `condition1 && condition2`. Such rules are still rejetected properly. Just that the error message is different, as follows:

| Expression | Error |
|---|---|
|`cond1 != cond2` | `found '!=' expected end of input` |
| `cond1 && cond2` | `Lexer error: found '&' expected '$', '}', '.', comparison operator (e.g. '=='), single-quoted string literal, identifier, or end of input` |

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -399,5 +306,230 @@ fn mqtt_bridge_config(
...
+) -> anyhow::Result<BridgeConfig> {
+    let bridge_config_dir = tedge_config
+        .mapper_config_dir::<C8yMapperSpecificConfig>(cloud_profile)
+        .join("bridge");
+
+    // Persist the built-in bridge configuration templates
+    persist_bridge_config_file(
+        &bridge_config_dir,
+        "mqtt-core",
+        include_str!("bridge/mqtt-core.toml"),
```
albinsuresh, line 324: My earlier naming suggestions were probably contradicting with each other. I suggested the name mqtt-core.toml for the earlier smartrest.toml file when the mqtt-service.toml was present separately. Now that they're both merged into one, I don't know what the right name would be. In a way core sounds fine, as the mqtt-servcie would become the new "core" in the near future.

docs/src/references/mappers/configurable-bridge.md
```diff
@@ -0,0 +1,173 @@
...
+sidebar_position: 3
+---
+
+import ProposalBanner from '@site/src/components/ProposalBanner'
+
+<ProposalBanner/>
+
+## Overview
+
+The built-in bridge forwards MQTT messages between the local broker and the cloud. By default, thin-edge provides bridge rules that handle standard communication patterns for Cumulocity, Azure, and AWS. However, you may need to customize this behavior—for example, to bridge additional topics, disable unused rules, or adapt to your cloud tenant's specific topic structure.
```
rina23q: A very minor thing:

```suggestion
The built-in bridge forwards MQTT messages between the local broker and the cloud. By default, %%te%% provides bridge rules that handle standard communication patterns for Cumulocity, Azure, and AWS. However, you may need to customize this behavior—for example, to bridge additional topics, disable unused rules, or adapt to your cloud tenant's specific topic structure.
```

jarhodes314: Done 44c6638

crates/core/tedge_mapper/src/c8y/mapper.rs
jarhodes314: Yeah, I don't entirely know either, although it's a pain to change it once we've merged this. I think `mqtt-core` is probably sufficient as that is still most of what this does (and it's what the mapper generally relies on).

didier-wenzek: `mqtt-core` is okay

PR #3920, opened 2026-01-12
test: re-enable c8y mqtt service basic auth test

albinsuresh:
## Proposed changes

The issue was on the server side with mqtt-service and they have fixed it now on the `eu-latest` env. Hence re-enabling the test.

- [x] re-enable c8y mqtt service basic auth test
- [x] switch vi to nocompatible mode in system test image (to avoid annoying characters while navigating with arrow keys)

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/images/debian-systemd/debian-systemd.dockerfile
+++ b/tests/images/debian-systemd/debian-systemd.dockerfile
@@ -36,6 +36,11 @@ RUN apt-get -y update \
     mosquitto \
     mosquitto-clients
 
+# Since vim.tiny is installed, vi command runs it under the hood, but in compatibility mode.
+# Disabling this classic compatibility mode to make it more user friendly (like vim) enabling features like:
+# arrow keys support in insert mode, multiple undo levels, etc.
+RUN echo "set nocompatible" >> /root/.vimrc
+
 # Note: Avoid using mosquitto 2.0.18 due to a session persistence bug when using `per_listener_settings true
 # Only comment out the custom install logic to make it easier to re-enable once the bug is resolved
 # See https://github.com/thin-edge/thin-edge.io/issues/3185 for more details


tests/images/debian-systemd/debian-systemd.dockerfile
```diff
@@ -36,6 +36,8 @@ RUN apt-get -y update \
     mosquitto \
     mosquitto-clients
 
+RUN echo "set nocompatible" >> /root/.vimrc
```
reubenmiller, line 42: Do you have an example of the behaviour without this setting? As I can't seem to reproduce the problem, and I'm surprised that you're using `vi` instead of `vim.tiny`...and if so, then maybe switching to vim.tiny (which is already installed) might fix the issue instead, and if so, then we could just set an alternative so that vi points to vim.tiny (e.g. `sudo update-alternatives --set vi /usr/bin/vim.tiny`)

albinsuresh: When I open a file in INSERT mode in `vi` and use the arrow keys to navigate the curos, it prints letters like `D`, `B` etc instead of moving the cursor. Only in the `INSERT` mode and not the normal mode. The navigation works just fine in the normal mode. This issue is only limited to `vi` and not `vim` or `vim.tiny`. 

I had also tried `update-alternatives --set vi /usr/bin/vim.tiny`, but that didn't fix the issue. Apparently, when `vim.tiny` is invoked via `vi`, it still behaves like `vi` in `compatible` mode, unlike `vim` which default to `nocompatible` mode whether or not it is invoked directly or via `vi` command.

reubenmiller: though then the question is, do you really need to use vi instead of vim.tiny?

reubenmiller: But I guess we can keep it, but please add a comment explaining why the line is needed within the code

reubenmiller: Or an alternative to create a vim shortcut to vim.tiny (using alternatives) which doesn't activate the vi compatibility mode:

```sh
update-alternatives --install /usr/bin/vim vim /usr/bin/vim.tiny 1

vim /etc/tedge/tedge.toml
```

albinsuresh: > though then the question is, do you really need to use vi instead of vim.tiny?

Just muscle memory. In most of the other systems that I've used, `vi` was always bound to `vim` anyway and hence I've never had this issue.

> Or an alternative to create a vim shortcut to vim.tiny (using alternatives) which doesn't activate the vi compatibility mode:

Sure, that can be also be done. But, I thought disabling the `compatible` mode was an even simpler solution. Is there any reason why we're trying to avoid that?

reubenmiller: Ok, we can stick with the compatibility mode, but just add a comment explaining what it is doing as it isn't obvious

albinsuresh: Added the comment in 9a17c61.

In fact the `vi` command was already running `vim.tiny` under-the-hood, but it is configured to behave like classic `vi` in compatibiity mode as configured in `/etc/vim/vimrc.tiny`. So, we're just overriding that. If `vim.tiny` was invoked using `vim` command, it would have run in `nocompatible` mode by default.

Issue #3905, opened 2026-01-05
HSM/`tedge-p11-server` Version 2 improvements


Bravo555:
Current `tedge-p11-server` and PKCS11 implementation has some problems that make it hard to debug, test and improve while maintaining compatibility:

# Error handling

## Serialisation format does not expose detailed error messages

If `tedge-p11-server` version is too old to support a given feature, when handling a request for that feature, the error message looks something like this:
```
Error: failed to Generate a keypair. # from tedge

Caused by:
	0: Failed to parse the received frame # from tedge-p11-server
	1: Hit the end of buffer, expected more data # from tedge-p11-server
```

### Proposed solution

We would like it to look like more like this:

```
Error: failed to Generate a keypair. # from tedge

Caused by:
	0: tedge-p11-server failed to process request: GenerateKeypair { ... } # from tedge
	1: tedge-p11-server response: Unrecognized request type "generate_keypair": Failed to parse payload: { ... }
```

This is currently impossible because currently used serialisation format, postcard, does not expose precise enough error information: for example, if `tedge-p11-server` receives a request id that it doesn't recognise, it returns `postcard::Error::SerdeDeCustom`, a generic error saying there was an error in de-serialisation, but nothing more.

To fix that, we would have to use a different format where this info is returned by the serialiser/de-serialiser, or try to effectively try to use a second, custom deserialize implementation to manually recover the error. There are other issues with postcard for our usecase though, so replacing it is probably better.

## Error interfaces are stringly/unstable

The errors that don't come from serde (i.e. PKCS11/invalid input errors) look like this:

```
$ tedge cert renew c8y
Error: failed to renew the device certificate via Cumulocity HTTP proxy http://127.0.0.1:8001
	
	Caused by:
		protocol error: bad response, expected get_public_key_pem, received: Error(ProtocolError("PKCS #11 service failed: Failed to find a key"))	
```

But on `tedge reconnect c8y`, the error is slightly different:
```
protocol error: bad response, expected chose scheme, received: Error(ProtocolError("PKCS #11 service failed: Failed to find a signing key: Failed to find a key"))
```

Top level tedge-p11-server request handlers currently return `anyhow::Error`s to which they can add context, and in the example above, in `reconnect` it's added and in `renew` it's not added. This makes it harder to properly match on error messages as the same root error can produce slightly different error messages because the errors are built bottom-up with string contexts without any stable top-level error type, so these errors messages are also changing as the lower-level code changes.

### Proposed solution

[...]

PR #3925, opened 2026-01-13
feat: tedge-p11-server JSON serialization

Bravo555:
## Proposed changes

Use JSON serialization in `Frame::Version2`, which allows us to maintain forward and backward compatibility much easier, because new optional fields can be added to current requests, unlike in `postcard`.

## Types of changes

## Paste Link to the issue

#3905 

## Checklist

### DIFF
--- a/crates/extensions/tedge-p11-server/src/proxy/frame2.rs
+++ b/crates/extensions/tedge-p11-server/src/proxy/frame2.rs
@@ -2,7 +2,7 @@ use anyhow::Context as _;
 use serde::{Deserialize, Serialize};
 
 use crate::{
-    proxy::{frame::ProtocolError, frame1::VersionInfo},
+    proxy::{error::Error, frame::ProtocolError, frame1::VersionInfo},
     service::{
         ChooseSchemeRequest, ChooseSchemeResponse, CreateKeyRequest, CreateKeyResponse,
         SignRequest, SignRequestWithSigScheme, SignResponse,
@@ -11,7 +11,7 @@ use crate::{
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Frame2 {
-    Error(ProtocolError),
+    Error(Error),
     ChooseSchemeRequest(ChooseSchemeRequest),
     SignRequest(SignRequest),
     ChooseSchemeResponse(ChooseSchemeResponse),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4847,6 +4847,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tempfile",
+ "thiserror 2.0.12",
  "tokio",
  "toml 0.9.6",
  "tracing",
--- a/crates/extensions/tedge-p11-server/Cargo.toml
+++ b/crates/extensions/tedge-p11-server/Cargo.toml
@@ -25,6 +25,7 @@ rustls.workspace = true
 sd-listen-fds.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+thiserror.workspace = true
 tokio = { workspace = true, features = [
     "macros",
     "rt-multi-thread",
--- a/crates/extensions/tedge-p11-server/src/proxy/client.rs
+++ b/crates/extensions/tedge-p11-server/src/proxy/client.rs
@@ -258,7 +258,7 @@ impl TedgeP11Client {
         debug!("Connected to socket");
 
         trace!(?request);
-        connection.write_frame(&request)?;
+        connection.write_frame(&Frame::Version1(request))?;
 
         let Frame::Version1(response) = connection.read_frame()? else {
             bail!("protocol error: bad response, expected version 1 frame");
--- a/crates/extensions/tedge-p11-server/src/proxy/connection.rs
+++ b/crates/extensions/tedge-p11-server/src/proxy/connection.rs
@@ -12,6 +12,9 @@ use anyhow::Context;
 use tracing::warn;
 
 use crate::proxy::frame::Frame2;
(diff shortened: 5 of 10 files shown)


didier-wenzek (commented):
I'm not really happy with idea to change the message format from `postcard` to `json`, even if I appreciate that `json` is a bit more flexible for our use case. Having to maintain both formats for backward compatibility reasons will make things complicated and somehow defeating the benefits of `json` and `serde`.

If I understand correctly your [motivations](https://github.com/thin-edge/thin-edge.io/issues/3905), there are two independent points you would like to improve.

- Error reporting from the server, to the client
- Forward and backward compatibility between the client and the server

I see these two points mostly unrelated, even if backward compatibility constraints might limit the freedom to change the error type. Furthermore, the feature is quite complete and I'm wondering if this is an appropriate time for these changes.

My questions:

- Can error reporting be improved through simpler means?  Even if a bit tedious, this might be more effective.
- Why adding `Frame2`, `Frame3` ... to the `enum Frame` is not enough to maintain backward compatibility?
   - `ClientV1 -> ServerV2` where a new field has been added, the server should fall back to `Frame1` de-serialization providing a default value for the new field.
  - Same for `ClientV2 -> ServerV1 -> ClientV2`.
- `serde-json` can help to provide default values for missing fields. But how `json` can help when a new field is not deserialized because the receiver is simply not aware of it?

crates/extensions/tedge-p11-server/src/proxy/frame1.rs
```diff
@@ -24,13 +25,35 @@ pub enum Frame1 {
     GetPublicKeyPemRequest(Option<String>),
     GetPublicKeyPemResponse(String),
     Ping,
-    Pong,
+    Pong(Option<VersionInfo>),
```
didier-wenzek, line 28: I would avoid to introduce such changes that hide a protocol version change behind optional arguments. This might work here because this is the optional version info is the last field of a `Pong` message, but this might also confuse the receiver of a `Pong(None)`: is the sender running an old version or not providing its version?

Hence, I would keep `Frame1` unchanged and add a `Frame2` variant:

```
pub enum Frame2 {
    Ping,
    Pong(VersionInfo),
    ...
}
```

A client V2 tries then using `Frame2` and fallbacks to  `Frame1` in case of errors:

```
    pub fn ping(&self) -> anyhow::Result<Option<VersionInfo>> {
        if let Frame2::Pong(version_info) = self.do_request(Frame2::Ping)? {
            return Ok(Some(version_info))
        };
        
        if let Frame1::Pong = self.do_request(Frame1::Ping)? {
            return Ok(None)
        };

        bail!("protocol error: bad response, expected pong, received: {response:?}")
    }
```

A server V2 returns a response using the same frame type:

```
   Frame2::Ping => Ok(Frame2::Pong(VersionInfo { version: 2 })),
   Frame1::Ping => Ok(Frame1::Pong),
```

crates/extensions/tedge-p11-server/src/proxy/frame2.rs
```diff
@@ -0,0 +1,34 @@
...
+    SignRequestWithSigScheme(SignRequestWithSigScheme),
+    GetPublicKeyPemRequest(Option<String>),
+    GetPublicKeyPemResponse(String),
+    Ping,
+    Pong(Option<VersionInfo>),
+    CreateKeyRequest(CreateKeyRequest),
+    CreateKeyResponse(CreateKeyResponse),
+    GetTokensUrisRequest,
+    GetTokensUrisResponse(Vec<String>),
+}
```
didier-wenzek, line 28: I would only add new or updated response variants. For the requests things are more subtle: if the new response might be more exhaustive then we must also include a new variant for the request. This is for instance the case for the `Ping` request which response includes  now a server version. 

```suggestion
pub enum Frame2 {
    Ping,
    Pong(VersionInfo),
}
```

For a request, where both the request and response payloads are unchanged, nothing has to be added in `Frame2`. This is for instance the case for `SignRequest` and `SignResponse`  The idea is that whatever their version, a client and a server can exchange  `SignRequest` and `SignResponse` using `Frame1`.

__In case a new error type is introduced__

If a new error type is introduced, we need to take extra care. Indeed, then the response to a request may be an error of the new v2 type. So even if the payload of a  `SignResponse` is unchanged, the client has to send a `Frame2::SignRequest` to get a extra info in the error case.

Hence the following suggestion for the `enum Frame2`:

```
pub enum Frame2 {
    Error(Error),
    ChooseSchemeRequest(ChooseSchemeRequest),
    SignRequest(SignRequest),


    SignRequestWithSigScheme(SignRequestWithSigScheme),
    GetPublicKeyPemRequest(Option<String>),

    Ping,
    Pong(Option<VersionInfo>),
    CreateKeyRequest(CreateKeyRequest),

    GetTokensUrisRequest,

}
```

A server V2 will have to handle both frame kind of requests, returning errors accordingly to the request kind:

```
   Frame2::SignRequest(r) => process(r).map(Frame1::SignResponse).map_err(Frame2::Error)
   Frame1::SignRequest(r) => process(r).map(Frame1::SignResponse).map_err(Frame1::Error)
```

A client V2 can either try first to guess the server version with a ping or to send directly a `Frame2` request and fallback to a `Frame1` request in case of an error.

That said I'm not convinced that we need to change the error type.

Bravo555: Okay, I now understand better what you have in mind and indeed only new requests can be overridden in new frames.
Only one change to that I'd like to make is to make server blind to specific frame version, but read `Request`s and return `Response` types, which would be then mapped to correct `Frame` at the `connection` module level.

PR #3926, opened 2026-01-13
feat: AWS mapper is configurable using flows

didier-wenzek:
## Proposed changes

- [x] AWS mapper mapping are implemented using tedge flows

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/docs/src/references/mappers/builtin-flows.md
+++ b/docs/src/references/mappers/builtin-flows.md
@@ -91,3 +91,61 @@ $ mv /etc/tedge/mappers/az/flows/mea.toml /etc/tedge/mappers/az/flows/mea.toml.d
 
 Alternatively, a builtin flow can be disabled by simply removing its definition
 and keeping the associated `.toml.template` file as a witness.
+
+## The AWS mapper
+
+The AWS mapper behavior is defined by a builtin flow located at `/etc/tedge/mappers/aws/flows/mea.toml`:
+
+```toml
+input.mqtt.topics = ["te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+/+/+/+/status/health"]
+
+steps = [
+  { builtin = "skip-mosquitto-health-status" },
+  { builtin = "add-timestamp", config = { property = "time", format = "unix", reformat = true } },
+  { builtin = "limit-payload-size", config = { max_size = 131072 } },
+  { builtin = "set-aws-topic", config = { prefix = "aws" } },
+]
+
+errors.mqtt.topic = "te/errors"
+
+```
+
+This file can be tuned using the `tedge config` command and the following settings:
+
+- `aws.topics`: the mapper input topics
+- `aws.mapper.timestamp`: whether the mapper should add a timestamp or not
+- `aws.mapper.timestamp_format`: the date format to be used for the timestamp (`unix` or `rfc-3339`)
+- `aws.mapper.mqtt.max_payload_size`: the maximum payload for a message (`131072` by default)
+- `aws.bridge.topic_prefix`: the prefix used for the bridge local MQTT topic  (`aws` by default)
+
+:::note
+For a change of one of these settings to be effective, the AWS mapper has to be restarted:
+```
+$ sudo tedge config set aws.mapper.timestamp_format rfc-3339
+$ sudo systemctl restart tedge-mapper-aws
+```
+:::
+
+This file can also be manually edited to:
+- Change step configurations
+- Add or remove steps
+- Substitute a JavaScript implementation for the builtin transformation.
+
+:::note
+If the builtin flow is updated, then the AWS mapper will not override its content,
+even if some `tedge config` settings have been updated. In such a case, the AWS mapper only
+update its flow template, i.e. the file `/etc/tedge/mappers/aws/flows/mea.toml.template`.
+:::
+
+If other flow definitions are provided along the builtin flow in the `/etc/tedge/mappers/aws/flows/` directory,
+then these flows are loaded by the AWS mapper.
+
+Finally, the builtin flow is disabled when replaced by a file with the same name and a `.toml.disabled` extension
+
+```
+# Disable the AWS mapper builtin flow:
+$ mv /etc/tedge/mappers/aws/flows/mea.toml /etc/tedge/mappers/aws/flows/mea.toml.disabled
+```
+
+Alternatively, a builtin flow can be disabled by simply removing its definition
+and keeping the associated `.toml.template` file as a witness.


crates/extensions/aws_mapper_ext/src/lib.rs
```diff
@@ -1,3 +1,810 @@
...
+
+    fn builtin_flow(&self) -> String {
+        let timestamp_step = if self.add_timestamp {
+            format!(
+                r#"{{ builtin = "add-timestamp", config = {{ property = "time", format = "{time_format}", reformat = true }} }},"#,
+                time_format = self.time_format,
+            )
+        } else {
+            "".to_string()
+        };
```
albinsuresh, line 72: Seeing this conditional addition of a step makes me wonder if all flow steps (builtin as well as scripts) should have an `enabled` option as well, so that the user can `enable` or `disable` a step without having to remove them from the file. It doesn't matter to those editing these files by-hand, as they can just comment a line. But it might be useful for programmatic manipluation.

crates/extensions/aws_mapper_ext/src/lib.rs
```diff
@@ -1,3 +1,810 @@
...
+        let result = converter.try_convert(&new_tedge_message(input)).await;
+
+        assert_eq!(result.unwrap()[0].payload_str().unwrap(), input);
+    }
+
+    #[tokio::test]
+    async fn try_convert_exceeding_threshold_returns_error() {
+        let mut converter = create_test_converter(false, Some(1), None).await;
+
+        let _topic = "te/device/main///m/".to_string();
```
albinsuresh: Why keep that unused value?

crates/extensions/aws_mapper_ext/src/lib.rs
```diff
@@ -1,3 +1,810 @@
...
+        let input = r#"{"temperature": 21.3}"#;
+        let _input_size = input.len();
```
albinsuresh: Why keep that unused value?

crates/extensions/aws_mapper_ext/src/lib.rs
```diff
@@ -1,3 +1,810 @@
...
+        let result = converter.try_convert(&new_tedge_message(input)).await;
+
+        assert_matches!(result, Err(FlowError::UnsupportedMessage(_)));
+    }
+
+    #[tokio::test]
+    async fn converting_input_without_timestamp_produces_output_without_timestamp_given_add_timestamp_is_false(
```
albinsuresh, line 198:
```suggestion
    async fn converting_measurement_without_timestamp_produces_output_without_timestamp_given_add_timestamp_is_false(
```
Just to be consistent with the other `convert_event_xyz` and `convert_alarm_xyz` tests. Same for the rest of the measurement tests. But, I understand that these tests are not new and were just moved from the old converter as-is.

crates/extensions/aws_mapper_ext/src/lib.rs
```diff
@@ -1,3 +1,810 @@
...
+    use tedge_flows::ConnectedFlowRegistry;
+    use tedge_flows::FlowResult;
+    use tedge_flows::MessageProcessor;
+    use tedge_flows::SourceTag;
+    use tedge_mqtt_ext::MqttMessage;
+    use time::macros::datetime;
+    static TE_MEA_TOPICS: &str =
+        r#"["te/+/+/+/+/m/+", "te/+/+/+/+/e/+", "te/+/+/+/+/a/+", "te/+/+/+/+/status/health"]"#;
+
+    fn new_tedge_message(input: &str) -> MqttMessage {
```
albinsuresh, line 169: Optional:
```suggestion
    fn new_tedge_message(topic: &str, input: &str) -> MqttMessage {
```
So that the same function can be reused in the tests for child devices, events, alarms etc.

crates/extensions/tedge_flows/src/transformers/skip_mosquitto_health_status.rs
```diff
@@ -0,0 +1,40 @@
...
+        Ok(())
+    }
+
+    fn on_message(
+        &self,
+        _timestamp: SystemTime,
+        message: &Message,
+    ) -> Result<Vec<Message>, FlowError> {
+        // don't convert mosquitto bridge notification topic
+        // https://github.com/thin-edge/thin-edge.io/issues/2236
```
albinsuresh, line 26: Not a request for change(for backward compatibility reasons), but I'm just wondering why we chose to ignore these and not to convert the payload into valid json like `{"status": "up"}` or `{"status": "up"}`. Yes, the `pid` field would be missing, but I don't think that was mandatory anyway.

albinsuresh (approved):
LGTM. The failed tests appear unrelated.

crates/extensions/aws_mapper_ext/src/lib.rs
didier-wenzek: This is a refactoring leftover. Removed  https://github.com/thin-edge/thin-edge.io/pull/3926/changes/e7d51107deaa7a03ce01265f84dc6c19a557f84d

crates/extensions/aws_mapper_ext/src/lib.rs
didier-wenzek: This is a refactoring leftover. Removed  https://github.com/thin-edge/thin-edge.io/pull/3926/changes/e7d51107deaa7a03ce01265f84dc6c19a557f84d

crates/extensions/aws_mapper_ext/src/lib.rs
didier-wenzek: > But, I understand that these tests are not new and were just moved from the old converter as-is.

This is precisely why I avoided to update the test logic.

crates/extensions/aws_mapper_ext/src/lib.rs
didier-wenzek: This is a legacy of the tests as written before the refactoring. I avoided to update the test logic.

crates/extensions/tedge_flows/src/transformers/skip_mosquitto_health_status.rs
didier-wenzek: Yeah, why not.

crates/extensions/aws_mapper_ext/src/lib.rs
didier-wenzek: Why not. There is already a request to enable / disable a flow.

That said, I'm not a big fan of this approach to disable a feature, a flow or a step.
Having explicit toggles in the flow definition adds extra code to ignore stuff that can easily be ignored using regular mechanisms:

- A step can be disabled by commenting out a line from the flow definition.
- A flow can be disabled by adding a non toml extension to the flow definition.

albinsuresh: > A step can be disabled by commenting out a line from the flow definition.

Manually doing that is fine. But doing the same with a script would be cumbersome, especially if the step definition spans over multiple lines as follows:

```toml
steps = [
    {
        enabled = false,
        builtin = "add-timestamp",
        config = {
            property = "time",
            format = "unix",
            reformat = false
        }
    }
]
```

With an `enabled` flag, we can use tools like `tomlq` to append that `enabled` setting entry easily.

PR #3931, opened 2026-01-15
feat: service restart support for file config plugin

albinsuresh:
## Proposed changes

- [x] extract `system_services` module from `tedge` into a standalone crate to use from the plugin
- [x] service restart support for file config plugin
- [x] update `config_update` workflow with granular operation steps
- [x] handle plugins restarting the `tedge-agent` itself
- [ ] ~how to detect a "reload" when service action is `reload` instead if `restart`~ to be handled using dedicated plugins as a generic logic in the `file` plugin is not possible

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
+++ b/tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
@@ -1,6 +1,7 @@
 *** Settings ***
 Resource            ../../../resources/common.resource
 Library             OperatingSystem
+Library             String
 Library             ThinEdgeIO
 Library             Cumulocity
 
@@ -127,7 +128,7 @@ Set Configuration Should Restart Service
     ${updated_tag}=    Execute Command    curl -I http://localhost 2>/dev/null | grep -i '^Server:'
     Should Contain    ${updated_tag}    tedge-lighttpd
 
-Set Configuration Should Restart Tedge Agent Service
+Set Configuration For system.toml Should Restart Tedge Agent Service
     Cumulocity.Set Device    ${PARENT_SN}
     ThinEdgeIO.Set Device Context    ${PARENT_SN}
 
@@ -155,6 +156,53 @@ Set Configuration Should Restart Tedge Agent Service
     # Verify the tedge-agent service was restarted by checking the PID changed
     ${pid_after}=    Get Service PID    tedge-agent
     Should Not Be Equal    ${pid_before}    ${pid_after}
+
+Set Configuration For tedge.toml Should Restart Tedge Agent Service
+    Cumulocity.Set Device    ${PARENT_SN}
+    ThinEdgeIO.Set Device Context    ${PARENT_SN}
+
+    # Config plugin TOML with service restart enabled for tedge-agent
+    ThinEdgeIO.Transfer To Device
+    ...    ${CURDIR}/tedge-configuration-plugin-with-service.toml
+    ...    /etc/tedge/plugins/tedge-configuration-plugin.toml
+    Should Contain Supported Configuration Types    tedge.toml
+
+    # Snapshot the current tedge.toml from the device via the cloud
+    ${timestamp}=    Get Unix Timestamp
+    ${snapshot_op}=    Cumulocity.Get Configuration    tedge.toml
+    ${snapshot_op}=    Operation Should Be SUCCESSFUL    ${snapshot_op}    timeout=60
+
+    # Retrieve the attachment that was uploaded by the snapshot operation
+    ${events}=    Cumulocity.Device Should Have Event/s
+    ...    type=tedge.toml
+    ...    with_attachment=${True}
+    ${contents}=    Cumulocity.Event Should Have An Attachment    ${events[0]["id"]}
+    ${config_content}=    Decode Bytes To String    ${contents}    UTF-8
+
+    # Update the existing config by disabling config snapshot capability,
+    # but keep config_update enabled so that the update operation resumes even after the agent restart.
+    # Selectively disabling config_snapshot to test that even that doesn't impact an ongoing config_update.
+    ${updated_content}=    Set Variable    ${config_content}\n\n[agent.enable]\nconfig_snapshot = false\n
+
+    ${pid_before}=    Get Service PID    tedge-agent
+
+    # Upload the updated content to Cumulocity and push it to the device
+    ${config_url}=    Cumulocity.Create Inventory Binary
+    ...    tedge.toml
+    ...    tedge.toml
+    ...    contents=${updated_content}
+    ${update_op}=    Cumulocity.Set Configuration    tedge.toml    url=${config_url}
+    ${update_op}=    Operation Should Be SUCCESSFUL    ${update_op}    timeout=60
+
+    # Verify the modified content was applied on the device
+    ${device_contents}=    ThinEdgeIO.Execute Command    cat /etc/tedge/tedge.toml    strip=${True}
+    Should Contain    ${device_contents}    config_snapshot = false
+
+    # Verify the tedge-agent service was restarted by checking the PID changed
+    ${pid_after}=    Get Service PID    tedge-agent
+    Should Not Be Equal    ${pid_before}    ${pid_after}
+
+    [Teardown]    Restore tedge.toml
 
 #
 # Get configuration
@@ -830,3 +878,7 @@ Enable config snapshot capability of tedge-agent
     Execute Command    tedge config set agent.enable.config_snapshot true
     ThinEdgeIO.Restart Service    tedge-agent
     ThinEdgeIO.Service Should Be Running    tedge-agent
+
+Restore tedge.toml
+    ThinEdgeIO.Set Device Context    ${PARENT_SN}
+    Execute Command    tedge config unset agent.enable.config_snapshot
--- a/tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin-with-service.toml
+++ b/tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin-with-service.toml
@@ -1,4 +1,5 @@
 files = [
     { path = '/etc/tedge/system.toml', type = 'system.toml', service = 'tedge-agent' },
+    { path = '/etc/tedge/tedge.toml', type = 'tedge.toml', service = 'tedge-agent' },
     { path = '/etc/lighttpd/lighttpd.conf', type = 'lighttpd-conf', service = 'lighttpd', service_action = 'restart' }
 ]
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -72,7 +72,6 @@ use tedge_uploader_ext::UploaderActor;
 use tedge_utils::file::create_directory_with_defaults;
 use tracing::info;
 use tracing::instrument;
-use tracing::warn;
 
 pub const TEDGE_AGENT: &str = "tedge-agent";
 
@@ -357,9 +356,9 @@ impl Agent {
             &self.config.service,
         );
 
-        // Instantiate config manager actor if config_snapshot or both operations are enabled
+        // Instantiate config manager actor if either config_snapshot or config_update operation is enabled
         let config_actor_builder: Option<ConfigManagerBuilder> =
-            if self.config.capabilities.config_snapshot {
+            if self.config.capabilities.config_snapshot || self.config.capabilities.config_update {
                 let manager_config = ConfigManagerConfig::from_options(ConfigManagerOptions {
                     config_dir: self.config.config_dir.clone().into(),
                     mqtt_topic_root: mqtt_schema.clone(),
@@ -368,6 +367,7 @@ impl Agent {
                     tmp_path: self.config.tmp_dir.clone(),
                     ops_dir: self.config.operations_dir.clone(),
                     is_sudo_enabled: self.config.is_sudo_enabled,
+                    config_snapshot_enabled: self.config.capabilities.config_snapshot,
                     config_update_enabled: self.config.capabilities.config_update,
                     plugin_dirs: self.config.config_plugin_dirs,
                 })?;
@@ -381,15 +381,16 @@ impl Agent {
                 .await?;
                 workflow_actor_builder.register_builtin_operation(&mut config_manager);
                 workflow_actor_builder.register_builtin_operation_step_handler(&mut config_manager);
-                workflow_actor_builder
-                    .register_sync_signal_sink(OperationType::ConfigSnapshot, &config_manager);
-                workflow_actor_builder
-                    .register_sync_signal_sink(OperationType::ConfigUpdate, &config_manager);
+                if self.config.capabilities.config_snapshot {
+                    workflow_actor_builder
+                        .register_sync_signal_sink(OperationType::ConfigSnapshot, &config_manager);
+                }
+                if self.config.capabilities.config_update {
+                    workflow_actor_builder
+                        .register_sync_signal_sink(OperationType::ConfigUpdate, &config_manager);
+                }
 
                 Some(config_manager)
-            } else if self.config.capabilities.config_update {
-                warn!("Config_snapshot operation must be enabled to run config_update!");
-                None
             } else {
                 None
             };
--- a/crates/extensions/tedge_config_manager/src/actor.rs
+++ b/crates/extensions/tedge_config_manager/src/actor.rs
@@ -865,12 +865,21 @@ impl ConfigManagerWorker {
 
         config_types.sort();
 
-        for topic in self.config.config_reload_topics.iter() {
-            let metadata = ConfigOperationData::Metadata {
-                topic: topic.clone(),
-                types: config_types.clone(),
-            };
-            self.output_sender.send(metadata).await?;
+        if self.config.config_snapshot_enabled {
+            self.output_sender
+                .send(ConfigOperationData::Metadata {
+                    topic: self.config.config_snapshot_meta_topic.clone(),
+                    types: config_types.clone(),
+                })
+                .await?;
+        }
+        if self.config.config_update_enabled {
+            self.output_sender
+                .send(ConfigOperationData::Metadata {
+                    topic: self.config.config_update_meta_topic.clone(),
+                    types: config_types.clone(),
+                })
+                .await?;
         }
         Ok(())
     }
--- a/crates/extensions/tedge_config_manager/src/config.rs
+++ b/crates/extensions/tedge_config_manager/src/config.rs
@@ -25,10 +25,12 @@ pub struct ConfigManagerConfig {
     pub tmp_path: Arc<Utf8Path>,
     pub ops_dir: Utf8PathBuf,
     pub mqtt_schema: MqttSchema,
-    pub config_reload_topics: Vec<Topic>,
+    pub config_snapshot_meta_topic: Topic,
+    pub config_update_meta_topic: Topic,
     pub config_update_topic: TopicFilter,
     pub config_snapshot_topic: TopicFilter,
     pub tedge_http_host: Arc<str>,
+    pub config_snapshot_enabled: bool,
     pub config_update_enabled: bool,
     pub sudo_enabled: bool,
 }
@@ -41,6 +43,7 @@ pub struct ConfigManagerOptions {
     pub tmp_path: Arc<Utf8Path>,
     pub ops_dir: Utf8PathBuf,
     pub is_sudo_enabled: bool,
+    pub config_snapshot_enabled: bool,
     pub config_update_enabled: bool,
     pub plugin_dirs: Vec<Utf8PathBuf>,
 }
@@ -54,10 +57,10 @@ impl ConfigManagerConfig {
         let plugin_config_dir = config_dir.join(DEFAULT_OPERATION_DIR_NAME);
         let plugin_config_path = plugin_config_dir.join(DEFAULT_PLUGIN_CONFIG_FILE_NAME);
 
-        let config_reload_topics = [OperationType::ConfigSnapshot, OperationType::ConfigUpdate]
-            .into_iter()
-            .map(|cmd| mqtt_topic_root.capability_topic_for(&mqtt_device_topic_id, cmd))
-            .collect();
+        let config_snapshot_meta_topic = mqtt_topic_root
+            .capability_topic_for(&mqtt_device_topic_id, OperationType::ConfigSnapshot);
+        let config_update_meta_topic = mqtt_topic_root
+            .capability_topic_for(&mqtt_device_topic_id, OperationType::ConfigUpdate);
 
         let config_update_topic = mqtt_topic_root.topics(
             EntityFilter::Entity(&mqtt_device_topic_id),
@@ -77,10 +80,12 @@ impl ConfigManagerConfig {
             tmp_path: cliopts.tmp_path,
             ops_dir: cliopts.ops_dir,
             mqtt_schema: mqtt_topic_root,
-            config_reload_topics,
+            config_snapshot_meta_topic,
+            config_update_meta_topic,
             config_update_topic,
             config_snapshot_topic,
             tedge_http_host: cliopts.tedge_http_host,
+            config_snapshot_enabled: cliopts.config_snapshot_enabled,
             config_update_enabled: cliopts.config_update_enabled,
             sudo_enabled: cliopts.is_sudo_enabled,
         })
--- a/crates/extensions/tedge_config_manager/src/lib.rs
+++ b/crates/extensions/tedge_config_manager/src/lib.rs
@@ -165,7 +165,10 @@ impl ConfigManagerBuilder {
 
     /// List of MQTT topic filters the log actor has to subscribe to
     fn subscriptions(config: &ConfigManagerConfig) -> TopicFilter {
-        let mut topic_filter = config.config_snapshot_topic.clone();
+        let mut topic_filter = TopicFilter::empty();
+        if config.config_snapshot_enabled {
+            topic_filter += config.config_snapshot_topic.clone();
+        }
         if config.config_update_enabled {
             topic_filter += config.config_update_topic.clone();
         }
@@ -261,14 +264,17 @@ impl IntoIterator for &ConfigManagerBuilder {
     type IntoIter = std::vec::IntoIter<Self::Item>;
 
     fn into_iter(self) -> Self::IntoIter {
-        let mut operation_senders = vec![(
-            OperationType::ConfigSnapshot.to_string(),
-            MappingSender::new(
-                self.box_builder.get_sender(),
-                generic_command_into_snapshot_request,
-            )
-            .into(),
-        )];
+        let mut operation_senders = vec![];
+        if self.config.config_snapshot_enabled {
+            operation_senders.push((
+                OperationType::ConfigSnapshot.to_string(),
+                MappingSender::new(
+                    self.box_builder.get_sender(),
+                    generic_command_into_snapshot_request,
+                )
+                .into(),
+            ));
+        }
         if self.config.config_update_enabled {
             operation_senders.push((
                 OperationType::ConfigUpdate.to_string(),
--- a/crates/extensions/tedge_config_manager/src/tests.rs
+++ b/crates/extensions/tedge_config_manager/src/tests.rs
@@ -116,19 +116,15 @@ async fn new_config_manager_builder(
             .unwrap()],
         plugin_config_dir: temp_dir.to_path_buf(),
         plugin_config_path: temp_dir.path().join("tedge-configuration-plugin.toml"),
-        config_reload_topics: [
-            "te/device/main///cmd/config_snapshot",
-            "te/device/main///cmd/config_update",
-        ]
-        .into_iter()
-        .map(Topic::new_unchecked)
-        .collect(),
+        config_snapshot_meta_topic: Topic::new_unchecked("te/device/main///cmd/config_snapshot"),
+        config_update_meta_topic: Topic::new_unchecked("te/device/main///cmd/config_update"),
         tmp_path: Arc::from(Utf8Path::from_path(&std::env::temp_dir()).unwrap()),
         ops_dir: temp_dir.dir("operations").utf8_path_buf(),
         mqtt_schema: MqttSchema::new(),
         config_snapshot_topic: TopicFilter::new_unchecked("te/device/main///cmd/config_snapshot/+"),
         config_update_topic: TopicFilter::new_unchecked("te/device/main///cmd/config_update/+"),
         tedge_http_host: "127.0.0.1:3000".into(),
+        config_snapshot_enabled: true,
         config_update_enabled: true,
         sudo_enabled: false,
     };


crates/common/tedge_system_services/src/services.rs
```diff
@@ -19,6 +19,8 @@ pub enum SystemService<'a> {
     #[strum(serialize = "tedge-agent")]
     /// TEdge SM agent
     TEdgeSMAgent,
+    /// Custom service with arbitrary name
+    Custom(&'a str),
```
didier-wenzek: I wonder if it makes sense to have all these enum types with a `Custom(String)` variant. This sounds type safe. But, these are only glorified strings and the `Custom(String)` variant makes this obvious.

tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin-with-service.toml
```diff
@@ -0,0 +1,3 @@
+files = [
+    { path = '/etc/lighttpd/lighttpd.conf', type = 'lighttpd-conf', service = 'lighttpd' }
```
reubenmiller: `service` isn't very clear that it is related to restarting the service, so I would recommend renaming it to `restart_service`.

plugins/tedge_file_config_plugin/src/lib.rs
```diff
@@ -15,20 +15,29 @@ use std::fs::File;
...
+use tedge_system_services::SystemService;
+use tedge_system_services::SystemServiceManager;
 use tedge_utils::atomic::MaybePermissions;
 use tedge_write::CopyOptions;
 use tedge_write::CreateDirsOptions;
 
 pub struct FileConfigPlugin {
     config: PluginConfig,
     use_tedge_write: TedgeWriteStatus,
+    config_dir: Utf8PathBuf,
```
didier-wenzek: ~Why do we need to know the config directory to restart a service?~

Oh, this is actually to get the `system.toml` file and not `tedge.toml`. I would consider to directly pass a `GeneralServiceManager`.

tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin-with-service.toml
reubenmiller: Alternatively have two properties, service and service_action where we could extend service_action in the future to also support "reload" (if we were to add support for that in the system.toml service abstraction):

```toml
[[files]]
path = '/etc/lighttpd/lighttpd.conf'
type = 'lighttpd-conf'
service = 'lighttpd'
service_action = 'restart'
```

plugins/tedge_file_config_plugin/src/config.rs
```diff
@@ -10,6 +10,12 @@ use std::hash::Hasher;
 use tedge_config::SudoCommandBuilder;
 use tedge_utils::file::PermissionEntry;
 
+/// Valid service actions in the context of configuration file management.
+/// Only the actions like and restart (and reload in future) are supported.
```
didier-wenzek, line 14:
```suggestion
/// Only the "restart" action is supported (and reload in future).
```

tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin-with-service.toml
albinsuresh: Introduced the `service_action` in [14e138f](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/14e138fdfe1e884a08c58ef9778f2c284fad05ce)

crates/common/tedge_system_services/src/services.rs
albinsuresh: Replaced the enum with a struct in [faed40f](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/faed40ffa4ef07d550fae7467d0fc46764085b53) that better captures our requirement to have both regular and templated services in a generic manner.

plugins/tedge_file_config_plugin/src/lib.rs
albinsuresh: Resolved by [14e138f](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/14e138fdfe1e884a08c58ef9778f2c284fad05ce)

tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin-with-service.toml
reubenmiller: The docs should be updated to reflect the new `service_action` property. Is there a default value for service_action?

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,38 @@
...
+[download]
+script = "curl --fail --show-error --output /tmp/config_update_${.topic.cmd_id} ${.payload.tedgeUrl}"
+on_success = "update"
+
+[update]
+script = "sudo /usr/share/tedge/config-plugins/file set ${.payload.type} /tmp/config_update_${.topic.cmd_id}"
+on_success = "apply"
+
+[apply]
+background_script = "sudo /usr/share/tedge/config-plugins/file apply ${.payload.type}"
```
albinsuresh: Even when we convert this step into a builtin task, this `plugin apply` command should be executed as a background task, irrespectve of the type of the config and whethere that type restarts `tedge-agent` or not. That'll smplify this workflow.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,38 @@
...
+on_exec = "verify"
+
+[verify]
+script = "sudo /usr/share/tedge/config-plugins/file verify ${.payload.type}"
```
albinsuresh: This verification here needs to be bullet-proof and this is something the user might as well get wrong. I myself encountered a situtaion where the `apply` state that does `systemctl restart tedge-agent` failed due to some sudo misconfiguration and this `verify` step that just does `systemctl is-active` couldn't catch that. The restart failure was ignored by the workflow (as designed) as it is a background script and the workflow happily moved to this `verify` state. And the `is-actve` check still passed, even though the restart never happend.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,38 @@
...
+on_success = "successful"
+
```
albinsuresh: Backup and rollback states would be added later.

tests/RobotFramework/tests/cumulocity/configuration/plugins/file
```diff
@@ -0,0 +1,84 @@
...
+    target_config="$2"
+    mv "$target_config" "$CONFIG_PATH"
+}
+
+apply_config() {
+    systemctl restart tedge-agent
+}
+
+verify_config() {
+    systemctl is-active tedge-agent
```
albinsuresh: This is not good enough, as `is-active` would pass even if the service never got restarted. The script should really have validated if the restart really happend. But, for some services, a reload would have happened instead of a restart. How can we detect that?

Add a `prepare` state to the workflow to fetch some runtime metadata (e.g: current pid for the restart case and specific service log statements for reload case) using the plugin **before** the update and pass that back to this `verify` state so that it can compare?

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,38 @@
...
+[download]
+script = "curl --fail --show-error --output /tmp/config_update_${.topic.cmd_id} ${.payload.tedgeUrl}"
+on_success = "update"
+
+[update]
+script = "sudo /usr/share/tedge/config-plugins/file set ${.payload.type} /tmp/config_update_${.topic.cmd_id}"
+on_success = "apply"
+
+[apply]
```
albinsuresh: The only reason why this `apply` state is separate from the `update` state is because it must be executed as a `background_script` to handle `tedge-agent` restart, while the `update` could be done as a regular action with immeidate feedback.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
reubenmiller: Though the downside of doing the background task is that the "verify" state will be executed before the "apply" command is done.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,38 @@
...
+[scheduled]
+action = "proceed"
+on_success = "executing"
+
+[executing]
+action = "proceed"
+on_success = "download"
+
+[download]
+script = "curl --fail --show-error --output /tmp/config_update_${.topic.cmd_id} ${.payload.tedgeUrl}"
```
reubenmiller: I assume this step will be replaced with internal logic which downloads the url if required and presents it to the "set" state...we're just using curl for now in the rapid prototyping.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,38 @@
...
+on_success = "update"
+
+[update]
```
reubenmiller: Since this is following the config plugin interface, I'd choose a name that follows the subcommand closely.

```suggestion
[set]
```

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,55 @@
...
+[init]
+action = "proceed"
+on_success = "scheduled"
+
+[scheduled]
+action = "proceed"
+on_success = "executing"
+
+[executing]
+script = "/etc/tedge/operations/workdir.sh"
```
albinsuresh: This script generates a temporary "working directory" that is passed to the plugin at different states. The agent will be doing this in the real impl. This directory can be used by the plugin to share any metadata between the states. Like saving backup files or any other matadata files, which are created in one state and used in another.

The agent will automatically remove this directory as well, once the operation completes.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,58 @@
...
+on_success = "download"
+
+[download]
+script = "curl --fail --show-error --output ${.payload.workdir}/config_update_${.topic.cmd_id} ${.payload.tedgeUrl}"
+on_success = "update"
+on_error = "failed"
+
+[prepare]
+script = "sudo /usr/share/tedge/config-plugins/file prepare ${.payload.type} ${.payload.workdir}"
```
albinsuresh: This state uses the working directory to backup the exitsng config and capture any metadat about the current running service (`pid` in this example).

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
```diff
@@ -0,0 +1,58 @@
...
+script = "sudo /usr/share/tedge/config-plugins/file set ${.payload.type} ${.payload.workdir}/config_update_${.topic.cmd_id}"
+on_success = "apply"
+
+[apply]
+background_script = "sudo /usr/share/tedge/config-plugins/file apply ${.payload.type} ${.payload.workdir}"
+on_exec = "verify"
+on_error = "rollback"
+
+[verify]
+script = "sudo /usr/share/tedge/config-plugins/file verify ${.payload.type} ${.payload.workdir}"
```
albinsuresh: The metadata captured in the `prepare` state (the `pid` of the running service) is used here to validate if the restart really succeeded or not.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
albinsuresh: Done in d270a9866ed7fc50dcc3720c812b2c55723bee7e as `prepare` and `finalize`.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
albinsuresh: Resolved by [55c2d76](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/55c2d767cee482ee9b759b6051a8aead3d16700b)

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
albinsuresh: Exactly.

tests/RobotFramework/tests/cumulocity/configuration/tedge-configuration-plugin-with-service.toml
albinsuresh: > Is there a default value for service_action?

Yes, `restart` is the default.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
albinsuresh: Since we have a `workdir` available (introduced in 55c2d76), how about using that in the `apply` state to write a "completion marker file" to that directory, which the `verify` phase can check for afterwards with a timeout? We just need to formalize the "completion marker file" contract.

reubenmiller: Though can't we just tell people not to restart the tedge-agent from within a plugin? And if they want to restart a service then pass back some data via the workflow state exchange mechanism (e.g. `:::begin-tedge:::`)?

albinsuresh: > And if they want to restart a service then pass back some data via the workflow state exchange mechanism (e.g. :::begin-tedge:::)?

Yeah, I had considered that option. But, these are the reasons why I tried to avoid this "restart request" exchange from the plugin to the agent via the command state:
1. We'll have to introduce yet another state after `apply` for the the agent to receive the restart request from the plugin, and then process it by restarting the agent. It feels like an optional state in the workflow only there for once specific case (the agent restart).
2. We'll be exposing the workflow state update contract (`:::begin-tedge:::` and `:::begin-tedge:::`) with the plugin, which was otherwise unaware of the workflow contracts. That's why I was looking for a simple file-system based contract via the `workdir`.
3. Consistency in the sense that they don't need to care if they are restarting the agent or any other service and don't need to do things differently for one vs the other. It would have been different if all service restarts were done by the agent. But since the agent running as `tedge` can't do that, it brings that inconsistency. Though, we can get arround this problem by making the agent use `tedgectl`.

But, I agree that this `workdir` approach introduces yet another mechanism for metdata exchange between states, when the MQTT state is already available. I thought this would be less of a problem as the `workdir` would anyway be used for config backups and even other things like service restart detection using persisted metadata(pids). So, this would be just one additional thing to that list.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
albinsuresh: Fixed in d270a98 with an improved pid check logic in the `verify` state.

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
albinsuresh: Resolved by [f5acace](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/f5acacef30acd49404bd1e0b94138b05790e4ee1) where the `workDir` is now generated and added to the payload in the `prepare` phase and cleaned up in the `finalize` or `rollback` phase.

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -41,7 +42,7 @@ pub struct OperationStepRequest {
     pub command_state: GenericCommandState,
 }
 
-pub type OperationStepResponse = Result<(), String>;
+pub type OperationStepResponse = Result<Option<Value>, String>;
```
didier-wenzek: The idea being to interpret `Ok(None)` as `Ok(Null)`, I would get rid of the `Option`.


```suggestion
pub type OperationStepResponse = Result<Value, String>;
```

tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
```diff
@@ -4,9 +4,14 @@ set -eu
...
-    echo "  get <type>                              Print the config for the specified type to stdout"
-    echo "  set <type> <new-config-path>            Update the config for the specified type from the new config path"
+    echo "  list                                                 List all config types supported by this plugin"
+    echo "  get <type>                                           Print the config for the specified type to stdout"
+    echo "  prepare <type> <new-config-path> --work-dir <dir>    Prepare for configuration update (create backup)"
+    echo "  set <type> <new-config-path>                         Update the config for the specified type from the new config path"
+    echo "  apply <type> --work-dir <dir>                        Apply configuration (restart services)"
+    echo "  verify <type> --work-dir <dir>                       Verify configuration was applied successfully"
+    echo "  finalize <type> --work-dir <dir>                     Finalize configuration update (cleanup)"
+    echo "  rollback <type> --work-dir <dir>                     Rollback configuration to previous state"
```
didier-wenzek, line 12: I failed to find the documentation for all the  sub-commands a configuration plugin has to implement, what is mandatory vs optional? Do we have such docs?

plugins/tedge_file_config_plugin/src/lib.rs
```diff
@@ -83,6 +92,112 @@ impl FileConfigPlugin {
...
+    }
+
+    /// Finalize configuration update
+    pub async fn finalize(
+        &self,
+        config_type: &str,
+        _workdir: &Utf8Path,
+    ) -> Result<(), PluginError> {
+        info!("Configuration update finalized successfully for {config_type}");
+        // No-op due for backward compatibility
```
didier-wenzek: Wouldn't be better for backward compatibility  to make the `prepare`, `finalize` and `rollback` sub-commands optional. I.e. Not mandating a plugin to provide a no-op implementation.

plugins/tedge_file_config_plugin/src/bin.rs
```diff
@@ -104,5 +180,69 @@ pub fn run(cli: FileConfigCli, tedge_config: TEdgeConfigView) -> anyhow::Result<
...
+            let workdir_path = Utf8PathBuf::from(work_dir);
+            match plugin.finalize(&config_type, &workdir_path).await {
+                Ok(()) => {
+                    log::info!("Successfully finalized configuration for {}", config_type);
+                    Ok(())
+                }
+                Err(err) => {
+                    log::error!("Failed to finalize configuration: {err}");
+                    Err(err.into())
+                }
```
didier-wenzek: I would consider as `Ok` the case where the plugin provides no specific implementation for `finalize`. Same for the other optional sub-commands.

didier-wenzek (commented):
I really like the new design for the communications between the workflow engine and an operation actor and this ability to extend the set of of atomic steps, supported for an operation, without touching the workflow engine.

plugins/tedge_file_config_plugin/src/lib.rs
albinsuresh: Well the backward compatibility is not in terms of the plugin API but in terms of the the previous config update behaviour. The plugin API (the subcommands) don't yet have a backward compatibility requirement because we haven't yet made a release yet with the existing config plugin APIs (`list`, `get` and `set`). Only the next release will introduce these APIs to the users and we can have the complete set (including `prepare`, `apply` and `verify` as well) included in the first release itself.

But in terms of behaviour, we're forced stay consistent with the previous contract (without plugins) where there was no `prepare` or `apply` or `finalize` logic performed. Even the `apply` for the service restart is performed only for the users that opts into that new feature with the `service = <service-name>` config. I can add another config setting for them to opt-into the `backup/restore` functionality as well.

tests/RobotFramework/tests/cumulocity/configuration/plugins/lighttpd
albinsuresh: All are mandatory subcommands. I'll add the missing docs.

That being said, we can make the subcommands like `prepare`, `apply`, finalize`, `verify` and `rollback` optional, if we find those excessive and feel that most users wouldn't need them. Not for backwards-compatibility reason, but purely for API simplicity. The backward-compatibility aspect is explained in this comment: https://github.com/thin-edge/thin-edge.io/pull/3931#discussion_r2839100230

plugins/tedge_file_config_plugin/src/bin.rs
albinsuresh: Will do that based on the decision on mandatory vs optional subcommands discussed in https://github.com/thin-edge/thin-edge.io/pull/3931#discussion_r2833246953

plugins/tedge_file_config_plugin/src/lib.rs
reubenmiller: I think we're doing way too many steps here. We should aim at having a very minimal set of actions so that we don't overcomplicate things (even more).

Can we review if all of the state are 100% required? For example the following states come to mind:
* verify (can't this be done in "set")
* apply (can't this be done in "finalize"?)

Maybe having an example of the sequence would be useful (e.g. mermaidejs flow diagram) to show the states would be useful to then try and cut-out the extra stuff.

albinsuresh: Here is the diagram:

```mermaid
stateDiagram-v2
    [*] --> executing

    executing --> prepare
    prepare --> set : successful
    prepare --> failed : failed

    set --> apply : successful
    set --> failed : failed

    apply --> verify : successful
    apply --> rollback : failed

    verify --> finalize : successful
    verify --> rollback : failed

    finalize --> successful: successful
    finalize --> failed: failed
    rollback --> failed

    successful --> [*]
    failed --> [*]
 ```

Comment (albinsuresh) [reply plugins/tedge_file_config_plugin/src/lib.rs]: Here are the options that I had considered to merge some states:
1. Merge `apply` into `set`:
   Decided against that as the failure state transition is different for both steps. If `set` fails (the existing config file is not even replaced with the new one), we can move to `failed` straightaway, as we don't need to attempt a rollback. But if the failure happens during `apply` after the config file is `set` (replaced), then we must try a rollback.
2. Combine `apply` and `verify`:
    Verify always felt like a standalone step as the logic would be independent from `apply`. Especially since this step needs to determine whether we `await-agent-restart` in the next state (if `tedge-agent` was restarted) or proceed directly to the `finalize` state (for other services).
3. `finalize` into `verify`:
     Though there isn't much relevance for `finalize` for the use-cases that I had considered, left it as an independent state as a natural counterpart of `rollback`. Here we can short-circuit the transition from `verify -> successful`.

Comment (reubenmiller) [reply plugins/tedge_file_config_plugin/src/lib.rs]: > 1

Maybe it is a naming thing as both "set" and "apply" are the same thing for a user. "I'm setting the configuration, and I'm applying the configuration to the device"...where in your example it is more doing some post actions afterwards...but still I'm not sure what value it brings sep

> 2

In its current state, I'd say finalize is useless here. It seems that it is only executed on the happy-path...I'd be interested if the sm-plugins behave in a similar way that the finalize command is never executed when a proceeding error occurred

File: crates/extensions/tedge_config_manager/src/resources/config_update.toml [hunk 19]
```diff
@@ -10,14 +11,33 @@ on_success = "download"
...
+on_success = "apply"
+
+[apply]
+action = "builtin:config_update:apply"
+on_success = "verify"
+
+[verify]
+action = "builtin:config_update:verify"
+on_success = "finalize"
+on_error = "rollback"
```
albinsuresh: To handle `tedge-agent` self-restart during a config update, we'll introduce a `restartAgent` flag in the payload that must be set by the plugin executed in the `apply` state to request an agent restart. The `apply` subcommand of the plugins can just print `:::begin-tedge:::\n {"restartAgent": true} \n :::end-tedge:::` to the stdout.

For handling this restart request we have the following alternatives: 

**Option A (Explicit states for agent restart)**

```
[apply]
action = "builtin:config_update:apply"
on_success = "evaluate-agent-restart"
on_error = "rollback"

[evaluate-agent-restart]
condition = "${.payload.restartAgent}"
on_success = "restart-agent"
on_error = "verify"

[restart-agent]
action = "restart-agent"
on_exec = "await-agent-restart"

[await-agent-restart]
action = "await-agent-restart"
timeout_second = 30
on_timeout = "rollback"
on_success = "finalize"

[verify]
action = "builtin:config_update:verify"
on_success = "finalize"
on_error = "rollback"
```

The major additions would be the new built-in tasks: `condition` and `restart-agent`.

We can even combine both `evaluate-agent-restart` and `restart-agent` into a single action as follows:

```
[conditional-agent-restart]
action = "restart-agent"
input.condition = "${.payload.restartAgent}"
on_exec = "await-agent-restart"
on_error = "verify"
```


**Option B (handle agent restart internally)**

Here the workflow remains the same and the restart in handled internally by the agent in the `apply` and `verify` states.

```
[apply]
action = "builtin:config_update:apply"
on_success = "verify"
on_error = "rollback"

[verify]
action = "builtin:config_update:verify"
on_success = "finalize"
on_error = "rollback"
```

In the apply phase, after the plugin's `apply` command is executed, if the agent receives the `restartAgent` flag, then it will move the workflow to the `verify` state and then shutdown itself. On restart, the workflow will continue from `verify` and the agent would check if a restart was requested (the `restartAgent` flag value) and deem it successful before preforming the plugin's `verify` command.

Unlike a `bg_script` that may or may not have triggered an agent restart, which forced the addition of the `await-agent-restart` action with timeout, here we don't need that, as the agent is guaranteed to perform its self-shutdown in the `apply` phase if `restartAgent` flag was set.

I prefer **Option B**, as my PoV is that the agent restart is just an internal business and hence those steps need not be exposed via workflow steps which are overridable and I don't see anyone realistically overriding the agent restart states at all.

didier-wenzek: The tricky part with an agent restart is not the action itself (`action = "restart-agent"`) but detecting that the restart has been effective and that the workflow can proceed. For that purpose there is an ["await-agent-restart"](https://thin-edge.github.io/thin-edge.io/next/references/agent/operation-workflow/#awaiting-the-agent-to-restart) action.

plugins/tedge_file_config_plugin/src/lib.rs
albinsuresh: With [this proposal](https://github.com/thin-edge/thin-edge.io/pull/3931#discussion_r2851196691) for handling agent self restart, I am now convinced that `apply` can be merged into `set` and `finalize` can be dropped as well. `set/apply` and `verify` must stay separate as there is the possibility of the agent restarting itself between them.

crates/extensions/tedge_config_manager/src/resources/config_update.toml
albinsuresh: > The tricky part with an agent restart is not the action itself (`action = "restart-agent"`) but detecting that the restart has been effective and that the workflow can proceed. For that purpose there is an ["await-agent-restart"](https://thin-edge.github.io/thin-edge.io/next/references/agent/operation-workflow/#awaiting-the-agent-to-restart) action.

Yes. And that's why `Option A` is using `await-agent-restart` combined with the `restart-agent` action. 

> Unlike a bg_script that may or may not have triggered an agent restart, which forced the addition of the await-agent-restart action with timeout, here we don't need that, as the agent is guaranteed to perform its self-shutdown in the apply phase if restartAgent flag was set.

What I was trying to say here is that `Option B` doesn't need that awaiting logic as the `verify` step is guaranteed to be executed only after the agent performs the shutdown in the `apply` state. Unless the agent fails to shutdown itself and still proceeds to process the next command state update for `verify`, which I assume is not possible.

albinsuresh: Further discussions revealed that **Option B** suffers from a race condition where the `verify` state transition from the `apply` state may not have reached the broker before the agent goes down. Hence, the assumption that the workflow would always resume from the `verify` state after the restart, is broken. So, for now we'll proceed with a simpler version of **Option A** as follows:

```
[apply]
action = "builtin:config_update:apply"
on_success = "evaluate-agent-restart"
on_error = "rollback"

[evaluate-agent-restart]
script = "test ${.payload.restartAgent} = \"true\""
on_exit.0 = "restart-agent"
on_exit.1 = "verify"

[restart-agent]
action = "restart-agent"
on_exec = "await-agent-restart"

[await-agent-restart]
action = "await-agent-restart"
timeout_second = 30
on_timeout = "rollback"
on_success = "finalize"

[verify]
action = "builtin:config_update:verify"
on_success = "successful"
on_error = "rollback"
```

albinsuresh: Resolved by commits [18af86c](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/18af86cbd97ebf3393d10b90187aebc8d1cdcd9d) and [7b83421](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/7b83421988714c4994c6ed6ecb4a56f77309b895)

plugins/tedge_file_config_plugin/src/bin.rs
albinsuresh: Obsolete as the smaller set of subcommands (after removing `apply` and `finalize`) are all mandatory.

plugins/tedge_file_config_plugin/src/lib.rs
albinsuresh: Resolved by [04e66f7](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/04e66f752fda84e0921c99c75bb84a00318b2771)

crates/core/tedge_api/src/workflow/mod.rs
albinsuresh: Resolved by [14ae61d](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/14ae61d4f91e4b4e7136db0f3209655999ac5837)

tests/RobotFramework/tests/cumulocity/configuration/composite_config_update.toml
albinsuresh: Resolved by [7b83421](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/7b83421988714c4994c6ed6ecb4a56f77309b895)

crates/core/tedge_actors/src/errors.rs
```diff
@@ -39,6 +39,9 @@ pub enum RuntimeError {
 
     #[error(transparent)]
     LinkError(#[from] LinkError),
+
+    #[error("A shutdown has been requested")]
+    Shutdown,
```
didier-wenzek: This doesn't look like an error.

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -348,6 +348,20 @@ impl WorkflowActor {
...
+
+                // First, move to the on_exec state and persist it
+                let new_state = state.clone().update(handlers.on_exec);
+                self.publish_command_state(new_state, &mut log_file).await?;
+
+                // Now request a shutdown to the runtime for a graceful shutdown
+                // The agent will be restarted by systemd and resume from the persisted state
+                log_file.log_info("Shutting down agent for restart").await;
+
+                return Err(RuntimeError::Shutdown);
```
didier-wenzek: This might be working but this is a hack.

The proper way to request a shutdown is to [send a RuntimeAction::Shutdown to the runtime](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/tedge_signal_ext/src/lib.rs#L81).

albinsuresh: I had tried that approach using `RuntimeAction::Shutdown` first. But the problem was that the runtime was treating it as a "graceful shutdown" and exiting with exit code 0, preventing the service from getting restarted. That's why I switched to this approach, by throwing an error, similar to the approach that we took for handling self-restart when `tedge` is self-updated. In the software management code, we are throwing a `RuntimeError::ActorError(Box::new(SoftwareManagerError::NotRunningLatestVersion))`. It felt like an indirect way of requesting for a shutdown. So, I thought of formalising it with a dedicated `RuntimeError::Shutdown` variant.

So the difference with both approaches in my mind is the following:
* Graceful shutdown without restart: `RuntimeAction::Shutdown`
* Shutdown to be restarted: `RuntimeError::Shutdown`

didier-wenzek: Okay, it makes sense. Somehow we are sending a `ConfigManagementError::NotUsingLatestConfiguration`. But we aim here to be less specific with an *agent restart* action.

As `RuntimeError::Shutdown` really doesn't sound as an error, what about `RuntimeError::RestartRequired`?

crates/core/tedge_actors/src/errors.rs
didier-wenzek:
```suggestion
    #[error("A restart is required: not running latest version and configuration ")]
    RestartRequired,
```

reubenmiller (changes requested):
I ran into an issue when trying to trigger a tedge-agent restart after applying a new tedge.toml file, but the operation fails for some reason, this line was found in the operation log at least:

```
ERROR: builtin action 'builtin:config_update:verify' failed: No builtin operation step handler registered for config_update operation verify step
```

Below is tedge-configuration-plugin.toml that I configured in my test:

**file: /etc/tedge/plugins/tedge-configuration-plugin.toml**

```toml
files = [
    { path = '/etc/tedge/system.toml', type = 'system.toml', service = 'tedge-agent' },
    { path = '/etc/tedge/tedge.toml', type = 'tedge.toml', service = 'tedge-agent' },
    { path = '/etc/lighttpd/lighttpd.conf', type = 'lighttpd-conf', service = 'lighttpd', service_action = 'restart' }
]
```
And the full operation log of the failed operation:
[TST_actualize_shiny_bend_workflow-config_update-c8y-mapper-53005959.log](https://github.com/user-attachments/files/25880830/TST_actualize_shiny_bend_workflow-config_update-c8y-mapper-53005959.log)

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: Resolved by [b1e3edb](https://github.com/thin-edge/thin-edge.io/pull/3931/commits/b1e3edb77aa33af944f1f89216de1b233cdb0a34)

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -279,6 +393,13 @@ Default plugin configuration
     ...    tedge.toml
     ...    tedge-log-plugin
 
+Config update restarts service if configured in plugin configuration
```
albinsuresh: To be removed, as I can't remember why I added this back then and the `Set Configuration Should Restart Service` test above covers what was intended anyway.

plugins/tedge_file_config_plugin/src/lib.rs
reubenmiller: @albinsuresh I just realized something, since we're now creating the `/etc/tedge/operations/config_update.toml` file within some initialization (or on startup), so we might have some subtle behaviour changes that we need to be aware of before merging, e.g.

* How is the config_update.toml updated over time when we have workflow changes?
* Do we allow users to have a their own custom config_update.toml which overrides the defualt behavior (I don't think we should, but we should document this to prevent users from trying to do it)?

reubenmiller: After an offline discussion both of these points will be addressed by https://github.com/thin-edge/thin-edge.io/issues/4000

crates/core/tedge_agent/src/software_manager/actor.rs
```diff
@@ -140,7 +140,7 @@ impl Actor for SoftwareManagerActor {
                         warn!("Tedge-agent is no more running the latest-version => a restart is required");
                         // Make sure the operation status is properly reported before the restart
                         tokio::time::sleep(Duration::from_secs(5)).await;
-                        return Err(RuntimeError::ActorError(Box::new(SoftwareManagerError::NotRunningLatestVersion)));
+                        return Err(RuntimeError::RestartRequired);
```
didier-wenzek, line 143: Good point.

PR #3934, opened 2026-01-21
feat: flow enabled c8y mapper

didier-wenzek:
## Proposed changes

The c8y mapper can be extended using flows.

- [x] The c8y mapper runs the flows defined in /etc/tedge/mappers/c8y/flows
- [x] The c8y mapper flow context holds entity definitions indexed by topic ids as well as external ids.
- [x] The builtin flow for measurements can be customized by users
- [x] The builtin flow for events can be customized by users
- [x] The builtin flow for alarms can be customized by users
- [x] The builtin flow for health status can be customized by users
- [x] Fix the broken unit tests
- [x] Cache MEA messages till the source entity is registered
- [x] Publish large events over HTTP 
- [x] Honor the `c8y.topics` settings to process only configured MEA messages
- [x] Remove dead code

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/flows.rs
+++ b/crates/extensions/c8y_mapper_ext/src/flows.rs
@@ -5,13 +5,20 @@ use tedge_flows::FlowRegistryExt;
 use tedge_flows::UpdateFlowRegistryError;
 use tedge_mqtt_ext::TopicFilter;
 use tedge_utils::file::create_directory_with_defaults;
+use tracing::error;
 
 impl C8yMapperBuilder {
     pub async fn flow_registry(
         &self,
         flows_dir: impl AsRef<Utf8Path>,
     ) -> Result<ConnectedFlowRegistry, UpdateFlowRegistryError> {
-        create_directory_with_defaults(flows_dir.as_ref()).await?;
+        if let Err(err) = create_directory_with_defaults(flows_dir.as_ref()).await {
+            error!(
+                "failed to create flow directory '{}': {err}",
+                flows_dir.as_ref()
+            );
+            return Err(err)?;
+        };
         let mut flows = ConnectedFlowRegistry::new(flows_dir);
 
         let mapper_topic_id = self.config.service_topic_id.clone();
--- a/crates/core/tedge_mapper/src/c8y/mapper.rs
+++ b/crates/core/tedge_mapper/src/c8y/mapper.rs
@@ -42,6 +42,7 @@ use tedge_timer_ext::TimerActor;
 use tedge_uploader_ext::UploaderActor;
 use tedge_utils::file::change_mode;
 use tedge_utils::file::change_user_and_group;
+use tedge_utils::file::create_directory_with_user_group;
 use tedge_watch_ext::WatchActorBuilder;
 use tracing::warn;
 use yansi::Paint;
@@ -336,9 +337,15 @@ pub async fn bridge_rules(
     tedge_config: &TEdgeConfig,
     cloud_profile: Option<&ProfileName>,
 ) -> anyhow::Result<BridgeConfig> {
-    let bridge_config_dir = tedge_config
-        .mapper_config_dir::<C8yMapperSpecificConfig>(cloud_profile)
-        .join("bridge");
+    let mapper_config_dir =
+        tedge_config.mapper_config_dir::<C8yMapperSpecificConfig>(cloud_profile);
+    if let Err(err) =
+        create_directory_with_user_group(mapper_config_dir.clone(), "tedge", "tedge", 0o755).await
+    {
+        warn!("failed to set file ownership for '{mapper_config_dir}': {err}");
+    }
+
+    let bridge_config_dir = mapper_config_dir.join("bridge");
 
     // Persist the built-in bridge configuration templates
     persist_bridge_config_file(


tests/RobotFramework/tests/cumulocity/flows/custom-measurements.js
```diff
@@ -0,0 +1,31 @@
...
+        "type": "custom",
+        [name]: {
+            [name]: {
+                "value": value
+             }
+        },
+    }
+
+    let entity_id = `${topic[1]}/${topic[2]}/${topic[3]}/${topic[4]}`
+    let entity = context.mapper.get(entity_id)
```
albinsuresh:
```suggestion
    let entity = context.mapper.entities.get(entity_id)
```
Unless we don't plan on exposing any other mapper specific context objects in the future.

Since this is just exposing the `entity_cache` maintained by the mapper, which is just a smaller subset of the real `entity_store` maintained by the agent, do we need to expose the global `entity_store` as well in the future? Or we expect users to query the entity store REST APIs directly from the javascript logic? Even the mapper will have to do the same behind the scenes, as not all mappers would be maintaining a local `entity_cache` with all info about the entities.

didier-wenzek: > `let entity = context.mapper.entities.get(entity_id)`

Sure, this would be more explicit and would avoid namespace issues when the context is  independently populated by the mapper and user scripts, However, this not so easy to implement in an extensible manner (i.e. a mapper being able to add its own namespaces). => I will see what are the options.

> Unless we don't plan on exposing any other mapper specific context objects in the future.

We already have to expose measurement units.

> Since this is just exposing the `entity_cache` maintained by the mapper, which is just a smaller subset of the real `entity_store` maintained by the agent, do we need to expose the global `entity_store` as well in the future? Or we expect users to query the entity store REST APIs directly from the javascript logic? Even the mapper will have to do the same behind the scenes, as not all mappers would be maintaining a local `entity_cache` with all info about the entities.

First, no REST APIs will ever be available from the javascript logic, which is by design effect free.

Second, what's the point to expose the whole entity store and not the cache? The c8y mapper is already working without these extra data.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -202,6 +203,9 @@ impl C8yMapperActor {
 
             // Convert and publish cached data messages
             for pending_data_message in pending_entity.data_messages {
+                // FIXME: The c8y converter is no processing these messages
+                //        - How to forward them to the builtin mea flows?
+                //        - An option is to let the flows manage the pending messages.
```
didier-wenzek: To be fixed before merge. If a measurement/event/alarm is received before the child device registration, then this mea is lost.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -195,13 +197,24 @@ impl C8yMapperActor {
...
         for pending_entity in pending_entities {
+            let entity_topic_id = pending_entity.reg_message.topic_id.to_string();
             let mut reg_message = pending_entity.reg_message;
             self.converter.append_id_if_not_given(&mut reg_message);
             let reg_message = reg_message.to_mqtt_message(&self.converter.mqtt_schema);
             self.process_message(reg_message).await?;
 
+            // TODO use proper JSON payload
+            let birth_message = MqttMessage::new(
+                &Topic::new_unchecked("te/device/main///status/entities"),
```
didier-wenzek: It would be better to use a topic that is owned by the mapper


```suggestion
                &Topic::new_unchecked("te/device/main/service/tedge-mapper-c8y/status/entities"),
```

crates/extensions/c8y_mapper_ext/src/actor.rs
didier-wenzek: Is this still useful? 

MEA messages are no more cached by the c8y converter but by the flows. The latter react on entity birth messages.

crates/extensions/c8y_mapper_ext/src/actor.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/346e155b19690898ebcc058e3bc9baf493dcdaa8

tests/RobotFramework/tests/cumulocity/flows/custom-measurements.js
albinsuresh: > Second, what's the point to expose the whole entity store and not the cache? The c8y mapper is already working without these extra data.

The only use-case that I can think of is for flow to the access to the command metadata for an entity, while it is processing a command. But, giving access to the entity store doesn't solve that either as the entity store doesn't keep track of all command metadata either.

crates/extensions/c8y_mapper_ext/src/mea/events.rs
```diff
@@ -0,0 +1,232 @@
...
+    fn on_message(
+        &mut self,
+        _timestamp: SystemTime,
+        message: &Message,
+        context: &FlowContextHandle,
+    ) -> Result<Vec<Message>, FlowError> {
+        match self.mqtt_schema.entity_channel_of(&message.topic) {
+            Ok((entity_id, Channel::Event { event_type })) => {
+                let Some(entity) = get_entity_metadata(context, entity_id.as_str()) else {
+                    self.cache.push(message.clone());
```
albinsuresh: We've had to repeat this caching logic across all builtin mea transformers. Instead of limiting this caching capability only to the builtin trasformers, I'm wondering if we could have extracted it out into a another standalone builtin step, which can be reused by user-provided flows as well.

crates/extensions/c8y_mapper_ext/src/mea/events.rs
```diff
@@ -0,0 +1,232 @@
...
+        }
+    }
+
+    pub fn http_event_topic_filter(c8y_prefix: &TopicPrefix) -> TopicFilter {
+        TopicFilter::new_unchecked(&format!("{c8y_prefix}/{C8Y_JSON_HTTP_EVENTS_TOPIC}/+"))
+    }
+
+    fn http_event_topic(&self, device: &EntityExternalId) -> String {
+        format!(
+            "{}/{C8Y_JSON_HTTP_EVENTS_TOPIC}/{}",
```
albinsuresh, line 156: How do we selectively stop the bridge listening on `c8y/#` messages  from picking this up?

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -195,13 +197,24 @@ impl C8yMapperActor {
...
             let mut reg_message = pending_entity.reg_message;
             self.converter.append_id_if_not_given(&mut reg_message);
             let reg_message = reg_message.to_mqtt_message(&self.converter.mqtt_schema);
             self.process_message(reg_message).await?;
 
+            self.mqtt_publisher.send(birth_message).await?;
+
             // Convert and publish cached data messages
             for pending_data_message in pending_entity.data_messages {
+                // TODO: Is this still useful?
```
albinsuresh, line 197: With the current impl, yes I don't see any value in keeping this.

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: Yes this can be done. The idea would be to have a first step that injects a source property to MEA messages , this property being set to the c8y external id of the source.

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: There is no bridge rules for this prefix (`c8y/http/` in practice). Hence, there are ignored.

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: > a first step that injects a source property to MEA messages

Actually the initial step can only check the source entity is fully registered. => I will implement that.

docs/src/references/mappers/builtin-flows.md
```diff
@@ -35,6 +35,118 @@ Their definitions can be customized in the same way as any user-defined flow.
...
+-rw-r--r-- 1 tedge   tedge    394 Jan 23 16:50 alarms.toml
+-rw-r--r-- 1 tedge   tedge    386 Jan 29 13:17 events.toml
+-rw-r--r-- 1 tedge   tedge    299 Jan 29 13:17 health.toml
+-rw-r--r-- 1 tedge   tedge    455 Jan 29 13:17 measurements.toml
+-rw-r--r-- 1 tedge   tedge    100 Jan 29 13:17 units.toml
+```
+
+These files can be customized at different levels:
+
+- Using the `tedge config` command to tune parameters such the maximum payload of messages sent to Cumulocity
```
rina23q:
```suggestion
- Using the `tedge config` command to tune parameters such as the maximum payload of messages sent to Cumulocity
```

tests/RobotFramework/tests/azure/azure_telemetry.robot
```diff
@@ -170,11 +170,12 @@ Updated builtin flows should not be overridden by the az mapper
...
     Disable Builtin Flow
-    Execute Command    (sleep 1; tedge mqtt pub te/device/main///e/ '{"text": "az builtin flow as been turned off"}')&
-    ${message}=    Execute Command
-    ...    tedge mqtt sub 'az/#' --duration 2s
-    ...    ignore_exit_code=${True}
-    Should Be Empty    ${message}
+    ${start}=    Get Unix Timestamp
+    Execute Command    tedge mqtt pub te/device/main///e/ '{"text": "az builtin flow as been turned off"}'
+    Should Not Have MQTT Messages
+    ...    topic=az
```
rina23q: A wildcard is missing?

```suggestion
    ...    topic=az/#
```

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: Implemented by https://github.com/thin-edge/thin-edge.io/pull/3934/changes/4c8e5269dc512c47fb5a82ac995aa5f2cf2e3092

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -109,6 +110,19 @@ pub struct Message {
...
+
+#[derive(Clone, serde::Deserialize, serde::Serialize, Eq, PartialEq)]
+pub enum Transport {
+    Mqtt {
+        #[serde(
+            serialize_with = "tedge_mqtt_ext::serialize_qos",
+            deserialize_with = "tedge_mqtt_ext::deserialize_qos"
+        )]
+        qos: QoS,
+        retain: bool,
```
rina23q, line 124: Nicee, I wanted this!

docs/src/references/mappers/builtin-flows.md
```diff
@@ -35,6 +35,118 @@ Their definitions can be customized in the same way as any user-defined flow.
 Currently, only the Azure mapper is defined using flows.
```
albinsuresh: That note can be removed now.

docs/src/references/mappers/builtin-flows.md
```diff
@@ -35,6 +35,118 @@ Their definitions can be customized in the same way as any user-defined flow.
...
 :::
 
+## The Cumulocity mapper
```
albinsuresh, line 34: The main user-facing mapper docs for each cloud should also link to this doc, right?

docs/src/references/mappers/builtin-flows.md
```diff
@@ -35,6 +35,118 @@ Their definitions can be customized in the same way as any user-defined flow.
...
+$ sudo systemctl restart tedge-mapper-c8y
+```
+:::
+
+### Editing builtin flows
+
+The builtin flows can be edited by users adding extra transformation steps.
+
+For instance, for builtin flow for measurements is defined by the file `/etc/tedge/mappers/c8y/flows/measurements.toml`,
+which default content is the following:
```
albinsuresh:
```suggestion
For instance, the builtin flow for measurements is defined by the file `/etc/tedge/mappers/c8y/flows/measurements.toml`,
with the default content as follows:
```

docs/src/references/mappers/builtin-flows.md
```diff
@@ -35,6 +35,118 @@ Their definitions can be customized in the same way as any user-defined flow.
...
+
+```toml
+input.mqtt.topics = ["te/+/+/+/+/m/+", "te/device/main/service/tedge-mapper-c8y/status/entities"]
+
+steps = [
+    { builtin = "add-timestamp", config = { property = "time", format = "unix", reformat = false } },
+    { builtin = "into_c8y_measurements", config = { topic_root = "te" } },
```
albinsuresh: Adding the `cache-early-messages` step as well:
```suggestion
    { builtin = "add-timestamp", config = { property = "time", format = "unix", reformat = false } },
    { builtin = "cache-early-messages",  config = { topic_root = "te" } },
    { builtin = "into_c8y_measurements", config = { topic_root = "te" } },
```

docs/src/references/mappers/builtin-flows.md
```diff
@@ -35,6 +35,118 @@ Their definitions can be customized in the same way as any user-defined flow.
...
+with files with the same name and a `.toml.disabled` extension
+
+```
+# Disable the Cumulocity mapper builtin flow for alarms:
+$ mv /etc/tedge/mappers/c8y/flows/alarms.toml /etc/tedge/mappers/c8y/flows/alarms.toml.disabled
+```
+
+Alternatively, a builtin flow can be disabled by simply removing its definition
+and keeping the associated `.toml.template` file as a witness.
+
```
albinsuresh, line 160:
```suggestion
:::note 
If the template file is removed, it will be recreated by the mapper on the next restart.
:::
```

crates/extensions/c8y_mapper_ext/src/flows.rs
```diff
@@ -0,0 +1,208 @@
...
+        let mqtt_schema = &self.config.mqtt_schema;
+        let topic_prefix = mqtt_schema.root.as_str();
+        let errors_topic = mqtt_schema.error_topic();
+
+        format!(
+            r#"
+[input.mqtt]
+topics = ["{topic_prefix}/+/+/+/+/m/+/meta"]
+
+[output.context]
```
albinsuresh: How about:
```suggestion
steps = [
    { builtin = "update-context" }
]
```

To showcase that built-in transformer as well, which is otherwise not used/documented anywhere.

crates/extensions/tedge_flows/src/transformers/update_context.rs
```diff
@@ -0,0 +1,58 @@
...
+    fn default() -> Self {
+        UpdateContext {
+            topics: TopicFilter::new_unchecked("#"),
+        }
+    }
+}
+
+impl Transformer for UpdateContext {
+    fn name(&self) -> &str {
+        "update-context"
```
albinsuresh, line 25: To be documented in the `flows.md` file along with other built-in transformers.

crates/extensions/c8y_mapper_ext/src/mea/entities.rs
```diff
@@ -0,0 +1,52 @@
...
+        te: &MqttSchema,
+        mapper: &EntityTopicId,
+        entity: &EntityTopicId,
+    ) -> MqttMessage {
+        let message_topic = te.topic_for(
+            mapper,
+            &Channel::Status {
+                component: "entities".to_string(),
+            },
+        );
```
albinsuresh, line 34: I'm starting to question if `te/device/main/service/tedge-mapper-c8y/status/entities` is the right target topic for these birth messages as it doesn't really represent any kind of `status` of `device/main/service/tedge-mapper-c8y` entity itself. Wondering if an internal topic like `c8y/entities` would have been more appropriate.

albinsuresh:
> @albinsuresh
> 
> Why do we skip twin data during serialization?
> 
> ```
> pub struct EntityMetadata {
>    ...
> 
>     #[serde(skip)]
>     pub twin_data: Map<String, JsonValue>,
> }
> ```
> 
> Because of that the `display_name` and `display_type` of an entity are not stored in the flows context nor passed to flow transformations. This is why the system test "Check if a service using configured service type" is failing, i.e. the service type reported to Cumulocity is always the default value.

Remnants of a historic mistake where the same `EntityMetadata` object was used as DMO to store all data associated with an entity (including the twin data) in the entity store, as well as the DTO that is passed around as the "entity metadata" for a single entity (even in REST APIs). For example, we didn't want to return the twin data of an entity for an entity metadata query (`/v1/entities/<topic-id>`), as the twin data had its own REST enpoint (`/v1/entities/<topic-id>/twin`).

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: For now, this is only available for builtin transformations, though.

docs/src/references/mappers/builtin-flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/fdb4eff41aaa3c8fa6557ae00f60664a7484ca2f

docs/src/references/mappers/builtin-flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/fdb4eff41aaa3c8fa6557ae00f60664a7484ca2f

docs/src/references/mappers/builtin-flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/fdb4eff41aaa3c8fa6557ae00f60664a7484ca2f

docs/src/references/mappers/builtin-flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/fdb4eff41aaa3c8fa6557ae00f60664a7484ca2f

tests/RobotFramework/tests/azure/azure_telemetry.robot
didier-wenzek: Oops. Thanks!

*Should not have* checks are really fragile. 

Fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/54d4634dae20196607e01e3e358c34a1ba175e8a

docs/src/references/mappers/builtin-flows.md
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/64ad98cbe43b9cd14fa2b79c6822c9b6521dbfc6

crates/extensions/c8y_mapper_ext/src/flows.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/64ad98cbe43b9cd14fa2b79c6822c9b6521dbfc6

crates/extensions/tedge_flows/src/transformers/update_context.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/64ad98cbe43b9cd14fa2b79c6822c9b6521dbfc6

docs/src/references/mappers/builtin-flows.md
didier-wenzek: Sure. But what would be the best location for these links? Any ideas?

albinsuresh: The only one that we have is https://thin-edge.github.io/thin-edge.io/understand/tedge-mapper/ under the `Conepts` section. Not the best location to document the extensibiity, though the counter argument would be that "extensibility" also is a concept.

Ideally I'd have added a document under the `Extend` section specifically for the mappers.

docs/src/references/mappers/builtin-flows.md
```diff
@@ -31,8 +31,124 @@ Their definitions can be customized in the same way as any user-defined flow.
...
+  - Messages published on this local topic are forwarded by the bridge to Cumulocity 
+- `c8y.mapper.mqtt.max_payload_size`
+  - Any message with a larger payload (after transformation) will be rejected
+  - In the case of events, larger events are published to Cumulocity over HTTP
+
+:::note
+For a change of one of these settings to be effective, the Cumulocity mapper has to be restarted:
+```
+$ sudo tedge config remove c8y.topics "te/+/+/+/+/m/+"
+$ sudo systemctl restart tedge-mapper-c8y
```
jarhodes314, line 76: This example feels either a bit weirdly incomplete (I'd expect something at the end where we observe the change in the generated flow), or it reads (to me at least) as if `tedge config remove ...` is part of restarting the mapper.

docs/src/references/mappers/builtin-flows.md
```diff
@@ -31,8 +31,124 @@ Their definitions can be customized in the same way as any user-defined flow.
...
+
+[errors.mqtt]
+topic = "te/errors"
+```
+
+Any modification to this file (adding, editing, removing a step as well as editing input and output)
+is dynamically reloaded and immediately made effective by the mapper.
+
+As an example, the `cache-early-messages` can be removed.
+Doing so any measurements received from a child device or a service that has not been properly registered yet.
```
jarhodes314: This doesn't make sense, I think it's missing some words?

crates/extensions/c8y_mapper_ext/src/flows.rs
```diff
@@ -0,0 +1,212 @@
...
+        let c8y_prefix = &self.config.bridge_config.c8y_prefix;
+        let max_size = self.config.max_mqtt_payload_size;
+
+        format!(
+            r#"input.mqtt.topics = ["{topic_prefix}/+/+/+/+/m/+", "{topic_prefix}/{mapper_topic_id}/status/entities"]
+
+steps = [
+    {{ builtin = "add-timestamp", config = {{ property = "time", format = "unix", reformat = false }} }},
+    {{ builtin = "cache-early-messages", config = {{ topic_root = "{topic_prefix}" }} }},
+    {{ builtin = "into_c8y_measurements", config = {{ topic_root = "{topic_prefix}" }} }},
```
jarhodes314: Is there a reason why this is underscore separated rather than hypen separated?

didier-wenzek: No reason. => I will fix that using hyphens for all builtin transformers

docs/src/references/mappers/builtin-flows.md
didier-wenzek: See https://github.com/thin-edge/thin-edge.io/pull/3934/changes/7c677cd94ea2dcf9c6bb42233718ccb50546b7d2

docs/src/references/mappers/builtin-flows.md
didier-wenzek: Fixed https://github.com/thin-edge/thin-edge.io/pull/3934/changes/7c677cd94ea2dcf9c6bb42233718ccb50546b7d2

crates/extensions/c8y_mapper_ext/src/flows.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/7bd5dba6d858adcde8f5fcabc54ca993d067f99b

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
```diff
@@ -0,0 +1,82 @@
...
+                FlowError::UnsupportedMessage(format!(
+                    "Invalid entity birth message received on {}: {err}",
+                    message.topic
+                ))
+            })?;
+
+        let mut messages = vec![];
+        let pending_messages = self.cache.take();
+        for message in pending_messages.into_iter() {
+            if message.topic.starts_with(&birth_message.topic) {
```
jarhodes314: This slightly threw me until I realised `birth_message.topic` is the topic prefix of the entity in question, not the topic the birth message was sent on.

I wonder whether the field could be renamed to clarify this:
```suggestion
            if message.topic.starts_with(&birth_message.entity_topic) {
```

and/or this could be encapsulated in a method on the `C8yEntityBirth`, e.g.
```suggestion
            if birth_message.matches_entity(&message.topic) {
```

crates/extensions/c8y_mapper_ext/src/mea/events.rs
```diff
@@ -0,0 +1,182 @@
...
+                error: e.to_string(),
+                payload: mqtt_payload.chars().take(50).collect(),
+            },
+        )
+        .map_err(|e| FlowError::UnsupportedMessage(format!("Not an event payload: {e}")))?;
+
+        let c8y_event = C8yCreateEvent::from(tedge_event);
+
+        // If the message doesn't contain any fields other than `text` and `time`, convert to SmartREST
+        let message = if c8y_event.extras.is_empty() {
```
jarhodes314: Very minor, but for symmetry with the else comment

```suggestion
        let message = if c8y_event.extras.is_empty() {
            // If the message doesn't contain any fields other than `text` and `time`, convert to SmartREST
```

crates/extensions/c8y_mapper_ext/src/mea/events.rs
```diff
@@ -0,0 +1,182 @@
...
+            let json_mqtt_topic = &format!("{}/{C8Y_JSON_MQTT_EVENTS_TOPIC}", self.c8y_prefix);
+            Message::new(json_mqtt_topic, cumulocity_event_json)
+        };
+
+        if self.can_send_over_mqtt(&message) {
+            // The message can be sent via MQTT
+            Ok(message)
+        } else {
+            // The message must be sent over HTTP
+            // Actually this converter forwards this message over MQTT to the c8y mapper which does the HTTP request
```
jarhodes314:
```suggestion
            // Actually this converter forwards this message over MQTT to the c8y converter which does the HTTP request
```

crates/extensions/c8y_mapper_ext/src/mea/events.rs
```diff
@@ -0,0 +1,182 @@
...
+use tedge_flows::FlowContextHandle;
+use tedge_flows::FlowError;
+use tedge_flows::JsonValue;
+use tedge_flows::Message;
+use tedge_mqtt_ext::TopicFilter;
+
+const DEFAULT_EVENT_TYPE: &str = "ThinEdgeEvent";
+const C8Y_JSON_MQTT_EVENTS_TOPIC: &str = "event/events/create";
+const C8Y_JSON_HTTP_EVENTS_TOPIC: &str = "http/events/create";
+const CREATE_EVENT_SMARTREST_CODE: u16 = 400;
```
jarhodes314, line 22: Is there a particular reason we're redefining this here instead of relying on `c8y_api::smartrest`?

crates/extensions/c8y_mapper_ext/src/mea/events.rs
```diff
@@ -0,0 +1,182 @@
...
+            .time
+            .format(&time::format_description::well_known::Rfc3339)
+            .map_err(|e| {
+                FlowError::UnsupportedMessage(format!("Fail to format timestamp as Rfc3339: {e}"))
+            })?;
+
+        Ok(format!(
+            "{},{},\"{}\",{}",
+            CREATE_EVENT_SMARTREST_CODE, c8y_event.event_type, c8y_event.text, time
+        ))
```
jarhodes314, line 173: Somewhat related to my duplication comment above about the smartrest message id, we should be using `SmartrestPayload::serialize` here to generate valid smartrest messages with proper escaping (since if `c8y_event.text` contains `"`, this will cause the generated message to be invalid).

crates/core/tedge_api/src/entity.rs
```diff
@@ -126,15 +126,22 @@ impl EntityMetadata {
     }
 
     pub fn display_name(&self) -> Option<&str> {
-        self.twin_data.get("name").and_then(|v| v.as_str())
+        self.twin_data
+            .get("name")
+            .and_then(|v| v.as_str())
+            .or_else(|| match self.r#type {
+                EntityType::MainDevice => None,
```
albinsuresh:
```suggestion
                EntityType::MainDevice => self.external_id.as_deref(),
```
Provided we add `Deref<str>` impl for `EntityExternalId`.

Or else:
```suggestion
                EntityType::MainDevice => self.external_id.as_ref().map(|id| id.as_ref())
```

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -274,22 +300,25 @@ impl EntityCache {
...
-                return Ok(false);
-            }
-        } else {
-            let existing = entity
-                .metadata
-                .twin_data
-                .insert(fragment_key, fragment_value.clone());
-            if existing.is_some_and(|v| v.eq(&fragment_value)) {
-                return Ok(false);
+        {
```
albinsuresh, line 309: Why this new scope?

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -373,6 +402,13 @@ impl EntityCache {
         self.entities.external_ids()
     }
 
+    fn update_flow_context(&self, topic_id: &EntityTopicId) {
+        let Some(entity) = self.get(topic_id) else {
+            return;
+        };
+        self.entities.update_context(topic_id, entity.clone())
```
albinsuresh, line 415: Why not call this directly from the `update_twin_data` function I wonder, as that function would have reached line 321 only if the entity was present in the cache.

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -468,6 +503,69 @@ impl EntityIndexes {
...
+    pub r#type: EntityType,
+
+    #[serde(rename = "@parent", skip_serializing_if = "Option::is_none")]
+    pub parent: Option<EntityTopicId>,
+
+    #[serde(rename = "@name")]
+    pub display_name: String,
+
+    #[serde(rename = "@type-name", skip_serializing_if = "Option::is_none")]
+    pub display_type: Option<String>,
```
albinsuresh: We never had those special `@` prefixes for `name` and `type` values in the `EntityRegistrationMessage` and `EntityMetadata`. So, if we introduce this special prefix only for C8Y, I'm wondering if it would confuse the users as they switch between using tedge registration APIs and writing these javascript transformers.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
```diff
@@ -31,22 +31,19 @@ pub(crate) fn convert_health_status_message(
...
     let external_id = entity.external_id.as_ref();
-    let display_name = entity
-        .metadata
-        .display_name()
-        .or_else(|| entity.metadata.topic_id.default_service_name())
-        .unwrap_or(external_id);
-
-    let display_type = entity.metadata.display_type().unwrap_or("service");
+    let display_name = entity.display_name();
+    let display_type = entity.display_type().unwrap_or("service");
```
albinsuresh: Wouldn't it be better if that `.unwrap_or("service")` fallback was also moved into the `CloudEntityMetadata::display_type()`?

docs/src/references/mappers/builtin-flows.md
```diff
@@ -101,8 +106,14 @@ topic = "te/errors"
...
 is dynamically reloaded and immediately made effective by the mapper.
 
-As an example, the `cache-early-messages` can be removed.
-Doing so any measurements received from a child device or a service that has not been properly registered yet.
+As an example, by default, the Cumulocity mapper caches measurements, events and alarms when received too soon,
+i.e. before the source child-device or service has been properly registered.
+These messages are cached till their source has been properly registered.
+The behavior is implemented by the `cache-early-messages` builtin transformation step.
+If this behavior is not desired, one can simply remove that step from the `measurements.toml` flow definition.
+A child device will have then to be properly registered for its measurements to be forwarded to Cumulocity.
```
albinsuresh, line 114: This `cache-early-messages` definition also needs be added to the `Builtin transformations` section in `flows.md` as this is not a Cumulocity specific builtin action anymore but a generic one, unlike `into-c8y-measurements`.

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: No specific reasons, except I just moved code around. This was previously in the c8y `converter.rs`.

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: You are right. However, this is not new and the focus of this PR has been to change the way message transformations are triggered and not on the processing details.

We can consider to improve this code. But I think, it's better to do it independently.

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/ef026d1bccbd3626a96bb70c05b9fea1bd26de6d

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/a5ac786ac2accdb1020560535930f7b3e47aafee

crates/extensions/c8y_mapper_ext/src/mea/events.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/a5ac786ac2accdb1020560535930f7b3e47aafee

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
didier-wenzek: To release the exclusive reference on `self` acquired with `let entity = self.try_get_mut()`, so we can proceed with `self.update_flow_context()`

An alternative is to extract a method.

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
didier-wenzek: For similar reasons as for https://github.com/thin-edge/thin-edge.io/pull/3934#discussion_r2767892983

The `entity` line `321` is bind to an exclusive `&mut`.

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
didier-wenzek: What would be your proposal?

An alternative, I considered, is to serialize the twin data in the context and doing so to avoid any new properties - which purpose is just to pass around this fields which are mandatory but skipped by `EntityMetadata` serialization.

About users being confused,  I have to confess that I'm confused by these metadata, twin data, and `@` prefixes.

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
didier-wenzek: Sure this would be cleaner, and this have been my first attempt. But what should be the default for the main device and for child devices?

What is the difference between the `type` and the `display_type`?

crates/extensions/c8y_mapper_ext/src/mea/events.rs
jarhodes314: > We can consider to improve this code. But I think, it's better to do it independently.

Makes sense

docs/src/references/mappers/builtin-flows.md
didier-wenzek: This is something specific to Cumulocity: messages are cached till a c8y entity birth message is published.

That said, docs are indeed missing for all Cumulocity builtin transformers.

crates/core/tedge_api/src/entity.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/04cae36f55ce0484a0c58da42ad66d4940b0f63e

crates/extensions/c8y_mapper_ext/src/service_monitor.rs
albinsuresh: Here are the mapping rules:

* The `@type` value determines the type of the MO in C8Y
* The `type` value is mapped differently for child devices and services:
   - For child devices, tedge `type` -> c8y `type`
   - For services, tedge `type` -> c8y `serviceType` as the `type` is hardcoded as `c8y_Service`
   
And hence the C8Y "display type" is derived from twin `type` of the entity. For child devices it is the c8y `type` itself. For services it is the c8y `serviceType`.

didier-wenzek: Suggestion applied: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/3058e85073d04191cd362951e867b949b25f2f1f

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -63,6 +68,34 @@ impl CloudEntityMetadata {
...
+            .unwrap_or_else(|| self.external_id.as_ref())
+    }
+
+    pub fn display_type(&self) -> &str {
+        self.metadata
+            .display_type()
+            .unwrap_or(match self.metadata.r#type {
+                EntityType::MainDevice => "thin-edge.io",
+                EntityType::ChildDevice => "thin-edge.io-child",
+                EntityType::Service => "service",
```
albinsuresh, line 96:
```suggestion
                EntityType::ChildDevice => `tedge_config.device.type`,
                EntityType::Service => `tedge_config.service.type`,
```
The defaults are already defined in the `tedge config`.

albinsuresh: Ignore the previoud comment. What you had done was correct. This `display_type` fallback is used when am entity is registered without an explicit `type` in the payload. The  `device.type` and `service.type` configs are not meant to be used for this purpose. Those configs are only to be used by tedge services (mapper, agent etc) in their registration messages that they publish during their startup.

In other words, those config values are the types of tedge devices and services only and not the  fallback default values for any random entity that is registered/converted with thin-edge.

crates/extensions/c8y_mapper_ext/src/flows.rs
```diff
@@ -0,0 +1,212 @@
...
+                .persist_builtin_flow("health", self.health_flow().as_str())
+                .await?
+        } else {
+            flows.disable_builtin_flow("health").await?;
+        }
+
+        Ok(())
+    }
+
+    fn units_flow(&self) -> String {
```
jarhodes314, line 75: It feels like we could do with some simple unit tests to just go through these flows and check each one generates valid toml syntax

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
```diff
@@ -40,43 +40,47 @@ impl tedge_flows::Transformer for MessageCache {
...
                 if get_entity_metadata(context, entity_id.as_str()).is_none() {
                     self.cache.push(message.clone());
                     return Ok(vec![]);
                 };
+
+                // In case the current message has been received before the entity birth message
+                // all the messages cached for that entity have to be processed first
+                let mut messages = self.restore_messages(&message.topic);
+                messages.push(message.clone());
+                Ok(messages)
```
jarhodes314, line 79: It feels like this rather defeats the point of the birth messages in the first place, but I'm guessing the issue here is that messages on the same topic end up out of order?

My gut feeling is that this is an issue because this code has multiple sources of truth. On one branch, we're waiting to receive a birth message, but this branch is relying on the entity store to decide when an entity exists. Is there some way this flow can maintain its own state about which birth messages have been received so we can use that as the source of truth?

(I appreciate that this may also be a quick solution to try and solve the system test issues)

crates/extensions/c8y_mapper_ext/src/flows.rs
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/120ad140e0e1ff26cb29ab48456e421f46fefae6

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
didier-wenzek: > It feels like this rather defeats the point of the birth messages in the first place, but I'm guessing the issue here is that messages on the same topic end up out of order?

This half defeats the birth messages. They are still required for singleton messages.

The issue is that messages on *different* topics an not necessarily received in order. Here, birth and status messages can be received out of order. The sequence of events before that fix was:

1.  "c8y mapper down" received as the very first message of the mapper (sent by the previous instance of the mapper) => cached because the mapper service is not yet registered.
2. The mapper registers itself and the flow context is populated with the c8y service metadata.
3. The mapper sends two messages: service birth and status
4. "c8y mapper up"  is received *first* by the flow, and successfully processed because the context is correctly populated.
5. "c8y service register" is received *second* by the flow. This triggers the recovery of the "c8y mapper down" message which is then processed too late overriding the up message.

> My gut feeling is that this is an issue because this code has multiple sources of truth. On one branch, we're waiting to receive a birth message, but this branch is relying on the entity store to decide when an entity exists. Is there some way this flow can maintain its own state about which birth messages have been received so we can use that as the source of truth?

I agree that too many related points are handled by too many independent actors: registration on c8y (c8y converter), birth messages (c8y converter), health status (availability actor), populating the context (c8y converter), using the context (flows).

However, the issue is because of a lack of knowledge of which birth messages have been received. Somehow the flows have this knowledge: the cache is cleared when a  birth message is processed.

That said, I don't have a clear idea of what would be a better design. Having a single component managing the context and the cache would remove the race conditions. But where should live this component? In the mapper? the flows ?
> (I appreciate that this may also be a quick solution to try and solve the system test issues)

Yes, this is the good news: health messages are now timely processed.

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+impl MessageCache {
+    /// Retrieve from the cache all the messages related to the entity with the given topic
+    pub fn restore_messages(&mut self, entity_topic: &str) -> Vec<Message> {
+        let mut messages = vec![];
+        let pending_messages = self.cache.take();
+        for message in pending_messages.into_iter() {
+            if message.topic.starts_with(entity_topic) {
```
albinsuresh: With this logic, if there are some cached messages for some entities (e.g: `child1` and `child2`) in this cache, and when messages are received for another entity (e.g: `child0`) that is already registered, we'll be unnecessarily draining the cache first, only to put them all back. And we repeat that for every single message.

We could make this logic more efficient, by changing the `cache` from a simple `RingBuffer<Message>` to `HasMap<EntityTopicId, RingBuffer<Message>>`  so that this step only consumes the messages for the current entity and leave the ones from other entities untouched. But, this is just a matter of efficiency and not correctness. So, it can be improved in a later PR.

crates/extensions/c8y_mapper_ext/src/mea/entities.rs
albinsuresh: Decision deferred for later.

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
didier-wenzek: > With this logic, if there are some cached messages for some entities (e.g: `child1` and `child2`) in this cache, and when messages are received for another entity (e.g: `child0`) that is already registered, we'll be unnecessarily draining the cache first, only to put them all back. And we repeat that for every single message.

The normal case is to have an empty cache and if not empty this will be for not too long, the main purpose of this cache being to solve races between registration and mea messages.

I see one annoying case though. A measurement which has been sent by mistake on behalf of a non-existing entity on a device configured not to use auto-registration. No registration message will ever be sent and the measurement sent by mistake will stay in the cache for ever. Worse, this measurement will be considered again and again for all cache drains.

In other word, I see more a cache pollution problem than a performance problem.

> We could make this logic more efficient, by changing the `cache` from a simple `RingBuffer<Message>` to `HasMap<EntityTopicId, RingBuffer<Message>>`  so that this step only consumes the messages for the current entity and leave the ones from other entities untouched. But, this is just a matter of efficiency and not correctness. So, it can be improved in a later PR.

Indeed, this would solve the issue of the unknown-source measurement polluting the drain process for all messages and all entities.

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
jarhodes314: Another thought, I'm not sure the comment is accurate. Because we're using the message topic and not the entity topic prefix for clearing the cache, we're only retrieving messages on that topic (which is perfectly adequate for the fix), but the comment suggests we retrieve _all_ messages for the entity, and I don't believe we do.

didier-wenzek: > the comment suggests we retrieve all messages for the entity, and I don't believe we do.

Indeed, we don't.

I decided to fix the behavior rather than the comment: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/af732feb1892469976d743a4fec5794ba99d749d

This has also been an opportunity to address @albinsuresh 's comment: https://github.com/thin-edge/thin-edge.io/pull/3934#discussion_r2773853718

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
didier-wenzek: > We could make this logic more efficient, by changing the `cache` from a simple `RingBuffer<Message>` to `HasMap<EntityTopicId, RingBuffer<Message>>`  so that this step only consumes the messages for the current entity and leave the ones from other entities untouched.

Done https://github.com/thin-edge/thin-edge.io/pull/3934/changes/af732feb1892469976d743a4fec5794ba99d749d

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    fn on_message(
+        &mut self,
+        _timestamp: SystemTime,
+        message: &Message,
+        context: &FlowContextHandle,
+    ) -> Result<Vec<Message>, FlowError> {
+        match self.mqtt_schema.entity_channel_of(&message.topic) {
+            Ok((_, Channel::Status { component })) if component == "entities" => {
```
jarhodes314: Do we also run into issues here if we have multiple c8y mappers running simultaneously (e.g. with multiple profiles)? Should we not be checking the service topic ID is for this mapper, and ignoring birth messages for other mappers?

tests/RobotFramework/tests/cumulocity/flows/c8y_flows.robot
```diff
@@ -0,0 +1,135 @@
+*** Settings ***
+Resource            ../../../resources/common.resource
+Resource            ../../azure/azure_telemetry.robot
```
reubenmiller: This line looks to be accidentally added in and should be deleted. Sometimes the autocomplete from the RobotFramework plugin is not so helpful and shows you keywords provided by other .robot files, and not just from the libraries, so it automatically adds the associated import if you select the keyword from another file (which is not so obvious from the autocompletion list).

reubenmiller: I pushed a fixup commit to remove this line. c3108ed16e177ffa42d1786436a1e5eaaedba2a8

crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
didier-wenzek: Agree. This is related to https://github.com/thin-edge/thin-edge.io/pull/3934#discussion_r2757599567

As the birth messages are related to a c8y mapper instance, we have to use a c8y related topic for those. Precisely this needs to be a sub-topic of `c8y.bridge.topic_prefix` for the appropriate profile.

didier-wenzek: Actually, the current code is working because the flows using this cache subscribe to a c8y mapper instance topic (aka ` te/device/main/service/tedge-mapper-c8y/status/entities`).

- This can made more obvious with a check on the entity sending the birth message (instead of ignoring is it with `_`).
- Independently , we can also change the topic as proposed by @albinsuresh https://github.com/thin-edge/thin-edge.io/pull/3934#discussion_r2757599567

jarhodes314: > - This can made more obvious with a check on the entity sending the birth message (instead of ignoring is it with `_`).

I think this is wise. It feels like a bit of a hack to rely on "we don't subscribe to other conflicting topics" to ensure we only respond to this mapper's messages.

didier-wenzek: >> - This can made more obvious with a check on the entity sending the birth message (instead of ignoring is it with `_`).

> I think this is wise. It feels like a bit of a hack to rely on "we don't subscribe to other conflicting topics" to ensure we only respond to this mapper's messages.

Done: https://github.com/thin-edge/thin-edge.io/pull/3934/changes/61d725992e562a07c7d5de5873567aab7577ffac

Issue #3939, opened 2026-01-26
Allow users to disable support for the c8y_Restart operation


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Some users would like to disable the support for the c8y_Restart command so that users can't restart the device remotely.

Currently if the user deletes the c8y_Restart operation declaration and then restarts the mapper (or agent), then the file is recreated (as it is created when `tedge init` is called).

```sh
sudo rm -f /etc/tedge/operations/c8y/c8y_Restart
sudo systemctl restart tedge-mapper-c8y
test -f /etc/tedge/operations/c8y/c8y_Restart
```

**Describe the solution you'd like**

Add a new configuration called `c8y.enable.restart` which can be used to disable support for the declaration of the c8y_Restart operation handler for Cumulocity.

The new config item would align with the other similarly named configuration items that already exist and serve a similar purpose.

```sh
$ tedge config list c8y.enable
c8y.enable.log_upload=true
c8y.enable.config_snapshot=true
c8y.enable.config_update=true
c8y.enable.firmware_update=true
c8y.enable.device_profile=true
```

**Describe alternatives you've considered**

**Additional context**

PR #3940, opened 2026-01-26
feat: support disabling cloud handling of c8y_Restart operation

reubenmiller:
## Proposed changes

Support new tedge configuration (`c8y.enable.device_restart`) to disable the handling of the the `c8y_Restart` operation from Cumulocity.

Other Cumulocity cloud operations already support this mechanism, so the PR just aligns the c8y_Restart with the other cloud operation handlers.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3939

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -699,6 +699,10 @@ define_tedge_config! {
             /// Enable device_profile feature
             #[tedge_config(example = "true", default(value = true))]
             device_profile: bool,
+
+            /// Enable device restart feature
+            #[tedge_config(example = "true", default(value = true))]
+            device_restart: bool,
         },
 
         mapper: {
--- a/crates/extensions/c8y_mapper_ext/src/converter.rs
+++ b/crates/extensions/c8y_mapper_ext/src/converter.rs
@@ -657,7 +657,12 @@ impl CumulocityConverter {
 
         let msgs = match C8yDeviceControlOperation::from_json_object(extras)? {
             C8yDeviceControlOperation::Restart(_) => {
-                self.forward_restart_request(device_xid, cmd_id)?
+                if self.config.capabilities.device_restart {
+                    self.forward_restart_request(device_xid, cmd_id)?
+                } else {
+                    warn!("Received a c8y_Restart operation, however, device_restart feature is disabled");
+                    vec![]
+                }
             }
             C8yDeviceControlOperation::SoftwareUpdate(request) => {
                 self.forward_software_request(device_xid, cmd_id, request)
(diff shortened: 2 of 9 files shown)


crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -699,6 +699,10 @@ define_tedge_config! {
             /// Enable device_profile feature
             #[tedge_config(example = "true", default(value = true))]
             device_profile: bool,
+
+            /// Enable device restart feature
+            #[tedge_config(example = "true", default(value = true))]
+            device_restart: bool,
```
reubenmiller, line 705: I'm not sure whether we should use `device_restart` or just `restart`. I opted for a more explicit approach given that some time in the future we'll probably have a "service_restart" command.

tests/RobotFramework/tests/cumulocity/restart/restart_device.robot
```diff
@@ -32,6 +32,23 @@ tedge-mapper-c8y does not react to local restart operations transitions
     topic=te/device/main///cmd/restart/local-2222    payload={"status":"failed"}    expected_status=failed    c8y_fragment=c8y_Restart
     topic=te/device/main///cmd/restart/local-3333    payload={"status":"successful"}    expected_status=successful    c8y_fragment=c8y_Restart
 
+Supports disabling the Cumulocity c8y_Restart Command
+    Execute Command    tedge config set c8y.enable.device_restart false
+    Execute Command    rm -f /etc/tedge/operations/c8y/c8y_Restart
```
jarhodes314, line 37: Does a user need to do both these actions to disable the operation, or is deleting the file just done so we can verify it doesn't reappear after restarting the mapper?

reubenmiller: Yes I think the file still needs to be deleted if the tedge-mapper-c8y service has already been started (with the setting on)...though I didn't verify this, so maybe I'm wrong.

reubenmiller: I just confirmed that once the c8y_Restart file is created, then the user has to manually remove it as disabling the feature does not affect the supported operations list that is sent to Cumulocity. However the mapper won't react to the operation if it is sent to the cloud.

We could delete the operation if present on startup though. Previously we avoided this as we thought that people might want to customize the operation handler (e.g. create the file themselves), however I don't think anyone is doing this now that we can customize the tedge-agent workflows instead.

The code shows that there is only an "add_operation" function, and nothing to remove an already existing one.

[pub async fn add_operation(](https://github.com/reubenmiller/thin-edge.io/blob/98947118a5051f91574491df2593b907fe117282/crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs#L72)

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -657,7 +657,12 @@ impl CumulocityConverter {
 
         let msgs = match C8yDeviceControlOperation::from_json_object(extras)? {
             C8yDeviceControlOperation::Restart(_) => {
-                self.forward_restart_request(device_xid, cmd_id)?
+                if self.config.capabilities.device_restart {
+                    self.forward_restart_request(device_xid, cmd_id)?
+                } else {
+                    warn!("Received a device restart operation, however, device_restart feature is disabled");
```
rina23q: To keep consistency with other warning messages.

```suggestion
                    warn!("Received a c8y_Restart operation, however, device_restart feature is disabled");
```

reubenmiller: Good catch. Fixed in 53259f7226bad991a1a8d88b521dc302e79832c9

PR #3941, opened 2026-01-26
fix: Disable partial download if file being downloaded is modified between retries

Bravo555:
## Proposed changes

In current download implementation, where in case of interruption we try to resume the download using a HTTP range request to avoid re-downloading parts we already downloaded, there is a potential issue where if the file is updated by the server between retries, we can miss this and corrupt the file.

This PR adds a check where if the file is modified, we abort the range request and request the full file again with a normal `GET` request.

In particular, the check if file was modified compares `ETag` header value if it exists, between current and previous request. If it is different than ETag from previous request, we request full range of the file again. If there's no `ETag`, behaviour is unchanged and we proceed with the range request, so as to not disable partial requests for servers that don't send an `ETag`, for example Cumulocity Inventory Binaries.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

One more thing we could do is check if the `Content-Length` of the resource is the same, hopefully catching some updates where size of the file changes. Unfortunately currently I am unable to test this because to test partial requests we use chunked transfer encoding, and when it's in use, `reqwest` ignores `Content-Length` header and doesn't report it (maybe idea being that chunked transfer encoding is used inherently for streaming requests, where you most often don't know the size ahead of time). So currently, this check is not added, and we only check ETag if it's present.

Some servers also send a `Last-Modified` header, which can be added (unfortunately Cumulocity also doesn't support it).

There's also option to use `Want-Content-Digest` header to request a digest from the server, which we could then also compute locally to perform an integration check before finishing the download, but so far I haven't seen a server that supports it.

### DIFF
--- a/crates/common/download/src/download.rs
+++ b/crates/common/download/src/download.rs
@@ -6,6 +6,7 @@ use anyhow::anyhow;
 use backoff::future::retry_notify;
 use backoff::ExponentialBackoff;
 use certificate::CloudHttpConfig;
+use http::StatusCode;
 use log::debug;
 use log::info;
 use log::warn;
@@ -348,7 +349,15 @@ impl Downloader {
             request
                 .send()
                 .await
-                .and_then(|response| response.error_for_status())
+                .and_then(|response| {
+                    if response.status() != StatusCode::RANGE_NOT_SATISFIABLE {
+                        response.error_for_status()
+                    } else {
+                        // if range not satisfiable, request should be retried with different range
+                        // (or without)
+                        Ok(response)
+                    }
+                })
                 .map_err(reqwest_err_to_backoff)
         };
 
@@ -385,6 +394,8 @@ async fn save_chunks_to_file_at(
         writer.write_all(&bytes)?;
     }
     writer.flush()?;
+    let end_pos = writer.stream_position()?;
+    writer.set_len(end_pos)?;
     Ok(())
 }
 
--- a/crates/common/download/src/download/partial_response.rs
+++ b/crates/common/download/src/download/partial_response.rs
@@ -43,6 +43,10 @@ pub(super) fn response_range_start(
             Ok(PartialResponse::PartialContent(pos))
         }
 
+        // Resource could've been modified such that range we requested is invalid (e.g. because new resource is
+        // smaller)
+        StatusCode::RANGE_NOT_SATISFIABLE => Ok(PartialResponse::ResourceModified),
+
         // We don't expect to receive any other 200-299 status code, but if we
         // do, treat it the same as OK
         status_code if status_code.is_success() => Ok(PartialResponse::CompleteContent),
--- a/crates/common/download/src/download/tests/partial_response.rs
+++ b/crates/common/download/src/download/tests/partial_response.rs
@@ -208,3 +208,125 @@ async fn resume_download_with_etag_changed() {
 
     server_task.abort();
 }
+
+/// If after retrying ETag of the resource is different, we should download it
+/// from scratch again.
+#[tokio::test]
+async fn resumed_download_doesnt_leave_leftovers() {
+    let file_v1 = "AAAABBBBCCCCDDDD";
+    let file_v2 = "XXXXYYYY";
+
+    let listener = TcpListener::bind("localhost:0").await.unwrap();
+    let port = listener.local_addr().unwrap().port();
+
+    // server should do 3 things:
+    // - only send a portion of the first request
+    // - after the first request update the resource
+    // - serve 2nd request normally (but we expect it to be a range request)
+    let server_task = tokio::spawn(async move {
+        let mut request_count = 0;
+        while let Ok((mut stream, _addr)) = listener.accept().await {
+            let response_task = async move {
+                let (reader, mut writer) = stream.split();
+                let mut lines = BufReader::new(reader).lines();
+                let mut range: Option<std::ops::Range<usize>> = None;
+
+                // We got an HTTP request, read the lines of the request
+                while let Ok(Some(line)) = lines.next_line().await {
+                    if line.to_ascii_lowercase().contains("range:") {
+                        let (_, bytes) = line.split_once('=').unwrap();
+                        let (start, end) = bytes.split_once('-').unwrap();
+                        let start = start.parse().unwrap_or(0);
+                        let end = end.parse().unwrap_or(file_v2.len());
+                        range = Some(start..end)
+                    }
+                    // On `\r\n\r\n` (empty line) stop reading the request
+                    // and start responding
+                    if line.is_empty() {
+                        break;
+                    }
+                }
+
+                let file = if request_count == 0 { file_v1 } else { file_v2 };
+                let etag = if request_count == 0 {
+                    "v1-initial"
+                } else {
+                    "v2-changed"
+                };
+
+                if let Some(range) = range {
+                    let start = range.start;
+                    let end = range.end;
+                    let msg = if start >= end || end >= file.len() {
+                        let header = format!(
+                            "HTTP/1.1 416 Range Not Satisfiable\r\n\
+                            content-range: bytes */{}\r\n",
+                            file.len()
+                        );
+                        format!("{header}\r\n")
+                    } else {
+                        // Return range for both first and subsequent requests
+                        let header = format!(
+                            "HTTP/1.1 206 Partial Content\r\n\
+                            transfer-encoding: chunked\r\n\
+                            connection: close\r\n\
+                            content-type: application/octet-stream\r\n\
+                            content-range: bytes {start}-{end}/*\r\n\
+                            accept-ranges: bytes\r\n\
+                            etag: \"{etag}\"\r\n"
+                        );
+                        let body = &file[start..end];
+                        let size = body.len();
+                        format!("{header}\r\n{size:x}\r\n{body}\r\n0\r\n\r\n")
+                    };
+
+                    writer.write_all(msg.as_bytes()).await.unwrap();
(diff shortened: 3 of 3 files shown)


crates/common/download/src/download.rs
```diff
@@ -191,26 +186,29 @@ impl Downloader {
...
     ) -> Result<(), DownloadError> {
         loop {
-            let file_pos = file
-                .stream_position()
-                .context("Can't get file cursor position".to_string())?;
-
-            let mut response = self.request_range_from(url, file_pos).await?;
+            let offset = next_request_offset(&prev_response, file);
+            let mut response = self.request_range_from(url, offset).await?;
+            if was_resource_modified(&response, &prev_response) {
```
albinsuresh: There is nothing wrong with this impl. But, instead of this manual check and adjusting the start position, using the `IfRange` header along with the `ETag` seems like the "accepted convention" to resume a partial download. Avoids that additional HTTP call as well (not that it really matters in the context of a file download).

crates/common/download/src/download.rs
```diff
@@ -389,6 +357,46 @@ impl Downloader {
...
+            true
+        }
+        (Some(etag), Some(prev_etag)) => {
+            // Examples:
+            // ETag: "xyzzy"
+            // ETag: W/"xyzzy"
+            // ETag: ""
+            if etag.starts_with("W/") {
+                // validator is weak, but in range requests tags must match using strong comparison
+                // https://www.rfc-editor.org/rfc/rfc9110#section-13.1.5-12.1
```
didier-wenzek: The following link would be more helpful:


```suggestion
                // validator is weak, but in range requests tags must match using strong comparison
                // https://www.rfc-editor.org/rfc/rfc9110#entity.tag.comparison
```

crates/common/download/src/download.rs
```diff
@@ -389,6 +357,46 @@ impl Downloader {
...
+    let prev_etag = prev_response
+        .headers()
+        .get(header::ETAG)
+        .and_then(|h| h.to_str().ok());
+
+    match (etag, prev_etag) {
+        (None, None) => {
+            // no etags in either request, assume resource is unchanged
+            false
+        }
```
didier-wenzek: It would be safer to assume the contrary. But, I also understand this choice: if the source doesn't even try to provide support to detect changes, then why bother?

Bravo555: Yeah, particularly with downloading files from Cumulocity where ETag is not given, I didn't want to disable partial downloads there, as people might be surprised why it stopped working... But in cases where ETag is actually available we shouldn't ignore it.

crates/common/download/src/download.rs
Bravo555: `IfRange` indeed is something that I wanted to add as well, but in a follow-up PR.

crates/common/download/src/download.rs
Bravo555: fixed in 7187df84d

crates/common/download/src/download/tests.rs
```diff
@@ -0,0 +1,473 @@
...
+        None,
+        CloudHttpConfig::new(Arc::from(vec![req_cert]), None),
+    );
+    let err = downloader.download(&url).await.unwrap_err();
+    let err = anyhow::Error::new(err);
+
+    assert!(dbg!(format!("{err:#}")).contains("received fatal alert: CertificateRequired"));
+}
+
+#[test_case::test_case(Some(r#""xyzzy"#), Some(r#""xyzzy"#), false)]
```
albinsuresh:
```suggestion
#[test_case::test_case(Some(r#""xyzzy""#), Some(r#""xyzzy""#), false)]
```
Right? Unless the literal with that single `"` was intentional for some reason. Same for the rest of the test cases.

crates/common/download/Cargo.toml
```diff
@@ -13,6 +13,7 @@ repository = { workspace = true }
 anyhow = { workspace = true, features = ["backtrace"] }
 backoff = { workspace = true }
 certificate = { workspace = true, features = ["reqwest"] }
+http.workspace = true
```
albinsuresh:
```suggestion
http = { workspace = true }
```
For consistency.

crates/common/download/src/download.rs
```diff
@@ -191,26 +186,29 @@ impl Downloader {
...
     }
 
-    /// Retries the download requesting only the remaining file part.
+    /// If interrupted, continues ongoing download.
     ///
-    /// If the server does support it, a range request is used to download only
-    /// the remaining range of the file. If the range request could not be used,
-    /// [`retry`](Downloader::retry) is used instead.
-    async fn download_remaining(
+    /// If the server supports is, a range request is used to download only the
```
albinsuresh:
```suggestion
    /// If the server supports it, a range request is used to download only the
```

crates/common/download/src/download/tests.rs
Bravo555: fixed in 8ce0723

crates/common/download/Cargo.toml
Bravo555: fixed in 8ce0723

crates/common/download/src/download.rs
Bravo555: fixed in 8ce0723

crates/common/download/src/download/partial_response.rs
```diff
@@ -1,8 +1,19 @@
 //! Utilities related to parsing responses to partial requests.
 
-use reqwest::header;
 use reqwest::header::HeaderValue;
 use reqwest::StatusCode;
+use reqwest::{header, Response};
+
+pub(super) enum PartialResponse {
```
albinsuresh, line 8: Optional:
```suggestion
pub(super) enum RangeResponse {
```

crates/common/download/src/download/partial_response.rs
```diff
@@ -11,26 +22,77 @@ use reqwest::StatusCode;
...
+        StatusCode::OK => Ok(PartialResponse::CompleteContent),
 
         // Partial response, the range might be different from what we
         // requested, so we need to parse it. Because we only request a single
         // range from the current position to the end of the document, we can
         // ignore multipart/byteranges media type.
-        StatusCode::PARTIAL_CONTENT => partial_response_start_range(response)?,
+        StatusCode::PARTIAL_CONTENT => {
+            if was_resource_modified(response, prev_response) {
+                return Ok(PartialResponse::ResourceModified);
```
albinsuresh, line 40: At first, I thought this inefficiency (where the already fetched partial content has to be discarded after the Etag check) could be completely avoided with the usage of `If-Range` header. But I just realised that a server supporting `ETag` doesn't guarantee that it supports `If-Range` as well. So, we'll need this to handle that corner case. But yeah, for the servers that support it, this path would be skipped.

crates/common/download/src/download/partial_response.rs
```diff
@@ -1,8 +1,19 @@
 //! Utilities related to parsing responses to partial requests.
 
-use reqwest::header;
 use reqwest::header::HeaderValue;
 use reqwest::StatusCode;
+use reqwest::{header, Response};
```
albinsuresh: To be fixed to make the formatter happy.

crates/common/download/src/download/partial_response.rs
Bravo555: IMO better to keep the name `PartialResponse` as it's related to the _Partial Content_ status code.

crates/common/download/src/download.rs
```diff
@@ -191,26 +186,32 @@ impl Downloader {
...
-            let file_pos = file
-                .stream_position()
-                .context("Can't get file cursor position".to_string())?;
-
-            let mut response = self.request_range_from(url, file_pos).await?;
-            let offset = partial_response::response_range_start(&response)?;
+            let request_offset = next_request_offset(&prev_response, file)?;
+            let mut response = self.request_range_from(url, request_offset).await?;
+            let offset = match partial_response::response_range_start(&response, &prev_response)? {
+                PartialResponse::CompleteContent => 0,
```
didier-wenzek, line 205: I don't fully get how we are sure the download is making progress, as we restart here from zero in a loop and the backoff retry being inner to this loop.

Bravo555: We restart from zero only if the server returns `200 OK` and we have to download the entire file again. `self.request_range_from` performs the HTTP request, and is subject to the backoff retry policy, but reading response body and writing it to file happens in `save_chunks_to_file_at`, which if it completes without errors, breaks out of the loop.

didier-wenzek: > which if it completes without errors, breaks out of the loop.

That *if* is my concern. What if we repeatedly fail to consume from the network the last bytes of a file that is served in its entirety each time? Having an infinite retry loop is never safe, even when the likelihood is low.

Bravo555: Indeed you've convinced me that the loop being unbounded is a problem and it should be bounded somehow, but how? Should we just limit the number of iterations the loop can do, such that if we retry too many times we fail, even if theoretically every retry could make a little progress before e.g. timing out? Or should we try to do something smarter, like only count towards the limit requests that haven't made any progress (although it's not clear to me how to precisely define it)?

Ah, this whole thing turns out to be more complicated that initially anticipated, everywhere there is some little edge case. Maybe there are some dependencies that could help with this, or if not I should take a more thorough look at how projects like wget are doing this.

didier-wenzek: > Should we just limit the number of iterations the loop can do

Simple and effective. I don't think we need something smarter as the point is only to avoid insane traps.

PR #3950, opened 2026-01-29
feat: config_update workflow with granular builtin actions

albinsuresh:
## Proposed changes

Persistent config_update workflow with granular built-in actions to handle the configuration download and update as separate steps. This split makes the workflow more extensible where users can keep the existing builtin download logic but can override the builtin config set logic.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -317,16 +317,16 @@ impl Agent {
         // Software update actor
         let mut software_update_builder = SoftwareManagerBuilder::new(self.config.sw_update_config);
 
-        // Converter actor
-        let mut converter_actor_builder = WorkflowActorBuilder::new(
+        // Workflow actor
+        let mut workflow_actor_builder = WorkflowActorBuilder::new(
             self.config.operation_config,
             &mut mqtt_actor_builder,
             &mut script_runner,
             &mut fs_watch_actor_builder,
             &mut downloader_actor_builder,
         );
-        converter_actor_builder.register_builtin_operation(&mut restart_actor_builder);
-        converter_actor_builder.register_builtin_operation(&mut software_update_builder);
+        workflow_actor_builder.register_builtin_operation(&mut restart_actor_builder);
+        workflow_actor_builder.register_builtin_operation(&mut software_update_builder);
 
         // Shutdown on SIGINT
         let signal_actor_builder = SignalActor::builder(&runtime.get_handle());
@@ -379,12 +379,11 @@ impl Agent {
                     &mut uploader_actor_builder,
                 )
                 .await?;
-                converter_actor_builder.register_builtin_operation(&mut config_manager);
-                converter_actor_builder
-                    .register_builtin_operation_step_handler(&mut config_manager);
-                converter_actor_builder
+                workflow_actor_builder.register_builtin_operation(&mut config_manager);
+                workflow_actor_builder.register_builtin_operation_step_handler(&mut config_manager);
+                workflow_actor_builder
                     .register_sync_signal_sink(OperationType::ConfigSnapshot, &config_manager);
-                converter_actor_builder
+                workflow_actor_builder
                     .register_sync_signal_sink(OperationType::ConfigUpdate, &config_manager);
 
                 Some(config_manager)
@@ -415,8 +414,8 @@ impl Agent {
                 &mut uploader_actor_builder,
             )
             .await?;
-            converter_actor_builder.register_builtin_operation(&mut log_actor);
-            converter_actor_builder.register_sync_signal_sink(OperationType::LogUpload, &log_actor);
+            workflow_actor_builder.register_builtin_operation(&mut log_actor);
+            workflow_actor_builder.register_sync_signal_sink(OperationType::LogUpload, &log_actor);
             Some(log_actor)
         } else {
             None
@@ -500,7 +499,7 @@ impl Agent {
         runtime.spawn(restart_actor_builder).await?;
         runtime.spawn(software_update_builder).await?;
         runtime.spawn(script_runner).await?;
-        runtime.spawn(converter_actor_builder).await?;
+        runtime.spawn(workflow_actor_builder).await?;
         runtime.spawn(health_actor).await?;
 
         runtime.run_to_completion().await?;
--- a/crates/core/tedge_agent/src/operation_workflows/tests.rs
+++ b/crates/core/tedge_agent/src/operation_workflows/tests.rs
@@ -803,16 +803,16 @@ async fn spawn_mqtt_operation_converter(
         tmp_dir: tmp_path.into(),
         capabilities: Capabilities::default(),
     };
-    let mut converter_actor_builder = WorkflowActorBuilder::new(
+    let mut workflow_actor_builder = WorkflowActorBuilder::new(
         config,
         &mut mqtt_builder,
         &mut script_builder,
         &mut inotify_builder,
         &mut downloade_builder,
     );
-    converter_actor_builder.register_builtin_operation(&mut restart_builder);
-    converter_actor_builder.register_builtin_operation(&mut software_builder);
-    converter_actor_builder.register_builtin_operation_step_handler(&mut config_builder);
+    workflow_actor_builder.register_builtin_operation(&mut restart_builder);
+    workflow_actor_builder.register_builtin_operation(&mut software_builder);
+    workflow_actor_builder.register_builtin_operation_step_handler(&mut config_builder);
 
     let config_box = config_builder.0.build().with_timeout(TEST_TIMEOUT_MS);
     let software_box = software_builder.0.build().with_timeout(TEST_TIMEOUT_MS);
@@ -821,12 +821,12 @@ async fn spawn_mqtt_operation_converter(
     let downloader_box = downloade_builder.build().with_timeout(TEST_TIMEOUT_MS);
     let _inotify_box = inotify_builder.build().with_timeout(TEST_TIMEOUT_MS);
 
-    let converter_actor = converter_actor_builder.build();
+    let workflow_actor = workflow_actor_builder.build();
     let tmp_dir_guard = Arc::clone(&tmp_dir);
     let actor_handle = tokio::spawn(async move {
         // Keep tmp_dir alive for the full actor lifetime.
         let _tmp_dir_guard = tmp_dir_guard;
-        converter_actor.run().await
+        workflow_actor.run().await
     });
 
     Ok(TestHandler {


crates/extensions/tedge_config_manager/src/config_plugins.rs
```diff
@@ -0,0 +1,74 @@
...
+        Self { external_plugins }
+    }
+
+    pub fn builder(self) -> ServerActorBuilder<ConfigPluginServer, Sequential> {
+        ServerActorBuilder::new(self, &ServerConfig::default(), Sequential)
+    }
+}
+
+#[async_trait]
+impl Server for ConfigPluginServer {
```
albinsuresh: Having to introduce this server actor, instead of reusing the existing actor is what I don't like. I was forced to do this as an actor (the existing one) couldn't be a regular actor as well as a server actor at the same time.

If we proceed this way, then I'll at least make the config manager actor to use this actor for all plugin invocations.

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -369,6 +381,128 @@ impl WorkflowActor {
...
             }
+            OperationAction::Download(handlers) => {
+                let step = &state.status;
+                info!("Processing {operation} operation {step} step with download builtin action");
+
+                let Some(tedge_url) = state
+                    .payload
+                    .get("url")
+                    .or_else(|| state.payload.get("remoteUrl"))
+                    .or_else(|| state.payload.get("tedgeUrl"))
```
albinsuresh: Here we have many fallbacks from `url` to `remoteUrl` to `tedgeUrl` because different actors use different URLs for download. For example, the config actor uses `remoteUrl` while software actor uses `url`.

A better alternative would be to parametrise this `download` action to accept an `input.url`.

crates/extensions/tedge_config_manager/src/config_plugins.rs
albinsuresh: Resolved in 9190c55 by using `RequestEnevelope` for any request-response interaction with the existing actor.

crates/extensions/tedge_config_manager/src/lib.rs
```diff
@@ -99,6 +103,13 @@ impl ConfigManagerBuilder {
     }
 
     pub async fn init(config: &ConfigManagerConfig) -> Result<(), FileError> {
+        let workflow_file = config.ops_dir.join("config_update.toml");
+        if !workflow_file.exists() {
+            let workflow_definition = include_str!("resources/config_update.toml");
```
albinsuresh, line 111: Need to create a persistent copy of `config_snapshot.toml` as well, once the current approach for breaking up the monolithic operation is finalized.

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -102,6 +102,30 @@ pub enum OperationAction {
...
+    ///
+    /// Deploys a config file using the appropriate config plugin.
+    /// Expects `downloaded_path` and config `type` in the payload.
+    ///
+    /// ```toml
+    /// action = "config_set"
+    /// on_success = "<state>"
+    /// on_error = "<state>"
+    /// ```
+    ConfigSet(ExitHandlers),
```
didier-wenzek: Exposing a `download` action at the workflow level makes sense as this is useful in other context.

However, I wonder about this `config_set` action which sounds really specific to the configuration plugin.

albinsuresh: > However, I wonder about this config_set action which sounds really specific to the configuration plugin.

Yeah, I had a similar feeling when I introduced an operation specific action into an otherwise generic workflow definition. And this set will expand with even more actions like `config_apply`, `config_verify`, `install_software` etc. Then I started questioning if all actions really need to be generic. My view was that since operation actors are also built into the agent, and those commands are well known and part of tedge-api, why not expose their specific builtin actions as well.

An alternative would be to not route the `config_set` action via this workflow actor, but have a pre-defined states like `config_set`, `config_apply` etc with a `builtin` action and expect the config manager actor to handle that state "externally" (not as a direct invocation by the workflow actor), as follows:

```
[config_set]
action = "builtin"
on_success = "config_apply"

[config_apply]
action = "builtin"
on_success = "config_verify"
```

With this design, once the config manager actor is done processing the `config_set` state, who would move the workflow to the next `config_apply` state? Either the config manager actor needs to be aware of the workflow definition to figureout which state to move to next, or we need to introduce a new mechanism for builtin actors "executed externally" to communicate with the workflow actor that they are done, so that it can move to the next state.

didier-wenzek: >  And this set will expand with even more actions like `config_apply`, `config_verify`, `install_software` etc. 

This is precisely my concern. The workflow engine will soon become a monster doing everything and the operation actors empty boxes.

> An alternative would be to not route the `config_set` action via this workflow actor, but have a pre-defined states like `config_set`, `config_apply` etc with a `builtin` action and expect the config manager actor to handle that state "externally" (not as a direct invocation by the workflow actor),

Yes, I would take this route. States and actions being defined and handled by the operation actors, and the workflow engine responsibility being on orchestration.

> With this design, once the config manager actor is done processing the `config_set` state, who would move the workflow to the next `config_apply` state? 

Indeed, the main issue to be solved.

> Either the config manager actor needs to be aware of the workflow definition to figure out which state to move to next

A strong no go! This would introduce a cyclic dependency, the engine and the operation being both driven by the other.

> or we need to introduce a new mechanism for builtin actors "executed externally" to communicate with the workflow actor that they are done, so that it can move to the next state.

This has to be explored. The engine could pass to the operation actor two state names corresponding to the success and error cases.

albinsuresh: > This is precisely my concern. The workflow engine will soon become a monster doing everything and the operation actors empty boxes.

Just to clarify on that remark on "operation actors being empty boxes", that's not the case even with the current impl. The workflow actor only accepts/recognises the operation-specific `action` values like `config_set`, `config_apply` etc. But, it doesn't perform that action. That is delegated to the operation actor itself. Just like calling a `script` and waiting for it to finish.

That being said, it would be nice if actors could declare/register their actions without the operation actor being bloated with all those `action` values, but just routes those actions to the appropriate actor. So, I'll definitely explore that path.

tests/RobotFramework/tests/cumulocity/configuration/config_operation_workflow/config_update.toml
```diff
@@ -0,0 +1,29 @@
+operation = "config_update"
+
+[init]
+action = "proceed"
+on_success = "scheduled"
+
+[scheduled]
+action = "proceed"
```
didier-wenzek: I think we can deprecate the `scheduled` state and this cascade of *proceed* steps.


```suggestion
```

tests/RobotFramework/tests/cumulocity/configuration/config_operation_workflow/config_update.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[download]
+action = "download"
+on_success = "set"
+on_error = "failed"
+
+[set]
+action = "config_set"
+on_success = "successful"
+on_error = "failed"
```
didier-wenzek: With the aim to decouple workflow orchestration (done by the workflow actor) from workflow execution (delegated to operation plugins and actors), I would explore the following where an action name points out two parts: the operation and a specific action / step.

The shape of a workflow definition could be:

```suggestion
[download]
action = "config_update:download"
on_success = "set"
on_error = "failed"

[set]
action = "config_update:config_set"
on_success = "successful"
on_error = "failed"
```

An alternative shape could be:

```suggestion
[download]
operation = "config_update"
action = "download"
on_success = "set"
on_error = "failed"

[set]
operation = "config_update"
action = "config_set"
on_success = "successful"
on_error = "failed"
```

From an implementation perspective:

1. The workflow engine uses the operation name to find the appropriate worker
2. The state of the operation message is set to the specified action (`download` or `config_set`) before being passed to the worker. So the latter don't need to know what are externally used for the states (`download` or `set`).
3. The worker (aka the operation actor) executes the actions expected at that state and return an updated operation state as either `successful` or `failed`.
4. The workflow engine maps the  `successful` or `failed` to those configured in the workflow definition.
5. Those state name substitutions are okay because not visible over MQTT but only along the in-memory channel between the engine and the worker.
6. A worker should also be able to process all the steps of an operation when invoke using a `builtin:config_update` operation.

albinsuresh: I have a slight preference for the first suggestion with explicitly named actions like `config_update:download`, `config_update:set` etc as it follows the existing style of workflow definitions where each state has either an `action` or an `operation` but not both. It also reads better I feel. Combining both would have complicated the parsing logic as well.  So, I'll proceed with that.

reubenmiller: For the download action, I'd go with the `<config_update:download>` style, as previously the "operation" field is reserved for a proper workflow, not a portion of a workflow.

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -369,6 +381,143 @@ impl WorkflowActor {
...
+
+                        let new_state = state.update_with_builtin_action_result(
+                            "download",
+                            Ok(json!({"downloaded_path": downloaded_path})),
+                            handlers,
+                        );
+                        self.publish_command_state(new_state, &mut log_file).await
+                    }
+                    Err(err) => {
+                        let err = format!("Download failed: {}", err);
```
Bravo555: note: Looks like we don't include error sources here, but looks like important context is already included in top-level error.

In particular, when invalid URL is used for download, error message looks like this:

```
FinalAssertionError: Expected operation (id=51946167) to be SUCCESSFUL, but got: FAILED (failureReason: tedge-agent failed downloading a file: HTTP status client error (404 Not Found) for url (http://127.0.0.1:8001/c8y/inventory/binaries/invalid))
```

To be fair, the error message is emitted in the FileCacheActor and not here, but it's interpolated in the same way, so it's ok.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -363,18 +407,73 @@ impl ConfigManagerWorker {
...
+        {
+            Ok(()) => ConfigSetResponse::Success,
+            Err(err) => ConfigSetResponse::Error(err.to_string()),
+        };
+        if let Err(error) = req.reply_to.send(result).await {
+            error!(
+                "Failed to send ConfigSetResponse for command on topic: {} due to: {}",
+                topic, error
+            );
+        }
```
Bravo555: thought: I think if there are errors sending messages between the actors or channels, or other runtime errors, we shouldn't try to be smart about them and swallow them, but return them to the runtime. Unless there's a particular scenario we need to handle, e.g. sender getting out of scope and dropping their end of the channel?

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -235,11 +264,26 @@ impl ConfigManagerWorker {
             OffsetDateTime::now_utc().unix_timestamp()
         ));
 
+        // Extract cmd_id from topic for CommandLog
+        let cmd_id = match self.config.mqtt_schema.entity_channel_of(topic) {
+            Ok((_, Channel::Command { cmd_id, .. })) => cmd_id,
+            _ => {
+                return Err(ConfigManagementError::InvalidTopicError);
+            }
+        };
```
Bravo555: suggestion: use `extract_command_id` function that is already used below

tests/RobotFramework/tests/cumulocity/configuration/config_operation_workflow/config_update.toml
albinsuresh: Resolved by [c79d190](https://github.com/thin-edge/thin-edge.io/pull/3950/commits/c79d1907450de3536a867362cc7d960c4ddbe17a)

crates/core/tedge_api/src/workflow/mod.rs
albinsuresh: Resolved by [c79d190](https://github.com/thin-edge/thin-edge.io/pull/3950/commits/c79d1907450de3536a867362cc7d960c4ddbe17a)

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -369,6 +387,101 @@ impl WorkflowActor {
...
+                    let err = format!(
+                        "No builtin operation step handler registered for {operation} operation {operation_step} step"
+                    );
+                    return self
+                        .fail_command(&action, &err, state, handlers, &mut log_file)
+                        .await;
+                };
+
+                let new_state = state.update_with_builtin_action_result(&action, result, handlers);
+                self.publish_command_state(new_state, &mut log_file).await
```
didier-wenzek, line 487: It's better to avoid early returns, especially in such a long method.


```suggestion
                let new_state = if let Some(handle) = self
                    .builtin_operation_step_executor
                    .get_mut(&(operation_type, operation_step.clone()))
                {
                    let result = match handle.await_response(step_request).await? {
                        OperationStepResponse::Success => Ok(Value::Null),
                        OperationStepResponse::Error(err) => Err(err),
                    }
                    state.update_with_builtin_action_result(&action, result, handlers)
                } else {
                    let err = format!(
                        "No builtin operation step handler registered for {operation} operation {operation_step} step"
                    );
                    log_file.log_error(&error).await;
                    state.update_with_builtin_action_result(action, Err(error.to_string()), handlers)
                };

                self.publish_command_state(new_state, &mut log_file).await
```

crates/core/tedge_agent/src/agent.rs
```diff
@@ -365,22 +366,27 @@ impl Agent {
...
                 let mut config_manager = ConfigManagerBuilder::try_new(
-                    manager_config,
+                    manager_config.clone(),
                     &mut fs_watch_actor_builder,
                     &mut downloader_actor_builder,
                     &mut uploader_actor_builder,
                 )
                 .await?;
                 converter_actor_builder.register_builtin_operation(&mut config_manager);
+                converter_actor_builder
```
didier-wenzek: Not necessarily in this PR, but it would be could to rename this var. Reading the code out-of-context, I was wondering what converter do we have in the agent?


```suggestion
                workflow_actor_builder
```

crates/core/tedge_api/src/workflow/mod.rs
```diff
@@ -25,11 +26,34 @@ use std::fmt::Formatter;
...
+}
+
+/// Response from a builtin operation step.
+///
+/// This enum is identical to `ConfigSetResponse` which it replaces.
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
+pub enum OperationStepResponse {
+    Success,
+    Error(String),
+}
```
didier-wenzek: As the main purpose of this struct is to be consumed by `update_with_builtin_action_result` which expects a `Result<Value, String>`, I would go with:


```suggestion
pub enum OperationStepResponse {
    Success(Value),
    Error(String),
}

impl From<OperationStepResponse> for Result<Value, String> {
   ...
}
```

This will give an opportunity for operation steps to update the operation state.

crates/core/tedge_api/src/workflow/toml_config.rs
```diff
@@ -193,7 +193,23 @@ impl TryFrom<(TomlOperationState, DefaultHandlers)> for OperationAction {
                     ))?;
                     Ok(OperationAction::BuiltIn(exec_handlers, await_handlers))
                 }
-                _ => Err(WorkflowDefinitionError::UnknownAction { action: command }),
+                "download" => {
+                    let handlers = ExitHandlers::try_from(input.handlers)?;
+                    Ok(OperationAction::Download(handlers))
+                }
+                _ => {
+                    if let Some(builtin) = command.strip_prefix("builtin:") {
```
didier-wenzek, line 202: This is a matter of taste, but I would not require the `builtin` prefix here. `"config_update:set" is clear enough with no risk of name conflicts; as no reserved keywords contain a column `:`.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -363,18 +395,116 @@ impl ConfigManagerWorker {
...
+                    (None, _) => Err(ConfigManagementError::MissingKey {
+                        key: "type".to_string(),
+                    }),
+                    (_, None) => Err(ConfigManagementError::MissingKey {
+                        key: "downloaded_path".to_string(),
+                    }),
+                }
+            }
+            _ => Err(ConfigManagementError::InvalidOperationStep {
+                step: "set".to_string(),
```
didier-wenzek: This error message gives little  hint on what's wrong. It's even confusing because `"set` is the only valid action here.

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -363,18 +395,116 @@ impl ConfigManagerWorker {
...
-        plugin.set(config_type, from_path).await?;
+    async fn process_operation_step_request(
+        &mut self,
+        mut req: RequestEnvelope<OperationStepRequest, OperationStepResponse>,
+    ) {
+        let topic = req.request.command_state.topic.clone();
+        let command_step = req.request.command_step;
+        let command = req.request.command_state;
+        let result = match command_step.as_str() {
+            "set" => {
```
didier-wenzek: I'm surprised to see that so much new code is required to *set* the config freshly downloaded. Wasn't this already something implemented by the config manager?

crates/core/tedge_api/src/workflow/toml_config.rs
albinsuresh: Agree. I was just trying to keep it consistent with `operation = "builtin:config_update"`.

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: I kept the error `InvalidOperationStep` generic, taking that target `step` as an argument to accomodate future steps that will be added like `prepare`, `apply`, `verify` etc.

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: What we already had has been extracted into `execute_config_set_request` and reused here (line no. 442) as well as in the legacy flow. The rest of the fluff is all about input validation and error handling.

crates/core/tedge_actors/src/test_helpers.rs
```diff
@@ -255,12 +255,14 @@ where
         }
     }
 
+    #[track_caller]
```
didier-wenzek: `#[track_caller]` is not working yet for `async fn`. See https://github.com/rust-lang/rust/issues/110011

crates/extensions/tedge_config_manager/src/actor.rs
didier-wenzek: The current error message doesn't match your intent.

A user trying to trigger the `action = "config_update:foo"` would be warned with an `Invalid operation step:  set`.

```
    #[error("Invalid operation step: {0}")]
    InvalidOperationStep(String),
```

crates/core/tedge_actors/src/test_helpers.rs
albinsuresh: Thanks. No wonder why it wasn't working. Forgot to remove those.

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: That was fixed in de1e0bde as follows:
```
"set" => self.process_config_set_request(command).await,
            _ => Err(ConfigManagementError::InvalidOperationStep(
                command_step.clone(),
            )),
```

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: Did some more refactoring in https://github.com/thin-edge/thin-edge.io/commit/de1e0bde987ae7eab7f23b56870fc3bb349bf424 to reduce the fluff.

crates/core/tedge_api/src/workflow/toml_config.rs
albinsuresh: @reubenmiller Please advise. Keep the longer version `builtin:config_update:set`, consistent with the `builtin:config_update` operation name, or keep it simple as `config_update:set`?

crates/core/tedge_api/src/workflow/mod.rs
albinsuresh: Redefined `OperationStepResponse` as `Result<(), String>` in de1e0bde987ae7eab7f23b56870fc3bb349bf424 to simplify this.

crates/core/tedge_agent/src/agent.rs
albinsuresh: Will do that right before the merge.

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: Refactored the whole logic in de1e0bde987ae7eab7f23b56870fc3bb349bf424.

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: Resolved by de1e0bde987ae7eab7f23b56870fc3bb349bf424

crates/extensions/tedge_config_manager/src/actor.rs
albinsuresh: That's a good point. But the fact that this `process_operation_step_request` and the existing `process_operation_request` functions are spawned as tokio tasks from the actor for parallel processing, it complicates the error handling as I can't just return the `RuntimeError`. I may have to use error channels or `JoinSet` to handle this properly.

I also see the following line intentionally added to `ConfigManagerActor::run` method to prevent exactly what you're asking for:
```
            if let Err(err: RuntimeError) = result {
                error!("Error processing event: {err:?}");
            }
```
I'm not quite sure why it was done that way.

Since the fix is not trivial and may have impact on previous assumptions, I'd prefer to explore better solutions for this in a standalone PR.

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
```diff
@@ -206,7 +207,7 @@ Manual config_update operation request
     # Don't worry about the command failing, that is expected since the tedgeUrl path does not exist
     Publish and Verify Local Command
     ...    topic=te/device/main///cmd/config_update/local-2222
-    ...    payload={"status":"init","tedgeUrl":"http://${PARENT_IP}:8000/te/v1/files/${PARENT_SN}/config_update/local-2222","remoteUrl":"","serverUrl":"","type":"tedge-configuration-plugin"}
+    ...    payload={"status":"init","remoteUrl":"http://${PARENT_IP}:8000/te/v1/files/${PARENT_SN}/config_update/local-2222","tedgeUrl":"","serverUrl":"","type":"tedge-configuration-plugin"}
```
didier-wenzek: Why this change? Were the tests previously buggy?

crates/extensions/tedge_config_manager/src/resources/config_update.toml
```diff
@@ -0,0 +1,27 @@
...
+action = "proceed"
+on_success = "executing"
+
+[executing]
+action = "proceed"
+on_success = "download"
+
+[download]
+action = "download"
+input.url = "${.payload.remoteUrl}"
```
albinsuresh: Currently this `download` action adds the `downloadedPath` to the state. I can make that also configurable with an `output` directive, if useful.

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
albinsuresh: This was done to fix some inconsistency with the current materialised workflow and the legacy builtin workflow. The `download` action in the new workflow definition expects the `input.url` to be provided and we provide the `remoteUrl` field from the payload to it, as the `init` state message sent by the mapper only contains `remoteUrl` and `serverUrl` fields and **not** the `tedgeUrl` field.

Our legacy workflow was simply copying the `remoteUrl` value into the `tedgeUrl` field before the `executing` state is published to the broker as follows:

```
                if request.tedge_url.is_none() {
                    request.tedge_url = Some(request.remote_url.clone());
                };
                request.executing();
```

 I still don't know why, as we could have just used the `remoteUrl` value directly.

In the new workflow, I just avoided that unnecessary `remoteUrl` to `tedgeUrl` copy step and updated only the local config_updates operation tests that were providing only the `tedgeUrl` instead of the `remoteUrl`. But, if we want to preserve that old behaviour exactly as-is, then I can add a `scheduled` state to the new workflow, just to do this copying.

albinsuresh: May be I should update the legacy code (that would still be used if someone has a workflow with `operation = "builtin:config_update"`) to fallback to `remoteUrl` as well, when `tedgeUrl` is not present.

reubenmiller: Yeah we need to make sure we don't break the contract here, as this is already public api. So can you confirm that the original test works?

reubenmiller: Plus the reason for the tedgeUrl is to have a URL that is reachable for child devices, as these might not have access to the internet, so the idea was to potentially translate the remoteUrl to a tedgeUrl which could reference the http server...this will become important when we want to introduce a caching feature.

crates/extensions/tedge_config_manager/src/resources/config_update.toml
reubenmiller: I wouldn't be against that, as at the moment it is hard to follow the contract between the states as it isn't clear that 'downloadedPath' is being provided by the download step.

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: Resolved by [5cad446](https://github.com/thin-edge/thin-edge.io/pull/3950/commits/5cad4469e57434f27fc382003872fa511efe4c1b)

tests/RobotFramework/tests/cumulocity/configuration/configuration_operation.robot
albinsuresh: > Plus the reason for the tedgeUrl is to have a URL that is reachable for child devices

Okay, I've always been a bit confused those 3 URL variants. Here is my understanding on the purpose of each:

1. `serverUrl`: The url received from the cloud as-is
2. `remoteUrl`: The proxied version of the `serverUrl`
3. `tedgeUrl`: Artefacts stored/managed by the file-transfer service (downloaded from `remoteUrl` and cached on the thin-edge device), or the `remoteUrl` as-is (when there is no caching).

Please confirm if this understanding is correct.

Based on this assumption, I'll update the `config_update` workflow by passing the `tedgeUrl` as the `input.url` for the `download` action and do the copying from `remoteUrl` to `tedgeUrl` in the `scheduled` state.

albinsuresh: Resolved in a5caeb3 by implementing a fallback mechanism from `input.url` to `tedgeUrl` to `remoteUrl` in the `download` action.

crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -369,6 +396,109 @@ impl WorkflowActor {
...
+                    .get("url")
+                    .and_then(|v| v.as_str())
+                    .filter(|v| !v.is_empty())
+                {
+                    (url, "input.url")
+                } else if let Some(url) = state
+                    .payload
+                    .get("tedgeUrl")
+                    .and_then(|v| v.as_str())
+                    .filter(|v| !v.is_empty())
```
didier-wenzek: I would consider to add a helper method.


```suggestion
                } else if let Some(url) = state.get_text_property("tedgeUrl")
```

crates/extensions/tedge_config_manager/src/resources/config_update.toml
albinsuresh: I made an attempt using `output` excerpt but it didn't look intuitive at all:

```
action = "download"
output.downloadedPath = "${.payload. downloadedPath}"
```

A user looking at this would be wondering where the `downloadedPath` in the `payload` come from. It feels like `output` excerpt is the wrong choice here. I even feel that an inline documentation as follows would be better than this artificial `output` parameter.

```
action = "download"    # downloads from `input.url` into `downloadedPath` 
input.url = "${.payload.remoteUrl}"
```

Or introduce a new `into` directive to specify the target key for that action:

```
action = "download"
into = "downloadedPath"
```

crates/core/tedge_api/src/workflow/toml_config.rs
albinsuresh: Since there were no clear favourites, I'll keep the `builtin:` prefix just as a dedicated namespace for the builtin operation actions.

crates/core/tedge_agent/src/operation_workflows/actor.rs
albinsuresh: Resolved by https://github.com/thin-edge/thin-edge.io/pull/3950/changes/986518890cfbf43b23aef2ffd340166c9dcb4848

crates/extensions/tedge_config_manager/src/lib.rs
```diff
@@ -99,6 +106,13 @@ impl ConfigManagerBuilder {
     }
 
     pub async fn init(config: &ConfigManagerConfig) -> Result<(), FileError> {
+        let workflow_file = config.ops_dir.join("config_update.toml");
```
reubenmiller, line 109: Do we have an upgrade strategy if we make any changes to the config_update.toml file within the tedge binary?

For instance, if the initial config_update.toml is created by the tedge-agent on startup, and then we upgrade tedge to a new version which includes some improvements to the default config_update.toml file, then will these changes be applied if the config_update.toml file already exists? (I believe the answer is no).

reubenmiller: Though looking at the code for the device_profile.toml, I think that suffers from the same flaw, so we can handle this problem in a different PR since it is not a change in behaviour, but it is definitely something we can't forget about.

crates/core/tedge_api/src/workflow/toml_config.rs
reubenmiller: Yes I'm ok with the `builtin:` prefix.

crates/extensions/tedge_config_manager/src/lib.rs
albinsuresh: > For instance, if the initial config_update.toml is created by the tedge-agent on startup, and then we upgrade tedge to a new version which includes some improvements to the default config_update.toml file, then will these changes be applied if the config_update.toml file already exists? (I believe the answer is no).

It doesn't, as we persist this default workflow only if a previous copy does not exist.

The tricky thing on upgrade is to figure out if the existing workflow is an **untouched** workflow from the previous version (which can be safely updated), or one that the user has already updated (which must not be updated).

reubenmiller: Though we have a similar problem for some of the built-in flows and I believe Didier has solved this by using a .template reference file and using it to determine if it is a user file change or a tedge upgrade file change.

albinsuresh: Agree. We can replicate that approach here as well. We just need to make sure that the comparison between the template and the actual file is done before the template file itself is updated with the newer version.

didier-wenzek: > Though we have a similar problem for some of the built-in flows and I believe Didier has solved this by using a .template reference file and using it to determine if it is a user file change or a tedge upgrade file change.

Yes, this is the approach used for tedge flows and builtin bridge configuration files.

crates/core/tedge_agent/src/agent.rs
albinsuresh: Resolved by [829dfcf](https://github.com/thin-edge/thin-edge.io/pull/3950/commits/829dfcfaf7b4fd33345b277ce38f48a021101a41).

PR #3953, opened 2026-01-30
feat: flow minor improvements

reubenmiller:
## Proposed changes

* support loading flows from nested folders on startup
* resolve script path relative to the flow definition file
* support meta information in flow definitions and display them in the `flows list` command

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/tedge_flows/sensor-flows/sensor-to-te.js
+++ b/tests/RobotFramework/tests/tedge_flows/sensor-flows/sensor-to-te.js
@@ -1,0 +1,15 @@
+export function onMessage(message) {
+    let topic = message.topic.split('/')
+    let data = message.payload
+
+    if (topic.length < 2) {
+        throw new Error("Not a sensor topic");
+    }
+
+    let measurement = topic[1]
+
+    return [{
+        topic: "te/sensor",
+        payload: `{"${measurement}": ${data}}`
+    }]
+}
--- a/tests/RobotFramework/tests/tedge_flows/sensor-flows/sensor.toml
+++ b/tests/RobotFramework/tests/tedge_flows/sensor-flows/sensor.toml
@@ -1,0 +1,8 @@
+input.mqtt.topics = ["sensor/+"]
+
+steps = [
+    { script = "sensor-to-te.js" },
+]
+
+output.mqtt.topic = "te/sensor"
+errors.mqtt.topic = "te/errors"
--- a/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
+++ b/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
@@ -223,6 +223,18 @@ Publishing transformation errors
     Should Be Equal As Integers    ${message["b"]["c"]}    6789
     [Teardown]    Uninstall Flow    publish-js-errors.toml
 
+Nested flow directory
+    Install Nested Flow    sensor-flows
+    ${start}    Get Unix Timestamp
+    Execute Command    tedge mqtt pub sensor/temperature 42
+    ${messages}    Should Have MQTT Messages
+    ...    topic=te/sensor
+    ...    minimum=1
+    ...    date_from=${start}
+    ${message}    JSONLibrary.Convert String To Json    ${messages[0]}
+    Should Be Equal As Integers    ${message["temperature"]}    42
+    [Teardown]    Uninstall Nested Flow    sensor-flows
+
 Monitor flow definition updates
     ${start}    Get Unix Timestamp
     Execute Command    touch /etc/tedge/mappers/flows/flows/collectd.toml
@@ -261,6 +273,20 @@ Install Flow
 Uninstall Flow
     [Arguments]    ${definition_file}
     Execute Command    cmd=rm -f /etc/tedge/mappers/flows/flows/${definition_file}
+
+Install Nested Flow
+    [Arguments]    ${directory}
+    ${start}    Get Unix Timestamp
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/${directory}/*    /etc/tedge/mappers/flows/flows/${directory}/
+    Execute Command    ls -lh /etc/tedge/mappers/flows/flows/${directory}
+    Should Have MQTT Messages
+    ...    topic=te/device/main/service/tedge-flows/status/flows
+    ...    date_from=${start}
+    ...    message_contains=${directory}
+
+Uninstall Nested Flow
+    [Arguments]    ${directory}
+    Execute Command    cmd=rm -fr /etc/tedge/mappers/flows/flows/${directory}
 
 Configure flows
     # Required by tail-named-pipe.toml


crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -22,6 +22,18 @@ use tokio::time::Instant;
...
+    pub name: Option<String>,
+
+    /// Version
+    pub version: Option<String>,
+
+    /// Description of the purpose
+    pub description: Option<String>,
+
+    /// User-defined tags
+    pub tags: Option<Vec<String>>,
```
didier-wenzek: As long as `description` and `tags` are not used by the engine, these properties can be kept in the `FlowConfig` definition of a flow. Only the latter has to match the `.toml` file structure.

The `name` can be computed when the `FlowConfig` is loaded into a `Flow`. This name can be a combination of the user provided name and version - using the source file name as a default.

 
```suggestion
    pub name: String,
```

reubenmiller: Actually I think we do need it as these are used in the `tedge flows list` command to display the meta information about the flow (and we could also reference some of the information in log entries when a new flow is loaded, for instance to at least include the name and version along with the file location).

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -112,27 +118,37 @@ pub enum ConfigError {
...
-            };
-            if let Ok(file_type) = entry.file_type().await {
-                if file_type.is_file() {
-                    if let Some("toml") = path.extension() {
-                        info!(target: "flows", "Loading flow: {path}");
-                        if let Some(flow) = FlowConfig::load_single_flow(&path).await {
-                            flows.insert(path.clone(), flow);
+        let config_dir_clone = config_dir.to_path_buf();
+        let paths = tokio::task::spawn_blocking(move || {
+            let pattern = format!("{}/**/*.toml", config_dir_clone);
```
didier-wenzek:
```suggestion
        let pattern = format!("{}/**/*.toml", config_dir);
        let paths = tokio::task::spawn_blocking(move || {
```

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -112,27 +118,37 @@ pub enum ConfigError {
...
+            let mut paths = Vec::new();
+            match glob(&pattern) {
+                Ok(entries) => {
+                    for result in entries {
+                        let Ok(entry) = result else { continue; };
```
didier-wenzek: Yeah, it works. But not really clear.

Actually, I don't think that using `glob` helps a lot. Notably this leads to use `spawn_blocking` and the former code was just a bit away from loading nested directories (recursing when the entry `is_dir()`).

=> I will propose a fixup commit

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -23,6 +23,18 @@ use tokio::time::Instant;
 /// processes those along a chain of transformation steps,
 /// and finally produces the derived messages to a sink of type [FlowOutput].
 pub struct Flow {
+    // Name
+    pub name: Option<String>,
```
didier-wenzek: I would be better to have a non-optional name, using the source file name if no name is explicitly given by the `FlowConfig`.


```suggestion
    pub name: String,
```

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3953/changes/7f793f9fc20f448e64221aae1f60b1aaf8a9689a

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3953/changes/7f793f9fc20f448e64221aae1f60b1aaf8a9689a

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/3953/changes/b8bb4b1c2652eab4419f62d7e0fb2ab23a18ea02

crates/core/tedge/src/cli/flows/cli.rs
```diff
@@ -182,13 +182,40 @@ impl TEdgeFlowsCli {
...
+            path.clone()
+        } else if tokio::fs::try_exists(path).await.unwrap_or(false) {
+            // Exists relative to current working directory
+            path.canonicalize_utf8().unwrap_or(path.clone())
         } else {
-            processor.load_single_script(path).await;
+            // Try to find the file under flows_dir using glob
+            Self::find_in_flows_dir(flows_dir, path)
+                .await
+                .unwrap_or(path.clone())
```
didier-wenzek: What's the motivation here? It would be good to warn the user if more than one file is found.

didier-wenzek: Raise an error if the given name is ambiguous: https://github.com/thin-edge/thin-edge.io/pull/3953/changes/29efeb98b4d3fb2c2a79a449c1a64be5b53df515

crates/core/tedge/src/cli/flows/list.rs
```diff
@@ -37,10 +37,33 @@ impl Command for ListCommand {
...
-            println!("\t{}", step.source());
+        let name = flow.name();
+        let version = flow
+            .version
+            .as_ref()
+            .map(|v| format!("v{v}"))
+            .unwrap_or_else(|| "unversioned".to_string());
+        let location = flow.source.to_string();
+
+        println!("Flow        : {name} ({version})");
```
albinsuresh, line 48: We could use `tabwriter` crate for this tab-aligned output. Its footprint is almost negligible.

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -112,27 +118,35 @@ pub enum ConfigError {
...
-                error!(target: "flows", "Skipping non UTF8 path: {}", entry.path().display());
-                continue;
-            };
-            if let Ok(file_type) = entry.file_type().await {
-                if file_type.is_file() {
-                    if let Some("toml") = path.extension() {
-                        info!(target: "flows", "Loading flow: {path}");
-                        if let Some(flow) = FlowConfig::load_single_flow(&path).await {
-                            flows.insert(path.clone(), flow);
+        let pattern = format!("{}/**/*.toml", config_dir);
```
albinsuresh, line 121: Do we really need the complication of deeply nested directories? Wouldn't it be simpler to limit it to a single level?

albinsuresh (commented):
For system test coverage, how about updating some existing tests in `tedge_flows.robot` to use this nesting feature?

crates/core/tedge/src/cli/flows/list.rs
didier-wenzek: I don't think it's a good idea to add a dependency on a crate for half a dozen of words to align.

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: We decided to go with deeply nested flow directories

PR #3955, opened 2026-02-02
fix: Limit retry attempts on download partial responses

Bravo555:
## Proposed changes

Add a maximum retry limit when retrying after failing due to network error when downloading response chunks.

Notably, this is a separate limit from the backoff policy used when initially connecting to the server, and which doesn't have a single number of attempts, but rather a total time limit.

This new limit is intended to prevent infinite retry loops in cases where the server is consistently failing to deliver the full response, for example due to a misconfiguration or a bug.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/download/src/download/tests/partial_response.rs
+++ b/crates/common/download/src/download/tests/partial_response.rs
@@ -1,5 +1,7 @@
 use super::*;
 
+use std::future::Future;
+use std::ops::Range;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 use tokio::io::AsyncBufReadExt;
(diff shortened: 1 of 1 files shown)


crates/common/download/src/download/tests/partial_response.rs
```diff
@@ -330,3 +338,112 @@ async fn resumed_download_doesnt_leave_leftovers() {
...
+                    let len = file.len();
+                    format!("{header}\r\n{len:x}\r\n{body}\r\n0\r\n\r\n")
+                };
+
+                writer.write_all(msg.as_bytes()).await.unwrap();
+                writer.flush().await.unwrap();
+            };
+            tokio::spawn(response_task);
+            rc.fetch_add(1, Ordering::SeqCst);
+        }
```
didier-wenzek: To make the tests clearer it would be good to extract a test helper here, passing as a parameter the number of retries before a successful response.

crates/common/download/src/download.rs
```diff
@@ -198,7 +198,8 @@ impl Downloader {
         file: &mut File,
         mut prev_response: Response,
     ) -> Result<(), DownloadError> {
-        loop {
+        let mut last_result = Ok(());
+        for _ in 0..4 {
```
didier-wenzek, line 202: Yeah, simple and effective.

crates/common/download/src/download/tests/partial_response.rs
Bravo555: Can you write a short snippet of how you'd use that helper you have in mind? I'm preparing a helper that extracts all the server setup boilerplate for all the tests, but it sounds like what you have in mind is a bit less generic and simpler.

didier-wenzek:
```
let server_task = tokio::spawn(async move {
   serve_file(listener, file, rc, num_of_errors_before_successful).await
}
```

PR #3967, opened 2026-02-05
fix: Make log level overrides for `tedge` binary work again

Bravo555:
## Proposed changes

Fix an issue where setting log level using some methods (e.g. `RUST_LOG`, `--debug`) didn't have an effect for `tedge` binary because a variable containing log level selected by CLI flags was being set if empty as a way to override the default log level, but caused other user-controlled log level overrides to stop working.

Instead, another parameter was introduced for caller to select a default log level that will be selected if no overrides are used.

### How to check

Before the PR, using a flag like `--debug` had no effect, no `DEBUG` logs are printed:

```
$ tedge reconnect c8y --debug
Disconnecting from Cumulocity:
Removing bridge config file... ✓

[...]

Checking Cumulocity is connected to intended tenant... ✓
Enabling tedge-agent... ✓
```

With the fix, `DEBUG` logs are printed:

```
$ cargo run --bin tedge -- reconnect c8y --debug
2026-02-05T17:40:22.868861455Z DEBUG tedge_config::tedge_toml::tedge_config_location: Loading configuration from "/etc/tedge/tedge.toml"

[...]

2026-02-05T17:40:25.734396371Z DEBUG rumqttc::state: Disconnect
Checking Cumulocity is connected to intended tenant... ✓
Enabling tedge-agent... ✓
```

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/system_toml/log_level.rs
+++ b/crates/common/tedge_config/src/system_toml/log_level.rs
@@ -7,6 +7,8 @@ use std::str::FromStr;
 use std::sync::Arc;
 use tracing_subscriber::util::SubscriberInitExt;
 
+const DEFAULT_LEVEL: tracing::Level = tracing::Level::INFO;
+
 #[macro_export]
 /// The basic subscriber
 macro_rules! subscriber_builder {
@@ -20,9 +22,11 @@ macro_rules! subscriber_builder {
 
 /// Configures and enables logging taking into account flags, env variables and file config.
 ///
-/// 1. Log config is taken from the file configuration first
-/// 2. If `RUST_LOG` variable is set, it overrides file-based configuration
-/// 3. If `--debug` or `--log-level` flags are set, they override previous steps
+/// Priority order (highest to lowest):
+/// 1. `--debug` or `--log-level` flags
+/// 2. `RUST_LOG` environment variable
+/// 3. File configuration (`system.toml`)
+/// 4. Default level (INFO)
 ///
 /// Reports all the log events sent either with the `log` crate or the `tracing`
 /// crate.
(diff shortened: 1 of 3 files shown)


crates/common/tedge_config/src/system_toml/log_level.rs
```diff
@@ -5,6 +5,9 @@ use camino::Utf8Path;
 use std::io::IsTerminal;
 use std::str::FromStr;
 use std::sync::Arc;
+use tracing_subscriber::util::SubscriberInitExt;
+
+const DEFAULT_MAX_LEVEL: tracing::Level = tracing::Level::INFO;
```
jarhodes314: Minor: I think the "max" part of the name is not helpful, since we don't use it in any of the other variable names. It makes me think there is something special about this, when really it is _just_ a default value.

```suggestion
const DEFAULT_LEVEL: tracing::Level = tracing::Level::INFO;
```

Edit: I've also realised that you just moved this line, you didn't create it

crates/common/tedge_config/src/system_toml/log_level.rs
jarhodes314: Minor: Again this isn't a new change in this PR, but the wording to me sounds like it's:

a) Reading the config file even if we have a flag set (which I believe isn't the case)
b) Reads as if it's highest to lowest priority when in actual fact it's the reverse

I think highest to lowest priority is what I'd expect this to say:

```suggestion
/// Priority order (highest to lowest):
/// 1. `--debug` or `--log-level` flags
/// 2. `RUST_LOG` environment variable
/// 3. File configuration (`system.toml`)
/// 4. Default level (INFO)
```

Bravo555: Indeed it's simpler just to state priority highest to lowest.

fixed in 63270e46

crates/common/tedge_config/src/system_toml/log_level.rs
```diff
@@ -19,9 +22,11 @@ macro_rules! subscriber_builder {
 
 /// Configures and enables logging taking into account flags, env variables and file config.
 ///
-/// 1. Log config is taken from the file configuration first
-/// 2. If `RUST_LOG` variable is set, it overrides file-based configuration
-/// 3. If `--debug` or `--log-level` flags are set, they override previous steps
+/// Priority order (highest to lowest):
+/// 1. `--debug` or `--log-level` flags
+/// 2. `RUST_LOG` environment variable
```
albinsuresh, line 27: You haven't changed the existing behaviour, but I'm wondering if this priority order should have been inverted for the top two. IMO `RUST_LOG` should have been at the top, overriding even the cli args `--debug` and `--log-level`, as this would be required to override the log levels in containerised environments where the cli arguments in the image definition would have been hardcoded. @reubenmiller thoughts?

Bravo555: AFAIK when we were previously discussing this with Reuben, we agreed that flags should override the `RUST_LOG` environment variable, although I don't remember the precise reason.

PR #3972, opened 2026-02-09
deps: bump cryptoki crate

Bravo555:
## Proposed changes

cryptoki 0.11 fixes UB when using `CKA_ALLOWED_MECHANISMS` with softhsm and also removes unmaintained paste crate in 0.12.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -108,7 +108,7 @@ clap = { version = "4.5", features = [
     "unstable-styles",
 ] }
 clap_complete = { version = "4.5", features = ["unstable-dynamic"] }
-cryptoki = "0.10.0"
+cryptoki = "0.12.0"
 csv = "1.1"
 darling = "0.21"
 doku = "0.21"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -241,7 +241,7 @@ zeroize = "1.5"
 # dlopen is directly under libc, so we need to modify the extern block to not link libdl
 # TODO: remove once fix is upstreamed to libloading
 [patch.crates-io]
-libloading = { git = "https://github.com/Bravo555/rust_libloading.git", branch = "0.8.8" }
+libloading = { git = "https://github.com/Bravo555/rust_libloading.git", branch = "0.8.9" }
```
didier-wenzek, line 244: Okay. This is to remove `dl`  linker flag under `musl`: https://github.com/nagisa/rust_libloading/compare/master...Bravo555:rust_libloading:0.8.9

What's the plan to get rid of this patch?

Bravo555: Once I manage to reproduce the error message that the change is fixing, so I can include it in the PR, I will create a PR to the libloading repository.

I was neglecting to create the PR because I don't really have a good idea of how our build process for musl works such that this error appears, so I can't reproduce the error without actually using our CI pipelline. For this reason, I'm not able to produce a minimal example of this problem showing that this change "fixes" it, or even really describe the problem in detail at all. All i know is that it works for us, but I'm not completely sure that this change is "correct" for the upstream repository, so I was worrying that I won't be able to provide enough context for the maintainer describing why the change is correct.

Nevertheless, I've been putting it off for too long so I will create PR including a best effort description and the error message. If maintainer asks for more details about building for musl, I'll ask @reubenmiller for help.

Bravo555: After a bit more research it turns out that perhaps patching libloading was not necessary and we should make the changes in our build process instead: https://github.com/nagisa/rust_libloading/pull/195#issuecomment-3880257142

Issue #3964, opened 2026-02-05
Debugging tooling for configurable bridge rules


jarhodes314:
The configurable bridge rules (once #3918 is merged) will be quite powerful, but also therefore somewhat possible to screw up. As a result, I think we need some tooling to understand what's going on (and #3918 includes a very simplistic initial attempt at this directly in the `tedge_mqtt_bridge` crate).

My vision for this is a `tedge` subcommand, e.g. `tedge bridge inspect <cloud>`. I think it should have the following functionality:
- [ ] Prints out the active bridge rules from a template (e.g. `tedge bridge inspect c8y`)
- [ ] Shows any disabled rules, explaining why they're disabled
- [ ] Shows any `template_rule`s that are effectively disabled (because the iterator in question is empty)
- [ ] Warn that the entire thing is effectively disabled if the built-in bridge isn't enabled
- [ ] Warns about unknown files in the bridge directory
- [ ] Warns about unrecognised fields
- [ ] Errors on `template_rule`s that don't reference `${item}` (and thus just generate duplicate rules)
- [ ] Warns about duplicate rules (/maybe possible duplicate rules)

The linting aspects of this should be shared with the mapper code so that passing the lints for the templates we supply is part of our tests.

The active rules should probably also be logged by tedge-mapper, so there is probably some logic to be shared there too.

Here is an example of what I came up with during some quick experimentation in #3918:
<img width="707" height="527" alt="image" src="https://github.com/user-attachments/assets/776a4100-0993-4413-82a9-0ec0b5c990ba" />

PR #3975, opened 2026-02-10
feat: `tedge bridge` cli for inspecting and testing configurable bridge rules

jarhodes314:
## Proposed changes
Adds a new `tedge bridge inspect c8y` subcommand to explain what bridge rules exist in the configurable bridge config (or explain where the config is for the cases where the configurable config isn't being used, e.g. mosquitto or built-in bridge + aws/az mappers atm).

It has a `--debug` flag, which will detail why rules are disabled (both via `if` and via `for` over an empty iterator).

Also adds `tedge bridge test c8y <TOPIC>` for finding rules matching a given topic (in either direction).

TODO:
- [x] Deduplicate file loading (mostly in `load_rules_from_directory`) as that shouldn't need to live in the `tedge` crate

## Types of changes

## Paste Link to the issue
- #3964

## Checklist

- [x] l I used `just check` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)

### DIFF
--- a/crates/core/tedge_mapper/src/c8y/mapper.rs
+++ b/crates/core/tedge_mapper/src/c8y/mapper.rs
@@ -86,9 +86,7 @@ impl TEdgeComponent for CumulocityMapper {
             &c8y_config,
             service_topic_id.clone(),
         )?;
-      
         let auth_method = auth_method(&c8y_config);
-      
         if tedge_config.mqtt.bridge.built_in {
             let (tc, cloud_config, reconnect_message_mapper) = mqtt_bridge_config(
                 &tedge_config,


jarhodes314:
> * I suggest to collect the output of `tedge bridge inspect c8y` with `tedge diag collect`

Will do this tomorrow

> * I would also be good to add some words about this tool in the docs for the configurable bridge.

Done 4248f01

didier-wenzek (approved):
I confirm my approval, now that docs and diag tools have been updated.

crates/extensions/tedge_mqtt_bridge/src/persist.rs
```diff
@@ -70,45 +73,104 @@ pub async fn persist_bridge_config_file(
...
+/// This function reads all `.toml` files in the given directory and expands
+/// them using the provided configuration context. Files with other extensions
+/// (like `.toml.template` or `.toml.disabled`) are ignored.
+pub async fn load_bridge_rules_from_directory(
+    dir: &Utf8Path,
+    tedge_config: &TEdgeConfig,
+    auth_method: AuthMethod,
+    cloud_profile: Option<&ProfileName>,
+) -> anyhow::Result<BridgeConfig> {
+    struct RuntimeVisitor(BridgeConfig);
```
albinsuresh, line 156: Interesting pattern.

configuration/contrib/diag-plugins/01_tedge.sh
```diff
@@ -52,6 +52,7 @@ collect() {
     for cloud in $CLOUDS; do
         if tedge config get "${cloud}.url" >/dev/null 2>&1; then
             collect_logs "tedge-mapper-${cloud}"
+            tedge bridge inspect "${cloud}" > "$OUTPUT_DIR/tedge-bridge-inspect-${cloud}.log"
```
albinsuresh, line 55:
```suggestion
            tedge bridge inspect "${cloud}"  --debug > "$OUTPUT_DIR/tedge-bridge-inspect-${cloud}.log"
```
So that we get those quick insights on what was skipped (due to misconfigurations).

jarhodes314: Yes. I also wonder whether we would benefit from collecting the bridge configurations too or whether that's overkill?

jarhodes314:
> Unrelated to the current PR, but do we not support the `if` condition for `config` values? When I tried to apply the following rule:
> 
> ```
> [[rule]]
> if = "${config.device.type} == 'thin-edge.io'"
> topic = "main/device/only/topic"
> direction = "inbound"
> ```
> 
> I get this error:
> 
> ```
> Error: Failed to expand bridge configuration
>      ╭─[ /etc/tedge/mappers/c8y/bridge/mqtt-core.toml:111:29 ]
>      │
>  111 │ if = "${config.device.type} == 'thin-edge.io'"
>      │                             ─┬  
>      │                              ╰── found '==' expected end of input
> ─────╯
> 2026-02-12T15:15:21.240263617Z ERROR tedge::cli::bridge::common: Invalid bridge config template
> Failed to read bridge config files
> ```
> 
> Equality checks are only working for `${connection.auth_method}` key. Intentional?

Yes, this is intentional as we don't have a use-case ourselves for it. However it's very easy to add support if it is a thing we do wish to add support for (I've pretty much implemented it previously to test the extensibility of the parsing logic), where the behaviour would be to compare the value against the stringified output in the format we get from `tedge config get`.

tests/RobotFramework/tests/cumulocity/bridge_config/builtin_bridge.robot
```diff
@@ -1,13 +1,53 @@
...
+
+Bridge test rejects wildcard topics
+    [Documentation]    Verify that tedge bridge test c8y rejects wildcard topics with exit code 1
+    ${output}=    Execute Command
+    ...    tedge bridge test c8y 'c8y/s/us/#'
+    ...    exp_exit_code=1
+    ...    stderr=${True}
+    ...    timeout=0
+    ...    retries=0
+    Should Contain    ${output}[1]    Wildcard characters (#, +) are not supported
```
albinsuresh, line 39: Wondering why you added that restriction, when we have many outbound rules with wildcard topics like: `c8y/s/us/# ->  s/us/#` . Esp when the `rumqttc::topic::matches` function, that is used under-the-hood for these checks, supports wildcards topics as well.

jarhodes314: The motivation for the restriction is that `rumqttc::topic::matches` takes a concrete topic and a filter, and the bridge rule is a filter, so the thing we need to test against it necessarily must be a topic. Otherwise you run into ambiguity (e.g. does `a/+/c` match `a/b/+`?).

I have noticed though that we probably could handle matching wildcard topics better. Currently we just show the concrete topic forwarding (which is important), but we probably ought to show the user the matching wildcard filter as well.

configuration/contrib/diag-plugins/01_tedge.sh
albinsuresh: Yes, I think that'd also be useful, especially to correlate the skipped rules for which the only line numbers are highlighted by the `inspect` command and not the rules themselves. So, we'll need the config files as well for the full picture to deduce which exact rule from that line number was skipped.

tests/RobotFramework/tests/cumulocity/bridge_config/builtin_bridge.robot
jarhodes314: I've decided that a better format for the output (so it can accomodate this information) is:
image

Issue #3976, opened 2026-02-10
tedge-flows input.command is only triggered at its interval after the next statistics dump interval elapses


reubenmiller:
**Describe the bug**

When using a tedge flow, the `input.command.interval` doesn't take effect until the next statistics dump interval occurs even though the flow definition is loaded by the tedge-flows service (as indicated in the logs). Since the statistics is configured to dump every 5 minutes, so it means that the timers are generally delayed up-to 5 mins after every change.

The same behaviour (the delayed execution of the first interval) can also be observed when the `tedge-flows` service is restarted.

I suspect the reason is around the following line, where the next deadline is taken as a minimum between the minimum interval from all flows, and the interval of the statistics dump. Either the reset line on the select statement isn't triggered when flows are loaded/reloaded, or the .min() calculation isn't respecting the `input.command.interval` value (but this is only a guess):

[let deadline_future = sleep_until(deadline).map(|_| Some(InputMessage::Tick(Tick)));](https://github.com/reubenmiller/thin-edge.io/blob/4d6d9dc1f58edaf30bd01d9cdc892e189c61165e/crates/extensions/tedge_flows/src/actor.rs#L115)

**To Reproduce**

The bug (/undesired behaviour) can be reproduced using the following procedure:

1. Stop the tedge-flows service

    ```sh
    systemctl stop tedge-flows
    ```

2. Create the following files under the `/etc/tedge/mappers/flows/flows` folder

    **file: example.toml**

    ```toml
    [input.process]
    command = """sh -c \"date\""""
    interval = "1s"

    [[steps]]
    script = "example.js"
    ```

    **file: example.js**
    ```js
    export function onMessage(message, context) {
        console.log(`Received input: ${message.payload}`);
        return [];
    }
    ```

3. Start the tedge-flows service

    ```sh
    systemctl start tedge-flows
    ```

4. Monitor the log entries

    ```sh
    journalctl -fu tedge-flows -n 20
    ```

    The bug is observed when you don't see the "Received input" log entries until about 5 minutes after the service starts.

5. You can see similar behaviour when you edit the flow changing the interval from `1s` to `2s`, then the first interval is only triggered after up-to 5 minutes later.

**Expected behavior**

Ideally the `input.command.interval` should be respected after the flow is loaded or reloaded.

[...]

PR #3977, opened 2026-02-10
fix: set the poller deadline upon instantiation to ensure it triggers after being loaded for the first time

reubenmiller:
## Proposed changes

Fix a bug where a flow's interval wasn't being triggered as expected and was coupled to the statistics dump interval.

Now the poller is set to trigger on first load (with a small delay of 2s). The delay is intended to prevent flows from trigger commands too frequently when the flow is reloaded too frequently (e.g. someone just touching the file).

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3976

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/input_source.rs
+++ b/crates/extensions/tedge_flows/src/input_source.rs
@@ -19,8 +19,7 @@ pub trait PollingSource: Send + Sync {
     async fn poll(&mut self, timestamp: SystemTime) -> Result<Vec<Message>, PollingSourceError>;
 
     /// Get the next deadline when this source should be polled
-    /// Returns None if the source doesn't have scheduled polling
-    fn next_deadline(&self) -> Option<Instant>;
+    fn next_deadline(&self) -> Instant;
 
     /// Check if this source is ready to be polled at the current time
     fn is_ready(&self, now: Instant) -> bool;
@@ -68,7 +67,7 @@ impl PollingSource for CommandPollingSource {
         Ok(messages)
     }
 
-    fn next_deadline(&self) -> Option<Instant> {
+    fn next_deadline(&self) -> Instant {
         self.poll.next_deadline
     }
 
@@ -133,7 +132,7 @@ impl PollingSource for FilePollingSource {
         Ok(messages)
     }
 
-    fn next_deadline(&self) -> Option<Instant> {
+    fn next_deadline(&self) -> Instant {
         self.poll.next_deadline
     }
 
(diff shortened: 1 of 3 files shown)


crates/extensions/tedge_flows/src/input_source.rs
```diff
@@ -179,7 +179,10 @@ impl PollInterval {
         );
         PollInterval {
             polling_interval,
-            next_deadline: None,
+            // Use a small delay before the first interval to prevent
+            // reduce noise when the poller is instantiated in quick succession
+            // but don't wait for entire interval before the first deadline
+            next_deadline: Some(Instant::now() + Duration::from_secs(2)),
```
albinsuresh: Wouldn't that be better?
```suggestion
            next_deadline: Some(Instant::now() + min(polling_interval, Duration::from_secs(2))),
```

crates/extensions/tedge_flows/src/input_source.rs
```diff
@@ -179,7 +179,10 @@ impl PollInterval {
         );
         PollInterval {
             polling_interval,
-            next_deadline: None,
+            // Use a small delay before the first interval to prevent
+            // reduce noise when the poller is instantiated in quick succession
```
albinsuresh:
```suggestion
            // Use a small delay before the first interval to
            // reduce noise when the poller is instantiated in quick succession
```

crates/extensions/tedge_flows/src/input_source.rs
albinsuresh: If the 2s delay was added to prevent the "thundering-herd" on startup, i'd consider adding a jitter as well. But if it was done just to prevent immediate polling on reload, then this is sufficient.

reubenmiller: Yes i was also considering jitter, but I think we need to spend more time on that to decide if that is the right thing to do as they makes the system more unpredictable (as users might wonder why some services are faster than other to startup, and the startup order is randomized).

And we should't use the minimum polling time as that could be less than 2 seconds, so the idea was to have a fixed "handbrake".

crates/extensions/tedge_flows/src/input_source.rs
reubenmiller: Fixed in fa45dacc3a201d15e84d3acd952587fc863505b7

crates/extensions/tedge_flows/src/input_source.rs
didier-wenzek: ~~This `next_deadline` is related to sources not transformers.~~ (edited, my first reading was too quick)

If we go that route, we can simplify the struct: the `next_deadline` is no more optional.

didier-wenzek (approved):
Approved

Issue #3984, opened 2026-02-16
Allow setting retain flags and QoS levels for outgoing MQTT messages


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

A flow currently capable of setting the MQTT topic and payload, however it would be useful for a flow to also define the following:
* QoS (0, 1, 2)
* Retain flag

**Describe the solution you'd like**

Allow flows to define the following properties of MQTT output messages:
* QoS (0, 1, 2)
* Retain flag

Below shows an example of the related properties used in a simple flow:

```js
export function onMessage(message, context) {
    return [{
        topic: "foo",
        payload: JSON.stringify({
            text: "bar"
        }),
        qos: 1,
        retain: true,
    }];
}
```

**Describe alternatives you've considered**

**Additional context**

PR #3994, opened 2026-02-16
feat: setting retain flags and QoS levels for outgoing MQTT messages

didier-wenzek:
## Proposed changes

* [x] setting retain flags and QoS levels for outgoing MQTT messages
* [x] getting retain flags and QoS levels for incoming MQTT messages
* [ ] ~getting command line for polled and streamed incoming messages~

## Types of changes

## Paste Link to the issue

See https://github.com/thin-edge/thin-edge.io/issues/3984

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/js_script.rs
+++ b/crates/extensions/tedge_flows/src/js_script.rs
@@ -104,6 +104,8 @@ mod tests {
     use serde_json::json;
     use std::time::Duration;
     use tedge_mqtt_ext::MqttMessage;
+    use tedge_mqtt_ext::QoS::*;
+    use tedge_mqtt_ext::Topic;
 
     #[tokio::test]
     async fn identity_script() {
@@ -656,17 +658,56 @@ export function onMessage(message, context) {
     #[tokio::test]
     async fn setting_protocol_specific_properties() {
         let js = r#"
+const utf8 = new TextDecoder();
 export function onMessage(message) {
-    message.mqtt = {
-        "qos": 2,
-        "retain": true,
-    };
+    message.mqtt = JSON.parse(utf8.decode(message.payload))
     return message
 }
         "#;
         let (runtime, mut script) = runtime_with(js).await;
 
-        let input = Message::new("foo/bar", "some message");
+        for (payload, qos, retain) in [
+            (r#"{ "qos": 2, "retain": true }"#, ExactlyOnce, true),
+            (r#"{ "retain": true }"#, AtLeastOnce, true),
+            (r#"{ "qos": 0 }"#, AtMostOnce, false),
+            ("{ }", AtLeastOnce, false),
+        ] {
+            let input = Message::new("", payload);
+            let output = script
+                .on_message(&runtime, SystemTime::now(), &input)
+                .await
+                .unwrap()
+                .pop()
+                .unwrap();
+
+            let mqtt_message = MqttMessage::try_from(output).unwrap();
+            assert_eq!(mqtt_message.qos, qos);
+            assert_eq!(mqtt_message.retain, retain);
+        }
+    }
+
+    #[tokio::test]
+    async fn getting_protocol_specific_properties() {
+        let js = r#"
+export function onMessage(message) {
+    let payload
+    if (message.mqtt) {
+         let mqtt = message.mqtt
+         let qos = mqtt.qos || 1
+         let retain = mqtt.retain || false
+        payload = `message received from MQTT, using QoS ${qos} with retain ${retain}`
+    } else {
+        payload = "message not received from MQTT"
+    }
+    return { "topic": "test", "payload": payload}
+}
+        "#;
+        let (runtime, mut script) = runtime_with(js).await;
+
+        let mqtt_message = MqttMessage::new(&Topic::new_unchecked("foo/bar"), "some message")
+            .with_qos(tedge_mqtt_ext::QoS::ExactlyOnce)
+            .with_retain_flag(true);
+        let input = mqtt_message.into();
         let output = script
             .on_message(&runtime, SystemTime::now(), &input)
             .await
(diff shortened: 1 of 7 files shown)


crates/extensions/tedge_flows/src/js_value.rs
```diff
@@ -119,11 +121,21 @@ impl From<Message> for JsonValue {
...
+                ]);
+                JsonValue::object([("mqtt", mqtt)])
+            }
+        });
         JsonValue::object([
             ("topic", JsonValue::string(value.topic)),
             ("payload", payload),
             ("raw_payload", raw_payload),
             ("time", JsonValue::option(value.timestamp)),
+            ("transport", JsonValue::option(transport)),
```
albinsuresh: Optional:
```suggestion
            ("source", JsonValue::option(source)),
```
As it is more about the source of the message: mqtt or file.

didier-wenzek: > As it is more about the source of the message: mqtt or file.

For an input message, this is the source, yes. But a transformed message, this is the target.

That said I'm not fully happy with the naming. An option is to simply remove this intermediate level that is not easy to name. By the way, [this is what is proposed in the issue](https://github.com/thin-edge/thin-edge.io/issues/3984):

```
    return [{
        topic: "foo",
        payload: JSON.stringify({
            text: "bar"
        }),
        qos: 1,
        retain: true,
    }];
}
```

albinsuresh: I also agree that moving the `qos` and `retain` to an MQTT specific key feels like the right thing to avoid polluting the root namespace, since we support non-mqtt sources/sinks as well. Yes, `topic` being in the `message` is contentious in that sense, but I guess we can't go back on that one now and we have already defined what a topic means for file sources as well.

didier-wenzek: > Yes, `topic` being in the `message` is contentious in that sense

I don't think so. This is a concept shared by all messaging systems. QoS and retained messages is not.

rina23q: I can understand we don't want to add too many things to the root level. Especially, QoS and retained are only related to mqtt input. However, then I would have one nested layer. It would look as below.

```javascript
    return [{
        topic: "foo",
        payload: JSON.stringify({
            text: "bar"
        }),
        mqtt: { 
          qos: 1,
          retain: true,
       }
    }];
}
```

The current approach (two nested layers) is a bit too nested for me.
```javascript
    return [{
        ...message, 
        transport: { mqtt: { qos: 2, retain: true } }
    }];
}
```

albinsuresh: > > Yes, `topic` being in the `message` is contentious in that sense
> 
> I don't think so. This is a concept shared by all messaging systems. QoS and retained messages is not.

I was referring to file sources where the path is conflated into `topic`.

didier-wenzek: > I was referring to file sources where the path is conflated into `topic`.

Yeah. This is one  of the point I wish to improve with this PR.

The path or command of a polling source is not fully conflated into `topic`. The flow definition can provide a topic name for its source. And the path or command is only used as a default.

To fix that, the proposal is to provide both, topic and source info. For a line received from a polling command:

```
{
    "topic": "logical name of the source",               # default to the command
    "payload": "line read from the source",
    "process": {
         "command": "/some/polling/command"
    }
}
```

crates/extensions/tedge_flows/src/js_value.rs
```diff
@@ -195,12 +207,20 @@ impl TryFrom<BTreeMap<String, JsonValue>> for Message {
                 .into())
             }
         };
+        let Ok(transport) = value
+            .get("transport")
+            .filter(|t| *t != &JsonValue::Null)
+            .map(|v| v.to_owned().into_value::<Transport>())
+            .transpose()
+        else {
+            return Err(anyhow::anyhow!("Unexpected transport data").into());
```
Bravo555: question: The error itself doesn't include the context for which message caused the problem, but I'm not sure where it's printed, is the message that caused the problem reported there? Otherwise it may be tricky to debug.

didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/3994/changes/95928aa8f49be321f64c26135776e2e0fe5c703f

crates/extensions/tedge_flows/src/js_value.rs
didier-wenzek: Rina's suggestion has been applied: https://github.com/thin-edge/thin-edge.io/pull/3994/changes/95928aa8f49be321f64c26135776e2e0fe5c703f
```
    return [{
        topic: "foo",
        payload: JSON.stringify({
            text: "bar"
        }),
        mqtt: { 
          qos: 1,
          retain: true,
       }
    }];
}
```

reubenmiller (approved):
Approved. It works nicely.

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -646,6 +649,71 @@ export function onMessage(message, context) {
...
+        "#;
+        let (runtime, mut script) = runtime_with(js).await;
+
+        for (payload, qos, retain) in [
+            (r#"{ "qos": 2, "retain": true }"#, ExactlyOnce, true),
+            (r#"{ "retain": true }"#, AtLeastOnce, true),
+            (r#"{ "qos": 0 }"#, AtMostOnce, false),
+            ("{ }", AtLeastOnce, false),
+        ] {
+            let input = Message::new("", payload);
```
albinsuresh, line 675: Unrelated to this PR, but I'm surprised that we don't reject empty `topic`. For the generic `Message` which is protocol-agnostic, it is probably fine (or is it?) but I expected it to fail at least during the conversion into `MqttMessage`.

albinsuresh (approved):
Though there is sufficient unit test coverage, since this is public API, a system test as well would have been nice (may be for the `retain` flag which is easier to test).

crates/extensions/tedge_flows/src/js_script.rs
didier-wenzek: As long as the message is not published over MQTT, tedge flows set no constraints on the topic and I see no reasons why we should.

PR #4001, opened 2026-02-19
breaking change: message payloads are passed as binary to flow scripts

didier-wenzek:
## Proposed changes

- [x] `Message.payload_raw` has been renamed `payload`
   - The payloads passed to flow scripts are no more strings but bytes arrays
   - To decode a payload, a flow script has to use a `TextDecoder`
- [ ] ~Consider to add a global `utf8` non-standard helper with two `decode()`, `encode()` methods.~

## Types of changes

## Paste Link to the issue

## Checklist


docs/src/references/mappers/flows.md
```diff
@@ -74,18 +74,29 @@ interface FlowStep {
...
 ```
 
+:::note
+The message `payload` is an array of unsigned bytes that has to be explicitly converted to a string when appropriate:
+
+```js
+export function onMessage(message) {
+    const utf8 = new TextDecoder();
+    let string_payload = utf8.decode(message.payload)
+    let json_paylod = JSON.parse(string_payload)
```
rina23q: As the typo check points out:
```suggestion
    let json_payload = JSON.parse(string_payload)
```

docs/src/references/mappers/flows.md
```diff
@@ -74,18 +74,29 @@ interface FlowStep {
...
+  payload: Uint8Array,
   time: Date
 }
 ```
 
+:::note
+The message `payload` is an array of unsigned bytes that has to be explicitly converted to a string when appropriate:
+
+```js
+export function onMessage(message) {
```
albinsuresh, line 93: One wild though for future: Since the decoding to both string and JSON formats now have to be done at the javascript level, I'm wondering if it'd have a significance impact on the performance compared to the deserialisation happening at the rust level.

To get around that issue and to make the life easier for the user, can we expose 3 different variants of this message handler function to the user:

* onMessage(binary)
* onStringMessage(string)
* onJsonMessage(json)

with the restriction that the user can only have one of these defined in his script file? That way, we can do the deserialisation at the rust level and give the user the message in the exact form that he expects the payload in?

Is this even feasible?

docs/src/references/mappers/flows.md
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/4001/commits/e044f70222aebb37bc79953460a1f6f392c25317

docs/src/references/mappers/flows.md
```diff
@@ -74,18 +74,29 @@ interface FlowStep {
...
+    const utf8 = new TextDecoder();
```
```suggestion
    const utf8 = State.decoder;
```
Wouldn't it be better to encourage the user to initialise the `TextDecoder` once in the `State` or the global scope and reuse it to avoid this cost on every message?

didier-wenzek: Yes, good point. The decoder doesn't even need to be part of some state.

```js
const utf8 = new TextDecoder();

export async function onMessage(message) {
   let string_payload = utf8.decode(message.payload)
   // ..
}
```

docs/src/references/mappers/flows.md
didier-wenzek: > Is this even feasible?

Yes, but I would not bother too much. In both case the utf8 decoding stuff is handled by Rust. Even better now: the decoding is done only when required.

docs/src/references/mappers/flows.md
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/4001/changes/bad74e3a983b955876183e8567c57714968463a1

docs/src/references/mappers/flows.md
albinsuresh: Okay I just understood the difference between the global `JSON` object (native version) and `TextDecoder.decode()` (implemented in Rust and exported). For some reason, the difference was not obvious to me purely from the syntax. I'm guessing it is, for an experienced JS developer.

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -184,9 +184,10 @@ mod tests {
     async fn collectd_script() {
         let js = r#"
 export function onMessage(message, context) {
-    const { topic = "topic/not/set" } = context.config;
+    const { topic = "topic/not/set" } = context.config
+    const td = new globalThis.TextDecoder()
```
reubenmiller, line 188: Out of interest, do we need to use `globalThis` here, or can we just use `new TextDecoder()` (like in the other examples).

didier-wenzek: This is not required. Just added in this test as I noticed some JS examples  use it.

Issue #3978, opened 2026-02-11
tedge-flows service continues to use the old script logic


rina23q:
**Describe the bug**
`tedge-flows` service detects the file changes automatically and reloads the scripts.
However, I found a corner case, where tedge-flows still uses an old script path instead of the new one after specific steps.

**To Reproduce**
### Preparation:
 a flow `test.toml`
```toml
input.mqtt.topics = ["test/in"]
steps = [
    { script = "test-old.js" },
]
```
and `test-old.js`
```js
export function onMessage(message, context) {
    return [{
        topic: "test/out",
        payload: `{"old":"I am from test-old.js"}`
    }]
}
```

### Step to reproduce

1. Change the script path to `test-new.js`, which doesn't exist yet.
  ```
  steps = [
      { script = "test-new.js" },
  ]
  ```

2. You can confirm the script doesn't exist by `tedge flow test test/in ''`
  ```sh
  $ tedge flows test test/in ''
  2026-02-11T13:05:43.574928405Z ERROR flows: Failed to compile flow /etc/tedge/mappers/flows/flows/test.toml: Cannot read file /etc/tedge/mappers/flows/flows/test-new.js: No such file or directory (os error 2)
  ```

3. `tedge-flows` service still converts the message based on the old `test-old.js` definition.
```sh
[test/in] 
[test/out] {"old":"I am from test-old.js"}
```

4. Create `test-new.js` with this contents.
```javascript
export function onMessage(message, context) {
    return [{
        topic: "test/out",
        payload: `{"new":"I am from test-new.js"}`
    }]
}
```

5. `tedge flows test` command uses the new script.
```sh
$ tedge flows test test/in ''
[test/out] {"new":"I am from test-new.js"}
```

6. However, `tedge-flows` service still uses the old definition.
```sh
[test/in] 
[test/out] {"old":"I am from test-old.js"}
```

7. To recover this, restart the `tedge-flows` service. After restart,
```sh
[test/in] 
[test/out] {"new":"I am from test-new.js"}
```

**Expected behavior**
`tedge-flows` service uses the new script contents without restarting it.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture `Linux 2c8819180ed5 6.8.0-47-generic #47-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 27 22:03:50 UTC 2024 aarch64 GNU/Linux`
[...]

PR #4002, opened 2026-02-19
fix: order of flow updates should not matter

didier-wenzek:
## Proposed changes

- [x] track paths of flow definition that cannot be loaded
- [x] when a script is added/updated retry to load the broken flow definitions
- [x] publish a flows/status message when a flow is broken   

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3978

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/registry.rs
+++ b/crates/extensions/tedge_flows/src/registry.rs
@@ -367,6 +367,9 @@ impl<F> FlowStore<F> {
     }
 
     pub fn add_unloaded(&mut self, path: Utf8PathBuf) {
+        // When the on-disk version of a flow is broken,
+        // its in-memory version is removed even if still valid.
+        self.flows.remove(&path);
         self.unloaded_flows.insert(path);
     }
 
--- a/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
+++ b/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
@@ -304,16 +304,15 @@ Order of flow definition updates does not matter
     ...    message_contains=issue-3978/test.toml
     Should Contain    ${messages[0]}    "broken"
 
-    # The new version of the flow has not been updated because broken
-    # However, the old version is still running
+    # The new version of the flow has not been loaded because broken
+    # and is no more running
     ${start}    Get Unix Timestamp
     Execute Command    sleep 0.1
     Execute Command    tedge mqtt pub test/in '' --qos 1
-    ${messages}    Should Have MQTT Messages
+    Should Not Have MQTT Messages
     ...    topic=test/out
-    ...    minimum=1
     ...    date_from=${start}
-    Should Be Equal    ${messages[0]}    {"old":"I am from test-old.js"}
+    ...    message_contains={"old":"I am from test-old.js"}
 
     # Push the new script
     ${start}    Get Unix Timestamp


Bravo555 (changes requested):
I think there are 2 problems, first where the fix is doing something it probably shouldn't so the fix should be changed, but 2nd isn't really related to the fix but still something that can be improved and maybe should be considered for a separate PR:

1. Changes in the script are reloaded after it's no longer used by the flow
    To reproduce, create a flow and script:
    
    `test.toml`:
    ```
    input.mqtt.topics = ["dummy"]
    steps = [
        { script = "main.js", interval = "1s" },
    ]
    ```
    
    `main.js`:
    ```
        return [];
    }
    
    export function onInterval(time, context) {
        console.debug("myflow: onInterval called");
        return [];
    }
    ```
    The flow will cause `JavaScript.Console: "myflow: onInterval called"` to be printed when `tedge-flows` is run directly.
    After changing script in `test.toml` from `main.js` to `main-new.js`, as expected, the error is logged but the flow continues logging.
    But, change the text printed in `main.js` to `new: onInterval called` and save the script.
    
    **Observed behaviour**: The new text is logged, change in `main.js` was picked up even though the flow no longer refers to it.

    **Expected behaviour**: After flow stops referring to a script, changes in that script shouldn't change the flow's behaviour.

2. `tedge flows list` returns a single `ERROR` log for a flow that refers to a script that does not exist.
    After doing steps above, call `tedge flows list` (also I created another flow `flow2.toml` to make sure output for other flows is not affected):
    ```
        2026-02-20T17:31:28.660494443Z ERROR flows: Failed to compile flow /etc/tedge/mappers/flows/flows/myflow/flow.toml: Cannot read file /etc/tedge/mappers/flows/flows/myflow/new-main.js: No such file or directory (os error 2)
    Flow        : flow2.toml (unversioned)
    Description : none
    File        : /etc/tedge/mappers/flows/flows/myflow/flow2.toml
    Step 1      : /etc/tedge/mappers/flows/flows/myflow/main.js
    ```

    The other information for the flow is no longer printed, but I think it should still be available. At the very least we could emit an error _in addition_ to printing the other details of the flow, but ideally I think we should display the error as part of information about the log itself, like:

    ```
    Flow        : flow.toml (unversioned)
    Status: <yellow, bold>running, errors in the definition</yellow, bold>
    Description : none
    File        : /etc/tedge/mappers/flows/flows/myflow/flow.toml
    Step 1      : /etc/tedge/mappers/flows/flows/myflow/main.js
    Errors:
      - <red, bold>Failed to compile: Cannot read file /etc/tedge/mappers/flows/flows/myflow/new-main.js: No such file or directory (os error 2)</red, bold>
    ```

    Or even more generally, we could borrow some other stuff from `systemctl status` command, to show important information about the flow like systemd shows information about services:

    ```
    $ systemctl status tedge-agent
    Warning: The unit file, source configuration file or drop-ins of tedge-agent.service changed on disk. Run 'systemctl daemon-reload' to reload units.
    ● tedge-agent.service - tedge-agent is a thin-edge.io component to support operations.
         Loaded: bad-setting (Reason: Unit tedge-agent.service has a bad unit file setting.)
        Drop-In: /etc/systemd/system/tedge-agent.service.d
                 └─override.conf
         Active: active (running) since Thu 2026-02-19 12:29:32 UTC; 1 day 5h ago
       Main PID: 33927 (tedge-agent)
          Tasks: 19 (limit: 5479)
         Memory: 9.4M
            CPU: 46.463s
         CGroup: /system.slice/tedge-agent.service
                 └─33927 /usr/bin/tedge-agent
    ```

So I think issue 1 should be resolved in this PR. For issue 2, I would like your feedback if you agree, and if so I'll create a dedicated issue for it.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -153,13 +154,24 @@ impl FlowsMapper {
...
-        let status = if self.processor.registry.contains_flow(flow) {
-            "updated"
-        } else {
-            "removed"
+        let status = match self.processor.registry.contains_flow(flow) {
+            RegistrationStatus::Unregistered => "removed",
+            RegistrationStatus::Registered => "updated",
+            RegistrationStatus::Broken => "broken",
         };
+        eprintln!("update_flow_status {flow}: {}", status);
```
Bravo555: todo: replace with tracing event

I think we should also clearly log what the flow engine is doing in response to the flow being broken: something like "INFO Unloading flow xxx"/"INFO Stopping flow xxx"/"INFO Starting flow xxx".
Currently we have:

```
2026-02-23T17:45:50.521930478Z  INFO flows: Loading flow /etc/tedge/mappers/flows/flows/test.toml
```

And if the flow definition is broken, also:

```
2026-02-23T17:41:04.098832472Z ERROR flows: Failed to compile flow /etc/tedge/mappers/flows/flows/test.toml: Cannot read file /etc/tedge/mappers/flows/flows/test-new-2.js: No such file or directory (os error 2)
```

But given that in this PR the behaviour meaningfully changed, but the logs are the same as before, the logs are probably lacking, so I'd like to see "Starting"/"Stopping"/etc. logs added to whatever function is invoked when the flow changes state.

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -152,14 +163,20 @@ impl<T: FlowRegistry + Send> FlowRegistryExt for T {
 
     async fn load_all_flows(&mut self, js_runtime: &mut JsRuntime) {
         let config_dir = self.config_dir().to_owned();
-        for (path, config) in FlowConfig::load_all_flows(&config_dir).await.into_iter() {
+        let (loaded_flows, unloaded_flows) = FlowConfig::load_all_flows(&config_dir).await;
```
Bravo555, line 166: thought: `config_dir` here is _usually_ `/etc/tedge/mappers/{mapper}/flows/`, right? If so, I think `config_dir`, and `FlowConfig` for that matter are a bit imprecise names; One could think that `FlowConfig` is a list of parameters/settings to the flow, but it's the flow definition itself.

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -176,30 +193,52 @@ impl<T: FlowRegistry + Send> FlowRegistryExt for T {
...
+            if reloaded {
+                reloaded_flows.insert(flow.as_ref().source.clone());
+            }
         }
+
+        // reload unloaded flows: they might be fixed by the new script
+        let unloaded_flows = self.store_mut().drain_unloaded();
+        for path in unloaded_flows {
+            self.add_flow(js_runtime, &path).await;
+            if self.store().flow(&path).is_some() {
```
albinsuresh, line 236: Rather than these 2 step: `add` first and then `check if added`, it would have been nicer if the `add_flow` itself returned the status of that add.

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -75,11 +76,17 @@ impl FlowRegistry for BaseFlowRegistry {
...
+    Registered,
+    Broken,
+}
+
 #[async_trait]
 pub trait FlowRegistryExt: FlowRegistry {
     fn config_dir(&self) -> Utf8PathBuf;
 
-    fn contains_flow(&self, path: &Utf8Path) -> bool;
+    fn contains_flow(&self, path: &Utf8Path) -> RegistrationStatus;
```
albinsuresh: **Optional (Purely opinionated**):
```suggestion
    fn flow_status(&self, path: &Utf8Path) -> RegistrationStatus;
```
As I find a function named `contains` returning anything other than a `bool` a bit weird.

crates/extensions/tedge_flows/src/actor.rs
didier-wenzek: Oops, this is debugging leftover. => removed

crates/extensions/tedge_flows/src/registry.rs
didier-wenzek: I agree that naming has slipped over time. What used to be named *FlowConfig* (i.e. the configuration of a flow) is now closer to a flow definition. However, I see no rush to change that.

crates/extensions/tedge_flows/src/registry.rs
didier-wenzek: Yeah. In this precise case, yes errors could be directly detected by the `add_flow` method. However, I opted for a design where actions and checks are independent because of the cascade of consequences reloading a file can trigger. For instance, reloading a script might fix or break a flow.

crates/extensions/tedge_flows/src/registry.rs
didier-wenzek: renamed

PR #4004, opened 2026-02-23
fix: prevent topics with null characters or empty

albinsuresh:
## Proposed changes

Reject topics that are empty or contains null characters as forbidden by the MQTT protocol: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718109

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/topics.rs
+++ b/crates/common/mqtt_channel/src/topics.rs
@@ -27,7 +27,7 @@ impl Topic {
     /// Check if the topic name is valid and build a new topic.
     pub fn new(name: &str) -> Result<Topic, MqttError> {
         let name = String::from(name);
-        if rumqttc::valid_topic(&name) {
+        if !name.is_empty() && !name.contains('\0') && rumqttc::valid_topic(&name) {
             Ok(Topic { name })
         } else {
             Err(MqttError::InvalidTopic { name })
@@ -67,7 +67,7 @@ impl TopicFilter {
     pub fn new(pattern: &str) -> Result<TopicFilter, MqttError> {
         let pattern = String::from(pattern);
         let qos = QoS::AtLeastOnce;
-        if rumqttc::valid_filter(&pattern) {
+        if !pattern.is_empty() && !pattern.contains('\0') && rumqttc::valid_filter(&pattern) {
             Ok(TopicFilter {
                 patterns: vec![pattern],
                 qos,
@@ -309,6 +309,8 @@ mod tests {
     fn check_invalid_topic() {
         assert!(Topic::new("/temp/+").is_err());
         assert!(Topic::new("/temp/#").is_err());
+        assert!(Topic::new("").is_err());
+        assert!(Topic::new("bad\0topic").is_err());
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/common/mqtt_channel/src/topics.rs
```diff
@@ -27,7 +27,7 @@ impl Topic {
     /// Check if the topic name is valid and build a new topic.
     pub fn new(name: &str) -> Result<Topic, MqttError> {
         let name = String::from(name);
-        if rumqttc::valid_topic(&name) {
+        if !name.is_empty() && !name.contains('\0') && rumqttc::valid_topic(&name) {
```
Bravo555, line 30: Huh, if it's disallowed by the MQTT protocol, it's weird that this `rumqttc` function doesn't handle it.

Looking at the specification, [this requirement is a normative conformance statement](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106:~:text=MQTT%2D4%2E7%2E3%2D1), while some others aren't, not entirely sure what that means. Maybe it was added by some later revision?

crates/common/mqtt_channel/src/topics.rs
```diff
@@ -67,7 +67,7 @@ impl TopicFilter {
     pub fn new(pattern: &str) -> Result<TopicFilter, MqttError> {
         let pattern = String::from(pattern);
         let qos = QoS::AtLeastOnce;
-        if rumqttc::valid_filter(&pattern) {
+        if !pattern.contains('\0') && rumqttc::valid_filter(&pattern) {
```
Bravo555: And here `rumqttc::valid_filter` rejects empty filters as expected, which we were already testing.

crates/common/mqtt_channel/src/topics.rs
rina23q: Regarding the specs, v3.1 specs says in their appendix:
> A topic must be at least one character long.

So, v3.1.1 promotes the requirement from appendix to one chapter.

Regarding rumqttc, I really think what @Bravo555 says is valid.
> Huh, if it's disallowed by the MQTT protocol, it's weird that this rumqttc function doesn't handle it.

I guess we can create a PR to `rumqttc`. The `valid_topic` implementation checks only if the value can't contain wildcards.
https://github.com/bytebeamio/rumqtt/blob/2167da05a66137cf4933ab7b2aba8b205671f58b/rumqttc/src/mqttbytes/topic.rs#L7-L14

albinsuresh: Yeah it's weird that they haven't fixed in despite [this issue](https://github.com/bytebeamio/rumqtt/issues/595) getting raised back in 2023 and someone even [raising this PR](https://github.com/bytebeamio/rumqtt/pull/813) in 2024.

crates/common/mqtt_channel/src/topics.rs
albinsuresh: Exactly. may be, just for code consistency between `Topic` and `TopicFilter` in our lib, I'll add the empty check here as well, though it's already done in `rumqttc::valid_filter`.

PR #4005, opened 2026-02-23
test: ensure mapper consistently handles arbitrarily ordered messages

jarhodes314:
## Proposed change
This PR adds Rust integration tests to verify that the mapper correctly processes MQTT messages across different topics, regardless of the order in which they are received.

Developing these tests enabled a refactor of the birth message handling, establishing the birth message as the single source of truth for entity existence.

Additionally, the new tests exposed a bug where receiving a service status before a child device message causes a failure. This PR includes a fix for that issue in a6439a04b82ecb8964ed6fe4c995e01b056a5be1 (you can observe the original failure by commenting out the fix and running `just test`).

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
+++ b/crates/extensions/c8y_mapper_ext/src/mea/message_cache.rs
@@ -95,3 +95,119 @@ impl MessageCache {
             .unwrap_or_default()
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::mea::entities::C8yEntityBirth;
+    use crate::mea::entities::C8yEntityStatus;
+    use tedge_flows::FlowContextHandle;
+    use tedge_flows::Transformer;
+
+    #[test]
+    fn main_device_messages_are_never_cached() {
+        let mut cache = make_cache();
+        let msg = make_message("te/device/main///m/temperature", r#"{"temp":42.0}"#);
+        let result = on_message(&mut cache, &msg);
+        assert_eq!(result, vec![msg]);
+    }
+
+    #[test]
+    fn messages_from_unregistered_entity_are_cached() {
+        let mut cache = make_cache();
+        let msg = make_message("te/device/child1///m/temperature", r#"{"temp":42.0}"#);
+        let result = on_message(&mut cache, &msg);
+        assert!(result.is_empty());
+    }
+
+    #[test]
+    fn cached_messages_are_released_on_entity_birth() {
+        let mut cache = make_cache();
+        let measurement = make_message("te/device/child1///m/temperature", r#"{"temp":42.0}"#);
+        assert!(on_message(&mut cache, &measurement).is_empty());
+
+        let birth = make_birth_message("device/child1//");
+        let result = on_message(&mut cache, &birth);
+        assert_eq!(result, vec![measurement]);
+    }
+
+    #[test]
+    fn messages_from_birthed_entity_pass_through() {
+        let mut cache = make_cache();
+
+        let birth = make_birth_message("device/child1//");
+        let _ = on_message(&mut cache, &birth);
+
+        let measurement = make_message("te/device/child1///m/temperature", r#"{"temp":99.0}"#);
+        let result = on_message(&mut cache, &measurement);
+        assert_eq!(result, vec![measurement]);
+    }
+
+    #[test]
+    fn messages_on_unrecognized_topics_pass_through() {
+        let mut cache = make_cache();
+        let msg = make_message("some/unrelated/topic", r#"{"data":"value"}"#);
+        let result = on_message(&mut cache, &msg);
+        assert_eq!(result, vec![msg]);
+    }
+
+    #[test]
+    fn messages_for_different_entities_are_cached_independently() {
+        let mut cache = make_cache();
+        let msg1 = make_message("te/device/child1///m/temperature", r#"{"temp":1.0}"#);
+        let msg2 = make_message("te/device/child2///m/temperature", r#"{"temp":2.0}"#);
+        assert!(on_message(&mut cache, &msg1).is_empty());
+        assert!(on_message(&mut cache, &msg2).is_empty());
+
+        let birth1 = make_birth_message("device/child1//");
+        assert_eq!(on_message(&mut cache, &birth1), vec![msg1]);
+
+        let birth2 = make_birth_message("device/child2//");
+        assert_eq!(on_message(&mut cache, &birth2), vec![msg2]);
+    }
+
+    #[test]
+    fn all_cached_messages_for_an_entity_are_released_on_birth() {
+        let mut cache = make_cache();
+        let msg1 = make_message("te/device/child1///m/temperature", r#"{"temp":1.0}"#);
+        let msg2 = make_message("te/device/child1///m/temperature", r#"{"temp":2.0}"#);
+        let msg3 = make_message("te/device/child1///e/click", r#"{"text":"clicked"}"#);
+        assert!(on_message(&mut cache, &msg1).is_empty());
+        assert!(on_message(&mut cache, &msg2).is_empty());
+        assert!(on_message(&mut cache, &msg3).is_empty());
+
+        let birth = make_birth_message("device/child1//");
+        let result = on_message(&mut cache, &birth);
+        assert_eq!(result, vec![msg1, msg2, msg3]);
+    }
+
+    fn make_cache() -> MessageCache {
+        let mapper_id: EntityTopicId = "device/main/service/tedge-mapper-c8y".parse().unwrap();
+        MessageCache::new(mapper_id)
+    }
+
+    fn make_message(topic: &str, payload: &str) -> Message {
+        Message::new(topic, payload)
+    }
+
+    const BIRTH_TOPIC: &str = "te/device/main/service/tedge-mapper-c8y/status/entities";
+
+    fn make_birth_message(entity_id: &str) -> Message {
+        let birth = C8yEntityBirth {
+            entity: entity_id.parse().unwrap(),
+            status: C8yEntityStatus::Registered,
+            time: 0.0,
+        };
+        make_message(BIRTH_TOPIC, &birth.to_json())
+    }
+
+    fn on_message(cache: &mut MessageCache, message: &Message) -> Vec<Message> {
+        cache
+            .on_message(
+                SystemTime::UNIX_EPOCH,
+                message,
+                &FlowContextHandle::default(),
+            )
+            .unwrap()
+    }
+}


crates/extensions/c8y_mapper_ext/src/entity_cache.rs
```diff
@@ -209,6 +209,9 @@ impl EntityCache {
                     .expect("At least a default parent exists for child entities"),
             )
         {
+            if entity.r#type == EntityType::Service {
+                entity.parent = parent.clone();
+            }
```
didier-wenzek, line 214: I'm surprised to see that something specific needs to be done `Service` but not for `ChildDevice`.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -0,0 +1,654 @@
...
+    }
+
+    /// Queue a message for lazy delivery. It will be sent to the mapper at a
+    /// random point (respecting per-topic order) as the test reads output.
+    pub fn queue(&mut self, message: MqttMessage) {
+        let topic = message.topic.name.clone();
+        let queue = self.pending.entry(topic.clone()).or_default();
+        if queue.is_empty() {
+            self.pending_topics.push(topic);
+        }
```
didier-wenzek: Why do we need to maintain this set of topics for which there are pending messages? Wouldn't the set of keys of  `self.pending` be enough?

didier-wenzek (approved):
Approved.

I really like these new tests checking that registration messages are actually sent to c8ybefore mea messages, despite being sent unordered. The test scripts are linear (sending this, then that and those) (hence easy to read). Coverage is provided by `proptest` combined with a shuffling MQTT message box with an explicit fault model (message reordering across topics, messages being actors with no specific order, derived messages being possibly published before the source message is processed by all actors).

I have a question though. @albinsuresh do you think it's normal that some registration were missing for services but not for child-devices?

crates/extensions/c8y_mapper_ext/src/config.rs
```diff
@@ -77,6 +77,7 @@ pub struct C8yMapperConfig {
     pub tmp_dir: Arc<Utf8Path>,
 
     pub max_mqtt_payload_size: u32,
+    pub alarm_interval: &'static str,
```
albinsuresh:
```suggestion
    pub alarm_sync_interval: &'static str,
```

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
albinsuresh: Yeah, I don't understand why this is done only for `service` either. The same makes sense for child device as well.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -0,0 +1,306 @@
...
+/// just like a real MQTT broker would.
+///
+/// ## Randomized per-actor delivery order
+/// When delivering a message to multiple actors (either an injected message
+/// or a looped-back publish), the actors receive it in a random order with
+/// a `yield_now()` between each delivery. This means an actor that receives
+/// the message first may process it and publish a response *before* another
+/// actor has even received the original message — creating realistic
+/// interleaving without any timing hacks.
+pub struct ShuffledMqttBox<'a, R> {
```
albinsuresh, line 56: That's a very powerful addition to our MQTT level testing, especially with the randomness that's built into it, which I don't even see much with the real brokers. I even feel that this, along with `MockMqttBox` can be promoted into a standalone test-helper that pretty much all the tedge components can use for testing.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -0,0 +1,654 @@
...
+    pub fn queue(&mut self, message: MqttMessage) {
+        let topic = message.topic.name.clone();
+        let queue = self.pending.entry(topic.clone()).or_default();
+        if queue.is_empty() {
+            self.pending_topics.push(topic);
+        }
+        queue.push_back(message);
+    }
+
+    /// Send one randomly-chosen pending message to actors in shuffled order,
```
albinsuresh:
```suggestion
    /// Send one pending message from a randomly-chosen topic to actors in shuffled order,
```

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
albinsuresh: It looks like it was done for easier indexed random access, to choose a random topic.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -0,0 +1,654 @@
...
+    /// yielding between each delivery. Returns Ok(true) if a message was
+    /// sent, Ok(false) if there are no pending messages.
+    async fn inject_one(&mut self) -> Result<bool, ChannelError> {
```
albinsuresh:
```suggestion
    async fn flush_one(&mut self) -> Result<bool, ChannelError> {
```
As we're not injecting a new message into the broker but rather taking one out of it.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -0,0 +1,654 @@
...
+
+        Ok(())
+    }
+}
+
+#[async_trait::async_trait]
+impl<R: Rng + Send> MessageReceiver<MqttMessage> for ShuffledMqttBox<'_, R> {
+    async fn try_recv(&mut self) -> Result<Option<MqttMessage>, RuntimeRequest> {
+        // Inject one pending message before reading output
+        let _ = self.inject_one().await;
```
albinsuresh: Instead of always injecting/flushing exactly one message, wondering if we should introduce some randomness here as well on the number of messages that are flushed.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -0,0 +1,654 @@
...
+
+#[proptest::property_test(config = proptest_config())]
+fn child_alarm_with_shuffled_entity_registration(seed: u64) {
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+    rt.block_on(async move {
+        tokio::time::timeout(
+            Duration::from_secs(2),
```
albinsuresh: That additional timeout just for this alarm test is to make sure that the alarm sync definitely happens (since test `alarm_sync_interval` = 100ms )?

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -0,0 +1,654 @@
...
+/// Set up the mapper actors and return the MQTT box ready for shuffled testing.
+///
+/// This handles the common boilerplate: building actors, spawning them,
+/// sending the bridge health message, and ignoring noisy topics.
+/// The caller creates a `ShuffledMqttBox` from the returned `MockMqttBox`.
+///
+/// Returns `(MockMqttBox, KeepAlive)` — the second value must be kept alive
+/// (bound to a `_variable`) for the duration of the test so that actor
+/// channels and the temp directory are not dropped.
+async fn setup_mapper() -> (MockMqttBox, Box<dyn std::any::Any>) {
```
albinsuresh, line 588: Instead of that `Box`, why not return the `TestHandle` itself? Just include the `ttd` also in that handle and make sure that the handle itself is not dropped from the tests.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
jarhodes314: I think maybe `send_one` is better. `flush_one` sounds a bit self-contradictory to me (I'd tend to think of flushing as getting rid of all messages).

crates/extensions/c8y_mapper_ext/src/entity_cache.rs
jarhodes314: Yes, this is probably highlighting a gap in the testing, which I think assumes `@parent` is always supplied for child devices, but that's only the case for nested child devices.

jarhodes314: The wider context for this code is:

```rust
        if entity.r#type != EntityType::MainDevice
            && !self.entities.contains(
                parent
                    .as_ref()
                    .expect("At least a default parent exists for child entities"),
            )
        {
            if entity.r#type == EntityType::Service {
                entity.parent = parent.clone();
            }
            self.pending_entities
                .cache_early_registration_message(entity);
            return Ok(false);
        }
```

If we're registering a top-level child device, the parent is the main device and therefore always exists, thus we don't enter the outer if block and skip the check. The specific problem this is fixing is that a service is defined for a child device that doesn't yet exist.

For nested child devices, the parent will be explicitly defined already, so only services can have a parent that is inferred but may not exist in the entity store.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
jarhodes314: I think this is a good point. Since we're currently yielding between messages, we are very much assuming each actor receives the same message at broadly the same time, but in practice one actor could easily receive multiple messages before another receives any. I'll try and see how best to make this possible.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
jarhodes314: We move `http` when we call `spawn_dummy_c8y_http_proxy`, so after that we no longer have the `TestHandle`. I think the `Box<dyn std::any::Any>` also has the nice touch of communicating that the tests don't to know anything about what's going on there, they just keep the value alive.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
jarhodes314: Somewhat. I think this is probably no longer needed. I think I might have introduced it when I was trying to get the alarm sync to happen faster to avoid the bottleneck on that. I'll remove it as I don't think it has any value now.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
albinsuresh: Yeah I agree that `flush` implies removing everything. The problem with both `inject_one()` and `send_one()` is that they both kinda sounds like we're sending something to the broker. That's why I was consisering words like `flush` or `drain` that implies it being taken out of the broker. So, here are few more suggestions:

1. `send_one_to_actors`, so that it's clear whom it's sent to
2. `deliver_one()`, for a shorter name

Anyway, since this is just a test API, it probably doesn't deserve a longer debate 😜 . So, you may pick one that you find appropriate.

albinsuresh (approved):
LGTM

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
jarhodes314: > they both kinda sounds like we're sending something to the broker.

In effect, that's exactly what we're doing. We are sending the messages directly to actors, but as far as I care from the perspective of a test in this context, that is what "sending a message to the broker" means.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
jarhodes314: I've had a good go at implementing this and can't come up with motivating example for the added complexity (i.e. a bit of code I can change that will cause a test failure that won't fail with the current approach).

That said, there are definitely some gaps in this current approach. For example, one case I hoped to catch was if the message cache continues to cache messages after the entity is birthed. I couldn't make a test that caught this (i.e. a test that sends some MEA data and an entity registration message and checks that all the measurements are received) despite some significant effort. One of the nice things with the current approach is that it's guaranteed to make progress and execute as quickly as possible. A more thorough approach would probably involve adding some intentional delays to messages, but this would be tricky to achieve in a way that's fast to execute, doesn't cause spurious failures and actually covers more paths.

I have just pushed some unit tests for the message cache that do cover this and other potential bugs (now we have a single source of truth in the implementation, it is nice and simple to test the contract).

PR #4009, opened 2026-02-24
feat: custom cloud mappers

jarhodes314:
## Proposed changes
In addition to the openspec proposal, this PR adds a custom cloud mapper that can be run using `tedge-mapper custom` (the idea being that the user names their custom cloud implementation and then runs this like any other mapper).

The mapper reads its config from a mapper config directory (e.g. `/etc/tedge/mappers/thingsboard` for `tedge-mapper thingsboard`). You can (optionally) connect a built-in bridge by adding a `mapper.toml` file:

```toml
url = "127.0.0.1:8883"
```

You can then add bridge config templates to the `bridge/` subdirectory and flows to the `flows/` subdirectory.

### Still TODO
- [ ] Ability to disable TLS for the bridge connection (i.e. connect to insecure cloud brokers)
- [ ] Handle profiles somewhat gracefully in `tedge mapper`/`tedge bridge` subcommands (currently they are totally unsupported)
- [ ] Basic `tedge connect --test` support
- [ ] Configurable bridge rules for aws/az mappers
- [ ] Improve relative path support - `tedge mapper config get` will currently display the naively generated path including any `..`
- [ ] Consider renaming `${config.}` to `${tedge.}` in bridge templates
- [ ] Disable mapper.toml without bridge directory warning for built-in mappers

### Proposal
A proposal for how to implement a generic mapper using [Openspec](https://github.com/Fission-AI/OpenSpec). In short, openspec is an IMO very useful tool for making design decisions with AI agents that can later serve as a source of truth.

The first commit is a bunch of files generated by openspec to make it actually work. It's probably worth some discussion about whether we think it's a useful tool.

The (at the time of writing, only) other commit is the actual proposal and design docs for implementing a generic mapper that can tie together the flows and configurable bridge. I would suggest reading it in the order `proposal.md`, `design.md` then the two `spec.md`s.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/bridge/test_command.rs
+++ b/crates/core/tedge/src/cli/bridge/test_command.rs
@@ -49,6 +49,7 @@ impl Command for BridgeTestCmd {
 
     #[mutants::skip]
     async fn execute(&self, config: TEdgeConfig) -> Result<(), MaybeFancy<anyhow::Error>> {
+        tedge_mapper::warn_misconfigured_mapper_dirs(&config.root_dir().join("mappers")).await;
         let detail = if self.debug {
             DetailLevel::Debug
         } else {
--- a/crates/core/tedge/src/cli/mapper/cli.rs
+++ b/crates/core/tedge/src/cli/mapper/cli.rs
@@ -6,13 +6,11 @@ use crate::log::MaybeFancy;
 use crate::ConfigError;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
-use tedge_config::cli::format_config_set_cmd;
 use tedge_config::TEdgeConfig;
 use tedge_mapper::custom_mapper_config::load_mapper_config;
 use tedge_mapper::custom_mapper_config::scan_mappers_shallow;
 use tedge_mapper::custom_mapper_resolve::resolve_effective_config;
 use tedge_mapper::custom_mapper_resolve::ConfigGetResult;
-use tedge_mapper::custom_mapper_resolve::ConfigSource;
 use tedge_mqtt_bridge::AuthMethod;
 use yansi::Paint;
 
@@ -139,6 +137,7 @@ impl Command for ListMappersCommand {
     }
 
     async fn execute(&self, config: TEdgeConfig) -> Result<(), MaybeFancy<anyhow::Error>> {
+        tedge_mapper::warn_misconfigured_mapper_dirs(&self.mappers_root).await;
         let mappers = scan_mappers_shallow(&self.mappers_root).await;
         if mappers.is_empty() {
             eprintln!("No mappers found under '{}'", self.mappers_root);
@@ -274,13 +273,6 @@ impl MapperConfigGetCommand {
         };
         writeln!(out, "{}", sourced.value)?;
         writeln!(err, "# {}", sourced.source)?;
-        if matches!(sourced.source, ConfigSource::TedgeConfig) {
-            writeln!(
-                err,
-                "# To change this, use: {}",
-                format_config_set_cmd(&self.mapper_name, &self.toml_key)
-            )?;
-        }
 
         Ok(())
     }
--- a/crates/core/tedge_mapper/src/lib.rs
+++ b/crates/core/tedge_mapper/src/lib.rs
@@ -66,6 +66,8 @@ pub mod c8y;
 mod collectd;
 mod core;
 mod custom;
+/// Re-export mapper directory warnings for use by CLI commands.
+pub use core::mappers_dir::warn_misconfigured_mapper_dirs;
 /// Re-export custom mapper config for use by bridge inspection commands.
 pub use custom::config as custom_mapper_config;
 /// Re-export custom mapper config resolution for use by CLI commands.
@@ -217,7 +219,7 @@ pub async fn run(mapper_opt: MapperOpt, config: TEdgeConfig) -> anyhow::Result<(
     )?;
 
     let mappers_dir = mapper_opt.common.config_dir.join("mappers");
-    core::mappers_dir::warn_unrecognised_mapper_dirs(&mappers_dir).await;
+    core::mappers_dir::warn_misconfigured_mapper_dirs(&mappers_dir).await;
 
     // Run only one instance of a mapper (if enabled)
     let mut _flock = None;
--- a/crates/core/tedge/src/cli/bridge/inspect.rs
+++ b/crates/core/tedge/src/cli/bridge/inspect.rs
@@ -42,6 +42,7 @@ impl Command for BridgeInspectCmd {
 
     #[mutants::skip]
     async fn execute(&self, config: TEdgeConfig) -> Result<(), MaybeFancy<anyhow::Error>> {
+        tedge_mapper::warn_misconfigured_mapper_dirs(&config.root_dir().join("mappers")).await;
         let detail = if self.debug {
             DetailLevel::Debug
         } else {
--- a/crates/core/tedge_mapper/src/core/mappers_dir.rs
+++ b/crates/core/tedge_mapper/src/core/mappers_dir.rs
@@ -1,32 +1,23 @@
-//! Utilities for scanning the mappers directory and warning about unrecognised entries.
+//! Utilities for scanning the mappers directory and warning about misconfigured entries.
 //!
-//! A directory under `/etc/tedge/mappers/` is a recognised mapper if it contains at least
-//! one of: a `mapper.toml` file or a `flows/` subdirectory. Directories with neither are
-//! unrecognised and generate a warning. Flows-only mapper directories (with `flows/` but no
-//! `mapper.toml`) generate a separate warning if their `flows/` directory is empty.
+//! A mapper directory under `/etc/tedge/mappers/` is valid if it is empty (the `flows/`
+//! directory will be created automatically on startup), has a `flows/` subdirectory, has a
+//! `mapper.toml`, or any combination thereof.
 //!
-//! Mappers that have a `bridge/` directory but no `mapper.toml` will always fail to start,
-//! since bridge rules require the connection settings from `mapper.toml`. These generate a
-//! targeted warning pointing the user at the fix.
+//! The only configuration that is always broken is a `bridge/` directory without a
+//! `mapper.toml`, since bridge rules require the connection settings from `mapper.toml`.
+//! This generates a targeted warning pointing the user at the fix.
 
 use camino::Utf8Path;
 use tracing::warn;
 
-/// Scans the mappers directory and emits warnings for:
-/// - directories that have neither `mapper.toml` nor a `flows/` subdirectory (unrecognised)
-/// - directories that have a `bridge/` subdirectory but no `mapper.toml` (will always fail
-///   to start — bridge rules require the connection settings in `mapper.toml`)
-/// - directories that have an empty `flows/` subdirectory and no `mapper.toml` (flows-only
-///   mapper with no scripts defined)
+/// Scans the mappers directory and emits warnings for directories that have a `bridge/`
+/// subdirectory but no `mapper.toml`. These mappers will always fail to start because
+/// bridge rules require the connection settings provided by `mapper.toml`.
 ///
-/// This is called on mapper startup to help users spot typos or stale entries.
-pub async fn warn_unrecognised_mapper_dirs(mappers_dir: &Utf8Path) {
-    for name in collect_unrecognised_mapper_dirs(mappers_dir).await {
-        warn!(
-            "Unrecognised mapper directory '{mappers_dir}/{name}': no 'mapper.toml' or 'flows/' found. \
-             This directory will be ignored.",
-        );
-    }
+/// Empty mapper directories and empty `flows/` directories are valid — the mapper creates
+/// the `flows/` directory automatically on startup.
+pub async fn warn_misconfigured_mapper_dirs(mappers_dir: &Utf8Path) {
     for name in collect_bridge_without_mapper_dirs(mappers_dir).await {
         warn!(
             "Mapper '{name}' has a 'bridge/' directory but no 'mapper.toml': the mapper will \
@@ -34,51 +25,11 @@ pub async fn warn_unrecognised_mapper_dirs(mappers_dir: &Utf8Path) {
              'bridge/' directory.",
         );
     }
-    for name in collect_empty_flows_mapper_dirs(mappers_dir).await {
-        warn!("Mapper '{name}' has a 'flows/' directory but no flow scripts are defined.",);
-    }
-}
-
-/// Returns the names of subdirectories in `mappers_dir` that have neither a `mapper.toml`
-/// nor a `flows/` subdirectory.
-///
-/// A directory with a `flows/` subdirectory is a recognised flows-only mapper even without
-/// `mapper.toml`. Only directories with neither are considered unrecognised.
-///
-/// This is exposed for testing; callers should normally use `warn_unrecognised_mapper_dirs`.
-pub(crate) async fn collect_unrecognised_mapper_dirs(mappers_dir: &Utf8Path) -> Vec<String> {
-    let Ok(mut entries) = tokio::fs::read_dir(mappers_dir).await else {
-        return Vec::new(); // Directory doesn't exist yet — nothing to warn about
-    };
-
-    let mut unrecognised = Vec::new();
-    while let Ok(Some(entry)) = entries.next_entry().await {
-        let Ok(file_type) = entry.file_type().await else {
-            continue;
-        };
-        if !file_type.is_dir() {
-            continue;
-        }
-        let name = entry.file_name().to_string_lossy().into_owned();
-        let path = camino::Utf8PathBuf::from(entry.path().to_string_lossy().into_owned());
-        let has_mapper_toml = tokio::fs::try_exists(path.join("mapper.toml"))
-            .await
-            .unwrap_or(false);
-        let has_flows_dir = tokio::fs::try_exists(path.join("flows"))
-            .await
-            .unwrap_or(false);
-        if !has_mapper_toml && !has_flows_dir {
-            unrecognised.push(name);
-        }
-    }
-    unrecognised
 }
 
 /// Returns the names of mapper directories that have a `bridge/` subdirectory but no
 /// `mapper.toml`. These mappers will always fail to start because bridge rules require
 /// the connection settings provided by `mapper.toml`.
-///
-/// This is exposed for testing; callers should normally use `warn_unrecognised_mapper_dirs`.
 pub(crate) async fn collect_bridge_without_mapper_dirs(mappers_dir: &Utf8Path) -> Vec<String> {
     let Ok(mut entries) = tokio::fs::read_dir(mappers_dir).await else {
         return Vec::new();
@@ -107,231 +58,9 @@ pub(crate) async fn collect_bridge_without_mapper_dirs(mappers_dir: &Utf8Path) -
     broken
 }
 
-/// Returns the names of flows-only mapper directories whose `flows/` subdirectory is empty.
-///
-/// A directory qualifies when it has a `flows/` subdirectory but no `mapper.toml` and the
-/// `flows/` directory contains no entries. This suggests the user set up the structure but
-/// has not yet added any flow scripts.
-///
-/// This is exposed for testing; callers should normally use `warn_unrecognised_mapper_dirs`.
-pub(crate) async fn collect_empty_flows_mapper_dirs(mappers_dir: &Utf8Path) -> Vec<String> {
-    let Ok(mut entries) = tokio::fs::read_dir(mappers_dir).await else {
-        return Vec::new();
-    };
-
-    let mut empty_flows = Vec::new();
-    while let Ok(Some(entry)) = entries.next_entry().await {
-        let Ok(file_type) = entry.file_type().await else {
-            continue;
-        };
-        if !file_type.is_dir() {
-            continue;
-        }
-        let name = entry.file_name().to_string_lossy().into_owned();
-        let path = camino::Utf8PathBuf::from(entry.path().to_string_lossy().into_owned());
-        let has_mapper_toml = tokio::fs::try_exists(path.join("mapper.toml"))
-            .await
-            .unwrap_or(false);
-        if has_mapper_toml {
-            continue;
-        }
-        let Ok(mut flows_entries) = tokio::fs::read_dir(path.join("flows")).await else {
-            continue; // no flows/ dir — handled by collect_unrecognised_mapper_dirs
-        };
-        if flows_entries.next_entry().await.ok().flatten().is_none() {
-            empty_flows.push(name);
-        }
-    }
-    empty_flows
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    mod warn_unrecognised {
-        use super::*;
-        use tedge_test_utils::fs::TempTedgeDir;
-
-        #[tokio::test]
-        async fn does_not_warn_for_dirs_with_mapper_toml() {
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-            for name in ["c8y", "az", "thingsboard", "production"] {
-                let dir = mappers_dir.join(name);
-                tokio::fs::create_dir_all(&dir).await.unwrap();
-                tokio::fs::write(dir.join("mapper.toml"), "").await.unwrap();
-            }
-
-            let unrecognised = collect_unrecognised_mapper_dirs(&mappers_dir).await;
-            assert!(
-                unrecognised.is_empty(),
-                "Dirs with mapper.toml should not be unrecognised, got: {unrecognised:?}"
-            );
-        }
-
-        #[tokio::test]
-        async fn handles_nonexistent_mappers_dir_gracefully() {
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("nonexistent/mappers");
-            assert!(collect_unrecognised_mapper_dirs(&mappers_dir)
-                .await
-                .is_empty());
-        }
-
-        #[tokio::test]
-        async fn dir_without_mapper_toml_or_flows_is_flagged() {
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-            tokio::fs::create_dir_all(mappers_dir.join("thingsboard"))
-                .await
-                .unwrap();
-
-            let unrecognised = collect_unrecognised_mapper_dirs(&mappers_dir).await;
-            assert!(
-                unrecognised.contains(&"thingsboard".to_string()),
-                "'thingsboard' (no mapper.toml, no flows/) should be flagged: {unrecognised:?}"
-            );
-        }
-
-        #[tokio::test]
-        async fn dir_with_flows_subdir_is_not_flagged_as_unrecognised() {
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-            // flows-only mapper: has flows/ but no mapper.toml
-            tokio::fs::create_dir_all(mappers_dir.join("thingsboard/flows"))
-                .await
-                .unwrap();
-            tokio::fs::write(
-                mappers_dir.join("thingsboard/flows/telemetry.toml"),
-                "input.mqtt.topics = [\"te/+/+/+/+/m/+\"]",
-            )
-            .await
-            .unwrap();
-
-            let unrecognised = collect_unrecognised_mapper_dirs(&mappers_dir).await;
-            assert!(
-                !unrecognised.contains(&"thingsboard".to_string()),
-                "'thingsboard' (flows-only) should not be flagged as unrecognised: {unrecognised:?}"
-            );
-        }
-
-        #[tokio::test]
-        async fn dir_with_empty_flows_subdir_is_not_flagged_as_unrecognised() {
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-            // flows/ dir exists but is empty — still recognised (not unrecognised)
-            tokio::fs::create_dir_all(mappers_dir.join("thingsboard/flows"))
-                .await
-                .unwrap();
-
-            let unrecognised = collect_unrecognised_mapper_dirs(&mappers_dir).await;
-            assert!(
-                !unrecognised.contains(&"thingsboard".to_string()),
-                "'thingsboard' (empty flows/) should not be flagged as unrecognised: {unrecognised:?}"
-            );
-        }
-
-        #[tokio::test]
-        async fn mixed_directories_only_flags_those_without_mapper_toml_or_flows() {
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-
-            // c8y: simulates a real built-in mapper dir (has flows/, no mapper.toml)
-            tokio::fs::create_dir_all(mappers_dir.join("c8y/flows"))
-                .await
-                .unwrap();
-
-            // Neither mapper.toml nor flows/ — unrecognised
-            tokio::fs::create_dir_all(mappers_dir.join("thingsboard"))
-                .await
-                .unwrap();
-            tokio::fs::create_dir_all(mappers_dir.join("stale-dir"))
-                .await
-                .unwrap();
-
-            let unrecognised = collect_unrecognised_mapper_dirs(&mappers_dir).await;
-            assert_eq!(
-                unrecognised.len(),
-                2,
-                "Should flag exactly 2 unrecognised: {unrecognised:?}"
-            );
-            assert!(unrecognised.contains(&"thingsboard".to_string()));
-            assert!(unrecognised.contains(&"stale-dir".to_string()));
-        }
-
-        #[tokio::test]
-        async fn dir_without_mapper_toml_or_flows_is_flagged_regardless_of_name() {
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-            // Even a name that looks like a built-in is flagged if it has neither
-            // mapper.toml nor flows/ — built-in mapper dirs always have flows/ in practice.
-            tokio::fs::create_dir_all(mappers_dir.join("c8y-extra"))
-                .await
-                .unwrap();
-
-            let unrecognised = collect_unrecognised_mapper_dirs(&mappers_dir).await;
-            assert!(
-                unrecognised.contains(&"c8y-extra".to_string()),
-                "'c8y-extra' (no mapper.toml, no flows/) should be flagged: {unrecognised:?}"
-            );
-        }
-
-        /// Verifies that `warn_unrecognised_mapper_dirs` actually emits `WARN` log events.
-        #[tokio::test]
-        async fn unrecognised_dirs_emit_warn_log_events() {
-            let warnings = CapturedWarnings::install();
-
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-            tokio::fs::create_dir_all(mappers_dir.join("stale-dir"))
-                .await
-                .unwrap();
-            // c8y with flows/ and a flow script — simulates a real built-in mapper dir
-            tokio::fs::create_dir_all(mappers_dir.join("c8y/flows"))
-                .await
-                .unwrap();
-            tokio::fs::write(
-                mappers_dir.join("c8y/flows/default.toml"),
-                "input.mqtt.topics = [\"te/+/+/+/+/m/+\"]",
-            )
-            .await
-            .unwrap();
-
-            warn_unrecognised_mapper_dirs(&mappers_dir).await;
-
-            let captured = warnings.get();
-            assert!(
-                captured.iter().any(|m| m.contains("stale-dir")),
-                "Expected a warning for 'stale-dir', got: {captured:?}"
-            );
-            assert!(
-                captured.iter().all(|m| !m.contains("c8y")),
-                "Should not warn about 'c8y' (built-in mapper), got: {captured:?}"
-            );
-        }
-
-        /// Verifies that `warn_unrecognised_mapper_dirs` emits the empty-flows warning.
-        #[tokio::test]
-        async fn empty_flows_dir_emits_warn_log_event() {
-            let warnings = CapturedWarnings::install();
-
-            let ttd = TempTedgeDir::new();
-            let mappers_dir = ttd.utf8_path().join("mappers");
-            // flows-only mapper with empty flows/ — should warn about missing scripts
-            tokio::fs::create_dir_all(mappers_dir.join("thingsboard/flows"))
-                .await
-                .unwrap();
-
-            warn_unrecognised_mapper_dirs(&mappers_dir).await;
-
-            let captured = warnings.get();
-            assert!(
-                captured.iter().any(|m| m.contains("thingsboard")),
-                "Expected a warning for 'thingsboard' (empty flows/), got: {captured:?}"
-            );
-        }
-    }
 
     mod bridge_without_mapper_toml {
         use super::*;
@@ -386,6 +115,45 @@ mod tests {
         }
 
         #[tokio::test]
+        async fn empty_mapper_dir_is_not_flagged() {
+            let ttd = TempTedgeDir::new();
+            let mappers_dir = ttd.utf8_path().join("mappers");
+            tokio::fs::create_dir_all(mappers_dir.join("thingsboard"))
+                .await
+                .unwrap();
+
+            let broken = collect_bridge_without_mapper_dirs(&mappers_dir).await;
+            assert!(
+                broken.is_empty(),
+                "Empty mapper dir should not be flagged: {broken:?}"
+            );
+        }
+
+        #[tokio::test]
+        async fn empty_flows_dir_is_not_flagged() {
+            let ttd = TempTedgeDir::new();
+            let mappers_dir = ttd.utf8_path().join("mappers");
+            tokio::fs::create_dir_all(mappers_dir.join("thingsboard/flows"))
+                .await
+                .unwrap();
+
+            let broken = collect_bridge_without_mapper_dirs(&mappers_dir).await;
+            assert!(
+                broken.is_empty(),
+                "Empty flows/ dir should not be flagged: {broken:?}"
+            );
+        }
+
+        #[tokio::test]
+        async fn handles_nonexistent_mappers_dir_gracefully() {
+            let ttd = TempTedgeDir::new();
+            let mappers_dir = ttd.utf8_path().join("nonexistent/mappers");
+            assert!(collect_bridge_without_mapper_dirs(&mappers_dir)
+                .await
+                .is_empty());
+        }
+
+        #[tokio::test]
         async fn bridge_without_mapper_toml_emits_warn_log() {
             let warnings = CapturedWarnings::install();
 
@@ -395,7 +163,7 @@ mod tests {
                 .await
                 .unwrap();
 
-            warn_unrecognised_mapper_dirs(&mappers_dir).await;
+            warn_misconfigured_mapper_dirs(&mappers_dir).await;
 
             let captured = warnings.get();
             assert!(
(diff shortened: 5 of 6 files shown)


openspec/changes/generic-mapper/proposal.md
```diff
@@ -0,0 +1,92 @@
...
+
+These questions affect the user-facing design and need to be resolved before or during the design phase.
+
+### How should custom mappers be invoked?
+
+Two main approaches, with different trade-offs:
+
+**Option A: Dedicated `custom` subcommand with profiles.** `tedge-mapper custom --profile thingsboard`. Simple to implement — the profile system already exists. But it makes custom mappers feel second-class compared to `tedge-mapper c8y`, and it prevents users from using profiles for their intended purpose (e.g. connecting to multiple ThingsBoard instances).
+
+**Option B: Custom mapper name as a first-class subcommand.** `tedge-mapper thingsboard`. The user experience is identical to built-in mappers, and profiles work naturally (`tedge-mapper thingsboard --profile staging`). This is more elegant, but raises further questions (see below).
```
albinsuresh: Do, we have plans to support tedge config templates per cloud type as well? Otherwise I don't see any value in that distinction between "yet another profile of some cloud type" vs "profile of a specific cloud type".

The "cloud type" knowledge is only useful for defining the tedge config structure for a cloud, so that all profiled mappers for the cloud type could use the same config template? Or there is something more?

If we could "templatise/externalise" a mapper config structure as well so that adding a new cloud type also doesn't need Rust code changes in `tedge config`, that'd be ideal.

openspec/changes/generic-mapper/proposal.md
```diff
@@ -0,0 +1,92 @@
...
+
+We favour Option B for the user experience, but it needs solutions for disambiguation, name collisions, and discoverability.
+
+### How do we disambiguate custom and built-in mappers?
+
+With Option B, both custom and built-in mappers live under `/etc/tedge/mappers/` and are invoked the same way. We need a reliable way to tell them apart, both on the filesystem and in CLI parsing:
+
+- **On the filesystem**: If thin-edge later adds a built-in `thingsboard` mapper, there's nothing intrinsic about `/etc/tedge/mappers/thingsboard/` that says whether it's custom or built-in. We need a signal.
```
albinsuresh: If we commit to this path of having a distinction between built-in cloud types and their profiles, why not just use subdirectories for the same? Like
```
/etc/tedge/mappers/
|-- c8y
|   |-- unnamed (the default profile)
|   |-- profile1
|-- az
|   |-- unnamed
|-- aws
|   |-- unnamed
|-- thingsboard
|   |-- unnamed
|   |-- profile1
|   |-- profile2
|-- custom
|   |-- profile1
|   |-- profile2
```

This clearly distinguishes the top level as built-in cloud types and all custom mappers go under `custom` as profiles.

openspec/changes/generic-mapper/proposal.md
```diff
@@ -0,0 +1,92 @@
...
+
+### How should the config namespace work?
+
+Custom mapper config is isolated — only the custom mapper itself uses it, unlike the global `tedge_config` schema which is referenced throughout thin-edge. This suggests custom mapper settings should live in the mapper's own `tedge.toml` (e.g. `/etc/tedge/mappers/_thingsboard/tedge.toml`), separate from the `define_tedge_config!` macro. Users edit this file directly rather than using `tedge config set/get`.
+
+This avoids the complexity of fitting dynamic/unknown mapper names into the compile-time config macro, and is acceptable because the users writing custom mappers are technical enough to edit TOML files. The bridge template system would need a new `${mapper.*}` variable namespace to let bridge rule templates reference the mapper's own config (analogous to how `${config.c8y.url}` references global config today). `tedge config` support could potentially be added later if demand warrants it (though this might prove complex to implement in practice).
```
albinsuresh: IIRC, the primary use-case for `tedge config` over editing these `toml` files by hand was to enable easier iterative updates to the config from automation scripts. So, I wouldn't rule out its utility.

openspec/changes/archive/2026-03-18-generic-mapper/proposal.md
```diff
@@ -0,0 +1,92 @@
...
+
+- **Flows** allow users to write custom mapping rules in JavaScript. There is already proof-of-concept work simulating a ThingsBoard connection using flows, and flows can be run today via `tedge-mapper flows` (soon to be `tedge-mapper local`).
+- **The built-in MQTT bridge** replaces the previous reliance on mosquitto for bridging. It lives inside tedge-mapper, supports proxied cloud connections, and makes thin-edge broker-agnostic. A recent PR added support for configuring bridge rules via a TOML file with basic templating, which the Cumulocity mapper already uses.
+
+The built-in mappers already use a unified directory structure per mapper (e.g. `/etc/tedge/mappers/c8y/` contains `tedge.toml`, `bridge/`, and `flows/` side by side). What's missing is the ability for users to create their own mapper using this same pattern — defining a new mapper directory with connection settings, bridge rules, and flows, and having thin-edge run it as a first-class service. This change makes that possible so users can create custom cloud mappers without writing any Rust code — only thin-edge configuration files and flow scripts.
```
didier-wenzek: What is currently missing is a way to configure a non-builtin connection (URL, ports, certificates, device identity). And even for the builtin mappers (c8y, az, aws), things are opaque (config and code being scattered around  `tedge config` and `tedge connect`).

This is the main point to solve: give the users the ability to *directly* define connection info for a mapper (and possibly a profile) in the same spirit as for bridge rules.

openspec/changes/generic-mapper/proposal.md
jarhodes314: We could do, although we've started to use

```
/etc/tedge/mappers/
|-- c8y (the default profile)
|-- c8y.profile1
|-- az
|-- aws
|-- thingsboard
|-- thingsboard.profile1
|-- thingsboard.profile2
|-- custom.profile1
|-- custom.profile2
```

because otherwise we get the question of what do we call the unnamed directory. 

That said, the general idea of translating the syntactic sugar command into a custom mapper is likely feasible. This would give us mapper config access via `tedge config get custom.url --profile mycloud`. While this does work, I think the syntactic sugar for running the mapper is a bit pointless, since `tedge config` is surely more user-facing.

I think for me the value in the two options is:
1. `tedge-mapper custom --profile mycloud`: supports `tedge config` and thin-edge provided systemd definitions (`tedge-mapper-custom@mycloud`) out of the box. Slightly cumbersome syntax to use the mapper, and no way for the user to have different profiles of their custom cloud
2. `tedge-mapper _mycloud`: more convenient syntax for running the mapper (though I don't think users are doing this directly). Easy to add support for profiles in this custom mapper.

The more I think about it, the more I'm coming to the conclusion that profiles don't actually provide any value. For the built-in clouds, it gives users the option to reuse functionality, but (at least at the moment), there is no sharing of functionality between different profiles of the same mapper, so profiles don't really provide intrinsic value in this case.

openspec/changes/generic-mapper/proposal.md
jarhodes314: Treating the different built-in clouds as different types is meaningful so that `tedge config` can have different schemas for the different clouds (e.g. smartrest specific config on c8y), but for the custom clouds I'm envisaging that the schema is more focused around simply connecting to an MQTT broker.

openspec/changes/generic-mapper/design.md
```diff
@@ -0,0 +1,206 @@
...
+- A packaging/distribution format for custom mappers
+- Custom code-based converters (all conversion is via flows)
+
+**Aspirational (may or may not be worth pursuing):**
+
+- First-class `tedge-mapper <name>` subcommands for custom mappers, so they feel identical to built-in mappers. This would be ideal UX, but may pose insurmountable implementation hurdles (see D2). If it proves too costly, `tedge-mapper custom --profile <name>` is an acceptable fallback
+
+## Decisions
+
+### D1: Custom mapper names use `_` prefix
```
didier-wenzek: I don't share this concern about name clash. Wouldn't be enough to consider as custom any mapper that is not c8y, az or aws?

openspec/changes/generic-mapper/design.md
```diff
@@ -0,0 +1,206 @@
...
+
+Invoked as `tedge-mapper custom --profile _thingsboard`. This is straightforward to implement — clap handles all argument parsing, help text, and validation. The trade-off is that custom mappers feel second-class. However, it is significantly more searchable and discoverable — a user seeing `tedge-mapper custom` in docs or help output immediately knows where to look.
+
+The design should proceed with the fallback approach for the initial implementation, with the first-class approach as an upgrade if it proves feasible.
+
+### D3: Config is separate from `define_tedge_config!`, with `${mapper.*}` template namespace
+
+Custom mapper config lives entirely in the mapper's own directory (`/etc/tedge/mappers/_thingsboard/tedge.toml`), separate from the global `define_tedge_config!` schema interacted with through the `tedge config` CLI. The global config schema is large because it's referenced throughout thin-edge; custom mapper config is isolated (only the custom mapper itself uses it), so pulling it out is both sensible and simpler.
+
+The mapper's `tedge.toml` is parsed directly by the custom mapper, and is invisible to `tedge_config`. A minimal example:
```
didier-wenzek: Simple and effective.

openspec/changes/generic-mapper/design.md
```diff
@@ -0,0 +1,206 @@
...
+- Service name: `tedge-mapper-_thingsboard` (consistent with `tedge-mapper-c8y`)
+- Health topic: `te/device/main/service/tedge-mapper-_thingsboard/status/health`
+- Lock file: `/run/tedge-mapper-_thingsboard.lock`
+- Bridge service name: `tedge-mapper-bridge-_thingsboard`
+
+### D6: Warn about unrecognised mapper directories
+
+On mapper startup (and optionally on `tedge config` operations), scan `/etc/tedge/mappers/` and warn about directories that are not:
+
+- A known built-in mapper name (`c8y`, `az`, `aws`, `collectd`, `flows`)
```
didier-wenzek: I don't see any difference between a so called custom mapper and the `flows` or `local` mapper, except that the former might have extra configuration to connection a cloud and bridge the MQTT bus with that cloud.

openspec/changes/generic-mapper/specs/custom-mapper-config/spec.md
```diff
@@ -0,0 +1,80 @@
+## ADDED Requirements
+
+### Requirement: Custom mapper directory layout
+A custom mapper SHALL be defined by a directory under `/etc/tedge/mappers/` containing a `tedge.toml` configuration file. The directory MAY also contain a `bridge/` subdirectory with bridge rule TOML files and a `flows/` subdirectory with flow scripts. This layout mirrors the structure used by built-in mappers.
```
didier-wenzek: A mapper could also have some `flows` but no cloud connection, hence no `tedge.toml`

openspec/changes/generic-mapper/design.md
jarhodes314: Yes, this was my question the other day about whether we need make a distinction between a `local` mapper and a custom mapper.

openspec/changes/generic-mapper/design.md
jarhodes314: We could do that, it just potentially causes a conflict in the future if we added a new builtin mapper.

openspec/changes/generic-mapper/proposal.md
albinsuresh: Yeah it is the user-define config schema that I was referring to as "config templates".  It would have been nice if we provide a way for custom mappers to specify their config schema (beyond just the connection settings) as well, so that `tedge config` utility validates configs for that mapper using that schema.

> but for the custom clouds I'm envisaging that the schema is more focused around simply connecting to an MQTT broker.

If the configs are just limited to the connection settings, which would be the same for all custom mappers, then I don't really feel the cloud type is that useful. This cloud type is only used for this config schema distinction, right? Without that, wouldn't `thinksborad.profile1` and `thinksboard.profile2` be equivalent to `custom.thingsboard-profile1` and `custom.thingsboard-profile2`? What additional value do we get by categorising them under `thingsboard`?

jarhodes314: Aside from the nicer `tedge-mapper` interface, I think there probably isn't much value here.

Config schemas could be a way of providing value associated with a specific cloud, but I think, at least at the moment, there is no reason why anyone would gain anything by putting their own cloud-specific configuration inside `tedge.toml`. Currently all the cloud-specific configurations we have are motivated by built-in behaviour that the config values control, so I don't see what an custom cloud mapping would use `tedge.toml` for other than controlling the connection the built-in bridge makes.

openspec/changes/generic-mapper/proposal.md
albinsuresh: The main question in my mind is, what would the cloud type distinction help us (or the user) with? In my view, when multiple profiles are categorised under the same cloud type, it means that they all share something that is specific to that cloud. 

This is the reason why I was proposing that intermediate cloud type directory to accommodate any shared configurations/settings for that given cloud type. For example, if an externalised config schema existed for each cloud type, it could have been placed directly under that cloud directory, which is shared by all profiles of that cloud type as follows:

```
/etc/tedge/mappers/
|-- c8y
|   |-- schema.json
|   |-- main (the default profile)
|   |-- profile1
|-- thingsboard
|   |-- schema.json
|   |-- main
|   |-- profile1
|   |-- profile2
```

For now, the config schema is the only thing that is shared. In future we can even think of shared flow definitions or flow scripts that can be placed under this root directory.

> because otherwise we get the question of what do we call the unnamed directory.

Naming it `main` is one option, as most users would be familiar with that from `git`.

openspec/changes/generic-mapper/proposal.md
albinsuresh: > I don't see what an custom cloud mapping would use tedge.toml for other than controlling the connection the built-in bridge makes.

I was under the assumption that the `tedge.toml` file would be used to define the config settings as well as any other configurations controlling the mapper behaviour as well.

Even for the connection part, I was thinking that may be there are things like conditional topic subscriptions, or last will support that changes from one cloud to another. For example, if a cloud supports the last will feature, it would be good if he can capture all that configs in a single config file itself (`tedge.toml`), right? Otherwise he'll have to configure the main connection settings in `tedge.toml` and the last will settings in some other file, right?

But, if `tedge.toml` was controlled by a cloud specific schema, then I'd expect the user to use the `tedge config set` command to set everything as follows:

```
tedge config set thingsboard.url demo.thingsboard.io
tedge config set thingsboard.last_will_topic my/last/will/topic
tedge config set thingsboard.last_will_payload my_last_will_payload
```

For some other cloud that does not support the last will feature, then its schema wouldn't have defined those lastwill fields and hence a user attempting `tedge config set other_cloud.last_will_topic my/last/will/topic` would get an error.

So, my vision was that a mapper developer should be able to define the configs supported by his mapper using a schema and the users can use `tedge config` utility itself to set everything that is defined in that schema. The mapper then just reads the `tedge.toml` file and has access to everything that it needs. From a user's PoV, he'll have a single configuration interface (`tedge config` utility) as well to configure everything.

jarhodes314: I think we can consider it totally infeasible to support `tedge config set thingsboard.url demo.thingsboard.io` due to possible name collisions between static configuration keys and custom mappers.

We could do something like `tedge config set --profile thingsboard custom.url demo.thingsboard.io` and potentially add in some syntactic sugar like `tedge config set +thingsboard.url ...` (or something like `mapper-thingsboard.url`, essentially anything that clearly disambiguates the first segment of the config path from a static configuration key).

Supporting things like last_will messages is an important feature to capture.

I think the answer for the first implementation step is to work under the following assumptions:
1. Custom mappers are executed as `tedge-mapper custom --profile mapper-name`, with configuration in `/etc/tedge/mappers/custom.mapper-name`
2. For the initial implementation, these mappers deserialise their configuration directly
3. This gives us the ability to add `tedge config` support as a follow-up step, and we can consider syntactic sugar that makes things more convenient without having to constrain ourselves before we embark on implementing anything.

openspec/changes/generic-mapper/proposal.md
jarhodes314: > Naming it `main` is one option, as most users would be familiar with that from `git`.

This is true, but with git branches, you have to have a name. With mapper profiles, there is no requirement to have a name, so any potential profile name (including `main`, which I could absolutely believe someone wanting to use as a profile name) is currently available for users to use.

I suspect the case of needing to connect multiple custom mappers that share configuration is pretty niche, and I imagine good tooling for packaging/using symlinks to share logic between directories is likely a perfectly adequate solution here.

I also think the other design constraints are pushing us in the direction of profiles of a single custom mapper, rather than custom cloud types which can have their own profiles.

openspec/changes/generic-mapper/design.md
jarhodes314: Following the discussions with Albin in https://github.com/thin-edge/thin-edge.io/pull/4009#discussion_r2850668366, I think there probably is some value in supporting `tedge config` (or at least allowing for that), but I think implementing this for the initial solution is more than adequate (and I'm slightly wondering whether the `${mapper.key}` syntax is worthwhile in any case since I don't think people implementing custom mappers will actually need to care about non-mapper config variables, so a terse syntax is probably quite nice).

openspec/changes/generic-mapper/proposal.md
albinsuresh: > I suspect the case of needing to connect multiple custom mappers that share configuration is pretty niche, and I imagine good tooling for packaging/using symlinks to share logic between directories is likely a perfectly adequate solution here.

Sure. We just need to confirm if there is the possibility of shared configs or shared flow definitions in the immediate future, that we can accommodate with that symlink approach or any other approach. We don't necessarily need to provide that feature now, but just have a way forward to accommodate shared artefacts if/when the need arises. I was just stressing on it now itself,  while we still have time to plan for the future before the release, as I didn't want to risk locking down the filesystem layout with limited scope for extensibility in future. 

> I also think the other design constraints are pushing us in the direction of profiles of a single custom mapper, rather than custom cloud types which can have their own profiles.

Fair point. TBH, I also feel that the only cloud where users are more likely to use multiple profiles is C8Y.

crates/core/tedge_mapper/src/custom/config.rs
```diff
@@ -0,0 +1,170 @@
...
+//!
+//! The full TOML table is available for `${mapper.*}` template expansion in bridge rules.
+
+use anyhow::Context;
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+use tedge_config::models::HostPort;
+use tedge_config::models::MQTT_TLS_PORT;
+
+/// Parsed custom mapper `tedge.toml` configuration.
```
albinsuresh: Since the contents of this file are read from the `mapper.*` namespace, wondering if naming the config file `mapper.toml` instead of `tedge.toml` would make it easier for the users even us to easily do that association in our heads. I might even go one step further and change the `config.*` namespace into `tedge.*.

jarhodes314: I have now renamed `tedge.toml` to `mapper.toml` following discussion with @reubenmiller on Friday

> I might even go one step further and change the config.* namespace into `tedge.*.

That's an interesting thought, I'll give it some consideration

openspec/changes/generic-mapper/specs/custom-mapper-config/spec.md
jarhodes314: Now fixed so this is considered valid. We warn if neither ~`tedge.toml`~`mapper.toml` or `flows/` exist, and if flows exists but is empty:
```
2026-03-18T16:51:49.75413481Z  WARN tedge_mapper::core::mappers_dir: Unrecognised mapper directory '/etc/tedge/mappers/emptydir': no 'mapper.toml' or 'flows/' found. This directory will be ignored.
2026-03-18T16:51:49.755822334Z  WARN tedge_mapper::core::mappers_dir: Mapper 'local' has a 'flows/' directory but no flow scripts are defined.
```

docs/src/references/mappers/configurable-bridge.md
```diff
@@ -95,6 +105,36 @@ topic = "shadow/#"
...
+This is useful for values specific to your cloud integration that don't belong in the global `tedge.toml`:
+
+```toml
+# /etc/tedge/mappers/thingsboard/mapper.toml
+url = "mqtt.thingsboard.io:8883"
+
+[bridge]
+topic_prefix = "v1/devices/me"  # custom field, accessible as ${mapper.bridge.topic_prefix}
+```
+
```
didier-wenzek, line 121: I would add a sentence to stress the relationship between the two files.

```suggestion
The values defined by `mapper.toml` can then be used to adapt the bridge rules:
```

docs/src/references/mappers/user-defined-mappers.md
```diff
@@ -0,0 +1,263 @@
...
+        └── telemetry.toml
+```
+
+The mapper name must:
+- start with a lowercase ASCII letter (`a`–`z`)
+- contain only lowercase letters, digits, and hyphens (`-`)
+- not contain underscores or uppercase letters
+
+Valid names: `thingsboard`, `my-cloud`, `edge-platform`
+Invalid names: `ThingsBoard`, `my_cloud`, `1cloud`
```
didier-wenzek, line 40: Otherwise, markdown renders this on a single line.

```suggestion
Valid names: `thingsboard`, `my-cloud`, `edge-platform`

Invalid names: `ThingsBoard`, `my_cloud`, `1cloud`
```

crates/core/tedge/src/cli/mapper/cli.rs
```diff
@@ -0,0 +1,371 @@
...
+use crate::log::MaybeFancy;
+use crate::ConfigError;
+use camino::Utf8Path;
+use camino::Utf8PathBuf;
+use tedge_config::TEdgeConfig;
+
+#[derive(clap::Subcommand, Debug)]
+pub enum MapperCli {
+    /// List available mappers and their cloud type
+    List,
```
didier-wenzek, line 20: What am I doing wrong?

```
$ tedge mapper list
No mappers found under '/etc/tedge/mappers'
```

Despite having a setting looking correct:

```
$ ls /etc/tedge/mappers                                             
aws  az  c8y  c8y.foo  local  thingsboard

$ fd . /etc/tedge/mappers/thingsboard 
/etc/tedge/mappers/thingsboard/flows/
/etc/tedge/mappers/thingsboard/flows/greeting-flows/
/etc/tedge/mappers/thingsboard/flows/greeting-flows/flow.toml
/etc/tedge/mappers/thingsboard/flows/greeting-flows/hello.js
/etc/tedge/mappers/thingsboard/flows/greeting-flows/params.toml
/etc/tedge/mappers/thingsboard/rules/

$ tedge-mapper thingsboard 
... running
```

Is this supposed to work only with the legacy mappers? I noticed the following:

```
$ tedge mapper list
No mappers found under '/etc/tedge/mappers'

$ mv /etc/tedge/mappers/c8y.foo/tedge.toml /etc/tedge/mappers/c8y.foo/mapper.toml

$ tedge mapper list                     
c8y.foo
```

Okay, I got it. This expects a `mapper.toml`

```
$ touch /etc/tedge/mappers/thingsboard/mapper.toml            

$  tedge mapper list                               
c8y.foo
thingsboard
```

This is a bit misleading, as this detects less mappers than working mappers.

docs/src/references/mappers/user-defined-mappers.md
```diff
@@ -0,0 +1,263 @@
...
+```sh
+tedge mapper list
+```
+
+```text title="Example output"
+c8y       cloud_type=c8y
+thingsboard
+```
+
+### `tedge mapper config get`
```
didier-wenzek, line 168: Nice to see that this is also working for the legacy mappers:

```
$ tedge mapper config get c8y.url
didier.latest.stage.c8y.io
```

And compatible with `tedge config get`:

```
$ tedge config get c8y.url
didier.latest.stage.c8y.io
```

__However, is the plan to deprecate cloud profile?__

```
$ tedge config get c8y.url --profile foo 
foo.cumulocity.com

$ tedge mapper config get c8y.url --profile foo
error: unexpected argument '--profile' found

$ tedge mapper config get c8y.foo.url          
Error: failed to get config key 'foo.url' for mapper 'c8y'

```

crates/core/tedge/src/cli/mapper/cli.rs
jarhodes314: > This is a bit misleading, as this detects less mappers than working mappers.

Yes, this is a relic from the faulty "all mappers have a `mapper.toml`" assumption I was working from. I'll address this now

docs/src/references/mappers/configurable-bridge.md
jarhodes314: I would like to fix this so that the `${mapper.` namespace works everywhere (i.e. for built-in mappers as well as custom mappers) before I document this here, but I'm working on that now

docs/src/references/mappers/user-defined-mappers.md
reubenmiller: > However, is the plan to deprecate cloud profile?

Yes, I would like to get rid of cloud profiles eventually...I think cloud profiles hasn't had the traction that we'd initially thought, and now that we have custom cloud naming, then the advantage of having a cloud profile is reduced...and it just adds more complications than the streamlined generic mapper.

didier-wenzek: > Yes, I would like to get rid of cloud profiles eventually..

I'm fully aligned with this plan. As a user, this has always been difficult to remember the small differences between a main mapper and its profiles. The move done by James a few months ago with separate tedge config improved things a bit.

docs/src/references/mappers/user-defined-mappers.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+#### Template variables (`${mapper.*}`)
+
+The entire `mapper.toml` is available for template expansion in bridge rule files via the `${mapper.*}` namespace.
+This means you can reference any field — including user-defined ones — directly in your bridge rules without repeating values:
+
+```toml
+# bridge/rules.toml — uses ${mapper.bridge.topic_prefix} from mapper.toml
+remote_prefix = "${mapper.bridge.topic_prefix}/"
```
didier-wenzek, line 102: There is slight inconsistency with this syntax for substitution expressions.

For mapper rules, these expressions don't start with a dot while they do in the case of  tedge flows and workflows.

- Somehow this makes sense. In some context of tedge workflows, the expression `${.}` is valid, while this is never the case for mapper rules.

docs/src/references/mappers/user-defined-mappers.md
```diff
@@ -0,0 +1,265 @@
...
+    │   └── rules.toml
+    └── flows/            ← optional: message transformation flows
+        └── telemetry.toml
+```
+
+The mapper name must:
+
+- start with a lowercase ASCII letter (`a`–`z`)
+- contain only lowercase letters, digits, and hyphens (`-`)
+- not contain underscores or uppercase letters
```
albinsuresh, line 36: Not arguing against that restriction, but any specific reason to restrict upper-case?

docs/src/references/mappers/user-defined-mappers.md
```diff
@@ -0,0 +1,265 @@
...
+sidebar_position: 5
+---
+
+import ProposalBanner from '@site/src/components/ProposalBanner'
+
+<ProposalBanner/>
+
+## Overview
+
+In addition to the built-in cloud mappers (Cumulocity, Azure, AWS), thin-edge supports user-defined mappers that can bridge a device to any cloud platform.
```
albinsuresh:
```suggestion
In addition to the built-in cloud mappers (Cumulocity, Azure, AWS), thin-edge supports user-defined mappers that can bridge a device to any IoT cloud platform with an MQTT interface.
```

docs/src/references/mappers/user-defined-mappers.md
```diff
@@ -0,0 +1,265 @@
...
+
+When `device.cert_path` and `device.key_path` are absent from `mapper.toml`, the mapper falls back to `device.cert_path` and `device.key_path` from the root `tedge.toml`.
+Explicit `mapper.toml` values always take precedence.
+
+#### `cloud_type` field
+
+The `cloud_type` field identifies which built-in cloud integration this mapper belongs to.
+Valid values are: `c8y`, `az`, `aws`.
+
+For custom cloud platforms (e.g. ThingsBoard), you should not set this field.
```
albinsuresh, line 129: That conflicts with the concrete example shown above, where the `cloud_type` value is  defined as `thingsboard`.

tests/RobotFramework/tests/custom_mapper/mapper-password.toml
```diff
@@ -0,0 +1,6 @@
+url = "localhost:18884"
+auth_method = "password"
```
albinsuresh, line 2: While testing a custom mapper connection to the public [EQMX broker](https://www.emqx.com/en/mqtt-dashboard#/dashboard/overview), I encountered an issue where the mapper was trying to establish a TLS connection, despite setting the `auth_method = "password"` as follows:

```
2026-03-25T07:27:07.618366352Z  WARN MQTT bridge: [cloud] Connection error: Tls(Io(Custom { kind: UnexpectedEof, error: "tls handshake eof" }))
2026-03-25T07:27:07.618438579Z ERROR MQTT bridge: [cloud] MQTT bridge failed to connect to cloud broker: TLS: I/O: tls handshake eof
```

Here is my `mapper.toml`:

```
url = "broker.emqx.io:1883"

auth_method = "password"
credentials_path = "credentials.toml"
```

and the `credentials.toml`:
```
[credentials]
username = ""
password = ""
```

Is there any other setting, other than the port specified explicitly, that I need to use to enforce an insecure connection?

The following works just fine:
```
mosquitto_sub -v -h broker.emqx.io -p 1883 -t "albin/#"
```
or
```
mosquitto_sub -v -h broker.emqx.io -p 1883 -u "" -P "" -t "albin/#"
```

tests/RobotFramework/tests/custom_mapper/custom_mapper.robot
```diff
@@ -0,0 +1,278 @@
...
+    ${output}=    Should Have MQTT Messages
+    ...    custom/test/out
+    ...    minimum=1
+    ...    date_from=${start}
+    Should Contain    ${output[0]}    temperature
+
+    [Teardown]    Stop And Remove Mapper    test-full
+
+Two custom mapper profiles run concurrently without interfering
+    [Documentation]    Two custom mapper profiles (alpha and beta) can run simultaneously as
```
didier-wenzek, line 125: The term *profile* is now confusing.

```suggestion
    [Documentation]    Two custom mappers (alpha and beta) can run simultaneously as
```

tests/RobotFramework/tests/custom_mapper/mapper-password.toml
jarhodes314: Yes, I need to add a configuration that can disable TLS (because certificate/password just controls whether we use client certificates as authentication in that TLS connection).

docs/src/references/mappers/user-defined-mappers.md
jarhodes314: I don't want issues with some things being case-sensitive and some things being case-insensitive. Blocking uppercase is the simplest way to avoid this.

docs/src/references/mappers/user-defined-mappers.md
jarhodes314: Now done

docs/src/references/mappers/user-defined-mappers.md
jarhodes314: Oops, now removed the contradictary bits of the concrete examples

docs/src/references/mappers/user-defined-mappers.md
```diff
@@ -0,0 +1,261 @@
...
+[bridge]
+# Whether to use a clean MQTT session (default: false)
+clean_session = false
+
+# MQTT keepalive interval (e.g. "60s", "2m")
+# keepalive_interval = "60s"
+
+# Any additional fields you add here are available as ${mapper.*} in bridge rules.
+# For example, this field is accessible as ${mapper.bridge.topic_prefix}.
+topic_prefix = "v1/devices/me"
```
albinsuresh, line 92: Just wondering if adding support for some in-line bridge rules in this config file as follows:

```
[rules]
local_prefix = "custom-local/"
remote_prefix = "custom-remote/"

[[rule]]
topic = "test"
direction = "outbound"

```
 would make it convenient for users to deploy just one config with all cloud connectivity aspects. Since a cloud connection isn't really useful without the bridge rules, it'd also force the users define at least one bridge rule to have a meaningful mapper.

We'll still keep the `bridge` directory as an extension location for the users to dynamically install more rules, without touching this main config file.

crates/core/tedge_mapper/src/lib.rs
```diff
@@ -121,6 +167,11 @@ pub enum MapperName {
     },
     Collectd,
     Local,
```
didier-wenzek, line 171: I would consider to deprecate the local mapper from the rust code base (i.e. the `MapperName::Local` enum case and the `tedge_mapper::flows` module).

- The local mapper is simply a custom mapper without any cloud connection.
- This would also help to address https://github.com/thin-edge/thin-edge.io/issues/4048

@reubenmiller what's your point of view on that?

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -583,6 +585,11 @@ define_tedge_config! {
         #[serde(skip)]
         mapper_config_file: Utf8PathBuf,
 
+        /// Identifies this mapper directory as belonging to the Cumulocity cloud integration.
+        /// Written into `mapper.toml` automatically; not user-configurable via the CLI.
+        #[tedge_config(reader(skip))]
+        cloud_type: CloudType,
```
albinsuresh, line 591: Are we adding this now, despite the same information being available from the mapper directory name, to allows users to name their mapper directories anything (without enforcing the `c8y.`, `az.` or `aws.` prefixes) in the future?

crates/core/tedge/src/cli/bridge/test_command.rs
```diff
@@ -124,42 +64,51 @@ impl Command for BridgeTestCmd {
...
+                    }
+                    if !rules.is_empty() {
+                        return Ok(print_topic_matches(w, &cmd.topic, &rules));
+                    }
+                }
+                Ok(Status::NoMatches)
+            }
+            #[cfg(feature = "aws")]
+            Cloud::Aws(_) => {
+                print_non_configurable_or_disabled(w, config, &cloud);
```
albinsuresh, line 93: Are we planning to fill this gap before the release? Or is there a strategic reason why we don't support the same specifically for `aws` and `az`?

crates/core/tedge/src/cli/mapper/cli.rs
```diff
@@ -0,0 +1,936 @@
...
+        #[clap(subcommand)]
+        cmd: MapperConfigCmd,
+    },
+}
+
+#[derive(clap::Subcommand, Debug)]
+pub enum MapperConfigCmd {
+    /// Get a config value from a mapper's mapper.toml
+    ///
+    /// The key is in the form `<mapper-name>.<toml-key-path>`, e.g. `thingsboard.url`
```
albinsuresh, line 33: No profile support planned for this `get` command?

crates/core/tedge/src/cli/mapper/cli.rs
```diff
@@ -0,0 +1,936 @@
...
+                    row.device_id.dim(),
+                    cloud_type.dim(),
+                );
+            }
+        }
+        Ok(())
+    }
+}
+
+/// `tedge mapper config get thingsboard.url` — returns an effective config value.
```
albinsuresh, line 169: The `get` command isn't working for me for `c8y`, where the config values are still kept in `tedge.toml` and not `mapper.toml`. I'm getting the following output, with that `WARN` indicating the same:

```
# tedge mapper config get c8y.url
2026-03-25T15:47:10.213502317Z  WARN tedge_config::tedge_toml::tedge_config: c8y configuration found in `tedge.toml`, but this will be ignored in favour of configuration in /etc/tedge/mappers/c8y/mapper.toml and /etc/tedge/mappers/c8y.*/mapper.toml
Error: failed to get config key 'url' for mapper 'c8y'

Caused by:
    Mapper 'c8y' has no mapper.toml at '/etc/tedge/mappers/c8y/mapper.toml'
```

If this restriction is intentional, it would be better to give an addtional hint to use the `tedge config get` command instead of `tedge mapper conifg get` for such cases.

BTW `tedge config get` also appears broken, as I'm getting the following error, despite the value being there in `tedge.toml`:

```
# tedge config get c8y.url
2026-03-25T15:49:19.228108792Z  WARN tedge_config::tedge_toml::tedge_config: c8y configuration found in `tedge.toml`, but this will be ignored in favour of configuration in /etc/tedge/mappers/c8y/mapper.toml and /etc/tedge/mappers/c8y.*/mapper.toml
The provided config key: 'c8y.url' is not set
```

crates/core/tedge_mapper/src/c8y/bridge/mqtt-core.toml
```diff
@@ -3,21 +3,21 @@
 # Cumulocity MQTT Quick Reference: https://cumulocity.com/docs/smartrest/quick-reference/
 # thin-edge.io Configurable built-in bridge reference: https://thin-edge.github.io/thin-edge.io/next/references/mappers/configurable-bridge/
 
-local_prefix = "${config.c8y.bridge.topic_prefix}/"
+local_prefix = "${mapper.bridge.topic_prefix}/"
```
albinsuresh, line 6: That doesn't seem to be working when the `bridge.topic_prefix` is defined in `tedge.toml` instead of `mapper.toml` as follows:

```
# tedge bridge inspect c8y
Bridge configuration for Cumulocity
Reading from: /etc/tedge/mappers/c8y/bridge

Error: Failed to expand bridge configuration
   ╭─[ /etc/tedge/mappers/c8y/bridge/rules.toml:1:26 ]
   │
 1 │ local_prefix = "${mapper.bridge.topic_prefix}/"
   │                          ─────────┬─────────  
   │                                   ╰─────────── Failed to expand local_prefix: Key 'bridge.topic_prefix' not found in mapper config
   │ 
   │ Note: Check that 'bridge' exists in your mapper's mapper.toml
───╯
2026-03-25T17:02:17.506654072Z ERROR tedge::cli::bridge::common: Invalid bridge config template
Failed to read bridge config files
```

while I have `c8y.bridge.topic_prefix` defined in `tedge.toml` as follows:

```
# tedge config get c8y.bridge.topic_prefix
c8y
```

crates/core/tedge/src/cli/mapper/cli.rs
```diff
@@ -0,0 +1,936 @@
...
+#[derive(clap::Subcommand, Debug)]
+pub enum MapperConfigCmd {
+    /// Get a config value from a mapper's mapper.toml
+    ///
+    /// The key is in the form `<mapper-name>.<toml-key-path>`, e.g. `thingsboard.url`
+    /// or `thingsboard.device.cert_path`.
+    ///
+    /// Schema-level keys (`url`, `device.id`, `device.cert_path`, `device.key_path`,
+    /// `device.root_cert_path`) return the *effective* value — including cert CN inference
+    /// and `tedge.toml` fallbacks — with a source annotation on stderr.
```
albinsuresh, line 38: The following `WARN` appearing even when I am not querying a `c8y` config value is mildly annoying.

```
# tedge mapper config get tb.url
2026-03-25T16:55:39.328786349Z  WARN tedge_config::tedge_toml::tedge_config: c8y configuration found in `tedge.toml`, but this will be ignored in favour of configuration in /etc/tedge/mappers/c8y/mapper.toml and /etc/tedge/mappers/c8y.*/mapper.toml
me.tb.io:8883
# from mapper.toml
```

crates/core/tedge/src/cli/mapper/cli.rs
```diff
@@ -0,0 +1,936 @@
...
+    "device.key_path",
+    "device.root_cert_path",
+];
+
+#[async_trait::async_trait]
+impl BuildCommand for MapperCli {
+    async fn build_command(self, config: &TEdgeConfig) -> Result<Box<dyn Command>, ConfigError> {
+        let mappers_root = config.root_dir().join("mappers");
+        match self {
+            MapperCli::List => Ok(ListMappersCommand { mappers_root }.into_boxed()),
```
reubenmiller, line 51: Whilst the `tedge mapper list` is pretty but it makes it harder for tools to parse due to the leading whitespace,

```sh
tedge mapper list
        c8y
thingsboard  mqtt.thingsboard.cloud:8883  TST_score_thundering_cream [tedge.toml]
```

The list should be easily parsable for scripts as it will make it easy for monitoring tools to check which bridges are configured etc.

Using `cut` (which only has a simple delimiter parsing) doesn't yield consistent results across the mappers (the c8y from above is missing due the leading whitespace).

```sh
tedge mapper list | cut -d' ' -f1

thingsboard
```

reubenmiller: There are a few strategies to solve this
* keep the output simple single column output
* detect when writing to a console, and only do the fancy multi column output then, otherwise stick with some tab delimited output (as cut's default delimiter is tab)

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
jarhodes314: Yes, this should (hopefully not far in the future) allow us to essentially make profiles redundant in favour of custom clouds with this `cloud_type` field set. For now, we're just automatically setting this from the built-in mappers to the relevant value.

crates/core/tedge/src/cli/mapper/cli.rs
jarhodes314: It sounds like from the warning that you've got a mapper.toml for a c8y profile but not for your main c8y config, so I think this is just a broken config on your end (the change that introduced mapper.toml explictly forbids cloud config simultaneously in `mapper.toml` files and `tedge.toml`). If you use `mapper.toml` for built-in clouds, you have to use only that for built-in clouds.

crates/core/tedge/src/cli/mapper/cli.rs
jarhodes314: That's just a function of the way our config warnings always appear when running a `tedge` command

crates/core/tedge_mapper/src/c8y/bridge/mqtt-core.toml
jarhodes314: This is the same issue observed by Didier and the system tests. I'm working on fixing it locally

crates/core/tedge/src/cli/mapper/cli.rs
reubenmiller: I'd like to get rid of cloud profiles all-together as the configurable mapper/bridge delivers much better flexibility with less complexity exposed to the user.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
didier-wenzek: > Yes, this should (hopefully not far in the future) allow us to essentially make profiles redundant in favour of custom clouds with this `cloud_type` field set.

I really like that way of telling a mapper has some c8y specifics, compared to deriving this from the mapper name.

crates/core/tedge/src/cli/mapper/cli.rs
albinsuresh: Okay. I did **not** have configs for the default `c8y` profile split between `tedge.toml` and `mapper.toml`. The default `c8y` profile configs were only present in `tedge.toml`. But then I added a new `c8y.edge` profile with a `mapper.toml`. That's when the `tedge config get` started breaking for the default `c8y` profile as well.

> If you use mapper.toml for built-in clouds, you have to use only that for built-in clouds.

All builtin clouds as well as their profiles, right? So, you can't have the `default` profile configs at one location (`tedge.toml`) and other profiles in a different location (`mapper.toml`). Either put all in `tedge.toml` or all in `mapper,toml`, right?

crates/core/tedge/src/cli/bridge/test_command.rs
jarhodes314: Yes, I am planning to, but in a separate PR. This should be pretty simple to implement

crates/core/tedge_mapper/src/c8y/bridge/mqtt-core.toml
jarhodes314: This should now be fixed @albinsuresh @didier-wenzek

crates/core/tedge/src/cli/mapper/cli.rs
jarhodes314: Now solved (by just using tab-delimited output)

docs/src/references/mappers/user-defined-mappers.md
didier-wenzek: We decided offline to fix that for flows, making the dot optional there.

crates/core/tedge_mapper/src/lib.rs
didier-wenzek: We might consider this. But in a follow-up PR.

crates/core/tedge_mapper/src/c8y/bridge/mqtt-core.toml
didier-wenzek: This is now working nicely.

crates/core/tedge/src/cli/mapper/cli.rs
didier-wenzek: I'm happy to deprecate cloud profile, however we might have to adapt our conventions.

On my laptop I have a configured cloud profile for c8y (created before that PR and simply updated with a `mv tedge.toml mapper.toml`).

This mapper is correctly listed:

```
$ tedge mapper list                      
c8y	didier.latest.stage.c8y.io:8883	didier-789 [cert CN]	
c8y.foo	broken.cumulocity.com:8883	didier-789 [cert CN]	
local
```

But the profile suffix confuses `tedge mapper config`

```
$ tedge mapper config get c8y.device.id 
didier-789
# inferred from certificate CN (/etc/tedge/device-certs/demo-device-888.pem)


$ tedge mapper config get c8y.foo.device.id
Error: failed to get config key 'foo.device.id' for mapper 'c8y'

Caused by:
    Unknown mapper config key: c8y.foo.device.id
```

crates/core/tedge/src/cli/mapper/cli.rs
didier-wenzek: This is now working with no surprises: all the working mappers are detected.

docs/src/references/mappers/user-defined-mappers.md
didier-wenzek: I would avoid to inline some bridge rules in the mapper config.

Sure ,it makes no sense to have a cloud connection without a bridge
and, sure, users are forced to create an other file to define those bridge rules.

However, clarity and simplicity will make life easier for everyone.

docs/src/references/mappers/user-defined-mappers.md
didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/4082

crates/core/tedge/src/cli/mapper/cli.rs
jarhodes314: > All builtin clouds as well as their profiles, right? So, you can't have the `default` profile configs at one location (`tedge.toml`) and other profiles in a different location (`mapper.toml`). Either put all in `tedge.toml` or all in `mapper,toml`, right?

Yes, this is correct

Issue #3998, opened 2026-02-17
Add `onStartup()` function


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
Add a function `onStartup()`, which is triggered when the `tedge-flows` process starts.
This would be helpful when user wants to do specific initialization only on the startup, like adding specific items to the KV store.

**Describe the solution you'd like**
Below shows the example how to initialize the mapper KV store for the main device on the startup.
```javascript
export function onStartup(time, context) {
  // Set main device property to KV store
  context.mapper.set("key::device/main//", "mainDeviceName");
  context.mapper.set("rev::mainDeviceName", "device/main//");
}
```

**Describe alternatives you've considered**

**Additional context**

PR #4011, opened 2026-02-25
feat: Execute `onStartup` function for loaded js modules

Bravo555:
# TODO

- [x] should `onStartup` be able to publish messages?
- [x] fix `onStartup` scheduling in `tedge flow test`
- [x] rerun `onStartup` on script reload
- [x] should `onStartup` run strictly before updating status of the flow? If `onStartup` fails, maybe we should avoid publishing status of the flow until it succeeds.
- [x] cleanup scheduling for transformer steps
- [x] cleanup
- [x] `onStartup` needs to run before `onMessage`
- [x] add more tests for running onStartup only starting from affected steps
- [x] add to `Transformer`

## Proposed changes

If it is exported by the module, execute `onStartup` function immediately after the module is loaded.

It is passed a timestamp and the context object, allowing the user to read and set KVstore values on script startup.

For every step where `on_startup` needs to run, it runs before `on_message` (if `on_message` also needs to run) and can react to `on_startup` messages from previous steps.

## Types of changes

## Paste Link to the issue
#3998

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -1390,6 +1390,10 @@ define_tedge_config! {
             /// Enable or disable statistics for step onInterval
             #[tedge_config(default(value = false))]
             on_interval: bool,
+
+            /// Enable or disable statistics for step onStartup
+            #[tedge_config(default(value = false))]
+            on_startup: bool,
         },
     },
 }
--- a/crates/core/tedge_mapper/src/lib.rs
+++ b/crates/core/tedge_mapper/src/lib.rs
@@ -188,6 +188,7 @@ pub(crate) fn flows_config(
         stats_config.interval.duration(),
         stats_config.on_message,
         stats_config.on_interval,
+        stats_config.on_startup,
     )
     .with_js_config(
         mem_config.heap_size as usize,
--- a/crates/extensions/c8y_mapper_ext/src/tests.rs
+++ b/crates/extensions/c8y_mapper_ext/src/tests.rs
@@ -3281,6 +3281,7 @@ pub(crate) async fn c8y_mapper_builder(
         Duration::from_secs(300),
         false,
         false,
+        false,
     );
 
     let mut flows_mapper = FlowsMapperBuilder::try_new(flows, service_config)
--- a/crates/extensions/tedge_flows/src/lib.rs
+++ b/crates/extensions/tedge_flows/src/lib.rs
@@ -71,6 +71,7 @@ impl Default for FlowsMapperConfig {
             Duration::from_secs(300),
             false,
             false,
+            false,
         )
     }
 }
@@ -82,6 +83,7 @@ impl FlowsMapperConfig {
         stats_dump_interval: Duration,
         publish_on_message_stats: bool,
         publish_on_interval_stats: bool,
+        publish_on_startup_stats: bool,
     ) -> Self {
         let statistics_topic = format!("{topic_prefix}/status/metrics");
         let status_topic = format!("{topic_prefix}/status/flows");
@@ -101,6 +103,7 @@ impl FlowsMapperConfig {
             stats_filter: StatsFilter {
                 publish_on_message_stats,
                 publish_on_interval_stats,
+                publish_on_startup_stats,
             },
             js_config: JsRuntimeConfig::default(),
         }
--- a/crates/extensions/tedge_flows/src/stats.rs
+++ b/crates/extensions/tedge_flows/src/stats.rs
@@ -17,6 +17,7 @@ pub enum Dimension {
     Flow(String),
     OnMessage(String),
     OnInterval(String),
+    OnStartup(String),
 }
 
 pub enum Sample {
@@ -30,6 +31,7 @@ pub enum Sample {
 pub struct StatsFilter {
     pub(crate) publish_on_message_stats: bool,
     pub(crate) publish_on_interval_stats: bool,
+    pub(crate) publish_on_startup_stats: bool,
 }
 
 #[derive(Default)]
@@ -203,6 +205,7 @@ impl StatsFilter {
             Dimension::Flow(_) => true,
             Dimension::OnMessage(_) => self.publish_on_message_stats,
             Dimension::OnInterval(_) => self.publish_on_interval_stats,
+            Dimension::OnStartup(_) => self.publish_on_startup_stats,
         }
     }
 }
@@ -232,6 +235,7 @@ impl Display for Dimension {
             Dimension::Flow(toml) => write!(f, "{}", Self::filename(toml)),
             Dimension::OnMessage(js) => write!(f, "{}", Self::filename(js)),
             Dimension::OnInterval(js) => write!(f, "{}.onInterval", Self::filename(js)),
+            Dimension::OnStartup(js) => write!(f, "{}.onStartup", Self::filename(js)),
         }
     }
 }
@@ -241,6 +245,7 @@ impl Dimension {
         match f {
             "onMessage" => Some(Dimension::OnMessage(js.to_owned())),
             "onInterval" => Some(Dimension::OnInterval(js.to_owned())),
+            "onStartup" => Some(Dimension::OnStartup(js.to_owned())),
             _ => None,
         }
     }
@@ -255,6 +260,7 @@ impl Dimension {
             Dimension::Flow(_) => "flow",
             Dimension::OnMessage(_) => "onMessage",
             Dimension::OnInterval(_) => "onInterval",
+            Dimension::OnStartup(_) => "onStartup",
         }
     }
 }


crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -74,6 +74,7 @@ impl JsRuntime {
             match *export {
                 "onMessage" => script.is_defined = true,
                 "onInterval" => script.is_periodic = true,
+                "onStartup" => script.has_startup = true,
```
reubenmiller, line 111: Any objection to changing the name to `onInit`?

reubenmiller: Or maybe not if we add a similar lifecycle hook for when a flow is removed, e.g. "onShutdown"...as I would favour onShutdown over onInit/onDestroy.

reubenmiller: Decision: Let's stick with "onStartup"

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -82,6 +82,19 @@ Units are configured using topic metadata
     ...    ${expected_msg}
 
 Entity registration data are passed around using the context
+    # NOTE(flaky, #unrelated-msgs): this test can fail if there are unrelated messages posted between messages the test
```
rina23q: I'm not sure if I understood the problem correctly.
The tests marked "flaky" are using the `tedge flows test` CLI command, not `tedge-mapper-local` flow service. The output of the command should be deterministic and not be affected by other input source.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -150,6 +165,8 @@ Consuming messages from the tail of file
 
 Consuming messages from a file, periodically
     Install Flow    input-flows    read-file-periodically.toml
+    # marcel: here, after this flow is loaded, it spikes CPU and publishes thousands (on my machine ~50k) of error messages:
+    # [test/file/input] Error in /etc/tedge/mappers/local/flows/read-file-periodically.toml: Fail to poll /tmp/file.input: No such file or directory (os error 2)
```
rina23q: Probably it's because the interval is set to `100ms`.

rina23q (commented):
The Rust implementation and the system tests for `onStartup()` look good to me.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
Bravo555: Okay, now I see that the comment is wrong about tests being flaky, it fails consistently every time.

To reproduce the scenario I'm talking about, add `Install Nested Flow    onstartup` keyword to `Custom Setup`, after `Start Service    tedge-mapper-local`, so `onstartup` flow starts running immediately after setup.

The problem is that `tedge flows test` invocations in these tests don't use `--flow` argument to select only the flow they test, so they get messages from all the flows, and if some publish messages in `onStartup`, they get these as well. These tests should probably use `--flow` argument to test only the flow installed in the test and not any other flows.

I'll make a separate PR changing these tests and remove the note in this PR. Thanks for the feedback!

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
Bravo555: But if the messages were really published every `100ms`, there wouldn't be so many, right? I'll look at this more thoroughly.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
rina23q: > The problem is that tedge flows test invocations in these tests don't use `--flow` argument to select only the flow they test, so they get messages from all the flows, and if some publish messages in onStartup, they get these as well. These tests should probably use --flow argument to test only the flow installed in the test and not any other flows.

Ah! This one really makes sense. So, test isolation is not enough as of now. Good catch!

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
rina23q: True. Indeed, I changed the interval to `10s`, but it didn't take effect on the frequency of publishing the error message. It looks like something unexpected.

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -93,6 +95,30 @@ impl JsScript {
             .map_err(flow::error_from_js)?
             .try_into()
     }
+
+    /// Trigger the onStartup function of the JS module
+    ///
+    /// The "onInterval" function is passed 2 arguments
```
didier-wenzek:
```suggestion
    /// The "onStartup" function is passed 2 arguments
```

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -17,6 +17,11 @@ pub struct FlowStep {
     handler: StepHandler,
     interval: Duration,
     pub(crate) next_execution: Option<Instant>,
+    /// The point in time at which the step was loaded and startup executed.
+    ///
+    /// Used to invoke the step's `onStartup` function, if applicable. If so, `onStartup` function
+    /// is scheduled to execute immediately and `started_up` is set to an instant it completed.
+    started_at: Option<Instant>,
 }
```
didier-wenzek, line 25: I don't get why we need to hold in-memory this *started at* timestamp. Shouldn't this be a dynamic value?

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -179,6 +209,31 @@ impl FlowStep {
...
+        &mut self,
+        js: &JsRuntime,
+        timestamp: SystemTime,
+    ) -> Result<Vec<Message>, FlowError> {
+        let output = match &mut self.handler {
+            StepHandler::JsScript(script, config) => script.on_startup(js, timestamp, config).await,
+            StepHandler::Transformer(..) => {
+                // startup function is exclusive to js scripts, do nothing but ensure we're not going to try to run
+                // onstartup again
+                Ok(vec![])
```
didier-wenzek: Why not having startup functions for `Transformer`?

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -116,14 +126,28 @@ impl FlowStep {
...
         }
     }
 
     /// Check if this script should execute its interval function now
     /// Returns true and updates next_execution if it's time to execute
     pub fn should_execute_interval(&mut self, now: Instant) -> bool {
+        if self.started_at.is_none() {
+            // needs to execute startup function
+            return true;
+        }
```
didier-wenzek: I'm really confused by this mix of logic (`onInterval` vs `onStartup`).

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
Bravo555: Even rerunning the test on the main branch there are probably too many messages, so it's probably unrelated to changes in this PR.

```
$ just integration-test '-s tedge_flows'
$ grep -c "Fail to poll /tmp/file.input" results/output.xml 
37256
```

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -170,6 +194,12 @@ impl FlowStep {
         js: &JsRuntime,
         timestamp: SystemTime,
     ) -> Result<Vec<Message>, FlowError> {
+        if self.started_at.is_none() {
+            let messages = self.on_startup(js, timestamp).await?;
+            self.init_next_execution();
+            return Ok(messages);
+        }
```
didier-wenzek: Better to put side by side the check and recording of the started status.

```suggestion
        if self.started_at.is_none() {
            let messages = self.on_startup(js, timestamp).await?;
            self.started_at = Some(Instant::now());
            self.init_next_execution();
            return Ok(messages);
        }
```

crates/extensions/tedge_flows/src/js_script.rs
Bravo555: fixed in ae13f7344e

crates/extensions/tedge_flows/src/steps.rs
Bravo555: Currently, when the step is loaded we schedule the `on_interval` handler to be run immediately, which then looks at `started_at` to determine if it should run `onInterval` or `onStartup` function.

Indeed it's probably confusing for `on_interval` to handle both these cases. Two alternatives I see immediately are:

- rename `shoud_execute_interval` to `should_execute_scheduled` or similar, where we decide if the function we should execute is `on_interval` or `on_startup` by looking at the `started_at` field.
- do the rename as above, but remove `started_at` field and include what function should be executed in `next_execution`, then something like `on_scheduled` would be trivial (would just execute `on_interval` or `on_startup` depending on what's in `next_execution`)

crates/extensions/tedge_flows/src/steps.rs
Bravo555: The issue only mentioned adding it to js scripts and I wasn't sure how to use it with transformers, but indeed it makes sense to have it for transformers as well. I'll include that in the PR.

crates/extensions/tedge_flows/src/steps.rs
Bravo555: fixed in f073bae655

crates/extensions/tedge_flows/src/steps.rs
Bravo555: Not sure what you mean by _dynamic value_. As for the general approach and alternatives, see https://github.com/thin-edge/thin-edge.io/pull/4011/changes#r2904580901. Here I'll also add that it also could've just been a boolean, but I figured this information might be useful, e.g. to log, but it currently isn't being done.

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -170,6 +194,13 @@ impl FlowStep {
         js: &JsRuntime,
         timestamp: SystemTime,
     ) -> Result<Vec<Message>, FlowError> {
+        if self.started_at.is_none() {
+            let messages = self.on_startup(js, timestamp).await?;
```
albinsuresh: I don't really understand why we decided to piggy-back the `onStartup` call with `on_interval`. That connection is not very intuitive and that was evident from the artificial `tick` that was required from the test as well. Why couldn't we simply call the `onStartup` on all the steps (for which it is configured), right after the flow is loaded?

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -440,6 +442,18 @@ Params are dynamically reloaded
     ...    date_from=${start}
     ...    message_contains=Hello World!
 
+Script runs onStartup function
+    Install Nested Flow    onstartup
+    Logs Should Contain    JavaScript.Console: "hello from on-startup.js"    max_matches=1
+    Should Have MQTT Messages    topic=test/onstartup    message_contains=hello from on-startup.js    maximum=1
+
+    ${start}    Get Unix Timestamp
+    Execute Command    sleep 1
```
albinsuresh, line 531:
```suggestion
    Execute Command    sleep 1
    ${start}    Get Unix Timestamp
```
You wanted to make sure that the start time for the new message is definitely higher than the previous message, right?

crates/extensions/tedge_flows/src/steps.rs
didier-wenzek: > Not sure what you mean by _dynamic value_.

I agree this is not clear. I mean a value that only passed to some function call and never stored in some state.

> As for the general approach and alternatives, see https://github.com/thin-edge/thin-edge.io/pull/4011/changes#r2904580901.

The point I missed during my first review is that the `onStartup` callback must be called not only when the mapper starts but also when a flow or step is reloaded.

So you are correct, we need to know which script has already been started or not.

> Here I'll also add that it also could've just been a boolean, but I figured this information might be useful, e.g. to log, but it currently isn't being done.

A timestamp is fine. That said I think the code might be simpler if this value is set to some loading timestamp for the scripts with *no* `onStartup` callback.

crates/extensions/tedge_flows/src/steps.rs
Bravo555: > Why couldn't we simply call the `onStartup` on all the steps (for which it is configured), right after the flow is loaded?

It was being done before but had to be changed because `onStartup` needs to be able to return messages, and these messages need to end up in the top-level actor which publishes messages. But loading steps happens at the lower level and isn't directly invoked by the actor, like `on_message` or `on_interval` where the actor calls these functions and they return messages. With `on_startup` we could either:

- make `FlowsMapper` actor aware of steps being loaded so it can get these messages by invoking functions (increases API surface area and overall complexity)
- use some workaround where we piggyback these messages using some separate channel (add extra actor input for startup messages or pass a message publisher to flows directly, bad because we lose the property of the behaviour being driven completely by the top-level actor by invoking processor's functions)
- (this was chosen) reuse the interval/ticks mechanism to publish startup messages. That way there's no modifications to the actor and we retain the property that the actor drives the process; the difference is that sometimes when the actor runs its `on_interval`, it doesn't always run `onInterval`, but can run `onStartup`, which, after the script is loaded, is scheduled to be run immediately (on the next tick).

Last Tuesday we had a call with Didier where we decided to go with that approach, but indeed running `on_startup` from within `on_interval` is confusing and they should be kept distinct. For this, we can just change the naming such that the actor invokes `processor.on_scheduled()` and have `onInterval` and `onStartup` be functions that can be scheduled and on the lower level decide which should be run, as proposed in [this comment](https://github.com/thin-edge/thin-edge.io/pull/4011#discussion_r2904580901).

crates/extensions/tedge_flows/src/steps.rs
didier-wenzek: I'm more confused by the mix of logic than by the name.

- This method resets the timer for `onInterval` but not for `onStartup`.
- There is some late initialization (when `next_execution` is `None`). This should be done sooner.
- The interaction between an `onStartup` callback and the first `onInterval` is not clear. 

To be fair, this mix of logic predates this PR. This is more obvious now.

crates/extensions/tedge_flows/src/steps.rs
didier-wenzek: Actually, we need to find an intermediate solution.

The `onStartup` callback of a script *must* be called before any call to `onMessage`. This is not currently the case. Given a flow with two steps, the `onMessage` callback of the second step will be called on the message produced by the `onStartup` callback of the first step before  the `onStartup` callback of the second step.

So I think we have to introduce a `Flow::on_startup()` method which is similar to `Flow::on_interval()` but invokes the `onStartup` callback starting from the last step down to the fist.

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -108,6 +115,9 @@ impl FlowStep {
             // FIXME: there is bug here when the updated version adds an on_interval method
             // This method will be ignored, because the interval is zero (because there no on_interval method before)
             // => The configured interval must not be erased
+
+            // after reload trigger onStartup again
+            self.started_at = None;
```
didier-wenzek:
```suggestion
            if script.has_startup {
                self.started_at = None;
                self.next_execution = Some(Instant::now());
            } else {
                self.started_at = Some(Instant::now());
            }
```

crates/extensions/tedge_flows/src/steps.rs
Bravo555: Changed in f575a6d.

Now if a step in the flow is reloaded, we run `onStartup` for this and next steps from `Flow::on_startup`, so output messages aren't passed to `onMessage` but are published individually, and it all happens before any other calls to `onMessage`.

However, I don't understand why `onStartup` should be run starting from the last step down to the first. Wouldn't it be more intuitive to the users for startup to run in the usual ascending order?

crates/extensions/tedge_flows/src/steps.rs
Bravo555: fixed in f575a6d.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
Bravo555: fixed in b52b66965e

crates/extensions/tedge_flows/src/steps.rs
didier-wenzek: > However, I don't understand why `onStartup` should be run starting from the last step down to the first. Wouldn't it be more intuitive to the users for startup to run in the usual ascending order?

The mental model of a flow is a pipeline.

For a regular message without any `onStartup` callback, we have the following

```
input -> msg-0
---------------
step1.onMessage(msg-0) -> [msg-1,msg2]
---------------
step2.onMessage(msg-1) -> [msg-3,msg4] 
step2.onMessage(msg-2) -> [msg-5,msg6] 
----------------
[msg-3, msg4, msg-5,msg6] -> output
```

To start/init the pipline I expect the following,
i.e. each step calls first its own `onStartup` callback,
then pushes downstream the messages produced by `onStartup`
and finally uses `onMessage` to transform the messages that are received from upstream.

```
no input
---------------
step1.onStartup() -> [msg-1,msg2]
---------------
step2.onStartup() ->  [msg-3,msg4] 
step2.onMessage(msg-1) -> [msg-5,msg6] 
step2.onMessage(msg-2) -> [msg-7,msg8] 
----------------
[msg-3, msg4, msg-5,msg6, msg7, msg8] -> output
```

So you are correct, there is no requirement to run `onStartup` starting from the last step down to the first.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -298,6 +298,62 @@ impl Flow {
...
+    /// with onInterval, but are instead published individually.
+    pub async fn on_startup(
+        &mut self,
+        js_runtime: &JsRuntime,
+        stats: &mut Counter,
+        timestamp: SystemTime,
+        now: Instant,
+    ) -> FlowResult {
+        // TODO: update stats calls
+        let stated_at = stats.flow_on_interval_start(self.name());
```
jarhodes314: Typo

```suggestion
        let started_at = stats.flow_on_interval_start(self.name());
```

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -298,6 +298,62 @@ impl Flow {
...
+        let result = self
+            .on_startup_steps(js_runtime, stats, timestamp, now)
+            .await;
+        match &result {
+            Ok(messages) => {
+                stats.flow_on_interval_done(self.name(), stated_at, messages.len());
```
jarhodes314:
```suggestion
                stats.flow_on_interval_done(self.name(), started_at, messages.len());
```

crates/extensions/tedge_flows/src/runtime.rs
```diff
@@ -144,6 +144,22 @@ impl<Registry: FlowRegistryExt + Send> MessageProcessor<Registry> {
 
     pub async fn on_interval(&mut self, timestamp: SystemTime, now: Instant) -> Vec<FlowResult> {
         let mut out_messages = vec![];
+
+        let unstarted_flow = self
+            .registry
+            .flows_mut()
+            .find(|f| f.as_ref().steps.iter().any(|s| s.should_execute_startup()));
+        if let Some(flow) = unstarted_flow {
```
jarhodes314: Is there a reason why we only execute this for the first flow that needs `onStartup` calling, rather than all the flows that have the requirement?

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -298,6 +298,62 @@ impl Flow {
...
+        }
+        self.publish(result)
+    }
+
+    async fn on_startup_steps(
+        &mut self,
+        js_runtime: &JsRuntime,
+        stats: &mut Counter,
+        timestamp: SystemTime,
+        _: Instant,
```
jarhodes314: If this argument isn't needed, we should just remove it

crates/extensions/tedge_flows/tests/interval.rs
```diff
@@ -458,6 +458,106 @@ async fn interval_executes_when_time_exceeds_interval() {
...
+    captured_messages.retain(|_| false);
+
+    // Now send an input message to trigger onMessage
+    mqtt.publish("test/input", "hello").await;
+    tick(Duration::from_millis(50)).await;
+
+    let messages = get_messages();
+
+    // Verify only 1 onMessage after onStartup messages
+    assert_eq!(messages[0], "on_message 2: on_message 1: hello");
```
jarhodes314: The assertion doesn't actually verify there's only one message

```suggestion
    // Verify only 1 onMessage after onStartup messages
    assert_eq!(messages, vec!["on_message 2: on_message 1: hello"]);
```

crates/extensions/tedge_flows/src/runtime.rs
Bravo555: fixed in e534823a

crates/extensions/tedge_flows/tests/interval.rs
Bravo555: fixed in e2cebec246

crates/extensions/tedge_flows/src/flow.rs
Bravo555: fixed in 38cff8743b

crates/extensions/tedge_flows/src/flow.rs
Bravo555: fixed in 38cff8743b

crates/extensions/tedge_flows/src/flow.rs
Bravo555: fixed in 2675444

crates/extensions/tedge_flows/src/steps.rs
Bravo555: added in 21da79f.

Though there's not yet any builtin transformers that are shipped in the production configuration that could be used in system tests.

crates/extensions/tedge_flows/src/steps.rs
Bravo555: `on_message` now processes `on_startup` messages from previous steps, implemented in 0a187f3

crates/extensions/tedge_flows/src/runtime.rs
```diff
@@ -145,8 +145,24 @@ impl<Registry: FlowRegistryExt + Send> MessageProcessor<Registry> {
             .collect()
     }
 
-    pub async fn on_interval(&mut self, timestamp: SystemTime, now: Instant) -> Vec<FlowResult> {
+    /// Run any pending `on_startup` or `on_interval` callbacks.
+    pub async fn on_pending(&mut self, timestamp: SystemTime, now: Instant) -> Vec<FlowResult> {
```
didier-wenzek: I'm not convinced by this name change.

To me, this adds more confusion than help.

crates/extensions/tedge_flows/src/runtime.rs
```diff
@@ -145,8 +145,24 @@ impl<Registry: FlowRegistryExt + Send> MessageProcessor<Registry> {
...
+            .filter(|f| f.as_ref().steps.iter().any(|s| s.should_execute_startup()));
+
+        for flow in unstarted_flows {
+            // if there are any unstarted flows, run their on_startup now
+            let flow_output = flow
+                .as_mut()
+                .on_startup(&self.js_runtime, &mut self.stats, timestamp)
+                .await;
+            out_messages.push(flow_output);
+        }
```
didier-wenzek, line 162: Why not extract this in an `on_startup` method? Keeping the former `on_interval()` unchanged?

crates/extensions/tedge_flows/src/steps.rs
reubenmiller: > `on_message` now processes `on_startup` messages from previous steps, implemented in [0a187f3](https://github.com/thin-edge/thin-edge.io/commit/0a187f39d2a632fd778c559d2f4dbd1335e7392a)

This just caught me by surprise when testing our the PR.

I don't think that any messages that are produced by the onStartup message should be passed to downstream onMessage handlers as this breaks the fundamental control that the onMessage handler has. For instance the onMessage is designed to take in input from the flow definition file, e.g. `input.mqtt.topics`...if messages suddenly come from other steps, then it could easily break some scripts (javascript files).

Why can't the output of the onStartup() functions just be routed directly to the output? Or if the idea is to still be able to intercept / suppress onStartup messages from an upstream script, then I would think that changing the onStartup function signature to `onStartup(upstreamMessages, time, context)` (but tbh I don't think that is worth it and would be more confusion then anything).

For instance, I was trying to have a multi-step flow where each step would publish its own messages on startup.

**step1.js**
```js
export function onStartup(time, context) {
        return [{
                topic: "onstartup-health/step1",
                payload: JSON.stringify({text:'step 1 startup message', time: time.toISOString()}),
                mqtt: {retain:true},
        }, {
                topic: "te/device/main///e/startup",
                payload: JSON.stringify({text:'step 1 startup message'}),
        }]
}

        return [];
}
```

**step2.js**
```js
export function onStartup(time, context) {
        return [{
                topic: "onstartup-health/step2",
                payload: JSON.stringify({text:'step 2 startup message', time: time.toISOString()}),
                mqtt: {retain:true},
        }, {
                topic: "te/device/main///e/startup",
                payload: JSON.stringify({text:'step 2 startup message'}),
        }]
}

        return [];
}
```

And I was wondering why I was only receiving the retain message `onstartup-health/step2` from step 2 and not from step 1.

reubenmiller: Here is a test that I created on a branch to show what I as expecting in this situation.

https://github.com/reubenmiller/thin-edge.io/commit/5b09cba9704cf5796ff9b3dd8c9dd034862e1803

reubenmiller: However I guess if the onStartup message still go through the onMessage handlers of downstream steps, then it gives the flow a changes to apply things like `{ builtin = "limit-payload-size", config = { max_size = 16184 } },`

didier-wenzek: > I don't think that any messages that are produced by the onStartup message should be passed to downstream onMessage handlers as this breaks the fundamental control that the onMessage handler has.

From my point of view, this is the reverse. Not passing downstream the `onStartup` messages would break  the fundamental principle of flows and its mental model of a pipeline of transformations. 

> However I guess if the onStartup message still go through the onMessage handlers of downstream steps, then it gives the flow a changes to apply things like `{ builtin = "limit-payload-size", config = { max_size = 16184 } },`

This is why messages produced by `onStartup` *must* be passed to downstream to the `onMessage` handlers. If we want steps being easily combined into a pipeline of steps (aka a flow), then all the callbacks of a step (`onMessage`, `onStartup` and `onInterval`) must emit messages of the same shape.

As the example provided by this comment https://github.com/thin-edge/thin-edge.io/pull/4011#discussion_r2932780310 can indeed be a requirement, I would rather look for a solution that is at the flow level (something akin to a `tee`) and not an exception for the `onStartup`  callback.

A sketch of what I have in mind is the following:

```toml
[[steps]]
script = "step1.js"
tee.topic = "te/device/main///e/startup"
```

Such a solution, would also applies to all the callbacks giving also to the `onMessage` the ability to publish off-bound messages.

reubenmiller: Yeah that is a very good point. Thanks for clarifying it. So in summary, any message published by the onStartup() **MUST** be available to downstream onMessage handlers (which the PR is already doing)

crates/extensions/tedge_flows/src/runtime.rs
Bravo555: reverted to separate `on_startup` and `on_interval` in fe3263a7

crates/extensions/tedge_flows/src/runtime.rs
Bravo555: reverted to separate on_startup and on_interval in fe3263a

crates/core/tedge/src/cli/flows/test.rs
```diff
@@ -41,6 +41,9 @@ impl Command for TestCommand {
             None => TEdgeFlowsCli::load_flows(&self.flows_dir).await?,
             Some(flow) => TEdgeFlowsCli::load_file(&self.flows_dir, flow).await?,
         };
+        // we need a tick before calling on_message to run onStartup
+        self.tick(&mut processor, SystemTime::now(), Instant::now())
+            .await;
```
didier-wenzek: We can now directly call the `on_startup` method of the processor.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -256,7 +256,8 @@ impl FlowsMapper {
         Ok(())
     }
 
-    async fn on_interval(&mut self) -> Result<(), RuntimeError> {
+    /// Run any pending `on_startup` or `on_interval` handlers.
+    async fn on_pending(&mut self) -> Result<(), RuntimeError> {
```
didier-wenzek: I don't get why we still have an on-pending method for the actor.

Having a clear separation of the on-startup and on-interval callbacks; would make independent and clearer the timestamps controlling their invocation.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -277,6 +278,13 @@ impl FlowsMapper {
             }
             self.next_dump = now + self.config.stats_dump_interval;
         }
+
+        // on_startup handlers are scheduled to run on a tick, so check if we need to run any before on_interval, as
```
didier-wenzek: > // on_startup handlers are scheduled to run on a tick

This is the point I disagree. This conflates two independent aspects.

- Having no started timestamp is the criteria telling `o_startup` must be called.
- The next execution timestamp should only be used to tell when `on_interval` needs to be called

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -116,8 +140,17 @@ impl FlowStep {
     /// Initialize the next execution time for this script's interval
     /// Should be called after the script is loaded and interval is set
     fn init_next_execution(&mut self) {
+        if self.started_at.is_none() {
+            // execute onStartup when first loaded
+            self.next_execution = Some(Instant::now());
+            return;
+        }
```
didier-wenzek: Now that `on_startup` and `on_interval` are no more tangled, this should no more be required.


```suggestion
```

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -116,8 +140,17 @@ impl FlowStep {
...
+
         if !self.interval.is_zero() {
             self.next_execution = Some(Instant::now() + self.interval);
+        } else {
+            // next_execution may be still Some after startup
+            self.next_execution = None;
```
didier-wenzek: Now that `on_startup` and `on_interval` are no more tangled, the next execution will stay unset.


```suggestion
            self.next_execution = None;
```

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -179,6 +216,33 @@ impl FlowStep {
...
+        let output = match &mut self.handler {
+            StepHandler::JsScript(script, config) => script.on_startup(js, timestamp, config).await,
+            StepHandler::Transformer(_, builtin) => {
+                builtin.on_startup(timestamp, &js.context_handle())
+            }
+        };
+
+        // after onStartup is finished, make sure it's not run again and schedule onInterval
+        self.started_at = Some(Instant::now());
+        self.init_next_execution();
```
didier-wenzek: Now that `on_startup` and `on_interval` are no more tangled, this is useless.


```suggestion
```

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -179,6 +216,33 @@ impl FlowStep {
...
         }
     }
+
+    /// Trigger the onStartup function of the JS module.
+    ///
+    /// If the script has an onStartup, it should be executed only once, after it's loaded (or reloaded) but before any
+    /// other functions (onMessage and onInterval) (and before reporting the status of the flow?).
+    ///
+    /// Return zero, one or more messages
+    // TODO: if this should be executed only once, we should use the type system to ensure that it's impossible to call
```
didier-wenzek: This sounds to me complicated for little benefits.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
Bravo555: Made a PR fixing this issue: #4050

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -302,6 +302,79 @@ impl Flow {
...
+    async fn on_startup_steps(
+        &mut self,
+        js_runtime: &JsRuntime,
+        stats: &mut Counter,
+        timestamp: SystemTime,
+    ) -> Result<Vec<Message>, FlowError> {
+        let mut messages = vec![];
+
+        // if we call on_startup for any step, need to call it for all the ones after as well,
+        // as previous steps startups might set some stay we may want to read in next steps
```
albinsuresh:
```suggestion
        // as previous steps startups might set some state we may want to read in next steps
```

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -302,6 +302,79 @@ impl Flow {
         Ok(messages)
     }
 
+    /// Calls onStartup functions of steps of this flow that are starting.
+    ///
+    /// If the entire flow starts, we call onStartup of all steps, from the first to the last.
+    /// If some step got reloaded, we call onStartup for this step and all successive ones.
```
albinsuresh, line 308: I'm worried if this might come as a surprise to many users, who might expect `onStartup` function with some bootstrap logic to execute only once when it is loaded, but then sees it getting executed multiple times during the runtime when things around it changed. This makes it a poor candidate to capture any one-time bootstrap logic, unless it is idempotent. Not sure if that's the best contract. If it is, we'll have to document it clearly that their startup logic must be idempotent as it can be executed multiple times.

crates/extensions/tedge_flows/tests/interval.rs
```diff
@@ -458,6 +458,106 @@ async fn interval_executes_when_time_exceeds_interval() {
...
+            .map(|m| m.payload_str().unwrap_or_default().to_string())
+            .collect::<Vec<_>>()
+    };
+
+    // Give startup functions time to execute
+    tick(Duration::from_millis(50)).await;
+
+    let messages = get_messages();
+
+    // Verify onStartup runs for both steps and output from 1st is passed to on_message of 2nd
```
albinsuresh, line 611: I'm confused with that. The on_startup message from `step1` should have been passed to `step 2` only if the flow subscribed to that topic, right? Here flow is subscribing to `test/input` topic and `step1:onStartup` publishes to `test/flow/execution` and `step1:onMessage` publishes to `test/onstartup` and neither matches the topic filter that it has subscribed to.

Even if the input topic filter is corrected to `test/onstartup`, why would the message published to `test/flow/execution` be passed to `step2`?

tests/RobotFramework/tests/tedge_flows/onstartup/on-startup1.js
```diff
@@ -0,0 +1,14 @@
+export function onStartup(_time, context) {
+    context.mapper.set("on-startup.js", `onstartup on_startup 1`)
```
albinsuresh, line 2:
```suggestion
    context.mapper.set("on-startup1.js", `onstartup on_startup 1`)
```
Unless the intention was to replace the same key in `step1` from `step2`.

tests/RobotFramework/tests/tedge_flows/onstartup/on-startup2.js
```diff
@@ -0,0 +1,13 @@
+export function onStartup(_time, context) {
+    context.mapper.set("on-startup.js", `onstartup on_startup 2`)
```
albinsuresh, line 2:
```suggestion
    context.mapper.set("on-startup2.js", `onstartup on_startup 2`)
```
Unless the intention was to replace the same key in `step1` from `step2`

crates/extensions/tedge_flows/src/flow.rs
didier-wenzek: @albinsuresh is correct. The `onStartup` callback must be called only once.

However, the problem is not in this method, which focus is on message transformations along the pipeline.

The fix should be in `Step::on_startup` with a guard on step started_at timestamp

```rust
    pub async fn on_startup(
        &mut self,
        js: &JsRuntime,
        timestamp: SystemTime,
    if self.started_at.is_some() {
        return OK(vec![])
    }
    
    ...
}
```

That said, the comment can be improved. I suggest the following:

```
    /// Calls the onStartup callback of all steps freshly reloaded
    ///
    /// If some step produces onStartup messages, these messages are processed by all the subsequent steps.
```

crates/extensions/tedge_flows/tests/interval.rs
didier-wenzek: > The on_startup message from `step1` should have been passed to `step 2` only if the flow subscribed to that topic, right? Here flow is subscribing to `test/input` topic and `step1:onStartup` publishes to `test/flow/execution` and `step1:onMessage` publishes to `test/onstartup` and neither matches the topic filter that it has subscribed to.

The input topic of a flow is only about the message *source*. Once a message has been received from that source or produced by an `onStartup` function, this message flows through the pipeline being transformed by the subsequent steps.

There is (at least currently) no topic filters ruling out which messages is passed to a step.

I feel that your source of confusion is related to that comment https://github.com/thin-edge/thin-edge.io/pull/4011#discussion_r2932780310

Are you looking for a way to push messages directly to the output, bypassing downstream transformations?

crates/extensions/tedge_flows/src/flow.rs
Bravo555: Makes sense. My understanding before was that the startup function very much should be written in an idempotent way, because once a script from step 1 is changed to e.g. write value `B` instead of `A` into a context map, then step 2 might want to in turn reflect that in its own startup. But indeed `on_message` can easily differentiate between messages from input topics and from `on_startup`: just use a different topic for startup messages and then compare the topic in `on_message`. Then it's up to the script author whether this is a change they want to react to, or not. I guess I was thinking that having to handle `on_startup` stuff in next steps' `on_message` handlers was a bit ugly, mixing in the normal case and the startup case in a single function, so startup logic doesn't live only in `on_startup`, but parts of it also live in `on_message` of next steps (whereas with `on_startup` propagating it was more "tidy"), but indeed handling it using message flow is much more flexible and well worth the price.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -83,6 +83,7 @@ impl Actor for FlowsMapper {
     }
 
     async fn run(mut self) -> Result<(), RuntimeError> {
+        self.on_startup().await?;
         self.send_updated_subscriptions().await?;
         self.notify_flows_status().await?;
```
didier-wenzek: This can be discussed, but from my point of view, it would be better to first notify the status of all the flows and only then to start sending messages.


```suggestion
        self.send_updated_subscriptions().await?;
        self.notify_flows_status().await?;
        self.on_startup().await?;
```

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -277,13 +278,25 @@ impl FlowsMapper {
             }
             self.next_dump = now + self.config.stats_dump_interval;
         }
+
+        // on_startup handlers are scheduled to run on a tick, so check if we need to run any before on_interval, as
+        // startup should always run first
```
didier-wenzek: This comment can be removed now.


```suggestion
```

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -529,10 +542,12 @@ impl FlowsMapper {
     async fn on_file_updated(&mut self, path: &Utf8Path) -> Result<(), RuntimeError> {
         if matches!(path.extension(), Some("js" | "ts" | "mjs")) {
             let reloaded_flows = self.processor.reload_script(path).await;
+            self.on_startup().await?;
```
didier-wenzek: Instead of repeated ourselves telling to invoke the `onStartup` callbacks when a flow, a script or a directory has been added, I suggest to invoke these callbacks in a single place: the `run()` method of the actor after an inotify event (already suggested here https://github.com/thin-edge/thin-edge.io/pull/4011#pullrequestreview-3944324208).

We don't have performance issue (flow & script updates are infrequent events compared to message transformations). So we don't care if this is called when a script is removed. And the main focus should be maintainability: having a single invocation would help a lot.

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -39,15 +44,17 @@ impl FlowStep {
         FlowStep {
             handler: StepHandler::JsScript(script, config),
             interval: Duration::ZERO,
-            next_execution: None,
+            next_execution: Some(Instant::now()),
```
didier-wenzek: This will be properly initialized by `init_next_execution`.


```suggestion
            next_execution: None,
```

crates/extensions/tedge_flows/src/steps.rs
```diff
@@ -39,15 +44,17 @@ impl FlowStep {
...
+            started_at: None,
         }
     }
 
     pub fn new_transformer(instance_name: String, transformer: Box<dyn Transformer>) -> Self {
         FlowStep {
             handler: StepHandler::Transformer(instance_name, transformer),
             interval: Duration::ZERO,
-            next_execution: None,
+            next_execution: Some(Instant::now()),
```
didier-wenzek: This will be properly initialized by `init_next_execution`.


```suggestion
            next_execution: None,
```

crates/extensions/tedge_flows/src/flow.rs
Bravo555: changed the function impl and comment in b392861

tests/RobotFramework/tests/tedge_flows/onstartup/on-startup1.js
Bravo555: here just wanted to check if we can write and read from the context, so can be left as is

tests/RobotFramework/tests/tedge_flows/onstartup/on-startup2.js
Bravo555: here just wanted to check if we can write and read from the context, so can be left as is

crates/extensions/tedge_flows/src/actor.rs
Bravo555: fixed in 14921667c9

crates/extensions/tedge_flows/src/actor.rs
Bravo555: fixed in 14921667c9

crates/extensions/tedge_flows/src/actor.rs
Bravo555: fixed in 14921667c9

crates/extensions/tedge_flows/src/steps.rs
Bravo555: fixed in 14921667c9

crates/extensions/tedge_flows/src/steps.rs
Bravo555: fixed in 14921667c9

crates/extensions/tedge_flows/src/steps.rs
Bravo555: fixed in bd90ae6618

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -561,6 +574,7 @@ impl FlowsMapper {
     async fn on_directory_updated(&mut self, path: &Utf8Path) -> Result<(), RuntimeError> {
         self.processor.load_all_flows_from_dir(path).await;
 
+        self.on_startup().await?;
```
didier-wenzek: This is now useless

```suggestion
```

crates/extensions/tedge_flows/tests/interval.rs
albinsuresh: Yes, I was expecting the `onStartup` message to produce messages published to a different topic that's not necessarily the the same message type that the flow is expecting. For example, I wanted all the flows to emit a startup event message when they are started up. The goal was to use these events as a confirmation of them being ready, when they are installed from the cloud. So, I'd have added a simple `startup_event` step at the beginning of all my flows as follows:

```measuremensts.toml
input.mqtt.topics = ["custom/measurement/+"]

steps = [
    { script = "startup_event.js" config = { flow_name = "measurements" }
    { builtin = "add-timestamp", config = { format = "${.params.time.format}"} },
    { script = "to_te_measurement.js" }
]
```

The script would be a simple one as follows:

```
export function onStartup(_time, context) {
    let flow_name = context.config.get("flow_name")
    let payload = { "flow": flow_name }
    return { topic: "te/device/main///e/flow_ready", payload: payload }
}
```

I'll add the same `startup_status.js` as the first step of every other flow as well, esp the ones that I install from the cloud.

With the current design, where the "starup message" needs to flow through the pipeline, I'm forced to update each and every step of every single flow to accommodate these event messages as well (so that they pass-through and not get filtered).

The expectation for these event messages to go through the `limit-payload-size` filter in the built-in `events.toml` flow is perfectly valid. But, I would have expected the "startup event" emitted by all flows to go through that single `events.toml` flow to go through that filtering phase rather than it being passed through each and every flow from which they were emitted from.

didier-wenzek: > But, I would have expected the "startup event" emitted by all flows to go through that single `events.toml` flow to go through that filtering phase rather than it being passed through each and every flow from which they were emitted from.

My approach is the following. *All* the messages produced by one step of the flow are processed by the `onMessage` callbacks of the steps downstream.

Your expectation is to have *some* exceptions to that rule. It's then difficult to say what is an exception what is not. You propose to treat differently all the `onStartup` messages, but one user might have the reverse expectation (for instance to trigger a message for the main device as first step of a registration flow knowing that  nothing is published for the main device on `te/main/device//`).

So exception to the main rule should be configurable by the users. It see two approaches:

- Add a topic filter to each step, so only messages matching that filter will be actually processed by the step the others being pushed downstream unchanged. By default this would the match-all topic filter `["#']`.
- Add a topic shortcut to each step, so messages matching that topic are directly pushed to the flow output bypassing all the downstream steps. By default this would the match-nothing topic filter `[]`.

> For example, I wanted all the flows to emit a startup event message when they are started up.

Actually this can be implemented with the current design.

The key point is to emit this startup message from the *last* step of the flow:

```
input.mqtt.topics = ["custom/measurement/+"]

steps = [
    { builtin = "add-timestamp", config = { format = "${.params.time.format}"} },
    { script = "to_te_measurement.js" },
    { script = "startup_event.js" config = { flow_name = "measurement" } }
]
```

And then to add an `onMessage` that simply push downstream all the messages produced by the flow:

```
export function onMessage(message) {
     return message
}

export function onStartup(_time, context) {
    let flow_name = context.config.get("flow_name")
    let payload = { "flow": flow_name }
    return { topic: "te/device/main///e/flow_ready", payload: payload }
}
```

albinsuresh: > Your expectation is to have some exceptions to that rule. It's then difficult to say what is an exception what is not. You propose to treat differently all the onStartup messages, but one user might have the reverse     expectation (for instance to trigger a message for the main device as first step of a registration flow knowing  that nothing is published for the main device on te/main/device//).

Yeah, I agree. Both expectations are perfectly valid from a user's PoV and restricting to just one behaviour would be wrong. We should try to support both, if possible.
 
> So exception to the main rule should be configurable by the users. It see two approaches:
> 
> * Add a topic filter to each step, so only messages matching that filter will be actually processed by the step the others being pushed downstream unchanged. By default this would the match-all topic filter `["#']`.

I like this proposal, as it makes it explicit what each step's expectation is, especially in flows accepting multiple topics. The only concern that I have with that solution is whether the user would get confused when he sees a topic filter for a step and interpret that as a "filtering topic" rather than expecting the non-matching messages it to pass through (especially if the step itself is a filtering step).

> * Add a topic shortcut to each step, so messages matching that topic are directly pushed to the flow output bypassing all the downstream steps. By default this would the match-nothing topic filter `[]`.

That sounds like the simplest solution where I directly declare on the "startup step" whether the output messages should passthrough the flow or bypass it.

> Actually this can be implemented with the current design.
> The key point is to emit this startup message from the last step of the flow:

That might be a functional solution, but I'm not fully convinced with that. Keeping "startup step" as the last step of a flow feels counter-intuitive, as a user's natural expectation for any "startup step" would be to see it at the beginning.

didier-wenzek: >  a user's natural expectation for any "startup step" would be to see it at the beginning.

I don't share the same feeling. To me startup rhythms with  beginning of time not beginning of the pipeline.

Issue #3979, opened 2026-02-12
flows engine doesn't detect flows when an entire directory is moved within the flows engine directory


reubenmiller:
**Describe the bug**

**Note** This is probably more a feature request then a bug.

Flows aren't auto detected if the flows are contained in a nested folder and the parent folder is moved within the flow engine's flows directory.

For example,

```sh
mv /tmp/myflow /etc/tedge/mappers/flows/flows/
```

This is partially related to the flow improvements PR https://github.com/thin-edge/thin-edge.io/pull/3953, but also to align with the new flow packaging proposal which packages flows as a folder.

The root cause looks to be related to this section of code where the flow reloading logic is only triggered when receiving FSWatchEvent::Modified and FSWatchEvent::FileDeleted (which seem to be file related), and don't trigger when a new directory is added which DirectoryCreated event, or DirectoryDeleted event when a directory is deleted, as these operations don't send notifications for any files under the directory structure.

[InputMessage::FsWatchEvent(FsWatchEvent::Modified(path)) => {](https://github.com/reubenmiller/thin-edge.io/blob/c866c87412fd699115b76b5122a6f7553beb2f87/crates/extensions/tedge_flows/src/actor.rs#L77-L101)

**To Reproduce**

1. Start the tedge-flows (with no flows loaded)
2. Create the following flow definition under /tmp/myflow

    The following commands will create the folder and files with the desired contents:

    ```sh
    mkdir -p /tmp/myflow
    cat <<EOT > /tmp/myflow/flow.toml
    input.mqtt.topics = ["dummy"]
    steps = [
        { script = "main.js", interval = "1s" },
    ]
    EOT

    cat <<EOT > /tmp/myflow/main.js
    export function onMessage(message, context) {
        return [];
    }

    export function onInterval(time, context) {
        console.debug("myflow: onInterval called");
        return [];
    }
    EOT
    ```

3. Move the flow into the desired mapper's flows folder (here we're using the generic tedge-flows mapper)

    ```sh
    mv /tmp/myflow /etc/tedge/mappers/flows/flows/
    cp -Ra /tmp/myflow /etc/tedge/mappers/flows/flows/
    ```

4. Monitor the tedge-flows service for a sign that the myflow/flow.toml was detected

    ```sh
    journalctl -fu tedge-flows -n 100
    ```

5. You can repeat the test by creating the flow folder, and then moving the individual files, then the tedge-flows service detects this properly.

    ```sh
[...]

PR #4013, opened 2026-02-26
fix: flows are removed when directory is moved out of flows dir

Bravo555:
## TODO

- [x] handle scenario where we move out of flows directory when the flows process was started with a flow already inside
- [x] auto-refresh nested flows
- [x] update tests
- [ ] ~~move logic to `MessageProcessor`~~ will be done in a follow-up refactor PR

## Proposed changes

Flows are now properly removed when a directory is moved out of flows dir, and also properly added and removed when moving flows which were loaded at startup, and when moving a directories containing scripts a flow uses.

## Types of changes

## Paste Link to the issue
#3979 

## Checklist


tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -255,9 +255,14 @@ Flow is discovered when a directory is moved
...
-    ${start}    Get Unix Timestamp
     Execute Command    mv /tmp/myflow /etc/tedge/mappers/local/flows
-    Should Have MQTT Messages    topic=myflow    message_contains=myflow    date_from=${start}
+    ${output}    Execute Command    tedge flows test test/nested-flows/myflow 'flow started after moving directory into flows dir'
+    Should Contain    ${output}    flow started after moving directory into flows dir
+
+    # now moving the directory again should stop the flow
+    Execute Command    mv /etc/tedge/mappers/local/flows/myflow /tmp/myflow
+    ${output}    Execute Command    tedge flows test test/nested-flows/myflow 'flow removed after moving directory out of flows dir'
+    Should Not Contain    ${output}    flow removed after moving directory out of flows dir
```
didier-wenzek: This test doesn't exercise the inotify part of the code added by this PR. Only `tedge-mapper local` is monitoring file changes; while `tedge flows test` use a snapshot of the flows.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -49,6 +50,11 @@ pub struct FlowsMapper {
     watched_commands: HashSet<Utf8PathBuf>,
     processor: MessageProcessor<ConnectedFlowRegistry>,
     next_dump: Instant,
+    /// Paths to flow declaration and script files that are currently in use.
+    ///
+    /// When a directory containing flows is moved from the flows directory, we only get an fs event
+    /// containing the name of the directory, so we need to remember what files were loaded so we can unload them.
+    loaded_files: HashSet<Arc<Utf8Path>>,
```
didier-wenzek: To be consistent with existing code.


```suggestion
    loaded_files: HashSet<Utf8PathBuf>,
```

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -495,10 +515,12 @@ impl FlowsMapper {
 
     async fn on_file_updated(&mut self, path: &Utf8Path) -> Result<(), RuntimeError> {
         if matches!(path.extension(), Some("js" | "ts" | "mjs")) {
+            self.loaded_files.insert(path.into());
```
didier-wenzek: As this is done whatever the file extension, I would do that unconditionally.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -508,8 +530,10 @@ impl FlowsMapper {
 
     async fn on_file_removed(&mut self, path: &Utf8Path) -> Result<(), RuntimeError> {
         if matches!(path.extension(), Some("js" | "ts" | "mjs")) {
+            self.loaded_files.remove(path);
```
didier-wenzek: As this is done whatever the file extension, I would do that unconditionally.

crates/extensions/tedge_flows/src/actor.rs
Bravo555: This isn't done if the file is not javascript and not toml.

crates/extensions/tedge_flows/src/actor.rs
Bravo555: fixed in dff64fd

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
Bravo555: fixed in dff64fd

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -453,46 +454,92 @@ impl FlowsMapper {
...
+        removed_scripts.sort_unstable();
+        removed_scripts.dedup();
+
+        removed_flows.extend(
+            self.processor
+                .registry
+                .flows()
+                .map(|f| f.source_path().to_path_buf())
+                .filter(|p| p.starts_with(path)),
+        );
```
Bravo555: thought: since to properly implement the feature we need to query the registry from the actor, and then call the registry's reload/add/remove methods, perhaps it would be an improvement for some future refactoring PR to move the logic there: the actor would be calling registry's method to mark a directory as dirty, and then the registry itself would read dirty directories to find out what changed.

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -163,7 +164,11 @@ impl<T: FlowRegistry + Send> FlowRegistryExt for T {
 
     async fn load_all_flows(&mut self, js_runtime: &mut JsRuntime) {
         let config_dir = self.config_dir().to_owned();
-        let (loaded_flows, unloaded_flows) = FlowConfig::load_all_flows(&config_dir).await;
+        self.load_all_flows_from_dir(&config_dir, js_runtime).await
```
Bravo555, line 167: thought: also regarding paths - found a nice article about `openat` syscall for reducing need to manually craft paths, avoid TOCTOU issues or cleaning up paths like `../../../` (I see that with scripts in flows, we support absolute paths, so no issues there), etc. I wonder if there's anywhere in thin-edge where there would be value in using that approach.

https://val.packett.cool/blog/use-openat/

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -453,46 +454,85 @@ impl FlowsMapper {
...
+
+        removed_flows.sort_unstable();
+        removed_flows.dedup();
```
didier-wenzek: A better place for this code is the registry.

The current code exposes too many internal details, and doing so, hides a very subtle point: a flow can be listed here despite being stored in a directory unrelated to `path` but because using a script stored in that directory. __It would be good to document this subtlety__.

```suggestion
        // from all the flows, we need to select scripts used by these flows that are prefixed by the path
        let (removed_flows, removed_scripts)= self
            .processor
            .registry
            loaded_files(path);
```

Also. I would consider to use `HashSet` to make explicit the fact we want to remove duplicates.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -453,46 +454,85 @@ impl FlowsMapper {
...
+
+        trace!(?removed_flows, ?removed_scripts);
```
didier-wenzek: Is this useful, once you are convinced your code works as expected?

```suggestion
```

Bravo555: Added this because while we do log whenever flows are reloaded, we don't log when scripts are reloaded/removed. So I'll remove this log, but should we add add an `INFO` log for where scripts are reloaded/removed?

didier-wenzek: > So I'll remove this log, but should we add add an INFO log for where scripts are reloaded/removed?

Yes, why not. Better than a trace with no more info than two file names ;-)

crates/extensions/tedge_flows/src/actor.rs
Bravo555: decided during a 1:1 call to leave this refactor for a follow-up refactor PR

crates/extensions/tedge_flows/src/actor.rs
Bravo555: removed in 1518431

Instead of now looking where to place this new user-facing log to make the most sense, decided to just remove it for now and consider adding it back in a follow-up PR.

Issue #3989, opened 2026-02-16
Allow externalising the configuration used by a flow


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When updating an existing flow, the user does not want to lose any customized user config.

**Describe the solution you'd like**

Add support for an externalized "params.toml" file which will be loaded when the flow is loaded which is automatically merged into the flow definition so that it can be used to control the flows configuration.

* When installing a flow via the sm-plugin, the "params.toml" should not be overwritten

An example of the params.toml file is show in the [flow software management plugin design](https://github.com/thin-edge/tedge-oscar/blob/main/docs/FLOWS_PACKAGING.md#flow-packaging)

**Describe alternatives you've considered**

**Additional context**

PR #4015, opened 2026-02-27
feat: Externalising the configuration used by a flow

didier-wenzek:
## Proposed changes

- [x] The `params.toml` file (if any) is loaded and used to substitute user defined values for placeholders in the flow config.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3989

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/params.rs
+++ b/crates/extensions/tedge_flows/src/params.rs
@@ -41,7 +41,11 @@ impl Params {
         let Some(directory) = path.parent() else {
             return Ok(Params::default());
         };
-        Self::load_params(directory).await
+        // To avoid breaking flows, an ill-formed params.toml file is read as empty set of params
+        Self::load_params(directory)
+            .await
+            .map_err(|err| log::warn!("{err:?}"))
+            .or(Ok(Params::default()))
     }
 
     /// Load the `params.toml` used by all the flows of the given directory
@@ -78,6 +82,7 @@ impl Params {
     /// Substitute all the path expressions of the input value
     pub fn substitute(&self, value: &Value) -> Result<Value, LoadError> {
         match value {
+            // To avoid breaking flows, a null value is substituted for an ill-formed params path
             Value::String(expr) => self
                 .substitute_path(expr)
                 .map_err(|err| log::warn!("{err:?}"))
--- a/tests/RobotFramework/tests/tedge_flows/greeting-flows/broken-params.toml
+++ b/tests/RobotFramework/tests/tedge_flows/greeting-flows/broken-params.toml
@@ -1,0 +1,1 @@
+hello.message = "Have a nice day!"
--- a/tests/RobotFramework/tests/tedge_flows/greeting-flows/hello.js
+++ b/tests/RobotFramework/tests/tedge_flows/greeting-flows/hello.js
@@ -1,0 +1,6 @@
+export function onMessage(message, context) {
+    const greeting = context.config.greeting ?? "Hello World!"
+    message.payload = greeting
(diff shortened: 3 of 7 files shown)


crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -486,7 +470,11 @@ impl FlowsMapper {
                 let Ok(path) = Utf8PathBuf::try_from(path) else {
                     return Ok(());
                 };
-                self.on_file_removed(path.as_path()).await?;
+                if Params::is_params_file(&path) {
+                    self.on_params_updated(&path).await?;
```
albinsuresh, line 474: When a parameter file is deleted, wouldn't it be better to disable any flows using those parameters so that the user immediately notices that he has messed up?

crates/extensions/tedge_flows/src/params.rs
```diff
@@ -0,0 +1,220 @@
...
+/// debug = true
+/// ```
+///
+/// A flow definition can reference a parameter value using the template syntax
+///
+/// ```toml
+/// [[steps]]
+/// script = "main.js"
+/// config.y = "{.params.x.y}"
+/// config.debug = "{.params.debug}"
```
albinsuresh:
```suggestion
/// config.y = "${.params.x.y}"
/// config.debug = "${.params.debug}"
```

crates/extensions/tedge_flows/src/params.rs
```diff
@@ -0,0 +1,220 @@
...
+                Ok(value) => Ok((key.to_string(), value)),
+                Err(err) => Err(err),
+            })
+            .collect()
+    }
+
+    /// Substitute all the path expressions of the input value
+    pub fn substitute(&self, value: &Value) -> Result<Value, LoadError> {
+        match value {
+            Value::String(expr) => self.substitute_path(expr).or(Ok(Value::Null)),
```
albinsuresh: When it' an unknown parameter(could even be a typo) wouldn't it be better to return `UnknownParam` instead of defaulting to `Value::Null`?

reubenmiller: Though these are the values being pulled into the flow config, so if they are missing, then giving then a null value mean you can set default values from within the javascript code. e.g.

```js
   // try reading from the config, but use a default if it is nullish (null or undefined)
   const foo = context.config.foo ?? "mydefault";
}
```

Yes there is still the case where we could improve the user experience to print a warning that they are setting a param that isn't referenced anywhere, however we think we can do this as a second step as it gets a bit complex if the params.toml is potentially used by multiple flows as you'd have to scan all the flows

crates/extensions/tedge_flows/src/actor.rs
reubenmiller: No, a flow should be able to function without the user customized values. Defaults can (and should) be set in the code.

crates/extensions/tedge_flows/src/params.rs
didier-wenzek: fixed: https://github.com/thin-edge/thin-edge.io/pull/4015/changes/d5de7314f2998e3c2472a4442b4fda59d8e529db

crates/extensions/tedge_flows/src/params.rs
didier-wenzek: I added a warning when no value can be substituted for the expression https://github.com/thin-edge/thin-edge.io/pull/4015/changes/d5de7314f2998e3c2472a4442b4fda59d8e529db

Issue #3992, opened 2026-02-16
tedge flows handles mapper specific built-in functions


reubenmiller:
**Is your refactoring request related to a problem? Please describe.**

Since tedge-mapper-c8y now uses flows to provide the cloud mappings, it would be useful for users to be able to test out mapping using the `tedge flows test` and `tedge flows list` cli commands.

**Describe the solution you'd like**

Depending on the time available, either one of the options is to be implemented:

* Option 1: Print out a warning that Cumulocity mapper specific built-in functions aren't available to be tested via the `tedge flows test` command

* Option 2: Support loading the built-in mapper specific commands

**Describe alternatives you've considered**

**Additional context**

PR #4020, opened 2026-03-03
fix: tedge flows cli loads mapper specific builtin transformers

didier-wenzek:
## Proposed changes

- [x] `tedge flows` loads c8y specific builtin transformers
- [x] aws and az specific builtin transformers
- [x] builtin transformers are available to all mappers.
- [x] `tedge flows test --context <path-to-or-inlined-json>` initializes the tests context with a JSON object
- [ ] Documentation for all the builtin transformers 

This raises several questions:

- [x] Do we have to load all the builtin transformers, even when the flow under test is only related to one of the mappers?
    - Yes, all the builtin transformers are always available, whatever the mapper, cloud-specific or local.
- [x] Can we simply make all the transformers, available to any mapper, being cloud specific or local?
    - Yes, so there is no prerequisite on running a specific mapper to run a transformer.
    - Note there is no impact on `tedge` binary size as this executable already includes all the daemons and mapper variants.
- [x] What about creating the builtin flows? For all the mapper?
   - No, the builtin flows are only created by their respective mapper.
   - We want to avoid the creation of flow definitions that are never used. 
- [ ] What about transformers expecting a configuration that is mapper specific (e.g. the cache which has to be aware of the MQTT topics for birth messages)?
- [ ] What about the transformers that expect the context to be populated? Currently, the c8y mapper populates the context using a very c8y specific piece of code: the entity cache. Should this cache be promoted to a builtin transformer? 

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3992

## Checklist

### DIFF
--- a/docs/src/references/mappers/flows.md
+++ b/docs/src/references/mappers/flows.md
@@ -481,3 +481,101 @@ Store a message in the [mapper context](#context) shared by all the flows and tr
 - An `update-context` step can be given a topic filter used to store only a subset of the messages.
   - `{ builtin = "update-context", config.topics = "te/+/+/+/+/m/+/meta" }`
   - If a message doesn't match the configured topic, this message is passed unchanged to the subsequent transformation steps.
+
+### `into-c8y-measurements`
+
+Transform a [%%te%% measurement](../../../understand/thin-edge-json/#measurements) into a [Cumulocity measurement](../c8y-mapper/#measurement)
+
+- This transformer uses the [mapper context](#context) to retrieve metadata about the message sources.
+  - In the case of the Cumulocity mapper, this context is populated by the mapper.
+  - However, in the case of a user-defined mapper, this context has to populated by a flow consuming MQTT messages
+    or reading a configuration file. This is done by inserting [registration messages](../mqtt-api/#entity-registration)
+    using the [entity topic identifiers](../mqtt-api/#group-identifier) as keys.
+  - ```
+    context.mapper.set("device/child-xyz//", {
+        "@id":"raspberry-007-child-xyz",
+        "@type":"child-device",
+        "@topic-id":"device/child-xyz//",
+        "@name":"child-xyz",
+        "@type-name":"thin-edge"
+     })
+    ```
+- If the source that is sending alarms is not registered yet in the mapper context,
+  then this transformer discards the alarm. If this is not the desired behavior an approach is to add upstream a step
+  using the [`cache-early-messages`](#cache-early-messages) transformer.
+- Even if designed after Cumulocity requirements, this transformer can be used by any mappers.
+
+### `into-c8y-events`
+
+Transform a [%%te%% event](../../../understand/thin-edge-json/#events) into a [Cumulocity event](../c8y-mapper/#events)
+
+- This transformer uses the [mapper context](#context) to retrieve metadata about the message sources.
+  - In the case of the Cumulocity mapper, this context is populated by the mapper.
+  - However, in the case of a user-defined mapper, this context has to populated by a flow consuming MQTT messages
+    or reading a configuration file. This is done by inserting [registration messages](../mqtt-api/#entity-registration)
+    using the [entity topic identifiers](../mqtt-api/#group-identifier) as keys.
+  - ```
+    context.mapper.set("device/child-xyz//", {
+        "@id":"raspberry-007-child-xyz",
+        "@type":"child-device",
+        "@topic-id":"device/child-xyz//",
+        "@name":"child-xyz",
+        "@type-name":"thin-edge"
+     })
+    ```
+- If the source that is sending alarms is not registered yet in the mapper context,
+  then this transformer discards the alarm. If this is not the desired behavior an approach is to add upstream a step
+  using the [`cache-early-messages`](#cache-early-messages) transformer.
+- Even if designed after Cumulocity requirements, this transformer can be used by any mappers.
+
+### `into-c8y-alarms`
+
+Transform a [%%te%% alarm](../../../understand/thin-edge-json/#alarms) into a [Cumulocity alarm](../c8y-mapper/#alarms)
+
+- This transformer uses the [mapper context](#context) to retrieve metadata about the message sources.
+  - In the case of the Cumulocity mapper, this context is populated by the mapper.
+  - However, in the case of a user-defined mapper, this context has to populated by a flow consuming MQTT messages
+    or reading a configuration file. This is done by inserting [registration messages](../mqtt-api/#entity-registration)
+    using the [entity topic identifiers](../mqtt-api/#group-identifier) as keys.
+  - ```
+    context.mapper.set("device/child-xyz//", {
+        "@id":"raspberry-007-child-xyz",
+        "@type":"child-device",
+        "@topic-id":"device/child-xyz//",
+        "@name":"child-xyz",
+        "@type-name":"thin-edge"
+     })
+    ```
+- If the source that is sending alarms is not registered yet in the mapper context,
+  then this transformer discards the alarm. If this is not the desired behavior an approach is to add upstream a step
+  using the [`cache-early-messages`](#cache-early-messages) transformer.
+- Even if designed after Cumulocity requirements, this transformer can be used by any mappers.
+
+### `cache-early-messages`
+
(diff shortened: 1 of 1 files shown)


crates/core/tedge_mapper/src/lib.rs
```diff
@@ -195,3 +201,24 @@ pub(crate) fn flows_config(
     );
     Ok(flows_config)
 }
+
+pub fn load_builtin_transformers(flows: &mut impl FlowRegistryExt) {
+    c8y_mapper_ext::load_builtin_transformers(flows);
+    az_mapper_ext::load_builtin_transformers(flows);
+    aws_mapper_ext::load_builtin_transformers(flows);
```
Bravo555, line 209: todo: as it is now, it breaks conditional compilation (e.g. `cargo check -p tedge-mapper --no-default-features --features aws` for only aws mapper), but should be easy to fix with `#[cfg(...)]` attributes. What's interesting though is we're not testing that configuration in CI, so all the tests pass.

crates/extensions/az_mapper_ext/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "az_mapper_ext"
-description = "thin-edge aws extension adding support for azure data model"
+description = "thin-edge extension adding support for azure data model"
```
Bravo555, line 3: nice catch

tests/RobotFramework/tests/tedge_flows/builtin-transformers.robot
```diff
@@ -0,0 +1,69 @@
...
+Install Flow
+    [Arguments]    ${directory}
+    ${start}    Get Unix Timestamp
+    Execute Command    sleep 0.1
+    ThinEdgeIO.Transfer To Device    ${CURDIR}/${directory}/*    /etc/tedge/mappers/local/flows/${directory}/
+    Execute Command    ls -lh /etc/tedge/mappers/local/flows/${directory}
+    Should Have MQTT Messages
+    ...    topic=te/device/main/service/tedge-mapper-local/status/flows
+    ...    date_from=${start}
+    ...    message_contains=${directory}
```
Bravo555, line 96: suggestion: very similar to `Install Flow` in main `tedge_flows` suite, we could extract it to a shared resource file

tests/RobotFramework/tests/tedge_flows/builtin-transformers.robot
```diff
@@ -29,6 +29,27 @@ Test c8y specific transformers with tedge flows cli
     ...    ${expected_msg}
     [Teardown]    Uninstall Flow    custom-measurements
 
+Test c8y specific transformers with tedge flows cli and initial context
+    Install Flow    custom-measurements
+    ${transformed_msg}    Execute Command
+    ...    tedge flows test te/device/main///m/environment '{"temperature": 258}' --context '{ "device/main//": {"@id":"raspberry-007","@type":"device","@topic-id":"device/main//","@name":"raspberry-007","@type-name":"thin-edge"} }'
```
reubenmiller, line 35: Just for clarity, which scope is the context added to on the JavaScript level? I was half expecting to see either a `flow`, `script`, or `mapper` in the keys

didier-wenzek: The runtime can only inject values in the `mapper` context. So this is implied.

reubenmiller: I'm having a problem getting the test code to work for the c8y mapper and a child device.

I'm re-using the context.json you have in the test files.

```
# works
$ tedge flows test --mapper c8y te/device/main///m/environment '{"temperature": 42}' --context ./context.json
[c8y/measurement/measurements/create] {"temperature":{"temperature":{"value":42.0}},"time":"2026-03-19T15:44:53.991Z","type":"environment"}

# no output
$ tedge flows test --mapper c8y te/device/child-xyz///m/environment '{"temperature": 42}' --context ./context.json
```

I'm assuming something is just missing from the context object which is required when mapping data for a child device, but I'm not sure what.

This is the context.json that I'm using (copied from the test)

```sh
$ cat context.json 
{
  "device/main//": {
    "@id":"raspberry-007",
    "@type":"device",
    "@topic-id":"device/main//",
    "@name":"raspberry-007",
    "@type-name":"thin-edge"
  },
  "device/child-xyz//": {
    "@id":"raspberry-007-child-xyz",
    "@type":"child-device",
    "@topic-id":"device/child-xyz//",
    "@name":"child-xyz",
    "@type-name":"thin-edge"
  }
}
```

reubenmiller: After doing some more digging, it seems the `cache-early-messages` is the step which is causing the messages for child entities to be discarded (in this case probably cached).

didier-wenzek: > After doing some more digging, it seems the `cache-early-messages` is the step which is causing the messages for child entities to be discarded (in this case probably cached).

Oh yeah. The  `cache-early-messages` awaits for a child birth message before releasing the measurements to the next step. => We really need to document all these builtin-transformers.

docs/src/references/mappers/flows.md
```diff
@@ -481,3 +481,55 @@ Store a message in the [mapper context](#context) shared by all the flows and tr
 - An `update-context` step can be given a topic filter used to store only a subset of the messages.
   - `{ builtin = "update-context", config.topics = "te/+/+/+/+/m/+/meta" }`
   - If a message doesn't match the configured topic, this message is passed unchanged to the subsequent transformation steps.
+
+### `into-c8y-alarms`
```
didier-wenzek, line 531: Starting with alarms. I will add sections for measurements and events, after a first series of comments.

docs/src/references/mappers/flows.md
```diff
@@ -481,3 +481,55 @@ Store a message in the [mapper context](#context) shared by all the flows and tr
...
+config = { mapper_topic_id = "device/main/service/tedge-mapper-c8y" }
+
+[[steps]]
+builtin = "into-c8y-measurements"
+```
+
+- The default for the Cumulocity mapper is to publish birth messages on `te/device/main/service/tedge-mapper-c8y/status/entities`
+- This is configurable. A custom mapper (which topic id is `a/b/c/d`) is expected to publish finalized entity registration
+  on `te/a/b/c/d/status/entities`.
+- A birth message is composed of
```
didier-wenzek, line 577: Documenting the `cache-early-messages` transformer and birth messages highlights that something is wonky.

First the config parameter for `cache-early-messages` should be a topic not a topic-id from which is derived. Why a custom mapper should be forced to publish these messages on a specific topic?

Second and more importantly, the contract is difficult to explain because based on *fragile* assumptions including *another* component. This other component is either the c8y mapper or a user-provided flow which is responsible of populating the context *before* sending birth messages.

1. The  c8y mapper or a user-provided flow shares entity metadata in the `context.mapper`.
    And *only then* emits birth messages for these entities.
2.  `cache-early-messages`  reacts on birth messages releasing any message cached for these entities.
   *assuming* blindly that the cache is actually properly populated for those.
3. The step downstream proceeds transforming the messages released by the birth messages;
   but *discarding* them if step one has not be done properly.

I do think this must be fixed. I propose the following:

- Deprecate birth messages before release 1.8. This is a fragile mechanism aimed to signal the context is populated  *but* not directly related to the context.
- Add an `onContextUpdate` callback to flow scripts and transformers. This will be called when the `context.mapper` is updated.
- Re-implement `cache-early-messages` to leverage `onContextUpdate` callbacks to release cached messages when the context is *actually* populated and not when a tier says it.

tests/RobotFramework/tests/tedge_flows/builtin-transformers.robot
reubenmiller: The latest commit adds the example that I was after. Being able to replay multiple input messages is a very useful feature!

docs/src/references/mappers/flows.md
reubenmiller: Though we might need to talk about how the `onContextUpdate` handler would work, and if it would react from any changes in the context, or just a subset, and what order is the handler called etc....though it does sound interesting a slightly more fitting for what we are using it for (in this case at least)

PR #4022, opened 2026-03-03
fix: respect sudo.enable setting when running log plugins

reubenmiller:
## Proposed changes

Fix a bug where the log plugin was not respecting the `sudo.enable` setting, so it would always run the plugins using sudo.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
+++ b/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
@@ -232,6 +232,27 @@ Reporting sudo misconfiguration
     ...    current_only=${True}
     ...    text=ERROR log plugins: Skipping /etc/share/tedge/log-plugins/dummy_plugin: not properly configured to run with sudo
 
+Supported log types loaded when sudo disabled
+    [Documentation]    Log plugins should be discovered and listed even when sudo.enable is false.
+    ...    Verifies that the agent does not unconditionally invoked sudo during plugin detection.
+    [Tags]    adapter:docker
+    Sleep    1s    reason=Allow the agent to complete startup actions before stopping
+    Stop Service    tedge-agent
+
+    # remove tedge sudo rules to prevent any command from working
+    Execute Command    cmd=rm -f /etc/sudoers.d/tedge
+
+    Execute Command    tedge config set sudo.enable false
+
+    # install a new plugin
+    ThinEdgeIO.Transfer To Device
+    ...    ${CURDIR}/plugins/dummy_plugin
+    ...    /usr/share/tedge/log-plugins/dummy_plugin2
+    Execute Command    chmod a+x /usr/share/tedge/log-plugins/dummy_plugin2
+
+    Start Service    tedge-agent
+    Should Contain Supported Log Types    dummy_log::dummy_plugin2
+
 Agent resilient to plugin dirs removal
     ${date_from}=    Get Unix Timestamp
     Execute Command    rm -rf /usr/share/tedge/log-plugins
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -403,6 +403,7 @@ impl Agent {
                 mqtt_schema: mqtt_schema.clone(),
                 mqtt_device_topic_id: device_topic_id.clone(),
                 plugin_dirs: self.config.log_plugin_dirs,
+                is_sudo_enabled: self.config.is_sudo_enabled,
             })?;
 
             let plugin_config =
--- a/crates/extensions/tedge_log_manager/src/config.rs
+++ b/crates/extensions/tedge_log_manager/src/config.rs
@@ -45,6 +45,7 @@ pub struct LogManagerOptions {
     pub mqtt_schema: MqttSchema,
     pub mqtt_device_topic_id: EntityTopicId,
     pub plugin_dirs: Vec<Utf8PathBuf>,
+    pub is_sudo_enabled: bool,
 }
 
 impl LogManagerConfig {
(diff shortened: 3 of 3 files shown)


tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -232,6 +232,29 @@ Reporting sudo misconfiguration
     ...    current_only=${True}
     ...    text=ERROR log plugins: Skipping /etc/share/tedge/log-plugins/dummy_plugin: not properly configured to run with sudo
 
+Supported log types loaded when sudo disabled
+    [Documentation]    Log plugins should be discovered and listed even when sudo.enable is false.
+    ...    Verifies that the agent does not unconditionally invoked sudo during plugin detection.
+    [Tags]    adapter:docker
+    Sleep    1s    reason=Allow the agent to complete startup actions before stopping
+    Stop Service    tedge-agent
+
```
rina23q, line 241: Probably better to assert the check if `tedge-agent` is really stopped?
```suggestion
    Service Health Status Should Be Down    tedge-agent

```

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -232,6 +232,29 @@ Reporting sudo misconfiguration
...
+    # remove tedge sudo rules to prevent any command from working
+    Execute Command    cmd=rm -f /etc/sudoers.d/tedge
+
+    Execute Command    tedge config set sudo.enable false
+
+    # install a new plugin
+    Execute Command    mkdir -p /etc/share/tedge/log-plugins
```
rina23q: Mistake using `/etc` instead of `/usr`?

```suggestion
    Execute Command    mkdir -p /usr/share/tedge/log-plugins
```

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
reubenmiller: Though the command will fail if the tedge-agent stopped (as it is using systemctl under the hood), so a check isn't needed (as we don't need to check that systemctl did what it says it will do).

reubenmiller: Though the 1s sleep hack is debatable as I think we have a problem in some actors which result in a slow shutdown (~60s), so I figured if the extra second helps us avoid a 60s timeout, then it is worthwhile using until we address the underlying reason

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
reubenmiller: It was just copied from another test, but yeah you're right that using a custom folder isn't really adding anything to the test so I'll remove that part and just use the default folder.

reubenmiller: Resolved in d2b3cbd2c9d00d9a1253602a69ff753346b98383 where I just removed the folder creation and just used the default plugin directory

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
rina23q: > Though the 1s sleep hack is debatable as I think we have a problem in some actors which result in a slow shutdown (~60s),

Yeah, exactly my point is that, if the corner case occurs (tedge-agent doesn't shut down within 60s), then the keyword can tell at least the failure reason clearly.

reubenmiller: But in both cases the tedge-agent will be stopped after the "Stop Service" keyword returns successfully

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -232,6 +232,27 @@ Reporting sudo misconfiguration
     ...    current_only=${True}
     ...    text=ERROR log plugins: Skipping /etc/share/tedge/log-plugins/dummy_plugin: not properly configured to run with sudo
 
+Supported log types loaded when sudo disabled
+    [Documentation]    Log plugins should be discovered and listed even when sudo.enable is false.
+    ...    Verifies that the agent does not unconditionally invoked sudo during plugin detection.
+    [Tags]    adapter:docker
```
albinsuresh, line 238: Out of curiosity, any specific reason why this tag is applied specifically for this test, unlike the other tests where it is added to `Test Tags` under `Settings`?

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -232,6 +232,27 @@ Reporting sudo misconfiguration
...
+    Sleep    1s    reason=Allow the agent to complete startup actions before stopping
```
albinsuresh, line 239: Wouldn't that be enough?
```suggestion
    Service Health Status Should Be Up    tedge-agent
```

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
reubenmiller: The reason was because it is doing a destructive action (removing the sudoers rule). Since we don't have a nice interface to revert this action in all cases (without overriding the automated cleanup), then it would break another test environment if we were to use the ssh adapter etc.

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
reubenmiller: Unfortunately the health status doesn't tell us if the agent is actually doing any work or not.

Issue #3985, opened 2026-02-16
Mitigate infinite loops


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When creating a custom flow, it is easy for users to create an infinite loop, by accidentally publishing a message to the same topic the flow is listening to.

For instance, the following flow would create an infinite loop which results in CPU resources from being consumed as well as potentially overloading the local MQTT broker causing other more important message from being dropped.

```js
export function onMessage(message, context) {
    return [{
        topic: message.topic,
        payload: message.payload,
    }];
}
```

**Describe the solution you'd like**

Flows which publish a message to the same MQTT topic as the configured input MQTT topics, should result in the following behaviour:

* Drop the message  (i.e. don't publish it to mitigate the infinite loop) if a fl
* Publish an error message (and a log message?), for observability, to declare that a message was attempted to be published to the same input topic.

Note: We don't need to add a complex multi-flow infinite loop check (e.g. A -> B -> C -> A), as we are trying to eliminate the common user-error scenarios.

**Optional: Enable users to disable the infinite loop checker**

For advanced cases, it would be useful to allow more technical users the ability to disable the infinite loop checker, as users can't implement their own protection mechanism to avoid creating infinite loops.

**Describe alternatives you've considered**

**Additional context**

PR #4024, opened 2026-03-06
fix: detect message loops in a flow

albinsuresh:
## Proposed changes

- [x] Detect and skip loading flows with static loops 
- [x] Detect loops during runtime and drop those messages specifically
- [x] Flow definition setting to disable loop detection

## Types of changes

## Paste Link to the issue
<br/>

https://github.com/thin-edge/thin-edge.io/issues/3985

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/config.rs
+++ b/crates/extensions/tedge_flows/src/config.rs
@@ -41,6 +41,10 @@ pub struct FlowConfig {
     output: OutputConfig,
     #[serde(default = "default_errors")]
     errors: OutputConfig,
+
+    /// If true, output messages that match the input filter are not dropped
+    #[serde(default)]
+    expect_loop: bool,
 }
 
 #[derive(Deserialize)]
@@ -120,6 +124,16 @@ pub enum ConfigError {
 
     #[error("Not a valid step configuration: {0}")]
     IncorrectSetting(String),
+
+    #[error("Flow '{name}' defines an infinite loop: the output topic '{output_topic}' matches input filter '{input_filter}'")]
+    MqttInfiniteLoop {
+        name: String,
+        input_filter: String,
+        output_topic: String,
+    },
+
+    #[error("Flow '{name}' defines an infinite loop: the output file '{path}' is the same as the input file")]
+    FileInfiniteLoop { name: String, path: String },
 }
 
 impl FlowConfig {
@@ -209,6 +223,8 @@ impl FlowConfig {
             description: None,
             tags: None,
             config: Map::new(),
+            // Expect a loop when wrapping a single script as a flow, as there is no way to statically identify input and output topics
+            expect_loop: true,
             input: InputConfig::Mqtt {
                 topics: vec![input_topic],
             },
@@ -239,6 +255,9 @@ impl FlowConfig {
         let name = self
             .name
             .unwrap_or_else(|| source.file_name().unwrap_or_default().to_string());
+
+        detect_loop(&name, &input, &output, self.expect_loop)?;
+
         Ok(Flow {
             name,
             version: self.version,
@@ -249,6 +268,7 @@ impl FlowConfig {
             output,
             errors,
             source,
+            expect_loop: self.expect_loop,
         })
     }
 }
@@ -429,10 +449,48 @@ fn default_errors() -> OutputConfig {
     }
 }
 
+/// Checks whether `input` and `output` form an infinite loop,
+/// where the output of published to the same input source.
+/// When `expect_loop` is true, the check is skipped.
+fn detect_loop(
+    name: &str,
+    input: &FlowInput,
+    output: &FlowOutput,
+    expect_loop: bool,
+) -> Result<(), ConfigError> {
+    if expect_loop {
+        return Ok(());
+    }
+    match (input, output) {
+        (FlowInput::Mqtt { topics }, FlowOutput::Mqtt { topic: Some(out) })
+            if topics.accept_topic_name(&out.name) =>
+        {
+            Err(ConfigError::MqttInfiniteLoop {
+                name: name.to_string(),
+                input_filter: format!("{topics:?}"),
+                output_topic: out.name.clone(),
+            })
+        }
+        (
+            FlowInput::PollFile { path: in_path, .. } | FlowInput::StreamFile { path: in_path, .. },
+            FlowOutput::File { path: out_path },
+        ) if in_path == out_path => Err(ConfigError::FileInfiniteLoop {
+            name: name.to_string(),
+            path: in_path.to_string(),
+        }),
+        _ => Ok(()),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
+    use camino::Utf8PathBuf;
     use serde_json::json;
+    use std::time::Duration;
+    use tedge_mqtt_ext::Topic;
+    use tedge_mqtt_ext::TopicFilter;
+    use test_case::test_case;
 
     #[test]
     fn inherit_shared_config() {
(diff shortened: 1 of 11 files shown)


crates/extensions/tedge_flows/src/config.rs
```diff
@@ -239,6 +249,10 @@ impl FlowConfig {
         let name = self
             .name
             .unwrap_or_else(|| source.file_name().unwrap_or_default().to_string());
+
+        // Detect static infinite loops before constructing the flow.
+        detect_loop(&name, &input, &output)?;
```
albinsuresh: The 2 test failures are caused by this detection logic getting applied to single script tests using `tedge flows test` command, which needs to be bypassed. Planning to use a special tag to detect this case. Or a simpler option would be to detect the `.js` extension from the `name` of the flow.

albinsuresh: Resolved by the simpler name based detection in 6ed097a760

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -304,18 +305,49 @@ impl Flow {
...
-            Ok(messages) => FlowResult::Ok {
-                flow: self.source.clone(),
-                messages,
-                output: self.output.clone(),
-            },
+            Ok(mut messages) => {
+                // Perform loop detection, but only for full flows and not for single test scripts
+                if !self.name.ends_with(".ts") && !self.name.ends_with(".js") {
+                    messages = self.filter_looping_messages(messages);
+                }
```
didier-wenzek: I'm not happy with this fix using names to excludes some flows.

- The flows to be excluded are only used for testing purpose, the `tedge flows test` command wrapping a single step into a flow.
- This name check is done for all the messages in a production context where the flow is *never* a simple wrapper around a script.
- This is fragile. 

A better fix would be to adapt the script-to-flow wrapper.

For instance, to use a non MQTT output for these script-wrapper flows.

didier-wenzek:
```diff
--- a/crates/extensions/tedge_flows/src/config.rs
+++ b/crates/extensions/tedge_flows/src/config.rs
@@ -223,7 +223,7 @@ impl FlowConfig {
                 topics: vec![input_topic],
             },
             steps: vec![step],
-            output: default_output(),
+            output: OutputConfig::File { path: Default::default() },
             errors: default_errors(),
         }
     }

```

albinsuresh: I like that suggestion to use a different output type for this case. But, using a dummy `File` output also feels a bit hacky. How about introducing a dedicated `OutputConfig` variant, like `Console` or `Stdout`, so that it's clearer?

didier-wenzek: Indeed, a  `Console` or `Stdout` output would make much more sense, as this is what is actually done by the `tedge flows test` command.

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -239,6 +252,13 @@ impl FlowConfig {
         let name = self
             .name
             .unwrap_or_else(|| source.file_name().unwrap_or_default().to_string());
+
+        // Perform loop detection, but only for full flows and not for single test scripts
+        if !name.ends_with(".ts") && !name.ends_with(".js") {
```
albinsuresh: This check isn't needed anymore as the output type is different for wrapped flows with a single script.

albinsuresh: Resolved by [eb9ddd0](https://github.com/thin-edge/thin-edge.io/pull/4024/commits/eb9ddd03e7b243cea4a97a7c4c6fb82c0828bc90)

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -239,6 +254,9 @@ impl FlowConfig {
         let name = self
             .name
             .unwrap_or_else(|| source.file_name().unwrap_or_default().to_string());
+
+        detect_loop(&name, &input, &output)?;
```
albinsuresh: **Not** guarding this static loop check with the `expect_loop` flag is intentional, as I don't feel there is any value in disabling static loops in the flow definition that can never be recovered from. I am of the opinion that this flag is only valid for runtime loops, which can be broken out of, using some logic in the script.

crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -87,7 +88,14 @@ impl TEdgeComponent for AwsMapper {
         );
         let flows_dir =
             tedge_flows::flows_dir(config_dir, "aws", self.profile.as_ref().map(|p| p.as_ref()));
-        let flows = aws_converter.flow_registry(flows_dir).await?;
+        let flows = aws_converter
+            .flow_registry(
+                flows_dir,
+                FlowsConfig {
+                    loop_detection_enabled: tedge_config.flows.loop_detection_enabled,
```
albinsuresh: We can introduce cloud-specific config settings as well, if there is a requirement. Keeping it simple with a global setting for now.

crates/extensions/tedge_flows/src/flow.rs
albinsuresh: Went with `OutputConfig::Log` instead, as defining this as the output target (`[output.log]`) results in the messages getting logged in the local-mapper log.

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -49,6 +51,9 @@ pub struct Flow {
 
     /// Path to the configuration file for this flow
     pub source: Utf8PathBuf,
+
+    /// Runtime configuration for this flow
+    pub flows_config: FlowsConfig,
```
didier-wenzek: I disagree with this idea of a global flow configuration that is injected into each flow.

For many reasons:

- Enabling loop detection is a flow specific property, not a global property.
- Having `Flow`, `FlowConfig` and `FlowsConfig` is confusing.
- Why an inner level? Why not simply a `loop_detection` field? We have such Russian dolls configuration struct for the c8y mapper and this is really difficult to find what is set where.

So, I propose to:
- Add an optional `loop_detection` field in the flow `.toml` files
- Remove any global `tedge config` settings for loop detection. 

Also:
- If we have a way to disable loop detection on a flow, then the new `Log` output for a flow is useless.

tests/RobotFramework/tests/tedge_flows/fs_dynamic_reload.resource
```diff
@@ -1,5 +1,5 @@
 *** Settings ***
-Library    ThinEdgeIO
+Library     ThinEdgeIO
```
albinsuresh, line 2: `just format` command is doing that. Not quite sure why. I'm just surprised why it wasn't flagged earlier, when the original code was merged.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -261,7 +261,9 @@ Flow is added/removed when a directory is moved in/out
 
     # now test on the same filesystem
     Execute Command    mv /tmp/myflow /etc/tedge/mappers/local/myflow
-    Move Directory And Assert it's added and removed    /etc/tedge/mappers/local/myflow    after moving directory into flows dir
+    Move Directory And Assert it's added and removed
```
albinsuresh, line 264: Yet another formatting by `just format`.

crates/extensions/tedge_flows/src/flow.rs
albinsuresh: The new output type was replaced by the flow-level `expect_loop` flag.

crates/core/tedge_mapper/src/aws/mapper.rs
albinsuresh: Obsolete as we don't have a global config setting anymore.

crates/extensions/tedge_flows/src/flow.rs
albinsuresh: Resolved by the introduction of `expect_loop` config in the flow definition.

jarhodes314 (approved):

crates/extensions/tedge_flows/src/flow.rs
```diff
@@ -304,18 +308,48 @@ impl Flow {
...
+                                "Flow '{}' is dropping output message to '{}' to prevent an infinite loop",
+                                self.name,
+                                msg.topic
+                            );
+                            false
+                        } else {
+                            true
+                        }
+                    })
+                    .collect()
```
jarhodes314: Minor: we could use retain to modify the `Vec` in-place instead of destroying it and allocating a new one.
```suggestion
                messages
                    .retain(|msg| {
                        if input_topics.accept_topic_name(&msg.topic) {
                            error!(
                                target: "flows",
                                "Flow '{}' is dropping output message to '{}' to prevent an infinite loop",
                                self.name,
                                msg.topic
                            );
                            false
                        } else {
                            true
                        }
                    });
                messages
```

albinsuresh: Resolved by 97a7acf

Issue #4023, opened 2026-03-05
flow is removed from memory when it was updated


reubenmiller:
**Describe the bug**

It was observed that when installing a flow (`uptime`), using the [PoC sm-plugin](https://github.com/thin-edge/tedge-oscar/blob/main/sm-plugins/flow), after the installation, the flow was removed from the tedge-mapper-c8y.

Below shows log entries from the tedge-mapper-c8y service showing the log entries around the loading and removing of flows:

```log
Mar 05 21:42:44 tedge tedge-mapper[94020]: 2026-03-05T21:42:44.8944025Z  INFO flows: Loading flow /etc/tedge/mappers/c8y/flows/uptime/flow.toml
Mar 05 21:42:44 tedge tedge-mapper[94020]: 2026-03-05T21:42:44.895503173Z  INFO flows: Removing flow /etc/tedge/mappers/c8y/flows/uptime/flow.toml
Mar 05 21:42:44 tedge tedge-mapper[94020]: 2026-03-05T21:42:44.895671424Z  INFO flows: Loading flow /etc/tedge/mappers/c8y/flows/uptime/flow.toml
Mar 05 21:42:44 tedge tedge-mapper[94020]: 2026-03-05T21:42:44.896540596Z  INFO flows: Reloading flow script /etc/tedge/mappers/c8y/flows/uptime/lib/main.js
Mar 05 21:42:44 tedge tedge-mapper[94020]: 2026-03-05T21:42:44.89658043Z  INFO flows: Removing flow /etc/tedge/mappers/c8y/flows/uptime/flow.toml
```

And on the file system, the flow "uptime" is under the `c8y` flows directory:

```sh
c8y
└── flows
    ├── alarms.toml
    ├── alarms.toml.template
    ├── events.toml
    ├── events.toml.template
    ├── health.toml
    ├── health.toml.template
    ├── measurements.toml
    ├── measurements.toml.template
    ├── units.toml
    ├── units.toml.template
    └── uptime
        ├── flow.toml
        ├── lib
        │   └── main.js
        ├── params.toml
        └── params.toml.template
```

Then if the tedge-mapper-c8y service is restarted, then the "uptime" flow is automatically loaded (showing that it was just an runtime related issue).

```sh
systemctl restart tedge-mapper-c8y
```

**To Reproduce**

The exact steps are not yet know, but I suspect it might be a race condition between the inotify events whilst updating a flow.

**Expected behavior**

[...]

PR #4031, opened 2026-03-10
fix: improve handling of file and directory events in FlowsMapper

reubenmiller:
## Proposed changes

Fixes flow being permanently lost from memory after a package manager updates it (deletes then recreates the files), even though files are restored on disk (#4023).

Three targeted fixes:

### 1. `remove_script()` — unload instead of remove (`registry.rs`)

When a script file is deleted, flows referencing it are now **unloaded** (moved to `unloaded_flows`) instead of being silently ignored with a warning. This means when the script is restored and `reload_script()` calls `drain_unloaded()`, the flow is rediscovered and reloaded automatically.

Previously, `remove_script()` only logged a warning and returned early — the flow stayed in the `flows` map with a broken script reference but was never re-evaluated when the script came back.

### 2. `add_flow()` — no-op on missing file instead of destructive removal (`registry.rs`)

Changed from `if tokio::fs::read_to_string().is_err() → remove_flow()` to `if !path.is_file() → return`.

The old code would destroy a valid in-memory flow on a transient read failure (e.g. file briefly absent during a package update). The new code simply returns — if the file was truly deleted, a `FileDeleted` event will handle removal; if it's being replaced, a subsequent `Modified` event will reload it.

### 3. `on_file_removed()` — skip stale `FileDeleted` events (`actor.rs`)

Added a `path.exists()` guard at the top of `on_file_removed()`. During a package update, `FileDeleted` events can arrive after the file has already been re-created. Without this guard, the stale event would remove the freshly-reloaded flow.

Also added the missing `send_updated_subscriptions()` + `update_flow_status()` calls for script removals, which were previously only done for `.toml` removals.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4023

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
+++ b/tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
@@ -84,6 +84,45 @@ Flow script update changes processing behavior
     ...    message_contains="version":"v2"
     ...    date_from=${start}
 
+Removing script unloads flow and restoring script loads flow again
+    [Documentation]    After the JS script file is moved of the path referred by flow, the flow must be unloaded and loaded again when script is restored.
+    Install Lifecycle Flow    main-v1.js
+
+    # Verify v1 behavior
+    ${start}    Get Unix Timestamp
+    Execute Command    tedge mqtt pub ${INPUT_TOPIC} '{}'
+    Should Have MQTT Messages
+    ...    topic=${OUTPUT_TOPIC}
+    ...    message_contains="version":"v1"
+    ...    date_from=${start}
+
+    # Move script to main flows dir
+    Execute Command    sleep 1
+    ${start}    Get Unix Timestamp
+    Execute Command    mv ${FLOW_DIR}/main.js ${FLOWS_DIR}/main.js
+
+    # FIXME: when we get Modified event for a js file we currently load it immediately even if it's not referred by any
+    # loaded or unloaded flow. Then we try to load all unloaded flows, hoping loaded script fixed them, but if these
+    # flows are not using the script, we emit this error message. Instead we shouldn't try to reload these flows unless
+    # the script we're modifying is part of the flow, which will avoid this error message.
+    Logs Should Contain    Failed to compile flow    date_from=${start}
+    Should Have MQTT Messages
+    ...    topic=${FLOW_STATUS_TOPIC}
+    ...    date_from=${start}
+    ...    message_contains="flow":"/etc/tedge/mappers/local/flows/lifecycle-test/main.js","status":"removed"
+
+    # Move script back to the flow dir
+    Execute Command    sleep 1
+    Execute Command    mv ${FLOWS_DIR}/main.js ${FLOW_DIR}/main.js
+
+    # Verify v1 behavior
+    ${start}    Get Unix Timestamp
+    Execute Command    tedge mqtt pub ${INPUT_TOPIC} '{}'
+    Should Have MQTT Messages
+    ...    topic=${OUTPUT_TOPIC}
+    ...    message_contains="version":"v1"
+    ...    date_from=${start}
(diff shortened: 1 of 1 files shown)


tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
Bravo555, line 1: Nice to have a separate suite dedicated to handling fs events, similar tests currently in main `tedge_flows` suite could also be moved here.

tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
```diff
@@ -0,0 +1,267 @@
...
+    Execute Command    sleep 1
+
+    # Messages sent after removal must not appear on the output topic
+    ${start}    Get Unix Timestamp
+    Execute Command    tedge mqtt pub ${INPUT_TOPIC} '{}'
+    Should Not Have MQTT Messages
+    ...    topic=${OUTPUT_TOPIC}
+    ...    date_from=${start}
+
+Flow script update changes processing behavior
```
Bravo555, line 58: note: Without the changes in `.rs` files, this is the only test that fails, with `Flow is reloaded after package-style replacement` regression test, which refers to the issue, notably seems to be passing on my machine. Is it different on yours?

crates/extensions/tedge_flows/src/registry.rs
```diff
@@ -191,10 +191,15 @@ impl<T: FlowRegistry + Send> FlowRegistryExt for T {
...
+        if !path.is_file() {
+            // The file doesn't exist at this moment (TOCTOU: it was present when
+            // on_file_updated checked path.is_file(), or when drain_unloaded queued
+            // this path, but was removed before we got here).  Do nothing: if the
+            // file was truly deleted a FileDeleted event will arrive and call
+            // remove_flow; if it is being replaced a subsequent Modified event will
+            // reload it.
             return;
-        };
+        }
```
Bravo555, line 201: thought: both in the before, and in the after, this function is doing too much/knowing too much - before it sometimes removed the flow while being called `add_flow` - and now it depends on some specific behaviour of the caller it should be unaware of. But with the current interface and design, it probably can't be entirely correct.

crates/extensions/tedge_flows/src/actor.rs
Bravo555, line 524: While reviewing the change, I found another small issue - when the script is removed we remove the flow as if we removed the `.toml` file, but it's not right! The flow should be unloaded, not entirely removed, only then `reload_script` starts the flow back up again after the script is valid again.
It's possible applying only this fix would be sufficient to resolve the issue, which I'm testing now.

tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
reubenmiller: I pushed a recent change, and now all of the system test pass, and the CI passes as well. Did you check the latest commit?

crates/extensions/tedge_flows/src/actor.rs
didier-wenzek, line 555:
```
    self.processor.remove_script(path).await;
} else if path.extension() == Some("toml") {
    self.processor.remove_flow(path).await;
}
self.update_flow_status(path).await?;
```

reubenmiller: I guess that assumes that we don't get any other notifications which are from .js, .ts, .mjs nor .toml files?

didier-wenzek: You are correct, we have to handle the case of a flows unrelated file.

reubenmiller: I pushed the recommended change

PR #4036, opened 2026-03-11
fix: Unload flow when removing script instead of removing

Bravo555:
## Proposed changes

Unload flow when one of its scripts is removed instead of removing. Removing sometimes made flow not start back up when script was restored. Instead, unloading keeps it in an unloaded list that is checked when relevant script is reloaded.

This is an alternative fix to #4031 using the same test suite, but addresses flow not being reloaded without assuming stale, out-of-order `FileDeleted` messages, which I could not reproduce. Instead my hypothesis is that the flow wasn't being reloaded because `reload_script` looks at unloaded flows list to restart flows, but the flow wasn't in that list because it was removed instead of just unloaded.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/actor.rs
+++ b/crates/extensions/tedge_flows/src/actor.rs
@@ -504,9 +504,9 @@ impl FlowsMapper {
             })
             .collect();
 
-        for file in unloaded_flows {
+        for flow in unloaded_flows {
             // will add flow to list of unloaded if scripts are missing
-            self.processor.load_single_flow(&file).await;
+            self.processor.registry.unload_flow(&flow);
             self.send_updated_subscriptions().await?;
             self.update_flow_status(path).await?;
         }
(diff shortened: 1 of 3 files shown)


reubenmiller:
I was testing with the PoC [sm-plugin](https://raw.githubusercontent.com/thin-edge/tedge-oscar/refs/heads/main/sm-plugins/flow) for a flow, and upgrading an existing flow results in the following log entries:

```log
Mar 11 20:17:08 7e289213ec5f tedge-mapper[808]: 2026-03-11T20:17:08.746513095Z  WARN flows: Removing a script used by a flow protobuf-xform: /etc/tedge/mappers/local/flows/protobuf-xform/lib/main.js
Mar 11 20:17:08 7e289213ec5f tedge-mapper[808]: 2026-03-11T20:17:08.757372524Z  INFO flows: Reloading flow script /etc/tedge/mappers/local/flows/protobuf-xform/lib/main.js
```

The WARN level for the flow removal is overkill (INFO would be more appropriate), however interesting enough, https://github.com/thin-edge/thin-edge.io/pull/4031, only yields a single "Reloading flow" log entry on updates (which is preferable from a users perspective given an update for a user should trigger a reload and the "remove" operation is unnecessary (again from the users point of view).

Issue #3986, opened 2026-02-16
Create a software management plugin to managed flows


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Add a software management plugin to list, install and remove flows for different mappers.

**Describe the solution you'd like**

Implement as defined in the following document
* [sm-plugin design](https://github.com/thin-edge/tedge-oscar/blob/main/docs/FLOWS_PACKAGING.md)

**Describe alternatives you've considered**

**Additional context**

PR #4037, opened 2026-03-11
feat: sm-plugin for flows 

rina23q:
## Proposed changes
This pull request introduces the new `tedge-flows-plugin` package and a binary symlinked to `/etc/tedge/sm-plugin/flow`.

The plugin supports only 3 commands:
1. list
2. install
3. remove

Note: `update-list` is unsupported, therefore, it returns `1` as in the sm-plugin specification.

### Naming rules
Look at https://github.com/thin-edge/thin-edge.io/pull/4037#issuecomment-4074367221.

### Installation rules
- No hard check for the given `--module-version <version>`. If given, it will be used to check the provided version equals to the one installed. 
- Installing a flow onto the top level is unsupported whilst nested layers are possible; simply, because it's impossible to distinguish if user wants it at the top level or second level, from the naming rule `local/toplevel` and `local/second`.
- Attempts to unpack if a provided file path doesn't have `.tar` or `.tar.gz`. At first, I didn't implement this fallback, however, I found that the file extension is missing while downloading a file... If there was no fallback, integration with c8y won't work :/

### Removal rules
- ignores `--module-version <version>`.
- `params.toml` will be kept only if `flows.params.keep_on_delete` is `true` (default: `false`).

## Types of changes

## Paste Link to the issue
#3986 

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/tedge_agent/device_profile/device_profile_operation.robot
+++ b/tests/RobotFramework/tests/tedge_agent/device_profile/device_profile_operation.robot
@@ -206,6 +206,7 @@ Custom Setup
     ${DEVICE_SN}=    Setup
     Set Suite Variable    $DEVICE_SN
     Device Should Exist    ${DEVICE_SN}
+    Execute Command    tedge config set software.plugin.default apt
 
     Copy Configuration Files
     Restart Service    tedge-agent
--- a/tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
+++ b/tests/RobotFramework/tests/cumulocity/configuration/config_operation_plugins.robot
@@ -45,7 +45,7 @@ Config operation plugin
 
 Supported config types updated on software update
     ${start_time}=    Get Unix Timestamp
-    ${OPERATION}=    Install Software    cron
+    ${OPERATION}=    Install Software    {"name": "cron", "softwareType": "apt"}
     Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
 
     Command Metadata Should Have Refreshed    ${start_time}
--- a/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
+++ b/tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
@@ -28,7 +28,7 @@ Log operation journald plugin
 
 Supported log types updated on software update
     ${start_time}=    Get Unix Timestamp
-    ${OPERATION}=    Install Software    cron
+    ${OPERATION}=    Install Software    {"name": "cron", "softwareType": "apt"}
     Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
 
     Should Have MQTT Messages
--- a/tests/RobotFramework/tests/cumulocity/software_management/software.robot
+++ b/tests/RobotFramework/tests/cumulocity/software_management/software.robot
@@ -17,19 +17,22 @@ Supported software types should be declared during startup
     ...    topic=te/device/main///cmd/software_list
     ...    minimum=1
     ...    maximum=1
-    ...    message_contains="types":["apt"]
+    ...    message_contains="types":["apt","flow"]
     Should Have MQTT Messages
     ...    topic=te/device/main///cmd/software_update
     ...    minimum=1
     ...    maximum=1
-    ...    message_contains="types":["apt"]
-    Run Keyword And Expect Error    *    Device Should Have Fragment Values    c8y_SupportedSoftwareTypes\=["apt"]
+    ...    message_contains="types":["apt","flow"]
+    Run Keyword And Expect Error
+    ...    *
+    ...    Device Should Have Fragment Values
+    ...    c8y_SupportedSoftwareTypes\=["apt", "flow"]
 
 Supported software types and c8y_SupportedSoftwareTypes should be declared during startup
     [Documentation]    c8y_SupportedSoftwareTypes should be created if the relevant config is set to true #2654
     Execute Command    tedge config set c8y.software_management.with_types true
     Restart Service    tedge-mapper-c8y
-    Device Should Have Fragment Values    c8y_SupportedSoftwareTypes\=["apt"]
+    Device Should Have Fragment Values    c8y_SupportedSoftwareTypes\=["apt", "flow"]
 
 Software list should be populated during startup
     [Documentation]    The list is sent via HTTP by default.
@@ -211,6 +214,7 @@ Custom Setup
     Set Test Variable    $DEVICE_SN
     Should Have MQTT Messages    te/device/main/service/tedge-mapper-c8y/status/health
     Execute Command    sudo start-http-server.sh
+    Execute Command    tedge config set software.plugin.default apt
 
 Stop tedge-agent
     [Timeout]    5 seconds


tests/RobotFramework/tests/plugin_flows/test_plugin.robot
```diff
@@ -0,0 +1,51 @@
...
+List flows
+    ${packages_list}=    Execute Command
+    ...    sudo /etc/tedge/sm-plugins/flow list
+    ...    exp_exit_code=0
+    Should Contain    ${packages_list}    local/default-flow\t1.0.0
+
+Install a .tar flow
+    Execute Command    sudo /etc/tedge/sm-plugins/flow install local/hello-flow --module-version 1.0.0 --file /usr/local/tedge/hello-flow.tar    exp_exit_code=0
+    ${flows_list}=    Execute Command    sudo /etc/tedge/sm-plugins/flow list    exp_exit_code=0
+    Should Contain    ${flows_list}    local/hello-flow\t1.0.0
```
reubenmiller: Given this bug, https://github.com/thin-edge/thin-edge.io/issues/4023, we should check whether the flow is actually functional (or non-functional) after installation and removal.

Plus we're missing a case for "updating an existing flow" (at least in the c8y tests)

rina23q: Refactored the tests with b155f0efb17d03343cbb12cbe3b6f9d4ddc29082 and added a check if the flow is working (or not) in f1e6eaf9e2aa6edd54ebe8eb694132f728037aac.

tests/RobotFramework/tests/plugin_flows/test_plugin.robot
```diff
@@ -0,0 +1,112 @@
...
+    Directory Should Not Exist    /etc/tedge/mappers/local/flows/hello-flow
+    Verify a flow is not working
+
+Remove a flow with keeping params.toml
+    [Setup]    Transfer hello-flow    with_params_toml=True
+    Execute Command    sudo /etc/tedge/sm-plugins/flow remove local/hello-flow --module-version 1.0.0    exp_exit_code=0
+    ${flows_list}=    Execute Command    sudo /etc/tedge/sm-plugins/flow list    exp_exit_code=0
+    Should Not Contain    ${flows_list}    local/hello-flow\t1.0.0
+    Directory Should Exist    /etc/tedge/mappers/local/flows/hello-flow
+    File Should Exist    /etc/tedge/mappers/local/flows/hello-flow/params.toml
```
reubenmiller: We had an offline conversation about this, and we reached the conclusion that it would be useful if there was a tedge.toml setting that would control the behaviour to keep the params.toml or not when removing a flow.

Below shows a suggestion along with the recommended default value.

```
flows.params.keep_on_delete=false
```

The reason for deleting the params.toml on the package removal is so that it is easier for users to reset the flow by simply uninstalling it, and installing it again.

tests/RobotFramework/tests/plugin_flows/test_with_c8y.robot
```diff
@@ -0,0 +1,64 @@
...
+    Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
+    Device Should Have Installed Software    {"softwareType": "flow", "name": "local/hello-flow", "version": "1.0.0"}
+    [Teardown]    Execute Command    rm -rf /etc/tedge/mappers/local/flows/hello-flow
+
+Update a flow
+    ${OPERATION}=    Install Software
+    ...    {"name": "local/world-flow", "version": "2.0.0", "softwareType": "flow", "url": "${V2_URL}"}
+    Operation Should Be SUCCESSFUL    ${OPERATION}    timeout=60
+    Device Should Have Installed Software    {"softwareType": "flow", "name": "local/world-flow", "version": "2.0.0"}
+    [Teardown]    Execute Command    rm -rf /etc/tedge/mappers/local/flows/world-flow
```
Bravo555: question: how does the update work here?

I'd expect the update process to be: we're already running flow of that name with 1.0.0 and then we're installing 2.0.0.
But AFAIK I see that there's no flow with such name already there when we start installing 2.0.0.

plugins/tedge_flows_plugin/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+            let (mapper_name, flow_name) = parse_module_name(&module)?;
+            crate::install::install_flow(&mapper_dir, mapper_name, flow_name, version, &file_path)?;
+        }
+        PluginOp::Remove { module, version } => {
+            let (mapper_name, flow_name) = parse_module_name(&module)?;
+            crate::remove::remove_flow(&mapper_dir, mapper_name, flow_name, version)?;
+        }
+        PluginOp::UpdateList => return Err(FlowsPluginError::InvalidUsage),
+        PluginOp::Prepare => {}
+        PluginOp::Finalize => {}
```
Bravo555, line 126: question: where is the difference between `update-list` where we return `1`, and `prepare`/`finalize` where we return `0`, coming from?

plugins/tedge_flows_plugin/src/lib.rs
```diff
@@ -0,0 +1,158 @@
...
+use tedge_config::log_init;
+use tracing::debug;
+use tracing::error;
+use tracing::warn;
+
+use crate::error::io_error;
+use crate::error::FlowsPluginError;
+
+pub(crate) const PARAMS_FILE: &str = "params.toml";
+pub(crate) const DEFAULT_VERSION: &str = "0.0.0";
```
didier-wenzek, line 23: I'm not sure that `0.0.0` is the best way to convey that the flow has *no* version.

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,244 @@
...
+}
+
+fn search_flows_recursive(
+    current_path: &Path,
+    base_flows_dir: &Path,
+    mapper_name: &str,
+    flows: &mut Vec<FlowEntry>,
+) {
+    // Check if flow.toml exists in the current directory
+    let flow_toml_path = current_path.join("flow.toml");
```
Bravo555: question: why does the name need to be `flow.toml` in nested flows?

When I test flows I have some named e.g. `onstartup.toml` and the local mapper runs it normally.
Specifically for flows managed by sm-plugin, is there a restricition somewhere that nested flow definitions have to have name `flow.toml`?

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,244 @@
...
+        };
+
+        flows.push(FlowEntry {
+            mapper: mapper_name.to_string(),
+            name: relative_path,
+            version,
+        });
+        // Stop recursing this branch once flow.toml is found
+        return;
+    }
```
Bravo555: issue: `.exists()` includes directories, so directories named `flow.toml` get reported as flows

tests/RobotFramework/tests/plugin_flows/test_plugin.robot
rina23q: Done in 801306744da3b7e477eefa717b99b216808bb7c5

plugins/tedge_flows_plugin/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+
+/// Flow metadata extracted from the flow.toml file
+#[derive(Deserialize, Debug)]
+struct FlowMeta {
+    version: Option<String>,
```
didier-wenzek: As a flow can be given a name that is not its filename, we have to extract its name from the its `.toml` definition.

```suggestion
    name: Option<String>,
    version: Option<String>,
```

And why not extracting and displaying all the meta:


```suggestion
    name: Option<String>,
    version: Option<String>,
    description: Option<String>,
    tags: Option<Vec<String>>,
```

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,244 @@
...
+    Ok(())
+}
+
+pub fn retrieve_flows(mappers_dir: impl AsRef<Path>) -> Result<Vec<FlowEntry>, FlowsPluginError> {
+    let mut flows: Vec<FlowEntry> = Vec::new();
+
+    let mappers_dir_ref = mappers_dir.as_ref();
+    debug!("Listing flows from directory: {mappers_dir_ref:?}");
+
+    let mappers_dir_read = match fs::read_dir(mappers_dir_ref) {
```
didier-wenzek: Instead of these recursive functions to read mapper directories, then flow directories and sub-directories, it might be simpler to [use a `glob` pattern as done here](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/tedge_flows/src/config.rs#L147).

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+    };
+
+    for entry in dir_read {
+        let (entry, file_type) = match read_entry_with_file_type(entry) {
+            Some(pair) => pair,
+            None => continue,
+        };
+
+        if file_type.is_dir() {
```
Bravo555: note: symlinks are not included, so there's no risk of any infinite recursive loops, nice

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,244 @@
...
+    for entry in flows_dir_read {
+        let (file_entry, file_type) = match read_entry_with_file_type(entry) {
+            Some(pair) => pair,
+            None => continue,
+        };
+        let entry_name = file_entry.file_name().to_string_lossy().into_owned();
+
+        // Pattern 1: there is a .toml file => flow name is the file name without .toml suffix
+        if file_type.is_file() && entry_name.ends_with(".toml") && entry_name != PARAMS_FILE {
+            let name = entry_name.strip_suffix(".toml").unwrap().to_string(); // safe to unwrap since we just checked the suffix
```
didier-wenzek: To get the name of a flow, you have to read its toml definition. Its filename can only be used as a default.

plugins/tedge_flows_plugin/src/list.rs
didier-wenzek: This restriction to `flow.toml` came as a surprised to me. Yes, it's nice to let users use a generic filename for their flows (hence the name property within the toml definition). But, this should not be mandatory.

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,244 @@
...
+                return;
+            }
+        };
+
+        flows.push(FlowEntry {
+            mapper: mapper_name.to_string(),
+            name: relative_path,
+            version,
+        });
+        // Stop recursing this branch once flow.toml is found
```
didier-wenzek: I would not treat sub-directories differently than mapper main directories. This is more work for us and more surprises for users.

plugins/tedge_flows_plugin/src/install.rs
```diff
@@ -0,0 +1,291 @@
...
+    debug!("Moving unpacked flow from temporary directory {tmp_dest:?} to destination {dest:?}");
+    fs::remove_dir_all(&dest).ok(); // Remove existing flow if it exists
+    fs::rename(&tmp_dest, &dest).map_err(|e| io_error(&dest, e))?;
+
+    if let Some(version) = version {
+        let version_from_file = get_flow_version(&dest.join("flow.toml"));
+        if version != version_from_file {
+            warn!(
+                "Provided version {version} does not match version {version_from_file} in flow.toml for flow {mapper_name}/{flow_name}."
+            );
```
Bravo555, line 72: issue: if the version in file doesn't match the version we wanted to install, shouldn't we abort?

What if somebody made a typo and used a wrong tar file? If somebody accidentally downgrades their flow, isn't there any potentiall for data loss?

plugins/tedge_flows_plugin/src/install.rs
```diff
@@ -0,0 +1,291 @@
...
+        .join("flows")
+        .join(flow_name);
+    let tmp_dest = dest.with_extension("tmp");
+    let params_path = dest.join(PARAMS_FILE);
+    let params_backup = tmp_dest.join(PARAMS_FILE);
+
+    if tmp_dest.exists() {
+        debug!("Cleaning up existing temporary directory at {tmp_dest:?}");
+        fs::remove_dir_all(&tmp_dest).map_err(|e| io_error(&tmp_dest, e))?;
+    }
```
Bravo555: issue: I think if there were multiple processes attempting to install the same flow, they would step on each other's toes trying to use the same temp directory.

Instead we should generate a unique temporary dirname and immediately try to create it, failing if it already exists (which is what `tempfile::tempdir_in` does, it's in the crate's dependencies but it doesn't seem to be used)

plugins/tedge_flows_plugin/src/lib.rs
reubenmiller: Though we need to show some value, and I'd prefer if it was still semver compliant in case if any tooling is using the value for some check.

plugins/tedge_flows_plugin/src/install.rs
```diff
@@ -0,0 +1,291 @@
...
+        let params_content = fs::read_to_string(mapper1_flows_hello.path().join("params.toml"))
+            .expect("Failed to read params.toml");
+        let params_toml: toml::Value =
+            toml::from_str(&params_content).expect("Failed to parse params.toml");
+        assert_eq!(params_toml["param1"].as_str().unwrap(), "value1");
+    }
+
+    /// Helper function to create a tarball with the specified content
+    /// Returns the path to the created tarball
+    #[allow(dead_code)]
```
Bravo555: suggestion: doesn't seem to be necessary

```suggestion
```

plugins/tedge_flows_plugin/src/lib.rs
rina23q: I didn't add other properties as the plugin uses only `version` from `.toml` file. 

For `.name`, https://github.com/thin-edge/tedge-oscar/blob/main/docs/FLOWS_PACKAGING.md says
> By convention the `.name` property inside the `flow.toml` should be kept inline with the folder structure, however this is pure convention and should not be relied upon since there is no guarantee that it is unique (unlike the file path).

That's why I didn't extract `name` either.

plugins/tedge_flows_plugin/src/remove.rs
```diff
@@ -0,0 +1,118 @@
...
+    keep_params: bool,
+) -> Result<(), FlowsPluginError> {
+    let flow_dir = mappers_dir
+        .as_ref()
+        .join(mapper_name)
+        .join("flows")
+        .join(flow_name);
+
+    info!("Removing flow {mapper_name}/{flow_name} from {flow_dir:?}",);
+    if !flow_dir.exists() {
```
Bravo555: thought: if `flow_dir` is a file, we proceed but error is printed:

```
2026-03-13T17:12:06.371101005Z  INFO tedge_flows_plugin::remove: plugins/tedge_flows_plugin/src/remove.rs:26: Removing flow local/evil-flow from "/etc/tedge/mappers/local/flows/evil-flow"
ERROR: Could not access /etc/tedge/mappers/local/flows/evil-flow: Not a directory (os error 20)
```

So it's probably fine

plugins/tedge_flows_plugin/src/lib.rs
```diff
@@ -0,0 +1,181 @@
...
+    };
+
+    Ok(())
+}
+
+fn parse_module_name(module: &str) -> Result<(&str, &str), FlowsPluginError> {
+    match module.splitn(2, '/').collect::<Vec<_>>().as_slice() {
+        [mapper_name, flow_name] => Ok((mapper_name, flow_name)),
+        _ => Err(FlowsPluginError::InvalidModuleName(module.to_owned())),
+    }
```
Bravo555: nit: could also do `split_once`, which would be bit shorter and not allocate
```suggestion
    module
        .split_once('/')
        .ok_or(FlowsPluginError::InvalidModuleName(module.to_owned()))
```

plugins/tedge_flows_plugin/src/install.rs
reubenmiller: We decided to not be so strict with a version mismatch and just show a warning. The artifact is the single-source of truth and should be believed over the version given in the command.

plugins/tedge_flows_plugin/src/list.rs
reubenmiller: We haven't been super strict at enforcing a specific "flow.toml", however we're trying to encourage some normalization, otherwise the organization gets messy very quickly (imo). Think of it like docker-compose.yaml files...technically you can use another name, but docker-compose.yaml clearly communicates intent.

plugins/tedge_flows_plugin/src/lib.rs
reubenmiller: This is just part of the [software management plugin api](https://thin-edge.github.io/thin-edge.io/references/software-management-plugin-api/#the-update-list-command) where the `update-list` is optional, and returning an exit code `1` says to the engine that this sm-plugin doesn't support the `update-list` command and should proceed with calling individual `install` / `remove` subcommands.

tests/RobotFramework/tests/plugin_flows/test_with_c8y.robot
reubenmiller: The test is using a suite setup so the tests are run in sequence, so the 1.0.0 version is installed in the previous test case.

Bravo555: I'm not understanding something. In previous test we install to `local/flows/hello-flow`, but in this test we install to `local/flows/world-flow`

plugins/tedge_flows_plugin/src/install.rs
Bravo555: Since there's only one possible version in the tar archive, and the version argument doesn't abort if the version doesn't match, are there any other planned uses for it, or is this warning the only one?

tests/RobotFramework/tests/plugin_flows/test_with_c8y.robot
rina23q: Actually world-flow is a mistake, at very first version I used the flow but I decided to remove it and replace with hello-flow. So this test has to be fixed

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile_operation.robot
```diff
@@ -206,6 +206,7 @@ Custom Setup
     ${DEVICE_SN}=    Setup
     Set Suite Variable    $DEVICE_SN
     Device Should Exist    ${DEVICE_SN}
+    Execute Command    tedge config set software.plugin.default apt
```
albinsuresh, line 209: Since we are being forced to make this change everywhere, or set the default plugin explicitly, I wonder if existing production deployments (without the default explicitly set) would also start breaking due to this? Wondering if we should add some implicit defaults in the software management code to handle this backward compatibility.

plugins/tedge_flows_plugin/src/lib.rs
reubenmiller: I'm happy to hear other opinions about this point...but the main motivation was to use the filesystem to generate unique names rather than relying on a .name field which is easy to mess up

plugins/tedge_flows_plugin/src/install.rs
reubenmiller: A warning is fine for now.

plugins/tedge_flows_plugin/src/install.rs
reubenmiller: Do we need a temp directory at all? Couldn't we extract the files directly into the target folder?

tests/RobotFramework/tests/plugin_flows/test_with_c8y.robot
```diff
@@ -0,0 +1,64 @@
...
+*** Keywords ***
+Custom Setup
+    ${DEVICE_SN}=    Setup
+    Device Should Exist    ${DEVICE_SN}
+    Transfer To Device    ${CURDIR}/default-flow    /etc/tedge/mappers/local/flows/
+
+    ${V1_URL}=    Cumulocity.Create Inventory Binary
+    ...    local/hello-flow
+    ...    package
+    ...    file=${CURDIR}/hello-flow_1.0.0.tar.gz
```
albinsuresh, line 112: It would have been nicer if the flow files were checked-into the repo instead of these `tar.gz` files, and do the packaging into `tar.gz` as part of the this test setup so that it's easier for others to figure out the contents of this flow, without having to download and untar the binary. Or just repliicte the same that is done in the previous `test_plugin.robot`

tests/RobotFramework/tests/plugin_flows/hello-flow_1.0.0/flow.toml
```diff
@@ -0,0 +1,9 @@
+# meta information about the flow
+name = "hello-flow"
```
albinsuresh, line 2: But this `name` metadata feels like a source of inconsistency and confusion to me. I had the same `hello-flow_2.0.0.tar.gz` installed with 2 different names on the device: as 1) `local/hello-flow` and 2) `local/world-flow`.

Now the output of `tedge flows list` and the plugin `/etc/tedge/flow list` command are not consistent as follows:

```
# tedge flows list
2026-03-16T09:17:14.072076022Z  WARN tedge_flows::params: UnknownParam { path: ".params.call" }
2026-03-16T09:17:14.073040678Z  WARN tedge_flows::params: UnknownParam { path: ".params.call" }
Flow        : hello-flow (v2.0.0)
Description : none
File        : /etc/tedge/mappers/local/flows/hello-flow/flow.toml
Step 1      : /etc/tedge/mappers/local/flows/hello-flow/main.js

Flow        : hello-flow (v2.0.0)
Description : none
File        : /etc/tedge/mappers/local/flows/world-flow/flow.toml
Step 1      : /etc/tedge/mappers/local/flows/world-flow/main.js

Flow        : default-flow (v1.0.0)
Description : none
File        : /etc/tedge/mappers/local/flows/default-flow/flow.toml
Step 1      : /etc/tedge/mappers/local/flows/default-flow/main.js
```

```
# /etc/tedge/sm-plugins/flow list
c8y/alarms	0.0.0
c8y/events	0.0.0
c8y/health	0.0.0
c8y/measurements	0.0.0
c8y/units	0.0.0
local/default-flow	1.0.0
local/hello-flow	2.0.0
local/world-flow	2.0.0
```

Ignore that missing `c8y/...` flows, as they'll be handled spearately. But look at the names reported. The former shows the name defined in the `flow.toml` whereas the latter picks up the folder name.

It feels like dropping the `name` from the metadata and fully relying on the folder name would be easier.

plugins/tedge_flows_plugin/src/lib.rs
didier-wenzek: > I'm happy to hear other opinions about this point...but the main motivation was to use the filesystem to generate unique names rather than relying on a .name field which is easy to mess up

I get that. But then it's confusing to have an alternative name is the toml file that is ignored by `flow-sm-plugin` but used by the flow engine to report status and errors.

plugins/tedge_flows_plugin/src/install.rs
Bravo555: > Couldn't we extract the files directly into the target folder?

Most likely yes, but then 1) flow installation is not atomic (what if there's problems installing some files but not others) but we might not care about that since it should be easy to handle by just deleting the entire directory and reporting failure, and 2) we get more fsnotify events and log noise, which with current design we're sort of forced to handle sequentially, but I have some ideas for how to address that on the fsnotify level once I'm done with current stuff.

Still, I think just using `tempfile_in` would be a simple and sufficient fix.

tests/RobotFramework/tests/plugin_flows/test_with_c8y.robot
rina23q: I wish I could do, however, I realized that `Cumulocity.Create Inventory Binary` can only support uploading a file from the host machine. Cannot work with the path of a test container.

Exactly because of the visibility of the contents, I archive flows in `test_plugin.robot` setup.

plugins/tedge_flows_plugin/src/lib.rs
rina23q: We discussed it offline. The conclusion is summarized https://github.com/thin-edge/thin-edge.io/pull/4037#issuecomment-4074367221

plugins/tedge_flows_plugin/src/list.rs
rina23q: The metadata `name` won't be used as a name. We discussed it offline. The conclusion is summarized https://github.com/thin-edge/thin-edge.io/pull/4037#issuecomment-4074367221

plugins/tedge_flows_plugin/src/list.rs
rina23q: Opened to any `.toml` name except `params.toml`. We discussed it offline. The conclusion is summarized https://github.com/thin-edge/thin-edge.io/pull/4037#issuecomment-4074367221

plugins/tedge_flows_plugin/src/list.rs
rina23q: I think it's fixed by the new design and code 8a404e1dace4c11001ae4da89e2b9ba779f29b9b

tests/RobotFramework/tests/plugin_flows/test_with_c8y.robot
rina23q: Fixed in 8a404e1dace4c11001ae4da89e2b9ba779f29b9b

plugins/tedge_flows_plugin/src/list.rs
rina23q: Now the name is opened to any `.toml` file except `params.toml`. Summarized https://github.com/thin-edge/thin-edge.io/pull/4037#issuecomment-4074367221

plugins/tedge_flows_plugin/src/list.rs
rina23q: Very good catch. Fixed in https://github.com/thin-edge/thin-edge.io/pull/4037/changes/3d5d2cbaf1ac1bf28ef0df772ba06d8559c51f81.

plugins/tedge_flows_plugin/src/list.rs
rina23q: I guess this comment became invalid after using blob patterns instead of recursive search.

plugins/tedge_flows_plugin/src/install.rs
rina23q: Applied `tempfile_in` in 8a404e1dace4c11001ae4da89e2b9ba779f29b9b

plugins/tedge_flows_plugin/src/install.rs
rina23q: Removed: 8a404e1dace4c11001ae4da89e2b9ba779f29b9b

plugins/tedge_flows_plugin/src/lib.rs
rina23q: Applied in 8a404e1dace4c11001ae4da89e2b9ba779f29b9b

tests/RobotFramework/tests/tedge_agent/device_profile/device_profile_operation.robot
rina23q: As we discussed it offline, user should put softwareType in c8y. This field is available since long time.

tests/RobotFramework/tests/plugin_flows/test_with_c8y.robot
rina23q: I changed a bit for the clarity and remove the duplication. I keep only `test_with_c8y.robot` and create `hello-flow_1.0.0` and `hello-flow_2.0.0` directories. So, at least everyone can see what was the original folders for the compressed tarballs.

tests/RobotFramework/tests/plugin_flows/hello-flow_1.0.0/flow.toml
rina23q: I also prefer dropping `name` from the metadata, otherwise, adding `name` field to the `tedge flows list` output. Should be addressed in a different ticket & PR.

plugins/tedge_flows_plugin/src/install.rs
```diff
@@ -0,0 +1,301 @@
...
+        if version != version_from_file {
+            warn!(
+                "Provided version {version} does not match version {version_from_file} in flow.toml for flow {mapper_name}/{flow_name}."
+            );
+        }
+    }
+
+    info!("Successfully installed flow {mapper_name}/{flow_name} from archive {file_path}");
+
+    Ok(())
```
Bravo555, line 78: suggestion(style,optional): Looking now, we're using this `{mapper_name}/{flow_name}` construct so much that we could make it a struct with a display impl so that we don't have to keep repeating it. Or we could reuse `list::FlowEntry`, which additionally has a `version` field,  and then the function has 3 args instead of 5. Up to you.

plugins/tedge_flows_plugin/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+        }
+        PluginOp::UpdateList => return Err(FlowsPluginError::InvalidUsage),
+        PluginOp::Prepare => {}
+        PluginOp::Finalize => {}
+    };
+
+    Ok(())
+}
+
+fn parse_module_name(module: &str) -> Result<(&str, &str), FlowsPluginError> {
```
Bravo555: issue: Maybe I'm too paranoic, but there is a potential path traversal here.

Because we're not checking if module name doesn't contain any more `/` characters, a command like `tedge-flows-plugin remove local/onstartup-test-order/../../../../../../test` attempts to remove directory `/test`. Likewise, `tedge-flows-plugin install --file ~/hello.tar local/hello/../../../../../../hello` happily creates a directory `/hello`.

Since I understand flows could be nested, like `local/hello/world`, restricting `module_name` to not contain any `/` wouldn't work. Then we could use either `path-clean` approach from the File Transfer Service, but I think here we could try the [`openat`](https://val.packett.cool/blog/use-openat/#and-back-to-the-regular-syscall-interfaces) approach where we get a stable reference to `/etc/tedge/mappers/local/flows` inode, and then we're guaranteed to only be able to make changes in this directory.

I admit that this problem is probably unlikely to happen in practice, and I'm not sure about the threat model (user should control these arguments and also `tedge-flows-plugin` would probably run with `tedge` permissions), but still, it feels like the behaviour is wrong, as clearly we only want to write inside the mappers directory.

I'd be happy to hear your opinions on this.

EDIT: It seems getting stable references to directories is still experimental in `std`, so we'd have to add [new dependencies](https://github.com/bytecodealliance/cap-std#cap-std) to use the `openat` approach.

plugins/tedge_flows_plugin/src/install.rs
```diff
@@ -0,0 +1,301 @@
...
+    let dest = mapper_dir.join(mapper_name).join("flows").join(flow_name);
+    let params_path = dest.join(PARAMS_FILE);
+
+    // Create a temporary directory in mappers dir so that inotify doesn't trigger on the temp dir creation but only on the final move.
+    let tmp_dir = tempfile::tempdir_in(mapper_dir).map_err(|e| io_error(mapper_dir, e))?;
+    let tmp_dest =
+        Utf8PathBuf::try_from(tmp_dir.path().to_owned()).expect("tempdir path is valid UTF-8");
+    let params_backup = tmp_dest.join(PARAMS_FILE);
+
+    unpack_archive(target, file_path, &tmp_dest)?;
```
Bravo555, line 42: note: just fyi, we're not checking if the archive contains a flow, so if we install something that is not a flow by mistake, `list` command doesn't list it, so a user may get confused "i just installed a flow, why isn't it showing up?". Admittedly it's probably easy to find that mistake.

plugins/tedge_flows_plugin/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// List all installed flows
+    List,
+
+    /// Install a flow from a .tar or .tar.gz archive
+    Install {
+        /// Name of flow in the form <mapper>/<flow-name>
+        module: String,
+
+        #[clap(short = 'v', long = "module-version")]
+        version: Option<String>,
```
Bravo555, line 52: suggestion: Myself I was a bit confused what's the point of the version argument, so maybe it should be documented to make it clear.

reubenmiller: Its just part of the [software management plugin interface](https://thin-edge.github.io/thin-edge.io/references/software-management-plugin-api/#the-install-command)...but being more descriptive on flags is never a bad thing

plugins/tedge_flows_plugin/src/install.rs
reubenmiller: Having some "is it a flow" check would be helpful. We could keep it to a minimal and just check for at least one valid flow definition file (.toml), and we could limit what we deem "valid" to "is the toml file able to be read by our schema" (if doing a full check if not easy to implement)

tests/RobotFramework/tests/plugin_flows/hello-flow_1.0.0/flow.toml
reubenmiller: The .name is part of an interface that we need to align for as part of the cloud side of things.

Personally I would view the "name" as the "instance name", as the instance (deployment) name is dictated by the location of the flow.

tests/RobotFramework/tests/plugin_flows/test_plugin.robot
reubenmiller: I can confirm it works nicely, plus it is awesome that you don't need to restart any services, as the flow sm-plugin will read the tedge.toml file itself when it is launched.

plugins/tedge_flows_plugin/src/lib.rs
```diff
@@ -0,0 +1,183 @@
...
+    },
+
+    /// Install or remove multiple modules at once (not supported)
+    UpdateList,
+
+    /// Prepare a sequences of install/remove commands (not used)
+    Prepare,
+
+    /// Finalize a sequences of install/remove commands (not used)
+    Finalize,
```
didier-wenzek: I don't see these help messages as helpful. There are even a bit confusing.

I appreciate that these commands are parts of the sm-plugin API. Can we convey this better?


```suggestion
    /// Install or remove multiple modules at once (sm-plugin API): not supported
    UpdateList,

    /// Prepare a sequences of install/remove commands (sm-plugin API): do nothing
    Prepare,

    /// Finalize a sequences of install/remove commands (sm-plugin API): do nothing
    Finalize,
```

plugins/tedge_flows_plugin/src/lib.rs
reubenmiller: Though if we were to parse the name first and forbid characters relative paths within the name, then that should be sufficient for now.

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,351 @@
...
+/// Collect all flows from a mapper's flows directory using glob patterns.
+///
+/// Name rules:
+/// - `flows_dir/hello.toml`             => name `hello`
+/// - `flows_dir/hello/flow.toml`        => name `hello`
+/// - `flows_dir/hello/world.toml`       => name `hello/world`
+/// - `flows_dir/hello/world/flow.toml`  => name `hello/world`
+/// - `flows_dir/flow.toml`              => forbidden (skipped with a warning)
+fn list_mapper_flows(flows_dir: &Utf8Path, mapper_name: &str, flows: &mut Vec<FlowEntry>) {
+    // Pass 1: find all flow.toml files
```
didier-wenzek: Why not a single pass over `"{flows_dir}/**/*.toml"` and a method that return a `FlowEntry` from a `(flow_dir, path)` pair? I think this would make the naming rules clearer.

plugins/tedge_flows_plugin/src/list.rs
```diff
@@ -0,0 +1,351 @@
...
+            .file("flow.toml")
+            .with_toml_content(toml::toml! { version = "2.0" });
+
+        let flows = retrieve_flows(ttd.utf8_path());
+        // Sorted by (mapper, name, version): 1.0 < 2.0
+        assert_eq!(
+            entries(&flows),
+            [("local", "hello", "1.0"), ("local", "hello", "2.0")]
+        );
+    }
```
didier-wenzek, line 319: This is the only point I disagree with the new naming convention https://github.com/thin-edge/thin-edge.io/pull/4037#issuecomment-4074367221. Assigning the same name to two different flows will be a source of trouble. This corner case will rarely occur if ever, I agree, but if so it will be a pain to detect and solve.

I would prefer to use distinct names (say `local/flows/hello/flow.toml` => `local/hello/flow` or `local/hello/`) and to accept shorter names from users when there is no ambiguities.

plugins/tedge_flows_plugin/src/install.rs
rina23q: Yes. agree that "is it a valid flow" check is necessary. Actually, this check also requires for `list`. For
example, if a toml file is having only `version` field like the test here, `tedge flows list` gives an error message for missing field `input`. Another inconsistency. So I'll work on it.

plugins/tedge_flows_plugin/src/install.rs
```diff
@@ -0,0 +1,301 @@
...
+    let dest_parent = dest.parent().expect("dest always has a parent");
+    fs::create_dir_all(dest_parent).map_err(|e| io_error(dest_parent, e))?;
+
+    match fs::remove_dir_all(&dest) {
+        Ok(_) => debug!("Removed existing flow directory at {dest}"),
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            debug!("No existing flow directory at {dest}")
+        }
+        Err(e) => return Err(io_error(&dest, e)),
+    }
```
didier-wenzek, line 62: Do we have to remove the old flow directory content before renaming the temporary flow to be the actual flow? I don't think so.

plugins/tedge_flows_plugin/src/install.rs
didier-wenzek: > Yes. agree that "is it a valid flow" check is necessary. Actually, this check also requires for `list`. For example, if a toml file is having only `version` field like the test here, `tedge flows list` gives an error message for missing field `input`. Another inconsistency. So I'll work on it.

To avoid too many dependencies, an option is to invoke `tedge flows list`. We can also improve that command to take a path as an argument to only check that flow.

By the way, we can have this check done by the `Finalize` sub command of the plugin.

plugins/tedge_flows_plugin/src/install.rs
rina23q: It is necessary, `fs::rename()` will fail when the `dest` is a non-empty directory.

plugins/tedge_flows_plugin/src/install.rs
rina23q: Done in https://github.com/thin-edge/thin-edge.io/pull/4037/changes/06e8fe943069bbf42b8f94b69f7c765eff66c34d. Added `FlowRecord` struct that implements `Display`.

plugins/tedge_flows_plugin/src/lib.rs
rina23q: Added a check and a system test in https://github.com/thin-edge/thin-edge.io/pull/4037/changes/06e8fe943069bbf42b8f94b69f7c765eff66c34d

plugins/tedge_flows_plugin/src/lib.rs
rina23q: Yes, the plugin must have the argument even if it's not used because of the sm-plugin contract.

I updated the help outputs as suggested https://github.com/thin-edge/thin-edge.io/pull/4037#discussion_r2958704383 by the commit https://github.com/thin-edge/thin-edge.io/pull/4037/changes/06e8fe943069bbf42b8f94b69f7c765eff66c34d.

plugins/tedge_flows_plugin/src/lib.rs
rina23q: Applied in https://github.com/thin-edge/thin-edge.io/pull/4037/changes/06e8fe943069bbf42b8f94b69f7c765eff66c34d. I marked `(sm-plugin API)` to all subcommands/arguments that are mandatory by sm-plugin specifications.

plugins/tedge_flows_plugin/src/list.rs
rina23q: Yes, you have a point. Changed in https://github.com/thin-edge/thin-edge.io/pull/4037/changes/06e8fe943069bbf42b8f94b69f7c765eff66c34d.

plugins/tedge_flows_plugin/src/install.rs
rina23q: Added a validation in fb7b092c20443670c1b6c6fbe024e77af51d1c71. Waiting for a change of the exit code when `tedge flows list` finds an error or nothing to list.

didier-wenzek (approved):
Approved. A great addition that will support users writing, sharing and deploying domain specific flows. Thank you.

plugins/tedge_flows_plugin/src/install.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    Ok(())
+}
+
+fn validate_flow(tmp_dest: &Utf8Path) -> Result<(), FlowsPluginError> {
+    debug!("Validating flow at {tmp_dest} by running `tedge flows list --flows-dir {tmp_dest}`");
+    let output = match std::process::Command::new("tedge")
+        .args(["flows", "list", "--flows-dir"])
+        .arg(tmp_dest.as_str())
+        .output()
```
Bravo555, line 88: thought: A bit annoying that we have to spawn a process instead of directly using the same function that `tedge flows list` does, but there shouldn't be a problem either way.

reubenmiller: Calling the binary is ok, though we need to make sure we pass the loaded config_dir (that the agent is using), and pass it to the `tedge flows list` command, e.g.

```sh
tedge flows list --flows-dir "foo" --config-dir /tmp/tedge
```

rina23q: Commited here: 8be199604cbfcec448f1f8224a1080a8fcea2e6e

Issue #4028, opened 2026-03-10
flow hot reloading fails after a while with a JS runtime error


reubenmiller:
**Describe the bug**

Reloading a tedge flow multiple times eventually results in a JS exception which causes the flow to fail due to a "JS raised exception"

Below shows the log entries for the `tedge-mapper local` command where an example flow was forced to be reloaded multiple times (by just touching it), and eventually the tedge-mapper rejects the flow. After this point, it seems that the flow is never reloaded successfully until the service is restarted.

```log
$ tedge-mapper local

2026-03-10T17:09:22.897284Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:23.824232Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:24.621738Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:25.147655Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:25.606933Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:26.098071Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:26.589867Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:27.115207Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:27.579312Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:28.012198Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
2026-03-10T17:09:28.457384Z  INFO flows: Loading flow /opt/homebrew/etc/tedge/mappers/local/flows/debugger/flow.toml
[...]

PR #4039, opened 2026-03-12
fix: memory leak on js module reload

didier-wenzek:
## Proposed changes

As rquickjs fails to drop old module versions (`use rquickjs::Module` doesn't implement `Drop`), a fresh new `rquickjs::AsyncRuntime` has to be created when a js module is reloaded.

This PR is inspired by https://github.com/thin-edge/thin-edge.io/pull/4034 which addresses the main problem (rquickjs fails to drop old module versions) with a working solution (i.e. to re-create the rquickjs runtime), but with an over-complicated implementation.

- [x] fix the memory leak on module reload
- [x] return an explicit error message when JS reach out of memory
- [x] add tedge config setting to control JS memory limits  

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/4028

## Checklist


crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -721,6 +721,42 @@ export function onMessage(message) {
...
 
+    /// Regression test for https://github.com/thin-edge/thin-edge.io/issues/4028
+    ///
+    /// Repeatedly reloading a flow script used to exhaust the QuickJS 16 MB
+    /// runtime memory limit because every reload left a JSModuleDef object in
+    /// ctx->loaded_modules (rquickjs Module has no Drop impl that calls
+    /// JS_FreeModule). The fix recreates the AsyncContext on each hot-reload so
+    /// that JS_FreeContext releases all accumulated module memory.
+    #[tokio::test]
+    async fn repeated_hot_reload_does_not_exhaust_js_memory() {
```
didier-wenzek, line 756: This test has been copied without modification from https://github.com/thin-edge/thin-edge.io/pull/4034

crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -98,6 +106,28 @@ impl JsRuntime {
...
+    ) -> Result<Vec<&'static str>, LoadError> {
+        if self.module_sources.remove(&name).is_some() {
+            // As rquickjs fails to drop old module versions,
+            // a new worker has to be created with fresh new Async Runtime & Context
+            self.runtime = Self::new_runtime().await?;
+            let context = rquickjs::AsyncContext::full(&self.runtime).await?;
+            self.worker = JsWorker::spawn(context, self.store.clone()).await;
+            for (n, s) in &self.module_sources {
+                self.load_new_js(n.to_owned(), s.clone()).await?;
+            }
```
didier-wenzek: This is the fix. Commenting out these lines (as well as the `assert` line 277) makes fail the test `repeated_hot_reload_does_not_exhaust_js_memory`  with an OOM

crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -21,32 +21,40 @@ pub struct JsRuntime {
...
             worker,
             execution_timeout,
+            module_sources,
         })
     }
 
+    async fn new_runtime() -> Result<rquickjs::AsyncRuntime, LoadError> {
+        let runtime = rquickjs::AsyncRuntime::new()?;
+        runtime.set_memory_limit(16 * 1024 * 1024).await;
+        runtime.set_max_stack_size(256 * 1024).await;
```
didier-wenzek: I consider to add `tedge config` settings for those two limits.

reubenmiller: Yeah that would be great to give more control over how much can be allocated.

crates/extensions/tedge_flows/src/js_runtime.rs
reubenmiller: Can we somehow provide a more meaningful error when the OOM prevents the loading of the module? Or does rquickjs not support this? Currently as in the linked ticket, the only info we get is "JS raised exception" which isn't very obvious for users.

didier-wenzek: Yeah, that would be nicer.

didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/4039/changes/472846038f1a2082f72de028782f15ffedc8317e

crates/extensions/tedge_flows/src/js_runtime.rs
didier-wenzek: Added https://github.com/thin-edge/thin-edge.io/pull/4039/commits/84a253c5c3bb0e84bd0ad980d4885cdf16e88422

crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -721,9 +745,43 @@ export function onMessage(message) {
...
+
+        // First load
+        runtime
+            .load_script_literal(&mut script, large_js.as_bytes().to_vec())
+            .await
+            .expect("initial load should succeed");
+
+        // Reload the same large script many times — without the fix this would
+        // fail around reload 500 with "JS raised exception" due to OOM.
+        for i in 1..=1000 {
```
didier-wenzek: Now that the memory allocated to the JS runtime can be reduced, this test can be improved to take less time.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -1367,6 +1367,16 @@ define_tedge_config! {
     },
 
     flows: {
+        memory: {
+            /// The maximum number of bytes allocated to the JS runtime
+            #[tedge_config(example = "16777216", default(value = 16777216u32))]
+            heap_size: u32,
```
jarhodes314, line 1373: I wonder whether given we now support `humantime` for times, it makes sense to use a crate like `bytesize` for this so we can specify the amount as `"16KiB"` instead of a random looking number of bytes.

didier-wenzek: I will do that in a follow up PR. It's better to fix the memory leak asap.

Issue #4043, opened 2026-03-13
incorrect tedge config name for apt.dpk.options.config


reubenmiller:
**Describe the bug**

The tedge config has a typo in the key name, `apt.dpk.options.config` which makes the setting harder for users to find as it uses unconventional shortform, "dpk" instead of "dpkg".

"dpkg" is the official Debian package binary (`dpkg`) that the configuration option is related to, so it should align with the name so it is easier for users to find.

For example, this should of matched the dpkg related config items:

```sh
tedge config list dpkg
```

**To Reproduce**

The unexpected config name can be seen in the `tedge config list` command:

```sh
tedge config list apt.dpk.options.config --doc
```

**Expected behavior**

The following changes are recommended:

* Deprecate `apt.dpk.options.config` (as to not break existing settings)
* Rename `apt.dpk.options.config` to `apt.dpkg.options.config`

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `any`
 - Hardware [incl. revision]: `any`
 - System-Architecture [e.g. result of "uname -a"]: `any`
 - thin-edge.io version [e.g. 0.1.0]: `tedge 1.7.2~358+gf3ff860`

**Additional context**

PR #4044, opened 2026-03-13
fix: rename tedge config property `apt.dpk.options.config` to `apt.dpkg.options.config`

reubenmiller:
## Proposed changes

Rename tedge config property `apt.dpk.options.config` to `apt.dpkg.options.config`, so that the `dpkg` related settings are easier for users to find.

The old value is marked as deprecated so it won't break anyones setup, but it will generate a warning marking that the config item has been renamed.

```sh
# old name still works
tedge config set apt.dpk.options.config keepold
```

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4043

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config/version.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config/version.rs
@@ -1,4 +1,4 @@
-//! Versioning and migrations of the main config file.
+// Versioning and migrations of the main config file.
 
 use std::borrow::Cow;
 
@@ -22,7 +22,7 @@ impl TryFrom<String> for TEdgeTomlVersion {
         match value.as_str() {
             "1" => Ok(Self::One),
             "2" => Ok(Self::Two),
-            _ => todo!(),
+            _ => Err(format!("Unknown tedge.toml version: {value}")),
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tests/RobotFramework/tests/tedge/tedge_completions.robot
```diff
@@ -13,7 +13,7 @@ ${thing}=
 ...             SEPARATOR=
 ...             apt.name\tThe filtering criterion that is used to filter packages list output by name\n\
 ...             apt.maintainer\tThe filtering criterion that is used to filter packages list output by maintainer\n\
-...             apt.dpk.options.config\tdpkg configuration option used to control the dpkg options "--force-confold" and "--force-confnew" and are applied when installing apt packages via the tedge-apt-plugin. Accepts either 'keepold' or 'keepnew'.
+...             apt.dpkg.options.config\tdpkg configuration option used to control the dpkg options "--force-confold" and "--force-confnew" and are applied when installing apt packages via the tedge-apt-plugin. Accepts either 'keepold' or 'keepnew'.
```
rina23q, line 16: (out of scope) It's a bit odd that only this tab completion test is affected by the change. This means we don't have a system test that is really using the config key.

reubenmiller: It seems we don't have a system test to cover this. I assume as it is a tricky one to create a test for.

reubenmiller: wait, I just found a test that I forgot to commit :/

reubenmiller: But it is good that we just tested the backwards compatibility :)

reubenmiller: This is the test showing coverage, tests/RobotFramework/tests/plugin_apt/install_overwrite_config_files.robot.

Though the test was still passing because we just flagged the property name as being deprecated (not removed).

Issue #3855, opened 2025-11-07
add device cert common name validation logic when connecting to Cumulocity


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

If someone creates a device certificate outside of thin-edge.io with a Common Name that is not Cumulocity compliant, then the device will fail to connect to Cumulocity, but the error message is not obvious to the user that the Common Name is valid.

**Describe the solution you'd like**

Add a warning message which is printed during the `tedge connect c8y` command that warns the user when an invalid Common Name is detected (the same validation which is already done when using `tedge cert create --device-id "myinvalid:name"`)

The validation is already present in the `tedge cert create c8y` command, however if users are creating their own device certificates (via some custom PKI integration), then they don't go through the same validation process, and hence it makes it harder for the users to spot the invalid common name.

```sh
$ tedge cert create c8y --device-id "myinvalid:name"
Error: failed to create a test certificate for the device myinvalid:name.

Caused by:
    0: DeviceID Error
    1: The string '"myinvalid:name"' contains characters which cannot be used in a name [use only A-Z, a-z, 0-9, ' = ( ) , - . ? % * _ ! @]
```

**Describe alternatives you've considered**

**Additional context**

PR #4052, opened 2026-03-16
feat: warn users about invalid certificate common name for Cumulocity connections

reubenmiller:
## Proposed changes

Improve the warning for users that have generated the device certificate using external command (e.g. not `tedge cert create`). In this scenerio the user doesn't get warning during the cert creation, so they are unaware of the problem.

Now the `tedge connect c8y` will display a warning if an invalid Common Name is detected at connection time, though it will not prevent trying to connect to Cumulocity (in case if this restriction is lifted at some point).

It is only valid for Cumulocity connections as Cumulocity has a defined list of allowed characters in the Common Name.

Below is an example of the warning that is shown to users:

```
# tedge connect c8y
Connecting to Cumulocity:
        mapper configuration file: /etc/tedge/tedge.toml
        device id: TST_develop_chalky_bullet:invalid
        cloud profile: <none>
        cloud host: example.cumulocity.com:8883
        auth type: Certificate
        cryptoki: off
        bridge: mosquitto
        service manager: systemd
        mosquitto version: 2.0.11
        proxy: Not configured
warning: The device certificate Common Name may not be accepted by the cloud: The string '"TST_develop_chalky_bullet:invalid"' contains characters which cannot be used in a name [use only A-Z, a-z, 0-9, ' = ( ) , - . ? % * _ ! @]
Creating device in Cumulocity cloud... ✗
error: Connection error while creating device in Cumulocity: Mqtt state: Mqtt serialization/deserialization error: IO: received fatal alert: CertificateUnknown: Mqtt serialization/deserialization error: IO: received fatal alert: CertificateUnknown: IO: received fatal alert: CertificateUnknown: received fatal alert: CertificateUnknown
```

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3855

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/connect/command.rs
+++ b/crates/core/tedge/src/cli/connect/command.rs
@@ -28,6 +28,7 @@ use anyhow::bail;
 use c8y_api::http_proxy::read_c8y_credentials;
 use camino::Utf8PathBuf;
 use certificate::parse_root_certificate::CryptokiConfig;
+use certificate::PemCertificate;
 use mqtt_channel::Topic;
 use std::borrow::Cow;
 use std::collections::HashMap;
@@ -146,6 +147,9 @@ impl Command for ConnectCommand {
         );
 
         validate_config(&tedge_config, &self.cloud)?;
+        if matches!(self.cloud, Cloud::C8y(_)) {
+            warn_if_cert_cn_invalid(&bridge_config);
+        }
 
         if self.is_test_connection {
             self.check_bridge(&tedge_config, &bridge_config)
@@ -520,6 +524,33 @@ impl ConnectCommand {
             );
         }
     }
+}
+
+fn warn_if_cert_cn_invalid(bridge_config: &BridgeConfig) {
+    if let Err(e) = check_cert_cn(bridge_config) {
+        warning!("{e}");
+    }
+}
+
+fn check_cert_cn(bridge_config: &BridgeConfig) -> anyhow::Result<()> {
+    if bridge_config.auth_type != AuthType::Certificate {
+        return Ok(());
+    }
+    let cert = PemCertificate::from_pem_file(&bridge_config.bridge_certfile).map_err(|e| {
+        anyhow::anyhow!(
+            "Failed to parse device certificate {}: {e}",
+            bridge_config.bridge_certfile
+        )
+    })?;
+    let cn = cert.subject_common_name().map_err(|e| {
+        anyhow::anyhow!(
+            "Failed to read Common Name from device certificate {}: {e}",
+            bridge_config.bridge_certfile
+        )
+    })?;
+    certificate::device_id::is_valid_device_id(&cn, 64).map_err(|e| {
+        anyhow::anyhow!("The device certificate Common Name may not be accepted by the cloud: {e}")
+    })
 }
 
 fn validate_config(config: &TEdgeConfig, cloud: &MaybeBorrowedCloud<'_>) -> anyhow::Result<()> {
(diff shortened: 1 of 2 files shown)


crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -522,6 +526,27 @@ impl ConnectCommand {
...
+        return Ok(());
+    }
+    let Some(cn) = PemCertificate::from_pem_file(&bridge_config.bridge_certfile)
+        .ok()
+        .and_then(|cert| cert.subject_common_name().ok())
+    else {
+        return Ok(());
+    };
+    certificate::device_id::is_valid_device_id(&cn, 64).map_err(|e| {
+        anyhow::anyhow!("The device certificate Common Name may not be accepted by the cloud: {e}")
```
albinsuresh, line 552: If this is a hard restriction for C8Y currently and there proceeding with a warning is eventually going to fail anyway, when it tries to establish the connection, why not fail here? Esp since we are doing this check only for C8Y anyway.

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -146,6 +147,9 @@ impl Command for ConnectCommand {
         );
 
         validate_config(&tedge_config, &self.cloud)?;
+        if matches!(self.cloud, Cloud::C8y(_)) {
```
albinsuresh: I'd move that `auth_type` check from `check_cert_cn` here:
```suggestion
        if bridge_config.auth_type == AuthType::Certificate && matches!(self.cloud, Cloud::C8y(_)) {
```

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -522,6 +526,27 @@ impl ConnectCommand {
...
+
+fn check_cert_cn(bridge_config: &BridgeConfig) -> anyhow::Result<()> {
+    if bridge_config.auth_type != AuthType::Certificate {
+        return Ok(());
+    }
+    let Some(cn) = PemCertificate::from_pem_file(&bridge_config.bridge_certfile)
+        .ok()
+        .and_then(|cert| cert.subject_common_name().ok())
+    else {
+        return Ok(());
```
albinsuresh: Is this intentional to skip any failures while parsing the certificate or when the common name can not be retrieved from it?

crates/core/tedge/src/cli/connect/command.rs
albinsuresh, line 150: I'd move that `auth_type` check from `check_cert_cn` here:
```suggestion
        if bridge_config.auth_type == AuthType::Certificate && matches!(self.cloud, Cloud::C8y(_)) {
```

crates/core/tedge/src/cli/connect/command.rs
reubenmiller: The reason for not failing hard as such name restrictions can always change so we should always at least try the connection. It is only intended as a hint to help users out.

crates/core/tedge/src/cli/connect/command.rs
reubenmiller: Yeah it was but I guess we can add an error here as well though I wouldn't fail because of this as our connection logic should be the main driver determining if the certificate is valid.

crates/core/tedge/src/cli/connect/command.rs
didier-wenzek: > I'd move that `auth_type` check from `check_cert_cn` here:

This is a matter of taste. It makes sense to group all the checks in a single function as of now.

crates/core/tedge/src/cli/connect/command.rs
reubenmiller: Resolved by 6540dbb58b2ecfe4258d0b1b33b92e5c8872d049

crates/core/tedge/src/cli/connect/command.rs
reubenmiller: ok, then keeping it as is...at least this has the advantage for being a reusable piece of code where everything about the c8y specific logic (e.g. use base auth) is encapsulated in the validation function

Issue #3374, opened 2025-01-31
Permissions error in file-transfer service is treated as transient when it will never resolve itself


jarhodes314:
**Describe the bug**
When triggering a configuration snapshot where `tedge` does not have permissions to write to/create `/var/tedge/file-transfer/<device_id>/`, the file-transfer service generates a 500 internal server error. This appears to be treated as a transient error by the uploader code, so it attempts to retry the upload even though it is essentially guaranteed to fail without some human intervention.

**To Reproduce**
1. Delete `/var/tedge/file-transfer/<device_id>/`
2. `sudo chown root "/var/tedge/file-transfer/ && sudo chmod 755 /var/tedge/file-transfer`
3. Trigger a configuration-snapshot operation in Cumulocity

The operation gets stuck in the executing state. You can confirm there is a permissions error through observing the following:

```
$ systemctl status tedge-agent
Jan 31 17:45:21 james-Latitude-5511 tedge-agent[687003]:     Permission denied (os error 13)
Jan 31 17:45:21 james-Latitude-5511 tedge-agent[687003]: 2025-01-31T17:45:21.24358042Z  WARN upload::upload: Temporary failure: HTTP status server error (500 Internal Server Error) for url (http://127.0.0.1:8000/tedge/file-transfer/jrh-test.5/config_snapshot/tedge-configuration-plugin-c8y-mapper-141376). Retrying in 22s
Jan 31 17:45:43 james-Latitude-5511 tedge-agent[687003]: 2025-01-31T17:45:43.741483458Z ERROR tedge_agent::http_server::error: Request to upload to "jrh-test.5/config_snapshot/tedge-configuration-plugin-c8y-mapper-141376" failed: Directory Error. Check permissions for /var/tedge/file-transfer/jrh-test.5/config_snapshot.
Jan 31 17:45:43 james-Latitude-5511 tedge-agent[687003]: Caused by:
Jan 31 17:45:43 james-Latitude-5511 tedge-agent[687003]:     Permission denied (os error 13)
Jan 31 17:45:43 james-Latitude-5511 tedge-agent[687003]: 2025-01-31T17:45:43.741968038Z  WARN upload::upload: Temporary failure: HTTP status server error (500 Internal Server Error) for url (http://127.0.0.1:8000/tedge/file-transfer/jrh-test.5/config_snapshot/tedge-configuration-plugin-c8y-mapper-141376). Retrying in 36s
[...]

PR #4054, opened 2026-03-16
fix: don't retry http server 500/501 errors when uploading files

reubenmiller:
## Proposed changes

When uploading files, treat server errors (status codes, 500 and 501) as non-transient errors so they aren't retried as generally a retry won't help. Status Codes 502/503 are still retried as before as generally these errors are transient.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3374

## Checklist

### DIFF
--- a/crates/common/upload/src/upload.rs
+++ b/crates/common/upload/src/upload.rs
@@ -261,9 +261,15 @@ impl Uploader {
                 })?
                 .error_for_status()
                 .map_err(|err| match err.status() {
-                    Some(status_error) if status_error.is_client_error() => {
+                    // 4xx and 500/501 server-side errors are permanent and retrying won't help
+                    Some(status)
+                        if status.is_client_error()
+                            || status == reqwest::StatusCode::INTERNAL_SERVER_ERROR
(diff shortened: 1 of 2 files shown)


crates/common/upload/src/upload.rs
```diff
@@ -261,9 +261,15 @@ impl Uploader {
                 })?
                 .error_for_status()
                 .map_err(|err| match err.status() {
-                    Some(status_error) if status_error.is_client_error() => {
+                    // 4xx and 500/501 server-side errors are permanent and retrying won't help
+                    Some(status)
+                        if status.is_client_error()
```
albinsuresh, line 266: Even 429 is a transient error that might get resolved if retried later, right?

albinsuresh (approved):

didier-wenzek (approved):
Approved

crates/common/upload/src/upload.rs
reubenmiller: I don't think we should retry 429 status codes until we would respect the associated `Retry-After` header, but that would require more involved changes

Issue #4038, opened 2026-03-12
Supported operations for child devices aren't resent on mapper startup (inconsistent behaviour)


reubenmiller:
**Describe the bug**

It was observed that the supported operations of a child device (using the tedge-agent as a child device) were not published to Cumulocity.

This results in the child device having only limited functionality displayed in the Cumulocity Device Management UI (e.g. device profile tab and log tab were not displayed for the user).

It seems that the tedge-mapper-c8y does not republish the `114` SmartREST 2.0 message for child devices on startup-up, which makes it difficult for users to force a republishing of this message.

**To Reproduce**

The exact steps to reproduce this situation is not yet known, but I suspect either the child device was removed at some point in time (after the original `114` message was sent).

**Expected behavior**

On the tedge-mapper-c8y startup, it should resend supported operations of child devices and services

**Note**

* The tedge-mapper-c8y already resends the supported operations for the main device on startup, e.g. the below message was observed when running `systemctl restart tedge-mapper-c8y`.

    ```sh
    [c8y/s/us] 114,c8y_Command,c8y_DeviceProfile,c8y_DownloadConfigFile,c8y_Firmware,c8y_LogfileRequest,c8y_RemoteAccessConnect,c8y_Restart,c8y_SoftwareUpdate,c8y_UploadConfigFile
    ```

**Screenshots**

**Environment (please complete the following information):**

|Property|Value|
|--------|-----|
|OS [incl. version]|`Debian GNU/Linux 12 (bookworm)`|
|Hardware [incl. revision]|`unknown`|
|System-Architecture|`Linux tedge 6.8.0-64-generic #67-Ubuntu SMP PREEMPT_DYNAMIC Sun Jun 15 20:23:40 UTC 2025 aarch64 GNU/Linux`|
|thin-edge.io version|`tedge 1.7.1`|

**Additional context**

**Workaround (if using tedge >= 1.7.1)**

Just force a re-sync of the tedge-mapper-c8y to force the republishing of the 114 messages by running the following command:

```sh
tedge mqtt pub te/device/main/service/tedge-mapper-c8y/signal/sync '{}'
```

**Workaround (if using tedge <= 1.7.1)**

You can manually force the updating

1. Change directory into `/etc/tedge/operations/c8y/` (assuming you're using the default directory)

    ```sh
    cd /etc/tedge/operations/c8y/
    ```

2. Locate the directory which is matching the problematic child device

    ```sh
    ls -l
    ```

    **Example output**

[...]

PR #4064, opened 2026-03-20
fix: send supported_operations for child devices on startup

Bravo555:
## Proposed changes

Previously on startup we only sent supported_operations(114) smartrest message on startup for the main device, so if supported_operations for child device were lost somehow on Cumulocity (maybe accidentally deleted) we wouldn't resend operations of child devices until they changed or we manually emitted a sync signal.

Now we send supported_operations after startup for all child devices that have any supported operations.

## Types of changes

## Paste Link to the issue
#4038 

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
+++ b/tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
@@ -134,13 +134,8 @@ Recreate nested child supported operations after cloud deletion on mapper restar
     Log    ${mo}
 
     Delete Managed Object    ${DEVICE_SN}:device:childA
-    Delete Managed Object    ${DEVICE_SN}:device:child0
-    Delete Managed Object    ${DEVICE_SN}:device:child1
-    Delete Managed Object    ${DEVICE_SN}:device:child2
-    Delete Managed Object    ${DEVICE_SN}:device:child3
-    Delete Managed Object    ${DEVICE_SN}:device:child4
 
-    Cumulocity.Device Should Have A Child Devices
+    Cumulocity.Device Should Not Have Any Child Devices
 
     # we restart the mapper, devices should be re-registered and have the same hierarchy
     Start Service    tedge-mapper-c8y


crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1256,28 +1254,27 @@ impl CumulocityConverter {
     fn try_init_messages(&mut self) -> Result<Vec<MqttMessage>, ConversionError> {
         let mut messages = vec![self.c8y_agent_inventory_fragment()?];
 
-        // supported operations for the main device
-        let supported_operations_message =
-            self.load_and_create_supported_operations_messages(&self.config.device_id.clone())?;
+        // supported operations for all devices
+        let supported_operations_messages = self.load_and_create_supported_operations_messages()?;
+
+        // TODO: not sure if here we should also request pending operations for all devices, probably yes?
```
reubenmiller: No, the `500` (request for pending operations) SmartREST 2.0 message is only needed to be sent from the main device (as that is the agent), it should receive the operations for all child devices/services.

crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs
```diff
@@ -256,12 +257,43 @@ impl SupportedOperations {
...
+            .filter_map(|(device_xid, operations)| {
+                self.supported_operations_from_xid_and_operations(
+                    device_xid, operations, c8y_prefix,
+                )
+                .inspect_err(|err| error!(?err, "Failed to create operations message"))
+                .ok()
+            })
+            .collect();
+
+        messages.sort_by(|m1, m2| m1.topic.name.cmp(&m2.topic.name));
```
albinsuresh: For the case where the child devices are deleted in the cloud and we expect this fix to re-create those devices, I don't think it'll work for nested child devices. The operations directory at `/etc/tedge/operations/c8y` or the `operations_by_xid: HashMap<String, Operations>` does not differentiate between child devices and nested child devices. So, if we had a device hierarchy as follows:

```
main
|-- childA
|-- |-- child0
|-- |-- child1
```

the operations directory would be flat as follows:
```
/etc/tedge/operations/c8y
|-- childA
|-- child0
|-- child1
```

without the hierarchy information in it. So, if we go on to recreate the child devices from this purely using the sorted order, we'll get the hierarchy wrong (`child0` an `child1` will get re-created as immediate child devices).

IIRC, this was one of the reasons why we chose not to resend child-device inventory update messages on every startup. The other reason was to reduce the number of "bootstrap  messages" on every mapper startup for fleets with 100s or 1000s of child devices.

Bravo555: > The other reason was to reduce the number of "bootstrap messages" on every mapper startup for fleets with 100s or 1000s of child devices.

This is the 1st point, and, hm, it suggests the current behaviour shouldn't be changed at all, but then the problem of out-of-date operations would remain. If Cumulocity can't ask for operations for devices it's missing, and if we don't want to send redundant messages, then it seems we can't do anything. We want to keep state in sync on both sides, but don't have any lean mechanism to determine the subset of the state that changed and send only that, and we also don't want to send the state every time. If bandwidth or broker load were a concern, maybe we could get by with some compaction of these messages, sending capabilities of all devices in 1 message if C8y allowed something like this, maybe some compression, but that's probably not practical. So AFAIK with features allowed by C8y you can either have lean startup, or 100% up-to-date device properties, but not both. 

> For the case where the child devices are deleted in the cloud and we expect this fix to re-create those devices, I don't think it'll work for nested child devices.

I misread the issue description and thought that the scope of this fix is just when operations of a device were removed, but indeed it's about when devices themselves are also removed. In that case, I was under the impression that registration of devices is being managed by the mapper in combination with the entity store such that if we attempt to send a supported operation message for a device that has a parent before registering the parent, mapper will delay it and insert parent registration first.

But indeed it doesn't work if we emit `114` directly, which is what we're doing here. C8y messages are published immediately, only local registrations get processed by the store. The ideal solution would be to go through the hierarchy-aware entity store first, which delays delays messages for devices that aren't yet cloud-registered, or to know that entity store already cloud-registered all the child devices.

Bravo555: > So, if we go on to recreate the child devices from this purely using the sorted order, we'll get the hierarchy wrong (child0 an child1 will get re-created as immediate child devices).

Made a test for this in 34fbd40ff3, and it passes, I think, because registration messages happen to be processed and registered first, which i think should be the case unless somebody deletes mosquitto message database and entity store, in which case this can probably still happen.

albinsuresh: > But indeed it doesn't work if we emit 114 directly, which is what we're doing here. C8y messages are published immediately, only local registrations get processed by the store. The ideal solution would be to go through the hierarchy-aware entity store first, which delays delays messages for devices that aren't yet cloud-registered, or to know that entity store already cloud-registered all the child devices.

Yes. To avoid resending the child-device registration messages for the devices that are already there in the cloud, we can also use the [SmartREST 105](https://cumulocity.com/docs/smartrest/mqtt-static-templates/#105) message to fetch the list of child devices for the main device and then recursively do the same for the child devices as well. Then we'll have both the cloud-view as well as the local-view (from the entity store) to perform any reconciliation.

> Made a test for this in https://github.com/thin-edge/thin-edge.io/commit/34fbd40ff357553c0cef91aac1f8943dea2528d6, and it passes, I think, because registration messages happen to be processed and registered first, which i think should be the case unless somebody deletes mosquitto message database and entity store, in which case this can probably still happen.

Well, that must have been a timing thing, as I had done a manual test with that exact same test case and it failed on the very first attempt itself.

tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
```diff
@@ -96,6 +96,52 @@ Recreate child supported operations after cloud deletion on mapper restart
...
+
+    # set operations for all child devices
+    Execute Command    tedge mqtt pub -r te/device/childA///cmd/restart '{}'
+    Execute Command    tedge mqtt pub -r te/device/child0///cmd/restart '{}'
+    Execute Command    tedge mqtt pub -r te/device/child1///cmd/restart '{}'
+    Execute Command    tedge mqtt pub -r te/device/child2///cmd/restart '{}'
+    Execute Command    tedge mqtt pub -r te/device/child3///cmd/restart '{}'
+    Execute Command    tedge mqtt pub -r te/device/child4///cmd/restart '{}'
+
+    # send registration messages again to have them be more recent than operations messages
```
albinsuresh: I didn't really get the reasoning here.

Bravo555: I hypothesised that the broker sends the retained messages to the starting mapper in order they were published, so tried to send them again after so they processed later but it doesn't seem to make a difference + I could've just not register them before, so I'll remove that.

albinsuresh: > I hypothesised that the broker sends the retained messages to the starting mapper in order they were published

The MQTT protocol provides no such guarantee for messages across multiple topics.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -180,6 +180,7 @@ pub struct CumulocityConverter {
 
     pub supported_operations: SupportedOperations,
     pub operation_handler: OperationHandler,
+    pending_c8y_messages: HashMap<EntityExternalId, Vec<MqttMessage>>,
```
albinsuresh: You don't need to maintain this copy of pending messages here, as the `entity_cache.pending_entities` is already tracking these.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -354,6 +354,14 @@ impl CumulocityConverter {
             messages.push(reg_message);
         }
 
+        // if there are any pending c8y messages for that device, include them after the registration message
+        let pending_c8y_messages = self
+            .pending_c8y_messages
+            .remove(external_id)
+            .unwrap_or_default();
+
+        messages.extend(pending_c8y_messages);
```
albinsuresh: Why not just do this?
```suggestion
        let supported_ops = self
            .supported_operations
            .create_supported_operations(external_id, c8y_prefix)?;

        messages.extend(supported_ops);
```

That way, we don't need to track `pending_c8y_messages` anymore.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -1256,30 +1264,57 @@ impl CumulocityConverter {
...
 
     pub fn load_and_create_supported_operations_messages(
         &mut self,
-        external_id: &str,
-    ) -> Result<MqttMessage, ConversionError> {
-        self.supported_operations
-            .load_all(external_id, &self.config.bridge_config)?;
-        let supported_operations_message = self
+    ) -> Result<Vec<MqttMessage>, ConversionError> {
+        // at this point list of supported operations for all devices should've been initialized in converter
```
albinsuresh: We wouldn't need any of these changes if what's proposed [here](https://github.com/thin-edge/thin-edge.io/pull/4064/changes#r2986967193) works.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: fixed in a7e743b13

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: fixed in a7e743b13

At first I was confused and thought this snippet was inside the startup function, but it's really in the function that converts registration messages, and you are right, we can easily do it here.

crates/extensions/c8y_mapper_ext/src/converter.rs
Bravo555: fixed in a7e743b13

tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
```diff
@@ -70,9 +74,103 @@ Re-publish supported operations by signal channel
...
+    ${mo}=    Cumulocity.Set Managed Object    ${DEVICE_SN}:device:childA
+    Cumulocity.Set Managed Object    ${DEVICE_SN}
+    Log    ${mo}
+
+    Delete Managed Object    ${DEVICE_SN}:device:childA
+    Delete Managed Object    ${DEVICE_SN}:device:child0
+    Delete Managed Object    ${DEVICE_SN}:device:child1
+    Delete Managed Object    ${DEVICE_SN}:device:child2
+    Delete Managed Object    ${DEVICE_SN}:device:child3
+    Delete Managed Object    ${DEVICE_SN}:device:child4
```
albinsuresh: Was that required? Deleting the parent (`childA`) should have deleted all nested child devices as well, right?

tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
```diff
@@ -70,9 +74,103 @@ Re-publish supported operations by signal channel
...
+
+    Cumulocity.Device Should Have A Child Devices
```
albinsuresh:
```suggestion
    Cumulocity.Device Should Not Have Any Child Devices
```
Right?

tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
Bravo555: removed extra deletes in afafba8b

tests/RobotFramework/tests/cumulocity/supported_operations/mapper-publishing-agent-supported-ops.robot
Bravo555: fixed in afafba8b

Issue #4073, opened 2026-03-23
Add basic crypto functions getRandomValues and randomUUID


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

flows allows users to come up with some really powerful features, however the lack of access to a suitable random number generator limits the usage for some more security related uses, e.g. like [x509-cert-issuer](https://github.com/thin-edge/tedge-flows-examples/tree/main/flows/x509-cert-issuer).

**Describe the solution you'd like**

There seems to be a common set of [crypto](https://developer.mozilla.org/en-US/docs/Web/API/Crypto) functions, whilst supporting all of them would be in the future very useful, just supporting a two core functions would be very useful:

A global `crypto` object, with the following functions:

* `crypto.getRandomValues()` - [rquickjs example](https://github.com/awslabs/llrt/blob/dd5d8db65994da55438b2cbf928031e2ca2ef900/modules/llrt_crypto/src/lib.rs#L260)
* `crypto.randomUUID()` - [rquickjs example](https://github.com/awslabs/llrt/blob/dd5d8db65994da55438b2cbf928031e2ca2ef900/modules/llrt_crypto/src/lib.rs#L257)

With this functions, then libraries like the following will be enabled to work:

The `@noble/ed25519` repo documents that they just use the `crypto.getRandomValues` api for their source of randomness
* https://www.npmjs.com/package/@noble/ed25519

Other properties like [crypto.subtle](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) could be supported in the future when new use-cases come up.

**Describe alternatives you've considered**

**Additional context**

PR #4078, opened 2026-03-24
feat:  basic crypto functions available from flow scripts

didier-wenzek:
## Proposed changes

- [x] `crypto.randomUUID` method can be invoked from flow scripts
- [x] `crypto.getRandomValues` method can be invoked from flow scripts
- [ ] ~Fix issues with `getRandomValues` on `Uint8ClampedArray`, `BigInt64Array`, `BigUint64Array`~

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/4073

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/js_lib/crypto.rs
+++ b/crates/extensions/tedge_flows/src/js_lib/crypto.rs
@@ -1,6 +1,9 @@
+use rand::Rng;
 use rquickjs::class::Trace;
 use rquickjs::Ctx;
+use rquickjs::IntoJs;
 use rquickjs::JsLifetime;
+use rquickjs::Object;
 
 #[derive(Clone, Trace, JsLifetime)]
 #[rquickjs::class(frozen)]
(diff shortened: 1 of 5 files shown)


crates/extensions/tedge_flows/src/js_script.rs
```diff
@@ -824,6 +824,79 @@ export function onMessage(message) {
...
+            "Float64Array(5)", // Not required
+        ] {
+            let js = format!(
+                r#"
+    export function onMessage(message) {{
+        const array = new {typed_array};
+        for (const num of array) {{
+          if (num != 0) {{ throw new Error("Expect {typed_array} to be 0-initialized"); }}
+        }}
+        crypto.getRandomValues(array);
```
reubenmiller: There is a bug (or more difference in behaviour here) regarding the return type (or lack of one).

The function signature of [crypto.getRandomValues()](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues) should actually also return a reference to the array. I noticed whilst checking which uses `@noble/curves/ed25519.js` where it uses it within a function like so:

```js
function safeRandomBytes(n) {
  return crypto.getRandomValues(new Uint8Array(n));
}
```

When run using our implementation, it returns `null` instead of a link to the array.

The linked spec says this about the **Return Type**

> The same array passed as typedArray but with its contents replaced with the newly generated random numbers. Note that typedArray is modified in-place, and no copy is made.

crates/extensions/tedge_flows/src/js_lib/crypto.rs
```diff
@@ -0,0 +1,61 @@
...
+        } else if let Some(array) = object.as_typed_array::<u64>() {
+            set_random_values!(ctx, object: u64[array.len()]);
+        } else if let Some(array) = object.as_typed_array::<i64>() {
+            set_random_values!(ctx, object: i64[array.len()]);
+        } else if let Some(array) = object.as_typed_array::<f32>() {
+            set_random_values!(ctx, object: f32[array.len()]);
+        } else if let Some(array) = object.as_typed_array::<f64>() {
+            set_random_values!(ctx, object: f64[array.len()]);
+        }
+    }
```
reubenmiller: Suggestion for https://github.com/thin-edge/thin-edge.io/pull/4078/changes#r2990835801: this should align the function signature (though this hasn't been tested)

```suggestion
    fn random_u32(ctx: Ctx<'js>, object: Object<'js>) -> Result<Object<'js>> {
        if let Some(array) = object.as_typed_array::<u8>() {
            set_random_values!(ctx, object: u8[array.len()]);
        } else if let Some(array) = object.as_typed_array::<i8>() {
            set_random_values!(ctx, object: i8[array.len()]);
        } else if let Some(array) = object.as_typed_array::<u16>() {
            set_random_values!(ctx, object: u16[array.len()]);
        } else if let Some(array) = object.as_typed_array::<i16>() {
            set_random_values!(ctx, object: i16[array.len()]);
        } else if let Some(array) = object.as_typed_array::<u32>() {
            set_random_values!(ctx, object: u32[array.len()]);
        } else if let Some(array) = object.as_typed_array::<i32>() {
            set_random_values!(ctx, object: u32[array.len()]);
        } else if let Some(array) = object.as_typed_array::<u64>() {
            set_random_values!(ctx, object: u64[array.len()]);
        } else if let Some(array) = object.as_typed_array::<i64>() {
            set_random_values!(ctx, object: i64[array.len()]);
        } else if let Some(array) = object.as_typed_array::<f32>() {
            set_random_values!(ctx, object: f32[array.len()]);
        } else if let Some(array) = object.as_typed_array::<f64>() {
            set_random_values!(ctx, object: f64[array.len()]);
        }
        
        Ok(object)
    }
```

crates/extensions/tedge_flows/src/js_script.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/4078/changes/418dd83e04ef10c4253e5e7a6756de22c9d57fc7

crates/extensions/tedge_flows/src/js_lib/crypto.rs
didier-wenzek: Done (with no `Result` wrapper)

Issue #4041, opened 2026-03-13
tedge-agent restart can be blocked by the TwinManagerActor actor not responding


reubenmiller:
**Describe the bug**

tedge-agent can be slow at restarting sometimes as some of the Actors don't shutdown cleanly. This is mostly observed in system tests where restarting the tedge-agent is often done to retrigger specific actions (like refreshing the software list after manually adding a software management plugin).

Below shows an example of a restart of the tedge-agent that takes 60 seconds due to the `TwinManagerActor-7` actor not cleanly shutting down.

```log
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.399277889Z  INFO tedge_agent::operation_workflows::actor: Waiting successful software_list opera
tion to be cleared
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.399679143Z  INFO tedge_agent::software_manager::actor: Installed binary version: 1.7.2~358+gf3ff
860
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.399699435Z  INFO tedge_agent::software_manager::actor: Checking if tedge got self updated
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.399702143Z  INFO tedge_agent::software_manager::actor: Current running version: 1.7.2~358+gf3ff8
60
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.401459533Z  INFO tedge_agent::software_manager::actor: Installed binary version: 1.7.2~358+gf3ff
860
Mar 13 13:40:33 61317d461740 systemd[1]: Stopping tedge-agent.service - tedge-agent is a thin-edge.io component to support operations....
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.620836701Z  INFO Runtime: Shutting down
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.620944077Z  INFO Runtime: Actor has finished: Signal-Handler-4
Mar 13 13:40:33 61317d461740 tedge-agent[723]: 2026-03-13T13:40:33.620991911Z  INFO Runtime: Actor has finished: FsWatcher-6
[...]

PR #4079, opened 2026-03-24
fix: prevent deadlock on shutdown in TwinManagerActor

rina23q:
## Proposed changes
The `TwinManagerActor` did not handle the `Shutdown` signal properly. This causes the 60s timeout as reported in the issue #4041.

The root cause was `recv()` was used instead of `try_recv()`. Internally, `recv()` maps `Err(RuntimeRequest::Shutdown)` to `None`, silently consuming the shutdown signal. 

```rust
#[async_trait]
impl<Input: Send> MessageReceiver<Input> for CombinedReceiver<Input> {
    async fn recv(&mut self) -> Option<Input> {
        match self.try_recv().await {
            Ok(Some(message)) => Some(message),
            _ => None,  // <- Err(Shutdown) is swallowed here!
        }
    }
```

The actor would then fall through to the drain loop at the bottom of `run()`, which waits for all senders to be dropped. However, the runtime holds a sender open while waiting for the actor to finish — so both sides were waiting on each other indefinitely, until the 60-second shutdown timeout expired.

The fix replaces `recv()` with `try_recv()`  and explicitly returning `Ok(())` on `Err(RuntimeRequest::Shutdown)` so the actor exits cleanly before ever reaching the drain loop.

## Types of changes

## Paste Link to the issue
#4041 

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/twin_manager/actor.rs
+++ b/crates/core/tedge_agent/src/twin_manager/actor.rs
@@ -8,6 +8,7 @@ use tedge_actors::Actor;
 use tedge_actors::LoggingSender;
 use tedge_actors::MessageReceiver;
 use tedge_actors::RuntimeError;
+use tedge_actors::RuntimeRequest;
 use tedge_actors::Sender;
 use tedge_actors::SimpleMessageBox;
 use tedge_api::mqtt_topics::Channel;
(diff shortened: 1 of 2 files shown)


albinsuresh (approved):
Good finding.

PR #4082, opened 2026-03-26
fix: flow params substitution expression doesn't need a leading dot

didier-wenzek:
## Proposed changes

Fllow params substitution expressions don't need a leading dot as in `"${params.hello.greeting}"`

__warning__: This is a breaking change. The syntax  `"${.params.hello.greeting}"` is not more accepted.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/extensions/tedge_flows/src/params.rs
```diff
@@ -206,11 +206,12 @@ mod tests {
         .unwrap();
 
         for (expr, value) in [
-            ("${.params.x}", json!(42)),
-            ("${.params.y.a}", json!("foo")),
-            ("${.params.z}", json!([1, 2, 3])),
+            ("${.params.x}", json!(42)), // a leading dot is accepted as for workflow substitution expression
```
didier-wenzek: This can be fully deprecated.

@rina23q and @reubenmiller what would be the impact on already implemented flows?

reubenmiller: technically very little, we could do a global search and replace as all of them are located in a single repo.

crates/extensions/tedge_flows/src/params.rs
```diff
@@ -124,15 +124,15 @@ impl Params {
         else {
             return Ok(expr.into());
         };
-        let Some(path) = path.trim().strip_prefix(".params") else {
+        let Some(path) = path.trim().trim_start_matches('.').strip_prefix("params") else {
```
Bravo555: note: `trim_start_matches` removes repeatedly, so `.....params` will trim all the dots, which I assume is the intent (could also add it to tests for completeness).

crates/extensions/tedge_flows/src/params.rs
didier-wenzek: So, let us deprecate it!

crates/extensions/tedge_flows/src/params.rs
didier-wenzek: Yeah, I know. But, I prefer that to a `strip_prefix('.').unwrap_or(path)`.

I consider to fully deprecate the leading dot. https://github.com/thin-edge/thin-edge.io/pull/4082#discussion_r2995269034

didier-wenzek: I finally removed backward compatibility as this use of a dot has never been released.

PR #4084, opened 2026-03-27
fix: Handle more possible cases in JS exception handling due to OOM

jarhodes314:
## Proposed changes
I was observing some flakiness in the OOM unit test for the flows runtime locally.

I threw the problem in Claude's direction and this is what it produced, and that seems to have resolved the flakiness.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/js_runtime.rs
+++ b/crates/extensions/tedge_flows/src/js_runtime.rs
@@ -367,10 +367,24 @@ impl<'js> JsModules<'js> {
 
 impl LoadError {
     fn from_js(ctx: &Ctx<'_>, err: Error) -> Self {
-        if let Some(ex) = ctx.catch().as_exception() {
+        let caught = ctx.catch();
+        if let Some(ex) = caught.as_exception() {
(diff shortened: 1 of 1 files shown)


crates/extensions/tedge_flows/src/js_runtime.rs
```diff
@@ -367,11 +367,25 @@ impl<'js> JsModules<'js> {
...
+        let caught = ctx.catch();
+        if let Some(ex) = caught.as_exception() {
             LoadError::JsException {
                 message: ex.message().unwrap_or_default(),
                 stack: ex.stack().unwrap_or_default(),
             }
+        } else if let Some(s) = caught.as_string() {
+            // OOM: QuickJS couldn't allocate the Exception object and fell back to
+            // storing "out of memory" as a string atom.
+            LoadError::JsException {
```
didier-wenzek, line 379: This makes sense. I observed that on very tight memory constraint QuickJS was not throwing exceptions, but failed to understand it was a simple string then.

Issue #4048, opened 2026-03-15
allow users to run multiple tedge-mapper-local on the same mqtt broker


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

A user which is running tedge-agent on a child device, should be able to run their own tedge-mapper-local instance on the same device without interfering with the tedge-mapper-local which is running on the main device.

Currently if the user tries to run the `tedge-mapper local` on a child device it will fight with any other instance which is running on the same MQTT broker due to it using the same MQTT Client id, `tedge-mapper-local`.

**Describe the solution you'd like**

Since the tedge-mapper-local is a generic component, it should be able to be used to run on multiple devices so that users can also deploy custom flows directly on child devices as well.

Suggestion: To enable this, it would be useful to allow the user to specify the following:

* the client id should be derived from the mqtt device topic id (e.g. `device/main//`)
* support the `--mqtt-device-topic-id` and `--mqtt-topic-root` flags (and inherit from the respective tedge.toml values)
* the tedge-mapper local should respect the `mqtt.device_topic_id` configuration setting

**Describe alternatives you've considered**

**Additional context**

PR #4086, opened 2026-03-27
refactor: use a custom mapper to run the local flows

didier-wenzek:
## Proposed changes

- [x] `tegde-mapper local` runs using the same code as `tegde-mapper thingsboard`
- [x] check the impacts on initialization  of a local flows mapper: `mkdir -p /etc/tedge/mappers/local/flows`
- [x] fix `tedge mapper <custom>` to load cloud specific builtin transformers
- [x] `tedge mapper <custom>` MQTT session name includes `mqtt.device_topic_id`
- [x] Flow status are published on `te/<child-specific-topic-id>`

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/4048

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
+++ b/tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
@@ -5,7 +5,7 @@ Suite Setup         Custom Setup
 Suite Teardown      Get Logs
 Test Teardown       Remove Lifecycle Flow
 
-Test Tags           theme:tedge_flows
+Test Tags           theme:flows
 
 
 *** Variables ***
--- a/tests/RobotFramework/tests/tedge_flows/local-flows-on-child-devices.robot
+++ b/tests/RobotFramework/tests/tedge_flows/local-flows-on-child-devices.robot
@@ -5,7 +5,7 @@ Library             Cumulocity
 Suite Setup         Custom Setup
 Suite Teardown      Custom Teardown
 
-Test Tags           theme:tedge_flows
+Test Tags           theme:flows
 
 
 *** Test Cases ***
--- a/tests/RobotFramework/tests/tedge_flows/builtin-transformers.robot
+++ b/tests/RobotFramework/tests/tedge_flows/builtin-transformers.robot
@@ -6,7 +6,7 @@ Resource            fs_dynamic_reload.resource
 Suite Setup         Custom Setup
 Suite Teardown      Get Logs
 
-Test Tags           theme:tedge_flows
+Test Tags           theme:flows
 
 
 *** Test Cases ***
--- a/tests/RobotFramework/tests/tedge_flows/flows_hot_reloading.robot
+++ b/tests/RobotFramework/tests/tedge_flows/flows_hot_reloading.robot
@@ -4,7 +4,7 @@ Library             ThinEdgeIO
 Suite Setup         Custom Setup
 Suite Teardown      Get Logs
 
-Test Tags           theme:tedge_flows
+Test Tags           theme:flows
 
 
 *** Variables ***
--- a/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
+++ b/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
@@ -6,7 +6,7 @@ Resource            fs_dynamic_reload.resource
 Suite Setup         Custom Setup
 Suite Teardown      Get Logs
 
-Test Tags           theme:tedge_flows
+Test Tags           theme:flows
 
 
 *** Test Cases ***


tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
```diff
@@ -268,6 +268,7 @@ Flow is reloaded after package-style replacement
 Custom Setup
     ${DEVICE_SN}    Setup
     Set Suite Variable    $DEVICE_SN
+    Execute Command    mkdir -p /etc/tedge/mappers/local/flows
```
reubenmiller: The setup of this flow is a bit annoying since you have to create the folder first before using, but it also means you need to know a bit about the mapper directories in order to create it (though the tedge-mapper is helpful there).

I would suggest that since we include the `tedge-mapper-local` service, that we extend the `ExecStartPre` section to also handle the creation of the folder (note that only the mapper root folder needs to be created), and the `flows` folder will be created by the service. So then we could remove the highlighted line. Below shows the lines we could add to the service.

```sh
ExecStartPre=+-/usr/bin/tedge init
ExecStartPre=+mkdir -p /etc/tedge/mappers/local
ExecStartPre=+chown -R tedge:tedge /etc/tedge/mappers/local
```

The ownership is set to **tedge** user so it aligns also to the other mappers like c8y:

```
$ ls -ld /etc/tedge/mappers/c8y/
drwxr-xr-x 3 tedge tedge 4096 Mar 30 20:00 /etc/tedge/mappers/c8y/
```

reubenmiller: An alternative approach would be to have a "tedge-mapper local --init" command.

tests/RobotFramework/tests/tedge_flows/local-flows-on-child-devices.robot
```diff
@@ -0,0 +1,63 @@
...
+
+Test Tags           theme:tedge_flows
+
+
+*** Test Cases ***
+Child devices can run their own tedge-mapper-local
+    Set Device Context    ${PARENT_SN}
+    Execute Command    mkdir -p /etc/tedge/mappers/local/flows
+    Execute Command    rm -fr /etc/tedge/mappers/local/flows/*
+    Restart Service    tedge-mapper-local
```
albinsuresh, line 15: Do we really want to mainatin a hardcoded mapper service named `tedge-mapper-local`? Since we'll be adding service support for custom mappers as well, we could just let the users choose the names, right? That way, even if the user wants to run multiple instances of the local mapper for whatever reasons, they'll just need to run two named custom mapper instances, right?

tests/RobotFramework/tests/tedge_flows/flow_lifecycle.robot
albinsuresh: Why not just let the custom mapper create the necessary directories in its bootstrap logic (like the agent and the other mappers do for their required directories like the `operations/c8y` directory etc), if they don't exist already? Why do we need to externalise into the service definition or even introduce a dedicated `init` command?

didier-wenzek: > An alternative approach would be to have a "tedge-mapper local --init" command.

I prefer the local directory be created by `ExecStartPre`. This `--init` would have no more to do than an `mkdir` and a `chown`. Easy to do in `ExecStartPre`. A bit more painful and adhoc within the mapper.

didier-wenzek: > Why not just let the custom mapper create the necessary directories in its bootstrap logic (like the agent and the other mappers do for their required directories like the `operations/c8y` directory etc), if they don't exist already? Why do we need to externalise into the service definition or even introduce a dedicated `init` command?

I would prefer not because the mapper name can be arbitrary. A typo and oops you are running a different mapper than you thought.

reubenmiller: The `tedge-mapper-local` service serves a nice example on how to run a local non-cloud connected mapper. and it also has the advantage of normalizing services names for use-cases where you want to deploy flows that are independent of a cloud connection...but yes it is purely convention, however an out-of-the-box one which will encourage adoption.

reubenmiller: I pushed the proposed changes (to create the folders on startup of the tedge-mapper-local service) in f68d80a5904bd84d54f58a5a9746476aac7ac492

crates/core/tedge_mapper/src/core/mapper.rs
```diff
@@ -17,7 +17,16 @@ pub async fn start_basic_actors(
...
 
-    let mut mqtt_actor = get_mqtt_actor(mapper_name, config).await?;
+    let device_topic_id = &config.mqtt.device_topic_id;
+    let session_name = if device_topic_id.is_default_main_device() {
+        mapper_name.to_string()
+    } else {
+        format!(
+            "{mapper_name}-{}",
+            device_topic_id.as_str().trim_matches('/').replace('/', "-")
+        )
```
reubenmiller: Should we be consistent here and follow something similar to how the tedge-agent uses the topic id in the client name (see below for an example):

```
INFO MQTT: Connecting to broker: host=172.18.0.3:1883, session_name=Some("tedge-agent#te/device/TST_color_bordeaux_breakout//")
```

Or does using mqtt chars cause problems with MQTT ACL rules if people do reference client ids from with the ACL rules?

didier-wenzek: > Or does using mqtt chars cause problems with MQTT ACL rules if people do reference client ids from with the ACL rules?

This name is only used for the MQTT session.

reubenmiller: ok, though for consistency, can we change it to follow the same style as the tedge-agent? e.g. `tedge-mapper-local#{topic-id}`? It makes it easier to trace things in the logs

tests/RobotFramework/tests/tedge_flows/local-flows-on-child-devices.robot
```diff
@@ -0,0 +1,63 @@
+*** Settings ***
+Library             ThinEdgeIO
+Library             Cumulocity
+
+Suite Setup         Custom Setup
+Suite Teardown      Custom Teardown
+
+Test Tags           theme:tedge_flows
```
rina23q: I think `theme:flows` is better as it matches the GitHub issue's label. Though we mix `theme:tedge_flows` and `theme:flows`.

```suggestion
Test Tags           theme:flows
```

albinsuresh (approved):

crates/core/tedge_mapper/src/core/mapper.rs
didier-wenzek: Done

tests/RobotFramework/tests/tedge_flows/local-flows-on-child-devices.robot
didier-wenzek: Done

PR #4089, opened 2026-03-31
feat: custom mapper improvements

jarhodes314:
## Proposed changes
- [x] Ability to disable TLS for the bridge connection (i.e. connect to insecure cloud brokers) via `bridge.tls` config in `mapper.toml`
- [x] Handle profiles somewhat gracefully in `tedge mapper`/`tedge bridge` subcommands, e.g. `tedge mapper config get c8y.new.url`
- [ ] ~Basic `tedge connect --test` support~ - Going to do this in a separate PR
- [x] Configurable bridge rules for aws/az mappers
- [x] Improve relative path support, `tedge mapper config get` will resolve relative paths in mapper.toml before displaying them
- [x] Rename `${config.}` to `${tedge.}` in bridge templates
- [x] Add tab-completion support to `tedge mapper config get` and `tedge bridge inspect`
- [x] Disable mapper.toml without bridge directory warning for built-in mappers
- [x] Ensure docs are up-to-date with these changes - particularly the bridge config renaming

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_mapper/src/az/mapper.rs
+++ b/crates/core/tedge_mapper/src/az/mapper.rs
@@ -223,6 +223,8 @@ mod tests {
 
         // Device-to-cloud messages
         assert!(has_local_subscription(&rules, "az/messages/events/#"));
+
+        // Cloud-to-device messages
         assert!(has_remote_subscription(
             &rules,
             "devices/test-device-id/messages/devicebound/#"


crates/core/tedge_mapper/src/custom/resolve.rs
```diff
@@ -276,6 +277,22 @@ impl EffectiveMapperConfig {
                 },
                 source: self.effective_auth.source.clone(),
             }),
+            "bridge.tls" => {
```
reubenmiller: I'm wondering if we should rename this property slightly from `bridge.tls` to something like or `bridge.tls_enabled`, as the name is a bit ambiguous whether it is referring to enabling the tls or specifying the tls version (e.g. like the mosquitto `tls_version`).

Or am I overthinking this?

reubenmiller: Just for reference, mosquitto does allow users to control the TLS version used, via the [tls_version](https://mosquitto.org/man/mosquitto-conf-5.html), however I'm not proposing that we need to do this now, but I just want to make sure the name doesn't preclude us from supporting it in the future.

crates/core/tedge/src/cli/bridge/inspect.rs
```diff
@@ -77,11 +79,31 @@ async fn run_inspect(
...
             #[cfg(feature = "aws")]
             Cloud::Aws(_) => {
-                print_non_configurable_or_disabled(w, config, &cloud);
+                use tedge_config::tedge_toml::mapper_config::AwsMapperSpecificConfig;
+                if let Some((rules, non_expansions)) =
+                    load_bridge_rules::<AwsMapperSpecificConfig>(w, config, &cloud, detail).await?
+                {
+                    if detail == DetailLevel::Debug {
+                        print_non_expansions(w, &non_expansions);
+                    }
```
Bravo555, line 88: note: not introduced in this commit, but this is controlled by `--debug` flag, which is a bit confusing since that flag is also already used to control the level of general logs.

crates/core/tedge_mapper/src/aws/bridge/rules.toml
```diff
@@ -0,0 +1,37 @@
+# Bridge rules for AWS IoT Core.
+#
+# AWS IoT Core MQTT reference: https://docs.aws.amazon.com/iot/latest/developerguide/mqtt.html
```
Bravo555, line 3: very nice to include reference relevant documentation in the comment

tests/RobotFramework/tests/custom_mapper/custom_mapper.robot
```diff
@@ -160,6 +160,84 @@ Two custom mapper profiles run concurrently without interfering
...
+    # Verify by checking that `tedge mapper test` does NOT report the bridge as healthy
+    # (the mapper health check should pass, but no bridge check should occur since
+    # there is no bridge/ directory).
+
+    # Wait for mapper to report healthy first
+    Execute Command    tedge mqtt pub 'te/device/main/service/tedge-mapper-test-stale/cmd/health/check' ''
+    Should Have MQTT Messages
+    ...    te/device/main/service/tedge-mapper-test-stale/status/health
+    ...    minimum=1
+    ...    message_contains="status":"up"
```
Bravo555, line 247: todo: this doesn't have a `date_from` so it can pick up the same stale up message from earlier

crates/core/tedge_mapper/src/az/mapper.rs
```diff
@@ -114,37 +118,182 @@ impl TEdgeComponent for AzureMapper {
...
+        let config = TEdgeConfig::load(ttd.path()).await.unwrap();
+
+        let rules = bridge_rules(&config, None).await.unwrap();
+
+        // Device-to-cloud messages
+        assert!(has_local_subscription(&rules, "az/messages/events/#"));
+        assert!(has_remote_subscription(
+            &rules,
+            "devices/test-device-id/messages/devicebound/#"
+        ));
```
Bravo555, line 231:
```suggestion
        // Device-to-cloud messages
        assert!(has_local_subscription(&rules, "az/messages/events/#"));
        
        // Cloud-to-device messages
        assert!(has_remote_subscription(
            &rules,
            "devices/test-device-id/messages/devicebound/#"
        ));

```

crates/core/tedge_mapper/src/custom/config.rs
```diff
@@ -81,6 +81,19 @@ pub struct DeviceConfig {
...
+#[serde(rename_all = "lowercase")]
+pub enum BridgeTls {
+    /// Always use TLS, regardless of port.
+    On,
+    /// Never use TLS (plain TCP). Incompatible with certificate authentication.
+    Off,
+    /// Infer from port: 8883 → TLS on, 1883 → TLS off, other → TLS on.
+    #[default]
+    Auto,
+}
```
Bravo555: todo: I agree with [Reuben](https://github.com/thin-edge/thin-edge.io/pull/4089#discussion_r3016642887) that we should probably leave `bridge.tls` to be a map, in case we want to add any other tls settings later, and for the enabled case do something like `bridge.tls.enable`, as that's the convention we use for other options (`tedge config list enable` lists 13 options), though it might be confusing that it's not just true/false, but also has auto, in which case i think the best way to handle this would be to map `true => on`, `false => Off` as aliases.

crates/core/tedge_mapper/src/custom/mapper.rs
```diff
@@ -936,5 +1003,140 @@ AwEHoUQDQgAEdklRDw9+AAMRbpNMWJutKe4QO/tUlvrBR2swUYN9onxXdKNjJ/k3\n\
...
+                let tedge_config = TEdgeConfig::load_toml_str("device.id = \"test-device\"");
+                let mut config = make_config(Some("mqtt.example.com:1883"));
+                config.bridge.tls = BridgeTls::Off;
+                config.credentials_path = Some(creds_path);
+                let effective = resolve(&config, &tedge_config).await;
+
+                build_cloud_mqtt_options(&effective, "svc", &mapper_dir, &tedge_config)
+                    .await
+                    .unwrap();
+            }
```
Bravo555, line 1042: question: In these tests, we don't actually assert whether TLS is used or not, but only if the config is [_valid_](https://mmapped.blog/posts/40-three-degress-of-validity#configurations-as-programs:~:text=Static,valid%2E), i.e. if no mutually-exclusive options are selected?

crates/core/tedge_mapper/src/custom/resolve.rs
jarhodes314: I've changed it to `bridge.tls.enable` as per @Bravo555's suggestion, with true/false/auto instead of on/off/auto

crates/core/tedge_mapper/src/custom/mapper.rs
jarhodes314: Oops, I've fixed that to check `MqttOptions::transport` is the correct type

crates/core/tedge_mapper/src/custom/config.rs
jarhodes314: Now done, and I've changed it to `"true"`/`"false"`/`"auto"` since on/off doesn't make much sense here

crates/core/tedge/src/cli/bridge/inspect.rs
jarhodes314: Yeah, this is a reasonable point. Perhaps it should be a `-v` flag instead? @reubenmiller, any thoughts on this?

tests/RobotFramework/tests/custom_mapper/custom_mapper.robot
jarhodes314: Now fixed

didier-wenzek (approved):
Approved. A nice set of improvements on top a great addition. 

It would have been better for the reviewers to split this PRs into smaller ones.

crates/core/tedge/src/cli/bridge/inspect.rs
reubenmiller: `-v` isn't very clear and it is too similar to also other standard of verbose mode which is more related to log levels.

What about `--show-all` or `--all` since the flag is about showing all the rules included the skipped ones (at least that is what I understood from the `Show skipped rules (e.g. due to unmet conditions or empty template loops)` docstring.

reubenmiller: It was decided to put this change in a follow up PR (to unblock other dependent PRs)

PR #4091, opened 2026-03-31
docs: add tutorial on writing flows

reubenmiller:
## Proposed changes

Preface:
* The original doc structure and scope of examples was created manually and then polished with AI
* All examples but "Example 6: Adaptive data mode", have been manually verified to check if they work as expected

Add a flows tutorial page to guide users on how to extend thin-edge.io using flows.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/docs/src/extend/flows.md
+++ b/docs/src/extend/flows.md
@@ -1,0 +1,1107 @@
+---
+title: Flows
+tags: [Extend, Flows]
+sidebar_position: 5
+description: Write custom IoT data logic on the device using JavaScript
+---
+
+This tutorial walks through how to write your own custom logic using %%te%% flows; from a simple data mapping example to more advanced topics like aggregation and cross-flow communication. If you're looking for just the technical details, see the [reference](../references/mappers/flows.md) page.
+
+## Overview {#overview}
+
+A **flow** is a small piece of logic that runs inside one of %%te%%'s mappers. Each flow is triggered by an input source — such as incoming MQTT messages, a file, or output from a process — processes the data using a JavaScript function, and optionally produces output — for example, publishing to the local MQTT broker or appending to a file.
+
+Flows are well-suited for use-cases like:
+
+- Remapping data between formats — for example, converting a custom sensor payload into the %%te%% format
+- Filtering or dropping unwanted messages before they reach the cloud
+- Running local analytics on the device, such as triggering alarms based on sensor data
+- Aggregating high-frequency sensor data to reduce bandwidth usage
+- Monitoring cloud message rates and raising local alerts
+
+Flows are designed to be lightweight and self-contained. Compared to writing a standalone application, they offer several practical advantages:
+
+- No service configuration (systemd, OpenRC, etc.) needed — flows run inside an existing mapper
+- No MQTT client to manage — connection, reconnection and QoS handling are provided automatically
+- Sandboxed execution with restricted CPU and memory usage
+- Packaged as a simple gzip tarball and hot-reloaded without any service restarts
+
+## Anatomy of a flow {#anatomy}
+
+A flow is defined by a TOML definition file (typically a `flow.toml`) that declares its inputs, outputs, and the steps that process messages. Steps can be JavaScript scripts (ES2020) or built-in functions implemented in Rust. The runtime loads each JavaScript script as an [ECMAScript Module](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) and calls the exported handler functions when events occur.
+
+The example below shows a typical flow package where all related files are kept together in a single directory. This is a convention rather than a requirement, but it helps make clear which files belong to which flow:
+
+```text
+my-flow/
+├── flow.toml             # Flow definition: inputs, steps, configuration
+├── main.js               # JavaScript logic
+└── params.toml.template  # (optional) Configuration defaults; copy to params.toml to customize
+```
+
+**flow.toml** declares what messages the flow receives, which script steps process them, and any configuration values.
+
+**main.js** exports handler functions that are called when messages arrive or timers fire.
+
+**params.toml.template** is the flow author's way of communicating what can be configured by users. The %%te%% runtime reads `params.toml.template` directly as a source of default values — if a `params.toml` file does not yet exist, the template values are used automatically. Users only need to create a `params.toml` (by copying the template) if they want to override specific values. When the flow package is updated, any existing `params.toml` is preserved so user-configured values are not overwritten.
+
+### Handlers {#handlers}
+
+JavaScript steps export one or more handler functions that the runtime calls as events occur. Three handlers are available:
+
+| Handler | When it is called | Typical use |
+|---------|------------------|-------------|
+| `onMessage(message, context)` | Once per incoming message | Data mapping, filtering, threshold checks |
+| `onInterval(time, context)` | At a fixed interval | Aggregation, periodic reporting, rate checks |
+| `onStartup(time, context)` | When the flow is first loaded | Initialization |
+
+Each handler returns an array of output messages. Return an empty array (`[]`) to emit nothing. Where each message is sent — MQTT topic or file — is determined by the flow's output connector configuration.
+
+The `message` object provides:
+- `message.topic` — source topic (an MQTT topic, file path, or process name depending on the input connector)
+- `message.payload` — raw bytes; use `new TextDecoder().decode(message.payload)` to get a string
+- `message.time` — message timestamp as a JavaScript `Date`
+
+The `context` object provides:
+- `context.config` — configuration values declared in `flow.toml`, and typically mapped to values from the `params.toml` file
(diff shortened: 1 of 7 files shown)


docs/src/extend/flows.md
```diff
@@ -0,0 +1,916 @@
...
+
+Each handler returns an array of output messages. Return an empty array (`[]`) to emit nothing. Where each message is sent — MQTT topic or file — is determined by the flow's output connector configuration.
+
+The `message` object provides:
+- `message.topic` — source topic (an MQTT topic, file path, or process name depending on the input connector)
+- `message.payload` — raw bytes; use `new TextDecoder().decode(message.payload)` to get a string
+- `message.time` — message timestamp as a JavaScript `Date`
+
+The `context` object provides:
+- `context.config` — configuration values declared in `flow.toml`, and typically mapped values from the `params.toml` file (though mapping is done explicitly)
```
didier-wenzek: > though mapping is done explicitly

This is not really clear. I would either remove this remark (giving details later) or be more explicit:

```suggestion
- `context.config` — configuration values declared in `flow.toml`, and typically mapped values from the `params.toml` file (using substitution expressions as `${params.my.flow.parameter}`)
```

Giving details later sounds better for a tutorial.

docs/src/extend/flows.md
```diff
@@ -0,0 +1,916 @@
+---
+title: Flows
+tags: [Extend, Flows]
+sidebar_position: 2
+description: Write custom IoT data logic on the device using JavaScript
+---
+
+This tutorial walks through how to write your own custom logic using %%te%% flows; from a simple data mapping example to more advanced topics like aggregation and cross-flow communication.
```
didier-wenzek: A link to the reference documentation might help

docs/src/extend/flows.md
```diff
@@ -0,0 +1,916 @@
...
+
+    if (typeof temperature !== "number") {
+        // Skip messages that do not carry a numeric temperature
+        return [];
+    }
+
+    // Re-publish in the thin-edge measurement format
+    return [{
+        topic: "te/device/main///m/environment",
+        payload: JSON.stringify({ temperature }),
```
didier-wenzek, line 108:
```suggestion
        payload: JSON.stringify({ "temperatue": temperature }),
```

docs/src/extend/flows.md
```diff
@@ -0,0 +1,916 @@
...
+            mqtt: { retain: true, qos: 1 },
+        }];
+    }
+
+    // Clear the alarm by publishing an empty retained message
+    return [{
+        topic: alarmTopic,
+        payload: "",
+        mqtt: { retain: true, qos: 1 },
+    }];
```
didier-wenzek, line 171: We can introduce a state here to avoid clearing the alarm for all measurements.

I agree that this can be introduced as a next step - using some static value or `context.step`.

docs/src/extend/flows.md
```diff
@@ -0,0 +1,916 @@
...
+| tedge flows test --flows-dir ./cloud-rate-monitor/ --final-on-interval
+```
+
+```text title="Output"
+[te/device/main///e/cloud_rate_exceeded] {"text":"Hourly cloud message limit reached: 10001 messages (limit: 1000)"}
+[te/device/main///e/cloud_rate_exceeded] {"text":"Daily cloud message limit reached: 10001 messages (limit: 10000)"}
+```
+
+:::note Context is not persisted
+`context.mapper` values are in-memory only. If the mapper process restarts, all counters reset to zero. For this monitoring use case that is acceptable — the rate monitor resumes counting from restart. If permanent state is required, consider writing to a file instead.
```
didier-wenzek: > If permanent state is required, consider writing to a file instead.

No easy way to implement this for now.

docs/src/extend/flows.md
reubenmiller: The two statements are equivalent though, and the former is usually recommend by JavaScript linting rules (e.g. this rule is called [object-shorthand](https://eslint.org/docs/latest/rules/object-shorthand))

docs/src/extend/flows.md
reubenmiller: Yes using context would be a much better way, though I didn't want to introduce context immediately, or do you think we should use it still as it is "less naive" than clearing the alarm everytime?

docs/src/extend/flows.md
reubenmiller: I missed that, sorry. I'll remove the last sentence.

docs/src/extend/flows.md
reubenmiller: I removed the additional info in 65e3d7848e5f31e24a7e5b4826fb8705600abfe1

docs/src/extend/flows.md
reubenmiller: Added a link to the reference page in a new sentence in 65e3d7848e5f31e24a7e5b4826fb8705600abfe1

docs/src/extend/flows.md
reubenmiller: Removed in 65e3d7848e5f31e24a7e5b4826fb8705600abfe1

docs/src/extend/flows.md
didier-wenzek: > I didn't want to introduce context immediately

Yes, it's better to be progressive. This can be introduced by a 3rd example.

1. Raising alarms on high temperature (basic)
2. Configure temperature threshold (using config and params)
3. Clearing alarms when below an hysteresis threshold (using context.step)

docs/src/extend/flows.md
didier-wenzek: I was not aware of that for JS.

docs/src/extend/flows.md
reubenmiller: Just pushed a new example which builds on example 3 and takes it a step further introducing `context.script` in 42c527ef53528eb9841e60413e157cb525cf8943

docs/src/extend/flows.md
```diff
@@ -0,0 +1,1049 @@
...
+
+---
+
+## Example 8: Multi-step debounce alarm
+
+**Goal:** Raise a temperature alarm only after a configurable number of _consecutive_ above-threshold readings. A single spike should not trigger an alarm — only sustained elevated temperatures are actionable.
+
+This example introduces two new concepts:
+- **Multi-step pipelines** — a flow can chain multiple scripts, where each step's output becomes the next step's input
+- **`context.flow`** — a key-value store shared between all scripts within the same flow, useful for inter-step coordination
```
didier-wenzek: Using a `context.flow` here is not really helpful: it is used by a single flow.

reubenmiller: Resolved by d8824d07ac4d7733dc3dfd2de216fe18a51e7ccf (switched to context.script)

docs/src/extend/flows.md
```diff
@@ -0,0 +1,1100 @@
...
+EOF
+```
+
+### Testing deployed flows
+
+Once a flow is installed (under `/etc/tedge/mappers/<mapper>/flows/`), you can test it without `--flows-dir`. The command uses flows already loaded by the running mapper:
+
+```sh
+echo '[te/device/main///m/environment] {"temperature": 75.0}' \
+    | tedge flows test --mapper local
```
rina23q, line 537: A surprise suddenly `local` appears.

```suggestion
    | tedge flows test --mapper <mapper>
```

docs/src/extend/flows.md
```diff
@@ -0,0 +1,1100 @@
...
+
+```sh
+cp -Ra dev/myflow /etc/tedge/mappers/c8y/flows/
+```
+
+Before deploying, validate the flow logic with `tedge flows test`. It catches syntax errors and lets you confirm that the output is correct before anything is published to a live broker.
+
+### On a device via Cumulocity
+
+Install a packaged flow through Cumulocity software management. The software name must use the `<mapper>/<flow>` format — for example, `c8y/high-temp-alert` installs the flow under the Cumulocity mapper's flows directory, while `local/monitor` targets the local mapper. If no mapper prefix is provided, the local mapper is used by default.
```
rina23q: > If no mapper prefix is provided, the local mapper is used by default.

This is incorrect. The flow sm-plugin returns an error due to missing the mapper name.

reubenmiller: that's also ok, that was the point I was unsure of, so great that you picked that up...I'll push a fix shortly

docs/src/extend/flows.md
reubenmiller: Fixed in e4c10b64803fe0b4e0cc7a56041f5e625aa99900. I ended up using the `<mapper>` and then providing an example.

docs/src/extend/flows.md
reubenmiller: Fixed in e4c10b64803fe0b4e0cc7a56041f5e625aa99900

Issue #3528, opened 2025-04-03
allow users to run thin-edge.io as a custom user (e.g. under the current userspace)


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When running thin-edge.io under a customer user (e.g. on macOS it is run under the user account), then some commands print a warning when the `tedge` user does not exist.

Most notably this is visible when using the macOS [tedge homebrew](https://github.com/thin-edge/homebrew-tedge) formula which runs thin-edge.io under the user's account (as it does not have permissions to create new users as the installation runs as a non-root user).

Below shows an example of setting a tedge config item.

```sh
$ tedge config set c8y.url $C8Y_DOMAIN
2025-04-03T07:07:26.409359Z  WARN tedge_config::tedge_toml::tedge_config_location: failed to set file ownership for '/opt/homebrew/etc/tedge/tedge.toml': User not found: "tedge".
```

And another example when using the `tedge reconnect c8y`.

```sh
$ tedge reconnect c8y
Disconnecting from Cumulocity
Removing bridge config file... ✓
Disabling tedge-mapper-c8y... ✓
Reconnecting with config:
	device id: rmi_helloworld001
	cloud profile: <none>
	cloud host: thin-edge-io.eu-latest.cumulocity.com:8883
	auth type: Certificate
	certificate file: /opt/homebrew/etc/tedge/device-certs/tedge-certificate.pem
	cryptoki: false
	bridge: built-in
	service manager: homebrew
	mosquitto version: 2.0.21
Creating device in Cumulocity cloud... ✓
2025-04-03T07:27:11.484431Z  WARN tedge::cli::connect::command: Failed to change ownership of /opt/homebrew/etc/tedge/device-certs/tedge-certificate.pem to tedge:tedge: User not found: "tedge".
2025-04-03T07:27:11.48477Z  WARN tedge::cli::connect::command: Failed to change ownership of /opt/homebrew/etc/tedge/device-certs/tedge-private-key.pem to tedge:tedge: User not found: "tedge".
Restarting mosquitto... ✓
Waiting for mosquitto to be listening for connections... ✓
Enabling tedge-mapper-c8y... ✓
Verifying device is connected to cloud... ✓
Checking Cumulocity is connected to intended tenant... ✓
Enabling tedge-agent... ✓
```

**Describe the solution you'd like**

Provide a mechanism for users to run thin-edge.io under a custom non-root user (and not assume the `tedge` user always exists).

[...]

PR #4094, opened 2026-04-01
feat: allow non-tedge users to own and run tedge

albinsuresh:
## Proposed changes

Allow non-tedge users to own and run tedge.

* allow users to control the user/group via the system.toml
* support running everything as the "current user" (when setting empty user/group values in the system.toml)
* all commands should respect these settings, e.g. `tedge init`, `c8y-remote-access-plugin --init`, `tedge cert` etc.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/3528

## Checklist


configuration/package_scripts/tedge/preinst
```diff
@@ -29,31 +29,51 @@ user_exists() {
...
+        fi
     fi
 fi
 
-### Create file in /etc/sudoers.d directory. With this configuration, the tedge user have the right to call the tedge command with sudo rights, which is required for system-wide configuration in "/etc/tedge"
-if [ -d /etc/sudoers.d ]; then
+### Create file in /etc/sudoers.d directory. With this configuration,
+###the configured user has the right to call the tedge command with sudo rights,
+### which is required for system-wide configuration in "/etc/tedge"
+if [ "$TEDGE_USER" = "tedge" ] && [ -d /etc/sudoers.d ]; then
```
reubenmiller, line 92: I'm wondering if we really need the "tedge" specific check here? Couldn't we just reference the `$TEDGE_USER` variable within the sudoers.d file creation instead of using the hard coded "tedge" value?

configuration/package_scripts/tedge/preinst
```diff
@@ -29,31 +29,51 @@ user_exists() {
...
 
+# Read user/group from /etc/tedge/system.toml if it exists
+TEDGE_USER="tedge"
+TEDGE_GROUP="tedge"
+SYSTEM_TOML="/etc/tedge/system.toml"
+if [ -f "$SYSTEM_TOML" ]; then
+    _val=$(grep -E '^[[:space:]]*user[[:space:]]*=' "$SYSTEM_TOML" | head -1 | sed 's/.*=[[:space:]]*"\(.*\)".*/\1/')
+    [ -n "$_val" ] && TEDGE_USER="$_val"
+    _val=$(grep -E '^[[:space:]]*group[[:space:]]*=' "$SYSTEM_TOML" | head -1 | sed 's/.*=[[:space:]]*"\(.*\)".*/\1/')
+    [ -n "$_val" ] && TEDGE_GROUP="$_val"
```
reubenmiller: Generally we use `head -n1` instead of `head -1` as the former is more compatible.

The grep and sed are pretty fancy, but I guess we can live without given that we don't have any good toml parsing tools available.

configuration/package_scripts/tedge/preinst
albinsuresh: Yeah, I was also not fully convinced what the "secure practice" here is. Initially I had it as you're suggesting, esp for convenience (less things for him to do). But then I was wondering if granting sudo permissions to a user that we didn't create is okay or not, from a security point of view. Ideally, the user should be managing his own permissions, right? Or is it okay for a tool that he installs to make those changes on his behalf? 

But may be, we can argue that we're only giving the user the necessary sudo rights to run `tedge` tools only and nothing more.

tests/RobotFramework/tests/installation/install_tedge.robot
```diff
@@ -49,10 +55,57 @@ Install then uninstall latest tedge via script (from main branch)
...
+    Transfer To Device    ${CURDIR}/system.toml    /etc/tedge/system.toml
+
+    # Install tedge
+    Install tedge
+
+    Execute Command    cmd=sed -i s/User=tedge/User=tedge-custom/ /lib/systemd/system/tedge-agent.service
+    Execute Command    cmd=sed -i s/User=tedge/User=tedge-custom/ /lib/systemd/system/tedge-mapper-c8y.service
+    Execute Command    cmd=systemctl daemon-reload
+
+    Connect device    ${DEVICE_SN}
```
albinsuresh: Something weird is happening here. Before `tedge connect` is executed here, that starts the `tedge-agent`, the service definition has `User=tedge-custom` as I have set in the previous steps. But after `tedge connect` that is getting switched back to `User=tedge`. That's weird and I wonder why. Investigating.

albinsuresh: Damn, that was a silly dev error. The `Connect device` call, instead of just connecting the device, was re-executing the boostrap script as well, reinstalling `tedge` again, wiping that systemd definition changes made in the previous step.

albinsuresh: Resolved by [c738cbf](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c738cbf0840d9d820eb584e64a95d9256e51f6b7)

configuration/package_scripts/tedge/preinst
reubenmiller: Though these permissions can be critical to some of the tedge functionality, e.g. if we can't use the tedge-write command, then the config plugin will not work in a lot of cases. If the system.toml is set to root permissions, then a non-root user can't change these values anyway.

Adding a check for the "tedge" user specifically feels a bit arbitrary, as the user would also get created automatically (which is normally an action reserved for root users)

reubenmiller: ah sorry I misread the part in the postinst script regarding the user creation! I thought it was creating the users automatically, and that's why I thought if were were doing that, then why not set the sudoers rules!

So ignore my comment. Sorry for the confusion

configuration/package_scripts/tedge/preinst
reubenmiller: We'll also need to improve the snippet regarding parsing the system.toml as it assumes a well-formed file (e.g. if i forget to have closing quotes, `user = "foo`, then the awk still matches, and results in an unexpected value.

And `[ -n "$_val" ] && TEDGE_GROUP="$_val"` will also stop the postinst execution if there is no value matched due to the usage of `set -e` at the top of the script. We can just use a plain if statement rather than using `<test> && true || false` logic.

Though I can refine this logic, and check all the corner case (if you want).

crates/core/tedge/src/cli/init.rs
```diff
@@ -76,12 +92,10 @@ impl TEdgeInitCmd {
...
-        let permissions = {
-            PermissionEntry::new(
-                Some(self.user.clone()),
-                Some(self.group.clone()),
-                Some(0o775),
-            )
+        let permissions = if user.is_empty() && group.is_empty() {
+            PermissionEntry::new(None, None, Some(0o775))
+        } else {
+            PermissionEntry::new(Some(user.clone()), Some(group.clone()), Some(0o775))
```
rina23q: I think the logic is incomplete. It's missing to address one of `user` or `group` is empty.

```suggestion
        let permissions = PermissionEntry::new(
            (!user.is_empty()).then(|| user.clone()),
            (!group.is_empty()).then(|| group.clone()),
            Some(0o775),
        );
```

crates/core/tedge/src/cli/init.rs
```diff
@@ -103,14 +117,46 @@ impl TEdgeInitCmd {
         create_directory_and_update_ownership(&config.logs.path, &permissions).await?;
         create_directory_and_update_ownership(&config.data.path, &permissions).await?;
 
-        let entity_store_file = config_dir.join(".agent").join("entity_store.jsonl");
+        let system_toml = config_dir.join("system.toml");
+        if system_toml.exists() && !user.is_empty() && !group.is_empty() {
```
rina23q: Why does it limit to the case **both** `user` and `group` exist?

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -405,8 +406,11 @@ pub async fn bridge_rules(
 ) -> anyhow::Result<BridgeConfig> {
     let mapper_config_dir =
         tedge_config.mapper_config_dir::<C8yMapperSpecificConfig>(cloud_profile);
+    let system_config = SystemConfig::try_new(tedge_config.root_dir()).unwrap_or_default();
+    let (tedge_user, tedge_group) = (system_config.user.as_str(), system_config.group.as_str());
     if let Err(err) =
-        create_directory_with_user_group(mapper_config_dir.clone(), "tedge", "tedge", 0o755).await
+        create_directory_with_user_group(mapper_config_dir.clone(), tedge_user, tedge_group, 0o755)
```
rina23q: And here there is no guard to check if `tedge_user.is_empty()` or `tedge_group.is_empty()`. So, the directory creation will fail in that case.

rina23q (commented):
I have minor comments regarding how to address a corner case where toml file has `user` or `group` key but empty value.

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -951,21 +951,28 @@ impl ConnectCommand {
...
     if bridge_config.auth_type == AuthType::Basic {
         return;
     }
 
     let (user, group) = match bridge_config.bridge_location {
-        BridgeLocation::BuiltIn => ("tedge", "tedge"),
-        BridgeLocation::Mosquitto => (crate::BROKER_USER, crate::BROKER_GROUP),
+        BridgeLocation::BuiltIn => {
+            let system_config =
+                tedge_config::SystemConfig::try_new(tedge_config.root_dir()).unwrap_or_default();
```
didier-wenzek: This deserves a `TedgeConfig` method

```suggestion
               tedge_config.read_system_config();
```

crates/core/tedge/src/cli/mod.rs
```diff
@@ -97,13 +97,17 @@ pub enum Component {
 pub enum TEdgeOpt {
     /// Initialize Thin Edge
     Init {
-        /// The user who will own the directories created
-        #[clap(long, default_value = "tedge")]
-        user: String,
+        /// The user who will own the directories created.
+        /// Overrides the value from system.toml [user].
+        /// Defaults to "tedge".
```
didier-wenzek: This is not accurate. `tedge` is the default only if no user is provided by `system.toml`

crates/core/tedge/src/cli/mod.rs
```diff
@@ -97,13 +97,17 @@ pub enum Component {
...
+        #[clap(long)]
+        user: Option<String>,
 
-        /// The group who will own the directories created
-        #[clap(long, default_value = "tedge")]
-        group: String,
+        /// The group who will own the directories created.
+        /// Overrides the value from system.toml [group].
+        /// Defaults to "tedge".
```
didier-wenzek: As for `--user`, the default value is not accurate.

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -405,8 +406,11 @@ pub async fn bridge_rules(
 ) -> anyhow::Result<BridgeConfig> {
     let mapper_config_dir =
         tedge_config.mapper_config_dir::<C8yMapperSpecificConfig>(cloud_profile);
+    let system_config = SystemConfig::try_new(tedge_config.root_dir()).unwrap_or_default();
```
didier-wenzek:
```suggestion
    let system_config = tedge_config.read_system_config();
```

albinsuresh: Resolved by [c22b8f8](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c22b8f878b574e75af8a156f40e310b0986f5559)

crates/core/tedge/src/cli/mod.rs
albinsuresh: Resolved by [c22b8f8](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c22b8f878b574e75af8a156f40e310b0986f5559)

crates/core/tedge/src/cli/mod.rs
albinsuresh: Resolved by [c22b8f8](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c22b8f878b574e75af8a156f40e310b0986f5559)

crates/core/tedge/src/cli/connect/command.rs
albinsuresh: Resolved by [c22b8f8](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c22b8f878b574e75af8a156f40e310b0986f5559)

crates/core/tedge_mapper/src/c8y/mapper.rs
albinsuresh: Resolved by [c22b8f8](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c22b8f878b574e75af8a156f40e310b0986f5559)

crates/core/tedge/src/cli/init.rs
albinsuresh: Resolved by [c22b8f8](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c22b8f878b574e75af8a156f40e310b0986f5559)

configuration/package_scripts/tedge/preinst
albinsuresh: Resolved by [c22b8f8](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/c22b8f878b574e75af8a156f40e310b0986f5559)

crates/core/tedge/src/cli/init.rs
albinsuresh: Resolved by [0c56812](https://github.com/thin-edge/thin-edge.io/pull/4094/commits/0c56812a46185c3fc1ccb7dd578915147be7147a)

configuration/package_scripts/tedge/preinst
```diff
@@ -29,31 +29,66 @@ user_exists() {
...
+
+# Read user/group from /etc/tedge/system.toml if it exists
+TEDGE_USER="tedge"
+TEDGE_GROUP="tedge"
+SYSTEM_TOML="/etc/tedge/system.toml"
+if [ -f "$SYSTEM_TOML" ]; then
+    _val=$(parse_toml_string_value "user" "$SYSTEM_TOML")
+    TEDGE_USER="${_val:-$TEDGE_USER}"
+    _val=$(parse_toml_string_value "group" "$SYSTEM_TOML")
+    TEDGE_GROUP="${_val:-$TEDGE_GROUP}"
```
reubenmiller: There is a slight logic error here when the system.toml has empty values, e.g.

```sh
user = ""
group = ""
```

But I can apply the fix for this.

crates/common/tedge_config/src/tedge_toml/tedge_config_location.rs
```diff
@@ -470,7 +471,10 @@ impl TEdgeConfigLocation {
 
         atomically_write_file_async(toml_path, toml.as_bytes()).await?;
 
-        if let Err(err) = change_user_and_group(toml_path, "tedge", "tedge").await {
+        let system_config = SystemConfig::try_new(&self.tedge_config_root_path).unwrap_or_default();
+        if let Err(err) =
+            change_user_and_group(toml_path, &system_config.user, &system_config.group).await
```
reubenmiller: I suspect we could move the guard around the empty user and/or group condition to inside the `change_user_and_group` function (technically inside the private `change_user_and_group_sync` function).

As doing so would ensure that any parts of the code that use that function directly (which there are a few places), would get the same handling.

Examples of direct calls to the `change_user_and_group`:
* https://github.com/albinsuresh/thin-edge.io/blob/9c464e6444c1a017eae04428ad0c86833688bb9a/crates/core/tedge/src/cli/connect/command.rs#L974
* https://github.com/albinsuresh/thin-edge.io/blob/9c464e6444c1a017eae04428ad0c86833688bb9a/plugins/c8y_remote_access_plugin/src/lib.rs#L107

configuration/package_scripts/tedge/preinst
```diff
@@ -29,31 +29,67 @@ user_exists() {
...
+    if [ -n "$MATCHES" ]; then
+        # property is defined, so respect all values, even empty ones
+        echo "$MATCHES" | sed 's/.*=[[:space:]]*"\([^"]*\)".*/\1/'
     else
-        echo "WARNING: Could not create group: tedge" >&2
+        # not-defined so the given default value
+        echo "Using default as no valid value for '$property' was found in '$file_path'" >&2
+        echo "$default_value"
+    fi
+}
```
Bravo555, line 48: note: this could fail if in the future we were to add `user` or `group` key under any other table (or for any other key if there's more than one)

crates/core/tedge/src/cli/init.rs
```diff
@@ -103,14 +102,52 @@ impl TEdgeInitCmd {
...
+        } else {
+            let agent_state_dir = config_dir.join(".agent");
+            create_directory_and_update_ownership(&agent_state_dir, &permissions).await?;
+            agent_state_dir
+        };
 
-        if entity_store_file.exists() {
-            change_user_and_group(entity_store_file, &self.user, &self.group).await?;
+        let entity_store_file = agent_state_dir.join("entity_store.jsonl");
+        if entity_store_file.exists() && !user.is_empty() && !group.is_empty() {
```
Bravo555: suggestion: don't have to check if `user`/`group` are empty because it's already done by `PermissionEntry::apply`.

```suggestion
        if entity_store_file.exists() {
```

crates/core/tedge/src/cli/certificate/cli.rs
```diff
@@ -229,9 +230,10 @@ pub enum CA {
 impl BuildCommand for TEdgeCertCli {
     async fn build_command(self, config: &TEdgeConfig) -> Result<Box<dyn Command>, ConfigError> {
         let (user, group) = if config.mqtt.bridge.built_in {
-            ("tedge", "tedge")
+            let system_config = SystemConfig::try_new(config.root_dir()).unwrap_or_default();
```
rina23q: It looks forgotten to change:

```suggestion
            let system_config = config.read_system_config();
```

crates/common/tedge_utils/src/file.rs
```diff
@@ -399,6 +399,18 @@ impl PermissionEntry {
     }
 }
 
+/// Creates a `PermissionEntry` from plain `user` and `group` strings.
+///
+/// Empty strings are treated as unset and mapped to `None`, so ownership
+/// is left unchanged for whichever field is empty.
+pub fn permissions(user: &str, group: &str, mode: u32) -> PermissionEntry {
```
rina23q, line 406: (trivial) I think this function deserves staying under `PermissionEntry` struct, however, it's fine.

tests/RobotFramework/tests/installation/install_tedge.robot
```diff
@@ -49,10 +55,41 @@ Install then uninstall latest tedge via script (from main branch)
...
+    ...    before installation forces tedge to use that user for all of its file ownership.
+
+    Create Custom User And Group
+    Install Tedge With System Config    empty_user_system.toml
+
+    # Set the desired tedge user manually, instead of specifying it in system.toml
+    Execute Command    tedge init --user ${CUSTOM_USER} --group ${CUSTOM_GROUP}
+
+    Setup Systemd Overrides For Custom User
+    Verify Tedge Running With Custom User
```
Bravo555: question: Not sure exactly what this is testing.

I assume the description is out of date and what it _should_ say instead is that we check that when user/group is empty then installation shouldn't change current ownership. But to test that we'd have to have some previous files and after install verify owner didn't change. But here we're doing `tedge init` which explicitly overwrites the owner.

tests/RobotFramework/tests/customizing/tedge_init.robot
```diff
@@ -67,10 +68,113 @@ Tedge init and check if default values are restored
...
+    Execute Command    sudo c8y-remote-access-plugin --init
+    All Tedge Directories Should Be Owned By    tedge:tedge
+    [Teardown]    Restore Initial State
+
+Tedge init leaves ownership unchanged when user and group are empty in system.toml
+    Transfer To Device    ${CURDIR}/resources/empty_user_empty_group_system.toml    /etc/tedge/system.toml
+    Execute Command    sudo tedge init
+    Execute Command    sudo c8y-remote-access-plugin --init
+    All Tedge Directories Should Be Owned By    root:root
+    [Teardown]    Restore Initial State
```
Bravo555, line 90: question: By "unchanged" we mean that `init` commands recreate the files so their new owner becomes whatever user runs the `init` command, and **not** that existing files do not have their owner changed?

reubenmiller: In this case, generally an explicit "empty" user/group means that the process can't change ownership of files/folders anyway, so attempting to would be anyway pointless.

So yes, new files should then have the owner set to the current processes owner...but I think this happens  without explicitly doing anything.

tests/RobotFramework/tests/installation/install_tedge.robot
reubenmiller: It looks to be testing two things, the pre-seeding (e.g. making sure the preinst part of the package doesn't try to apply the "tedge" user), ~and the second is to check that the `--user` and `--group` flags override what ever is set in the system.toml~

update:

Looking at it again, it is testing a corner case where the user wants to handle the entire initialization themselves whilst still using the debian package to install thin-edge.io. The empty values in the system.toml should control the preinst and postinst logic...though tbh this test case doesn't provide much value, as this is not the recommended sequence. If you want to use a custom user, then just set explicit values in the system.toml...if you want to run as the current user, then you should not be using the debian package.

configuration/package_scripts/tedge/preinst
reubenmiller: This can't happen as we're piping to ` | head -n1` (after the grep) ;)

reubenmiller: But yes, the parsing is not ideal, but we can live with it since we don't generally add new properties to the system.toml, and until we have a more generic toml cli commands, then we can't do anything else (the downside of the toml format tbh).

crates/common/tedge_utils/src/file.rs
reubenmiller: I can't judge tbh, so if this is to be done, then someone else will need to push the fix.

crates/core/tedge/src/cli/init.rs
reubenmiller: Resolved by 62b36c459b3258b3cb3debb29bae3e9ee11b2b17

crates/core/tedge/src/cli/certificate/cli.rs
reubenmiller: Resolved by 714ba74268011d69f990fbf77172372031a2f23c

crates/common/tedge_utils/src/file.rs
rina23q: I would say keep it as it is for now. I would rather like to suggest a refactoring of this file after a release.

tests/RobotFramework/tests/customizing/tedge_init.robot
```diff
@@ -67,10 +68,113 @@ Tedge init and check if default values are restored
...
+
+Tedge init defaults both user and group to tedge when system.toml has an invalid user value
```
rina23q, line 92: I am a bit surprised to see that we force to change `tedge:tedge` when `system.toml` contains invalid user or group value.

```toml
user = 42
group = "tedge"
```

I think it would be better if it doesn't change anything and throws an error, so that users can aware of their mistake and modify the wrong `system.toml`.

tests/RobotFramework/tests/installation/install_tedge.robot
rina23q: I don't think this test is useful either, since the behaviour when defining empty values for `user`/`group` are well covered in `customizing/tedge_init.robot`.

rina23q (approved):
I think the code and test coverage are good. I left a comment, but it doesn't intend to block the PR. I just wanted to express my opinion.

tests/RobotFramework/tests/customizing/tedge_init.robot
reubenmiller: Your suggestion might be more correct, however given the lack of a formal toml parser on the command line, we won't be able to do this easily, and generally the `tedge init` will fail when using tedge:tedge as the user won't exist (hence why the user was editing the user/group in the system.toml in the first place).

reubenmiller: so I'd just suggest leaving it as is for now

tests/RobotFramework/tests/installation/install_tedge.robot
reubenmiller: Removed in 9f82209e72c075c37dbc1acf8831f99fe2258f50

Issue #4019, opened 2026-03-03
Allow flow to lookup the device.id of any cloud identity for a given device, child-device or service


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

When writing a custom flow to provide mapping logic for a custom cloud, it is often that the flow will require access to the `device.id` (Common Name used in the device certificate).

Though the same requirement could also be extended to getting the external id of any device, child-device or service of a given mapper.

For example, the [ThingsBoard flow](https://github.com/thin-edge/tedge-flows-examples/blob/main/flows/thingsboard-registration/flow.toml) requires the device.id when registering itself.

At the moment, the ThingsBoard flow example requires the user to define it in the flows config manually:

```toml
[[steps]]
script = "lib/main.js"
config.main_device_name = "{{MAIN_DEVICE_NAME}}"
config.default_device_profile = "default"
config.max_pending_messages = 100
```

**Describe the solution you'd like**

A few potential options would be (though feel free to add your own):

#### Option 1: Provide the device.id value via the context.

```js
export function onMessage(message, context) {
    console.log(`device.id: ${context.device_id}`);
    return [];
}
```

* The `device_id` would be populated with the `device.id` applicable to the mapper's context. For example, when running under the `tedge-mapper c8y`, the value from `c8y.device.id`, and the `tedge-mapper aws` would be `az.device.id`. For the `tedge-mapper local`, the `device.id` value could be used.

#### Option 2: Expose all tedge configuration values under `context.tedge`

This would be slightly more complicated for users to access as they would need to be aware in which context they're running in, e.g. as c8y mapper, az mapper etc.

```js
export function onMessage(message, context) {
    console.log(`device.id: ${context.tedge.device.id}`);
    return [];
}
```

#### Option 3: Expose the entity store (and extend the entity store to also include the mapper specific ids)

```js
export function onMessage(message, context) {
    const [_, part1, part2, part3, part4, ...other] = message.split("/");
    const identity = context.entity.get(`${part1}/${part2}/${part3}/${part4}`);
    console.log(`identity: ${identity}`);
    return [];
}
```

**Describe alternatives you've considered**

**Additional context**

PR #4095, opened 2026-04-01
feat: flows configured after mapper config

didier-wenzek:
## Proposed changes

- [x] `${mapper.*}` expressions in flow definitions
- [x] mappers uses their `mapper.toml` to substitute `${mapper.*}` expressions of their flows
- [x] `tedge flows test` uses `mapper.toml` of the mapper under test

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/4019

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/params.rs
+++ b/crates/extensions/tedge_flows/src/params.rs
@@ -168,7 +168,7 @@ impl<'a, T: MapperParams + ?Sized> Params<&'a T> {
         };
 
         if let Some(path) = path.strip_prefix("params") {
-            match Self::get(&self.params, path.strip_prefix('.')) {
+            match get_inner_value(&self.params, path.strip_prefix('.')) {
                 Ok(value) => Ok(value),
                 Err(unknown_path) => Err(LoadError::UnknownParam {
                     path: format!("params.{unknown_path}",),
@@ -218,28 +218,28 @@ impl<'a, T: MapperParams + ?Sized> Params<&'a T> {
             })
             .collect()
     }
+}
 
-    fn get(values: &Map<String, Value>, steps: Option<&str>) -> Result<Value, String> {
-        let Some(path) = steps else {
-            return Ok(values.clone().into());
-        };
-        let (key, next_steps) = match path.split_once(".") {
-            None => (path, None),
-            Some((key, inner_path)) => (key, Some(inner_path)),
-        };
-        let Some(value) = values.get(key) else {
-            return Err(key.to_string());
-        };
-        if next_steps.is_none() {
-            return Ok(value.clone());
-        }
-        let Some(inner_values) = value.as_object() else {
-            return Err(format!("{key}.*"));
-        };
-        match Self::get(inner_values, next_steps) {
-            Ok(value) => Ok(value),
-            Err(unknown_path) => Err(format!("{key}.{unknown_path}")),
-        }
+fn get_inner_value(values: &Map<String, Value>, steps: Option<&str>) -> Result<Value, String> {
+    let Some(path) = steps else {
+        return Ok(values.clone().into());
+    };
+    let (key, next_steps) = match path.split_once(".") {
+        None => (path, None),
+        Some((key, inner_path)) => (key, Some(inner_path)),
+    };
+    let Some(value) = values.get(key) else {
+        return Err(key.to_string());
+    };
+    if next_steps.is_none() {
+        return Ok(value.clone());
+    }
+    let Some(inner_values) = value.as_object() else {
+        return Err(format!("{key}.*"));
+    };
+    match get_inner_value(inner_values, next_steps) {
+        Ok(value) => Ok(value),
+        Err(unknown_path) => Err(format!("{key}.{unknown_path}")),
     }
 }
 


crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -162,9 +162,15 @@ impl TEdgeComponent for CumulocityMapper {
...
+        let mapper_dir = cfg_dir.join("mappers").join("c8y");
         let flows_dir =
             tedge_flows::flows_dir(cfg_dir, "c8y", self.profile.as_ref().map(|p| p.as_ref()));
-        let mut flows = crate::flow_registry(flows_dir).await?;
+        let mapper_config = crate::effective_mapper_config(
+            &tedge_config,
+            "c8y",
+            mapper_dir,
+        ).await?;
+        let mut flows = crate::flow_registry(mapper_config, flows_dir).await?;
```
didier-wenzek: This sequence of function calls is repeated for all the mappers and is fragile (repeated cloud name and redundant paths). => I will refactor that.

crates/core/tedge/src/cli/flows/cli.rs
```diff
@@ -198,7 +199,8 @@ impl TEdgeFlowsCli {
         flows_dir: &Utf8PathBuf,
         js_config: JsRuntimeConfig,
     ) -> Result<MessageProcessor<BaseFlowRegistry>, LoadError> {
-        let mut registry = BaseFlowRegistry::new(flows_dir);
+        let mapper_config = HashMap::new();
```
didier-wenzek: The proper `mapper.toml` must be used instead of an empty set of mapper parameters.

crates/core/tedge_mapper/src/lib.rs
```diff
@@ -275,11 +281,41 @@ pub(crate) async fn flow_registry(
...
+    };
+    match mapper_name {
+        #[cfg(feature = "c8y")]
+        "c8y" => Ok(Some(
+            c8y::mapper::resolve_effective_mapper_config(tedge_config, None).await?,
+        )),
+        #[cfg(feature = "aws")]
+        "aws" => Ok(None),
+        #[cfg(feature = "azure")]
+        "az" => Ok(None),
```
didier-wenzek: This will be done once merged https://github.com/thin-edge/thin-edge.io/pull/4089

tests/RobotFramework/tests/cumulocity/flows/c8y_flows.robot
```diff
@@ -54,6 +54,44 @@ Get entity metadata from the c8y mapper context
     ...    series=temperature
     Should Be Equal As Numbers    ${measurements[0]["temperature"]["temperature"]["value"]}    23.1
 
+Get device.id from the c8y mapper config
+    Execute Command    tedge config upgrade
+    Restart Service    tedge-mapper-c8y
```
didier-wenzek, line 59: Updating the mapper configuration is mandatory, as this feature is using the `mapper.toml` files.

crates/core/tedge/src/cli/flows/cli.rs
didier-wenzek: Done

crates/core/tedge_mapper/src/lib.rs
didier-wenzek: Done

docs/src/references/mappers/flows.md
```diff
@@ -316,6 +316,46 @@ For topics, commands, paths and intervals, several substitutions can be applied
...
+key_path  = "key.pem"
+
+# derived
+# id = derived from the device certificate
+
+# user defined
+arch = "ARMv8.7-A"
+```
+
+the following makes the mapper cloud url and the %%te%% device id and arch
```
reubenmiller:
```suggestion
The following makes the mapper cloud url and the %%te%% device id and arch
```

docs/src/references/mappers/flows.md
```diff
@@ -316,6 +316,46 @@ For topics, commands, paths and intervals, several substitutions can be applied
...
+available to the transformation scripts via the `context.config` object:
+
+```toml
+input.mqtt.topics = ["sensor/#"]
+
+[config]
+cloud.url = "${mapper.url}"
+device.id = "${mapper.device.id}"
+device.arch = "${mapper.device.arch}"
```
reubenmiller: (optional) It might be worth adding a toml comment above the `${mapper.device.arch}` indicating it is a custom user-defined value just in case if people didn't read the prelude and think that it is part of thin-edge.io


```suggestion
# user defined value
device.arch = "${mapper.device.arch}"
```

docs/src/references/mappers/flows.md
```diff
@@ -316,6 +316,46 @@ For topics, commands, paths and intervals, several substitutions can be applied
 as in `input.mqtt.topics = [ "te/device/${params.child}/service/${params.service}/m/${params.measurement.type}" ]`
 :::
 
+### Mapper Configuration Values
+
+Flows can be parameterized with configuration values specific to the mapper running these flows.
+
+All the values defined in the [`mapper.toml`](../user-defined-mappers/#mappertoml) file — including user-defined ones —  
+can substitute [`${mapper.*}` expressions](../user-defined-mappers/#template-variables-mapper)
+in a flow `.toml` definition, in a similar manner as for [parameters](#parameters).
```
reubenmiller, line 325: Something that was slightly unexpected (but might be ok), the `${mapper.device.id}` value is only available for the c8y mapper is if the users has migrated the config using `tedge config upgrade`, and restarted the tedge-mapper-c8y.

But I guess annoyance would be removes if we were to automatically execute the `upgrade` when installing the new version as detailed in https://github.com/thin-edge/thin-edge.io/issues/4097

docs/src/references/mappers/flows.md
reubenmiller: And personally I was a bit surprised that we support user-defined values, but a nice touch.

crates/extensions/tedge_flows/src/params.rs
```diff
@@ -121,19 +162,29 @@ impl Params {
...
             .and_then(|path| path.strip_suffix("}"))
+            .map(|path| path.trim())
         else {
             return Ok(self.substitute_inner_paths(expr).into());
         };
-        let Some(path) = path.trim().strip_prefix("params") else {
-            return Err(LoadError::UnknownParam {
+
+        if let Some(path) = path.strip_prefix("params") {
+            match Self::get(&self.params, path.strip_prefix('.')) {
```
jarhodes314: It feels a bit weird to be passing data from `self` to a static method on `Self`. Is there a reason why `get` doesn't take `&self`?

docs/src/references/mappers/flows.md
didier-wenzek: @jarhodes314

I'm using `resolve_effective_mapper_config` to read the mapper configuration. However this works only after a `tedge config upgrade`.

- Do I have to do something specific to read the mapper config from the main `tedge.toml`  when there is no `mapper.toml`?
- Would it be simpler to simply force a config upgrade for the 2.0?

docs/src/references/mappers/flows.md
didier-wenzek: > And personally I was a bit surprised that we support user-defined values, but a nice touch.

This is using the same code as for bridge rules, hence inheriting the same features (as that one) and constraints (as assuming a `tedge config upgrade`)

didier-wenzek: Added https://github.com/thin-edge/thin-edge.io/pull/4095/commits/e851c1e40d2b34d694149c2473fc57b1bd545a48

docs/src/references/mappers/flows.md
didier-wenzek: Fixed https://github.com/thin-edge/thin-edge.io/pull/4095/commits/e851c1e40d2b34d694149c2473fc57b1bd545a48

docs/src/references/mappers/flows.md
jarhodes314: Yeah, I've tried quite hard in the custom mapper to support `tedge.toml` as well as it supports `mapper.toml`, but I think we're now very much at the point where we have a lot of functionality that is more likely to work as expected if the config is upgraded rather than not.

docs/src/references/mappers/flows.md
```diff
@@ -316,6 +316,46 @@ For topics, commands, paths and intervals, several substitutions can be applied
...
+
+For instance, given a `mapper.toml`, which combines connection settings, derived values (such as the device id)
+as well as user-defined values (here about the sensor):
+
+```toml
```
Bravo555: suggestion:

```suggestion
```toml title="mapper.toml"
```

docs/src/references/mappers/flows.md
```diff
@@ -316,6 +316,46 @@ For topics, commands, paths and intervals, several substitutions can be applied
...
+# id = derived from the device certificate
+
+# user defined
+arch = "ARMv8.7-A"
+```
+
+The following makes the mapper cloud url and the %%te%% device id and arch
+available to the transformation scripts via the `context.config` object:
+
+```toml
```
Bravo555: suggestion:

```suggestion
```toml title="flow.toml"
```

docs/src/references/mappers/flows.md
reubenmiller: Yeah I'm fine leaving it as is (requiring the `tedge config upgrade` to be run). So don't worry about finding a solution, as it will just complicate the code, and we'll push to get the changes to run the upgrade automatically at install time next week anyway, which will remove the need for this all-together.

crates/extensions/tedge_flows/src/params.rs
didier-wenzek: The reason is that this function is recursively called on sub-objects of the former params object.

Renamed to `get_inner_value` and move out of `Params`:  https://github.com/thin-edge/thin-edge.io/pull/4095/changes/d5a6ac8bccec3759c8a5a57be7b0b823b61d3de2

crates/extensions/tedge_flows/src/params.rs
```diff
@@ -23,61 +50,75 @@ use tracing::log;
 /// config.y = "${params.x.y}"
 /// config.debug = "${params.debug}"
 /// ```
-#[derive(serde::Deserialize, Default)]
-pub struct Params {
+pub struct Params<T> {
+    mapper: T,
     params: Map<String, Value>,
 }
```
Bravo555, line 56: thought: I understand the use now, but long-term coupling of `Params` to mapper-related concepts, seems suboptimal.

It makes sense that `Params` was supposed to read only from `params.toml` and since we now support `${mapper.*}` lookups to add it directly here, since this is where the substitution logic is, but it's starting to get a bit big and to be able to better add various new paths in the future, after the relase we could separate it as follows:

- reading any toml file and converting it to map available for lookups is one thing
- combining multiple maps into bigger maps to add new data is one thing 
- substitution logic should just work for any map and not special-case `params.`, `mapper.` or any special fields, and that's another thing

didier-wenzek: > thought: I understand the use now, but long-term coupling of `Params` to mapper-related concepts, seems suboptimal.

Yeah, the dependency is not the direct as `Params` exposes a trait that is implemented by the mappers without making the flows aware of any mapper aspects.

But, I get your point.

> It makes sense that `Params` was supposed to read only from `params.toml` and since we now support `${mapper.*}` lookups to add it directly here, since this is where the substitution logic is, but it's starting to get a bit big

Indeed bigger than I would expect. The main reason is that the mapper parameters have a different timeline (read once, and used every time a `params.toml`  file is read).

> to be able to better add various new paths in the future, after the relase we could separate it as follows:
> 
>     * reading any toml file and converting it to map available for lookups is one thing
> 
>     * combining multiple maps into bigger maps to add new data is one thing
> 
>     * substitution logic should just work for any map and not special-case `params.`, `mapper.` or any special fields, and that's another thing

That is a good suggestion. The complicated part will be to combine the `mapper` sub-map with the `params` sub-map using a reference to the former and not a copy. But this should be feasible.

docs/src/references/mappers/flows.md
didier-wenzek: Done - but directly squashed.

docs/src/references/mappers/flows.md
didier-wenzek: Done - but directly squashed.

PR #4101, opened 2026-04-02
feat: `tedge connect --test` support for custom mappers

jarhodes314:
## Proposed changes
- [x] Add support for `tedge connect --test` to custom mappers
- [x] Rename `--debug` to `--show-all` in `tedge bridge inspect` to prevent conflict with existing `--debug` option

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge/src/cli/connect/command.rs
+++ b/crates/core/tedge/src/cli/connect/command.rs
@@ -1816,40 +1816,6 @@ Each cloud profile requires either a unique URL or unique device ID, so it corre
         }
 
         #[tokio::test]
-        async fn skips_bridge_phase_for_flows_only_mapper() {
-            let broker = mqtt_tests::test_mqtt_broker();
-            let ttd = TempTedgeDir::new();
-            // No bridge/ dir — flows-only mapper
-            let config = config_pointing_at_broker(broker.port, &ttd);
-
-            broker
-                .publish_with_opts(
-                    "te/device/main/service/tedge-mapper-flowsonly/status/health",
-                    r#"{"status":"up"}"#,
-                    rumqttc::QoS::AtLeastOnce,
-                    true,
-                )
-                .await
-                .unwrap();
-
-            // Should return Ok immediately after mapper health, without waiting
-            // for a bridge health that never comes
-            wait_for_custom_mapper_health(&config, "flowsonly")
-                .await
-                .expect("health check should succeed without bridge");
-
-            broker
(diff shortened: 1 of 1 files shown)


reubenmiller:
@jarhodes314 Everything worked nicely, the only suggestion would be with the `tedge connect <name> --test` command if we could use a different exit code when the mapper service is up, but the bridge is not connected, as the follow up action would be different, but only if this is simple todo.

didier-wenzek (approved):
Approved

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -1541,4 +1733,173 @@ Each cloud profile requires either a unique URL or unique device ID, so it corre
...
+            broker
+                .publish_with_opts(
+                    "te/device/main/service/tedge-mapper-flowsonly/status/health",
+                    "",
+                    rumqttc::QoS::AtLeastOnce,
+                    true,
+                )
+                .await
+                .unwrap();
+        }
```
Bravo555: question: How is this test different from `succeeds_when_mapper_health_arrives`? Seems only comments and name of the mapper are different.

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -526,6 +548,171 @@ impl ConnectCommand {
...
+            }
+            Err(e) => {
+                fatal = Some(ConnectError::Anyhow(
+                    anyhow::Error::from(e)
+                        .context("Failed to connect to mosquitto for connection check"),
+                ));
+                break;
+            }
+            _ => {}
+        }
```
Bravo555, line 638: thought: With another case (apart from other cloud-specific connection clients) of an MQTT client that connects, only waits for 1 or 2 messages, and disconnects, it would be nice to create small layer on rumqttc that provides a simpler API for these common operations.

crates/core/tedge/src/cli/connect/command.rs
jarhodes314: Deleted this now because it is just a duplicate

crates/core/tedge/src/cli/connect/command.rs
jarhodes314: Good suggestion, I'll handle this in a separate PR as some care is probably needed to ensure this all works properly and what we have at the moment is sufficient for the release.

PR #4111, opened 2026-04-08
refactor: Split into tedge-p11 and tedge-p11-server crates

Bravo555:
## Proposed changes

Make `tedge-p11-server` only contain the binary target and extract the everything else into a separate `tedge-p11` crate.

This is done because tedge-config depends on the p11 connectors via the certificate crate, so we'd like to avoid pulling in unnecessary dependencies into tedge-config that are used by the tedge-p11-server binary.

This allows us to use tedge-config as a dependency in tedge-p11-server, whereas previously this would result in an circular dependency error (`tedge-p11-server <- certificate <- tedge-config <- tedge-p11-server`).

The PR should have no functional changes.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/common/certificate/Cargo.toml
```diff
@@ -25,7 +25,7 @@ reqwest = { workspace = true, optional = true, features = [
 rustls = { workspace = true }
 rustls-native-certs = { workspace = true }
 sha1 = { workspace = true }
-tedge-p11-server = { workspace = true }
+tedge-p11 = { workspace = true }
```
albinsuresh, line 28: Though this is definitely an improvement over what we previously had, seeing a `common` crate having a dependency on an `extensions` crate still doesn't feel right. Can we go one step further and extract a very minimal crate the `tedge-p11` crate into another `common` crate (e.g: `tedge-p11-api`) and have this crate depend on that one?

Bravo555: > seeing a common crate having a dependency on an `extensions` crate still doesn't feel right

I agree, moved the `tedge-p11` into `common`.

> Can we go one step further and extract a very minimal crate the `tedge-p11` crate into another `common` crate (e.g: `tedge-p11-api`) and have this crate depend on that one?

Here I don't understand the motivation of why split it again instead of just putting the entirety of `tedge-p11` in `common`. The way I see this, `tedge-p11` is pretty much analogous to `certificate` crate, which is in `common`, so it should be fine to put `tedge-p11` there as well, right?

I'd agree with this additional split if there were either 1) clear dependency graph problems that would be fixed by doing this (like splitting `tedge-p11-server` achieves), or 2) if some of the code in `tedge-p11` clearly didn't fit as _common_ code, or was mostly _extensions_ code so common is a wrong place for it.

albinsuresh: > Here I don't understand the motivation of why split it again instead of just putting the entirety of tedge-p11 in common.

I was suggesting that on seeing the `certificate` crate having only 2 lightweight dependencies on `tedge_p11::CryptokiConfig` and `tedge_p11::pkcs11::SigScheme`. So, I thought we could just extract that instead of pulling in everything the `proxy::client` bits, which didn't feel "common"

> The way I see this, tedge-p11 is pretty much analogous to certificate crate, which is in common, so it should be fine to put tedge-p11 there as well, right?

When you put it that way, yeah I think I agree.

albinsuresh (approved):

Issue #4115, opened 2026-04-09
Refactor `tedge_utils::file` and improve APIs


rina23q:
**Is your refactoring request related to a problem? Please describe.**
Refactor [`tedge_utils::file`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/common/tedge_utils/src/file.rs) to clean up public APIs. 

While reviewing [this PR](https://github.com/thin-edge/thin-edge.io/pull/4094#discussion_r3027901049), I discovered that the APIs are partially duplicated and offers too many options. I believe some APIs can be changed to private, and some can be merged.

Just for an example, currently, we have those public APIs to create a directory:
* create_directory()
* create_directory_and_update_ownership()
* create_directory_with_defaults()
* create_directory_with_user_group()
* create_directory_with_user_group_and_update_ownership()
* PermissionEntry::create_directory()

In addtion, `PermissionEntry` can be improved. It does not handle when `user` or `group` is empty string. Recently added `permissions()` function does this correctly, however, I believe, this should have been addressed inside `PermissionEntry`. 
https://github.com/thin-edge/thin-edge.io/blob/219081c539ae3c3e8986115670e5426405a5a8b6/crates/common/tedge_utils/src/file.rs#L405-L411

Also, recently updated `change_user_and_group_sync()` addresses the case where `user` or `group` is an empty string. 
https://github.com/thin-edge/thin-edge.io/blob/219081c539ae3c3e8986115670e5426405a5a8b6/crates/common/tedge_utils/src/file.rs#L458-L464

**Describe the solution you'd like**
Visit `tedge_utils::file` and re-design private/public APIs, and update the APIs usages in the Rust code.

**Describe alternatives you've considered**

**Additional context**
This should be a pure refactoring, no behavior change is expected.

rina23q:
My goal has been achieved by #4128 and #4142 refactoring. Now we all have to use only `TedgePaths` and `ManagedDir::ensure()` API.

PR #3983, opened 2026-02-13
fix: Don't take ownership of existing operations directory in remote access plugin init

jarhodes314:
## Proposed changes
Fix an issue where the `c8y-remote-access-plugin --init` command changes ownership of the `/etc/tedge/operations/c8y` directory when it already exists.

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
<br/>

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s. You can activate automatic signing by running `just prepare-dev` once)
- [x] I ran `just format` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `just check` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #4117, opened 2026-04-09
refactor: consolidate file/directory/permissions utilities

rina23q:
## Proposed changes
The purpose is to clean up APIs available in `tedge_utils::file`.

This PR does:
* reduces the number of public APIs to create a directory.
* introduces a wrapper function `apply_permissions()` to apply permissions to keep consistent with `create_directory()` and `create_file()`.
* changes most of methods under `PermissionEntry` to private.
* also changes a lot of functions to private, which are called only by the same crate.
* and makes `change_user_and_group()` private as this one can be replaced with `apply_permissions()`.
* removes unused APIs (`get_filename()`, `get_uid_by_name()`, `get_gid_by_name()`)

## Types of changes

## Paste Link to the issue
#4115 

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/file.rs
+++ b/crates/common/tedge_utils/src/file.rs
@@ -73,52 +73,32 @@ pub enum FileError {
     },
 }
 
+/// Create a directory with the given permissions.
+///
+/// If the directory already exists, permissions are left unchanged.
 pub async fn create_directory(
     dir: impl AsRef<Path>,
     permissions: &PermissionEntry,
 ) -> Result<(), FileError> {
-    permissions.create_directory(dir.as_ref()).await
+    permissions.create_directory(dir.as_ref(), false).await
 }
 
+/// Create a directory with the given permissions, updating ownership even if it already exists.
 pub async fn create_directory_and_update_ownership(
     dir: impl AsRef<Path>,
     permissions: &PermissionEntry,
 ) -> Result<(), FileError> {
-    permissions
-        .clone()
-        .force_dir_ownership()
-        .create_directory(dir.as_ref())
-        .await
+    permissions.create_directory(dir.as_ref(), true).await
 }
 
-/// Create the directory owned by the user running this API with default directory permissions
+/// Create a directory owned by the user running this process with default permissions.
 pub async fn create_directory_with_defaults(dir: impl AsRef<Path>) -> Result<(), FileError> {
     create_directory(dir, &PermissionEntry::default()).await
 }
 
-pub async fn create_directory_with_user_group(
-    dir: impl AsRef<Path>,
-    user: &str,
-    group: &str,
-    mode: u32,
-) -> Result<(), FileError> {
-    let perm_entry = PermissionEntry::new(Some(user.into()), Some(group.into()), Some(mode));
-    perm_entry.create_directory(dir.as_ref()).await
-}
-
-pub async fn create_directory_with_user_group_and_update_ownership(
-    dir: impl AsRef<Path>,
-    user: &str,
-    group: &str,
-    mode: u32,
-) -> Result<(), FileError> {
-    let perm_entry = PermissionEntry::new(Some(user.into()), Some(group.into()), Some(mode));
-    perm_entry
-        .force_dir_ownership()
-        .create_directory(dir.as_ref())
-        .await
-}
-
+/// Create a file with the given permissions and optional default content.
+///
+/// If the file already exists it is left unchanged.
 pub async fn create_file(
     file: impl AsRef<Path>,
     content: Option<&str>,
@@ -127,36 +107,34 @@ pub async fn create_file(
     permissions.create_file(file.as_ref(), content).await
 }
 
-/// Create the directory owned by the user running this API with default file permissions
-pub async fn create_file_with_defaults(
-    file: impl AsRef<Path>,
-    content: Option<&str>,
+/// Apply the given permissions (owner, group, mode) to an existing path.
+pub async fn apply_permissions(
+    path: impl AsRef<Path>,
+    permissions: &PermissionEntry,
 ) -> Result<(), FileError> {
-    create_file(file, content, PermissionEntry::default()).await
+    permissions.apply(path.as_ref()).await
 }
 
-pub async fn create_file_with_mode(
-    file: impl AsRef<Path>,
-    content: Option<&str>,
-    mode: u32,
+/// Synchronous version of [`apply_permissions`].
+pub fn apply_permissions_sync(
+    path: impl AsRef<Path>,
+    permissions: &PermissionEntry,
 ) -> Result<(), FileError> {
-    let perm_entry = PermissionEntry::new(None, None, Some(mode));
-    perm_entry.create_file(file.as_ref(), content).await
+    permissions.apply_sync(path.as_ref())
+}
+
+/// Shorthand to create a [`PermissionEntry`] with all three fields set.
+///
+/// Empty strings are treated as unset, leaving the corresponding ownership unchanged.
+pub fn permissions(user: &str, group: &str, mode: u32) -> PermissionEntry {
+    PermissionEntry::default()
+        .with_user(user)
+        .with_group(group)
+        .with_mode(mode)
 }
 
 pub async fn path_exists(path: impl AsRef<Path>) -> bool {
     tokio::fs::try_exists(path).await.unwrap_or(false)
-}
-
-pub async fn create_file_with_user_group(
-    file: impl AsRef<Path>,
-    user: &str,
-    group: &str,
-    mode: u32,
-    default_content: Option<&str>,
-) -> Result<(), FileError> {
-    let perm_entry = PermissionEntry::new(Some(user.into()), Some(group.into()), Some(mode));
-    perm_entry.create_file(file.as_ref(), default_content).await
 }
 
 /// Moves a file to a destination path.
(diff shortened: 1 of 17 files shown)


crates/common/tedge_utils/src/file.rs
```diff
@@ -320,12 +327,12 @@ impl PermissionEntry {
         Ok(())
     }
 
-    async fn create_directory(&self, dir: &Path) -> Result<(), FileError> {
+    async fn create_directory(&self, dir: &Path, force_update: bool) -> Result<(), FileError> {
```
jarhodes314, line 330: The bool argument is somewhat opaque if you're just reading the function call without the definition also visible. I think this ought to be an enum:

```suggestion
    async fn create_directory(&self, dir: &Path, update_existing: UpdateExisting) -> Result<(), FileError> {
```

where the enum definition is:

```rust
#[derive(Debug, Copy, Clone)]
pub enum UpdateExisting {
    Yes,
    No,
}
```

rina23q: However, this `PermissionEntry::create_directory()` is private. Publicly, there are two different APIs to switch this boolean:`create_directory()` and `create_directory_and_update_ownership()`. Isn't it clear enough?

albinsuresh: I'm fine with the current explicitly named public APIs. An enum would have been better if we have plans to add more variations in the future, something like  "create_directory_with_parent_ownership()". In that case, an enum as follows might be marginally better:
```
enum PermissionMode {
    KeepExisting,
    ForceUpdate,
    MatchParent
}
```

crates/common/tedge_utils/src/file.rs
```diff
@@ -257,28 +235,57 @@ impl FileMoveError {
...
+    }
+
+    /// Set the file mode.
+    pub fn with_mode(mut self, mode: u32) -> Self {
+        self.mode = Some(mode);
         self
     }
 
-    pub async fn apply(&self, path: &Path) -> Result<(), FileError> {
+    async fn apply(&self, path: &Path) -> Result<(), FileError> {
```
albinsuresh, line 288: Not quite sure if that change was necessary. It forced many changes as follows:

**From:** 
```
permissions.apply(bridge_config_dir.as_std_path()).await
```
**To:** 
```
file::apply_permissions(bridge_config_dir.as_std_path(), &permissions).await
```
which I'm not really convinced is an improvement. The former version was definitely more concise. But I agree that the latter reads better.

albinsuresh (approved):
This refactoring/cleanup is much appreciated.

didier-wenzek:
A different approach to the same problem has been proposed here: https://github.com/thin-edge/thin-edge.io/pull/3983#pullrequestreview-3859352996

crates/common/tedge_utils/src/file.rs
```diff
@@ -257,28 +235,57 @@ impl FileMoveError {
 
 #[derive(Debug, PartialEq, Eq, Default, Clone)]
 pub struct PermissionEntry {
-    pub user: Option<String>,
-    pub group: Option<String>,
-    pub mode: Option<u32>,
-    pub reassert_dir_ownership: bool,
```
didier-wenzek, line 263: What has been the motivation for removing this flag?

I see this change as a regression. At least something is incoherent.

We should:
- either wrap all the file creation parameters in a struct driving the file creation
- or pass all these parameters to a file creation function

I have a preference for the former.

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -1165,7 +1165,14 @@ pub async fn chown_certificate_and_key(bridge_config: &BridgeConfig, tedge_confi
     // - When `tedge cert create` is not run as root, a certificate is created but owned by the user running the command.
     // - A better approach could be to remove this `chown` and run the command as mosquitto.
     let path = &bridge_config.bridge_certfile;
-    if let Err(err) = tedge_utils::file::change_user_and_group(path, &user, &group).await {
+    if let Err(err) = tedge_utils::file::apply_permissions(
+        path,
+        &tedge_utils::file::PermissionEntry::default()
+            .with_user(&user)
+            .with_group(&group),
+    )
```
didier-wenzek, line 1173: These lines highlight a mix of API style.

We have to choose between an API with a bunch of functions such as:

```
tedge_utils::file::change_user_and_group(path, &user, &group).await
```

or the builder pattern:

```
tedge_utils::file::PermissionEntry::default()
            .with_user(&user)
            .with_group(&group
            .apply(path)
           .await
```

The latter is more flexible.

didier-wenzek (commented):
I appreciate the motivation for this refactoring. However, we have first to agree on a style for the new API, the current proposal being a mix of approaches.

crates/core/tedge/src/cli/connect/command.rs
albinsuresh: Here is my PoV:

Wherever we expect to create new files/directories, use the helper functions like `create_directory_with_defaults`, `create_directory` and `create_directory_and_update_ownership` etc as it conveys the intent clearly, and is also concise.

But for updating the permissions of an existing file/directory is involved, I'd prefer to use `PermissionEntry::apply()` function directly rather than more helpers like `change_user_and_group`, `change_user_group_and_mode`, `change_mode` etc.

crates/common/tedge_utils/src/file.rs
rina23q: This `reassert_dir_ownership` is used only to create a directory. Hence, I think having the flag in `PermissionEntry` is too much, and it is not really a part of a permission.

After reading your comment, I am also in favor of the former approach.

crates/common/tedge_utils/src/file.rs
rina23q: Yeah, it was one part that I was struggling. On the one hand, there is `create_directory()` and similar functions. Then, making only `PermissionEntry::apply()` pubic? Inconsistency here.

---

Now I think `PermissionEntry` should not have had `create_directory()` and `create_file()`.

`PermissionEntry::apply(&path)` => Ok, permission is applied to the path.
`PermissionEntry::create_directory(&path)` => Sounds weird. Permission creates a directory? 

I am thinking to introduce two new structs, one for directory, and the other for file. I'll update later.

didier-wenzek: > Now I think `PermissionEntry` should not have had `create_directory()` and `create_file()`.
> 
> `PermissionEntry::apply(&path)` => Ok, permission is applied to the path. `PermissionEntry::create_directory(&path)` => Sounds weird. Permission creates a directory?
> 
> I am thinking to introduce two new structs, one for directory, and the other for file. I'll update later.

This can be solved with a single struct and appropriate naming. I suggest `FileUser`:

- `FileUser::create_directory(path)`
- `FileUser::create_file(path, content)`
- `FileUser::take_ownership(path)`

crates/common/tedge_utils/src/file.rs
jarhodes314: I agree with @rina23q here that the distinction between "a set of permissions for a file/directory" and "the mechanism we are using to apply said permissions" is fuzzy, and that's one of the problems with the existing API.

I feel one of the things the API seems to strive for is to be declarative - ignoring the `reassert_dir_ownership`, this API achieves that. Reasserting ownership is a reasonable thing to do, but I think the underlying issue is more that for a given file, we don't know which (if any) of the parent directories are conceptually "owned" by thin-edge. This solution to that problem feels like a hack - instead of the code naturally expressing what state the directories should be in, we're responding to a permissions issue observed in practice by walking the directory tree and setting some somewhat arbitrary permissions on each parent until we succeed at the task we originally set out to do.

didier-wenzek: > I feel one of the things the API seems to strive for is to be declarative

I feel the same and this is the style I would like to enforce: telling the files and directories the system expects to proceed, being new or not, user created or not.

> ignoring the reassert_dir_ownership

This can be done in a declarative style too, provided we distinguish expected ownership from default ownership. The former being optional and the latter used only if the file or directory has to be created.

>  we don't know which (if any) of the parent directories are conceptually "owned" by thin-edge.

Agree. We need to add a root directory to the concept of `PermissionEntry` to define expectations only for files and directories below that root.

jarhodes314:
Overall, there's some broad agreement that this code is problematic and this code does do some nice things to tidy up the code. However, I feel it's a bit surface-level - it's focused on tidying up some points where the code doesn't beautifully express itself, but leaving behind a lot of the core issues that prevent us from changing file-related code with confidence.

Recently with the `mapper.toml` changes, I've found I run into costly issues (e.g. waiting for the system tests to highlight obvious regressions in how we handle permissions) trying to modify the code. I think there are two true issues highlighted by this example:
1. Many different APIs for handling file permissions
2. Lack of ability to test the permissions in Rust code and an inability for that Rust code to truly express its intent

As for 1, this PR does somewhat solve this problem, but both @albinsuresh and @didier-wenzek have expressed some concerns about the approach of using free functions as opposed to methods and how that can impact code readability. In addition, there is still a problem in that some of the code uses `tokio::fs` and `std::fs` to create files and directories directly, giving us yet another way to handle some of these problems.

With 2, I think there's an acute issue in the `tedge_config` unit tests, where they don't assume `/etc/tedge` exists and this causes spurious permissions issues if we don't create the full directory structure when writing `mapper.toml` files. This meant that code that passed the system tests failed the unit tests and it wasn't obvious whether the unit tests were overly strict, or the system tests were missing something. It would be really useful if we could have specific tests that verify things like "this component is responsible for ensuring this directory exists with the default tedge ownership".

As for making the code express its intent better, for me this naturally makes me want a more declarative API, focusing on specifying the files and directories that should exist independently of the mechanism used to create them, along with a well thought through interface for the different actions we can take to ensure those exist, covering things like "should we create the parent directory for this file or fail", "should we take ownership of an existing directory" etc. I think this is similar to what @didier-wenzek proposed in https://github.com/thin-edge/thin-edge.io/pull/3983#pullrequestreview-3859352996

rina23q:
#4128 and follow-up PRs did/will cover the topic.

PR #4120, opened 2026-04-10
tests: use specific version instead of the latest available version

reubenmiller:
## Proposed changes

After the release of 2.0.0, this broke a system test which didn't specify an exact tedge version to upgrade to, leading to the 2.0.0 version being installed instead of the version under test.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/self-update/software_update.toml
+++ b/tests/RobotFramework/tests/cumulocity/self-update/software_update.toml
@@ -10,12 +10,19 @@ on_success = "prepare"
 
 [prepare]
 script = "sudo /etc/tedge/sm-plugins/apt prepare"
+on_success = "get-version"
+
+# Extract the exact version to upgrade to
+# otherwise the server will just get the latest available version which maybe newer than the
+# version under test
+[get-version]
+script = """bash -c "printf ':::begin-tedge:::\\n{\\"version\\":\\"%s\\"}\\n:::end-tedge:::\\n' \\"$(jq -r '.updateList[].modules[].version' <<< '${.payload}')\\"""""
 on_success = "self-update"
 
 [self-update]
 # Note: update tedge as it is used by the tedge-agent (as tedge is a multi-call binary)
 # Note: installing the new tedge package will not restart the tedge-agent service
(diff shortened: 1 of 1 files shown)


tests/RobotFramework/tests/cumulocity/self-update/software_update.toml
```diff
@@ -10,12 +10,19 @@ on_success = "prepare"
 
 [prepare]
 script = "sudo /etc/tedge/sm-plugins/apt prepare"
+on_success = "get-version"
+
+# Extract the exact version to upgrade to
+# otherwise the server will just get the latest available version which maybe newer than the
+# version under test
+[get-version]
+script = """bash -c "printf ':::begin-tedge:::\\n{\\"version\\":\\"%s\\"}\\n:::end-tedge:::\\n' \\"$(jq -r '.updateList[].modules[].version' <<< '${.payload}')\\"""""
```
reubenmiller, line 19: Note: This was very difficult to get all of the escaping correct as there is both shell and toml escaping going on...

If we supported accessing array items via the `${.}` syntax, then I could of avoided this step, and just referenced the version using: `${.payload.updateList[0].modules[0].version}`

albinsuresh: Why not externalise that into script? Yes, it'd be a one-liner script, but would still be simpler than this, right?

didier-wenzek: > If we supported accessing array items via the `${.}` syntax, then I could of avoided this step, and just referenced the version using: `${.payload.updateList[0].modules[0].version}`

This is something we could support.

reubenmiller: > Why not externalise that into script? Yes, it'd be a one-liner script, but would still be simpler than this, right?

Using an external script to do this is overcomplicating it since the more logic feature would be just to support the syntax to remove the need for any jq logic.

Bravo555 (approved):

PR #4122, opened 2026-04-13
feat: accessing array items via the ${.} syntax

didier-wenzek:
## Proposed changes

Add support for accessing array items using brackets `[]` via the `${.}` syntax in workflow definitions.

```toml
[self-update]
script = "sudo /etc/tedge/sm-plugins/apt install tedge --module-version '${.payload.updateList[0].modules[0].version}'"
on_success = "restart-agent"
```

__warning__ this syntax cannot be used to [fix the system test for self-update](https://github.com/thin-edge/thin-edge.io/pull/4120), because this test starts from an old version of thin-edge (aka 1.3.1).

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/pull/4120#discussion_r3065338391

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/substitution.rs
+++ b/crates/core/tedge_api/src/substitution.rs
@@ -117,9 +117,15 @@ fn json_array_excerpt<'a>(value: &'a Value, path: &'a str) -> Option<&'a Value> 
         None => value.get(path),
         Some(0) => {
             let (key, path) = path[1..].split_once(']')?;
-            let index = key.parse::<usize>().ok()?;
-            value
-                .get(index)
+            let index = key.parse::<i64>().ok()?;
+            let items = value.as_array()?;
+            let abs_index = if index >= 0 {
+                index
+            } else {
+                (items.len() as i64) + index
+            };
+            items
+                .get(abs_index as usize)
                 .and_then(|value| json_array_excerpt(value, path))
         }
         Some(pos) => {
@@ -228,6 +234,24 @@ mod tests {
             json!(2)
         );
         assert_eq!(
+            cmd.inject_values_into_template("${.payload.bar.extra[3]}"),
+            ""
+        );
+        assert_eq!(
+            cmd.inject_values_into_template("${.payload.bar.extra[-1]}")
+                .to_json(),
+            json!(3)
+        );
+        assert_eq!(
+            cmd.inject_values_into_template("${.payload.bar.extra[-3]}")
+                .to_json(),
+            json!(1)
+        );
+        assert_eq!(
+            cmd.inject_values_into_template("${.payload.bar.extra[-4]}"),
+            ""
+        );
+        assert_eq!(
             cmd.inject_values_into_template("${.payload.nested[0].matrix[1]}")
                 .to_json(),
             json!([7, 8, 9])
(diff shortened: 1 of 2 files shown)


crates/core/tedge_api/src/substitution.rs
```diff
@@ -97,10 +96,38 @@ fn json_as_string(value: &Value) -> String {
...
 fn json_excerpt<'a>(value: &'a Value, path: &'a str) -> Option<&'a Value> {
+    if path.is_empty() {
+        return Some(value);
+    }
+
     match path.split_once('.') {
-        None if path.is_empty() => Some(value),
+        None => json_array_excerpt(value, path),
+        Some((key, path)) => {
+            json_array_excerpt(value, key).and_then(|value| json_excerpt(value, path))
```
albinsuresh, line 106: The name `json_array_excerpt` is slightly misleading at this level, where we haven't even established if the path has an array index or not. Something generic like `json_path_excerpt` would have been better.

albinsuresh (approved):
Nice improvement.

Unrelated to this PR specifically, but a question on handling of non-existent indexes as well as keys: Currently we return empty string for both cases. Wouldn't it be better to return `null` so that the developers can catch those bugs quickly rather than silently proceeding with an empty string that may or may not be acceptable value?

reubenmiller:
> Nice improvement.
> 
> Unrelated to this PR specifically, but a question on handling of non-existent indexes as well as keys: Currently we return empty string for both cases. Wouldn't it be better to return `null` so that the developers can catch those bugs quickly rather than silently proceeding with an empty string that may or may not be acceptable value?

an empty string is much easier to handle in shell rather than "null", as most of the time the users don't care if it was set or missing, as both need to be handled the same.

So if we had two different values for "the value is not there", then we'd have to handle using:

```sh
if [ -z "$value" ] || [ "$value" = null ]; then
    echo "value is not set"
fi
```

crates/core/tedge_api/src/substitution.rs
didier-wenzek: Yeah, `json_array_excerpt` is not perfect. But `json_path_excerpt` is not either: this function only handles the brackets and assumes there is no dots.

Anyway, this function is private and only used from `json_excerpt`.

crates/core/tedge_api/src/substitution.rs
```diff
@@ -97,10 +96,38 @@ fn json_as_string(value: &Value) -> String {
...
+
+    match path.find('[') {
         None => value.get(path),
-        Some((key, path)) => value.get(key).and_then(|value| json_excerpt(value, path)),
+        Some(0) => {
+            let (key, path) = path[1..].split_once(']')?;
+            let index = key.parse::<usize>().ok()?;
+            value
+                .get(index)
+                .and_then(|value| json_array_excerpt(value, path))
```
reubenmiller: Would the be any chance to support referencing the last item in the array using `-1` (similar to python)...so we could just treat negative numbers as `length - index`

didier-wenzek: Sure. Except the correct formula is `length + index` when the `index` is negative.

didier-wenzek: Done: https://github.com/thin-edge/thin-edge.io/pull/4122/changes/36fbfc06b24d9a66de06c82f075b85114a0126d2

PR #4128, opened 2026-04-14
refactor: Simplify the `tedge_utils` APIs to ensure we handle permissions consistently

jarhodes314:
## Proposed changes
A start on refactoring the `tedge_utils` API for creating and changing ownership of files/directories. This is inspired by the builder-pattern approach originally suggested in https://github.com/thin-edge/thin-edge.io/pull/3983#pullrequestreview-3859352996. The initial implementation doesn't cover all the possible cases where we create files/directories, but addresses most (if not all) of the handling of the contents of `/etc/tedge`.

The commits are conceptually separate steps to make reviewing easier.

A brief summary of the main changes:
- Add a new builder-based API to `paths.rs` focused around the paths we are manipulating, with sensible defaults to handle typical use-cases with ease
- Add `TEdgeConfig::config_root` for easy manipulation of the `/etc/tedge` directory (or the appropriate config root dir) with the system.toml specified user/group. This needs to live in `tedge_config` rather than `tedge_utils` to avoid a cyclic dependency
- Update the main call sites to use this API
- Remove some no longer needed functions from `tedge_utils`

TODO:
- [ ] Update the remaining call sites using functions from `file.rs` in tedge config
- [ ] Delete the now unwanted helper functions and `PermissionsEntry` interface from `file.rs`
- [ ] Update clippy rules to prevent usage of `std::fs` and `tokio::fs` to create files/directories outside of the `tedge_utils` methods (this is probably ought to be a separate PR)

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4115

## Checklist


crates/common/tedge_utils/src/paths.rs
```diff
@@ -37,114 +28,499 @@ pub enum PathsError {
...
+
+    pub async fn ensure(&self) -> Result<(), PathsError> {
+        for dir in self.ancestors_from_root() {
+            let permissions = self.permission_entry().force_dir_ownership();
+            permissions.create_directory(&dir).await?;
+        }
+        Ok(())
+    }
+
+    pub async fn create_if_missing(&self) -> Result<(), PathsError> {
```
albinsuresh, line 252: Do we really need this API that fails if the parent does not exist? The more defensive `ensure` would cover all our needs, no?

crates/common/tedge_utils/src/paths.rs
```diff
@@ -37,114 +28,499 @@ pub enum PathsError {
...
+        let parent = self
+            .path
             .parent()
             .ok_or_else(|| PathsError::ParentDirNotFound {
-                path: target.as_os_str().into(),
+                path: self.path.as_os_str().into(),
             })?;
-        let file = BufWriter::new(TempFile::new_in(dir).await?);
+        tokio::fs::metadata(parent).await?;
```
albinsuresh: Leftover from some debugging?

crates/common/tedge_utils/src/paths.rs
```diff
@@ -37,114 +28,499 @@ pub enum PathsError {
...
+        self
+    }
 
+    pub async fn replace_atomic(&self, content: impl AsRef<[u8]>) -> Result<(), PathsError> {
+        atomically_write_file_async(&self.path, content.as_ref()).await?;
+        self.permission_entry().apply(&self.path).await?;
         Ok(())
     }
+
+    pub async fn create_if_missing(&self, content: impl AsRef<[u8]>) -> Result<(), PathsError> {
```
albinsuresh, line 319: Can we not replace these 2 public APIs with just `ensure` and handle these 2 cases internally? Or do we really need to differentiate?

It'd be ideal if `ManagedFile::ensure()` can also create the parent directories if they are missing, where the parent directories can either inherit the ownership parameters of the parent that is already present in the hierarchy, or use the same `user:group` as specified for this file, and use sensible defaults for mode. For most cases, I feel that'd be sufficient.

crates/common/tedge_config/src/lib.rs
```diff
@@ -36,6 +37,11 @@ impl TEdgeConfig {
         SystemConfig::try_new(self.root_dir()).unwrap_or_default()
     }
 
+    pub fn config_root(&self) -> TedgePaths {
+        let system = self.read_system_config();
+        TedgePaths::from_root_with_defaults(self.root_dir(), system.user, system.group)
+    }
```
didier-wenzek, line 43: These lines are a good hint that this PR is heading in the correct direction.

crates/core/tedge/src/cli/init.rs
```diff
@@ -414,4 +416,26 @@ mod tests {
...
         }
     }
+
+    mod config_root_directories_tests {
+        use super::*;
+
+        #[test]
+        fn contains_expected_managed_directories() {
+            assert_eq!(
+                config_root_directories(),
```
didier-wenzek: I see no value to such unit test repeating the function definition.

crates/core/tedge_mapper/src/aws/mapper.rs
```diff
@@ -159,12 +157,16 @@ async fn bridge_rules(
...
-    let permissions = file::permissions(&user, &group, 0o755);
-
-    if let Err(err) =
-        create_directory_and_update_ownership(mapper_config_dir.clone(), &permissions).await
+    let config_root = tedge_config.config_root();
+    let relative_mapper_config_dir = mapper_config_dir
+        .strip_prefix(tedge_config.root_dir())
+        .context("mapper config directory must stay under the config root")?;
+    if let Err(err) = config_root
+        .dir(relative_mapper_config_dir.as_std_path())
```
didier-wenzek: I would consider to `strip_prefix` from within `config_root.dir()`.

crates/core/tedge/src/cli/connect/command.rs
```diff
@@ -1329,16 +1322,29 @@ async fn write_bridge_config_to_file(
...
+    config_root
+        .dir(TEDGE_BRIDGE_CONF_DIR_PATH)?
+        .with_mode(0o755)
+        .ensure()
+        .await?;
+    config_root
+        .file(format!("{TEDGE_BRIDGE_CONF_DIR_PATH}/{config_file}"))?
+        .with_owner(Owner::user_group(crate::BROKER_USER, crate::BROKER_GROUP))
+        .replace_atomic(contents)
+        .await?;
```
didier-wenzek: Crystal clear: focused on what we must ensure with all the defaults properly read from the config.

didier-wenzek (commented):
I really like this new API. Notably:

- The `config_root: TedgePaths` struct that encapsulates configured defaults for directory location and ownership.
- The declarative way to tell and enforce the requirements for a directory or a file

crates/common/tedge_utils/src/paths.rs
jarhodes314: This is intended for plugins (I think in practice just the remote access plugin) to ensure the relevant directories exist, but don't update the ownership if not the case. Perhaps though this change in behaviour wants capturing in a builder method before calling `ensure` instead of a totally separate terminal method.

crates/core/tedge/src/bridge/mod.rs
```diff
@@ -15,3 +19,22 @@ pub use config::BridgeConfig;
...
+
+pub(crate) async fn write_mosquitto_owned_config(
+    tedge_config: &TEdgeConfig,
+    config_file: &str,
+    contents: &[u8],
+) -> Result<(), PathsError> {
+    let config_root = tedge_config.config_root();
+    config_root
+        .dir(TEDGE_BRIDGE_CONF_DIR_PATH)?
+        .with_mode(0o755)
```
Bravo555: thought: For an optional step in the future, maybe we might want to look into restricting the ability of setting the mode to any value?

Often when I see these calls where we set the particular mode, I'm not sure why this particular value, why `755` and not `775`, if we shouldn't partially respect user's umask instead of setting all the bits, etc.

crates/core/tedge/src/cli/init.rs
jarhodes314: I'm updating this to run the command in a temporary directory and verify it actually creates the directories in question

crates/core/tedge/src/cli/init.rs
```diff
@@ -81,23 +81,11 @@ impl TEdgeInitCmd {
...
-        create_directory_and_update_ownership(config_dir.join("plugins"), &permissions).await?;
-        create_directory_and_update_ownership(config_dir.join("sm-plugins"), &permissions).await?;
-        create_directory_and_update_ownership(config_dir.join("device-certs"), &permissions)
-            .await?;
-        create_directory_and_update_ownership(config_dir.join("mappers"), &permissions).await?;
-        create_directory_and_update_ownership(config_dir.join(".tedge-mapper-c8y"), &permissions)
-            .await?;
+        for dir in config_root_directories() {
+            config_root.dir(dir)?.with_mode(0o775).ensure().await?;
+        }
```
Bravo555, line 88: thought: I find myself having a slight preference the former notation, actually 

It was all in one place, easy to add/remove/(though it'll probably stay like this for a while), tweak selectively if one directory is slightly different (though again, the new form is intentional to show that they're all the same). Not a strong opinion though, both forms are still fine.

crates/common/tedge_utils/src/paths.rs
jarhodes314: Yep, likely the refactoring to add the `mode_bits` helper function

crates/common/tedge_utils/src/paths.rs
```diff
@@ -37,114 +28,499 @@ pub enum PathsError {
...
 
-pub fn create_directories(dir_path: impl AsRef<Path>) -> Result<(), PathsError> {
-    let dir_path = dir_path.as_ref();
-    std::fs::create_dir_all(dir_path)
-        .map_err(|error| PathsError::DirCreationFailed(error, dir_path.into()))
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Owner {
+    pub user: String,
+    pub group: String,
 }
```
Bravo555, line 63: suggestion: Make clear that valid values for the fields are only usernames of valid users.

This seems to be eventually used in `PermissionsEntry::apply` and then in `change_user_and_group` where we assume this is a name and lookup a uid and gid, so this probably won't work with `user = "1000"`?

Without any other comments I'm thinking `String` here means "a name of the user that exists", which is an entirely reasonable assumption, and to be fair the previous API didn't make this very clear (and probably doesn't work when given a uid instead of username), but since this is a new API it'd be good to make it clear and test this case (i.e. that it fails as expected, AFAIK).

Then it's another matter to make it actually work with uids/gids (since this is technically a refactoring PR, so the change it behaviour would be better to include in a follow-up). It think it's desirable property if numeric ids could be used interchangably with names. We don't support it yet (in `system.toml` we also assume it's a name), but I think we should once the paths API handles it correctly.

plugins/c8y_remote_access_plugin/src/lib.rs
```diff
@@ -380,6 +370,76 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn supported_operation_file_uses_explicit_owner_override() {
```
rina23q, line 374: Since `supported_operation_file()` is just calling the `ManagedFile` API, this test seems just testing the API.

plugins/c8y_remote_access_plugin/src/lib.rs
```diff
@@ -380,6 +370,76 @@ mod tests {
...
+            file.path(),
+            dir.path().join("operations/c8y/c8y_RemoteAccessConnect")
+        );
+    }
+
+    #[tokio::test]
+    async fn init_creates_operation_file_when_parent_exists() {
+        let dir = tempfile::tempdir().unwrap();
+        std::fs::write(dir.path().join("tedge.toml"), "").unwrap();
+        std::fs::create_dir_all(dir.path().join("operations/c8y")).unwrap();
```
rina23q, line 392: All tests in this crate would be better if they used `TempTedgeDir` instead of `tempdir`.

crates/extensions/tedge_mqtt_bridge/src/persist.rs
```diff
@@ -47,7 +47,7 @@ pub async fn persist_bridge_config_file(
 
     // Set file ownership and permissions on both files
     for path in [&config_path, &template_path] {
-        let permissions = file::permissions(user, group, 0o644);
+        let permissions = file::permissions(&owner.user, &owner.group, 0o644);
```
rina23q: To be honest, I'm not fan of `file::permissions()`, it covers the problem (addressing `Some("")`) on the top of `PermissionEntry`, which `PermissionEntry` should have addressed. I hope you are planning to update this part in the next commits.

crates/core/tedge/src/cli/init.rs
```diff
@@ -115,7 +103,7 @@ impl TEdgeInitCmd {
             config.agent.state.path.to_path_buf()
         } else {
             let agent_state_dir = config_dir.join(".agent");
-            create_directory_and_update_ownership(&agent_state_dir, &permissions).await?;
+            config_root.dir(".agent")?.with_mode(0o775).ensure().await?;
```
rina23q: Also user and group have to be "ensured" I guess.

Bravo555 (commented):
In general I agree that this new API is a good direction. It makes it easy to ensure we have certain files and directories with correct owners and modes but another thing is all the clients using it correctly, not setting more than they have to, figuring why they use certain mode bits, etc., though I'm not sure if others feel similar.

crates/core/tedge/src/cli/init.rs
didier-wenzek: That's one of the nice additions of this PR: `config_root` encapsulates default user and group as defined in `system.toml` https://github.com/thin-edge/thin-edge.io/pull/4128#discussion_r3084570612

crates/common/tedge_utils/src/paths.rs
didier-wenzek: > It think it's desirable property if numeric ids could be used interchangably with names.

Using numeric ids is really fragile on a fleet of devices: the configuration can no more be shared.

crates/core/tedge/src/cli/init.rs
jarhodes314: Yeah, I think I see it as the most likely thing for someone to want to do when reading the code is establish which directories are created by this code, and I think listing them makes that much simpler

crates/core/tedge/src/cli/init.rs
jarhodes314: Arguably it might be worth not requiring the user to specify modes as that mostly just depends on whether the subject is a directory or a file so we set executable bits as appropriate. Perhaps even abstracting this out away from modes entirely (e.g. methods like `.group_writable()`, `.executable()`, `.sensitive()` as a more intent-focused API rather than the slightly cryptic nature of mode bits)

crates/core/tedge/src/bridge/mod.rs
jarhodes314: Yeah, I think this ties into the thoughts I had here about possibly just abstracting away modes entirely: https://github.com/thin-edge/thin-edge.io/pull/4128/changes/BASE..610a12a9bebb4d8ee0e9c70661883e75f1ec3938#r3092737419

crates/extensions/tedge_mqtt_bridge/src/persist.rs
jarhodes314: I've now got rid of this function in favour of using the builder API for these files to handle the ownership. Once everything is using the builder API, we then shouldn't need `PermissionEntry` whatsoever, and everything can follow the pattern of the target file/dir is the subject that we apply to.

crates/core/tedge_mapper/src/aws/mapper.rs
jarhodes314: Now done

crates/common/tedge_utils/src/paths.rs
jarhodes314: I've now changed the directory handling to have a single `ensure` method in d021bbab

crates/common/tedge_utils/src/paths.rs
jarhodes314: I think for files there's much more of a case for different terminal methods. With directories, there's broadly just the question of  "do we update the ownership/permissions if the directory already exists". With files on the other hand, we have a bunch of things e.g. replace the file every time, create it if it doesn't exist and update the metadata on a user provided file, but don't write to it (not an exhaustive list). Given the wide variability in what we do, I think it's important the API draws attention to that.

As for automatically creating parent directories, it's not totally infeasible to do this, but I think it needs a bit of consideration as to how it ought to work (e.g. to what extent do we want to traverse parents of parents to reassert ownership). I think an important part of this refactoring is understanding what variations we have in how we handle things and whether we actually need the complexity resulting from that variability. I think keeping the directory handling manual and separate helps somewhat with that goal.

I also think there's something to be said for not making APIs that automatically do everything possible. In this case, I think my concern is around the fact we should be able to trust that `tedge init` has run so components shouldn't really _need_ to ensure directories exist, and the fact they do is a bit surprising (and is a sort of thing that can make debugging stuff really difficult).

crates/core/tedge_mapper/src/c8y/mapper.rs
```diff
@@ -445,8 +439,29 @@ mod tests {
...
+            .unwrap();
+        let group = std::process::Command::new("id")
+            .arg("-gn")
+            .output()
+            .ok()
+            .and_then(|o| String::from_utf8(o.stdout).ok())
+            .map(|s| s.trim().to_owned())
+            .unwrap();
+        (user, group)
+    }
```
didier-wenzek: I would consider to let the 3 mapper test module share this function.

crates/common/tedge_utils/src/paths.rs
```diff
@@ -37,23 +42,474 @@ pub enum PathsError {
...
+    pub fn use_process_ownership(self) -> Self {
+        self.with_owner(Owner::user_group("", ""))
+    }
+
+    /// Leave an existing directory unchanged.
+    ///
+    /// If the directory is missing, it is created and configured with this
+    /// builder's owner and mode. If it already exists, its ownership and mode
+    /// are left as-is.
+    pub fn respect_existing(mut self) -> Self {
```
didier-wenzek, line 208: Definitely better than the former `force_dir_ownership` which was not clear enough on *when* the ownership was forced.

crates/common/tedge_utils/src/paths.rs
```diff
@@ -37,23 +42,474 @@ pub enum PathsError {
...
+        self.respect_existing = true;
+        self
+    }
+
+    pub fn group_writable(mut self) -> Self {
```
didier-wenzek, line 213: Having ownership kind instead of adhoc numeric mod is a good idea. I'm just surprised to see only this `group_writable` case.

crates/common/tedge_utils/src/paths.rs
```diff
@@ -37,23 +42,474 @@ pub enum PathsError {
...
+        Ok(ManagedFile {
+            root: self.root.clone(),
+            path: self.resolve(path)?,
+            owner: self.default_owner.clone(),
+            mode: DEFAULT_FILE_MODE,
+            warn_and_ignore_permission_errors: false,
+        })
+    }
+
+    pub fn template_file(&self, path: impl AsRef<Path>) -> Result<ManagedTemplateFile, PathsError> {
```
didier-wenzek: It's a great idea to merge the logic for template files with the core logic to create tedge configuration files and directories. The implementation is less transparent though. => Some doc comment  would be helpful.

crates/common/tedge_utils/src/fs.rs
```diff
@@ -178,50 +178,10 @@ fn parent_dir(file: &Path) -> PathBuf {
...
 
-/// Persists a file using the template pattern:
-///
-/// - `{name}` - active file (updated only if not overridden or disabled)
-/// - `{name}.template` - canonical template (always updated)
-/// - `{name}.disabled` - marker file indicating the file is disabled
-///
-/// If a user has modified the `name` file (making it differ from the `name.template`),
-/// or if a `.disabled` marker exists, the active file will not be updated.
-/// However, the template will always be refreshed with the latest definition.
```
didier-wenzek, line 189: This doc comment deserves to be kept somewhere. `TedgePaths::template_file()` seems to be a good candidate.

didier-wenzek (commented):
A very nice clarification of the API to ensure consistent file ownership across thin-edge components.

crates/common/tedge_utils/src/paths.rs
jarhodes314: Yeah, we don't actually do much wrt different permissions

crates/common/tedge_utils/src/fs.rs
jarhodes314: Done in 55d692ca. Slightly reworded it to communicate the correct intent (we aren't persisting it there, just creating a builder for it)

crates/common/tedge_utils/src/paths.rs
jarhodes314: Done in 55d692ca

crates/core/tedge_mapper/src/c8y/mapper.rs
jarhodes314: Done 55d692ca

didier-wenzek (approved):
Approved. A very nice refactoring making clear thin-edge usage and conventions on configuration files and directories.

rina23q (approved):

crates/common/tedge_utils/src/paths.rs
albinsuresh: > With files on the other hand, we have a bunch of things e.g. replace the file every time, create it if it doesn't exist and update the metadata on a user provided file, but don't write to it (not an exhaustive list). Given the wide variability in what we do, I think it's important the API draws attention to that.

Yeah I was actually hoping that an approach similar to `ManagedDir:respect_existing` can be applied to `ManagedFile` as well. But I agree that it may not be as elegant as I hoped it would be, and doesn't even have an equivalent for some combinations:

| Contract | Current | Proposed |
| --- | --- | --- |
| Ensure provided content and permissions | file.replace_atomic(content).await | file.ensure(content).await |
| Create new file with permissions or leave existing one as-is | file.create_if_missing(content).await | file.respect_existing_content().ensure(content).await |
| Create new file with permissions or update permissions of existing one | file.ensure_permissions().await | file.respect_existing_permissions().ensure(content).await |
| Update permissions of existing | file.ensure_permissions().await | ??? |

So, I agree that what we've got is fine.

> As for automatically creating parent directories, it's not totally infeasible to do this, but I think it needs a bit of consideration as to how it ought to work (e.g. to what extent do we want to traverse parents of parents to reassert ownership).

Sure, we can explore that separately. Since it's a new API anyway, it doesn't have to be part of this refactoring. Now that we have the builder-esque API, adding the `parent_owner` to it would be just trivial work.

> I also think there's something to be said for not making APIs that automatically do everything possible.

Agree. Expecting `tedge init` to do the fair minimum like creating the immediate directories under `/etc` or `/var` is a fair expectation. The parent directory creation will come in handy mostly in the context of config management and nested flows.

crates/common/tedge_utils/src/file.rs
```diff
@@ -73,66 +73,44 @@ pub enum FileError {
     },
 }
 
-pub async fn create_directory(
+pub async fn create_directory_and_update_ownership(
```
albinsuresh, line 76: Not used anymore. I'm assuming you're planning to delete this along with `create_directory_with_defaults` in a follow-up PR.

crates/common/tedge_utils/src/file.rs
```diff
@@ -73,66 +73,44 @@ pub enum FileError {
...
-    permissions.create_directory(dir.as_ref()).await
+    permissions
+        .clone()
+        .force_dir_ownership()
+        .create_directory(dir.as_ref())
+        .await
 }
 
-pub async fn create_directory_and_update_ownership(
+pub async fn create_directory_and_update_ownership_with_root(
```
albinsuresh, line 87: Same as above. Not used anymore.

albinsuresh (approved):

PR #4130, opened 2026-04-15
Add `just debug` command for a faster local system test cycle

jarhodes314:
## Proposed changes
Adds a new command to the justfile. This is like `just release` except it builds in debug mode with debug symbols stripped. This is intended to speed up running the system tests locally as the release builds always take a long time. The symbol stripping is required to keep the packaging step fast, as that slows down considerably if provided with huge binaries.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/justfile
+++ b/justfile
@@ -10,6 +10,10 @@ DEFAULT_TARGET := `./ci/build_scripts/detect_target.sh`
 
 CARGO := `command -V cargo-zigbuild >/dev/null 2>&1 && printf "cargo-zigbuild" || printf "cargo"`
 
+# Build profile used when building the Rust binaries. Defaults to 'release'
+# Available values: [release, dev-stripped, debug]
+BUILD_PROFILE := env("BUILD_PROFILE", "release")
+
 # Print project and host machine info
 info:
     @echo "OS:             {{os()}}"
@@ -126,9 +130,14 @@ check-dependencies:
     cargo deny fetch
     cargo deny --all-features check --allow duplicate --allow unmaintained
 
-# Release, building all binaries and debian packages
+# Build all binaries and packages using the set profile
+# or dev, or something in-between
+build *ARGS:
+    ci/build_scripts/build.sh --profile {{BUILD_PROFILE}} {{ARGS}}
+
+# Release, building all binaries and packages (optimized, stripped of debug symbols)
 release *ARGS:
-    ci/build_scripts/build.sh {{ARGS}}
+    just -f {{justfile()}} BUILD_PROFILE=release build {{ARGS}}
 
 # Run unit and doc tests
 test *ARGS:
@@ -147,8 +156,8 @@ test-docs *ARGS:
 setup-integration-test *ARGS:
     tests/RobotFramework/bin/setup.sh {{ARGS}}
 
-# Run integration tests (using local build)
-integration-test *ARGS: release
+# Build the integration test image
+build-integration-test: build
     #!/usr/bin/env bash
     set -e
     if [ ! -d tests/RobotFramework/.venv ]; then
@@ -156,7 +165,10 @@ integration-test *ARGS: release
     fi
     cd tests/RobotFramework
     source .venv/bin/activate
-    invoke build --local
+    invoke build
+
+# Run integration tests (using local build)
+integration-test *ARGS: build-integration-test
     invoke tests {{ARGS}}
 
 # Generate linux package scripts from templates
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -240,6 +240,10 @@ yansi = { version = "1.0.1", default-features = false, features = [
 ] }
 zeroize = "1.5"
 
+[profile.dev-stripped]
+inherits = "dev"
+strip = "symbols"
+
 [profile.release]
 codegen-units = 1
 lto = true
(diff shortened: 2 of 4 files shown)


justfile
```diff
@@ -130,6 +130,10 @@ check-dependencies:
 release *ARGS:
     ci/build_scripts/build.sh {{ARGS}}
 
+# Debug, building all binaries and debian packages (unoptimized, stripped of debug symbols)
+debug *ARGS:
+    ci/build_scripts/build.sh --profile dev-stripped {{ARGS}}
```
didier-wenzek: As this build is uniquely intended to run system tests, I would add the following:

```suggestion
    ci/build_scripts/build.sh --profile dev-stripped {{ARGS}}
    cd tests/RobotFramework
    source .venv/bin/activate
    invoke build --local
```

didier-wenzek (approved):
Approved

justfile
albinsuresh:
```suggestion
integration-test *ARGS: debug
```
Isn't that also required? Since the goal was to use the same `debug` artefacts from the system tests?

justfile
```diff
@@ -156,7 +160,7 @@ integration-test *ARGS: release
     fi
     cd tests/RobotFramework
     source .venv/bin/activate
-    invoke build --local
+    invoke build
```
albinsuresh, line 168: Was that intentional?

reubenmiller: `--local` is anyway the default since a while now (about a year?)

justfile
albinsuresh: Or have a new recipe defined with the `debug` recipe dependency, as @didier-wenzek has suggested in https://github.com/thin-edge/thin-edge.io/pull/4130/changes#r3087013132

reubenmiller: I'd prefer an independent task, or just parameterize the task so that we still run the official system tests using the proper release artifacts.

reubenmiller: However, I suspect that "integration-test" is not used that often locally (or at least I never run it), as usually the dev is focused on specific test cases, and will run the required ones using the IDE "run test" function

reubenmiller: @jarhodes314 @albinsuresh  I can make a proposal here if you don't object

didier-wenzek: > I suspect that "integration-test" is not used that often locally (or at least I never run it),

Same for me

> as usually the dev is focused on specific test cases,

precisely my use case

justfile
jarhodes314: Yeah, this was intentionally done so if people consult the justfile for how we should be building these things, they get the example of the simplest possible command.

justfile
reubenmiller: I have this working locally as described below...Once agreed I can push the fixup commit.

Task changes:

* Removed `debug` and replaced it with a generic `build` task where the profile is controlled by the `BUILD_PROFILE` variable (which defaults to the existing behaviour `release`)
* Added `build-integration-test` - build the integration test image but don't run the tests. This is 
* All tasks use the generic `build` task when building the artifacts as it allows the BUILD_PROFILE to propagate through-out the tasks

The new `BUILD_PROFILE` variable can either be set using `just BUILD_PROFILE ...` or setting an environment variable (if you want to set it in your .env file).

Below shows the different usages:

Setting the value from the environment variable:

```sh
export BUILD_PROFILE=dev-stripped

# build-integration-test
just build-integration-test

# or just build the binaries
just build
```

Or setting the BUILD_PROFILE when calling just

```sh
# build rust binaries and integration test image
just BUILD_PROFILE=dev-stripped build-integration-test

# build rust binaries, integration test image, and run integration tests
just BUILD_PROFILE=dev-stripped integration-test

just BUILD_PROFILE=dev-stripped build

# or if you want to force building the release binaries (as was the previous behavior)
just release
```

jarhodes314: @reubenmiller that looks good to me. Can we default to `dev-stripped` and require the `BUILD_PROFILE` (or `just release`) only for release? Scripts can easily pass the extra argument, and it'll save us the hassle of typing it out when running it by hand.

albinsuresh: > However, I suspect that "integration-test" is not used that often locally (or at least I never run it), as usually the dev is focused on specific test cases, and will run the required ones using the IDE "run test" function

I used to use that recipe just to get the test image built (by passing a dummy test name). So your proposal to add a dedicated `build-integration-test` recipe perfectly meets my requirement.

reubenmiller: > dev-stripped

We could, though that would only offer value if you're just wanting to build it and the packages, and not then use it as part of the integration tests...otherwise `just BUILD_PROFILE=dev-stripped integration-test` does both (build binaries and system test).

And if it is about not wanting to remember each of these commands, then I'd recommend just setting the `BUILD_PROFILE` in your .env file, and be done with it ;)

reubenmiller: pushed, we can discuss on the code, 5e97265d755ff9ce556032cb749e92a766ec985e

reubenmiller: @jarhodes314 There is some new just features where you can also add flags style arguments, where you would at least get some help information if you provide the wrong value...however no tab completion yet...

```
$ just build --profile asdf       
error: Argument `asdf` passed to recipe `build` parameter `profile` does not match pattern 'release|debug|dev-profile'
```

But since this is fairly new just syntax, I'd stick with the current proposal for now before switching to allow `--profile <profile>`.

albinsuresh (approved):
Re-confirming my approval

Issue #1954, opened 2023-05-05
Make collectd mapper batching externally testable


albinsuresh:
**Is your feature improvement request related to a problem? Please describe.**

The batching of collectd mapper is controlled by two batching parameters: `batching_window` (`event_jitter` in code) and `max message delay` (`delivery_jitter`) with default values of `500ms` and `400ms` respectively. But, these parameters can not be changed from outside and that makes it extremely difficult to deterministically test the batching behaviour from outside as it is almost impossible to send multiple collectd messages from outside and make sure that they are delivered within that `500ms` delay. The fixed `max_message_delay` of `400ms` makes it worse as it can lead to any one of the input messages to be dropped if they arrive just `400ms` late.

**Describe the solution you'd like**

The most obvious solution is to make both `batching window` and `max message delay` configurable via `tedge config` so that we can update these values to testable limits. 

The problem with this solution is the overlap that it has with the configurability provided by collectd for metrics collection interval, with which the user can control how frequently these measurements are generated and hence influence the batching behaviour. Providing additional configurations to control the batching behaviour possesses the risk of exposing too many parameters to the end user leading them to get it wrong (We've had this experience in the past, while trying to tune these parameters for our RaspberryPi slaves).

**Describe alternatives you've considered**

Drop the concept of `max message delay` completely and batch every input that arrives within the `batching window` after the receipt of the first message.  With this, we can limit the configurability to just the `batching window`.

The main risk with dropping this parameter is the possibility of newer batches being polluted with older measurements in some extremely slow environments.

didier-wenzek:
> Drop the concept of max message delay completely and batch every input that arrives within the batching window after the receipt of the first message. With this, we can limit the configurability to just the batching window.

To be more precise, one has to stop throwing away messages that are *processed too late*. I stressed the words *processed too late* because I see that as the main issue : messages are discarded only because the mapper, is late for some reason making things really to understand and test.

- Currently a message is dropped if `wall-clock-processing-time > message-source-timestamp + max-delay`

> The main risk with dropping this parameter is the possibility of newer batches being polluted with older measurements in some extremely slow environments.

There is no such risk if the `batching time window` is relative to the message source timestamps and *not* the processing time. A batch will only contain messages for a given time window. However, one can have a 2 batches for the same time window, the second batch containing the laggers.

[...]

PR #4136, opened 2026-04-16
refactor: replace the collectd mapper with local flows

didier-wenzek:
## Proposed changes

The goal of this PR is to deprecate the collectd mapper providing the tools to implement the same using flows

- [x] Flow script translating collectd messages into te measurements
- [x] Flow transformer batching messages
- [ ] ~Merge Collectd Flow example: https://github.com/thin-edge/tedge-flows-examples/pull/121~
- [x] Merge Collectd setup: https://github.com/thin-edge/tedge-collectd-setup/pull/6#pullrequestreview-4382562089
- [x] Rewrite system tests to use a flow-based collectd mapper
- [x] Update sections of the docs that use collectd to illustrate device monitoring 
- [x] Deprecate the batcher actor
- [x] Deprecate the collectd mapper

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


tests/RobotFramework/tests/cumulocity/monitoring/monitor_device_collectd.robot
```diff
@@ -27,19 +27,10 @@ Test Tags           theme:monitoring    theme:c8y    theme:collectd
 
 
 *** Test Cases ***
-Check running collectd
-    Service Should Be Running    collectd
-
-Is collectd publishing MQTT messages?
-    ${messages}=    Should Have MQTT Messages    topic=collectd/#    minimum=1    maximum=None
-
```
didier-wenzek: These checks have been moved to the suite setup. They are not system tests but prerequisite checks.

tests/RobotFramework/tests/cumulocity/monitoring/collectd/group-te-measurements.js
```diff
@@ -0,0 +1,18 @@
+// Translate an array of thin-edge measurements into a single message with grouped measurements
```
didier-wenzek: I consider promoting this script to a builtin transformer: grouping measurements is a quite common use case.

crates/common/batcher/src/batcher.rs
```diff
@@ -17,6 +18,17 @@ pub struct Batcher<B: Batchable> {
     batches: Vec<Batch<B>>,
 }
 
+impl<B: Batchable> Default for Batcher<B> {
+    fn default() -> Self {
+        let batch_config = BatchConfigBuilder::new()
+            .event_jitter(500)
+            .delivery_jitter(400) // Heuristic delay that should work out well on a Rpi
+            .message_leap_limit(500)
+            .build();
```
didier-wenzek: These configuration parameters are simply hairy to get correct, the batching algorithm being complicated and fragile with two independent clocks (batching time and event time). This complexity is furthermore useless in practice, the best way to avoid too many collectd measurements being to reduce the collection frequency.

crates/common/batcher/src/flows.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    fn key(&self) -> Self::Key {
+        self.topic.clone()
+    }
+
+    fn event_time(&self) -> OffsetDateTime {
+        self.timestamp
+            .map(|t| t.into())
+            .unwrap_or_else(OffsetDateTime::now_utc)
+    }
```
didier-wenzek: This defeats the purpose of the batcher: i.e using the processing time to batch message, ignoring the event time.

I will fix that with a flow config giving a JSON path to the message event time.

crates/common/batcher/src/flows.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct MessageBatcher {
+    batcher: Batcher<Message>,
+    batch_topic: String,
+}
+
+impl Batchable for Message {
+    type Key = String;
+
+    fn key(&self) -> Self::Key {
+        self.topic.clone()
```
didier-wenzek: Using the topic as a key make the batcher transformer a bit complicated to use: the previous step needs to be aware that the topic is used as a key by the batcher.

I will fix that with a way to configure the transformer with a path to the key.

docs/src/operate/monitoring/systemd-watchdog.md
```diff
@@ -18,7 +18,7 @@ This document describes how the systemd watchdog mechanism can be enabled for %%
 
 ## Enabling the systemd watchdog feature for a tedge service
 
-Enabling systemd watchdog for a %%te%% service (tedge-agent, tedge-mapper-c8y/az/collectd) is a two-step process.
+Enabling systemd watchdog for a %%te%% service (tedge-agent, tedge-mapper-c8y/az) is a two-step process.
```
rina23q, line 21: (unrelated) Out of curiosity, The watchdog feature doesn't work also for aws and local mappers?
I found this document is actually unlisted, so maybe it doesn't matter how precise information we have here.

docs/src/operate/troubleshooting/log-files.md
```diff
@@ -50,20 +50,6 @@ journalctl -u tedge-mapper-aws
 Run `tedge_mapper --debug aws` to log more debug messages
```
rina23q: (unrelated) underscore is legacy, should be replaced with `-`.

```suggestion
Run `tedge-mapper --debug aws` to log more debug messages
```

docs/src/operate/configuration/mosquitto-configuration.md
```diff
@@ -80,36 +80,14 @@ tedge connect aws
 This will configure all the services (mosquitto, tedge-mapper-c8y.service, tedge-mapper-az.service,
 tedge-mapper-aws.service, tedge-agent.service) to use the newly set port and the bind address.
 
-## Common Errors
+## Step 4: Restart the local services using MQTT
```
rina23q: Need to add one more level to align with the other steps.

```suggestion
### Step 4: Restart the local services using MQTT
```

docs/src/operate/troubleshooting/log-files.md
```diff
@@ -50,20 +50,6 @@ journalctl -u tedge-mapper-aws
...
 :::
 
-### Device monitoring logs {#device-logs}
```
rina23q, line 53: Probably we can also remove the section "Collectd logs" below (inside "Third-party component logs"). Instead, we can mention how to see the log of collectd in the new repository for the collectd flow.

tests/RobotFramework/tests/cumulocity/monitoring/monitor_device_collectd.robot
```diff
@@ -75,8 +68,11 @@ Check grouping of measurements
...
-    # 1. Install collectd
-    # 2. Configure collectd
+    Start Service    tedge-mapper-local
+    Install Flow    collectd
+    Install Collectd
+
+Install Collectd
     Execute Command
     ...    sudo apt-get update && sudo apt-get install -y --no-install-recommends collectd-core libmosquitto1
     Execute Command    sudo cp /etc/tedge/contrib/collectd/collectd.conf /etc/collectd/collectd.conf
```
rina23q, line 76: I guess this `collectd.conf` file is also a candidate to move out of the thin-edge.io repository later.

rina23q (commented):
I guess still code is in progress, so I reviewed the documentation changes.

docs/src/operate/troubleshooting/log-files.md
didier-wenzek: unrelated, but fix as this is the only occurrence of this typo

docs/src/operate/configuration/mosquitto-configuration.md
didier-wenzek: fixed

tests/RobotFramework/tests/cumulocity/monitoring/monitor_device_collectd.robot
```diff
@@ -48,20 +39,22 @@ Check thin-edge monitoring
...
     # This test step is only partially checking the grouping of the messages, because of the timeouts and the current design
     # if proper checks will be implemented this test would be failing from time to time
-
     Execute Command    sudo systemctl stop collectd
     Sleep    5s    reason=Needed because of the batching
+
     ${start_time}=    Get Unix Timestamp
-    Execute Command
-    ...    tedge mqtt pub collectd/localhost/temperature/temp1 "`date +%s.%N`:50" && tedge mqtt pub collectd/localhost/temperature/temp2 "`date +%s.%N`:40" && tedge mqtt pub collectd/localhost/pressure/pres1 "`date +%s.%N`:10" && tedge mqtt pub collectd/localhost/pressure/pres2 "`date +%s.%N`:20"
+    Execute Command    tedge mqtt pub collectd/localhost/temperature/temp1 "`date +%s.%N`:50"
```
albinsuresh: IIRC, the chaining of the `pub` commands in the former version was done to minimise the timestamp drift between the first measurement and the last measurement, to make sure that they get batched together. So, this version might end up being flaky.

This was the point that we always wanted to improve, at least by [making the batch window configurable](https://github.com/thin-edge/thin-edge.io/issues/1954), so that we don't have to rely on fragile tricks like that.

tests/RobotFramework/tests/cumulocity/monitoring/collectd/group-te-measurements.js
```diff
@@ -0,0 +1,18 @@
...
+
+const utf8 = new TextDecoder();
+
+export function onMessage(message, context) {
+    const { topic = "te/device/main///m/" } = context.config;
+    let measurements = JSON.parse(utf8.decode(message.payload))
+    let grouped_measurements = {}
+
+    for (let measurement of measurements) {
+        Object.assign(grouped_measurements, measurement.payload)
```
albinsuresh: That would get the timestamp wrong, no? The timestamp of the last measurement would replace the former ones, while the expectation is for the batch to keep the timestamp of the first measurement that triggered the batch.

tests/RobotFramework/tests/cumulocity/monitoring/collectd/group-te-measurements.js
albinsuresh: That would be a good addition. The input topic would also have to be in the tedge measurements format, right?

tests/RobotFramework/tests/cumulocity/monitoring/collectd/group-te-measurements.js
didier-wenzek: I would not say this is wrong. Not your expectation apparently, but as meaningful.

Anyway, this script is only used for testing purposes and will be replaced by a builtin transformer.

On the joke side. According this [comment](https://github.com/thin-edge/thin-edge.io/pull/4136#discussion_r3132007173), messages must happen in a blink to be batched! So first or last timestamp should not be a big deal ;-)

tests/RobotFramework/tests/cumulocity/monitoring/monitor_device_collectd.robot
didier-wenzek: > So, this version might end up being flaky.

This is not what I observe:

```
$ invoke flake-finder --test-name  "Check grouping of measurements" --iterations 25 --outputdir output_ff --clean

Overall: PASSED
Results: 25 iterations, 25 passed, 0 failed
Elapsed time: 0:10:34.572869
```

> This was the point that we always wanted to improve, at least by [making the batch window configurable](https://github.com/thin-edge/thin-edge.io/issues/1954),

I will do that in this PR.

crates/common/batcher/src/flows.rs
didier-wenzek: Finally, this code will be let unchanged, i.e. the timestamp of the message being used as the event time.

For that purpose the flow engine has been improved. Now, [message timestamps can be set by a script](https://github.com/thin-edge/thin-edge.io/pull/4136/commits/89338e2dc8941b774a8da3997d74a562aab95f6c) and used by the subsequent steps.

tests/RobotFramework/tests/cumulocity/monitoring/collectd/group-te-measurements.js
didier-wenzek: > The input topic would also have to be in the tedge measurements format, right?

I still have to see what's the best approach to decompose the flow into independent steps that can be re-used / re-combined.

crates/common/batcher/src/flows.rs
```diff
@@ -0,0 +1,210 @@
...
+        message: &Message,
+        _context: &FlowContextHandle,
+    ) -> Result<Vec<Message>, FlowError> {
+        let batches = self
+            .batcher
+            .event(timestamp.into(), message.clone())
+            .into_iter()
+            .filter_map(|action| match action {
+                BatcherOutput::Batch(batch) => Some(batch),
+                BatcherOutput::Timer(_) => None,
```
albinsuresh: As far as I understood, ignoring that `Timer` timestamp is the cause of the major change in behaviour. It would have been better if this timer could be used to control the interval of the flow, rather than it being just periodic.

docs/src/references/mappers/mqtt-topics.md
```diff
@@ -82,19 +82,3 @@ The AWS topics are prefixed by `aws/`.
 * `aws/shadow/#` Use this topic to interact with unnamed and named shadows of the device. It's mapped to
   `$aws/things/{device_id}/shadow`.
 
-## Collectd topics
-
-When the [device monitoring feature is enabled](../../start/device-monitoring.md),
```
albinsuresh, line 87: Forgot to remove that getting started `device-monitoring.md`? Or the plan is to keep it with the documentation of the new location of the collectd flow mapper?

albinsuresh (approved):
From what I understood, there is a behavioural difference between the old collectd batcher and this one. In the old one, the first incoming message of a batch used to trigger batching and the batching would continue until the batch interval elapses. The primary reason behind this approach was to avoid artificially splitting a natural incoming batch of message across multiple outgoing batches.

For example assuming the following batching scenario:
Source message batches are triggered at T0, T10, T20, T30.. and so on (the numbers mean seconds) and each batch of messages lasts for upto 5 seconds. So, ideally we should have had batches every 10 seconds with messages from `T0-T5`, `T10-T15`, `T20-T25` etc in their own respective batches. This "message triggered batching" was the key requirement given when the original collectd mapper was written.

With the new approach where the flow engine's timer dictates when the batching starts and when it ends, I fear that it might lead to these batches getting split due to the drift between the even source timeline and the batcher's timeline.

For example, if the flow engine's timer first triggers at T3 instead of T0, then the intervals change to T3, T8, T13, T18, T23, ... and so on. So effectively the first batch might have the last few messages from the `T0-T5` batch and the next batch would have the remaining and so on. The same split would happen to the rest of the batches as well.

Most likely, the end-user wouldn't notice the difference or not even care. I'm just writing this to be sure that we accept this change in behaviour, unless I completely got it all wrong. Happy to approve after a confirmation on this.

docs/src/references/mappers/mqtt-topics.md
didier-wenzek: Yeah, this is not clear yet.

didier-wenzek:
> From what I understood, there is a behavioural difference between the old collectd batcher and this one.

I don't think so - even if I agree that this is not obvious and that I might even be wrong.

> In the old one, the first incoming message of a batch used to trigger batching and the batching would continue until the batch interval elapses. The primary reason behind this approach was to avoid artificially splitting a natural incoming batch of message across multiple outgoing batches.

Nothing has been changed on that side. The batcher is feed when a message is received and the logic to close batches is fully managed by the batcher.

What changed is *when* the batcher is pooled for finalized batches. Previously, this was done using a set of timers to to release batches asap - i.e. as soon as time moved forward enough to tell for sure that no more messages will be added to some batches. Now, this is done at a regular pace. If pooled too soon, the batcher will release no batches. 

The only difference that might be observed is that a batches is produced a bit later with the new mapper compared to the previous one.

> With the new approach where the flow engine's timer dictates when the batching starts and when it ends, I fear that it might lead to these batches getting split due to the drift between the even source timeline and the batcher's timeline.

No, the batching logic as not been changed at all.

> For example, if the flow engine's timer first triggers at T3 instead of T0, then the intervals change to T3, T8, T13, T18, T23, ... and so on. So effectively the first batch might have the last few messages from the `T0-T5` batch and the next batch would have the remaining and so on. The same split would happen to the rest of the batches as well.

The batches are the same when pooled a bit too late. However, this is not be the case when a message is received a bit too late: it might be discarded. This is not a new issue. Since its inception, the collectd batcher is brittle when faced to message delivery delays. Mainly because of wrong assumptions on when no more messages will be received for a batch. Notably, if the collectd mapper stops for a while all the messages received meantime are discarded. They should not.

> Most likely, the end-user wouldn't notice the difference or not even care. I'm just writing this to be sure that we accept this change in behaviour, unless I completely got it all wrong. Happy to approve after a confirmation on this.

This is the main problem! The collectd batching logic is complicated and brittle, but nobody care.

albinsuresh:
> What changed is when the batcher is pooled for finalized batches. Previously, this was done using a set of timers to to release batches asap - i.e. as soon as time moved forward enough to tell for sure that no more messages will be added to some batches. Now, this is done at a regular pace. If pooled too soon, the batcher will release no batches.

Okay, I think I got confused by that `BatcherOutput::Timer(_) => None` statement. On taking a deeper look at the code after your explanation, it is clear that that even a premature or random call to the `Batcher::time()` (as called from the `on_interval`) would still leave the open batch open, without ever splitting it. The `Batcher` logic ensures that the messages that belong to the same batch (emitted within the 500ms window) would always be batched together. While some of them could still be dropped if delivered late, they'll never be split into multiple batches. So, the flow's interval only dictates how frequently the batches would be published and does not dictate when a batch gets closed.

didier-wenzek:
> So, the flow's interval only dictates how frequently the batches would be published and does not dictate when a batch gets closed.

Very well stated. This clarifies my long explanation.

tests/RobotFramework/tests/cumulocity/monitoring/collectd/flow.toml
```diff
@@ -0,0 +1,7 @@
+input.mqtt.topics = ["collectd/+/+/+"]
+
+steps = [
+    { script = "collectd-to-te-measurements.js" },
+    { builtin = "time-window-batcher", interval = "1s" },
+    { script = "group-te-measurements.js", config = { topic = "te/device/main///m/collectd"} },
+]
```
didier-wenzek: I decided to fully deprecate the `batcher`  crate and this test highlights well why.

The `time-window-batcher` flow transformation, exposed by the `batcher`  crate, could be used unchanged (this test is working as expected), but using it is artificially complicated. You have to pre and post process the data in and out to adapt the messages to the batcher constraints.

- The batcher has been specifically designed to work on top of `collectd`. A measurement is expected to be a single value which type is fully identified by the topic. By contrast, on a thin-edge measurement topic, measurements of different types can be published and grouped.
- The pre-processing step  (here `collectd-to-te-measurements.js`) cannot be a [generic collectd to te measurement script](https://github.com/thin-edge/tedge-flows-examples/pull/121). The format of the message payload can be a thin-edge measurement, but grouped measurements must be splitted into separate measurements and the topic must reflect the measurement type.
- The  post-processing step (here `group-te-measurements.js`) has then to do the reverse, i.e to make a grouped measurement from an array of atomic measurement.

I see as a must the ability to use a [generic collectd to te measurement script](https://github.com/thin-edge/tedge-flows-examples/pull/121) giving users the ability to group or not the measurements produced by collectd.

Also, if thin-edge provides a `time-window-batcher`, it makes more sense to a flow transformation which operates on thin-edge measurements and not only collectd measurements.

=> I will work on a new `time-window-batcher`
- consuming thin-edge measurements
- grouping into a single measurement all the measurements received during a configurable time-window 
- avoiding to group conflicting measurements (providing different values for a key)
- returning unchanged any measurement that cannot be grouped  (instead of being thrown away as currently done by the batcher)

didier-wenzek: Implemented: https://github.com/thin-edge/thin-edge.io/pull/4136/changes/a2569f7e0173a20ae5efedbe0d602199c85a193c

tests/RobotFramework/tests/cumulocity/monitoring/collectd/group-te-measurements.js
didier-wenzek: This script has been removed and replace by a builtin transformer which uses the smallest event-time of a batch as the timestamp for that batch. See https://github.com/thin-edge/thin-edge.io/pull/4136/changes/6251fd67c9a0cfa2d6e601aa149cc759e5d93a43

tests/RobotFramework/tests/cumulocity/monitoring/collectd/group-te-measurements.js
didier-wenzek: Implemented by https://github.com/thin-edge/thin-edge.io/pull/4136/changes/6251fd67c9a0cfa2d6e601aa149cc759e5d93a43

tests/RobotFramework/tests/cumulocity/monitoring/monitor_device_collectd.robot
didier-wenzek: The batch window is now configurable 
https://github.com/thin-edge/thin-edge.io/pull/4136/changes/6251fd67c9a0cfa2d6e601aa149cc759e5d93a43

crates/common/batcher/src/flows.rs
didier-wenzek: This has been fully deprecated and replaced by https://github.com/thin-edge/thin-edge.io/pull/4136/changes/6251fd67c9a0cfa2d6e601aa149cc759e5d93a43

didier-wenzek:
Updating the docs

The following sections use `collectd` as an example of an app that can be managed with thin-edge => okay, unchanged
- https://thin-edge.github.io/thin-edge.io/references/agent/tedge-configuration-management/
- https://thin-edge.github.io/thin-edge.io/references/agent/device-management-api/
- https://thin-edge.github.io/thin-edge.io/references/agent/software-management/
- https://thin-edge.github.io/thin-edge.io/references/agent/device-profiles/
- https://thin-edge.github.io/thin-edge.io/references/mappers/c8y-mapper/
- https://thin-edge.github.io/thin-edge.io/references/mappers/flows/#flow-configuration
- https://thin-edge.github.io/thin-edge.io/references/mqtt-api/
- https://thin-edge.github.io/thin-edge.io/references/software-management-plugin-api/
- https://thin-edge.github.io/thin-edge.io/understand/faq/ 

Partially updated
- https://thin-edge.github.io/thin-edge.io/next/operate/troubleshooting/log-files/

Fully removed because only focused on `collectd`
- https://thin-edge.github.io/thin-edge.io/operate/troubleshooting/device-monitoring/

What about those ones?
- https://thin-edge.github.io/thin-edge.io/start/getting-started/#step-4-monitor-the-device
- https://thin-edge.github.io/thin-edge.io/start/device-monitoring/

I decided to fully remove the first one i.e. from ["A tour of thin-edge.io"](https://thin-edge.github.io/thin-edge.io/start/getting-started/#step-4-monitor-the-device). Indeed, the point of this tour is to highlight key features, not extensions.
Not perfect, because `collectd` is used as an example for configuration management :-(

The section ["device monitoring"](https://thin-edge.github.io/thin-edge.io/start/device-monitoring/) can be kept.
But, I propose to move it under https://thin-edge.github.io/thin-edge.io/operate/monitoring/.

reubenmiller (changes requested):
Great that we can split the collectd stuff and decouple it from the thin-edge.io project, though moving this functionality outside of the core package will required a multi-step transition to ensure we don't break the existing package and can cleaning transition the core collectd logic to the tedge-collectd-setup package (and rename the package in the future).

So I would propose the following actions to facilitate the transition:

1. Split out the Rust related commits from this PR so that we can merge the new "group-measurements" Rust based step for future use by the collectd flow
2. Create a naive "group-measurement" JavaScript implementation of the "group-measurements" so that it can be used in the tedge-collectd-setup in this transition period until the next thin-edge.io release
3. After the next official thin-edge.io release, merge the collectd packaging changes (and docs) to remove the tedge-mapper-collectd service handling. This will require a change in the maintainer script generator to also cleanly stop any existing tedge-mapper-collectd service before removing it so that we don't leave a stale service afterwards (this is something I can help)
4. In the tedge-collectd-setup, transition to use the Rust based "group-measurements" step instead of the naive Javascript implementation (and add the specific thin-edge.io package version as a dependency to the package)

didier-wenzek:
>     1. Split out the Rust related commits from this PR so that we can merge the new "group-measurements" Rust based step for future use by the collectd flow

Done: https://github.com/thin-edge/thin-edge.io/pull/4201

>     2. Create a naive "group-measurement" JavaScript implementation of the "group-measurements" so that it can be used in the tedge-collectd-setup in this transition period until the next thin-edge.io release

Done: https://github.com/thin-edge/tedge-collectd-setup/pull/6/changes/c7e0b734e42b8ef9b5f8038f5750424115a98cd7

>     3. After the next official thin-edge.io release, merge the collectd packaging changes (and docs) to remove the tedge-mapper-collectd service handling. This will require a change in the maintainer script generator to also cleanly stop any existing tedge-mapper-collectd service before removing it so that we don't leave a stale service afterwards (this is something I can help)

I will clean up that PR, once merged https://github.com/thin-edge/thin-edge.io/pull/4201

>     4. In the tedge-collectd-setup, transition to use the Rust based "group-measurements" step instead of the naive Javascript implementation (and add the specific thin-edge.io package version as a dependency to the package)

This will be a simple change reverting https://github.com/thin-edge/tedge-collectd-setup/pull/6/changes/c7e0b734e42b8ef9b5f8038f5750424115a98cd7

```diff
-   { script = "group-measurements.js", interval = "1s" },
+   { builtin = "group-measurements", interval = "1s", config = { time_window = "500ms" } },
```

albinsuresh (approved):
The refactoring has significantly improved the testability of the batching logic and that itself is a huge win.

The only thing that I found weird is batches getting split when conflicting measurements arrived within a batch window. I was wondering if replacing the old values with newer would have been better, so that batches are always emitted at the configured intervals. But, I understand our current rationale to push those into separate batches so that no mesurements are lost and the batch timestamps don't need dynamic re-compute either, which would have further complicated the batching logic.

didier-wenzek:
> The only thing that I found weird is batches getting split when conflicting measurements arrived within a batch window. I was wondering if replacing the old values with newer would have been better, so that batches are always emitted at the configured intervals.

Yeah, the main difference with the deprecated batcher is that there is no attempt to concurrently build more than one batch of messages per topic. So when a message is received for a topic but cannot be added to the associated batch we are forced to take a decision:

- either to keep the batch unchanged and to forward the new message: this is what is done for late messages
- or to close the current batch and to create a new one with the new messages: this is what is done for conflicting messages and those more recent than the batch time-window

> But, I understand our current rationale to push those into separate batches so that no mesurements are lost

Indeed, avoiding message lost has been the driver.

> the batch timestamps don't need dynamic re-compute either, which would have further complicated the batching logic.

There is a bit of window re-computation logic though. When a message is received late (i.e. its event time is older than the oldest batched so far) but not too late (i.e. the difference between that message and the newest batched so far is within the configured time window), then that message is accepted and the batch time window adjusted accordingly.

crates/core/tedge/src/supervisor.rs
```diff
@@ -561,12 +561,12 @@ mod tests {
...
     fn run_all_logging_considers_supervisor_agent_and_mapper_services() {
         assert_eq!(
-            log_service_names(Some(&MapperName::Collectd)),
+            log_service_names(Some(&MapperName::UserDefined(vec!["local".to_string()]))),
             vec![
                 "tedge".to_string(),
                 tedge_agent::AGENT_NAME.to_string(),
                 "tedge-mapper".to_string(),
-                "tedge-mapper-collectd".to_string(),
+                "tedge-mapper".to_string(),
```
didier-wenzek, line 569: @jarhodes314 This is a naive fix to your recent change that is no more applicable once the collectd mapper is deprecated.

Shouldn't `log_service_names()` push a specific mapper name for each `UserDefined` name?

- I don't get why  `MapperName::UserDefined` expects a *list* of names instead of a single name.
- I don't get the impacts of changing the behavior of `log_service_names()`.

Issue #4125, opened 2026-04-14
Create backup of existing tedge.toml before it is modified using tedge upgrade


albinsuresh:
**Is your feature improvement request related to a problem? Please describe.**

The `tedge config upgrade` command currently extracts `mapper.toml` files out of a `tedge.toml` file that has had mapper configurations in it. Since `tedge config upgrade` is implicitly done as part of tedge upgrade using any one of the supported package managers (post installation script), if the upgrade fails for some reason, the user won't be able to simply revert back to the old version and continue using it.

**Describe the solution you'd like**

Take a backup of the `tedge.toml` as `tedge.toml.backup` before the upgrade and inform the user about the presence of the same so that he can use it to restore to the old state, or simply delete it, if everything is working fine. `tedge config get/set` commands could also warn the user of any stale backup files.

PR #4138, opened 2026-04-21
feat: take tedge.toml backup on tedge config upgrade

albinsuresh:
## Proposed changes

Take `tedge.toml` backup during `tedge config upgrade` and notify the user about the existence of any stale backup in `tedge config get/set`.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4125

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/lib.rs
+++ b/crates/common/tedge_config/src/lib.rs
@@ -65,17 +65,30 @@ impl TEdgeConfig {
         self.location().update_toml(update).await
     }
 
-    pub async fn migrate_mapper_configs(self) -> Result<(), TEdgeConfigError> {
+    pub async fn migrate_mapper_configs(self) -> Result<Option<PathBuf>, TEdgeConfigError> {
+        // Check if migration is actually required
+        if !self.location().is_migration_required().await {
+            tracing::info!("No migration required, skipping backup creation");
+            return Ok(None);
+        }
+
+        // Create a backup of tedge.toml before starting migration
+        let backup_path = self
+            .location()
+            .backup_tedge_config(self.config_root())
+            .await?;
+        tracing::info!("Created backup of tedge.toml at: {}", backup_path);
+
         for cloud_type in CloudType::iter() {
             self.location().migrate_mapper_config(cloud_type).await?;
         }
-        Ok(())
+        Ok(Some(backup_path))
     }
 
     pub fn mapper_config_dir<T: ExpectedCloudType>(
         &self,
         profile: Option<&ProfileName>,
-    ) -> camino::Utf8PathBuf {
+    ) -> PathBuf {
         self.location().config_dir::<T>(profile)
     }
 
--- a/crates/core/tedge/src/cli/config/commands/upgrade.rs
+++ b/crates/core/tedge/src/cli/config/commands/upgrade.rs
@@ -11,14 +11,29 @@ impl Command for UpgradeConfigCommand {
     }
 
     async fn execute(&self, tedge_config: TEdgeConfig) -> Result<(), MaybeFancy<anyhow::Error>> {
-        tedge_config.migrate_mapper_configs().await.map_err(|e| {
+        let backup_path = tedge_config.migrate_mapper_configs().await.map_err(|e| {
             MaybeFancy::Unfancy(anyhow::Error::new(e).context(
                 "Failed to migrate mapper configurations. \
                      Fix the underlying issue and run 'tedge config upgrade' again to retry.",
             ))
         })?;
 
-        eprintln!("Configuration updates completed successfully.");
+        match backup_path {
+            Some(path) => {
+                eprintln!("Configuration updates completed successfully.");
+                eprintln!(
+                    "Your original configuration has been backed up to: {} in case the old configuration must be restored.",
+                    path
+                );
+                eprintln!(
+                    "You may delete it after validating that your upgraded configuration is working."
+                );
+            }
+            None => {
+                eprintln!("Configuration is already up to date, nothing was migrated.");
+            }
+        }
+
         Ok(())
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5217,6 +5217,7 @@ dependencies = [
  "strum",
  "strum_macros",
  "tedge_api",
+ "tedge_config",
  "tedge_config_macros",
  "tedge_test_utils",
  "tedge_utils",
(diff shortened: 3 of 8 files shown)


crates/core/tedge/src/cli/config/commands/upgrade.rs
```diff
@@ -11,14 +11,21 @@ impl Command for UpgradeConfigCommand {
...
         })?;
 
         eprintln!("Configuration updates completed successfully.");
+        eprintln!(
+            "Your original configuration has been backed up to: {} in case the old configuration must be restored.",
+            backup_path
+        );
+        eprintln!(
+            "You may delete it after validating that your upgraded configuration is working."
+        );
```
reubenmiller: I'm wondering how loud we have to be about this backup file.

If we're opening communicating about the file, whilst it is good to make users aware of it, I'm wondering if just will prompt more questions from users. And the downgrade scenario.

For instance, to restore the file, we'd assume just need the tedge.toml from the tedge.toml.bak file, but; what services would need to be restarted, does the new version of the mapper really look at the values within the tedge.toml if there are c8y values present, and the mapper.toml file exists? How does the user know which file is controlling the mapper tedge.toml vs. mapper.toml?

@jarhodes314 any thoughts on this?

jarhodes314: It does feel a bit weird to suddenly require a filesystem interaction from the CLI we built largely to prevent the user hand editing files.

I'm also wondering:
1. Do we always create this file, even if there is no change made to `tedge.toml` (either through a lack of cloud configs or calling it after already upgrading to `mapper.toml`)?
2. Do we want to enable downgrading? If that's the motivating use case, we probably want to write tedge.toml.bak like we write to tedge.toml, just before we strip it of cloud configs. That way, we don't make this feature essentially pointless is a user runs `tedge config upgrade` multiple times.

As for what config is read for a cloud, that should be specified in the `tedge connect` config summary. The logic is that if there are any cloud configs in `tedge.toml`, then use that (since we only update `tedge.toml` when the `mapper.toml` files have been written).

albinsuresh: > If we're opening communicating about the file, whilst it is good to make users aware of it, I'm wondering if just will prompt more questions from users. And the downgrade scenario.

How about keeping these messages here in the `config upgrade` but remove the warnings about the backup file from `config get/set` to reduce the noise?

> Do we always create this file, even if there is no change made to tedge.toml (either through a lack of cloud configs or calling it after already upgrading to mapper.toml)?

Currently, yes. That rationale was to enable `config upgrade` invocations from future `tedge` versions as well so that we always have the backup of the previous version. But yeah, that introduced the unexpected side effect of the already backed up "original v1 config" getting replaced by an "already migrated v2 config", if `config upgrade` is invoked multiple times with the same `tedge` version.

> Do we want to enable downgrading? If that's the motivating use case, we probably want to write tedge.toml.bak like we write to tedge.toml, just before we strip it of cloud configs.

The `backup_tedge_config` logic is called right before `migrate_mapper_config`. We just need to avoid replacing an existing `tedge.toml.bak`, if that `upgrade` invocation did not result in any changes to the `tedge.toml`.

albinsuresh: Resolved by [f8c5dfb](https://github.com/thin-edge/thin-edge.io/pull/4138/commits/f8c5dfb9660428ed6af4e2dccff68d7ff40a1114)

Removed the warnings from `tedge config get/set` when a backup is present and also fixed the issue where an existing backup is not replaced if there is nothing to migrate in the config.

crates/common/tedge_config/tests/mapper_config.rs
```diff
@@ -161,6 +161,12 @@ async fn mapper_config_is_not_created_if_another_cloud_exists_in_tedge_toml() {
 async fn mapper_config_is_created_for_new_profile_of_existing_separate_config_cloud() {
     let ttd = TempTedgeDir::new();
 
+    // Create system.toml with empty user and group to use current process ownership
+    ttd.file("system.toml").with_toml_content(toml::toml! {
+        user = ""
+        group = ""
+    });
+
```
didier-wenzek: This change is unrelated and useless

crates/common/tedge_config/src/lib.rs
```diff
@@ -65,17 +65,37 @@ impl TEdgeConfig {
...
+    /// Returns Some(path) if a backup file is found, None otherwise.
+    pub fn check_backup_exists(&self) -> Option<PathBuf> {
+        self.location().check_backup_exists()
     }
 
     pub fn mapper_config_dir<T: ExpectedCloudType>(
         &self,
         profile: Option<&ProfileName>,
-    ) -> camino::Utf8PathBuf {
+    ) -> PathBuf {
```
didier-wenzek, line 91: This change is useless and moving along the reverse direction of other changes assuming Utf8Path.

crates/common/tedge_config/src/lib.rs
```diff
@@ -65,17 +65,37 @@ impl TEdgeConfig {
...
             self.location().migrate_mapper_config(cloud_type).await?;
         }
-        Ok(())
+        Ok(Some(backup_path))
+    }
+
+    /// Check if a stale tedge.toml.bak backup file exists
+    ///
+    /// Returns Some(path) if a backup file is found, None otherwise.
+    pub fn check_backup_exists(&self) -> Option<PathBuf> {
```
didier-wenzek: Better to not have this function `pub` and to keep backup as an internal detail. By the way only used one by a test.

crates/common/tedge_config/src/lib.rs
albinsuresh: It's the same. We have the following import in line no.20:
```
pub use camino::Utf8PathBuf as PathBuf;
```
So, I was just trying to keep it consistent.

didier-wenzek: Okay. Such aliases are confusing though.

didier-wenzek (approved):
Approved

crates/common/tedge_config/src/lib.rs
albinsuresh: Agree. We have the following aliases:

```
pub use camino::Utf8Path as Path;
pub use camino::Utf8PathBuf as PathBuf;
use std::path::Path as StdPath;
```

While I understand the intent to formally declare `tedge_config::Path` as `Utf8Path`, it definitely is confusing, when code snippets are read in isolation, esp without the full path.

Issue #4139, opened 2026-04-21
Log file request leaves temporary files in `/tmp` (`tmp.path`)


d-timotijevic:
**Describe the bug**
When a log file request is triggered, thin-edge.io creates temporary files in `/tmp` (I'm assuming directory is originating from `tmp.path` configuration setting). These files are not removed after the request has completed.

As a result, repeated log file requests leave stale temporary files behind in `/tmp`, which can accumulate over time and unnecessarily consume disk space, or in some cases RAM.

**To Reproduce**
1. Run a Linux system with thin-edge.io version `1.7.1`.
2. Trigger a log file request from the system or UI.
3. Access the device shell.
4. Check the `/tmp` directory for newly created temporary files.
5. Trigger additional log file requests.
6. Observe that the temporary files remain in `/tmp` after the requests complete.

**Expected behavior**
Temporary files created during a log file request should be automatically cleaned up once they are no longer needed.

**Screenshots**

<img width="1025" height="814" alt="Image" src="https://github.com/user-attachments/assets/d811487f-d68a-4a0f-9129-a5472f521a4c" />

<img width="1466" height="293" alt="Image" src="https://github.com/user-attachments/assets/8c652ef3-65a2-4e44-917c-8d3f6e216b2d" />

**Environment (please complete the following information):**
* OS [incl. version]: Linux
* thin-edge.io version [e.g. 0.1.0]: `1.7.1`

d-timotijevic:
Similar to issue described in #3786.

reubenmiller:
@d-timotijevic thanks for the ticket.

What log handler are you using? The filename format would suggest that it might be coming from a custom log handler.

reubenmiller:
I stand corrected, requesting a log file via a command definitely leaves a copy of the files.

With the following tedge-log-plugin.toml configuration:

```toml
# cat /etc/tedge/plugins/tedge-log-plugin.toml
[[files]]
type = "device_profile"
path = "/var/log/tedge/agent/workflow-device_profile-*"

[[files]]
type = "software-management"
path = "/var/log/tedge/agent/workflow-software*"

[[files]]
type = "firmware_update"
path = "/var/log/tedge/agent/workflow-firmware_update-*"

[[files]]
type = "restart"
path = "/var/log/tedge/agent/workflow-restart-*"

[[files]]
type = "all-workflows"
path = "/var/log/tedge/agent/workflow-*"

[[files]]
type = "shell"
[...]

PR #4141, opened 2026-04-23
fix: cleanup temp log artefacts created for log_upload

albinsuresh:
## Proposed changes

Cleanup the following temp files created during the `log_upload` operation handling:
- [x] The temp file created by the `file` plugin combining all target files
- [x] The uploaded log file in the file transfer service, cleaned up by the mapper

The `agent` was already deleting the tmp file that it creates after applying the filters like line count and search text.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4139

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
@@ -42,82 +42,95 @@ impl OperationContext {
             }
         };
 
-        let smartrest_topic = &target.smartrest_publish_topic;
-
         match command.status() {
             CommandStatus::Executing => Ok(OperationOutcome::Executing {
                 extra_messages: vec![],
             }),
             CommandStatus::Successful => {
-                // Send a request to the Downloader to download the file asynchronously from FTS
-                let log_filename = format!("{}-{}", command.payload.log_type, cmd_id);
-
-                let tedge_file_url = &command.payload.tedge_url;
-
-                let destination_dir = tempfile::tempdir_in(self.tmp_dir.as_std_path())
-                    .context("Failed to create a temporary directory")?;
-                let destination_path = destination_dir.path().join(log_filename);
-
-                let download_request = DownloadRequest::new(tedge_file_url, &destination_path);
-                let (_, download_result) = self
-                    .downloader
-                    .clone()
-                    .await_response((cmd_id.into(), download_request))
-                    .await
-                    .context("Unexpected ChannelError")?;
-
-                let download_response = download_result.context(
-                    "tedge-mapper-c8y failed to download log from file transfer service",
-                )?;
-
-                let file_path = Utf8PathBuf::try_from(download_response.file_path)
-                    .map_err(|e| e.into_io_error())
-                    .context("Could not parse file path as Utf-8")?;
-
-                let event_type = &command.payload.log_type;
-
-                let (binary_upload_event_url, upload_result) = self
-                    .upload_file(
-                        &target.external_id,
-                        &file_path,
-                        None,
-                        Some(mime::TEXT_PLAIN),
-                        cmd_id,
-                        event_type.clone(),
-                        None,
-                    )
-                    .await
-                    .context("Could not upload log file to C8y")?;
-
-                let smartrest_response = super::get_smartrest_response_for_upload_result(
-                    upload_result,
-                    binary_upload_event_url.as_str(),
-                    CumulocitySupportedOperations::C8yLogFileRequest,
-                    self.smart_rest_use_operation_id,
-                    self.get_operation_id(cmd_id),
-                );
-
-                let c8y_notification = MqttMessage::new(smartrest_topic, smartrest_response);
-
-                self.upload_operation_log(
-                    &target.external_id,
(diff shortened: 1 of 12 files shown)


crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
```diff
@@ -108,6 +108,9 @@ impl OperationContext {
                 .await
                 .context("Could not upload operation log")?;
 
+                // Delete the file from the file transfer service after successful upload
+                self.http_delete(tedge_file_url).await;
```
didier-wenzek: Shouldn't the file also be deleted after a *failed* transfer to C8Y?

Actually, the code for the `CommandStatus::Successful` case is so long and with so many error cases (aka `?`) before reaching that file deletion, that it would deserve to be extracted in some helper method, say `transfer_to_c8y`.

This would help to improve error handling:

```rust
      CommandStatus::Successful => {
          let tedge_file_url = &command.payload.tedge_url;
          let maybe_c8y_notification = self.transfer_to_c8y(tedge_file_url, command).await;

          self.http_delete(tedge_file_url).await;

          let c8y_notification = maybe_c8y_notification.context("...")?;
          Ok(OperationOutcome::Finished {
              messages: vec![c8y_notification],
          })
      }
```

plugins/tedge_file_log_plugin/src/bin.rs
```diff
@@ -125,6 +125,8 @@ pub fn run(cli: FileLogCli, plugin_config: TEdgeConfigView) -> anyhow::Result<()
                         let line = line?;
                         println!("{}", line);
                     }
+
+                    std::fs::remove_file(log_path)?;
```
rina23q, line 129: You can ignore this. What I was thinking here:

If we use `tempfile` instead of regular file creation for the temporary file, we probably don't need to have this `remove_file()`. However, then the return type of `FileLogPlugin::get()` and the actual function `new_read_logs()` would need to be changed to `NamedTempFile`. Doable, but not sure if it's worth it.

rina23q (approved):
Approved. Confirmed it fixed the reported issue.

Bravo555 (commented):
NOTE: The same issue happens with config snapshot operation

When the same check is applied to config snapshot test:
```
diff --git a/tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot b/tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot
index 824c5536c5..a8c08e0de6 100644
--- a/tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot
+++ b/tests/RobotFramework/tests/cumulocity/configuration/configuration_with_file_transfer_https.robot
@@ -54,6 +54,11 @@ File Transfer Service is accessible over HTTPS from child device
 Configuration snapshots are uploaded to File Transfer Service via HTTPS
     Get Configuration Should Succeed    device=${CHILD_SN}    external_id=${PARENT_SN}:device:${CHILD_SN}
 
+    # Validate that all temporary files created for the log upload operation are cleaned up
+    ThinEdgeIO.Set Device Context    ${FTS_SN}
+    Execute Command    ls /tmp/CONFIG1*    exp_exit_code=2
+    Execute Command    ls /var/tedge/file-transfer/${PARENT_SN}:device:${CHILD_SN}/config_snapshot/CONFIG1*    exp_exit_code=2
+
 Configuration snapshots are uploaded to File Transfer Service via HTTPS with client certificate
     Enable Certificate Authentication for File Transfer Service
     Get Configuration Should Succeed    device=${CHILD_SN}    external_id=${PARENT_SN}:device:${CHILD_SN}
```

It fails:

```
`ls /var/tedge/file-transfer/TS...` returned an unexpected exit code
stdout:
/var/tedge/file-transfer/TST_hoard_massive_carrier:device:TST_detour_smart_gang/config_snapshot/CONFIG1-c8y-mapper-54216380

stderr:
```

I'd say since this is the same issue, it should should also be fixed in this PR.

albinsuresh:
> NOTE: The same issue happens with config snapshot operation
> 
> I'd say since this is the same issue, it should should also be fixed in this PR.

Resolved by [6d3966f](https://github.com/thin-edge/thin-edge.io/pull/4141/commits/6d3966f739085509f6a99a648713f14a0dcef730)

Bravo555 (approved):
Fix also addressed configuration operation, so approving.

plugins/tedge_file_log_plugin/src/bin.rs
rina23q: I see now `tedge_config_manager` is using `tempfile`. Then, it would be good also `tedge_log_manager` is doing the same.

crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
albinsuresh: Agree, that the operation may have failed after a partial upload, which still needs to be deleted. And since DELETE is idempotent, it is safe it do it even in the failed case.

didier-wenzek (approved):
Approved

PR #4142, opened 2026-04-23
refactor: cleanup unused unwanted code from file api

rina23q:
## Proposed changes
Refactor the code to remove unused code and old APIs.

- [x] Get rid of unused code
- [x] Get rid of helper functions to create a directory or a file
- [x] Refactor usages of `config_dir`, `data_dir`, `tmp_dir`, `state_path`, `logs_path`, `run_path` to take `TedgePaths` instead of `Utf8PathBuf`, so that easy to manipulate their subdirectories and files. 

Out of scope of this PR:
Need to think how to do because they are used by the new `ManagedDir` and `ManagedFile`...
- [ ] Get rid of `create_directory_xxx()` under `PermissionEntry`
- [ ] Get rid of the whole `PermissionEntry` interface (?)

## Types of changes

## Paste Link to the issue
#4115 

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/lib.rs
+++ b/crates/common/tedge_config/src/lib.rs
@@ -42,6 +42,30 @@ impl TEdgeConfig {
         TedgePaths::from_root_with_defaults(self.root_dir(), system.user, system.group)
     }
 
+    pub fn data_root(&self) -> TedgePaths {
+        let system = self.read_system_config();
+        TedgePaths::from_root_with_defaults(self.data.path.clone(), system.user, system.group)
+    }
+
+    pub fn state_root(&self) -> TedgePaths {
+        let system = self.read_system_config();
+        TedgePaths::from_root_with_defaults(
+            self.agent.state.path.clone(),
+            system.user,
+            system.group,
+        )
+    }
+
+    pub fn logs_root(&self) -> TedgePaths {
+        let system = self.read_system_config();
+        TedgePaths::from_root_with_defaults(self.logs.path.clone(), system.user, system.group)
+    }
+
+    pub fn run_root(&self) -> TedgePaths {
+        let system = self.read_system_config();
+        TedgePaths::from_root_with_defaults(self.run.path.clone(), system.user, system.group)
+    }
+
     /// Load [TEdgeConfig], using a separate mapper config file as the default
     /// behaviour if no clouds are already configured
     ///
--- a/crates/common/tedge_config/src/lib.rs
+++ b/crates/common/tedge_config/src/lib.rs
@@ -37,16 +37,31 @@ impl TEdgeConfig {
         SystemConfig::try_new(self.root_dir()).unwrap_or_default()
     }
 
+    /// The root directory for thin-edge configuration files such as `tedge.toml`
+    /// and subdirectories.
+    ///
+    /// Defaults to `/etc/tedge`.
     pub fn config_root(&self) -> TedgePaths {
         let system = self.read_system_config();
         TedgePaths::from_root_with_defaults(self.root_dir(), system.user, system.group)
     }
 
+    /// The root directory for persistent thin-edge data that needs to survive
+    /// service restarts. This includes the file-transfer repository used by
+    /// `tedge-agent` to stage files for child devices (stored under
+    /// `file-transfer/`), firmware caches, and other service-specific data.
+    ///
+    /// Defaults to `/var/tedge`. (configurable via `data.path`)
     pub fn data_root(&self) -> TedgePaths {
         let system = self.read_system_config();
         TedgePaths::from_root_with_defaults(self.data.path.clone(), system.user, system.group)
     }
 
+    /// The root directory where `tedge-agent` persists its operational state
+    /// across reboots, such as in-progress operation records used by the
+    /// software manager, restart manager, and operation workflow engine.
+    ///
+    /// Defaults to `/data/tedge/agent` (configurable via `agent.state.path`).
     pub fn state_root(&self) -> TedgePaths {
         let system = self.read_system_config();
         TedgePaths::from_root_with_defaults(
@@ -56,14 +71,32 @@ impl TEdgeConfig {
         )
     }
 
+    /// The root directory for thin-edge log files.
+    ///
+    /// Defaults to `/var/log/tedge`. (configurable via `logs.path`)
     pub fn logs_root(&self) -> TedgePaths {
         let system = self.read_system_config();
         TedgePaths::from_root_with_defaults(self.logs.path.clone(), system.user, system.group)
     }
 
+    /// The root directory for runtime files such as lock files (stored under
+    /// `lock/`) that prevent multiple instances of the same service from
+    /// running simultaneously, and restart marker files used by the restart
+    /// operation handler.
+    ///
+    /// Defaults to `/run`. (configurable via `run.path`)
     pub fn run_root(&self) -> TedgePaths {
         let system = self.read_system_config();
         TedgePaths::from_root_with_defaults(self.run.path.clone(), system.user, system.group)
+    }
+
+    /// The root directory for temporary files created during operations, such
+    /// as files staged before being moved to their final destination.
+    ///
+    /// Defaults to `/tmp`. (configurable via `tmp.path`)
+    pub fn tmp_root(&self) -> TedgePaths {
+        let system = self.read_system_config();
+        TedgePaths::from_root_with_defaults(self.tmp.path.clone(), system.user, system.group)
     }
 
     /// Load [TEdgeConfig], using a separate mapper config file as the default
--- a/crates/common/tedge_config/src/lib.rs
+++ b/crates/common/tedge_config/src/lib.rs
@@ -6,6 +6,7 @@ mod system_toml;
 pub use system_toml::*;
 
 pub mod tedge_toml;
+use tedge_api::path::DataDir;
 use tedge_config_macros::ProfileName;
 pub use tedge_toml::error::*;
 pub use tedge_toml::models;
@@ -52,9 +53,11 @@ impl TEdgeConfig {
     /// `file-transfer/`), firmware caches, and other service-specific data.
     ///
     /// Defaults to `/var/tedge`. (configurable via `data.path`)
-    pub fn data_root(&self) -> TedgePaths {
+    pub fn data_root(&self) -> DataDir {
         let system = self.read_system_config();
-        TedgePaths::from_root_with_defaults(self.data.path.clone(), system.user, system.group)
+        let paths =
+            TedgePaths::from_root_with_defaults(self.data.path.clone(), system.user, system.group);
+        DataDir::from(paths)
     }
 
     /// The root directory where `tedge-agent` persists its operational state
--- a/crates/common/tedge_utils/src/paths.rs
+++ b/crates/common/tedge_utils/src/paths.rs
@@ -340,6 +340,11 @@ impl ManagedFile {
         self
     }
 
+    pub fn with_mode(mut self, mode: u32) -> Self {
+        self.mode = mode;
+        self
+    }
+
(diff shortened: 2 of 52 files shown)


crates/common/tedge_config/src/lib.rs
```diff
@@ -42,6 +42,30 @@ impl TEdgeConfig {
         TedgePaths::from_root_with_defaults(self.root_dir(), system.user, system.group)
     }
 
+    pub fn data_root(&self) -> TedgePaths {
```
jarhodes314: It might be good to add some doc comments on these functions explaining things like what "data" we store in data_root, and what path that is by default?

crates/common/tedge_utils/src/paths.rs
```diff
@@ -220,6 +220,33 @@ impl ManagedDir {
...
+            root: self.root.clone(),
+            path: self.path.join(path),
+            owner: self.owner.clone(),
+            mode: DEFAULT_DIR_MODE,
+            respect_existing: false,
+            warn_and_ignore_permission_errors: false,
+        })
+    }
+
+    // TODO: create `dir_unchecked`?
```
didier-wenzek: To remove or implement before merging.

What would be the point of this `dir_unchecked` method?

crates/core/tedge_agent/src/agent.rs
```diff
@@ -75,15 +76,15 @@ pub(crate) struct AgentConfig {
     pub restart_config: RestartManagerConfig,
     pub sw_update_config: SoftwareManagerConfig,
     pub operation_config: OperationConfig,
-    pub config_dir: Utf8PathBuf,
-    pub tmp_dir: Arc<Utf8Path>,
-    pub run_dir: Utf8PathBuf,
+    pub config_dir: TedgePaths,
+    pub tmp_dir: Arc<Utf8Path>, // TODO: change it to TedgePaths
```
didier-wenzek: Yeah, it would be good to implement that change is this PR.

crates/core/tedge_api/src/path.rs
```diff
@@ -52,6 +60,10 @@ impl DataDir {
         self.0.join(path.as_ref().to_path_buf())
     }
 
+    pub fn root_dir(&self) -> Utf8PathBuf {
+        self.0.clone()
+    }
```
didier-wenzek, line 32: This new method is unused.

crates/core/tedge_api/src/path.rs
```diff
@@ -2,6 +2,8 @@ use camino::Utf8Path;
 use camino::Utf8PathBuf;
 use std::path::Path;
 
+// TODO: We might consider removing this file entirely.
```
didier-wenzek: You mean to remove the `path` module?

crates/extensions/c8y_firmware_manager/src/worker.rs
```diff
@@ -268,8 +269,10 @@ impl FirmwareManagerWorker {
             attempt: 1,
         };
 
+        let data_root = TedgePaths::from_root_with_defaults(&self.config.data_dir, "", "");
```
didier-wenzek: I was expecting here a call to `TEdgeConfig::data_root`.

didier-wenzek (commented):
I will be happy to see this PR merged. There are still some TODOs but I think we must avoid taking too much time polishing. Many improvements can be deferred to other PRs. This will be easier to review.

jarhodes314 (approved):

crates/common/tedge_utils/src/paths.rs
rina23q: The `dir()` method returns `Result` because if the input path refers outside of the root directory, e.g., `../../some`, the `resolve()` method returns an error. 
However, if the input path is hardcoded like `.dir("cache")`, this never fails. 

In my refactoring, there are many `.unwrap()` or `?`, but they actually never fail. That's why I was considering a shortcut like we have for `Topic::new_unchecked()`.

didier-wenzek: Okay, I get your intent. However, better to keep explicit `root.dir("mappers").expect("Valid sub directory")`.

crates/core/tedge_api/src/path.rs
rina23q: Yes, the entire module. I found it was not so brilliant together with `TedgePaths` and `ManagedDir. 

To address your comment below, I actually did refactoring this `path` module. Previously, it wrapped `Utf8PathBuf`, however, this makes difficult to work with the new `TedgePaths` and others. 
I changed this to wrap `TedgePaths`. 

[This refactoring](https://github.com/thin-edge/thin-edge.io/pull/4142/changes/7934acedbba2c2c4f882fb2dae78e5c5ff598c3e) leads a lot of changes, but I believe the code readability is clearer than before.

crates/common/tedge_utils/src/paths.rs
rina23q: Ok, then will not implement this.

crates/core/tedge_agent/src/agent.rs
rina23q: Done in https://github.com/thin-edge/thin-edge.io/pull/4142/changes/ad72f9602be4bb12fa322797864543299ef44012 and 34423311187daad75f879c285fdcbc7e0a150a6c

crates/extensions/c8y_firmware_manager/src/worker.rs
rina23q: As I mentioned above, I did it together with further refactoring on the `DataDir` struct.

didier-wenzek (approved):
Approved. Thanks. The tedge file API is now more consistent.

crates/core/tedge_agent/src/software_manager/config.rs
```diff
@@ -1,16 +1,17 @@
...
-    pub tmp_dir: Utf8PathBuf,
-    pub config_dir: Utf8PathBuf,
-    pub state_dir: Utf8PathBuf,
-    pub sm_plugins_dir: Utf8PathBuf,
-    pub log_dir: Utf8PathBuf,
+    pub tmp_dir: TedgePaths,
+    pub config_dir: TedgePaths,
+    pub state_dir: TedgePaths,
+    pub sm_plugins_dir: ManagedDir,
+    pub log_dir: ManagedDir,
```
albinsuresh, line 14: That inconsistency between a few being `TedgePaths` and the rest being `ManagedDir` looks a bit weird. It would have been nice if that differentiation between root directories and sub-directories were not visible everywhere. I mean, I was wondering why every field couldn't be a `ManagedDir`. But, that's just my stylistic preference. Since it's just a matter of optics, I'm fine with this.

crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs
```diff
@@ -81,7 +82,8 @@ impl SupportedOperations {
         };
 
         // Create directory for a device if it doesn't exist yet
-        file::create_directory_with_defaults(ops_file.parent().expect("should never fail")).await?;
+        let parent_dir = TedgePaths::from_root_with_defaults(ops_file.parent().unwrap(), "", "");
```
albinsuresh, line 85: Since this is production code, shouldn't the real `user` and `group` values from `system.toml` be passed here instead of `""`?

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -407,11 +408,17 @@ impl C8yMapperBuilder {
         self.flow_context = Some(flow_context);
     }
 
-    pub async fn init(config: &C8yMapperConfig) -> Result<(), FileError> {
+    pub async fn init(config: &C8yMapperConfig) -> Result<(), PathsError> {
         // Create c8y operations directory
-        create_directory_with_defaults(config.ops_dir.as_std_path()).await?;
+        TedgePaths::from_root_with_defaults(&config.ops_dir, "", "")
```
albinsuresh, line 413: Same comment as above.

crates/core/tedge/src/cli/diag/collect.rs
```diff
@@ -39,9 +39,13 @@ impl Command for DiagCollectCommand {
     }
 
     async fn execute(&self, _: TEdgeConfig) -> Result<(), MaybeFancy<anyhow::Error>> {
-        file::create_directory_with_defaults(&self.diag_dir)
+        let diag_dir_root = TedgePaths::from_root_with_defaults(&self.diag_dir, "", "");
```
albinsuresh, line 42: Same comment as above.

crates/core/tedge_agent/src/device_profile_manager/mod.rs
albinsuresh, line 13: Same comment as above.

crates/core/tedge_agent/src/software_manager/config.rs
rina23q: Here `tmp_dir`, `config_dir`, `state_dir` are the root directory that user can configure them via CLI or `tedge config` directly. So, they are absolutely "root" directories. => `TedgePaths`
On the other hand, `sm_plugin_dir` is `{config_dir}/sm-plugins` (the name "sm-plugins" user cannot change) and here's `log_dir` is actually `{logs.path}/agent`. => `ManagedDir` since they are subdirectories of the "root" directories.

crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs
rina23q: (To all the related comments) I think it's a valid point. However, the previous code didn't take care of it either. So, I prefer addressing in a separated PR as it will be a "fix".

crates/core/tedge_agent/src/software_manager/config.rs
albinsuresh: > differentiation between root directories and sub-directories

Yeah yeah, that's what I meant by that "differentiation". The question was whether that differentiation matters from the perspective of a software manager or config manager. May be it does. The distinction that some of them were provided by the user and the others were derived. I just felt that even `TedgePaths` being represented as `ManagedDir` would have made things more consistent looking without losing any capabilities. But yeah, this is also fine.

crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs
albinsuresh: Sure, I'm fine with that.

albinsuresh (approved):
The refactoring looks fine. The bug that was discovered around the mapper and a few other components ignoring the `user` and `group` from `system.toml` can be fixed in an independent PR, as it is not directly linked to the refactoring.

crates/core/tedge_agent/src/software_manager/config.rs
rina23q: The point is how to interpret the "root" directory depends on a context. In the context of software manager or config manager, their root can be `{config_dir}/sm-plugins` or `{config_dir}/plugins` or whatever.

However, after your review about `system.toml` settings for `user`/`group`, In technical point of view, treating the user config directories as root regardless of the contexts is simpler to address them. `TEdgeConfig::config_root()` honors `system.toml` settings, and the following `ManagedDir` will inherit the configuration.

crates/core/tedge/src/cli/diag/collect.rs
rina23q: I think `diag` is a special. This `diag_dir` is the output directory:

> Directory where output tarball and temporary output files are stored. The path from tmp.path will be used by default

where should be owned by the current process's user.

crates/core/tedge_agent/src/software_manager/config.rs
rina23q: Ah I think I got what you meant now. `TedgePaths` can be transformed to `ManagedDir` by `.root_dir()`. So, indeed, all can be `ManagedDir`.

PR #4151, opened 2026-04-30
fix: Don't write time if logging to journald

Bravo555:
## Proposed changes

Journald already records the timestamp of each log and allows the user to easily customise the format using journalctl's `--output` flag, and to filter the logs using `--since/--until`. So only format time if not logging to journald.

Before:

Screenshot From 2026-04-30 10-00-58

After:

Screenshot From 2026-04-30 10-11-30

In this PR, while we use `tracing-journald` to detect if we're logging to journald and change our formatting based on that, we aren't yet fully using its functionality to send structured logs containing log levels and fields, because the default formatting of the journald's `MESSAGE` field (i.e. the log message shown by `journalctl` after the timestamp and identifier) does not format attached fields and the target to be present in the message.

Given an event like this:

```rust
info!(field = 42, "event happened");
```

our current logging to `stderr` gives us this:

```
Apr 30 08:27:10 32df1eef46c8 tedge-mapper[399038]:   INFO tedge_config::system_toml::log_level: event happened field=42
```

Logging with journald gives us this:
```
Apr 30 08:30:26 32df1eef46c8 tedge-mapper[406816]: **event happened**
```

Though field are accessible when selecting a more verbose output level:

```
{
        "_HOSTNAME" : "32df1eef46c8",
        "_SYSTEMD_INVOCATION_ID" : "192c32fe9c8f4fdf9dd20a9667569419",
        "_CMDLINE" : "/workspace/target/debug/tedge-mapper c8y",
        ...
        "CODE_FILE" : "crates/common/tedge_config/src/system_toml/log.rs",
        "__MONOTONIC_TIMESTAMP" : "196768238204",
        "MESSAGE" : "event happened",
        "CODE_LINE" : "66",
        "F_FIELD" : "42",
        "_PID" : "406816",
        "SYSLOG_IDENTIFIER" : "tedge-mapper",
        "_GID" : "995",
}
```

Given that this is a change from the current behaviour where in some cases we assume that the field should appear next to the message where it's relevant, which could be surprising to the users, we will still log to stderr and only switch once we can put embed relevant fields into the message.

There are some proposed solutions (https://github.com/tokio-rs/tracing/pull/3474), but they haven't been merged yet. 

Still, one nice thing about logging to journald is that information relevant to debugging, like `CODE_FILE` and `CODE_LINE` are attached to the logs by default and can be shown with `--output json-pretty/verbose` and at the same time they don't have to shown to users by default. With these and other fields it can make our debugging easier while not making logs too noisy for the users. 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/system_toml/log_level.rs
+++ b/crates/common/tedge_config/src/system_toml/log_level.rs
@@ -5,6 +5,7 @@ use camino::Utf8Path;
 use std::io::IsTerminal;
 use std::str::FromStr;
 use std::sync::Arc;
+use tracing_subscriber::fmt::time::FormatTime;
 use tracing_subscriber::util::SubscriberInitExt;
 
 const DEFAULT_LEVEL: tracing::Level = tracing::Level::INFO;
@@ -69,6 +70,17 @@ fn logger(
     default_level: tracing::Level,
 ) -> Result<Arc<dyn tracing::Subscriber + Send + Sync>, SystemTomlError> {
     let subscriber = subscriber_builder!();
+
+    // Added by systemd if the process is running as systemd unit
+    // https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#%24INVOCATION_ID
+    let is_running_as_systemd_unit = std::env::var("INVOCATION_ID").is_ok();
+
+    // disable time formatting if journald because journald already records the time
+    let subscriber = if is_running_as_systemd_unit {
+        subscriber.with_timer(TimeFormat::Disabled)
+    } else {
+        subscriber.with_timer(TimeFormat::Enabled)
+    };
 
     // first use log level from flags
     let log_level = flags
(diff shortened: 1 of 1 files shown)


crates/common/tedge_config/src/system_toml/log_level.rs
```diff
@@ -70,6 +71,15 @@ fn logger(
 ) -> Result<Arc<dyn tracing::Subscriber + Send + Sync>, SystemTomlError> {
     let subscriber = subscriber_builder!();
 
+    let journald_layer = tracing_journald::layer().ok();
+
+    // disable time formatting if journald because journald already records the time
+    let subscriber = if journald_layer.is_some() {
```
didier-wenzek: This check is brittle. Actually, this is only checking that `journald` is running and not that `tedge` is using it.

Directly running the mapper (as I often do during development)  on a systemd box, leads to the following where all timestamps are discarded:

```
$ sudo tedge-mapper c8y               
  INFO Runtime: Running Signal-Handler-0
  INFO Runtime: Running HealthMonitorActor-1
  ...
```

didier-wenzek (commented):
I'm not really happy to add a dependency just to deal with journald weirdness (a logger should not redact its input - adding timestamp and, worse, discarding fields which violate its conventions).

Also note that, with the `--output=cat` option, journald doesn't add any extra timestamp:

```
$ journalctl --output=cat  --unit tedge-mapper-c8y 
2025-06-04T09:34:59.130172495Z  INFO tedge_mqtt_bridge::health: MQTT bridge connected to cloud broker
```

crates/common/tedge_config/src/system_toml/log_level.rs
Bravo555: fixed in f78aedc.

Dependency was removed and now we're only check `INVOCATION_ID` envvar that is added by systemd if process is running as part of a systemd unit, so it correctly doesn't remove the clock if the binary is run directly.

didier-wenzek (approved):
Approved

PR #4145, opened 2026-04-24
feat: switch to using rumqttc fork to resolve potential security issues

reubenmiller:
## Proposed changes

Switch to using a rumqttc fork (https://github.com/thehouseisonfire/rumqtt). There is a discussion here on the upstream project about the need for a fork, https://github.com/bytebeamio/rumqtt/issues/1029.

Currently below shows the impact of switching to the fork (it may be incomplete):

* rumqttc fork only contains rumqttc (not rumqttd), so we'll also have to fork the upstream rumqtt project to update the rumqttd test dependencies (to avoid having to still have project wide CVE ignore rules)
* Requires upgrading MSRV to 1.89 (from 1.85)
* Some rumqttc interfaces that we use for logging have moved from being public to private making them inaccessible (loss of functionality)

## Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue
<br/>

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s. You can activate automatic signing by running `just prepare-dev` once)
- [ ] I ran `just format` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I used `just check` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #4152, opened 2026-05-04
deps: upgrade to rustls-webpki 0.103.13

didier-wenzek:
## Proposed changes

This is an alternative to https://github.com/thin-edge/thin-edge.io/pull/4145

- Use a fork of rumqttc where the dependencies to rutsls-webpki have been updated to the latest.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -17,11 +17,6 @@ ignore = [
     # appropriate entries should be added to clippy.toml to disallow these methods.
     "RUSTSEC-2023-0071",
     "RUSTSEC-2026-0009",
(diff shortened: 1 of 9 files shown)


jarhodes314 (approved):
I'm happy with these changes, and pretty confident the cryptoprovider handling is correct. The [docs for `install_default`](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default) say:

> This can be called successfully at most once in any process execution.

This sounds to me like "if you try and call the method multiple times it'll fail but nothing bad will happen beyond that", so we can't change the provider once it's set, but we can call this as many times as we like to ensure it's definitely been called at least once. As a result, ignoring the error sounds perfectly reasonable to me. If we were actually mishandling this, rustls would be throwing errors when we try and use it.

Bravo555 (approved):

PR #4155, opened 2026-05-07
fix: hot reloading of flows over symlinks

didier-wenzek:
## Proposed changes

When the flow directory passed to a mapper is a symlink (e.g. `/etc/tedge/mapper` -> `/data/tedge/mapper`), then flows can no more be added or updated while the mapper is running. Errors are reported telling the flow filename is "not a valid filename for a flow", which is more puzzling than helpful.

The root cause of the issue is that the name of a flow is derived from its file path, checking first that this file sits in the mapper flows directory. However this check was done without properly canonicalizing first the flow and directory paths.

- [x] Improve the error message when a flow name cannot be derived from its file path.
- [x] Reproduce the error - combining symlinks and dynamic flow reloading
- [x] Reproduce a similar error - combining symlinks and invoking `tedge flows test` on a specific `--flow`
- [x] Ensure that a flow registry root path is canonicalized. 

## Types of changes

## Paste Link to the issue

- https://community.cumulocity.com/t/error-not-a-valid-filename-for-a-flow-deploying-tedge-flow/14538

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/config.rs
+++ b/crates/extensions/tedge_flows/src/config.rs
@@ -154,7 +154,6 @@ pub enum ConfigError {
 
 /// ```
 /// use tedge_flows::derive_flow_name;
-/// assert_eq!(derive_flow_name("/flows".into(), "/flows/flow.toml".into()), Some("flow".into()));
 /// assert_eq!(derive_flow_name("/flows".into(), "/flows/hello.toml".into()), Some("hello".into()));
 /// assert_eq!(derive_flow_name("/flows".into(), "/flows/hello/flow.toml".into()), Some("hello".into()));
 /// assert_eq!(derive_flow_name("/flows".into(), "/flows/hello/world.toml".into()), Some("hello/world".into()));
@@ -163,6 +162,10 @@ pub enum ConfigError {
 /// assert_eq!(derive_flow_name("/flows".into(), "/flows/hello/params.toml".into()), None);
 /// assert_eq!(derive_flow_name("/flows".into(), "/flows/hello/world.js".into()), None);
 /// assert_eq!(derive_flow_name("/flows".into(), "/unrelated/flows/hello.toml".into()), None);
+/// // Not recommended but working:
+/// assert_eq!(derive_flow_name("/flows".into(), "/flows/flow.toml".into()), Some("flow".into()));
+/// assert_eq!(derive_flow_name("/flows".into(), "/flows/params/flow.toml".into()), Some("params".into()));
+/// assert_eq!(derive_flow_name("/flows".into(), "/flows/flow/flow.toml".into()), Some("flow".into()));
 /// ```
 pub fn derive_flow_name(flows_dir: &Utf8Path, flow_path: &Utf8Path) -> Option<String> {
     let path = flow_path.strip_prefix(flows_dir).ok()?;
--- a/crates/extensions/tedge_flows/src/config.rs
+++ b/crates/extensions/tedge_flows/src/config.rs
@@ -123,8 +123,8 @@ pub enum IntervalConfig {
 
 #[derive(thiserror::Error, Debug)]
 pub enum ConfigError {
-    #[error("Not a valid filename for a flow")]
-    IncorrectFlowFilename,
+    #[error("The flow name cannot be derived from its relative path to {dir}")]
+    FlowNameCannotBeDerived { dir: Utf8PathBuf, path: Utf8PathBuf },
 
     #[error("Not a valid MQTT topic: {0}")]
     IncorrectTopic(String),
@@ -317,7 +317,10 @@ impl FlowConfig {
         }
 
         let Some(name) = derive_flow_name(flows_dir, &source) else {
-            return Err(ConfigError::IncorrectFlowFilename);
+            return Err(ConfigError::FlowNameCannotBeDerived {
+                dir: flows_dir.to_owned(),
+                path: source.to_owned(),
+            });
         };
 
         detect_loop(&name, &input, &output, self.expect_loop)?;
--- a/crates/core/tedge/src/cli/flows/cli.rs
+++ b/crates/core/tedge/src/cli/flows/cli.rs
@@ -212,10 +212,7 @@ impl TEdgeFlowsCli {
         flows_dir: &Utf8PathBuf,
         js_config: JsRuntimeConfig,
     ) -> Result<MessageProcessor<BaseFlowRegistry>, Error> {
-        let flows_dir = flows_dir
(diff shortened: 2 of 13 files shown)


crates/extensions/tedge_flows/src/config.rs
```diff
@@ -123,8 +123,8 @@ pub enum IntervalConfig {
 
 #[derive(thiserror::Error, Debug)]
 pub enum ConfigError {
-    #[error("Not a valid filename for a flow")]
-    IncorrectFlowFilename,
+    #[error("The flow name cannot be derived from its relative path to {dir}")]
```
Bravo555, line 126: thought: Before and after, the error says that the name of the flow is wrong, but I'm not sure if the message is not too indirect for users who aren't sure how the name of the flow is derived.

Maybe it is documented somewhere, but even if it is, maybe the message should be more direct and say exactly how the path is derived and what we expect: that the canonical path of the flow `.toml` file should be inside the canonical path of `dir`, or in other words, that after resolving all symlinks, in the path of `dir` and in the path of the flow, `path` should be inside `dir`.

Bravo555 (approved):
Approved, as it fixes the issue. Some thoughts:

1. I see it was my fix for following symlinks for `tedge flows` that was not entirely correct as it created a difference from the behaviour of commands like `tedge flows test` and how flows were used in the mapper. It's not an excuse, but I think similar issues may happen in the future as long as `tedge flows` CLI and flows processing in the mapper use different codepaths to initialize and start the flows processing. Ideally we would have only one way of loading flows from `flows_dir`, and it would be used for both startup and live reload. It would have access to currently loaded flows and would detect new flows to load. The startup would then just be a regular case where no flows are currently running and we load all of them.
2. We could be more clear in [_Installing Flows_ section in the docs](https://thin-edge.github.io/thin-edge.io/extend/flows/#installing-flows) about how symlinks are handled (currently the word "symlink" doesn't appear anywhere in the article)
2. I'm not sure if we're not being too restrictive in how we're dealing with symlinks. Currently, as we canonicalize both `flows_dir` and paths of individual flows, we're requiring that all all flow files must be located in `flows_dir`, and we're using flow file's _canonical_ path to generate it's name. So if there's a symlink in `flows_dir` that points to a flow file in some different directory, it wouldn't work. But some users could perhaps like the flexibility of using a symlink in `/etc/tedge/mappers/MAPPER/flows`, which is owned by e.g. root, and point it to a file which could be owned by a less privileged user.
Instead of requiring that the flow file must be placed in `flows_dir`, we could require that the flow file is just _reachable_ from `flows_dir`, and the regular files could be somewhere else. We would then use the path inside `flows_dir` (which could not be fully resolved!) to generate the name of the flow.
There could be other difficulties though, but in any case I think it would be good to discuss because currently how symlinks are handled seems a bit unspecified.

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: > Maybe it is documented somewhere

No, this is not documented. Adding a few sentences is a good idea, either in the reference doc or in the [installing flows section](https://thin-edge.github.io/thin-edge.io/extend/flows/#installing-flows) as you suggested. 

> Maybe the message should be more direct and say exactly how the path is derived

This is not the purpose of an error message to tell all the subtleties.

didier-wenzek:
> Approved, as it fixes the issue. Some thoughts:
> 
>     1. I see it was my fix for following symlinks for `tedge flows` that was not entirely correct as it created a difference from the behaviour of commands like `tedge flows test` and how flows were used in the mapper. It's not an excuse, but I think similar issues may happen in the future as long as `tedge flows` CLI and flows processing in the mapper use different codepaths to initialize and start the flows processing. Ideally we would have only one way of loading flows from `flows_dir`, and it would be used for both startup and live reload. It would have access to currently loaded flows and would detect new flows to load. The startup would then just be a regular case where no flows are currently running and we load all of them.

This is why I moved the canonicalization deeper in `FlowStore::new()`.

>     2. We could be more clear in [_Installing Flows_ section in the docs](https://thin-edge.github.io/thin-edge.io/extend/flows/#installing-flows) about how symlinks are handled (currently the word "symlink" doesn't appear anywhere in the article)

Agree. I will add a sentence there.

>     3. I'm not sure if we're not being too restrictive in how we're dealing with symlinks. Currently, as we canonicalize both `flows_dir` and paths of individual flows, we're requiring that all all flow files must be located in `flows_dir`, and we're using flow file's _canonical_ path to generate it's name. So if there's a symlink in `flows_dir` that points to a flow file in some different directory, it wouldn't work. But some users could perhaps like the flexibility of using a symlink in `/etc/tedge/mappers/MAPPER/flows`, which is owned by e.g. root, and point it to a file which could be owned by a less privileged user.
>        Instead of requiring that the flow file must be placed in `flows_dir`, we could require that the flow file is just _reachable_ from `flows_dir`, and the regular files could be somewhere else. We would then use the path inside `flows_dir` (which could not be fully resolved!) to generate the name of the flow.
>        There could be other difficulties though, but in any case I think it would be good to discuss because currently how symlinks are handled seems a bit unspecified.

What you describe is actually working as long as we don't try to derive the name of a flow from its file path. We added this logic to ease flow packaging & installation. And I think this is not too bad. Sure, you cannot install a flow any where but this should not be a restriction in practice. This even sounds as a good practice to keep all the extension points under the same directory.

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: Added some docs: https://github.com/thin-edge/thin-edge.io/pull/4155/changes/cc1762140f07211b7cd798e0e2833582a1b29caa

On purpose, I said nothing about symlinks, because using symlinks should work as long as the flow file is under the mapper flow directory.

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -163,6 +163,9 @@ pub enum ConfigError {
 /// assert_eq!(derive_flow_name("/flows".into(), "/flows/hello/params.toml".into()), None);
 /// assert_eq!(derive_flow_name("/flows".into(), "/flows/hello/world.js".into()), None);
 /// assert_eq!(derive_flow_name("/flows".into(), "/unrelated/flows/hello.toml".into()), None);
+/// // Not recommanded by working:
```
Bravo555: typo?
```suggestion
/// // Not recommended but working:
```

rina23q (approved):

crates/extensions/tedge_flows/src/config.rs
didier-wenzek: fixed

Issue #3837, opened 2025-10-28
Add an option to `tedge connect` to prevent listener creation


rina23q:
**Is your feature improvement request related to a problem? Please describe.**
When users have an already configured local MQTT broker, the listener settings created by `tedge connect` can conflict with the existing broker configuration.

For example, settings like `mqtt.bind.address` and `mqtt.bind.port` always have default values.
```
mqtt.bind.address=127.0.0.1
mqtt.bind.port=1883
```

Then, `tedge connect` always creates this mosquitto configuration file. Currently, there is **no way** to prevent this file's creation.

/etc/tedge/mosquitto-conf/tedge-mosquitto.conf
```text
per_listener_settings true
connection_messages true
log_type error
log_type warning
log_type notice
log_type information
log_type subscribe
log_type unsubscribe
message_size_limit 268435455
listener 1883 127.0.0.1
allow_anonymous true
require_certificate false
```

If a user has already a configured broker, this forced common configuration file might create issues such as:
* Listener collision (same IP and port)
* Creating a new listener and opening a port with unwanted security settings
* Unwanted mosquitto log settings

**Describe the solution you'd like**
Introduce a new option to `tedge connect` that prevents the creation of the `tedge-mosquitto.conf` file.

**Describe alternatives you've considered**

**Additional context**
This problem was discovered while working on #3823. In that scenario, users were using already configured listener with username and password authentication on the local broker. The mosquitto configuration file created by `tedge connect` forces the creation of a new listener that lacks username/password authentication. This is unwanted behavior.

Issue #3599, opened 2025-05-08
add mosquitto configuration under the mosquitto packages conf.d folder


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

Currently in tedge <= 1.5.0, the mosquitto configuration files are stored under `/etc/tedge/mosquitto-conf/`, and an `include_dir <path>` directive is placed in the mosquitto.conf file like so:

**file: /etc/mosquitto/mosquitto.conf**

```
include_dir /etc/tedge/mosquitto-conf
```

The above relies on mosquitto having access to a path outside of the default `/etc/mosquitto` path which presents a problem when AppArmor is being used, as it will block any attempt by mosquitto to read the path. mosquitto will just log a "permission denied" error and then exit. Luckily, AppArmor does add an audit log entry to journald so that a user can see the reason for the permission denied (but if they don't look at the journald log, then the error can be very hard to spot as the standard linux permissions look ok).

Below shows an example of the log entry printed by AppArmor to journald:

```log
May 07 15:12:44 32df1eef46c8 kernel: audit: type=1400 audit(1746630764.792:686): apparmor="DENIED" operation="open" class="file" profile="mosquitto" name="/etc/tedge/mosquitto-conf/" pid=419394 comm="mosquitto
```

In recent versions of Ubuntu, e.g. Ubuntu 25.04, its seems that the AppArmor policies are being more strictly enforced, and applications aren't allowed to even read files/folders outside of those defined in the AppArmor config (e.g. `/etc/apparmor.d/`). For an example of the AppArmor definition for mosquitto (`mosquitto 2.0.20-1~bpo12+1`) under Ubuntu 25.04, check out the "Additional Context" section.

**Describe the solution you'd like**

**Pre-requisite**

* Switch to using built-in bridge by default, otherwise the device's certificates will also have to be moved under `/etc/mosquitto/` as they are currently located under `/etc/tedge/device-certs/`, which would suffer from the same AppArmor policy problem when using the mosquitto bridge to connect to the cloud

**Changes**

* Place mosquitto configuration files under `/etc/mosquitto/conf.d` instead of `/etc/tedge/mosquitto-conf`

Since the `/etc/tedge/mosquitto-conf` folder has been used for a long time under thin-edge.io, the change will also have to take care of migrating to the new folder structure. This generally would include:

[...]

Issue #3459, opened 2025-03-10
built-in bridge should generate separate logs for debugging connection issues


jarhodes314:
**Is your feature improvement request related to a problem? Please describe.**
Debugging a Cumulocity core issue with thin-edge recently has led me to see how useful the existing mosquitto logs are for explaining what is going on with MQTT communication (with the correct settings in mosquitto, `cat /var/log/mosquitto/mosquitto.log | grep Cumulocity` will show all the packets sent to/from the bridge). In the implementation of the built-in bridge, we've generally avoided logging in too much detail since that can pollute the mapper logs which contain a lot of other detail. I initially developed the bridge with concerns around debugging thin-edge.io specifically, but the bridge should also enable us to debug communication problems arising from the cloud too.

**Describe the solution you'd like**
I think the built-in bridge could perform detailed logging in a separate file. It would make sense to give the user some control over this (/the ability to disable it entirely since it could occupy rather a lot of disk space). There should probably be some basic logs in the main mapper logs, like what we have now which gives a broad indication of bridge health/backoffs in reconnecting, but doesn't give packet-by-packet descriptions of what's going on unless `debug` level logging is enabled.

**Describe alternatives you've considered**

**Additional context**

PR #4157, opened 2026-05-07
chore: improve MQTT log messages

rina23q:
## Proposed changes
With INFO level, `tedge-mapper-c8y` does not record any message conversions (e.g. when receiving an operation, converting to `init` cmd message, when receiving a final state cmd message, converting c8y operation update message and clearing the local cmd message).

This PR adds INFO level messages to improve the debuggabilty.

This is an example log of `tedge-mapper-c8y` when a `c8y_Command` operation is created.

```
2026-05-12T17:36:30.642908384Z  INFO C8Y: Forwarding c8y operation to device topic=te/device/main///cmd/shell_execute/c8y-mapper-448226 payload={"c8y-mapper":{"on_fragment":"c8y_Command","output":"${.payload.result}"},"command":"ls","status":"init"}
2026-05-12T17:36:30.726603327Z  INFO C8Y: Notifying c8y: operation set to EXECUTING topic=te/device/main///cmd/shell_execute/c8y-mapper-448226
2026-05-12T17:36:30.767538398Z  INFO C8Y: Notifying c8y: operation completed topic=te/device/main///cmd/shell_execute/c8y-mapper-448226
2026-05-12T17:36:30.767622146Z  INFO C8Y: Clearing command topic topic=te/device/main///cmd/shell_execute/c8y-mapper-448226
```

Before this PR, nothing was recorded.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/operations/convert.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/convert.rs
@@ -28,6 +28,7 @@ use tedge_api::workflow::StateExcerpt;
 use tedge_api::Jsonify;
 use tedge_mqtt_ext::MqttMessage;
 use tracing::error;
+use tracing::info;
 use tracing::warn;
 
 use crate::converter::CumulocityConverter;
@@ -103,9 +104,9 @@ impl CumulocityConverter {
         };
 
         // Command messages must be retained
-        Ok(vec![
-            MqttMessage::new(&topic, request.to_json()).with_retain()
-        ])
+        let payload = request.to_json();
+        info!(target: "C8Y", topic = %topic.name, payload = %payload, "Forwarding c8y operation to device");
+        Ok(vec![MqttMessage::new(&topic, payload).with_retain()])
     }
 
     /// Convert c8y_LogfileRequest operation to a ThinEdge log_upload command
@@ -145,9 +146,9 @@ impl CumulocityConverter {
         };
 
         // Command messages must be retained
-        Ok(vec![
-            MqttMessage::new(&topic, request.to_json()).with_retain()
-        ])
+        let payload = request.to_json();
+        info!(target: "C8Y", topic = %topic.name, payload = %payload, "Forwarding c8y operation to device");
+        Ok(vec![MqttMessage::new(&topic, payload).with_retain()])
     }
 
     /// Converts a log_upload metadata message to
@@ -222,9 +223,9 @@ impl CumulocityConverter {
         };
 
         // Command messages must be retained
-        Ok(vec![
-            MqttMessage::new(&topic, request.to_json()).with_retain()
-        ])
+        let payload = request.to_json();
+        info!(target: "C8Y", topic = %topic.name, payload = %payload, "Forwarding c8y operation to device");
+        Ok(vec![MqttMessage::new(&topic, payload).with_retain()])
     }
 
     pub async fn register_firmware_update_operation(
@@ -295,7 +296,9 @@ impl CumulocityConverter {
         };
 
         // Command messages must be retained
-        vec![MqttMessage::new(&topic, request.to_json()).with_retain()]
+        let payload = request.to_json();
+        info!(target: "C8Y", topic = %topic.name, payload = %payload, "Forwarding c8y operation to device");
+        vec![MqttMessage::new(&topic, payload).with_retain()]
     }
 
     async fn convert_config_metadata(
@@ -396,9 +399,9 @@ impl CumulocityConverter {
         }
 
         // Command messages must be retained
-        Ok(vec![
-            MqttMessage::new(&topic, request.to_json()).with_retain()
-        ])
+        let payload = request.to_json();
+        info!(target: "C8Y", topic = %topic.name, payload = %payload, "Forwarding c8y operation to device");
+        Ok(vec![MqttMessage::new(&topic, payload).with_retain()])
     }
 
     /// Converts a device_profile metadata message to supported operation "c8y_DeviceProfile"
@@ -475,6 +478,7 @@ impl CumulocityConverter {
         let request = GenericCommandState::new(topic, CommandStatus::Init.to_string(), payload)
             .update_with_json(inject_object);
 
+        info!(target: "C8Y", topic = %request.topic, payload = %request.payload, "Forwarding c8y operation to device");
         Ok(vec![request.into_message()])
     }
 }
--- a/crates/extensions/c8y_mapper_ext/src/operations/handler.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/handler.rs
@@ -18,8 +18,8 @@ use tedge_api::mqtt_topics::ChannelFilter;
 use tedge_api::mqtt_topics::EntityFilter;
 use tedge_api::workflow::GenericCommandState;
 use tedge_mqtt_ext::MqttMessage;
-use tracing::debug;
 use tracing::error;
+use tracing::info;
 use tracing::warn;
 
 /// Handles operations.
@@ -152,7 +152,7 @@ impl OperationHandler {
         match current_operation {
             Entry::Vacant(entry) => {
                 let Some(status) = status else {
-                    debug!(topic = %entry.key(), "unexpected clearing message");
+                    info!(target: "C8Y", topic = %entry.key(), "unexpected clearing message");
                     return;
                 };
 
@@ -167,9 +167,8 @@ impl OperationHandler {
             Entry::Occupied(entry) => {
                 let previous_status = entry.get().status.as_str();
                 if status.as_ref().is_some_and(|s| *s == previous_status) {
-                    debug!(
-                        "already handling operation message with this topic and status, ignoring"
-                    );
+                    info!(target: "C8Y", topic = %entry.key(), status = %previous_status,
+                        "already handling operation message with this topic and status, ignoring");
                     return;
                 }
 
--- a/crates/extensions/c8y_mapper_ext/src/operations/handlers/mod.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/handlers/mod.rs
@@ -50,6 +50,7 @@ use tedge_mqtt_ext::QoS;
 use tedge_mqtt_ext::Topic;
 use tracing::debug;
 use tracing::error;
+use tracing::info;
 
 /// State required by the operation handlers.
 pub(super) struct OperationContext {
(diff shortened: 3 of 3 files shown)


crates/extensions/tedge_mqtt_ext/src/lib.rs
```diff
@@ -472,7 +472,7 @@ impl FromPeers {
         while let Ok(Some(message)) = self.try_recv(rx_to_peers).await {
             match message {
                 MqttRequest::Publish(message) => {
-                    tracing::debug!(target: "MQTT pub", "{message}");
+                    tracing::info!(target: "MQTT pub", "{message}");
```
rina23q: Note: `MQTT sub:` is actually already INFO level, so I believe it makes it consistent to change `MQTT pub:` log message from DEBUG to INFO.

didier-wenzek: I see that as terribly noisy. All the messages sent by a mapper are then logged by default.

rina23q: Yes, actually this change affects also `tedge-agent`.
But this information is necessary to debug. ~~With built-in bridge, we cannot track when and who publishes messages. This was possible with mosquitto bridge with debug level.~~

~~If `tedge-mapper` and `tedge-agent` are not appropriate places to log them, then we should consider https://github.com/thin-edge/thin-edge.io/issues/3459.~~

=> I did misunderstand here. It should be able to see them in mosquitto.log as long as there is `log_type debug`. And, the customer's mosquitto config settings doesn't have this, that's why it was not visible... so I will revert this part.

rina23q: I found that `/etc/tedge/mosquitto-conf/tedge-mosquitto.conf` does not have `log_type debug`. And we recommend to add this in [doc](https://thin-edge.github.io/thin-edge.io/operate/troubleshooting/log-files/#mosquitto). Can't we add  `log_type debug` to `tedge-mosquitto.conf` by default?

didier-wenzek: > I found that `/etc/tedge/mosquitto-conf/tedge-mosquitto.conf` does not have `log_type debug`. And we recommend to add this in [doc](https://thin-edge.github.io/thin-edge.io/operate/troubleshooting/log-files/#mosquitto). Can't we add `log_type debug` to `tedge-mosquitto.conf` by default?

I understand your need but I also think we must stop hacking `mosquitto` configuration with `tedge-mosquitto.conf`. This is a source of confusion and problems (see https://github.com/thin-edge/thin-edge.io/issues/3837 or https://github.com/thin-edge/thin-edge.io/issues/3599).

- Mosquitto log configuration have to be put apart from listener and authentication configuration
- On install we can write a `tedge-recommendation.conf`  under `/etc/mosquitto/conf.d` but being aware that users are free to edit these settings.

crates/extensions/c8y_mapper_ext/src/operations/handlers/mod.rs
```diff
@@ -188,6 +190,7 @@ impl OperationContext {
                 }
             }
             OperationOutcome::Finished { messages } => {
+                info!(topic = %message.topic.name, "Notifying c8y: operation complete: ");
```
Bravo555: suggestion: remove trailing `:` in all added logs

Indeed the current formatter prints the fields after the message, so I see the temptation of adding `:` after the message, but IMO I wouldn't add `:` suffix to the message just to control how it's printed: when structured logging `:` ends up as part of the message field, the formatting can be inconsistent (extra space like here or not), and I don't think it's really necessary to include `:` for presentation - when logging to the terminal fields are _italicised_, but I can understand how when viewing in `journalctl` one would like to include it. If so, we could potentially solve it by [tweaking the `FormatEvent` implementation](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/trait.FormatEvent.html) so it would be solved for all events, not just a few.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
```diff
@@ -152,7 +152,7 @@ impl OperationHandler {
         match current_operation {
             Entry::Vacant(entry) => {
                 let Some(status) = status else {
-                    debug!(topic = %entry.key(), "unexpected clearing message");
+                    info!(topic = %entry.key(), "unexpected clearing message");
```
Bravo555: thought: I understand this is changed because this information is useful for debugging so we'd like it to be captured in that case, even though when viewing the logs interactively, maybe we're not interested in such details as when topics are cleared.

So we change it to `INFO` because we log `INFO` by default, even though one could use an env filter like `RUST_LOG=info,c8y_mapper_ext::operations=debug` to show these operation events, but one would have to set that beforehand.

IMO the ideal solution for such problems is to have different event streams for different targets and direct them to different sinks, such that we don't have to make an event appear to a user, just so it can be captured for debugging by us. Events from different parts of the system already are attached with a `target` attribute, we could use tracing-subscriber's [per-layer-filtering](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/layer/index.html#per-layer-filtering) to e.g. print the logs to the journal with one filter in one layer, and log events as JSON to file with a different filter in another layer. We would just have to systematize the currently used event `target`s and define the proper filters, in which we could be more flexible than just doing it by level.

Bravo555 (approved):
I see the changes as not too big and changes to MQTT logging which would make it too noisy were reverted, so while I'd like to see trailing `:` suggestion applied for consistency, it's not necessary so approving.

This PR reminds me that [log levels kinda suck because of unclear semantics](https://mmapped.blog/posts/44-tag-based-logging) and how we could potentially improve it: https://github.com/thin-edge/thin-edge.io/pull/4157#discussion_r3219419137

crates/extensions/c8y_mapper_ext/src/operations/convert.rs
```diff
@@ -103,6 +104,7 @@ impl CumulocityConverter {
         };
 
         // Command messages must be retained
+        info!(topic = %topic.name, "Forwarding c8y operation to device:");
```
albinsuresh: Why not log the payload as well?

rina23q: Because I thought that payloads are sometimes too huge (e.g. `software_list`), but here since this is `init` state, it cannot be as huge as I imaged with software_list response. 
=> I will add payloads here.

crates/extensions/c8y_mapper_ext/src/operations/handler.rs
rina23q: This message is something unexpected by tedge-mapper, so I think user should be aware of it (and fix it potentially).

But in general, as you pointed out, there is definitely a dilemma. It's a common practice, running with INFO level in the production level. That's why INFO is the default level. Then, when a problem occurs in the user side, they provide us their logs. We have to debug them, but since it's INFO level, useful logs are omitted.

crates/extensions/c8y_mapper_ext/src/operations/handlers/mod.rs
rina23q: Removed trailing `:` in dd8c1dcf7e39c509be283f49d096fe5bbb677a4c

crates/extensions/c8y_mapper_ext/src/operations/convert.rs
rina23q: Added payload for the incoming request in dd8c1dcf7e39c509be283f49d096fe5bbb677a4c.

FYI: I didn't add payloads to the logs of outgoing messages with the same reason above, to avoid too huge logs. Since the topic contains the operation ID, users should be able to what was the result of the operation in their c8y tenant.

didier-wenzek (approved):
Approved.

A suggestion is to add a `target = "C8Y"` to all logging in the `c8y_mapper_ext` crate:

```
2026-05-07T15:44:40.760225043Z  INFO C8Y: Forwarding c8y operation to device: topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
2026-05-07T15:44:40.840647119Z  INFO C8Y: Notifying c8y: operation set to EXECUTING: topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
2026-05-07T15:44:40.898653098Z  INFO C8Y: Notifying c8y: operation complete:  topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
2026-05-07T15:44:40.898747387Z  INFO C8Y: Clearing command topic: topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
```

instead of 

```
2026-05-07T15:44:40.760225043Z  INFO c8y_mapper_ext::operations::convert: Forwarding c8y operation to device: topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
2026-05-07T15:44:40.840647119Z  INFO c8y_mapper_ext::operations::handlers: Notifying c8y: operation set to EXECUTING: topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
2026-05-07T15:44:40.898653098Z  INFO c8y_mapper_ext::operations::handlers: Notifying c8y: operation complete:  topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
2026-05-07T15:44:40.898747387Z  INFO c8y_mapper_ext::operations::handlers: Clearing command topic: topic=te/device/main///cmd/shell_execute/c8y-mapper-443224
```

rina23q:
> Approved.
> 
> A suggestion is to add a `target = "C8Y"` to all logging in the `c8y_mapper_ext` crate:

Good idea. Applied in f37f3fb72ba8a1d8564582efe5157d95f8fba663.

crates/extensions/c8y_mapper_ext/src/operations/convert.rs
```diff
@@ -104,7 +104,7 @@ impl CumulocityConverter {
         };
 
         // Command messages must be retained
-        info!(topic = %topic.name, "Forwarding c8y operation to device:");
+        info!(topic = %topic.name, payload = %request.to_json(), "Forwarding c8y operation to device");
         Ok(vec![
             MqttMessage::new(&topic, request.to_json()).with_retain()
```
albinsuresh: Minor: We could avoid the re-computation of the payload with `request.to_json()` here, as it is already computed once in the previous step for the logging. The same comment applies to other places where the same logging pattern is used.

rina23q: Done [f997b17](https://github.com/thin-edge/thin-edge.io/pull/4157/commits/f997b17d9ab455a2d71fd53c6d75110895b38143)

Issue #4166, opened 2026-05-11
tedge flows running on a short interval can starve other flows


jarhodes314:
**Describe the bug**
If a flow is set to run on an interval shorter than the time it takes to execute, it executes constantly in a loop, starving other flows which respond to messages from executing. I discovered this when publishing measurements. If I set the polling interval for the measurements really short, the measurements do not get forwarded to Cumulocity.

I believe this is the result of `next_interval_deadline` being in the past when we call `FlowsMapper::next_message`. This means the `sleep_until` future returns without polling, and gets reliably polled first by `futures::future::sleep`. As a result, `next_message` always returns tick, and the flows mapper never processes incoming messages.

**To Reproduce**

**Expected behavior**
Setting the interval too short might have the potential to cause performance issues, but we should always ensure messages get processed in a reasonably timely manner. It shouldn't be possible for customer-written flows to completely break built-in thin-edge behaviour.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]
 - Hardware [incl. revision]
 - System-Architecture [e.g. result of "uname -a"]
 - thin-edge.io version [e.g. 0.1.0]

**Additional context**

PR #4168, opened 2026-05-12
fix: Prevent flow starvation from repeated interval scheduling

jarhodes314:
## Proposed changes
Prevent misconfigured flows from starving other flows. I initially discovered this when running a metrics collection flow every 100ms to speed up some integration tests and found that if the interval the flow runs on is smaller than the time taken for the flow to execute, `tedge_flows` only runs the interval handlers, it doesn't call any onMessage functions.

This PR makes two changes to prevent this from causing problems in the future:
1. The implicitly biased `futures::future::select` call that was causing the starvation issue at the root of the problem has been replaced with an unbiased `tokio::select`, with special handling to ensure fairness in cases where we already have a backlog of messages and interval deadlines.
2. Interval handling has been modified so that if an interval action takes too long, we avoid rescheduling it for the past, as this doesn't seem like useful behaviour. Care has been taken to ensure when we do execute intervals, we do our best to adhere to the interval, so flows that take time to execute don't disrupt the schedule (unless the schedule is totally unrealistic).

## Types of changes

## Paste Link to the issue
- #4166

## Checklist

### DIFF
--- a/crates/extensions/tedge_flows/src/actor.rs
+++ b/crates/extensions/tedge_flows/src/actor.rs
@@ -14,7 +14,6 @@ use crate::Tick;
 use async_trait::async_trait;
 use camino::Utf8Path;
 use camino::Utf8PathBuf;
-use futures::FutureExt;
 use serde_json::json;
 use std::cmp::min;
 use std::collections::HashSet;
@@ -50,6 +49,7 @@ pub struct FlowsMapper {
     watched_commands: HashSet<Utf8PathBuf>,
     processor: MessageProcessor<ConnectedFlowRegistry>,
     next_dump: Instant,
+    deferred_tick: bool,
 }
 
 impl FlowsMapper {
@@ -72,6 +72,7 @@ impl FlowsMapper {
             watched_commands,
             processor,
             next_dump,
+            deferred_tick: false,
         }
     }
 }
@@ -121,16 +122,14 @@ impl FlowsMapper {
             .processor
             .next_interval_deadline()
             .map_or(self.next_dump, |deadline| min(deadline, self.next_dump));
-        let deadline_future = sleep_until(deadline).map(|_| Some(InputMessage::Tick(Tick)));
-        let incoming_message_future = self.messages.recv();
 
-        futures::pin_mut!(incoming_message_future);
-        futures::pin_mut!(deadline_future);
-
-        futures::future::select(deadline_future, incoming_message_future)
-            .await
-            .factor_first()
-            .0
+        tokio::select! {
+            message = self.messages.recv() => {
+                self.deferred_tick = false;
+                message
+            }
+            _ = sleep_until(deadline) => Some(InputMessage::Tick(Tick)),
+        }
     }
 
     async fn send_updated_subscriptions(&mut self) -> Result<(), RuntimeError> {
--- a/crates/extensions/tedge_flows/src/lib.rs
+++ b/crates/extensions/tedge_flows/src/lib.rs
@@ -58,6 +58,7 @@ use tedge_mqtt_ext::TopicFilter;
 use tedge_watch_ext::WatchActorBuilder;
 use tedge_watch_ext::WatchEvent;
 use tedge_watch_ext::WatchRequest;
+use tokio::time::Instant;
 pub use transformers::Transformer;
 
 pub struct FlowsMapperConfig {
@@ -131,6 +132,22 @@ pub fn flows_dir(mapper_dir: &Utf8Path) -> Utf8PathBuf {
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 struct Tick;
+
+pub(crate) fn next_deadline_after(deadline: Instant, interval: Duration, now: Instant) -> Instant {
+    debug_assert!(!interval.is_zero());
+    if deadline > now {
+        return deadline;
+    }
+
+    let missed_intervals = now.duration_since(deadline).as_nanos() / interval.as_nanos() + 1;
+
+    match u32::try_from(missed_intervals) {
+        // Keep the original cadence, but skip any deadlines that have already passed.
+        Ok(missed_intervals) => deadline + interval * missed_intervals,
+        // If we're this far (~4.2million invocations) behind, just restart the schedule from now.
+        Err(_) => now + interval,
+    }
+}
 
 pub struct FlowsMapperBuilder {
     config: FlowsMapperConfig,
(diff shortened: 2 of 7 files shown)


crates/extensions/tedge_flows/src/lib.rs
```diff
@@ -132,6 +133,22 @@ pub fn flows_dir(mapper_dir: &Utf8Path) -> Utf8PathBuf {
...
+    if deadline > now {
+        return deadline;
+    }
+
+    let missed_intervals = now.duration_since(deadline).as_nanos() / interval.as_nanos() + 1;
+
+    match u32::try_from(missed_intervals) {
+        // Keep the original cadence, but skip any deadlines that have already passed.
+        Ok(missed_intervals) => deadline + interval * missed_intervals,
+        // If we're this far (~4.2million invocations) behind, just restart the schedule from now.
```
didier-wenzek, line 147: :-)

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -121,16 +123,33 @@ impl FlowsMapper {
...
-            .factor_first()
-            .0
+            match self.messages.try_recv().now_or_never() {
+                Some(Ok(Some(message))) => {
+                    self.deferred_tick = true;
+                    return Some(message);
+                }
+                Some(Ok(None)) | None => {}
+                Some(Err(_)) => return None,
+            }
```
didier-wenzek: I don't get why this code is required. I though it was to avoid small intervals starving message processing, but I'm puzzled by the condition on `deadline <= Instant::now() && ! self.deferred_tick`.


Actually, the tests are still green when removed. That said, I fully appreciate that it's not easy to test the two starvation cases.

jarhodes314: It's intended to force us to alternate between processing messages and intervals to guarantee fairness, but it's probably overkill and doesn't practically anything that the random polling order in `tokio::select` achieves

didier-wenzek: Okay. Let's go with that.

didier-wenzek (approved):
Approved. Thank you for spotting and fixing that subtle issue.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -121,16 +123,33 @@ impl FlowsMapper {
...
-        futures::pin_mut!(incoming_message_future);
-        futures::pin_mut!(deadline_future);
+        // A tick and a message can be ready at the same time. If that happens
+        // repeatedly, always picking the same side can keep the other from running.
+        // Let one queued message go first, then force the tick on the next loop.
+        if deadline <= Instant::now() {
+            if self.deferred_tick {
+                self.deferred_tick = false;
+                return Some(InputMessage::Tick(Tick));
+            }
```
Bravo555: suggestion: It would be nice to `warn!` the user if the ticks start falling behind

crates/extensions/tedge_flows/src/actor.rs
Bravo555: In fact, even after `git restore --source main crates/extensions/tedge_flows/src/` the new test seems to pass :/ We would probably need heavier guns like proptest to exhaustively test that scenario.

crates/extensions/tedge_flows/src/actor.rs
```diff
@@ -121,16 +123,33 @@ impl FlowsMapper {
...
+        }
+
+        tokio::select! {
+            message = self.messages.recv() => {
+                self.deferred_tick = false;
+                message
+            }
+            _ = sleep_until(deadline) => Some(InputMessage::Tick(Tick)),
+        }
```
Bravo555, line 132: thought: This _probably_ solves the issue of intervals (tbh I wasn't able to reproduce it yet), but I'm wondering if we could still be suffering from some bad flows starving other flows.

If I understand correctly, a flow could still potentially starve other flows if its e.g. `onMessage` handler took really long? Because in `self.on_message` we're calling the flow's `on_message` handler synchronously, wait until it returns a message, and then publish it? In effect, until one flow runs its handler, we can't route messages to other flows. Also because this all happens on one task, we only use 1 thread to run flows, which, true, should be sufficient for most scenarios, but if someone has a really big and complex flow, it can still slow down others. JS processing AFAIK runs on a separate thread (not sure if it's per-flow), but still we only run 1 thread at a time and wait for `on_message` handler to complete, so in effect it also runs synchronously.

I think that in the main processing loop of `FlowsMapper` we should try to take as little time to possible to process a message and consider running flows as separate tasks to minimize the chances that flows impact other flows. It would still be unavoidable of course if one runs so many flows it saturates the CPU, but the tokio runtime should schedule the flows fairly.

There would also be disadvantages in that it would complicate our current simple synchronous model of: just run `on_message`, which will run `on_message` of a single flow, which returns a `Message` as a regular value that you then publish, so we may not want to do that. Still, running flows really in parallel seems like somewhat low hanging fruit to me.

Bravo555 (approved):

crates/extensions/tedge_flows/src/actor.rs
jarhodes314: > In fact, even after `git restore --source main crates/extensions/tedge_flows/src/` the new test seems to pass :/ We would probably need heavier guns like proptest to exhaustively test that scenario.

I see it fail reliably with that change (in both `cargo test` and `cargo nextest`):

```
     Running tests/interval.rs (target/debug/deps/interval-146221fc3b50f1cc)

running 10 tests
test script_with_oninterval_but_no_config_gets_default_1s_interval ... ok
test file_polling_interval_is_respected_on_errors ... ok
test multiple_scripts_execute_at_independent_frequencies ... ok
test interval_executes_at_configured_frequency ... ok
test very_short_intervals_execute_correctly ... ok
test sources_poll_at_configured_frequency ... ok
test interval_executes_independently_from_message_processing ... ok
test interval_executes_when_time_exceeds_interval ... ok
test onstartup_before_onmessage_in_two_step_flow ... ok
test process_polling_flows_do_not_starve_on_message_flows ... FAILED

failures:

---- process_polling_flows_do_not_starve_on_message_flows stdout ----

thread 'process_polling_flows_do_not_starve_on_message_flows' (17364) panicked at crates/extensions/tedge_flows/tests/interval.rs:179:5:
Message flow should process MQTT input published by the process-polling flow; got 17 mongo_disk outputs but 0 processed messages
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    process_polling_flows_do_not_starve_on_message_flows

test result: FAILED. 9 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.52s
```

crates/extensions/tedge_flows/src/actor.rs
jarhodes314: I think we also run the risk of just moving the problem by spawning tasks - if I tried to solve the interval problem by spawning tasks, we would just end up having an explosion in the number of tasks running at once. The problem of "how do we prevent anything any users can do with flows preventing the execution of other flows" is a very different problem to the one I am targeting here, which is more along the lines "how do we prevent the configuration of a flow from preventing other flows from executing".

Running flows in parallel/concurrently could come with advantages for sure, but I think it's a considerably larger undertaking than this fix, and not guaranteed to be something we can implement for sure.

didier-wenzek: > If I understand correctly, a flow could still potentially starve other flows if its e.g. onMessage handler took really long? 

An `onMessage` handler taking too much time is cancelled but the flow runtime, moving to the next flow having to process the same message or to the next message. So, sure, a slow handler will slow down all the flows but not prevent those to work.

crates/extensions/tedge_flows/src/actor.rs
jarhodes314: I've now removed this code as it doesn't really serve any particularly useful purpose (there's nothing I can concretely identify as a problem it will solve in itself that `tokio::select` doesn't).

I've also modified the test to include an explicit sleep in the test as this resolves the issue  @Bravo555 was seeing with the regression test still passing even with the fix applied (as well as more accurately emulating what was going on when I discovered the bug).

PR #4170, opened 2026-05-13
tests: Remove hardcoded skips of certain number of messages from c8y mapper Rust unit tests

Bravo555:
## Proposed changes

Having hardcoded "skip `N` messages" in tests makes it harder to change what messages are published, e.g. in #4064 where adding `114` message after registration makes a bunch of tests fail because now instead of skipping `N` messages they need to skip `N+1` messages.

This skipping was necessary as long as the tests were operating on regular `MessageReceiver`s, which only provide an interface to receive a new message and have no way to store already received messages. I'd argue that while it's a useful foundation for expressing different types of receivers which are used in production code, it's probably a bad abstraction for tests where we can easily store incoming messages to a `Vec` as they come and then run asserts not on one message that we just received, but on all messages received.

Thus, the PR provides a new `MockMqttBox` (alternative proposed names: `TestMqttBox`/`BufferedMqttBox`) which wraps any receiver and stores its received messages. The assertions then run on all captures messages.

The assertions then can be made more flexible, but currently they still need to be adjusted to some of the intuitive usage patterns. For example, after asserting message `m1` and then asserting message `m2`, the writer of the test probably expects that they asserted that `m2` happens after `m1`. So for simple and common asserts we could update a cursor pointing to the least recently asserted message and only assert additional messages after that. 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
+++ b/crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
@@ -138,6 +138,8 @@ impl OperationContext {
 mod tests {
     use super::*;
     use crate::config::C8yMapperConfig;
+    use crate::tests::helpers::assert_received_contains_str;
+    use crate::tests::helpers::assert_received_includes_json;
     use crate::tests::*;
     use c8y_api::json_c8y_deserializer::C8yDeviceControlTopic;
     use serde_json::json;
@@ -146,8 +148,6 @@ mod tests {
     use tedge_actors::MessageReceiver;
     use tedge_actors::Sender;
     use tedge_downloader_ext::DownloadResponse;
-    use tedge_mqtt_ext::test_helpers::assert_received_contains_str;
-    use tedge_mqtt_ext::test_helpers::assert_received_includes_json;
     use tedge_mqtt_ext::Topic;
     use tedge_test_utils::fs::TempTedgeDir;
     use tedge_uploader_ext::UploadResponse;
@@ -161,8 +161,6 @@ mod tests {
 
         let TestHandle { mqtt, .. } = test_handle;
         let mut mqtt = mqtt.with_timeout(TEST_TIMEOUT_MS);
-
-        skip_init_messages(&mut mqtt).await;
 
         // Simulate c8y_LogfileRequest JSON over MQTT request
         mqtt.send(MqttMessage::new(
@@ -211,8 +209,6 @@ mod tests {
         let TestHandle { mqtt, .. } = test_handle;
         let mut mqtt = mqtt.with_timeout(TEST_TIMEOUT_MS);
 
-        skip_init_messages(&mut mqtt).await;
-
         // Register the device upfront
         mqtt.send(MqttMessage::new(
             &Topic::new_unchecked("te/device/child1//"),
@@ -220,7 +216,6 @@ mod tests {
         ))
         .await
         .expect("Send failed");
-        mqtt.skip(1).await; // Skip the mapped registration message
 
         // Simulate log_upload cmd metadata message
         mqtt.send(MqttMessage::new(
@@ -229,8 +224,6 @@ mod tests {
         ))
         .await
         .expect("Send failed");
-
-        mqtt.skip(2).await; //Skip supported ops and supported log types messages
 
         // Simulate c8y_LogfileRequest JSON over MQTT request
         mqtt.send(MqttMessage::new(
@@ -278,8 +271,6 @@ mod tests {
         let TestHandle { mqtt, .. } = test_handle;
         let mut mqtt = mqtt.with_timeout(TEST_TIMEOUT_MS);
 
-        skip_init_messages(&mut mqtt).await;
-
         // Simulate log_upload cmd metadata message
         mqtt.send(MqttMessage::new(
             &Topic::new_unchecked("te/device/main///cmd/log_upload"),
@@ -311,8 +302,6 @@ mod tests {
         let TestHandle { mqtt, .. } = test_handle;
         let mut mqtt = mqtt.with_timeout(TEST_TIMEOUT_MS);
 
-        skip_init_messages(&mut mqtt).await;
-
         // Register the device upfront
         mqtt.send(MqttMessage::new(
             &Topic::new_unchecked("te/device/child1//"),
@@ -320,7 +309,6 @@ mod tests {
         ))
         .await
         .expect("Send failed");
-        mqtt.skip(1).await; // Skip the mapped registration message
 
         // Simulate log_upload cmd metadata message
         mqtt.send(MqttMessage::new(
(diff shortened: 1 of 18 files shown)


didier-wenzek (commented):
This is really nice proposal! I encourage you to finalize it.

didier-wenzek:
That commit https://github.com/thin-edge/thin-edge.io/pull/4170/changes/202b0a6b76757afecaf9939fcdf915bf40b55e85 seems to have been included by accident.

crates/extensions/c8y_mapper_ext/src/tests/mock_mqtt_box.rs
```diff
@@ -0,0 +1,165 @@
...
+        true
+    }
+
+    #[track_caller]
+    pub fn none<P>(&mut self, predicate: P) -> bool
+    where
+        P: Fn(&MqttMessage) -> bool,
+    {
+        let messages = self.messages.lock().unwrap();
+        messages.iter().position(predicate).is_none()
```
didier-wenzek: Suggested by clippy


```suggestion
       !messages.iter().any(predicate)
```

crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
```diff
@@ -146,8 +148,6 @@ mod tests {
     use tedge_actors::MessageReceiver;
     use tedge_actors::Sender;
     use tedge_downloader_ext::DownloadResponse;
-    use tedge_mqtt_ext::test_helpers::assert_received_contains_str;
-    use tedge_mqtt_ext::test_helpers::assert_received_includes_json;
```
didier-wenzek, line 150: Have you considered to fully deprecate these test helpers - or, better, to move your improved implementation there? What are the blockers, if any?

The former `tedge_mqtt_ext::test_helpers` are still used to test the agent and `c8y_mapper_ext::availability`. Since the purpose is the same it would be good to have a single abstraction.

crates/extensions/c8y_mapper_ext/src/tests.rs
```diff
@@ -1281,7 +1242,10 @@ async fn mapper_dynamically_updates_supported_operations_for_child_device() {
     ))
     .await
     .expect("Send failed");
-    mqtt.skip(1).await; // Skip the mapped registration message
+
+    // ignore all messages up to this point
+    mqtt.recv_short().await;
+    mqtt.messages.lock().unwrap().clear();
```
didier-wenzek: This is not clear. What is done? Why?

This might be a good candidate for a new test helper

```suggestion
    mqtt.ignore_messages_received_so_far();
```

crates/extensions/c8y_mapper_ext/src/tests/mock_mqtt_box.rs
Bravo555: fixed in f64978f729

crates/extensions/c8y_mapper_ext/src/tests.rs
Bravo555: fixed in f64978f729

Good catch, indeed that was something I replaced too quickly and didn't realize the test was doing too much - instead of ignoring all messages, we should only ignore any received `114` messages for the child device topic to not mistake it for a `114` message that should be produced when sending command metadata message later on in the test.

crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs
Bravo555: > Have you considered to fully deprecate these test helpers - or, better, to move your improved implementation there?

Indeed the new assertions could probably be moved in and replace the current ones in `tedge_mqtt_ext`, but there's still quite a few usages of the old asserts in the agent tests and I'd like to avoid having this PR becoming too big and replacing everything in one go. Asserts in the agent tests could be changed in a follow-up PR.

> The former tedge_mqtt_ext::test_helpers are still used to test the agent and c8y_mapper_ext::availability

Indeed `availability` module still uses old asserts and skips messages by number. Initially I thought to leave it as is, as the availability module tests don't test the entire mapper but only the availability actor, so it's possible that the availability messages won't change that much _and_ skipped messages are sent by the same actor thus they're not unrelated noise from other actors, so skipping them is meaningful, so i though that _maybe_ the current asserts can stay, but you changed my mind and I'll change those as well because:

- as you say, it's better not to duplicate abstractions, but build one good abstraction and use it everywhere
- the skips often have the form `mqtt.skip(1) // 117`, i.e. we're expecting the skipped message to be 117 but we're not checking that, which we probably should

So I'll replace these skips with the proper asserts or ignore method from f64978f72 as required.

didier-wenzek (approved):
Approved. Thank you. This is great improvement and refactoring c8y mapper tests will be less error prone.

crates/extensions/tedge_mqtt_ext/src/test_helpers/test_mqtt_box/asserts.rs
```diff
@@ -0,0 +1,109 @@
+use std::time::Duration;
```
albinsuresh, line 3: Why not move these test helpers to `tedge_mqtt_ext` module itself as suggested [here](https://github.com/thin-edge/thin-edge.io/pull/4170#discussion_r3335330781), so that these enhancements are available to agent and other components as well?

I wouldn't remove the existing helpers, but just add these to a separate module.

crates/extensions/c8y_mapper_ext/src/tests/mock_mqtt_box.rs
```diff
@@ -0,0 +1,174 @@
...
+/// An MQTT message box that stores received messages in a buffer and supports
+/// assertions for stored messages.
+///
+/// To enable maximum flexibility in test assertions, where most tests don't
+/// care about the all the messages, only ones they're directly testing, we want
+/// to allow these tests to ignore these messages without losing them (in case
+/// something else asserts them later). This is most easily done by gathering
+/// all received messages to a buffer and then choosing to drop selected
+/// messages only once they're asserted.
+pub struct MockMqttBox<Box> {
```
albinsuresh: Could we use a different parameter name than `Box` to disambiguate it from the `std::boxed::Box` type?

albinsuresh (approved):
Great enhancement. It was nice to see the seamless transition in all existing tests.

crates/extensions/tedge_mqtt_ext/src/test_helpers/test_mqtt_box/asserts.rs
Bravo555: Moved into `tedge_mqtt_ext::test_helpers::TestMqttBox` in 07069fc96d

I've been putting it off for some reason, but indeed it makes sense and would have needed to be done sooner or later.

crates/extensions/c8y_mapper_ext/src/tests/mock_mqtt_box.rs
Bravo555: fixed in 07069fc96d

Issue #4172, opened 2026-05-13
`entity_store.jsonl` grows indefinitely when retained messages are repeatedly published to the same topic


dtimotijevic:
**Describe the bug**
`entity_store.jsonl` appears to grow continuously when retained MQTT messages are published repeatedly to the same twin topic with different payloads.

We are aware that `entity_store.jsonl` may not be intended to be a strict MQTT broker retained-message database. But the practical consequence is that retained-message usage can cause this file to grow indefinitely. From our perspective, this makes the MQTT retained feature unusable for state-like messages that are updated frequently.

In our case, retained messages are published to the same topic, but with a changing payload containing a unique sequence number. Instead of the previous retained state effectively being replaced in the persisted store, new entries appear to be appended, causing `entity_store.jsonl` to grow significantly over time.

According to the MQTT v5 specification, retained messages are expected to represent the latest retained application message per topic:

> If the RETAIN flag is set to 1 in a PUBLISH packet sent by a Client to a Server, the Server MUST replace any existing retained message for this topic and store the Application Message

Even if thin-edge stores additional entity state outside the MQTT broker retained database, the current behavior has the same practical impact as retaining historical payloads for the same topic.

**To Reproduce**

```
tedge mqtt pub 'te/device/my-device///twin/example_retained_state' '{"test_value":"first"}' --retain
tedge mqtt pub 'te/device/my-device///twin/example_retained_state' '{"test_value":"second"}' --retain

grep 'example_retained_state' /etc/tedge/.agent/entity_store.jsonl
```

**Environment (please complete the following information):**
 - thin-edge.io version 1.7.1

didier-wenzek:
I confirm that currently `entity_store.jsonl` keeps growing without any cleanup and that is has to be fixed, recreating a clean content of this file after x twin related messages.

albinsuresh:
@dtimotijevic What's the frequency of these twin updates? The number of times per minute or per hour or per day?

dtimotijevic:
@albinsuresh @albinsuresh The updates are event based and can very different on system, but mostly they occur on a per-second/per-minute basis rather than hourly, which is why the file grows so rapidly.

Given this high frequency, a time-based cleanup / compaction isn't ideal since it assumes a predictable fill rate. A size-based compaction triggered by file size limits would be much more reliable.

Ideally this compaction mechanism should simply preserve the latest local state for each topic, without requiring a state re-sync from the cloud twin.

[...]

PR #4173, opened 2026-05-13
fix: Automatically compact `entity_store.jsonl` after a large number of repeat topics

jarhodes314:
## Proposed changes
Adds a compaction mechanism to the entity store file. This will trigger after we have written a significant number (currently 100) messages to topics that already exist in the file, preventing unbounded growth.

It also refactors the existing test for the readers/writers into well-defined unit tests.

## Types of changes

## Paste Link to the issue
- #4172

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/store/message_log.rs
+++ b/crates/core/tedge_api/src/store/message_log.rs
@@ -1,8 +1,10 @@
 //! The message log is a persistent append-only log of MQTT messages.
 //! Each line is the JSON representation of that MQTT message.
 //! The underlying file is a JSON lines file.
+use indexmap::IndexMap;
 use mqtt_channel::MqttMessage;
 use serde_json::json;
+use std::collections::HashSet;
 use std::fs::File;
 use std::fs::OpenOptions;
 use std::io::BufRead;
@@ -10,9 +12,12 @@ use std::io::BufReader;
 use std::io::BufWriter;
 use std::io::Write;
 use std::path::Path;
+use std::path::PathBuf;
 
 const LOG_FILE_NAME: &str = "entity_store.jsonl";
+const LOG_FILE_TEMP_NAME: &str = "entity_store.jsonl.tmp";
 const LOG_FORMAT_VERSION: &str = "1.0";
+const DEFAULT_REDUNDANCY_THRESHOLD: usize = 100;
 
 #[derive(thiserror::Error, Debug)]
 pub enum LogEntryError {
@@ -65,6 +70,10 @@ impl MessageLogReader {
 /// A writer to append new MQTT messages to the end of the log
 pub(crate) struct MessageLogWriter {
     writer: BufWriter<File>,
+    log_dir: PathBuf,
+    redundancy_threshold: usize,
+    unique_topics: HashSet<String>,
+    total_entries: usize,
 }
 
 impl MessageLogWriter {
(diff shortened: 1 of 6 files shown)


crates/core/tedge_api/src/store/message_log.rs
```diff
@@ -65,43 +69,85 @@ impl MessageLogReader {
...
+
+        // Read the existing file to build the topic index for accurate redundancy tracking.
+        let mut unique_topics = HashSet::new();
+        let mut total_entries = 0;
+        if let Ok(mut reader) = MessageLogReader::new(log_dir) {
+            while let Some(msg) = reader.next_message().map_err(std::io::Error::other)? {
+                unique_topics.insert(msg.topic.name.clone());
+                total_entries += 1;
+            }
+        }
```
albinsuresh: We could avoid this when `trunacte == true`

crates/core/tedge_api/src/store/message_log.rs
```diff
@@ -111,6 +157,66 @@ impl MessageLogWriter {
...
+            if redundant_count >= self.redundancy_threshold {
+                self.compact()?;
+            }
+        }
+
+        Ok(())
+    }
+
+    fn compact(&mut self) -> Result<(), std::io::Error> {
+        self.writer.flush()?;
```
albinsuresh: Was that really needed? I'm wondering why, since the `compact` is called from `append_message` after both `flush` and `sync_all` are done.

crates/core/tedge_api/src/store/message_log.rs
```diff
@@ -111,6 +157,66 @@ impl MessageLogWriter {
...
+
+        let mut all_messages = Vec::new();
+        let mut reader = MessageLogReader::new(&self.log_dir)?;
+        while let Some(msg) = reader.next_message().map_err(std::io::Error::other)? {
+            all_messages.push(msg);
```
albinsuresh: Optional: The current impl is intuitive enough. To simplify it even further, we could use an `IndexMap` from the `indexmap` crate (minimal overhead) to build the deduplicated message set on this first pass itself.

crates/core/tedge_api/src/store/message_log.rs
```diff
@@ -111,6 +157,66 @@ impl MessageLogWriter {
...
+        }
+        drop(reader);
+
+        // Keep the last occurrence of each topic, in order of last write (move-to-end semantics).
+        let mut seen = HashSet::new();
+        let mut deduped: Vec<_> = all_messages
+            .into_iter()
+            .rev()
+            .filter(|msg| seen.insert(msg.topic.name.clone()))
+            .collect();
```
albinsuresh: Could we cleanup of cleared messages as well during this compaction? The current impl would still keep the last "clear message" in the log, right?

albinsuresh (commented):
The code looks fine. Happy to approve once cleared message handling is also fixed. Also as an additional enhancement, how about doing an unconditional compaction on startup?

crates/core/tedge_api/src/store/message_log.rs
jarhodes314: I don't think so, I'll remove it

crates/core/tedge_api/src/store/message_log.rs
jarhodes314: `indexmap` will change the ordering semantics so we preserve the order in which the topics first appeared rather than the order in which the most recent message on the topic in question appeared. However, I believe this is actually more correct - we directly register entities from this log, so if we have the sequence messages:
1. `device/child0` registers
2. `device/child1` register specifying `device/child0` as the parent
3. `device/child0` is updated with a name

In the current implementation, we'd have the registration message for `device/child1` come before the registration message for `device/child0`, which is broken. With `indexmap`, we'd store the latest message for `device/child0` in the place of the original registration message, which is the behaviour we actually want.

I'll change it to use `indexmap` right away.

crates/core/tedge_api/src/store/message_log.rs
jarhodes314: Now fixed as part of the indexmap usage

crates/core/tedge_api/src/store/message_log.rs
```diff
@@ -65,43 +70,87 @@ impl MessageLogReader {
...
+
+            OpenOptions::new()
+                .create(true)
+                .append(true)
+                .open(&log_path)?
+        } else {
+            OpenOptions::new()
+                .create(true)
+                .write(true)
+                .truncate(true)
```
jarhodes314, line 97: The behaviour of this is slightly altered. Previously, truncate mode meant we opened the file with `truncate(true)`, then closed it, then opened it again (via `MessageLogWriter::new`). Now we just open it the one time with truncation. I believe this won't cause any dramatic behaviour changes as we didn't do anything to make the truncation fsync-durable, and it seemed from reading the code that we mostly did it for easy code reuse between the truncate and non-truncate paths.

This has the side-effect of us also failing if we fail to open the file with truncation (previously we ignored the result of opening the truncated file). I don't believe this will cause any major issues - if we can't open a file to truncate, we probably can't open it in append mode. There's a possible edge case where file system attributes might mean we can only append to a file, but in this case I think we're better off failing rather than pretending everything's working?

crates/core/tedge_api/src/store/message_log.rs
```diff
@@ -289,30 +287,145 @@ mod tests {
...
+            Some(make_message(
+                "te/device/child2//",
+                r#"{"@type":"child-device","name":"Child 2"}"#
+            ))
+        );
         assert_eq!(
             reader.next_message().unwrap(),
-            Some(make_message("topic1", "v3"))
+            Some(make_message(
+                "te/device/child01//",
```
albinsuresh: That looked opposite to what I had expected where the `child0` (the parent) should have appeared in the log before `child01` (the child of `child0`). The is one of the invariances of this persistent entity store file. The test is appending them in the wrong order leading to this state. In practice, the `child01` will only be persisted after `child0` is. Child registrations are cached in the `PendingEntityStore` until the parent is registered.

jarhodes314: Actually this whole test is a bit mad. We deregister the parent and then register it's children (while it's deregistered). I think the test is demonstrating useful behaviour, but it's accidental spillover from asking copilot for a motivating example in the case where we update the parent data.

The behaviour we're testing though is sensible IMO. If we receive an empty message this effectively deletes the topic, so a new message on that topic can reasonably be treated as totally novel.

albinsuresh (approved):
LGTM

crates/core/tedge_api/src/store/message_log.rs
```diff
@@ -65,43 +70,87 @@ impl MessageLogReader {
 /// A writer to append new MQTT messages to the end of the log
 pub(crate) struct MessageLogWriter {
     writer: BufWriter<File>,
+    log_dir: PathBuf,
+    redundancy_threshold: usize,
+    unique_topics: HashSet<String>,
+    total_entries: usize,
 }
```
didier-wenzek: I would consider the following. Using a map and not set would ease message compaction, that compaction being done in-memory on the fly.

```suggestion
/// A writer to append new MQTT messages to the end of the log
pub(crate) struct MessageLogWriter {
    writer: BufWriter<File>,
    log_dir: PathBuf,
    redundancy_threshold: usize,
    messages: HashMap<String,String>,
    redundancy_count: usize,
}
```

didier-wenzek (commented):
Okay, this PR fixes the issue. However, this leads to a design that seems overly complicated.

- Does the separation of `MessageLog` reader and writer add something? These structs are now fully tangled.
- Wouldn't be simpler to have a writer that maintains a map from topics to payloads rather that a set of topics?

albinsuresh:
> Okay, this PR fixes the issue. However, this leads to a design that seems overly complicated.
> 
> * Does the separation of `MessageLog` reader and writer add something? These structs are now fully tangled.

Not anymore with the current refactoring, as the writer is now doing reading as well. The earlier split was to isolate readers from writers, with only what they need.

> * Wouldn't be simpler to have a writer that maintains a map from topics to payloads rather that a set of topics?

That sounds like a mini-entity store at this level as well. Not quite sure about that. The mapper already has an `EntityCache` which it maintains with pretty much the same data. Maintaining another copy at this level, I'm not sure.

didier-wenzek (approved):
Approved. Thank you for the refactoring done along the fix.

Issue #4124, opened 2026-04-13
c8y_Agent is version does not update after thin-edge.io is upgraded


reubenmiller:
**Describe the bug**

The `c8y_Agent` fragment doesn't seem to be updated after tedge is updated. This results in the c8y_Agent shown in Cumulocity to give the user the wrong impression the version of thin-edge.io running on the device (though the version under the software tab is correct).

Below shows the stale "twin/c8y_Agent" fragment which shows the older 1.3.2 version instead of the expected, 2.0.0.

```sh
# tedge mqtt sub '#' | grep c8y_Agent
[te/device/main///twin/c8y_Agent] {"name":"thin-edge.io","url":"https://thin-edge.io","version":"1.3.2~328+g98b0f1d"}
^C
root@debian-fra1-01:~# tedge --version
tedge 2.0.0
```

**To Reproduce**

The steps are unclear how to reproduce as of yet, but I would assume if an older version would be installed, and then upgrading to the latest version.

**Expected behavior**

The `te/device/main///twin/c8y_Agent` fragment should always show the current version of thin-edge.io.

**Screenshots**

**Environment (please complete the following information):**
 - OS [incl. version]: `any`
 - Hardware [incl. revision]: `any`
 - System-Architecture [e.g. result of "uname -a"]: `any`
 - thin-edge.io version [e.g. 0.1.0]: `2.0.0`

**Additional context**

**entity store**

```sh
# cat /etc/tedge/.agent/entity_store.jsonl
{"version":"1.0"}
{"topic":{"name":"te/device/main/service/tedge-agent"},"payload":"{\"@parent\":\"device/main//\",\"@type\":\"service\",\"name\":\"tedge-agent\"}","qos":1,"retain":true}
{"topic":{"name":"te/device/main/service/c8y-firmware-plugin"},"payload":"{\"@parent\":\"device/main//\",\"@type\":\"service\",\"name\":\"c8y-firmware-plugin\"}","qos":1,"retain":true}
{"topic":{"name":"te/device/main/service/c8y-firmware-plugin/twin/type"},"payload":"\"service\"","qos":1,"retain":true}
{"topic":{"name":"te/device/main/service/mosquitto-c8y-bridge"},"payload":"{\"@parent\":\"device/main//\",\"@type\":\"service\",\"name\":\"mosquitto-c8y-bridge\"}","qos":1,"retain":true}
[...]

Issue #4093, opened 2026-04-01
flaky unit test: child_alarm_with_shuffled_entity_registration


reubenmiller:
**Describe the bug**

A [PR check run](https://github.com/thin-edge/thin-edge.io/actions/runs/23835119633/job/69477130249?pr=4092) found a flaky unit test regarding the ordering of child messages. Below shows the error output:

```
FAIL [   8.468s] ( 502/2497) c8y_mapper_ext shuffled_tests::child_alarm_with_shuffled_entity_registration
  stdout ───

    running 1 test
    test shuffled_tests::child_alarm_with_shuffled_entity_registration ... FAILED

    failures:

    failures:
        shuffled_tests::child_alarm_with_shuffled_entity_registration

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 296 filtered out; finished in 8.46s
    
  stderr ───

    thread 'shuffled_tests::child_alarm_with_shuffled_entity_registration' (94207) panicked at crates/extensions/tedge_mqtt_ext/src/test_helpers.rs:50:5:
    Payload assertion failed.
     Actual: "303,temperature_high,Temperature high,2026-04-01T06:32:50.496492327Z" 
    Expected message with topic: c8y/s/us/test-device:device:external_sensor, payload: 303,temperature_high,Still high
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
    proptest: Saving this and future failures in /home/runner/work/thin-edge.io/thin-edge.io/crates/extensions/c8y_mapper_ext/proptest-regressions/shuffled_tests.txt
    proptest: If this test was run on a CI system, you may wish to add the following line to your copy of the file.
    cc edb57ff902f26a1738fdceaf0d22fe7d8d4e910e1da4054944e73bd1a0867b58

    thread 'shuffled_tests::child_alarm_with_shuffled_entity_registration' (94207) panicked at crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs:349:1:
    Test failed: Payload assertion failed.
     Actual: "303,temperature_high,Temperature high,2026-04-01T06:32:50.496492327Z" 
    Expected message with topic: c8y/s/us/test-device:device:external_sensor, payload: 303,temperature_high,Still high.
    minimal failing input: ChildAlarmWithShuffledEntityRegistrationArgs = ChildAlarmWithShuffledEntityRegistrationArgs {
        field0: 15668371114638989990,
    }
```

**To Reproduce**

[...]

PR #4174, opened 2026-05-14
fix: tedge-agent publishes agent twin data on startup

albinsuresh:
## Proposed changes

`tedge-agent` publishes agent twin data fragment with the current installed version to `te/device/main///twin/agent` topic on startup.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4124

## Checklist


albinsuresh:
The remaining unit test failure of `nested_child_service_alarm_with_shuffled_ordering` is not a flaky one similar to #4093 . It seems to have regressed with the changes in this PR. Investigating. Most likely some updates to the test would be required. The rest of the code can be reviewed in the mean time.

crates/extensions/c8y_mapper_ext/src/shuffled_tests.rs
```diff
@@ -124,8 +124,8 @@ impl<'a, R: Rng> ShuffledMqttBox<'a, R> {
 #[async_trait::async_trait]
 impl<R: Rng + Send> MessageReceiver<MqttMessage> for ShuffledMqttBox<'_, R> {
     async fn try_recv(&mut self) -> Result<Option<MqttMessage>, RuntimeRequest> {
-        // Inject one pending message before reading output
-        let _ = self.send_one().await;
+        // Inject all pending messages to actors before reading output
+        while let Ok(true) = self.send_one().await {}
```
albinsuresh: Is this acceptable? Or draining just one message was intentional? If that's the case, I can make the test drain a single message, as per its expectations, on a test-by-test basis.

didier-wenzek: This weakens the shuffle logic.

Looking at the failing test fixed by this change, I wonder why *all* the `te/#` messages must be sent for a message received on `c8y/s/us`? Indeed, only  `te/#` messages are released by `self.send_one()` and the test is expecting a message on `c8y/s/us`. It seems that the test is missing a trigger.

albinsuresh: Here is the alternative approach: [8e6cb02](https://github.com/thin-edge/thin-edge.io/pull/4174/commits/8e6cb02476000e8467940367387abdb5c7ae0a37) where the messages are released from the test itself.

We're forced to release all messages at once, because of the nested entity registration ordering problem. The test registers an `immediate_child`, `nested_child` and `nested_child_service` followed by publishing the `nested_child_service` alarm. Since `try_recv` releases only one of these randomly, and if that ends up being any message other than the `immediate_child` registration (e.g: `nested_child` or `nested_child_service`), then that message will only be queued by the mapper, as the parents are not registered. So, to ensure that the entire hierarchy is registered, especially with random delivery of messages, we should release them all.

I agree that this issue is specific to tests that involves registration of hierarchical entities and hence I agree that a test-focussed fix is better than the earlier fix that released all messages implicitly from the `try_recv` itself.

didier-wenzek (approved):
Approved

PR #4175, opened 2026-05-15
refactor: fix retry behaviour in `tedge_flows.robot`

jarhodes314:
## Proposed changes
Modify the tests in `tedge_flows.robot` so retries are handled correctly.

For example, this failure shows a `Directory not empty` error when we retried the test: https://github.com/thin-edge/thin-edge.io/actions/runs/25811734528/job/75832244151. This is due to the directory not being cleaned-up at the end of the test, so the retry tries to install the already-installed flow and fails at that point.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
+++ b/tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
@@ -250,7 +250,7 @@ Reloading a broken script when its permission is fixed
     ...    minimum=1
     ...    message_contains="temperature":{"temperature":{"value":259.12}}
     ...    date_from=${start}
-    [Teardown]    Execute Command    cmd=chmod a-r /etc/tedge/mappers/local/flows/te_to_c8y.js
+    [Teardown]    Execute Command    cmd=chmod a+r /etc/tedge/mappers/local/flows/te_to_c8y.js
 
 Publishing transformation errors
     Install Flow    input-flows    publish-js-errors.toml
@@ -625,6 +625,7 @@ Script runs onStartup function
     ...    maximum=1
     ...    date_from=${start}
     Should Be Equal    ${messages}[0]    onstartup on_startup 2
+    [Teardown]    Uninstall Nested Flow    onstartup
 
 
 *** Keywords ***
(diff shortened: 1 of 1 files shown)


tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -33,7 +34,8 @@ Normalizes flows-dir path
     Should Not Contain    ${out3}    ${fail_due_to_invalid_path}
     Should Not Contain    ${out4}    ${fail_due_to_invalid_path}
 
-    Execute Command    rm -r ${child_dir} ${symlink}
+    Execute Command    rm -rf ${child_dir} ${symlink}
```
albinsuresh: Redundant as the teardown is doing the same.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -33,7 +34,8 @@ Normalizes flows-dir path
...
+    [Teardown]    Execute Command    cmd=rm -rf ${child_dir} ${symlink}
```
albinsuresh, line 37: When the same cleanup is performed at the beginning of the test, which will get executed on retries, isn't this teardown command redundant? Setting a clean slate at the beginning of the test has the benefit of being resilient to any side effects from other tests as well and not just retries of this same test.

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
```diff
@@ -187,12 +189,13 @@ Consuming messages from a file, periodically
...
 
 Appending messages to a file
+    Execute Command    cmd=rm -f /tmp/events.log
     Install Flow    input-flows    append-to-file.toml
     Execute Command    for i in $(seq 3); do tedge mqtt pub seq/events "$i"; done
     Execute Command    grep '\\[seq/events\\] 1' /tmp/events.log
     Execute Command    grep '\\[seq/events\\] 2' /tmp/events.log
     Execute Command    grep '\\[seq/events\\] 3' /tmp/events.log
-    [Teardown]    Uninstall Flow    append-to-file.toml
+    [Teardown]    Run Keywords    Uninstall Flow    append-to-file.toml    AND    Execute Command    cmd=rm -f /tmp/events.log
```
albinsuresh, line 230: Same comment as above regarding the redundancy of the cleanup logic.

didier-wenzek (approved):
Approved

tests/RobotFramework/tests/tedge_flows/tedge_flows.robot
reubenmiller: I think its because all tests in the suite are using the same container, so it is to reduce the amount of state which is left behind should it fail.

Issue #4167, opened 2026-05-11
allow relative path resolution for commands set in input.process.command


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

If a flow is packaged as a tarball that includes a shell script which is to be called from the `input.process` property, then the flow.toml definition, must use an explicit path when referencing that shell script.

For example, given a flow called `foo`, which contains `flow.toml`, `example.sh` script, then the only way the `example.sh` script can be executed is if the full path to the script is used, e.g. `/etc/tedge/mappers/c8y/flows/foo/example.sh`, however this assumes that the flow is being installed on the default `TEDGE_CONFIG_DIR` path, e.g. `/etc/tedge`. For customer installations, then the path will be invalid and the flow will fail.

```toml
name = "foo"
version = "1.0.0"
description = "Call a script packaged with the 'foo' flow"

[input.process]
command = """/etc/tedge/mappers/c8y/flows/foo/example.sh"""
interval = "60s"

[[steps]]
script = "main.js"
```

**Describe the solution you'd like**

When executing a command as defined in the `input.process.command` property, the path either be resolved relative to the flow definition file (e.g. typically this is the `flow.toml` file), or the command should be executed with the current working directory (CWD) set to the flow's root directory, e.g. `/etc/tedge/mappers/c8y/flows/foo` in this case. This would also be consistent with the relative path resolution of the javascript files.

Below shows an example of the same flow file using a relative path (where `./` is inferred).

**File: foo/flow.toml**

```toml
name = "foo"
version = "1.0.0"
description = "Call a script packaged with the 'foo' flow"

[input.process]
# here, the example.sh will be resolved as ${TEDGE_CONFIG_DIR}/mappers/<name>/flows/<flow_name>/example.sh
command = """./example.sh"""
interval = "60s"

[[steps]]
script = "main.js"
```

**Describe alternatives you've considered**

**Additional context**

PR #4176, opened 2026-05-15
feat: Allow `[input.process]` to take a relative path for `command`

jarhodes314:
## Proposed changes
Allow users to specify a relative path for the `command` in an `[input.process]`, enabling people to ship the input script as part of the flow itself.

## Types of changes

## Paste Link to the issue
- #4167

## Checklist


albinsuresh (approved):
LGTM.

crates/extensions/tedge_flows/Cargo.toml
```diff
@@ -15,6 +15,7 @@ camino = { workspace = true, features = ["serde1"] }
 futures = { workspace = true }
 glob = { workspace = true }
 humantime = { workspace = true }
+path-clean = { workspace = true }
```
didier-wenzek, line 18: Even if controlling dependencies seems to be a lost battle, I wonder. Wouldn't `canonicalize` be enough?

crates/extensions/tedge_flows/src/config.rs
```diff
@@ -472,6 +475,27 @@ impl InputConfig {
...
+        match self {
+            InputConfig::Process {
+                command,
+                topic,
+                interval,
+            } => InputConfig::Process {
+                command: if Utf8Path::new(&command).is_relative() {
+                    path_clean::clean(flows_dir.join(&command))
+                        .to_string_lossy()
+                        .into_owned()
```
didier-wenzek: Beware, the `command` is not a path but a command line possibly with arguments.

We to call first `shell_words::split().

We can also consider to:

- normalize not only the command path but any argument looking akin to a path.
- properly type `InputConfig::Process::command` as an array of `String`.

didier-wenzek: For instance the following test is red:

```rust
    #[test]
    fn relative_process_paths_are_cleaned_up() {
        let flow_toml = r#"
        [input.process]
        command = "./../flows/./my_flow/../test.sh foo /../bar"
        interval = "1s"
        "#;

        let flow: FlowConfig = toml::from_str(flow_toml).unwrap();
        let resolved = flow
            .input
            .resolve_relative_process_paths(Utf8Path::new("/flows"));
        assert_eq!(
            resolved,
            InputConfig::Process {
                topic: None,
                command: "/flows/test.sh foo /../bar".into(),
                interval: Some(IntervalConfig::Duration(Duration::from_secs(1))),
            }
        )
    }
```

didier-wenzek (changes requested):

crates/extensions/tedge_flows/Cargo.toml
jarhodes314: Almost certainly, I didn't realise we already depended on two crates that do this same thing (`path-clean` is not a new dependency)

crates/extensions/tedge_flows/src/config.rs
jarhodes314: I've now added this (among other tests) in 013ff0c

didier-wenzek (approved):
Approved

PR #4178, opened 2026-05-18
refactor: respect user group values from system toml for directory creation

rina23q:
## Proposed changes
Follow-up PR to address the issue found in https://github.com/thin-edge/thin-edge.io/pull/4142#discussion_r3257736431.
The fix leads the further refactoring in the same way as done in #4142.

Also, this PR removes the unused `/etc/tedge/.tedge-mapper-c8y` directory from tedge-mapper-c8y.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist


crates/common/tedge_utils/src/paths.rs
```diff
@@ -128,27 +128,10 @@ impl TedgePaths {
...
-                root: self.root.clone(),
-                path,
-                owner: self.default_owner.clone(),
-                mode: DEFAULT_FILE_MODE,
-                warn_and_ignore_permission_errors: false,
-            },
-            parent,
-            warn_and_ignore_permission_errors: false,
-        })
+        let absolute = self.resolve(path)?;
```
rina23q: There was a need to call `template_file()` from `ManagedDir`, so I did this refactoring. However, I am not 100% confident if I did correctly.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -272,10 +274,6 @@ impl Agent {
         // under log_dir (/var/log/tedge)
         self.config.agent_log_dir.ensure().await?;
 
-        // under state_dir (/data/tedge/agent)
```
rina23q, line 275: This is a mistake from the previous refactoring - this directory is supposed to be under `/etc/tedge`.

crates/extensions/c8y_mapper_ext/src/actor.rs
```diff
@@ -410,15 +406,7 @@ impl C8yMapperBuilder {
...
-        TedgePaths::from_root_with_defaults(&config.ops_dir, "", "")
-            .root_dir()
-            .ensure()
-            .await?;
-        // Create directory for persistent entity store
-        TedgePaths::from_root_with_defaults(&config.state_dir, "", "")
-            .root_dir()
-            .ensure()
-            .await?;
+        config.ops_dir.ensure().await?;
```
rina23q, line 409: This is the key of this refactoring. Since the desired user/group of `config.ops_dir` is already derived from the root directory (= `config_dir` = `/etc/tedge`) settings, and the values from `system.toml` are already taken into consideration when `config_dir` is constructed, we don't need to read `system.toml` here.

crates/core/tedge_agent/src/agent.rs
didier-wenzek: I don't get what the mistake was. Apart the comment that is gone, the code is unchanged - only moved few lines above.

crates/common/tedge_utils/src/paths.rs
didier-wenzek: The refactoring seems to be up-side down. Wouldn't be simpler to invoke `TedgePaths::template_file()` from `ManagedDir::template_file`, instead of the reverse as proposed here?

This would remove the need to strip a prefix in `TedgePaths::template_file()` and then to add back this prefix in  `ManagedDir::template_file` - with some unnecessary checks and `unwrap()` along the way.

crates/core/tedge_agent/src/agent.rs
rina23q: The comment itself was the mistake :D

crates/common/tedge_utils/src/paths.rs
rina23q: Changed in 9f580701af383f91f3ac53f28e59f150a81d8f46. Reverted `TedgePaths::template_file()`. Now `ManagedDir::template_file()` calls `TedgePaths::template_file()`.

crates/common/tedge_utils/src/paths.rs
```diff
@@ -245,6 +245,25 @@ impl ManagedDir {
...
+    /// See [`TedgePaths::template_file`] for details.
+    pub fn template_file(
+        &self,
+        path: impl AsRef<Utf8Path>,
+    ) -> Result<ManagedTemplateFile, PathsError> {
+        let relative_from_root = self
+            .path
+            .strip_prefix(&self.root)
+            .expect("managed dir path must be inside root")
+            .join(path);
```
didier-wenzek: This logic combining `self.root` and `self.path` seems complicated at first, but is actually correct.

- Do we have to `validate_managed_path(path)`?
- It might be good to add a `ManagedDir` method to get the `relative_path_from_root()` for a path.

didier-wenzek (approved):
Approved

crates/common/tedge_utils/src/paths.rs
rina23q: > Do we have to `validate_managed_path(path)`?

In this context, no. Because the first line of `TedgePaths::template_file()` does `self.resolve(path)`. This `resolve` calls `validate_managed_path()`.

> t might be good to add a `ManagedDir` method to get the `relative_path_from_root()` for a path.

I think this is a quick win and nice as API. I will do this.

crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs
```diff
@@ -197,7 +197,7 @@ impl SupportedOperations {
     /// If a given directory is a c8y operations directory of a device, return that device's external id.
     pub fn xid_from_path(&self, ops_dir: &Path) -> Result<ExternalId, anyhow::Error> {
         ensure!(
-            ops_dir.starts_with(self.base_ops_dir.as_std_path()),
+            ops_dir.starts_with(self.base_ops_dir.path().as_std_path()),
```
jarhodes314: Do we actually need the `as_std_path()` here? If we do, can we avoid that by taking `impl AsRef<Path>`, like we do elsewhere so `Utf8Path` just worsk?

crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs
```diff
@@ -213,7 +213,7 @@ impl SupportedOperations {
             )
         );
 
-        let suffix = ops_dir.strip_prefix(self.base_ops_dir.as_std_path())?;
+        let suffix = ops_dir.strip_prefix(self.base_ops_dir.path().as_std_path())?;
```
jarhodes314: I suspect this can also take `impl AsRef<Path>` to avoid the `as_std_path()` conversion

crates/extensions/tedge_config_manager/src/tests.rs
```diff
@@ -932,9 +933,9 @@ async fn test_init_updates_both_when_identical() -> anyhow::Result<()> {
     let config = test_config(&tempdir);
 
     let old_template = "# Old workflow\noperation = \"config_update\"";
-    let template_file = config.ops_dir.join("config_update.toml.template");
+    let template_file = config.ops_dir.path().join("config_update.toml.template");
     tokio::fs::write(&template_file, old_template).await?;
-    let toml_file = config.ops_dir.join("config_update.toml");
+    let toml_file = config.ops_dir.path().join("config_update.toml");
     tokio::fs::write(&toml_file, old_template).await?;
```
jarhodes314: My ultimate goal was to enforce (via `clippy::disallowed_methods`) that we don't directly call things like `tokio::fs::write` to ensure we use the APIs we've created (`TedgePaths` and `TempTedgeDir`) as those should be able to handle our file writing needs, and it helps ensure we actually use those APIs.

On that note, we probably ought to rewrite these to use the proper `TedgePaths` APIs for writing these files in these tests we're already touching to make them work with the refactored code.

jarhodes314 (approved):
A few minor comments, but broadly happy with the changes

crates/common/tedge_utils/src/paths.rs
rina23q: Added `relative_path_from_root()` in https://github.com/thin-edge/thin-edge.io/pull/4178/changes/a2229aac63c88b2b4aa88e996e6ce2e8df55f142.

crates/extensions/c8y_mapper_ext/src/supported_operations/mod.rs
rina23q: Yes, you are right. I removed `as_std_path()` in https://github.com/thin-edge/thin-edge.io/pull/4178/changes/a2229aac63c88b2b4aa88e996e6ce2e8df55f142.

crates/extensions/tedge_config_manager/src/tests.rs
rina23q: It makes sense what you suggest and it was actually easy to get rid of the usages of `tokio::fs::write` from the touched tests in a0293afd0a7fcf5d003832c0965622d36cdddbd7. 

I see there are still a lot of usages of `tokio::fs::write` in our unit tests. This is a candidate for the further refactoring in the future. (won't be addressed in this PR)

albinsuresh (commented):
A few instances of `TedgePaths::from_root_with_defaults(dir, "", "")` were missed:

1. `persist_builtin_flow()` in `tedge_flows::registry.rs`
2. `DiagCollectCommand::execute()` in `tedge::cli::diag::collect.rs`

Feel free to merge after those are also fixed.

rina23q:
> A few instances of `TedgePaths::from_root_with_defaults(dir, "", "")` were missed:
> 
> 1. `persist_builtin_flow()` in `tedge_flows::registry.rs`
> 2. `DiagCollectCommand::execute()` in `tedge::cli::diag::collect.rs`
> 
> Feel free to merge after those are also fixed.

For 1., yes it has to be fixed as well, and will be a lot refactoring again... => I will address in a separated PR as the previous and this PRs didn't touch `tedge_flows` code, it requires more work.

For 2., I don't think it has to change. The diag output directory is where the tarball will be stored. The process user/group should be appropriate.

Issue #1764, opened 2023-02-27
Avoid or limit explicit init required by tedge components


albinsuresh:
**Is your feature improvement request related to a problem? Please describe.**

Most of the thin-edge daemons support an `--init` command with which the initialization of these daemons are performed. They use this option to prepare their initial state by creating the necessary files, directories etc. Doing this `-init` for each and every component is tedious and we should try to avoid such explicit inits as much as possible by doing as many initialization steps as possible dynamically during the runtime of these daemons.

**Describe the solution you'd like**

Deprecate the `--init` options of all the daemons to no-ops, so that they can be completely removed in future. These daemons may create their init state(necessary directories, files etc) dynamically, as and when they are required.

For initialization steps that can't be performed during the normal runtime of a component, like creating `/etc/tedge` directory that can only be done with elevated privileges, provide a `tedge init --user <user>` command which will perform these "essential" initializations with the provided `user`. If a `--user` value is not provided, use the default `tedge` user.

albinsuresh:
### Test Plan

1. The `sudo tedge init --user <user-name> --group <group-name>` subcommand is used to initialize the following top-level directories such as `/etc/tedge`, `/var/tedge` and `/var/tedge/logs` which require elevated privileges to be created. The `--user` and `--group` args are optional and defaults to `tedge` user and group.
2. The `tedge --init` arg is deprecated with a warning in favour of the `tedge init` subcommand.
3. The `--init` arg was deprecated from all tedge daemons as well, namely:
    - `tedge-mapper`
    - `tedge-agent`
    - `c8y-configuration-plugin`
    - `c8y-log-plugin`
    - `c8y-fimware-plugin`
4. The directories/files created by these daemons using `--init` earlier(e.g: the `/etc/tedge/c8y/c8y-configuration-plugin` file created by the `c8y-configuration-plugin`) are now created dynamically during the startup of these services. Even if some of these files are deleted at some point, they are recreated on next startup.

gligorisaev:
> ### Test Plan
> 1. The `sudo tedge init --user <user-name> --group <group-name>` subcommand is used to initialize the following top-level directories which require elevated privileges to be created:
>    
>    * /etc/tedge
>    * /etc/tedge/mosquitto-conf
>    * /etc/tedge/operations
>    * /etc/tedge/plugins
>    * /etc/tedge/device-certs
>    * /var/tedge
>    * /var/tedge/logs
[...]

PR #4179, opened 2026-05-19
feat: remove deprecated `--init` option from tedge-agent

rina23q:
## Proposed changes

`tedge-agent --init` is deprecated since 2023, ticket [#1764](https://github.com/thin-edge/thin-edge.io/issues/1764). I think it's good time to remove the option. 

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/core/tedge_agent/src/lib.rs
+++ b/crates/core/tedge_agent/src/lib.rs
@@ -16,7 +16,6 @@ use std::sync::Arc;
 use agent::AgentConfig;
 use tedge_config::cli::CommonArgs;
 use tedge_config::log_init;
-use tracing::log::warn;
 
 mod agent;
 mod device_profile_manager;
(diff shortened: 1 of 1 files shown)


jarhodes314 (approved):
Approved.

Just a small note about my feelings on the deprecation: when I think of "deprecated", I generally expect that to mean "I'll warn that I don't like you doing this as there's a new way to do it, but it still works". In our case, it seems to mean "we're going to tell you this is deprecated and silently do nothing, but exit with a success status".

didier-wenzek (approved):
Approved

didier-wenzek:
> Just a small note about my feelings on the deprecation: when I think of "deprecated", I generally expect that to mean "I'll warn that I don't like you doing this as there's a new way to do it, but it still works". In our case, it seems to mean "we're going to tell you this is deprecated and silently do nothing, but exit with a success status".

Yeah, in that case, it was more "there is no more point to call --init; and that option will be removed in the future"

Issue #4169, opened 2026-05-13
Operations stay in pending after installing container-group which include a container in a restart loop


k-butz:
**Describe the bug**
Once I'm installing a certain docker-compose project in tedge-demo-container, all following operations stay in pending until tedge-agent is restarted manually. See reproduction steps for details. 

**To Reproduce**
1. Upload this docker-compose project to the Cumulocity Software Repository. Use "container-group" as software type. 
```yaml

services:
  gateway:
    # Image is maintained under: https://github.com/thin-edge/opcua-device-gateway-container
    image: ghcr.io/thin-edge/opcua-device-gateway:latest
    container_name: opc-device-gateway
    restart: always
    environment:
    # Use configuration from https://cumulocity.com/docs/device-integration/opcua/#gateway-configuration-and-registration 
      - GATEWAY_IDENTIFIER=OPCUAGateway
      - GATEWAY_NAME=OPCUAGateway
      - GATEWAY_THINEDGE_MQTTSERVERURL=tcp://host.containers.internal:1883
      - GATEWAY_THINEDGE_USEFORDATAFORWARDING=true
      - GATEWAY_THINEDGE_ENABLED=true
      - GATEWAY_MAPPINGS_MERGECYCLICREAD=false
      - GATEWAY_MONITORING_INTERVAL=3600000

    volumes:
      - opcua_data:/app/data
      # Provide access to thin-edge.io configuration
      - /etc/tedge:/etc/tedge
      - /etc/tedge/device-certs:/etc/tedge/device-certs
    networks:
      - backend

volumes:
  opcua_data:
    name: opcua_data
networks:
  backend:
    name: opc-ua-gateway_backend
```
2. Spawn a tedge-demo-container on localhost: `c8y tedge demo start tedge-dbg-01`
3. Go to Software Tab and install the container group from Step 1 => Operation succeeds. It got installed, however, this project runs into endless restart loop (there's also lots of Events about the restarts in Event Tab)
4. Run any other operation, e.g. a shell command with "df -h" => Operation stays in pending
5. Wait (I waited few minutes), Operation still in pending
6. Now do a manual restart of tedge-agent via `systemctl restart tedge-agent` => Operation is executed and moves into successful
7. Do any other operations => They again stay in pending until tedge-agent is restarted manually

I've did a `tedge diag collect` after step 5, see attached. 

**Expected behavior**
A misbehaving container should not block thin-edge from executing operations.

**Environment (please complete the following information):**
[...]

PR #4180, opened 2026-05-20
fix: deduplicate agent sync messages so only most recent one is processed

Bravo555:
## Proposed changes

As part of #4169, where part of the problem is that sync messages are getting queued up because they're processed much slower than they are arriving, ~~rate limit sync messages to only process 1 sync message of each type every 10s.~~ deduplicate sync messages so we don't waste time processing stale messages.

I haven't made a test for that since it would be annoying to do as it relies somewhat on the environment and its performance characteristics and is very bandwidth heavy if CPU is not limited, so I've been testing it manually.
The scenario before the fix is applied looks as follows:

1. Run a single thin-edge instance in a docker container (can be the project's default devcontainer)
2. Limit CPU time for the container using `docker update --cpus=".05" CONTAINER_ID` (tweak as necessary to make it as slow as you want, you could also perhaps use systemd's CPU quotas to only slow down `tedge-agent` and `tedge-mapper-c8y` and not the entire container, but I haven't tried that)
3. Run a publish loop: `while true; do tedge mqtt pub te/device/main/service/tedge-agent/signal/sync_log_upload {}; sleep 0.5; done`
4. Then invoke a shell command from c8y (`tedge-command-plugin` must be installed)
5. The command stays in `PENDING` for a long time, potentially indefinitely, if it progresses to `EXECUTING` then it stays there indefinitely
6. When running `tedge mqtt sub "#"` one can see that sync messages are sent much faster than `tedge-agent` can respond to them
7. After stopping the publish loop, `tedge-agent` continues to respond to queued messages

With the fix, because we're immediately discarding additional sync messages ~~if one has already been sent in the last 10s~~, sent commands complete promptly and once publish loop is stopped, `tedge-agent` does not continue replying (because sync messages we didn't have bandwidth to process were dropped). We can do it because `sync...` messages do not have any payload and are used only to trigger the agent to possibly run some other commands, save their output, and sync this with the cloud.
In other words, `sync...` is a type of request that only the most recent one has to be handled and previous ones can be safely dropped, but so far we're processing them like regular messages that contain payloads and have to be processed individually.

~~This fix is thus a simple mechanism to reduce the number of sync messages that are processed. It's main disadvantage is that if there is some update that changes the state we're syncing but the sync message falls within the rate limit it will not be processed, i.e. the most recent sync message will not be processed. Potential alternatives that address this are discussed in _Further comments_.~~ See https://github.com/thin-edge/thin-edge.io/pull/4180#issuecomment-4510407130.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

## Further comments

The current mechanism only slightly changes `process_signal_message` to implement rate limiting, however, it has the disadvantage that the most recent sync message is not processed if it falls within the rate limit window, but we'd like to always process the most recent one and just reduce the amount of duplicate syncs we have to process.
Addressing that though would have to involve some changes to the actor's main processing loop. From the least to most complete:

### Instead of discarding, delay the message until after rate limit interval and deduplicate

Actor's even loop could be changed as follows:

```rust
        while let Some(input) = self.input_receiver.recv().await {
            match input {
                AgentInput::MqttMessage(message) => {
                    self.process_mqtt_message(message).await?;
                }
                AgentInput::InternalCommandState(InternalCommandState(command_state)) => {
                    self.process_command_update(command_state).await?;
                }
                AgentInput::GenericCommandData(GenericCommandData::State(new_state)) => {
                    self.process_builtin_command_update(new_state).await?;
                }
                AgentInput::GenericCommandData(GenericCommandData::Metadata(
                    GenericCommandMetadata { operation, payload },
                )) => {
                    self.publish_builtin_capability(operation, payload).await?;
                }
                AgentInput::FsWatchEvent(file_update) => {
                    if let Some(updated_capability) = self
                        .workflow_repository
                        .update_operation_workflows(
                            &self.mqtt_schema,
                            &self.device_topic_id,
                            file_update,
                        )
                        .await
                    {
                        self.mqtt_publisher.send(updated_capability).await?
                    }
                }
            }
           self.process_pending_sync_messages().await?; // NEW
        }
```

The in the main sync processing function we could deduplicate and schedule sync message to be run some interval after last processed message and `process_pending_messages` process only after that interval. However there would still have to be some message that triggers it and without any new messages the most recent sync will not be published.

### Instead of rate limiting peek the next message in the receiver and deduplicate if it's the same type of sync message

Instead of rate limiting try to consume and deduplicate as much sync messages of the same type to process most recent one. In main event loop we would peek one additional message and if we could deduplicate it with the current one, we would consume it and potentially repeat until we can't deduplicate. We could still fall behind if sync messages are interleaved with some other messages or are shuffled sync messages of different types. Also our current receiver doesn't provide a `peek()` method that doesn't immediately take the message off the receiver.

### Instead of processing one message at a time, fetch all current messages into a list and then decide what to do with them.

Inspired by https://matklad.github.io/2025/05/14/scalar-select-aniti-pattern.html

If instead of processing 1 message at a time we could consume all messages that are ready we could more effectively deduplicate them. For sync messages especially, interleaved messages of different types could be reordered and deduplicated. However the receiver would have to provide a method to get many messages that are currently ready without awaiting the next one once the current channel is ready, or provide a way to query for the current number of messages in the channel so we can receive set number of messages.

### DIFF
--- a/crates/core/tedge_actors/src/message_boxes.rs
+++ b/crates/core/tedge_actors/src/message_boxes.rs
@@ -92,6 +92,7 @@ use crate::Message;
 use crate::RuntimeRequest;
 use async_trait::async_trait;
 use futures::channel::mpsc;
+use futures::channel::mpsc::TryRecvError;
 use futures::StreamExt;
 use log::debug;
 use std::fmt::Debug;
@@ -396,5 +397,75 @@ impl<Input: Send> MessageReceiver<Input> for CombinedReceiver<Input> {
 
     async fn recv_signal(&mut self) -> Option<RuntimeRequest> {
         self.signal_receiver.next().await
+    }
+}
+
+pub trait MessageReceiverNoblock<T>: MessageReceiver<T> {
+    /// Tries to receive a message from the channel without blocking.
+    /// If the channel is empty, or empty and closed, this method returns an error.
+    fn recv_noblock(&mut self) -> Result<T, RecvNoblockError>;
+
+    /// Returns all the currently received messages in the message box.
+    ///
+    /// Returns `None` if input channel is closed or RuntimeRequest is recieved. Returns
+    /// `Some(messages)` if there is at least 1 message. If there are currently no messages, we
+    /// await until one is received.
+    ///
+    /// If there are many messages currently ready, it can be useful to collect them all for
+    /// inspection/filtering/deduplication/reordering instead of strictly processing them in order 1
+    /// at a time.
+    ///
+    /// https://matklad.github.io/2025/05/14/scalar-select-aniti-pattern
+    #[expect(async_fn_in_trait)]
+    async fn recv_many(&mut self) -> Option<Vec<T>> {
+        // don't return until we have at least 1 message
+        let message = self.recv().await?;
+        // instead of always allocating and returning a vec from this function we could consider
+        // taking one as an &mut argument
+        let mut messages = vec![message];
+        while let Ok(message) = self.recv_noblock() {
+            messages.push(message);
+        }
+        Some(messages)
+    }
+}
+
+pub enum RecvNoblockError {
+    Empty,
+    Closed,
+}
+
+impl From<TryRecvError> for RecvNoblockError {
+    fn from(value: TryRecvError) -> Self {
+        match value {
+            TryRecvError::Closed => Self::Closed,
+            TryRecvError::Empty => Self::Empty,
+        }
+    }
+}
+
+impl<T: Send> MessageReceiverNoblock<T> for CombinedReceiver<T> {
+    /// Tries to receive a message from the channel without blocking.
+    /// If the channel is empty, or empty and closed, this method returns an error.
+    fn recv_noblock(&mut self) -> Result<T, RecvNoblockError> {
+        self.input_receiver
+            .try_recv()
+            .map_err(RecvNoblockError::from)
+    }
+}
+
+impl<T: Send + Debug> MessageReceiverNoblock<T> for LoggingReceiver<T> {
+    fn recv_noblock(&mut self) -> Result<T, RecvNoblockError> {
+        self.receiver.recv_noblock()
+    }
+}
+
+impl<I, O> MessageReceiverNoblock<I> for SimpleMessageBox<I, O>
+where
+    I: Send + Debug + 'static,
+    O: Send + Debug + 'static,
+{
+    fn recv_noblock(&mut self) -> Result<I, RecvNoblockError> {
+        self.input_receiver.recv_noblock()
     }
 }
--- a/crates/extensions/tedge_config_manager/src/actor.rs
+++ b/crates/extensions/tedge_config_manager/src/actor.rs
@@ -14,7 +14,7 @@ use tedge_actors::ChannelError;
 use tedge_actors::ClientMessageBox;
 use tedge_actors::LoggingReceiver;
 use tedge_actors::LoggingSender;
-use tedge_actors::MessageReceiver;
+use tedge_actors::MessageReceiverNoblock;
 use tedge_actors::RequestEnvelope;
 use tedge_actors::RuntimeError;
 use tedge_actors::Sender;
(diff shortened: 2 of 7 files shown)


crates/core/tedge_agent/src/operation_workflows/actor.rs
```diff
@@ -78,6 +80,12 @@ pub struct WorkflowActor {
     pub(crate) script_runner: ClientMessageBox<Execute, std::io::Result<Output>>,
     pub(crate) downloader: ClientMessageBox<DownloaderRequest, DownloaderResult>,
     pub(crate) tmp_dir: Utf8PathBuf,
+
+    /// Timestamps of different types of most recently processed sync messages. If different types
+    /// of sync messages like `sync_log_upload` and `sync_device_profile` are published, we want to
+    /// process them all, so we rate-limit them by topic.
```
didier-wenzek: Minor: the following seems more appropriate

```suggestion
    /// Timestamps of different types of most recently processed sync messages. If different types
    /// of sync messages like `sync_log_upload` and `sync_device_profile` are published, we want to
    /// process them independtly, so we rate-limit them by topic.
```

didier-wenzek (commented):
It's not clear to me if that PR is a fix or only a partial fix to https://github.com/thin-edge/thin-edge.io/issues/4169.

>  (This PR)'s main disadvantage is that if there is some update that changes the state we're syncing but the sync message falls within the rate limit it will not be processed, i.e. the most recent sync message will not be processed.

I think this is okay-ish, even if there is a race condition. Indeed, there is no contract on how `sync` signals interact with effect-full operation requests.

> Instead of rate limiting peek the next message in the receiver and deduplicate if it's the same type of sync message

There is currently no `peek` on the workflow actor receiver, making this not easy to implement. However, that strategy could be used on the actors actually reacting to the synchronization requests (e.g  [`tedge_config_manager`](https://github.com/thin-edge/thin-edge.io/blob/main/crates/extensions/tedge_config_manager/src/actor.rs#L167)) using two independent receivers.

> Instead of processing one message at a time, fetch all current messages into a list and then decide what to do with them.

A very good place to implement this would by the final actors such as the `tedge_config_manager`. These actors have the more appropriate knowledge of which operation requests might or might not interact with a sync command.

crates/core/tedge_agent/src/operation_workflows/actor.rs
Bravo555: fixed in 11f6cc8e19 (amended)

Bravo555:
> It's not clear to me if that PR is a fix or only a partial fix to https://github.com/thin-edge/thin-edge.io/issues/4169

My bad, somehow I missed this important information. Indeed that is only a partial fix of #4169, where the root of the problem was that service entered a restart loop which caused `tedge-container-plugin` to send a flood of `sync_log_upload` messages, which caused the degradation of `tedge-agent` not being able to progress other operations. AFAIK @reubenmiller implemented a fix on `tedge-container-plugin` side to avoid sending the flood of messages, which is probably a sufficient fix, but still we can try to have `tedge-agent` handle the floods of these types of messages better, providing a more graceful degradation, in case it arrives by some other means.

> A very good place to implement this would by the final actors such as the tedge_config_manager. These actors have the more appropriate knowledge of which operation requests might or might not interact with a sync command.

Indeed the main workflow actor only dispatches the sync commands to `log` and `config` actors and doesn't wait for a response, so these actors are where the deduplication logic should live. Before merging as is, I can try to incorporate one of these approaches into this PR. Do you have any preference for which would be the best?

didier-wenzek:
> > A very good place to implement this would by the final actors such as the tedge_config_manager. These actors have the more appropriate knowledge of which operation requests might or might not interact with a sync command.
> 
> Indeed the main workflow actor only dispatches the sync commands to `log` and `config` actors and doesn't wait for a response, so these actors are where the deduplication logic should live. Before merging as is, I can try to incorporate one of these approaches into this PR. Do you have any preference for which would be the best?

I don't have strong preference. Implementation constraints might drive the choice.

Bravo555:
I have implemented the approach where the log and config actors receive all the currently available messages and deduplicate the sync messages, and it works nicely.

Running `while true; do tedge mqtt pub te/device/main/service/tedge-agent/signal/sync {}; sleep 0.01; done` which syncs all the state, not only `log_upload`, the operations are consistently processed in a timely manner (in about 5 seconds on my system when limiting the CPU of the devcontainer to `0.1`). Had to implement it for the config actor as well as the `config_update` triggers a sync for `log_upload` and after that, when a publish loop is stopped, `tedge-agent` stops publishing cloudbound sync messages almost immediately.

 So I think this smarter approach can be merged. Once approved, I will remove the leftover previous fix + its revert.

crates/core/tedge_actors/src/message_boxes.rs
```diff
@@ -157,6 +158,12 @@ impl<Input: Debug> LoggingReceiver<Input> {
     pub fn close_input(&mut self) {
         self.receiver.close_input();
     }
+
+    /// Tries to receive a message from the channel without blocking.
+    /// If the channel is empty, or empty and closed, this method returns an error.
+    pub async fn recv_noblock(&mut self) -> Result<Input, RecvNoblockError> {
```
didier-wenzek: Making this method non`async` would better convey the fact this method is non blocking.
Also, the name `try_recv` would be more usual.


```suggestion
    pub fn try_recv(&mut self) -> Result<Input, TryRecvError> {
```

crates/extensions/tedge_config_manager/src/actor.rs
```diff
@@ -201,6 +210,31 @@ impl ConfigManagerActor {
...
+    /// Returns `None` if input channel is closed or RuntimeRequest is recieved. Returns
+    /// `Some(messages)` if there is at least 1 message. If there are currently no messages, we
+    /// await until one is received.
+    ///
+    /// If there are many messages currently ready, it can be useful to collect them all for
+    /// inspection/filtering/deduplication/reordering instead of strictly processing them in order 1
+    /// at a time.
+    ///
+    /// https://matklad.github.io/2025/05/14/scalar-select-aniti-pattern
+    // if more actors end up using this approach, we could consider putting it in the actors crate
```
didier-wenzek: Better to do that is this PR, adding a `recv_many()` method along the `try_recv()` (currently named  `recv_noblock()` )

That method delegates everything to `self.input_receiver`. This is a hint that the code has to be moved to the `input_receiver`.

didier-wenzek (commented):
> I have implemented the approach where the log and config actors receive all the currently available messages and deduplicate the sync messages, and it works nicely.

Indeed, really nice and nicely implemented.

I have some suggestions though.

crates/core/tedge_actors/src/message_boxes.rs
Bravo555: Removed `async` in 88c50daeb0.

I can't rename it to `try_recv` however, because it conflicts with `MessageReceiver::try_recv`. One could call the type-specific implementation by fully qualifying it, but I think it's better to have separate names. Alternatively we could consider renaming `MessageReceiver::try_recv`, but that would be a big refactoring touching many files, so I'd opt for keeping the current name for now.

didier-wenzek: Ok so let the name unchanged

crates/extensions/tedge_config_manager/src/actor.rs
Bravo555: Implementation moved into `tedge_actors` crate in e7b773a313.

crates/core/tedge_actors/src/message_boxes.rs
```diff
@@ -398,3 +399,73 @@ impl<Input: Send> MessageReceiver<Input> for CombinedReceiver<Input> {
...
+    /// inspection/filtering/deduplication/reordering instead of strictly processing them in order 1
+    /// at a time.
+    ///
+    /// https://matklad.github.io/2025/05/14/scalar-select-aniti-pattern
+    #[expect(async_fn_in_trait)]
+    async fn recv_many(&mut self) -> Option<Vec<T>> {
+        // don't return until we have at least 1 message
+        let message = self.recv().await?;
+        // instead of always allocating and returning a vec from this function we could consider
+        // taking one as an &mut argument
```
didier-wenzek, line 424: Let's go with the current API returning a fresh new `Vec` on each call

didier-wenzek (approved):
Approved. Root issue fixed and nicely implemented.

Issue #4184, opened 2026-05-21
ubuntu 26.04 mosquitto package apparmor settings are incompatible with thin-edge.io mosquitto-conf settings


reubenmiller:
**Describe the bug**

When using thin-edge.io on Ubuntu 26.04, the default "mosquitto" built for this specific Ubuntu version has additional AppArmor configuration when breaks the `include_dir include_dir /etc/tedge/mosquitto-conf` line that the "tedge" .deb package adds to the `/etc/mosquitto/mosquitto.conf`. This results in the mosquitto failing to function correctly as it can't load the `*.conf` files inside the `/etc/tedge/mosquitto-conf/` which contains the listener configuration making the mosquitto service non-functional.

The apparmor mosquitto config problem can be confirmed by looking at the kernel log and for apparmor denied log entries:

```sh
$ sudo grep -i "apparmor" /var/log/kern.log | grep -i "denied"

2026-05-22T09:53:46.318476+02:00 lima-ubuntu-26 kernel: audit: type=1400 audit(1779436426.317:400): apparmor="DENIED" operation="open" class="file" profile="mosquitto" name="/etc/tedge/mosquitto-conf/" pid=51741 comm="mosquitto" requested_mask="r" denied_mask="r" fsuid=0 ouid=999
```

Below shows the default mosquitto apparmor configuration that comes with the most recent mosquitto package:

```sh
cat /etc/apparmor.d/mosquitto
#------------------------------------------------------------------
#    Copyright (C) 2025 Canonical Ltd.
#
#    This program is free software; you can redistribute it and/or
#    modify it under the terms of version 2 of the GNU General Public
#    License published by the Free Software Foundation.
#------------------------------------------------------------------
# vim: ft=apparmor
#
abi <abi/5.0>,

include <tunables/global>

profile mosquitto /usr/sbin/mosquitto {
  include <abstractions/base>
  include <abstractions/nameservice-strict>
  include <abstractions/hosts_access>

  # If run as a root user, drop privileges to mosquitto/nobody/custom-user
  capability setgid,
  capability setuid,

  network inet  stream,
  network inet6 stream,
  network inet  dgram,
  network inet6 dgram,
  network netlink raw,

  file @{run}/.nscd_socket rw,
  file @{run}/nscd/socket rw,

  # nss can be configured to use libvirt in host resolution
  file /var/lib/libvirt/dnsmasq/ r,
  file /var/lib/libvirt/dnsmasq/*.status r,

[...]

PR #4185, opened 2026-05-21
fix: add tedge settings to mosquitto apparmor config

reubenmiller:
## Proposed changes

In Ubuntu 26.04, the mosquitto is packaged with AppArmor settings which result in mosquitto being non-functional. This PR adds some customized AppArmor settings to allow the mosquitto process access to the `/etc/tedge/mosquitto-conf` folder.

The AppArmor `/etc/apparmor.d/local/mosquitto` file (the user settings file), is modified within the postinst maintainer script to add the required rules, and also removes the section when the tedge package is removed. Since this file could be modified by the user, the file can't be part of the nfpm package configuration and needs to be handled dynamically (as it could also be provided by another package).

The tests include a simple smoketests to verify the postinst and postrm logic, as we can't install AppArmor within a docker container as it relies on host functionality, so this level of testing will have to be sufficient for now.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4184

## Checklist

### DIFF
--- a/configuration/package_scripts/tedge/postrm
+++ b/configuration/package_scripts/tedge/postrm
@@ -75,6 +75,22 @@ remove_mosquitto_edit() {
     fi
 }
 
+remove_apparmor_edit() {
+    APPARMOR_LOCAL_MOSQUITTO=/etc/apparmor.d/local/mosquitto
+    if [ -f "$APPARMOR_LOCAL_MOSQUITTO" ]; then
+        sed -i '/# BEGIN tedge/,/# END tedge/d' "$APPARMOR_LOCAL_MOSQUITTO"
+        # Remove the file entirely if nothing else was in it
+        if ! grep -qv '^[[:space:]]*$' "$APPARMOR_LOCAL_MOSQUITTO" 2>/dev/null; then
+            rm -f "$APPARMOR_LOCAL_MOSQUITTO"
+        fi
+        if [ -x /usr/sbin/apparmor_parser ] && [ -d /sys/kernel/security/apparmor ]; then
+            if [ -f /etc/apparmor.d/mosquitto ]; then
+                apparmor_parser -r /etc/apparmor.d/mosquitto || true
+            fi
+        fi
+    fi
+}
+
 purge_var_log() {
     if [ -d "/var/log/tedge" ]; then
         rm -rf /var/log/tedge
@@ -92,6 +108,7 @@ case "$1" in
         remove_user_tedge
         remove_tedge_group
         remove_mosquitto_edit
+        remove_apparmor_edit
         remove_sudoers_file
         purge_configs
         purge_var_log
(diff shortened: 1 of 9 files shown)


configuration/package_scripts/tedge/postrm
```diff
@@ -75,6 +75,21 @@ remove_mosquitto_edit() {
     fi
 }
 
+remove_apparmor_edit() {
+    if [ -f /etc/apparmor.d/local/mosquitto ]; then
+        sed -i '/# BEGIN tedge/,/# END tedge/d' /etc/apparmor.d/local/mosquitto
```
didier-wenzek: We could add `APPARMOR_LOCAL_MOSQUITTO=/etc/apparmor.d/local/mosquitto` to be consistent with `postinst`

configuration/package_scripts/tedge/postrm
```diff
@@ -75,6 +75,21 @@ remove_mosquitto_edit() {
...
+        # Remove the file entirely if nothing else was in it
+        if ! grep -qv '^[[:space:]]*$' /etc/apparmor.d/local/mosquitto 2>/dev/null; then
+            rm -f /etc/apparmor.d/local/mosquitto
```
didier-wenzek: Otherwise the following `apparmor_parser -r` will fail and the updates silently ignored

```suggestion
            apparmor_parser -R /etc/apparmor.d/mosquitto || true
            rm -f /etc/apparmor.d/local/mosquitto
```

reubenmiller: I don't this this is true, as the refresh of the apparmor is triggered on `/etc/apparmor.d/mosquitto` and not `/etc/apparmor.d/local/mosquitto` (the drop-in file).

Or maybe I'm misinterpreting your comment.

reubenmiller: And the apparmor `-r` means replace (reload) the profile, so I think it will then unload the imported local/mosquitto file as it is reloading the entire mosquitto apparmor profile.

```
$ apparmor_parser --help
-r, --replace		Replace apparmor definitions
```

configuration/package_scripts/tedge/postrm
reubenmiller: resolved in 12152a7579ac78be1688deb7d2cf7b94bc9182b0

configuration/package_scripts/tedge/postrm
didier-wenzek: Yeah, I was wrong. Reloading `/etc/apparmor.d/mosquitto` will also reload the local version (aka `/etc/apparmor.d/local/mosquitto`) and the latter doesn't need to be explicitly removed from apparmor definitions.

didier-wenzek (approved):
Approved

PR #4188, opened 2026-05-22
chore: make unit tests macos compatible

albinsuresh:
## Proposed changes

Fix some unit tests that were not compatible with macos, with certain Linux specific assumptions.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/c8y_auth_proxy/src/server.rs
+++ b/crates/extensions/c8y_auth_proxy/src/server.rs
@@ -555,11 +555,11 @@ mod tests {
                 })
             }),
         );
-        (
-            test_app,
-            tokio::time::timeout(Duration::from_secs(5), async move { rx.next().await })
-                .map(|e| e.context("Waiting for ConnectionClosed from server")),
-        )
+        (test_app, async move {
+            tokio::time::timeout(Duration::from_secs(5), rx.next())
+                .await
+                .context("Waiting for ConnectionClosed from server")
+        })
     }
 
     async fn receive_all_messages(mut ws: impl Stream + Unpin) {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -850,7 +850,6 @@ dependencies = [
  "tokio",
  "tokio-rustls",
  "url",
- "whoami",
  "ws_stream_tungstenite 0.15.0",
 ]
 
@@ -4890,7 +4889,6 @@ dependencies = [
  "url",
  "uzers",
  "which",
- "whoami",
  "x509-parser 0.18.0",
  "yansi",
 ]
@@ -5468,7 +5466,6 @@ dependencies = [
  "toml 0.9.6",
  "tracing",
  "tracing-subscriber",
- "whoami",
  "yansi",
 ]
 
@@ -5532,6 +5529,7 @@ dependencies = [
  "camino",
  "tempfile",
  "toml 0.9.6",
+ "uzers",
 ]
 
 [[package]]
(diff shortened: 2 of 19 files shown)


.config/nextest.toml
```diff
@@ -14,3 +14,11 @@ retries = 4
 filter = 'package(c8y_firmware_manager)'
 test-group = 'serial'
 
+[[profile.default.overrides]]
+# c8y_auth_proxy tests spin up TLS proxy + mock servers and do per-test cert
+# generation + TLS handshakes; running many in parallel starves their
+# single-threaded runtimes and the proxy<->c8y handshake times out
+# Hence serializing them.
+filter = 'package(c8y_auth_proxy)'
+test-group = 'serial'
```
albinsuresh: The tests are all individually marked `serial` as well for `cargo test` users like me.

PS: I'm forced to use `cargo test` instead of `nextest` due to its process-per-test model, which doesn't play well with the Crowdstrike falcon agent on my system that scans every spawned process adding an extra 1s delay to each test. With `cargo test` since the number of processes are far lesser, it fares better.

crates/extensions/c8y_auth_proxy/src/server.rs
```diff
@@ -573,6 +574,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
```
jarhodes314: I'm confused as to why these tests need to run serially? There isn't anything that should be causing them to conflict, and indeed, on my machine running `cargo test -p c8y_auth_proxy` on `main` works just fine.

albinsuresh: On my machine with 8 cores, when these tests were all spawned in parallel, they were all getting starved of CPU cycles to make progress presumably due to the heavy load with the keygen, TLS crypto etc. Under such heavy load, these tests were timing out after ~20s even though the test timeouts are only 5s. So, that indicated that the event loops required for these tests to make progress were not getting scheduled as anticipated. So, this is what forced me to mark is `serial` to avoid CPU contention.

TBH it didn't really need serial execution of all tests. When I had the worker threads less than the number of cores (e.g: `--test-threads 4`) the tests were passing. But since there was no easy way to formally set the max number of threads at the test suite level, `serial` was the only option.

didier-wenzek: > On my machine with 8 cores, when these tests were all spawned in parallel, they were all getting starved of CPU cycles to make progress presumably due to the heavy load with the keygen, TLS crypto etc. Under such heavy load, these tests were timing out after ~20s even though the test timeouts are only 5s.

Strange and counter intuitive. It seems that something else than the CPU is blocking. By the way, `--test-threads 32` works as expected on a 16 core linux laptop.

albinsuresh: I had tested with 2, 4, 8 and 16 test threads, and the flaky failures always started with 8 and 100% failure with 16 threads. This observation, combined with tests taking far too long under load, is what lead to this conclusion.

That being said, if this isn't an issue for anyone else in the team while using `cargo test`, then not everyone needs to pay the price for it either. So, I'm open to revert this change and just handle these failure myself locally.

albinsuresh: Here is a more reasonable alternate that's more lenient to the initial setup delays: e5a8ea1

didier-wenzek (approved):
Approved

jarhodes314 (approved):

Issue #4187, opened 2026-05-21
tedge-mapper-c8y does not delete files from file transfer service after log file upload


reubenmiller:
**Describe the bug**

After a log file request is processed by the tedge-mapper-c8y, it doesn't clean up temporary file from the file transfer service when https is being used to secure the endpoint.

Using the tedge-demo-container, the tedge file transfer service is configured to use cert based authentication (using https), the config is shown below:

```sh
$ tedge config list http
http.key_path=/etc/tedge/device-certs/local-tedge.key
http.ca_path=/etc/step-ca/certs
http.bind.port=8000
http.bind.address=0.0.0.0
http.client.port=8000
http.client.host=tedge
http.client.auth.cert_file=/etc/tedge/device-certs/local-tedge.crt
http.client.auth.key_file=/etc/tedge/device-certs/local-tedge.key
```

And in the tedge-mapper-c8y, there is a warning that the removal of the file from the file transfer service (on port 8000) failed due to use a http instead of a https address.

```log
WARN c8y_mapper_ext::operations::upload: Failed to delete log file from file transfer service at http://tedge:8000/te/v1/files/rmi_prereleasecheck_2.0.1/log_upload/dpkg-c8y-mapper-55195886: Failed with HTTP error status 307 Temporary Redirect for DELETE request to endpoint /te/v1/files/rmi_prereleasecheck_2.0.1/log_upload/dpkg-c8y-mapper-55195886
```

The full tedge-mapper-c8y logs related to the log file request are shown below:

```log
May 21 20:39:21 tedge tedge-mapper[20509]:   INFO C8Y: Forwarding c8y operation to device topic=te/device/main///cmd/log_upload/c8y-mapper-55197687 payload={"status":"init","tedgeUrl":"http://tedge:8000/te/v1/files/rmi_prereleasecheck_2.0.1/log_upload/dpkg-c8y-mapper-55197687","type":"dpkg","dateFrom":"2026-05-20T22:39:16+02:00","dateTo":"2026-05-21T22:39:16+02:00","lines":1000}
May 21 20:39:21 tedge tedge-mapper[20509]:   INFO C8Y: Notifying c8y: operation set to EXECUTING topic=te/device/main///cmd/log_upload/c8y-mapper-55197687
[...]

PR #4189, opened 2026-05-22
fix: generate tedgeUrl with https when TLS enabled in file-transfer service

rina23q:
## Proposed changes
#4141 added the http file deletion from the file transfer service after a`config_snapshot` or `log_upload` command completed.

https://github.com/thin-edge/thin-edge.io/blob/dda1013edc2724ef5b57e04b79fc6d2235d842c1/crates/extensions/c8y_mapper_ext/src/operations/handlers/log_upload.rs#L53

This HTTP request works when TLS is disabled in the file transfer service (that's why the existing test passes). However, when TLS is enabled, a warning message appeared.
```
Failed with HTTP error status 307 Temporary Redirect for DELETE request to endpoint
```

This warning (error) indicates that the http redirect to https occurs. The `hyper` backed client doesn't follow the redirect automatically.

Hence, this PR changes `tedgeUrl` generated by tedge-mapper to have a URL with `https://` when `http.cert_path` and `http.key_path` are configured.

## Types of changes

## Paste Link to the issue
#4187

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/http.rs
+++ b/crates/common/tedge_utils/src/http.rs
@@ -1,0 +1,15 @@
+/// The scheme used to connect to an HTTP server.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Protocol {
+    Http,
+    Https,
+}
+
+impl Protocol {
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Http => "http",
+            Self::Https => "https",
+        }
+    }
+}
--- a/crates/core/c8y_api/src/proxy_url.rs
+++ b/crates/core/c8y_api/src/proxy_url.rs
@@ -1,5 +1,7 @@
 use std::sync::Arc;
 
+pub use tedge_utils::http::Protocol;
+
 #[derive(Clone, Debug)]
 pub struct ProxyUrlGenerator {
     host: Arc<str>,
@@ -10,21 +12,6 @@ pub struct ProxyUrlGenerator {
 impl Default for ProxyUrlGenerator {
     fn default() -> Self {
         ProxyUrlGenerator::new("localhost".into(), 8001, Protocol::Http)
-    }
-}
-
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
-pub enum Protocol {
-    Http,
-    Https,
-}
-
-impl Protocol {
-    pub fn as_str(self) -> &'static str {
-        match self {
-            Self::Http => "http",
-            Self::Https => "https",
-        }
     }
 }
 
--- a/crates/core/tedge_agent/src/agent.rs
+++ b/crates/core/tedge_agent/src/agent.rs
@@ -61,6 +61,7 @@ use tedge_mqtt_ext::MqttConfig;
 use tedge_script_ext::ScriptActor;
 use tedge_signal_ext::SignalActor;
 use tedge_uploader_ext::UploaderActor;
+use tedge_utils::http::Protocol;
 use tedge_utils::paths::ManagedDir;
 use tedge_utils::paths::TedgePaths;
 use tracing::info;
@@ -88,6 +89,7 @@ pub(crate) struct AgentConfig {
     pub service_topic_id: ServiceTopicId,
     pub mqtt_topic_root: Arc<str>,
     pub tedge_http_host: Arc<str>,
+    pub tedge_http_protocol: Protocol,
     pub service: TEdgeConfigReaderService,
     pub identity: Option<Identity>,
     pub cloud_root_certs: CloudHttpConfig,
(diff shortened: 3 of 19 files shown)


crates/common/tedge_utils/src/http.rs
```diff
@@ -0,0 +1,15 @@
+/// The scheme used to connect to an HTTP server.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Protocol {
```
albinsuresh, line 3: Wondering if we could just use the standard `http::uri::Scheme` type instead of introducing this new type.

crates/core/c8y_api/src/proxy_url.rs
```diff
@@ -13,21 +15,6 @@ impl Default for ProxyUrlGenerator {
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
albinsuresh, line 16: Aah I see that you just refactored this type out and not introduce it.

tests/RobotFramework/tests/cumulocity/log/generate_certificates.sh
```diff
@@ -0,0 +1,120 @@
+#!/bin/sh
```
albinsuresh, line 1: Could we not move the existing `generate_certificates.sh‎` to a common location and reuse it?

tests/RobotFramework/tests/cumulocity/log/log_operation_with_file_transfer_https.robot
```diff
@@ -0,0 +1,74 @@
...
+
+Test Tags           theme:c8y    theme:log
+
+
+*** Variables ***
+${DEVICE_SN}    ${EMPTY}
+
+
+*** Test Cases ***
+Log upload operation succeeds when File Transfer Service has TLS enabled
```
albinsuresh, line 21: Not for this PR, but I just wonder if there was some Robot framework trick that we could use to run the same test against multiple setups like with SSL enabled/disabled, to avoid test logic duplication like this. I understand that we have followed this duplication pattern in other tests as well. Hence I feel we could explore if there's a better way.

albinsuresh (approved):
LGTM. Just a few suggestions on test organisational hygiene.

crates/common/tedge_utils/src/http.rs
didier-wenzek: This `http::Protocol` is just glorified strings, adding as such no value to `tedge_utils`.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -132,6 +134,13 @@ impl AgentConfig {
         let tedge_http_address = tedge_config.http.client.host.clone();
         let tedge_http_port = tedge_config.http.client.port;
         let tedge_http_host = format!("{}:{}", tedge_http_address, tedge_http_port).into();
+        let tedge_http_protocol = if tedge_config.http.cert_path.or_none().is_some()
+            && tedge_config.http.key_path.or_none().is_some()
+        {
+            Protocol::Https
+        } else {
+            Protocol::Http
+        };
```
didier-wenzek, line 143: Those lines bring more value than the thin abstraction provided by `tedge_utils::http::protocol`. They tell when we have to use TLS or not, and should not be repeated for the mapper and the agent.

Would `tedge_config` be a better location?

crates/extensions/c8y_mapper_ext/src/operations/handlers/config_snapshot.rs
```diff
@@ -95,7 +95,8 @@ impl OperationContext {
             Some(ref tedge_file_url) => Cow::Borrowed(tedge_file_url),
             None => {
                 let tedge_file_url = format!(
-                    "http://{}/te/v1/files/{external_id}/config_snapshot/{config_filename}",
+                    "{}://{}/te/v1/files/{external_id}/config_snapshot/{config_filename}",
+                    self.tedge_http_protocol.as_str(),
```
didier-wenzek, line 99: Here is the heart of the issue and hence of the abstraction that could be introduced.

The current fix is direct, replacing the hard-coded `http` protocol with a dynamic property *string*. However, how this property is derived (from tedge config) and used (to connect the file transfer service) is scattered in the agent, the mapper, the config manager.

We could do better. What has to be abstracted is the file transfer service and its urls.

didier-wenzek (approved):
Approved, because this fix the issue. However, it would be good to introduce a better abstraction once the 2.0.1 released.

reubenmiller (approved):
Approved. The fix has also been verified on the tedge demo container where the original bug was reported from.

crates/core/c8y_api/src/proxy_url.rs
rina23q: Yes. Since this was under `c8y_api`, I wanted to make it accessible by the crates that has no dependency on c8y.

rina23q:
There are several points that can polish the code. I will address in a follow-up PR.

crates/core/tedge_agent/src/agent.rs
rina23q: Yeah, I actually considered `tedge_config` while preparing this PR as the condition if using https or http is dependent on tedge config settings. Probably it's the right location.

Issue #4186, opened 2026-05-21
custom child device names are reverted when restarting the tedge-agent


reubenmiller:
**Describe the bug**

Custom child device names aren't consistently handled and are reset after the `tedge-agent` service is restarted.

It's possible that the regression is due to some recent work around the entity store fixes:

* https://github.com/thin-edge/thin-edge.io/pull/4183
* https://github.com/thin-edge/thin-edge.io/pull/4173

**To Reproduce**

The behaviour can be reproduced three different ways:

#### Scenario 1: Register a child device and change the twin/name value

1. Register a child device

    ```sh
    tedge mqtt pub "te/device/child100//" '{"@type":"child-device"}' --retain
    ```

    Cumulocity name afterwards: `example:device:child100`

2. Change the name of the child device

    ```sh
    tedge mqtt pub te/device/child100///twin/name '"child100-custom"' -r
    ```

    Cumulocity name afterwards: `child100-custom`

3. Restart the tedge-agent

    ```sh
    sudo systemctl restart tedge-agent
    ```

    Cumulocity name afterwards: `child100`

#### Scenario 2: Register a child device and re-register it with a custom name

1. Register a child device

    ```sh
    tedge mqtt pub "te/device/child101//" '{"@type":"child-device"}' --retain
    ```

    Cumulocity name afterwards: `rmi_prereleasecheck_2.0.1:device:child101`

2. Register a child device and specify a custom name

    ```
    tedge mqtt pub "te/device/child101//" "{\"@parent\":\"device/main//\",\"@type\":\"child-device\",\"name\":\"child101-custom\"}" --retain
    ```

    Cumulocity name afterwards: `child101-custom`

3. Restart the tedge-agent

    ```sh
    sudo systemctl restart tedge-agent
    ```

    Cumulocity name afterwards: `child101`

#### Scenario 3: Register a child device with a custom name

1. Register a child device with a customer name

    ```sh
    tedge mqtt pub "te/device/child103//" "{\"@parent\":\"device/main//\",\"@type\":\"child-device\",\"name\":\"child103-custom\"}" --retain
    ```

    Cumulocity name afterwards: `child103-custom`

2. Restart the tedge-agent

    ```sh
    sudo systemctl restart tedge-agent
    ```

    Cumulocity name afterwards: `child103`

**Expected behavior**

If the user opts in for a custom child device name, then it shouldn't change after the `tedge-agent` service is restarted.

**Screenshots**

[...]

PR #4190, opened 2026-05-26
fix: disable entity store clean_start of tedge-agent

albinsuresh:
## Proposed changes

Disable entity store clean_start of tedge-agent by default. This reduces the amount of message races that the agent has to deal with on startup, when all entity registrations and their data messages are delivered to the agent.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4186

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config.rs
@@ -1242,8 +1242,9 @@ define_tedge_config! {
             #[tedge_config(example = "true", default(value = true), deprecated_key = "c8y.entity_store.auto_register")]
             auto_register: bool,
 
-            /// On a clean start, the whole state of the device, services and child-devices is resent to the cloud
-            #[tedge_config(example = "true", default(value = true), deprecated_key = "c8y.entity_store.clean_start")]
+            /// When set to `false`, the entity store is persisted and restored over agent restarts.
+            /// When set to `true`, the entity store is re-created from the retained entity registration MQTT messages.
+            #[tedge_config(example = "true", default(value = false), deprecated_key = "c8y.entity_store.clean_start")]
             clean_start: bool,
         },
 
--- a/tests/RobotFramework/tests/tedge_agent/entity_store_persistence.robot
+++ b/tests/RobotFramework/tests/tedge_agent/entity_store_persistence.robot
@@ -1,0 +1,180 @@
+*** Settings ***
+Resource            ../../resources/common.resource
+Library             Cumulocity
+Library             ThinEdgeIO
+
+Test Setup          Custom Setup
+Test Teardown       Get Logs    ${DEVICE_SN}
+
+Test Tags           theme:c8y    theme:registration    theme:deregistration
+
+
+*** Variables ***
+${DEVICE_SN}    ${EMPTY}
+${CHILD_SN}     ${EMPTY}
+${CHILD_XID}    ${EMPTY}
+
+
+*** Test Cases ***
+Twin fragment in registration message survives agent restart
+    Execute Command
+    ...    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device","name":"${CHILD_SN}-custom"}'
+    Device Should Exist    ${CHILD_XID}
+    Device Should Have Fragment Values    name\=${CHILD_SN}-custom
+
+    # Re-publish registration with a different name — must NOT clobber the user's value
+    Restart Service    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
+
+    Should Have Retained MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/name
+    ...    message_contains=${CHILD_SN}-custom
+    Device Should Have Fragment Values    name\=${CHILD_SN}-custom
+
+Twin fragment set via twin topic survives agent restart
+    Execute Command
+    ...    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device","@id":"${CHILD_XID}"}'
+    Device Should Exist    ${CHILD_XID}
+    Device Should Have Fragment Values    name\=${DEVICE_SN}:device:${CHILD_SN}
+
+    Execute Command
+    ...    tedge mqtt pub --retain "te/device/${CHILD_SN}///twin/name" '"${CHILD_SN}-custom"'
+    Device Should Have Fragment Values    name\=${CHILD_SN}-custom
+
+    Restart Service    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
+
+    Should Have Retained MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/name
+    ...    message_contains=${CHILD_SN}-custom
+    Device Should Have Fragment Values    name\=${CHILD_SN}-custom
+
+Twin fragment in registration message updated via twin topic survives agent restart
+    Skip
+    ...    msg=When the registration message is re-delivered to the agent on startup, the twin value in it is accepted as the latest twin value, which causes the updated twin value to be lost.
+
+    Execute Command
+    ...    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device","@id":"${CHILD_XID}","name":"${CHILD_SN}-initial"}'
+    Device Should Exist    ${CHILD_XID}
+    Device Should Have Fragment Values    name\=${CHILD_SN}-initial
+
+    Execute Command
+    ...    tedge mqtt pub --retain "te/device/${CHILD_SN}///twin/name" '"${CHILD_SN}-updated"'
+    Device Should Have Fragment Values    name\=${CHILD_SN}-updated
+
+    Restart Service    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
+
+    Should Have Retained MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/name
+    ...    message_contains=${CHILD_SN}-updated
+    Device Should Have Fragment Values    name\=${CHILD_SN}-updated
+
+Twin fragment in registration message updated via re-registration survives agent restart
+    Execute Command
+    ...    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device","@id":"${CHILD_XID}","name":"${CHILD_SN}-initial"}'
+    Device Should Exist    ${CHILD_XID}
+    Device Should Have Fragment Values    name\=${CHILD_SN}-initial
+
+    # Re-publish registration with a different name
+    Execute Command
+    ...    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device","@id":"${CHILD_XID}","name":"${CHILD_SN}-updated"}'
+    Device Should Have Fragment Values    name\=${CHILD_SN}-updated
+
+    Restart Service    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
+
+    Should Have Retained MQTT Messages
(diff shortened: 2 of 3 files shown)


crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -1242,8 +1242,9 @@ define_tedge_config! {
             #[tedge_config(example = "true", default(value = true), deprecated_key = "c8y.entity_store.auto_register")]
             auto_register: bool,
 
-            /// On a clean start, the whole state of the device, services and child-devices is resent to the cloud
-            #[tedge_config(example = "true", default(value = true), deprecated_key = "c8y.entity_store.clean_start")]
+            /// When set to `false`, the entity store is persisted and restored over agent restarts.
+            /// When set to `true`, the entity store is re-created from the retained entity registration MQTT messages.
+            #[tedge_config(example = "true", default(value = false), deprecated_key = "c8y.entity_store.clean_start")]
```
rina23q, line 1247: I think the deprecation of `c8y.entity_store.*` is incomplete.
We have to 
1. remove the old keys `c8y.entity_store.auto_register` and `c8y.entity_store.clean_start` from this file and
2. add toml migration steps to `TedgeTomlVersion::steps()` 
Afterwards, finally this `deprecatred_key` macro takes care of the old keys.

Also, if we remove the old keys, c8y-mapper needs a small cleanup. I found there are still construction of the two old keys, which are no longer used.

=> Created a PR #4191

tests/RobotFramework/tests/tedge_agent/entity_store_persistence.robot
```diff
@@ -0,0 +1,99 @@
...
+    Execute Command
+    ...    tedge mqtt pub --retain "te/device/${CHILD_SN}///twin/name" '"${CHILD_SN}-updated"'
+    Device Should Have Fragment Values    name\=${CHILD_SN}-updated
+
+    Restart Service    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
+
+    Should Have Retained MQTT Messages
+    ...    te/device/${CHILD_SN}///twin/name
+    ...    message_contains=${CHILD_SN}-updated
```
albinsuresh, line 70: This test is still flaky as the twin value might become `${CHILD_SN}-initial` if the following sequence of events happen:

1. Before the agent restart, the broker has the following two retained messages:
    ```
    [te/device/child100//] {"@type":"child-device","@id":"main_device:device:child100","name":"child100-initial"}
    [te/device/child100///twin/name] "child100-updated"
2. On agent restart, the agent recreates the entity store from the persistent `entity_store.jsonl` file. But since we don't persist the `twin` messages anymore, the in-memory entity store would have the old name in it, which it picks up from `name` value in the registration message.
3. Now, if the twin message is delivered to the agent, before the registration message, the name will get updated from `child100-initial` to `child100-updated`.
4. Now, when the registration message is delivered afterwards, it will carry the old `name: child100-initial` and the agent would interpret this as a `name` update and update the name back from `child100-updated` to `child100-initial`.

This is the reason why I was saying that the `clean_start` flag alone won't be enough. Either we revert the change that stopped persisting the `twin` messages in the persistent `entity_store.jsonl` file or stop accepting `twin` updates via secondary registration messages.

didier-wenzek: > Either we revert the change that stopped persisting the `twin` messages in the persistent `entity_store.jsonl` file

This would be the most conservative and safest solution. Trying to be factual and honest without blaming anyone, the current design is so brittle that any move forward (such as no more accepting twin updates via secondary registration messages) is at risk of introducing yet another unexpected side effect. A redesign is required, but that's not the time.

reubenmiller: I'm not very keen to use the entity_store.jsonl to keep all twin information though...we could potentially limit it to `twin/name` and `twin/type` data, as that would prevent issues with conflict between name/type info being defined in the registration messages and updated via the twin topic right?

albinsuresh: No, it's not just limited to `name` and `type`. A user is free to include any/all of his twin data in the original registration message as follows:

```
tedge mqtt pub -r "te/device/child100//" '{"@type":"child-device", "x": "X", "y": "Y", "z": "Z"}'
```

Now the user could have updated any of these twin values via their `twin/` topics and end up with the same bug as described above on a restart.

Introducing an additional inconsistency in behaviour between `name/type` and other twin data would only add to the confusion, right? The system is already fragile. Having more `if-else` branches would complicate it even more, both for us and the users.

reubenmiller: I wasn't fully ware that that is the case, but I don't think we need to take care of those cases...so I would be ok if we only listen and store messages on the /twin/name and /twin/type as these have special meanings anyway.

albinsuresh: Unfortunately, persisting the `name` and `type` did not fix the issue, as the test can still fail with a different sequence of events:

1. Before the agent restart, the broker has the following two retained messages:
    ```
    [te/device/child100//] {"@type":"child-device","@id":"main_device:device:child100","name":"child100-initial"}
    [te/device/child100///twin/name] "child100-updated"
2. On agent restart, the agent recreates the entity store from the persistent `entity_store.jsonl` file. With the selective persistence of `name` and `twin` logic, the in-memory entity store would have the latest name (`child100-updated`) in it.
3. If the `twin` message gets delivered first, it leaves the `name` unchanged as the in-memory value is already the latest.
4. But now, when the registration message is delivered afterwards, it will carry the old name: `child100-initial` and the agent would interpret this as a name update and update the name back from `child100-updated` to `child100-initial`.

So, I still feel, rejecting any twin updates via secondary registration messages would still be the better. It would avoid inconsistencies where the `name` value is updated to different values, both via the registration message as well as twin message when the agent is down, and then on restart, the agent won't be able to distinguish which one is the latest. So, we'll keep `twin/` topic as the sole interface for twin updates. The twin messages in registration messages are only considered during the initial registration and not afterwards. Hope that would satisfy most users.

Is that acceptable? Since that might be considered a behavioural change, if anyone was relying on that behaviour.

didier-wenzek: > rejecting any twin updates via secondary registration [*retained*] messages would still be the better.

This sentence highlights several design mistakes we made:

- Using MQTT *retained* messages for partial updates. As with `HTTP PUT` requests, MQTT retained messages must *only* be used to set the whole content of a value for a key. 
- Using two different channels to publish the `name` and the `type` of an entity. These data should be published only on a specific channel (entity registration *or* twin data).
- Having no clear understanding of the purpose of the `name` and `type` of an entity. Are these mandatory properties (like `@parent` and `@type`), or info (like a `location` or `time-zone` info). 

This is why I think your proposal is going in the wrong direction, just adding corner cases. What does it mean for a retained message to be "secondary". How, as a user, can I know that a message has already been seen by the agent.

albinsuresh: > What does it mean for a retained message to be "secondary"

By secondary, I meant any registration messages that are sent to replace the existing registration message of an entity that is already registered.

> How, as a user, can I know that a message has already been seen by the agent.

Just checking if a retained message already exists on the target topic isn't good enough, as that could be a bad registration message that the agent may have rejected. So, the only authoritative way is to use the HTTP API to query of an entity with the target topic id exists already or not.

> Using MQTT retained messages for partial updates. As with HTTP PUT requests, MQTT retained messages must only be used to set the whole content of a value for a key.

Agree. We should not have supported entity updates over the same registration channel. Unfortunately, it was provided before the `twin` era and we were forced to provide backward compatibility.

> Using two different channels to publish the name and the type of an entity. These data should be published only on a specific channel (entity registration or twin data).

Again, agree. Even when we started accepting twin data via registration messages, and published `twin/` messages for all of them on behalf of the user, we probably should have replaced the registration message as well, by stripping the twin data fragments, as they are already present in the twin topics. So, from then on, that `twin/` topic is the sole source of truth. But, would it have been okay to replace a registration message sent by the user with the agent's view of it? I'm not sure. Again highlights the issue with using the same channel as the source-of-truth as well as for user requests.

> Having no clear understanding of the purpose of the name and type of an entity. Are these mandatory properties (like @parent and @type), or info (like a location or time-zone info).

At least the agent's PoV was that they are optional fields as the agent doesn't care if an entity has a `name` and `type` or not. But the c8y mapper cares. But then, it should have been responsible for generating any sensible defaults when the user hand't provided one. Hence, the agent generating a `name` for auto-registered entities was one more mistake.

tests/RobotFramework/tests/tedge_agent/entity_store_persistence.robot
```diff
@@ -0,0 +1,141 @@
...
+    Device Should Exist    ${CHILD_XID}
+    Device Should Have Fragment Values    name\=${DEVICE_SN}:device:${CHILD_SN}
+
+    Execute Command
+    ...    tedge mqtt pub --retain "te/device/${CHILD_SN}///twin/name" '"${CHILD_SN}-custom"'
+    Device Should Have Fragment Values    name\=${CHILD_SN}-custom
+
+    Restart Service    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
```
reubenmiller: Duplicate line that could be deleted (I guess)

tests/RobotFramework/tests/tedge_agent/entity_store_persistence.robot
```diff
@@ -0,0 +1,141 @@
...
+
+    Execute Command
+    ...    tedge mqtt pub --retain 'te/device/${CHILD_SN}//' '{"@type":"child-device","@id":"${CHILD_XID}","name":"${CHILD_SN}-updated"}'
+    Execute Command
+    ...    tedge mqtt pub --retain "te/device/${CHILD_SN}///twin/name" '"${CHILD_SN}-custom"'
+
+    Restart Service    tedge-agent
+    Service Health Status Should Be Up    tedge-agent
+
+    Should Have Retained MQTT Messages
```
reubenmiller, line 108: I'm a bit surprised about this assertion. I would of expected that the value coming from the `/twin/name` would win, so the expected value would be "${CHILD_SN}-custom".

But if that is just how it is now, then we can live with it. But I guess the `/twin/name` is overwritten due to the registration message being processed again and resulting in a new `/twin/name` message.

albinsuresh: > But I guess the /twin/name is overwritten due to the registration message being processed again and resulting in a new /twin/name message.

Exactly. On restart, when the registration message with the `name` in it is processed, that `twin/name` value is published back to the broker, which would overwrite the `${CHILD_SN}-custom`. I also agree that I don't like this behaviour. But at least it is deterministic in who would win.

tests/RobotFramework/tests/tedge_agent/entity_store_persistence.robot
albinsuresh: Resolved in [aa33f97](https://github.com/thin-edge/thin-edge.io/pull/4190/commits/aa33f977a7d163a1d6cde276b4939ad56f24b448)

reubenmiller (approved):
This solves the most common cases, and the remaining indeterminate behaviour we can live with for now as we decided we anyway need a re-think on how this should work in the future.

didier-wenzek (approved):
Approved. Even if the fix is partial, we have now a better understanding of the entanglement between entity registrations, twin data updates, agent restart and retained messages.

PR #4191, opened 2026-05-26
fix: complete deprecation of c8y.entity_store.* keys

rina23q:
## Proposed changes

The deprecation of
* `c8y.entity_store.auto_register`
* `c8y.entity_store.clean_start`
was partially done. That's why it didn't mask those deprecated keys in `tedge config list`.

This PR finalizes the deprecation of the keys. Below are examples of CLI output:

```sh
# No `c8y.entity_store.*` listed
root@60dca65925f8:/setup# tedge config list entity --all
agent.entity_store.auto_register=true
agent.entity_store.clean_start=false
```

```sh
# Deprecation announcement in the output
root@60dca65925f8:/setup# tedge config get c8y.entity_store.auto_register
2026-05-26T15:06:27.739503035Z  WARN tedge_config::tedge_toml::tedge_config: The key 'c8y.entity_store.auto_register' is deprecated. Use 'agent.entity_store.auto_register' instead.
true
```

## Types of changes

## Paste Link to the issue
Found during the review of https://github.com/thin-edge/thin-edge.io/pull/4190#discussion_r3303717385

## Checklist

### DIFF
--- a/crates/common/tedge_config/src/tedge_toml/tedge_config/version.rs
+++ b/crates/common/tedge_config/src/tedge_toml/tedge_config/version.rs
@@ -196,10 +196,18 @@ impl TEdgeTomlVersion {
                     value: Self::Two.into(),
                 },
             ],
-            Self::Two => vec![TomlMigrationStep::MoveKey {
-                original: "apt.dpk",
-                target: "apt.dpkg".into(),
-            }],
+            Self::Two => vec![
+                TomlMigrationStep::MoveKey {
+                    original: "apt.dpk",
+                    target: "apt.dpkg".into(),
+                },
+                mv(
+                    "c8y.entity_store.auto_register",
+                    AgentEntityStoreAutoRegister,
+                ),
+                mv("c8y.entity_store.clean_start", AgentEntityStoreCleanStart),
+                rm("c8y.entity_store"),
+            ],
         }
     }
 
@@ -368,5 +376,42 @@ mod tests {
             Some("keepold")
         );
         assert!(migrated["apt"].as_table().unwrap().get("dpk").is_none());
+    }
+
+    #[test]
+    fn v2_migration_moves_c8y_entity_store_to_agent_entity_store() {
+        let input = toml::toml!(
+            [config]
+            version = "2"
+
+            [c8y.entity_store]
+            auto_register = false
+            clean_start = false
+        );
+        let (migrated, any_keys_migrated) = TEdgeTomlVersion::Two
+            .migrations()
+            .try_fold(
+                (toml::Value::Table(input), false),
+                |(toml, any_migrated), step| {
+                    let (toml, step_migrated) = step.apply_to(toml)?;
+                    Ok::<_, String>((toml, any_migrated || step_migrated))
+                },
+            )
+            .unwrap();
+
+        assert!(any_keys_migrated);
+        assert_eq!(
+            migrated["agent"]["entity_store"]["auto_register"].as_bool(),
+            Some(false)
+        );
+        assert_eq!(
+            migrated["agent"]["entity_store"]["clean_start"].as_bool(),
+            Some(false)
+        );
+        assert!(migrated["c8y"]
+            .as_table()
+            .unwrap()
+            .get("entity_store")
+            .is_none());
     }
 }
--- a/tests/RobotFramework/tests/tedge/tedge_config_deprecated_keys.robot
+++ b/tests/RobotFramework/tests/tedge/tedge_config_deprecated_keys.robot
@@ -2,8 +2,8 @@
 Resource            ../../resources/common.resource
 Library             ThinEdgeIO
 
-Test Setup         Custom Setup
-Test Teardown      Get Logs
+Test Setup          Custom Setup
+Test Teardown       Get Logs
 
 Test Tags           theme:cli    theme:configuration
 
@@ -41,7 +41,7 @@ Migrate c8y.entity_store.* keys to agent.entity_store.* keys 1.4->1.5->2.0.0->la
     # The values from agent.entity_store.* keys should be effective, not the values from c8y.entity_store.* keys
     ${auto_register}=    Execute Command    tedge config get agent.entity_store.auto_register    strip=True
     Should Be Equal    ${auto_register}    true
-    ${clean_start}=    Execute Command    tedge config get agent.entity_store.clean_start   strip=True
+    ${clean_start}=    Execute Command    tedge config get agent.entity_store.clean_start    strip=True
     Should Be Equal    ${clean_start}    true
 
     # Update to the test version
--- a/tests/RobotFramework/tests/tedge/tedge_config_deprecated_keys.robot
+++ b/tests/RobotFramework/tests/tedge/tedge_config_deprecated_keys.robot
@@ -1,0 +1,94 @@
+*** Settings ***
+Resource            ../../resources/common.resource
+Library             ThinEdgeIO
+
+Test Setup         Custom Setup
+Test Teardown      Get Logs
+
+Test Tags           theme:cli    theme:configuration
+
+
+*** Test Cases ***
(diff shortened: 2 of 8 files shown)


crates/common/tedge_config/src/tedge_toml/tedge_config/version.rs
```diff
@@ -196,10 +196,18 @@ impl TEdgeTomlVersion {
...
-                target: "apt.dpkg".into(),
-            }],
+            Self::Two => vec![
+                TomlMigrationStep::MoveKey {
+                    original: "apt.dpk",
+                    target: "apt.dpkg".into(),
+                },
+                mv(
+                    "c8y.entity_store.auto_register",
+                    AgentEntityStoreAutoRegister,
```
didier-wenzek: I'm not sure the purpose is the same - despite similar names. @albinsuresh can you confirm?

albinsuresh: While migrating a configuration that only had `c8y.entity_store.auto_register` in it (a pre 1.5 configuration) and not the `agent.entity_store.auto_register` to the current version, then this is fine, as the user would expect the same auto-registration behaviour even though the responsibility moved from the mapper to agent.

But, if a configuration has both `c8y.entity_store.auto_register` (stale since 1.5) and `agent.entity_store.auto_register` settings, then I hope this migration wouldn't happen, overwriting the existing `agent.entity_store.auto_register` value with what was there in `c8y.entity_store.auto_register`. That would be wrong. We might have to add some test coverage to ensure that this isn't happening.

rina23q: The latter case is a good point. It will not overwrite, parsing an error. 

```
"Cannot migrate 'c8y.entity_store.auto_register' to 'agent.entity_store.auto_register': target key already exists. Please manually remove one of them from tedge.toml."
```

This is actually bad. In our normal deprecation scenario, this case cannot happen like `apt.dpk` => `apt.dpkg`, as always either old name or new name exists. But this case is a special since both configuration could exists in `tedge.toml`.  

Since it's `Err`, if this collision occurs, process crashes.
```
May 27 14:24:26 b27a2e6dfdd8 tedge[1427]:   WARN tedge_config::tedge_toml::tedge_config: The key 'c8y.entity_store.auto_register' is deprecated. Use 'agent.entity_store.auto_register' instead.
May 27 14:24:26 b27a2e6dfdd8 tedge[1427]:   WARN tedge_config::tedge_toml::tedge_config: The key 'c8y.entity_store.clean_start' is deprecated. Use 'agent.entity_store.clean_start' instead.
May 27 14:24:26 b27a2e6dfdd8 tedge[1427]: Error: Cannot migrate 'c8y.entity_store.auto_register' to 'agent.entity_store.auto_register': target key already exists. Please manually remove one of them from tedge.toml.
```

---

Wait. It could be the situation "both old and new config exist" cannot happen unless user modifies `tedge.toml` directly. 

---

###  ✅ Configured `c8y.entity_store.*` in version >= 1.5.0 
It's safe, as CLI changes `agent.entity_store.*` values. 
```
root@84f68b96597e:/etc/tedge# tedge config set c8y.entity_store.clean_start false
root@84f68b96597e:/etc/tedge# tedge config set c8y.entity_store.auto_register false
root@84f68b96597e:/etc/tedge# cat /etc/tedge/tedge.toml 
[agent.entity_store]
auto_register = false
clean_start = false
```  

### ❌   Configured `c8y.entity_store.*` in version <= 1.4.0 and changed it in version 1.5.0
* CLI modifies values of `agent.entity_store.*` even if you set for `c8y.entity_store.*`.
* `[c8y.entity_store]` left in `tedge.toml` - like a **zombie**, cannot be set or remove by CLI

#### current behaviour - no toml migration
* after upgrading to 2.0.0, this zombie config migrated to `/etc/tedge/mappers/c8y/mappper.toml`.
```
root@84f68b96597e:/etc/tedge# cat /etc/tedge/mappers/c8y/mapper.toml 
cloud_type = "c8y"
url = "your-tenant.c8y.io"

[entity_store]
auto_register = false
clean_start = false
```
* Restarting `tedge-agent` or `tedge-mapper-c8y` doesn't fail. It seems simply the zombie config are silently ignored.

#### this PR's behaviour - with toml migration
* Migration fails:
```
root@ef7801c9c6a5:/etc/tedge# tedge config upgrade
2026-05-27T15:38:30.760603116Z  WARN tedge_config::tedge_toml::tedge_config: The key 'c8y.entity_store.auto_register' is deprecated. Use 'agent.entity_store.auto_register' instead.
2026-05-27T15:38:30.760631949Z  WARN tedge_config::tedge_toml::tedge_config: The key 'c8y.entity_store.clean_start' is deprecated. Use 'agent.entity_store.clean_start' instead.
Error: Cannot migrate 'c8y.entity_store.auto_register' to 'agent.entity_store.auto_register': target key already exists. Please manually remove one of them from tedge.toml.
```
* Also, `tedge-agent` and `tedge-mapper-c8y` fails to start.

albinsuresh: Yeah, that's not good, because the migration happens as part of version upgrade AFAIK. So, we need a mitigation strategy.

I'm also wondering if such a key migration warrants a change of version of the config itself.

rina23q: In short, I should revert the migration part. That will leave the zombie config in the toml file, but it won't hurt. Or, I should give a patch for only for this specific keys, if both `agent.xxx` and `c8y.xxx` exists, delete the `c8y.xxx`.

albinsuresh: > Or, I should give a patch for only for this specific keys, if both agent.xxx and c8y.xxx exists, delete the c8y.xxx

That would be ideal.

rina23q: Currently, if the target key has already exists, it returns an error. If I change this from error to warning, is it too brutal? This solves this problem.

`crates/common/tedge_config/src/tedge_toml/tedge_config/version.rs`
```rust
impl TomlMigrationStep {
    pub fn apply_to(self, mut toml: toml::Value) -> Result<(toml::Value, bool), String> {
        match self {
            TomlMigrationStep::MoveKey { original, target } => {
                    ...
                    if table.contains_key(target_field) {
                        return Err(format!(
                            "Cannot migrate '{original}' to '{target}': target key already \
                             exists. Please manually remove one of them from tedge.toml."
                        ));
                    }
   ...
}
```

tests/RobotFramework/tests/tedge/tedge_config_deprecated_keys.robot
```diff
@@ -0,0 +1,94 @@
...
+    # If tedge config migration error happens, these services will fail to start
+    Service Health Status Should Be Up    tedge-agent
+    Service Health Status Should Be Up    tedge-mapper-c8y
+
+    # The zombie config is gone
+    ${output_c8y_mapper_toml}=    Execute Command    cat /etc/tedge/mappers/c8y/mapper.toml
+    Should Not Contain    ${output_c8y_mapper_toml}    auto_register = false
+    Should Not Contain    ${output_c8y_mapper_toml}    clean_start = false
+
+Migrate c8y.entity_store.* keys to agent.entity_store.* keys 1.4->1.5->latest
```
rina23q, line 62: Currently, this test is supposed to fail due to the migration issue discussed in https://github.com/thin-edge/thin-edge.io/pull/4191#discussion_r3306162489.

rina23q:
It seems more complicated than initially I thought. I would close this PR and further work will be split into:

- a new issue #4193: to find a proper way to deprecate the obsolete `c8y.entity_store.*` keys
- a new PR #4194: to extract the refactoring part from this PR to remove unused code in c8y-mapper.

PR #4196, opened 2026-05-29
fix: tedge init ensures all configured paths exist

didier-wenzek:
## Proposed changes

`tedge init` creates all missing path components - *even* those above tedge root.

-  E.g. ensure `/var/log`  exists, before creating `/var/log/tedge`
-  While folders below tedge root (e.g. `/etc/tedge/mappers`) are `chown` to be owned by the configured tedge user
-  folders above tedge root (e.g. `/var/log`) are owned by the user running `tedge init` (root in practice)

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/tedge_utils/src/file.rs
+++ b/crates/common/tedge_utils/src/file.rs
@@ -242,6 +242,9 @@ impl PermissionEntry {
         dir: &Path,
         root: Option<&Path>,
     ) -> Result<(), FileError> {
+        if let Some(root_dir) = root {
+            ensure_parents_exist(root_dir).await?;
+        }
         match dir.parent() {
             None => return Ok(()),
             Some(_parent) if Some(dir) == root => {}
@@ -278,6 +281,24 @@ impl PermissionEntry {
             }),
         }
     }
+}
+
+async fn ensure_parents_exist(dir: &Path) -> Result<(), FileError> {
+    if let Some(parent) = dir.parent() {
+        if !path_exists(parent).await {
+            Box::pin(ensure_parents_exist(parent)).await?;
+
+            if let Err(e) = fs::create_dir(&parent).await {
+                if e.kind() != io::ErrorKind::AlreadyExists {
+                    return Err(FileError::DirectoryCreateFailed {
+                        dir: parent.display().to_string(),
+                        from: e,
+                    });
+                }
+            }
+        }
+    }
+    Ok(())
 }
 
 /// Overwrite the content of existing file. The file permissions will be kept.
--- a/crates/common/tedge_utils/src/paths.rs
+++ b/crates/common/tedge_utils/src/paths.rs
@@ -601,7 +601,6 @@ pub fn ok_if_not_found(err: std::io::Error) -> std::io::Result<()> {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::file::FileError;
     use nix::unistd::Uid;
     use std::os::unix::fs::PermissionsExt;
     use tedge_test_utils::fs::TempTedgeDir;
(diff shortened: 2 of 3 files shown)


crates/common/tedge_utils/src/file.rs
```diff
@@ -242,6 +242,9 @@ impl PermissionEntry {
         dir: &Path,
         root: Option<&Path>,
     ) -> Result<(), FileError> {
+        if let Some(root_dir) = root {
+            ensure_parents_exist(root_dir).await?;
```
rina23q, line 246: It fixes the discussed issue.

However, as the unit test failure `ensure_does_not_create_directories_above_the_root()` tells, this change repurposes the API. Previously, `ManagedDir::ensure()` does not create the directories above the managed root directory on purpose. 
I think it's better to provide the API covers both patterns: 1) not create directories above the root, 2) create directories above the root. Probably only `tedge init` wants the pattern 2).

So what I expected was something like below (name can be changed ofc):
```rust
let root = ttd.utf8_path().join("missing-parent").join("managed-root");
let config_root = TedgePaths::from_root_with_defaults(&root, "", "");

let managed_dir = config_root.dir("operations/c8y").unwrap();

managed_dir.ensure().await.unwrap(); // <- default doesn't ensure the parents above the root
managed_dir.ensure_above_root().ensure().await.unwrap(); // <- ensure the parents above the root
```

crates/common/tedge_utils/src/file.rs
```diff
@@ -280,6 +283,24 @@ impl PermissionEntry {
...
+                    return Err(FileError::DirectoryCreateFailed {
+                        dir: parent.display().to_string(),
+                        from: e,
+                    });
+                }
+            }
+        }
+    }
+    Ok(())
+}
```
Bravo555, line 302: As I understand this is just `fs::create_dir_all`, but if `create_dir_all` can't create some directory it can't report _which one_ because `io::Error` doesn't have space for it, so makes sense to do it this way.

Bravo555 (approved):

crates/common/tedge_utils/src/file.rs
Bravo555: > I think it's better to provide the API covers both patterns: 1) not create directories above the root, 2) create directories above the root. Probably only tedge init wants the pattern 2).

If we think of `tedge init` as a shortcut that creates all the required paths defined in `tedge.toml`, that's probably correct. If that's the approach, we could make the API for `tedge init` specifically more separate, such that normal tedge services don't use it by mistake. 
But it also isn't that bad if some service running as `tedge` attempts to create any directories if missing. The most important thing is to make sure:

1. `tedge init` and services ensure their root directories exist and are owned by `tedge`
2. nothing tries to `chown` anything above the root directories
3. that we don't `chown` anything if we're running as a user that doesn't have `CAP_CHOWN` (or even if we try, that errors resulting from failed `chown`s are ignored) 

Then as long as all the directories are derived from the root directories and keep track of paths so they're accurately reported on errors, users should be able to fix any permissions issues.

crates/common/tedge_utils/src/file.rs
didier-wenzek: Yeah. This is simply `fs::create_dir_all` with improved error messages.

crates/common/tedge_utils/src/file.rs
didier-wenzek: Sure, I will have to remove or fix the failing test `ensure_does_not_create_directories_above_the_root`,
but this is not a breaking change. With that fix, the API has only more liberal pre-requisites.

Having two methods to create or not missing directories above the root, would only make things more confusing and all the invocations will finally end up to be written as `ensure_above_root().ensure()`, just to be sure.

Does this change defeat the recent refactoring introducing the concept of tedge root directory? I don't think so. On the contrary, the fix is simple thanks to that refactoring. We have now a clear understanding of what directories and files are owned by `tedge` and those that only have to be accessible.

It's true that my first reaction has been against that idea to create missing directories above tedge root. Such a strict policy would avoid mistakes (as creating `va/log/tedge` after a typo). However, if the `tedge init` command has then to be wrapped in a script creating any missing directories beforehand, it's better to make things simple, even if not ideal, with a `tedge init` command that creates the missing directories.

didier-wenzek: Test updated: https://github.com/thin-edge/thin-edge.io/pull/4196/commits/3d56d310d01a51f91ee21ef13a45bf368316a4d8

rina23q: > On the contrary, the fix is simple thanks to that refactoring. We have now a clear understanding of what directories and files are owned by tedge and those that only have to be accessible.

That’s true and it relieves me.

What I want to highlight here is, this change affects not only `tedge init`, but also all running services. If directories above a managed root directory are missing for some reason while services like tedge-agent are running, the process will create the missing directories.

jarhodes314: Yeah, it was somewhat of a deliberate decision in the refactoring _not_ to create these directories, but that was mostly for the purposes of just deciding that. We were doing these things somewhat arbitrarily and that felt like a bad solution (and ultimately led to code that at least looked like it was handling permissions for directories that aren't  owned by tedge).

My understanding of the situation is that we were implicitly relying on the directories we access existing by including `tedge init` in service definitions, and if we just want the default behaviour to be "create any directory we depend on", putting this in the application code is a very sensible idea.

My concerns around doing things like this generally stems from if the application is likely to misbehave if we temporarily lose access to a directory. For example, I've had issues with jellyfin deleting all of my metadata after a failed mount causing it to see an empty directory, but I don't think thin-edge tends to have such issues -- if we temporarily screw something up with the filesystem, tedge generally won't do anything that can't be resolved by reinstating the correct files and restarting the relevant services.

jarhodes314 (commented):
I think this is a good solution to the problem. It's a bit annoying that we can't really write unit tests for the permissions issues (without e.g. mocking the filesystem), but this is not really a problem we ought to solve here.

Issue #3786, opened 2025-09-18
Rotate operation logs under /var/log/tedge/agent


reubenmiller:
**Is your feature request related to a problem? Please describe.**

Currently the `/var/log/tedge/agent/` directory contains both the workflow logs and the custom operation handler logs, however these directories are not cleaned up over time. Whilst the directory only contains logs from operations (and not explicitly time-dependent), the logs should still be cleaned up over time to prevent unnecessary disk usage.

**Describe the solution you'd like**

When a workflow or custom operation handler is triggered, then t
* Only keep the last N number of logs matching the given pattern (based on time, where the oldest files are deleted)
* Allow user to configure how many logs to retain in the tedge.toml (controlled by `tedge config set`)
* Optional: Allow the log pattern to be configured to only delete logs files matching a specific pattern

Things to be aware of:
* do the cleanup before the workflow log is created for the current workflow so that problems with time jumps don't result in the current workflow log being deleted (e.g. if the device time jumps back due to the time not being in-sync)

**Describe alternatives you've considered**

A logrotate rule could be defined however it has the following limitations:
* Adds a dependency to logrotate (on some smaller embedded devices, this might not be available)
* The location of the `/var/log/tedge/agent` folder is not fully part of the public interface (e.g. only the `/var/log/tedge`, `logs.path,` directory is defined, so the user needs to know that the logs are stored under a sub folder, `agent`).

**Additional context**

d-timotijevic:
> No immediate plans considering that a from job can be done to delete older files.
> 
> Though this would also be a good first issue if you would like to contribute.

That's understandable. We'll add a custom workaround for now.
I'll see if I can free up some time to contribute here, though things are very hectic in my day job at the moment. If I do find the time, I will request you to assign it to me.

reubenmiller:
The current functionality works well, though I would propose a slight improvement.

The workflow log rotate should run when the tedge-agent starts up (in additional to what it currently does). Currently the workflow log cleanup only occurs when receiving a given log workflow command, so this means that if the device already has a large amount of workflow logs, then it won't get rotated until the user sends a new operation, so the space saving won't be immediate when upgrading the thin-edge.io version.

reubenmiller:
We'll skip the rotation of logs on startup for now as most of the time a software_list is sent when the mapper starts so it isn't worth the effort to trigger this on startup as well.

PR #2895, opened 2024-05-22
fix: Include software plugin output in workflow log

albinsuresh:
## Proposed changes

- [x] Include software plugin logs in their respective workflow logs
- [x] Update the log location of `software-management` log type in `tedge-log-plugin.toml` to point to the workflow logs
- [x] ~Handle this log type path change during upgrade~: Won't do

## Types of changes

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Improvement (general improvements like code refactoring that doesn't explicitly fix a bug or add any new functionality)
- [ ] Documentation Update (if none of the other choices apply)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/2835

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTING.md) doc
- [x] I have signed the [CLA](https://github.com/thin-edge/thin-edge.io/blob/main/CONTRIBUTOR-LICENSE-AGREEMENT.md) (in all commits with git commit -s)
- [x] I ran `cargo fmt` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [x] I used `cargo clippy` as mentioned in [CODING_GUIDELINES](https://github.com/thin-edge/thin-edge.io/blob/main/CODING_GUIDELINES.md)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)

PR #4205, opened 2026-06-04
feat: rotate operation logs

didier-wenzek:
## Proposed changes

- [x] Keep only N log files per operations under /var/log/tedge/agent/  (based on time, where the oldest files are deleted)
- [x] Add `tedge config` options to configure number of log files to keep per operation

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/3786

## Checklist


crates/core/plugin_sm/src/operation_logs.rs
```diff
@@ -27,17 +27,12 @@ pub struct OperationLogs {
 }
 
 pub enum LogKind {
-    SoftwareUpdate,
-    SoftwareList,
```
didier-wenzek, line 31: The file rotation logic once implemented for software operations has been removed (by accident?) in that commit https://github.com/thin-edge/thin-edge.io/pull/2895/changes#r3356159116

The code can be re-used for all operations but having specific log kinds, for the software update and list operations, is no more useful as the agent handles arbitrary operations.

crates/extensions/c8y_mapper_ext/src/converter.rs
```diff
@@ -221,7 +221,7 @@ impl CumulocityConverter {
 
         let log_dir = config
             .logs_path
-            .dir(TEDGE_AGENT_LOG_DIR)
+            .dir(TEDGE_AGENT_LOG_DIR) // FIXME-DIDIER Why is the agent logs dir created by the C8Y mapper?
```
didier-wenzek: @albinsuresh Do you know why the mapper is logging operations in the agent log directory ?

reubenmiller:
@didier-wenzek Can we make sure we sort the files based on creation time and not last modified time (if not already done so)?

The reason is that when doing firmware upgrades, generally the device's clock will be reset and it could potentially be set back to 2011, so we want to ensure that the these files aren't lost when a rollback occurs and the "seemingly old" log file is rotated out. Below is a rough description of the scenario:

1. Start firmware update on Partition A
2. Switch active partition to B to boot, then reboot device
3. OS starts on Partition B but the clock hasn't synced yet due to some network issue, this causes a rollback to occur
4. OS Starts on Partition A (due to the rollback), and at the startup the, other operations list "software list" will be triggered which would potentially cause a log rotation

crates/extensions/c8y_mapper_ext/src/converter.rs
reubenmiller: I believe this is for when the tedge-mapper has a custom operation handler which doesn't use a workflow, so the tedge-mapper process is the executing and logging the script/binary defined in the operation definition.

For example, https://thin-edge.github.io/thin-edge.io/operate/c8y/smartrest-templates/#step-3-creating-the-operation-handler

crates/core/tedge_api/src/workflow/log/log_dir.rs
```diff
@@ -0,0 +1,287 @@
...
+        assert!(swlist_log_1.exists());
+        assert!(swlist_log_2.exists());
+
+        Ok(())
+    }
+
+    async fn create_file(dir: &Path, name: &str) -> PathBuf {
+        let file_path = dir.join(name);
+        let _log_file = File::create(file_path.clone()).expect("fail to create a test file");
+        tokio::time::sleep(Duration::from_millis(1000)).await;
```
didier-wenzek: This is really painful, making the tests terribly slow.

Since the file creation timestamps are only accurate to the second, one has to wait before creating another file (during testing).

=> A suggestion from @reubenmiller is to suffix the log filenames with a monotonic sequence number.
A bit more involved but easier to test.

didier-wenzek:
> @didier-wenzek Can we make sure we sort the files based on creation time and not last modified time (if not already done so)?

Implemented using the creation time as suggested: https://github.com/thin-edge/thin-edge.io/pull/4205/changes/46af513f8439719da1c13826818574a46bd2c573

crates/core/tedge_api/src/workflow/log/log_dir.rs
didier-wenzek: I was wrong. File creation timestamps are accurate below the second and a few milliseconds between file creations is enough for the test. https://github.com/thin-edge/thin-edge.io/pull/4205/changes/448fabdcbbe7493aee4cc878d50760a83ae420f0

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -197,6 +197,63 @@ Placeholder workflow created for ill-defined operations
...
+    ${max_per_command}    Execute Command    tedge config get logs.max_per_operation
+    Should Contain    ${max_per_command}    *:5
+    Execute Operation Workflow    lite_software_update    n001
+    Execute Operation Workflow    lite_software_update    n002
+    Execute Operation Workflow    lite_software_update    n003
+    Execute Operation Workflow    lite_software_update    n004
+    Execute Operation Workflow    lite_software_update    n005
+    Execute Operation Workflow    lite_software_update    n006
+    ${log_files}    Execute Command    ls /var/log/tedge/agent/*-lite_software_update-*
+    Should Not Contain    ${log_files}    workflow-lite_software_update-n001.log
```
didier-wenzek, line 210: This test is failing because this file is actually not removed,
the root cause being that there is no creation date attached to the file.

A modification date is attached, though. => The fix is to use the modification date when no creation date is provided by the file system. Done: https://github.com/thin-edge/thin-edge.io/pull/4205/changes/a1a2d3842981807e93d9abe0221c4d5a185d7ec0

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -1289,6 +1290,12 @@ define_tedge_config! {
         /// The directory used to store logs
         #[tedge_config(example = "/var/log/tedge", default(from_str = "/var/log/tedge"))]
         path: AbsolutePath,
+
+        /// The maximum count of log files per operation
+        #[tedge_config(default(from_str = "*:5, software_list:1"))]
+        #[tedge_config(example = "*:5, software_list:1")]
+        #[tedge_config(note = "Expect a list of <operation-name>:<max-count> pairs, using *:<max-count> for a default maximum")]
+        max_per_operation: CountPerGroup,
```
jarhodes314, line 1298: This is an elegant way to incorporate this into `tedge config`, particularly with the `add/remove` support :)

docs/src/operate/troubleshooting/log-files.md
```diff
@@ -64,15 +64,34 @@ journalctl -u tedge-mapper-collectd
...
-This section describes how to access the software management component logs
+### Operations logs {#software-management}
 
-#### Software update operation log {#software-update}
-For every new software operation (list/update), a new log file will be created at `/var/log/tedge/agent`.
-For each `plugin command` like prepare, update-list (install, remove), finalize, and list,
-the log file captures `exit status, stdout, and stderr` messages.
+The agent creates a log file per operation execution, logging all the steps, sub-commands output and errors.
+This is done for all operations: software, config and firmware updates but also software or config list requests
+as well as any [user-defined operation defined using a workflow](../../../references/agent/operation-workflow).
```
Bravo555, line 71: note: When I tried to open this link directly in VSCode it failed because of wrong path, and indeed the path here has 3x `../` because to get from this file to references we should only go up twice (once to move up from `/operate/troubleshooting` to `/operate` and then again to go to `/`), but I was surprised this wouldn't work in the `tedge-docs` website, and 3x `../` was actually required!

It's not ideal because it makes it harder to view the docs in any other markdown-capable text editor because the links are slightly wrong there. That's probably because the URL for this page is `http://localhost:3000/thin-edge.io/next/operate/troubleshooting/log-files/` so first `../` actually goes to `troubleshooting/` dir and the fix would be to either not append final `/` or rewrite the link when generating HTML to fixup the path.

docs/src/operate/troubleshooting/log-files.md
```diff
@@ -64,15 +64,34 @@ journalctl -u tedge-mapper-collectd
...
-#### tedge-agent logs {#tedge-agent-logs}
+These log files are stored in `/var/log/tedge/agent`.
+
+- This directory can be configured with `tedge config set logs.path <tedge-log-directory>`,
+- the `agent` sub-directory being created by the agent or `tedge init`.
+
+Each log file is named after the [command identifier as published over MQTT](../../../references/mqtt-api/#command-examples).
+
+- A software update command published on `te/<device-id>/cmd/<command-name>/<cmd_id>`
+- is monitored by the log file `workflow-<command-name>-<cmd_id>.md`
```
Bravo555: nit: a bit inconsistent word separation

```suggestion
- A software update command published on `te/<device-id>/cmd/<command-name>/<cmd-id>`
- is monitored by the log file `workflow-<command-name>-<cmd-id>.md`
```

crates/core/tedge_api/src/workflow/log/log_dir.rs
```diff
@@ -0,0 +1,293 @@
...
+
+        let mut command_log = CommandLog::from_log_path(file_path, operation_name, cmd_id);
+        self.ensure_file_exists(&mut command_log).await;
+        Ok(command_log)
+    }
+
+    pub async fn remove_all_outdated_logs(&self) {
+        for (operation, keep) in self.keep_per_operation.iter() {
+            if let Err(err) = self.remove_outdated_logs(operation, *keep).await {
+                log::warn!("Fail to remove out-dated log files: {}", err);
```
Bravo555, line 87: thought: wanted to say to prefer `tracing` over `log`, but it appears `tedge_api` still uses `log` and not `tracing`, so it's ok, better to not change it in this PR

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
```diff
@@ -197,6 +197,63 @@ Placeholder workflow created for ill-defined operations
     ${workflow_log}    Execute Command    cat /var/log/tedge/agent/workflow-issue-3079-test-1.log
     Should Contain    ${workflow_log}    item=TOML parse error
 
+The default is to keep a maximum of 5 log files per operation
+    ${max_per_command}    Execute Command    tedge config get logs.max_per_operation
```
reubenmiller, line 201: @didier-wenzek This question is more out of my curiosity is regarding the visualization of the values shown by the `tedge config get` command, and how the values are rendered on the console.

For example, when getting the `logs.max_per_operation` values, then the rendered array is not showing double quotes around each item in the array.

```sh
$ tedge config get logs.max_per_operation
logs.max_per_operation=[*:5, software_list:1]
```

But with other array config items, it wraps the values with quotes.

```sh
$  tedge config get c8y.smartrest.templates
["bar", "foo"]
```

Is this expected?

docs/src/operate/troubleshooting/log-files.md
```diff
@@ -64,15 +64,34 @@ journalctl -u tedge-mapper-collectd
...
+- stored in the log directory of the agent running on the device `<device-id>`.
+
+Over time, these log files are purged by the agent.
+- The default agent behavior is to keep a maximum of 5 log files per operation type.
+- This default can be configured using `tedge config add logs.max_per_operation "*:4"` where `*` stands for any operation,
+  and the number is the new default (`4` in that case).
```
reubenmiller, line 87: If someone adds a new entry without the `:N` format, e.g.

```sh
$ tedge config add logs.max_per_operation "software_list"
```

It defaults to `software_list:0`, however it behaves more like `:1`.

```
# tedge config get logs.max_per_operation
[*:5, software_list:0]
```

So the question is, if the user does not provide a `:N` value, then should we default to `:1` instead of `:0`?

reubenmiller: (though this is a really really minor point)

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
didier-wenzek: This is what I intended. But, yes I introduced some inconsistency.

On purpose, I introduced some differences between input and output of `logs.max_per_operation`.

- `get` always displays the list inside brackets `[ ... ]`, while there are not required by `set`, `add` and `remove`.
  - The rational is to ease addition: `tedge config add logs.max_per_operation software_update:3`
- A count is not mandatory, and is interpreted as `0` then.
  - The idea is to ease removal:  `tedge config remove logs.max_per_operation software_update`

docs/src/operate/troubleshooting/log-files.md
didier-wenzek: Yeah, `0` is interpreted as `1`, because it makes no sense to keep 0 log files for a command.

> should we default to :1 instead of :0?

This is a good suggestion.

docs/src/operate/troubleshooting/log-files.md
```diff
@@ -64,15 +64,34 @@ journalctl -u tedge-mapper-collectd
...
+- A max log file count can be assigned specifically to an operation with
+ `tedge config add logs.max_per_operation "<operation-name>:<max-count>"`,
+  for example `software_list:1`.
+- The default can be reset for an operation with `tedge config remove logs.max_per_operation "<operation-name>"`
+- Note that for these configuration updates to be effective, the agent must be restarted.
```
Bravo555, line 95: todo: I would also add here how to completely disable the purging.

A user determined to disable the feature completely, e.g. because they're testing something and want to see how many logfiles are created, so they don't want any to be removed, might be confused about how to do so.
There are some traps:
- `tedge config unset` resets to the default, which is enabled
- one could set something like `*:9999` - not particularly elegant and most importantly won't work if there's some more specific operation which has a smaller limit
- you could do `tedge config remove '*'`, but then still some more specific limits may affect removal.

AFAIK, with the current implementation, the recommended way would be to run `tedge config set logs.max_per_operation []`, but then `tedge config get logs.max_per_operation` returns `[:0]`, which could confuse the user that a limit has been set to `0`, i.e. no messages are kept, and also it doesn't work, see comment in `log_dir.rs`.

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
reubenmiller: Ok, I'm fine with the current behaviour. Just adding quotes around it to look visually consistent isn't really worth it.

It was also a very nice touch with allowing users to remove an entry via `tedge config remove logs.max_per_operation software_update`.

crates/core/tedge_api/src/workflow/log/log_dir.rs
```diff
@@ -0,0 +1,293 @@
...
+        let path = Utf8Path::from_path(path).unwrap();
+        TedgePaths::from_root_with_defaults(path, "", "").root_dir()
+    }
+
+    #[tokio::test]
+    async fn on_start_keeps_only_the_latest_logs() -> Result<(), anyhow::Error> {
+        // Create a log dir with a bunch of fake log files
+        let log_dir = TempDir::new()?;
+
+        let swlist_log_1 = create_file(log_dir.path(), "software-list-1996-02-22T16:39:57z").await;
```
jarhodes314: I find the use of a timestamp in the filename a little bit confusing in this context. This is because we do date-based log rotation, but it's the creation time that matters, not any date in the filename. Also, I believe the actual filenames from c8y operations use operation ids in the name, not a timestamp.

crates/core/tedge_api/src/workflow/log/log_dir.rs
```diff
@@ -0,0 +1,293 @@
...
+    pub log_dir: ManagedDir,
+    keep_default: u32,
+    keep_per_operation: HashMap<String, u32>,
+}
+
+impl OperationLogs {
+    pub fn new(log_dir: ManagedDir) -> OperationLogs {
+        OperationLogs {
+            log_dir,
+            keep_default: 5,
```
Bravo555: issue: When a user runs `tedge config set logs.max_per_operation []`, believing it should disable the removal feature, we're still removing files, surprising the user!

It's additionally confusing because then `tedge config get logs.max_per_operation` returns `[:0]`, so the user may think that they wanted to remove the limit, but `tedge config` tells them the new limit is `0`, while the real limit is still `5`!

Because in `tedge_config/src/lib.rs` we're only calling `keep_default` if `*` is in `max_per_operation`, if `*` is removed we're using this default here, which the user has no idea why it's `5`, because AFAIK we're not logging it anywhere here.

crates/core/tedge_api/src/workflow/log/log_dir.rs
didier-wenzek: Agree. I will fix the names. My point was to minimize changes on existing tests during the refactoring, but it's no more just a refactoring.

Bravo555 (dismissed):
Trying to disable this feature appears to have some traps, so I think this warrants requesting the change here.

When a user runs `tedge config set logs.max_per_operation []`, believing it should disable the removal feature, we're still removing files, surprising the user!

It's additionally confusing because then `tedge config get logs.max_per_operation` returns `[:0]`, so the user may think that they wanted to remove the limit, but tedge config tells them the new limit is `0`, while the real limit is still `5`!

To achieve a design where these issues are prevented, I think it should satisfy some additional properties (blocking issues):

- no any intermediate defaults between components that are not visible to the user, any value for the limit must come only from the config, and nowhere else, so there's no situation where we limit to a different number than what `tedge config` says. The only way to create `OperationLogs` should be to explicitly provide all the values
- currently in `OperationsLogs`, for the feature to be effectively disabled, one'd have to set `keep_default` to `u32::MAX` and `keep_per_operation` to `HashMap::new()`. In `CountPerGroup`, that would be achieved by setting it to `[*:u32::MAX]`, but slightly different and less annoying `[]`, that some users might try, sets `keep_default` to 5. Some meaning is lost in translation between config layer and `OperationLogs` layer. Ideally layers should have one canonical `disabled` state and would be somewhat similar by structure.

Additionally (non-blocking):
- In `OperationLogs`, maybe instead of embedding it directly, we could have very clear `enum LogRetentionPolicy {Disabled, Enabled {keep_default: u32, keep_per_operation: HashMap<String, u32>}}`, which if `LogRetentionPolicy::Disabled`, no removal-related code would ever run. This conflicts somewhat with the idea to [eliminate special cases by changing them into regular cases](https://felipec.github.io/good-taste/parts/1.html), so I'll let you judge whether to make this change
- Similarly in tedge config, the _disabled_ state should be clear. Current `[]` for `max_per_operation` could work, but in tedge config we have many other simple boolean `enable` flags for other features, so maybe we could also provide one here for symmetry, just so we don't start adding any features where each is disabled in a completely different manner

tests/RobotFramework/tests/tedge_agent/workflows/custom_operation.robot
didier-wenzek: However, there is something weird in `tedge.toml`: The whole list is stored as a string, not as a list of strings.

The mistake being to deserialize with `#[serde(try_from = "String")]` instead of `#[serde(try_from = "Vec<String>")]`.

=> fixed: https://github.com/thin-edge/thin-edge.io/pull/4205/changes/fef480aa6a52f735fd72dadb6720cf2a1b7597f1

docs/src/operate/troubleshooting/log-files.md
didier-wenzek: > todo: I would also add here how to completely disable the purging.

Intentionally the feature cannot be disabled. That said, I agree that the current behavior of `tedge config set logs.max_per_operation []` is confusing.

docs/src/operate/troubleshooting/log-files.md
Bravo555: > If someone adds a new entry without the :N format, e.g.

I'd say this is a result of trying to map the data which is inherently a map of strings to numbers, from a list, which necessitates additional parsing and related error handling and choosing of default values.

I think it would be nicer to read `per_operation` directly as a map instead because it better fits to what the data actually is and we reduce some special domain-errors into more general config schema errors.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -1289,6 +1290,12 @@ define_tedge_config! {
         /// The directory used to store logs
         #[tedge_config(example = "/var/log/tedge", default(from_str = "/var/log/tedge"))]
         path: AbsolutePath,
+
+        /// The maximum count of log files per operation
+        #[tedge_config(default(from_str = "*:5, software_list:1"))]
```
Bravo555, line 1295: thought: Since I didn't see any discussions about this, because I assume there are no immediately apparent reasons for changing it, I was wondering if the fallback element should be in the map or as a separate option.

In the `max_per_operation` map, `*` is the fallback element - it applies if the map doesn't have entry that matches the operation type we're currently processing. I think it would make sense to allow having it be optional if the user wants to enable rotation for some operation types, but keep it disabled for all others. This maps neatly to map entries not having a fallback, which simply means don't consider for removal (i.e. keep all logs).

And all of the above holds regardless of if the fallback is part of, or outside of, the map. I'd say with it inside _maybe_ it's a bit more annoying to clear all the per-type limits without affecting the main limit, (one'd have to either save it, clear list and restore it, or remove all types separately), but other than that I can't think of any clear reasons for changing it.

docs/src/operate/troubleshooting/log-files.md
Bravo555: > Intentionally the feature cannot be disabled

Ah, if that's intentional then indeed much of my feedback isn't really applicable. But why are we not allowing users to disable the feature?

didier-wenzek: Your comment is applicable, and made me changed my mind. I will properly handle the case where no value is provided for an operation means that all the log for that operation have to be kept.

docs/src/operate/troubleshooting/log-files.md
didier-wenzek: Without the 3x `../` the links are broken

crates/core/tedge_api/src/workflow/log/log_dir.rs
didier-wenzek: Fixed: https://github.com/thin-edge/thin-edge.io/pull/4205/changes/102475fa00330519e331ba211be06917b7d19d91

docs/src/operate/troubleshooting/log-files.md
didier-wenzek: "1" is now used as the default.

This has been done along a complete refactoring of the serialization/deserialization of the `logs.max_per_operation`: https://github.com/thin-edge/thin-edge.io/pull/4205/changes/fef480aa6a52f735fd72dadb6720cf2a1b7597f1

docs/src/operate/troubleshooting/log-files.md
didier-wenzek: fixed

docs/src/operate/troubleshooting/log-files.md
didier-wenzek: > todo: I would also add here how to completely disable the purging.

Added: https://github.com/thin-edge/thin-edge.io/pull/4205/changes/a425ee5d0dbdb654428fe38b441a97eb7a6940eb

crates/core/tedge_api/src/workflow/log/log_dir.rs
didier-wenzek: Fixed, by https://github.com/thin-edge/thin-edge.io/pull/4205/changes/fef480aa6a52f735fd72dadb6720cf2a1b7597f1 and https://github.com/thin-edge/thin-edge.io/pull/4205/changes/a425ee5d0dbdb654428fe38b441a97eb7a6940eb

```
$ tedge config set logs.max_per_operation "[]"
$ tedge config get logs.max_per_operation     
[]
$tedge config set logs.max_per_operation ""  
$ tedge config get logs.max_per_operation   
[]
$ tedge config unset logs.max_per_operation
$ tedge config get logs.max_per_operation  
[*:5, software_list:1]
$ tedge config remove logs.max_per_operation "*"
$ tedge config get logs.max_per_operation       
[software_list:1]
```

reubenmiller (approved):
Approved. Very nice addition with a good level of control

PR #4207, opened 2026-06-05
feat: add single-process supervisor run mode

jarhodes314:
## Proposed changes

Add a run mode that supervises the agent and mappers within a single process, with a view to ultimately removing the need for an external init system to start and keep the components running, making `tedge-standalone` and containerised deployments simpler and reducing resource consumption.

Currently this runs `tedge-agent` and a single mapper. The next step would be to support multiple mappers and other tedge services, and allow the user to configure which services run.

To try this out, run `tedge run all <cloud>` where `<cloud>` is the mapper you wish to run, e.g. `tedge run all c8y`. Once multiple mappers are supported, this cloud argument will go away in favour of configuration for which "services" tedge run all should execute.

The supervisor isolates component failures with automatic restart and bounded backoff, owns process signal handling for graceful shutdown (SIGTERM/SIGINT/SIGQUIT) and mapper-only restart (SIGUSR1), and starts the agent before mappers without imposing a readiness dependency. Each component keeps its own MQTT connection and session so restarts preserve broker-side offline message queueing, and logs are attributed per component. Standalone invocation of each component is unaffected.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/common/flockfile/src/unix.rs
+++ b/crates/common/flockfile/src/unix.rs
@@ -12,7 +12,6 @@ use std::path::Path;
 use std::path::PathBuf;
 use tracing::debug;
 use tracing::error;
-use tracing::info;
 use tracing::warn;
 
 const LOCK_CHILD_DIRECTORY: &str = "lock/";
@@ -148,8 +147,11 @@ pub fn check_another_instance_is_not_running(
             Err(err)
         }
         Err(_) => {
-            info!(
-                "The locking mechanism is not active, as {:?} is not a directory with write access.",
+            warn!(
+                "Single-instance locking is NOT active for {app_name}: {:?} is not a writable \
+                 directory. Another instance running concurrently will not be detected. Ensure \
+                 the run directory exists and is writable by the tedge user, or disable locking \
+                 explicitly with run.lock_files=false.",
                 lock_path.parent().unwrap() // This is true as run_dir and lock directories are always provided
             );
             Ok(None)
--- a/crates/core/tedge/src/cli/mod.rs
+++ b/crates/core/tedge/src/cli/mod.rs
@@ -65,6 +65,10 @@ pub struct TEdgeCli {
 
 #[derive(clap::Parser, Debug)]
 pub enum Component {
+    /// Run the agent and a mapper together under a single-process supervisor
+    #[clap(hide = true)]
+    All(crate::supervisor::RunAllOpt),
+
     C8yFirmwarePlugin(FirmwarePluginOpt),
 
     C8yRemoteAccessPlugin(C8yRemoteAccessPluginOpt),
--- a/crates/core/tedge/src/supervisor.rs
+++ b/crates/core/tedge/src/supervisor.rs
@@ -738,6 +738,63 @@ mod tests {
             .unwrap();
     }
 
+    #[tokio::test(start_paused = true)]
+    async fn rapid_mapper_restart_requests_do_not_overlap_live_actors() {
+        let mapper_builds = Arc::new(AtomicUsize::new(0));
+        let live_actors = Arc::new(AtomicUsize::new(0));
+        let max_live_actors = Arc::new(AtomicUsize::new(0));
+        let policy = test_policy(1000);
+
+        let units = vec![make_unit(
+            "mapper",
+            UnitKind::Mapper,
+            leaky_shutdown_factory(
+                mapper_builds.clone(),
+                live_actors.clone(),
+                max_live_actors.clone(),
+            ),
+            policy,
+        )];
+        let mut supervisor = Supervisor::new(units);
+        supervisor.drain_timeout = Duration::from_millis(100);
+        let commands = supervisor.commands();
+        let handle = tokio::spawn(supervisor.run_loop());
+
+        wait_until(
+            || mapper_builds.load(Ordering::SeqCst) == 1,
+            "the mapper to start",
+        )
+        .await;
+
+        commands.send(Command::RestartMappers).await.unwrap();
+        commands.send(Command::RestartMappers).await.unwrap();
+
+        tokio::task::yield_now().await;
+        tokio::task::yield_now().await;
+        tokio::time::advance(Duration::from_secs(61)).await;
+        tokio::task::yield_now().await;
+        tokio::task::yield_now().await;
+        wait_until(
+            || mapper_builds.load(Ordering::SeqCst) >= 2,
+            "the mapper to be restarted",
+        )
+        .await;
+
+        assert_eq!(
+            max_live_actors.load(Ordering::SeqCst),
+            1,
+            "a mapper restart must not start a replacement while the previous actor is still alive"
+        );
+
+        commands.send(Command::ShutdownAll).await.unwrap();
+        commands.send(Command::ShutdownAll).await.unwrap();
+        timeout(Duration::from_secs(5), handle)
+            .await
+            .expect("supervisor should exit after shutdown")
+            .unwrap()
+            .unwrap();
+    }
+
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn shutdown_drains_all_units_and_exits() {
         let agent_builds = Arc::new(AtomicUsize::new(0));
@@ -1053,6 +1110,32 @@ mod tests {
         })
     }
 
+    /// A factory whose actor stays alive after shutdown is requested. This models an
+    /// actor that has not actually released its external resource (e.g. an MQTT
+    /// client id) by the time the runtime gives up waiting for cleanup.
+    fn leaky_shutdown_factory(
+        builds: Arc<AtomicUsize>,
+        live_actors: Arc<AtomicUsize>,
+        max_live_actors: Arc<AtomicUsize>,
+    ) -> RuntimeFactory {
+        Box::new(move || {
+            builds.fetch_add(1, Ordering::SeqCst);
+            let live_actors = live_actors.clone();
+            let max_live_actors = max_live_actors.clone();
+            async move {
+                let mut runtime = Runtime::new();
+                runtime
+                    .spawn(LeakyShutdownActorBuilder {
+                        live_actors,
+                        max_live_actors,
+                    })
+                    .await?;
+                Ok(runtime)
+            }
+            .boxed()
+        })
+    }
+
     /// Polls `cond` until it holds, failing the test if it does not within 5s.
     async fn wait_until(mut cond: impl FnMut() -> bool, what: &str) {
         let deadline = tokio::time::Instant::now() + Duration::from_secs(5);
@@ -1130,6 +1213,61 @@ mod tests {
     }
 
     impl RuntimeRequestSink for StubbornActorBuilder {
+        fn get_signal_sender(&self) -> DynSender<RuntimeRequest> {
+            NullSender.into()
+        }
+    }
+
+    struct LeakyShutdownActor {
+        live_actors: Arc<AtomicUsize>,
+        max_live_actors: Arc<AtomicUsize>,
+    }
+
+    #[async_trait::async_trait]
+    impl Actor for LeakyShutdownActor {
+        fn name(&self) -> &str {
+            "leaky-shutdown"
+        }
+
+        async fn run(self) -> Result<(), RuntimeError> {
+            let live = self.live_actors.fetch_add(1, Ordering::SeqCst) + 1;
+            self.max_live_actors.fetch_max(live, Ordering::SeqCst);
+            let _guard = LiveActorGuard(self.live_actors.clone());
+            std::future::pending::<()>().await;
+            #[allow(unreachable_code)]
+            Ok(())
+        }
+    }
+
+    struct LiveActorGuard(Arc<AtomicUsize>);
+
+    impl Drop for LiveActorGuard {
+        fn drop(&mut self) {
+            self.0.fetch_sub(1, Ordering::SeqCst);
+        }
+    }
+
+    struct LeakyShutdownActorBuilder {
+        live_actors: Arc<AtomicUsize>,
+        max_live_actors: Arc<AtomicUsize>,
+    }
+
+    impl Builder<LeakyShutdownActor> for LeakyShutdownActorBuilder {
+        type Error = std::convert::Infallible;
+
+        fn try_build(self) -> Result<LeakyShutdownActor, Self::Error> {
+            Ok(self.build())
+        }
+
+        fn build(self) -> LeakyShutdownActor {
+            LeakyShutdownActor {
+                live_actors: self.live_actors,
+                max_live_actors: self.max_live_actors,
+            }
+        }
+    }
+
+    impl RuntimeRequestSink for LeakyShutdownActorBuilder {
         fn get_signal_sender(&self) -> DynSender<RuntimeRequest> {
             NullSender.into()
         }
--- a/crates/core/tedge/src/supervisor.rs
+++ b/crates/core/tedge/src/supervisor.rs
@@ -1,0 +1,1137 @@
+//! Single-process supervisor reached through `tedge run all`.
+//!
+//! Hosts the core thin-edge components (the agent and a mapper) inside one process
+//! so no external init system is required. Each component is a *unit*: a rebuildable
+//! factory plus a restart policy. The supervisor owns process-wide signal handling,
+//! isolates a crashing unit from its co-hosted peers (rebuilding it under a bounded
+//! backoff), and drains everything cleanly on termination.
+//!
+//! Each component's `build()` factory is also used by the standalone per-component
+//! path (`tedge-agent`, `tedge-mapper c8y`, …); only the way the resulting runtime
+//! is driven differs.
+
+// Crash isolation depends on tokio catching a panicking actor task per-task, which
+// only happens with the unwinding panic runtime. Refuse to build the supervisor if
+// the release profile is ever switched to `panic = "abort"`, which would let one
+// component's panic abort every co-hosted component.
+#[cfg(panic = "abort")]
+compile_error!(
+    "the single-process supervisor (`tedge run all`) requires `panic = \"unwind\"` for \
+     per-component crash isolation; `panic = \"abort\"` would let one component's panic \
+     abort the whole process"
+);
+
+use anyhow::Context;
+use futures::future::BoxFuture;
+use futures::stream::FuturesUnordered;
+use futures::FutureExt;
+use futures::StreamExt;
+use std::collections::VecDeque;
+use std::time::Duration;
+use std::time::Instant;
+use tedge_actors::Runtime;
+use tedge_actors::RuntimeHandle;
+use tedge_agent::AgentOpt;
+use tedge_config::cli::CommonArgs;
+use tedge_config::log_init_for_services;
+use tedge_config::TEdgeConfig;
+use tedge_mapper::MapperName;
+use tokio::signal::unix;
+use tokio::sync::mpsc;
+use tracing::error;
+use tracing::info;
+use tracing::warn;
+use tracing::Instrument;
+
+/// `tedge run all` — run the agent and (optionally) a mapper under one supervisor.
+#[derive(Debug, clap::Parser)]
+pub struct RunAllOpt {
+    /// The mapper to run alongside the agent (e.g. `c8y`, `aws`, `az`).
+    #[clap(subcommand)]
+    pub mapper: Option<MapperName>,
+
+    #[command(flatten)]
+    pub common: CommonArgs,
+}
+
+/// Kind of a supervised unit. Drives start ordering (agent before mappers) and
+/// which signals target it (SIGUSR1 restarts only mappers).
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum UnitKind {
+    Agent,
+    Mapper,
+}
+
+#[derive(Clone, Copy, Debug)]
+struct RestartPolicy {
+    initial_backoff: Duration,
+    max_backoff: Duration,
+    max_restarts: usize,
+    window: Duration,
+}
+
+impl Default for RestartPolicy {
+    fn default() -> Self {
+        Self {
+            initial_backoff: Duration::from_secs(1),
+            max_backoff: Duration::from_secs(60),
+            max_restarts: 5,
+            window: Duration::from_secs(60),
+        }
+    }
+}
+
+/// Produces a fresh component runtime on each call (the rebuildable factory).
+type RuntimeFactory = Box<dyn Fn() -> BoxFuture<'static, anyhow::Result<Runtime>> + Send>;
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum UnitStatus {
+    /// A runtime is built and its completion task is being awaited.
+    Running,
+    /// Asked to stop so it can be rebuilt (on-demand restart): rebuild immediately
+    /// once it finishes draining.
+    Restarting,
+    /// Crashed and waiting for its backoff timer before being rebuilt.
+    BackingOff,
+    /// Exceeded the restart cap; left down without exiting the process.
+    GaveUp,
+    /// Finished and not to be restarted (clean exit, or supervisor shutdown).
+    Stopped,
+}
+
+struct Unit {
+    name: String,
+    kind: UnitKind,
+    factory: RuntimeFactory,
+    policy: RestartPolicy,
+    status: UnitStatus,
+    /// `Some` while a runtime is live — used to request a graceful drain.
+    handle: Option<RuntimeHandle>,
+    /// Timestamps of recent restarts, pruned to the policy window.
+    restarts: VecDeque<Instant>,
+    /// Single-instance lock, held for the unit's whole lifetime (across restarts).
+    _lock: Option<flockfile::Flockfile>,
+}
+
+/// Message sent back to the supervisor when a unit's runtime finishes.
+struct UnitFinished {
+    id: usize,
+    result: Result<(), tedge_actors::RuntimeError>,
+}
+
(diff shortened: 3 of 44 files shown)


openspec/specs/single-process-supervisor/spec.md
```diff
@@ -0,0 +1,114 @@
...
+
+The single-process run mode SHALL initialise one logging subscriber and attribute
+each log record to the component that produced it.
+
+#### Scenario: Logs identify their component
+
+- **WHEN** multiple components log to the shared process log stream
+- **THEN** each record identifies the originating component
+</content>
+</invoke>
```
Bravo555: question: what does this do?

crates/core/tedge/src/cli/mod.rs
```diff
@@ -65,6 +65,10 @@ pub struct TEdgeCli {
 
 #[derive(clap::Parser, Debug)]
 pub enum Component {
+    /// Run the agent and a mapper together under a single-process supervisor
+    #[clap(hide = true)]
```
Bravo555, line 69: question: why hide the option? Is this something that's not supposed to be used by the users yet?

crates/core/tedge/src/cli/mod.rs
```diff
@@ -65,6 +65,10 @@ pub struct TEdgeCli {
...
+    All(crate::supervisor::RunAllOpt),
```
Bravo555, line 70: issue: not sure if `all` is the best name, one'd assume all means all, and not only two components. But I'm not sure what to replace it with, `agent-and-mapper` doesn't have the same ring to it.

crates/core/tedge/src/supervisor.rs
```diff
@@ -0,0 +1,1121 @@
...
+// Crash isolation depends on tokio catching a panicking actor task per-task, which
+// only happens with the unwinding panic runtime. Refuse to build the supervisor if
+// the release profile is ever switched to `panic = "abort"`, which would let one
+// component's panic abort every co-hosted component.
+#[cfg(panic = "abort")]
+compile_error!(
+    "the single-process supervisor (`tedge run all`) requires `panic = \"unwind\"` for \
+     per-component crash isolation; `panic = \"abort\"` would let one component's panic \
+     abort the whole process"
+);
```
Bravo555, line 22: thought: nice safety check to make sure we fail to build rather than have a nasty surprise at runtime.

crates/core/tedge/src/supervisor.rs
```diff
@@ -0,0 +1,1121 @@
...
+    });
+
+    Ok(())
+}
+
+/// Entry point for `tedge run all`: assembles the units and runs the supervisor.
+pub async fn run(opt: RunAllOpt) -> anyhow::Result<()> {
+    // A single tracing subscriber for the whole process; each unit's logs are
+    // attributed to its component via the `component` span field.
+    log_init("tedge", &opt.common.log_args, &opt.common.config_dir)?;
```
Bravo555: todo: `log_init` can read from files the logging configuration for each component, currently that appears to be ignored.

Create `/etc/tedge/system.toml` with following contents:

```toml
[log]
tedge-agent = "trace"
```

Now `tedge run tedge-agent` should print debug and trace logs but `tedge run all` does not.

If this feature isn't meant to be used by users yet we could file it as a separate issue and do as a follow-up, otherwise it would be good to have it fixed.

crates/common/flockfile/src/unix.rs
```diff
@@ -148,8 +147,11 @@ pub fn check_another_instance_is_not_running(
             Err(err)
         }
         Err(_) => {
-            info!(
-                "The locking mechanism is not active, as {:?} is not a directory with write access.",
+            warn!(
+                "Single-instance locking is NOT active for {app_name}: {:?} is not a writable \
+                 directory. Another instance running concurrently will not be detected. Ensure \
+                 the run directory exists and is writable by the tedge user, or disable locking \
+                 explicitly with run.lock_files=false.",
```
Bravo555, line 154: todo: If I'm not wrong, this message will now be used for locks even if components are not launched as a single instance? If so, we'll have to change it a bit if we want to use a different message for single-instance locking.

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
```diff
@@ -0,0 +1,118 @@
...
+Documentation       Smoke tests for the single-process supervisor (`tedge run all`).
+...                 Runs the agent and the c8y mapper together inside one process and
+...                 checks both come up healthy and the device keeps talking to
+...                 Cumulocity, that the supervisor's locks make it mutually exclusive
+...                 with the standalone components, and that SIGUSR1 really restarts
+...                 the mapper (proven via a certificate rotation that only takes
+...                 effect on a restart).
+...
+...                 The suite is sequential: the supervisor started by the first test
+...                 is reused by the later ones.
```
Bravo555: suggestion: I'd prefer if the newly created test suites didn't rely on a sequential test execution order.

It makes it impossible to run only one test if it has dependencies on earlier tests.

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
```diff
@@ -0,0 +1,118 @@
...
+Suite Teardown      Custom Teardown
+
+Test Tags           theme:supervisor    theme:c8y
+
+
+*** Test Cases ***
+Run the agent and c8y mapper under a single supervisor
+    # Establish connectivity the normal way first, to prove a working baseline.
+    Execute Command    tedge connect c8y
+    Device Should Exist    ${DEVICE_SN}
```
Bravo555: todo: `Device Should Exist` is deprecated

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
```diff
@@ -0,0 +1,118 @@
...
+    # Baseline: the supervised bridge is connected with the current cert (C1).
+    Wait Until Keyword Succeeds
+    ...    60s    2s    Service Health Status Should Be Up    tedge-mapper-bridge-c8y
+    ${c1_fingerprint}=    Current Certificate Fingerprint
+
+    # Rotate the on-disk certificate to a fresh, *untrusted* C2 and revoke C1 in
+    # Cumulocity. The running mapper still holds C1 in memory, so nothing has
+    # changed for it yet.
+    Execute Command    tedge cert renew --ca self-signed
+    Delete Device Certificate From Platform    fingerprint=${c1_fingerprint}
```
Bravo555: suggestion: I'd like to avoid relying on Cumulocity APIs just to verify if we've restarted the mapper.

Could we perhaps change some config value like mappers's MQTT client id and then check a client with that id connected to the MQTT broker?

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
Bravo555, line 1: question: Will the other requirements from `spec.md` be tested here?

Or are the other requirements supposed to be tested using Rust unit tests?

Bravo555 (commented):
Overall in any case this PR has some nice structure improvements (separating building of actor connections from actually running them, which could make it more easy to write integration tests where we could run the entire runtime in the test in close-to-production configuration, only mocking input/output).
As for the feature itself, indeed the idea is very nice and the implementation shows that our actor architecture enables us to implement this relatively easily. There are still some TODOs and also some questions (e.g. should we run separate runtimes for separate components, but could we try to use one and perhaps spawn some actors only once and have them work with both agent and mapper). Though it still is some complexity (I haven't looked at the supervisor very closely yet), it's not _that_ much and as I understand it's not yet presented to be used by the user, so we can experiment with it and fix any issues before it's publicly documented. Very nice.

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
jarhodes314: Certainly the requirements around things like "does X get restarted if it crashes" and "do components subscribe to mosquitto in a persistent way" are pretty difficult to reliably test in the robot tests, so I think we need to lean on unit tests for those. I think the robot tests will be more valuable when the feature is more established, e.g. when we can run multiple mappers in a single process.

crates/core/tedge/src/cli/mod.rs
jarhodes314: I want this to be hidden/undocumented for the time being as it isn't very useful right now, and I think the interface is going to change (e.g. it takes a cloud argument at the moment and won't in the future once it's got multi-mapper support).

openspec/specs/single-process-supervisor/spec.md
jarhodes314: Good question, I think that's just something making a mistake

crates/common/flockfile/src/unix.rs
jarhodes314: Single-instance in this context means a single instance of any component, so this is meaningful for any way of running a component. I think this should be a warning - it's a very deliberate feature that can almost silently break (and it's easy to disable if it isn't desired).

I'm not sure how representative of a typical installation my local setup is, but my /run/lock directory isn't writable by tedge, so the locking was just not working for me and with this very subtle message to indicate that, hence my motivation to make it louder when it fails.

crates/core/tedge/src/cli/mod.rs
jarhodes314: I think the UX of the feature needs some consideration, but it's easy for this to descend into bikeshedding. Ultimately, the motivation for the feature is to not depend on an init system for running tedge services and possibly even control mosquitto too. With that in mind, this will run all the tedge services at some point in the future.

crates/core/tedge/src/supervisor.rs
jarhodes314: Yeah, I'll look to fix this now because this is quite important

crates/common/flockfile/src/unix.rs
didier-wenzek: > I'm not sure how representative of a typical installation my local setup is, but my /run/lock directory isn't writable by tedge, so the locking was just not working for me and with this very subtle message to indicate that, hence my motivation to make it louder when it fails.

Indeed, I would expect /run/lock to be open to any user

```
drwxrwxrwt 3 root root 120 Jun  8 09:07 /run/lock
```

The error is a bit verbose but that's okay.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -287,8 +265,41 @@ impl Agent {
         Ok(())
     }
 
+    /// Standalone entry point: a thin wrapper over [`Agent::build`].
+    ///
+    /// Preserves the pre-supervisor semantics exactly — it installs the agent's own
```
didier-wenzek: > Preserves the pre-supervisor semantics exactly

What does that mean?

crates/core/tedge_mapper/src/lib.rs
```diff
@@ -236,7 +231,45 @@ pub async fn run(mapper_opt: MapperOpt, config: TEdgeConfig) -> anyhow::Result<(
...
+    }
+}
+
+/// Builds a mapper runtime for supervised use.
+///
+/// Resolves the named mapper component, then wires every actor and spawns the
+/// runtime without installing signal handling, taking the single-instance lock, or
+/// running to completion. This is the rebuildable factory the single-process
+/// supervisor calls (repeatedly, on each restart) for a mapper unit.
+pub async fn build(name: MapperName, config: TEdgeConfig) -> anyhow::Result<Runtime> {
```
didier-wenzek, line 263: Could this be used to reload the config on `--SIGHUP`? Even for a non-supervised  mapper or agent? This would be great.

openspec/specs/single-process-supervisor/spec.md
```diff
@@ -0,0 +1,114 @@
...
+restarts failed components, owns process signal handling, and attributes logs and
+MQTT connections per component.
+
+## Requirements
+
+### Requirement: Single-process run mode
+
+tedge SHALL provide a run mode that supervises multiple core components (the agent
+and a mapper) within a single process, so that no external init system is required
+to start or keep the components running.
```
didier-wenzek, line 17: I appreciate this is a matter of taste, but I dislike this specification style and have a strong preference for specifications written as a regular user documentation as if the feature were already implemented


```suggestion
`tedge run all` provides a run mode that supervises multiple core components (the agent
and a mapper) within a single process, so that no external init system is required
to start or keep the components running.
```

didier-wenzek:
With `tedge run all`, *two* software list commands are systematically triggered on start - on by the agent, the other by the mapper.

```
2026-06-09T07:11:40.806686474Z  INFO C8Y: Clearing command topic topic=te/device/main///cmd/software_list/c8y-mapper-2026-06-09T07:11:40.125716803Z
2026-06-09T07:11:40.813730376Z  INFO C8Y: Clearing command topic topic=te/device/main///cmd/software_list/c8y-mapper-2026-06-09T07:11:40.114948066Z
```

This is also the case when the mapper and the agent are started independently. Couldn't we remove that behavior from the agent? After all, only the mapper care about that list of software.

crates/core/tedge_actors/src/runtime.rs
```diff
@@ -64,10 +66,20 @@ impl Runtime {
...
     fn with_events_sender(events_sender: Option<DynSender<RuntimeEvent>>) -> Runtime {
         let (actions_sender, actions_receiver) = mpsc::channel(16);
-        let runtime_actor =
-            RuntimeActor::new(actions_receiver, events_sender, Duration::from_secs(60));
+        // Capture the ambient span at construction so that every actor this runtime
+        // later spawns is attributed to it. Under the single-process supervisor each
+        // component is built inside a `component` span, which gives per-component log
+        // attribution in the shared process log stream. Standalone components build
+        // with no ambient span, so this is a no-op and their logs are unchanged.
+        let span = Span::current();
```
didier-wenzek: The intention is good, but this leads to bloated logs. Or more precisely, this makes the logs even more bloated. => We have to assign more systematically a target to tracing events (not in that PR, though).

```
2026-06-09T07:11:40.107609715Z  WARN component{name=tedge-mapper-c8y}: c8y_mapper_ext::converter: Failed to find the template file for `native-reboot`. Registration skipped
2026-06-09T07:11:40.122078919Z  INFO component{name=tedge-agent}:sm-agent: tedge_agent::operation_workflows::actor: Moving software_list operation to state: scheduled
```

crates/core/tedge_agent/src/agent.rs
```diff
@@ -287,8 +265,41 @@ impl Agent {
...
+    /// signal handling and runs to completion, exiting the process on a runtime
+    /// error (via [`Runtime::run_to_completion`]). The single-instance lock is taken
+    /// by the caller (see [`Agent::acquire_instance_lock`]) and held for the agent's
+    /// lifetime. Under the supervisor this wrapper is bypassed: the supervisor calls
+    /// [`Agent::build`] directly, owns signals centrally, and applies a restart
+    /// policy instead of exiting.
```
didier-wenzek: This generated comment is correct but not easy to understand because specifically written in the context of that PR (e.g. *pre-supervisor semantics* stands for the *behavior before that PR*).

=> This comment has to be rewritten.




```suggestion
    /// Start the agent as a standalone process and run it to completion.
    ///
    /// The agent is equipped with a SIGINT/SIGTERM/SIGQUIT handler.
    /// The process is terminated on runtime errors.
```

crates/core/tedge/src/supervisor.rs
```diff
@@ -0,0 +1,1121 @@
...
+//!
+//! Hosts the core thin-edge components (the agent and a mapper) inside one process
+//! so no external init system is required. Each component is a *unit*: a rebuildable
+//! factory plus a restart policy. The supervisor owns process-wide signal handling,
+//! isolates a crashing unit from its co-hosted peers (rebuilding it under a bounded
+//! backoff), and drains everything cleanly on termination.
+//!
+//! The standalone per-component path (`tedge-agent`, `tedge-mapper c8y`, …) is
+//! unaffected; this is a purely additive second consumer of each component's
+//! `build()` factory.
```
didier-wenzek: I really dislike this kind of wording: *a purely additive second consumer of each component's build factory*. Really?


```suggestion
//! By contrast, when the agent and the mappers run as standalone processes
//! (aka `tedge-agent`, `tedge-mapper c8y`, …) ,
//! runtime errors as well as SIGINT/SIGTERM/SIGQUIT signals terminate the process.
//! Any restart policy has then to be delegated to an external init system.
```

didier-wenzek (commented):
I really like the idea of supervisor. This is used in that PR to launch a mapper and the agent in a single process with the ability to restart one or the other after an error or a configuration update. I see many other use cases: 

- Using a supervisor even with standalone services (tedge-agent and tedge-mapper-c8y) to enable config reloading on SIGHUP.
- Running the builtin-bridge and a mapper under the control of a supervisor, so the bridge can be restarted independently on certificate update.

crates/core/tedge/src/supervisor.rs
jarhodes314: This should now be fixed

reubenmiller:
FYI: I ran into an issue when testing this PR where a subcomponent of the single process got itself into a MQTT reconnect loop, where it would connect to the local broker, but be immediately disconnected.

Given that only one `tedge` process was running, there is suspicion that maybe a specific actor wasn't shutdown fully and it resulted in a duplicate connection. Killing the tedge process was able to "resolve" the issue, but we need to better understand how this situation occurred.

I had a call with @jarhodes314 about it and provided some logs, and he's investigating it.

Device: Luckfox pico with 32MB RAM (armv7l)

albinsuresh (commented):
I haven't managed to simulate this with a test or anything, but I'm slightly concerned about the handling of ungraceful shutdown of actors in a component like mapper and its impact on the supervisor when it tries to replace the old one with the new one. We have often seen actors not always shutting down within the 60s grace period and the fallback was an `exit(1)`. If the same happens in this supervised mode, and the supervisor proceeds to replacing the old incarnation with a new one, wouldn't it lead to resource leaks that accumulates over time? Or we've got some safety nets that'll avoid this case?

reubenmiller:
> I haven't managed to simulate this with a test or anything, but I'm slightly concerned about the handling of ungraceful shutdown of actors in a component like mapper and its impact on the supervisor when it tries to replace the old one with the new one. We have often seen actors not always shutting down within the 60s grace period and the fallback was an `exit(1)`. If the same happens in this supervised mode, and the supervisor proceeds to replacing the old incarnation with a new one, wouldn't it lead to resource leaks that accumulates over time? Or we've got some safety nets that'll avoid this case?

I thought in this case we just hung because we were waiting for a graceful shutdown of the actor, and not because we couldn't forcefully shut it down. Or was the forceful shutdown really only doing exiting the process?

jarhodes314:
> I haven't managed to simulate this with a test or anything, but I'm slightly concerned about the handling of ungraceful shutdown of actors in a component like mapper and its impact on the supervisor when it tries to replace the old one with the new one. We have often seen actors not always shutting down within the 60s grace period and the fallback was an `exit(1)`. If the same happens in this supervised mode, and the supervisor proceeds to replacing the old incarnation with a new one, wouldn't it lead to resource leaks that accumulates over time? Or we've got some safety nets that'll avoid this case?

As of cc15828e6633f264d2ae3ce2875c23b20106c3dd, we should be killing the tasks after the timeout elapses, so there shouldn't be issues with this.

didier-wenzek:
One of the point of this PR being to reduce memory pressure, here are some figures showing this is promising.

Running under a load of 1k measurement per second (processed by the c8y mapper and a user-defined collectd flow) leading to 20kB/s of data sent to C8Y:

```
$ tedge mqtt sub 'c8y/#' | pv >/dev/null
1.50MiB 0:01:17 [19.9KiB/s] 
```

The memory foot print of `tedge run all c8y` is reduced by 2:

```
ps --long | where name == tedge | each {|p| {command: $p.command, cpu: $p.cpu, mem: $p.mem } }
╭───┬──────────────────────┬──────┬─────────╮
│ # │       command        │ cpu  │   mem   │
├───┼──────────────────────┼──────┼─────────┤
│ 0 │ tedge run all c8y    │ 8.62 │ 70.9 MB │
│ 1 │ tedge mqtt sub c8y/# │ 0.00 │ 38.6 MB │
╰───┴──────────────────────┴──────┴─────────╯
```

compared to the agent and the mapper running concurrently:

```
ps --long | where name == tedge | each {|p| {command: $p.command, cpu: $p.cpu, mem: $p.mem } }
╭───┬────────────────────────────┬──────┬─────────╮
│ # │          command           │ cpu  │   mem   │
├───┼────────────────────────────┼──────┼─────────┤
│ 0 │ tedge mqtt sub c8y/#       │ 0.00 │ 39.9 MB │
│ 1 │ tedge run tedge-agent      │ 0.00 │ 59.9 MB │
│ 2 │ tedge run tedge-mapper c8y │ 8.47 │ 67.3 MB │
╰───┴────────────────────────────┴──────┴─────────╯
```

reubenmiller:
> FYI: I ran into an issue when testing this PR where a subcomponent of the single process got itself into a MQTT reconnect loop, where it would connect to the local broker, but be immediately disconnected.
> 
> Given that only one `tedge` process was running, there is suspicion that maybe a specific actor wasn't shutdown fully and it resulted in a duplicate connection. Killing the tedge process was able to "resolve" the issue, but we need to better understand how this situation occurred.
> 
> I had a call with @jarhodes314 about it and provided some logs, and he's investigating it.
> 
> Device: Luckfox pico with 32MB RAM (armv7l)

The issue of spawning duplicate actors has been resolved, and it works nicely now.

didier-wenzek:
> > I haven't managed to simulate this with a test or anything, but I'm slightly concerned about the handling of ungraceful shutdown of actors in a component like mapper and its impact on the supervisor when it tries to replace the old one with the new one. We have often seen actors not always shutting down within the 60s grace period and the fallback was an `exit(1)`. If the same happens in this supervised mode, and the supervisor proceeds to replacing the old incarnation with a new one, wouldn't it lead to resource leaks that accumulates over time? Or we've got some safety nets that'll avoid this case?
> 
> As of [cc15828](https://github.com/thin-edge/thin-edge.io/commit/cc15828e6633f264d2ae3ce2875c23b20106c3dd), we should be killing the tasks after the timeout elapses, so there shouldn't be issues with this.

I confirm this is working properly.

- Tested with a broken `tedge_flows::Actor` that waits for ever on shutdown request.
- When a `SIGUSR1` is sent to `tedge run all c8y`
- The supervisor stops the mapper runtime, waiting for 60s before killing the non-responsive flow actor.
- And then properly restarts the mapper: `tedge connect c8y --test` being okay.

```
...
2026-06-11T13:12:23.456698577Z  INFO tedge::supervisor: SIGUSR1: restarting all mapper components
2026-06-11T13:12:23.457115328Z  INFO component{name=tedge-mapper-c8y}: Runtime: Shutting down
...
2026-06-11T13:13:23.459121916Z ERROR component{name=tedge-mapper-c8y}: Runtime: Timeout waiting for all actors to shutdown
2026-06-11T13:13:23.459412553Z ERROR component{name=tedge-mapper-c8y}: Runtime: Failed to shutdown: FlowsMapper-2
2026-06-11T13:13:23.460271248Z ERROR component{name=tedge-mapper-c8y}: Runtime: Failed to execute actor: task 704 was cancelled
2026-06-11T13:13:23.460809761Z  INFO tedge::supervisor: restarting component=tedge-mapper-c8y
...
```

reubenmiller (approved):
Approved (from a functional point of view)

We can do follow up extensions in follow up PRs, but is a great step forward to simplifying the deployment in specific scenarios like tedge-standalone and container deployments.

didier-wenzek (approved):
Approved.

The supervisor is a great addition that opens the door to many improvements.

crates/core/tedge/src/supervisor.rs
Bravo555: Ah, I had thought that only a simple `EnvFilter` would be required to fix this, but indeed it wouldn't suffice because with the old approach **we were setting a global log level** based on the service we were running, so if `tedge-agent` set to `DEBUG` called e.g. `reqwest`, we wouldn't show any of `reqwest`'s `DEBUG` events. 
To replicate the same behaviour with the supervisor and both components running under a single process required a more complete but more verbose solution of using a filter that looks for the `name` attribute of a span called `component` and applies the log level to everything below encompassed by that span, which is different from filtering based on `target` attribute that `EnvFilter` does. Initially I thought it could be done simpler, but now I see that probably not.

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
Bravo555: fixed

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
Bravo555: fixed: now we look at `time` timestamp of the mapper's up message, which doesn't rely on changing certificates or any C8y APIS

tests/RobotFramework/tests/tedge_run_all/run_all_c8y.robot
Bravo555: fixed: when test cases are run in random order with

```
/workspace/tests/RobotFramework/.venv/bin/python3 -m pabot.pabot --randomize tests --processes 10 --outputdir /workspace/tests/RobotFramework/output --listener RetryFa
iled:0 --exclude 'test:on_demand OR theme:benchmarks' --suite run_all_c8y --consolecolors on --consolemarkers on tests
```

The tests pass.

Bravo555 (approved):
My suggestions were also addressed, so approving, nicely done.

PR #4213, opened 2026-06-15
feat: Only publish messages from the bridge after all subscriptions have been acknowledged

jarhodes314:
## Proposed changes
If there is a poison message causing the connection to fail immediately after connection, this currently will likely be sent before we have subscribed to any topics on the remote broker (assuming we have a clean session, as is the case for c8y). This means that if the remote broker sends us an error message to help us debug the faulty message, we don't receive said message.

This PR adds a gate to the bridge that will collect messages until the subscriptions have been acknowledged, after which the gate will open and the bridge will publish any queued messages. In the case of non-clean (i.e. persistent) sessions, the gate will not be enabled. The gate is closed after any disconnection event.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/crates/extensions/tedge_mqtt_bridge/src/lib.rs
+++ b/crates/extensions/tedge_mqtt_bridge/src/lib.rs
@@ -42,6 +42,7 @@ use tedge_actors::RuntimeError;
 use tedge_actors::RuntimeRequest;
 use tedge_actors::RuntimeRequestSink;
 use tokio::sync::mpsc;
+use tokio::sync::watch;
 use tokio::task::JoinHandle;
 use tracing::Instrument;
 
@@ -163,7 +164,7 @@ impl MqttBridgeActorBuilder {
             .map(|t| SubscribeFilter::new(t.to_owned(), QoS::AtLeastOnce))
             .collect();
 
-        let [cloud_target, local_target] =
+        let ([cloud_target, local_target], [cloud_gate_controller, local_gate_controller]) =
             bidirectional_channel(cloud_client.clone(), local_client.clone(), in_flight.into());
         let [(convert_local, bidir_local), (convert_cloud, bidir_cloud)] =
             rules.converters_and_bidirectional_topic_filters();
@@ -194,6 +195,7 @@ impl MqttBridgeActorBuilder {
                     local_tx,
                     // The local→cloud direction enforces the cloud broker's limit
                     Some(max_payload_size),
+                    local_gate_controller,
                 )
                 .instrument(tracing::Span::current()),
             ),
@@ -213,6 +215,7 @@ impl MqttBridgeActorBuilder {
                     // The cloud→local direction is left unguarded: the local broker
                     // accepts large messages and cloud messages already met the cloud's limit
                     None,
+                    cloud_gate_controller,
                 )
                 .instrument(tracing::Span::current()),
             ),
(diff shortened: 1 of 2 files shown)


crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -232,13 +235,79 @@ fn bidirectional_channel<Client: MqttClient + 'static>(
...
+    /// Returns whether outbound publishes may currently proceed
+    fn is_open(&self) -> bool {
+        *self.open.borrow()
+    }
+
+    /// Resolves when the gate's open/closed state next changes
+    async fn await_change(&mut self) {
+        if self.open.changed().await.is_err() {
+            // The controller has been dropped: the state can no longer change
+            std::future::pending::<()>().await;
```
didier-wenzek, line 305: Could a caller be blocked waiting for a change? Actually that's okay, because the only caller used it as one the case of `select!` race.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -333,37 +407,39 @@ impl<Client: MqttClient + 'static> BridgeAsyncClient<Client> {
...
+                            Some(BridgeMessage::BridgeAck { publish }) => {
+                                target.ack(&publish).await.unwrap();
+                                acknowledged.fetch_add(1, Ordering::Relaxed);
+                            }
+                            Some(message) => {
+                                if gate.is_open() && buffer.is_empty() {
+                                    publish_to_target(&target, &tx, &published, message).await;
+                                } else {
+                                    buffer.push_back(message);
+                                }
```
didier-wenzek: I understand the check that `buffer.is_empty()` as a safety net - the buffer being local to the loop and only populated when the gate is closed. Am I missing a race condition?

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -678,8 +816,27 @@ async fn half_bridge(
...
+                {
+                    log_event!(warn: name, "Broker rejected one or more subscriptions: {suback:?}");
+                }
+                // Count return codes, not packets: a broker may acknowledge several filters
+                // in one SUBACK, and a rejected filter still counts since we must never wait
+                // on an acknowledgement that will never arrive.
+                subacks_seen += suback.return_codes.len();
+                if subacks_seen >= subacks_needed {
+                    gate.open();
+                }
```
didier-wenzek, line 857: Simply counting acks seems a bit weak. But yeah I appreciate that the `SubAck` payload doesn't ease a proper check against a topic set. Furthermore, this is only for opening a gate.

didier-wenzek (approved):
Approved

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: If a message is published just as the gate opens, we could find ourselves in this branch without hitting the `await_change` branch prior and flushing the buffer, so yes, it is technically possible for this to fail.

I think the simplest solution is to just lean into the queue, so the publish branch always pushes to the back of the queue, then the top of the loop clears this queue. It's less logic and clearer what's going on as we only publish messages from a single place.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: We could check for the various pkids, but that seems somewhat overkill here. And I'm cautious about making the check too strict given it has the potential to completely break the bridge if either broker misbehaves.

didier-wenzek: >  that seems somewhat overkill here.

Agree

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: > I think the simplest solution is to just lean into the queue, so the publish branch always pushes to the back of the queue, then the top of the loop clears this queue. It's less logic and clearer what's going on as we only publish messages from a single place.

Now done in bd356a2

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: I've added some small comments to communicate this in bd356a2

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -339,37 +416,36 @@ impl<Client: MqttClient + 'static> BridgeAsyncClient<Client> {
...
-                                .publish(target_topic, publish.qos, publish.retain, publish.payload)
-                                .await
-                                .unwrap();
-                            published.fetch_add(1, Ordering::Relaxed);
+                // While the gate is closed (the target's subscriptions are not yet
+                // acknowledged on a clean session), outbound publishes are held here in
+                // order and flushed once the gate opens. Every publish goes through the
+                // buffer, so they are always forwarded in arrival order. Acks are never
+                // gated.
+                let mut buffer: VecDeque<BridgeMessage> = VecDeque::new();
```
albinsuresh, line 427: I was initially worried that this unbounded queue introduced the risk of OOME during a prolonged offline period where telemetry messages (likely QoS 0) would have just filled this queue up and pushed the mapper to an OOME. Unlike the QoS 1 messages which are bounded by the `in_flight` limits and the cloud PUBACK chain back to the local half bridge, the QoS 0 messages, aren't "bounded" that way right?

But on looking back at the old code, I realised that this fix did not introduce this OOME case, but just moved the source of that OOME from the old cloud half bridge's `unbounded_rx` to this `VecDeque`. While it's not a regression, we'll have to address this separately as offline periods would be quite common for these devices.

albinsuresh (approved):
Impressive work, with excellent test coverage. Just one remark on an OOME risk that I feel is still there, but fixing that is definitely not under the scope of this PR.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: I'll check, but I'm pretty sure (and working in the assumption that the broker publishing messages will only allow a certain number of in-flight messages at once, so publishing nore messages than this to the local broker will just cause them to queue with mosquitto (where they are probably held in memory but I'm not convinced there's a lot we can do there).

jarhodes314: Yep, mosquitto at least has configurations for: 
- `max_inflight_messages`: it will send us up to 20 >QoS1 messages at a time if we don't ack them
- `max_queued_messages`: it will queue up to 1000 messages for us, after that it will drop the oldest messages to make room for new messages

So that should protect us from such events.

Issue #3165, opened 2024-10-07
avoid using colons in tedge-agent command log file names


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The current tedge-agent log file names under `/var/log/tedge/agent/*.log` use the colon (`:`) character which results in slight annoyances when selecting files names on the console, or transferring the log files to NTFS storage mediums.

This mostly affects the log files created when the command id is based on a timestamp (e.g. RFC3339).

Below shows an example of some of the files which make use of colons (`:`):

```sh
$ ls -ltr /var/log/tedge/agent

total 80
-rw-r--r-- 1 tedge tedge 14652 Oct  7 06:47 workflow-software_list-c8y-mapper-2024-10-07T06:47:36.491494387Z.log
-rw-r--r-- 1 tedge tedge 10185 Oct  7 06:47 workflow-software_list-c8y-mapper-2024-10-07T06:47:40.881845183Z.log
-rw-r--r-- 1 tedge tedge 10185 Oct  7 06:47 workflow-software_list-c8y-mapper-2024-10-07T06:47:41.047587459Z.log
-rw-r--r-- 1 tedge tedge   519 Oct  7 06:49 c8y_RemoteAccessConnect-2024-10-07T06:49:52.706238934Z.log
-rw-r--r-- 1 tedge tedge 10170 Oct  7 06:53 workflow-software_list-c8y-mapper-2024-10-07T065303.502675425Z.log
-rw-r--r-- 1 tedge tedge 10173 Oct  7 06:53 workflow-software_list-c8y-mapper-2024-10-07T065303.578293335Z.log
-rw-r--r-- 1 tedge tedge   519 Oct  7 06:53 c8y_RemoteAccessConnect-2024-10-07T06:53:08.020012572Z.log
-rw-r--r-- 1 tedge tedge  2487 Oct  7 07:43 workflow-log_upload-c8y-mapper-40290121.log
-rw-r--r-- 1 tedge tedge  2433 Oct  7 07:43 workflow-log_upload-c8y-mapper-40290123.log
```

See the **Additional Context** section for more examples where this incompatibility was discovered.

**Describe the solution you'd like**

Avoid using colons (`:`) in the log file names to improve portability and usability of the log files.

It should be noted that it might be also worthwhile changing the tedge MQTT command id (which is where the filename is being derived from) to also avoid using colons (`:`) so that the command id and filename remain in sync.

There are a few options to consider how to format the log file names:

[...]

PR #4214, opened 2026-06-16
fix: avoid using colons in file names

didier-wenzek:
## Proposed changes

- [x] Use unix timestamps instead of Rfc3339 in log filenames
- [ ] What about log files for cloud operations?
- [ ] What about child device names used as directory names?
- [ ] What about `.agent`?

## Types of changes

## Paste Link to the issue

- https://github.com/thin-edge/thin-edge.io/issues/3165

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/mqtt_topics.rs
+++ b/crates/core/tedge_api/src/mqtt_topics.rs
@@ -12,7 +12,6 @@ use std::convert::Infallible;
 use std::fmt::Display;
 use std::fmt::Formatter;
 use std::str::FromStr;
-use time::format_description;
 use time::OffsetDateTime;
 
 const ENTITY_ID_SEGMENTS: usize = 4;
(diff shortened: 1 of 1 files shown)


crates/core/tedge_api/src/mqtt_topics.rs
```diff
@@ -960,13 +959,9 @@ impl IdGenerator {
...
-        format!(
-            "{}-{}",
-            self.prefix,
-            OffsetDateTime::now_utc()
-                .format(&format_description::well_known::Rfc3339)
-                .unwrap(),
-        )
+        let now = OffsetDateTime::now_utc();
+        let millis = now.unix_timestamp_nanos() / 1_000_000;
+        format!("{}-{}.{}", self.prefix, millis / 1000, millis % 1000)
```
albinsuresh, line 964: I understand that we agreed on UNIX timestamp for this earlier and that's completely fine for something like a command `id`.

But, when used in log file names, I'm a bit worried if it'll make debugging harder for **humans**. For example, when you get a diagnostic zip with multiple log files and a rough estimate of the incident time (which the customer would usually communicate in standard time like "the incident happened on 5th July, between 10 AM and 11 AM"), we'll be forced to do that UNIX timestamp to human timestamp conversion across multiple log files to even narrow down the right one to look at.

So, I'd still be in favour of something more human readable like `2026-06-16T10h10m48s` using `h`, `m` and `s` separators for the time value:
```suggestion
        let format = time::macros::format_description!("[year]-[month]-[day]T[hour]h[minute]m[second]s");
        
        now.format(&format).expect("valid format")
```

Issue #4216, opened 2026-06-16
enforce strict message payload limits on the built-in bridge


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

During the investigation of https://github.com/thin-edge/thin-edge.io/issues/4209, it was discovered that:

* publishing directly to the `c8y/` topics bypasses message payload checks as the checks only occur within the flows (before the message is published to `c8y/`
* publishing large messages to the tedge `/twin` topic doesn't have any payload size checks.

In the event when a large payload is sent to Cumulocity, the cloud connection is either disconnected, or some other cases it becomes unresponsive, resulting in a reconnection loop until the problematic messages are purged from the MQTT session history.

**Describe the solution you'd like**

Make the bridge configuration aware of payload size restrictions, and reject messages before sending them to the cloud. Different clouds generally have different size limits, so should be configurable. In such cases where a message is rejected, it should be visible in the logs (and MQTT output?) that the message has been dropped due to exceeding the max allowed cloud MQTT payload size.

**Describe alternatives you've considered**

**Additional context**

The problem can be reproduced with the following one-liner (using python):

```sh
tedge mqtt pub -r te/device/main///twin/large_payload "$(python3 -c 'import json; print(json.dumps({f"key{i}": "x"*1000 for i in range(1, 100)}))')"
```

If you need to purge the problematic messages, then you can run the following code

```
echo "Stopping thin-edge.io and mosquitto"
sudo tedge disconnect c8y
sudo systemctl stop tedge-agent
sudo systemctl stop mosquitto

echo "Purging large messages from cache"
sed '/^.\{15000\}./d' /etc/tedge/.agent/entity_store.jsonl | sudo tee /etc/tedge/.agent/entity_store.jsonl.tmp >/dev/null
mv /etc/tedge/.agent/entity_store.jsonl.tmp /etc/tedge/.agent/entity_store.jsonl

mv /var/lib/mosquitto/mosquitto.db /var/lib/mosquitto/mosquitto.db.corrupt.large_messages 2>/dev/null ||:

echo "Starting thin-edge.io and mosquitto"
sudo tedge reconnect c8y
```

reubenmiller:
The change works nicely, and now the message is dropped if it exceeds the defined. A log entry on the tedge-mapper will indicate the max payload violation:

```sh
sudo journalct -fu tedge-mapper-c8y
[...]

PR #4217, opened 2026-06-16
feat: Respect `max_payload_size` in tedge-mqtt-bridge

jarhodes314:
## Proposed changes
- Enforce a strict limit of the max payload size for the builtin bridge, dropping messages (with a warning) if they exceed the limit
- Refine calculation for the size of a message when deciding in `tedge-mapper-c8y` whether to use MQTT or HTTP, taking into account the full packet size, instead of just the payload itself

## Types of changes

## Paste Link to the issue
- #4216

## Checklist

### DIFF
--- a/crates/common/mqtt_channel/src/messages.rs
+++ b/crates/common/mqtt_channel/src/messages.rs
@@ -181,6 +181,32 @@ impl DebugPayload {
 /// A message payload
 pub type Payload = Vec<u8>;
 
+/// The number of bytes the MQTT "remaining length" field occupies for a packet
+/// whose remaining length is `len`
+fn remaining_length_size(len: usize) -> usize {
+    if len >= 2_097_152 {
+        4
+    } else if len >= 16_384 {
+        3
+    } else if len >= 128 {
+        2
+    } else {
+        1
+    }
+}
+
+/// Computes the wire size, in bytes, of an MQTT PUBLISH packet with the given
+/// topic, QoS and payload length
+///
+/// Matches `rumqttc::Publish::size()`
+pub fn publish_packet_size(topic: &str, qos: QoS, payload_len: usize) -> usize {
+    const TOPIC_LENGTH_SIZE: usize = 2; // 2 bytes for the topic length prefix
+    const FIXED_HEADER_SIZE: usize = 1; // 1 byte for the fixed header
+    let packet_id_size = if qos == QoS::AtMostOnce { 0 } else { 2 };
+    let remaining_length = TOPIC_LENGTH_SIZE + topic.len() + packet_id_size + payload_len;
+    FIXED_HEADER_SIZE + remaining_length_size(remaining_length) + remaining_length
+}
+
 impl MqttMessage {
     pub fn new<B>(topic: &Topic, payload: B) -> MqttMessage
     where
@@ -227,6 +253,11 @@ impl MqttMessage {
     /// Split the message into a (topic, payload) pair
     pub fn split(self) -> (String, Payload) {
         (self.topic.name, self.payload.0)
+    }
+
+    /// The size, in bytes, this message occupies as an MQTT PUBLISH packet on the wire
+    pub fn wire_size(&self) -> usize {
+        publish_packet_size(&self.topic.name, self.qos, self.payload().len())
     }
 }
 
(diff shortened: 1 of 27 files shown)


crates/core/tedge_mapper/src/custom/config.rs
```diff
@@ -118,6 +120,26 @@ pub struct BridgeConfig {
     /// TLS transport control: on, off, or auto (default).
     #[serde(default)]
     pub tls: BridgeTls,
+    /// Maximum MQTT packet size the bridge forwards to the cloud for this mapper.
+    /// Defaults to the MQTT maximum, leaving the limit effectively disabled for
+    /// brokers whose limit is unknown.
+    #[serde(default = "default_max_payload_size")]
+    pub max_payload_size: MqttPayloadLimit,
```
albinsuresh, line 127: I was tempted to suggest the name `max_packet_size`, but I'm assuming we were forced to stick to this name because of the legacy mapper configs named `max_payload_size`.

crates/extensions/tedge_flows/src/transformers/limit_payload_size.rs
```diff
@@ -33,7 +33,7 @@ impl Transformer for LimitPayloadSize {
         _context: &FlowContextHandle,
     ) -> Result<Vec<Message>, FlowError> {
         if let Some(max_size) = self.max_size {
-            if message.payload.len() > max_size {
+            if message.wire_size() > max_size {
```
albinsuresh, line 36: Isn't that a change in semantics for this transport agnostic transformer, especially for non-MQTT message types (unless this transformer was only intended to be used on MQTT messages)?

crates/core/tedge_mapper/src/custom/config.rs
didier-wenzek, line 127: It would be simpler to `impl Default for MqttPayloadLimit`, as both that type and its actual default value (i.e. `QTT_MAX_PAYLOAD_SIZE`) are defined in `tedge_config::models`.

crates/common/mqtt_channel/src/messages.rs
```diff
@@ -181,6 +181,30 @@ impl DebugPayload {
...
+}
+
+/// Computes the wire size, in bytes, of an MQTT PUBLISH packet with the given
+/// topic, QoS and payload length
+///
+/// Matches `rumqttc::Publish::size()`
+pub fn publish_packet_size(topic: &str, qos: QoS, payload_len: usize) -> usize {
+    let packet_id_size = if qos == QoS::AtMostOnce { 0 } else { 2 };
+    let remaining_length = 2 + topic.len() + packet_id_size + payload_len;
+    1 + remaining_length_size(remaining_length) + remaining_length
```
albinsuresh: Documenting those magic numbers `1` (fixed-header byte) and `2` (topic-length) would be nice.

crates/extensions/tedge_mqtt_bridge/src/lib.rs
```diff
@@ -617,8 +624,24 @@ async fn half_bridge(
...
-                        target.publish(topic.to_string(), publish);
+                        let wire_size = mqtt_channel::publish_packet_size(
+                            topic.as_ref(),
+                            publish.qos,
+                            publish.payload.len(),
+                        );
+                        if let Some(limit) = max_payload_size.filter(|&limit| wire_size > limit) {
+                            // The message is too large to ever be accepted by the cloud broker.
+                            // Acknowledge it locally so it is not redelivered, and drop it rather
+                            // than let it wedge the cloud connection.
```
didier-wenzek: What do you mean by *wedge*? I'm not used to that term.

didier-wenzek (approved):
Approved

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
```diff
@@ -77,7 +77,9 @@ pub use mqtt_config::TEdgeMqttClientAuthConfig;
 
 const DEFAULT_ROOT_CERT_PATH: &str = "/etc/ssl/certs";
 
-pub const C8Y_MQTT_PAYLOAD_LIMIT: u32 = 16184; // 16 KB
+// 200 bytes below Cumulocity's real 16384 limit, as headroom for payload growth
```
albinsuresh, line 80: Is this safety net still relevant, as we check the `wire_size` on the converted message itself?

albinsuresh (commented):
The impl looks fine. Just a few clarification comments.

PS: I'm really liking this spec driven development workflow that gives an overview of what was done the reviewer (the spec) before diving into the code

crates/core/tedge_mapper/src/custom/config.rs
jarhodes314: We could rename it (and I agree it's a more sensible name) as we have the deprecated key logic so the old name can still work if we rename it, though that's probably best as a separate PR for changelog purposes.

crates/common/tedge_config/src/tedge_toml/tedge_config.rs
jarhodes314: I think it isn't relevant now. There's still a potential difference between the messages the mapper checks and the messages that are actually sent, because the topic changes from `c8y/s/us` to `s/us` when the message goes to Cumulocity, although that always reduces the message size so that shouldn't be a problem.

crates/extensions/tedge_flows/src/transformers/limit_payload_size.rs
jarhodes314: I think at the moment, this can only be an MQTT message. I think in practical terms, (currently) this is only usefully a step in processing MQTT messages so it probably doesn't matter.

albinsuresh (approved):

crates/common/mqtt_channel/src/messages.rs
jarhodes314: Now done

crates/extensions/tedge_mqtt_bridge/src/lib.rs
jarhodes314: I've now changed this to _block_

crates/core/tedge_mapper/src/custom/config.rs
jarhodes314: Now done

Issue #4210, opened 2026-06-11
Reduce memory usage by log plugins when calling the get command


reubenmiller:
**Is your feature improvement request related to a problem? Please describe.**

The log plugin inefficiently uses memory when processing the `get` command, and leads to the plugin being killed by the Out-of-Memory (OOM) Killer, resulting in retrieval of the logs being impossible for larger files.

The code snippet from [crates/extensions/tedge_log_manager/src/plugin.rs#L125-L130](https://github.com/jarhodes314/thin-edge.io/blob/b68731e8318de48a1df2f7df9f30d09ec5e4f957/crates/extensions/tedge_log_manager/src/plugin.rs#L125-L130), shows that the `log-plugin get` executes and loads the entire stdout/stderr into memory. Given that the function of a log plugin is to present log entries, this could be very large (i.e. many many MBs and in worst case scenarios GBs).

```
        let output = self.execute(command, None).await?;

        if !output.status.success() {
            let stderr = String::from_utf8_lossy(&output.stderr);
            return Err(self.plugin_error(format!("Get command error: {}", stderr)));
        }
```

**Describe the solution you'd like**

The log plugin should stream the output of the `get` command to minimize the amount of memory used. The log file upload should also be checked if it is also processing the file upload in a streaming manner as well to avoid similar OOM killer issues.

**Describe alternatives you've considered**

**Additional context**

PR #4222, opened 2026-06-18
fix: avoid reading external log plugin's stdout to memory

Bravo555:
## Proposed changes

When spawning an external log plugin process and running its "get" command, instead of using pipes to keep its output in memory, attach stdout to a file and read the file after process exits. This is done to avoid reading the entire output into memory as on some limited RAM environments it might be too big.

## Types of changes

## Paste Link to the issue

- #4210

## Checklist

### DIFF
--- a/crates/extensions/tedge_log_manager/src/plugin.rs
+++ b/crates/extensions/tedge_log_manager/src/plugin.rs
@@ -3,8 +3,12 @@ use camino::Utf8Path;
 use std::collections::BTreeSet;
 use std::collections::VecDeque;
 use std::fs::File;
+use std::fs::OpenOptions;
+use std::io::BufRead;
+use std::io::BufReader;
 use std::io::Write;
 use std::path::PathBuf;
+use std::process::ExitStatus;
 use std::process::Output;
 use std::sync::Arc;
 use tedge_api::CommandLog;
@@ -58,6 +62,14 @@ impl ExternalPluginCommand {
             .await
             .map_err(|err| self.plugin_error(err))?;
         Ok(output)
+    }
+
+    pub async fn status(&self, command: LoggedCommand) -> Result<ExitStatus, LogManagementError> {
+        let status = command
+            .status()
+            .await
+            .map_err(|err| self.plugin_error(err))?;
+        Ok(status)
     }
 
     pub fn plugin_error(&self, err: impl std::fmt::Display) -> LogManagementError {
@@ -118,18 +130,67 @@ impl ExternalPluginCommand {
             command.arg(until_time.unix_timestamp().to_string());
         }
 
+        let stdout_file =
+            tempfile::NamedTempFile::new_in(self.tmp_dir.as_std_path()).map_err(|err| {
+                self.plugin_error(format!(
+                    "Failed to create temporary file in {}: {err}",
+                    tempfile::env::temp_dir().to_string_lossy()
+                ))
+            })?;
+        let stderr_file =
+            tempfile::NamedTempFile::new_in(self.tmp_dir.as_std_path()).map_err(|err| {
+                self.plugin_error(format!(
+                    "Failed to create temporary file in {}: {err}",
+                    tempfile::env::temp_dir().to_string_lossy()
+                ))
+            })?;
+
+        let stdout = OpenOptions::new()
+            .create(false)
+            .read(false)
+            .write(true)
+            .open(stdout_file.path())
+            .map_err(|err| {
+                self.plugin_error(format!(
+                    "failed to open temporary file for writing process output '{}': {err}",
+                    stdout_file.path().to_string_lossy()
+                ))
+            })?;
+        let stderr = OpenOptions::new()
+            .create(false)
+            .read(false)
+            .write(true)
+            .open(stderr_file.path())
+            .map_err(|err| {
+                self.plugin_error(format!(
+                    "failed to open temporary file for writing process output'{}': {err}",
+                    stderr_file.path().to_string_lossy()
+                ))
+            })?;
+
+        command.stdout(stdout);
+        command.stderr(stderr);
+
         debug!(
             target: "log plugins",
             "Fetching log using command: {}", command
         );
-        let output = self.execute(command, None).await?;
+        let status = self.status(command).await?;
 
-        if !output.status.success() {
-            let stderr = String::from_utf8_lossy(&output.stderr);
-            return Err(self.plugin_error(format!("Get command error: {}", stderr)));
+        if !status.success() {
+            // we assume stderr is short enough to load in memory
+            let stderr = std::fs::read_to_string(stderr_file.path())?;
+            return Err(self.plugin_error(format!("Get command error: {stderr}",)));
         }
 
-        let stdout = String::from_utf8_lossy(&output.stdout);
+        let stdout = File::open(stdout_file.path()).map_err(|err| {
+            self.plugin_error(format!(
+                "failed to open temporary file for writing process output'{}': {err}",
+                stdout_file.path().to_string_lossy()
+            ))
+        })?;
+        let stdout = BufReader::new(stdout);
+
         let file = File::create(output_file_path).map_err(|err| {
             self.plugin_error(format!(
                 "Failed to create plugin output file at {} due to {}",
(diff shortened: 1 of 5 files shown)


tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -335,6 +335,49 @@ Plugin dynamic filter addition
...
+    ...    ${start_timestamp}
+    ...    ${end_timestamp}
+    ...    log_type=huge::huge
+
+    # WARNING(flaky)
+    Sleep    15s
+
+    # make sure unit wasn't killed by systemd OOM killer (journalctl --grep returns 1 if no match)
+    Execute Command
+    ...    cmd=journalctl -u tedge-agent -e --grep="Failed with result 'oom-kill'"
```
jarhodes314: Instead of actually trying to recreate the OOM, can we not just use `systemctl status` to find the peak memory usage of the program and check it's a vaguely sensible number?

Bravo555: Indeed at first I wanted to check peak memory usage, but `systemctl status` in the devcontainer showed me only current memory usage, without peak. I vaguely remember changing config options to try to see the peak usages, but wasn't successful. I'll try again with the test image to make sure if it can be done, or what's the reason if it can't.

crates/extensions/tedge_log_manager/src/plugin.rs
```diff
@@ -118,18 +130,65 @@ impl ExternalPluginCommand {
             command.arg(until_time.unix_timestamp().to_string());
         }
 
+        let stdout_file = tempfile::NamedTempFile::new().map_err(|err| {
```
albinsuresh:
```suggestion
        let stdout_file = tempfile::NamedTempFile::new_in(&self.tmp_dir).map_err(|err| {
```

crates/extensions/tedge_log_manager/src/plugin.rs
```diff
@@ -118,18 +130,65 @@ impl ExternalPluginCommand {
...
+            self.plugin_error(format!(
+                "Failed to create temporary file in {}: {err}",
+                tempfile::env::temp_dir().to_string_lossy()
+            ))
+        })?;
+        let stderr_file = tempfile::NamedTempFile::new().map_err(|err| {
```
albinsuresh: Same as above.

tests/RobotFramework/tests/cumulocity/log/huge_plugin.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+case "$1" in
+  list)
+    echo huge
+    ;;
+  get)
+    if [ "$2" = "huge" ]; then
+      # Emit ~10000MB to stdout to simulate an oversized logfile
+      head -c 10000000000 /dev/zero | tr '\0' 'A'
```
albinsuresh: A bit sceptical about the impact of this 10 GB worth of bytes being emitted from a process on a CI machine. Instead, can we run the `tedge-agent` with very limited RAM and reproduce the same test with a much smaller output?

albinsuresh (commented):
The fix looks fine. Just a few minor concerns.

tests/RobotFramework/tests/cumulocity/log/huge_plugin.sh
Bravo555: systemd's [`MemoryMax`](https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html?__goaway_challenge=meta-refresh&__goaway_id=f91afe6013168625311e629e379b8655&__goaway_referer=https%3A%2F%2Fduckduckgo.com%2F#MemoryMax=bytes) option sets a hard limit of memory used by a unit. The test sets `MemoryMax` to `25M`, and when I looked at `systemctl status tedge-agent` while the test was running, I saw the `MemoryCurrent`, i.e. actual memory used, to be `25M`. The big size is mainly to be sure that we're attempting to use more memory for long enough of a window for OOM killer to be invoked, which could be flaky if the window is too brief. But indeed maybe `10G` could be a problem if for some reason systemd wouldn't do its job actually limit it, impacting other tests, so I'll see if I can bring it down to `100M-1G` range and maybe use some sleeps in the scripts to make sure OOM killer is consistently invoked (or to not rely on it and just check peak usage as James suggested)

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
Bravo555: fixed in 65e329bed

Changed the test to not depend on systemd and do manual RSS measurement at a known point when stdout is filled.

crates/extensions/tedge_log_manager/src/plugin.rs
Bravo555: fixed in 65e329bed

crates/extensions/tedge_log_manager/src/plugin.rs
Bravo555: fixed in 65e329bed

tests/RobotFramework/tests/cumulocity/log/huge_plugin.sh
Bravo555: fixed in 65e329bed

Now we generate only a 50MB file in the test image tmpfs and limit memory usage to 20MB and measure at a well-known point where stdout should be filled.

Actually the previous version of the test had a bug where because tmpfs was only 64MB, full 10G wasn't even generated and it failed on the `head ...` command with `no more space left on the device` and proceeded running the other steps because `set -euo ...` was missing. Now the script properly fails on first failed command and we verify that it ran without errors and that tedge-agent stays alive.

tests/RobotFramework/tests/cumulocity/log/huge_plugin.sh
```diff
@@ -0,0 +1,34 @@
...
+      len=50M
+      dd if=/dev/zero bs=$len count=1 2>/dev/null | tr '\0' 'A'
+
+      # once the entire log is generated output a message so we know we can measure memory used
+      echo "script generated $len of data" >&2
+
+      # create a pipe and read it to pause executing, allowing test to measure memory usage and
+      # decide when the script should exit
+      mkfifo /tmp/huge-plugin-pipe
+      read -r _x < /tmp/huge-plugin-pipe
```
albinsuresh, line 20: Neat trick.

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
```diff
@@ -335,6 +335,44 @@ Plugin dynamic filter addition
...
+    Should Contain Supported Log Types    huge::huge
+    ${start_timestamp}=    Get Current Date    UTC    -1 hours    result_format=%Y-%m-%dT%H:%M:%S+0000
+    ${end_timestamp}=    Get Current Date    UTC    +1 hours    result_format=%Y-%m-%dT%H:%M:%S+0000
+    ${operation}=    Create Log Request Operation
+    ...    ${start_timestamp}
+    ...    ${end_timestamp}
+    ...    log_type=huge::huge
+
+    # Wait until huge plugin creates a pipe which means stdout was filled and we can measure memory usage
+    Execute Command    ls /tmp/huge-plugin-pipe
```
albinsuresh: Does the `Execute Command` keyword really "wait" until the given command is successful as the comment seems to suggest? Not that I'm aware of. Since we're doing this check immediately after triggering the operation from the cloud in the previous step, depending on the time taken for the command to reach the agent from the cloud via the mapper, and the agent executing it, this pipe may or may not exist. So, I feel this could be flaky.

albinsuresh (commented):
Nice work. Happy to approve after one test concern is clarified.

tests/RobotFramework/tests/cumulocity/log/log_operation_plugins.robot
Bravo555: Technically I'm relying here on the automatic retries with the `Execute Command` keyword, which by default retry every 2s up to 15 times, so it's more or less guaranteed to work, but indeed I could've added `retry=` parameter here to make it fully explicit or alternatively some sort of blocking bash command that waits until the file exists but then if for some reason it doesn't appear then it waits indefinitely. So I'd opt in for keeping the current form and just add retry for clarification.

jarhodes314: As someone that's frequently surprised that retries are opt-out for `Execute Command`, I think it's sensible to spell the retries and timeout out here as this step in particular relies on them.

albinsuresh (approved):
LGTM.

The handling of `config_snapshot` operation using plugins would also have the same limitation where the plugin output is read into memory by the agent. It'd be nice to fix that as well, in a different PR of course, to keep things consistent across components.

Bravo555:
> The handling of config_snapshot operation using plugins would also have the same limitation where the plugin output is read into memory by the agent

Though arguably it probably wouldn't be as useful, as if you have a config file that's bigger than available RAM, then you probably have other problems ;)

PR #4233, opened 2026-06-25
spec: entity registration MQTT API revamp

albinsuresh:
## Proposed changes

A design doc detailing the existing issues with the entity registration MQTT APIs and a proposal to address them, with minimal backward compatibility breakage.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/design/decisions/0009-entity-registration-api-revamp.md
+++ b/design/decisions/0009-entity-registration-api-revamp.md
@@ -1,0 +1,280 @@
+# Entity registration API revamp
+
+* Date: __2026-06-24__
+* Status: __Draft__
+
+## Background
+
+%%te%% exposes two APIs to manage entities (devices and services):
+
+* an **MQTT API**
+  ([docs](../../docs/src/operate/entity-management/mqtt_api.md),
+  [reference](../../docs/src/references/mqtt-api.md)),
+  where an entity is registered by publishing a **retained** message
+  to its entity-entity metadata topic (e.g. `te/device/child01//`), and
+* an **HTTP API** ([docs](../../docs/src/operate/entity-management/rest_api.md)) served by `tedge-agent`,
+  which validates the request and, on success, publishes the same retained registration message to the broker.
+  With this, the `tedge-agent` was made the central owner of the entity store;
+
+The original MQTT API remains the one most users rely on.
+The two APIs have drifted slightly in both behaviour and validation,
+and the underlying retained-message model has shown structural problems that this document captures,
+before proposing a way forward.
+
+This document is concerned with the **registration / update / delete** lifecycle
+and the **twin data** that rides along with it.
+
+## Root cause: the broker plays three incompatible roles
+
+In the current design the MQTT broker is, at the same time:
+
+1. the **ingest channel** — any client may publish a retained registration message to any entity metadata topic;
+2. the **persistence layer** — the retained messages are what components like the mappers rely on for persistence,
+   despite the agent maintaining a persistent entity store.
+3. the **queryable source of truth** —
+   components treat the retained messages on `te/+/+/+/+` as the list of registered entities.
+   They don't query the `tedge-agent`, which has an authoritative entity store.
+
+There is **no validation gate** between role 1 and role 3.
+A client write becomes "the truth" the instant it is retained,
+regardless of whether `tedge-agent` accepts it.
+The agent validates *after the fact* and has no authority to correct what is already on the broker.
+Almost every issue below is a symptom of collapsing these three roles into one writable, unvalidated namespace.
+
+The HTTP API does not escape this:
+a successful HTTP registration is still published to the broker as a retained message
+so an HTTP-created entity can be overwritten by any MQTT client afterwards.
+HTTP buys validated *writes*, not a protected *source of truth*.
+
+## Catalogue of issues
+
+### A. Source-of-truth integrity
+
+1. **Rejected registrations leave broken retained messages.**
+   A bad registration message, rejected by the agent, stays with the broker even after the rejection.
+   There is no compensating publish.
+   The broker keeps serving a message the agent refuses to honour,
+   so the broker's view and the agent's view disagree permanently and never reconverge.
+
+2. **A valid retained message can be replaced by an invalid one.**
+   Because the entity metadata topic is writable by anyone,
+   a second publisher can clobber a good registration with a bad one.
+   The agent rejects the new message but, per issue 1, the bad message stays retained.
+
+3. **Registration and update share one channel, with surprising replace semantics.**
+   An MQTT "update" is a **full struct replace**, not a merge:
+   the code builds `EntityMetadata { twin_data: merged, ..entity_metadata }`,
+   so `@id`, `@type`, `@parent` and `@health` all come from the new message.
+   * An update that omits `@id` **silently drops** the external id (`@id` becomes `None`).
+   * This contradicts the documentation, which states `@type` and `@id` "cannot be updated after the registration".
+     Nothing enforces that — both can be silently changed or dropped on re-publish.
+   * The MQTT update example in the docs republishes only `@type`+`@parent` (no `@id`),
+     which would wipe the id of an entity originally registered with one.
+   * Twin data is *merged* while entity metadata is *replaced* —
+     two different merge semantics in one message, undocumented.
+
+4. **Empty payload is overloaded as "delete".**
+   An entity can never have an intentionally empty registration,
+   and any accidental empty retained publish (a common scripting mistake)
+   deletes the entity and its entire subtree, with no confirmation.
+   Combined with issue 2 (anyone can publish), this is a sharp footgun.
+
+### B. Twin data carried in the registration message
+
+5. **Inline twin data creates a second, never-maintained copy.**
+   The twin fragments inside the retained registration message are republished to `twin/<key>`.
+   The fragment now lives in two places.
+   Twin-topic updates (HTTP or MQTT) update only `twin/<key>` —
+   they never rewrite the inline copy in the registration message.
+
+6. **On restart the stale inline copy wins.**
+   `load_from_message_log` deliberately **skips** replaying `twin/` topics
+   ("Twin fragments are intentionally not restored from the entity store log")
+   but **does** replay the registration message,
+   so the only twin value the rebuilt store sees is the stale inline copy, and it overwrites the newer value.
+   This is exactly the family of bugs documented by the `Skip`-ped tests in
+   [`entity_store_persistence.robot`](../../tests/RobotFramework/tests/tedge_agent/entity_store_persistence.robot):
+   * a twin updated via the `twin/` topic is lost on restart;
+   * a twin updated while the agent is offline is lost on restart;
+   * a twin **deleted** via an empty `twin/` payload is **resurrected** on restart by the inline copy.
+
+### C. MQTT vs HTTP behavioural divergence (same operation, different result)
+
+7. **`@`-prefixed keys are rejected in one path and silently absorbed in another.**
+    The twin HTTP API rejects any fragment key starting with `@` (`400`).
+    But the registration payload has no `deny_unknown_fields`
+    and `#[serde(flatten)]`s everything else into `twin_data`,
+    so a `@whatever` key in a registration message is silently accepted
+    *and then published to a `twin/@whatever` topic* — the very thing the twin API forbids.
+    (`EntityUpdateMessage` *does* set `deny_unknown_fields`, so the three payload types validate inconsistently.)
+    The MQTT docs only "discourage" `@` twin keys; the HTTP docs hard-reject them.
+
+8. **Parent-dependency model differs.**
+   HTTP registration is *strict*: an unknown `@parent` returns `400 NoParent`.
+   MQTT registration is *eventual*:
+   a child-before-parent is cached in `pending_entity_store` and silently held
+   (`update()` catches `NoParent` and returns `Ok(vec![])`).
+   The same registration succeeds over MQTT and fails over HTTP.
+
+## Design constraints
+
+Any solution must respect these constraints (mostly for backward compatibility):
+
+1. **The retained registration message on the entity metadata topic is sacred.**
+   The overwhelming majority of users register entities
+   by publishing a retained message to `te/device/<id>//`.
+   This behaviour cannot be broken.
+   It is the one hard constraint.
+
+2. **Inline twin data in the registration message is expendable if there is no other way.**
+   Its only purpose was convenience:
+   register a device with all its initial twin data in one shot,
+   instead of publishing each twin fragment separately afterwards.
+   It is rarely used and is the direct cause of the issue-B family.
+   Breaking or constraining it is acceptable as a last resort,
+   ideally while preserving the one-shot convenience.
+
+3. **Mappers validate eventually, independently, and possibly differently from each other.**
+   A registration that is valid for the agent may be invalid for a mapper
+   (e.g. an external id that conflicts in the cloud),
+   and one that is valid for the c8y mapper may be invalid for the Azure mapper.
+   Mappers may not even be running at registration time —
+   they can come up much later
+   and must be able to fetch and replicate everything registered so far.
+   Therefore **mapper-level validation can never gate local registration**;
+   it happens asynchronously and reports back out-of-band.
+
+4. **The agent must eventually reconcile state from both the device side and the cloud side.**
+   Local registrations/twins originate on the device;
+   some twin values and managed objects originate in the cloud.
+   The agent needs a defined reconciliation with explicit precedence
+   so the two sources converge instead of clobbering each other
+   (a generalisation of the issue-B restart problem to the cloud dimension).
+
+5. **The parent dependency model difference between HTTP and MQTT APIs was an intentional design choice.**
+   Both APIs are designed differently as per their strenths/limitations.
+   The HTTP API can immediately reject a child without a parent with feedback while MQTT can not.
+   There is no desire to align the behaviour on this aspect.
+
+6. A **malicious actor with write access to the broker is explicitly a non-goal.**
+   Such an actor can already delete any device by clearing its topics,
+   impersonate any component, or flood the broker —
+   none of which any registration-API redesign can prevent.
+   This is an **authorization** problem,
+   and authorization is only solvable at the transport layer:
+   broker ACLs and per-client authentication (mosquitto, tedge's default broker, supports both).
+   Because backward compatibility requires that ordinary clients keep publishing to `te/+/+/+/+`,
+   nothing at the API layer can distinguish a malicious authorized writer from a legitimate one.
+   We therefore do not try to; we point operators at broker ACLs and move on.
+
+## Proposal A — Agent owns the entity metadata topic; user publishes treated as requests
+
+The core idea is to separate the **desired-state input** from the **authoritative-state output**,
+and make `tedge-agent` the single writer of the **authoritative state**.
+
+* Clients express *intent* (register / update / delete / set-twin).
+* The agent is the only component that validates intent and publishes the resulting authoritative entity state.
+* Mappers and external observers read only the authoritative state;
+  they never see unvalidated input as truth.
+* Mapper/cloud validation is a separate, eventual layer
+  that reports status on a dedicated channel and feeds back into reconciliation,
+  never blocking local acceptance.
+
+### A.1 Distinguishing intent from validated truth
+
+This is the crux of keeping a single shared topic.
+If the agent re-publishes validated data to the *same* entity metadata topic the user writes to,
+how does a mapper — or any other listener — tell a raw, not-yet-validated user message
+apart from agent-validated truth?
+On a single topic with last-writer-wins retention, it cannot, for two reasons:
+
+* there is a window between a user's publish and the agent's validation
+  in which the retained message is unvalidated input; and
+* after the agent re-publishes, nothing in the payload says "this one is the validated version".
+
+A listener that acts on whatever it reads
+will act on transient, unvalidated, or about-to-be-rolled-back input.
+So yes — distinguishing the two is required not just for mappers but for *every* listener
+trying to determine the real truth.
+
+The minimal-breakage answer is a reserved marker that only the agent ever sets on the output.
+A bad actor can mimic this marker, but we aren't immune to bad actors anyway as already established.
+
+* On a valid registration message, the agent re-publishes the normalized entity **with** the marker
+  like `@source: "tedge-agent"`.
+* On a **rejected** message, the agent
+  (a) publishes a structured error to `te/errors` describing why, and
+  (b) **restores the last-known-good** marked message if one exists,
+  or **clears** the topic if the entity never existed.
+  This directly fixes issues 1 and 2:
+  the broker can no longer be left holding a message the agent rejected.
+* **Listeners act only on messages bearing the agent's marker.**
+  An unmarked message is treated as "pending validation"
+  and ignored until the agent confirms it (or rolls it back).
+
+The same mechanism solves the agent's own echo problem:
+when the agent's re-published message comes back to it
+(the agent is subscribed to the same topics),
+the marker plus a match against its store identifies it as its own echo,
+so it is a no-op and there is no republish loop.
+
+**Legacy Clients**
+
+Third-party components that reads `te/+/+/+/+` today would get notified twice on those topics.
+They should have expected it anyway, as upates via the same topics were allowed earlier as well.
+The clients publishing the registrations wouldn't feel any difference,
+as they are less likely to be subscribed to the same topic.
+They can subscribe to them now, to get notified about the agent accepting/rejecting them.
+
+:::note
+Restoring or clearing a retained message that the agent did not author is a fundamental change.
+It must be documented and comunicated heavily on roll-out.
+:::
+
+### A.2 Inline twin: accept at intake, never retain a second copy
+
+* The agent keeps **accepting** twin fragments inside a registration message
+  (the one-shot convenience is preserved).
+* On intake it **splits** them out:
+  identity is retained on the entity metadata topic;
+  each twin fragment is published to its `twin/<key>` topic.
+* The **normalized registration message the agent republishes contains identity only — no twin.**
+  So there is never a second retained copy of a twin fragment,
+  and `load_from_message_log` can no longer resurrect a stale value.
+* Restart behaviour changes so the `twin/` topics are authoritative for twin values
+  (the registration message no longer carries any), fixing the entire issue-B family.
+
+**Breakage.**
+A client that re-reads its *own* registration retained message
+and expects to find the twin fields it sent will no longer find them there
+(they are on the `twin/` topics instead).
+This is the acceptable last-resort breakage the constraints permit.
+The convenience of one-shot registration-with-twin itself is retained.
+
+### A.3 Enforce immutability and unify validation
+
+* Reject changes to `@id`/`@type` on re-publish (update).
+  The only update that is allowed is the updation of `@parent` and `@health` just like the HTTP API.
+* Apply the same twin-key validation (no leading `@`, no `/`, non-empty)
+  at registration intake as the twin API already applies,
+  so `@`-keys can no longer sneak onto `twin/` topics via registration (issue 10).
+  Unknown `@`-keys in a registration are rejected rather than absorbed.
+* These restrictions can't be considered breaking changes,
+  as we are just tightening previously undefined behaviour.
+
+
+### Summary of behavioural/breaking changes
+
+* Re-publishing of entity metadata messages with `@source` marker leading to duplicate notifications.
+* Stripping twin fields from a "user owned" registration* message.
+* Previously-silent malformed inputs
+  (changing `@id`/`@type`, `@`-prefixed twin keys, unknown `@`-keys)
+  now get rejected with feedback instead of being silently mishandled.
+
+## Proposal B - Agent owns the entity metadata topic; separate channel for registration requests
+
+Not refined further as the primary backward compatibility requirement itself is not met.
+
+## Proposal C - User publishes to entity metadata topic; Agent owns separate entity metedata channel
+
+Not refined further as the duplication of the same data across multiple topic trees isn't ideal.
--- a/design/decisions/0009-entity-registration-api-revamp.md
+++ b/design/decisions/0009-entity-registration-api-revamp.md
@@ -48,20 +48,20 @@ HTTP buys validated *writes*, not a protected *source of truth*.
 
 ## Catalogue of issues
 
-### A. Source-of-truth integrity
+### A. The entity data on the broker is easily corrupted
 
-1. **Rejected registrations leave broken retained messages.**
+* **A1 — Rejected registrations leave broken retained messages.**
    A bad registration message, rejected by the agent, stays with the broker even after the rejection.
    There is no compensating publish.
    The broker keeps serving a message the agent refuses to honour,
    so the broker's view and the agent's view disagree permanently and never reconverge.
 
-2. **A valid retained message can be replaced by an invalid one.**
+* **A2 — A valid retained message can be replaced by an invalid one.**
    Because the entity metadata topic is writable by anyone,
    a second publisher can clobber a good registration with a bad one.
-   The agent rejects the new message but, per issue 1, the bad message stays retained.
+   The agent rejects the new message but, per issue A1, the bad message stays retained.
 
-3. **Registration and update share one channel, with surprising replace semantics.**
+* **A3 — Registration and update share one channel, with surprising replace semantics.**
    An MQTT "update" is a **full struct replace**, not a merge:
    the code builds `EntityMetadata { twin_data: merged, ..entity_metadata }`,
    so `@id`, `@type`, `@parent` and `@health` all come from the new message.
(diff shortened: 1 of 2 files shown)


design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,280 @@
...
+A listener that acts on whatever it reads
+will act on transient, unvalidated, or about-to-be-rolled-back input.
+So yes — distinguishing the two is required not just for mappers but for *every* listener
+trying to determine the real truth.
+
+The minimal-breakage answer is a reserved marker that only the agent ever sets on the output.
+A bad actor can mimic this marker, but we aren't immune to bad actors anyway as already established.
+
+* On a valid registration message, the agent re-publishes the normalized entity **with** the marker
+  like `@source: "tedge-agent"`.
```
rina23q, line 292: From the client's perspective (I'm imaging a custom cloud mapper development), it is a hassle because client always has to do validation on `@source` from the payload to distinguish the marked entity.
If a registered channel is separated from a request channel, it would be easy for clients as it just to subscribe the registered entity topic. However, I saw this idea is rejected in Proposal C.

rina23q, line 292: I have a suggestion: add a `@last-updated` timestamp to validated entity messages

Given that a custom mapper has its own entity store to manage their device IDs. When the mapper restarts and receives a burst of retained messages, it can compare the timestamp against its own cache to decide whether to update or skip.
Without timestamp, it may need to validate all keys and values in the cache if they are up-to-date.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,280 @@
...
+  (changing `@id`/`@type`, `@`-prefixed twin keys, unknown `@`-keys)
+  now get rejected with feedback instead of being silently mishandled.
+
+## Proposal B - Agent owns the entity metadata topic; separate channel for registration requests
+
+Not refined further as the primary backward compatibility requirement itself is not met.
+
+## Proposal C - User publishes to entity metadata topic; Agent owns separate entity metedata channel
+
+Not refined further as the duplication of the same data across multiple topic trees isn't ideal.
```
rina23q: Actually I like the Proposal C. Is it really a problem to have two channels, one for registration and the other for truth of source? Their data might be same, but might be different if the registration message is malformed. I'm pretty sure that third-party services wants to subscribe only the truth of source channel.

rina23q (commented):
Thank you for your accurate analysis and problem breakdown; it is greatly appreciated.

The proposal is debatable - hope as you expected. I left my opinions on the proposal.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: I wouldn't say rejected. The duplication of the same data across 2 topic channels was my concern. We can't change the existing request channels and their "retained" requirement. So, the the clients will continue publishing retained registration messages and if the agent duplicates the same (or a normalised version) on a different channel, the original retained request is now redundant and stays with the broker with no value. That was my concern with proposal C and the only reason why I put this as the preferred one. If we weren't limited by backward compatibility, I also would have picked proposal C.

We can explore that proposal further if the duplication isn't a concern OR, if the agent is allowed to clear the registration message published by the users once the normalised version is published to its own channel. Even with the proposal A, the agent is updating the user's request messages. So, clearing them can also be treated as another mutation. These are all options and nothing is accepted/rejected until we all agree. So, let's discuss, debate and finalise.

albinsuresh: The mappers currently don't maintain a **persistent** entity store. What they have is a transient entity cache and it is rebuilt on every restart from the retained messages on the broker, as that's the source of truth. But yeah, once the agent becomes the owner of those retained entity metadata messages, a `@last-updated` field could be useful to persistent clients. Especially if we make the mappers persistent again, the way they used to be in the past. Earlier they were relying on the equality of the entity metadata between the entity store state and the message received.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: I don't think leaving an already processed request retained on the broker would be acceptable. That's just wasted resources. I'd say proposal C is feasible if the agent is allowed to clear the registration messages published by the clients once the normalised version is retained on its own channel.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,280 @@
...
+regardless of whether `tedge-agent` accepts it.
+The agent validates *after the fact* and has no authority to correct what is already on the broker.
+Almost every issue below is a symptom of collapsing these three roles into one writable, unvalidated namespace.
+
+The HTTP API does not escape this:
+a successful HTTP registration is still published to the broker as a retained message
+so an HTTP-created entity can be overwritten by any MQTT client afterwards.
+HTTP buys validated *writes*, not a protected *source of truth*.
+
+## Catalogue of issues
```
reubenmiller, line 49: I would of thought we also have some issues with processing the order of messages when sent to the cloud.

For instance, with the service registration and status updates can result in the status sometimes being ignored.

design/decisions/0009-entity-registration-api-revamp.md
reubenmiller: > What they have is a transient entity cache and it is rebuilt on every restart from the retained messages on the broker, as that's the source of truth.

We can change this so that the mapper's request the entity store information via the http interface provided by the tedge-agent though?

Bravo555: Once we're at it, also would be nice to provide both `@created-at` and `@last-updated-at` to know if the entity was created previously and updated, or just has been created (when both timestamps are equal)

albinsuresh: Yes. That would make the startup logic of the mapper more straightforward, where it doesn't have to deal with any of the ordering problems between parent-child entities during that initial surge. The only issue here would be the added strong dependency on the agent process, in addition to mosquitto, during mapper startups, as it won't be able to do anything until the agent is also available. TBH, for starup I don't think that's a big deal.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,280 @@
...
+1. **Rejected registrations leave broken retained messages.**
+   A bad registration message, rejected by the agent, stays with the broker even after the rejection.
+   There is no compensating publish.
+   The broker keeps serving a message the agent refuses to honour,
+   so the broker's view and the agent's view disagree permanently and never reconverge.
+
+2. **A valid retained message can be replaced by an invalid one.**
+   Because the entity metadata topic is writable by anyone,
+   a second publisher can clobber a good registration with a bad one.
+   The agent rejects the new message but, per issue 1, the bad message stays retained.
```
Bravo555: question: would the agent overwriting these topics on bad input by itself, without adding `@source: agent`, be sufficient to solve these issues?

Indeed without `@source: agent` mappers wouldn't know if the entity is valid so they could send invalid entities to the cloud, but if these entities were then overwritten by the agent to clear the topic, they could be immediately removed. This is suboptimal to be sure, but the the problem described here is only about leftover broken retained messages, which I think could be solved that way?

In the proposal A, there's a note:

> Restoring or clearing a retained message that the agent did not author is a fundamental change. It must be documented and comunicated heavily on roll-out.

Could this potentially be a breaking change?

Asking just for clarity of which portions of the proposal address which issues and how the change could be decomposed into a number of smaller action items.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,280 @@
...
+   * The MQTT update example in the docs republishes only `@type`+`@parent` (no `@id`),
+     which would wipe the id of an entity originally registered with one.
+   * Twin data is *merged* while entity metadata is *replaced* —
+     two different merge semantics in one message, undocumented.
+
+4. **Empty payload is overloaded as "delete".**
+   An entity can never have an intentionally empty registration,
+   and any accidental empty retained publish (a common scripting mistake)
+   deletes the entity and its entire subtree, with no confirmation.
+   Combined with issue 2 (anyone can publish), this is a sharp footgun.
```
Bravo555: As far as I understand, we can't do anything about this issue for the entity metadata topic that wouldn't be a breaking change, so the solution is to just use HTTP API.
Or, for clients who for some reason can't/won't use HTTP, we _could_ consider providing additional methods for registering/updating entity metadata under some new `te/device/d_id/service/s_id/meta` topic.

Bravo555 (commented):
After thinking on it a bit, I'm convinced by the **Proposal A**. Adding a marker property that only later clients expect and verify is an elegant way to solve the problem with as few breaking changes as possible.
Though I do expect most of the value to come from resolving the inconsistencies between MQTT and HTTP API and updating the tests to cover these missing cases.
If the benefits were ever large enough for clients to update, we could also provide additional mechanisms under a meta topic like `te/device/d_id/service/s_id/meta`, but "providing additional registration/patch methods that eliminate some of the footguns" is probably not big enough of a benefit to justify the complexity of having to support and document both these methods.
For the same reason, I considered an _A+B_ solution, where, instead of removing the registration at entity metadata topic, we could still process them while also allowing registration at a different topic like `te/device/d_id/service/s_id/meta/register`. One advantage of that approach is that new clients could opt into this new topic and their registration wouldn't trigger as duplicate "registration" message for older clients who observe this topic but aren't aware of the `@source` property, but that also probably be a doubtful proposition of bringing in more complexity than it removes.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: > question: would the agent overwriting these topics on bad input by itself, without adding `@source: agent`, be sufficient to solve these issues?
> 
> Indeed without `@source: agent` mappers wouldn't know if the entity is valid so they could send invalid entities to the cloud, but if these entities were then overwritten by the agent to clear the topic, they could be immediately removed. This is suboptimal to be sure, but the the problem described here is only about leftover broken retained messages, which I think could be solved that way?

If we're re-using the same `te/+/+/+/+` topic both as the request channel as well as the agent's authoritative channel, then the `@source` field is critical to avoid the mappers or clients from reacting to an invalid request. Yes, the agent would **eventually** replace the invalid message with a valid one as per proposal 1. But, what if the agent is down when the request was sent? The mapper would have already reacted to the original request and it wouldn't get that "correction message" until the agent comes back up. By making the mappers and other clients react only to messages with the `@source` field, we are preventing such issues.

> In the proposal A, there's a note:
> 
> > Restoring or clearing a retained message that the agent did not author is a fundamental change. It must be documented and comunicated heavily on roll-out.
> 
> Could this potentially be a breaking change?

This is one part that I'm not fully sure about. I don't expect any normal "publishing clients" that registers themselves and pushes data or receiving commands (which is likely the majority of clients) to be affected at all by this change, as they are less likely to subscribe to the `te/+/+/+/+` topics that they are publishing to.

But, if there are other clients like mappers that were tracking the registration messages, they would definitely get notified. But, here also I'm not sure if it can be considered a "breaking behaviour" as those clients should have been prepared to received **entity update messages** on the same channel, as we've supported that (though in a broken manner). So, I'd expect those clients to treat the addition of the `@source` field as yet another update.

> Asking just for clarity of which portions of the proposal address which issues and how the change could be decomposed into a number of smaller action items.

Sure, I'll add a section or a table after each proposal capturing how it addresses each and every listed issue.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: > As far as I understand, we can't do anything about this issue for the entity metadata topic that wouldn't be a breaking change, so the solution is to just use HTTP API.

The fix considered for this, over the MQTT API, was for the agent to restore the original entity registration message if the cleared (deleted) entity had any children that are not deleted yet. So, the clear message would be considered as a delete request only for "leaf" entities. But yeah, this clearly is a breaking change and it would introduce yet another point of divergence from the HTTP API.

> Or, for clients who for some reason can't/won't use HTTP, we could consider providing additional methods for registering/updating entity metadata under some new te/device/d_id/service/s_id/meta topic.

This was proposal B. But, introducing a new channel for registration is a strict no-go as we will have to support registrations/deregistrations over the existing `te/+/+/+/+` topic anyway for backward compatibility.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+  the first data of a freshly registered entity
+  is delayed by the creation round-trip;
+  during a cloud/HTTP outage the hold queues are bounded,
+  so the overflow policy must be defined (and observable, per A.5);
+  and the gate must cover *every* outbound channel for the entity —
+  a single ungated channel reintroduces the race.
+
+This change is internal to the c8y mapper
+and is orthogonal to which proposal is chosen;
+it is listed here to complete the end-to-end story for issue D2.
```
reubenmiller, line 501: I'm having my doubts here about just switching to using the Cumulocity REST API to do device creation. Mostly because we don't have any good buffering mechanism when the device is offline.

reubenmiller: I'm wondering if we could take a slightly different approach to the ordering issue on the cloud side, e.g. what do we need to do to ensure the cloud device is created properly regardless of sending a registration message, telemetry, or digital twin info?

albinsuresh: Had considered that option, where a placeholder entity is created first, followed by updating it when the actual registration message arrives. The main hindrance for this approach is the external id. Since there aren't any MQTT APIs available to update the external ID, that was a dead end. The only other option was HTTP API. But there isn't an atomic PUT API available to update the external ID either. It will be a multi-step process with a POST followed by a DELETE. Since both solutions require HTTP, there isn't really much difference as some buffering solution for offline handling will be required either way.

reubenmiller: In newer Cumulocity versions there should be an atomic creation of a managed object along with an external id, see [Cumulocity API](https://cumulocity.com/api/core/#operation/postManagedObjectCollectionResource). I'm not sure if we could use it there since we still need to support older Cumulocity versions, however it will give us a future simplification.

Below shows an example how to use the new api to create a managed object with a given external id in the same call, where if the external id already exists, then the managed object will not be created.

```sh
$ c8y inventory create --template '{externalIds:[{externalId:"rmi_test_exid_001",type:"c8y_Serial"}]}' --name rmi_test_exid_001
```

Then when trying it create it again, it will return a HTTP status code 409 (conflict/duplicate):

```
$ c8y inventory create --template '{externalIds:[{externalId:"rmi_test_exid_001",type:"c8y_Serial"}]}' --name rmi_test_exid_001

{
  "error": "identity/Duplicate",
  "info": "https://cumulocity.com/api/core/",
  "message": "ID [type=c8y_Serial, value=rmi_test_exid_001] was already bound to different Global ID."
}
```

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+    the same missing-barrier problem as issue D1, one hop further.
+
+### E. Service status at registration
+
+* **E1 — Service registration fabricates an initial status.**
+    A service's registration message declares its *existence* but carries no *liveness* information;
+    the actual status lives on the separate `status/health` channel.
+    The c8y mapper, however, must supply a status in the service creation message
+    (SmartREST `102`), and it hardcodes **`up`** —
+    so every service appears as "up" in the cloud from the moment of registration,
```
rina23q, line 160: Isn't it rather a problem of our c8y-mapper implementation? We can first send `unknown` status for initial creation. Then, update the status either `up` or `down` if it's known. I think it is more reasonable to decouple "registration" and "status update". 

> `status` can be `up`, `down`, `unknown`, or any custom status
https://cumulocity.com/docs/device-integration/fragment-library/#announcing-a-service-to-the-platform

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+## Proposal B - User publishes to entity metadata topic; Agent owns separate entity metadata channel
+
+Not refined further as the duplication of the same data across multiple topic trees isn't ideal.
+
+The duplication could be mitigated by clearing the retained message on the user channel
+once the agent has replicated it to its own channel,
+turning `te/+/+/+/+` into a pure ingest channel.
+But the evaluation still doesn't favour it:
+
+* **No significant ordering advantage over Proposal A.**
```
rina23q, line 630: Agreed — ordering is the same in both proposals, since it comes from MQTT's per-topic ordering, not from the topic design. But then it should not count as a reason to reject Proposal B. Note also that per the A.4 caveats, Proposal A alone even makes the window a bit worse, because the marked republish always comes after the raw publish.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+  The ordering problem is registration-vs-*data*,
+  and data stays on its own topics either way;
+  the confirmation barrier (wait for the agent-channel message)
+  and the consumer-side tolerance
+  are identical in shape to A.4/A.5.
+* Its genuine advantage is a cleaner trust model:
+  a channel that only ever carries validated truth,
+  with no `@source` marker inspection
+  and no window in which a retained message is unvalidated input.
```
rina23q, line 639: Another big advantage: a separate source-of-truth channel can be protected by a broker ACL, only `tedge-agent` can publish to it. In Proposal A this is impossible, because the same topic must stay writable by all clients for registration. Constraint 6 says bad writers should be handled by broker ACLs, but actually only Proposal B makes that possible for the truth topic.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+* The clearing mitigation, however, breaks the most relied-upon *consumer* behaviour:
+  entity discovery via retained replay on `te/+/+/+/+`
+  (explicitly advertised in the docs as the entity-store view)
+  would come up empty —
+  violating the spirit of constraint 1 on the read side.
+  Without clearing, the duplication drawback stands.
```
rina23q, line 645: I agree clearing is not acceptable. But then I don't understand why the remaining duplication is a serious drawback. If the old topic is treated as input only, it does not matter whether a message stays there or not:

* Migrated consumers read only the new truth channel, so the leftover copy on the old topic does not affect them.
* Legacy consumers get exactly today's behaviour: same view, same known problems. No regression.
* The retained input even keeps one useful property: a registration published while tedge-agent is down is still delivered when the agent comes back.

So the real cost of "B without clearing" is documentation ("publish here, read truth there"), not a technical problem. I think this bullet should say concretely what harm the duplication causes. Probably the storage cost, but is it really that bad? It is only one extra retained copy per entity (Proposal A also publishes each registration twice anyway).

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+* Every consumer must migrate to the new channel to benefit,
+  whereas Proposal A upgrades the semantics of the topics they already watch.
```
rina23q, line 647: Consumers must migrate in Proposal A too: every consumer has to parse each payload and check the `@source` marker just to know if the message is trustworthy. With a separate channel, the topic itself is the filter — the broker does the work, no payload inspection needed. So the migration cost is the same or higher in A.

design/decisions/0009-entity-registration-api-revamp.md
rina23q: Note: What I said **Proposal C** (new output channel: registration stays on `te/+/+/+/+` exactly as today.) is renamed to **Proposal B** in this commit. 7eb4597ab462960ec348488090ca4c729f707fc3

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+  are no longer cached and replayed — they are dropped,
+  like all other pre-registration measurements already are;
+  the remaining event buffer reports its drops instead of staying silent (A.5).
+* The c8y mapper delays the first data of a newly registered entity
+  until its cloud creation is confirmed (A.6).
+* Newly registered services appear in the cloud with `unknown` status
+  until their first (inlined or published) health status,
+  instead of a fabricated `up` (A.7).
+
+## Proposal B - User publishes to entity metadata topic; Agent owns separate entity metadata channel
```
rina23q, line 621: I prefer this proposal (the former Proposal C), and I'd like to propose a refined version:

1. **Registration stays on the existing `te/+/+/+/+`, unchanged.** Retained is not required for requests, but still accepted (and useful: a registration published while `tedge-agent` is down is delivered when it comes back). The retained state of this topic just carries no authority anymore — it is input, not truth.
2. **tedge-agent subscribes, validates, and on success publishes the normalized entity retained to a result channel** (rejections go to `te/errors`). The agent's entity store stays the real source of truth; the result channel is its projection for MQTT consumers.
3. **The result channel is ACL-protected: only `tedge-agent` may publish, everyone else subscribes.** No client can overwrite validated state — enforced by the broker, not by convention.

With this, the subscription itself becomes the trust filter — no payload inspection, no marker, no restore/clear machinery, and it works even while the agent is down.

To be clear, this is **not** the earlier idea of a new channel *for registration* (former Proposal B): registration stays where it is; only the validated **output** gets the new channel. So constraint 1 holds fully, and legacy consumers of `te/+/+/+/+` see today's behaviour unchanged.

rina23q (commented):
I'm definitely in favor of Proposal B (former Proposal C) — my comments argue it deserves as deep an exploration as Proposal A, and I've sketched a refined version in the Proposal B section. Take this as a friendly counter-proposal: everything else in the document (issue catalogue, A.2–A.7) has my full support.

design/decisions/0009-entity-registration-api-revamp.md
reubenmiller: There is another order of messages issue, but more related to the difference to the Cumulocity MQTT and HTTP , where a child device software list can easily fail if the REST API updating the software list is sent before the child device has been registered via MQTT. This race condition has been seen within the "Software list should be populated during startup" test on the "software-update-child.robot" suite.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: That's a good addition on the C8Y side, which we can use to create the device with the initial external id when an entity is registered.

But just to be clear, the problem that I was highlighting in my previous commit wasn't about providing external id during creation. I was talking about the case where an entity registration message is sent with an explicit `@id` to be used as the external id followed by data messages, but the mapper receives its data messages first. If we go with the approach of creating a provisional entity first for that data message, only to correct its metadata when the actual registration message arrives, we'll be forced to **generate** an external id for that provisional entity first (derived from the topic scheme) and then later change it to the one provided in the registration message. This can't be done over MQTT but only using REST APIs (multiple API calls).

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: Yup. That's the default approach proposed in A.7. The rest of A.7 is about other enhancements like support for initial status in the registration message itself, if some users don't want to see the `unknown` to `up` transition at all.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: > Legacy consumers get exactly today's behaviour: same view, same known problems. No regression.

Agree. That's the true backward compatibility where existing users are fully unaffected, though they'd continue suffering from today's limitations.

> The retained input even keeps one useful property: a registration published while tedge-agent is down is still delivered when the agent comes back.

We don't necessarily need retained for that. Just the persistent session of the agent would be sufficient.

> I think this bullet should say concretely what harm the duplication causes. Probably the storage cost, but is it really that bad? It is only one extra retained copy per entity (Proposal A also publishes each registration twice anyway).

Having two copies itself is the disadvantage IMO. Yes, it'd be just a few 100 messages or a few 1000s in larger fleets. The storage impact might be minimal as these are small JSON records. But still, it's unnecessary wastage of resources, which is why I was insisting on the cleanup. But that comes at the cost of breaking backward compatibility as well. So, that duplication might be the cost that we must pay to maintain backward compatibility.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: Agree. Though the topic doesn't change, that alone doesn't provide backward compatibility to the user as they'll still have to update their logic to accommodate the `@source` marker.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: Yes. The former Proposal B with a new registration topic was moved into the "Ideal Solution" section earlier in the doc, which triggered this rename.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: You've raised some really convincing arguments to make me consider Proposal B as the preferred solution.

The argument that `Proposal A` is more backward compatible than `Proposal B` doesn't hold true for the reasons you highlighted: re-using the same topic alone doesn't guarantee backward compatibility in this case, as the new contract also forces existing clients to update their logic handling updated messages published back to that topic. I made a wrong assumption that most users would only be publishing to those topics and not subscribed to the same. But I myself have contradicted on that point where publishing back to the same topic was also done to keep existing subscribers happy.

Maintaining a separate tedge-agent owned retained topic definitely feels like the cleaner approach now, with zero impact to existing clients that might already be subscribed to the `te/+/+/+/+` topics. So, the only 2 outstanding points are:
**1. Duplication of data: do we care?** I'd say yes, we care, but I guess this is the cost we must pay for backward compatibility. There's no other way around it.
**2. Can we clear the duplicate data on the request channel?** Unfortunately no, as it'd affect all existing clients subscribed to those topic and might interpret that clear message as an entity deletion request.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+regardless of whether `tedge-agent` accepts it.
+The agent validates *after the fact* and has no authority to correct what is already on the broker.
+Almost every issue below is a symptom of collapsing these three roles into one writable, unvalidated namespace.
+
+The HTTP API does not escape this:
+a successful HTTP registration is still published to the broker as a retained message
+so an HTTP-created entity can be overwritten by any MQTT client afterwards.
+HTTP buys validated *writes*, not a protected *source of truth*.
+
+## Catalogue of issues
```
didier-wenzek, line 49: Not sure that the following is listed some where:

- `name` and `type` are published on two independent topics as *entity registration* or *twin data*

design/decisions/0009-entity-registration-api-revamp.md
rina23q: > We don't necessarily need retained for that. Just the persistent session of the agent would be sufficient.

True. And that's actually nice. It means the duplication is not permanent in principle. Once backward compatibility is dropped in some future major version, requests can become non-retained (and the leftover retained copies safely cleaned up).

> So, that duplication might be the cost that we must pay to maintain backward compatibility.

Agreed, and the constraint comes from today's design.

design/decisions/0009-entity-registration-api-revamp.md
rina23q: Thanks for reconsidering, much appreciated!

Point 1: As discussed in [this thread](https://github.com/thin-edge/thin-edge.io/pull/4233#discussion_r3575214368), the duplication is not permanent. It goes away once the old contract is eventually dropped.

Point 2: Good point. An empty retained message on the input channel is a delete request in today's API, so we cannot use it just to remove the duplicated data. Legacy subscribers could even propagate it as an entity deletion. It's worth capturing in the doc when revising.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+  provide this exchange natively.)
+* **Authoritative state is retained on agent-owned entity metadata topics.**
+  Only the agent writes to the `te/device/<entity>/<topic>/<id>` topic.
+  Consumers never see unvalidated retained input,
+  so no marker is needed to tell input apart from truth.
+* **The broker is not the queryable source of truth for consumers.**
+  Late-starting consumers (e.g. mappers) bootstrap from the agent's HTTP query API —
+  a consistent snapshot — and then follow the live topics.
+  Retained messages remain a notification/cache layer, not the store,
+  which removes the retained-replay ordering problem for late joiners entirely.
```
didier-wenzek, line 248: I don't see how a mix of protocol (HTTP + MQTT) can remove ordering problem entirely.

Without a *version* attached to registration messages and snapshots, how  a late joiner can detect that a registration message has to be applied or not to its current snapshot?

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+* **Data messages are order-tolerant by specification, not by per-consumer improvisation.**
+  Data received for an unknown entity creates a *provisional* entity,
+  upgraded by merge when its authoritative registration arrives
+  (a generalisation of today's auto-registration).
+  With this mandated by the spec, out-of-order data stops being every consumer's private problem.
```
didier-wenzek, line 253: Sorry, but to me, that sounds like gibberish.

design/decisions/0009-entity-registration-api-revamp.md
```diff
@@ -0,0 +1,650 @@
...
+  and `load_from_message_log` can no longer resurrect a stale value.
+* Restart behaviour changes so the `twin/` topics are authoritative for twin values
+  (the registration message no longer carries any), fixing both category-B issues.
+
+**Breakage.**
+A client that re-reads its *own* registration retained message
+and expects to find the twin fields it sent will no longer find them there
+(they are on the `twin/` topics instead).
+This is the acceptable last-resort breakage the constraints permit.
+The convenience of one-shot registration-with-twin itself is retained.
```
didier-wenzek, line 340: Is this *one-shot registration-with-twin* really convenient?

As a user I would prefer an API that let me see what I `SET` with a subsequent  `GET`, over an API that let me save a request by conflating two concepts.

didier-wenzek (commented):
These specs are difficult to grasp. Not only because they are lengthy, but mainly because they are focused on mechanisms without a clear data model nor user points of view. What is a device registration? Why a registration can be rejected by the agent? Shouldn't a device be in a better position than the agent to provide its registration data? What is auto-registration? How auto-registration interacts with registration? Why `@id` cannot be changed, while `@parent` can? What is the relationship between registration metadata and twin data?  

- If I understand correctly, the main difference between proposal A and B is where the sanitized registration messages are published to. On the same topic with a source field added by the agent for A, on a second topic owned by the agent in the case of proposal B.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: Not captured separately as `name` and `type` are just `twin` data and these 2 were the primary ones for which we introduced the "inline twin data in registration message" for, so that the name of the device is set properly right from the start. I agree that it's worth capturing that point separately.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: Yeah, I get that from an external reader's pov. The sophisticated sounding gibberish has piled up over time, after multiple back-and-forth conversations with the AI, to a point where I had to explicitly prompt it to "simplify the language" right before this final commit. A few like this still slipped through. I'll do another pass.

design/decisions/0009-entity-registration-api-revamp.md
albinsuresh: The `name` and `type` fragments were the primary drivers behind this requirement, where this would let the C8Y mapper create the cloud entity with the "right name and type" right from the start rather than using provisional values first, which gets corrected later. But, since we have considered the notion of provisional entities as a solution for ordering problems, may be it isn't such a big deal to "correct things eventually".

albinsuresh:
> These specs are difficult to grasp. Not only because they are lengthy, but mainly because they are focused on mechanisms without a clear data model nor user points of view. What is a device registration? Why a registration can be rejected by the agent? Shouldn't a device be in a better position than the agent to provide its registration data? What is auto-registration? How auto-registration interacts with registration? Why `@id` cannot be changed, while `@parent` can? What is the relationship between registration metadata and twin data?

Fair criticism. I didn't include a detailed section on the data model because neither proposals were fundamentally changing the existing data model concepts like registration, auto registration, twin metadata etc. The focus was on reshaping the API to addressing the catalogue of issues that have been identified, without fundamentally changing the core concepts maintaining backward compatibility.

However, you've highlighted a major gap: though concepts like registration and auto-registration were documented in the original specs, several questions that you've raised above like the "difference between registration metadata vs twin data" or the "relevance of @id in the entity store" aren't clearly documented anywhere. Over time, these concepts got defined purely through an 'API-driven' lens, which has muddled our core data model. So, yes, I agree that we should address those gaps at least in this round.

> * If I understand correctly, the main difference between proposal A and B is where the sanitized registration messages are published to. On the same topic with a source field added by the agent for A, on a second topic owned by the agent in the case of proposal B.

Yes.

PR #4234, opened 2026-06-30
fix: avoid reading external plugin stdout to memory

Bravo555:
## Proposed changes

Similar to 5fda661, stream result of external config plugin instead of loading everything to memory. Not as impactful as log plugin, because usually people won't be having config files bigger than available RAM, but nevertheless behaviour was changed to keep consistency between the two plugins.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/pull/4222#pullrequestreview-4561595928

## Checklist

### DIFF
--- a/crates/core/tedge_api/src/workflow/log/logged_command.rs
+++ b/crates/core/tedge_api/src/workflow/log/logged_command.rs
@@ -1,5 +1,6 @@
 use crate::CommandLog;
 use std::ffi::OsStr;
+use std::io::Cursor;
 use std::os::unix::process::ExitStatusExt;
 use std::path::Path;
 use std::process::ExitStatus;
@@ -8,6 +9,8 @@ use std::process::Stdio;
 use std::time::Duration;
 use tedge_utils::signals::terminate_process;
 use tedge_utils::signals::Signal;
+use tokio::io::AsyncBufRead;
+use tokio::io::AsyncBufReadExt as _;
 use tokio::io::AsyncWrite;
 use tokio::io::AsyncWriteExt;
 use tokio::process::Child;
@@ -52,7 +55,7 @@ impl LoggingChild {
         let outcome = self.inner_child.wait_with_output().await;
         if let Some(command_log) = command_log {
             command_log
-                .log_command_and_output(&self.command_line, &outcome)
+                .log_command_and_output(&self.command_line, outcome.as_ref())
                 .await;
         }
         outcome
@@ -72,7 +75,7 @@ impl LoggingChild {
             }
         };
         command_log
-            .log_command_and_output(&command_line, &outcome)
+            .log_command_and_output(&command_line, outcome.as_ref())
             .await;
         outcome
     }
@@ -219,7 +222,7 @@ impl LoggedCommand {
         let outcome = self.command.output().await;
         if let Some(command_log) = command_log {
             command_log
-                .log_command_and_output(&self.to_string(), &outcome)
+                .log_command_and_output(&self.to_string(), outcome.as_ref())
                 .await;
         }
         outcome
@@ -251,7 +254,7 @@ impl LoggedCommand {
 
     pub async fn log_outcome(
         command_line: &str,
-        result: &Result<Output, std::io::Error>,
+        result: Result<impl Into<CommandOutput<'_>>, &std::io::Error>,
         logger: &mut (impl AsyncWrite + Unpin),
     ) -> Result<(), std::io::Error> {
         if !command_line.is_empty() {
@@ -260,8 +263,10 @@ impl LoggedCommand {
                 .await?;
         }
 
-        match result.as_ref() {
+        match result {
             Ok(output) => {
+                let output: CommandOutput<'_> = output.into();
+
                 if let Some(code) = &output.status.code() {
                     let exit_code_msg = if *code == 0 { "OK" } else { "ERROR" };
                     logger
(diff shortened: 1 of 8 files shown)


crates/extensions/tedge_config_manager/src/plugin.rs
```diff
@@ -110,13 +109,74 @@ impl ExternalPlugin {
         let mut command = self.command(GET)?;
         command.arg(config_type);
 
-        let output = self.execute(command, command_log).await?;
-        let mut file = File::create(target_file_path).map_err(|err| {
+        let stderr_file =
+            tempfile::NamedTempFile::new_in(self.tmp_dir.as_std_path()).map_err(|err| {
+                self.plugin_error(format!(
+                    "Failed to create temporary file in {}: {err}",
+                    tempfile::env::temp_dir().to_string_lossy()
```
albinsuresh:
```suggestion
                    self.tmp_dir.as_str()
```
I see that it was inherited from `crates/extensions/tedge_log_manager/src/plugin.rs` in 5fda6611e4d6a7499db1585cf5d7b5527dc29310.

crates/extensions/tedge_config_manager/src/plugin.rs
```diff
@@ -110,13 +109,74 @@ impl ExternalPlugin {
...
+                    stderr_file.path().to_string_lossy()
+                ))
+            })?;
+
+        command.stdout(target_stdout);
+        command.stderr(stderr);
+
+        let child = command.spawn().map_err(|err| self.plugin_error(err))?;
+        let output = child
+            .wait_with_output(command_log)
```
albinsuresh: Wondering why you took this "spawn child wait for output" approach here as opposed to the `command.status()` approach taken in `crates/extensions/tedge_log_manager/src/plugin.rs`. The `command.status()` approach felt more straightforward, especially about the intent of not caring about the `stdout` or `stderr`. I understand that the `output.stdout/stderr` would be `None` in this case as well, as they are redirected to the files. But, its' not as direct as the other.

Since you only wrote the simpler status based version for the `log_manager/plugin.rs`, there must have been a specific reason why you took a different approach here. Just trying to understand what that is.

albinsuresh (approved):
LGTM. Just a few minor remarks/queries.

crates/extensions/tedge_config_manager/src/plugin.rs
Bravo555: fixed in b4c1681b

crates/extensions/tedge_config_manager/src/plugin.rs
Bravo555: Okay, the problem is that log plugin's get operation isn't logged, but config plugin's get operation is logged.
To verify that config plugin's get operation keeps being logged, I added a check to the config plugin's test in 782d86b6.
The `spawn + wait_with_output` method here was chosen to pass the `command_log` to log stdout to the operation log, however this just doesn't work because stdout is routed to file, not pipe.
So to avoid reading all of stdout to memory and also log stdout to the operation log, we'll have to do something like this:
1. create tempfile for stdout
1. route process stdout to tempfile
1. call `.status()` to wait for process to exit
1. open both stdout and stderr tempfiles and copy their contents to operation logfile
1. persist operation logfile
1. then open and read stdout file again for upload

We write the file, then read it twice, keeping all its contents in a logfile, but we don't load all stdout into memory, so it's okay. I will implement these changes.

rina23q (approved):
LGTM

crates/core/tedge_api/src/workflow/log/logged_command.rs
```diff
@@ -302,3 +310,20 @@ impl LoggedCommand {
         Ok(())
     }
 }
+
+/// A command output that allows reading from it, no matter if output was captured to memory or attached to a file.
+pub struct CommandOutput<'a> {
```
albinsuresh, line 317: That's a nice abstraction.

crates/core/tedge_api/src/workflow/log/logged_command.rs
```diff
@@ -273,22 +278,25 @@ impl LoggedCommand {
...
                     logger.write_all(b"stderr (EMPTY)\n\n").await?;
                 }
 
-                if !output.stdout.is_empty() {
-                    logger.write_all(b"stdout <<EOF\n").await?;
-                    logger.write_all(&output.stdout).await?;
-                    logger.write_all(b"EOF\n").await?;
+                let mut stdout = output.stdout;
+                if !stdout.fill_buf().await?.is_empty() {
+                    logger.write_all(b"stderr <<EOF\n").await?;
```
albinsuresh:
```suggestion
                    logger.write_all(b"stdout <<EOF\n").await?;
```

crates/core/tedge_api/src/workflow/log/logged_command.rs
```diff
@@ -273,22 +278,25 @@ impl LoggedCommand {
...
+                    tokio::io::copy_buf(&mut stdout, logger).await?;
+                    logger.write_all(b"EOF\n\n").await?;
                 } else {
-                    logger.write_all(b"stdout (EMPTY)\n").await?;
+                    logger.write_all(b"stderr (EMPTY)\n\n").await?;
```
albinsuresh:
```suggestion
                    logger.write_all(b"stdout (EMPTY)\n\n").await?;
```

albinsuresh (approved):
Approving again with some minor suggestions.

crates/core/tedge_api/src/workflow/log/logged_command.rs
rina23q: I think both this and below comments are referring to old commits. These don't exist now and also when I reviewed it.

PR #4237, opened 2026-07-02
tests: Make `Tests.Cumulocity.Software Management.Software` suite faster

Bravo555:
## Proposed changes

`Tests.Cumulocity.Software Management.Software` suite is our slowest test suite, running at over 10 minutes on the test runner, I thought to try to speed it up a bit to get some sense what kind of changes would be necessary in other suites to speed up the entire test suite. Currently it runs in about 30 minutes, I think it would be neat if it ran closer to 5-10 minutes.

This PR speeds up software test suite by:

- rewriting some tests that spawn new containers to not spawn + restore previous state
- removing some `Sleep` keywords that were useless because keywords after retry anyway

On my pc
```
before: 00:03:49.143
after: 00:01:33.541
diff: 00:02:16 (-59%)
```

on test runner
```
before: 00:10:41.672
(projected) after: 00:06:18
```

We're running the suites in parallel, so it will have very limited impact on total runtime, but after doing it I have a better idea of what type of changes would be required to speed up other tests. Apart from repeating these changes for other suites, one other speedup would be to try to replace retry mechanisms that always wait `2s` with blocking mechanisms that return as soon as assertion succeeds. In some tests there are multiple of these kinds of checks, for example in test case `Operation gets updated regardless of the order of the creation time` we hit a `2s` retry 4 times, so `8s` in a test that takes `14s` (on my machine), little over 50%, is spent waiting, so I expect it to also inflate runtime of other suites a bit.

Though speeding it up does take a lot of time, so next time I'll try to task other suites to an AI model.

## Types of changes

## Paste Link to the issue
<br/>

## Checklist

### DIFF
--- a/tests/RobotFramework/tests/cumulocity/software_management/software.robot
+++ b/tests/RobotFramework/tests/cumulocity/software_management/software.robot
@@ -4,15 +4,27 @@ Library             Collections
 Library             Cumulocity
 Library             ThinEdgeIO
 
-Suite Setup          Custom Setup
-Suite Teardown       Custom Teardown
+Suite Setup         Custom Setup
+Suite Teardown      Custom Teardown
+Test Setup          Set Managed Object    ${DEVICE_SN}
 
 Test Tags           theme:c8y    theme:software    theme:plugins
 
 
+*** Variables ***
+${DEVICE_SN}    ${EMPTY}
+
+
 *** Test Cases ***
 Supported software types should not be declared during startup when disabled
-    [Setup]    Custom Setup With Software Types Disabled
+    # explicitly reset c8y_SupportedSoftwareTypes if it has been already set on tedge startup
+    Execute Command    tedge mqtt pub c8y/inventory/managedObjects/update/${DEVICE_SN} '{"c8y_SupportedSoftwareTypes":null}'
+    Managed Object Should Not Have Fragments    c8y_SupportedSoftwareTypes
+
+    # check that the mapper doesn't send c8y_SupportedSoftwareTypes if it's disabled
+    Set Mapper Config And Restart    tedge config set c8y.software_management.with_types false
+    Restart Service    tedge-agent
+
     Should Have MQTT Messages
     ...    topic=te/device/main///cmd/software_list
     ...    minimum=1
@@ -25,10 +37,26 @@ Supported software types should not be declared during startup when disabled
     ...    message_contains="types":["apt","flow"]
     Managed Object Should Not Have Fragments    c8y_SupportedSoftwareTypes
 
+    [Teardown]    Run Keywords
+    ...    Set Mapper Config And Restart    tedge config unset c8y.software_management.with_types
+    ...    AND    Restart Service    tedge-agent
+
 Supported software types and c8y_SupportedSoftwareTypes should be declared during startup
     [Documentation]    c8y_SupportedSoftwareTypes should be created if the relevant config is set to true #2654
-    [Setup]    Custom Setup With Software Types Enabled
+
+    # explicitly reset c8y_SupportedSoftwareTypes if it has been already set on tedge startup
+    Execute Command    tedge mqtt pub c8y/inventory/managedObjects/update/${DEVICE_SN} '{"c8y_SupportedSoftwareTypes":null}'
+    Managed Object Should Not Have Fragments    c8y_SupportedSoftwareTypes
+
+    # check that the mapper sends c8y_SupportedSoftwareTypes if it's enabled
+    Set Mapper Config And Restart    tedge config set c8y.software_management.with_types true
+    Restart Service    tedge-agent
+
     Device Should Have Fragment Values    c8y_SupportedSoftwareTypes\=["apt", "flow"]
+
+    [Teardown]    Run Keywords
+    ...    Set Mapper Config And Restart    tedge config unset c8y.software_management.with_types
+    ...    AND    Restart Service    tedge-agent
 
 Software list should be populated during startup
     [Documentation]    The list is sent via HTTP by default.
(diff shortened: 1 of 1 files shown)


tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -4,14 +4,22 @@ Library             Collections
...
 
 
+*** Variables ***
+${DEVICE_SN}    ${EMPTY}
+
+
 *** Test Cases ***
 Supported software types should not be declared during startup when disabled
+    # This Setup could be removed and test could run on a suite device, making it faster
+    # if we either actively sent an empty software type fragment on startup (currently we just don't send anything)
```
rina23q: Having the `c8y_SupportedSoftwareTypes` fragment in ManagedObject triggers filtering out "Installed Software" list in the "Software" tab. To avoid an unexpected surprise, `c8y.software_management.with_types` is set to `false` by default.

I tested how to reset the fragment in c8y. We can send either JSON over MQTT or SmartREST.
- `tedge mqtt pub c8y/inventory/managedObjects/update/${DEVICE_SN} '{"c8y_SupportedSoftwareTypes":null}'`
- `tedge mqtt pub c8y/s/us 143`
After the MQTT request is sent, if we ensure the fragment does not exist in MO, this is the clean enough state for this test.

In short, yes, we can remove this `Setup` by replacing it as I wrote above.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -275,3 +305,14 @@ Validate operation log not uploaded
     ...    minimum=0
     ...    maximum=0
     ...    type=software_update_op_log
+
+Set Managed Object
+    [Arguments]    ${device_sn}
+    Cumulocity.Set Managed Object    ${device_sn}
+    ThinEdgeIO.Set Device Context    ${device_sn}
```
rina23q, line 323: (trivial) I think we can promote this to library level 🤔

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -205,22 +227,30 @@ Supports disabling the Cumulocity c8y_SoftwareUpdate Command
     Execute Command    tedge mqtt pub c8y/s/us '505,${operation.to_json()["id"]},Cancelled operation'
     Operation Should Be FAILED    ${operation}
 
+    [Teardown]    Set Mapper Config And Restart    tedge config unset c8y.enable.software_update
+
 
 *** Keywords ***
+Custom Suite Setup
+    ${device_sn}=    Custom Setup With Software Types Enabled
```
rina23q: (note) This change makes most of tests run with `c8y.software_management.with_types = true`, which is against default. Previously, they were tested with the default (`false`) value. Eventually, this value doesn't affect tests unless it uses advanced software management.

rina23q (commented):
Nice experiment.

I think one problem of this test file is mixed of 3 level things:
1. testing against c8y with legacy software management
2. testing against c8y with advanced software management
3. testing locally, without using c8y

I agree that the previous setup, creating a test container per each test is time consuming and too much costs. 
However, I sometimes think we put too many tests in one robot file. For this case, if we split the file into three (or at least two, covering point 1&2-c8y and point 3-local) files, maybe the tests would look short and easy to read.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
```diff
@@ -239,11 +269,11 @@ Publish and Verify Local Command
     ...    maximum=1
     ...    message_contains="status":"${expected_status}"
 
-    Sleep    5s    reason=Given mapper a chance to react, if it does not react with 5 seconds it never will
     ${retained_message}=    Execute Command
     ...    tedge mqtt sub --no-topic '${topic}' --duration 1
     ...    ignore_exit_code=${True}
     ...    strip=${True}
+    ...    timeout=5
```
reubenmiller, line 287: I think this change is problematic as you might get the message before it has been cleared by the mapper, so the static sleep was in-place to ensure the mapper is not clearing the message. Removing the timeout means that the `tedge mqtt sub` might see the previously manually posted message before the tedge-mapper reacts to it.

Bravo555: restored in 28595acfdd

Ah, so the sleep here is to make sure the mapper doesn't clear the topic because it's supposed to be cleared by the caller?
Then it makes sense why the `Sleep` is required, probably the only thing we can reasonably do to test that things **don't** happen.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
Bravo555: After looking at the usages of these keywords, there aren't many where we use both like here (only other suite where we do is `tedge_c8y_upload`), so I think it can stay local for now, but if we see this pattern more often in other suites, then I'd agree.

tests/RobotFramework/tests/cumulocity/software_management/software.robot
Bravo555: thanks for checking, implemented in 28595acfdd

tests/RobotFramework/tests/cumulocity/software_management/software.robot
Bravo555: cleaned and removed in d6f87b6c2c

After changing these test cases to not rely on startup and properly set and cleanup related configuration settings, decided that we don't have to set any option in startup and we should just use the default, so I removed these setup variants

rina23q (approved):
Good improvement, thank you.

Issue #4229, opened 2026-06-24
Allow custom mapper configs to be edited via CLI


jarhodes314:
# Background
In thin-edge.io v2.0.0, we added support for custom mappers, and a very basic shell of a `tedge mapper` CLI. This included `tedge mapper list` to see what mappers are defined, and `tedge mapper config get` to read config keys for a custom mapper.

At the moment, `tedge config` supports much more than just reading configuration: it has a `set` command for writing values, `add`/`remove` for per-item updates of list types, listing config keys and their documentation. Ultimately, there should be an equivalent for the mapper config. This could take the form of `tedge mapper config {action} {name} {key}` (like we have at the moment for `mapper config get`), or something like `tedge config {action} mappers.{name}.{key}` (e.g. `tedge config set mappers.tb.url example.com`)

Any solution we come up with should be compatible with the thin-edge v2 config and the existing CLI structure as that is familiar to users. This means a root tedge.toml with non-mapper configs and a tedge.toml for each configured mapper (plus the ability to upgrade the config from tedge1 to tedge2, but that is a relatively simple process that needn't depend on the tedge_config APIs). Ideally, the change should result in no outward-facing differences in what the CLI looks like to use, even once we've simplified the main tedge config implementation.

## Current implementation notes
The existing `tedge config` CLI is largely powered by a bunch of custom logic in the `define_tedge_config!` macro, allowing us to generate structs from the basic config shape alongside the tools needed to manipulate this configuration. This works well from a user perspective, but does limit us when it comes to adding new features -- at the moment I am the only team member comfortable making changes to the macro implementation. As a result, I have thus far avoided reusing the macro in the custom mappers configuration

# Proposed solution
[facet](https://facet.rs/) is a new crate designed to solve problems very much like this. We have an established schema for our config file, and we want to manipulate the data within that schema programmatically. This is a fairly significant subset of what the macro currently does, and certainly accounts for a fair amount of complexity.

I believe based on some PoC implementation I have been working on, facet can support at the very least most of what the macro currently does. I would still suggest we have some sort of macro implementation so we can keep the elegant nested struct syntax, e.g.

```rust
define_tedge_config! {
    c8y: {
         proxy: {
              bind: {
                  port: u16,
              }
         }
    }
}
```

instead of 

```rust
struct Config {
   c8y: ConfigC8y
}
struct ConfigC8y {
    proxy: ConfigC8yProxy,
}
struct ConfigC8yProxy {
[...]

PR #4238, opened 2026-07-03
feat: Add prototype for facet-powered tedge config

jarhodes314:
## Proposed changes
(Not intended to be merged, just for showing the shape of the proposed tedge/mapper config code ahead of implementing it bit-by-bit)

An initial prototype for #4229. This isn't changes to the thin-edge code itself, just a comprehensive prototype demonstrating the features a config refactoring would need to support in order to (a) achieve feature parity with the existing `tedge config` interface and (b) allow that to be extended to cover mappers.

The design is based around [facet](https://facet.rs), a toolkit for manipulating and introspecting on Rust types via reflection. The goal of using this is to push logic currently inside the `define_tedge_config` macro out of the macro, making it easier to maintain and (hopefully) improving compilation times.

If you `cd` into `prototypes/facet-config`, you should find that there is a roughly equivalent binary to `tedge config` (for a few keys):

```shell
cargo run -- get device.key_path
cargo run -- set device.key_path /a/new/path
cargo run -- unset device.key_path
cargo run -- add c8y.smartrest_templates new_template
cargo run -- remove c8y.smartrest_templates new_template
cargo run -- list
```

It doesn't support all keys, though this is purely because the code to define them doesn't exist rather than an inherent limitation. Some of the formatting of the output is changed somewhat from `tedge config`, notably the list output is less nicely formatted and doesn't (at the time of writing) support the filter argument to search the keys.

The new features it enables are configuring custom mappers and cloud-type dispatch. For custom mappers, first ensure you have a custom mapper defined by creating the mapper dir (e.g. `/etc/tedge/mappers/tb`), then you can interact with the config:

```
cargo run -- get mappers.tb.url
cargo run -- set mappers.tb.url example.com
```

As for the cloud-type dispatch feature, if you run the `list` command, you will observe that `c8y` has a different config schema with more options than the other mappers. You can change the cloud_type field in those mapper configs to change which schema will be used (for the PoC, it only has a c8y schema and a custom mapper schema). This probably needs protections in a real implementation, as it has the potential to remove a lot of config data, but the feature to change the schema dynamically exists. This means we can use this feature as a basis for deprecating profiles in favour of users just creating extra built-in clouds.

Reading prototypes/facet-config/design/facet-usage-guide.md will give you more of an idea of the technical shape (and will repeat some of this information here).

## Types of changes

## Paste Link to the issue
- #4229

## Checklist

## Further comments
Facet is a fairly new crate, but it is incredibly actively maintained by @fasterthanlime. It is _incredibly_ comprehensive, and at least on recent versions, it has been quite easy to solve the config manipulation problems we need to solve here.


prototypes/facet-config/STATUS.md
```diff
@@ -0,0 +1,44 @@
...
+
+### ~~6. TOML nested group serialization~~ DONE
+
+Explicit tests confirm `skip_serializing_if = "Option::is_none"` on intermediate groups works correctly: setting `c8y.proxy.bind.port` only serializes `[c8y.proxy.bind]`, not `[c8y.proxy.client]` or `[mqtt]`. Round-trip preserves values and doesn't create spurious groups.
+
+## Out of scope
+
+- **Multi-file config** — we want this, but it's a separate investigation not tied to this PoC. Detecting cloud key access and rerouting to the mapper config implementation should be straightforward. For cross-config defaults (e.g. mapper's `c8y.device.cert_path` falling back to root tedge.toml's `device.cert_path`), a `DefaultSpec::FromParentKey` variant would resolve against an optional parent DTO passed at reader construction time. Validation at init ensures `FromParentKey` entries require a parent to be provided. The safety guarantee shifts from compile-time (macro) to runtime (registry validation at startup), which is equivalent in practice as long as both tests and prod use the same `build_reader` entrypoint — and we add a unit test that constructs the registry.
```
albinsuresh, line 49: To address this type safety guarantee shifting from compile-time to runtime, we could consider an approach where the `define_config!` macro generates additional functions for the keys that are cross referenced (e.g: `device.cert_path`) as follows:
```
    pub mod device {
        pub fn cert_path(root: &TEdgeConfig) -> &camino::Utf8PathBuf {
            &root.device.cert_path
        }
    }
```

And then replace the `from_root` references in the mapper configs from the raw strings like `device.id` to typed functions in the `tedge-config` crate as follows:

```
device: {
    #[tedge_config(default(from_root = tedge_config::keys::device::cert_path))]
    cert_path: camino::Utf8PathBuf,
}
```

That way, we still get type-safe validations for these references rather than at runtime. Yes, this comes with the cost of an additional dependency from the `mapper-config` to `tedge-config`, but that only sounds reasonable, as the runtime dependency exists anyway.

Though this was marked out-of-scope, I'm raising it now as this might even partly override parts of the proposal in `federated-read-with-root.md`.

prototypes/facet-config/crates/tedge-config/src/lib.rs
```diff
@@ -0,0 +1,84 @@
+use facet_config_runtime::*;
+
+facet_config_macro::define_config! {
+    TEdge {
+        device: {
+            /// Unique device identifier
+            #[tedge_config(example = "my-device-001", example = "AINA12345678")]
+            id: String,
```
albinsuresh, line 12: I'm assuming the "id from certificate" logic for this key would be handled by the `default(function = path::to::fn)` attribute.

albinsuresh (commented):
I can only appreciate the simplicity brought by this facet model with a clear split between the core tedge and mapper DTOs, unlike the merged monolith of the past, even at the cost of the runtime resolution that's required for cross config references. That really feels like the right mental model that mimics the mapper -> agent dependency as well.

I've added one suggestion on how to restore the compile-time type safety in cross-config references. But, it felt slightly going against the spirit of the current work that's keen on minimising per-key code-gen. So, I'll leave it to your judgement.

I noticed some wrinkles with misleading errors while using the `set` command, or the mapper directories not getting auto-created while setting the very first value for a mapper etc. But, expected from a prototype. I was also surprised to the see the `add` command working for a non-collection concrete types as well, where a secondary `add` functions like a replace. But then found out that the same behaviour exists in the current `tedge config` as well.

prototypes/facet-config/crates/tedge-config/src/lib.rs
jarhodes314: ~It'll need some slightly different logic to that as the function default doesn't currently provide any data to the function it calls, but yes, it will be something similar to that and is definitely possible.~ Now committed this in aa8bfcc

prototypes/facet-config/STATUS.md
jarhodes314: This might be useful. Equally, we might be able to resolve this sort of runtime issue with some simple automated tests (e.g. do a config get of every key with a blank config file and make sure we don't hit an error). These are definitely possible genuine runtime errors though if we don't properly test stuff (e.g. if you use from_root with a u16 value it just treats the key as not found at the moment).

jarhodes314: As part of aa8bfcc, I have added validation when we try and resolve the defaults for the configuration to check that (a) if a config has `from_root` configs, a root config has been supplied and (b) those root configs are resolvable. This should be fairly robust - it's still runtime rather than compile time, but should be easy to spot by simply creating config objects _somewhere_ in the unit test suite, rather than relying on exercising the exact code path.

Another small remark: I think the reason why the proposed compile-time checked API (aka a method taking the root `TEdgeConfig`) is non-ideal in my mind is that it's an encapsulation failure. The mapper should be able to read the mapper config without needing to worry about how that config is constructed, and resolving defaults via TEdgeConfig very much is a concern about how the config is constructed.

prototypes/facet-config/crates/config-macro-impl/src/input.rs
```diff
@@ -0,0 +1,375 @@
...
+pub struct Configuration {
+    pub name: syn::Ident,
+    pub groups: Punctuated<FieldOrGroup, Token![,]>,
+}
+
+#[derive(Debug)]
+pub enum FieldOrGroup {
+    Field(Box<ConfigField>),
+    Group(ConfigGroup),
+}
```
Bravo555, line 23: question: do you think it would be useful to extend the prototype to enable reuse of groups in the config? For example:

```rs
struct MqttListenerConfig {
    port: u16,
    bind_address: IpAddr,
    tls_config: TlsConfig
}


facet_config_macro::define_config! {
    TEdge {
        listener_a: MqttListenerConfig,
        listener_b: MqttListenerConfig,
        listener_c: MqttListenerConfig,
    }
}
```

As I understand, currently we really only can use custom models as leaf config options (e.g. `HostPort`), and the idea would be to be able to use custom models for groups as well.
This would possibly would increase the size of the implementation, but I think it's nice to provide some mechanism for some config paths to share a model to define leafs only in a single place and avoid repeating them under multiple groups, like we used to have for `c8y.`, `aws.`, `az.`, etc. where, indeed extracting them into separate files was crucial to allow more progress for custom mappers, etc. but until that happened it was a bit tedious where we had a lot of code to make them fit a common interface because they didn't share any part of the model.

jarhodes314: I'll have an explore of how possible this is. I think the answer is that it's non-trivial, but likely doable. I don't think it's a dealbreaker, but agree it would be a nice to have, particularly with the different cloud mapper configs having lots in common.

jarhodes314: Implemented in ee8d81b (currently for the mapper device section)

prototypes/facet-config/crates/mapper-config/src/shared.rs
```diff
@@ -0,0 +1,28 @@
...
+
+        /// Path to the device certificate for this mapper
+        #[tedge_config(default(from_root = "device.cert_path"))]
+        cert_path: camino::Utf8PathBuf,
+
+        /// Path to the device private key for this mapper
+        #[tedge_config(default(from_root = "device.key_path"))]
+        key_path: camino::Utf8PathBuf,
+    }
+}
```
Bravo555, line 28: Nice to see that groups be can be reused now.

jarhodes314:
The real implementation is now underway in #4265 so closing this

Issue #4235, opened 2026-07-01
API to access to the device.id field


reubenmiller:
**Is your feature request related to a problem? Please describe.**

The [Cumulocity opcua-device-gateway](https://github.com/thin-edge/opcua-device-gateway-container) needs to get access to the device's Cumulocity external ID, however it runs in a container and does not generally have access to the filesystem where thin-edge.io is running. Without access to the file system, it is difficult for it to get the device's external ID except via using the Cumulocity proxy which requires the device to be currently connected to Cumulocity.

**Describe the solution you'd like**

Extend the entity store api to return the external id (`@id`) field of the devices.

**Describe alternatives you've considered**

**Additional context**

reubenmiller:
Actually the use-cases also requires getting access to configured c8y.http value as well (in some modes of the opcua-device-gateway). Currently it gains access by mounting the `/etc/tedge` directory inside the container, and installing the `tedge` cli.

PR #4241, opened 2026-07-06
feat: expose tedge configs to external clients

albinsuresh:
## Proposed changes

Proposal to exposed selected tedge/mapper configs to external clients that don't have access to the config files via the same file system.

## Types of changes

## Paste Link to the issue

https://github.com/thin-edge/thin-edge.io/issues/4235

## Checklist


design/decisions/0010-config-exposure-to-external-processes.md
```diff
@@ -0,0 +1,134 @@
...
+Configuration holds secrets (`device.key_pin`, `cryptoki.pin`, credential-file paths),
+so the exposed set must be curated — by allowlist (opt-in) or denylist (opt-out).
+
+We choose **opt-in**, for *safe by default*.
+With a denylist, a newly added setting is exposed the moment it is introduced,
+and a new secret leaks unless someone remembers to deny it — a silent failure.
+With an allowlist, a new setting is invisible until a maintainer marks it exposable,
+so the worst case is a missing value (loud, easy to fix), not a leaked secret.
+Marking exposability where each setting is defined also keeps the decision
+reviewable alongside the value.
```
reubenmiller: I'm pretty sure we use a special type for secrets, e.g. "Redacted" (or something like that) which could be used to disable automatic exposure via this interface. That would make a more convenient/automatic mapping of new values rather than having to explicitly opting in to each new config key

design/decisions/0010-config-exposure-to-external-processes.md
```diff
@@ -0,0 +1,134 @@
...
+Each owner publishes each exposable value as a **retained** message,
+one value per topic under its own service topic — mirroring the per-fragment twin-data model.
+Per-value topics let a consumer subscribe to exactly the value(s) it needs
+and give clean per-value updates (empty retained payload = removal);
+retention keeps a value available while its publisher is down.
+
+The topic adds a `config` channel with the configuration key as its final segment:
+
+```
+te/device/<device>/service/<service>/config/<key>
```
reubenmiller: Do we know how many settings would be potentially published here? Or since we're using opt-in, then we only expose "useful" config keys, so the number would be under 10?

design/decisions/0010-config-exposure-to-external-processes.md
reubenmiller: The answer is also dependent on how many config keys we are exposing; whether we go for a "expose everything but secrets" or not.

albinsuresh: I was thinking that we'll only expose a very limited number of configs, that we feel would be relevant to the user. For example, I don't see any reason to expose configs like `device.key_path`, `device.cert_path`, or even entire config trees like `agent.*`, `apt.*` etc that an external client would have no business with.

design/decisions/0010-config-exposure-to-external-processes.md
albinsuresh: I've documented the list of configs that I feel worth exposing in this commit: 71e5a8741d0a6195f0e317ebd27f496f1c212006 and 
fc751b5

design/decisions/0010-config-exposure-to-external-processes.md
jarhodes314: > I'm pretty sure we use a special type for secrets, e.g. "Redacted" (or something like that) which could be used to disable automatic exposure via this interface. 

I'm fairly sure we don't have this at the moment as we store a reference to a secret (e.g. key path or password file) rather than the secret itself, so there isn't any sensitive information to redact. Though since those are just file paths, they aren't relevant for a public API as they're meaningless once they're not on the main device.

I guess with the hsm configurations we do have secrets that we store (key_pin, for example), but we don't actually have any mechanism in place to redact this info as far as I'm aware.

design/decisions/0010-config-exposure-to-external-processes.md
```diff
@@ -0,0 +1,295 @@
...
+| `service.type` | ✗ |
+| `software.plugin.default` | ✗ |
+| `software.plugin.exclude` | ✗ |
+| `software.plugin.include` | ✗ |
+| `software.plugin.max_packages` | ✗ |
+| `sudo.enable` | ✗ |
+| `tmp.path` | ✗ |
+
+#### Pull — HTTP (convenience view).**
+The agent aggregates these retained topics — subscribing like any other consumer,
```
rina23q: The "opt-in" concept is also applied to HTTP?
For example, If `GET /te/v1/entities/tedge-mapper-c8y/config/c8y.proxy.bind.address` (this key is marked "not exposed" in the list) is called, the HTTP returns the value?

albinsuresh: Only the data published on those MQTT `config/+` retained topics are exposed via the HTTP API. Just like our `twin/` data.

design/decisions/0010-config-exposure-to-external-processes.md
Bravo555: > I guess with the hsm configurations we do have secrets that we store (key_pin, for example), but we don't actually have any mechanism in place to redact this info as far as I'm aware.

AFAIK secrets like `device.key_pin` are indeed loaded into config as regular string types, but are converted into a `SecretString` wrapper on the boundary with the HSM module, so they're not accidentally logger after that point.

The optimal solution would be to not have any secrets in the config (indeed putting HSM PIN there was a mistake I regret making), but since the milk is already spilt, and we can _potentially_ have secret data in the config, it seems to me the most complete way to proceed would be to:

1. add a mechanism to load the PIN from an external source (e.g. file so it doesn't appear in config directly)
2. encourage users to use this new method by logging a warning when usage of the less secure secret-embedded-in-config setting is detected
3. provide a way to disable the warning for users who don't want to change anything (probably by adding another setting)
4. somehow cope with some users using the older setting which embeds secret in the config - either by introducing the notion of a secret field to `tedge-config` and having to handle it differently, or by not doing that and assuming that if the user silenced the warning, they must know what they're doing or don't care (fwiw I don't think adding this secret wrapper to the config would be that much work so we should still try it and only abandon the idea if we encounter unknown problems).

But indeed the proposal only covers step 4 and we don't have to do 1-3 for the proposed feature to work, but I just wanted to mention for completeness. I remember us talking about the secret management more generally, since there were also some other secret values that use another mechanism entirely, but if I remember correctly we haven't agreed on any general solution to that yet?

design/decisions/0010-config-exposure-to-external-processes.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Ownership: each component exposes only what it owns
+
+* the **agent** exposes core/device settings (`device.id`, the local MQTT/HTTP endpoints);
+* each **mapper** exposes its own cloud settings (`c8y.device.id`, `c8y.mqtt`, `az.url`).
+
+No component reads another's file to expose it on its behalf:
+the agent does **not** read the mappers' `mapper.toml`,
+which would couple it to every cloud's configuration shape.
+The source of truth for each setting stays with the component that owns it.
```
Bravo555: Since the agent/mapper might run with a different values for some fields than what's inside `.toml` files, because we allow overriding them using environment variables like `TEDGE_C8Y_HTTP`, I assume what we mean here is that agent/mapper don't read the file again, but just publish values for these fields that they've been ran with, so any env var overrides are included.

Bravo555 (commented):
Initially I've been bummed out that because of secrets like `key_pin` we can't just dump the entire config as a payload to a single topic and not have to think about it again, but I see that out of 156 config options we only expose 43 of them, so I'm glad that it doesn't seem to be inhibiting the feature much. Indeed not exposing options which are considered to be, if not outright secrets, then at least considered somewhat private, is maybe better than exposing everything.

design/decisions/0010-config-exposure-to-external-processes.md
albinsuresh: Yes, the component that consumes the config file is responsible for publishing them as well. So, they should factor-in the overrides as well.

design/decisions/0010-config-exposure-to-external-processes.md
reubenmiller: Ok, let's go with the current proposal, however after the implementation if the `/config/<key>` topics are too spammy, then we might just go with a single `/config` topic which has the entire list of keys and avoid the individual per key topics. Having the HTTP endpoints are fine, but the MQTT topics per key will most likely just pollute the retained messages

design/decisions/0010-config-exposure-to-external-processes.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+## Security considerations
+
+* Only marked values are ever published or served; secrets are never marked.
+* Unexposed and non-existent are indistinguishable (no topic / `404`),
+  so the boundary reveals nothing about which secret settings exist.
+* Retained messages are overwritable by any local client — a tampering nuisance,
+  not disclosure; owners republish the authoritative value on change/restart.
+  Stronger guarantees come from broker ACLs and the agent's existing TLS/client-cert options.
```
rina23q: After the second read, this is a real mess even though the values will be corrected after components change/restart.

Also, I realized that the design can open up a bit hacky usage. 

* a random component publishes a value to `/config/foo.bar`, and another component in the same network can get the value from HTTP, although this key does not exist in `tedge config`.

It would be nice to think about ACL during the implementation phase, even as a follow-up of the initial implementation.

rina23q (dismissed):
I think we can move to implementation, so I would approve.

design/decisions/0010-config-exposure-to-external-processes.md
albinsuresh: I feel the current proposal plays really well with ACLs to prevent any unauthorised writes as you highlighted. The services publish their configs to topics scoped within their entity topic ids: `te/device/main/service/tedge-agent/config/some.tedge.config` for the agent or `te/device/main/service/tedge-agent/config/some.mapper.config` for the c8y mapper. This "config topic as sub-topic of the entity topic id" design makes it easier to apply ACL rules to limit the write permissions to these topics only to those owning components. So, I hope that addresses your concern.

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -0,0 +1,432 @@
...
+**Alternative considered**
+
+The agent reads all `mapper.toml` files and exposes everything centrally.
+Rejected — it couples the agent to every cloud's config shape and profiles,
+breaking the rule that each setting's source of truth stays with its owner.
+
+### Allowlist marking is a per-field macro attribute, not a group-level or file-level flag
+
+`#[tedge_config(exposable)]` goes directly on each setting in `define_tedge_config!`,
+generating a `ReadableKey::is_exposable()` accessor — the same pattern already used for
```
jarhodes314: This has the potential to cause some slight conflicts with the config migration, but it should be fine as I'm anticipating the facet migration will take longer to complete than this will, and the old API will still be in use until the new one is complete.

This has also made me go and check how we use facet at the moment for these sorts of things, and I've realised we can lean even more heavily on facet than the prototype implementation did. **Edit:** now implemented

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -0,0 +1,432 @@
...
+
+### One shared publisher, one shared collection helper
+
+The (key, value) collection for "everything this component is allowed to expose" is a pure function in
+`tedge_config` (`exposed_core_config()` for the agent, `exposed_cloud_config(cloud, profile)` for a
+mapper), built on the existing `readable_keys()` / `read_string()` used by `tedge config list`.
+The publishing itself (topic construction, retain flag, the actor) is one small shared actor used by
+the agent and all three mappers, instead of three near-duplicate implementations.
+
+Caveat: some values are overridden at launch by a CLI flag or env var not written back to
```
jarhodes314: This should only be true for CLI flags, not for environment variables (which are taken into account when reading the config). I wonder whether it's worth adding in API to populate the readable tedge configuration with CLI overrides (to the facet-powered config refactoring).

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -0,0 +1,432 @@
...
+payload (which clears the retained message) if unset.
+This means a value removed from config can't linger from a previous run — twin-data already relies on
+the same behavior.
+
+One consequence: to any subscriber connecting later — including the agent's own aggregating
+subscription — an exposable-but-unset key and a never-exposed key look identical: no retained message
+at all.
+See "Unexposed and non-existent keys are indistinguishable" below.
+
+### The shared publisher subscribes to its own `config/+` and reconciles every message it sees
```
jarhodes314: It feels like this section is possibly slightly lacking detail on how it retrieves the information about what's currently set, i.e. "how do we work out when we've received all the retain messages we're subscribed to?" as knowing when you've received all the retain messages is somewhat nontrivial.

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -0,0 +1,432 @@
...
+
+Neither the retained topic nor the HTTP `404` gives any signal about whether a key is a real,
+non-exposed setting or doesn't exist at all — on purpose, so the boundary between "exists but secret"
+and "doesn't exist" leaks nothing.
+This isn't a special rule bolted onto the HTTP layer — the agent has no other source of truth for a key
+besides what arrives over MQTT.
+It collects every service's config purely by subscribing, including its own, so it can never know more
+than "did a retained message arrive for this key."
+
+### Deprecated aliases are never exposed; ownership decides who publishes
```
jarhodes314: I'm completely confused as to how deprecated keys are related to ownership?

openspec/changes/expose-tedge-configs/specs/config-exposure/spec.md
```diff
@@ -0,0 +1,153 @@
...
+#### Scenario: A demoted key's retained value is cleared the same way
+- **WHEN** a setting's `exposable` marking is removed while the setting itself still exists
+- **THEN** it SHALL be treated the same as a removed key: its old retained value SHALL be cleared
+
+#### Scenario: A cleared topic does not trigger further clears
+- **WHEN** the component receives an empty payload on a `config/<key>` topic whose expected state is
+  absent
+- **THEN** it SHALL take no action, so that a clear cannot echo into another clear
+
+### Requirement: The agent serves exposed configuration over HTTP
```
jarhodes314: Is this specifically retained messages only (i.e. only at startup), or does it include updates when services publish those (e.g. if the mapper starts up shortly after the agent)?

openspec/changes/expose-tedge-configs/design.md
albinsuresh: It's actually about a very specific case of the currently exposed `agent.entity_store.auto_register` which replaced the old deprecated `c8y.entity_store.auto_register` setting. The only thing that I wanted to highlight with this case was that the agent's setting being exposed does not transitively expose the mapper's value as well as both configs are now owned by two different components. I'll update that text as I realise that 2 different concepts (general handling of deprecated keys vs this specific case about cross component deprecation) got conflated into this single section.

openspec/changes/expose-tedge-configs/specs/config-exposure/spec.md
albinsuresh: The config values by other services can be published at any time, as they can be started any time. The agent will then start building that service's config map in its own entity store. So, once all the `config/+` retained messages for that external service is delivered to the agent, then it its config map is also complete and can be queried over HTTP. Unfortunately, there is no "readiness" cue here in terms of when all the configs of a given service is available, as that is completely dependent on the time taken for all those retained messages to be delivered and processed by the agent. So, the config endpoint becomes "available" from the moment the first `config/+` message is received by the agent.

openspec/changes/expose-tedge-configs/design.md
albinsuresh: This shared publisher is not concerned about the end of the retained messages delivery as it is equipped to do reconciliation dynamically at any point in time, as tampering attempts can also happen anytime.

design/decisions/0010-config-exposure-to-external-processes.md
albinsuresh: I'm starting to rethink this decision to publish individual config values to separate topics. While it provides the convenience of getting a simple config value even to pure MQTT clients, without having to parse the entire config as a JSON document, it came with some complexity as well as follows:

1. If an exposed value that was set before was unset and the service is restarted, the old value will still remain with the broker. To handle this case, we have introduced a reconciliation logic where explicitly send clear messages for all "unset but exposed" values as well on every startup. While this works, if there was a single config topic with the entire configuration, we could simply replace that with the new one.
2. Since there is an arbitrary duration for all retained config messages of a service to be delivered to the agent, the agent effectively exposes partial config of that entity from the moment the first retained messages is delivered until the last. If there was a single `config` message, then it's always all or nothing.

So, which one matters? The convenience of single config key subscribes or the simplicity of all configs at a single topic?

reubenmiller: Yeah I think publishing individual config keys to individual topic is not required...though maybe we only need to implement the http endpoint (which can do both all the config or individual keys), and just have a flow that calls the api and then publishes a single retained message with all of the keys in one message.

If users want to publish to multiple topics, then they can also do this by just modifying the flow, or creating their own.

albinsuresh: We get the MQTT API for free, as the mappers/components are using the MQTT interface to "share their configs" with the tegde-agent. So, the choice was between a single `te/<entity-topic-id>/config` topic vs multiple per-key: `te/<entity-topic-id>/config/<config.key>` topics. So, we go with the single topic solution.

openspec/changes/archive/2026-08-14-expose-tedge-configs/proposal.md
```diff
@@ -0,0 +1,86 @@
+## Why
```
jarhodes314, line 1: This is now doing a much better job of explaining the context around the problem and the core issue we're trying to solve.

openspec/changes/expose-tedge-configs/proposal.md
```diff
@@ -0,0 +1,86 @@
...
+
+%%te%% config lives in files (`tedge.toml`, `mapper.toml`) that only the owning component can read,
+and `tedge config get` needs the device certificate as well.
+So, that external process can not read any of the configs, not even the ones that are not secrets,
+without file system access to those files and device certificates.
+
+See [thin-edge/thin-edge.io#4235](https://github.com/thin-edge/thin-edge.io/issues/4235) for the full
+requirement and the motivating use cases.
+
+## What you can do
```
jarhodes314: I don't love this as a heading. It's not immediately clear (or frankly even after reading the section) what it's trying to convey. I think part of this is the "you" aspect: who is this really addressed to? the person proposing the change? reviewers? ai agents? a combination of those?

I think we could just call this `## Proposed Solution` or something

jarhodes314 (commented):
I think the proposed structuring for the proposal is nice and helps frame the problem correctly. I've suggested some improvements to the config in #4729 (as a PR seemed to be the easiest way to share it along with context by way of what the resulting artifacts look like after following those config suggestions). The broad scope of those changes is to focus more on the motivation for why we want our specs to be a certain way and less on just telling the agent to write fewer words.

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -1,12 +1,15 @@
...
-own them.
-An external process — a plugin, or a %%te%% component in its own container — has no access to the
-config directory or the device certificate, since these aren't mounted into other containers, and
-`tedge config get` needs both.
-So even a non-secret, widely-needed value like `device.id` is out of reach for anything outside the
-component that owns the config file.
+The problem and the user-facing surface are in the [proposal](./proposal.md):
+an external process can't read even a non-secret value like `device.id` without file access.
+
+This document covers how that surface is built.
```
didier-wenzek: I find this usage of the term *surface* unusual and weird.

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -1,12 +1,15 @@
...
 
-%%te%% config lives in files (`tedge.toml`, per-mapper `mapper.toml`) read only by the components that
-own them.
-An external process — a plugin, or a %%te%% component in its own container — has no access to the
-config directory or the device certificate, since these aren't mounted into other containers, and
-`tedge config get` needs both.
-So even a non-secret, widely-needed value like `device.id` is out of reach for anything outside the
-component that owns the config file.
+The problem and the user-facing surface are in the [proposal](./proposal.md):
+an external process can't read even a non-secret value like `device.id` without file access.
```
didier-wenzek: The wording used in `proposal.md` is clearer:

```suggestion
A plugin, or a %%te%% component running in its own container, or a child device,
cannot read any configuration setting from the `tedge.toml` configuration file stored on the main device disk.
```

openspec/config.yaml
```diff
@@ -27,3 +27,18 @@ context: |
...
 #     tasks:
 #       - Break tasks into chunks of max 2 hours
+
+rules:
+  proposal:
+    - After "Why", add a "What you can do" section that introduces the feature from the
+      user's point of view with short usage examples — before any internal mechanism.
+    - Keep "What Changes" for the mechanism and code impact. Never lead the solution with an
+      implementation detail (a macro flag, a struct, a topic-construction rule); those belong
+      under "What Changes", after the reader has seen how the feature is used.
```
didier-wenzek: I'm astonished that you delegate even that to an agenr :-(

docs/src/operate/entity-management/rest_api.md
```diff
@@ -488,6 +488,69 @@ Remove all of the child device's twin data using a single request.
 curl -f -X DELETE http://localhost:8000/te/v1/entities/device/child01///twin
 ```
 
+## Get exposed configuration
+
+Each entity that owns `tedge_config` settings (e.g. `tedge-agent`, or a cloud mapper such as
+`tedge-mapper-c8y`) publishes any setting explicitly marked exposable as a retained `config` MQTT
+message under its own service topic (see the [MQTT API reference](../../references/mqtt-api.md#configuration)).
+The agent collects these as an ordinary MQTT subscriber and serves them as a read-only view over HTTP.
```
didier-wenzek: The tone of this text is not appropriate for a user manual. This is more akin to a reference or a design doc.


```suggestion
The agent and the mappers expose part of their configuration over MQTT as  a retained `config` MQTT
messages.
The agent also gather all these configuration settings and serves them as a read-only view over HTTP.
```

docs/src/operate/entity-management/rest_api.md
```diff
@@ -488,6 +488,69 @@ Remove all of the child device's twin data using a single request.
...
+
+This view is read-only: `PUT`, `PATCH`, and `DELETE` on a `config` resource path are rejected.
+Configuration can only be changed by its owning component, via `tedge config set` and a restart.
+
+A configuration key that is not marked exposable and a key that does not exist at all are
+indistinguishable here: both return `404 Not Found`, so the response reveals nothing about which
+non-exposed settings exist.
```
didier-wenzek: Speaking about  *keys that are not marked exposable* might confuse the users asking themselves how to expose a key.


```suggestion
The agent and the mappers don't expose all their configuration settings (some are secret values, others are meaningless when used from another box). 
A configuration key that is not exposed or does not exist at all are
indistinguishable here: both return `404 Not Found`, so the response reveals nothing about which
non-exposed settings exist.
```

docs/src/references/mqtt-api.md
```diff
@@ -560,6 +562,35 @@ te/device/main///twin/device_OS
 }
 ```
 
+## Configuration
+
+A service that owns a `tedge_config` setting (`tedge-agent` for core/device settings, each cloud
+mapper for its own cloud's settings) publishes any setting explicitly marked exposable as a retained
```
didier-wenzek: *Explicitly marked exposable* is confusing for the users as they have no mean to mark a setting as exposable. This is a internal concept.


```suggestion
mapper for its own cloud's settings) publishes a non-secret subset of its setting as a retained
```

docs/src/references/mqtt-api.md
```diff
@@ -560,6 +562,35 @@ te/device/main///twin/device_OS
...
+```text title="Topic (retain=true)"
+te/device/main/service/tedge-agent/config/device.id
+```
+
+```text title="Payload"
+my-device-01
+```
+
+An empty payload means the value is unset, or was removed from configuration since the last time the
+owning service started. A setting is only published this way if it was explicitly marked exposable
```
didier-wenzek: Are setting updates dynamically published? When the file is updated? When `tedge config set` is used?

What about default values - not explicitly set in `tedge.toml`? What is the effect of `tedge config unset`?

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
```diff
@@ -0,0 +1,241 @@
...
+        let reader = config_reader("");
+        // mqtt.client.host has a default, so it should always be set
+        let core = exposed_core_config(&reader);
+        let (_, host) = core.iter().find(|(k, _)| k == "mqtt.client.host").unwrap();
+        assert_eq!(host.as_deref(), Some("127.0.0.1"));
+
+        // c8y.url has no default, so with no config it should be unset
+        let cloud = exposed_cloud_config(&reader, CloudType::C8y, None).unwrap();
+        let (_, url) = cloud.iter().find(|(k, _)| k == "url").unwrap();
+        assert_eq!(url, &None);
```
jarhodes314: This feels like it ought to be broken out into a separate test

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
```diff
@@ -0,0 +1,241 @@
...
+            "mqtt.client.auth.key_file",
+            "mqtt.client.auth.password_file",
+            "http.client.auth.cert_file",
+            "http.client.auth.key_file",
+        ];
+
+        for key in secret_keys {
+            let parsed: ReadableKey = key
+                .parse()
+                .unwrap_or_else(|e| panic!("failed to parse known configuration key '{key}': {e}"));
```
jarhodes314:
```suggestion
                .expect(format!("failed to parse known configuration key '{key}'"));
```

should do the same job

openspec/config.yaml
albinsuresh: The `proposal.md` generated by Openspec in this case was heavily solution oriented. So, I was just trying to tune it to our liking so that we don't have to rely on yet another document to understand the "user impact" of the proposed solution before getting into the details of the solution.

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
albinsuresh: Resolved in 17fb8b1

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
albinsuresh: Resolved in 17fb8b1

docs/src/references/mqtt-api.md
albinsuresh: There is no dynamic update. The published config represents the config the service was started with. Any subsequent changes take effect (and published) only on the next restart. Default values are published and unset values are omitted. Since we switched to publishing a single `config` JSON document message instead of a message-per-key, that published document only contains the set values. I have updated the user facing docs to explicitly state these as well in 911c34d.

docs/src/references/mqtt-api.md
albinsuresh: Re-written in 911c34de80d4cfca023c8968c573a52642376e19

docs/src/operate/entity-management/rest_api.md
albinsuresh: Re-written in 911c34de80d4cfca023c8968c573a52642376e19

docs/src/operate/entity-management/rest_api.md
albinsuresh: Rewritten in 911c34de80d4cfca023c8968c573a52642376e19

design/decisions/0010-config-exposure-to-external-processes.md
albinsuresh: Switched to a single config document per service in [419d007](https://github.com/thin-edge/thin-edge.io/pull/4241/commits/419d007baac6bf4d8e3253b5ba4aab90f7783dc7)

openspec/changes/expose-tedge-configs/design.md
albinsuresh: This isn't a concern anymore as we have switched to a single config document per service model in [419d007](https://github.com/thin-edge/thin-edge.io/pull/4241/commits/419d007baac6bf4d8e3253b5ba4aab90f7783dc7)

openspec/changes/expose-tedge-configs/proposal.md
albinsuresh: Merged your suggestions from #4279 (except for the changes to the `tasks` list) in [6272b71](https://github.com/thin-edge/thin-edge.io/pull/4241/commits/6272b71b5a886c7fe0b348362a67ef6d79aa22fa)

docs/src/operate/entity-management/rest_api.md
```diff
@@ -488,6 +488,75 @@ Remove all of the child device's twin data using a single request.
...
+GET /te/v1/entities/{topic-id}/config/{key}
+```
+
+**Response status codes**
+
+* 200: OK
+* 404: Not Found (the key is not exposed, or does not exist)
+
+:::note
+A key that is not exposed and a key that does not exist at all
```
rina23q, line 522: From a user perspective, there is no way to know which keys are exposed. My suggestion is to mark "exposed" in `tedge config list --doc`. It can be done later in follow-up.

docs/src/operate/entity-management/rest_api.md
```diff
@@ -488,6 +488,75 @@ Remove all of the child device's twin data using a single request.
...
+#### Example: Get all of a c8y mapper's exposed configuration
+
+```sh
+curl http://localhost:8000/te/v1/entities/device/main/service/tedge-mapper-c8y/config
+```
+
+```json title="Response"
+{
+    "url": "example.cumulocity.com",
+    "device.id": "my-device-01"
```
rina23q: Maybe worth noting that all values are represented as string. For example, boolean becomes `"true"`, array becomes `"[]"`, with double-quotes.

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
```diff
@@ -0,0 +1,66 @@
...
+
+Resource            ../resources/common.resource
+Library             Cumulocity
+Library             ThinEdgeIO
+Library             JSONLibrary
+
+Suite Setup         Custom Setup
+Test Teardown       Get Logs    ${DEVICE_SN}
+
+Test Tags           theme:c8y
```
rina23q: Probably adding configuration tag makes sense?

```suggestion
Test Tags           theme:c8y   theme:configuration
```

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
```diff
@@ -0,0 +1,66 @@
...
+    ${retained}=    Should Have Retained MQTT Messages    te/device/main/service/tedge-mapper-c8y/config
+
+    ${config}=    JSONLibrary.Convert String To Json    ${retained}[0]
+    Should Be Equal As Strings    ${config["url"]}    ${url}
+    Should Be Equal As Strings    ${config["bridge.topic_prefix"]}    ${topic_prefix}
+
+Mapper does not publish another cloud's settings
+    ${retained}=    Should Have Retained MQTT Messages    te/device/main/service/tedge-mapper-c8y/config
+    Should Not Contain    ${retained}[0]    az.url
+    Should Not Contain    ${retained}[0]    aws.url
```
rina23q: I think the test name doesn't match what this test actually checks.

* If it was intended to check that no key has a cloud prefix, then I would understand. In that case I would add a check that c8y.url is not included:

```
    Should Not Contain    ${retained}[0]    c8y.url
```

* If it was intended to check that the value of the `url` key in `${retained}` does not come from another cloud, then it should check the value, not the key name.

tests/RobotFramework/tests/tedge_agent/http_server/entity_config_api.robot
```diff
@@ -0,0 +1,78 @@
...
+
+
+*** Variables ***
+${DEVICE_SN}    ${EMPTY}    # Main device serial number
+
+
+*** Test Cases ***
+Agent publishes its exposed core settings as one retained MQTT JSON message
+    ${device_id}=    Execute Command    tedge config get device.id    strip=${True}
+    ${mqtt_port}=    Execute Command    tedge config get mqtt.client.port    strip=${True}
```
rina23q: Nit: They can move to `Custom Setup` and set them as Suite Variable, then we can avoid repeating the same commands in different tests.

openspec/changes/expose-tedge-configs/proposal.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Impact
+
+- `tedge_config_macros` — new `#[tedge_config(exposable)]` field attribute,
+  generates `ReadableKey::is_exposable()`
+- `tedge_config` — allowlist marking in `define_tedge_config!`,
+  `exposed_core_config()` / `exposed_cloud_config()` helpers
+- `tedge_api` — new `Channel::Config` topic variant,
+  config storage on the entity store
+- New shared publisher crate used by the agent and all three mappers
```
rina23q: "all three mappers" is inaccurate. I confirmed `tedge-mapper-local` and `tedge-mapper-collectd` also publish to `/config` now.

```
[te/device/main/service/tedge-mapper-local/config] {}
[te/device/main/service/tedge-mapper-collectd/config] {}
```

openspec/changes/archive/2026-08-14-expose-tedge-configs/specs/config-exposure/spec.md
```diff
@@ -0,0 +1,84 @@
...
+#### Scenario: Core and cloud settings on the right service topics
+- **WHEN** the agent starts with `device.id = my-device-01`
+  and the c8y mapper starts with `c8y.url = example.cumulocity.com`
+- **THEN** the agent SHALL publish retained on `.../tedge-agent/config`
+  a JSON object containing `"device.id":"my-device-01"`
+- **AND** the c8y mapper SHALL publish retained on
+  `.../tedge-mapper-c8y/config` a JSON object containing
+  `"url":"example.cumulocity.com"`
+
+#### Scenario: A profiled mapper instance publishes under its own service topic
```
rina23q, line 46: No system test coverage for profiled mapper.

openspec/changes/archive/2026-08-14-expose-tedge-configs/specs/config-exposure/spec.md
```diff
@@ -0,0 +1,84 @@
...
+`404 Not Found`, indistinguishably.
+Writes (PUT/PATCH/DELETE) SHALL be rejected.
+
+#### Scenario: Whole-service and single-key queries
+- **WHEN** a client sends `GET .../tedge-mapper-c8y/config`
+- **THEN** the response SHALL be a JSON object of every exposed key-value pair
+- **WHEN** a client sends `GET .../tedge-agent/config/device.id`
+- **THEN** the response SHALL be the value `my-device-01`
+
+### Requirement: Self-correcting publisher
```
rina23q, line 76: This requirement's coverage is entirely missing in system test.

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -0,0 +1,324 @@
...
+| `aws.url` | ✓ |
+| `az.bridge.keepalive_interval` | ✗ |
+| `az.bridge.topic_prefix` | ✓ |
+| `az.device.cert_path` | ✗ |
+| `az.device.csr_path` | ✗ |
+| `az.device.id` | ✓ |
+| `az.device.key_path` | ✗ |
+| `az.device.key_pin` | ✗ |
+| `az.device.key_uri` | ✗ |
+| `az.mapper.mqtt.max_payload_size` | ✗ |
```
rina23q: It seems marked "no" by mistake as the same key for `aws` and `c8y` are marked ✓. Also code.

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -0,0 +1,324 @@
...
+comparison also clears stale keys left behind by renames, demotions, or
+version upgrades — they're simply absent from the republished object, with
+no separate clearing path to implement or test.
+
+### Config stored parallel to twin data, not merged into it
+
+The entity store gets a `config` map on each entity, separate from
+`twin_data`. Config values come from one source of truth (the owning
+component's `tedge_config`) and nothing external can set them.
+Merging into twin data would blur that distinction.
```
rina23q: To be honest, I do not understand this section. Suddenly `twin` is mentioned. `twin` is something that is supposed to be reported to the cloud. `tedge config` key-values are not supposed to be exposed to the cloud. This is already enough reason why we introduced the new topic segment `config`.

While I was writing above, now I found the title is misleading. The section title should tell something "why we introduce the new topic endpoint, not reusing the existing `twin`"

openspec/changes/expose-tedge-configs/design.md
```diff
@@ -0,0 +1,324 @@
...
+That name comes from `bridge.topic_prefix`: `format!("tedge-mapper-{prefix}")`
+
+### One shared publisher
+
+At startup, a component publishes one retained message: a JSON object of
+every currently-set exposed key-value pair. Unset exposable keys are simply
+omitted from the object.
+Because the whole document is replaced as a unit, a value removed from config, 
+or a key no longer marked exposable, can't linger from a previous run:
+the next publish just doesn't include it.
```
rina23q: Also I could not understand this chapter... the title and the body doesn't match?

rina23q (commented):
Firstly reviewed other than Rust code. 
After reading `design.md` of OpenSpec, I still believe the initial design doc is worth reviving to this PR. The story of the decisions are missing there, while the initial doc had.

docs/src/operate/entity-management/rest_api.md
albinsuresh: Good catch. That's a side-effect of using the same `readable_keys()` plumbing that's used by the `tedge config list` command that treats every value as `String`. I'll see what I can do to preserve the types.

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
albinsuresh: It was the latter and yes, because of the `prefix` stripping, you're right that the values should have been checked.

openspec/changes/archive/2026-08-14-expose-tedge-configs/specs/config-exposure/spec.md
albinsuresh: Yeah, I also wanted to add one, but skipped it due to the lack of a second c8y tenant that I can use for the second profile. I'll check again if any other trick possible to work around that limitation.

openspec/changes/expose-tedge-configs/design.md
albinsuresh: Since configs are also just key-value pairs, just like twin data, it was just trying to explain that the same can't be re-used to store configs as well, as `twin` is published by the users while configs are published by the owning component itself. Now I feel it would have been better not to write anything about it, to avoid any unnecessary confusion.

openspec/changes/expose-tedge-configs/design.md
albinsuresh: It's not really a mismatch, but trying to explain a convoluted concept ("a value removed from config, 
or a key no longer marked exposable") straightaway. I'll re-write that section to explain that case better.

albinsuresh: Resolved by a2b0a72

openspec/changes/expose-tedge-configs/design.md
albinsuresh: Resolved by a2b0a72

openspec/changes/expose-tedge-configs/design.md
albinsuresh: Resolved by a2b0a72

openspec/changes/archive/2026-08-14-expose-tedge-configs/specs/config-exposure/spec.md
albinsuresh: Resolved by 4535ebb

openspec/changes/archive/2026-08-14-expose-tedge-configs/specs/config-exposure/spec.md
albinsuresh: Resolved by 4535ebb. There was a similar trick used in `tests/RobotFramework/tests/cumulocity/remote-access/test_remote_access_with_cloud_profiles.robot`

tests/RobotFramework/tests/tedge_agent/http_server/entity_config_api.robot
albinsuresh: Resolved by 4535ebb

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
albinsuresh: Resolved by 4535ebb

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
albinsuresh: Resolved by 4535ebb

openspec/changes/expose-tedge-configs/proposal.md
albinsuresh: Resolved by a2b0a72

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
```diff
@@ -0,0 +1,139 @@
...
+    Should Have MQTT Messages
+    ...    te/device/main/service/tedge-mapper-c8y/config
+    ...    minimum=1
+    ...    date_from=${start}
+    ...    message_contains="url":"${url}"
+
+    # The corrected document, not the injected one, is what ends up retained
+    ${retained}=    Should Have Retained MQTT Messages
+    ...    te/device/main/service/tedge-mapper-c8y/config
+    Should Not Contain    ${retained}[0]    bogus
```
rina23q: `bogus` isn't used anywhere.

```suggestion
    Should Not Contain    ${retained}[0]    bad
```

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
```diff
@@ -0,0 +1,139 @@
...
+    ${config}=    JSONLibrary.Convert String To Json    ${retained}[0]
+    Should Be Equal As Strings    ${config["url"]}    ${url}
```
rina23q, line 94: This check doesn't really make sense unless the bad actor publishes to overwrite `url`. Something like `{"url":"bad-url"}`, in addition to `{"bad":"config"}`.

albinsuresh: Here we are validating that the mapper re-publishes the original config back to the broker, overwriting the "bad config" message.

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
albinsuresh: Resolved by [47f10ed](https://github.com/thin-edge/thin-edge.io/pull/4241/commits/47f10edc1d926119739144c73faf588eb93a0dc4)

docs/src/operate/entity-management/rest_api.md
albinsuresh: Resolved by a series of commits from bdfaa75 to 6f4763c

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
rina23q: > Here we are validating that the mapper re-publishes the original config back to the broker, overwriting the "bad config" message.

This is line 94 - validating `"bad":"config"` is gone. 

Line 96 checks the value of `url`, which hasn't been changed in this test. If we want to check "overwriting" the existing key-value, the test should also publish a bad value to `url`. This is what I meant.

albinsuresh: The mappers and agents don't publish different config keys as separate `config/<key>` messages anymore. We had switched to a single `config` message (with all the config keys) per service model in  419d007 following [this discussion](https://github.com/thin-edge/thin-edge.io/pull/4241#discussion_r3527288937). 

In line no. 81, we're overwriting the entire config (that contains the `url` key as well) with a single "bad config". And then asserting in line. no 85 that the mapper would have replace this bad config with the original config where the `url` is present, in the background. That's why we're re-asserting that the final retained message still contains the original `url` and not the bad config.

If the current logic is confusing, I can read the original config before the bad config published and then re-assert that the eventual retained message (after the mapper reconciliation) is still the original config. Rather than comparing the entire config, I was just trying to validate it with just the `url` fragment in it.

tests/RobotFramework/tests/tedge/diag/predefined_plugins.robot
```diff
@@ -46,7 +46,7 @@ Test Tags           theme:troubleshooting    theme:cli    theme:plugins
     Log Should Contain    tedge-mqtt-sub.log    [te/device/main///cmd/restart] {}
     Log Should Contain
     ...    tedge-mqtt-sub-retained-only.log
-    ...    [te/device/main/service/tedge-agent] {"@parent":"device/main//","@type":"service","name":"tedge-agent","type":"service"}
+    ...    [te/device/main/service/tedge-agent] {"@parent":"device/main//","@type":"service","name":"tedge-agent"
```
albinsuresh, line 49: This change was required to handle the race between the entity registration and the `config` messages of the `tedge-agent` service. This is a classic side-effect of mixing up auto-registration with explicit registration.

For all tedge services, their entity registration and `config` messages are published by two independent actors. This can lead to the auto-registration of that service, if the config message gets delivered first. This will also result in the auto-registration message overwriting the explicit registration. Auto-registration does not include the `type` field by default, and hence this change.

The other options considered were the following:
1. Make the same `HealthMonitorActor` that generates and publishes the entity registration message to publish the `config` message as well, so that the ordered publishing (registration followed by config) can be ensured.
2. Pre-register the `tedge-agent` service entity in the entity store so that an auto-registration never happens for itself. While this would work for the `agent`, as it can pre-register itself, it won't work for mappers and other services.
3. Find a proper solution for mixed-up auto and explicit registrations in #4233 as this can affect any entity that parallelly sends registration messages and other messages.

Just went ahead with this cheap fix for now.

tests/RobotFramework/tests/configuration/mapper_config_exposure.robot
rina23q: > The mappers and agents don't publish different config keys as separate `config/<key>` messages anymore.

You are right. My bad. I misunderstood the design. I thought somehow publishing `{"bad":"config"}` applied only partially. You don't need to change here at all.

crates/core/tedge_agent/src/agent.rs
```diff
@@ -441,6 +465,7 @@ impl Agent {
         runtime.spawn(mqtt_actor_builder).await?;
         runtime.spawn(fs_watch_actor_builder).await?;
         runtime.spawn(twin_manager_builder).await?;
+        runtime.spawn(config_publisher_builder).await?;
```
rina23q: Related to https://github.com/thin-edge/thin-edge.io/pull/4241#discussion_r3735813642, probably moving spawning the config actor after `runtime.spawn(health_actor).await?;` might help. It is also a cheap solution though.

openspec/changes/archive/2026-08-14-expose-tedge-configs/.openspec.yaml
```diff
@@ -0,0 +1,2 @@
+schema: spec-driven
```
albinsuresh, line 1: This will be archived once the work is approved. Not archived yet, just in case something fundamental needs changing during the reviews, like the switch from the message-per-config-key model to aggregate-config-message-per-service model after the fist round of reviews.

rina23q (approved):
Approved.

reubenmiller (approved):
Everything worked nicely and a good overall UX

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
```diff
@@ -0,0 +1,302 @@
...
+        let reader = config_reader("");
+        let core = exposed_core_config(&reader);
+        assert!(!core.iter().any(|(k, _)| k.starts_with("c8y.entity_store")));
+        assert!(core
+            .iter()
+            .any(|(k, _)| k == "agent.entity_store.auto_register"));
+    }
+
+    #[test]
+    fn core_config_reports_none_for_unset_exposable_keys() {
```
jarhodes314: This test name doesn't really match up with what the test is doing - it suggests the value will be null, but we're actually testing that the default value appears.

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
```diff
@@ -0,0 +1,302 @@
...
+        let reader = config_reader("");
+        // mqtt.client.host has a default, so it should always be set
+        let core = exposed_core_config(&reader);
+        let (_, host) = core.iter().find(|(k, _)| k == "mqtt.client.host").unwrap();
+        assert_eq!(host, &Some(json!("127.0.0.1")));
+    }
+
+    /// Values keep the type they have in `tedge.toml` rather than all collapsing to strings, so a
+    /// consumer of the published document doesn't have to re-parse them.
```
jarhodes314: I think this comment is a bit pointless - the test is well-named and it's clear to me from reading the code what this is doing

crates/common/tedge_config_macros/impl/src/query.rs
```diff
@@ -807,6 +820,88 @@ fn generate_string_readers(paths: &[VecDeque<&FieldOrGroup>]) -> TokenStream {
...
+        .filter(|path| {
+            path.back()
+                .and_then(|fog| fog.field())
+                .is_some_and(|field| field.exposable())
+        })
+        .collect::<Vec<_>>();
+    let enum_variants = exposable_paths.iter().map(|path| enum_variant(path));
+    let arms =
+        exposable_paths
+            .iter()
```
jarhodes314, line 842: I don't care too much about fixing this as these macro changes will need to be reworked as part of the facet work regardless, but it feels like a smell the fact we keep iterating over the same things.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -222,6 +222,116 @@ async fn clear_entity_twin_data() {
     assert_eq!(entity.twin_data.get("x"), None);
 }
 
+#[tokio::test]
+async fn config_document_ingested_from_mqtt() {
```
jarhodes314: This test name doesn't describe the behaviour we're looking to observe

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -222,6 +222,116 @@ async fn clear_entity_twin_data() {
...
+
+    entity_store
+        .process_mqtt_message(
+            MqttMessage::from(("te/device/main///config", r#"{"device.id":"my-device"}"#))
+                .with_retain(),
+        )
+        .await;
+
+    // A payload that isn't a JSON object of strings must be ignored rather than panicking or
+    // clearing the previously ingested document.
```
jarhodes314: What about on the off chance the value is valid JSON but not an object, e.g. `["a", "b"]` or `"not an object"` or `true` or `2`? That's probably worth testing too if we're testing this.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -222,6 +222,116 @@ async fn clear_entity_twin_data() {
...
+        &mut entity_store,
+        "device/child0//",
+        EntityType::ChildDevice,
+        None,
+    )
+    .await
+    .unwrap();
+    mqtt_output.skip(1).await; // Skip the registration message
+
+    // Simulate a retained config document published by the entity's owning component
```
jarhodes314: This comment feels a bit redundant, it is basically repeating what the code below it/the test name are telling me.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -222,6 +222,116 @@ async fn clear_entity_twin_data() {
...
 
+#[tokio::test]
+async fn config_document_ingested_from_mqtt() {
+    let handle = entity::server("device-under-test");
+    let mut entity_store = handle.entity_store;
+
+    entity_store
+        .process_mqtt_message(
+            MqttMessage::from(("te/device/main///config", r#"{"device.id":"my-device"}"#))
+                .with_retain(),
```
jarhodes314: If we're simulating how the entity store reacts when we send it messages once it's in steady-state, it isn't very realistic to set the retain flag. In MQTT 3.1.1, the retain flag is only set for received messages received because they are retained (i.e. existing messages that are published to the client when it subscribes). If you are subscribed when a retain message is published, the retain flag is not set for the incoming message.

crates/core/tedge_agent/src/entity_manager/tests.rs
```diff
@@ -222,6 +222,116 @@ async fn clear_entity_twin_data() {
...
+    mqtt_output.skip(1).await;
+
+    entity::delete_entity(&mut entity_store, "device/child0//")
+        .await
+        .unwrap();
+
+    mqtt_output
+        .assert_received([
+            MqttMessage::from(("te/device/child0///config", "")).with_retain(),
+            MqttMessage::from(("te/device/child0//", "")).with_retain(),
```
jarhodes314: This test feels quite noisy. A lot of what's going on here: skipping mqtt messages and asserting the deregistration message, are not actually related to what the test is trying to demonstrate. Is there a different assertion mechanism we can use (e.g. `assert_received_unordered`) that will allow us to trim down the test so it focuses more specifically on what we're trying to test here?

crates/core/tedge_agent/src/http_server/entity_store.rs
```diff
@@ -132,6 +132,11 @@ enum Error {
 
     #[error("Actions on channel: {0} are not supported")]
     UnsupportedChannel(String),
+
+    // Deliberately as generic as EntityTwinDataNotFound: the response must not reveal whether the
+    // key is a real, non-exposed setting or does not exist at all.
```
jarhodes314: This comment is not helpful. Firstly, the entity store doesn't know whether a key doesn't exist or simply isn't exposed, that information isn't provided to it. Secondly, this isn't secret information -- anyone can read the thin-edge docs and establish that there is a setting called `device.key_pin`. The key itself isn't sensitive. I guess perhaps this was trying to get at the fact an attacker might be able to use the fact that an HSM is configured to do something nefarious?

jarhodes314 (commented):
Haven't finished reviewing, but have read through about half the files so far. Will resume tomorrow, but here are my thoughts for now.

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -477,6 +484,46 @@ impl EntityStore {
...
+            .and_then(|entity| entity.config.get(key))
+    }
+
+    /// Returns every exposed configuration value collected from an entity's own
+    /// retained `config` MQTT topic.
+    pub fn get_config(
+        &self,
+        topic_id: &EntityTopicId,
+    ) -> Result<&BTreeMap<String, JsonValue>, entity_store::Error> {
+        let entity = self.try_get(topic_id)?;
```
jarhodes314, line 507: I'm slightly intrigued as to the difference between `try_get` used here and `get` used in `get_config_value` above? I would expect the two functions to be doing the same thing. Is it just that this can return a meaningful result whereas the values themselves are inherently optional (as the key may not exist)o?

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -1747,6 +1802,133 @@ mod tests {
...
+            .unwrap();
+
+        let old = store
+            .set_config(
+                &topic_id,
+                BTreeMap::from([("device.id".to_string(), json!("other-device"))]),
+            )
+            .unwrap();
+
+        assert_eq!(old.get("mqtt.client.port"), Some(&json!(1883)));
```
jarhodes314: This shouldn't need to exist, we should be able to assume from the other passing tests that this holds

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -1747,6 +1802,133 @@ mod tests {
...
+        let mut store = new_entity_store(&temp_dir, true);
+        let topic_id = EntityTopicId::default_main_device();
+
+        store
+            .set_config(
+                &topic_id,
+                BTreeMap::from([
+                    ("device.id".to_string(), json!("my-device")),
+                    ("mqtt.client.port".to_string(), json!(1883)),
+                ]),
```
jarhodes314: The target type `BTreeMap` is determined by the `set_config` signature, so we can reduce noise here (and in the rest of these tests) by using `.into()` instead.

```suggestion
                [
                    ("device.id".to_string(), json!("my-device")),
                    ("mqtt.client.port".to_string(), json!(1883)),
                ].into(),
```

crates/core/tedge_api/src/entity_store.rs
```diff
@@ -477,6 +484,46 @@ impl EntityStore {
...
+        if let Some(key) = config
+            .keys()
+            .find(|key| key.is_empty() || key.starts_with('@') || key.contains('/'))
+        {
+            return Err(Error::InvalidConfigKey(key.clone()));
+        }
+
+        let entity = self.try_get_mut(topic_id)?;
+        let old = mem::replace(&mut entity.config, config);
+        Ok(old)
```
jarhodes314: Is there a particular motivation for returning the old value from this method? It feels like a slightly weird API to return the stale value. I appreciate in the standard library, methods do this somewhat regularly (`mem::replace`, `HashMap::insert`, etc.), but unless we _need_ the old value, it feels like we don't gain anything from this and it's a bit confusing.

crates/extensions/tedge_config_ext/src/actor.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// comparison also clears a stale key left behind by a rename, removal, or demotion — it is
+    /// simply absent from the republished object.
+    async fn reconcile(&mut self, message: MqttMessage) {
+        let Ok((_, Channel::Config)) = self.mqtt_schema.entity_channel_of(&message.topic) else {
+            return;
+        };
+
+        if !self.payload_matches_expected(message.payload_bytes()) {
+            self.publish_current_config().await;
+        }
```
jarhodes314: The `let`...`else` here is really an implicit `if`, and I think this would read better if it were made explicit:

```suggestion
        let channel = self.mqtt_schema.entity_channel_of(&message.topic);
        if matches!(Ok((_, Channel::Config)), channel)
            && !self.payload_matches_expected(message.payload_bytes())
        {
            self.publish_current_config().await;
        }
```

crates/extensions/tedge_config_ext/src/actor.rs
```diff
@@ -0,0 +1,103 @@
...
+use tedge_mqtt_ext::QoS;
+use tracing::error;
+
+pub struct ConfigPublisherActor {
+    mqtt_schema: MqttSchema,
+    service_topic_id: ServiceTopicId,
+    /// Every currently-set exposed key-value pair, published as one retained JSON object. Each
+    /// value keeps the type it has in `tedge.toml`, so a port stays a number and a flag stays a
+    /// boolean. A key that is exposable but unset is simply absent here.
+    expected: BTreeMap<String, Value>,
```
jarhodes314: I'm struggling with the name `expected`. It implies that the thing will become true in the future, but it seems like it actually means a previously true value that will become stale in the future.

Maybe something like: `last_published` (though that does sound a little bit like I'd expect it to be the time the value was last published, not the value itself)?

crates/extensions/tedge_config_ext/src/actor.rs
```diff
@@ -0,0 +1,103 @@
...
+use tedge_mqtt_ext::MqttMessage;
+use tedge_mqtt_ext::QoS;
+use tracing::error;
+
+pub struct ConfigPublisherActor {
+    mqtt_schema: MqttSchema,
+    service_topic_id: ServiceTopicId,
+    /// Every currently-set exposed key-value pair, published as one retained JSON object. Each
+    /// value keeps the type it has in `tedge.toml`, so a port stays a number and a flag stays a
+    /// boolean. A key that is exposable but unset is simply absent here.
```
jarhodes314: I think this is pretty much all possible to infer from the code, especially if `expected` is named more clearly.

crates/extensions/tedge_config_ext/src/tests.rs
```diff
@@ -0,0 +1,147 @@
...
+async fn publishes_the_set_keys_as_one_json_object_on_startup() {
+    let mut mqtt_box = setup(vec![("device.id", Some(json!("my-device")))]);
+
+    mqtt_box
+        .assert_received([config_message(r#"{"device.id":"my-device"}"#)])
+        .await;
+}
+
+/// Values are published with the type they have in `tedge.toml`, so a consumer doesn't have to
+/// re-parse a number or a flag out of a string.
```
jarhodes314: This feels a bit redundant

crates/extensions/tedge_config_ext/src/tests.rs
```diff
@@ -0,0 +1,147 @@
...
+    mqtt_box
+        .assert_received([config_message(r#"{"device.id":"my-device"}"#)])
+        .await;
+}
+
+#[tokio::test]
+async fn republishes_a_diverged_document() {
+    let mut mqtt_box = setup(vec![("device.id", Some(json!("my-device")))]);
+
+    // Skip the startup publish
```
jarhodes314: If we're going to comment this, we should do it in the same way in every test below this, too.

docs/src/references/mqtt-api.md
```diff
@@ -560,6 +562,82 @@ te/device/main///twin/device_OS
...
+* `te/device/main/service/tedge-agent/config`: the core device settings from `tedge.toml`
+* `te/device/main/service/tedge-mapper-<cloud>/config`: cloud connection settings from `mapper.toml`
+
+The whole set is published as a single JSON document, as a retained message.
+Each field name is the corresponding `tedge config` key:
+the agent publishes `device.id` as `device.id`.
+
+Each value keeps the type the setting has in `tedge.toml`.
+A port is published as a number, a flag as a boolean and a template set as an array,
+so a consumer does not have to parse them back out of strings.
```
jarhodes314: I'm slightly unsure whether this is helpful or not. To me, this represents exactly how I would expect the API to behave - why would a port be represented as a string? Obviously if someone writes a script to consume the output of `tedge config get`, then they will need to parse the information from a string, but that's because command output is always a string, not anything we're in control of.

I think the examples also help show that this is what's happening wrt types, and in a less abstract manner.

jarhodes314 (commented):
A few more thoughts from the rest of the review.

Overall, the feature is in really good shape, just a few relatively minor things about the details of the code.

crates/core/tedge_agent/src/http_server/entity_store.rs
albinsuresh: That's true. The entity store only knows what the services tells it about. So, there is no need to highlight the difference between non-existent vs unexposed, at least in this context.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Yeah, I think the impl just kept it consistent with `set_twin_fragments` but unlike that API, the retuned value of this one isn't really used anywhere. So, we can keep it simple.

crates/extensions/tedge_config_ext/src/actor.rs
albinsuresh: Yeah, that name only makes sense in the context of config message reconciliation (`received == self.expected`), where the incoming config message is compared against the current/expected config. I guess I'll just call it `config` to keep it simple. Checking `received == self.config` isn't that bad either.

crates/extensions/tedge_config_ext/src/actor.rs
albinsuresh: Yeah, these are side-effects of changing an openspec "spec" after it has already done an implementation against an older spec (where all config values where published as strings). Then it starts highlighting the changes. I'll remove that.

docs/src/references/mqtt-api.md
albinsuresh: Same reason as https://github.com/thin-edge/thin-edge.io/pull/4241#discussion_r3781659840. I'll clean that up.

crates/core/tedge_api/src/entity_store.rs
albinsuresh: No, they should have been consistent.

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
albinsuresh: Resolved by df661ef5deeca51c64e014f0d18a3e956449d0e5

crates/common/tedge_config/src/tedge_toml/tedge_config/config_exposure.rs
albinsuresh: Resolved by df661ef5deeca51c64e014f0d18a3e956449d0e5

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: Resolved by a5fd79009336507065067c9c853b90fbea5c4d32

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: Resolved by a5fd79009336507065067c9c853b90fbea5c4d32

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: Resolved by a5fd79009336507065067c9c853b90fbea5c4d32

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: Resolved by a5fd79009336507065067c9c853b90fbea5c4d32

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: Resolved by a5fd79009336507065067c9c853b90fbea5c4d32

crates/core/tedge_agent/src/http_server/entity_store.rs
albinsuresh: Resolved by 82c37bc3bf02db5a93b526a5ee7a6aae5e436207

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Resolved by 4f40ba0f6b5c07cec437c92392c5b06ecb972307

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Resolved by 4f40ba0f6b5c07cec437c92392c5b06ecb972307

crates/extensions/tedge_config_ext/src/actor.rs
albinsuresh: Resolved by 201e275d8c5f37a3305de040fbeb405d269b6722

crates/extensions/tedge_config_ext/src/actor.rs
albinsuresh: Resolved by 201e275d8c5f37a3305de040fbeb405d269b6722

crates/extensions/tedge_config_ext/src/actor.rs
albinsuresh: Resolved by 201e275d8c5f37a3305de040fbeb405d269b6722

crates/extensions/tedge_config_ext/src/tests.rs
albinsuresh: Resolved by 201e275d8c5f37a3305de040fbeb405d269b6722

crates/extensions/tedge_config_ext/src/tests.rs
albinsuresh: Resolved by a5fd79009336507065067c9c853b90fbea5c4d32

docs/src/references/mqtt-api.md
albinsuresh: Resolved by aae9032d86d5a81ff82b07e303cbf94f644039ba

crates/core/tedge_api/src/entity_store.rs
albinsuresh: Resolved by df5e412a8b71c166c694ee93a0341b3842c1df29

docs/src/operate/entity-management/rest_api.md
albinsuresh: In an offline discussion, we decided to skip this for now, as the decision on whether to expose some configs or all configs (minus the secrets) will be revisited before the release of this feature.

crates/core/tedge_agent/src/entity_manager/tests.rs
jarhodes314: I think asserting the second message is still unnecessary

jarhodes314 (approved):

crates/core/tedge_agent/src/entity_manager/tests.rs
albinsuresh: Resolved and squashed into 4e535640ee87ec148b20ddc098a1dd3b483edbbb

